Issue #12347, opened 2020-11-06
Magic is broken when outlook safe links are enabled


xoxoxo:
### Issue Summary
This is basically as in https://github.com/FusionAuth/fusionauth-issues/issues/629 If you have safe links enabled whenever users clicks on it token has already been used. Seems there is some scanning on office 365/outlook side. Some folks are using our membership with corporate emails so seems like its not possible for them to disable it.

### To Reproduce

1. Login as member
2. Open link from inside outlook with safe links enabled

PR #12519, opened 2021-01-08
Added multiple use grace period to tokens

allouis:
no-issue

This change allows a token to be used multiple times for the first 10
seconds after its initial use, this should stop dynamic link checking
software from invaliding magic links.


matthanley:
Tested to work with outlook.com 🎉

matthanley:
refs #12347

Issue #10318, opened 2019-01-02
Settings table cleanup


ErisDS:
Our settings table needs a little bit of cleanup. The intended conventions got lost and broken, and chaos has ensued! 

To fix this, we've agreed on the following conventions:

**Key**

- must be snake_case
- should be descriptive and unique
- for keys that appear in APIs, they should be named as they will be in the API

**Value**

- can be one of: String, Number, Boolean or Array (a list of similar items)
- must not be a JSON Object
- must be nullable or have a sensible default

**Type**

- Explicitly used to declare what type the `value` contains
- String, Number, Boolean or Array (a list of similar items)
- Array could be extended to specify the type of item they will contain - e.g. `Array<String/Number/Boolean/StripePlan>`

**Group**

- Groups should be used to represent settings that are fetched and updated from the API together.
- The groups have special implied access rules:
    - `site` - available via the Admin and Content Settings APIs
    - `core` - not available via any API
    - anything else - available via the Admin Settings API
- New groups can be added as needed, if we need a new group with a new implied access rule, we can also do this.

**Flags**

- NULL - default
- PUBLIC - will appear in the unauthenticated Admin and Members Site APIs
- RO - cannot be written directly via the Admin API - is read only
- PUBLIC,RO - is both public and read only

----

## TODO

To bring our existing settings into line with these conventions we need to make the following changes

General cleanup:
- [x] Remove `force_i18n`, `permalinks` and `members_session_secret`
- [x] Rename `default_locale` to `lang` and `active_timezone` to `timezone`
- [x] Rename `ghost_head` and `ghost_foot` to `codeinjection_head` and `codeinjection_foot`
- [x] Rename `brand.publicationColor` to `accent_color`

New `group` and `special` concept:
- [x] Add a new column called `group`, give each setting a new correct group name (inc changing blog -> site)
- [x] Add group support to the API and update Ghost Admin to use group, not type
- [x] Add a special column to indicate extra API rules
- [x] Implement special rules at API level and in Ghost admin

Types and Validations:
- [x] Repurpose the old type column to contain the value type (String, Boolean, Array)
- [x] Ensure that default-settings.json has useful validations for as many settings as possible

Bigger reworks:
- [x] Rework members_subscription_settings to a set of key-values in a shared group
- [x] Rework stripe_connect_integration to a set of key-values in a shared group
- [x] Rework bulk_email_settings to a set of key-values in a shared group
[...]

PR #12540, opened 2021-01-14
Added FirstPromoter Integration

rshbhgrg:
FirstPromoter allows a site to build a referral program and ask other members to promote your work to other people. Currently, to integrate FirstPromoter on a site, it needs steps from the guide [here](https://ghost.org/integrations/firstpromoter/). 
The new native integration with FirstPromoter makes it easy for as site to enable FirstPromoter referral program on their site, by only needing to add their FirstPromoter Tracking ID and enabling it

- [x] Adds new FirstPromoter settings
- [x] Update tests

### DIFF
--- a/core/server/api/canary/utils/serializers/output/utils/settings-type-group-mapper.js
+++ b/core/server/api/canary/utils/serializers/output/utils/settings-type-group-mapper.js
@@ -11,7 +11,8 @@ const groupTypeMapping = {
     private: 'private',
     portal: 'portal',
     email: 'bulk_email',
-    newsletter: 'newsletter'
+    newsletter: 'newsletter',
+    firstpromoter: 'firstpromoter'
 };
 
 const mapGroupToType = (group) => {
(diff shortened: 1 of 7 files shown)


daniellockyer (approved):

kevinansfield (approved):

tpatel (approved):
Looks great, I have two questions:

1. I've read in the engineering handbook that it's [prohibited to add new objects to the settings table](https://www.notion.so/ghost/API-Core-c7f1926386304eedb3d8fd4304bfecf8#ca32316c885c46198b1825e5ede31d73). I'm not sure to understand what is prohibited? I guess adding keys is fine.
2. Is that a good thing to update the v2 API code? If it's something new I guess it's generally good to not update "frozen" apis? although in this case it's just changing the shape of the object that is returned if I understand correctly.

rishabhgrg:
@tpatel 

> it's prohibited to add new objects to the settings table.

That basically refers to the `type` of Setting value, which shouldn't be a JSON object anymore as we moved away from JSON stringified objects assigned to setting keys. [This](https://github.com/TryGhost/Ghost/issues/10318) issue is a good reference when we undertook the Setting table cleanup and covers this :)

> If it's something new I guess it's generally good to not update "frozen" apis

Yeah so settings getting new value has been the exception so far, afair primarily because the whole default-settings setup where we end up adding all new settings on Ghost startup and then the v2 API is setup to just pick and send all settings by default instead of having a proper include-list. Ghost Admin always explicitly asks for settings by group so new settings are not added in its response.

PR #12560, opened 2021-01-22
Remove not null constraint from `plan_nickname` column

allouis:
This value is nullable from the Stripe API so we're making sure that we
can store exact values locally.

SQLite3 does not supports altering columns so instead we have to:
1. Create a temp table and copy the data to it
2. Delete the original table
3. Recreate the original table with the necessary modifications
4. Copy the data from the temp table
5. Drop the temp table

### DIFF
--- a/core/server/data/migrations/versions/4.0/01-remove-not-null-constraint-from-plan-name-column.js
+++ b/core/server/data/migrations/versions/4.0/01-remove-not-null-constraint-from-plan-name-column.js
@@ -1,0 +1,64 @@
+const {createNonTransactionalMigration} = require('../../utils');
+const {createTable, deleteTable} = require('../../../schema/commands');
+const logging = require('../../../../../shared/logging');
+
+const tableDef = {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    customer_id: {type: 'string', maxlength: 255, nullable: false, unique: false, references: 'members_stripe_customers.customer_id', cascadeDelete: true},
+    subscription_id: {type: 'string', maxlength: 255, nullable: false, unique: true},
+    plan_id: {type: 'string', maxlength: 255, nullable: false, unique: false},
+    status: {type: 'string', maxlength: 50, nullable: false},
+    cancel_at_period_end: {type: 'bool', nullable: false, defaultTo: false},
+    cancellation_reason: {type: 'string', maxlength: 500, nullable: true},
+    current_period_end: {type: 'dateTime', nullable: false},
+    start_date: {type: 'dateTime', nullable: false},
+    default_payment_card_last4: {type: 'string', maxlength: 4, nullable: true},
+    created_at: {type: 'dateTime', nullable: false},
+    created_by: {type: 'string', maxlength: 24, nullable: false},
+    updated_at: {type: 'dateTime', nullable: true},
+    updated_by: {type: 'string', maxlength: 24, nullable: true},
+    plan_nickname: {type: 'string', maxlength: 50, nullable: true},
+    plan_interval: {type: 'string', maxlength: 50, nullable: false},
+    plan_amount: {type: 'integer', nullable: false},
+    plan_currency: {type: 'string', maxLength: 3, nullable: false}
+};
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        if (knex.client.config.client === 'mysql') {
+            logging.info('Removing NOT_NULL constraint from members_stripe_customers_subscriptions:plan_nickname');
+            await knex.schema.alterTable('members_stripe_customers_subscriptions', (table) => {
+                table.string('plan_nickname').nullable().alter();
+            });
+        } else {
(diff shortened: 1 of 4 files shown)


core/server/data/migrations/versions/4.0/01-remove-not-null-constraint-from-plan-name-column.js
```diff
@@ -0,0 +1,64 @@
+const {createNonTransactionalMigration} = require('../../utils');
+const {createTable, deleteTable} = require('../../../schema/commands');
```
allouis, line 2: Not sure if I should be using the commands directly? Like should I create a helper in the utils for this? :thinking:

ErisDS:
Hang on... sqlite3 doesn’t support changing constraints, but I thought knex did?

Under the hood the table has to be recreated but that behaviour is baked into knex. We still don’t do it on a minor, cos it’s so destructive, but we shouldn’t be writing the code to do this ourselves.

kevinansfield:
> sqlite3 doesn’t support changing constraints, but I thought knex did?

We had [some discussion](https://ghost.slack.com/archives/C02G9E68C/p1606311044482500) recently about this, Knex does do the table dance automatically but it doesn't [fully re-apply the schema](https://github.com/knex/knex/issues/631). Indexes end up getting lost which means we also lose unique constraints, it's not something we've really noticed before because we tend to have in-app checks for uniqueness but we've started using indexes more heavily recently to keep email sending and analytics working past a certain scale.

@allouis was this an attempt to work around that issue or was the manual table dance for something else?

> We still don’t do it on a minor, cos it’s so destructive

Knex does the same table dance when dropping columns on sqlite, we've done that in a few minors recently 😬

allouis:
:thinking: I was going from the docs for the `alter` method here: http://knexjs.org/#Schema-alter

I did just try to run this migration using that syntax for SQLite and received an error like:

```
[2021-01-25 10:37:15] ERROR Sqlite does not support alter column.
```

Perhaps I've missed something and there's a different approach within knex that will work?

kevinansfield:
> Perhaps I've missed something and there's a different approach within knex that will work?

No, re-reading the docs it looks I was mistaken on knex's abilities here

core/server/data/migrations/versions/4.0/01-remove-not-null-constraint-from-plan-name-column.js
rishabhgrg: @allouis Just wondering if we should include the whole table deletion/temp table/re-creation dance in the util instead of just these 2 methods. The whole flow for re-creation with correct tableDef is pretty generic and having it out in a util will mean easier to re-use as well as update it if we/knex comes across with a better way in future.

allouis: Yeah that's what I was thinking, but it might be best to do that if we run into this a second time, not sure.

@daniellockyer what do you think?

daniellockyer: Yeah I'd say leave it for now and implement it if we need it again in the future. Are we likely to see this again soon?

core/server/data/schema/commands.js
```diff
@@ -79,7 +79,7 @@ function dropUnique(tableName, column, transaction) {
  * https://github.com/tgriesser/knex/issues/1303
  * createTableIfNotExists can throw error if indexes are already in place
  */
-function createTable(table, transaction) {
+function createTable(table, transaction, tableSpec = schema[table]) {
```
kevinansfield, line 82: TIL default arguments are evaluated at call time 😄

core/server/data/migrations/versions/4.0/01-remove-not-null-constraint-from-plan-name-column.js
kevinansfield: A util for whole table dance would be good, especially as we'll want one for dropping columns too, but I don't think it's a blocker to getting this PR merged.

kevinansfield (approved):

core/server/data/schema/commands.js
allouis: It's awesome!

daniellockyer: Yes I had to go check that too 😁

daniellockyer (approved):
LGTM - I don't see an alternative to the temp table dance whilst keeping indices so this'll have to do 🙂

core/server/data/schema/commands.js
allouis: You can do cool stuff with it like:
```js
function required(name) {
  throw new Error(`Missing parameter ${name}`);
}

function myFn(name = required('name'), age = required('age')) {
  // Can guarantee a non-undefined value for name and age
}
```

daniellockyer: 😮

rshbhgrg (approved):

PR #12563, opened 2021-01-23
Add mexican peso currency

gusreyes01:
Added Mexican Peso currency and abbreviations to enable currency in Stripe.


ErisDS:
Thank you for your contribution! We are planning some work that will improve our currency support in a more complete way and so I'm closing this in lieu of the proper solution.

Issue #12160, opened 2020-08-25
Members with cancelled comp plans aren't listed with the free members filter


matthanley:
### Issue Summary

Members who have had complimentary plans removed don't appear in the *free members* filter on the members list.

### To Reproduce

1. Create a new member in a members-enabled site with Stripe connected
1. Observe member appears in list of free members (at `/ghost/#/members?paid=false`)
1. Enable "Complimentary premium plan" for the member
1. Observe member appears in list of paid members (at `/ghost/#/members?paid=true`)
1. Disable "Complimentary premium plan" for the member
1. Observe member no longer appears in *either* free or paid filtered member list, but does appear in *all members* list

**What did you expect to happen instead?**
Member should be reverted to a free member, and appear in the free members filtered list.

We should also verify the behaviour for cancelled and expired paid members (they should also appear as free members).

### Technical details:

* Ghost Version: 3.31.1 (on Pro)
* Browser/OS: Chrome

kevinansfield:
The issue is that our "free member" query works on the assumption that members do not have a Stripe customer record if they are unpaid.

We can adjust the generated SQL query to something like this:

```sql
select
    `members`.*
from
    `members`
    left join `members_stripe_customers` on `members`.`id` = `members_stripe_customers`.`member_id`
    left join `members_stripe_customers_subscriptions` on `members_stripe_customers`.`customer_id` = `members_stripe_customers_subscriptions`.`customer_id`
where
    `members_stripe_customers`.`member_id` is null
    or `members_stripe_customers_subscriptions`.`customer_id` is null
    or `members_stripe_customers_subscriptions`.`status` not in ("active", "trialed");
```

However that query involves fetching all members/customers/subscriptions no matter what limit is applied meaning performance isn't great. For ~100k members with subscriptions and 3 test members with no customer, no subscription, or subscription that isn't active or trialed the query takes ~1sec locally even for paged result sets.

I'm not sure if there's a way to improve the query or if we'd need to look at a different way of fetching free members. We could possibly run multiple queries checking:
1. no customer
2. no subscription
3. subscription with inactive status
We'd need to re-architect how the `?paid=false` query param is handled in that case.

kevinansfield:
Alternative quick-fix would be to ensure that stripe customer data is cleared from the database when a member has a complimentary plan removed or otherwise has their subscription cancelled. That would still leave us with a potentially brittle handling of `?paid=false`

naz:
[...]

PR #12570, opened 2021-01-25
Add status flag to members

allouis:
refs https://github.com/TryGhost/Ghost/issues/12160

This lays the groundwork for being able to filter members by free/paid - without having to do complicated joins. It also allows us to, correctly, filter then via standard NQL filters rather than with our custom `?paid` query param at the moment.

### DIFF
--- a/core/server/data/migrations/versions/4.0/03-populate-status-column-for-members.js
+++ b/core/server/data/migrations/versions/4.0/03-populate-status-column-for-members.js
@@ -1,0 +1,45 @@
+const {chunk} = require('lodash');
+const {createTransactionalMigration} = require('../../utils');
+const logging = require('../../../../../shared/logging');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Updating members.status based on members_stripe_customers_subscriptions.status');
+        const paidMemberIds = (await knex('members')
+            .select('members.id')
+            .innerJoin(
+                'members_stripe_customers',
+                'members.id',
+                'members_stripe_customers.member_id'
+            ).innerJoin(
+                'members_stripe_customers_subscriptions',
+                function () {
+                    this.on(
+                        'members_stripe_customers.customer_id',
+                        'members_stripe_customers_subscriptions.customer_id'
+                    ).onIn(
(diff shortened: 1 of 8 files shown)


core/server/data/migrations/versions/4.0/03-populate-status-column-for-members.js
```diff
@@ -0,0 +1,45 @@
...
+        const chunkSize = 998;
+        const paidMemberIdChunks = chunk(paidMemberIds, chunkSize);
+
+        for (const paidMemberIdsChunk of paidMemberIdChunks) {
+            await knex('members')
+                .update('status', 'paid')
+                .whereIn('id', paidMemberIdsChunk);
+        }
+    },
+    async function down(knex) {
```
naz, line 39: This down migration makes sense to me. Wanted to confirm there's nothing more that I could be missing, the main reasoning is: we roll back to a `default` value introduced in `02-add-status-column-to-members.js`, otherwise we might end up in partially paid/free members set.

On the other hand, is there even a need to have the down migration as `02-add-status-column-to-members.js` would remove the column completely if rolled back?

naz (approved):
Tested on MySQL LTGM :+1:

core/server/data/migrations/versions/4.0/03-populate-status-column-for-members.js
rishabhgrg: +1, just wondering the same and if it's better to leave out the down migration than set back the default value once we go forward as setting back incorrect status is probably worse ? 🤷 Again not sure which is better but just thinking out loud.

rshbhgrg (approved):

core/server/data/migrations/versions/4.0/03-populate-status-column-for-members.js
allouis: The thinking was that each `down` should leave the db in the state it was before the `up!

Happy to remove the `down` migration though

allouis: Seeing as you've both approved I'm going to merge as-is, and we can update it if necessary in future

daniellockyer (approved):

PR #12588, opened 2021-02-01
Added a git hook to prevent relative references

ErisDS:
Is there a better way to do this 🤔 

- Using relative references causes difficulties when moving code
- Enforcing absolute references will help us remember this!

### DIFF
--- a/.github/hooks/commit-msg
+++ b/.github/hooks/commit-msg
@@ -1,0 +1,22 @@
+#!/bin/bash
+# Modified from https://github.com/chaitanyagupta/gitutils
+
+red='\033[0;31m'
+no_color='\033[0m'
+grey='\033[0;90m'
+
+echo -e "Checking commit message for bad references ${grey}(commit-msg hook)${no_color}"
+
+echo "Message:"
+echo `cat $1`
+
+BAD_MATCH=`grep -E '#\d+' $1 | grep -v 'https://github.com/'`
+
+if [[ ! -z "$BAD_MATCH" ]]; then
+    echo "Your commit message contains relative references, please make them absolute"
+    echo -e "${red}${BAD_MATCH}${no_color}"
+    echo "E.g. 'refs: #1234' should be 'refs: https://github.com/TryGhost/Ghost#1234'"
+    exit 1
+else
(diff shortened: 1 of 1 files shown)


daniellockyer:
@ErisDS Oddly enough, `-E` doesn't seem to work with this on Linux, as I get the following:

```
0 neo@ariel ~/code/Ghost λ grep -E '#\d+' abc | grep -v 'https://github.com/'
grep config error: unknown encoding: #\d+
```

without `-E`:
```
1 neo@ariel ~/code/Ghost λ grep '#\d+' abc | grep -v 'https://github.com/'
refs #123
```

After that change, it all seems to work as expected 🙂 Not sure if it's required for Mac though.

Also, `grunt symlink` (run initially as part of `yarn setup`) won't work for people who have already set up the other symlink, unless we change it to [`overwrite: true`](https://github.com/TryGhost/Ghost/blob/main/Gruntfile.js#L348) (or they manually symlink).

.github/hooks/commit-msg
```diff
@@ -0,0 +1,22 @@
...
+# Modified from https://github.com/chaitanyagupta/gitutils
+
+red='\033[0;31m'
+no_color='\033[0m'
+grey='\033[0;90m'
+
+echo -e "Checking commit message for bad references ${grey}(commit-msg hook)${no_color}"
+
+echo "Message:"
+echo `cat $1`
```
daniellockyer, line 11: Would we be able to nuke these two lines to keep it clean? 😀

ErisDS: oh for sure - I just chucked this together as an idea!

ErisDS:
> The remaining question is how to do this globally for all repos where we care about it 🤔

Yeah I'm not sure if there's something more we can do to lint commit messages

ErisDS:
This was an accident but @daniellockyer is there a sensible way forward with this, or a better way to enforce it? 🤔

daniellockyer:
@ErisDS I was investigating how we could set up "global hooks" using our internal setup tooling, and then only trigger the hook if the repo is `TryGhost` owned, but Git doesn't have support for this and the hacks I found were pretty overkill. 😬

Issue #12565, opened 2021-01-25
Remove "apps" database tables


naz:
refs https://github.com/TryGhost/Team/issues/221

## Context
The concept of "apps" has been removed in Ghost some time ago but there were database tables left as these would be potentially long running migrations. With a preparation of Ghost 4.0 we can add migrations which would need longer time to run. 

### Solution
Remove the tables apps, app_settings, app_fields and permissions_apps - as they are not used.
 
 - [x] drop tables (https://github.com/TryGhost/Ghost/commit/606212d18767d7ed5a61b62be7a68341224c7ed7)
 - [x] cleanup old data (https://github.com/TryGhost/Ghost/pull/12596) 
 
 
### Related
- https://github.com/TryGhost/Ghost/pull/10682

Fhiltyphil88:
> refs https://github.com/TryGhost/Team/issues/221
> 
> 
> 
> ## Context
> 
> The concept of "apps" has been removed in Ghost some time ago but there were database tables left as these would be potentially long running migrations. With a preparation of Ghost 4.0 we can add migrations which would need longer time to run. 
> 
> 
> 
> ### Solution
> 
> Remove the tables apps, app_settings, app_fields and permissions_apps - as they are not used.
> 
>  
> 
> ### Related
> 
> - https://github.com/TryGhost/Ghost/pull/10682

naz:
The https://github.com/TryGhost/Ghost/pull/12596 landed in master, which was last bit needed for full cleanup.

PR #12592, opened 2021-02-02
🏗  Dropped apps related tables

naz:
refs https://github.com/TryGhost/Ghost/issues/12565

- Dropped apps_* tables which were not used since the concept of "apps" has been abandoned
- Uses newly introduced "dropTables" migration utility

Note to reviewers, would pay a special attention to dropTables which I've introduced with [earlier commit](https://github.com/TryGhost/Ghost/commit/b5b65b3a60e87df03e906a53ea723493e5b7d0ca).


naz:
@daniellockyer the CI is failing because it tries to rollback to v3. Is that intentional? Why would we want to do this?

daniellockyer:
@naz The test was for v3 to make sure we could roll back to any of the previous versions. I'll update it to ignore v4 for now

tpatel (changes requested):
I think `dropTables` should use `createIrreversibleMigration` because it's only implementing `up` and not `down`. That way the migration will be `irreversible` so that it's compatible with the CLI!

tpatel (approved):
Just seen that the issue I've mentioned got fixed in https://github.com/naz/Ghost/commit/80481d4076b147858b9d45fe86fa2a39ca3317a7

daniellockyer (approved):

daniellockyer:
@naz The test was updated - feel free to rebase again 🙂

allouis (approved):
so smol!!

naz:
Thanks guys! Merged into main as 606212d

PR #12593, opened 2021-02-02
🏗 Cleanedup app-related settings

naz:
Fields removed in this migrations should have been cleaned up as a result of changed done in https://github.com/TryGhost/Ghost/commit/f74d459dad68ad23a5229cfce3ab9bcde76f5802


daniellockyer:
@naz Would you be open to using `createTransactionalMigration` as per other migrations? It doesn't really make the code shorter but it'd be nice to follow the same pattern 🙂

allouis (approved):
Agree w/ @daniellockyer but otherwise looks good!

Issue #12568, opened 2021-01-25
Fix execution order of settings populateDefaults


naz:
refs https://github.com/TryGhost/Team/issues/221

## Context
[populateDefaults](https://github.com/TryGhost/Ghost/blob/cbfdf79661fb06ea10f687f7c1409d43d002d67d/core/server/data/migrations/hooks/migrate/before.js#L8) call in Ghost's pre-boot phase causes `settings` table population before migrations have a chance to set correct values first (longer explanation available [here](https://github.com/TryGhost/Ghost/issues/12026#issuecomment-656140060)). The method is called in pre-boot only to allow "frontend" to initialize in a minimal way to serve _styled_ 503 page.

### Solution
The problem of the "frontend" boot before the "server" has a chance to initialize will be solved by flipping the dependence of the "server" on the "frontend". "server" will serve it's own 503 page until the boot is complete ("db.ready" event?) and then will notify the "frontend" to start it's initialization. 

### Todos
- [x] introduce independent 503 page served directly from Ghost server
- [x] initialize "frontend" after migrations and populateDefaults have finished execution 
- [x] change boot sequence to call [populateDefaults](https://github.com/TryGhost/Ghost/blob/cbfdf79661fb06ea10f687f7c1409d43d002d67d/core/server/data/migrations/hooks/migrate/before.js#L8) after migrations (https://github.com/TryGhost/Ghost/pull/12595)
- [x] design review of maintenance page
- [x] ~allow to add site icon to maintenance page (reference: https://github.com/TryGhost/Core/commit/fb0d0f09ba58703ee54683140acd07acea470316)~ the plan has been scrapped because of too much effort needed to reach the effect
  - [x] clean up the maintenance page

naz:
Related to the issue is Ghost's boot sequence rewrite - https://github.com/TryGhost/Ghost/compare/main...ErisDS:boot-spike.

naz:
An updated boot sequence has [landed in main](https://github.com/TryGhost/Ghost/commit/0b79abf5b252ab561a6024368d663eeff8da8727) and this initializes the settings defaults after the migrations are done.

Issue #12026, opened 2020-07-09
Permissions error when trying to update settings on 3.23.x


daniellockyer:
reports:

* https://forum.ghost.org/t/saving-settings-configs-error-permission-error-cannot-edit-setting-attempted-to-access-core-setting-from-external-request/15731
* https://forum.ghost.org/t/upgrade-3-22-issues/15667/11

### Issue Summary

Updating to 3.23.x from 3.21.1, and then trying to edit settings will result in the following error:

![https://aws1.discourse-cdn.com/business4/uploads/ghost2/original/2X/9/9f5f5244d0e6f19d57a374fb99d1432f2dda42f7.png](https://aws1.discourse-cdn.com/business4/uploads/ghost2/original/2X/9/9f5f5244d0e6f19d57a374fb99d1432f2dda42f7.png)

### To Reproduce

1. `ghost install 3.21.1 --local`
1. `ghost upgrade 3.23.0 --local`
1. Run the following query:

```
sqlite> select `key`,`type`,`group` from settings where key like 'portal_%';
portal_name|boolean|portal
portal_button|boolean|portal
portal_plans|array|portal
portal_button_style|string|core
portal_button_icon|string|core
portal_button_signup_text|string|core
```

* `portal_button_*` settings seem to have been placed in the `core` group, instead of the desired `portal` group.
* `core` settings cannot be set from outside Ghost, hence the permissions error.

### Technical details:

* Ghost Version: 3.23.x
* Node Version: any
* Browser/OS: any
* Database: SQLite (probably any)

---

Tasks

- [ ] We must to add explicit group and flag migrations for any new setting.
- [ ] We should add a new test to ensure this happens. The quick version for this would be an integrity check for default-settings.json - as this at least stops edits to that file without thinking.

rishabhgrg:
The root cause for this behavior is related to change in settings table and init flow in 3.22.

Any new setting added since version 3.23 (and future versions) , when migrating from any < 3.22 version will always end up with `group` column value as `core` due to this issue. Reason ->

1. The before hook of knex-migrate [here](https://github.com/TryGhost/Ghost/blob/master/core/server/data/migrations/hooks/migrate/before.js#L8)  populates all default settings from default-settings.json, which is called as soon as DB backup is done and before running through migrations.

[...]

PR #12595, opened 2021-02-02
Removed settings populateDefaults pre-migration

naz:
refs https://github.com/TryGhost/Ghost/issues/12568
refs https://github.com/TryGhost/Ghost/commit/5fbc40430bde2e2f975d95e7beb4f9ac2a546ccf
reincarnation :zombie: of https://github.com/TryGhost/Ghost/pull/12587

Having populateDefaults run before migrations creates a chicken and egg problem where populate defaults can create records that are "non-migratable" as happened in https://github.com/TryGhost/Ghost/issues/12026.

Note fore reviewers. This is not a small change (see references)! I have tested it when creating a new instance with clean 4.0, when migrating from 3.21 to 4.0 and also when migrating from 3.21 to 4.0 with disabled [migration in 3.23](https://github.com/TryGhost/Ghost/commit/5fbc40430bde2e2f975d95e7beb4f9ac2a546ccf). Imo we could remove the mentioned migration, but it doesn't harm to stay and would keep the history clean.

Note no. 2. The only place that is now executing populateDefaults is during `settings.init` method execution (server boot). This should always happens **after** the migrations are done. This way the main **source of truth is migrations**.

### DIFF
--- a/core/server/data/migrations/hooks/migrate/before.js
+++ b/core/server/data/migrations/hooks/migrate/before.js
@@ -3,8 +3,5 @@ const models = require('../../../../models');
 
 module.exports = function before() {
     models.init();
-    return dbBackup.backup().then(() => {
-        // ensure that our default settings are created to limit possible db states in migrations
-        return models.Settings.populateDefaults();
-    });
+    return dbBackup.backup();
 };


tpatel (approved):
Looks great!

naz:
@daniellockyer @allouis any thoughts on this change?

daniellockyer (approved):
Have also had a play migrating between different versions & it looks ok to me

allouis (approved):

Issue #12567, opened 2021-01-25
Ghost 4.0 schema changes and migrations


naz:
refs https://github.com/TryGhost/Team/issues/221

This issue is meant to track migrations which will go into upcoming Ghost 4.0 release.

## Context
Some migrations were not added to Ghost 3.x deliberately because of the time it would take to run them. With the release of new major Ghost version it is possible to expect longer migration times.

## 4.0 Migrations
- [ ] Change slug unique constraint on `post` table to be slug+type
- [ ] Clean up webhooks table from "orphaned" and no belonging to any integration records ([source](https://github.com/TryGhost/Team/issues/388#issuecomment-693024105))
  - [ ] Create missing integration/api_key records for orphaned or without valid 'integration_id' webhooks
  - [ ] Add foreign key constraint to `webhooks.integration_id` column
- [ ] Update `canonical_url` type in `posts` table to be `string` instead of `text`
- [ ] TBD audit the rest of tables for any possible changes

naz:
By changing `posts.canonical_url` type to be `string` instead of `text` we still have enough space to dodge MySQLs row size limit. Calculation available [here](https://docs.google.com/spreadsheets/d/1em4aB18GnS86UnW0nz9a5nJ7R0L38eZZMDu-cS_7hrc/edit#gid=0):
![Screenshot from 2021-02-05 19-25-26](https://user-images.githubusercontent.com/675397/106997555-f62a3680-67e7-11eb-8797-44426d9fc1e3.png)

tpatel:
Based on my research in https://github.com/TryGhost/Team/issues/452 I think this column type change doesn't changing anything, because we're using innodb's dynamic row format.

naz:
@tpatel for the settings' table cleanup task, this is current table definition when it has been initialized pre 4.0:
```
mysql> show create table `settings` \G
*************************** 1. row ***************************
       Table: settings
Create Table: CREATE TABLE `settings` (
  `id` varchar(24) NOT NULL,
  `group` varchar(50) NOT NULL DEFAULT 'core',
  `key` varchar(50) NOT NULL,
  `value` text,
  `type` varchar(50) NOT NULL DEFAULT 'core',
  `flags` varchar(50) DEFAULT NULL,
  `created_at` datetime NOT NULL,
  `created_by` varchar(24) NOT NULL,
  `updated_at` datetime DEFAULT NULL,
  `updated_by` varchar(24) DEFAULT NULL,
[...]

PR #12598, opened 2021-02-04
🏗  Altered slug unique constraint in posts table

naz:
refs https://github.com/TryGhost/Ghost/issues/12567

- Changing unique constraint from `slug` to `slug+type` should allow for posts and pages to be created with the same slug
- The constraint will be present on application layer for API v4 while we figure out how to deal with it in API v5

Migrations! Migrations! Migrations!

:warning:  SQLite should be checked with special attention :building_construction: 

:bulb: pro tip, to check unique constraints in SQLite use: `pragma index_list('posts');`, for MySQL: `SHOW CREATE TABLE posts`


tpatel (approved):
Great work 🙌

rshbhgrg (approved):
Integrity spec needs update otherwise looks good 😄

naz:
@daniellockyer ping :wink:

daniellockyer (approved):

PR #12601, opened 2021-02-04
💾 Created & Populated members_subscribe_events table

allouis:

This adds the members_subscribe_events table, and the migrations to create it and populate it. We've also got the beginnings of a basic model to use with it.

@rishabhgrg @daniellockyer @kevinansfield For review - the DB changes are all in https://github.com/TryGhost/Ghost/pull/12601/commits/6b49f3caeb3842700c9812019027a655829fb8aa and https://github.com/TryGhost/Ghost/pull/12601/commits/d82c85cc0f03d0f58adaaa8bbc646a0935d1a89e

### DIFF
--- a/core/server/data/migrations/versions/4.0/06-populate-members-subscribe-events-table.js
+++ b/core/server/data/migrations/versions/4.0/06-populate-members-subscribe-events-table.js
@@ -1,0 +1,49 @@
+const {chunk} = require('lodash');
+const ObjectID = require('bson-objectid');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const allMembers = await knex.select(
+            'id as member_id',
+            'created_at'
+        ).from('members');
+
+        const unsubscribedMembers = await knex.select(
+            'id as member_id',
+            'updated_at as created_at'
+        ).from('members').where('subscribed', false);
+
+        const membersSubscribedEvents = allMembers.map((event) => {
+            return {
+                ...event,
+                id: ObjectID.generate(),
+                subscribed: true,
+                source: 'member'
+            };
+        });
+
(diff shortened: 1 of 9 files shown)


core/server/data/migrations/versions/4.0/06-populate-members-subscribe-events-table.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const allMembers = await knex.select(
+            'id as member_id',
+            'created_at'
+        ).from('members');
+
+        const unsubscribedMembers = await knex.select(
+            'id as member_id',
+            'updated_at as created_at'
```
allouis, line 13: Makes the assumption that the last time the member was updated was when they unsubscribed - not accurate, but it'll do IMO

core/server/data/migrations/versions/4.0/06-populate-members-subscribe-events-table.js
```diff
@@ -0,0 +1,49 @@
...
+        const chunkSize = 199;
+
+        const eventChunks = chunk(allEvents, chunkSize);
+
+        for (const events of eventChunks) {
+            await knex.insert(events).into('members_subscribe_events');
+        }
+    },
+    async function down(knex) {
+        return knex('members_subscribe_events').del();
```
tpatel, line 47: you might want to `return await knex('members_subscribe_events').del();` here?

tpatel (approved):
Looks great 🙌

core/server/data/migrations/versions/4.0/06-populate-members-subscribe-events-table.js
allouis: Chatted offline, gonna just leave as is!

rshbhgrg (approved):

daniellockyer (approved):

PR #12604, opened 2021-02-05
🐛 Fixed Member model removing labels when unset

### LINKED ISSUE #12600: member unsubscribing from newsletter erases their membership labels
member unsubscribing from newsletter erases their membership labels
ghost 3.40.5

apply label(s) to a member and have them unsubscribe from newsletter (by clicking on link in newsletter email, didn't test via portal)

-> label(s) have disappeared

allouis:
closes https://github.com/TryGhost/Ghost/issues/12600

The bookshelf-relations plugin which we use will **remove** all
relations when they are set to an empty array, but will leave them alone
if it's set to undefined.

Our logic to deduplicate uppercase & lowercase version of the same label
was in advertently always setting the labels to an array, but when the
model was saved without passing the labels, this array would be empty.

Here we've added a check which will skip all label handling, if there
are no labels set.

### DIFF
--- a/core/server/api/canary/members.js
+++ b/core/server/api/canary/members.js
@@ -184,7 +184,7 @@ module.exports = {
         permissions: true,
         async query(frame) {
             try {
-                frame.options.withRelated = ['stripeSubscriptions'];
+                frame.options.withRelated = ['stripeSubscriptions', 'labels'];
                 const member = await membersService.api.members.update(frame.data.members[0], frame.options);
 
                 const hasCompedSubscription = !!member.related('stripeSubscriptions').find(sub => sub.get('plan_nickname') === 'Complimentary' && sub.get('status') === 'active');
--- a/core/server/models/member.js
+++ b/core/server/models/member.js
@@ -106,6 +106,11 @@ const Member = ghostBookshelf.Model.extend({
     onSaving: function onSaving(model, attr, options) {
         let labelsToSave = [];
 
+        if (_.isUndefined(this.get('labels'))) {
+            this.unset('labels');
+            return;
+        }
+
         // CASE: detect lowercase/uppercase label slugs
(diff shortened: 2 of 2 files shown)


core/server/api/canary/members.js
```diff
@@ -184,7 +184,7 @@ module.exports = {
         permissions: true,
         async query(frame) {
             try {
-                frame.options.withRelated = ['stripeSubscriptions'];
+                frame.options.withRelated = ['stripeSubscriptions', 'labels'];
```
allouis, line 187: Now that we don't automatically overwrite labels (and thus include them in the result) - we have to explicitly request labels in the result

core/server/models/member.js
```diff
@@ -106,6 +106,11 @@ const Member = ghostBookshelf.Model.extend({
     onSaving: function onSaving(model, attr, options) {
         let labelsToSave = [];
 
+        if (_.isUndefined(this.get('labels'))) {
+            this.unset('labels');
```
allouis, line 110: This is to ensure that an undefined value is treated the same as the absence of a value e.g.
```js
Member.add({name: 'egg', email: 'egg@monst.er'})
Member.add({name: 'egg', email: 'egg@monst.er', labels: undefined})
```
From what I can tell, both calls will be skipped over by `bookshelf-relations` and then inserted - however the second call needs to be looked at by bookshelf relations, so that it can have the property removed before being sent to the database! 

This change ensures that consumers of the Model have a more understandable interface

rishabhgrg: Makes sense 👍

rshbhgrg (approved):
Its been a while so i am not 100%, but I think the explicit handling for null /undefined  seems only around thinking of cleaning up for case-insensitive labels but definitely seems to have missed the case of no labels set in API which caused the problem. The fix looks good to me 👍

Issue #12602, opened 2021-02-04
💾 Member Events 


allouis:
We want to store historical events related to members so that we can visualise this data in the admin. This issue focuses on the data side of things!

#### Tasks

- [x] Newsletter subscription status changes
  - [x] DB Table
  - [x] Model
  - [x] Event capturing
  - [x] Aggregate query
- [x] Paid subscription changes
  - [x] DB Table
  - [x] Model
  - [x] Event capturing
  - [x] Aggregate query
- [x] Status changes
  - [x] DB Table
  - [x] Model
  - [x] Addition of the 'comped' status 
  - [x] Event capturing
  - [x] Aggregate query
- [x] Email address changes
  - [x] DB Table
  - [x] Model
  - [x] Event capturing
- [x] Logins
  - [x] DB Table
  - [x] Model
  - [x] Event capturing
- [x] Payments
  - [x] DB Table
  - [x] Model
  - [x] Aggregate query
- [x] Event Repository 
  - [x] Member timeline
  - [x] Global timeline
  - [x] Subscriptions over time
  - [x] Gross volume over time
  - [x] MRR over time

allouis:
This has been completed in all of the above linked PR's

PR #12618, opened 2021-02-09
Add member event tables

allouis:

This adds the tables&migrations necessary for storing member events

### DIFF
--- a/core/server/data/migrations/versions/4.0/08-add-members-login-events-table.js
+++ b/core/server/data/migrations/versions/4.0/08-add-members-login-events-table.js
@@ -1,0 +1,6 @@
+const {addTable} = require('../../utils');
+module.exports = addTable('members_login_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+    created_at: {type: 'dateTime', nullable: false}
+});
--- a/core/server/data/migrations/versions/4.0/09-add-members-email-change-events-table.js
+++ b/core/server/data/migrations/versions/4.0/09-add-members-email-change-events-table.js
@@ -1,0 +1,8 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('members_email_change_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+    email: {type: 'string', maxlength: 191, nullable: false, unique: false, validations: {isEmail: true}},
+    created_at: {type: 'dateTime', nullable: false}
+});
--- a/core/server/data/migrations/versions/4.0/10-add-members-status-events-table.js
+++ b/core/server/data/migrations/versions/4.0/10-add-members-status-events-table.js
@@ -1,0 +1,8 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('members_status_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+    status: {type: 'string', maxlength: 50, nullable: false},
+    created_at: {type: 'dateTime', nullable: false}
+});
--- a/core/server/data/migrations/versions/4.0/11-add-members-paid-subscription-events-table.js
+++ b/core/server/data/migrations/versions/4.0/11-add-members-paid-subscription-events-table.js
@@ -1,0 +1,12 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('members_paid_subscription_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+    from_plan: {type: 'string', maxlength: 50, nullable: true},
+    to_plan: {type: 'string', maxlength: 50, nullable: true},
+    currency: {type: 'string', maxLength: 3, nullable: false},
+    source: {type: 'string', maxlength: 50, nullable: false},
+    mrr_delta: {type: 'integer', nullable: false},
+    created_at: {type: 'dateTime', nullable: false}
+});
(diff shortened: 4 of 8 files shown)


core/server/data/migrations/versions/4.0/08-add-members-login-events-table.js
```diff
@@ -0,0 +1,6 @@
+const {addTable} = require('../../utils');
+module.exports = addTable('members_login_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+    created_at: {type: 'dateTime', nullable: false}
```
rishabhgrg, line 5: Just wondering if it makes sense to have a `source` here which can have the token/link that was used to login ? No strong usecase except just storing extra info for/when needed

rshbhgrg (approved):
Looks good, just a small comment on possible extra column on login events 👍

core/server/data/migrations/versions/4.0/08-add-members-login-events-table.js
allouis: Hmm, I think we should add stuff as/when we have the usecase for it, at this stage. Wondering what that could be used for?

rishabhgrg: > I think we should add stuff as/when we have the usecase for it, at this stage

Cool yea, definitely no immediate usecase in mind except any audit purposes, happy to leave it out :)

daniellockyer (approved):

core/server/data/migrations/versions/4.0/08-add-members-login-events-table.js
matthanley: imo there'd probably be a stronger use-case for geolocation or IP here 🤔

core/server/data/migrations/versions/4.0/09-add-members-email-change-events-table.js
```diff
@@ -0,0 +1,9 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('members_email_change_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+    previous_email: {type: 'string', maxlength: 191, nullable: false, unique: false, validations: {isEmail: true}},
+    email: {type: 'string', maxlength: 191, nullable: false, unique: false, validations: {isEmail: true}},
```
matthanley, line 7: `members_paid_subscription_events` uses `from_*` and `to_*` prefixes - should it be the same here for email fields?

core/server/data/migrations/versions/4.0/10-add-members-status-events-table.js
```diff
@@ -0,0 +1,8 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('members_status_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+    status: {type: 'string', maxlength: 50, nullable: false},
```
matthanley, line 6: As above - this is the only table that doesn't store both old and new values. Should we have `from_status` and `to_status` here?

core/server/data/migrations/versions/4.0/11-add-members-paid-subscription-events-table.js
```diff
@@ -0,0 +1,13 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('members_paid_subscription_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+    from_plan: {type: 'string', maxlength: 50, nullable: true},
+    to_plan: {type: 'string', maxlength: 50, nullable: true},
+    amount: {type: 'integer', nullable: false},
```
matthanley, line 8: Not sure what purpose this serves here. Is this new amount or prev amount? Do we need it as well as MRR delta? or should we store old and new amounts instead of MRR delta?

core/server/data/schema/schema.js
```diff
@@ -355,6 +355,47 @@ module.exports = {
         updated_at: {type: 'dateTime', nullable: true},
         updated_by: {type: 'string', maxlength: 24, nullable: true}
     },
+    members_payment_events: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+        amount: {type: 'integer', nullable: false},
+        currency: {type: 'string', maxLength: 3, nullable: false},
+        source: {type: 'string', maxlength: 50, nullable: false},
```
matthanley, line 363: Possibly needs `isIn` validations as we have expected values for this (just stripe?)

core/server/data/migrations/versions/4.0/09-add-members-email-change-events-table.js
allouis: Sure, I can update that!

core/server/data/migrations/versions/4.0/10-add-members-status-events-table.js
allouis: Sure thing, will add this

core/server/data/migrations/versions/4.0/11-add-members-paid-subscription-events-table.js
allouis: Will remove :+1:

core/server/data/migrations/versions/4.0/10-add-members-status-events-table.js
allouis: I'm wondering actually - do what's the usecase for this?

core/server/data/migrations/versions/4.0/09-add-members-email-change-events-table.js
allouis: Removed previous_email as it's superfluous for now

core/server/data/migrations/versions/4.0/10-add-members-status-events-table.js
allouis: As discussed, this isn't necessary to store

matthanley (approved):

PR #12615, opened 2021-02-09
🏗 Refactor slack unsplash settings

naz:
reincarnated :zombie: https://github.com/TryGhost/Ghost/pull/12050
~:x: blocked by https://github.com/TryGhost/Ghost/pull/12595~
closes https://github.com/TryGhost/Team/issues/331
refs https://github.com/TryGhost/Ghost/issues/10318

:construction: this is WIP and creating this PR for visibility as it will need migration reviews from few people

The changes are a rebase of existing changes over `main` branch and moved migrations into `4.0` set.

Note, this PR will most likely be split into two separate parts - one for slack setting and one for unsplash. For now keeping it as one changeset to keep changes visible wrt what has already been implemented.

Left to do:
- [ ] unsplash setting 'object' notation deprecateion in canary/v4 APIs
- [ ] new slack options handling in v2/v3 APIs 
- [ ] importer handling for both unsplash/slack properties

PR #12632, opened 2021-02-10
🏗 Migrated unsplash setting to up to date convention

naz:
splits https://github.com/TryGhost/Ghost/pull/12615 into "slack" and "unsplash" parts
refs closes TryGhost/Team#331
refs #10318

Note to reviewers: the part that needs thorough review is the migration located in [10-refactor-unsplash-setting.js  file](https://github.com/TryGhost/Ghost/pull/12632/files#diff-523b93c05310cc634755e8fc294838e2c2181112e196931c057ec5eda09aedab). The rest is optional if you feel like having a look :wink: 

TODO:
- [x] handling for v2/v3/v4(canary) APIs
- [x] importer compatibility
- [x] clarify the `Object` return type in v4 api

Change/compatibility table for this PR:

### importer

`unsplash` :heavy_check_mark: import from both boolean AND object

#### `GET /settings/` (browse)

`unsplash` :heavy_check_mark: present in response **with `boolean` value**

#### `GET /settings/:settingName` (read)

`unsplash` :heavy_check_mark: present in response **with `boolean` value**

#### `PUT /settings/` (edit)

`unsplash` :heavy_check_mark: updates setting, accepts a boolean

#### `GET /settings/` (browse)

`unsplash` :heavy_check_mark: present in response **with `Object` value**

#### `GET /settings/:settingName` (read)

`unsplash` :heavy_check_mark: present in response **with `Object` value**

#### `PUT /settings/` (edit)

`unsplash` :heavy_check_mark: updates setting, accepts either boolean or object

#### `GET /settings/` (browse)

`unsplash` :heavy_check_mark: present in response with object value

#### `GET /settings/:settingName` (read)

`unsplash` :heavy_check_mark: present in response with object value

#### `PUT /settings/` (edit)

`unsplash` :heavy_check_mark: updates setting, accepts object


tpatel (changes requested):
I would either add `irreversible: true` in the config (that would require an update to the helper) or `return Promise.reject();` in `down()` like in https://github.com/TryGhost/Ghost/blob/873d21d9751345fe7c983719a4a52aed1ba11cab/core/server/data/migrations/utils.js#L261

I think it's better than doing nothing so that we can't run this migration twice.

naz:
@tpatel good spot! Changed the migration :wink:

tpatel (approved):
awesome 🙌

core/server/data/schema/default-settings.json
```diff
@@ -408,8 +408,11 @@
     },
     "unsplash": {
         "unsplash": {
-            "defaultValue": "{\"isActive\": true}",
-            "type": "object"
+            "defaultValue": "true",
+            "validations": {
+                "isIn": [["true", "false"]]
```
allouis, line 413: Do we want to include the `isEmpty` validation too? 
```suggestion
                "isEmpty": false,
                "isIn": [["true", "false"]]
```

core/server/data/migrations/versions/4.0/10-refactor-unsplash-setting.js
```diff
@@ -0,0 +1,29 @@
+const {createIrreversibleMigration} = require('../../utils');
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    const unsplashSetting = await knex('settings')
```
allouis, line 4: Daniel reminded me on another PR, we probs wanna ass a little log message for visibility!

core/server/data/importer/importers/data/settings.js
```diff
@@ -90,6 +102,48 @@ class SettingsImporter extends BaseImporter {
...
+                }
+
+                if (deprecatedSettingValue) {
+                    deprecatedSupportedSettingsOneToManyMap[key].forEach(({from, to}) => {
+                        let value = _.isObject(deprecatedSettingValue)
+                            ? _.get(deprecatedSettingValue, from)
+                            : deprecatedSetting.value;
+
+                        this.dataToImport.push({
+                            id: deprecatedSetting.id || ObjectId.generate(),
```
allouis, line 131: This could cause duplicate `id`'s if the deprecated setting is split into multiple settings

core/server/data/importer/importers/data/settings.js
```diff
@@ -17,6 +18,17 @@ const deprecatedSupportedSettingsMap = {
     ghost_foot: 'codeinjection_foot'
 };
 
+const deprecatedSupportedSettingsOneToManyMap = {
```
allouis, line 21: I like the generalised approach of this! Though I wonder if it might be a little easier to follow with an object like:
```js
unsplash: [{
  getValue(valueFromDb) {
    return JSON.parse(valueFromDb).isActive;
  },
  key: 'unsplash',
  group: 'unsplash',
  type: 'boolean'
]]
```

core/server/data/importer/importers/data/settings.js
```diff
@@ -90,6 +102,48 @@ class SettingsImporter extends BaseImporter {
...
+                            ? _.get(deprecatedSettingValue, from)
+                            : deprecatedSetting.value;
+
+                        this.dataToImport.push({
+                            id: deprecatedSetting.id || ObjectId.generate(),
+                            key: to.key,
+                            value: value,
+                            group: to.group,
+                            type: to.type,
+                            flags: null,
```
allouis, line 136: probably might as well do `to.flags || null`?

allouis (changes requested):
All looks good, only the comment about duplicate `id`'s is essential

core/server/data/schema/default-settings.json
naz: Yeah, I think that wouln't hurt and would be more explicit :+1:  Although I'm a bit torn on what difference that would make in most cases where we use `isIn` validation in combination with `isEmpty`. If `isIn` does not specify empty string or `null` as an acceptable value that should mean anything but specified values should fail.

core/server/data/migrations/versions/4.0/10-refactor-unsplash-setting.js
naz: Yeeees, agreed :relaxed:

core/server/data/importer/importers/data/settings.js
naz: Hm.. :thinking:  Yup, very good point!

core/server/data/importer/importers/data/settings.js
naz: yupp :+1:

core/server/data/importer/importers/data/settings.js
naz: I like this notation! It's smaller and agree, it's way more readable :+1:

naz: Tried playing with it a little, but having multiple fields (slack has multiple mappings as well) then there's lots of duplicate code that comes from having to do error handling for `JSON.parse`. Might come up with something nice later on, but for now leaving it as "good enough, does the job" code :sweat_smile:

allouis: Makes perfect sense to me :relaxed:

allouis (approved):

daniellockyer (approved):

naz:
@daniellockyer @tpatel @allouis I've updated the migration here to follow idempotence rule. The commit which fixes is is this one - https://github.com/TryGhost/Ghost/pull/12632/commits/0bb7be38623d6d4fcd3d5335ad862dd70f2f9ee5

tpatel (approved):
The migration looks great 🙌

allouis (approved):

daniellockyer (approved):

PR #12636, opened 2021-02-11
Updated members_paid_subscription_events table

allouis:

This allows us to store the stripe_id of the subscription events


allouis:
Hey @daniellockyer What do we want to do about modifying unreleased migrations? It seems OTT to have to do a column migration when only devs have been running these? :thinking:

matthanley (approved):

rshbhgrg (approved):

daniellockyer:
@allouis Oh yeah, I'd say it's totally fine. Everything unreleased in `main` is WIP and shouldn't be relied on. Internally, we can just make sure people nuke their DB if needed.

daniellockyer (approved):

PR #12638, opened 2021-02-11
Added the frontmatter column to meta

tpatel:
issue https://github.com/TryGhost/Team/issues/452

I've just added a new column to the `post_meta` table (future `meta` table) to prepare for adding a frontmatter feature to posts.

### DIFF
--- a/core/server/data/migrations/versions/4.0/15-add-frontmatter-column-to-meta.js
+++ b/core/server/data/migrations/versions/4.0/15-add-frontmatter-column-to-meta.js
@@ -1,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('posts_meta', 'frontmatter', {
+    type: 'text',
+    maxlength: 65535,
+    nullable: true
+});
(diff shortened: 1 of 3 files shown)


core/server/data/migrations/versions/4.0/15-add-frontmatter-column-to-meta.js
```diff
@@ -0,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('posts_meta', 'frontmatter', {
+    type: 'text',
+    maxlength: 65535,
+    nullable: false
```
daniellockyer, line 6: `nullable: false` vs `nullable: true` in schema definition 🙂

tpatel: good catch 🙊

tpatel: fixed ✔️

naz (approved):
:rocket:

daniellockyer (approved):

allouis (approved):

PR #12639, opened 2021-02-12
🏗  Removed labs from settings table

naz:
refs https://github.com/TryGhost/Team/issues/332

By default members will be always "on" starting Ghost 4.0, so there's no need for this flag anymore. 

Note for reviewers. the migration under review here is - https://github.com/TryGhost/Ghost/pull/12639/commits/c86e674a3507f037fdce79646b017dd87600b4bf#diff-0c3b0fd661dd42a7397fbd89076b436b028481c32ea5f169b022b6508def52ae. 

The rest of changes are needed because the migration breaks quite a lot of tests after execution.

### DIFF
--- a/core/server/api/canary/utils/serializers/input/settings.js
+++ b/core/server/api/canary/utils/serializers/input/settings.js
@@ -5,7 +5,8 @@ const settingsCache = require('../../../../../services/settings/cache');
 
 const DEPRECATED_SETTINGS = [
     'bulk_email_settings',
-    'slack'
+    'slack',
+    'labs'
 ];
 
 const deprecatedSupportedSettingsOneToManyMap = {
--- a/core/server/api/v2/utils/serializers/input/settings.js
+++ b/core/server/api/v2/utils/serializers/input/settings.js
@@ -5,7 +5,8 @@ const settingsCache = require('../../../../../services/settings/cache');
 
 const DEPRECATED_SETTINGS = [
     'bulk_email_settings',
-    'slack'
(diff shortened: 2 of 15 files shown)


core/server/data/migrations/versions/4.0/16-remove-labs-setting.js
```diff
@@ -0,0 +1,23 @@
...
+const {createIrreversibleMigration} = require('../../utils');
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    let result = await knex('settings')
+        .where('key', '=', 'labs')
+        .select('value');
+
+    if (!result || !result[0]) {
+        logging.warn(`Could not find labs setting`);
+        result = [{}];
```
tpatel, line 11: I think you should `return` here, otherwise the migration will break on `result[0].value`.

tpatel (changes requested):
I think there could be an issue when the lab settings isn't there (although I'm not sure when it can happen). I wrote the comment inline 📝

core/server/data/migrations/versions/4.0/16-remove-labs-setting.js
naz: @tpatel I've decided the best thing here would be to completely remove the labs as a key altogether :sweat_smile:

tpatel (approved):
Looks great 👌

daniellockyer (approved):

rshbhgrg (approved):

PR #12640, opened 2021-02-12
Member event models

allouis:

This adds models for each of the member event tables, as well as a migration to backfill members_paid_subscription_events table

### DIFF
--- a/core/server/data/migrations/versions/4.0/14-populate-members-paid-subscription-events-table.js
+++ b/core/server/data/migrations/versions/4.0/14-populate-members-paid-subscription-events-table.js
@@ -1,0 +1,104 @@
+const {chunk} = require('lodash');
+const ObjectID = require('bson-objectid');
+const {createTransactionalMigration} = require('../../utils');
+const logging = require('../../../../../shared/logging');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Populating members_paid_subscription_events from members_stripe_customers_subscriptions');
+        const allSubscriptions = await knex
+            .select(
+                'c.member_id',
+                's.status',
+                's.start_date',
+                's.updated_at',
+                's.created_at',
+                's.plan_id',
+                's.plan_amount',
+                's.plan_currency',
+                's.plan_interval'
+            )
+            .from('members_stripe_customers_subscriptions as s')
+            .join('members_stripe_customers as c', 'c.customer_id', '=', 's.customer_id');
+
+        function calculateMrrDelta({interval, amount}) {
+            if (interval === 'year') {
+                return Math.floor(amount / 12);
+            }
+
+            if (interval === 'month') {
+                return amount;
+            }
+        }
+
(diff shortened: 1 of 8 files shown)


allouis:
Migration here: https://github.com/TryGhost/Ghost/pull/12640/commits/1121e109600cbc94fddb6f11fee73f05a55bc2f1

rshbhgrg (approved):

core/server/models/member-payment-event.js
```diff
@@ -0,0 +1,47 @@
+const errors = require('@tryghost/errors');
+const ghostBookshelf = require('./base');
+
+const MemberPaymentEvent = ghostBookshelf.Model.extend({
```
matthanley, line 4: Doesn't this model use the `members_payment_events` table [here](https://github.com/TryGhost/Ghost/blob/main/core/server/data/schema/schema.js#L358)? Also `.select(knex.raw('SUM(volume) as gross_volume'))` should be `.select(knex.raw('SUM(amount) as gross_volume'))`

core/server/data/migrations/versions/4.0/14-populate-members-paid-subscription-events-table.js
```diff
@@ -0,0 +1,112 @@
...
+                        status: subscription.status,
+                        amount: subscription.plan_amount,
+                        interval: subscription.plan_interval
+                    }),
+                    created_at: subscription.start_date
+                };
+                events.push(subscriptionCreatedEvent);
+            }
+
+            if (subscription.status === 'canceled') {
```
matthanley, line 78: I don't think this will work as expected for cancelled events: if all of my subscriptions are cancelled, this migration will give me a negative aggregated MRR because we don't have the corresponding `created` events to cancel out. I think MRR delta for backfilling cancelled events should be 0.

core/server/models/member-payment-event.js
allouis: THANKS

core/server/data/migrations/versions/4.0/14-populate-members-paid-subscription-events-table.js
allouis: Okay maybe it's a bit too implicit, but the `else` block above will handle the initial subscription event so we would have an event with `+X` MRR and then the `-X` MRR afterwards

matthanley: I completely missed that - but won't we actually get two lots of `-X` MRR since the status is still cancelled?

allouis: Yes you're totally right, I'll move the modifier out of the calculateMrrDelta function

daniellockyer:
@allouis I think this looks ok but can we add a simple logging line somewhere like "Filling in historical members paid subscription events", just to make it easier to track migrations 🙂

allouis:
@daniellockyer dun!!

daniellockyer (approved):

Issue #12646, opened 2021-02-15
Update mobiledoc/markdown rendering for 4.0


kevinansfield:
We want the ability to make changes in our rendered HTML output without breaking old content. An example of such a change in 4.0 will be modifying header elements' `id` attribute value to better serve non-English languages. If we were to make that change without considering old content then any time old content is re-rendered (through editing old posts, or forced re-render in a major upgrade) then deep-links could break from `id` attributes adopting the newer scheme.

By introducing the concept of "Ghost version" to our mobiledoc (separate to the underlying mobiledoc spec version) we can use it in our card renderers and post-render modifiers to change the output in new content without breaking the old.

We also want to remove the outdated/deprecated markdown card alias (Ghost 1.0 had a different name for markdown cards) and the custom card-only version that was added to facilitate adding the 1.0 wrapper element. This is a cleanup task.

- [x] Test `mobiledoc-kit` and `mobiledoc-dom-renderer` to make sure there are no problems adding top-level properties to mobiledoc documents
- [x] Decide on versioning scheme
    - tied to Ghost major or separate so we can modify HTML output without breaking changes outside of major releases?
- [x] Migration to modify mobiledoc of all posts
  - [x] add a `ghostVersion` property to mobiledoc of all posts
  - [x] change markdown card names from `card-markdown` to `markdown` (cleanup of 1.0 content)
- [x] Ensure newer `ghostVersion` is set for all newly created mobiledoc (Ghost + Admin)
  - the alternative would be setting the version at time of publication _but_ we run into potential problems there with content being unpublished and republished for any reason (send as email, force automations to run, etc)
- [x] Remove card-markdown card alias in `kg-default-cards`
- [x] Remove handling of `cardOptions.version` that is no longer used
- [x] Update mobiledoc renderers to use top-level `ghostVersion`
- [x] Update header id generation to use a less destructive algorithm (closes https://github.com/TryGhost/Ghost/issues/9740)
  - [x] [Top-level DOM modifier](https://github.com/TryGhost/Koenig/blob/master/packages/kg-mobiledoc-html-renderer/lib/mobiledoc-html-renderer.js#L34-L55)
  - [x] [Markdown renderer](https://github.com/TryGhost/Koenig/blob/master/packages/kg-markdown-html-renderer/lib/markdown-html-renderer.js#L11-L20)

kevinansfield:
[...]

PR #12648, opened 2021-02-15
Added 4.0 mobiledoc update migration

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/12646

- adds `ghostVersion` property to mobiledoc of all posts so that we can pin the rendering output of existing content to protect against future breaking changes
- cleans any deprecated markdown card names from old Ghost 1.0 content
- cleans up HR cards that had an [accidental payload added](https://github.com/TryGhost/Koenig/commit/610d801bcf83c7159215e56af3638da24424fb1a)

### DIFF
--- a/core/server/data/migrations/versions/4.0/17-update-mobiledoc.js
+++ b/core/server/data/migrations/versions/4.0/17-update-mobiledoc.js
@@ -1,0 +1,59 @@
+const logging = require('../../../../../shared/logging');
+const {createIrreversibleMigration} = require('../../utils');
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    logging.info('Adding version to posts.mobiledoc objects and cleaning up deprecated card names');
+
+    await knex.transaction(async (trx) => {
+        // get list of posts ids, use .forUpdate to lock rows until the transaction is finished
+        const postIdRows = await knex('posts')
+            .transacting(trx)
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.0/17-update-mobiledoc.js
```diff
@@ -0,0 +1,53 @@
...
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    logging.info('Adding version to posts.mobiledoc objects and cleaning up deprecated card names');
+
+    await knex.transaction(async (trx) => {
+        // get list of posts ids, use .forUpdate to lock rows until the transaction is finished
+        const postIdRows = await knex('posts')
+            .transacting(trx)
+            .forUpdate()
+            .select('id');
```
kevinansfield, line 12: I opted for grabbing all post ids rather than id+mobiledoc because some sites have 1000's of posts and 100's of megabytes of post content which doesn't seem like a good candidate for pre-loading into memory.

Interested to hear if anyone things this is optimising for edge cases at the cost of slower migrations for the majority.

matthanley: This is the best approach imo 👍

tpatel (approved):
Looks great 🙌

PR #12649, opened 2021-02-15
💡 Resolve orphaned webhooks

tpatel:
issue https://github.com/TryGhost/Ghost/issues/12567

1. Get the orphaned webhooks
2. Create new integrations for each orphaned webhook
3. Update each orphaned webhook with their new integration

An orphaned webhook is a webhook that has no matching integration, it was possible to create them before https://github.com/TryGhost/Team/issues/388. Resolving this issue enable users to see and manage these webhooks via the ghost admin interface.

### DIFF
--- a/core/server/data/migrations/versions/4.0/21-solve-orphaned-webhooks.js
+++ b/core/server/data/migrations/versions/4.0/21-solve-orphaned-webhooks.js
@@ -1,0 +1,92 @@
+const {createIrreversibleMigration} = require('../../utils');
+const logging = require('../../../../../shared/logging');
+const ObjectID = require('bson-objectid');
+const crypto = require('crypto');
+
+// createSecret is copied from core/server/models/api-key.js
+const createSecret = (type) => {
+    const bytes = type === 'content' ? 13 : 32;
+    return crypto.randomBytes(bytes).toString('hex');
+};
+
+module.exports = createIrreversibleMigration(
+    async function up(knex) {
+        logging.info('Resolving the orphaned webhooks');
+
+        const orphanedWebhooks = await knex('webhooks')
+            .select('webhooks.id')
+            .leftJoin('integrations', 'integrations.id', 'webhooks.integration_id')
+            .where('integrations.id', null);
+
+        if (orphanedWebhooks.length === 0) {
+            logging.info('No orphaned webhooks found, skipping');
+            return;
+        }
+
+        const [role] = await knex('roles')
+            .select('roles.id')
+            .where('name', 'Admin Integration');
+
+        if (!role) {
+            logging.warn('Failed to apply migration because of missing role.');
+            return;
+        }
+
+        const now = knex.raw('CURRENT_TIMESTAMP');
+        const id = ObjectID.generate();
+
+        const integration = {
+            id: id,
+            type: 'custom',
+            name: `Legacy webhooks`,
+            slug: `legacy-webhooks-${id}`,
+            icon_image: null,
+            description: `This integration was created as part of the 4.0 migration. It contains all webhooks created via the API that weren't visible in the admin interface previously.`,
+            created_at: now,
+            created_by: 1,
+            updated_at: now,
+            updated_by: 1
+        };
+
+        await knex('integrations')
+            .insert(integration);
+
+        const contentKey = {
+            id: ObjectID.generate(),
+            type: 'content',
+            secret: createSecret('content'),
+            role_id: null,
+            integration_id: integration.id,
+            created_at: now,
+            created_by: 1,
+            updated_at: now,
+            updated_by: 1
+        };
+        await knex('api_keys').insert(contentKey);
+
+        const adminKey = {
+            id: ObjectID.generate(),
+            type: 'admin',
+            secret: createSecret('admin'),
+            role_id: role.id,
+            integration_id: integration.id,
+            created_at: now,
+            created_by: 1,
+            updated_at: now,
+            updated_by: 1
+        };
+        await knex('api_keys').insert(adminKey);
+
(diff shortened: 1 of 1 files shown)


Fhiltyphil88 (commented):
Cyphering bot

core/server/data/migrations/versions/4.0/17-solve-orphaned-webhooks.js
```diff
@@ -0,0 +1,47 @@
...
+        for (let i = 0; i < orphanedWebhooks.length; i++) {
+            // Create new integration
+
+            const webhook = orphanedWebhooks[i];
+            const now = knex.raw('CURRENT_TIMESTAMP');
+
+            const integration = {
+                id: ObjectID.generate(),
+                type: 'custom',
+                name: `Generated integration ${i + 1}`,
```
naz, line 23: @tpatel because the `name`, `slug` and `description` parts will be visible in the UI they should get a copy review from someone from design team. My suggestion, reuse `name` from the webhook if it exists and point our that the integration was generated automatically for specific webhook in the `description.

core/server/data/migrations/versions/4.0/17-solve-orphaned-webhooks.js
```diff
@@ -0,0 +1,47 @@
...
+                slug: `generated-integration-${i + 1}`,
```
naz, line 24: Because of a low but still a possibility to hit `slug` unique constraint using `i` as a number is suboptimal :thinking: A randomly generated string with a number should be less prone to hitting the constraint.

core/server/data/migrations/versions/4.0/17-solve-orphaned-webhooks.js
tpatel: I've asked for help for the copy 👍

core/server/data/migrations/versions/4.0/17-solve-orphaned-webhooks.js
tpatel: Good point, I've switched to the id of the object. Is it ok to share the id of the object? If not I could do another `ObjectID.generate()` for the name/slug.

matthanley:
What’s the benefit in creating a new integration for each webhook vs nesting all of the webhooks under a single new integration?

tpatel:
That's a good point, I'll move to a single integration instead 👍

tpatel:
Updates

1. Instead of creating multiple integrations, I've just generated one.
2. I've created the api keys for the generated integration.

⚠️  The admin api key should have a role ID, I'm not sure which one to assign yet.

naz:
@tpatel in this situation the orphaned webhooks were meant to belong to custom integration, which I believe belong to "external apps" role :thinking: 
```
       name: Admin Integration
description: External Apps
```

core/server/data/migrations/versions/4.0/17-solve-orphaned-webhooks.js
tpatel: The copy has been validated by Matt and Zimo ✅

tpatel:
@naz @daniellockyer @allouis the PR is ready for review

core/server/data/migrations/versions/4.0/21-solve-orphaned-webhooks.js
```diff
@@ -0,0 +1,92 @@
...
+            created_at: now,
+            created_by: 1,
+            updated_at: now,
+            updated_by: 1
+        };
+
+        await knex('integrations')
+            .insert(integration);
+
+        const contentKey = {
```
allouis, line 54: Out of interest, what's the reason for creating API keys for this integration?

allouis (approved):

core/server/data/migrations/versions/4.0/21-solve-orphaned-webhooks.js
tpatel: I haven't found a place in the code that created the API key, so I assumed that I had to do it by hand, given that I'm not using any code from the integration service.

core/server/data/migrations/versions/4.0/21-solve-orphaned-webhooks.js
```diff
@@ -0,0 +1,92 @@
...
+            logging.info('No orphaned webhooks found, skipping');
+            return;
+        }
+
+        const [role] = await knex('roles')
+            .select('roles.id')
+            .where('name', 'Admin Integration');
+
+        if (!role) {
+            logging.info('Failed to apply migration because of missing role.');
```
daniellockyer, line 31: I would maybe make this `logging.warn/error`? My internal thinking for `info` would be messages where thing are happening as expected. However given the message, I'm not sure which one out of `error`/`warn` it should be. Technically the migration cannot continue due to bad state, so I'd expect error, but we don't want to start logging out errors if there's nothing they can do. @naz What do you think?

daniellockyer (approved):

core/server/data/migrations/versions/4.0/21-solve-orphaned-webhooks.js
naz: We don't have a precedent for logging `error` levels within migrations. I think `warn` is the level we usually go with when the state is not what one expects.

naz (changes requested):
Only a small change regarding logging level [here](https://github.com/TryGhost/Ghost/pull/12649#discussion_r580668582). Otherwise it's good to hit the `main` :dash:

core/server/data/migrations/versions/4.0/21-solve-orphaned-webhooks.js
tpatel: I've switched to `warn` 👍

PR #12655, opened 2021-02-16
Split the "Too many requests" error in ping service

tpatel:
issue https://github.com/TryGhost/Team/issues/362

It prevents the "429 Too many requests" http errors to generate 500 http errors in ghost logs.


naz:
@tpatel this is absolutely a right direction. Merging this commit as it's already much more useful to be in `main` to our current implementation!

naz (approved):

PR #12658, opened 2021-02-16
Added migration to transform all post content urls from absolute to relative

kevinansfield:
refs https://github.com/TryGhost/Team/issues/467
refs https://github.com/TryGhost/Team/issues/221

- we added robust url transformation in Ghost 3.0 which ensures newly created or edited content is stored with relative URLs in the database
- this migration loops over all posts (and posts_meta) in the database performing the absolute->relative transformation to normalise data for old posts
- `posts.html` is skipped because a later migration will regenerate that from `posts.mobiledoc`


tpatel (approved):
Looks great 🙌

daniellockyer (approved):

PR #12631, opened 2021-02-10
🏗 Migrated slack setting to up to date convention

naz:
splits https://github.com/TryGhost/Ghost/pull/12615 into "slack" and "unsplash" parts
refs closes TryGhost/Team#331
refs #10318

Note to reviewers: the part that needs thorough review is the migration located in [09-refactor-slack-setting.js file](https://github.com/TryGhost/Ghost/pull/12631/files#diff-afd4a833f1ba2f19533e829814151b7db43e1b4760586001b5130139444d8313) Used irreversible migration just like pointed in the other review for similar ([unsplash](https://github.com/TryGhost/Ghost/pull/12632#pullrequestreview-589317545)) migration. The rest is optional if you feel like having a look :wink: 

TODO:
- [x] handling for v2/v3/v4(canary) APIs
- [x] importer compatibility

Change/compatibility table for this PR:
### importer

`slack_username` :heavy_check_mark: import
`slack_url` :no_entry_sign: ignore
`slack` :heavy_check_mark: import `slack.username`
`slack` :no_entry_sign: ignore `slack.url`

#### `GET /settings/` (browse)

`slack_username` :heavy_check_mark: present in response
`slack_url` :heavy_check_mark: present in response
`slack` :heavy_check_mark: present in response

#### `GET /settings/:settingName` (read)

`slack_username` :heavy_check_mark: present in response
`slack_url` :heavy_check_mark:  present in response
`slack` :heavy_check_mark: present in response

#### `PUT /settings/` (edit)

`slack_username` :heavy_check_mark: updates setting
`slack_url` :heavy_check_mark: updates setting
`slack` ⚠️ : updates setting **unless** `slack_username` or `slack_url` are set (new keys take priority)

#### `GET /settings/` (browse)

`slack_username` :no_entry_sign: not present in response
`slack_url` :no_entry_sign: not present in response
`slack` :heavy_check_mark:  present in response

#### `GET /settings/:settingName` (read)

`slack_username` :no_entry_sign: ignored (404)
`slack_url` :no_entry_sign: ignored (404)
`slack` :heavy_check_mark:  present in response

#### `PUT /settings/` (edit)

`slack_username` :no_entry_sign: ignored (no error)
`slack_url` :no_entry_sign: ignored (no error)
`slack` :heavy_check_mark:  updates setting

PR #12659, opened 2021-02-17
🏗  Fixed slack setting migration to follow idempotance rule

naz:
refs https://github.com/TryGhost/Ghost/pull/12631

- Initialy writen migration would break if executed twice
- Also accounted for situation if only part of the migration executed leaving only one of the values migrated

### DIFF
--- a/core/server/data/migrations/versions/4.0/16-refactor-slack-setting.js
+++ b/core/server/data/migrations/versions/4.0/16-refactor-slack-setting.js
@@ -1,7 +1,29 @@
 const ObjectId = require('bson-objectid').default;
 const {createIrreversibleMigration} = require('../../utils');
+const logging = require('../../../../../shared/logging');
 
 module.exports = createIrreversibleMigration(async (knex) => {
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.0/16-refactor-slack-setting.js
```diff
@@ -28,33 +50,43 @@ module.exports = createIrreversibleMigration(async (knex) => {
...
+                value: slackUrl,
+                created_by: 1,
+                created_at: now,
+                updated_by: 1,
+                updated_at: now
+            });
+    }
+
+    if (!slackUsernameSetting) {
+        logging.info(`Populating slack_url setting with value: ${slackUsername}`);
```
tpatel, line 72: nitpick: it's `slack_username` here

tpatel (approved):
Looks great 🙌  just one small comment

kevinansfield (approved):

daniellockyer (approved):
Same comment as @tpatel but LGTM otherwise 🙂

PR #12660, opened 2021-02-17
Added posts html regeneration migration

kevinansfield:
refs https://github.com/TryGhost/Team/issues/467
refs https://github.com/TryGhost/Team/issues/221

- we've introduced backwards-compatible changes to rendering in 3.0 such as srcset and sizes which will only have taken effect on posts created or edited since the changes were made
- 4.0 brings additional changes such as image card width/height
- re-generating the `html` field of all posts from the `mobiledoc` brings all content up to latest rendering output

### DIFF
--- a/core/server/data/migrations/versions/4.0/23-regenerate-posts-html.js
+++ b/core/server/data/migrations/versions/4.0/23-regenerate-posts-html.js
@@ -1,0 +1,72 @@
+const logging = require('../../../../../shared/logging');
+const {createIrreversibleMigration} = require('../../utils');
+const mobiledocLib = require('../../../../lib/mobiledoc');
+const htmlToText = require('html-to-text');
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    logging.info('Starting re-generation of posts html.');
+
+    await knex.transaction(async (trx) => {
+        // get list of posts ids, use .forUpdate to lock rows until the transaction is finished
+        const postIdRows = await knex('posts')
+            .transacting(trx)
+            .forUpdate()
+            .select('id');
+
+        // transform each post individually to avoid dumping all posts into memory and
+        // pushing all queries into the query builder buffer in parallel
+        // https://stackoverflow.com/questions/54105280/how-to-loop-through-multi-line-sql-query-and-use-them-in-knex-transactions
+
+        for (const postIdRow of postIdRows) {
+            const {id} = postIdRow;
+            const [post] = await knex('posts')
+                .transacting(trx)
+                .where({id})
+                .select('mobiledoc', 'html', 'plaintext');
+
+            let mobiledoc;
+
+            try {
+                mobiledoc = JSON.parse(post.mobiledoc || null);
+
+                if (!mobiledoc) {
+                    return logging.warn(`No mobiledoc for ${id}. Skipping.`);
+                }
+            } catch (err) {
+                return logging.warn(`Invalid JSON structure for ${id}. Skipping`);
+            }
+
+            const html = mobiledocLib.mobiledocHtmlRenderer.render(mobiledoc);
+
+            const updatedAttrs = {
+                html: html
+            };
+
+            // NOTE: block comes straight from the Post model
+            // https://github.com/TryGhost/Ghost/blob/4.0.0-alpha.2/core/server/models/post.js#L484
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.0/20-regenerate-posts-html.js
```diff
@@ -0,0 +1,75 @@
...
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    logging.info('Starting re-generation of posts html.');
+
+    await knex.transaction(async (trx) => {
+        // get list of posts ids, use .forUpdate to lock rows until the transaction is finished
+        const postIdRows = await knex('posts')
+            .transacting(trx)
+            .forUpdate()
+            .select('id');
```
tpatel, line 14: You could exclude the ids of post without mobiledoc here, but I guess it's not really worth the trouble because it's a micro-optimization.

core/server/data/migrations/versions/4.0/20-regenerate-posts-html.js
```diff
@@ -0,0 +1,75 @@
...
+
+            const html = mobiledocLib.mobiledocHtmlRenderer.render(mobiledoc);
+
+            const updatedAttrs = {
+                html: html
+            };
+
+            // NOTE: block comes straight from the Post model
+            // https://github.com/TryGhost/Ghost/blob/4.0.0-alpha.2/core/server/models/post.js#L484
+            if (html !== post.html || !post.plaintext) {
```
tpatel, line 47: You're trying to access `post.html` and `post.plaintext` here but it's not there in the select at [line 25](https://github.com/TryGhost/Ghost/pull/12660/files#diff-13dab3d53b700bbe52a3246066a4ba53f513c23d9d3f09cdd54e71101b8c4b73R25). I'd add these columns to the select.

tpatel (changes requested):
I think there are two columns missing in the individual post select

core/server/data/migrations/versions/4.0/20-regenerate-posts-html.js
kevinansfield: Good catch! 🎣 Force-pushed with a fix, thanks!

core/server/data/migrations/versions/4.0/20-regenerate-posts-html.js
kevinansfield: I left it in because the logs for invalid/missing mobiledoc could be useful, missing mobiledoc means an invalid post record that could be indicative of other issues.

tpatel (approved):
Looks great 🙌

core/server/data/migrations/versions/4.0/23-regenerate-posts-html.js
```diff
@@ -0,0 +1,75 @@
...
+            const html = mobiledocLib.mobiledocHtmlRenderer.render(mobiledoc);
+
+            const updatedAttrs = {
+                html: html
+            };
+
+            // NOTE: block comes straight from the Post model
+            // https://github.com/TryGhost/Ghost/blob/4.0.0-alpha.2/core/server/models/post.js#L484
+            if (html !== post.html || !post.plaintext) {
+                const plaintext = htmlToText.fromString(post.html, {
```
kevinansfield, line 48: We should be getting plaintext from the newly generated `html` not the existing `post.html`.

core/server/data/migrations/versions/4.0/23-regenerate-posts-html.js
```diff
@@ -0,0 +1,72 @@
...
+                const plaintext = htmlToText.fromString(html, {
+                    wordwrap: 80,
+                    ignoreImage: true,
+                    hideLinkHrefIfSameAsText: true,
+                    preserveNewlines: true,
+                    returnDomByDefault: true,
+                    uppercaseHeadings: false
+                });
+
+                if (plaintext || plaintext !== post.plaintext) {
```
daniellockyer, line 57: @kevinansfield Should this be `&&`?

kevinansfield: Hmm, the first part of the conditional can be removed I think. Same logic oddness exists in the Post model

PR #12666, opened 2021-02-18
Made addUnique/dropUnique migration utils idempodent

naz:
refs https://github.com/TryGhost/Ghost/pull/12598

- This changeset adds idepmotence to situations where unique constraint has to be dropped or added to the table
- Note '4.0/07-alter-unique-constraint-for-posts-slug.js` was migration that was effected by lack of idempotence

Note to reviewers, didn't find any better way to search for unique constraints then doing raw queries on `TABLE_CONSTRAINTS` in MySQL and `PRAGMA index_list` for SQLite. Let me know if you know any better ways.

### DIFF
--- a/core/server/data/schema/commands.js
+++ b/core/server/data/schema/commands.js
@@ -64,6 +64,46 @@ function dropColumn(tableName, column, transaction) {
 }
 
 /**
+ * Checks if unique index exists in a table over the given columns.
+ *
+ * @param {string} tableName - name of the table to add unique constraint to
+ * @param {string|[string]} columns - column(s) to form unique constraint with
+ * @param {Object} transaction - connnection object containing knex reference
+ * @param {Object} transaction.knex - knex instance
+ */
+async function hasUnique(tableName, columns, transaction) {
+    const knex = (transaction || db.knex);
+    const client = knex.client.config.client;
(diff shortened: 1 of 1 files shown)


allouis (approved):
Might wanna add some logging for the case that we _do_ add/drop the constraint!

naz:
Very good point @allouis. Will add!

core/server/data/schema/commands.js
```diff
@@ -85,10 +131,16 @@ function addUnique(tableName, columns, transaction) {
...
-    });
+async function dropUnique(tableName, columns, transaction) {
+    const hasUniqueConstraint = await hasUnique(tableName, columns, transaction);
+
+    if (hasUniqueConstraint) {
+        return (transaction || db.knex).schema.table(tableName, function (table) {
+            table.dropUnique(columns);
+        });
+    } else {
+        logging.warn(`Constraint for ${columns} does not exists for table ${tableName}`);
```
daniellockyer, line 142: `s/exists/exist/`

daniellockyer (approved):
Same point as @allouis + my small logging message nit, but LGTM 👍🏻

rshbhgrg (approved):

PR #12667, opened 2021-02-18
🐛 Fixed 500 chars of members-only content appearing in excerpt field

kevinansfield:
closes https://github.com/TryGhost/Team/issues/468

- split `extraAttrs.forPost` into separate functions (`postReadingTime`, `postExcerpt`) to allow better ordering in post mapper
- updated `extraAttrs.postExcerpt` to read plaintext from attrs rather than directly from the model
- moved `extraAttrs.postExcerpt` call to be made after the post gating call to ensure the plaintext attr contains the gated contents from which to generate the excerpt

### DIFF
--- a/core/server/api/canary/email-preview.js
+++ b/core/server/api/canary/email-preview.js
@@ -21,7 +21,7 @@ module.exports = {
         ],
         permissions: true,
         query(frame) {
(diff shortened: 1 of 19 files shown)


kevinansfield:
Closing in favour of https://github.com/TryGhost/Ghost/pull/12670 because the approach in this PR had unintended side-effects on API output.

PR #12669, opened 2021-02-18
Renamed the post_meta table to meta

tpatel:
issue https://github.com/TryGhost/Team/issues/452

Work in progress: the tests don't pass yet.


naz:
`meta` naming didn't work out due to conflict with internal naming. A PR with `metadata` naming is coming shortly instead.

PR #12673, opened 2021-02-18
Added migration to remove surrounding <> in email_batches.provider_id

kevinansfield:
refs https://github.com/TryGhost/Team/issues/221#issuecomment-759105424

- Mailgun responds to an email send with a provider id in the format `<x@y.com>` but everywhere else it's used in their API it uses the format `x@y.com`
- updates email batch save to strip the brackets, and migration removes brackets from existing records so we no longer have to add special handling for the stored id any time we use it

### DIFF
--- a/core/server/services/bulk-email/index.js
+++ b/core/server/services/bulk-email/index.js
@@ -168,7 +168,7 @@ module.exports = {
             // update batch success status
             return await emailBatchModel.save({
                 status: 'submitted',
-                provider_id: sendResponse.id
+                provider_id: sendResponse.id.trim().replace(/^<|>$/g, '')
             }, Object.assign({}, knexOptions, {patch: true}));
         } catch (error) {
             // update batch failed status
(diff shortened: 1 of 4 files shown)


core/server/services/bulk-email/index.js
```diff
@@ -168,7 +168,7 @@ module.exports = {
             // update batch success status
             return await emailBatchModel.save({
                 status: 'submitted',
-                provider_id: sendResponse.id
+                provider_id: sendResponse.id.trim().replace(/^<|>$/g)
```
naz, line 171: This `replace` is buggy producing `undefined` on both sides as a result, example from debug session:
```
 sendResponse.id.trim().replace(/^<|>$/g)
'undefined20210219001342.1.2107B24C283F9A12@sandbox1a102ce1e8514fb597612c97d40b08ac.mailgun.orgundefined'
```
Creates faulty records in the db

naz (changes requested):

core/server/services/bulk-email/index.js
tpatel: I guess it's just missing an empty string as the second argument or the replace call

kevinansfield: 🤦🏻‍♂️ thanks, fixed!

tpatel (approved):
Looks good 👍

naz (approved):

PR #12696, opened 2021-02-23
Updated price helper to output well formatted prices

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/472

The current `{{price}}` helper only works with `amount` to convert it into right value but doesn't allow any formatting with currency etc, leaving most of the work to theme.  We want to be able to output well formatted prices. E.g. the API returns 5000 for 5 EUR but we want to output €5.

The updated {{price}} helper can take a plan object or amount, currency and use them to output a well formatted price. It works with JS's built in Intl.NumberFormat behaviour to return output in different formats, also allowing theme devs to override formatting with options.

Examples:

With Plan object => `{{price plan}} → "€5"`
With Plan object and custom number format =>  `{{price plan numberFormat="long"}} → "€5.00"`

### DIFF
--- a/core/frontend/helpers/price.js
+++ b/core/frontend/helpers/price.js
@@ -1,13 +1,79 @@
 // # {{price}} helper
 //
 // Usage: `{{price 2100}}`
+// Usage: `{{price plan}}`
+// Usage: `{{price plan numberFormat="long"}}`
+// Usage: `{{price plan currencyFormat="code"}}`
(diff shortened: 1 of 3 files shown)


rishabhgrg:
> I think this needs to remove , as well - not sure about other characters

@allouis Good point, updated regex for `,` but also added `minimum` and `maximum` fraction digits to 0 in case we are calculating currency symbol to completely avoid any other case as we are only interested in the symbol.

PR #12697, opened 2021-02-23
Added primary keys to the brute table

tpatel:
issue https://github.com/TryGhost/Team/issues/267

- Tables without primary keys cause issues with mysql replication and clustering

### DIFF
--- a/core/server/data/migrations/versions/4.0/23-add-primary-key-brute-migrations-lock.js
+++ b/core/server/data/migrations/versions/4.0/23-add-primary-key-brute-migrations-lock.js
@@ -1,0 +1,9 @@
+const logging = require('../../../../../shared/logging');
+const {createIrreversibleMigration} = require('../../utils');
+const {addPrimaryKey} = require('../../../schema/commands');
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    logging.info('Adding a primary key for the brute table');
+
+    await addPrimaryKey('brute', 'key', knex);
+});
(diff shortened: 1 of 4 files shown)


core/server/data/migrations/versions/4.0/23-add-primary-key-brute-migrations-lock.js
```diff
@@ -0,0 +1,11 @@
+const logging = require('../../../../../shared/logging');
+const {createIrreversibleMigration} = require('../../utils');
+const {addPrimaryKey} = require('../../../schema/commands');
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    logging.info('Adding a primary key for the brute and migrations_lock tables');
+
+    await addPrimaryKey('brute', 'key', knex);
+
+    await addPrimaryKey('migrations_lock', 'lock_key', knex);
```
naz, line 10: The `migrations_lock` table is not "owned" by Ghost. It is created and managed by the knex-migrator. Imo, this migration should be happening inside of the knex-migrator's [migrations](https://github.com/TryGhost/knex-migrator/tree/fdd58cb911b39103029ad8e977384ee0029ca97a/migrations).

tpatel: good point 💯

tpatel: it has been removed ✅

naz (approved):
Would split this PR into 2 separate commits: (1) the commands changes and (2) the migration itself. Also the change in integrity_spec seems to be missing :wink: Otherwise works well!

daniellockyer (approved):
Agree with Naz's points, but LGTM otherwise 👍🏻

allouis (approved):

PR #12704, opened 2021-02-25
Remove orphaned Stripe data from SQLite

allouis:
refs https://github.com/TryGhost/Team/issues/476

The migration to populate the paid subscription events relies on the Stripe customers & subscriptions data being associated with a member. Ghost running with SQLite does not clean up orphaned rows, and so we must remove them _before_ running the event population migration.


core/server/data/migrations/versions/4.0/14-remove-orphaned-stripe-records.js
```diff
@@ -0,0 +1,35 @@
+const {createIrreversibleMigration} = require('../../utils');
+const logging = require('../../../../../shared/logging');
+
+module.exports = createIrreversibleMigration(async function up(connection) {
+    if (connection.client.config.client === 'mysql') {
+        logging.info('Skipping removal of orphaned stripe records for MySQL');
+        return;
+    }
+
+    logging.info('Removed orphaned rows from members_stripe_customers');
```
tpatel, line 10: nitpick: you might want to say `removing` instead of `removed`, so that if an errrr happens after the log, users understand the action hasn't finished successfully yet.

tpatel (approved):
Looks great 🙌

daniellockyer (approved):
Agree with Thibaut, but LGTM!

rshbhgrg (approved):

core/server/data/migrations/versions/4.0/14-remove-orphaned-stripe-records.js
allouis: Good call! :relaxed:

Issue #11084, opened 2019-09-06
Feature: headers.yaml & default security headers


ErisDS:
In Ghost we have the ability to set a content type header for any custom route in routes.yaml, but otherwise, headers are not customisable.

There are many different headers that a user might want to set, the most common use case being to improve security. We should also provide a better default set of headers, as part of this.

## Ref material

- Netlify \_headers file docs: https://www.netlify.com/docs/headers-and-basic-auth/
- Ghost docs \_headers file example: https://github.com/TryGhost/docs/commit/8af94d2c92d455738b940150cba56bf741ed9568
- Statamic content type: https://docs.statamic.com/routing#content-types
- Ghost content type: https://docs.ghost.org/concepts/routing/#content_type
- Ghost CLI default headers: https://github.com/TryGhost/Ghost-CLI/blob/975049cd850f0cdf33555c73a2b675fe7b6e742e/extensions/nginx/templates/ssl-params.conf#L11

## Spec

We will provide the ability to upload a `headers.yaml` file. The file must contain valid YAML.

Headers will apply to the web application of Ghost only (same as redirects).

We will allow for setting headers per route with route matching exactly like [Netlify _headers](https://github.com/TryGhost/docs/commit/8af94d2c92d455738b940150cba56bf741ed9568)

## Format

The top level YAML keys represent rules, with wildcard matches.
Nested key value pairs should represent header names and values. Case should not matter.

```
/*:
  Header-Name: Value

/something/*:
   header-name: value
```

Real world example (taken from Ghost docs):

```
/*:
  Referrer-Policy: no-referrer-when-downgrade
  Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
  X-Content-Type-Options: nosniff
  X-Frame-Options: SAMEORIGIN
  X-Xss-Protection:	1; mode=block
  Feature-Policy: accelerometer 'none'; camera 'none'; geolocation 'none'; gyroscope 'none'; magnetometer 'none'; microphone 'none'; payment 'none'; usb 'none'

```

### Blacklisted / Ignored Headers

There may be some headers that Ghost sets which should be ignored if they are overridden via headers.yaml, to prevent any conflicts or issues on Ghost(Pro). The first that comes to mind is Cache-Control & Proxy headers.

- Cache-control
- X-Forwarded-*
- X-Ghost-*
- Location? 
- ???

### X-Powered-By

[...]

PR #12707, opened 2021-02-28
Feature/#11084

dmitrymarokhonov:
https://github.com/TryGhost/Ghost/issues/11084

Tasks
 Finalise header blacklist
 Decide on default headers for each area of Ghost
 Implement headers on admin & api
 Fully remove x-powered-by
 Implement frontend defaults & ability to override via headers.yaml file
 Add UI for uploading headers.yaml


ErisDS:
Doesn't look like this belongs here

PR #12702, opened 2021-02-24
Added a foreign key in migration utils

tpatel:
blocks https://github.com/TryGhost/Team/issues/476

- This makes it easy to add/remove foreign key in both mysql and sqlite

@naz this is a work in progress and hasn't been tested yet, but I wanted to make it available to you as soon as possible.

PR #12713, opened 2021-03-01
Added "on delete cascade" to several foreign keys in sqlite

tpatel:
issue https://github.com/TryGhost/Team/issues/476
blocked by https://github.com/TryGhost/Ghost/pull/12702

### DIFF
--- a/core/server/data/migrations/versions/4.0/26-add-cascade-on-delete.js
+++ b/core/server/data/migrations/versions/4.0/26-add-cascade-on-delete.js
@@ -1,0 +1,75 @@
+const logging = require('../../../../../shared/logging');
+const {createIrreversibleMigration} = require('../../utils');
+const {addForeign, dropForeign} = require('../../../schema/commands');
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    if (knex.client.config.client !== 'sqlite3') {
+        return logging.warn('Skipping adding "on delete cascade" - database is not SQLite3');
+    }
+
+    logging.info('Adding on delete cascade for members_labels, members_stripe_customers and members_stripe_customers_subscriptions');
+
+    await dropForeign({
+        fromTable: 'members_labels',
+        fromColumn: 'member_id',
+        toTable: 'members',
+        toColumn: 'id',
+        transaction: knex
+    });
+    await addForeign({
+        fromTable: 'members_labels',
+        fromColumn: 'member_id',
+        toTable: 'members',
+        toColumn: 'id',
+        cascade: true,
+        transaction: knex
+    });
+
+    await dropForeign({
+        fromTable: 'members_labels',
+        fromColumn: 'label_id',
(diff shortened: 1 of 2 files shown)


naz (approved):
Awesome work man!

core/server/data/migrations/versions/4.0/26-add-cascade-on-delete.js
```diff
@@ -0,0 +1,75 @@
+const logging = require('../../../../../shared/logging');
+const {createIrreversibleMigration} = require('../../utils');
+const {addForeign, dropForeign} = require('../../../schema/commands');
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    if (knex.client.config.client !== 'sqlite3') {
+        return logging.warn('Skipping  - database is not SQLite3');
```
daniellockyer, line 7: Can you make this something like "Skipping adding 'on cascade delete' - database is not SQLite3"? Just so it's even more explicit what we're early returning from 🙂

tpatel: good point! done ✅

daniellockyer (approved):

core/server/data/migrations/versions/4.0/26-add-cascade-on-delete.js
```diff
@@ -0,0 +1,75 @@
...
+        toTable: 'members',
+        toColumn: 'id',
+        transaction: knex
+    });
+    await addForeign({
+        fromTable: 'members_labels',
+        fromColumn: 'member_id',
+        toTable: 'members',
+        toColumn: 'id',
+        cascade: true,
```
allouis, line 24: Not a big deal, but considering `CASCADE` can be used in multiple contexts, it might make sense to have this option named `cascadeDelete` like in the schema. Doesn't matter though if you cba to change it

core/server/data/schema/commands.js
```diff
@@ -199,13 +199,29 @@ async function addForeign({fromTable, fromColumn, toTable, toColumn, cascade = f
 
     if (!hasForeignKey) {
         logging.info(`Adding foreign key for: ${fromColumn} in ${fromTable} to ${toColumn} in ${toTable}`);
-        return (transaction || db.knex).schema.table(fromTable, function (table) {
+
+        //disable and re-enable foreign key checks on sqlite because of https://github.com/knex/knex/issues/4155
```
allouis, line 203: Nice find!

allouis (approved):

daniellockyer (approved):

core/server/data/migrations/versions/4.0/26-add-cascade-on-delete.js
tpatel: just learned "cba" 😂  also I've changed it, thanks!

PR #12724, opened 2021-03-03
Added a foreign key from webhooks to integrations

tpatel:
issue https://github.com/TryGhost/Team/issues/477

Note: I had to change the position of the `webhooks` schema definition to make sure it comes after the `integrations` table. So that it initialise in the right order when creating a new database.


core/server/data/migrations/versions/4.0/28-add-webhook-intergrations-foreign-key.js
```diff
@@ -0,0 +1,23 @@
...
+        toTable: 'integrations',
+        toColumn: 'id',
+        transaction: knex
+    });
+    await addForeign({
+        fromTable: 'webhooks',
+        fromColumn: 'integration_id',
+        toTable: 'integrations',
+        toColumn: 'id',
+        cascadeDelete: true,
```
naz, line 20: @tpatel I think we don't need cascading here :thinking:  the migration was meant to only introduce the foreign key as it did not exist before :wink:

daniellockyer: If so, we don't need the `dropForeign` either right?

naz: yes, exactly!

naz: having a little more thought about it... `ON DELETE CASCADE` (or as we call it in schema `cascadeDelete`) might be ok here because it would prevent us from having accidental orphaned webhooks when the integration is deleted!

naz: I'm still not sure `dropForeign` would serve any purpose here except of being a very defensive way of making sure there's no foreign key already. would like to hear why @tpatel added it in the first place :)

naz:
@tpatel my understanding is that this PR is still missing the change in the `schema.js` file marking `integration_id` with `references` and possibly `cascadeDelete` properties:wink:

core/server/data/migrations/versions/4.0/28-add-webhook-intergrations-foreign-key.js
daniellockyer: Actually, we should clarify the naming of `cascadeDelete` or `cascade` - it was `cascade` in the previous PR

daniellockyer: What is the difference between `cascade` in the last PR vs `cascadeDelete` in this one? 😬

tpatel: yeah exactly I thought about an edge case where there would already be a foreign key, but it's not relevant here so I'm gonna remove the `dropForeign`.
The `ON DELETE CASCADE` seems relevant to avoid orphaned webhooks as you said, even more because there was an issue with orphaned webhooks first :)

naz (approved):
@tpatel from my perspective this is good to go!

rshbhgrg (approved):
Looks good

daniellockyer (approved):

PR #12733, opened 2021-03-05
Updated default format for date helper to locale-based date string

matthanley:
refs https://github.com/TryGhost/Casper/pull/741
closes https://github.com/TryGhost/Team/issues/524

### DIFF
--- a/core/frontend/helpers/date.js
+++ b/core/frontend/helpers/date.js
@@ -7,40 +7,37 @@ const {SafeString, themeI18n} = require('../services/proxy');
 const moment = require('moment-timezone');
 
 module.exports = function (date, options) {
-    let timezone;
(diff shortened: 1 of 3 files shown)


core/frontend/helpers/date.js
```diff
@@ -35,7 +35,7 @@ module.exports = function (date, options) {
     // Documentation: http://momentjs.com/docs/#/i18n/
     // Locales: https://github.com/moment/moment/tree/develop/locale
     const dateMoment = moment(date);
-    dateMoment.locale(themeI18n.locale());
+    dateMoment.locale(options.data.site.locale || themeI18n.locale());
```
matthanley, line 38: @ErisDS I'm not sure about this line, but it's the only way I could figure out how to get the tests to work with different locales

PR #12735, opened 2021-03-05
fix typo

makeitworks:
i think this is a typo


ErisDS:
It is, but the return value is not used. Would be better to change this to just `return;`

ErisDS:
I made this change here: https://github.com/TryGhost/Ghost/commit/c12edf4bbd956a5e6a9caf7ffa1f51ee1f3b2479

PR #12741, opened 2021-03-09
Fixed mobiledoc errors when upgrading from v1/v2 to v4

kevinansfield:
no issue

Upgrading from v1 or v2 can result in successful upgrades but with mobiledoc errors showing in the logs:

```
NAME: InternalServerError
MESSAGE: Mobiledoc card 'card-markdown' not found.
```

The errors do not signify a problem as long as the 4.0 migrations run because those rename the deprecated card before re-rendering.

- `@tryghost/kg-default-cards` dropped support for `card-markdown` cards. 4.0 migrations handled this by renaming all `card-markdown` cards to `markdown` before re-generating any content
- 2.0 and 3.0 also had migrations that re-generated content but they are run before the 4.0 card rename migration meaning that the mobiledoc renderer sees cards that it doesn't know about. The behaviour for unknown cards is to log an error and skip rendering of that card
- by NOOPing the 2.0 and 3.0 migrations we eliminate the incompatibility errors and reduce the amount of processing the upgrade needs to perform


kevinansfield:
There are 2 alternative and possibly preferable approaches to this:

1. Do nothing - the errors aren't fatal so migrations will continue and when the 4.0 migration is run everything is resolved by the card rename and re-render
2. NOOP the 2.0 and 3.0 migrations - they are repeating the same html re-generation that the 4.0 migration will do anyway

daniellockyer (approved):

PR #12742, opened 2021-03-09
🐛 Fixed a foreign key to point to the right column

tpatel:
pr https://github.com/TryGhost/Ghost/pull/12713

This is a fix to point the foreign key to the right column. I'm deleting both foreign keys to make sure the db is in a clean state and the foreign key is `ON DELETE CASCADE`.


daniellockyer (approved):

allouis:
Do we need the second migration? I would've thought we can just update the broken one and ask people to blow away their DB locally

allouis (approved):

matthanley (approved):

daniellockyer:
@allouis We discussed it and figured now there has been a RC, with a (small) chance that it's being used somewhere, we'd need a migration to correct it 🙂

PR #12761, opened 2021-03-14
Fix comped members content gating

kevinansfield:
no issue

Comped members were not able to view paid-member content because content gating was only looking for `member.status === 'paid'` which doesn't take into consideration members on a complimentary plan.

- added front-end acceptance tests for member access to posts
- updated content-gating check to take comped members into consideration

### DIFF
--- a/core/server/services/members/content-gating.js
+++ b/core/server/services/members/content-gating.js
@@ -23,7 +23,7 @@ function checkPostAccess(post, member) {
         return PERMIT_ACCESS;
     }
 
-    if (post.visibility === 'paid' && member.status === 'paid') {
+    if (post.visibility === 'paid' && (member.status === 'paid' || member.status === 'comped' || member.comped)) {
         return PERMIT_ACCESS;
     }
 
--- a/test/api-acceptance/admin/members_spec.js
+++ b/test/api-acceptance/admin/members_spec.js
@@ -36,7 +36,7 @@ describe('Members API', function () {
         const jsonResponse = res.body;
         should.exist(jsonResponse);
         should.exist(jsonResponse.members);
-        jsonResponse.members.should.have.length(4);
+        jsonResponse.members.should.have.length(5);
         localUtils.API.checkResponse(jsonResponse.members[0], 'member', 'subscriptions');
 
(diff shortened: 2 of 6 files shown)


core/server/services/members/content-gating.js
```diff
@@ -23,7 +23,7 @@ function checkPostAccess(post, member) {
         return PERMIT_ACCESS;
     }
 
-    if (post.visibility === 'paid' && member.status === 'paid') {
+    if (post.visibility === 'paid' && (member.status === 'paid' || member.status === 'comped' || member.comped)) {
```
kevinansfield, line 26: @allouis, @rishabhgrg this isn't working. If I `console.log(member)` at this point the comped member always has `.status = 'free'` and no `comped` property .

naz: The problem was in test fixtures. There were missing initialization steps for member's subscription/customer relations. We'll need to make these fixtures more straight forward to use. Right now it's a labyrinth to navigate when adding a single new record to the fixtures :grimacing:

rishabhgrg: Nice catch, minor thing that we should update the check here to use `!== free` instead of checking both comp and paid like rest of the codebase

naz: IMO, current approach - using "allowlist" of statuses is a better from security standpoint. It's much better to have bugs like current one where we don't expose sensitive data instead of accidentally exposing it to unwanted actors. 

If we are concerned with the logic consistency in other places that apply similar paid/free/comped filtering, like [email recipient list here](https://github.com/TryGhost/Ghost/blob/e87ded8d7206a2ca3ed3adf989b606df1e380c4d/core/server/services/mega/mega.js#L115-L117), I'd propose changing logic to one like in current PR, e.g: `status:['paid','comped']` ([NQL does support](https://github.com/NexesJS/NQL-Lang/blob/85ab577fe0cf7ff2bfaf49d24137d5222287655a/test/parser.test.js#L82-L90) `IN` style filtering).

naz (approved):
The style of using "allowlist" to detect content accessibility can be done in the cleanup phase. This implementation does the job for now :+1:

PR #12764, opened 2021-03-15
Updated @price template data to work with price helper

allouis:
no-issue

The price helper requires an object with amount & currency properties to
work correctly. This updates the `@price` data object to expose these.

In order to maintain backward compatibility with using the `@price` data
as primitive number values, we add a `valueOf` method which returns the
legacy dollar amount value.

This means you can use `{{price @price.monthly}}` OR
`{{@price.monthly}}` - the second of which will output the dollar
amount.


allouis:
To expand on what this does, suppose we have a site with a 5 USD a month and a 50 USD a month plan.

Currently, `@price.yearly` and `@price.monthly` contain a number (50 and 5 respectively), and `@price.currency` contains an ISO currency code (in this case "usd")

This means that in order to use them in conjunction with the price helper we would do something like:

```html
The monthly plan costs {{price currency=@price.currency}}{{@price.monthly}} a month.
The yearly plan costs {{price currency=@price.currency}}{{@price.yearly}} a year.
```

And this would product output like:

```html
The monthly plan costs $5 a month.
The yearly plan costs $50 a year.
```

This is different to how the `{{price}}` helper is used when displaying which plan a member is subscribed to (e.g. `{{price plan}}`


What this change does is update the `@price.monthly` and `@price.yearly` data to instead be an object with a `currency` property of `'usd'` and an `amount` property of either `500` (monthly) or `5000` (yearly) - which can be used with the `{{price}}` helper like so: 

```html
The monthly plan costs {{price @price.monthly}} a month.
The yearly plan costs {{price @price.yearly}} a year.
```

Which would also product output like:
```html
The monthly plan costs $5 a month.
The yearly plan costs $50 a year.
```

We add the `valueOf` method to preserve the existing functionality, so that when `{{@price.monthly}}` or `{{@price.yearly}}` are used independently, as in the first example - then they will work as they do currently.

matthanley:
refs https://github.com/TryGhost/Team/issues/545

test/utils/fixtures/themes/price-data-test-theme/index.hbs
```diff
@@ -0,0 +1,6 @@
+<p class="number">
```
allouis, line 1: This is the actual important theme file

allouis:
@rishabhgrg Asking for your review on this as you worked on the `{{price}}` helper - so I'm hoping you can check if I've missed anything

rshbhgrg (approved):

PR #12777, opened 2021-03-17
Fixed typo in migration to add backupContent permission

allouis:
refs https://github.com/TryGhost/Team/issues/553

A typo in a permission migration was introduced in https://github.com/TryGhost/Ghost/commit/79c3709f

This fixes the migration, and adds a new migration to fix affected sites

### DIFF
--- a/core/server/data/migrations/versions/4.1/01-fix-backup-content-permission-typo.js
+++ b/core/server/data/migrations/versions/4.1/01-fix-backup-content-permission-typo.js
@@ -1,0 +1,15 @@
+const logging = require('../../../../../shared/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(async function up(knex) {
+    const typoedPermisson = await knex.select('*').from('permissions').where('action_type', 'backupContect').first();
+
+    if (!typoedPermisson) {
+        return logging.warn('Not updating permissions, no typo found');
+    }
+
+    logging.info('Updating permissions, fixing typo by renaming "backupContect" to "backupContent"');
+    await knex('permissions').update('action_type', 'backupContent').where('action_type', 'backupContect');
+}, function down() {
+    // noop
+});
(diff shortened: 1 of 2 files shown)


tpatel:
@allouis will this migration run for folks that have already migrated to Ghost 4.0?

daniellockyer:
@tpatel It will because `knex-migrator` will diff between the migrations and detect this one won't have run yet. In theory it should be ok but it's still a little odd to do 😬

tpatel:
@daniellockyer does that mean you're suggesting to rename/move this migration to be in the 4.x folder ?

We could have this 3.x migration merged in the 3.x branch?

daniellockyer:
@tpatel We can't do that because if a site which is currently on 3.42 updates to 3.43, and then up to 4.0, the migrations won't line up and it'll complain __unless__ any that appear in 3.x appear in 4.x too.

tpatel:
Thanks for the explanation, I get it now 🙏

tpatel (approved):
Looks great 🙌

rshbhgrg (approved):

matthanley (approved):

core/server/data/migrations/versions/4.1/01-fix-backup-content-permission-typo.js
```diff
@@ -0,0 +1,15 @@
+const logging = require('../../../../../shared/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(async function up(knex) {
+    const typoedPermisson = await knex.select('*').from('permissions').where('action_type', 'backupContect').first();
```
daniellockyer, line 5: You typoed `Permission` in `typoedPermisson` 😝

allouis: :joy: fucks sake

daniellockyer (approved):

PR #12780, opened 2021-03-17
🐛 Replace __GHOST_URL__ imageUrl in sitemap-posts

alexcberk:
Ensures we get the actual ghost url instead of the placeholder string in the sitemap.

- [X] There's a clear use-case for this code change


naz:
Initially reported in https://forum.ghost.org/t/images-not-showing-url-says-ghost-url-content/20656/10

kevinansfield:
Thanks for taking a look at this @acberk! 🙌  You were on the right track here but we had a larger refactor in this area already planned and the fix was rolled into that in https://github.com/TryGhost/Ghost/pull/12787. It will go out with the next Ghost release 🙂

PR #12782, opened 2021-03-18
Export import test suite refactor

naz:
Temporary PR to run the checks on CI


naz:
Will fix this next week when I'm back from the long weekend.

PR #12784, opened 2021-03-18
🐛  Fixed unable to delete member

rshbhgrg:
closes https://github.com/TryGhost/Team/issues/557

After updating to v4 on SQLite from v3, we were unable to delete members due to a foreign key mismatch error. This is because the migrations which recreates the tables for `members_stripe_customers_subscriptions` and `members_stripe_customers` doesn't add back the unique constraint which is needed by FKs for reference.

The migration creates the missing UNIQUE constraints on the tables.


core/server/data/migrations/versions/4.1/01-add-unique-constraint-for-member-stripe-tables.js
```diff
@@ -0,0 +1,14 @@
+const logging = require('../../../../../shared/logging');
+const {createIrreversibleMigration} = require('../../utils');
+const {addUnique} = require('../../../schema/commands');
+
+module.exports = createIrreversibleMigration(async (knex) => {
```
allouis, line 5: We don't usually create irreversible migrations outside of a major, I think this should just have a `noop` for the down migration :thinking:

allouis (commented):
Smol comment

core/server/data/migrations/versions/4.1/01-add-unique-constraint-for-member-stripe-tables.js
rishabhgrg: @allouis Yeah good point, have changed it to `transactionalMigration` now with noop for down migration

daniellockyer (approved):

allouis (approved):

tpatel (approved):
looks great 👍

PR #12773, opened 2021-03-17
🐛 Explicitly use single quotes in `email_batches` SQL migration.

toddbirchard:
## Description

Self-hosted Ghost instances with MySQL databases fail to perform necessary migrations when upgrading from `3.42.2` -> `4.0.1`. The culprit is a SQL query found in [21-sanitize-email-batches-provider-id.js](https://github.com/TryGhost/Ghost/blob/08e1268aed862a062df7dea721e1948634b53cc8/core/server/data/migrations/versions/4.0/21-sanitize-email-batches-provider-id.js#L7).

MySQL interprets values enclosed in double-quote as columns, thus resulting in an invalid query. Reversing the usage of double quotes and single quotes in the **email_batches** query alleviates this:

```sql
UPDATE email_batches SET provider_id = REPLACE(REPLACE(provider_id, '<', ''), '>', '');
```

The malformed query is likely the result of **eslint** and **jshint** enforcing single outer quotes. For the purpose of performing this migration, these rules are temporarily disabled in **21-sanitize-email-batches-provider-id.js** to pass lint tests.

### Stack trace

```shell
[2021-03-17 00:02:16] WARN Database state requires migration.
[2021-03-17 00:02:16] INFO Creating database backup
[2021-03-17 00:02:25] INFO Database backup written to: /var/www/ghost/content/data/my.ghost.2021-03-17-00-02-16.json
[2021-03-17 00:02:25] INFO Sanitizing provider_id values in email_batches
[2021-03-17 00:02:25] INFO Sanitizing provider_id values in email_batches
[2021-03-17 00:02:25] ERROR UPDATE email_batches SET provider_id = REPLACE(REPLACE(provider_id, "<", ""), ">", "") - ER_BAD_FIELD_ERROR: Unknown column '<' in 'field list'

UPDATE email_batches SET provider_id = REPLACE(REPLACE(provider_id, "<", ""), ">", "") - ER_BAD_FIELD_ERROR: Unknown column '<' in 'field list'

{"config":{"irreversible":true},"name":"21-sanitize-email-batches-provider-id.js"}
"Error occurred while executing the following migration: 21-sanitize-email-batches-provider-id.js"

Error ID:
    300

Error Code: 
    ER_BAD_FIELD_ERROR

----------------------------------------

MigrationScriptError: UPDATE email_batches SET provider_id = REPLACE(REPLACE(provider_id, "<", ""), ">", "") - ER_BAD_FIELD_ERROR: Unknown column '<' in 'field list'
[...]

PR #12794, opened 2021-03-19
Update migration to use standard SQL strings instead of double quoted column names

cthornton:
Double quoted strings are used to indicate column names in standard SQL, may cause problems for some databases, and MySQL using strict settings. For example, the migration caused issues with hosted MySQL in digital ocean.

Also could someone take a look at the test failure? I'm not sure why this failure is happening

### DIFF
--- a/core/server/data/migrations/versions/4.0/21-sanitize-email-batches-provider-id.js
+++ b/core/server/data/migrations/versions/4.0/21-sanitize-email-batches-provider-id.js
@@ -4,5 +4,5 @@ const {createIrreversibleMigration} = require('../../utils');
 module.exports = createIrreversibleMigration(async (knex) => {
     logging.info('Sanitizing provider_id values in email_batches');
 
-    await knex.raw('UPDATE email_batches SET provider_id = REPLACE(REPLACE(provider_id, "<", ""), ">", "")');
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.0/21-sanitize-email-batches-provider-id.js
```diff
@@ -4,5 +4,5 @@ const {createIrreversibleMigration} = require('../../utils');
 module.exports = createIrreversibleMigration(async (knex) => {
     logging.info('Sanitizing provider_id values in email_batches');
 
-    await knex.raw('UPDATE email_batches SET provider_id = REPLACE(REPLACE(provider_id, "<", ""), ">", "")');
+    await knex.raw('UPDATE email_batches SET provider_id = REPLACE(REPLACE(provider_id, ?, ?), ?, ?)', '<', '', '<', '');
```
cthornton, line 7: The second `<` should be a `>`

daniellockyer:
Hey @cthornton! We recently received [another PR](https://github.com/TryGhost/Ghost/pull/12773) about this, but we have decided to update our documentation about the use of `ANSI` and `ANSI_QUOTES` SQL modes - https://ghost.org/docs/faq/misconfigured-mysql-database/#error-er_bad_field_error. There is more context in the other PR if you'd like to read more 🙂

PR #12797, opened 2021-03-20
Feature: Allow themes to have any configuration properties in package…

Doogiemuc:
….json

Provide a way so that theme authors can make their themes configurable.
For example: A theme might offer the possibility to optionally show a tag cloud. 
There should be a way so that the theme user can easily configure this.

And actually there already would be such a way. It's just simple forbidden in the current code. There already is a `config` section in the themes package.json  But the code filters and only allows the attributes `posts_per_page` [as documented](https://ghost.org/docs/themes/helpers/config/) and `image_sizes`. Themes can directly access these variables as `{{@config.posts_per_page}}

Why restrict this to just these two variables. Allow any variables for theme authors.

- [ ] There's a clear use-case for this code change
- [ ] The build will pass (run `yarn test` and `yarn lint`)

### DIFF
--- a/core/frontend/services/themes/config/index.js
+++ b/core/frontend/services/themes/config/index.js
@@ -1,12 +1,12 @@
 const _ = require('lodash');
 const defaultConfig = require('./defaults');
-const allowedKeys = ['posts_per_page', 'image_sizes'];
+//const allowedKeys = ['posts_per_page', 'image_sizes'];
 
 module.exports.create = function configLoader(packageJson) {
     let config = _.cloneDeep(defaultConfig);
 
     if (packageJson && Object.prototype.hasOwnProperty.call(packageJson, 'config')) {
-        config = _.assign(config, _.pick(packageJson.config, allowedKeys));
+        //config = _.assign(config, _.pick(packageJson.config, allowedKeys));
     }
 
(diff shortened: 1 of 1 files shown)


Doogiemuc:
Some more detailed description of the use case.

A theme author creates a large and sophisticated theme for ghost. The theme shall have some parts that are configurable. For example it shall be shown if some "widget" is shown on the page or _removed_ completely.    The important part is the "remove" not just "hidden"! Meaning: The removed parts shall be completely removed from the generated HTML that the static site generator generates. 

Example: The theme offers two different navigation bars. One at the top or alternatively a sidebar at the left. Only one of those navigation bars is used. It should be configurable which one to include. The other one is not just "hidden". Its completely not rendered in the generated HTML source.

ErisDS:
Hi @Doogiemuc,

Thanks for the PR, genuinely appreciated. However in this case, the config options are deliberately limited to prevent this getting away from us until we figure out the best approach for everyone.

We are actively looking into ways to allow more options for themes, the beginnings of which you can already see in Ghost 4.0. However, what we know already is this won't be done through config.

In the meantime, feel free to reach out and chat to us about your use cases on the forum :)

PR #12837, opened 2021-04-05
Added migration to fix incorrect mrr events

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/595

Due to a bug in mrr_delta calculation, we ended up reducing the MRR delta by twice the original amount when a subscription goes from active to canceled and storing it in `members_paid_subscription_events` table, which is used to show the MRR chart on Dashboard. The way we identify the incorrect events in the table which got the double negative value is by checking if they match certain criteria - Both `from_plan` and `to_plan` have same value as this is the only case this bug always happens in since a subscription changes status on same plan.

This migration halves the `mrr_delta` for incorrect events to restore the correct MRR change for the site.

### DIFF
--- a/core/server/data/migrations/versions/4.2/01-fix-incorrect-mrr-delta-events.js
+++ b/core/server/data/migrations/versions/4.2/01-fix-incorrect-mrr-delta-events.js
@@ -1,0 +1,34 @@
+const {createTransactionalMigration} = require('../../utils');
+const logging = require('../../../../../shared/logging');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Fixing incorrect mrr_delta in members_paid_subscription_events table');
+        const allIncorrectEvents = await knex
+            .select('id', 'mrr_delta')
+            .from('members_paid_subscription_events')
+            .where('from_plan', '=', knex.ref('to_plan'))
+            .where('mrr_delta', '<', 0);
+
+        const correctedEvents = allIncorrectEvents.filter(({mrr_delta: mrrDelta}) => {
+            return (mrrDelta % 2 === 0);
+        }).map(({id, mrr_delta: mrrDelta}) => {
+            return {
+                id,
+                mrrDelta: mrrDelta / 2
+            };
+        });
+
+        logging.info(`Updating mrr_delta for ${correctedEvents.length} events`);
+        for (const event of correctedEvents) {
+            await knex('members_paid_subscription_events')
+                .where({id: event.id})
+                .update({
+                    mrr_delta: event.mrrDelta
+                });
+        }
+    },
+    async function down() {
+        // noop
+    }
+);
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.2/01-fix-incorrect-mrr-delta-events.js
```diff
@@ -0,0 +1,34 @@
...
+        for (const event of correctedEvents) {
+            await knex('members_paid_subscription_events')
+                .where({id: event.id})
+                .update({
+                    mrr_delta: event.mrrDelta
+                });
+        }
+    },
+    async function down() {
+        // noop
```
rishabhgrg, line 32: @allouis I was wondering if a noop is correct in this case or we should look at reverting the above calculation for the `down` migration here 🤔 IMO noop is better because we don't want to (un)fix the mrr calculation, but not feeling 100% about the right thing

allouis:
Is it possible for a row to meet the first criteria (from_plan is the same as to_plan) and it not meet the second two?

rishabhgrg:
@allouis 🤔 We probably could, its unlikely there will be cases which match the first and not the other 2, the other 2 are more of a safety net. Though yeah the UPDATE query will much better if we just consider the first

rishabhgrg:
@allouis Had another 🤔 session and updated the migration to directly update based on `from_plan` = `to_plan` criteria as the others are redundant

daniellockyer:
@rishabhgrg I've noticed the following:

```
> select mrr_delta from members_paid_subscription_events where from_plan=to_plan limit 5;
+-----------+
| mrr_delta |
+-----------+
|     -1000 |
|     -1000 |
|     -1000 |
|      -916 |
|     -1000 |
+-----------+
5 rows in set (0.00 sec)

> select mrr_delta/2 from members_paid_subscription_events where from_plan=to_plan limit 5;
+-------------+
| mrr_delta/2 |
+-------------+
|   -500.0000 |
|   -500.0000 |
|   -500.0000 |
|   -458.0000 |
|   -500.0000 |
+-------------+
5 rows in set (0.00 sec)

> select round(mrr_delta/2) from members_paid_subscription_events where from_plan=to_plan limit 5;
+--------------------+
| round(mrr_delta/2) |
+--------------------+
|               -500 |
|               -500 |
|               -500 |
|               -458 |
|               -500 |
+--------------------+
```

Unless we do a `round`, we're storing the values as floats. Does this matter? 🤔

rishabhgrg:
@daniellockyer 🤔 I don't think it matters because it seems when the UPDATE sets it back its converted to INTEGER as its working properly. But, its probably on the safer side to introduce `ROUND` to the SET so i'll update that

daniellockyer (approved):

core/server/data/migrations/versions/4.2/01-fix-incorrect-mrr-delta-events.js
allouis: We need a `down` migration for sure, because otherwise a rollback will mean that running the migration again will break the values!

allouis (changes requested):
Looks good, just missing the `down` migration :)

allouis (approved):

tpatel (commented):
The code looks great, but I still have to run it locally 😬 doing my best to test it quickly

tpatel (approved):
Looks good to me, well done 🙌

PR #12844, opened 2021-04-06
Added `products` and `members_products` tables

allouis:

- Add the `products` table
- Add the `members_products` table
- Uses `sort_order` on the `members_products` table to follow the same convention in `members_labels`
- Populates the `products` table with a single product, using the name from the `stripe_product_name` setting
- Populates the `members_products` table with relations based on the `status` column of the members table


rshbhgrg (approved):

core/server/data/migrations/versions/4.3/01-add-products-table.js
```diff
@@ -0,0 +1,9 @@
+const addTable = require('../../utils').addTable;
```
daniellockyer, line 1: Not even worth mentioning but is there a reason you don't use object destructuring here? 😝

allouis: I couldn't remember what the function was called so I did `require('../../utils').` and let autocomplete show me the options :joy:

daniellockyer: Hahahaha 🤣

daniellockyer (approved):

core/server/data/migrations/versions/4.3/04-attach-members-to-product.js
```diff
@@ -0,0 +1,48 @@
...
+        logging.info(`Attaching product ${product.name} to ${membersWithProduct.length} members`);
+        const memberProductRelations = membersWithProduct.map((member) => {
+            return {
+                id: ObjectID.generate(),
+                member_id: member.id,
+                product_id: product.id
+            };
+        });
+
+        const chunkSize = 333;
```
tpatel, line 37: Not blocking: it could be nice to add the reason why it's 333 in a comment (I think I've seen such a comment in a few other migrations).

tpatel (approved):
Looks great 🙌

core/server/data/migrations/versions/4.3/01-add-products-table.js
allouis: Switched this out for destructuring when I added ze comment FYI!

core/server/data/migrations/versions/4.3/04-attach-members-to-product.js
allouis: Done :)

PR #12858, opened 2021-04-07
Added `stripe_products` and `stripe_prices` tables

rshbhgrg:
**NOTE** : This is rebased on https://github.com/TryGhost/Ghost/pull/12844 for referenced `product` table, migrations in this PR are - https://github.com/TryGhost/Ghost/pull/12858/files#diff-c6b6e53efc55917d7a2f0e04252de41f319b42bd618c932e91f226b0ea610bc4 and https://github.com/TryGhost/Ghost/pull/12858/files#diff-1180657b62a40f8416566ba1fa73fd64dcc5bb005c970a727996b89353f1daf9

- Add the `stripe_products` table
- Add the `stripe_prices` table

### DIFF
--- a/core/server/data/exporter/index.js
+++ b/core/server/data/exporter/index.js
@@ -33,6 +33,8 @@ const BACKUP_TABLES = [
     'snippets',
     'tokens',
     'sessions',
+    'stripe_products',
+    'stripe_prices',
     'mobiledoc_revisions',
     'email_batches',
     'email_recipients',
--- a/core/server/data/migrations/versions/4.3/05-add-stripe-products-table.js
+++ b/core/server/data/migrations/versions/4.3/05-add-stripe-products-table.js
@@ -1,0 +1,9 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('stripe_products', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    product_id: {type: 'string', maxlength: 24, nullable: false, unique: false, references: 'products.id'},
+    stripe_product_id: {type: 'string', maxlength: 255, nullable: false, unique: true},
+    created_at: {type: 'dateTime', nullable: false},
+    updated_at: {type: 'dateTime', nullable: true}
+});
--- a/core/server/data/migrations/versions/4.3/06-add-stripe-prices-table.js
+++ b/core/server/data/migrations/versions/4.3/06-add-stripe-prices-table.js
@@ -1,0 +1,16 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('stripe_prices', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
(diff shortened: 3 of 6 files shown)


core/server/data/migrations/versions/4.3/04-add-stripe-prices-table.js
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = addTable('stripe_prices', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    stripe_price_id: {type: 'string', maxlength: 255, nullable: false, unique: true},
+    stripe_product_id: {type: 'string', maxlength: 255, nullable: false, unique: false, references: 'stripe_products.stripe_product_id'},
+    active: {type: 'boolean', nullable: false},
+    livemode: {type: 'boolean', nullable: false},
+    nickname: {type: 'string', maxlength: 50, nullable: true},
+    currency: {type: 'string', maxLength: 3, nullable: false},
+    amount: {type: 'integer', nullable: false},
```
allouis, line 11: I wanted to check, there is `unit_amount` and `unit_amount_decimal` do you know which one we plan to use/store here?

core/server/data/migrations/versions/4.3/04-add-stripe-prices-table.js
```diff
@@ -0,0 +1,16 @@
+const addTable = require('../../utils').addTable;
+
+module.exports = addTable('stripe_prices', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    stripe_price_id: {type: 'string', maxlength: 255, nullable: false, unique: true},
+    stripe_product_id: {type: 'string', maxlength: 255, nullable: false, unique: false, references: 'stripe_products.stripe_product_id'},
+    active: {type: 'boolean', nullable: false},
+    livemode: {type: 'boolean', nullable: false},
```
allouis, line 8: Is this necessary? No harm, just wondering the usecase :relaxed:

allouis (approved):
Looks good!

core/server/data/migrations/versions/4.3/04-add-stripe-prices-table.js
rishabhgrg: So I had checked, and it should be `unit_amount` in our case as that maps to what we have been using with plans so far, which is the unit amount in paise represented as whole integer, while the `unit_amount_decimal` is the same thing represented as string and allowing upto 12 decimal points.

core/server/data/migrations/versions/4.3/04-add-stripe-prices-table.js
rishabhgrg: 😄 I was on the fence as well but decided to include because it _might_ come in as handy info when we pull in info for outside Ghost prices and needs a filtering on current connected mode or anything.

tpatel (approved):
Looks great!

daniellockyer (approved):

PR #12875, opened 2021-04-14
Added permissions for the Products resource

allouis:
refs https://github.com/TryGhost/Team/issues/616

This adds the BREAD permissions for the Product resource, which'll be used by the Admin API.

We currently give `Administrator` and `Owner` all of these permissions, as discussed with Zimo.

It may be that the "read" permissions need more roles attached

### DIFF
--- a/test/regression/migrations/migration_spec.js
+++ b/test/regression/migrations/migration_spec.js
@@ -28,203 +28,121 @@ describe('Database Migration (special functions)', function () {
             roleNames.should.eql(roles);
         });
 
+        should.Assertion.add('havePermission', function (name, roles = null) {
+            const permission = this.obj.find((permission) => {
+                return permission.name === name;
+            });
+            should.exist(permission, `Could not find permission ${name}`);
+
+            if (roles) {
+                permission.should.be.AssignedToRoles(roles);
+            }
+        });
+
         // Custom assertion to wrap all permissions
         should.Assertion.add('CompletePermissions', function () {
             this.params = {operator: 'to have a complete set of permissions'};
             const permissions = this.obj;
 
-            // DB
-            permissions[0].name.should.eql('Export database');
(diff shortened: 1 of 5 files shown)


allouis:
@kevinansfield As well as reviewing the migration, I was hoping for your input on which roles would need the "browse" and "read" permissions? I'm thinking in the context of the PSM you'll be adding to set the access level for posts!

kevinansfield:
> which roles would need the "browse" and "read" permissions?

I think at a minimum all roles other than Contributor would need read/browse permission because they can all publish posts and it doesn't make sense to restrict the access levels they can assign. I'm not 100% sure that Contributors should be denied read permissions but I can't think of any immediate need for them to have read permissions.

allouis:
@kevinansfield Thanks! I've updated the PR to reflect that, we can always add Contributor when we have a use-case :)

test/regression/migrations/migration_spec.js
```diff
@@ -28,203 +28,121 @@ describe('Database Migration (special functions)', function () {
...
+        should.Assertion.add('havePermission', function (name, roles = null) {
+            const permission = this.obj.find((p) => {
+                return p.name === name;
+            });
+            should.exist(permission, `Could not find permission ${name}`);
+
+            if (roles) {
+                permission.should.be.AssignedToRoles(roles);
+            }
+        });
```
kevinansfield, line 40: ❤️

daniellockyer: ❤️

allouis: :blush:

daniellockyer (approved):

rshbhgrg (approved):

kevinansfield (approved):

PR #12879, opened 2021-04-14
Added settings keys to store OAuth related data

naz:
refs https://github.com/TryGhost/Team/issues/618

- The `oauth_client_id` and `oauth_client_secret` are placeholders to store OAuths related data.
- The flag for `oauth_enabled` or anything along those lines was not added intentionally in favour of checking if the `oauth_client_id` & `oauth_client_secret` are null.

WIP:
- need to reconfirm if creating a new group `sso` (or maybe named in a different way) is a right way to go
- need to double check if we are still meant to do an `update` on the migration level or `create` after the [split in the initialization sequence](https://github.com/TryGhost/Ghost/pull/12595) has happened

### DIFF
--- a/core/server/api/canary/utils/serializers/output/utils/settings-type-group-mapper.js
+++ b/core/server/api/canary/utils/serializers/output/utils/settings-type-group-mapper.js
@@ -12,7 +12,8 @@ const groupTypeMapping = {
     portal: 'portal',
     email: 'bulk_email',
     newsletter: 'newsletter',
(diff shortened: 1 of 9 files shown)


naz:
@tpatel have left a todo list above and some comments in the design doc to consider before finishing this pr ;) Good luck!

tpatel:
I've renamed the settings group from `sso` to `oauth`, and there is no need for a migration anymore!

PR #12881, opened 2021-04-15
Added stripe_price_id column to subscriptions table

rshbhgrg:
Migrations in [1](https://github.com/TryGhost/Ghost/pull/12881/files#diff-991c2e8a8b5778a361d07ddaa9951d876b4ba8abaf08a27c1f806c9b5b95e9d9) and [2](https://github.com/TryGhost/Ghost/pull/12881/files#diff-c74946cf49a9599dc502ce7793158f6a10f6dcd7700197dff3b71139abddac82)

- Adds new `stripe_price_id` column to subscriptions table to store stripe price ids
- Populates `stripe_price_id` column value to current `plan_id`
- Updates tests


core/server/data/schema/schema.js
```diff
@@ -450,11 +450,31 @@ module.exports = {
         updated_at: {type: 'dateTime', nullable: true},
         updated_by: {type: 'string', maxlength: 24, nullable: true}
     },
+    stripe_products: {
```
allouis, line 453: Not 100% sure why these have moved, it's a little tricky to check that nothing has changed

rishabhgrg: Ah they had originally moved due to a test failure with FK constraint where schema had to be re-ordered to not fail Table creation. I'll push a patch.

kevinansfield (approved):
LGTM 👍  Only thing is the migration numbers need updating, there's already an `08` migration in `4.3`

allouis (approved):

daniellockyer (approved):

PR #12886, opened 2021-04-16
Migrated `members_allow_free_signup` setting to `members_signup_access`

kevinansfield:

Currently the members signup setting is explicitly yes/no to allowing free members signup, with the implication that when set to "no" members is still active but members have to be created via Stripe or the admin API ("invite only").

This change renames the setting and changes its type to allow more than a binary option.

- migration to create/update the new setting based on the old value
  - free signup = "any", no free signup = "invite"; matches the current UI for this setting
- rename setting everywhere it's used/tested against
- modify `getAllowSelfSignup()`, used to configure members packages, to only return `true` when the new setting is set to `'all'` to match behaviour to the older setting
- update importer to rename the setting when importing from an older Ghost version

### DIFF
--- a/core/server/data/migrations/versions/4.3/08-migrate-members-signup-setting.js
+++ b/core/server/data/migrations/versions/4.3/08-migrate-members-signup-setting.js
@@ -1,0 +1,109 @@
+const logging = require('../../../../../shared/logging');
+const {createTransactionalMigration} = require('../../utils');
+const ObjectId = require('bson-objectid');
+
+module.exports = createTransactionalMigration(
+    async function up(connection) {
+        const oldSetting = await connection('settings')
+            .where('key', 'members_allow_free_signup')
+            .select('value', 'created_by', 'updated_by')
+            .first();
+
+        // no need to create new setting in this case, Ghost will create through default settings
+        if (!oldSetting) {
+            logging.warn('Could not find setting `members_allow_free_signup`, skipping update of `members_signup_access` setting');
+            return;
+        }
+
+        const newSetting = await connection('settings')
+            .where('key', 'members_signup_access')
+            .select('value')
+            .first();
+
+        const migrateValue = oldSetting.value ? 'all' : 'invite';
+
+        if (newSetting) {
+            // new setting already exists, *update* with migrated value
+            logging.info('Updating `members_signup_access` setting with value from `members_allow_free_signup`');
+            await connection('settings')
+                .where('key', 'members_signup_access')
+                .update('value', migrateValue);
+        } else {
+            // setting does not exist yet, *create* with migrated value
+            logging.info('Creating `members_signup_access` setting with value from `members_allow_free_signup`');
+
+            const currentTimestamp = connection.raw('CURRENT_TIMESTAMP');
+
+            const newSettingData = {
+                id: ObjectId.generate(),
+                group: 'members',
+                key: 'members_signup_access',
+                value: migrateValue,
+                type: 'string',
+                created_at: currentTimestamp,
+                updated_at: currentTimestamp,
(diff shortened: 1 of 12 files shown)


core/server/data/migrations/versions/4.3/08-migrate-members-signup-setting.js
```diff
@@ -0,0 +1,96 @@
...
+
+        const newSetting = await options.transacting('settings')
+            .where('key', 'members_signup_access')
+            .select('value')
+            .first();
+
+        const migrateValue = oldSetting.value ? 'all' : 'invite';
+
+        if (newSetting) {
+            // new setting already exists, *update* with migrated value
```
kevinansfield, line 29: I was tripped up by this a few times whilst testing in development. Normally this state wouldn't be reachable but until 4.3 is released we can end up in a half-way state:

1. Local instance updated to `main` with new migration
2. Ghost started, warning shown `WARN Please run `knex-migrator migrate --v 4.3 --force` to proceed!` but boot continues and creates default settings
3. `knex-migrator migrate --v 4.3 --force` is run

allouis: Gross! Good catch though :grin:

allouis (approved):

core/server/data/migrations/versions/4.3/08-migrate-members-signup-setting.js
```diff
@@ -0,0 +1,96 @@
+const logging = require('../../../../../shared/logging');
+const ObjectId = require('bson-objectid');
+
+module.exports = {
```
allouis, line 4: Not a big deal at all, but you could have used `createTransactionalMigration` from the utils for ever so slightly more concise migration :relaxed:

rshbhgrg (approved):
Looks good :)

core/server/data/migrations/versions/4.3/08-migrate-members-signup-setting.js
```diff
@@ -0,0 +1,96 @@
...
+            return;
+        }
+
+        // this can potentially be lossy if going from "nobody" but it matches to the
+        // most appropriate setting available in earlier versions of Ghost
+        const rollbackValue = newSetting.value === 'all' ? true : false;
+
+        logging.info('Creating `members_allow_free_signup` based on value from `members_signup_access`');
+        await options.transacting('settings')
+            .where('key', 'members_signup_access')
```
daniellockyer, line 88: I believe this should be `members_allow_free_signup` as per the comment

core/server/data/migrations/versions/4.3/08-migrate-members-signup-setting.js
daniellockyer: ➕ on this 🙂

kevinansfield: Updated to use `createTransactionalMigration` 🙂

core/server/data/migrations/versions/4.3/08-migrate-members-signup-setting.js
kevinansfield: Good catch, this wasn't working as intended. Updated to insert/update based on presence of the old key

daniellockyer: Shouldn't the new version still be changed to `members_allow_free_signup`?

kevinansfield: 🤦 too much copy/paste there. Fixed

daniellockyer (approved):

PR #12892, opened 2021-04-20
Added Default Product fixture

allouis:
refs https://github.com/TryGhost/Team/issues/619

As part of the Custom Products work, we are linking members to products
when updating their subscriptions. This requires that we have at least
one product in the database. For existing sites that are using Members
this is handled by the v4.3 03 migration. But for new sites we must
include a fixture.

### DIFF
--- a/core/server/data/schema/fixtures/fixtures.json
+++ b/core/server/data/schema/fixtures/fixtures.json
@@ -1,6 +1,15 @@
 {
     "models": [
         {
+            "name": "Product",
+            "entries": [
+                {
+                    "name":             "Default Product",
+                    "slug":             "default-product"
+                }
+            ]
+        },
+        {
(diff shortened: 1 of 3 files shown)


core/server/data/schema/fixtures/fixtures.json
```diff
@@ -1,5 +1,14 @@
 {
     "models": [
+        {
+            "name": "Product",
+            "entries": [
+                {
+                    "name":             "Default Product",
+                    "slug":             "default-product"
```
rishabhgrg, line 8: Think we should get rid of the extra space here after `:` as we don't use it anywhere else in rest of the file.

rshbhgrg (approved):
LGTM, just needs the test fix

daniellockyer (approved):

PR #12899, opened 2021-04-23
Added OAuth user data

tpatel:
issue https://github.com/TryGhost/Team/issues/614

## String field sizes

The sizes are picked from the [Ghost string sizes](https://github.com/TryGhost/Ghost/blob/49f48c0f09ba2b07c0af9725ca24941fa2144ce3/core/server/data/schema/schema.js#L1-L9).

- `id` is a standard ID, so it gets the standard size (24)
- `user_id` is a foreign key so it gets the same size as the other column
- `provider` is a simple string that contains a provider name like `google`, `github`... From the [grant npm module](https://www.npmjs.com/package/grant) (compatible with 200+ oauth providers), the longest provider string is 16 chars, so we should be good with 50 characters.
- `provider_id` is an arbitrary string provided by the oauth provider. It's hard to find accurate data outside of [Twitter](https://developer.twitter.com/en/docs/twitter-ids), which has 64-bits unsigned integers user ids (max 19 characters). Given the magnitude of twitter, 191 characters to store user ids seems very safe. A recent google id example was 21 characters-long (all digits) so it confirms the 191-characters length.
- `access_token` from [google can be up to 2048 bytes](https://developers.google.com/identity/protocols/oauth2#size) so it get a 65535 length
- `refresh_token` from [google can be up to 512 bytes](https://developers.google.com/identity/protocols/oauth2#size), so it gets a 200 length

### DIFF
--- a/core/server/data/schema/schema.js
+++ b/core/server/data/schema/schema.js
@@ -112,6 +112,16 @@ module.exports = {
         updated_at: {type: 'dateTime', nullable: true},
         updated_by: {type: 'string', maxlength: 24, nullable: true}
     },
+    oauth: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        provider: {type: 'string', maxlength: 50, nullable: false},
+        provider_id: {type: 'string', maxlength: 191, nullable: false},
+        access_token: {type: 'text', maxlength: 65535, nullable: true},
+        refresh_token: {type: 'text', maxlength: 2000, nullable: true},
+        created_at: {type: 'dateTime', nullable: false},
+        updated_at: {type: 'dateTime', nullable: true},
+        user_id: {type: 'string', maxlength: 24, nullable: false, references: 'users.id'}
+    },
     posts_authors: {
         id: {type: 'string', maxlength: 24, nullable: false, primary: true},
         post_id: {type: 'string', maxlength: 24, nullable: false, references: 'posts.id'},
--- a/core/server/data/exporter/table-lists.js
+++ b/core/server/data/exporter/table-lists.js
@@ -14,6 +14,7 @@ const BACKUP_TABLES = [
     'members_stripe_customers_subscriptions',
     'migrations',
     'migrations_lock',
+    'oauth',
     'permissions',
     'permissions_roles',
     'permissions_users',
--- a/core/server/data/migrations/versions/4.11/01-add-oauth-user-data.js
+++ b/core/server/data/migrations/versions/4.11/01-add-oauth-user-data.js
@@ -1,0 +1,12 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('oauth', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    provider: {type: 'string', maxlength: 50, nullable: false},
+    provider_id: {type: 'string', maxlength: 191, nullable: false},
+    access_token: {type: 'text', maxlength: 65535, nullable: true},
+    refresh_token: {type: 'text', maxlength: 2000, nullable: true},
+    created_at: {type: 'dateTime', nullable: false},
+    updated_at: {type: 'dateTime', nullable: true},
+    user_id: {type: 'string', maxlength: 24, nullable: false, references: 'users.id'}
+});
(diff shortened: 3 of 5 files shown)


allouis (approved):
Looks good, are these column sizes definitely what we want? Will they work for all/most providers?

tpatel:
> Looks good, are these column sizes definitely what we want? Will they work for all/most providers?

@allouis good point, I've increased the size of the access and refresh tokens to match the maximum length [specced by google](https://developers.google.com/identity/protocols/oauth2#size):
> Tokens can vary in size, up to the following limits:
>   - Authorization codes: 256 bytes
>   - Access tokens: 2048 bytes
>   - Refresh tokens: 512 bytes

Changes made:

- access token: from 191 to 65535 chars
- refresh token: from 191 to 2000 chars

That is huge and should fit most OAuth providers.

daniellockyer (changes requested):
@tpatel Need to update the migration path to `4.4/01-...` 🙂

tpatel:
I'm creating another PR to reduce the `users` row size because I can't add these oauth fields

tpatel:
Hey all, after understanding that I wouldn't be able to modify the `users` table (because we're not yet releasing a major 😅 ), I've changed this PR to create another table linked to `users`.

It should be a 1:1 relation in the upcoming implementation, but I prefer to keep the door open to a user that has multiple OAuth providers configured, which is why the foreign key is from the `oauth` table to `users`.

core/server/data/migrations/versions/4.4/01-add-oauth-user-data.js
```diff
@@ -0,0 +1,17 @@
+const {
+    addTable,
+    combineNonTransactionalMigrations
+} = require('../../utils');
+
+module.exports = combineNonTransactionalMigrations(
```
daniellockyer, line 6: You can just do:

```suggestion
module.exports = addTable('oauth', {
```

daniellockyer:
@tpatel Reminder that we need to add an entry to https://github.com/TryGhost/Ghost/blob/main/core/server/data/exporter/index.js#L13-L47

core/server/data/migrations/versions/4.4/01-add-oauth-user-data.js
tpatel: yess good point 💯 , it's code from the previous implementation!

core/server/data/schema/schema.js
```diff
@@ -107,11 +107,22 @@ module.exports = {
         meta_description: {type: 'string', maxlength: 2000, nullable: true, validations: {isLength: {max: 500}}},
         tour: {type: 'text', maxlength: 65535, nullable: true},
         last_seen: {type: 'dateTime', nullable: true},
+
```
daniellockyer, line 110: Also, where did this newline come from? 😅

daniellockyer: 😁

allouis (approved):

daniellockyer (approved):

core/server/data/schema/schema.js
```diff
@@ -112,6 +112,16 @@ module.exports = {
         updated_at: {type: 'dateTime', nullable: true},
         updated_by: {type: 'string', maxlength: 24, nullable: true}
     },
+    oauth: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        provider: {type: 'string', maxlength: 191, nullable: false},
+        provider_id: {type: 'string', maxlength: 191, nullable: false},
```
naz, line 118: We'd need some references regarding this field because at the moment it's unclear if 191 characters would do a trick.

naz (changes requested):
Some fields still need documented references around the size.

core/server/data/schema/schema.js
tpatel: @naz I've updated the issue with a justification for each length + rebased it, it should be ready to be merged, let me know!

daniellockyer:
@tpatel We'll need to update this to 4.11 now 🙂

tpatel:
> @tpatel We'll need to update this to 4.11 now 🙂

Right! it's done ✅

naz (approved):
Looking good :+1:

daniellockyer (approved):

PR #12905, opened 2021-04-26
🐛 Restored correct "allow free signup" setting from backup after buggy 4.3 upgrade

kevinansfield:
refs https://github.com/TryGhost/Team/issues/634

- find earliest backup file created when a 4.3 migration was run, if found use the `members_allow_free_signup` value from there to change `members_signup_access` from `'all'` to `'invite'` if necessary

### DIFF
--- a/core/server/data/migrations/versions/4.4/01-restore-free-members-signup-setting-from-backup.js
+++ b/core/server/data/migrations/versions/4.4/01-restore-free-members-signup-setting-from-backup.js
@@ -1,0 +1,98 @@
+const logging = require('../../../../../shared/logging');
+const {createTransactionalMigration} = require('../../utils');
+const config = require('../../../../../shared/config');
+const fs = require('fs-extra');
+const path = require('path');
+const semver = require('semver');
+
+// 4.3.0 contained a buggy migration for members_allow_free_signup -> members_signup_access
+// meaning all sites switched to allowing free member signup.
+//
+// This migration attempts to fix that by finding the original value from the db backup file
+// created during migration and updating the new setting value to match
+
+// we only care about backups created since the bad release was published
+const MIN_DATE = new Date('2021-04-21T00:00:00.000Z');
+
+module.exports = createTransactionalMigration(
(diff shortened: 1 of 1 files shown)


daniellockyer:
@kevinansfield I think we should be try-catching some of this. The other time we've restored from backups [we did](https://github.com/TryGhost/Ghost/blob/main/core/server/data/migrations/versions/2.18/1-restore-settings-from-backup.js), and we don't want it to crash out if there's a problem reading the directory/requiring a file/parsing the JSON.

kevinansfield:
Added try/catch blocks around reading and parsing

daniellockyer (approved):

core/server/data/migrations/versions/4.4/01-restore-free-members-signup-setting-from-backup.js
```diff
@@ -0,0 +1,98 @@
...
+        const backupFiles = files
+            .filter(filename => backupFileRegex.test(filename))
+            .filter(filename => backupFileDate(filename) > MIN_DATE)
+            .sort((a, b) => backupFileDate(a) - backupFileDate(b));
+
+        let hasRestored = false;
+
+        for (const backupFile of backupFiles) {
+            try {
+                const backup = require(path.join(dataPath, backupFile));
```
naz, line 68: Because we might be processing multiple backup files here it would be useful to add logging about which exactly backup has been used. Something along the lines:
```
const backupPath = path.join(dataPath, backupFile);
const backup = require(path.join(backupPath));

logging.info(`Checking backup data stored at ${backupPath}`);
```

In case someone runs into problems with this migration it would be good to know which file exactly has caused them.

This is a minor improvement that might be useful, not critical to merging the PR :wink:

naz (approved):
:+1: like the new variation over the `1-restore-settings-from-backup` one! hopefully we come up with a testing framework and never have to deal with anything similar :grimacing:

allouis (approved):

PR #12907, opened 2021-04-27
Migrated 'invite' to 'all' for sites that were not truly invite-only

kevinansfield:

The previous `members_allow_free_signup` -> `members_signup_access` migration made a direct correlation between the toggle `true/false` to `all/invite` under the assumption that behaviour between the two settings would be identical. The assumption was incorrect and the behaviour is changing so `invite` forces invite-only mode, stopping all front-end signup to free or paid plans with the free plan now being disabled via the portal plans setting.

- check existing `members_signup_access` setting and if it's `'invite'` migrate it to `'all'` where signup should still be possible. The "invite-only" mode should only be active if certain conditions are met:
  - Stripe is not configured ("allow free member signup" off and no Stripe showed "invite-only" in portal)
  - Stripe is configured but no plans are selected in portal (no plans showed "invite-only" in portal)
- when migrating `'invite'` to `'all'`, also remove `'free'` plan from the `portal_plans` setting to avoid previously paid-only sites unexpectedly showing a free plan on signup

### DIFF
--- a/core/server/data/migrations/versions/4.4/02-migrate-members-signup-access.js
+++ b/core/server/data/migrations/versions/4.4/02-migrate-members-signup-access.js
@@ -1,0 +1,126 @@
+const logging = require('../../../../../shared/logging');
+const {createTransactionalMigration} = require('../../utils');
+const MembersConfigProvider = require('../../../../services/members/config');
+const settingsCache = require('../../../../services/settings/cache');
+const config = require('../../../../../shared/config');
+
+// members_allow_free_signup was migrated to members_signup_access in 4.3/08-migrate-members-signup-setting
+// using a direct binary migration of true=all and false=invite based on new setting behaviour matching the
+// old setting. That was innacurate and the behaviour is changing to a proper invite-only toggle which
+// requires a more nuanced migration.
+//
+// 'all' - no need to change, matches expected behaviour
+// 'invite' - change to 'all' based on members usage
+// 'none' - no need to change, it's a completely new option that wasn't previously migrated to
+//
+// For the majority of cases 'invite' should actually be 'all'. The true invite-only case is:
+// - stripe not configured
+// - portal configured with no plans
+//
+// When switching from signup access from 'invite' to 'all' we also need to remove the 'free' plan from
+// portal settings to avoid paid-only sites suddenly offering free plan signups
+
+module.exports = createTransactionalMigration(
+    async function up(connection) {
+        const currentSetting = await connection('settings')
+            .where('key', 'members_signup_access')
+            .select('value')
+            .first();
+
+        if (!currentSetting) {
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.4/02-migrate-members-signup-access.js
```diff
@@ -0,0 +1,127 @@
...
+            logging.info('Skipping update of members_signup_access setting. Portal configured as invite-only, staying as invite-only');
+            return;
+        }
+
+        logging.info(`Updating members_signup_access setting to 'all'`);
+        await connection('settings')
+            .where('key', 'members_signup_access')
+            .update({value: 'all'});
+
+        const currentPlansSetting = await connection('settings')
```
rishabhgrg, line 67: 🤔 Why do we need to fetch the portal _plans settings here again, we could use the one already fetched [here](https://github.com/TryGhost/Ghost/pull/12907/files#diff-9a8f70a21f28ddd4a9dd7958c1457734c9e983c1eaf12a09acf3ee5c5fc6498aR52) right  ?

kevinansfield: Good point 😄 was just an oversight, updating...

core/server/data/migrations/versions/4.4/02-migrate-members-signup-access.js
```diff
@@ -0,0 +1,126 @@
...
+            logging.info(`Skipping update of members_signup_access setting. Not set to 'invite', keeping current value`);
+            return;
+        }
+
+        const membersConfig = new MembersConfigProvider({
+            settingsCache,
+            config
+        });
+
+        const hasStripe = membersConfig.isStripeConnected();
```
rishabhgrg, line 45: 🤔 Just wondering if we should look into extracting this logic out directly via settings/config in the migration than depend on `membersConfig`, my only concern being any future changes to membersConfig or this method might break the migration. Ideally any updates to members config would mean this piece of code is updated as well so maybe its fine to leave as-is ?

kevinansfield: Yeah, I had similar concerns but it seemed like checking if Stripe is configured touched both settings/config with some additional logic which also seemed brittle to have in a migration.

Extracting the check could be useful but would that gain much? If `membersConfig` changes I'd expect usage of it to be checked too 🤔

rishabhgrg: Yeah that's fair, and its more likely we continue to have a method to check if stripe is connected so any changes can update this on usage check

rshbhgrg (approved):

daniellockyer (approved):

allouis (approved):

PR #12909, opened 2021-04-27
Updated members `allowSelfSignup()` to take portal plans into account

kevinansfield:

`members_signup_access = 'invite'` now forces invite-only mode so both free and paid setups both use the `'all'` setting. To ensure we're properly allowing/disabling free (self signup) signups in the members API we need to update `allowSelfSignup()` to take additional settings into account.

- `true` when Stripe is not connected. There are no paid plans available in this configuration so free signup is always enabled. To disable free signup on a site with no Stripe setup the members signup access should be set to `invite` or `none`.
- `true` when Stripe is configured and free plan is enabled in portal, without it Members API would not send magic link emails to signup requests
- `false` in all other situations such as invite-only and members-disabled signup access modes, or when the free plan has been disabled in portal configuration

### DIFF
--- a/core/server/services/members/config.js
+++ b/core/server/services/members/config.js
@@ -215,7 +215,25 @@ class MembersConfigProvider {
     }
 
     getAllowSelfSignup() {
-        return this._settingsCache.get('members_signup_access') === 'all';
+        // 'invite' and 'none' members signup access disables all signup
+        if (this._settingsCache.get('members_signup_access') !== 'all') {
+            return false;
+        }
+
+        // if stripe is not connected then selected plans mean nothing.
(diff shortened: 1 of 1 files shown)


rshbhgrg (approved):
This will mostly need update when changes for custom products is done and `portal_plans` is changed, but good for now :)

kevinansfield:
> This will mostly need update when changes for custom products is done and portal_plans is changed, but good for now :)

👍 I wonder if at that point it would be better to drop `allowSelfSignup` and let portal and members API work things out as needed from the direct `members_signup_access` and `portal_plans` (or product-enabled equivalent) settings?

Anyhow, future discussion 😄 I'll merge this as-is for the current release, thanks for the review.

PR #12910, opened 2021-04-27
🎨 Added `{{#has members="all|invite|none|true|false"}}`

kevinansfield:
refs https://github.com/TryGhost/Team/issues/626

We've introduced some new options for members signup access which allows for explicitly turning members off altogether or turning off signup to make an invite-only site.

Previously `@labs.members` was used to determine if members was enabled or disabled but that setting is deprecated and will be removed in the next major Ghost release. Instead the `{{#has}}` helper has been extended to provide more granular conditionals for themes.

Usage:
`{{#has members="true"}}` - members is active with either all functionality or sign-in only
`{{#has members="all"}}` - members is active with sign-in and sign-up enabled
`{{#has members="invite"}}` - members is active with only sign-in enabled
`{{#has members="false/none"}}` - all members functionality is disabled

### DIFF
--- a/core/frontend/helpers/has.js
+++ b/core/frontend/helpers/has.js
@@ -4,9 +4,9 @@
 //
 // Checks if a post has a particular property
 
-const {logging, i18n} = require('../services/proxy');
+const {logging, i18n, settingsCache} = require('../services/proxy');
 const _ = require('lodash');
-const validAttrs = ['tag', 'author', 'slug','visibility', 'id', 'number', 'index', 'any', 'all'];
+const validAttrs = ['tag', 'author', 'slug','visibility', 'id', 'number', 'index', 'any', 'all', 'members'];
 
 function handleCount(ctxAttr, data) {
     if (!data || !_.isFinite(data.length)) {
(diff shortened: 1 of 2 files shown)


ErisDS:
Implementation LGTM, design IMO needs a slight tweak. One of the tenets of the theme API is having one-clear-way to do things, which `{{#has members="false/none"}}` doesn't pass. I would only support false.

ErisDS:
@kevinansfield actually can I just double check - is this setting exposed via the Content API settings endpoint, and if so what values does it have?

kevinansfield:
> is this setting exposed via the Content API settings endpoint, and if so what values does it have?

It's not currently exposed by the Content API settings endpoint. If it was and we weren't modifying the value in serializers then the value would be `'all'`, `'invite'`, or `'none'`.

ErisDS:
To summarise a chat from slack, it feels like this is conflating two things into one, and we'll end up not being able to have consistency between the Content Settings API and this helper.

I think, whatever things are called, the suite of values should either match the members_signup_access  values or be true/false, but not merge both into a single property.

fueko:
Hi @kevinansfield 👋

Is there any chance that this helper will appear in future versions of Ghost? It would be great to have support for this in the themes :)

PR #12928, opened 2021-05-06
Removed unique constraint from Product names

allouis:

A discussion in the Members team resulted in us determining that we do
not need to enforce unique names for Products. Stripe does not enforce
uniqueness for their Products, and we feel it's not necessary for us to.

### DIFF
--- a/core/server/data/migrations/versions/4.5/04-remove-unique-constraint-from-product-name.js
+++ b/core/server/data/migrations/versions/4.5/04-remove-unique-constraint-from-product-name.js
@@ -1,0 +1,13 @@
+const {createNonTransactionalMigration} = require('../../utils');
+const {dropUnique} = require('../../../schema/commands');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        return await dropUnique('products', 'name', knex);
+    },
+    async function down() {
+        // We cannot roll back this migration because adding a unique constraint
+        // could error if we have duplicate names already in the table.
+        return;
+    }
+);
--- a/core/server/data/schema/schema.js
+++ b/core/server/data/schema/schema.js
@@ -363,7 +363,7 @@ module.exports = {
     },
     products: {
         id: {type: 'string', maxlength: 24, nullable: false, primary: true},
-        name: {type: 'string', maxlength: 191, nullable: false, unique: true},
+        name: {type: 'string', maxlength: 191, nullable: false},
         slug: {type: 'string', maxlength: 191, nullable: false, unique: true},
         description: {type: 'string', maxlength: 191, nullable: true},
         created_at: {type: 'dateTime', nullable: false},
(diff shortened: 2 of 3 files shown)


core/server/data/migrations/versions/4.5/03-remove-unique-constraint-from-product-name.js
```diff
@@ -0,0 +1,14 @@
+const {createNonTransactionalMigration} = require('../../utils');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        return knex.schema.table('products', (table) => {
```
rishabhgrg, line 5: @allouis Think this should be `schema.alterTable` here instead of `schema.table`

allouis: They're the same thing! https://github.com/knex/knex/blob/master/lib/schema/builder.js#L77

rishabhgrg: TIL 👍

rshbhgrg (approved):
Tests have a weird failure, rest looks good

core/server/data/migrations/versions/4.5/03-remove-unique-constraint-from-product-name.js
```diff
@@ -0,0 +1,16 @@
+const {logging} = require('../../../../../shared');
+const {createNonTransactionalMigration} = require('../../utils');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        logging.info('Dropping UNIQUE constraint from products.name');
+        return knex.schema.alterTable('products', (table) => {
```
naz, line 7: @allouis we should be reusing our existing helpers where possible :wink: 

Could be something like:
```

const {dropUnique} = require('../../../schema/commands');

...

return await dropUnique('products', 'name', knex);
```

naz (changes requested):

core/server/data/migrations/versions/4.5/03-remove-unique-constraint-from-product-name.js
```diff
@@ -0,0 +1,16 @@
+const {logging} = require('../../../../../shared');
```
daniellockyer, line 1: I don't think this works because we don't re-export `logging` from within shared. But if you switch to the helper function like Naz suggested, it does the logging for you 🙂

allouis:
@naz @daniellockyer updated to use `dropUnique`

core/server/data/migrations/versions/4.5/03-remove-unique-constraint-from-product-name.js
allouis: Nice, thanks!

daniellockyer (approved):

daniellockyer:
@allouis I've approved but we've just merged another `03-` so you'll need to update it 😅

allouis:
Done!

naz (approved):
This is good to go :rocket:

Issue #12860, opened 2021-04-09
Disable Gravatar Feature Not Working


rkafp:
### Issue Summary

When trying to disable the Gravatar feature by using "useGravatar": false it doesn't work.

### To Reproduce

1. Open config.production.json
2. Add this line

```
"privacy": {
     "useGravatar": false
}
```

3. ghost restart

### Technical details:

* Ghost Version: 4.2.0
* Node Version: 12.2.1
* Browser/OS: Ubuntu
* Database: MySQL

ErisDS:
Please provide more information about what isn't working how you would expect.

ErisDS:
Closing now as there's not enough info here.

PR #12929, opened 2021-05-06
Disable Gravatar Feature

Rashmildp:
Disable the Gravatar Feature  #12860


ErisDS:
I am not sure what this PR or the related issue are meant to be about - going to close both now. Please provide a reproduction case if there's something that needs fixing.

PR #12946, opened 2021-05-08
✨ Added support for gating content by member labels and products

kevinansfield:
refs https://github.com/TryGhost/Team/issues/581
closes https://github.com/TryGhost/Team/issues/582

Emails can now be sent to members with specific associated labels or products by specifying an NQL string. We want to bring the same members segment feature to content by allowing `visibility` to be an NQL filter string on top of the `public/members/paid` special-case strings.

As an example it's possible to set `posts.visibility` to `label:vip` to make a post available only to those members with the `vip` label.

- removed enum validations for `visibility` so it now accepts any string or `null`
    - bumped `@tryghost/admin-api-schema` for API-level validation changes
- added nql validation to API input validators by running the visibility query against the members model
- added transform of NQL to special-case visibility values when saving post model
    - ensures there's a single way of representing "members" and "paid" where NQL gives multiple ways of representing the same segment
    - useful for keeping theme-level checks such as `{{#has visibility="paid"}}` working as expected
- updated content-gating to parse nql from post's visibility and use it to query the currently logged in member to see if there's a match
    - bumped @tryghost/members-api to include label and product data when loading member

TODO:
- [x] fetch associated label and product data when fetching logged in member
  - [x] release and bump `@tryghost/members-api`
- [x] update content gating to run an NQL against the logged in member object
  - [x] tests
- [x] remove API schema validations to let NQL through in `visibility`
  - [x] release and bump `@tryghost/admin-api-schema`
- [x] validate NQL when saving a post
- [x] ~update `foreach` and `has` helpers to support NQL values~ (not necessary, exact matching still works. Advanced NQL parse+match is a separate issue)
- [x] transform NQL query values to known special cases on save (eg, `status:-free` to `paid` or `status:free,status:-free` to `members`)
- [x] ~update default content CTA wording to handle custom segments~ (re-visiting later if needed)

NOTE:
- transform of `paid` to `status:-free` has not been done here as it has a wider impact on API compatibility, for now it's kept as a special-case as per `public` and `members` (both similar to the special-case "all" in `posts.email_recipient_filter`)

### DIFF
--- a/test/utils/fixture-utils.js
+++ b/test/utils/fixture-utils.js
@@ -461,35 +461,36 @@ const fixtures = {
         return Promise.map(DataGenerator.forKnex.labels, function (label) {
             return models.Label.add(label, context.internal);
         }).then(function () {
-            return Promise.each(_.cloneDeep(DataGenerator.forKnex.members), function (member) {
-                let memberLabelRelations = _.filter(DataGenerator.forKnex.members_labels, {member_id: member.id});
-
-                memberLabelRelations = _.map(memberLabelRelations, function (memberLabelRelation) {
-                    return _.find(DataGenerator.forKnex.labels, {id: memberLabelRelation.label_id});
-                });
-
-                member.labels = memberLabelRelations;
-
-                // TODO: replace with full member/product associations
-                if (member.email === 'with-product@test.com') {
-                    member.products = [{slug: 'default-product'}];
-                }
-
-                return models.Member.add(member, context.internal);
-            });
-        }).then(function (member) {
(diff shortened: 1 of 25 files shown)


test/utils/fixture-utils.js
```diff
@@ -457,39 +457,41 @@ const fixtures = {
         });
     },
 
-    insertMembersAndLabels: function insertMembersAndLabels() {
+    insertMembersAndLabelsAndProducts: function insertMembersAndLabelsAndProducts() {
```
allouis, line 460: Thanks for renaming this :heart:

PR #12970, opened 2021-05-17
Remove the `'comped'` status from Members

allouis:
Since we've removed the concept of Complimentary, we are replacing the `'comped'` status for Members with `'paid'`, and considering previous use of the `'comped'` status, as if it were `'paid'`.

We update the v3 API to always considered a member to have `comped: false`, and we remove the `comped` flag from the canary/v4 API.

### DIFF
--- a/test/regression/api/canary/admin/utils.js
+++ b/test/regression/api/canary/admin/utils.js
@@ -2,15 +2,9 @@ const url = require('url');
 const _ = require('lodash');
 const testUtils = require('../../../../utils');
 
-// NOTE: the dependance on the schema here is wrong! It is a design flaw which is causing problems for API maintenance and compatibility
-//       whenever you need to modify any of the below property lists using schema - rework them into an "allowlist" array like it's done in
-//       the commit introducing this comment.
-const schema = require('../../../../../core/server/data/schema').tables;
-
 const API_URL = '/ghost/api/canary/admin/';
 
 const expectedProperties = {
-    // API top level
     posts: ['posts', 'meta'],
     tags: ['tags', 'meta'],
     users: ['users', 'meta'],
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.6/01-remove-comped-status.js
```diff
@@ -0,0 +1,39 @@
+const logging = require('../../../../../shared/logging');
```
allouis, line 1: @rishabhgrg @daniellockyer @kevinansfield This is the migration for review

core/server/models/member-status-event.js
```diff
@@ -17,15 +17,12 @@ const MemberStatusEvent = ghostBookshelf.Model.extend({
             return qb.clear('select')
```
allouis, line 17: I would also be interested in a double check of my logic for updating the handling of `member_status_events`

core/server/data/migrations/versions/4.6/01-remove-comped-status.js
```diff
@@ -0,0 +1,39 @@
...
+                        'members_stripe_customers_subscriptions.status',
+                        ['active', 'trialing', 'past_due', 'unpaid']
+                    );
+                }
+            ).where(
+                'members_stripe_customers_subscriptions.plan_nickname',
+                '=',
+                'Complimentary'
+            )).map(({id}) => id);
+
```
naz, line 34: Adding a log message here saying how many members are being updated would be useful. Also, do we even want to run a query if `compedMemberIds.length === 0` ?

test/regression/api/canary/admin/utils.js
```diff
@@ -86,7 +86,6 @@ const expectedProperties = {
     member: _(schema.members)
         .keys()
         .concat('avatar_image')
-        .concat('comped')
```
naz, line 89: As this is been touched can you please follow up with a small refactor into "allowlist" syntax as done [here](https://github.com/TryGhost/Ghost/commit/06dd9bac5957d4cd3415a3086c275b2a7501fe3e)? :raised_hands:

naz (approved):
Apart from very _minor_ nits the migration worked well on MySQL and SQlite :muscle:

test/regression/api/canary/admin/utils.js
allouis: Done!

core/server/data/migrations/versions/4.6/01-remove-comped-status.js
allouis: Done :+1:

rshbhgrg (approved):

daniellockyer (approved):

PR #12973, opened 2021-05-20
Handled week & day interval in population of mrr events

allouis:
refs https://github.com/TryGhost/Team/issues/635

This is to ensure we don't break migrations for any sites which have
imported external subscriptions which have an interval of 'week' or
'day

### DIFF
--- a/core/server/data/migrations/versions/4.0/25-populate-members-paid-subscription-events-table.js
+++ b/core/server/data/migrations/versions/4.0/25-populate-members-paid-subscription-events-table.js
@@ -29,10 +29,24 @@ module.exports = createTransactionalMigration(
             if (interval === 'month') {
                 return amount;
             }
+
+            if (interval === 'week') {
+                return amount * 4;
+            }
+
+            if (interval === 'day') {
+                return amount * 30;
+            }
+
+            throw new Error(`Unknown Subscription interval "${interval}" found.`);
         }
 
         const allEvents = allSubscriptions.reduce((allEventsAcc, subscription) => {
             if (['incomplete', 'incomplete_expired'].includes(subscription.status)) {
+                return allEventsAcc;
+            }
+
+            if (!['year', 'month', 'week', 'day'].includes(subscription.plan_interval)) {
                 return allEventsAcc;
             }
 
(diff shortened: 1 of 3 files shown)


core/server/data/migrations/versions/4.0/25-populate-members-paid-subscription-events-table.js
```diff
@@ -29,6 +29,14 @@ module.exports = createTransactionalMigration(
...
             }
+
+            if (interval === 'week') {
+                return amount * 4;
+            }
+
+            if (interval === 'day') {
+                return amount * 30;
+            }
         }
```
daniellockyer, line 40: @allouis So we've seen this migration fail before where the subscription interval was `week` and obviously this PR fixes that, but would it be useful to have some sort of error/ignoring for when it's still not handled? I realise we're probably handling all cases now but..... 😅

core/server/data/migrations/versions/4.0/25-populate-members-paid-subscription-events-table.js
```diff
@@ -29,6 +29,14 @@ module.exports = createTransactionalMigration(
             if (interval === 'month') {
                 return amount;
             }
+
+            if (interval === 'week') {
+                return amount * 4;
+            }
+
+            if (interval === 'day') {
+                return amount * 30;
```
daniellockyer, line 38: Are the `30` and `4` just above too general to "the average week or month" or doesn't it really matter for us?

core/server/data/migrations/versions/4.0/25-populate-members-paid-subscription-events-table.js
allouis: So the only valid values from stripe are: "day, week, month or year." - but I think you're right, we should probably error if we encounter an unknown one

core/server/data/migrations/versions/4.0/25-populate-members-paid-subscription-events-table.js
allouis: I spoke with Matt and he's happy with these - "Yep, they’re edge cases anyway"

core/server/data/migrations/versions/4.0/25-populate-members-paid-subscription-events-table.js
daniellockyer: I foresee `decade` or something 😆

daniellockyer (approved):

PR #13001, opened 2021-06-02
Added groundwork for storing additional feature image data

kevinansfield:
requires https://github.com/TryGhost/SDK/pull/387 (checkout and `yarn link`)

We want to allow setting captions for feature images but there are technical hurdles that prevent a straightforward implementation:

1. we can't push more columns into the `posts` table (we're hitting max row length)
2. we can't push more columns into the `posts_meta` table (this will become a generic meta data table in future)
3. we don't want to store feature images inside mobiledoc (feature images are used outside of post content)

To work around these the best solution will be to store images as their own model and link them to parent models. This change takes the approach of using Bookshelf's polymorphic associations to link an image to a parent via an `imageable` association.

- adds `images` table
- migrates existing `posts.feature_image` values to `image` records
  - `posts.feature_image` needs to be kept around until a major release where we are able to drop it
- adds `Image` model
- updates `Post` model
  - adds `feature_image()` association
  - updates default fetch behaviour to include `feature_image` association
- updates `v2` and `v3` API serializers to convert the embedded `feature_image` record back into a URL string to maintain backwards compatibility
- updates `{{img_url}}` helper to handle a plain URL string or an embedded image resource to maintain compatibility across APIs


kevinansfield:
Closing in favour of adding fields directly to `posts_meta` (https://github.com/TryGhost/Ghost/pull/13030).

PR #13005, opened 2021-06-03
🔒 Added a "reset all passwords" feature

tpatel:
issue https://github.com/TryGhost/Team/issues/750

- Only accessible by admins
- Resets all staff users' passwords and prevents them to log-in
- Sends them a reset email password to give them back access to their account
- Closes all existing staff user sessions


core/server/web/api/canary/admin/routes.js
```diff
@@ -67,6 +67,7 @@ module.exports = function apiRoutes() {
 
     // ## Users
     router.get('/users', mw.authAdminApi, http(apiCanary.users.browse));
+    router.get('/users/reset-all-passwords', mw.authAdminApi, http(apiCanary.users.resetAllPasswords));
```
naz, line 70: I don't think we use kebab casing anywhere else to name an endpoint :thinking:  It's usually underscore separated like: `/email_preview/posts/:id`,  `/members/stats/gross_volume`.

Also not sure about the name. What other names have you considered when picking this one?

core/server/api/canary/users.js
```diff
@@ -230,5 +233,31 @@ module.exports = {
                 return models.ApiKey.refreshSecret(model.toJSON(), Object.assign({}, {id: model.id}));
             });
         }
+    },
+
+    resetAllPasswords: {
+        headers: {
+            cacheInvalidate: true
```
naz, line 240: Why are we invalidating the cache?

core/server/api/canary/users.js
```diff
@@ -230,5 +233,31 @@ module.exports = {
...
+            const isAdmin = user.hasRole('Owner') || user.hasRole('Administrator');
+            if (!isAdmin) {
+                return Promise.reject();
+            }
+            return Promise.resolve();
+        },
+        async query(frame) {
+            await userService.resetAllPasswords(frame.options);
+
+            // TODO: make sure it's ok to return this kind of JSON
```
naz, line 257: Would be good to have references for this from other endpoints that are "actions" and not resource fetching.

core/server/api/canary/users.js
```diff
@@ -230,5 +233,31 @@ module.exports = {
...
         }
+    },
+
+    resetAllPasswords: {
+        headers: {
+            cacheInvalidate: true
+        },
+        permissions: async (frame) => {
+            // TODO: Is there an easier way than querying the roles?
+            const user = await models.User.findOne({
```
naz, line 244: Is there a related resource which has similar permission level needed? We might need to add entries to permissions table instead :thinking:  Definitely should not be doing checks like this here and leave it for the permission layer to handle  :wink:

core/server/services/users.js
```diff
@@ -17,17 +19,66 @@ const path = require('path');
...
+
+    async resetAllPasswords(frameOptions) {
+        return this.models.Base.transaction(async (t) => {
+            frameOptions.transacting = t;
+
+            // Reset all passwords
+            const users = await this.models.User.findAll({...frameOptions});
+            for (const user of users) {
+                await user.save({
+                    password: await Users.randomPassword(),
```
naz, line 66: Do we even need to set a random password or just setting the locked status is enough?

naz (commented):
Added quite few comments. I think the direction is good needs some more reference material and polishing up (it's not `main` ready for sure :sweat_smile: )

core/server/web/api/canary/admin/routes.js
matthanley: This should also **definitely not** be a GET route

tpatel: It's updated to `PUT /users/reset_all_passwords`

tpatel: Regarding the name, this is how we're talking about the feature in the designs so it feels like the most descriptive name, I'm open to suggestions though.

image

core/server/api/canary/users.js
tpatel: I couldn't find any docs on this property, I was thinking it would avoid the specific route from landing in the cache.
After looking through the Ghost Pro repositories, it looks like this header might clear the cache for the whole Ghost instance?

core/server/services/users.js
tpatel: good catch, the locked status should be enough!

core/server/api/canary/users.js
tpatel: The only other instances I could find were [these](https://github.com/TryGhost/Ghost/blob/f4017cc4392afee117a29ad11d47a76e3883799e/core/server/api/canary/members.js#L346
) [two](https://github.com/TryGhost/Ghost/blob/273e22032761d39aef10e0b9a90584f31b0271c9/core/server/api/canary/webhooks.js#L146) so I'm changing this to `null` .

core/server/web/api/canary/admin/routes.js
naz: Imo "all" part is not necessary for the non-user facing naming. Naming it `reset_password` leaves a door open to do things like passing a list of emails that have been compromised to this endpoint and doing a "bulk reset password". Also, worth giving a thought if we want to thing of a "password" as a separate resource that happens to belong to a user or leave it as is :thinking:

core/server/web/api/canary/admin/routes.js
```diff
@@ -75,6 +75,7 @@ module.exports = function apiRoutes() {
 
     router.put('/users/password', mw.authAdminApi, http(apiCanary.users.changePassword));
     router.put('/users/owner', mw.authAdminApi, http(apiCanary.users.transferOwnership));
+    router.put('/users/reset_all_passwords', mw.authAdminApi, http(apiCanary.users.resetAllPasswords));
```
matthanley, line 78: @tpatel imo this would make more sense as an auth endpoint like `/authentication/reset` - we already have `passwordreset` there: https://github.com/TryGhost/Ghost/blob/5ef9bdc91a36dbf46a413ce70b2f34e6e6fc495f/core/server/web/api/canary/admin/routes.js#L211

tpatel: Yes I was hesitating between the two with no strong opinion, I'm gonna move it under `/authentication` then 👍

core/server/api/canary/users.js
naz: > this header might clear the cache for the whole Ghost instance?

Yup, that's exactly what it does. We don't want to do that unless any of the content (posts, tags, etc.) has changed.

core/server/api/canary/users.js
naz: Yup. This is way better! :+1:

core/server/web/api/canary/admin/routes.js
naz: @tpatel to add to above comment I'm not sure the `PUT` is a correct method here either. I struggled to find a good reference. The best example we use often is Stipe's API. In [similar "action" situation](https://stripe.com/docs/api/setup_intents) they use `POST`:
```
  POST /v1/setup_intents/:id/confirm
  POST /v1/setup_intents/:id/cancel
```

What outside internal references did you check when deciding on the naming?

core/server/services/users.js
```diff
@@ -17,17 +17,59 @@ const path = require('path');
...
+        this.apiMail = apiMail;
+        this.apiSettings = apiSettings;
+    }
+
+    async resetAllPasswords(frameOptions) {
+        return this.models.Base.transaction(async (t) => {
+            frameOptions.transacting = t;
+
+            // Reset all passwords
+            const users = await this.models.User.findAll({...frameOptions});
```
naz, line 56: What's the difference of using `...` spread operator to just doing a `.findAll(frameOptions)` call?

core/server/services/users.js
```diff
@@ -17,17 +17,59 @@ const path = require('path');
...
+            for (const user of users) {
+                await user.save({
+                    status: 'locked' // Prevent signins before password reset
+                }, {...frameOptions});
+            }
+
+            // Delete all sessions
```
naz, line 63: I don't think this whole block belongs here. Anything to do with session data processing imo should be in the [SessionStore](https://github.com/TryGhost/Ghost/blob/44053553f5baf7a2a1e2f688ceef235afb9e3b82/core/server/services/auth/session/store.js). 

The session store base class we extend in Ghost has exactly the method that is mean to remove all the session data - https://github.com/expressjs/session/tree/0048bcac451ad867299d404aca94c79cc8bc751d#storeclearcallback.

Sidenote, maybe worth looking into truncating the table instead of doing a findAll-> iterate delete dance :thinking:

core/server/data/schema/fixtures/fixtures.json
```diff
@@ -486,6 +486,11 @@
                     "name": "Delete snippets",
                     "action_type": "destroy",
                     "object_type": "snippet"
+                },
+                {
+                    "name": "Reset all passwords",
+                    "action_type": "resetAllPasswords",
```
naz, line 492: I'd give the permission naming a hard think here. What alternative names have been considered here?

naz (changes requested):

core/server/services/users.js
tpatel: The two reasons I had chosen the findAll->iterate are:
1. It's only touching the staff users, so it shouldn't be more than a couple hundreds sessions
2. This way it triggers hooks if we want to use them in the future.

I'm fine with doing a truncate though, it's optimizing performance over the hooks feature which is ok.

Regarding the code placement, I've moved everything to the express session object but I had to export the express out of the session service, so let me know what you think about it.

core/server/data/schema/fixtures/fixtures.json
tpatel: Here is the full list of what's used as an `action_type` value: `"activate", "add", "assign", "auth", "backupContent", "browse", "deleteAllContent", "destroy", "download", "edit", "exportContent", "generate", "importContent", "publish", "read", "retry", "send", "sendTestEmail", "upload"`. With this context I feel it's ok to name the action that resets all passwords `resetAllPasswords` 😅 
Do you have an example of what would be more fitting?

naz: Yup! Now it totally makes sense, specially that we have `deleteAllContent` :+1:

core/server/models/session.js
```diff
@@ -48,6 +48,12 @@ const Session = ghostBookshelf.Model.extend({
             });
     },
 
+    destroyAll() {
+        // As of 2021, it's the recommended way to truncate a table
+        // https://github.com/bookshelf/bookshelf/issues/1740
```
naz, line 53: Noice :ok:

core/server/services/auth/session/express-session.js
```diff
@@ -22,7 +23,8 @@ const expressSessionMiddleware = session({
     }
 });
 
-exports.getSession = async function getSession(req, res) {
+module.exports = {};
```
naz, line 26: Is this line needed? :thinking:

core/server/services/auth/session/index.js
```diff
@@ -49,3 +49,4 @@ module.exports.createSessionFromToken = sessionFromToken({
 });
 
 module.exports.sessionService = sessionService;
+module.exports.sessionStore = expressSession.sessionStore;
```
naz, line 52: I was thinking about this bit and I think a cleaner approach here would be exposing a separate method on the `sessionsService` like `clear()` that would call the `store.clear()`. "Store" is an internal implementation detail of this module, don't think controllers should know about it. The other upside is having way smaller exposed API surface with a method vs whole Store instance. Wdyt?

core/server/web/api/canary/admin/routes.js
tpatel: I had started by looking at the internal usage of `POST`/`PUT` in Ghost to mimic the choices made by the team:

```
PUT /users/password => changePassword
PUT /authentication/passwordreset => resetPassword
POST /authentication/passwordreset => generateResetToken
```

As we're modifying the state of users, sending a password reset email AND deleting sessions, it's hard to know which one fits best and I arbitrarily chose `PUT`.

Now, reading [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9) seems to point towards using `POST` instead. The main reason is that `PUT` is mentioned as being idempotent vs `POST` not having to be idempotent.

I've updated it ✅

core/server/services/auth/session/index.js
tpatel: I didn't think about this approach, I'm looking into it now 👀

core/server/services/auth/session/express-session.js
tpatel: nope, TIL 😄

core/server/services/auth/session/index.js
tpatel: After a couple of attempts I think it's cleaner not to bring the concept of express-session into our already-extracted session library.
So I've simply reduced the surface area of what's exposed by only exposing a promisified version of `store.clear()`, I think it's the best of both worlds

core/server/services/users.js
tpatel: I've fixed it, I guess I was expecting to add more properties to the findAll configuration object but didn't end up doing so.

core/server/api/canary/authentication.js
```diff
@@ -190,5 +197,14 @@ module.exports = {
                     return models.Invite.findOne({email: email, status: 'sent'}, frame.options);
                 });
         }
+    },
+
+    resetAllPasswords: {
+        // validation: false,
```
naz, line 203: @tpatel what's the reason behind this being commented out?

core/server/web/api/canary/admin/routes.js
naz: Yup. Makes sense :+1:

core/server/services/auth/session/index.js
naz: Love it!

core/server/api/canary/authentication.js
tpatel: It's definitely a leftover from early attempts at learning the Ghost framework, I've removed this because uncommenting would skip the validation step and there is no data to validate!

core/server/data/migrations/versions/4.7/03-add-reset-all-passwords-permission.js
```diff
@@ -0,0 +1,11 @@
+const {
```
naz, line 1: @tpatel this file needs a bump as the next version is 4.8 I think ;)

naz (approved):
@tpatel I think it's good to go just few minor tweaks needed: 
1. Change the migration folder (and maybe the number?)
2. Make sure to write a test for this endpoint! This can be done post merge, but we have to have at least basic test coverage :wink:

tpatel:
I've updated the migration (most likely I'll have to update it again before merging) and I've added one regression test!

I'm thinking it's not a good thing to have it in the regression test folder though, I might have to convert that to a set of unit tests instead.

daniellockyer (approved):

allouis (approved):

PR #13009, opened 2021-06-04
🏗 Added "labs" setting enabling feature flags

naz:
refs https://github.com/TryGhost/Team/issues/757
refs https://github.com/TryGhost/Team/issues/332
refs https://github.com/TryGhost/Ghost/commit/ea6d656457eb9088c5a121fc8b3a4920a4e1b903

- We have a need a quick way to add features behind flags. The old way of "labs" is the quickest way to achieve this. It has ready tooling around it and well understood pitfalls. This change reintroduces "labs" group & key in settings table in the same shape it used to be (see reffed commit)
- Next step will be introducing very basic guard rails to protect from pitfalls previous implementation of "labs" had. This will include an allow-list based input validation for lab's object's data
-  The labs being an "object" type is an EXCEPTION. Even though it's an anti-pattern we aim to move away from, for now it's the lowest impact solution that will unblock the use of flags in the system. A proper solution will come at some point.

This PR contains a [migration](https://github.com/TryGhost/Ghost/pull/13009/commits/caf9761b0a83af4b3f35fb7ac8cb66f5a075013e) modifying settings table would appreciate a review :pray:  Above are the details of why and what is being changed :wink: 

### TODO:
- [x] remove `labs` write support from non-canary endpoints as per https://github.com/TryGhost/Ghost/pull/13009#discussion_r646412031

### DIFF
--- a/core/server/api/v2/utils/serializers/input/settings.js
+++ b/core/server/api/v2/utils/serializers/input/settings.js
@@ -2,7 +2,6 @@ const _ = require('lodash');
 const url = require('./utils/url');
 const typeGroupMapper = require('../../../../shared/serializers/input/utils/settings-filter-type-group-mapper');
 const settingsCache = require('../../../../../services/settings/cache');
-const {WRITABLE_KEYS_ALLOWLIST} = require('../../../../../services/labs');
 
 const DEPRECATED_SETTINGS = [
     'bulk_email_settings',
@@ -92,11 +91,12 @@ module.exports = {
 
         const settings = settingsCache.getAll();
 
-        // Ignore and drop all values with Read-only flag
         frame.data.settings = frame.data.settings.filter((setting) => {
             const settingFlagsStr = settings[setting.key] ? settings[setting.key].flags : '';
             const settingFlagsArr = settingFlagsStr ? settingFlagsStr.split(',') : [];
-            return !settingFlagsArr.includes('RO');
+
+            // Ignore and drop all values with Read-only flag AND 'labs' setting
+            return !settingFlagsArr.includes('RO') && (setting.key !== 'labs');
         });
 
         frame.data.settings.push(...getMappedDeprecatedSettings(frame.data.settings));
@@ -137,19 +137,6 @@ module.exports = {
 
             if (setting.key === 'unsplash') {
                 setting.value = JSON.parse(setting.value).isActive;
-            }
-
-            if (setting.key === 'labs') {
-                const inputLabsValue = JSON.parse(setting.value);
-                const filteredLabsValue = {};
-
-                for (const value in inputLabsValue) {
-                    if (WRITABLE_KEYS_ALLOWLIST.includes(value)) {
-                        filteredLabsValue[value] = inputLabsValue[value];
-                    }
-                }
-
-                setting.value = JSON.stringify(filteredLabsValue);
             }
 
             setting = url.forSetting(setting);
--- a/core/server/api/v2/utils/serializers/input/settings.js
+++ b/core/server/api/v2/utils/serializers/input/settings.js
@@ -2,6 +2,7 @@ const _ = require('lodash');
 const url = require('./utils/url');
 const typeGroupMapper = require('../../../../shared/serializers/input/utils/settings-filter-type-group-mapper');
 const settingsCache = require('../../../../../services/settings/cache');
+const {WRITABLE_KEYS_ALLOWLIST} = require('../../../../../services/labs');
 
 const DEPRECATED_SETTINGS = [
     'bulk_email_settings',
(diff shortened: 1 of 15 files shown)


core/server/services/labs.js
```diff
@@ -6,6 +6,14 @@ const i18n = require('../../shared/i18n');
 const logging = require('../../shared/logging');
 const settingsCache = require('../services/settings/cache');
 
+// NOTE: this allowlist is meant to be used to filter out any unexpected
+//       input for the "labs" setting value
+const WRITABLE_KEYS_ALLOWLIST = [
```
naz, line 11: @ErisDS I've put the allowlist in the service layer with a mindset that it's the best place to keep things central when it comes to reuse across API serialization, model, and importer layers. Was wondering if you have any thoughts on better placement of data like this?

core/server/api/v2/utils/serializers/input/settings.js
```diff
@@ -139,6 +139,19 @@ module.exports = {
                 setting.value = JSON.parse(setting.value).isActive;
             }
 
+            if (setting.key === 'labs') {
```
kevinansfield, line 142: Do we have a use case for allowing labs to be writable by any API version other than `canary`? Wondering if it makes sense to keep labs as a canary-only feature so there's less maintenance overhead

kevinansfield: That said, having the allowed keys list live in the service keeps maintenance minimal for older API versions 🤔

core/server/data/importer/importers/data/settings.js
```diff
@@ -204,6 +207,21 @@ class SettingsImporter extends BaseImporter {
...
+
+                for (const flag in inputLabsValue) {
+                    if (WRITABLE_KEYS_ALLOWLIST.includes(flag)) {
+                        filteredLabsValue[flag] = inputLabsValue[flag];
+                    }
+                }
+
+                // Overwrite the labs setting with our current defaults
+                // Ensures things that are enabled in new versions, are turned on
+                obj.value = JSON.stringify(_.assign({}, filteredLabsValue, labsDefaults));
```
kevinansfield, line 222: This would have the effect of re-enabling labs features that have been explicitly disabled, is that intentional?

core/server/api/v2/utils/serializers/input/settings.js
naz: This is an awesome point! "labs" is a synonym of "canary" and doesn't make sense to allow for it in the stable APIs. Additional upside - less code to maintain.

Reading is a different topic and totally makes sense to have labs available as readable property across all the clients.

The only reason why I can think of for leaving labs as a writable option is a convenience for the API client where they are pegged to a certain stable version and would need to reconfigure SDK (whatever way they access the API) with canary endpoint. But this is an edge case I think we can address IF there's anybody even complaining about it :thinking:

core/server/data/importer/importers/data/settings.js
naz: Yup. This is how the rules have been set up in the past as well - https://github.com/TryGhost/Ghost/commit/ea6d656457eb9088c5a121fc8b3a4920a4e1b903#diff-175e0a6e8ad6003197e8359cb8a50b37dfa14cd6eab87b67840c63cd1399386aL189. 

Maybe we'd want to change it, but for now I'm trying to keep the restoration to the minimum with minor improvements to avoid common pitfalls we had in the past.

kevinansfield (approved):

core/server/api/v2/utils/serializers/input/settings.js
ErisDS: Agreed I wouldn't expect any client other than admin to be writing to this setting.

core/server/services/labs.js
ErisDS: This seems fine to me for now. Maybe a better location will come to mind when I start to use it in anger - but this seems sensible.

core/server/data/migrations/versions/4.7/03-add-labs-setting.js
```diff
@@ -0,0 +1,44 @@
...
+
+const MIGRATION_USER = 1;
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const labsExists = await knex('settings')
+            .where('key', '=', 'labs')
+            .first();
+
+        logging.info(labsExists);
```
ErisDS, line 13: I don't think we should leave this here, the migration output should be easily readable

ErisDS (approved):
LGTM!

core/server/data/migrations/versions/4.7/03-add-labs-setting.js
naz: This is deffo a debugging leftover! Gonna :fire: now. Thanks for spotting!

rshbhgrg (approved):
LGTM

PR #13030, opened 2021-06-09
Added `posts_meta.feature_image_{alt,caption}` columns

kevinansfield:
refs https://github.com/TryGhost/Team/issues/770

We want post feature image functionality to better match what's available inside the editor, to do that we'll need somewhere to store alt and caption meta data. `posts_meta` chosen because even though we want to make this generic for other tables in the future those tables also have a `feature_image` (or closely related) field.

- updated schema with new columns
- added migration to create columns
- cleaned new columns from API output
  - not output on v2/v3
  - conditionally output on v4/canary output based on labs flag
- bumped `@tryghost/admin-api-schema` to allow new columns through in canary API requests
  - silently clean properties from input when labs flag is disabled
  - updated acceptance tests so they fail if `admin-api-schema` is not letting the new fields through


core/server/api/canary/utils/validators/input/posts.js
```diff
@@ -28,15 +29,38 @@ const validateVisibility = async function (frame) {
     }
 };
 
+const validateLabsFeatures = async function (frame) {
+    if (!frame.data.posts || !frame.data.posts[0]) {
+        return Promise.resolve();
+    }
+
+    if (!labs.isSet('featureImageMeta')) {
```
naz, line 37: Wondering if we should add a similar if block in the importer :thinking:

core/shared/i18n/translations/en.json
```diff
@@ -345,7 +345,8 @@
             "posts": {
                 "postNotFound": "Post not found.",
                 "invalidEmailRecipientFilter": "Invalid filter in email_recipient_filter param.",
-                "invalidVisibilityFilter": "Invalid filter in visibility property"
+                "invalidVisibilityFilter": "Invalid filter in visibility property",
```
naz, line 348: Worth burning these inf favor of "messages" object pattern that was [introduced recently](https://github.com/TryGhost/Utils/commit/e93164402536d8d43010bc4f62d2748254406955) :wink:

naz (approved):
The migration is :ok:  Left some non-critical comments along the way.

core/server/data/schema/schema.js
```diff
@@ -72,7 +72,9 @@ module.exports = {
         meta_title: {type: 'string', maxlength: 2000, nullable: true, validations: {isLength: {max: 300}}},
         meta_description: {type: 'string', maxlength: 2000, nullable: true, validations: {isLength: {max: 500}}},
         email_subject: {type: 'string', maxlength: 300, nullable: true},
-        frontmatter: {type: 'text', maxlength: 65535, nullable: true}
+        frontmatter: {type: 'text', maxlength: 65535, nullable: true},
+        feature_image_alt: {type: 'text', maxlength: 65535, nullable: true},
```
naz, line 76: I'm not sure if giving full `65535` characters is necessary for an "alt" text. The HTML spec doesn't put a limit on it, but the general rule that floats around is having around 125 characters to describe an image and using longdesc attribute for detailed descriptions. Do you have any references for this>

naz: This is the best I could find on the "125" number reason - https://moz.com/learn/seo/alt-text

kevinansfield: Good point. 125 sounds reasonable and looks widely referenced. The only other reference to a specific length I could find is a [WCAG 2.0 test suite for alt text](https://www.w3.org/WAI/GL/WCAG20/tests/test3.html) which will fail if text is >100chars but allows for longer if there's a good reason. In the [WAI images tutorial](https://www.w3.org/WAI/tutorials/images/complex/) they have an `alt` text with 143 chars but it looks like an outlier with alternative ways to reference a longer description.

Initially I went with `text` and the larger fields to avoid problems with row length. Do we have any issues around that with the `posts_meta` table?

core/server/api/canary/utils/validators/input/posts.js
kevinansfield: Couldn't that result in data loss? I don't see any harm importing the fields if the feature can be re-enabled or will be made GA in the future.

kevinansfield:
Updated to remove new fields from v2/v3 output and silently clean fields from input on v4/canary when the labs feature is disabled rather than erroring.

core/server/data/schema/schema.js
kevinansfield: Changed to a `191` length string column with max length validation of `125` chars

test/regression/api/canary/content/posts_spec.js
```diff
@@ -264,11 +264,6 @@ describe('api/canary/content/posts', function () {
         let paidPost;
         let membersPostWithPaywallCard;
 
-        before(function () {
-            // NOTE: ideally this would be set through Admin API request not a stub
-            sinon.stub(labs, 'isSet').withArgs('members').returns(true);
-        });
-
```
kevinansfield, line 271: Removed this as it's no longer needed (`members` is always `true`) and was interfering with `labs.isSet()` for other features.

daniellockyer (approved):

PR #13033, opened 2021-06-09
Updated package-json lib to not need DI

ErisDS:
- The underlying package-json package has had i18n ripped out using the new tpl utility instead
- It's also then been refactored to not be a class that needs instantiating
- This means it can be required directly and its public interface methods used where needed
- This is a much nicer, neater pattern for what is a mature utility library :)


ErisDS:
opened this for testing as I had local failures

PR #13036, opened 2021-06-09
Enabled all labs flags when testing

kevinansfield:
no issue

Shows impact of new code behind labs flags through the existing acceptance/regression tests. Allows for existing tests to be updated to match new behaviour rather than requiring separate tests with individual flags enabled. Should result in minimal test updating once code reaches GA.

- adds a forced `'labs:enabled'` fixture op that edits the `labs` setting to enable all flags then restarts the settings service to pick up the new setting
- modifies labs service to not remove ALPHA_FEATURE labs settings when running in a testing environment

### DIFF
--- a/core/server/data/schema/schema.js
+++ b/core/server/data/schema/schema.js
@@ -187,7 +187,8 @@ module.exports = {
                     'array',
                     'string',
                     'number',
-                    'boolean'
+                    'boolean',
+                    'object'
                 ]]
             }
         },
--- a/core/server/services/labs.js
+++ b/core/server/services/labs.js
@@ -24,7 +24,7 @@ module.exports.getAll = () => {
     const labs = _.cloneDeep(settingsCache.get('labs')) || {};
 
     ALPHA_FEATURES.forEach((alphaKey) => {
-        if (labs[alphaKey] && !(config.get('enableDeveloperExperiments'))) {
+        if (labs[alphaKey] && !(config.get('enableDeveloperExperiments') || process.env.NODE_ENV.match(/^testing/))) {
             delete labs[alphaKey];
         }
     });
--- a/test/regression/api/canary/admin/settings_spec.js
+++ b/test/regression/api/canary/admin/settings_spec.js
@@ -677,8 +677,10 @@ describe('Settings API (canary)', function () {
 
             jsonResponse.settings.length.should.eql(1);
             testUtils.API.checkResponseValue(jsonResponse.settings[0], ['id', 'group', 'key', 'value', 'type', 'flags', 'created_at', 'updated_at']);
+
+            const jsonObjectRegex = /^\{.*\}$/; // '{...}'
             jsonResponse.settings[0].key.should.eql('labs');
-            jsonResponse.settings[0].value.should.eql(JSON.stringify({}));
+            jsonResponse.settings[0].value.should.match(jsonObjectRegex);
         });
 
         it('Can read deprecated default_locale', function (done) {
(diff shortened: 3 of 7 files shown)


kevinansfield:
Approach isn't working when setting doesn't already exist - re-thinking.

core/server/data/schema/schema.js
```diff
@@ -187,7 +187,8 @@ module.exports = {
                     'array',
                     'string',
                     'number',
-                    'boolean'
+                    'boolean',
+                    'object'
```
kevinansfield, line 191: @naz this was needed to be able to add the labs setting using `models.Settings.add`. Noted the type wasn't added back in here when the labs setting was re-added with the `object` type - was that intentional?

kevinansfield:
[Example test where this was useful](https://github.com/TryGhost/Ghost/pull/13030/files#diff-2a4754aeb645c949edac3dbd0fd723352fd56373f32301717d2d2283412d98ae) - written as if the feature is in GA already, no need for test updates when flag becomes GA, will trigger an error if the feature behaviour changes or is reverted

core/server/data/schema/schema.js
naz: Not intentional. Good spot :+1: This makes me wonder if there's some test missing or some other way to check if the types listed here are in use anywhere :thinking:

test/regression/api/canary/admin/settings_spec.js
```diff
@@ -678,7 +678,7 @@ describe('Settings API (canary)', function () {
             jsonResponse.settings.length.should.eql(1);
             testUtils.API.checkResponseValue(jsonResponse.settings[0], ['id', 'group', 'key', 'value', 'type', 'flags', 'created_at', 'updated_at']);
             jsonResponse.settings[0].key.should.eql('labs');
-            jsonResponse.settings[0].value.should.eql(JSON.stringify({}));
+            jsonResponse.settings[0].value.should.match(/^\{.*\}$/);
```
naz, line 681: Why was this change needed?

kevinansfield: The value was populated with all of the labs settings so we couldn't do an `===` on `'{}'`. As the labs settings will change over time I figured it would be less of a headache to expect an object-like JSON string rather than a direct match.

core/server/services/labs.js
```diff
@@ -24,7 +24,7 @@ module.exports.getAll = () => {
     const labs = _.cloneDeep(settingsCache.get('labs')) || {};
 
     ALPHA_FEATURES.forEach((alphaKey) => {
-        if (labs[alphaKey] && !(config.get('enableDeveloperExperiments'))) {
+        if (labs[alphaKey] && !(config.get('enableDeveloperExperiments') || process.env.NODE_ENV.match(/^testing/))) {
```
kevinansfield, line 27: Perhaps there's a better way of avoiding this when testing? I was confused as to why the PR using this change was failing, I hadn't realised in CI we also use `testing-mysql` as the NODE_ENV

test/regression/api/canary/admin/settings_spec.js
naz: Makes sense :+1:  Let's extract this regex into a vairable. Find it hard looking at event the simplest regexes without breaking out of the context :sweat_smile:

kevinansfield: 👍 done

core/server/services/labs.js
kevinansfield: Going to merge with this in place as work is building up behind this PR. We can come back and change things later on if needed

PR #13052, opened 2021-06-15
Switch to @trghost/validator, remove validator

ErisDS:
@daniellockyer which one of us broke it? 😂


ErisDS:
Ok that was weird

The test

[failure] custom excerpt soft limit reached in test/regression/models/model_posts_spec.js was hanging for me, but a `yarn cache clean` seems to have fixed it

PR #13055, opened 2021-06-16
Added new `portal_products` setting

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/768

- `portal_products` stores list of products available in Portal, chosen via Admin
- adds new `portal_products` setting to default settings 
- adds migration to populate `portal_products` with current product so its available by default
- inserts or updates setting - its possible the setting doesn't exist as CLI and Pro run migrations before settings are initialized
- update tests


core/server/data/migrations/versions/4.9/01-add-default-product-portal-products.js
```diff
@@ -0,0 +1,56 @@
...
+            .where('key', 'portal_products')
+            .select('value')
+            .first();
+
+        if (!portalProductSetting) {
+            logging.info('Skipping adding default product to portal_products, setting does not exist');
+            return;
+        }
+
+        if (portalProductSetting.length > 0) {
```
allouis, line 27: I think you need to `JSON.parse` the value first, otherwise you're checking the length of a string!

core/server/data/migrations/versions/4.9/01-add-default-product-portal-products.js
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        const defaultProduct = products[0];
+
+        const portalProductSetting = await knex('settings')
+            .where('key', 'portal_products')
+            .select('value')
+            .first();
+
+        if (!portalProductSetting) {
```
allouis, line 22: Is this a case that is possible?

rishabhgrg: Unlikely, but saw couple of previous migrations checking for setting existence so added the condition

core/server/data/migrations/versions/4.9/01-add-default-product-portal-products.js
rishabhgrg: Good catch, updated! ✅

core/server/data/migrations/versions/4.9/01-add-default-product-portal-products.js
rishabhgrg: Removed the redundant checks.

core/server/data/migrations/versions/4.9/01-add-default-product-portal-products.js
```diff
@@ -0,0 +1,49 @@
...
+        const products = await knex
+            .select('id')
+            .from('products');
+
+        if (products.length === 0) {
+            logging.warn(`Skipping adding default product to portal_products, no default product exists`);
+            return;
+        }
+
+        const defaultProduct = products[0];
```
kevinansfield, line 15: Is it guaranteed that there's only 1 product in the db at this point?

core/server/data/migrations/versions/4.9/01-add-default-product-portal-products.js
```diff
@@ -0,0 +1,49 @@
+const {createTransactionalMigration} = require('../../utils');
```
kevinansfield, line 1: Is there a reason this is 4.9 instead of 4.8?

rishabhgrg: None other than me totally missing 4.8 is the next release but just has new migrations 🤦 Updating

core/server/data/migrations/versions/4.9/01-add-default-product-portal-products.js
rishabhgrg: Ideally yes, we never added support for adding more than 1 product and we are treating the first(and only) product as the default everywhere

core/server/data/migrations/versions/4.9/01-add-default-product-portal-products.js
rishabhgrg: Updated, thanks for catching 😄

core/server/data/migrations/versions/4.9/01-add-default-product-portal-products.js
```diff
@@ -19,17 +19,8 @@ module.exports = createTransactionalMigration(
...
             .first();
 
-        if (!portalProductsSetting) {
-            logging.info('Skipping adding default product to portal_products, setting does not exist');
-            return;
-        }
-
         const portalProducts = JSON.parse(portalProductsSetting.value);
 
-        if (portalProducts.length > 0) {
```
allouis: :thinking: I'm not sure that this check is redundant, without it I don't think the migration is idempotent, as if it were to run twice we would go from `[] -> [productId] -> [productId, productId]`

rishabhgrg: 🤔 Hmm good point, have added this back now, thanks.

kevinansfield (approved):

core/server/data/migrations/versions/4.8/03-add-default-product-portal-products.js
```diff
@@ -0,0 +1,69 @@
...
+        if (!portalProductSetting) {
+            logging.info('Skipping revert of portal_products setting, setting does not exist');
+            return;
+        }
+
+        logging.info(`Reverting portal_products setting to empty array`);
+
+        await knex('settings')
+            .where('key', 'portal_products')
+            .update({value: JSON.stringify([])});
```
naz, line 67: Why don't we do deletion of the record here? The expected state of pre 4.8.0 is that there is no record at all, right? :thinking: 

DISCLAIMER: NOT A STRONG OPINION! :joy:

naz (approved):
The migration is good. Added a small comment that might simplify the logic a little. It s not crucial for shipping this change!

core/server/data/migrations/versions/4.8/03-add-default-product-portal-products.js
rishabhgrg: 😄 Yeah thats a good alternate for `down`, originally since I wasn't sure if we need to create the setting here I had added it as a reset of value than reset of setting. I'll definitely go with the deletion route in such cases in future, but will leave it for now.

daniellockyer (approved):

temabu:
As an Owner the portal settings have changed? Now all I see as Admin/Owner is "Membership" and "Email Newsletter".  What I no longer see (with 4.5.0) is "Access", "Products", "Portal", and "Payments".  So how would I setup Stripe now?  Where have these options gone?

allouis:
@temabu These are all under the "Membership" settings now :relaxed: You can see them at https://yoursite.com/ghost/#/settings/members

temabu:
Thanks.  Some of it is there.  Under "Membership" I can no longer change the name of the subscription option, e.g. "Free" is hardcoded and there is no option to set a description.  Is this a design decision?  For instance, I prefer the term "Complimentary" rather than "Free".  Same feedback on "Premium".
Also, the new "Connect with Stripe" popup is modal and the cancel option (the "X") doesnt work.  nor does escape...so no way of getting out of this if you click on it.
I like the menu consolidation but the 'issues' are a tad problematic.

temabu:
..on the "Connect with Stripe" comment - of course one way to get out of it is to complete the Stripe sign up... :)

PR #13074, opened 2021-06-18
✨ Added additional newsletter customisation settings

kevinansfield:
refs https://github.com/TryGhost/Team/issues/793

New settings added for newsletter customisation options:

- `newsletter_header_image` - `null/"$url"`
- `newsletter_show_header_icon` - `"true/false"`
- `newsletter_show_header_title` - `"true/false"`
- `newsletter_title_alignment` - `"center/left"`
- `newsletter_title_font_category` - `"serif/sans_serif"`
- `newsletter_show_feature_image` - `"true/false"`

`newsletter_show_header` has been dropped because the same functionality can be achieved by setting both `newsletter_show_header_icon` and `newsletter_show_header_title` to `false`

---

- migration to convert and delete `newsletter_show_header` setting
- removed `newsletter_show_header` from default settings to ensure it doesn't get re-created
- replaced main labs template and template settings generation with the labs template
- deleted labs template


core/server/data/migrations/versions/4.8/04-migrate-show-newsletter-header-setting.js
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            await connection('settings')
+                .insert({
+                    id: ObjectId().toHexString(),
+                    key: 'newsletter_show_header',
+                    group: 'newsletter',
+                    type: 'boolean',
+                    value: showHeader,
+                    created_at: moment().toDate(),
```
daniellockyer, line 86: This inserts the wrong thing into the DB for SQLite for me:
* inserted: `1624267454065`
* other values: `2021-06-17 16:20:03`

core/server/data/migrations/versions/4.8/04-migrate-show-newsletter-header-setting.js
```diff
@@ -0,0 +1,98 @@
...
+                )
+                .where(
+                    'roles_users.role_id',
+                    connection('roles').select('id').where('name', 'Owner')
+                );
+
+            let showHeader = true; // default
+
+            if (showTitle && showIcon) {
+                showHeader = showTitle.value === 'false' && showIcon.value === 'false' ? false : true;
```
daniellockyer, line 76: Should we be inserting the string versions of true/false (and above in the default)? For SQLite, it inserts `1` instead of `true` like other `boolean` type settings

daniellockyer:
Also, do we need to remove `newsletter_show_header` from the default settings? Otherwise it'll get recreated upon boot

core/server/data/migrations/versions/4.8/04-migrate-show-newsletter-header-setting.js
kevinansfield: Good catch 🎣

core/server/data/migrations/versions/4.8/04-migrate-show-newsletter-header-setting.js
kevinansfield: Ugh, OK. I'll switch to the `knex.raw('CURRENT_TIMESTAMP')` pattern although that also has the non-UTC issue.

kevinansfield: Updated

core/server/data/migrations/versions/4.8/04-migrate-show-newsletter-header-setting.js
kevinansfield: Fixed

kevinansfield:
> Also, do we need to remove newsletter_show_header from the default settings? Otherwise it'll get recreated upon boot

Good point! Removed it 👍

rshbhgrg (approved):

core/server/data/migrations/versions/4.8/04-migrate-show-newsletter-header-setting.js
```diff
@@ -0,0 +1,123 @@
...
+                .select('users.id')
+                .innerJoin(
+                    'roles_users',
+                    'users.id',
+                    'roles_users.user_id'
+                )
+                .where(
+                    'roles_users.role_id',
+                    connection('roles').select('id').where('name', 'Owner')
+                );
```
rishabhgrg, line 94: This is a good candidate for helper/util extraction in future 😄

daniellockyer (approved):

PR #13092, opened 2021-06-24
Added `benefits` & `product_benefits` tables

allouis:
refs TryGhost/Team#804

Benefits are tag-like resources which will be associated with Products.
The first iteration just requires a name for the benefit, which will be
stored as plaintext.

The associative table is used to implement the many-to-many relationship
between Products and Benefits. The `sort_order` column is needed because
a product's benefits should be orderable by an admin.


tpatel (approved):
Great stuff 🙌

daniellockyer (approved):

rshbhgrg (approved):

PR #13100, opened 2021-06-29
Added member segment to email batches

tpatel:
issue https://github.com/TryGhost/Team/issues/826

### DIFF
--- a/core/server/data/schema/schema.js
+++ b/core/server/data/schema/schema.js
@@ -576,6 +576,7 @@ module.exports = {
             defaultTo: 'pending',
             validations: {isIn: [['pending', 'submitting', 'submitted', 'failed']]}
         },
+        member_segment: {type: 'text', maxlength: 2000, nullable: true},
         created_at: {type: 'dateTime', nullable: false},
         updated_at: {type: 'dateTime', nullable: false}
     },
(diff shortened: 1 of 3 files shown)


core/server/data/schema/schema.js
```diff
@@ -576,6 +576,7 @@ module.exports = {
             defaultTo: 'pending',
             validations: {isIn: [['pending', 'submitting', 'submitted', 'failed']]}
         },
+        member_segment: {type: 'text', maxlength: 1000, nullable: true},
```
naz, line 579: Why was this specific length of `1000` chosen?

tpatel: I had the choice between small, medium and large string (https://github.com/TryGhost/Ghost/blob/ff37398311ea9c72fcad70efb78d8e90a6b37a38/core/server/data/schema/schema.js#L6) and I'm always cautious not to prevent use cases due to a too-restrictive database constraint.

Looking at a couple of examples I come up with, we could lower the limit the string to 191:
- `status:paid+email_count:>=3` (27 characters)
- `status:free+labels.slug:vip` (27 characters)

naz: do we have references to storing anything similar? I remember there was a problem at some point Kevin faced with the field being too small when storing a filter like this :thinking:

allouis (approved):

daniellockyer (approved):

core/server/data/schema/schema.js
tpatel: After talking with Kevin, I've updated the issue with the reasoning behind using 2000 characters for this field.

naz (approved):

PR #13102, opened 2021-06-30
[WIP] Content cta products

rshbhgrg:
Note: This PR is only for the purpose of experimental demos with content cta with multiple products and not intended to be the final design/direction.

### DIFF
--- a/core/frontend/helpers/content.js
+++ b/core/frontend/helpers/content.js
@@ -22,13 +22,18 @@ function restrictedCta(options) {
     _.merge(this, {
         accentColor: (options.data.site && options.data.site.accent_color)
     });
-    if (this.visibility !== 'members' && this.visibility !== 'paid' && this.visibility.includes('product:')) {
+    if (this.visibility.includes('product:')) {
         const productSlugs = this.visibility.split(',').filter(qry => qry.startsWith('product:')).map(qry => qry.replace('product:', ''));
         const productNames = productSlugs.map((slug) => {
             return options.data.products.find(product => product.slug === slug);
-        }).filter(product => !!product).map(product => product.name);
-
-        const customVisibilityText = `This post is only for subscribers on products - ${productNames.join()}`;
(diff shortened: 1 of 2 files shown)


core/frontend/helpers/content.js
```diff
@@ -18,9 +18,26 @@ const createFrame = hbs.handlebars.createFrame;
 function restrictedCta(options) {
     options = options || {};
     options.data = options.data || {};
+
     _.merge(this, {
         accentColor: (options.data.site && options.data.site.accent_color)
     });
+    if (this.visibility.includes('product:')) {
+        const productSlugs = this.visibility.split(',').filter(qry => qry.startsWith('product:')).map(qry => qry.replace('product:', ''));
```
ErisDS, line 26: If visibility _is_ going to be used to store NQL strings, we should be using our language parser to parse them

core/frontend/helpers/tpl/content-cta.hbs
```diff
@@ -7,6 +7,9 @@
         {{#has visibility="members"}}
             <h2>This post is for subscribers only</h2>
         {{/has}}
+        {{#if customVisibilityText}}
```
ErisDS, line 10: If we're introducing this property as part of the theme API, has it passed design review?

core/frontend/helpers/content.js
rishabhgrg: @ErisDS Ah yea good point! This was just a first version hacked out for demo purposes, but yeah didn't strike me to use language parser as the first thing to use here :)

core/frontend/helpers/tpl/content-cta.hbs
rishabhgrg: @ErisDS Its pending design review and don't expect the property to go out in current form for theme API. This PR was purely for demo purposes, sorry should have added a note in description clarifying(done now).

rishabhgrg:
Closing for now as the PR was an early hack on how content cta can look with multiple products.

PR #13108, opened 2021-07-01
Migrated comped members to 'comped' status

allouis:
refs https://github.com/TryGhost/Team/issues/790

**Migration 4.9/06:**
In order to track when a member was comped, as well as to differentiate
paid members from comped, we are reintroducing the 'comped' status. This
migration will updated members with a Complimentary Stripe Subscription
to a status of 'comped'. It is essentially a reversal of the 4.6
migration.

**Migration 4.9/07:**
Since version 4.6 the 'comped' status has not been used. Any members
which were given complimentary plans since then will have had a `status`
of 'paid', and therefore the corresponding members_status_events row
would have a `to_status` of 'paid'.

This migration is designed to fix these members_status_events rows by
ensuring that the last (chronologically) members_status_event row for a
comped member has a to status of 'comped'.

Unfortunately this migration loses information which makes writing a
perfect inverse migration impossible. Alternative down migrations were
considered, but these would lose further information.

### DIFF
--- a/core/server/data/migrations/versions/4.9/06-add-comped-status.js
+++ b/core/server/data/migrations/versions/4.9/06-add-comped-status.js
@@ -1,0 +1,46 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils.js');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const compedMemberIds = (await knex('members')
+            .select('members.id')
+            .innerJoin(
+                'members_stripe_customers',
+                'members.id',
+                'members_stripe_customers.member_id'
+            ).innerJoin(
+                'members_stripe_customers_subscriptions',
+                function () {
+                    this.on(
+                        'members_stripe_customers.customer_id',
+                        'members_stripe_customers_subscriptions.customer_id'
+                    ).onIn(
+                        'members_stripe_customers_subscriptions.status',
+                        ['active', 'trialing', 'past_due', 'unpaid']
+                    );
+                }
+            ).where(
+                'members_stripe_customers_subscriptions.plan_nickname',
+                '=',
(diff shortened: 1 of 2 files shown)


core/server/data/migrations/versions/4.9/06-update-comped-members-status-events.js
```diff
@@ -0,0 +1,36 @@
...
+
+        if (compedMembers.length === 0) {
+            logging.info('No comped members found - skipping migration');
+            return;
+        } else {
+            logging.info(`Found ${compedMembers.length} comped members - checking members_status_events`);
+        }
+
+        for (const member of compedMembers) {
+            const mostRecentStatusEvent = await knex('members_status_events')
```
daniellockyer, line 19: @allouis @rishabhgrg Is it possible for there to be no `members_status_events` event for a member? 🤔 I don't think so *but* I think we should add a `if (!mostRecentStatusEvent) { logging.warn('huh?'); continue; }` below

rishabhgrg: Yeah i think its pretty unlikely for them to not have one, but no harm in adding the guard i think so sounds good 👍

rshbhgrg (approved):

core/server/data/migrations/versions/4.9/06-update-comped-members-status-events.js
allouis: It's not possible for them to not have one, but I'll add the check anyway :relaxed:

daniellockyer:
@allouis You'll need to bump the migration IDs because we merged a `05-` but LGTM otherwise 🙂

daniellockyer (approved):

kevinansfield (approved):

PR #13109, opened 2021-07-02
🐛 Fixed incorrectly stored URLs after migration from 3.x to 4.6.1-4.8.4

kevinansfield:
refs https://github.com/TryGhost/Team/issues/853

A refactor of `urlUtils` usage in 4.6.1 left a buggy 4.0 migration that did not transform URLs inside of mobiledoc cards. Anyone upgrading from 3.x to 4.6.1-4.8.4 would end up with inconsistent URL formats and potentially broken images.

- fixed 4.0 migration by passing our mobiledoc cards list in when transforming mobiledoc urls
- added a new migration that re-applies the missed URL transforms and content re-generation for any site that did a 3.x upgrade to a buggy 4.x version

### DIFF
--- a/core/server/data/migrations/versions/4.9/05-fix-missed-mobiledoc-url-transforms.js
+++ b/core/server/data/migrations/versions/4.9/05-fix-missed-mobiledoc-url-transforms.js
@@ -1,0 +1,86 @@
+const logging = require('@tryghost/logging');
+const urlUtils = require('../../../../../shared/url-utils');
+const htmlToPlaintext = require('../../../../../shared/html-to-plaintext');
+const mobiledocLib = require('../../../../lib/mobiledoc');
+const {createTransactionalMigration} = require('../../utils');
+
+// in Ghost versions 4.6.1-4.8.4 the 4.0 migration that transfored URLs had a bug
+// that meant urls inside cards in mobiledoc content was not being transformed
+//
+// if the migrations table indicates an upgrade was made from 3.x to 4.6-4.8 then
+// we'll re-run the transforms against post.mobiledoc and re-generate the html
+// and plaintext contents
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const badVersionUsedFor40Migration = await knex('migrations')
+            .where({
+                name: '18-transform-urls-absolute-to-transform-ready.js'
+            })
+            .whereIn('currentVersion', ['4.6', '4.7', '4.8'])
+            .first();
+
+        if (!badVersionUsedFor40Migration) {
+            logging.info('Skipping transform of mobiledoc URLs - original transform was good');
+            return;
+        }
+
+        logging.info('Transforming all internal URLs in posts.{mobiledoc,html,plaintext} to transform-ready');
+
+        await knex.transaction(async (trx) => {
+            const postIdRows = await knex('posts')
+                .transacting(trx)
+                .forUpdate()
+                .select('id');
+
+            for (const postIdRow of postIdRows) {
+                const {id} = postIdRow;
+                const [post] = await knex('posts')
+                    .transacting(trx)
+                    .where({id})
+                    .select([
+                        'mobiledoc'
+                    ]);
+
+                let mobiledoc;
+                let html;
+
+                try {
+                    mobiledoc = urlUtils.mobiledocToTransformReady(post.mobiledoc, {cardTransformers: mobiledocLib.cards});
+
+                    if (!mobiledoc) {
+                        logging.warn(`No mobiledoc for ${id}. Skipping.`);
+                        continue;
+                    }
+                } catch (err) {
+                    logging.warn(`Invalid mobiledoc JSON structure for ${id}. Skipping`);
+                    continue;
+                }
+
+                try {
+                    html = mobiledocLib.mobiledocHtmlRenderer.render(JSON.parse(mobiledoc));
+                } catch (err) {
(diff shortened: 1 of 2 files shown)


core/server/data/migrations/versions/4.9/05-fix-missed-mobiledoc-url-transforms.js
```diff
@@ -0,0 +1,73 @@
...
+    async function up(knex) {
+        const versionUsedFor40Migration = await knex('migrations')
+            .where({
+                name: '18-transform-urls-absolute-to-transform-ready.js',
+                version: '4.0'
+            })
+            .select('currentVersion')
+            .first();
+
+        // migrations table at time of writing only stores major.minor so we need to coerce
```
daniellockyer, line 25: @kevinansfield Is there a particular reason for doing this over `.whereIn('currentVersion', ['4.6', '4.7', '4.8'])` above?

core/server/data/migrations/versions/4.9/05-fix-missed-mobiledoc-url-transforms.js
```diff
@@ -0,0 +1,73 @@
...
+        // and be liberal with the bad version range
+        const currentVersion = semver.coerce(versionUsedFor40Migration.currentVersion);
+        const badVersionRange = '4.6.0 - 4.8.4';
+
+        if (!semver.satisfies(currentVersion, badVersionRange)) {
+            logging.info('Skipping transform of mobiledoc URLs - original transform was good');
+            return;
+        }
+
+        logging.info('Transforming all internal URLs in posts.{mobiledoc,html,plaintex} to transform-ready');
```
daniellockyer, line 35: tiny nit: typo on `plaintext`

core/server/data/migrations/versions/4.9/05-fix-missed-mobiledoc-url-transforms.js
```diff
@@ -0,0 +1,73 @@
...
+            for (const postIdRow of postIdRows) {
+                const {id} = postIdRow;
+                const [post] = await knex('posts')
+                    .transacting(trx)
+                    .where({id})
+                    .select([
+                        'mobiledoc'
+                    ]);
+
+                const mobiledoc = urlUtils.mobiledocToTransformReady(post.mobiledoc, {cardTransformers: mobiledocLib.cards});
```
daniellockyer, line 52: Would we be able to wrap this in try-catches like we did for a similar [4.0 migration](https://github.com/TryGhost/Ghost/blob/fec534f83374a4c4d0ddafb84d487b71eb96d2ba/core/server/data/migrations/versions/4.0/23-regenerate-posts-html.js#L29-L37)? It saves us from crashing out if the mobiledoc is in a bad format

core/server/data/migrations/versions/4.9/05-fix-missed-mobiledoc-url-transforms.js
kevinansfield: Nope, just started off thinking in terms of full versions where server comparison would be useful then continued with that 😬 I'll switch to the simpler comparison!

core/server/data/migrations/versions/4.9/05-fix-missed-mobiledoc-url-transforms.js
```diff
@@ -0,0 +1,73 @@
...
+
+                let mobiledoc;
+
+                try {
+                    mobiledoc = urlUtils.mobiledocToTransformReady(post.mobiledoc, {cardTransformers: mobiledocLib.cards});
+                } catch (err) {
+                    return logging.warn(`Invalid JSON structure for ${id}. Skipping`);
+                }
+
+                const html = mobiledocLib.mobiledocHtmlRenderer.render(JSON.parse(mobiledoc));
```
daniellockyer, line 53: I think we should include this in a try-catch too. If my mobiledoc is just `{}`, this migration crashes out with `MigrationScriptError: Unexpected Mobiledoc version "undefined"`

daniellockyer (approved):

core/server/data/migrations/versions/4.9/05-fix-missed-mobiledoc-url-transforms.js
```diff
@@ -0,0 +1,79 @@
...
+                        'mobiledoc'
+                    ]);
+
+                let mobiledoc;
+                let html;
+
+                try {
+                    mobiledoc = urlUtils.mobiledocToTransformReady(post.mobiledoc, {cardTransformers: mobiledocLib.cards});
+                } catch (err) {
+                    return logging.warn(`Invalid mobiledoc JSON structure for ${id}. Skipping`);
```
vikaspotluri123, line 51: Is the expectation here to cancel the remainder of the migration, or just the specific post? Right now, it would cancel the remainder 😄 Also applicable to the next block

```suggestion
                    logging.warn(`Invalid mobiledoc JSON structure for ${id}. Skipping`);
                    continue;
```

kevinansfield: Good catch @vikaspotluri123! 🎣 The 4.0 migration had the same issue 🙈

daniellockyer: 🤦🏻‍♂️ I've caught the same thing multiple times before and missed it here 😅

PR #13120, opened 2021-07-05
🐛 Fixed potential for partial content re-generation in 4.0 migrations

kevinansfield:
no issue

- incorrect syntax was used in the error handlers inside of the `for` loop, by using `return` when logging the whole for-loop was aborted whereas we want to log and continue processing the rest of the items


daniellockyer (approved):
😅

PR #13126, opened 2021-07-07
Updated WEBHOOK_SECRET check to warn rather than error

allouis:
refs https://github.com/TryGhost/Team/issues/841

When using our development tooling Ghost should always start, instead of
exiting with an error. This check for the WEBHOOK_SECRET env var was the
primary cause of Ghost erroring in development, so it's been switched
with a warning.

### DIFF
--- a/core/server/services/members/service.js
+++ b/core/server/services/members/service.js
@@ -68,7 +68,8 @@ const membersService = {
 
         if (env !== 'production') {
             if (!process.env.WEBHOOK_SECRET && membersConfig.isStripeConnected()) {
-                throw new Error('Cannot use remote webhooks in development. Please restart in production mode or see https://ghost.org/docs/webhooks/#stripe-webhooks for developing with Stripe.');
+                process.env.WEBHOOK_SECRET = 'DEFAULT_WEBHOOK_SECRET';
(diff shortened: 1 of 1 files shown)


allouis:
@ErisDS Does this make sense to do in `Gruntfile.js`? And do we need to handle this for `yarn start` too?

ErisDS:
Could it not live in overrides under an `if env == development` flag?

Trying to kill the Gruntfile 😉

allouis:
🤔 I thought overrides was only when we run tests? But maybe I've missed something?

> Trying to kill the Gruntfile 😉

Noice 😄 I can look at overrides, or I can update where we actually make the check and add the default there!

ErisDS:
I meant the [server overrides](https://github.com/TryGhost/Ghost/blob/main/core/server/overrides.js) not the test overrides! 

But TBH the better solution would be for this to be a default value in the development config? then you can use `config.get()`.. but then that only supports lowercase env vars :(  sadface :( 

Could put a config-level override in place for that though....


But other wise either server level overrides or in-place where its used work

allouis:
I've switched this back to adding the default in-place as that was the simplest, and IMO easiest to maintain going forward as all our checks are together!

PR #13137, opened 2021-07-12
Lazily instantiated express-session middleware

allouis:
refs https://github.com/TryGhost/Team/issues/756

When running the tests it was possible for this middleware to be
instantiated before the settings cache, resulting in an undefined
'session_secret' setting being passed. This would cause tests to fail.

Tracking this down proved difficult, so the fix was made here, by
instantiating the express-session middleware only once a request needs
to use it, we can be confident that Ghost has completely started.


allouis:
PR to double check CI results

PR #13140, opened 2021-07-14
Removed use of `new Error()` in migrations

naz:
:building_construction: @daniellockyer not a big change, just following our review principles ;)


daniellockyer:
@naz I think this is OK but I wonder if we should move to throwing a [`MigrationScriptError`](https://github.com/TryGhost/knex-migrator/blob/478a4fe901ac6b08d9206d4d824be836e7aa0340/lib/errors.js#L24) that's currently present in knex-migrator at some point in the future

daniellockyer (approved):

PR #13157, opened 2021-07-17
change meta_title and meta_description into TEXT

Wqrld:
Hi, mariaDB limits the max row size to 8k which ghost went over during a (recent) update. Having such big varchar's is seen as a bad practice either way. After changing the columns to text instead of huge varchars ghost is able to start again.

reference thread: https://forum.ghost.org/t/update-from-4-9-3-to-4-9-4-went-wrong/24240

(first time contributing to ghost, so let me know if this is the right file to edit ;) )

Error before fix:
> OuterError: alter table posts_meta add feature_image_caption text null - ER_TOO_BIG_ROWSIZE: Row size too large. The maximum row size for the used table type, not


Wqrld:
> Hi @Wqrld thanks for this PR and sorry about the delay in responding.
> 
> Please see our MySQL documentation on how to resolve this issue: https://ghost.org/docs/faq/troubleshooting-mysql-database/
> 
> This is a configuration issue, rather than needing fixing in the codebase.

As this affects more people and is easily fixed in the codebase i'd argue that it's worth thinking about. But if you already have docs that's a good step forward!

PR #13163, opened 2021-07-19
Exposed Members pub keys on /members/.well-known/

allouis:
refs https://github.com/TryGhost/Team/issues/664

The new WellKnownController and middleware handles exposing a JSON Web
Key Set for us.

In order to serve the keys on /members/.well-known/jwks.json without a
trailing slash, we must mount the wellKnown middleware before the
frontend.

### DIFF
--- a/core/server/web/site/app.js
+++ b/core/server/web/site/app.js
@@ -15,7 +15,7 @@ const sitemapHandler = require('../../../frontend/services/sitemap/handler');
 const appService = require('../../../frontend/services/apps');
 const themeEngine = require('../../../frontend/services/theme-engine');
 const themeMiddleware = themeEngine.middleware;
-const membersMiddleware = require('../../services/members').middleware;
+const membersService = require('../../services/members');
 const siteRoutes = require('./routes');
 const shared = require('../shared');
 const mw = require('./middleware');
@@ -118,7 +118,8 @@ module.exports = function setupSiteApp(options = {}) {
     debug('Helpers done');
 
     // Global handling for member session, ensures a member is logged in to the frontend
-    siteApp.use(membersMiddleware.loadMemberSession);
+    siteApp.use(membersService.middleware.loadMemberSession);
+    siteApp.use('/members/.well-known', (req, res, next) => membersService.api.middleware.wellKnown(req, res, next));
 
     // Theme middleware
     // This should happen AFTER any shared assets are served, as it only changes things to do with templates
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "@tryghost/limit-service": "0.6.1",
     "@tryghost/logging": "0.1.4",
     "@tryghost/magic-link": "1.0.7",
-    "@tryghost/members-api": "1.21.0",
+    "@tryghost/members-api": "1.22.0",
     "@tryghost/members-csv": "1.1.2",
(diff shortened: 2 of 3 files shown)


core/server/web/site/app.js
```diff
@@ -118,7 +118,8 @@ module.exports = function setupSiteApp(options = {}) {
     debug('Helpers done');
 
     // Global handling for member session, ensures a member is logged in to the frontend
-    siteApp.use(membersMiddleware.loadMemberSession);
+    siteApp.use(membersService.middleware.loadMemberSession);
+    siteApp.use('/members/.well-known', (req, res, next) => membersService.api.middleware.wellKnown(req, res, next));
```
allouis, line 122: @naz What do you think about mounting this here? The reason it's not in web/members/app.js is because we don't want the `connect-slashes` middleware on it

naz: What's happening with the other route mounted at `/ghost/.well-known` or is this a separate member-specific implementation?

naz: Is the [members/app](https://github.com/TryGhost/Ghost/blob/8425fb480d3975d9bdac4ad9a8b241ebeea9df15/core/server/web/members/app.js#L12) a bad place to put this route in? 

P.S. Damn.. we have quite a maze figuring out the tree of routes and which express app+middlewares are mounted where :sweat_smile:

allouis: > What's happening with the other route mounted at /ghost/.well-known or is this a separate member-specific implementation?

This is a members specific one

> Is the members/app a bad place to put this route in?

If we put it there we have to somehow disable the pretty-url middleware for just this route, which I _think_ means removing it from the siteApp and manually adding it to any app/route that wants/needs it!

> P.S. Damn.. we have quite a maze figuring out the tree of routes and which express app+middlewares are mounted where 😅

Yup 😬 Reminds me how much I like explicitness!

ErisDS: I don't think there's anything wrong with the location of the mount, but this is exactly the kind of detail that's useful to record in a code comment as well as in the commit message (which looks great btw).

_However_ 😬 this code pretty much duplicates the well-known middleware/controller thing we already have in the Ghost codebase deep in the members codebase, which really isn't ideal. I don't really want to block this now but we _have_ to start delivering smaller shared modules for members code, otherwise we're just walking into all the same traps the Ghost codebase is stuck in.

allouis: Cool thanks!

Agree on not blocking this, but just so that I'm clear, are you suggesting we should have created a "well-known" package which would be reused for both Members and Ghost?

ErisDS: Yar - that's exactly what I'm saying. In my head it. would be `Utils/mw-wellknown` -> `@tryghost/mw-wellknown because prefixing express middleware/controllers with `mw-` is a pattern we want to establish.

Then I'd assume needing to pass in the function to get keys:

```
const mw-wellknown = require('@tryghost/mw-wellknown');
app.use(path, mw-wellknown({getKeysFn})
```

Unless the token service / keystore was also a shared component, and you could just pass some key for which key in the keystore you want to serve, because I don't see a problem with tightly coupling a keystore component and wellknown middleware as they are fundamentally related.

But wellknown could be pulled out and take a function now, and then updated to take some sort of key identifier later.

PR #13179, opened 2021-07-26
Removed labels usage for content gating

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/909
refs https://github.com/TryGhost/Ghost/commit/c36e749820bae58309791f969c67c12f387b96b8

Content gating by labels was previously added as an alpha feature that allowed Admins to filter post access via both label and product segments. Going forward, we don't want to use labels for content segmentation and instead only use products. This change cleans up usage of `labels` for content gating, which allows us to remove unnecessary `labels` data/query on fetching a member identity by Portal on each page load.


rishabhgrg:
Closing as this is still under discussion and we haven't reached a conclusion yet if its something we want to remove immediately

PR #13181, opened 2021-07-26
🐛 Fixed unsubscribed members receiving email when a post is sent to all members

kevinansfield:
refs https://github.com/TryGhost/Team/issues/935

The problem was incorrect operator precedence when multiple statements existed in the filter original filter when we transform it to enforce `subscribed:true` before sending.

- free only - `subscribed:true+status:free` - no issue
- paid only - `subscribed:true+status:-free` - no issue
- all - `subscribed:true+status:-free,status:free` - the `,status:free` part is treated as a separate OR statement meaning the subscribed:true is not applied to it and free members that are unsubscribed will receive the email

Changes:
- extracted the filter transform into a separate function so it can be unit tested
- updated the transform to use `()` for operator precedence, eg: `subscribed:true+(status:-free,status:free)`
- used transform function in `addEmail()` and `getEmailMemberRows()`

### DIFF
--- a/core/server/services/mega/mega.js
+++ b/core/server/services/mega/mega.js
@@ -24,10 +24,8 @@ const events = require('../../lib/common/events');
 
 const messages = {
     invalidSegment: 'Invalid segment value. Use one of the valid:"status:free" or "status:-free" values.',
-    unexpectedEmailRecipientFilterError: 'Unexpected email_recipient_filter value "{emailRecipientFilter}", expected an NQL equivalent',
-    noneEmailRecipientFilterError: 'Cannot sent email to "none" email_recipient_filter',
-    unexpectedRecipientFilterError: 'Unexpected recipient_filter value "{recipientFilter}", expected an NQL equivalent',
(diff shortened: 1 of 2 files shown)


core/server/services/mega/mega.js
```diff
@@ -128,6 +158,7 @@ const addEmail = async (postModel, options) => {
     const filterOptions = Object.assign({}, knexOptions, {limit: 1});
 
     const emailRecipientFilter = postModel.get('email_recipient_filter');
+    filterOptions.filter = transformEmailRecipientFilter(emailRecipientFilter);
 
     switch (emailRecipientFilter) {
```
daniellockyer, line 163: @kevinansfield Does that mean we can clean this up?

daniellockyer:
@kevinansfield Does this need the same changes here: https://github.com/TryGhost/Ghost/blob/main/core/server/services/mega/mega.js#L353-L371

core/server/services/mega/mega.js
kevinansfield: Copy vs cut fail. Fixed!

daniellockyer (approved):

PR #13201, opened 2021-08-04
Added email_only column to posts_meta table

naz:
closes https://github.com/TryGhost/Team/issues/893
closes https://github.com/TryGhost/Team/issues/894

Reviewers, the only commit for review here is the very first one containing a migration "Added email_only column to posts_meta table " [8729eb2](https://github.com/TryGhost/Ghost/pull/13201/commits/8729eb253f7b266771fa91145cb67ed251873df4). Thanks :pray:

### DIFF
--- a/core/server/api/canary/utils/serializers/output/utils/mapper.js
+++ b/core/server/api/canary/utils/serializers/output/utils/mapper.js
@@ -52,6 +52,9 @@ const mapPost = (model, frame) => {
     _(metaAttrs).filter((k) => {
         return (!frame.options.columns || (frame.options.columns && frame.options.columns.includes(k)));
     }).each((attr) => {
+        // NOTE: the default of `email_only` is `false` which is why we default to `false` instead of `null`
+        //       The undefined value is possible because `posts_meta` table is lazily created only one of the
+        //       values is assigned.
         const defaultValue = (attr === 'email_only') ? false : null;
         jsonModel[attr] = _.get(jsonModel.posts_meta, attr) || defaultValue;
     });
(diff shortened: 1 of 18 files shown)


core/server/api/canary/utils/serializers/output/utils/mapper.js
```diff
@@ -52,7 +52,8 @@ const mapPost = (model, frame) => {
     _(metaAttrs).filter((k) => {
         return (!frame.options.columns || (frame.options.columns && frame.options.columns.includes(k)));
     }).each((attr) => {
-        jsonModel[attr] = _.get(jsonModel.posts_meta, attr) || null;
+        const defaultValue = (attr === 'email_only') ? false : null;
+        jsonModel[attr] = _.get(jsonModel.posts_meta, attr) || defaultValue;
```
allouis, line 56: I'm a bit confused as to the purpose of this, is this to stop a `false` value being turned into a `null`? Or is it because we expect that the value _could_ be `undefined` and we want to add a default?

If it's the first, I wonder if something like this is clearer, in that we check if the value exists, rather than if it's truthy, and use a default if it doesn't
```js
const value = _.get(jsonModel.posts_meta, attr);
jsonModel[attr] = _.isUndefined(value) ? null : value;
```

allouis (commented):
Looks good, just one question about conditional default value stuff

daniellockyer (approved):
Migration LGTM

core/server/api/canary/utils/serializers/output/utils/mapper.js
naz: Nice one @allouis! It's more of the latter one - having an `undefined` for whole posts_meta object and by extension the `email_only` value. The default for the `email_only` is `false` and this is why we want that `if (undefined) assign "false"` logic. 
I see why it's a bit confusing... the ternary operations are not the most readable thing in the world :sweat_smile: 

I'll refactor this into an if/else with a comment. If the question came out now it will definitely confuse us when we come back to this code in the future!

rshbhgrg (approved):

Issue #12945, opened 2021-05-08
Uploading an empty redirects.yaml causes `Cannot read property '302' of undefined`


xsdg:
Ghost should either reject an empty (0 bytes long) YAML file as containing zero redirects, or it should throw an exception advising that an empty file is not accepted.  My preference would be the former, since an empty YAML file appears to be technically valid according to random Googling.

That said, fundamentally, this code should be checking for whether those properties exist before accessing them for iteration:
https://github.com/TryGhost/Ghost/blob/273e22032761d39aef10e0b9a90584f31b0271c9/core/frontend/services/redirects/settings.js#L77

xsdg:
A slightly different manifestation of this bug:
If the uploaded file contains exactly the contents `---`, then this also blows up in the expected way, but this time with the slightly different error, `Cannot read property '302' of null`

Presumably, any valid YAML file that happens to not contain the keys `301` and `302` would cause this to fail.

In particular, I notice that the tests only validate a single sample file, which happens to have both keys defined, so it would not trigger this bug:
https://github.com/TryGhost/Ghost/blob/7528ec8c3b44226896dff1edb7cd03dfd77c4e0d/test/regression/api/canary/admin/redirects_spec.js#L257

ErisDS:
Hi @xsdg,

yaml support is quite a recent contribution to Ghost. I agree it seems the error handling could be improved. If that’s something you’re interested in working on let us know! 

Otherwise as mentioned already, if you need support getting your redirects setup on Ghost(Pro), please drop us an email on support@ghost.org and we'll be more than happy to give you a hand directly 🙂

hey24sheep:
This issue isn't stale as there is an updated PR referecning it.

PR #13215, opened 2021-08-06
Fixes issue #12945 - Validation checks for content, yaml and redirect keys in parseRedirectsFile

hey24sheep:
This PR will fix issue #12945, I have added checks for content, yaml and redirect keys. 

Please have a look and let me know if anything else you would like me to do.


ErisDS:
Hey @hey24sheep I'm so sorry I totally missed this PR 🙈 

These changes look good, only the file has since moved from frontend -> server: 

https://github.com/TryGhost/Ghost/blob/main/core/server/services/redirects/settings.js 

Would you be up for rebasing this PR? If you do drop a comment and I'll keep a close eye out for it.

PR #13219, opened 2021-08-11
Fix member statuses

allouis:
refs https://github.com/TryGhost/Team/issues/959

Due to a bug in the members-api code, when a subscription was canceled
the status would incorrectly be set to 'comped', rather than 'free'.

This PR updates the members-api to correctly set member statuses, as well
as migrates existing members to fix their status to 'free' if they do not have
access to any products.

### DIFF
--- a/core/server/data/migrations/versions/4.12/02-fix-member-statuses.js
+++ b/core/server/data/migrations/versions/4.12/02-fix-member-statuses.js
@@ -1,0 +1,38 @@
+const logging = require('@tryghost/logging');
+const {chunk: chunkArray} = require('lodash');
+const {createTransactionalMigration} = require('../../utils.js');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Updating members with no products to "free"');
+
+        const freeMembers = await knex('members')
+            .select('members.id')
+            .leftJoin('members_products', 'members.id', 'members_products.member_id')
+            .where('members.status', '!=', 'free')
+            .whereNull('members_products.member_id');
+
+        if (freeMembers.length === 0) {
+            logging.info('No free members found');
+            return;
+        }
+
+        // SQLite >= 3.32.0 can support 32766 host parameters
+        // We use one for the SET clause, and the rest can be
+        // used to populate the WHERE IN (?... clause.
+        const chunkSize = 32765;
+
+        const freeMembersIds = freeMembers.map(row => row.id);
+
+        const chunks = chunkArray(freeMembersIds, chunkSize);
+
+        for (const chunk of chunks) {
+            await knex('members')
+                .update('status', 'free')
+                .whereIn('id', chunk);
+        }
+    },
+    async function down() {
+        logging.warn('Not reverting members to incorrect status');
+    }
+);


core/server/data/migrations/versions/4.12/02-fix-member-statuses.js
```diff
@@ -5,10 +5,21 @@ module.exports = createTransactionalMigration(
     async function up(knex) {
         logging.info('Updating members with no products to "free"');
 
-        await knex('members')
+        const freeMembers = await knex('members')
+            .select('members.id')
             .leftJoin('members_products', 'members.id', 'members_products.member_id')
-            .update('status', 'free')
```
allouis: SQLite3 doesn't support updating with a `JOIN` so I've refactored this to use `WHERE IN`

core/server/data/migrations/versions/4.12/02-fix-member-statuses.js
```diff
@@ -0,0 +1,27 @@
...
+const {createTransactionalMigration} = require('../../utils.js');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Updating members with no products to "free"');
+
+        const freeMembers = await knex('members')
+            .select('members.id')
+            .leftJoin('members_products', 'members.id', 'members_products.member_id')
+            .whereNull('members_products.member_id');
```
daniellockyer, line 11: Should we also add a `.where('members.status', '!=', 'free')` to avoid doing extra work?

allouis: Yeah, that makes sense!

core/server/data/migrations/versions/4.12/02-fix-member-statuses.js
```diff
@@ -5,10 +5,21 @@ module.exports = createTransactionalMigration(
...
             .whereNull('members_products.member_id');
+
+        if (freeMembers.length === 0) {
+            logging.info('No free members found');
+            return;
+        }
+
+        const freeMembersIds = freeMembers.map(row => row.id);
+
+        await knex('members')
```
naz, line 20: Think it's best to keep the "chunking" [pattern](https://github.com/TryGhost/Ghost/blob/main/core/server/data/migrations/versions/4.0/03-populate-status-column-for-members.js#L65-L69) here even though the SQLite parameter limit is way bigger now.

core/server/data/migrations/versions/4.12/02-fix-member-statuses.js
naz: Is the `member_product` deleted in this case, as in is it `NULL` ? After going through reproduction steps in https://github.com/TryGhost/Team/issues/959 the `members_products` record still exists for me after cancelling the subscription from Admin UI :thinking:

allouis: @naz Do you have webhooks set up locally to handle the cancelled subscription request? The product should be removed from the member, so the `members_products` entry will be deleted

core/server/data/migrations/versions/4.12/02-fix-member-statuses.js
allouis: Hmm, so the new limit is 32766 - this seems like more than enough to me, unless we have sites out there that have 32k cancelled subscriptions!

I'll add the chunking anyway, but would be good to have a conversation about when that's considered premature optimisation or not

naz: I also think it's a long stretch to have this many cancelled subscriptions out there. The reason why I raised the topic is because we are meant to write code "as if there is a 100k members" in the system from perf perspective. With a big audience of ~100k the 32k limit starts becoming possible? Maybe I'm being too paranoid about having bullet proof code, or maybe we should make this kind of update helper and forget about it :sweat_smile:

ErisDS: 2 things 

1. the new version of sqlite3 that has the new limit isn't yet widely accessible or available in Node. We're trying to take over the node package to fix this, but it's taking time.
2. We are meant to write code as if there is 1million members in the system, as documented here: https://www.notion.so/ghost/Database-Migrations-Seeding-08a408efbd9140e691f46cb903592fd7

allouis: Oh, my bad - I thought since we'd bumped the `sqlite3` to 5.x we'd be using SQLite3 3.34 :grimacing:

So we should always assume that we're working with SQLite3 < 3.32, right?

core/server/data/migrations/versions/4.12/02-fix-member-statuses.js
rishabhgrg: Verified this and the product is being removed correctly from the member and are set as `free`

rshbhgrg (approved):

core/server/data/migrations/versions/4.12/02-fix-member-statuses.js
daniellockyer: We're currently shipping with `sqlite3` 5.0.2, which bundles SQLite 3.34 and the limit was bumped to 32766 in 3.32.0.

However, I'm not sure that the limit applies for `where in (...`? Still, I wouldn't fancy sending a LOT of un-chunked data in 1 query so I think this looks good.

allouis: The limit applies to host parameters, which knex uses under the hood like this:

```sql
SELECT column FROM table WHERE id IN (?, ?, ?) VALUES (1, 2, 3)
```

The limit is essentially on the number of `?` in the statement

ref https://sqlite.org/limits.html (point 9.)

daniellockyer: TIL - ok!

daniellockyer (approved):

naz (approved):
Should be sweet!

PR #13236, opened 2021-08-20
Added members_product_events table

allouis:
refs https://github.com/TryGhost/Team/issues/873

This table is to track events related to members be given or having
removed access to products. It will allow us to provide start dates for
access for complimentary members, as well as being able to track access
to products over time, either for individual members or for aggregates.

### DIFF
--- a/core/server/data/schema/schema.js
+++ b/core/server/data/schema/schema.js
@@ -438,6 +438,17 @@ module.exports = {
         },
         created_at: {type: 'dateTime', nullable: false}
     },
+    members_product_events: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+        product_id: {type: 'string', maxlength: 24, nullable: false, references: 'products.id', cascadeDelete: false},
+        action: {
+            type: 'string', maxlength: 50, nullable: true, validations: {
+                isIn: [['add', 'remove']]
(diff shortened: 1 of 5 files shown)


core/server/data/schema/schema.js
```diff
@@ -438,6 +438,17 @@ module.exports = {
         },
         created_at: {type: 'dateTime', nullable: false}
     },
+    members_product_events: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+        product_id: {type: 'string', maxlength: 24, nullable: false, references: 'products.id', cascadeDelete: false},
+        action: {
+            type: 'string', maxlength: 50, nullable: true, validations: {
+                isIn: [['add', 'remove']]
```
matthanley, line 447: On the [actions table](https://github.com/TryGhost/Ghost/blob/50354633487558e4e4520d10a173182b979018cf/core/server/data/schema/schema.js#L548) the actions are past tense afaict - these should be consistent

matthanley (changes requested):
Is the expectation that we get multiple events for a product change (one removal, one addition)?

allouis:
> Is the expectation that we get multiple events for a product change (one removal, one addition)?

Yeah that's right

rshbhgrg (approved):

daniellockyer (approved):

matthanley (approved):

PR #13242, opened 2021-08-20
🐛 Fix parallel upload of images with the same name

diastremskii:
- because fs interactions are not atomic, it was possible to get the same name for a file and overwrite a previosly created file
- to mitigate that, error handling and retry is added when trying to overwrite an existing file
- "native" fs module is used because fs-extra throws errors that are impossible to handle, just plain text description without any error codes
- additional fs.stat call is used to keep the uniform interface for saving files and directories
- the issue with this approach is that it expects Node to save files asyncronously, but in more or less atomic way, which is not the guarantee Node provides and it might break

While it sounds like an unlikely situation, it gets more likely when uploading files with non-ASCII characters, since they are being replaced by minus symbols, so this bug can be triggered when uploading files that have the same length of the filename.
Found this bug when was trying to track down an issue with images being swapped between the posts in production from time to time. Not yet sure if it happens because of this bug or some other thing, but I guess it's worth fixing it anyway :)
Test added to regression tests fails on current `main`.

### DIFF
--- a/core/server/adapters/storage/LocalFileStorage.js
+++ b/core/server/adapters/storage/LocalFileStorage.js
@@ -3,6 +3,7 @@
 const serveStatic = require('../../../shared/express').static;
 
 const fs = require('fs-extra');
+const nodeFS = require('fs').promises;
 const path = require('path');
 const Promise = require('bluebird');
 const moment = require('moment');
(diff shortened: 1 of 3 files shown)


TheBeastOfCaerbannog:
Oh no, sorry, didn't know `LocalFileStore.save` is used for directories too. Well, as much as it would be nice to avoid dealing with string errors, seems like that's the option to go if the uniform interface for files and directories is a must.

PR #13249, opened 2021-08-24
Populated members_product_events table

allouis:
refs https://github.com/TryGhost/Team/issues/873

In order to have the correct events for existing members we need to
ensure that there is at least an "added" event for each member<->product
relationship in the database.

### DIFF
--- a/core/server/data/migrations/versions/4.13/03-populate-members-products-events-table.js
+++ b/core/server/data/migrations/versions/4.13/03-populate-members-products-events-table.js
@@ -15,7 +15,7 @@ module.exports = createTransactionalMigration(
 
         const memberProductEvents = memberProductRelationships.map((row) => {
             return {
-                id: new ObjectId().toHexString(),
+                id: ObjectId().toHexString(),
                 created_at: connection.raw('CURRENT_TIMESTAMP'),
                 member_id: row.member_id,
                 product_id: row.product_id,
--- a/core/server/data/migrations/versions/4.13/03-populate-members-products-events-table.js
+++ b/core/server/data/migrations/versions/4.13/03-populate-members-products-events-table.js
@@ -5,12 +5,17 @@ const ObjectId = require('bson-objectid').default;
 
 module.exports = createTransactionalMigration(
     async function up(connection) {
-        logging.info('Adding members_product_events rows for existing members_products relationships');
         const memberProductRelationships = await connection('members_products').select('*');
+
+        if (memberProductRelationships.length === 0) {
+            logging.warn('Skipping population of members_product_events rows - no members_products relationships found.');
+            return;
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.13/03-populate-members-products-events-table.js
```diff
@@ -0,0 +1,33 @@
+const {chunk} = require('lodash');
+const {createTransactionalMigration} = require('../../utils');
+const logging = require('@tryghost/logging');
+const ObjectId = require('bson-objectid').default;
+
+module.exports = createTransactionalMigration(
+    async function up(connection) {
+        logging.info('Adding members_product_events rows for existing members_products relationships');
```
daniellockyer, line 8: Can we add logging to:

1. log and early-return if `memberProductRelationships.length === 0`
2. log with `memberProductRelationships.length` otherwise

core/server/data/migrations/versions/4.13/03-populate-members-products-events-table.js
```diff
@@ -0,0 +1,33 @@
...
+        const memberProductRelationships = await connection('members_products').select('*');
+
+        const memberProductEvents = memberProductRelationships.map((row) => {
+            return {
+                id: Reflect.construct(ObjectId, []).toHexString(),
```
daniellockyer, line 13: What's the reason for using `Reflect` here? 😛

core/server/data/migrations/versions/4.13/03-populate-members-products-events-table.js
allouis: Yus!! Will do this now!

core/server/data/migrations/versions/4.13/03-populate-members-products-events-table.js
allouis: Ugh just me trying to get rid of errors in my editor! The types for ObjectId say that it needs to be called with new, so like `new ObjectId().toHexString()` - I wasn't actually supposed to push up the Reflect change :grin:  Was experimenting locally with other ways of calling it - will remove :grin:

core/server/data/migrations/versions/4.13/03-populate-members-products-events-table.js
allouis: Done

core/server/data/migrations/versions/4.13/03-populate-members-products-events-table.js
allouis: Have replaced!

daniellockyer (approved):

rshbhgrg (approved):

PR #13285, opened 2021-09-06
Migrated members comped status to reflect subscriptions

allouis:
refs https://github.com/TryGhost/Team/issues/995

Due to a bug in subscription handling, members with Complimentary stripe
subscriptions were incorrectly given a status of 'paid'.

The goal of this migration is to fix existing broken members, and it
will be accompanied by a fix which prevents the bug for future members.


allouis:
I've reused logic from here: https://github.com/TryGhost/Ghost/blob/main/core/server/data/migrations/versions/4.9/06-add-comped-status.js

kevinansfield (approved):

rshbhgrg (approved):

daniellockyer (approved):

PR #13290, opened 2021-09-07
Used @tryghost/members-api Stripe disconnect logic

allouis:
refs https://github.com/TryGhost/Team/issues/1006

Moving the logic of disconnecting Stripe into the members-api module
decouples the Ghost API from the Members API internals. This method can
now be updated independently of Ghost, to implement the deletion of
webhooks from Stripe.

### DIFF
--- a/core/server/api/canary/settings.js
+++ b/core/server/api/canary/settings.js
@@ -192,19 +192,7 @@ module.exports = {
                 });
             }
 
-            /** Delete all Stripe data from DB */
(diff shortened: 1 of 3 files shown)


allouis:
@rishabhgrg Pinging you on this just for visibility - I've move this logic out of Ghost core here: https://github.com/TryGhost/Members/commit/2dc37b2988a61d675158488f39e6f78764f5036a

PR #13312, opened 2021-09-15
Added temporary database table for analytic events

allouis:
refs https://github.com/TryGhost/Team/issues/1053

This table is going to be completely deleted at some point in the
future. It serves as a persistent datastore for a spike into collection
analytic events for members. We've opted for a generic table, rather
than a table for each event, so that we can push the DB to the limit in
terms of the length of the table, and find out performance issues A$AP

### DIFF
--- a/test/unit/data/exporter/index.test.js
+++ b/test/unit/data/exporter/index.test.js
@@ -173,10 +173,16 @@ describe('Exporter', function () {
             } = require('../../../../core/server/data/exporter/table-lists.js');
 
             const nonSchemaTables = ['migrations', 'migrations_lock'];
+            const requiredTables = schemaTables.concat(nonSchemaTables);
+            // NOTE: You should not add tables to this list unless they are temporary
+            const ignoredTables = ['temp_member_analytic_events'];
+
+            const expectedTables = requiredTables.filter(table => !ignoredTables.includes(table)).sort();
+            const actualTables = BACKUP_TABLES.concat(TABLES_ALLOWLIST).sort();
 
             // NOTE: this test is serving a role of a reminder to have a look into exported tables allowlists
             //       if it failed you probably need to add or remove a table entry from table-lists module
-            [...Object.keys(schema.tables), ...nonSchemaTables].sort().should.eql([...BACKUP_TABLES, ...TABLES_ALLOWLIST].sort());
+            should.deepEqual(actualTables, expectedTables);
         });
 
         it('should be fixed when default settings is changed', function () {
--- a/core/server/data/migrations/versions/4.15/01-add-temp-members-analytic-events-table.js
+++ b/core/server/data/migrations/versions/4.15/01-add-temp-members-analytic-events-table.js
@@ -1,0 +1,12 @@
+const utils = require('../../utils');
+
+module.exports = utils.addTable('temp_member_analytic_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    event_name: {type: 'string', maxlength: 50, nullable: false},
+    created_at: {type: 'dateTime', nullable: false},
+    member_id: {type: 'string', maxlength: 24, nullable: false},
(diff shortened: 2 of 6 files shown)


core/server/data/exporter/table-lists.js
```diff
@@ -36,7 +36,8 @@ const BACKUP_TABLES = [
     'members_status_events',
     'members_paid_subscription_events',
     'members_subscribe_events',
-    'members_product_events'
+    'members_product_events',
+    'temp_member_analytic_events'
```
allouis, line 40: I think it's fine to include this in the export - but I was originally planning to not include it - I couldn't actually find a way to do that though - cc @naz is it possible?

There doesn't seem to be an "EXCLUDE_TABLES" list, and the tests would need updating to be able to handle excluding it - but I wonder if the tests were written that way because we _never_ want to exlude tables?

naz: I don't think it's worth including it in the export if that's a temporary table. If there's no special need for it would avoid adding future maintenance headache.

As for `EXCLUDE_TABLES`, there's no such concept at the moment because we assume be default to exclude every table that's not included in the `BACKUP_TABLES`. I probably don't fully understand the purpose of the `EXCLUDE_TABLES`, can you help me understand why we'd need them?

naz (approved):
The migration looks :+1: Needs small clarification around including new table in the export.

core/server/data/exporter/table-lists.js
allouis: Yeah I was thinking that as well, and then I realised that this is for the backups that Pro does - I think although it's temporary it might be around for a while whilst we test.

@ErisDS Do you have any thoughts on whether or not this table should be included in the backup export?

> I probably don't fully understand the purpose of the EXCLUDE_TABLES, can you help me understand why we'd need them?

The tests are written in such a way that they expect every table to be exported in the backup, so we'd need an exclude list for the tests to ignore - that's all :)

ErisDS: In these cases we should always work with allow lists, not block lists. The code should be refactored so the tests can work from the same allowlist IMO?

> Yeah I was thinking that as well, and then I realised that this is for the backups that Pro does

I don't follow! This is a temporary table, why would we need a backup? Also exports != backups 😬  backups are done with mysqldump. Exports are meant to be for content.

ErisDS: Urgh somehow hit a keyboard combo to submit, but in short, this table def should not be added to the allow list IMO.

allouis: 👌 Awesome okay - I'll get this removed and refactor ze tests to work with it!

test/unit/data/exporter/index.test.js
```diff
@@ -173,10 +173,16 @@ describe('Exporter', function () {
             } = require('../../../../core/server/data/exporter/table-lists.js');
 
             const nonSchemaTables = ['migrations', 'migrations_lock'];
+            const requiredTables = schemaTables.concat(nonSchemaTables);
+            // NOTE: You should not add tables to this list unless they are temporary
+            const ignoredTables = ['temp_member_analytic_events'];
```
allouis, line 178: We don't use the blocklist in the application code anywhere.

This only exists so that any addition to the schema will cause the test to fail still. Without it we would not be able to know that `temp_member_analytic_events` table should not be in the `BACKUP_TABLES` array

allouis: @ErisDS Heads up - I could find a way to refactor the tests without needing this.

The entire point of this test is to stop people from accidentally adding a table, and forgetting to add it to the `BACKUP_TABLES` or `TABLES_ALLOWLIST` - so to do that we need to compare the schema with those lists and error if there is a disrepancy.

I couldn't see another way around this, without knowing to ignore this table!

rshbhgrg (approved):

daniellockyer (approved):

PR #13318, opened 2021-09-17
Refactored migration to run faster

allouis:
no-issue

We're seeing problems with large sites taking a long time to run this
migration. The aim here is to refactor it so that it is faster to run.

We've achieved that by reducing the number of database queries needed,
firstly by selecting members with a join to their events (rather than
fetching the events on a member-by-member basis) we also batch the
necessary updates to further reduce db query time.

### DIFF
--- a/core/server/data/migrations/versions/4.14/02-fix-free-members-status-events.js
+++ b/core/server/data/migrations/versions/4.14/02-fix-free-members-status-events.js
@@ -30,7 +30,7 @@ module.exports = createTransactionalMigration(
             const events = eventsByMember[memberId];
 
             events.sort((a, b) => {
-                return new Date(b) - new Date(a);
+                return new Date(b.created_at) - new Date(a.created_at);
             });
 
             const mostRecentStatusEvent = events[0];
--- a/core/server/data/migrations/versions/4.14/02-fix-free-members-status-events.js
+++ b/core/server/data/migrations/versions/4.14/02-fix-free-members-status-events.js
@@ -19,6 +19,11 @@ module.exports = createTransactionalMigration(
             )
             .where('members.status', 'free');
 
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.14/02-fix-free-members-status-events.js
```diff
@@ -1,36 +1,54 @@
...
+                'members_status_events.created_at'
+            )
+            .innerJoin(
+                'members_status_events',
+                'members.id',
+                'members_status_events.member_id'
+            )
+            .where('members.status', 'free');
+
+        const eventsByMember = _.groupBy(freeMemberEvents, 'member_id');
```
allouis, line 22: Need to exit early if no members

core/server/data/migrations/versions/4.14/02-fix-free-members-status-events.js
```diff
@@ -1,36 +1,59 @@
...
-                await knex('members_status_events')
-                    .update('to_status', 'free')
-                    .where('id', mostRecentStatusEvent.id);
+        const eventsByMember = _.groupBy(freeMemberEvents, 'member_id');
+
+        const eventsToUpdate = Object.keys(eventsByMember).reduce((incorrectEvents, memberId) => {
+            const events = eventsByMember[memberId];
+
+            events.sort((a, b) => {
+                return new Date(b) - new Date(a);
```
daniellockyer, line 33: Should this be `b.created_at` and `a.created_at`?

allouis: EUGH YES!

rshbhgrg (approved):

matthanley (approved):

PR #13321, opened 2021-09-17
Updated OAuth middleware to use the correct URL helpers

matthanley:
no-ref

Hardcoded redirect URLs were breaking when admin and site URLs were different, or when Ghost is configured with a subdirectory.

### DIFF
--- a/core/server/web/oauth/app.js
+++ b/core/server/web/oauth/app.js
@@ -36,11 +36,8 @@ module.exports = function setupOAuthApp() {
      */
     function googleOAuthMiddleware(clientId, secret) {
         return (req, res, next) => {
-            // TODO: use url config instead of the string /ghost
-
             //Create the callback url to be sent to Google
-            const callbackUrl = new URL(urlUtils.getSiteUrl());
-            callbackUrl.pathname = '/ghost/oauth/google/callback';
+            const callbackUrl = new URL(`${urlUtils.urlFor('admin', true)}oauth/google/callback`);
 
             passport.authenticate(new GoogleStrategy({
                 clientID: clientId,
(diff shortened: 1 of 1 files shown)


core/server/web/oauth/app.js
```diff
@@ -36,11 +36,8 @@ module.exports = function setupOAuthApp() {
      */
     function googleOAuthMiddleware(clientId, secret) {
         return (req, res, next) => {
-            // TODO: use url config instead of the string /ghost
-
             //Create the callback url to be sent to Google
-            const callbackUrl = new URL(urlUtils.getSiteUrl());
-            callbackUrl.pathname = '/ghost/oauth/google/callback';
+            const callbackUrl = new URL(`${urlUtils.urlFor('admin', true)}oauth/google/callback`);
```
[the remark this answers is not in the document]
allouis: Also - it might be worth calling `urlUtils.urlFor('admin', true)` just once and storing it - I can't remember how expensive that call is though - so might not be worth it

allouis (approved):
Looks good - two smol comments

core/server/web/oauth/app.js
```diff
@@ -106,12 +103,11 @@ module.exports = function setupOAuthApp() {
 
                 await auth.session.sessionService.createSessionForUser(req, res, req.user);
 
-                return res.redirect('/ghost/');
+                return res.redirect(urlUtils.urlFor('admin'));
             }), {
                 scope: ['profile', 'email'],
                 session: false,
-                prompt: 'consent',
-                accessType: 'offline'
```
naz, line 114: Why was the `accessType: 'offline'` parameter removed? IIRC without this parameter we won't get a refresh token in the response :thinking:

matthanley: I'd removed it to stop my editor complaining and then forgot to add it back - fixed!

PR #13327, opened 2021-09-21
Added `custom_theme_settings` table

kevinansfield:
refs https://github.com/TryGhost/Team/issues/1070

- stores values of custom theme settings
  - will be merged with full settings data parsed from themes for API output
  - will be cached and made available for lookup in themes to avoid db roundtrips
- stores type of custom theme settings so we can coerce values and know if the type has changed when syncing
- records will be synced with themes upon activation

### DIFF
--- a/core/server/data/exporter/table-lists.js
+++ b/core/server/data/exporter/table-lists.js
@@ -50,6 +50,7 @@ const TABLES_ALLOWLIST = [
     'roles',
     'roles_users',
     'settings',
+    'custom_theme_settings',
     'tags',
     'users'
 ];
--- a/core/server/data/migrations/versions/4.16/01-add-custom-theme-settings-table.js
+++ b/core/server/data/migrations/versions/4.16/01-add-custom-theme-settings-table.js
@@ -1,0 +1,9 @@
+const utils = require('../../utils');
+
+module.exports = utils.addTable('custom_theme_settings', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    theme: {type: 'string', maxlength: 191, nullable: false},
+    key: {type: 'string', maxlength: 191, nullable: false},
+    type: {type: 'string', maxlength: 50, nullable: false},
+    value: {type: 'text', maxlength: 65535, nullable: true}
+});
--- a/core/server/data/schema/schema.js
+++ b/core/server/data/schema/schema.js
@@ -643,5 +643,21 @@ module.exports = {
         entry_id: {type: 'string', maxlength: 24, nullable: true},
         source_url: {type: 'string', maxlength: 2000, nullable: true},
         metadata: {type: 'string', maxlength: 191, nullable: true}
+    },
+    custom_theme_settings: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        theme: {type: 'string', maxlength: 191, nullable: false},
+        key: {type: 'string', maxlength: 191, nullable: false},
+        type: {
+            type: 'string',
+            maxlength: 50,
+            nullable: false,
+            validations: {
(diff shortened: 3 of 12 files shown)


core/server/data/exporter/table-lists.js
```diff
@@ -50,6 +50,7 @@ const TABLES_ALLOWLIST = [
     'roles',
     'roles_users',
     'settings',
+    'custom_theme_settings',
```
kevinansfield, line 53: Added to the general export tables because it felt like custom theme settings are an extension of site settings. The active theme is exported as part of site settings so having your customisations of that theme exported too felt like a reasonable expectation. Happy to revise to only include in the full "backup" export if that better fits the intention of export/backup split.

core/server/data/migrations/versions/4.16/01-add-custom-theme-settings-table.js
```diff
@@ -0,0 +1,12 @@
+const utils = require('../../utils');
+
+module.exports = utils.addTable('custom_theme_settings', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    theme: {type: 'string', maxlength: 191, nullable: false},
+    property: {type: 'string', maxlength: 50, nullable: false},
+    type: {type: 'string', maxlength: 50, nullable: false},
+    group: {type: 'string', maxlength: 50, nullable: true},
+    options: {type: 'text', maxlength: 65535, nullable: true}, // json
+    default_value: {type: 'text', maxlength: 65535, nullable: false},
```
kevinansfield, line 10: This data isn't strictly necessary to store because the source of truth will always be the theme and we'll extract that information when validating the theme at startup/activation. However, there are potentially some cases where it would be nice to have the data (eg, informing the user that new options have been added when upgrading a theme) so I've added kept them here for now.

ErisDS: In the spirit of Hyrem's law, if we don't have a usecase for it, I think it would be better to not add more columns?

core/server/data/schema/schema.js
```diff
@@ -643,5 +643,21 @@ module.exports = {
         entry_id: {type: 'string', maxlength: 24, nullable: true},
         source_url: {type: 'string', maxlength: 2000, nullable: true},
         metadata: {type: 'string', maxlength: 191, nullable: true}
+    },
+    custom_theme_settings: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        theme: {type: 'string', maxlength: 191, nullable: false},
+        key: {type: 'string', maxlength: 50, nullable: false},
```
kevinansfield, line 650: Is 50 chars enough for custom property names? `@custom.50_snake_case_chars_is_a_fairly_long_property_var`

core/server/data/migrations/versions/4.16/01-add-custom-theme-settings-table.js
kevinansfield: 👍 Removed them for now, they were complicating the syncing code unnecessarily too.

core/server/data/schema/schema.js
kevinansfield: Switched to 191

daniellockyer (approved):

test/regression/api/canary/admin/db.test.js
```diff
@@ -15,7 +15,7 @@ let ghost = testUtils.startGhost;
 let request;
 let eventsTriggered;
 
-describe('DB API', function () {
+describe('DB API (canary)', function () {
```
allouis, line 18: noice

allouis (approved):

PR #13354, opened 2021-09-22
Added syncing and theme exposure of custom theme settings

kevinansfield:
refs https://github.com/TryGhost/Team/issues/1070TryGhost/Team#1070

- added `@tryghost/custom-theme-settings-service` as a dependency
- `core/server/services/custom-theme-settings` creates an instance of the new service passing in the model used for storing the setting keys/values and a cache instance
- requiring `core/shared/services/custom-theme-settings-cache` creates a cache instance, it has no dependencies so can be required anywhere and the first require will initialize the shared instance
- updated the theme activation bridge to trigger the theme settings service to sync the newly activated theme settings and populate the cache
- updated theme validation to pass `labs` through as an option so that we get custom theme settings back as part of the checked theme as that's what is passed to the custom theme settings service

TODO:
- [x] fix failing gscan publish and bump version


core/boot.js
```diff
@@ -85,6 +85,13 @@ async function initCore({ghostServer, config}) {
     await settings.syncEmailSettings(config.get('hostSettings:emailVerification:verified'));
     debug('End: settings');
 
+    // Custom theme settings cache depends on it's backing model.
+    // Doesn't do anything except create a class instance but needs to be ready before it's required in frontend
+    debug('Begin: Custom theme settings');
+    const customThemeSettingsCache = require('./shared/custom-theme-settings-cache');
+    customThemeSettingsCache.init({model: models.CustomThemeSetting});
+    debug('End: Custom theme settings');
+
```
kevinansfield, line 94: I think that can be avoided if the cache population is done from the custom theme settings service rather than in the cache instance itself.

PR #13361, opened 2021-09-23
Added custom theme settings browse/edit permissions for Administrators

kevinansfield:
refs https://github.com/TryGhost/Team/issues/1070TryGhost/Team#1070

- initial implementation will only allow browse+edit via the API

### DIFF
--- a/core/server/data/migrations/versions/4.16/02-add-custom-theme-settings-permissions.js
+++ b/core/server/data/migrations/versions/4.16/02-add-custom-theme-settings-permissions.js
@@ -1,0 +1,21 @@
+const {
+    addPermissionWithRoles,
+    combineTransactionalMigrations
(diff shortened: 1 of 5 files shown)


allouis (approved):
Looks good!

Non-blocking Q - do we wanna give 'Admin Integration' this permission too?

kevinansfield:
> Non-blocking Q - do we wanna give 'Admin Integration' this permission too?

I thought about that but couldn't think of a use-case where an integration is changing theme settings 🤔 Only possible one I could think of would be an external service updating a setting value periodically or based on an event but I don't think that's the sort of thing we want to encourage being done through theme settings.

rshbhgrg (approved):

daniellockyer (approved):

PR #13377, opened 2021-09-30
Add offers table and model

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1088

- adds new offer table and model
- adds migrations

### DIFF
--- a/core/server/data/exporter/table-lists.js
+++ b/core/server/data/exporter/table-lists.js
@@ -25,6 +25,7 @@ const BACKUP_TABLES = [
     'snippets',
     'tokens',
     'sessions',
+    'offers',
     'stripe_products',
     'stripe_prices',
     'mobiledoc_revisions',
--- a/core/server/data/migrations/versions/4.17/02-add-offers-table.js
+++ b/core/server/data/migrations/versions/4.17/02-add-offers-table.js
@@ -1,0 +1,19 @@
+const utils = require('../../utils');
+
+module.exports = utils.addTable('offers', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    name: {type: 'string', maxlength: 191, nullable: false, unique: true},
+    code: {type: 'string', maxlength: 191, nullable: false, unique: true},
+    product_id: {type: 'string', maxlength: 24, nullable: false},
+    stripe_coupon_id: {type: 'string', maxlength: 255, nullable: false, unique: true},
+    interval: {type: 'string', maxlength: 50, nullable: false, validations: {isIn: [['month', 'year']]}},
+    currency: {type: 'string', maxlength: 50, nullable: true},
+    discount_type: {type: 'string', maxlength: 50, nullable: false, validations: {isIn: [['percent', 'amount']]}},
+    discount_amount: {type: 'integer', nullable: false},
+    duration: {type: 'string', maxlength: 50, nullable: false},
+    duration_in_months: {type: 'integer', nullable: true},
+    portal_title: {type: 'string', maxlength: 191, nullable: false},
+    portal_description: {type: 'string', maxlength: 2000, nullable: true},
+    created_at: {type: 'dateTime', nullable: false},
+    updated_at: {type: 'dateTime', nullable: true}
+});
--- a/core/server/data/migrations/versions/4.17/03-add-offers-permissions.js
+++ b/core/server/data/migrations/versions/4.17/03-add-offers-permissions.js
@@ -1,0 +1,35 @@
+const {
+    addPermissionWithRoles,
+    combineTransactionalMigrations
+} = require('../../utils');
+
+module.exports = combineTransactionalMigrations(
+    addPermissionWithRoles({
+        name: 'Browse offers',
+        action: 'browse',
+        object: 'offer'
+    }, [
+        'Administrator'
+    ]),
+    addPermissionWithRoles({
+        name: 'Read offers',
+        action: 'read',
+        object: 'offer'
+    }, [
+        'Administrator'
+    ]),
+    addPermissionWithRoles({
+        name: 'Edit offers',
+        action: 'edit',
+        object: 'offer'
+    }, [
+        'Administrator'
+    ]),
+    addPermissionWithRoles({
+        name: 'Add offers',
+        action: 'add',
+        object: 'offer'
+    }, [
+        'Administrator'
+    ])
+);
--- a/core/server/models/offer.js
+++ b/core/server/models/offer.js
@@ -1,0 +1,13 @@
+const ghostBookshelf = require('./base');
+
+const Offer = ghostBookshelf.Model.extend({
+    tableName: 'offers',
+
+    product() {
+        return this.belongsTo('Product', 'product_id', 'id');
+    }
+});
+
+module.exports = {
+    Offer: ghostBookshelf.model('Offer', Offer)
+};
--- a/core/server/data/schema/fixtures/fixtures.json
+++ b/core/server/data/schema/fixtures/fixtures.json
@@ -488,6 +488,26 @@
                     "object_type": "snippet"
                 },
                 {
+                    "name": "Browse offers",
+                    "action_type": "browse",
+                    "object_type": "offer"
+                },
+                {
+                    "name": "Read offers",
+                    "action_type": "read",
+                    "object_type": "offer"
+                },
+                {
+                    "name": "Edit offers",
+                    "action_type": "edit",
+                    "object_type": "offer"
+                },
+                {
+                    "name": "Add offers",
+                    "action_type": "add",
+                    "object_type": "offer"
+                },
+                {
                     "name": "Reset all passwords",
                     "action_type": "resetAllPasswords",
                     "object_type": "authentication"
@@ -749,6 +769,7 @@
                     "member_signin_url": "read",
                     "snippet": "all",
                     "custom_theme_setting": "all",
+                    "offer": "all",
                     "authentication": "resetAllPasswords",
                     "members_stripe_connect": "auth"
                 },
--- a/core/server/data/schema/schema.js
+++ b/core/server/data/schema/schema.js
@@ -384,6 +384,23 @@ module.exports = {
         created_at: {type: 'dateTime', nullable: false},
         updated_at: {type: 'dateTime', nullable: true}
     },
+    offers: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        name: {type: 'string', maxlength: 191, nullable: false, unique: true},
+        code: {type: 'string', maxlength: 191, nullable: false, unique: true},
+        product_id: {type: 'string', maxlength: 24, nullable: false},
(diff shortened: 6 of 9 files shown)


core/server/data/migrations/versions/4.17/02-add-offers-table.js
```diff
@@ -0,0 +1,17 @@
+const utils = require('../../utils');
+
+module.exports = utils.addTable('offers', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    name: {type: 'string', maxlength: 191, nullable: false},
+    slug: {type: 'string', maxlength: 191, nullable: false, unique: true},
+    product_id: {type: 'string', maxlength: 24, nullable: false},
+    interval: {type: 'string', maxlength: 50, nullable: false, validations: {isIn: [['monthly', 'yearly']]}},
+    currency: {type: 'string', maxlength: 50, nullable: false},
```
allouis, line 9: Percentage based offers won't have a currency - so I think we need to either have this as `nullable` or we store like `'ANY'` or something for the currency - what do you think?

core/server/data/migrations/versions/4.17/02-add-offers-table.js
```diff
@@ -0,0 +1,17 @@
...
+    discount_type: {type: 'string', maxlength: 50, nullable: false},
+    discount_amount: {type: 'integer', nullable: false},
+    duration: {type: 'string', maxlength: 50, nullable: false},
```
allouis, line 12: If we have a duration of `'repeating'` then we need to store for how long it repeats (e.g. 3 months) so I think we either need another column of like `duration_count` or something or we store a number in this column 🤔 

`0` - once
`Infinity` - forever
`1`,`2`,...`1000` - number of months

core/server/data/migrations/versions/4.17/02-add-offers-table.js
```diff
@@ -0,0 +1,17 @@
+const utils = require('../../utils');
+
+module.exports = utils.addTable('offers', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    name: {type: 'string', maxlength: 191, nullable: false},
+    slug: {type: 'string', maxlength: 191, nullable: false, unique: true},
```
allouis, line 6: Is this to store the Code for an Offer?

core/server/models/offer.js
```diff
@@ -0,0 +1,53 @@
...
+    tableName: 'offers',
+
+    async onSaving(model, _attr, options) {
+        ghostBookshelf.Model.prototype.onSaving.apply(this, arguments);
+
+        if (model.get('name')) {
+            model.set('name', model.get('name').trim());
+        }
+
+        if (model.hasChanged('slug') || !model.get('slug')) {
```
allouis, line 13: I'm wondering if we need this logic too - I think this can be handled by Offers module if necessary too!

core/server/models/offer.js
```diff
@@ -0,0 +1,53 @@
...
+
+        return filteredKeys;
+    },
+
+    product() {
+        return this.belongsTo('Product', 'product_id', 'id');
+    }
+});
+
+const Offers = ghostBookshelf.Collection.extend({
```
allouis, line 46: Can remove this line - we're not adding collection going forward!

core/server/models/offer.js
```diff
@@ -0,0 +1,53 @@
+const ghostBookshelf = require('./base');
+
+const Offer = ghostBookshelf.Model.extend({
+    tableName: 'offers',
+
+    async onSaving(model, _attr, options) {
+        ghostBookshelf.Model.prototype.onSaving.apply(this, arguments);
+
+        if (model.get('name')) {
```
allouis, line 9: I think this can be handled by the Offers module so all the logic is in one place - wdyt?

allouis (commented):
Looks good - few comments and questions!

core/server/data/migrations/versions/4.17/02-add-offers-table.js
rishabhgrg: Yeah this is to store the offer code, i realized the offer code is essentially as good as an offer slug since it will have all of the slug properties for a URL, so we can just store it as that. Was split between naming it `slug` or something like `code`, but `slug` seemed more consistent to me so went with that, not a strong pref tho, wdyt ?

core/server/data/migrations/versions/4.17/02-add-offers-table.js
rishabhgrg: Yea i had the thought, and my feeling was we can always just store the currency for the price at the time offer was created irrespective if it was for percentage based or amount based. For percentage based it will simply get ignored in the logic, while for amount based we can use it, but overall the logic can remain the same and we don't need any extra work for either type, wdyt ?

core/server/data/migrations/versions/4.17/02-add-offers-table.js
rishabhgrg: 🤔 Good catch, i missed the need for storing `duration_count` in case of `repeating`, i'll add a nullable `duration_in_months ` integer column that can store the duration count in number of months to stay consistent with stripe, don't think overloading the `duration` column with number is good

core/server/models/offer.js
rishabhgrg: Comment below ⬇️

core/server/models/offer.js
rishabhgrg: 🤔 Yeah this is basically a copy of old practice we have been following so far for slugs, we could move it to offers module tho which might be cleaner. We'll essentially just need to duplicate the logic in `generateSlug` plugin method, but might be a cleaner cut so on same page

core/server/models/offer.js
rishabhgrg: Ah ok cool, updating 👍

core/server/data/migrations/versions/4.17/02-add-offers-table.js
rishabhgrg: Added `duration_in_months` column now

core/server/models/offer.js
rishabhgrg: Done

core/server/models/offer.js
rishabhgrg: Done

core/server/models/offer.js
rishabhgrg: Trimmed down models now so we can move this logic to members codebase

core/server/data/migrations/versions/4.17/02-add-offers-table.js
allouis: Sure we can store it if you think we should ☺️

core/server/data/migrations/versions/4.17/02-add-offers-table.js
allouis: It wasn't immediately obvious to me what this was storing - but happy to leave it like this if it makes sense to you

core/server/models/offer.js
allouis: Cool - sounds good!

core/server/data/exporter/table-lists.js
```diff
@@ -25,6 +25,7 @@ const BACKUP_TABLES = [
     'snippets',
     'tokens',
     'sessions',
+    'offers',
```
allouis, line 28: I always forget this!

core/server/data/migrations/versions/4.17/03-add-offers-permissions.js
```diff
@@ -0,0 +1,42 @@
...
+        'Administrator'
+    ]),
+    addPermissionWithRoles({
+        name: 'Add offers',
+        action: 'add',
+        object: 'offer'
+    }, [
+        'Administrator'
+    ]),
+    addPermissionWithRoles({
```
allouis, line 35: Do we need/want this permission? AFAIK We're only going to implement archive not delete

core/server/data/migrations/versions/4.17/02-add-offers-table.js
rishabhgrg: It not being immediately obvious is definitely a problem, maybe it makes more sense to use the name as `code` in this case so we don't have same problem down the link. I'll rename it to `code` so its more clear, we can always make it behave like a slug behind the scenes :)

core/server/data/migrations/versions/4.17/02-add-offers-table.js
rishabhgrg: Have updated the `nullable` to `true` in this case as on second thought it does make sense to leave the option in DB to not store anything for it as you said, we can always ensure in logic we save currency for both anyways :)

core/server/data/migrations/versions/4.17/03-add-offers-permissions.js
rishabhgrg: 🤔 Hmm yeah thats true, can bring it back if/when needed, will remove it for now

rishabhgrg: Removed

core/server/data/migrations/versions/4.17/02-add-offers-table.js
```diff
@@ -0,0 +1,18 @@
+const utils = require('../../utils');
+
+module.exports = utils.addTable('offers', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    name: {type: 'string', maxlength: 191, nullable: false},
+    code: {type: 'string', maxlength: 191, nullable: false, unique: true},
+    product_id: {type: 'string', maxlength: 24, nullable: false},
+    interval: {type: 'string', maxlength: 50, nullable: false, validations: {isIn: [['monthly', 'yearly']]}},
```
allouis, line 8: These intervals are incorrect btw - it should be `'month'` and `'year'`

rishabhgrg: 👍 Updated

core/server/data/migrations/versions/4.17/02-add-offers-table.js
rishabhgrg: Done

core/server/data/migrations/versions/4.17/02-add-offers-table.js
```diff
@@ -0,0 +1,18 @@
+const utils = require('../../utils');
+
+module.exports = utils.addTable('offers', {
```
allouis, line 3: Oh I just realised - this is missing a column for `stripe_coupon_id`!

rishabhgrg: Done

rishabhgrg:
Closed in favor of https://github.com/TryGhost/Ghost/pull/13381

core/server/models/offer.js
```diff
@@ -0,0 +1,13 @@
+const ghostBookshelf = require('./base');
+
+const Offer = ghostBookshelf.Model.extend({
```
allouis, line 3: This is missing being exported from `models/index.js` btw!

PR #13379, opened 2021-09-30
Used DynamicRedirectManager for Custom Redirects

allouis:
refs https://github.com/TryGhost/Members/commit/9e59f5a9

Since we have a DynamicRedirectManager for handling adding/removing
redirects at runtime, we no longer need the custom-redirects middleware.
The redirects service does however need an init method now to add the
custom redirects at Ghost boot, so it's been refactored into our Class &
DI pattern.


naz:
these changes landed in main with https://github.com/TryGhost/Ghost/commit/91efa4605cad20c808ccc792a2893d4adc463934

PR #13436, opened 2021-10-04
Moved StripeAPIService to its own service

allouis:
refs https://github.com/TryGhost/Team/issues/1083

The Offers service is going to need access to the StripeAPIService too,
so we need to move it out of the @tryghost/members-api module and make
it accessible to both.


allouis:
https://github.com/TryGhost/Ghost/commit/cda041d424779b1376dc5d9e5fc248a0037207ab

PR #13454, opened 2021-10-05
Bringing  @tryghost/express-dynamic-redirects module into core

naz:
closes https://linear.app/tryghost/issue/CORE-84/have-a-look-at-the-eggs-redirects-refactor-branch


naz:
Note on failing regression tests.

The failing tests relying on stubbing config through "configUtils" should be slimmed down to just couple calls to see if the endpoints exists and respond with a 200s. The rest should be in unit tests for the CustomRedirectsAPI class.

PR #13455, opened 2021-10-05
Replaced i18n.t w/ tpl helper in posts.js

thick-hollins:
refs: TryGhost#13380
The i18n package is depracated. It is being replaced with the tpl package.


ErisDS:
Thank you for your PR :)

Happy Hacktober!

Issue #13380, opened 2021-09-30
Replace i18n.t() with tpl() everywhere


ErisDS:
Our internal, server-side i18n package is deprecated, and so is the use of `i18n.t()`. 

Instead we have a new package [@tryghost/tpl](https://www.npmjs.com/package/@tryghost/tpl) that does string interpolation and makes refactoring i18n out of the codebase easier.   

We have successfully removed `i18n.t()` from core/frontend, but the rest of the codebase has many many references that need to be updated. At the time of writing there are 404 instances of i18n.t in 143 files!

## Steps to refactor

1. Wherever there's an instance of i18n.t in a file we need to:

Replace  
`const i18n.t = require('../[some nesting]../core/shared/i18n)`

With

`const tpl = require('@tryghost/tpl');`

2. Then at the top of the document, after the last require, add a new messages object:

```
const messages = {
};
```

That's the setup done for this file.

3. For each instance of `i18n.t()` for example:

`i18n.t('common.warnings.helpers.flagMustBeEnabled', {flagName: 'something'});`

You'll need to go into `core/shared/i18n/translations/en.json` and find the correct string.

 E.g 
`common.warnings.helpers.flagMustBeEnabled` 

Will look like:
`"flagMustBeEnabled": "The {flagName} flag must be enabled in labs if you wish to use the \\{\\{{helperName}\\}\\} helper.",`

Copy this string into your messages object, as is. If you have your editor setup to use our eslint rules & set to auto fix or auto fix on save, it should automatically fix the quotes & commas.

⚠️ Don't cut or otherwise remove the string from en.json. Strings are reused in multiple places, and it's a waste of time trying to figure out if a string is now unused. We'll simply delete the file verbatim once this refactor is finished.

You should now have something like this:

```
const tpl = require('@tryghost/tpl');
// more requires

const messages = {
	flagMustBeEnabled: 'The {flagName} flag must be enabled in labs if you wish to use the \\{\\{{helperName}\\}\\} helper.'
};
```

Finally, you need to change the reference to `i18n.t` to `tpl` and the string path, to the messages variable:

So from:

`i18n.t('common.warnings.helpers.flagMustBeEnabled', {flagName: 'something'});`

To:

`tpl(messages.flagMustBeEnabled, {flagName: 'something'});`

And that's it!

There is no need to mess with the escaping or interpolation!

## How to work on this issue
[...]

PR #13458, opened 2021-10-05
Replaced i18n.t w/ tpl helper in core/server/services/auth

achromik:
refs #13380

This is a refactor to do everywhere


ErisDS:
Thanks for chipping away at this ❤️

PR #13465, opened 2021-10-06
Replaced i18n.t w/ tpl helper in members

thick-hollins:
refs: TryGhost#13380
The i18n package is deprecated. It is being replaced with the tpl package.


thick-hollins:
I duplicated someone else's work

PR #13554, opened 2021-10-09
Replaced i18n.t w/ tpl helper in api/v3

AlubuArt:
refs: TryGhost#13380
- this is to replace i18n.t with tpl because i18n.t is deprecated


AlubuArt:
refs: #13380

this is to replace i18n.t with tpl because i18n.t is deprecated

PR #13556, opened 2021-10-10
Removed unnecessary spaces in word "with" in GhostMailer.js on line 36

jsonhall:
no issue
- tidy code and comments


ErisDS:
Hi @jsonhall thanks for your interest in the Ghost project. I'm not going to merge this PR because it's Hacktober and I really don't want to encourage this kind of contribution. We do regularly have issues open that we do want help with though.

PR #13592, opened 2021-10-12
Added active column to offers table

allouis:
refs https://github.com/TryGhost/Team/issues/1131

This will be used as a flag to determine whether or not an Offer is
archived or active.


rshbhgrg (approved):

daniellockyer (approved):

tpatel (approved):
Looks great 🙌

PR #13594, opened 2021-10-12
Added boot-time metric shipping

sam-lord:
Ghost already measures boot-time, which is awesome. To properly track boot-time as well as other performance metrics it would be beneficial to have a way to store and analyse this data.

The new @tryghost/metrics package follows the same patterns as @tryghost/logging, and uses the same configuration file (loggingrc.js in the root of the project). It plugs into multiple backends (stdout & elasticsearch for now) and transforms the metric into a format which is best processed by that backend.

When no metric transports are specified, nothing will change. Adding a transport will start shipping the boot-time metric to that backend. For backends which support it (currently just ElasticSearch), additional metadata is shipped alongside metric.

### DIFF
--- a/core/boot.js
+++ b/core/boot.js
@@ -16,13 +16,18 @@ const debug = require('@tryghost/debug')('boot');
  * Helper class to create consistent log messages
  */
 class BootLogger {
-    constructor(logging, startTime) {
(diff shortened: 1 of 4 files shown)


ErisDS (approved):
LGTM 👍

PR #13609, opened 2021-10-14
Removed NOT NULL constraint from stripe_coupon_id

allouis:
no-issue

When handling disconnecting from Stripe - we remove all Stripe data from
our database to ensure we do not have bad/invalid data stored. Removing
this constraint will allow us to set the value to NULL.

This also updates the `offer_redemptions` table to cascade deletes for subscirption_id
and adds a created_at for ordered


tpatel (approved):
Looks great 🙌

rshbhgrg (approved):

daniellockyer (approved):

PR #13679, opened 2021-10-22
Implemented spike version for video content type support

naz:
refs https://linear.app/tryghost/issue/CORE-121/create-a-video-storage-adapter
refs https://github.com/TryGhost/Toolbox/issues/95

- Example source files downloaded from https://filesamples.com/
- This is an experimental implementation of video file upload support
- The storage adapter still needs more thinking as it's almost the same as the "LocalStorgeAdapter" that stores images.
- Also the output serializer skipped use of url utils in favor of inline implementatoin - this should almost certainly be it's own package


naz:
There will be follow up work with `/media/` endpoint that will be done directly in main

PR #13703, opened 2021-11-01
Moved `launchComplete` user setting as global editor setting

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/807

The launch wizard completed flag was previously stored at per user level in accessibility column of user table, so an administrator still got the option to complete the launch wizard even if the owner had completed it previously, which is not expected pattern. This change moves the launch complete flag for Admin to common settings from per user level so a site only needs to complete the launch wizard once irrespective of which user completes it

- Adds new `editor_is_launch_complete` setting to track if a site launch steps are completed in Admin
- Adds new migration util to easily allow adding new setting 
- Adds migration to introduce new `editor_is_launch_complete` setting
- Adds migration to update launch complete flag for a site if any of the users have already completed the launch steps

Migrations [here](https://github.com/TryGhost/Ghost/pull/13703/files#diff-95604734d8223e4f7afb0295f0eaacd3593fc34aa7321f4698106152c9b1e9dd) and [here](https://github.com/TryGhost/Ghost/pull/13703/files#diff-3f8ef89baacca449ee551f816b0c4f255fa1353a1b4d0ef1f23cbbbe4fcd0318)


core/server/data/migrations/utils.js
```diff
@@ -421,12 +421,63 @@ function createDropColumnMigration(table, column, columnDefinition) {
...
 }
 
+/**
+ * Creates a migration which will insert a new setting in settings table
+ * @param {object} settingSpec - setting key, value, group and type
+ *
+ * @returns {Object} migration object returning config/up/down properties
+ */
+function addSetting({key, value, type, group}) {
+    return createNonTransactionalMigration(
```
allouis, line 431: Why is this non transactional?

core/server/data/schema/default-settings.json
```diff
@@ -590,6 +590,13 @@
         "editor_default_email_recipients_filter": {
             "defaultValue": "all",
             "type": "string"
+        },
+        "editor_is_launch_complete": {
+            "defaultValue": "false",
+            "validations": {
```
allouis, line 596: Other boolean settings we have also have the `"isEmpty": false` validation - is it worth adding here?

allouis (commented):
Couple comments but otherwise looks good!

core/server/data/schema/default-settings.json
rishabhgrg: Yea good point, no reason to not add it here, updating

core/server/data/migrations/utils.js
rishabhgrg: @allouis Ah so this was picked up from the pattern we use for `createAddColumnMigration` which also used non transactional, but probably that's wrong here and we can/should use transactional instead ? 🤔

allouis: Ah okay - so there's two types of migrations DML and DDL - essentially are we changing the data, or are we changing the tables. If we're changing the data then it should always be in a transaction (so that any errors are rolled back completely), if we're changing the tables (e.g. adding a column) then these cannot be done inside a transaction!

Hopefully that all makes sense

rishabhgrg: This makes perfect sense! Thanks for the details 😄 I'll update this to wrap in a transaction

rishabhgrg: Done ✅

daniellockyer:
We're targetting 4.22 now 😀

daniellockyer (approved):

kevinansfield (approved):

allouis (approved):

daniellockyer:
Re-requested review from myself as we may be delaying the merge and we'll need version numbers updating

daniellockyer (approved):

PR #13711, opened 2021-11-04
Added migration review checklist

matthanley:
Migrations are dangerous and need to be treated with a great deal of care. This workflow adds a comment to PRs containing migrations with a checklist for the reviewer to complete, which covers known pitfalls with migrations that may compromise correctness, performance, or safety.

It runs on the `pull_request_target` event because `pull_request` events run in the context of the fork rather than the base repo, and don't have access to comment on the PR.

### DIFF
--- a/.github/workflows/migration-review.yml
+++ b/.github/workflows/migration-review.yml
@@ -1,0 +1,45 @@
+name: Migration Review
+on:
+  pull_request_target:
+    types: [opened]
+    paths:
+      - 'core/server/data/schema/**'
+      - 'core/server/data/migrations/versions/**'
+jobs:
+  createComment:
+    runs-on: ubuntu-latest
+    name: Create checklist comment
+    steps:
+      - uses: peter-evans/create-or-update-comment@v1
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            It looks like this PR contains a migration 👀
+            Here's the checklist for reviewing migrations:
+
+            ### General requirements
+
+            - [ ]  Satisfies idempotency requirement (both `up()` and `down()`)
+            - [ ]  Does not reference models
+            - [ ]  Filename is in the correct format
+            - [ ]  Targets the next minor version
+            - [ ]  All code paths have appropriate log messages
+            - [ ]  Uses the correct utils
+            - [ ]  Contains a minimal changeset
+            - [ ]  Does not mix DDL/DML operations
+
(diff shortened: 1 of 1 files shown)


.github/workflows/migration-review.yml
```diff
@@ -0,0 +1,43 @@
...
+          body: |
+            It looks like this PR contains a migration 👀
+            Here's the checklist for reviewing migrations:
+
+            ### General requirements
+
+            - [ ]  Satisfies idempotency requirement (both `up()` and `down()`)
+            - [ ]  Does not reference models
+            - [ ]  Filename is in the correct format
+            - [ ]  All early returns have appropriate log messages
```
daniellockyer, line 25: Can we change this to "All code paths have appropriate log messages"?

matthanley: ✅

.github/workflows/migration-review.yml
```diff
@@ -0,0 +1,43 @@
...
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            It looks like this PR contains a migration 👀
+            Here's the checklist for reviewing migrations:
+
+            ### General requirements
+
+            - [ ]  Satisfies idempotency requirement (both `up()` and `down()`)
+            - [ ]  Does not reference models
+            - [ ]  Filename is in the correct format
```
daniellockyer, line 24: Is it worth adding "Targets the next minor version" as an extra explicit check?

.github/workflows/migration-review.yml
```diff
@@ -0,0 +1,43 @@
...
+
+            - [ ]  Both schema change and related migration have been implemented
+            - [ ]  For column changes: indexes have been recreated for sqlite
+            - [ ]  For index changes: has been performance tested for large tables
+            - [ ]  For new tables/columns: fields use the appropriate predefined field lengths
+            - [ ]  For new tables/columns: field names follow the appropriate conventions
+            - [ ]  Does not drop a non-alpha table outside of a major version
+            - [ ]  For settings updates: follows the appropriate guidelines
+
+            ### Data changes
```
daniellockyer, line 40: Perhaps also something like "handles invalid data" (to cover JSON parsing errors) and "handles missing data" (checking for knex result length)?

daniellockyer (approved):
LGTM! 🕺🏻 Difficult to test before merging but let's see 👀

PR #13730, opened 2021-11-10
[WIP] Custom OEmbed providers

allouis:
no-issue

This PR is open to share code with Kevin and receive feedback


core/server/services/oembed.js
```diff
@@ -53,7 +59,26 @@ class OEmbed {
      */
     constructor({config, externalRequest}) {
         this.config = config;
-        this.externalRequest = externalRequest;
+        /** @type {IExternalRequest} */
+        this.externalRequest = async (url, requestConfig) => {
```
allouis, line 63: This wraps the external request with checks before & after for ip/localhost - this is so custom providers don't need to know/care about that. 

I want to check this looks correct, and if it does, then we should be able to remove these checks from the methods below and just use `this.externalRequest`

core/server/services/oembed.js
```diff
@@ -274,6 +299,16 @@ class OEmbed {
         let data;
 
         try {
+            const urlObject = new URL(url);
+            for (const provider of this.customProviders) {
+                if (await provider.canSupportRequest(urlObject)) {
+                    const result = await provider.getOEmbedData(urlObject, this.externalRequest);
+                    if (result !== null) {
```
allouis, line 306: Allows fallthrough

core/server/services/nft-oembed.js
```diff
@@ -0,0 +1,59 @@
...
+     */
+    async getOEmbedData(url, externalRequest) {
+        const [match, transaction, asset] = url.pathname.match(OPENSEA_PATH_REGEX);
+        if (!match) {
+            return null;
+        }
+        const result = await externalRequest(`https://api.opensea.io/api/v1/asset/${transaction}/${asset}/`, {
+            json: true
+        });
+        return {
```
allouis, line 41: I'm trying to stick to the OEmbed spec here - but I'm not sure if that's how you imagined it?

What's particularly tricky with the OEmbed spec is that we should provide the minimum width & height in pixels - which I've hardcoded to 1000 atm :grimacing:

core/server/services/nft-oembed.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * @implements ICustomProvider
+ */
+class NFTOEmbedProvider {
+    /**
+     * @param {object} dependencies
+     */
+    constructor(dependencies) {
+        this.dependencies = dependencies;
```
allouis, line 16: placeholder

core/server/services/nft-oembed.js
kevinansfield: Yeah, we should stick the the oembed spec as much as possible. We don't actually use the width/height for anything at the moment other than to verify that we have what looks like a [valid oembed response here](https://github.com/TryGhost/Ghost/blob/main/core/server/services/oembed.js#L255-L257) so hardcoded value is fine for now 👍

core/server/services/oembed.js
kevinansfield: 👍 LGTM and agree we can remove the other checks below

PR #13758, opened 2021-11-17
e2e tests state preparation for url-cache boots

naz:
refs https://github.com/TryGhost/Toolbox/issues/135


core/boot.js
```diff
@@ -346,11 +351,14 @@ async function bootGhost() {
...
-        await ghostServer.start(rootApp);
-        bootLogger.log('server started');
-        debug('End: load server + minimal app');
+
+        if (withBackend) {
+            const GhostServer = require('./server/ghost-server');
+            ghostServer = new GhostServer({url: config.getSiteUrl()});
+            await ghostServer.start(rootApp);
+            bootLogger.log('server started');
+            debug('End: load server + minimal app');
```
ErisDS, line 360: Little thing but this debug statement shouldn't be caught by the if statement

core/boot.js
```diff
@@ -297,7 +302,7 @@ async function initBackgroundServices({config}) {
 
  * @returns {Promise<object>} ghostServer
  */
-async function bootGhost() {
+async function bootGhost({withBackend = true, withFrontend = true} = {}) {
```
ErisDS, line 305: Also a minor comment, but these params could simply be `backend` and `frontend`. With makes it more confusing IMO, not less 😉

core/boot.js
ErisDS: A more hairy comment 😬 

IMO mounting the backend onto the app, and booting the server are not the same thing. I wouldn't expect setting backend/withBackend false to result in the server not being created 🤔 I'd expect it to result in a server without any backend paths mounted.

naz:
This PR still needs few last bits fixing up the test environment. Hoping to merge it on Monday :crossed_fingers:

PR #13790, opened 2021-11-29
Add lint rules for migrations folder

sam-lord:
refs: https://github.com/TryGhost/Toolbox/issues/105

The lint rules prevent:

* Invalid naming conventions for new migrations
* Loop constructs in migrations - these should be used with caution and are therefore a warning rule, use `// eslint-disable-next-line no-restricted-syntax` to prevent this rule from firing where a loop is required
* Returing within a loop - this is usually meant to be a continue/break
* Multiple joins - these can be badly performing migrations, so should be treated with caution, disable the rule for the line if the risk is understood / the migration cannot be written without it

Added `// eslint-disable-next-line no-restricted-syntax` in existing migrations to make linting pass, and also as an example for how to use it in future where these constructs are necessary.

For reviewers:
* Are the `2021-11-29-11-45-slugified-name.js` format and the `01-slugified-name.js` formats enough? I've used `(?:-[a-zA-Z]+)+` for the name segments, but if we also need numerical bits in the slugified name I will add those to the regex.
* Is the approach of disabling a rule for a single-line good for removing warnings for the developer around performance?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -23,6 +23,38 @@ module.exports = {
             }
         },
         {
+            files: 'core/server/data/migrations/versions/**',
+            excludedFiles: [
+                'core/server/data/migrations/versions/1.*/*',
+                'core/server/data/migrations/versions/2.*/*',
+                'core/server/data/migrations/versions/3.*/*'
+            ]
+        },
+        {
+            files: 'core/server/data/migrations/versions/**',
+            rules: {
+                'no-restricted-syntax': ['warn', {
+                    selector: 'ForStatement',
+                    message: 'For statements can perform badly in migrations'
+                }, {
+                    selector: 'WhileStatement',
+                    message: 'While statements can perform badly in migrations'
+                }, {
+                    selector: 'CallExpression[callee.property.name=\'forEach\']',
+                    message: 'Loop constructs like forEach can perform badly in migrations'
+                }, {
+                    selector: 'CallExpression[callee.object.name=\'_\'][callee.property.name=\'each\']',
+                    message: 'Loop constructs like _.each can perform badly in migrations'
+                }, {
+                    selector: 'ForStatement ReturnStatement',
+                    message: 'Invalid use of a return statement within for-loop in a migration'
+                }, {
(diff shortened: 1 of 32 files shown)


.eslintrc.js
```diff
@@ -22,6 +22,45 @@ module.exports = {
...
+                'core/server/data/migrations/versions/3.*/*'
+            ],
+            rules: {
+                'ghost/filenames/match-regex': ['error', '^(?:\\d{2}|\\d{4}(?:-\\d{2}){4})(?:-[a-zA-Z]+)+$', true]
+            }
+        },
+        {
+            files: 'core/server/data/migrations/versions/**',
+            rules: {
+                'no-restricted-syntax': ['warn', {
```
ErisDS, line 39: 🤔 does this get correctly merged with the no-restricted syntax errors in the underlying eslint-plugin-ghost without downgrading those errors to warnings?

.eslintrc.js
```diff
@@ -22,6 +22,45 @@ module.exports = {
...
+                    selector: 'CallExpression[callee.property.name=\'forEach\']',
+                    message: 'Loop constructs like forEach can perform badly in migrations'
+                }, {
+                    selector: 'CallExpression[callee.object.name=\'_\'][callee.property.name=\'each\']',
+                    message: 'Loop constructs like _.each can perform badly in migrations'
+                }, {
+                    selector: 'CallExpression[callee.property.name=/join|innerJoin|leftJoin/] CallExpression[callee.property.name=/join|innerJoin|leftJoin/] CallExpression[callee.name=\'knex\']',
+                    message: 'Use of multiple join statements in a single knex block'
+                }],
+                'ghost/no-return-in-loop/no-return-in-loop': ['error']
```
ErisDS, line 61: Really surprised this doesn't exist in eslint, eslint-plugin-node or eslint-plugin-unicorn... seems like a doozie?!

ErisDS (commented):
This looks great! Very cool :) 

I have just one question about how the no-restricted-syntax rules interact.

.eslintrc.js
sam-lord: Yeah I couldn't find it! It's more involved than a `no-restricted-syntax` plugin because it needs to avoid traversing into function expressions. I've not enabled it globally but it sounds like a good one to introduce

PR #13814, opened 2021-12-01
Moved vhost mounts into boot file

naz:
:see_no_evil: 

The baseline for merging these changes is getting the tests to pass and run in similar timeframe as they do on main. Right now the time to run all mock-express-style regression tests is: **41s**.


naz:
The current running time with the unified boot process is in the ballpark of **47s**. Merging this will be a tradeoff of <10s of runtime for much more maintainable approach to the instance boot.

Issue #13648, opened 2021-10-22
Replace moment with luxon in core/frontend


ErisDS:
Ghost makes heavy use of the moment and moment-timezone libraries all over our code bases. However, it's an extremely heavy dependency, we have difficulties getting the timezone to "stick" and the project itself essentially recommends not using it anymore: https://momentjs.com/docs/#/-project-status/

Instead of using Moment, we want to change to using [Luxon](https://www.npmjs.com/package/luxon). We're using Luxon because of its timezone capabilities - something the core of Ghost makes heavy use of. In smaller packages where we only need to manipulate dates, rather than creating, storing and retrieving them, we'll use [date-fns](https://www.npmjs.com/package/date-fns).

For the main Ghost codebase, @daniellockyer has already [laid some groundwork](https://github.com/TryGhost/Ghost/pull/13623/files), added Luxon as a dependency and ensured the default timezone is set to UTC.

From here, we need to start replacing usages of moment with luxon.

This is a new thing for us, we haven't used the library in many places yet and so we don't know a lot about it. For that reason, we want to start on areas of the codebase where we can do the least harm - the frontend. If we end up displaying dates incorrectly for some reason on the frontend that's _much easier_ to fix than if we store the wrong date!

## How to work on this issue

The slow and steady approach to refactoring is the only thing that really works. Branches should be very short lived (<1 day if possible). Therefore, please do not attempt to do this entire refactor in a single PR. It is not possible for us to review!

Instead, please find a single file inside of `core/frontend`, make the necessary changes, test and lint your work, then submit a PR. If you like, repeat that with another file.

Luxon provides a [useful guide](https://moment.github.io/luxon/#/moment) to help translate between code in `moment` and `luxon`, along with important functional differences.

Please follow our [contribution guidelines] closely, particularly being sure to reference this issue in your commit. Refactor commits should not use emojis as they are not user-facing changes.

Thank you 🙏 ❤️  🙏

danimajo:
@ErisDS @matthanley any feedback on my PRs?

ErisDS:
Hey @danimajo I still really want to get this done, but it turned out to require a lot more knowledge & context than I hoped. I don't think our tests are going to do a good job of picking up if formats change slightly. I also think we need more context of exactly which luxon functions to use where and why.

[...]

PR #13839, opened 2021-12-03
Replaced moment with luxon in prev_post.js

danimajo:
refs: #13648
Replace moment with luxon


core/frontend/services/sitemap/base-generator.js
```diff
@@ -104,7 +103,7 @@ class BaseSiteMapGenerator {
         node = {
             url: [
                 {loc: url},
-                {lastmod: moment(this.getLastModifiedForDatum(datum)).toISOString()}
```
ErisDS, line 107: Why is it ok to lose formatting here?

ErisDS (changes requested):
I realise you raised separate PRs but actually this one has all changes in it.  Further to my comments in https://github.com/TryGhost/Ghost/issues/13648, this PR raises some questions, which I've left as individual comments.

I don't know enough about luxon to be sure about these changes, and also I don't trust our tests are doing enough to ensure that the right date formats are being output, especially for sitemaps.

ErisDS:
Thanks for trying to help with this refactor. I have ended up closing the issue for this as, whilst we do want to migrate to luxon, I don't have enough context to help people get these changes over the line. 

It needs either deep, focused work or someone experienced to move it forward.

Issue #13832, opened 2021-12-02
RSS RFC-822 standard not valid for "fr" locale


frnode:
### Issue Summary

Hello,

I use Ghost CMS with the "French" ("fr") language, I wanted to implement the RSS feed available here: https://ghost.org/integrations/custom-rss/

But in French language it does not respect the "[RFC-822](https://www.rssboard.org/rss-validator/docs/error/InvalidRFC2822Date.html)" standard, because it translates the date into the defined language and there is no way to force the display in "English" mode.

If anyone has any idea how we can get around this so that we have a valid feed!

Thank you,
Regards

### Steps to Reproduce

1. [Implement RSS](https://ghost.org/integrations/custom-rss/)
2. Change lang to "fr" in Admin
3. Go to the RSS page
4. Look <lastBuildDate> and <pubDate>

### Ghost Version

4.24

### Node.js Version

14.18.2-0

### How did you install Ghost?

Own server with https://hub.docker.com/r/bitnami/ghost/

### Database type

Other

### Browser & OS version

All

### Relevant log / error output

```shell
<lastBuildDate>jeu., 02 déc. 2021 19:17:18 +0100</lastBuildDate>
<pubDate>mar., 30 nov. 2021 08:30:00 +0100</pubDate>

Good format is:

<pubDate>Wed, 02 Oct 2002 15:00:00 +0200</pubDate>
<pubDate>Wed, 02 Oct 2002 15:00:00 +0200</pubDate>
```

[ Feed Validation Service ](https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Finsys.fr%2Fblog%2Frss%2F)

### Code of Conduct

- [x] I agree to be friendly and polite to people in this repository

aq1:
True, true. Hoped there is solution but I guess there isn't

touzoku:
@ErisDS any chance to get #13892 merged? Our RSS is a mess 😿

ErisDS:
This has been fixed by @touzoku in PR #13892. The fix will go out in the next release, which is currently our standard Friday afternoon automated release.

PR #13892, opened 2021-12-15
Add timezone and locale to the options of the date helper

touzoku:
The `date` handlebars helper has only one option currently — `format`. It assumes the locale and timezone from the `options.data.site` object which is not always desired behavior.

The helper sometimes is used, for example, in custom RSS template where we always need the `en-US` locale, not the one that we have configured for the website globally. This change makes the two options configurable. The default values come from the `options.data.site` object values if omitted, and therefore this change is backwards compatible.

This pull request fixes the issue #13832.

### DIFF
--- a/core/frontend/helpers/date.js
+++ b/core/frontend/helpers/date.js
@@ -25,12 +25,11 @@ module.exports = function (...attrs) {
     // ensure that date is undefined, not null, as that can cause errors
     date = date === null ? undefined : date;
 
-    const timezone = options.data.site.timezone;
-    const locale = options.data.site.locale;
-
     const {
         format = 'll',
-        timeago
+        timeago,
+        timezone = options.data.site.timezone,
+        locale = options.data.site.locale
     } = options.hash;
 
     const timeNow = moment().tz(timezone);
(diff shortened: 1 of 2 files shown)


touzoku:
Tagging @ErisDS and @matthanley for visibility as you were the last two contributors to this part of code.

aq1:
This is simply brilliant. I need this ASAP

touzoku:
@ErisDS I know it is a holiday season, but is there any chance to merge this soon?

ErisDS:
Hey there @touzoku, really sorry for the delay. We close over Christmas and it takes a little while to catch up when we're back.

The change looks good, but could you please add a unit test [in this file](https://github.com/TryGhost/Ghost/blob/f0c680c044ee9916d42d742bf86786d474e8bef2/test/unit/frontend/helpers/date.test.js) to show that this fixes the problem you experienced? Thanks!

touzoku:
I have added a test, @ErisDS. Let me know if there are any other blockers.

ErisDS:
This is great, thank you so much for the PR! :)

touzoku:
@ErisDS Thank you for a prompt merge!

By the way, I would have updated the docs as well but can't locate the repository for them.
https://ghost.org/docs/themes/helpers/date/

matthanley:
Hey @touzoku - thanks for the PR! I've updated the docs mentioning the change.

PR #13907, opened 2021-12-16
Switched AMP to be 'off' by default in all new Ghost instances

allouis:
refs https://github.com/TryGhost/Team/issues/1189

### DIFF
--- a/core/server/data/schema/default-settings.json
+++ b/core/server/data/schema/default-settings.json
@@ -397,7 +397,7 @@
     },
     "amp": {
         "amp": {
-            "defaultValue": "true",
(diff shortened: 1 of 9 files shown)


yogeshbeniwal:
@allouis Any reason to disable AMP by default?

allouis:
@yogeshbeniwal We're going to be slowly moving away from AMP as it seems the rest of the internet is, it no longer has an effect on SEO anymore. It makes it more complicated to add new cards, and doesn't particularly play well with membership functionality and content gating.

The option to enable it will still be there, just off by default!

PR #13930, opened 2022-01-05
Improve the performance of prepareContentFolder

sam-lord:
The prepareContentFolder function in our test suite takes a long time to execute because it does so much unnecessary work.

This PR currently fails tests, creating for collaboration :)


ErisDS:
This is probably worth rebasing now, although I don't think it'll fix everything just yet :)

ErisDS:
Gonna close this for now, as having a bit of a cleanup before the branch rejig. We know we need to come back to it!

PR #13955, opened 2022-01-11
Allow custom helpers

mariusbolik:
# Allow custom helpers

This pull request makes it possible to use custom helpers with Ghost.

Custom helpers can be placed in `/content/helpers`, or inside a theme folder: `/content/themes/casper/helpers`.

### Example
Place the following code into `/content/helpers/stringify.js`:
```js
module.exports = function stringify(object) {
    return JSON.stringify(parseJson);
 }
```

Now you can use it like after restarting ghost:
```hbs
{{#get "posts" limit="all"}}
{{#stringify posts}}
```

### DIFF
--- a/core/frontend/services/helpers/index.js
+++ b/core/frontend/services/helpers/index.js
@@ -1,12 +1,24 @@
 const registry = require('./registry');
-
 const path = require('path');
+const config = require('../../../shared/config');
+const settingsCache = require('../../../shared/settings-cache');
 
-// Initialise Ghost's own helpers
 // This is a weird place for this to live!
-const init = async () => {
+const init = () => {
+    // Initialize Ghost's own helpers
     const helperPath = path.join(__dirname, '../../', 'helpers');
-    return await registry.registerDir(helperPath);
+    registry.registerDir(helperPath);
+
+    // Initialize custom helpers in /content/helpers
+    const customHelperPath = config.getContentPath('helpers');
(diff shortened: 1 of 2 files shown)


ErisDS:
Hi @mariusbolik thanks for supplying this PR. 

We don't have support for custom helpers in Ghost for a couple of key reasons:

- The first is that it would allow users/theme developers to run arbitrary custom code, which needs careful, in-depth consideration from both a usability and security point of view. We not only have to add support, but would need to maintain it, ensuring Ghost is secure and usable long term.
- The second is that to date, no one has provided a clear use case for custom helpers, such that there's value to offset the costs I mentioned above. Every time we dig into the underlying reason why someone is trying to add a custom helper they are trying to either do something that can already be done with Ghost's existing, extensive set of helpers, that wouldn't work as expected e.g. rendering dynamic content into a statically rendered template, or they were trying to do something that should have been possible and so we added a helper to accomplish the goal.

It appears you're trying to output Ghost data as JSON, which doesn't make much sense to me as a use case in production, but I'd hazard a guess you were trying to make development and debugging easier, in which case the closest thing we have right now is the `{{log}}` helper, that will output the JSON to your terminal. We have discussed adding a `{{json}}` helper for this as well, which we would happily accept a PR for, and makes more sense for all the reasons I've outlined than adding generic custom helper support.

If I'm wrong and your use case is something else, let me know :)

PR #13970, opened 2022-01-13
Added new free tier as proper tier

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1037

Migrations at - 
- https://github.com/TryGhost/Ghost/pull/13970/files#diff-549e3032ba241e39d0a645d7ec58a652bee03ff53ba36de6d3ef384843404724
- https://github.com/TryGhost/Ghost/pull/13970/files#diff-0fa91bff0d1a019a087be2fa1995abe98a092936741d712358768c904f9f5dc9

-----

At the moment, free tier is not actually structured/stored as a proper "tier", and has the description hard-coded. This doesn't allow the flexibility to attach benefits/description to a free tier like rest of the tiers to publishers. This change adds the free tier as a proper "tier" in database. It also introduces a new `type` column to differentiate between free and paid tiers, allowing free tier to have same first class support for description/benefits as other tiers.

### DIFF
--- a/core/frontend/services/theme-engine/middleware/update-global-template-options.js
+++ b/core/frontend/services/theme-engine/middleware/update-global-template-options.js
@@ -28,7 +28,9 @@ function calculateLegacyPriceData(products) {
         };
     }
 
-    const defaultProduct = products[0] || {};
+    const defaultProduct = products.find((product) => {
+        return product.type === 'paid';
+    }) || {};
 
     const monthlyPrice = makePriceObject(defaultProduct.monthly_price || defaultPrice);
 
--- a/core/server/api/canary/utils/serializers/output/products.js
+++ b/core/server/api/canary/utils/serializers/output/products.js
@@ -73,6 +73,7 @@ function serializeProduct(product, options, apiType) {
         name: json.name,
         description: json.description,
         slug: json.slug,
+        type: json.type,
         created_at: json.created_at,
         updated_at: json.updated_at,
         stripe_prices: json.stripePrices ? json.stripePrices.map(price => serializeStripePrice(price, hideStripeData)) : null,
--- a/core/server/data/migrations/versions/4.33/2022-01-14-11-50-add-type-column-to-products.js
+++ b/core/server/data/migrations/versions/4.33/2022-01-14-11-50-add-type-column-to-products.js
@@ -1,0 +1,12 @@
+const utils = require('../../utils');
+
+module.exports = utils.createAddColumnMigration(
+    'products',
+    'type',
+    {
+        type: 'string',
+        maxlength: 50,
+        nullable: false,
+        defaultTo: 'paid'
+    }
+);
--- a/core/server/data/migrations/versions/4.33/2022-01-14-11-51-add-default-free-tier.js
+++ b/core/server/data/migrations/versions/4.33/2022-01-14-11-51-add-default-free-tier.js
@@ -1,0 +1,37 @@
+const {createTransactionalMigration} = require('../../utils');
+const ObjectID = require('bson-objectid');
+const {slugify} = require('@tryghost/string');
+const logging = require('@tryghost/logging');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const [result] = await knex
+            .count('id', {as: 'total'})
+            .from('products')
+            .where('type', 'free');
+
+        if (result.total !== 0) {
+            logging.warn(`Not adding default free tier, a free tier already exists`);
+            return;
+        }
+
+        const name = 'Free';
+        const id = ObjectID().toHexString();
+
+        logging.info(`Adding tier "${name}"`);
+        await knex('products')
+            .insert({
+                id: id,
+                name: name,
+                type: 'free',
+                slug: slugify(id),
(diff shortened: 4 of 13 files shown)


core/server/data/migrations/versions/4.33/01-add-type-column-to-products.js
```diff
@@ -0,0 +1,13 @@
...
+
+module.exports = utils.createAddColumnMigration(
+    'products',
+    'type',
+    {
+        type: 'string',
+        maxlength: 50,
+        nullable: false,
+        defaultTo: 'paid',
+        validations: {isIn: [['paid', 'free']]}
```
allouis, line 11: This isn't necessary as it's not used to construct the column

core/server/data/migrations/versions/4.33/02-add-default-free-product.js
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        const name = 'Free';
+
+        logging.info(`Adding tier "${name}"`);
+        await knex('products')
+            .insert({
+                id: ObjectID().toHexString(),
+                name: name,
+                description: 'Free Preview',
```
allouis, line 25: The fixture is missing the `description` column data

core/server/data/migrations/versions/4.33/02-add-default-free-product.js
```diff
@@ -0,0 +1,37 @@
...
+                type: 'free',
+                slug: slugify(name),
+                created_at: knex.raw(`CURRENT_TIMESTAMP`)
```
allouis, line 28: Can't remember if we need a `created_by` too - do we?

allouis (commented):
Few comments but looks good - is there a PR for the Members repo too?

core/server/data/migrations/versions/4.33/02-add-default-free-product.js
rishabhgrg: We don't use `created_by` anymore in the newer schemas, and `products` table doesn't has any such column so we are good to skip it.

core/server/data/migrations/versions/4.33/01-add-type-column-to-products.js
rishabhgrg: Updated, thanks!

core/server/data/migrations/versions/4.33/02-add-default-free-product.js
rishabhgrg: We decided to not have any default description for free tier, updated to remove from here and kept fixtures as-is. Good catch, thanks.

rishabhgrg:
@allouis Thanks for the quick review, updated now :) Members repo PR is [here](https://github.com/TryGhost/Members/pull/356)

daniellockyer:
@rishabhgrg The filenames of the migrations should be `yyyy-mm-dd-hh-mm-migration-desc.js` as per the new docs and tooling 🙂

core/server/data/migrations/versions/4.33/02-add-default-free-product.js
```diff
@@ -0,0 +1,37 @@
...
+        const name = 'Free';
+        const id = ObjectID().toHexString();
+
+        logging.info(`Adding tier "${name}"`);
+        await knex('products')
+            .insert({
+                id: id,
+                name: name,
+                type: 'free',
+                slug: slugify(id),
```
daniellockyer, line 27: Do we want to be `slugify`ing the `id`? Would the `name` not make more sense? 🤔

core/frontend/services/theme-engine/middleware/update-global-template-options.js
```diff
@@ -28,7 +28,9 @@ function calculateLegacyPriceData(products) {
         };
     }
 
-    const defaultProduct = products[0] || {};
```
allouis, line 31: I think we use the first product as the default product elsewhere too, in the Members repo - would you mind having a check for that and updating it anywhere?

core/server/data/migrations/versions/4.33/02-add-default-free-product.js
rishabhgrg: Yeah that's what we originally did, but we realized that might actually have conflicts for anyone that might have already created a paid tier with name `Free`, which is unlikely but possible 😬  We don't use slugs anywhere right now for a product, so to be on a safe side with the migration running across all sites that might have some used `slug`, we went with using the `id` as the slug to keep it unique

daniellockyer: Ah!

core/frontend/services/theme-engine/middleware/update-global-template-options.js
rishabhgrg: Yeah definitely, I am actively testing and updating all places the default product could be a problem with this change, will update anywhere I find in Members repo as well 👍

rishabhgrg:
@daniellockyer Updated the filename for migrations, thanks for catching 🙌

daniellockyer (approved):

allouis (approved):

PR #13978, opened 2022-01-15
:bug: Fixed AssetHelper not working with svg

### LINKED ISSUE #13971: AssetHelper not working with svg
### Issue Summary

The asset helper does not work with svg images using # because it adds a query param.

### Steps to Reproduce

```
 <svg class="icon icon__primary quote__line">
        <use xlink:href="{{asset "images/solid.svg#minus"}}"></use>
</svg>
```

Using this does not show the svg images it renders to:
xlink:href="/assets/images/solid.svg#minus?v=094ca0c660"

When I remove the param it works fine and also of cource if  I don't use the helper but this raises a gscan error

### Ghost Version

4.32.0

### Node.js Version

14.18.2

Docker Image

### Database type

SQLite3

### Browser & OS version

Chrome / safari

_No response_

### Code of Conduct

torstenzander:
closes #13971

This fixes an issue with svg sprites using the # to select an icon, when a ?v=hash is rendered the svg images is not shown. The bugfix removes ?v=hash param for svg files with an #.


ErisDS:
Hey @torstenzander thanks for the PR :) 

I don't think this is the fix that you want. The `?v=` query param is for cache busting, so this change would mean any svg with a # would get stuck in cache indefinitely.

The correct fix would be to insert the `?v=` param before the #, and maintain both.

Also, this bug would affect absolutely any path with a #, although it currently only really makes sense to use with an SVG, there's no need to limit the solution if implemented this way  🙂

test/unit/frontend/meta/asset-url.test.js
```diff
@@ -34,6 +34,11 @@ describe('getAssetUrl', function () {
         testUrl.should.equal('/public/myfile.js?v=' + config.get('assetHash'));
     });
 
+    it('should return no hash when svg#', function () {
+        const testUrl = getAssetUrl('myfile.svg#arrow-up');
+        testUrl.should.equal('/assets/myfile.svg#arrow-up');
```
ErisDS, line 39:
```
testUrl.should.equal(`/assets/myfile.svg?v=${config.get('assetHash')}#arrow-up`);
```

I'd expect the test to look like this :)

ErisDS (changes requested):
Hey @torstenzander thanks for the PR :)

I don't think this is the fix that you want. The ?v= query param is for cache busting, so this change would mean any svg with a # would get stuck in cache indefinitely.

The correct fix would be to insert the ?v= param before the #, and maintain both.

Also, this bug would affect absolutely any path with a #, although it currently only really makes sense to use with an SVG, there's no need to limit the solution if implemented this way 🙂

torstenzander:
I am not sure if I have to do anything, my tests cover everything I added even though the coverage report looks different.

ErisDS (approved):
Hey @torstenzander no you don't need to do anything. Thanks for your work :)

Issue #9942, opened 2018-10-03
"Site changed" webhook


kevinansfield:
Alongside the [API v2](https://github.com/TryGhost/Ghost/issues/9866) and new [auth mechanisms](https://github.com/TryGhost/Ghost/issues/9865) we'll be exposing our webhooks functionality via the admin UI and adding a new "site changed" webhook.

Brief feature outline:
- allow registration of webhooks via the admin UI
- add a new "site changed" webhook that can be used to trigger rebuilds when using static site generators such as gatsby.js with Netlify

jloh:
It'd be great to have webhooks validated like GitHub's are! https://developer.github.com/webhooks/securing/#validating-payloads-from-github

Apologies if this has already been considered, I can't see anything hinting towards it.

kirrg001:
@rishabhgrg Can we close this issue?

rishabhgrg:
@kirrg001 Keeping it open till the UI is enabled for this, will close after.

PR #13980, opened 2022-01-16
Added secret handling for webhooks

Regrau:
This change will allow to verify the source of the POST request.
The idea is to have the same functionality as in githubs
guide on securing webhooks.

https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-your-webhooks

With this change there is an additional header that will be sent to the
webhooks targetUrl and that can then verify the authenticity of
the source.

This will require another change in the admin client. That is already prepared and I will submit a
pull request there to.

The idea was already discussed in issue #9942 but the implementation seemed to be left behind.

If there is need for any other changes please let me know.

Regards
Georg

### DIFF
--- a/ghost/admin/.lint-todo
+++ b/ghost/admin/.lint-todo
@@ -954,3 +954,4 @@ add|ember-template-lint|no-passed-in-event-handlers|176|44|176|44|dcb4785647a508
 add|ember-template-lint|no-passed-in-event-handlers|186|44|186|44|70487c008d7dda453fef82f0140699ee93c0055c|1660521600000|1670893200000|1676077200000|app/components/modal-tier.hbs
 add|ember-template-lint|style-concatenation|303|54|303|54|23293f0c3838b23432d2b2daaf04b34504896d91|1660608000000|1670979600000|1676163600000|app/components/modal-tier.hbs
 remove|ember-template-lint|no-nested-interactive|23|28|23|28|5cf783a5684dda036706ff7438472e99c60a88e7|1658102400000|1668474000000|1673658000000|app/templates/whatsnew.hbs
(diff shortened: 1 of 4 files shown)


hwlmatt:
Hey @Regrau, this is great.
What about adding the timestamp to the signature header so receiving endpoint can check against that as well? 
Consider: [Stripe webhooks: replay attacks  ](https://stripe.com/docs/webhooks/signatures#replay-attacks)
```
...
'X-Ghost-Signature': `sha256=${crypto.createHmac('sha256', secret)
                                    .update(reqPayload).digest('hex')},t=${Date.now()}`
...
```

Regrau:
Hi @hwlmatt sounds reasonable so I added the timestamp.

@ErisDS Does this PR still need some work to be merged? If it's the tests I'd need a pointer to where they are located as I'm not quite sure where to start.

ErisDS:
I'm so sorry, I thought I left a comment on this aaaaages ago. 

Secret is optional, we shouldn't assume one is set, and the header shouldn't be added unless there is a secret set. The current PR adds the header no-matter what.

Tests would be nice, but not a deal-breaker here.

ErisDS:
The admin client is now back in this repo under `ghost/admin` so it should be much easier to get this change over the line now 😄

Regrau:
Hi @ErisDS,
this is now soooo much more convenient than working with submodules. That's awesome!

Now the header will appear only when there is a secret set for the webhook like you suggested.
Please have a look, if there is something else that needs doing let me know.

ErisDS:
Hey @Regrau I'm glad you found the monorepo much easier to work with! I've added a few tweaks to get this through linting and tests and merged it. Thank you so much for persevering with us. This feature will go out in the release at the end of this week 😄

PR #13985, opened 2022-01-17
Added MySQL 8 to CI tests

daniellockyer:
refs https://github.com/TryGhost/Toolbox/issues/181

- we officially support MySQL 8 but we don't run tests for it -- oops
- this commit adds MySQL 8 to the DB matrix in CI tests


daniellockyer:
This is blocked until we switch to the `mysql2` library, or the `mysql-action` supports allowing the legacy auth method. I'd prefer to wait until `mysql2` because this would be a good test that the auth works

PR #14038, opened 2022-01-21
Added pivot table to store tiers on post

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1071

We used `posts.visibility` originally to store visibility as `free|paid` with a character limit of 50. This same field was repurposed to store an NQL filter when member tiers is enabled. The NQL filter uses the slug of the tier name, which can easily create a filter longer than 50 characters, adding an unwanted limitation on number of tiers that can be added to post's visibility.
Going forward, we'd like to store the visibility of posts for tiers in a separate pivot table and instead store the value of `visibility` as `tiers` when restricting post access to specific tiers. This change -

- adds a new pivot table fixture for storing relation between posts and products
- adds a migration for creating the new table
- adds new table to db exporter list
- updates tests


allouis (approved):

daniellockyer (approved):

daniellockyer:
Blocking merge until we update to 4.35 🙂

daniellockyer (approved):

PR #14070, opened 2022-01-25
Add welcome page url

allouis:
refs TryGhost/Team#1168

Rather than using a single url for paid signup redirects, we want to
support setting a welcome page on a tier by tier basis. This column will
be used to store the URL. A text column of length 2000 is how we have
stored URL's elsewhere in the schema.


rshbhgrg (approved):

allouis:
@matthanley Good Q! We have some column size guides here - which I think this adheres to https://github.com/TryGhost/Ghost/blob/6fa8809406a417b210aa34164cb8a2dbda20bc69/core/server/data/schema/schema.js#L1-L9

Whether or not that's still relevant, I'm not sure - but my thinking was to store it on table based on pure intuition 😁 I'd be happy to change it either way - I think argument for string is slight perf improvement, argument for text is smaller table size

daniellockyer (approved):

PR #14083, opened 2022-01-30
    Migrated redirect settings to welcome_page_urls

allouis:
refs https://github.com/TryGhost/Team/issues/1168

This migrates the existing settings onto the Tier objects, so that users
with Tiers enabled can seamless move from global settings to Tier level
settings - without losing/modifying data/functionality.

### DIFF
--- a/core/server/data/migrations/versions/4.35/2022-01-30-15-17-set-welcome-page-url-from-settings.js
+++ b/core/server/data/migrations/versions/4.35/2022-01-30-15-17-set-welcome-page-url-from-settings.js
@@ -1,0 +1,39 @@
+const logging = require('@tryghost/logging');
+
+// For DML - data changes
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Setting welcome_page_url from settings');
+
+        const paidSignupRedirect = await knex()
+            .select('value')
+            .from('settings')
+            .where('key', 'members_paid_signup_redirect')
+            .first();
+
+        const freeSignupRedirect = await knex()
+            .select('value')
+            .from('settings')
+            .where('key', 'members_free_signup_redirect')
+            .first();
+
+        logging.info(`Setting paid Tiers welcome_page_url to ${paidSignupRedirect.value}`);
+
+        await knex('products')
+            .update('welcome_page_url', paidSignupRedirect.value)
+            .where('type', 'paid');
+
+        logging.info(`Setting free Tiers welcome_page_url to ${freeSignupRedirect.value}`);
+
+        await knex('products')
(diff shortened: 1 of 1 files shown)


rshbhgrg (approved):

core/server/data/migrations/versions/4.35/2022-01-30-15-17-set-welcome-page-url-from-settings.js
```diff
@@ -0,0 +1,39 @@
...
+            .from('settings')
+            .where('key', 'members_paid_signup_redirect')
+            .first();
+
+        const freeSignupRedirect = await knex()
+            .select('value')
+            .from('settings')
+            .where('key', 'members_free_signup_redirect')
+            .first();
+
```
daniellockyer, line 21: Can we check for the existence of `paidSignupRedirect` and `freeSignupRedirect` here? They __should__ exist but I'd like to prevent migrations crashing if not

allouis: Sure thing - I'll exit early if they're not present!

core/server/data/migrations/versions/4.35/2022-01-30-15-17-set-welcome-page-url-from-settings.js
```diff
@@ -0,0 +1,39 @@
+const logging = require('@tryghost/logging');
+
+// For DML - data changes
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Setting welcome_page_url from settings');
+
+        const paidSignupRedirect = await knex()
```
daniellockyer, line 10: @allouis I think knex expects the format of `await knex('settings')...`

```
[2022-01-31 14:23:23] INFO Setting welcome_page_url from settings
calling knex without a tableName is deprecated. Use knex.queryBuilder() instead.
calling knex without a tableName is deprecated. Use knex.queryBuilder() instead.
```

core/server/data/migrations/versions/4.35/2022-01-30-15-17-set-welcome-page-url-from-settings.js
allouis: I decided to wrap the inserts with checks - I think it's better because we may have just one

core/server/data/migrations/versions/4.35/2022-01-30-15-17-set-welcome-page-url-from-settings.js
allouis: Done!

daniellockyer (approved):

PR #14089, opened 2022-02-01
🐛 Added missing payment events in the member activity feed

tpatel:
refs https://github.com/TryGhost/Team/issues/1297

- They were expected to be in the activity feed but were hidden


tpatel:
duplicate

PR #14091, opened 2022-02-01
Migrated email filter columns from VARCHAR to TEXT

allouis:
refs TryGhost/Team#1287

Currently we have a hard limit of how large an email filter can be,
which is very restrictive once a site starts using Tiers - by moving
toward a TEXT column, we essentially give the filters unlimited size.

This currently doesn't handle SQLite as I'm waiting on feedback from
Matt on whether or not the development database needs this migration.

### DIFF
--- a/core/server/data/migrations/versions/4.35/2022-02-01-11-48-update-email-recipient-filter-column-type.js
+++ b/core/server/data/migrations/versions/4.35/2022-02-01-11-48-update-email-recipient-filter-column-type.js
@@ -1,0 +1,25 @@
+const logging = require('@tryghost/logging');
+const {createNonTransactionalMigration} = require('../../utils');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        if (knex.client.config.client === 'sqlite3') {
+            logging.warn('Skipping migration for SQLite3');
+            return;
+        }
+        logging.info('Changing posts.email_recipient_filter column from VARCHAR(50) to TEXT');
+        await knex.schema.alterTable('posts', function (table) {
+            table.text('email_recipient_filter').alter();
+        });
+    },
+    async function down(knex) {
+        if (knex.client.config.client === 'sqlite3') {
+            logging.warn('Skipping migration for SQLite3');
+            return;
+        }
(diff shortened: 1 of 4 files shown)


core/server/data/migrations/versions/4.35/2022-02-01-11-48-update-email-recipient-filter-column-type.js
```diff
@@ -0,0 +1,25 @@
...
+        });
+    },
+    async function down(knex) {
+        if (knex.client.config.client === 'sqlite3') {
+            logging.warn('Skipping migration for SQLite3');
+            return;
+        }
+        logging.info('Changing posts.email_recipient_filter column from TEXT to VARCHAR(50)');
+        knex.schema.alterTable('posts', function (table) {
+            table.string('email_recipient_filter', 50).alter();
```
matthanley, line 22: These `down()` operations will fail if content is longer than 50 chars:
```mysql
mysql> alter table posts modify email_recipient_filter varchar(50);
ERROR 1406 (22001): Data too long for column 'email_recipient_filter' at row 1
```

matthanley (changes requested):

core/server/data/migrations/versions/4.35/2022-02-01-11-48-update-email-recipient-filter-column-type.js
allouis: Smart! What shall we do? noop on `down` I guess?

allouis: Done

rshbhgrg (approved):

daniellockyer:
@allouis Can I check that we'll be adding the default to the model? Changing to a `text` removes the default on the column

allouis:
@daniellockyer Good question! Both models have a default set https://github.com/TryGhost/Ghost/blob/main/core/server/models/post.js#L68 https://github.com/TryGhost/Ghost/blob/main/core/server/models/email.js#L11

daniellockyer (approved):

daniellockyer:
I'm happy with this with the understanding we aren't providing a `down` to revert the column change to keep things simpler

PR #14102, opened 2022-02-03
:bug: Added test case for bug #13971

luisda0:
- Added test case to check the url of svg assets with hash so it appears after ?v=

- refs #13971
- The test case was added to check that the svg hash is added at the end instead of the middle so the asset helper can work properly with svg files with hashes at the end


ErisDS:
Hi there @luisda0! thank you for taking the time to submit a PR. This test case was already added in a previous PR which was merged this week & fixes the underlying issue: https://github.com/TryGhost/Ghost/pull/13978

Let us know if you find any more broken cases though!

PR #14104, opened 2022-02-03
Added new default content visibility tiers setting

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1071

Default content visibility for a post can be one of `public|members|paid|tiers`, where `tiers` denotes visibility restricted to specific tiers. This change adds a new setting to store the tier ids when default content visibility is set to `tiers`. This closely matches how the visibility is stored on `posts` table as well, with `visibility` stored as `tiers` and tiers data is stored on tiers pivot table.

### DIFF
--- a/core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
+++ b/core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
@@ -1,0 +1,147 @@
+const logging = require('@tryghost/logging');
+
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Checking default_content_visibility for specific tiers');
+
+        const settings = await knex('settings')
+            .select()
+            .whereIn('key', ['default_content_visibility', 'default_content_visibility_tiers']);
+        const contentVisibilitySetting = settings.find(d => d.key === 'default_content_visibility');
+        const visibilityTiersSetting = settings.find(d => d.key === 'default_content_visibility_tiers');
+        if (!contentVisibilitySetting) {
+            logging.warn('No default_content_visibility setting found.');
+            return;
+        }
+
+        if (!visibilityTiersSetting) {
+            logging.warn('No default_content_visibility_tiers setting found.');
+            return;
+        }
+
+        const contentVisibility = contentVisibilitySetting.value;
+
+        if (['public', 'members', 'paid'].includes(contentVisibility)) {
+            logging.info(`Ignoring default_content_visibility change as already set to ${contentVisibility}.`);
+            return;
+        }
+        // Transform visibility to tiers when stored as nql string
+        const isValidProductNqlFilter = /^(?:product:[\w-]+,?)+$/.test(contentVisibility);
+        const now = knex.raw('CURRENT_TIMESTAMP');
+        // Reset visibility value to paid if invalid string/filter
+        if (!isValidProductNqlFilter) {
+            logging.warn(`Found invalid default_content_visibility value - ${contentVisibility}, resetting to paid`);
+            await knex('settings')
+                .where({
+                    key: 'default_content_visibility'
+                })
+                .update({
+                    value: 'paid',
+                    updated_at: now
+                });
+
+            logging.info(`Resetting default_content_visibility_tiers to []`);
+            await knex('settings')
+                .where({
+                    key: 'default_content_visibility_tiers'
+                })
+                .update({
+                    value: JSON.stringify([]),
+                    updated_at: now
+                });
+            return;
+        }
+
+        // fetch product slugs from nql filter
+        const productSlugs = contentVisibility.split(',').map((segment) => {
+            return segment.replace('product:', '');
+        });
+
+        // get product ids for slugs
+        const products = await knex('products')
+            .select('id')
+            .whereIn('slug', productSlugs);
+        const productList = products.map((product) => {
+            return product.id;
+        });
+
+        logging.info(`Updating default_content_visibility to tiers`);
+        await knex('settings')
+            .where({
+                key: 'default_content_visibility'
+            })
+            .update({
+                value: 'tiers',
+                updated_at: now
+            });
+
+        logging.info(`Updating default_content_visibility_tiers to ${productList}`);
+        await knex('settings')
+            .where({
+                key: 'default_content_visibility_tiers'
+            })
+            .update({
+                value: JSON.stringify(productList),
+                updated_at: now
+            });
+    },
+    async function down(knex) {
+        logging.info('Reverting default_content_visibility for specific tiers');
+
+        const settings = await knex('settings')
+            .select()
+            .whereIn('key', ['default_content_visibility', 'default_content_visibility_tiers']);
+        const contentVisibilitySetting = settings.find(d => d.key === 'default_content_visibility');
+        const visibilityTiersSetting = settings.find(d => d.key === 'default_content_visibility_tiers');
+
+        const visibilityValue = contentVisibilitySetting && contentVisibilitySetting.value;
+        const visibilityTiersValue = visibilityTiersSetting && visibilityTiersSetting.value;
+
+        if (visibilityValue !== 'tiers') {
+            logging.info(`Ignoring default_content_visibility as is set to ${visibilityValue}.`);
+            return;
+        }
+
+        if (!visibilityTiersValue) {
+            logging.warn(`Ignoring, found empty default_content_visibility_tiers value`);
+            return;
+        }
+
+        try {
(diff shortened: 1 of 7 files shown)


core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
```diff
@@ -0,0 +1,107 @@
...
+            logging.warn('No default_content_visibility_tiers setting found.');
+            return;
+        }
+
+        const contentVisibility = contentVisibilitySetting.value;
+
+        // Transform visibility to tiers when stored as nql string
+        if (!['public', 'members', 'paid'].includes(contentVisibility)) {
+            // fetch product slugs from nql filter
+            const productSlugs = contentVisibility.split(',').map((segment) => {
```
allouis, line 29: Should we be validating the format of this string? ATM it could be any arbitrary NQL filter right?


From what I can tell we're expecting something like `product:slug,product:other-slug,product:etc`

I think a regex like `/^(?:product:[\w\-]+,?)+$/` should work to test this, and we can just use default values if the filter doesn't match - what do you think?

rishabhgrg: @allouis 🤔 Yea that's a good point, while ideally sites should not have any arbitrary NQL filter other than products or label, it's a good idea to handle the invalid ones and fallback to `paid` as default instead which is the highest level instead.

rishabhgrg: Updated ✔️

core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        if (!visibilityTiersSetting) {
+            logging.warn('No default_content_visibility_tiers setting found.');
+            return;
+        }
+
+        const contentVisibility = contentVisibilitySetting.value;
+        // Transform visibility to tiers when stored as nql string
+        if (!['public', 'members', 'paid'].includes(contentVisibility)) {
```
daniellockyer, line 26: Can we invert the logic here to early return and avoid the `else` + an extra layer of indentation?

```js
if (['public', 'members', 'paid'].includes(contentVisibility)) {
    logging.info(`Ignoring default_content_visibility change as set to ${contentVisibility}.`);
    return;
}
```

core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
```diff
@@ -0,0 +1,114 @@
...
+                // get product ids for slugs
+                const products = await knex('products')
+                    .select('id')
+                    .whereIn('slug', productSlugs);
+                newVisibilityTiersValue = products.map((product) => {
+                    return product.id;
+                });
+            }
+
+            logging.info(`Updating default_content_visibility to tiers`);
```
daniellockyer, line 46: This isn't strictly true, it should probably be 

```js
logging.info(`Updating default_content_visibility to ${newVisibilityValue}`);
```

core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
```diff
@@ -0,0 +1,114 @@
...
+        if (!visibilityTiersSetting) {
+            logging.warn('No default_content_visibility_tiers setting found.');
+            return;
+        }
+
+        const contentVisibility = contentVisibilitySetting.value;
+        // Transform visibility to tiers when stored as nql string
+        if (!['public', 'members', 'paid'].includes(contentVisibility)) {
+            const isValidProductNqlFilter = /^(?:product:[\w-]+,?)+$/.test(contentVisibility);
+            let newVisibilityValue = 'paid';
```
daniellockyer, line 28: So am I right in saying that if `contentVisibility` is not one of `[public,members,paid]`, and it's not a valid product NQL filter, then we set `contentVisibility` to `paid` anyway?

daniellockyer: Or should the setting updates be inside the if-statement?

core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
```diff
@@ -0,0 +1,114 @@
...
+        const settings = await knex('settings')
+            .select()
+            .whereIn('key', ['default_content_visibility', 'default_content_visibility_tiers']);
+        const contentVisibilitySetting = settings.find(d => d.key === 'default_content_visibility');
+        const visibilityTiersSetting = settings.find(d => d.key === 'default_content_visibility_tiers');
+
+        const visibilityValue = contentVisibilitySetting && contentVisibilitySetting.value;
+        const visibilityTiersValue = visibilityTiersSetting && visibilityTiersSetting.value;
+
+        if (visibilityValue === 'tiers' && visibilityTiersValue) {
```
daniellockyer, line 79: Preferably the same here with inverting the if-statement to early return

core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
rishabhgrg: > if contentVisibility is not one of [public,members,paid], and it's not a valid product NQL filter, then we set contentVisibility to paid anyway

@daniellockyer Yea that is correct, the reasoning behind this is to set the highest visibility restriction as default if someone in an unlikely scenario doesn't has the right NQL filter in place. We could as well set it to `public` instead as it only affects the default visibility, but my idea of using `paid` was to fallback to tighter visibility in case of an invalid value, wdyt ?

daniellockyer: Gotcha - makes sense 👍🏻

core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
rishabhgrg: ✅

core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
rishabhgrg: ✅

core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
rishabhgrg: ✅

core/server/data/migrations/versions/4.35/2022-02-02-13-10-transform-specific-tiers-default-content-visibility.js
```diff
@@ -0,0 +1,147 @@
...
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Checking default_content_visibility for specific tiers');
+
+        const settings = await knex('settings')
+            .select()
+            .whereIn('key', ['default_content_visibility', 'default_content_visibility_tiers']);
+        const contentVisibilitySetting = settings.find(d => d.key === 'default_content_visibility');
+        const visibilityTiersSetting = settings.find(d => d.key === 'default_content_visibility_tiers');
+        if (!contentVisibilitySetting || !contentVisibilitySetting.value) {
```
daniellockyer, line 14: If `default_content_visibility` === `''` (empty string), we'll stop here, but we actually want to reset the value to `paid` below right?

rishabhgrg: 🤔 umm so this is a tricky one, so i had done this intentionally thinking that a site must have really messed up to have an empty `default_content_visibility` as there is still a valid path to invalid filter but none to this. My thinking here was if we are not creating the setting here if its not found, we shouldn't also update the tiers value if a site managed to set it empty.

I'll remove the `|| !contentVisibilitySetting.value` from this condition so empty string is treated the same as invalid filter as on second thought that does seem a bit more consistent 👍

daniellockyer (approved):

allouis (approved):

PR #14115, opened 2022-02-04
Fixed missing portal product data

rshbhgrg:
refs  https://github.com/TryGhost/Team/issues/1311

For some sites, the portal_products array was created without any value and due to a possible bug in older version of Ghost, it also didn't get filled on Stripe connect with default product. This causes a side-effect of sites not showing the prices in Portal when tiers beta is enabled or is out as GA. This change populates the missing product data in `portal_product` for sites that have a single tier (haven't enabled tiers beta), as they right now don't have an option to hide the tier.

### DIFF
--- a/core/server/data/migrations/versions/4.35/2022-02-04-04-34-populate-empty-portal-products.js
+++ b/core/server/data/migrations/versions/4.35/2022-02-04-04-34-populate-empty-portal-products.js
@@ -29,27 +29,30 @@ module.exports = createTransactionalMigration(
             logging.warn(`Missing portal_products setting`);
             return;
         }
+        try {
+            const currPortalProductsValue = JSON.parse(portalProductsSetting.value);
 
-        const currPortalProductsValue = JSON.parse(portalProductsSetting.value);
+            if (currPortalProductsValue.length > 0) {
+                logging.warn(`Ignoring - portal_products setting is not empty, - ${currPortalProductsValue}`);
+                return;
+            }
 
-        if (currPortalProductsValue.length > 0) {
-            logging.warn(`Ignoring - portal_products setting is not empty, - ${currPortalProductsValue}`);
-            return;
+            const defaultProduct = products[0];
+            const portalProductsValue = [defaultProduct.id];
+
+            logging.info(`Setting portal_products setting to have product - ${defaultProduct.id}`);
+
+            const now = knex.raw('CURRENT_TIMESTAMP');
+
+            await knex('settings')
+                .where('key', 'portal_products')
+                .update({
+                    value: JSON.stringify(portalProductsValue),
+                    updated_at: now
+                });
+        } catch (e) {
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.35/2022-02-04-04-34-populate-empty-portal-products.js
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        const defaultProduct = products[0];
+        const portalProductsValue = [defaultProduct.id];
+
+        logging.info(`Setting portal_products setting to have product - ${defaultProduct.id}`);
+
+        await knex('settings')
+            .where('key', 'portal_products')
+            .update({value: JSON.stringify(portalProductsValue)});
```
daniellockyer, line 47: Should we also update the `updated_at` date here?

rishabhgrg: Ah yea we should, fixing.

core/server/data/migrations/versions/4.35/2022-02-04-04-34-populate-empty-portal-products.js
```diff
@@ -0,0 +1,56 @@
...
+            .where('key', 'portal_products')
+            .select('value')
+            .first();
+
+        if (!portalProductsSetting) {
+            logging.warn(`Missing portal_products setting`);
+            return;
+        }
+
+        const currPortalProductsValue = JSON.parse(portalProductsSetting.value);
```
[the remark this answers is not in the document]
rishabhgrg: Yea totally, done now. Had done it in the other migration because of this reason but missed here 🙈

core/server/data/migrations/versions/4.35/2022-02-04-04-34-populate-empty-portal-products.js
rishabhgrg: Done

core/server/data/migrations/versions/4.35/2022-02-04-04-34-populate-empty-portal-products.js
```diff
@@ -0,0 +1,59 @@
...
+
+            const now = knex.raw('CURRENT_TIMESTAMP');
+
+            await knex('settings')
+                .where('key', 'portal_products')
+                .update({
+                    value: JSON.stringify(portalProductsValue),
+                    updated_at: now
+                });
+        } catch (e) {
```
daniellockyer, line 53: It might just be worth adding `e` into the logging line below so we can see it if it happens

rishabhgrg: Done ✔️

daniellockyer (approved):

allouis (approved):

PR #14128, opened 2022-02-07
Updated members regression tests to use e2e framework

ErisDS:
This is an example refactor to the new framework it does not yet properly include snapshot testing [except for this one test](https://github.com/TryGhost/Ghost/pull/14128/files#diff-b60603f89c419b12b96f670acef12d003c5e452b35fd90501bd5286cfeb55e3bR39) because I think it's handy to see these changes in isolation.

The refactor achieves the following things:

- Test initialisation now has 3 separate steps, hiding away boot code that we don't want to care about, whilst making authentication and fixture management clearer
   1. Get an agent (includes booting Ghost)
   2. Optional Fixture management
   3. Optional Authentication
- The framework has methods for creating and tearing down common mocks (API needs more work)
- The test agent automatically handles common request features, reducing the test overhead
   - API Base URL is set, meaning no need to calculate URLs per test
   - Origin Header is set
   - Host Header is set

The overall aim is to have tests that are easy to write and easy to read.

Here's where I'd like some feedback:

The `express-test` library has a chained API very very similar to `supertest`, but slightly different.

The reason for the difference is quite simply readability.

There are some downsides - not counting refactoring cos that is only a problem once. First is cognitive dissonance - the new API is new and different - the second is interoperability - it's harder to swap between `express-test` and `supertest`.

What I'd like feedback on is: does anyone feel that the readability improvements are hugely beneficial, for either of these 2 cases:

1. When creating requests .set -> .header and .send -> .body
2. When checking requests instead of a single overloaded `.expect()` we have explicit methods e.g. `.expectStatus()`

Please see [this single test](https://github.com/TryGhost/Ghost/pull/14128/files#diff-b60603f89c419b12b96f670acef12d003c5e452b35fd90501bd5286cfeb55e3bR39) as the example of how e2e tests are meant to look once snapshot testing is in place:

- Simple API for making a request
- Simple API for checking the result
- Body and Headers use snapshots to check they match what's expected
- Sinon is used to mock key components and check side effects happen as expected

### DIFF
--- a/test/regression/api/admin/members.test.js
+++ b/test/regression/api/admin/members.test.js
@@ -1,11 +1,25 @@
 const querystring = require('querystring');
-const should = require('should');
-const sinon = require('sinon');
-const testUtils = require('../../../utils');
-const {agentProvider, mockManager, fixtureManager} = require('../../../utils/e2e-framework');
-const localUtils = require('./utils');
+const assert = require('assert');
+const {agentProvider, mockManager, fixtureManager, matchers} = require('../../../utils/e2e-framework');
+const {anyString, anyArray, anyObjectId, anyEtag, anyUuid, anyErrorId, anyDate} = matchers;
 
 let agent;
+
+const memberMatcherNoIncludes = {
+    id: anyObjectId,
+    uuid: anyUuid,
+    created_at: anyDate,
+    updated_at: anyDate
+};
+
+const memberMatcherShallowIncludes = {
+    id: anyObjectId,
+    uuid: anyUuid,
+    created_at: anyDate,
+    updated_at: anyDate,
+    subscriptions: anyArray,
+    labels: anyArray
+};
 
 describe('Members API', function () {
     before(async function () {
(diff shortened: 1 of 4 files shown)


allouis (commented):
I like the API - it's way clearer what it does, the readability is worth the incompatibility IMO. 

I think I could convert one request to another in under a minute tbh.


One annoying thing with supertest, which would be amazing to fix if possible, is that it error on the first error, I don't know how much work it would be to fix that, but maybe something like this could make fixing/writing tests easier:

```js
await agent
    .get('members/?order=email_open_rate%20asc')
    .expectHeader('Content-Type', /json/)
    .expectHeader('Cache-Control', testUtils.cacheRules.private)
    .expectStatus(200)
    .execute() // This could throw _all_ the errors (& maybe error/body info)
               // usually if the status code doesn't match we're
               // interested in the body/error message as to why
    .then((res) => {
        // etc...
    });
```

UPDATE:

I've also just realised, I'm guessing that the reason we're using the `then` method is so that you know when to run everything, with something like the above we could stick to more traditional `async`/`await` style like:
```js
const res = await agent
    .method()
    .method()
    .execute()
```

UPDATE 2:

I realise this is a little off topic, and I have _zero_ idea how difficult this would be, so I'm not expecting anything - just thought you may find it interesting

test/regression/api/admin/members.test.js
```diff
@@ -4,34 +4,36 @@ const should = require('should');
...
-        request = supertest.agent(config.get('url'));
-        await localUtils.doAuth(request, 'members');
+        agent = await framework.getAgent('/ghost/api/canary/admin/');
+        await framework.initFixtures('members');
+        await agent.loginAsOwner();
     });
 
     beforeEach(function () {
-        sinon.stub(mailService.GhostMailer.prototype, 'send').resolves('Mail is disabled');
+        emailStub = framework.stubMail();
```
allouis, line 32: I guess we'll need to setup this to be able to stub the Stripe API - but I think that'll be straightforward

allouis, line 32: I guess we'll need to setup this to be able to stub the Stripe API - but I think that'll be straightforward

ErisDS:
> One annoying thing with supertest, which would be amazing to fix if possible, is that it error on the first error, 

@allouis I have had the same frustration with supertest and had the same thought whilst building this library, but I have had two realisations:

1. Only reporting a single error is a limitation of _mocha_ not supertest, and so the new library only returns one error even though it could return many.
2. (class what's the problem) The problem is not that it fails on the first error, the problem is that the first error is often a missing header or incorrect status, which isn't the most useful failure AND that the error message then does not include the useful information - the error - which in our context is often encoded as JSON in the response body.

With these two things in mind, what can be done is this :

1. automatically sort the assertions so that body assertions are done first, then headers, then status.
2. if there is a `res.body.error`, decode it and show that.

I'm not sure if 1 won't actually make it super confusing (maybe) but I definitely think 2 would be very useful.

Aside: If there are _any_ problems with the error output whilst you're building tests with the new framework, then you should make & push a branch and send it to me so that I can fix it.

allouis:
> Only reporting a single error is a limitation of mocha not supertest, and so the new library only returns one error even though it could return many.

Ah yes - that make total sense

IMO 1. wouldn't make it _that_ confusing and would be helpful, both of those sound great but neither are essential! 

Cool, will do - I'll get started with trying to write a test with this :)

ErisDS:
@allouis this refactor to using snapshots is done - take a look!

It includes a few other things like reducing unnecessary requests and renaming tests to make more sense in .snap files.

There are still a couple of @TODOs in here that I'd like you to tackle once this is merged please :D 

@allouis and @rishabhgrg lemme know if this makes sense and then I'll merge!

test/regression/api/admin/members.test.js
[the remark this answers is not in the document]
ErisDS: Yeah I would expect a new mockStripe function to be added to the mockManager. Mock manager is a very simple thing at the moment, once we get a few more examples in there it’ll probably need a refactor.

As long as we keep mock-related code in there that will help us to see what mocking tools we need and make sure they are reusable and reused.

One of the big problems we have had is tooling being created for one test but no one knows about it so the next person does it differently…

allouis:
@ErisDS This looks good - feel free to merge!

PR #14139, opened 2022-02-09
Validate tier prices

allouis:
refs https://github.com/TryGhost/Team/issues/1319

This updates the members-api to explicitly validate Tier prices rather
than deferring to Stripe and using whichever errors they throw.

### DIFF
--- a/test/e2e-api/admin/products.test.js
+++ b/test/e2e-api/admin/products.test.js
@@ -1,0 +1,69 @@
+const should = require('should');
+const sinon = require('sinon');
+const labs = require('../../../core/shared/labs');
+
+const {agentProvider, fixtureManager} = require('../../utils/e2e-framework');
+
+describe('Tiers API', function() {
+    let agent;
+
+    before(async function() {
+        agent = await agentProvider.getAdminAPIAgent();
+        await fixtureManager.init('members');
+        await agent.loginAsOwner();
+    });
+
+    beforeEach(function() {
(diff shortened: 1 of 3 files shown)


test/e2e-api/admin/products.test.js
```diff
@@ -0,0 +1,69 @@
+const should = require('should');
+const sinon = require('sinon');
+const labs = require('../../../core/shared/labs');
+
+const {agentProvider, fixtureManager} = require('../../utils/e2e-framework');
+
+describe('Tiers API', function () {
```
naz, line 7: @allouis, I was messing with this suite today and have found that it can't be run in isolation, e.g if one sets `it.only('Errors when price is non-integer'` the test fails with:
```
1) Tiers API
       "before all" hook for "Errors when price is non-integer":
     Unable to create test agent. Could not understand request.
  MigrationScriptError: Could not understand request.
      at MigrationScriptError.KnexMigrateError (node_modules/knex-migrator/lib/errors.js:7:26)
      at new MigrationScriptError (node_modules/knex-migrator/lib/errors.js:25:26)
      at /home/naz/Workspace/Ghost/Ghost/node_modules/knex-migrator/lib/index.js:1032:19
  
  Error: select `products`.* from `products` where `products`.`slug` = 'default-product' limit 1 - SQLITE_ERROR: no such table: products
      at Child.<anonymous> (core/server/models/base/plugins/crud.js:144:31)
```

Didn't look any further, just noting here in case it's something that was missed

PR #14141, opened 2022-02-11
Moved authentication tests to e2e-api

ErisDS:
refs: https://github.com/TryGhost/Ghost/commit/3e7039c47d5b7fdf52fe36517e8180aaa7e562c5

- Moving these tests as they are the reference material for how we should be writing e2e tests

NOTE: this required https://github.com/TryGhost/Ghost/commit/3e7039c47d5b7fdf52fe36517e8180aaa7e562c5
- the authentication tests initialise theme storage during setup when testing that a theme can be selected
- then when the theme tests try to test the theme api, theme storage had cached the reference to the content
  folder meaning that the theme was being saved to the old folder.
- however our package-json utils read the theme package.json using a fresh path, so it was looking for the
  theme in the new folder


ErisDS:
How it failed when I originally moved it: https://github.com/TryGhost/Ghost/runs/5155149066?check_suite_focus=true
How it failed after I applied what should have fixed it: https://github.com/TryGhost/Ghost/runs/5161333124?check_suite_focus=true
Turns out that's a DB issue: https://github.com/TryGhost/Ghost/runs/5161802513?check_suite_focus=true which I found out with [this ugly commit](https://github.com/TryGhost/Ghost/pull/14141/commits/dbe90f1ae53013080687259e763fc7517c5c0787)

And then I tried to fix that and now the unit tests fail https://github.com/TryGhost/Ghost/actions/runs/1831383395? 😱 

I am out of 🧠

PR #14145, opened 2022-02-14
🐛 Fixed {{t}} helper not working in combination with concat subexpression

albogdano:
The following Handlebars expression does not work as expected: `{{t (concat label "-en")}}`.
The `concat` helper returns a `SafeString` which does not pass the `isString` in [i18n.js#L169](https://github.com/albogdano/Ghost/blob/1b0a69efdbc8902c14c66b9e3ce7be12ea634daf/core/frontend/services/theme-engine/i18n/i18n.js#L169).

Since we have no easy way of switching to another language from the clientside, the only way to easily add i18n support in a theme is to have a single `locale.json` file with strings for two or more languages like so:
```json
{
    "Home-en": "Home",
    "Home-de": "Anfang",
}
```
Then inside the `.hbs` templates, we can add `{{t (concat key (block "lang"))}}`. This simple change makes it possible to achieve that.

### DIFF
--- a/core/frontend/services/theme-engine/i18n/i18n.js
+++ b/core/frontend/services/theme-engine/i18n/i18n.js
@@ -163,6 +163,7 @@ class I18n {
         const options = merge({log: true}, opts || {});
         let candidateString;
         let matchingString;
+        msgPath = msgPath.toString();
 
         // no path? no string
         if (msgPath.length === 0 || !isString(msgPath)) {


benjaminrancourt:
For the Ghost team, if you are able (or have any idea how) to fix this problem wherever where we want to use the concat helper, that would be really great! Maybe directly inside the concat helper? 🤔

Another case:

```handlebars
    {{!-- Correct !--}}
    {{~#get "posts" limit="1" filter="tag:en+id:-{{post.id}}" as |sidebar_latest_post|~}}
      {{~> "sidebar/sidebar-story" widgetTitle=(t "Latest Posts") widgetPosts=sidebar_latest_post~}}
    {{~/get~}}

    {{!-- Incorrect !--}}
    {{~#get "posts" limit="1" filter=(concat "tag:en+id:-{{post.id}}") as |sidebar_latest_post|~}}
      {{~> "sidebar/sidebar-story" widgetTitle=(t "Latest Posts") widgetPosts=sidebar_latest_post~}}
    {{~/get~}}

    {{!-- Incorrect !--}}
    {{~#get "posts" limit="1" filter=(concat "tag:" (block "lang") "+id:-{{post.id}}") as |sidebar_latest_post|~}}
      {{~> "sidebar/sidebar-story" widgetTitle=(t "Latest Posts") widgetPosts=sidebar_latest_post~}}
    {{~/get~}}
```

image

PR #14176, opened 2022-02-18
[WIP] Welcome page tests

allouis:
🍦

### DIFF
--- a/test/e2e-api/members/signin.test.js
+++ b/test/e2e-api/members/signin.test.js
@@ -2,7 +2,6 @@ const membersService = require('../../../core/server/services/members');
 const {agentProvider, mockManager, fixtureManager} = require('../../utils/e2e-framework');
 
 let membersAgent;
-let adminAgent;
 
 describe('Members Signin', function () {
     before(async function () {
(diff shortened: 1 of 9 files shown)


test/utils/e2e-framework-mock-manager.js
```diff
@@ -37,6 +37,9 @@ const setupStripe = () => {
         if (key === 'stripe_connect_account_name') {
             return 'Test Account';
         }
+        if (key === 'theme_session_secret') {
```
allouis, line 40: This should probably be in like `setupMembers` function, but I think this is showing a clear pattern that we need to be able to stub settings before getting an agent/booting Ghost

test/e2e-api/members/signin.test.js
```diff
@@ -0,0 +1,106 @@
...
+    it('Will redirect to the free welcome page for signup', async function () {
+        const magicLink = await membersService.api.getMagicLink('member1@test.com');
+        const magicLinkUrl = new URL(magicLink);
+        const token = magicLinkUrl.searchParams.get('token');
+
+        const {body: {products}} = await adminAgent.get('/products/');
+
+        const freeProduct = products.find(product => product.type === 'free');
+
+        await adminAgent.put(`/products/${freeProduct.id}/`).body({
```
allouis, line 59: I don't particularly like this, but the way product fixtures work are really weird because they are fixtures in the database too, this was the quickest way to get things working.

PR #14182, opened 2022-02-21
Backfilled the members `last_seen_at` column

tpatel:
refs https://github.com/TryGhost/Team/issues/1305

- Skipped sqlite3 compatibility due to the lack of join support in update
- Uses a raw migration to improve performance

This PR takes into account the cases where there were a member login or not. And if there is already a `last_seen_at` value or not.
There should never be a `last_seen_at` value but I'm not sure what kind of convoluted migration flows could happen.

I've added a `down` function, which is great to test the migration. However in the future, `last_seen_at` will be changed based on several events (login, email open), so running the down migration could delete data that couldn't be recovered by the backfilling. So I'm ok to delete/empty the `down` function if you think I should.

### DIFF
--- a/core/server/data/migrations/versions/4.37/2022-02-21-09-53-backfill-members-last-seen-at-column.js
+++ b/core/server/data/migrations/versions/4.37/2022-02-21-09-53-backfill-members-last-seen-at-column.js
@@ -1,0 +1,31 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        if (knex.client.config.client === 'sqlite3') {
+            logging.warn('Skipping migration for SQLite3');
+            return;
+        }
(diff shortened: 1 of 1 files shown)


allouis (commented):
It's possible, but unlikely, that a member does not have a login event associated with them - do you need to handle that case with a default date? Otherwise it looks good!

tpatel:
> It's possible, but unlikely, that a member does not have a login event associated with them - do you need to handle that case with a default date? Otherwise it looks good!

Good point! Do we want to avoid `NULLs` in this column?

The only other valid data would be the signup date, although the member might still not have been on the publication if they have signed-up without logging in (member import for example).

tpatel:
@allouis After thinking through it, I'm thinking I should merge this. There is still an option on the table to create a new migration to backfill remaining NULLs using the member `created_at` date! I'll ask on the product channel.

allouis (approved):
Cool sounds good - don't forget we need @daniellockyer to sign off before merging tho!

daniellockyer (approved):

PR #14194, opened 2022-02-21
Added `accentColor` and `description` parameters to `/authentication/setup` route

aileen:
refs https://github.com/TryGhost/Team/issues/1382

- Added two possible new setup values: `accentColor` and `description` to define the brand colour and site description on initial setup
- Updated tests to reflect those changes
- Only the params when passed and fall back to default site description

### DIFF
--- a/core/server/services/auth/setup.js
+++ b/core/server/services/auth/setup.js
@@ -77,7 +77,7 @@ async function doSettings(data, settingsAPI) {
     const context = {context: {user: data.user.id}};
     const user = data.user;
     const blogTitle = data.userData.blogTitle;
-    const description = data.userData.description;
+    const description = data.userData.description ? data.userData.description.trim() : null;
 
     let userSettings;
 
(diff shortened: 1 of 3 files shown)


core/server/services/auth/setup.js
```diff
@@ -77,18 +77,29 @@ async function doSettings(data, settingsAPI) {
...
 
     let userSettings;
 
     if (!blogTitle || typeof blogTitle !== 'string') {
         return user;
     }
 
+    if (!description || typeof description !== 'string') {
+        return user;
+    }
```
tpatel, line 90: I think this will cause an issue: a setup call with a title but without a description will prevent the title from being set. I would remove the IF completely (unless I've missed something).

tpatel (changes requested):
I think an `if` might break the install process for self hosters who can't set the description in the setup screen.
image

core/server/services/auth/setup.js
aileen: This should be good now! Thanks for checking @tpatel 🤗

tpatel (approved):
Looks great 🙌

PR #14197, opened 2022-02-22
Update the `last_seen_at` property on email open and member page view

tpatel:
refs https://github.com/TryGhost/Team/issues/1306

- Update on email open in the `email-analytics` event processor.
- Update on page view using a middleware in the frontend site setup.

More details in each commit.


core/server/services/email-analytics/lib/event-processor.js
```diff
@@ -88,6 +88,16 @@ class GhostEventProcessor extends EventProcessor {
                 opened_at: this.db.knex.raw('COALESCE(opened_at, ?)', [moment.utc(event.timestamp).format('YYYY-MM-DD HH:mm:ss')])
             });
 
+        await this.db.knex('members')
+            .where('email', '=', event.recipientEmail)
+            .andWhere(builder => builder
+                .where('last_seen_at', '<', moment.utc(event.timestamp).subtract(24, 'hours').format('YYYY-MM-DD HH:mm:ss'))
```
matthanley, line 94: I think we can improve this a little - we want the behaviour to be "updated once each day" rather than "updated once every 24 hours". i.e. if I visit at 2pm on Monday and then again at 1pm on Tuesday my last_seen should be Tuesday. So I think we can change this to something like (untested)

```suggestion
                .where('last_seen_at', '<', moment.utc(event.timestamp).format('YYYY-MM-DD 00:00:00'))
```

(we can treat last_seen as a date rather than a datetime)

tpatel: The problem you've mentioned would happen in the solution you've shared too. It would happen when a member is visiting at 00:01 UTC and then at 23:59 UTC.
Your solution also means there would be two writes in a short period of time for every member visiting around midnight UTC.

I think the best solution would be to use "updated once every 12 hours" if we require this level of precision, but then it wouldn't match the pitch requirements.

I would choose the "every 12 hours" solution, however I lack context on the reason behind the "last_seen only needs to be updated once in a given 24-hour period" sentence of the pitch. What do you think?

tpatel:
PR pending on publishing the module added in https://github.com/TryGhost/Members/pull/370

core/server/services/email-analytics/lib/event-processor.js
matthanley: The specific wording about a 24-hour period in the pitch is misleading - the thing we care about is the _date_, not the specific time. `last_seen_at` could easily have been just a date field:
> we will not need a write every time a user opens lands on a page, **just the first time that day**.

tpatel: I'm afraid moving to storing `last_seen_at` as a date will have negative consequences for the produce experience.

There will be different kinds of issues depending on how we display the date in the frontend, illustrated here by one member visiting a San Francisco publication (UTC-7) at 3pm and 5pm local time.

### Keeping the UTC date

In this example we just truncate the datetime to 00:00:00 UTC, and display it directly to the publication admins.

- member visits at 1am UTC on Feb, 24th
- last_seen_at is set to 2022-02-24
- The local datetime is 5pm Feb, 23th, the display time is 24th Feb, so the publication admins will see a `last_seen_at` in the future

### Displaying the date in local datetime

In this example we first truncate the datetime to 00:00:00 UTC, then convert it to local time and then display it to the publication admins.

- member visits at 11pm UTC on Feb, 24th
- last_seen_at is set to 2022-02-24
- we convert the date to local time 2022-02-24 00:00:00 UTC = 2022-02-23 16:00:00 UTC
- The local datetime is 3pm Feb, 24th, the display date is Feb, 23th, so the publication admins will see a `last_seen_at` on the previous day

### Conclusion

I recommend not to truncate the `last_seen_at` value. Instead I would recommend to switch to "every 12h" pattern as we have a 1 day precision when displaying the data. The tradeoff seems correct as records will be updated at most twice a day.

I'm fine implementing the date truncation if you are ok with the product limitations it will create.

core/frontend/web/site.js
```diff
@@ -171,6 +173,13 @@ module.exports = function setupSiteApp(options = {}) {
         }
     });
 
+    siteApp.use(function (req, res, next) {
+        if (req.member) {
+            DomainEvents.dispatch(MemberPageViewEvent.create({url: req.url, memberId: req.member.id}, new Date()));
```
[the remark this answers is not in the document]
tpatel: Passing a full member isn't a pattern I found in any events in `member-events`. What do you think about adding an extra `memberLastSeenAt` instead?
For example, `MemberPaidConversionEvent` uses `memberId` and `memberStatus` to pass data about a member: https://github.com/TryGhost/Members/blob/main/packages/member-events/lib/MemberPaidConversionEvent.js#L3-L4
I do like the idea of avoid database queries 👍

core/server/services/email-analytics/lib/event-processor.js
matthanley: Right - but my proposal wasn't to truncate the date that we store or to switch to a date field, it was to truncate the date that we use for comparison.

The difference is that it gives us a regular, consistent 24h window (== once a day) in which we update the member vs a sliding 12h window.

With the same example:
* member visits at 1am UTC on Feb, 24th
* last_seen_at is set to 2022-02-24 01:00:00
* The local datetime is 5pm Feb 23rd, the display time is 23rd Feb (we display in local timezone)

core/frontend/web/site.js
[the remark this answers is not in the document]
matthanley: That would work 👍🏼 I think it'll need a comment about why we're passing the last seen property, because it's not immediately obvious why it would be relevant for a MemberPageViewEvent

core/server/services/email-analytics/lib/event-processor.js
tpatel: Oh from your previous message I understood that you were suggesting to make `last_seen_at` a date, my bad.
> last_seen_at could easily have been just a date field

### comparing with a truncated date

Do you agree that the first problem you've mentioned will still be present when comparing to a truncated date?

> i.e. if I visit at 2pm on Monday and then again at 1pm on Tuesday my last_seen should be Tuesday.

In the above example that you wrote, if your publication is in UTC-10, then the visit at 1pm on Tuesday will be on the same UTC day.

So depending on the timezone of your publication, there will always be such a problem (although it will be bound to specific times while the 24h isn't).

### regular 24h window

Are consistent 24h windows a good thing? It will create more writes right after midnight UTC across all publications.

tpatel: I've chosen to disagree and commit to the UTC-based window. This has been implemented in the other PR first: https://github.com/TryGhost/Members/pull/370/commits/ff46449ad6069b53cef85a8c2f78daed672668ed

core/server/services/email-analytics/lib/event-processor.js
```diff
@@ -88,6 +88,17 @@ class GhostEventProcessor extends EventProcessor {
                 opened_at: this.db.knex.raw('COALESCE(opened_at, ?)', [moment.utc(event.timestamp).format('YYYY-MM-DD HH:mm:ss')])
             });
 
+        const {value: timezone} = await this.db.knex('settings').first('value').where('key', 'timezone');
+        await this.db.knex('members')
+            .where('email', '=', event.recipientEmail)
+            .andWhere(builder => builder
+                .where('last_seen_at', '<', moment.utc(event.timestamp).tz(timezone).startOf('day').format('YYYY-MM-DD HH:mm:ss'))
```
matthanley, line 95: There's an error here, because the `format()` is called on the date in the site's timezone so will always be `00:00:00`. It needs another call to `utc()` after the `startOf()` to compare against the UTC date in the db:

```
> moment.utc().tz('Europe/Paris').startOf('day').format('YYYY-MM-DD HH:mm:ss')
'2022-03-01 00:00:00'
> moment.utc().tz('Europe/Paris').startOf('day').utc().format('YYYY-MM-DD HH:mm:ss')
'2022-02-28 23:00:00'
```

matthanley (changes requested):

core/server/services/email-analytics/lib/event-processor.js
tpatel: I tried to make a failing test on https://github.com/TryGhost/Members/blob/main/packages/members-events-service/lib/last-seen-at-updater.js but it couldn't fail. This is a bug specific to the SQL implementation so I'm just adding this missing `.utc()` and we're good

core/server/services/email-analytics/lib/event-processor.js
```diff
@@ -88,6 +88,17 @@ class GhostEventProcessor extends EventProcessor {
                 opened_at: this.db.knex.raw('COALESCE(opened_at, ?)', [moment.utc(event.timestamp).format('YYYY-MM-DD HH:mm:ss')])
             });
 
+        const {value: timezone} = await this.db.knex('settings').first('value').where('key', 'timezone');
```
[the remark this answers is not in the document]
tpatel: I've added a fallback ✅ 

Could you share in which case it would happen?
I was assuming that a DB without a `timezone` setting could only be the result of a migration, in which case the person writing the migration would make sure nothing would break in the whole Ghost codebase.

matthanley (approved):
Not tested locally but lgtm

core/server/services/email-analytics/lib/event-processor.js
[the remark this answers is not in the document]
matthanley: It _should_ always be present, but I think this is one of those cases where it's better to be defensive, especially as we're not going through the settings cache for this.

PR #14215, opened 2022-02-28
✨ Added >, <, >=, and <= operators to match helper 

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1386

- The current match handler supports normal (in)equality operators, but no numeric comparisons (<, >, <=, >=)
- A use case for these new operators is to show the latest post in a separate way from other posts

### DIFF
--- a/core/frontend/helpers/match.js
+++ b/core/frontend/helpers/match.js
@@ -45,6 +45,9 @@ const handleMatch = (data, operator, value) => {
     case '!=':
         result = data !== value;
         break;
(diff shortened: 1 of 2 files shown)


matthanley (approved):

matthanley:
@SimonBackx really nice 👍🏼 Could you also PR a change to the docs ([here](https://github.com/TryGhost/Ghost.org/blob/main/content/docs/themes/helpers/match.md)) to add this functionality?

PR #14217, opened 2022-02-28
🐛 Fixed 500 error when creating product benefits without name

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1188

The products API currently returns a 500 error when you try to store product benefits with an empty name. This should throw a 422 Validation Error instead.

- Updated bookshelf-relations to 2.4.0, which includes a fix for this issue (validation error was not rethrown correctly)
- Added test when creating a new product with an empty benefit name
- Added test when updating an existing product with new benefits, with an empty name
- Added a test that creates a new product with benefits

### DIFF
--- a/test/e2e-api/admin/products.test.js
+++ b/test/e2e-api/admin/products.test.js
@@ -64,4 +64,34 @@ describe('Tiers API', function () {
             .body({products: [tier]})
             .expectStatus(422);
     });
+
+    it('Errors when a benefit has an empty name', async function () {
+        const tier = {
+            name: 'Blah',
+            monthly_price: {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -116,7 +116,7 @@
     "bluebird": "3.7.2",
     "body-parser": "1.19.2",
     "bookshelf": "1.2.0",
-    "bookshelf-relations": "2.3.0",
+    "bookshelf-relations": "2.4.0",
```
matthanley, line 119: This has since been updated, so just needs a rebase

matthanley (approved):

PR #14220, opened 2022-03-01
Added /tiers endpoint to Content API

allouis:
refs https://github.com/TryGhost/Team/issues/1313

When adding the tiers endpoint the Content API was missed, this is
needed so that themes can access Tiers via the `{{#get}}` helper.

### DIFF
--- a/core/frontend/helpers/get.js
+++ b/core/frontend/helpers/get.js
@@ -31,6 +31,9 @@ const RESOURCES = {
     },
     authors: {
         alias: 'authorsPublic'
+    },
+    tiers: {
+        alias: 'tiersPublic'
     }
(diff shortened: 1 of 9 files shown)


allouis:
@matthanley This still needs a test - but should I update the `{{#get}}` helper as part of this so that it works with themes?

matthanley (approved):
Not tested locally, but lgtm

matthanley:
@allouis could you also make sure to track related changes to the docs and GScan for switching from `@products` to `#get tiers` before this is merged

allouis:
Added an issue here: https://github.com/TryGhost/Team/issues/1405

PR #14221, opened 2022-03-01
Added visible column to products table

allouis:
refs https://github.com/TryGhost/Team/issues/1387

We are moving away from the portal_products setting to instead store
each tiers visiblity on the Tier itself. This column will be used for
that data.

Both of the default Tiers should be visible, but newly created tiers
should not be.

### DIFF
--- a/core/server/data/migrations/versions/4.38/2022-03-01-08-46-add-visibility-to-tiers.js
+++ b/core/server/data/migrations/versions/4.38/2022-03-01-08-46-add-visibility-to-tiers.js
@@ -3,5 +3,5 @@ const {createAddColumnMigration} = require('../../utils');
 module.exports = createAddColumnMigration('products', 'visible', {
     type: 'boolean',
     nullable: false,
-    defaultTo: true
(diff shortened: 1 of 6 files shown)


allouis:
A second migration will handle updating the `visible` column for existing Tiers, based off of the `portal_products` setting.

matthanley (commented):
I can see why you've implemented it that way, but having a different default behaviour in the schema vs the model feels like it could be super confusing later on.

Does it make more sense to default to false everywhere, and then flip them back to true in the second migration for each tier currently enabled in the setting?

allouis:
> Does it make more sense to default to false everywhere...

Yes it does :) Will update

matthanley (approved):
lgtm once the integrity test has been updated 👍🏼

rshbhgrg (approved):

daniellockyer (approved):

ErisDS:
Can we _please_ call this column `visibility`? We have a pattern for that, and none for `visible`.

PR #14249, opened 2022-03-01
Switched to util for retrieving DB info in migrations

daniellockyer:
refs https://github.com/TryGhost/Toolbox/issues/174

- right now, our migrations manually check the client of the knex
  instance to see whether we're running on MySQL or SQLite
- that's been working fine, but the problem is that we're due to switch
  to the mysql2 driver soon, so all these checks will be faulty
- i've altered the functionality of `@tryghost/database-info` to accept
  a knex instance, and it'll return if the DB is MySQL or SQLite in some
  helper functions
- this commit bumps the package and switches to that format
- originally I used a shared instance of the class within
  `@tryghost/database-info` but there's a chance that the knex instance
  inside migrations actually comes from knex-migrator, and not Ghost, so
  that wouldn't work

### DIFF
--- a/core/server/data/migrations/versions/3.29/01-remove-duplicate-subscriptions.js
+++ b/core/server/data/migrations/versions/3.29/01-remove-duplicate-subscriptions.js
@@ -1,4 +1,5 @@
 const logging = require('@tryghost/logging');
+const DatabaseInfo = require('@tryghost/database-info');
 
 module.exports = {
     config: {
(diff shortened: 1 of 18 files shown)


core/server/data/db/connection.js
```diff
@@ -9,7 +9,7 @@ let knexInstance;
 function configure(dbConfig) {
     const client = dbConfig.client;
 
-    if (client === 'sqlite3') {
```
daniellockyer, line 12: I can't use my util here because this is before we init the knex instance

sam-lord (approved):

PR #14253, opened 2022-03-02
Migrated visible column from portal settings

allouis:
https://github.com/TryGhost/Team/issues/1387

This is split into two migrations, one for the portal_products setting
and one for the portal_plans setting, as dealing with both of them in a
single migration led to too many branches.

### DIFF
--- a/core/server/data/migrations/versions/4.38/2022-03-01-15-42-update-products-visible-column.js
+++ b/core/server/data/migrations/versions/4.38/2022-03-01-15-42-update-products-visible-column.js
@@ -1,0 +1,36 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const portalProductSetting = await knex('settings').select('value').where('key', 'portal_products').first();
+
+        if (!portalProductSetting) {
+            logging.warn('Could not find portal_products setting - skipping migration');
+            return;
+        }
+
+        try {
+            const settingData = JSON.parse(portalProductSetting.value);
+
+            if (settingData.length === 0) {
+                logging.warn(`portal_product is empty, skipping migrations`);
+                return;
+            }
+
+            logging.info(`Updating ${settingData.length} products to visible, ${settingData}`);
+            await knex('products').update('visible', true).whereIn('id', settingData);
+        } catch (err) {
+            logging.warn('portal_products setting is invalid - skipping migration');
+            return;
+        }
+    },
+    async function down(knex) {
+        const visibleTiers = await knex('products').select('id').where('visible', true);
+
+        const settingData = JSON.stringify(visibleTiers.map(obj => obj.id));
+
+        logging.info(`Updating portal_products to ${settingData}`);
+        await knex('settings').update('value', settingData).where('key', 'portal_products');
+    }
+);
--- a/core/server/data/migrations/versions/4.38/2022-03-02-09-10-update-free-products-visible-column.js
+++ b/core/server/data/migrations/versions/4.38/2022-03-02-09-10-update-free-products-visible-column.js
@@ -21,6 +21,7 @@ module.exports = createTransactionalMigration(
             logging.info(`Updating free products to visible`);
             await knex('products').update('visible', true).where('type', 'free');
         } catch (err) {
+            logging.error(err);
             logging.warn('portal_plans setting is invalid - skipping migration');
             return;
         }
@@ -48,15 +49,16 @@ module.exports = createTransactionalMigration(
                     logging.info('portal_plans setting already contains "free" - skipping update');
                     return;
                 } else {
-                    settingData = existingSettingData.concat('free');
+                    settingData = JSON.stringify(existingSettingData.concat('free'));
                 }
             } else {
-                settingData = existingSettingData.filter(value => value !== 'free');
+                settingData = JSON.stringify(existingSettingData.filter(value => value !== 'free'));
             }
 
             logging.info(`Updating portal_plans to ${settingData}`);
-            await knex('settings').update('value', JSON.stringify(settingData)).where('key', 'portal_plans');
+            await knex('settings').update('value', settingData).where('key', 'portal_plans');
         } catch (err) {
+            logging.error(err);
             logging.warn('portal_plans setting is invalid - skipping migration');
             return;
         }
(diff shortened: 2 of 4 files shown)


core/server/data/migrations/versions/4.38/2022-03-02-09-10-update-free-products-visible-column.js
```diff
@@ -0,0 +1,63 @@
...
+            return;
+        }
+
+        try {
+            const existingSettingData = JSON.parse(portalPlanSetting.value);
+            let settingData;
+
+            if (freeTier.visible) {
+                if (existingSettingData.includes('free')) {
+                    settingData = existingSettingData;
```
matthanley, line 48: Would it be better to have an early return here?

allouis: Would save a DB query so yeah I guess

allouis, line 48:
```suggestion
                    logging.info('portal_plans setting already contains "free" - skipping update');
                    return;
```

core/server/data/migrations/versions/4.38/2022-03-02-09-10-update-free-products-visible-column.js
```diff
@@ -0,0 +1,64 @@
...
+
+            if (!settingData.includes('free')) {
+                logging.warn(`portal_plans does not include "free" - skipping migration`);
+                return;
+            }
+
+            logging.info(`Updating free products to visible`);
+            await knex('products').update('visible', true).where('type', 'free');
+        } catch (err) {
+            logging.warn('portal_products setting is invalid - skipping migration');
```
rishabhgrg, line 24: Small copy pasta typo - should be `portal_plans setting is invalid` instead of `portal_products`.

rshbhgrg (approved):
Needs small typo fixes but looks good overall :)

core/server/data/migrations/versions/4.38/2022-03-02-09-10-update-free-products-visible-column.js
allouis:
```suggestion
            logging.warn('portal_plans setting is invalid - skipping migration');
```

core/server/data/migrations/versions/4.38/2022-03-02-09-10-update-free-products-visible-column.js
```diff
@@ -0,0 +1,64 @@
...
+                    settingData = existingSettingData.concat('free');
+                }
+            } else {
+                settingData = existingSettingData.filter(value => value !== 'free');
+            }
+
+            logging.info(`Updating portal_plans to ${settingData}`);
+            await knex('settings').update('value', JSON.stringify(settingData)).where('key', 'portal_plans');
+        } catch (err) {
+            logging.warn('portal_products setting is invalid - skipping migration');
```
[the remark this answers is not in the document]
allouis:
```suggestion
            logging.warn('portal_plans setting is invalid - skipping migration');
```

core/server/data/migrations/versions/4.38/2022-03-01-15-42-update-products-visible-column.js
```diff
@@ -0,0 +1,36 @@
...
+            const settingData = JSON.parse(portalProductSetting.value);
+
+            if (settingData.length === 0) {
+                logging.warn(`portal_product is empty, skipping migrations`);
+                return;
+            }
+
+            logging.info(`Updating ${settingData.length} products to visible, ${settingData}`);
+            await knex('products').update('visible', true).whereIn('id', settingData);
+        } catch (err) {
```
daniellockyer, line 23: Can we log the `err` out here? If we don't, we won't know if it failed because the JSON.parse failed, or the DB update failed etc

core/server/data/migrations/versions/4.38/2022-03-02-09-10-update-free-products-visible-column.js
```diff
@@ -0,0 +1,64 @@
...
+            const settingData = JSON.parse(portalPlanSetting.value);
+
+            if (!settingData.includes('free')) {
+                logging.warn(`portal_plans does not include "free" - skipping migration`);
+                return;
+            }
+
+            logging.info(`Updating free products to visible`);
+            await knex('products').update('visible', true).where('type', 'free');
+        } catch (err) {
```
daniellockyer, line 23: Same here with logging the `err`

core/server/data/migrations/versions/4.38/2022-03-02-09-10-update-free-products-visible-column.js
```diff
@@ -0,0 +1,64 @@
...
+                } else {
+                    settingData = existingSettingData.concat('free');
+                }
+            } else {
+                settingData = existingSettingData.filter(value => value !== 'free');
+            }
+
+            logging.info(`Updating portal_plans to ${settingData}`);
+            await knex('settings').update('value', JSON.stringify(settingData)).where('key', 'portal_plans');
+        } catch (err) {
```
daniellockyer, line 59: And here 😛

core/server/data/migrations/versions/4.38/2022-03-02-09-10-update-free-products-visible-column.js
```diff
@@ -0,0 +1,64 @@
...
+                    logging.info('portal_plans setting already contains "free" - skipping update');
+                    return;
+                } else {
+                    settingData = existingSettingData.concat('free');
+                }
+            } else {
+                settingData = existingSettingData.filter(value => value !== 'free');
+            }
+
+            logging.info(`Updating portal_plans to ${settingData}`);
```
daniellockyer, line 57: Can we extract the `JSON.stringify(settingData)` out into a variable and use it in this log line? Without it, I get the following when rolling back:

```
[2022-03-02 20:58:13] INFO Updating portal_plans to
```

whereas I expect

```
[2022-03-02 20:58:13] INFO Updating portal_plans to []
```

allouis: Done

daniellockyer (approved):

allouis:
This cannot be merged until after the UI updates have been done

matthanley (changes requested):
This needs updating for the `visibility` column change (sorry!)

daniellockyer:
+ change the timestamps on the migrations 😁

matthanley (approved):

daniellockyer (approved):

rshbhgrg (approved):

PR #14256, opened 2022-03-02
🐛 Added the member context to the get, next-post and prev-post helpers 

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1367

- The {{access}} property of a post always returned false for non-public content
- The get, next-post and prev-post helpers didn't add the member context to the internal API calls
- Added the members context to these calls

Tests:
- Added tests that check if the member context is passed to the API calls
- Includes a refactor for both the next-post and prev-post helper tests to the async await syntax
- Added `const should = require('should');` to the next-post and prev-post helper tests to allow them to run on their own

### DIFF
--- a/core/frontend/helpers/get.js
+++ b/core/frontend/helpers/get.js
@@ -156,6 +156,7 @@ module.exports = function get(resource, options) {
 
     // Parse the options we're going to pass to the API
     apiOptions = parseOptions(ghostGlobals, this, apiOptions);
+    apiOptions.context = {member: data.member};
 
     // @TODO: https://github.com/TryGhost/Ghost/issues/10548
     return controller[action](apiOptions).then(function success(result) {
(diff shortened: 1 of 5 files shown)


core/frontend/helpers/get.js
```diff
@@ -153,6 +153,7 @@ module.exports = function get(resource, options) {
 
     // Parse the options we're going to pass to the API
     apiOptions = parseOptions(ghostGlobals, this, apiOptions);
+    apiOptions.context = {member: data.member};
```
matthanley, line 156: Is it correct/safe to always override context here, or should we be merging with an existing `context`?

matthanley (approved):

matthanley:
When this is squashed + merged, it would be great to update the commit title to communicate the problem that's fixed (access is incorrect when using get helper) rather than the fix itself so it makes sense to users who don't have context

PR #14264, opened 2022-03-03
Switched products.visible for products.visibility

matthanley:
- We have an existing pattern for using `visibility: public` instead of `visible: true|false`
- We no-op the existing migration and roll forward so that we don't have to manually revert db changes

### DIFF
--- a/core/server/data/migrations/versions/4.38/2022-03-01-08-46-add-visibility-to-tiers.js
+++ b/core/server/data/migrations/versions/4.38/2022-03-01-08-46-add-visibility-to-tiers.js
@@ -5,3 +5,7 @@ module.exports = createAddColumnMigration('products', 'visible', {
     nullable: false,
     defaultTo: false
 });
+
+module.exports.up = () => {
+    // noop - column will be replaced with `visibility` instead
+};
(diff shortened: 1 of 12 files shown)


core/server/data/migrations/versions/4.38/2022-03-01-08-46-add-visibility-to-tiers.js
```diff
@@ -5,3 +5,7 @@ module.exports = createAddColumnMigration('products', 'visible', {
     nullable: false,
     defaultTo: false
 });
+
+module.exports.up = async () => {
```
allouis, line 9: Why is this file not being deleted?

allouis (commented):
Changes look good - I don't understand why we're keeping the unreleased migrations though - isn't it simpler to completely remove them?

Is this to stop issues with local development on `main`?

allouis (approved):
Either way this looks good though!

daniellockyer (approved):

daniellockyer:
> Is this to stop issues with local development on main?

@allouis Yeah, that, and we decided to always move forwards with migrations once they've landed in main. We have some internal sites that are continuously deployed without resetting their DB, and deleting the migration will brick their DB because knex-migrator won't like the extra row

allouis:
Cool cool cool cool cool 👌

PR #14276, opened 2022-03-07
Fixed site setup correctly naming the default Tier

allouis:
no-issue

Since the Free tier was added, the first Tier was not the default one. This updates the site setup to check for the slug explicitly

### DIFF
--- a/core/server/services/auth/setup.js
+++ b/core/server/services/auth/setup.js
@@ -110,14 +110,15 @@ async function doProduct(data, productsAPI) {
         return user;
     }
     try {
-        const page = await productsAPI.browse({limit: 1});
+        const page = await productsAPI.browse({limit: 'all'});
 
-        const [product] = page.products;
+        const product = page.products.find(p => p.slug === 'default-product');
+
         if (!product) {
             return data;
         }
 
-        productsAPI.edit({products: [{name: blogTitle.trim()}]}, {context: context.context, id: product.id});
+        await productsAPI.edit({products: [{name: blogTitle.trim()}]}, {context: context.context, id: product.id});
     } catch (e) {
         return data;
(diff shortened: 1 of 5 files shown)


core/server/services/auth/setup.js
```diff
@@ -110,14 +110,15 @@ async function doProduct(data, productsAPI) {
         return user;
     }
     try {
-        const page = await productsAPI.browse({limit: 1});
+        const page = await productsAPI.browse({limit: 'all'});
```
allouis, line 113: Need to fetch all, because we cannot assume first one is free - generally there will be 2

core/server/services/auth/setup.js
```diff
@@ -110,14 +110,15 @@ async function doProduct(data, productsAPI) {
...
+
+        const product = page.products.find(product => product.slug === 'default-product');
 
-        const [product] = page.products;
         if (!product) {
             return data;
         }
 
-        productsAPI.edit({products: [{name: blogTitle.trim()}]}, {context: context.context, id: product.id});
+        await productsAPI.edit({products: [{name: blogTitle.trim()}]}, {context: context.context, id: product.id});
```
allouis, line 121: We await it so that we can catch errors

test/utils/e2e-framework-mock-manager.js
```diff
@@ -13,6 +13,14 @@ let emailCount = 0;
 const mailService = require('../../core/server/services/mail/index');
 const labs = require('../../core/shared/labs');
 
+const disableStripe = async () => {
```
naz, line 16: nice :ok_hand::ok_hand::ok_hand:

matthanley (approved):

PR #14282, opened 2022-03-08
🐛 Fixed post access in the get helper

allouis:
no-issue

Because we are passing through a different member object as the context
in the get helper, the content gating was not workign correctly, as the
member was missing a status property, this adds the property which fixes
content gating.


test/e2e-frontend/helpers/get.test.js
```diff
@@ -0,0 +1,290 @@
+const should = require('should');
+const sinon = require('sinon');
+const {startGhost} = require('../../utils/e2e-framework');
```
matthanley, line 3: Is there a reason `startGhost` was added to e2e-framework instead of using `testUtils.startGhost`?

SimonBackx: No, I've corrected it. Thanks!

matthanley (approved):

PR #14284, opened 2022-03-08
Fixed error when editing user with empty roles data

daniellockyer:
- we send the roles data array in when we're changing the role of the
  user
- if we send an empty array, we don't want to edit the user's role
- the code _thought_ that's what it was doing, but we only check the
  falsiness of the array, which is truthy for `[]`
- it also needs to check the length of the array
- this commit includes a test which would fail with a 500 error without
  the fix

### DIFF
--- a/core/server/models/user.js
+++ b/core/server/models/user.js
@@ -510,7 +510,7 @@ User = ghostBookshelf.Model.extend({
             return ghostBookshelf.Model.edit.call(self, data, options).then((user) => {
                 let roleId;
 
-                if (!data.roles) {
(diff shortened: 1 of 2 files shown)


daniellockyer:
@matthanley I found this when investigating the user roles bug the other day. The fix is very straightforward and I've written a regression test that fails without the change. I know we're trying to cut down on regression tests, but I'm unsure where else to put it.

Should this rather be a unit test (although it looks like I'd need to mock a bunch of things), or an E2E test? 🙂

PR #14298, opened 2022-03-10
Added the product name to member subscriptions

tpatel:
refs https://github.com/TryGhost/Team/issues/1141 https://github.com/TryGhost/Members/commit/da6a43bfcf9878e2711d6b28cb02f0b4186b50be

- Adds the `member.subscriptions.price.product.name` property to the member read endpoint

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "@tryghost/logging": "2.0.4",
     "@tryghost/magic-link": "1.0.20",
     "@tryghost/member-events": "0.4.0",
-    "@tryghost/members-api": "5.1.1",
+    "@tryghost/members-api": "5.2.0",
     "@tryghost/members-events-service": "0.3.1",
     "@tryghost/members-importer": "0.5.3",
     "@tryghost/members-offers": "0.10.8",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2169,10 +2169,10 @@
     "@tryghost/domain-events" "^0.1.8"
     "@tryghost/member-events" "^0.4.0"
 
-"@tryghost/members-api@5.1.1":
-  version "5.1.1"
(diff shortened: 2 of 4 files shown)


yarn.lock
```diff
@@ -2256,7 +2256,7 @@
     jsonwebtoken "^8.5.1"
     lodash "^4.17.11"
 
-"@tryghost/members-stripe-service@0.8.4", "@tryghost/members-stripe-service@^0.8.4":
+"@tryghost/members-stripe-service@0.8.4":
```
matthanley, line 2259: I think `@tryghost/members-stripe-service` needs bumping in Ghost as well, otherwise afaict we have two different versions of it 🤔 
cc @allouis is this expected?

package.json
```diff
@@ -83,7 +83,7 @@
     "@tryghost/logging": "2.0.4",
     "@tryghost/magic-link": "1.0.20",
     "@tryghost/member-events": "0.4.0",
-    "@tryghost/members-api": "5.1.1",
+    "@tryghost/members-api": "5.2.0",
```
matthanley, line 86: @allouis this also releases your `updateStripeProductNamesFromDefaultProduct` migration - is that cool?

matthanley (commented):
@tpatel I don't see any tests for the addition of the product to the subscription 👀 Either a unit test on the member service or an e2e test for the behaviour you're trying to achieve is fine.

package.json
allouis: Give me a bit of time and I'll merge my bump first if that's cool

yarn.lock
allouis: This is fine not to be bumped - yeah it's expected, I'll fix it though, it's because we still have it as a dependency of members-api

matthanley (approved):

PR #14300, opened 2022-03-10
Added the MemberCancelEvent model

tpatel:

- This event stores the members' subscription cancelations.


allouis (approved):

core/server/data/migrations/versions/4.39/2022-03-07-14-37-add-members-cancel-events-table.js
```diff
@@ -0,0 +1,8 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('members_cancel_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+    from_plan: {type: 'string', maxlength: 255, nullable: true},
```
daniellockyer, line 6: What is the rationale around `from_plan` being nullable? Is that something that can happen? 🤔

tpatel: Good point, it can't happen so I've moved `from_plan` so that it can't be nullable, thanks!

daniellockyer (approved):

daniellockyer:
@tpatel We've just released 4.39 so these migrations will need to be bumped. I'm going to block review, and I'm off next week, but you should be able to override it with another engineer's approval 🙂

daniellockyer (changes requested):
@tpatel We've just released 4.39 so these migrations will need to be bumped. I'm going to block review, and I'm off next week, but you should be able to override it with another engineer's approval 🙂

daniellockyer (approved):

PR #14313, opened 2022-03-11
Added the `MemberCancelEvent` model to MembersApi

tpatel:

- Triggered when a subscription is canceled by an admin or by the member


SimonBackx:
Some dependencies were updated in the members api package, but not in the Ghost main repo. I think it would make sense to also update these dependencies in the Ghost repo itself, as to avoid duplicate dependencies with different patch fixes. I don't think RenovateBot does this automatically.

SimonBackx (approved):

PR #14325, opened 2022-03-15
Enabled Admin integration for tiers and offers API

rshbhgrg:
As multiple tiers is now GA, we want to allow devs to be able to work with Tiers and offers via content/Admin API. This change -

- updates fixtures to add permission with roles for tiers and offers for new sites
- adds migration to update existing sites to have correct roles and permissions
- whitelists add/edit/read/browse on tiers and offers API for integrations


kevinansfield (approved):

matthanley (approved):

kevinansfield (approved):

matthanley:
Shouldn't this target 4.40 instead of 4.41?

PR #14329, opened 2022-03-16
Removed archived tiers from content api

rshbhgrg:
closes https://github.com/TryGhost/Team/issues/1426

When fetching tiers using the content API, we incorrectly returned all tiers including archived ones unless the active:true filter is passed. Correct behaviour is to always hide archived tiers, so this filter should not be required.

- forces `active:true` filter for tiers content api browse
- updates test to check for archived test removal in tiers content api

### DIFF
--- a/core/server/api/canary/utils/serializers/input/tiers.js
+++ b/core/server/api/canary/utils/serializers/input/tiers.js
@@ -1,3 +1,13 @@
+const localUtils = require('../../index');
+
+const forceActiveFilter = (frame) => {
+    if (frame.options.filter) {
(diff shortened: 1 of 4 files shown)


rishabhgrg:
@matthanley The codecov failure is a bit of red herring as the update to tiers content api test rightly covers the scenario where an archived tier is ignored in the final outcome. We have updated that test to add another tier which is archived to DB, and then we check that we still have actually the same outcome as before, and archived tier is not returned.

matthanley (approved):

PR #14335, opened 2022-03-18
Added e2e tests for offers API

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1429

Includes tests for most endpoints. Still needs one for filtering and some special cases.


test/e2e-api/admin/offers.test.js
```diff
@@ -0,0 +1,527 @@
...
+
+    it('Has no initial offers', async function () {
+        await agent
+            .get(`offers/`)
+            .expectStatus(200)
+            .matchHeaderSnapshot({
+                etag: anyEtag
+            })
+            .matchBodySnapshot({
+                offers: []
```
ErisDS, line 35: You only need to supply an object to matchBodySnapshot if you need to pass matchers for values that change like IDs and Dates.

SimonBackx:
Depends on one of these:
- https://github.com/TryGhost/Ghost/pull/14340 
- https://github.com/TryGhost/Members/pull/379
for the tests to work

PR #14349, opened 2022-03-21
Moved authentication tests to e2e-api

ErisDS:


ErisDS:
@naz please could you take over this PR? I was going to raise it as an issue but figured a PR made it easier to see the problems.

The first issue with moving these tests is the issue with the storage path being fixed in the wrong place.

I _believe_ the fix [here](https://github.com/TryGhost/Ghost/pull/14349/commits/93110ab7812b5379ee4e0c7bdfa48df8dc0591af) helps, but the content path changes across tests & the theme tests cannot use the new framework as it doesn't support file uploads yet, so there are possibly still issue to work out.

Note: do not attempt to get the framework to do fileuploads. That is too big a piece of work for right now. We'll tackle it later.

The current failure is [with this line](https://github.com/TryGhost/Ghost/blob/93110ab7812b5379ee4e0c7bdfa48df8dc0591af/test/e2e-api/admin/authentication.test.js#L81) but afaict this test should never have passed, because valid.zip is not the dawn theme, it's the casper theme. I'm very confused!

This may be related to https://github.com/TryGhost/Toolbox/issues/150 and also https://github.com/TryGhost/Toolbox/issues/252 and need a reasonable amount of rejigging.

naz:
Note on why the theme name works out to be "Dawn" instead of suspected "Capser" in the authentication test. The name for a theme pulled from GitHub is [derived from it's repository name](https://github.com/TryGhost/Ghost/blob/8b0676d9ff098edf1d20211d23cbe188794b27cf/core/server/services/themes/installer.js#L29). Because we only mock the response with a "valid.zip" the name is still taken from the Dawn repository and not the zip file itself.

naz:
Another note, locally the tests pass. Suspecting there's something tricky going on on the CI :thinking: Will need to dive deeper another time as this is not a blocking PR.

naz:
Had another stab at this today and could not figure out what causes the difference to load "dawn" on the local machine comparing to locaing "casper" on the CI. Would be good to find out if it also passes locally for others or is it just me with some non-standard configuration.

ErisDS:
@daniellockyer Asked @naz to put this down for now. As it's a CI-only issue, maybe you could take a quick look & see if you have any ideas to contribute for what could cause this? It's pretty low priority though.

daniellockyer:
I somehow get the test failure locally 😛

naz:
Had another look at this to figure out what exactly is causing the storage path to be reset to be a "local" one. 

What triggers the reset is the mockManager's restore, called in the afterEach block:
```js
        afterEach(function () {
        mockManager.restore();
```

this in turn culls the `cconfigUtils.restore()` which restores the "paths" configuration to default, local ones. when starting the Ghost instance for testing we set up paths to be a `tmp` directory:
```
    const contentFolder = path.join(os.tmpdir(), uuid.v4(), 'ghost-test');
    await prepareContentFolder({contentFolder});

    // NOTE: need to pass this config to the server instance
    configUtils.set('paths:contentPath', contentFolder);
```

The problem boils down to managing the `paths` configuration variable between test cases. 

Directions to explore next time we get to this:
1. Passing in `contentPath` information to the `startGhost` method from the test suite itself and ignore it's reset/restore in `mockManager.restore()` call.
2. Maybe we can have an even deeper change where we implement a special test-environment storage adapter that would be easier to initialize/reset.

PR #14363, opened 2022-03-22
Fixed 404 error handling in Ghost Admin

sam-lord:
no issue

Prevents errors from being uploaded to Sentry when a 404 happens in Ghost Admin. At the moment, 404s in Ghost Admin create an ENOENT error in express' static library. Our generic 404 handler at the end will only intercept requests that don't have any errors in the context, so a simple middleware can strip out 404 errors just before we add in our own.

The Ghost-specific error that we attach to requests does not get uploaded to Sentry :) 

---

Edit (additional context for changes other than the 404 fix):

Issue is that we were sending various errors to sentry that shouldn't have been, including some validation errors and 404s under /ghost/assets/.

The changes:

* Separated out prepareError and prepareStack
  * Prepare error still need to run before sending errors to sentry, since errors can be stored in an array (as is the case with validation errors), or can still be converted into a Ghost error at that stage
  * Prepare stack returns a new cloned error to avoid mutating the stack for the error which has already been sent to Sentry
* Prevent 404s in Ghost Admin from being sent to Sentry
  * Strips out the default error generated in Express so that Ghost can inject its own NotFoundError, which isn't sent to Sentry by default

### DIFF
--- a/core/frontend/web/middleware/error-handler.js
+++ b/core/frontend/web/middleware/error-handler.js
@@ -7,7 +7,7 @@ const config = require('../../../shared/config');
 const helpers = require('../../services/routing/helpers');
 
 // @TODO: make this properly shared code
-const {prepareError} = require('@tryghost/mw-error-handler');
+const {prepareError, prepareStack} = require('@tryghost/mw-error-handler');
 
 const messages = {
     oopsErrorTemplateHasError: 'Oops, seems there is an error in the error template.',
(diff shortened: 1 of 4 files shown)


sam-lord:
Merged with a bad commit title.

Issue is that we were sending various errors to sentry that shouldn't have been, including some validation errors and 404s under `/ghost/assets/`. 

The changes:
* Separated out `prepareError` and `prepareStack`
  * Prepare error still need to run before sending errors to sentry, since errors can be stored in an array (as is the case with validation errors), or can still be converted into a Ghost error at that stage
  * Prepare stack returns a new cloned error to avoid mutating the stack for the error which has already been sent to Sentry
* Prevent 404s in Ghost Admin from being sent to Sentry
  * Strips out the default error generated in Express so that Ghost can inject its own `NotFoundError`, which isn't sent to Sentry by default

PR #14364, opened 2022-03-22
Added the newsletters table

tpatel:
refs https://github.com/TryGhost/Team/issues/1433

- The `default` property stores whether a newsletter is set as default by the admin (the one `default` newsletter is the one selected by default when admins publish a new article)
- The `status` property stores whether a newsletter is archived or not
- The `recipient_filter` property is only storing whether a newsletter is "paid-only" or not for now, although it can be expanded to more specific filters in the future
- The `subscribe_on_signup` property stores whether a new member should be automatically signed up to the newsletter
- The `sort_order` property enables displaying the newsletter list in an order chosen by the admins

### DIFF
--- a/core/server/data/migrations/versions/4.42/2022-03-21-17-17-add.js
+++ b/core/server/data/migrations/versions/4.42/2022-03-21-17-17-add.js
@@ -1,0 +1,20 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('newsletters', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    name: {type: 'string', maxlength: 191, nullable: false},
+    description: {type: 'string', maxlength: 2000, nullable: true},
+    sender_name: {type: 'string', maxlength: 191, nullable: false},
+    sender_email: {type: 'string', maxlength: 191, nullable: false, validations: {isEmail: true}},
+    sender_reply_to: {type: 'string', maxlength: 191, nullable: false, validations: {isEmail: true}},
+    default: {type: 'bool', nullable: false, defaultTo: false},
+    status: {type: 'string', maxlength: 50, nullable: false, defaultTo: 'active'},
+    recipient_filter: {
+        type: 'text',
+        maxlength: 1000000000,
+        nullable: false,
+        defaultTo: ''
+    },
+    subscribe_on_signup: {type: 'bool', nullable: false, defaultTo: false},
+    sort_order: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0}
+});
--- a/core/server/data/exporter/table-lists.js
+++ b/core/server/data/exporter/table-lists.js
@@ -15,6 +15,7 @@ const BACKUP_TABLES = [
     'members_stripe_customers_subscriptions',
     'migrations',
     'migrations_lock',
+    'newsletters',
     'oauth',
     'permissions',
     'permissions_roles',
--- a/core/server/data/schema/schema.js
+++ b/core/server/data/schema/schema.js
@@ -711,5 +711,23 @@ module.exports = {
             }
         },
         value: {type: 'text', maxlength: 65535, nullable: true}
+    },
+    newsletters: {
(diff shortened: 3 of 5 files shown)


core/shared/config/env/config.development.json
```diff
@@ -22,5 +25,23 @@
...
+        "options": {
+            "service": "Mailgun",
+            "host": "localhost",
+            "port": 1025,
+            "secure": false,
+            "auth": {
+                "user": "username",
+                "pass": "password"
+            }
+        }
```
rishabhgrg, line 45: @tpatel Guess this change was accidental and meant for `config.local.json` ?

rshbhgrg (changes requested):

core/shared/config/env/config.development.json
tpatel: huh good catch 🙈

tpatel: @rishabhgrg It's fixed ✅

core/server/data/migrations/versions/4.42/2022-03-21-17-17-add.js
```diff
@@ -0,0 +1,20 @@
...
+    sender_name: {type: 'string', maxlength: 191, nullable: false},
+    sender_email: {type: 'string', maxlength: 191, nullable: false, validations: {isEmail: true}},
+    sender_reply_to: {type: 'string', maxlength: 191, nullable: false, validations: {isEmail: true}},
+    default: {type: 'bool', nullable: false, defaultTo: false},
+    status: {type: 'string', maxlength: 50, nullable: false, defaultTo: 'active'},
+    recipient_filter: {
+        type: 'text',
+        maxlength: 1000000000,
+        nullable: false,
+        defaultTo: 'status:-free'
```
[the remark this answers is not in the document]
tpatel: This field would be a filter to prevent non-matching users from being able to subscribe to the newsletter. I believe posts on a newsletter would inherit/copy this field indeed.

You're right we might want to leave the default open to all users, I'm gonna make it empty

rshbhgrg (approved):
@tpatel Looks good, the only thing i am a bit unsure about is the default value for `recipient_filter` long term, as the value used in other places in schema right now is either `status:-free` or `none`. Not 100% which is the right value here, so probably something we can come back to upgrade as we setup rest of the flow, wdyt ?

tpatel:
@rishabhgrg I agree, yesterday after your comment I've dived into the code to understand more about these two defaults. My thinking is that the default isn't really relevant, we're gonna need to always set it to something.

I did use `status:-free` at first as I was reusing the post declaration, although it doesn't feel the right value for newsletters.

I was thinking about using `none` but then it's the same as having an empty filter: you need to have code to handle this special case.

I agree with going forward and correcting this later if needed, especially as we'll have more migrations on the newsletter model.

daniellockyer (approved):

core/server/data/migrations/versions/4.42/2022-03-21-17-17-add.js
```diff
@@ -0,0 +1,20 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('newsletters', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    name: {type: 'string', maxlength: 191, nullable: false},
+    description: {type: 'string', maxlength: 2000, nullable: true},
+    sender_name: {type: 'string', maxlength: 191, nullable: false},
+    sender_email: {type: 'string', maxlength: 191, nullable: false, validations: {isEmail: true}},
+    sender_reply_to: {type: 'string', maxlength: 191, nullable: false, validations: {isEmail: true}},
+    default: {type: 'bool', nullable: false, defaultTo: false},
```
matthanley, line 10: Do we have an existing pattern of setting a default bool on a single row vs adding a setting?

matthanley: This field is to be removed ref https://github.com/TryGhost/Team/issues/1474

PR #14367, opened 2022-03-23
refactor: replace deprecated String.prototype.substr()

CommanderRoot:

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.


ErisDS:
Thanks for this PR 😄 It's merged and will go out in the next 4.46 release.

PR #14389, opened 2022-03-30
Added the newsletters endpoint

tpatel:
refs https://github.com/TryGhost/Team/issues/1463

- This enables listing, creating and editing newsletters

### DIFF
--- a/core/server/models/newsletter.js
+++ b/core/server/models/newsletter.js
@@ -1,0 +1,9 @@
+const ghostBookshelf = require('./base');
+
+const Newsletter = ghostBookshelf.Model.extend({
+    tableName: 'newsletters'
+});
+
+module.exports = {
+    Newsletter: ghostBookshelf.model('Newsletter', Newsletter)
+};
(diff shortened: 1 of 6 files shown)


core/server/models/newsletter.js
```diff
@@ -0,0 +1,14 @@
+const ghostBookshelf = require('./base');
+
+const Newsletter = ghostBookshelf.Model.extend({
+    tableName: 'newsletters'
+});
+
+const Newsletters = ghostBookshelf.Collection.extend({
```
allouis, line 7: I think you can remove this, I'm pretty sure this is a pattern we want to move away from

refs https://ghost.slack.com/archives/C02G9E68C/p1631721852044500

allouis (commented):
I guess we need to update the `permissions` property after the other PR is merged - but otherwise looks good!

One comment about some code you can delete

core/server/models/newsletter.js
tpatel: awesome, thanks!

PR #14391, opened 2022-03-30
Created new stats service and stats API to get member count history

SimonBackx:
refs TryGhost/Team#1458
refs TryGhost/Team#1459
refs TryGhost/Team#1372

- Added a new stats service, which is divided into several categories. Currently only the 'members' category for member related stats.
- When there are missing or corrupt members status events in the DB, the totals returned by the old member stats endpoint (`/members/stats/count`) were wrong. This is fixed in the new service by counting in reverse order and starting with the actual totals.
- New Stats API, with the new `/stats/members/count-history` endpoint.
- This new endpoint also returns the paid deltas -> dashboard 5.0 will show subscribed and canceled paid members for each day
- Includes tests for the new stats service and endpoint


core/server/api/canary/stats.js
```diff
@@ -0,0 +1,18 @@
...
+    memberCountHistory: {
+        permissions: {
+            docName: 'members',
+            method: 'browse'
+        },
+        async query() {
+            const data = await statsService.members.getCountHistory();
+
+            return {
+                data
```
allouis, line 14: Should we have a `meta` property here? I think we're treating each entry as a resource right? Maybe just something like
```js
"meta": {
    "pagination": {
        "page": 1,
        "limit": "all",
        "pages": 1,
        "total": data.length,
        "next": null,
        "prev": null
    }
}
```

I guess this would be better inside the service 🤔 

Not sure if we need this though

allouis (approved):
Looks good, one comment about the `meta` property - but I don't think it's blocking

rshbhgrg (approved):

core/server/api/canary/stats.js
SimonBackx: Okay, I've added it to the service 👍

allouis (approved):

PR #14392, opened 2022-03-30
Backfill MRR deltas based on offers

allouis:
refs https://github.com/TryGhost/Team/issues/1452

**UPDATE**

This migration has been simplified to only migrate MRR events for Members which satisfy

    1. They have an active Subscription with an Offer.
    2. This is the only Subscription they have ever had.

---

<details>
  <summary>Previous description</summary>

Okay, so this is gnarly one 😬 

Places with unsound logic have been marked with ⚠️

#### MRR Event

A row in the `members_paid_subscription_events` table

It can be associated with a Member, and has an mrr_delta value

It has a `from_plan` and `to_plan` value which should form a chain of events, e.g. `{f:0, t:1} > {f:1, t:4} > {f:4, t:5}`

It is created whenever a subscription is created, deleted, or has the price changed

#### Offer Redemption 

A row in the `offer_redemptions` table

It can be associated with a Member, Subscription & Offer

It is created whenever an Offer is used to create a Subscription

#### Offer

A row in the `offers` table

It has one of three durations: 'forever', 'once', 'repeating'
It has one of two types: 'percentage' or 'fixed'
If it is fixed it can have one of two intervals: 'yearly'  or 'monthly'
It has a value, which is in the same magnitude/format as the mrr_delta value in an MRR Event

### What we know

- MRR Events are _always_ created before a corresponding Offer Redemption
- MRR Events will have a `from_plan` of `null` if it is the start of a Subscription
- An Offer can only be used to start a subscription
- An Offer will be removed from a Subscription if the price is changed

### Setup

We first fetch all necessary information upfront to avoid multiple reads from the database.

We fetch all Offer Redemptions which are associated with an Offer with the 'forever' duration, we include with them, the status and price of the associated Subscription, as well as the type, interval and amount of the associated Offer.

We fetch all MRR Events which are associated with Members whom have an Offer Redemption at some point in time.

### What we need to do

We know that for every existing Offer Redemption we need to adjust MRR for at least one event, possibly two. The MRR Event associated with the creation of the Subscription will always need the mrr_delta adjusted negatively, to account for the Offer discount.

If there is a second MRR Event associated with the lifecycle of the Subscription with an Offer - it will be one of two cases:
1. The Subscription has been cancelled
2. The Subscription has had its price changed.
 
Both of these cases will require an mrr_delta adjustment by the same amount the first MRR Event was adjusted, but in the positive direction

### How we do it

Loop through all Offer Redemptions and determine which MRR Event is associated with the Subscription - because both Offer Redemptions and MRR Events are associated with a Member we can limit the search to only MRR Events which share a Member.

There are two cases:

1. A single "initial" MRR Event is present.
2. Multiple "initial" MRR Events are present.

#### Case 1. A single "initial" MRR Event is present.

This our "base case" and the easiest one to handle, we adjust the mrr_delta of the found MRR Event, and check for a second MRR Event - because we know only a single Subscription has been created there is no doubt of whether or not the second MRR Event is associated with the subscription.

If a second MRR Event is found, we also adjust the mrr_delta of that.

#### Case 2. Multiple "initial" MRR Events are present.

This is where things get tricky.

⚠️ The current approach is to guess which one is associated with the Offer Redemption based on the different in timestamps between the Offer Redemption and the MRR Events, choosing the ones closest together, but in the correct chronological order.  

From here we adjust the mrr_delta of the MRR Event and begin our search for a possible second event.

⚠️ If the Subscription is currently on the same price as the initial MRR Event, we make an assumption that a user has not changed price, lost their offer, and then changed back. We stop searching for a second event immediately

Otherwise we take the MRR Events, and filter them to a list of candidates for a second event, they must be chronologically ordered, and the `to_plan` and `from_plan` values must match.

⚠️ We then take the first matching event and treat it as the second, updating the mrr_delta

</details>

#### Optimisations

We've used a JOIN in the initial fetch which reduces 1.reads from the DB and 2.loops to match up the separate responses

Rather than updating individual MRR Events, which would require a lot of writes, we delete all the events which need to be modified, and then insert new ones (retaining the initial timestamps)


core/server/data/migrations/versions/4.42/2022-03-28-15-25-backfill-mrr-adjustments-for-offers.js
```diff
@@ -0,0 +1,170 @@
...
+            const firstEvent = getFirstEvent(firstEvents, redemption);
+            const mrrAdjustment = getMRRAdjustment(firstEvent, redemption);
+
+            updateEvent(firstEvent, -mrrAdjustment);
+
+            const mustHaveSecondEvent = redemption.subscription_status === 'canceled' || firstEvent.to_plan !== redemption.subscription_price;
+
+            const likelyDoesNotHaveSecondEvent = firstEvent.to_plan === redemption.subscription_price;
+
+            if (likelyDoesNotHaveSecondEvent) {
```
SimonBackx, line 118: Not sure, but might be worth to only skip if `mustHaveSecondEvent` is also false? Maybe that never happens, but I'm not sure.
```suggestion
            if (likelyDoesNotHaveSecondEvent && !mustHaveSecondEvent) {
```

ErisDS:
Isn’t this going to compound with missing offer data?

allouis:
@ErisDS I thought it would compound with the missing offer data, yeah! But now you say it, I suppose we can fix it when we fix the missing offer data 🤔 

The current "plan" however is to hold off on merging this until the missing offer data is fixed, which I've communicated with Peter in Slack

core/server/data/migrations/versions/5.0/2022-03-28-15-25-backfill-mrr-adjustments-for-offers.js
```diff
@@ -0,0 +1,103 @@
...
+                });
+            }
+
+            return calculateMRR({
+                interval: event.interval,
+                amount: event.amount - redemption.discount_amount
+            });
+        }
+
+        offerRedemptions.forEach((redemption) => {
```
matthanley, line 68: I thought this would need an eslint-ignore line 🤔

core/server/data/migrations/versions/5.0/2022-03-28-15-25-backfill-mrr-adjustments-for-offers.js
```diff
@@ -0,0 +1,103 @@
...
+
+            if (memberEvents.length !== 1) {
+                return;
+            }
+
+            const firstEvent = memberEvents[0];
+
+            const mrr = calculateMRR(firstEvent, redemption);
+
+            firstEvent.used = true;
```
matthanley, line 79: This isn't used anywhere

core/server/data/migrations/versions/5.0/2022-03-28-15-25-backfill-mrr-adjustments-for-offers.js
allouis: Doesn't seem to! maybe `forEach` isn't in the linting

matthanley (approved):

core/server/data/migrations/versions/5.0/2022-03-28-15-25-backfill-mrr-adjustments-for-offers.js
```diff
@@ -0,0 +1,101 @@
...
+        });
+
+        if (updatedEvents.length === 0) {
+            return;
+        }
+
+        const idsToDelete = updatedEvents.map(event => event.id);
+
+        await knex('members_paid_subscription_events').whereIn('id', idsToDelete).del();
+        await knex('members_paid_subscription_events').insert(updatedEvents);
```
daniellockyer, line 98: Does it make more sense for this to be `.batchInsert`?

core/server/data/migrations/versions/5.0/2022-03-28-15-25-backfill-mrr-adjustments-for-offers.js
```diff
@@ -0,0 +1,101 @@
...
+
+            if (memberEvents.length !== 1) {
+                return;
+            }
+
+            const firstEvent = memberEvents[0];
+
+            const mrr = calculateMRR(firstEvent, redemption);
+
+            updatedEvents.push({
```
daniellockyer, line 79: I know `type` is `nullable` but we seem to lose it here. Is this used at present?

allouis: Good catch this migration was written before those columns were added!

core/server/data/migrations/versions/5.0/2022-03-28-15-25-backfill-mrr-adjustments-for-offers.js
allouis: Yup!

allouis:

PR #14393, opened 2022-03-31
Added the newsletter API permissions

tpatel:
refs https://github.com/TryGhost/Team/issues/1463

- Allow admins to perform all newsletter operations
- We can adjust and be more permissive in the future if needed

### DIFF
--- a/test/utils/fixtures/fixtures.json
+++ b/test/utils/fixtures/fixtures.json
@@ -532,6 +532,21 @@
                     "name": "Edit custom theme settings",
                     "action_type": "edit",
                     "object_type": "custom_theme_setting"
+                },
+                {
+                    "name": "Browse newsletters",
+                    "action_type": "browse",
+                    "object_type": "newsletter"
+                },
+                {
+                    "name": "Add newsletters",
+                    "action_type": "add",
+                    "object_type": "newsletter"
+                },
+                {
+                    "name": "Edit newsletters",
+                    "action_type": "edit",
+                    "object_type": "newsletter"
                 }
             ]
         },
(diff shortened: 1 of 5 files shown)


core/server/api/canary/newsletters.js
```diff
@@ -11,15 +11,15 @@ module.exports = {
             'order',
             'page'
         ],
-        permissions: () => true, //FIXME: move to database permissions
```
allouis, line 14: IMO This change should be part of the other PR, I don't like that we've got commits which explicitly disable permissions on our endpoints. Not huge deal but 🤷 I would just merge this followed by the API implementation

allouis (approved):
Love migrations like these - super easy!

test/utils/fixtures/fixtures.json
```diff
@@ -532,6 +532,20 @@
                     "name": "Edit custom theme settings",
                     "action_type": "edit",
                     "object_type": "custom_theme_setting"
+                },
+                {
+                    "name": "Browse newsletters",
+                    "action_type": "browse",
+                    "object_type": "newsletter"
+                },{
```
daniellockyer, line 540: An unbelievably tiny nit but can we put the braces on separate lines? 😛

core/server/api/canary/newsletters.js
tpatel: I'm gonna merge the controller with `permissions: true` and with the tests commented out, otherwise I would have to mess with the commits too much. I agree it's better to not leave this endpoint open at any time 👍

test/utils/fixtures/fixtures.json
tpatel: good catch 😉

daniellockyer (approved):

PR #14408, opened 2022-04-01
Added table to store newsletters for a member

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1469

With multiple newsletters, members can have access to one or more newsletters on the site. Right now, the subscription to default newsletter for a member is controlled via a single boolean `subscribed` column on the member table.
We want to extend the relation between a newsletter and a member so that multiple newsletters can be attached to a member for different flows via new pivot table.

- adds new `members_newsletters` pivot table
- updates tests

### DIFF
--- a/core/server/data/schema/schema.js
+++ b/core/server/data/schema/schema.js
@@ -733,7 +733,6 @@ module.exports = {
     members_newsletters: {
         id: {type: 'string', maxlength: 24, nullable: false, primary: true},
         member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
-        newsletter_id: {type: 'string', maxlength: 24, nullable: false, references: 'newsletters.id', cascadeDelete: true},
-        sort_order: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0}
+        newsletter_id: {type: 'string', maxlength: 24, nullable: false, references: 'newsletters.id', cascadeDelete: true}
     }
 };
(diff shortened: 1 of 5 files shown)


core/server/data/schema/schema.js
```diff
@@ -729,5 +729,11 @@ module.exports = {
         },
         subscribe_on_signup: {type: 'bool', nullable: false, defaultTo: false},
         sort_order: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0}
+    },
+    members_newsletters: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+        newsletter_id: {type: 'string', maxlength: 24, nullable: false, references: 'newsletters.id', cascadeDelete: true},
+        sort_order: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0}
```
ErisDS, line 737: Why would we need sort order on this? 🤔

rishabhgrg: @ErisDS Yea we don't have a usecase for it rn, had only kept this in to have the same convention as other member/product pivot tables, but guess that's not needed ? Will get rid of it in that case.

rishabhgrg: Updated

kevinansfield (approved):

daniellockyer (approved):

PR #14411, opened 2022-04-01
Added the post-newsletter relation

tpatel:
refs https://github.com/TryGhost/Team/issues/1471

- This is a many-to-one relation so that many posts can be linked to a specific newsletter
- The `newsletters` table had to come first in the schema file so that it's initialized before the `posts` table (because of the foreign key)
- Updated the model to make sure the new field doesn't leak in the API for now
- This migration isn't using the existing utils because of a performance issue. In MySQL, adding a new row without `algorithm=copy` uses the INPLACE algorithm which was too slow on big `posts` tables (~3 minutes for 10k posts). Switching to the COPY algorithm fixed the issue (~3 seconds for 10k posts).

[Slack conversation](https://ghost.slack.com/archives/C02G9E68C/p1649151709623559) for the `algorithm=copy`

### DIFF
--- a/core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
+++ b/core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
@@ -1,0 +1,8 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('posts', 'newsletter_id', {
+    type: 'string',
+    maxlength: 24,
+    nullable: true,
+    references: 'newsletters.id'
+});
--- a/core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
+++ b/core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
@@ -55,7 +55,7 @@ module.exports = {
 
         await knex.raw('alter table `posts` drop foreign key `posts_newsletter_id_foreign`;');
 
-        await knex.raw('alter table `posts` drop `newsletter_id`;');
+        await knex.raw('alter table `posts` drop `newsletter_id`, algorithm=copy;');
     }
 };
 
--- a/core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
+++ b/core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
@@ -1,6 +1,7 @@
 const logging = require('@tryghost/logging');
 const DatabaseInfo = require('@tryghost/database-info');
 const commands = require('../../../schema/commands');
+const {createTransactionalMigration} = require('../../utils');
 
 const table = 'posts';
 const column = 'newsletter_id';
@@ -23,13 +24,8 @@ const columnDefinition = {
  * was too slow on big `posts` tables (~3 minutes for 10k posts). Switching to
  * the COPY algorithm fixed the issue (~3 seconds for 10k posts).
  */
-module.exports = {
-    config: {
-        transaction: true
-    },
-    async up(config) {
-        const knex = config.transacting;
-
+module.exports = createTransactionalMigration(
+    async function up(knex) {
         const hasColumn = await knex.schema.hasColumn(table, column);
 
         if (hasColumn) {
@@ -52,7 +48,8 @@ module.exports = {
         }).toSQL()[0].sql;
 
         if (DatabaseInfo.isMySQL(knex)) {
-            sql += ', algorithm=copy';
+            // Guard against an ending semicolon
+            sql = sql.replace(/;\s*$/, '') + ', algorithm=copy';
         }
 
         await knex.raw(sql);
@@ -68,9 +65,7 @@ module.exports = {
             transaction: knex
         });
     },
-    async down(config) {
-        const knex = config.transacting;
-
+    async function down(knex) {
         const hasColumn = await knex.schema.hasColumn(table, column);
 
         if (!hasColumn) {
@@ -103,10 +98,11 @@ module.exports = {
         }).toSQL()[0].sql;
 
         if (DatabaseInfo.isMySQL(knex)) {
-            sql += ', algorithm=copy';
+            // Guard against an ending semicolon
+            sql = sql.replace(/;\s*$/, '') + ', algorithm=copy';
         }
 
         await knex.raw(sql);
     }
-};
+);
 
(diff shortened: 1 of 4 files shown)


core/server/data/schema/schema.js
```diff
@@ -712,24 +731,6 @@ module.exports = {
         },
         value: {type: 'text', maxlength: 65535, nullable: true}
     },
-    newsletters: {
```
allouis, line 715: I guess this has to move so that the table exists when the post table references it for the foreign key?

core/server/models/post.js
```diff
@@ -878,6 +878,9 @@ Post = ghostBookshelf.Model.extend({
         // CASE: never expose the revisions
         delete attrs.mobiledoc_revisions;
 
+        // CASE: hide the newsletter_id for now
+        delete attrs.newsletter_id;
```
allouis, line 882: Arguably this could be done in the serialiser - but I think here works too 👍

allouis (approved):

allouis:
@daniellockyer Adding a FK to the posts table will cause the creation of a new index - is that right? If so, what's the protocol for testing that against a large dataset?

core/server/data/schema/schema.js
tpatel: Exactly, I did update the commit to mention this but forgot to update the PR!

core/server/models/post.js
tpatel: I wanted to avoid any side-effect in code in general for this PR, I'll look into the serializer next, thanks!

core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
```diff
@@ -48,14 +62,19 @@ module.exports = {
 
         logging.info(`Removing ${table}.${column} column`);
 
-        if (DatabaseInfo.isSQLite(knex)) {
-            await commands.dropColumn(table, column, knex, columnDefinition);
-            return;
-        }
+        await knex.schema.alterTable(table, function (t) {
```
allouis, line 64: Can we use the `dropForeign`[^1] command instead? That handles checking for the existence of the constraint and also handles SQLite cases 

[^1]: https://github.com/TryGhost/Ghost/blob/main/core/server/data/schema/commands.js#L220

core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
```diff
@@ -27,14 +36,19 @@ module.exports = {
...
+        if (DatabaseInfo.isMySQL(knex)) {
+            sql += ', algorithm=copy';
         }
 
-        await knex.raw('alter table `posts` add column `newsletter_id` varchar(24) null, algorithm=copy;');
+        await knex.raw(sql);
 
-        await knex.raw('alter table `posts` add constraint `posts_newsletter_id_foreign` foreign key (`newsletter_id`) references `newsletters` (`id`);');
+        await knex.schema.alterTable(table, function (t) {
+            t.foreign(column).references('newsletters.id');
```
allouis, line 47: Same as below except use the `addForeign`[^1] command

[^1]: https://github.com/TryGhost/Ghost/blob/main/core/server/data/schema/commands.js#L168

core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
```diff
@@ -27,14 +36,19 @@ module.exports = {
...
 
-        if (DatabaseInfo.isSQLite(knex)) {
-            await commands.addColumn(table, column, knex, columnDefinition);
-            return;
+        let sql = knex.schema.table(table, function (t) {
+            t.string(column, 24);
+        }).toSQL()[0].sql;
+
+        if (DatabaseInfo.isMySQL(knex)) {
+            sql += ', algorithm=copy';
```
allouis, line 39: Do we have any guarantee that `knex` will never return the SQL string with an `;` on the end? I think this is fine, but I can't see anything in the docs about it, so would be mildly worried about this changing in a minor. 

cc @daniellockyer If you think this is fine, then so do I 👍

allouis (changes requested):
Some comments about the use of commands for the FK constraints. and a smol question mostly for @daniellockyer about the consistency of the `toSQL` return values

core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
```diff
@@ -29,6 +39,12 @@ module.exports = {
 
         logging.info(`Adding ${table}.${column} column`);
 
+        // Use the default flow for SQLite because .toSQL() is tricky with SQLite
```
allouis, line 42: Out of interest, what happened with `toSQL()` with SQLite?

allouis (approved):

core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
matthanley: I agree it would be good to guard against, _just in case_

matthanley (approved):
Tiny change to guard against `;` but otherwise lgtm

core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * This migration is adding a new column `newsletter_id` to the table posts
+ * that is a foreign key to `newsletters.id`.
+ *
+ * It isn't using the existing utils because of a performance issue. In MySQL,
+ * adding a new row without `algorithm=copy` uses the INPLACE algorithm which
+ * was too slow on big `posts` tables (~3 minutes for 10k posts). Switching to
+ * the COPY algorithm fixed the issue (~3 seconds for 10k posts).
+ */
+module.exports = {
```
daniellockyer, line 26: Can we refactor to use the util for transactional migrations? Like https://github.com/TryGhost/Ghost/blob/caea330647b70c300428764de52a05c873ee5652/core/server/data/migrations/versions/4.3/10-populate-stripe-price-id-in-subscriptions.js#L4

tpatel: done ✅

core/server/data/migrations/versions/4.43/2022-04-01-10-13-add-post-newsletter-relation.js
tpatel: done ✅

tpatel:
All feedback has been addressed, I'm merging now!

PR #14427, opened 2022-04-05
Added MRR stats service and endpoint

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1470

Instead of counting the MRR by resolving all the deltas from the past until now, we should start with the current calculated MRR and resolve it until the first event. That would give a more accurate recent MRR (in exchange for a less accurate MRR for older data) and allows us to limit the amount of returned days in the future.

- Includes MRR stats service that can fetch the current MRR
- The service can return a history of the MRR for every day
- New admin API endpoint /stats/mrr that returns the MRR history
- Includes tests for these new service and endpoint

### DIFF
--- a/core/server/services/stats/lib/mrr-stats-service.js
+++ b/core/server/services/stats/lib/mrr-stats-service.js
@@ -1,0 +1,152 @@
+const {DateTime} = require('luxon');
+
+class MrrStatsService {
+    constructor({db}) {
+        this.db = db;
+    }
+
+    /**
+     * Get the current total MRR, grouped by currency (ascending order)
+     * @returns {Promise<MrrByCurrency[]>}
+     */
+    async getCurrentMrr() {
+        const knex = this.db.knex;
+        const rows = await knex('members_stripe_customers_subscriptions')
+            .select(knex.raw(`plan_currency as currency`))
+            .select(knex.raw(`SUM(
+                CASE WHEN plan_interval = 'year' THEN
+                    plan_amount / 12
+                ELSE 
+                    plan_amount
+                END
+            ) AS mrr`))
+            .whereIn('status', ['active', 'unpaid', 'past_due'])
+            .where('cancel_at_period_end', 0)
+            .groupBy('plan_currency')
+            .orderBy('currency');
+
+        if (rows.length === 0) {
+            // Add a USD placeholder to always have at least one currency
+            rows.push({
+                currency: 'usd',
+                mrr: 0
+            });
+        }
+
+        return rows;
+    }
+
+    /**
+     * Get the MRR deltas for all days (from old to new), grouped by currency (ascending alphabetically)
+     * @returns {Promise<MrrDelta[]>} The deltas sorted from new to old
+     */
+    async fetchAllDeltas() {
+        const knex = this.db.knex;
+        const rows = await knex('members_paid_subscription_events')
+            .select('currency')
+            .select(knex.raw('DATE(created_at) as date'))
+            .select(knex.raw(`SUM(mrr_delta) as delta`))
+            .groupByRaw('DATE(created_at), currency')
+            .orderByRaw('DATE(created_at), currency');
+        return rows;
+    }
+
+    /**
+     * Returns a list of the MRR history for each day and currency, including the current MRR per currency as meta data. 
+     * The respons is in ascending date order, and currencies for the same date are always in ascending order.
+     * @returns {Promise<MrrHistory>}
+     */
+    async getHistory() {
+        // Fetch current total amounts and start counting from there
+        const totals = await this.getCurrentMrr();
+
+        const rows = await this.fetchAllDeltas();
+
+        // Get today in UTC (default timezone for Luxon)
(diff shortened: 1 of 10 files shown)


SimonBackx:
Blocked by a pending version bump of `members-api` package (required for a test to succeed)

core/server/services/stats/lib/mrr-stats-service.js
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        return rows;
+    }
+
+    /**
+     * Get the MRR deltas for all days (from old to new), grouped by currency (ascending alphabetically)
+     * @returns {Promise<MrrDelta[]>} The deltas sorted from new to old
+     */
+    async fetchAllDeltas() {
```
matthanley, line 43: I thought part of the reason for switching to reverse-ordering was so that we didn't have to query ALL of the events to figure out MRR for e.g. last 30 days. Is that not the case?

core/server/services/stats/lib/mrr-stats-service.js
```diff
@@ -0,0 +1,154 @@
+const {DateTime} = require('luxon');
```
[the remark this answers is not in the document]
SimonBackx: Luxon was already in the package.json and the lock file, the timezone is set to UTC somewhere too. So I figured it would be better to already start using Luxon in new services too (which will probably also move to a different repository in the future). Should I change it to moment?

core/server/services/stats/lib/mrr-stats-service.js
SimonBackx: Yes, that was one of the advantages. The main reason was more to have a higher precision. I wanted to add the `limit` option to it earlier, but that sparked some discussion (whether or not to use a `filter` or limit, refs https://github.com/TryGhost/Team/issues/1458, which seemed to settle on the idea that a `limit` option makes the most sense with the addition of a filter in the future), so I hold of making any decisions on that. Currently it is in line with the members stats endpoint, so maybe we should create a separate PR that adds the `limit` option to both?

core/server/services/stats/lib/mrr-stats-service.js
```diff
@@ -0,0 +1,154 @@
...
+     * Get the current total MRR, grouped by currency (ascending order)
+     * @returns {Promise<MrrByCurrency[]>}
+     */
+    async getCurrentMrr() {
+        const knex = this.db.knex;
+        const rows = await knex('members_stripe_customers_subscriptions')
+            .select(knex.raw(`plan_currency as currency`))
+            .select(knex.raw(`SUM(
+                CASE WHEN plan_interval = 'year' THEN
+                    plan_amount / 12
```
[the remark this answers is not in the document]
SimonBackx: You are right. But maybe we should round the total instead of the individual calculations.

If you have 1,000 members that pay a subscription of $100 yearly:
* **Rounding inside the query**: You would get $8.33 * 1,000 = $8,330 MRR. But in fact 12 * 8,330 would give an ARR of $99,960, while it should be $100,000. 
* **Rounding at the end**: ROUND(1000*100/12), being $8,333.33 might be more correct. Giving a ARR of $99,999.96

core/server/services/stats/lib/mrr-stats-service.js
SimonBackx: Thinking further on this, limit might not be a good idea if we have multiple currencies.

core/server/services/stats/lib/mrr-stats-service.js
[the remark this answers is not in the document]
matthanley: Whichever makes most sense - the values are stored in cents so we'll only ever be at most $0.01/mo per user away from the correct value (0.05% in this example) so I don't think it matters 😄

core/server/services/stats/lib/mrr-stats-service.js
matthanley: We don't need it right now, so let's track it and leave it for later 👍🏼

core/server/services/stats/lib/mrr-stats-service.js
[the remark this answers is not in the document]
matthanley: Let's stick with moment for now and migrate everything in one go.

core/server/services/stats/lib/mrr-stats-service.js
SimonBackx: I added a new issue [here](https://github.com/TryGhost/Team/issues/1499)

SimonBackx:
@matthanley The rounding was a very interesting edge case. Apparently we `floor` MRR deltas, so it would be best to also floor individual results too. And since the `plan_amount` has an `int` type, SQLite already rounded intermediate results in the SUM (probably why I didn't notice it).  So I added some tests to check the rounding and added the rounding explicitly.

https://github.com/TryGhost/Members/blob/0eaae7cc83dba0484fe0b381f8fa60d5631ec9d3/packages/members-api/lib/repositories/member.js#L1164-L1166

matthanley (approved):

PR #14429, opened 2022-04-05
Removed pagination from members stats endpoint and added extra day to…

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1458

- Removed pagination from endpoint (refs slack https://ghost.slack.com/archives/C02G9E68C/p1649169172732619?thread_ts=1649157179.904669&cid=C02G9E68C)
- Fixes a special case for negative member counts (test included)
- Always adds an extra day to the output (should be zero for perfect events)
- Changed sort order of (private) fetchAllStatusDeltas method

### DIFF
--- a/core/server/services/stats/lib/members-stats-service.js
+++ b/core/server/services/stats/lib/members-stats-service.js
@@ -28,8 +28,8 @@ class MembersStatsService {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


matthanley (approved):
Not tested that it returns the correct data, but lgtm

PR #14431, opened 2022-04-06
Added default newsletter subscription for new members

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1469

Currently, all new members get auto subscribed to the default newsletters. This change adds same behavior with multiple newsletters by auto subscribing all available newsletters on site for new members(If flag is enabled).

- adds newsletters service for working with newsletter data
- bumps `@tryghost/members-api` package which handles default subscription
- adds new test fixture/data for newsletters

### DIFF
--- a/core/server/services/members/api.js
+++ b/core/server/services/members/api.js
@@ -13,6 +13,7 @@ const SingleUseTokenProvider = require('./SingleUseTokenProvider');
 const urlUtils = require('../../../shared/url-utils');
 const labsService = require('../../../shared/labs');
 const offersService = require('../offers');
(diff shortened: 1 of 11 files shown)


matthanley (commented):
Shouldn't the changes to the members API take into account the newsletter `subscribe_on_signup` flag?

rishabhgrg:
> Shouldn't the changes to the members API take into account the newsletter subscribe_on_signup flag?

@matthanley Based on the working doc, we want to attach all existing newsletters on a member signup for 🛹 version. This will be taken into account though down the line in cycle as we wire up other parts of newsletter management.

matthanley (approved):

PR #14433, opened 2022-04-06
Supported 'FOR UPDATE' lock for `findOne`

allouis:
refs https://github.com/TryGhost/Team/issues/1248

This is the underlying cause of the problems we've seen whilst handling
Stripe webhooks. A transaction ensures that the operations are atomic,
but not that they can run concurrently.

If you have some code which does this, concurrently:
1. Starts a transaction
2. Reads a value
3. Changes the values
4. Ends the transaction

Without applying the `FOR UPDATE` lock - you will have both sequences
read the same value at step 2.

With the `FOR UPDATE` lock - one of the sequences will hang at step 2,
waiting for the other transaction to end, at which point it will resume
and read the _changed_ value.

### DIFF
--- a/core/server/models/base/plugins/crud.js
+++ b/core/server/models/base/plugins/crud.js
@@ -183,6 +183,8 @@ module.exports = function (Bookshelf) {
                 model.hasTimestamps = false;
             }
 
+            options.lock = 'forUpdate';
+
             const object = await model.fetch(options);
             if (object) {
                 options.method = 'update';
--- a/core/server/models/base/plugins/crud.js
+++ b/core/server/models/base/plugins/crud.js
@@ -137,7 +137,7 @@ module.exports = function (Bookshelf) {
                 options.columns = _.intersection(options.columns, this.prototype.permittedAttributes());
             }
 
-            if (options.forUpdate) {
+            if (options.transacting && options.forUpdate) {
                 options.lock = 'forUpdate';
             }
 
@@ -183,7 +183,9 @@ module.exports = function (Bookshelf) {
                 model.hasTimestamps = false;
             }
 
-            options.lock = 'forUpdate';
+            if (options.transacting) {
+                options.lock = 'forUpdate';
+            }
 
             const object = await model.fetch(options);
             if (object) {
--- a/core/server/models/base/plugins/crud.js
+++ b/core/server/models/base/plugins/crud.js
@@ -137,6 +137,10 @@ module.exports = function (Bookshelf) {
                 options.columns = _.intersection(options.columns, this.prototype.permittedAttributes());
             }
 
+            if (options.forUpdate) {
+                options.lock = 'forUpdate';
+            }
+
             return model.fetch(options)
                 .catch((err) => {
                     // CASE: SQL syntax is incorrect
--- a/test/unit/server/models/base/crud.test.js
+++ b/test/unit/server/models/base/crud.test.js
@@ -105,13 +105,14 @@ describe('Models: crud', function () {
             });
         });
 
-        it('Sets the `lock` option to "forUpdate" when the `forUpdate` option is passed', function () {
+        it('Sets the `lock` option to "forUpdate" when the `forUpdate` and `transacting` options are passed', function () {
             const data = {
                 id: 670
             };
             const unfilteredOptions = {
                 donny: 'donson',
-                forUpdate: true
+                forUpdate: true,
+                transacting: {}
             };
             const model = models.Base.Model.forge({});
             const fetchedModel = models.Base.Model.forge({});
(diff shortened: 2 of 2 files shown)


allouis:
I think we may need to extend this to the other crud methods too, for example the Post model passes this option to the `edit` method[^1].

As far as I can tell we only support this for reading collections[^2] at the moment, which we apply using `knex`, rather than `bookshelf`

[^1]: https://github.com/TryGhost/Ghost/blob/abb3ee48e48dedd5c1bcc9f14429c32e64845a16/core/server/models/post.js#L1099-L1101
[^2]: https://github.com/TryGhost/Ghost/blob/abb3ee48e48dedd5c1bcc9f14429c32e64845a16/core/server/models/base/plugins/filtered-collection.js#L27-L29

allouis:
Note this also affects the work being done for: https://github.com/TryGhost/Team/issues/1453 as these events need to be inserted based on Stripe webhooks, which Stripe sends two for, resulting in concurrent DB operations.

SimonBackx:
I think this issue is also related: https://github.com/TryGhost/Team/issues/1372. Because when editing a member's status, it also does a read + write (inside the `edit` method), and compares the read with the changed value. If two concurrent calls read the same value, you end up with duplicate status events.

In `packages/members-api/lib/repositories/member.js` at lines 760 - 803 (members repo)
```js
updatedMember = await this._Member.edit({status: status, products: memberProducts}, {...options, id: data.id});
        
// (...)
        
if (updatedMember.attributes.status !== updatedMember._previousAttributes.status) {
    await this._MemberStatusEvent.add({
        member_id: data.id,
        from_status: updatedMember._previousAttributes.status,
        to_status: updatedMember.get('status'),
        ...eventData
    }, options);
}
```

allouis:
@SimonBackx Yeah good shout - so we need to update this for the `edit` method too - thanks!

SimonBackx:
@allouis I think that will be the safest thing to do. I might be wrong, but this might also prevent other issues. 

Say you have two code paths (A and B) that concurrently change different fields of a member (or other model). Currently the edit method does not use the `patch` option of bookshelf (ref [Bookshelf Save](https://bookshelfjs.org/api.html#Model-instance-save)).
1. A: Reads member with ID 123
2. B: Reads same member with ID 123
3. A: Wants to update the status of the member to `paid`, but does a full save (because we don't use the `patch` option), if I understand it correctly it will set all columns of this `member` again to the ones read at step 1.
```UPDATE members SET status = 'paid', email = 'email@gmail.com', name = '', ... WHERE id = '123'```
4. B: Wants to change a different column of a member. But ends up overriding the status change because it sets all columns to the ones read at step 2, which didn't include the status change yet.
```UPDATE members  SET status = 'free', email = 'changed-email@gmail.com', name = '', ... WHERE id = '123'```

**Comment by allouis** [comment]
@SimonBackx Yup that understanding looks correct to me, I agree. I can't think of a reason where we'd want to _not_ use `FOR UPDATE` when calling the `edit` method, but we can always add an option in the future to opt-out

**Review by SimonBackx — APPROVED** [review]

**Review by matthanley — APPROVED** [review]

File: core/server/models/base/plugins/crud.js [hunk 1]
```diff
@@ -179,6 +183,8 @@ module.exports = function (Bookshelf) {
                 model.hasTimestamps = false;
             }
 
+            options.lock = 'forUpdate';
```
naz, line 186: Based on the doc from Bookshelf:
> Type of row-level lock to use. Valid options are forShare and forUpdate. This only works in conjunction with the transacting option, and requires a database that supports it.

I think it's worth adding an explicit check for 'transacting' here. Otherwise we might give an impression that the operation is always protected by the lock.

We do have similar check in our collections. I'm assuming the thinking was similar:
```js
        onFetchingCollection: function onFetchingCollection(model, columns, options) {
            if (options.forUpdate && options.transacting) {
                options.query.forUpdate();
            }
        },
```

naz:
@allouis  I'm assuming because this is an "update" operation that will be effected the most we do not care that much about performance implications of adding the `forUpdate` as a default. As you have more context, does this sound as a reasonable assumption or something you've considered?

naz:
Another point, if we are introducing the locking behavior as a default would be awesome to have a follow up issue to check places that we could clean up and remove explicit use of `forUpdate` :thinking:

allouis:
> I'm assuming because this is an "update" operation that will be effected the most we do not care that much about performance implications of adding the forUpdate as a default. As you have more context, does this sound as a reasonable assumption or something you've considered?

Yeah that sounds reasonable to me, I also think that without using this lock by default in the `edit` method - all calls to `edit` are prone to the issue Simon outlined above.

> Another point, if we are introducing the locking behavior as a default would be awesome to have a follow up issue to check places that we could clean up and remove explicit use of forUpdate 

Yeah that's a good point - shall I assign it to you/core?

allouis:
@naz I'm adding a check for options.transaction, but IMO we should also consider forcing a transaction for the `edit` method - because we are doing a read and a write, we should force that to be atomic

PR #14436, opened 2022-04-07
Added slug column to newsletters table

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1494

We have a convention for resources used in filters to have a slug field, so we should add one for Newsletters. As always these should be unique at the db-schema level.

- updates schema for newsletters
- adds migration for new `slug` column
- updates tests and test fixtures


kevinansfield (approved):

kevinansfield (approved):

daniellockyer (approved):

rishabhgrg:
Closed in favor of https://github.com/TryGhost/Team/issues/1500.

PR #14437, opened 2022-04-07
Populated type column of existing events

allouis:

Currently we only have three event types: created, updated & expired.

A created event always has a `from_plan` of null

An updated event will always have a different `to_plan` and `from_plan`
as the subscription has changed tier/cadence.

An expired event _should_ have a `to_plan` of null, but due to a bug we
have events with the same `from_plan` and `to_plan`.


core/server/data/migrations/versions/4.43/2022-04-06-15-22-populate-type-column-for-paid-subscription-events.js
```diff
@@ -0,0 +1,21 @@
...
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Setting "type" to "updated" for events with different to_plan & from_plan');
+        await knex('members_paid_subscription_events').update('type', 'updated').whereNotNull('from_plan').whereNotNull('to_plan').whereRaw('to_plan != from_plan');
+
+        logging.info('Setting "type" to "expired" for events with null to_plan or the same to_plan & from_plan');
+        await knex('members_paid_subscription_events').update('type', 'expired').whereNull('to_plan').whereNotNull('from_plan');
+        await knex('members_paid_subscription_events').update('type', 'expired').whereRaw('from_plan = to_plan');
```
matthanley, line 12: If this is always the case, should we also fix these events by setting `to_plan=NULL`?

allouis: I was thinking about that yeah - but with the new `type` property we won't have to infer anything, and I'm concerned about deleting data - just incase there is something we haven't thought of which we later want to fix/create a type for 🤔

matthanley: Makes sense 👍🏼

matthanley (approved):

matthanley (changes requested):
Just needs filename updating to target 4.44

allouis:
Updated!

matthanley (approved):

allouis:
@daniellockyer This is ready for you to take a look at!

daniellockyer (approved):

PR #14438, opened 2022-04-07
Added migration to populate cancellation events

allouis:

This also includes the MRR delta changes to the events, which may need
to be split out into a separate migration.


core/server/data/migrations/versions/4.43/2022-04-07-11-18-add-cancelled-events.js
```diff
@@ -0,0 +1,48 @@
...
+
+        const eventsToInsert = cancelledSubscriptions.map((subscription) => {
+            const event = {
+                id: (new ObjectID()).toHexString(),
+                type: 'cancelled',
+                member_id: subscription.id,
+                from_plan: null,
+                to_plan: null,
+                currency: subscription.plan_currency,
+                source: 'migration',
```
allouis, line 33: Is this a good idea @matthanley? Should we use this as a constraint for the `down` migration? 🤔

core/server/data/migrations/versions/4.43/2022-04-08-11-54-add-cancelled-events.js
```diff
@@ -0,0 +1,50 @@
...
+
+        if (cancelledSubscriptions.length === 0) {
+            logging.info('No missing cancelled events - skipping migration');
+            return;
+        }
+
+        const eventsToInsert = cancelledSubscriptions.map((subscription) => {
+            const event = {
+                id: (new ObjectID()).toHexString(),
+                type: 'cancelled',
```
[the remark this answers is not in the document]
daniellockyer: vs above 😅 

```
.whereNot('members_stripe_customers_subscriptions.status', 'canceled');
```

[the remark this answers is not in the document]
matthanley: We have a precedent in the db from the Stripe subs, I think we should follow it for consistency. We also use US-English everywhere, so I think single-L makes sense.

core/server/data/migrations/versions/4.43/2022-04-07-11-18-add-cancelled-events.js
matthanley: Yeah I think this probably makes sense. We can always update it later if we need to.

matthanley (changes requested):

allouis:
@matthanley Updated spelling!

core/server/data/migrations/versions/4.43/2022-04-08-11-54-add-cancelled-events.js
```diff
@@ -0,0 +1,50 @@
...
+
+            return event;
+        });
+
+        logging.info(`Found ${eventsToInsert.length} missing cancellation events`);
+        await knex('members_paid_subscription_events').insert(eventsToInsert);
+    },
+    async function down(knex) {
+        logging.info('Deleting all members_paid_subscription_events with a "type" of "cancelled"');
+        await knex('members_paid_subscription_events').where('type', 'canceled').del();
```
matthanley, line 48: Needs an additional check on `source=migration`

core/server/data/migrations/versions/4.43/2022-04-08-11-54-add-cancelled-events.js
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        const eventsToInsert = cancelledSubscriptions.map((subscription) => {
+            const event = {
+                id: (new ObjectID()).toHexString(),
+                type: 'canceled',
+                member_id: subscription.member_id,
+                subscription_id: subscription.id,
+                from_plan: null,
+                to_plan: null,
```
matthanley, line 33: Rethinking this: will we be able to answer the question later on of "what plan were they on when they cancelled?" 🤔

matthanley (changes requested):

core/server/data/migrations/versions/4.43/2022-04-08-11-54-add-cancelled-events.js
allouis: 🤔 Hmm - no - I guess we should store it as `from_plan` and `to_plan` ?

matthanley: Yeah I think that makes sense

allouis: Updated!

allouis:
@daniellockyer This is ready for you now!

daniellockyer (approved):

matthanley (approved):

PR #14441, opened 2022-04-08
Cleaned up newsletter schema migrations

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1500

The newsletter table schema has bunch of changes to go through for new and existing columns, this consolidates the schema changes into a single re-create table migration that drops and adds the newsletter table with correct schema. The table re-create migration needs to run before any of the tables using newsletter as foreign key. The schema changes include -

- new columns for design related fields
- new slug column for filtering
- unique constraint to `name` column
- remove `default` column (noops the existing default column migration)
- `sender_reply_to` has a default of newsletter and a validation of ['newsletter', 'support']
- changed schema for `recipient_filter` column to use same behavior as post visibility
- renames `recipient_filter` to `visibility`
- updated default values for `subscribe_on_signup` and  `visibility`


kevinansfield (approved):

core/server/data/migrations/utils.js
```diff
@@ -65,6 +65,35 @@ function dropTables(names) {
...
+                logging.warn(`Failed to drop table: ${name} - table does not exist`);
+            } else {
+                logging.info(`Dropping table: ${name}`);
+                await commands.deleteTable(name, connection);
+                logging.info(`Re-adding table: ${name}`);
+                await commands.createTable(name, connection, tableSpec);
+            }
+        },
+        async function down() {
+            // noop: we cannot go back to old table schema
```
matthanley, line 91: Could you add a logging message here to say we're no-op'ing?

core/server/data/schema/schema.js
```diff
@@ -10,20 +10,31 @@
...
+            maxlength: 50,
             nullable: false,
-            defaultTo: ''
+            defaultTo: 'members'
         },
-        subscribe_on_signup: {type: 'bool', nullable: false, defaultTo: false},
-        sort_order: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0}
+        subscribe_on_signup: {type: 'bool', nullable: false, defaultTo: true},
+        sort_order: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0},
+
```
matthanley, line 28: There's an extra newline here

core/server/data/schema/schema.js
```diff
@@ -10,20 +10,31 @@
...
+            type: 'string',
+            maxlength: 50,
             nullable: false,
-            defaultTo: ''
+            defaultTo: 'members'
         },
-        subscribe_on_signup: {type: 'bool', nullable: false, defaultTo: false},
-        sort_order: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0}
+        subscribe_on_signup: {type: 'bool', nullable: false, defaultTo: true},
+        sort_order: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0},
```
matthanley, line 27: Should this be unique? If we default to zero we might accidentally change the default newsletter

core/server/data/migrations/utils.js
rishabhgrg: Done

matthanley, line 91: Could you add a logging message here to say we're no-op'ing?

core/server/data/schema/schema.js
rishabhgrg: ✅

core/server/data/migrations/utils.js
rishabhgrg: Done.

core/server/data/schema/schema.js
rishabhgrg: 🤔 We don't have any pattern for enforcing uniqueness on `sort_order`, i think it might make more guard against this in the model layer/business logic

tpatel (approved):

matthanley (approved):

daniellockyer (approved):

PR #14445, opened 2022-04-08
Added subscription_id to MRR events

allouis:
refs https://github.com/TryGhost/Team/issues/1453

As part of 5.0 we want to have cancelled evetns affect MRR. We are going
to backfill and begin populating cancelled events _before_ 5.0 however,
adding a migration to set the MRR deltas as part of 5.0

This migration will need to associate the cancelled events with a
subscription, so we are adding the subscription_id now. This will allow
us to guarantee that all cancelled events will have a subscription_id.

### DIFF
--- a/core/server/data/migrations/versions/4.43/2022-04-08-10-45-add-subscription-id-to-mrr-events.js
+++ b/core/server/data/migrations/versions/4.43/2022-04-08-10-45-add-subscription-id-to-mrr-events.js
@@ -3,7 +3,5 @@ const {createAddColumnMigration} = require('../../utils');
 module.exports = createAddColumnMigration('members_paid_subscription_events', 'subscription_id', {
     type: 'string',
     maxlength: 24,
-    nullable: true,
-    references: 'members_stripe_customers_subscriptions.id',
-    cascadeDelete: true
+    nullable: true
 });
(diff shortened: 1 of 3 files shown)


core/server/data/migrations/versions/4.43/2022-04-08-10-45-add-subscription-id-to-mrr-events.js
```diff
@@ -0,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('members_paid_subscription_events', 'subscription_id', {
+    type: 'string',
+    maxlength: 24,
+    nullable: true
+});
```
matthanley, line 7: Is this intentionally missing a FK to the subscription?

allouis: Yes - the FK was failing to add because the of the null values for the column.

The plan is to add a migration in the future to populate all missing values, at which point we can add a `NOT NULL` and `REFERENCES` constraint

matthanley (approved):

daniellockyer (approved):

PR #14447, opened 2022-04-08
Populated mrr_delta for cancelled subscriptions

allouis:
refs https://github.com/TryGhost/Team/issues/1455

We have to consider that a cancelled subscription may have been
cancelled and reactivated in the past, one option would be to only
update the most recent cancelled event, but this requires more
computation than updated all cancelled events, and then balancing out
any reactivated events with the inverse MRR.

This doesn't give us perfectly accurate historical data, but the current
data will be.

### DIFF
--- a/core/server/data/migrations/versions/5.0/2022-04-08-11-46-update-mrr-for-cancelled-events.js
+++ b/core/server/data/migrations/versions/5.0/2022-04-08-11-46-update-mrr-for-cancelled-events.js
@@ -9,7 +9,7 @@ module.exports = createTransactionalMigration(
                 'members_stripe_customers_subscriptions.id',
                 'members_stripe_customers_subscriptions.plan_currency',
                 'members_stripe_customers_subscriptions.plan_interval',
-                'members_stripe_customers_subscriptions.plan_amount',
+                'members_stripe_customers_subscriptions.plan_amount'
             )
             .from('members_stripe_customers_subscriptions')
             .join('members_stripe_customers', 'members_stripe_customers.customer_id', '=', 'members_stripe_customers_subscriptions.customer_id')
--- a/core/server/data/migrations/versions/5.0/2022-04-08-11-46-update-mrr-for-cancelled-events.js
+++ b/core/server/data/migrations/versions/5.0/2022-04-08-11-46-update-mrr-for-cancelled-events.js
@@ -40,7 +40,7 @@ module.exports = createTransactionalMigration(
 
             logging.info(`Updating cancelled events for ${subscription.id} to include mrr_delta`);
             await knex('members_paid_subscription_events').update({mrr_delta: mrrDelta}).where({
-                type: 'cancelled',
+                type: 'canceled',
                 subscription_id: subscription.id
             });
             logging.info(`Updating reactivated events for ${subscription.id} to include mrr_delta`);
@@ -51,6 +51,6 @@ module.exports = createTransactionalMigration(
         }
     },
     async function down(knex) {
-        await knex('members_paid_subscription_events').update({mrr_delta: 0}).whereIn('type', ['cancelled', 'reactivated']);
+        await knex('members_paid_subscription_events').update({mrr_delta: 0}).whereIn('type', ['canceled', 'reactivated']);
     }
 );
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/5.0/2022-04-08-11-46-update-mrr-for-cancelled-events.js
```diff
@@ -0,0 +1,56 @@
...
+        for (const subscription of cancelledSubscriptions) {
+            let mrrDelta;
+            if (subscription.plan_interval === 'year') {
+                mrrDelta = -1 * Math.floor(subscription.plan_amount / 12);
+            }
+            if (subscription.plan_interval === 'month') {
+                mrrDelta = -1 * subscription.plan_amount;
+            }
+            if (subscription.plan_interval === 'week') {
+                mrrDelta = -1 * subscription.plan_amount * 4;
```
matthanley, line 35: any reason to go with `* 4` rather than `* 52 / 12`?

core/server/data/migrations/versions/5.0/2022-04-08-11-46-update-mrr-for-cancelled-events.js
```diff
@@ -0,0 +1,56 @@
...
+            }
+            if (subscription.plan_interval === 'day') {
+                mrrDelta = -1 * subscription.plan_amount * 30;
```
matthanley, line 38: Similarly does `365/12` work better here?

core/server/data/migrations/versions/5.0/2022-04-08-11-46-update-mrr-for-cancelled-events.js
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            logging.info(`Updating cancelled events for ${subscription.id} to include mrr_delta`);
+            await knex('members_paid_subscription_events').update({mrr_delta: mrrDelta}).where({
+                type: 'canceled',
+                subscription_id: subscription.id
+            });
+            logging.info(`Updating reactivated events for ${subscription.id} to include mrr_delta`);
+            await knex('members_paid_subscription_events').update({mrr_delta: -1 * mrrDelta}).where({
+                type: 'reactivated',
```
matthanley, line 48: Do we have any reactivated events?

matthanley: nvm - I see this migration is gonna be deferred until 5.0.

core/server/data/migrations/versions/5.0/2022-04-08-11-46-update-mrr-for-cancelled-events.js
allouis: This is how we calculate the MRR for new events, so we're not deviating from the current practice https://github.com/TryGhost/Members/blob/main/packages/members-api/lib/repositories/member.js#L1186-L1192

core/server/data/migrations/versions/5.0/2022-04-08-11-46-update-mrr-for-cancelled-events.js
allouis: Not right now, but it's possible in 5.0 that we have some!

allouis:
@daniellockyer Are you good with this going into `main` as opposed to the 5.0 branch?

daniellockyer:
@allouis I've thought about it and a problem I can see is that it's going to continually flag up with the following upon boot:

```
[2022-04-11 11:32:40] WARN knex-migrator is skipping 5.0
[2022-04-11 11:32:40] WARN Current version in MigratorConfig.js is smaller then requested version, use --force to proceed!
[2022-04-11 11:32:40] WARN Please run `knex-migrator migrate --v 5.0 --force` to proceed!
```

I'm worried a self-hoster might run into some other problem and execute it, which isn't what we want. Would you mind putting it on 5.0?

allouis:
@daniellockyer Done!

allouis:
closed in favour of https://github.com/TryGhost/Ghost/pull/14558

PR #14448, opened 2022-04-08
Added mrr column to subscriptions

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1457

We want to save the MRR with a subscription to simplify the calculation of the total MRR once, in 5.0, we also take 'forever' offers into account into the MRR (so we can just SUM the MRR of all subscriptions).

- Sets the MRR to 0 for now.
- A migration (in a different PR) will fill in all the values in a data migration, but this needs to get merged first because we need this new column in order to update the members-api package (so we already save the MRR before doing the data migration).

### DIFF
--- a/core/server/data/migrations/versions/4.44/2022-04-11-10-54-add-mrr-to-subscriptions.js
+++ b/core/server/data/migrations/versions/4.44/2022-04-11-10-54-add-mrr-to-subscriptions.js
@@ -1,0 +1,8 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('members_stripe_customers_subscriptions', 'mrr', {
+    type: 'integer',
+    unsigned: true,
+    nullable: false,
+    defaultTo: 0
+});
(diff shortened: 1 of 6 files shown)


core/server/data/migrations/versions/4.43/2022-04-08-14-41-add-mrr-to-subscriptions.js
```diff
@@ -0,0 +1,8 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('members_stripe_customers_subscriptions', 'mrr', {
```
matthanley, line 3: Are we sure we want to go with MRR vs ARR?

SimonBackx: Discussed in https://ghost.slack.com/archives/C02G9E68C/p1649667605432439?thread_ts=1649408400.454149&cid=C02G9E68C

tl;dr: we'll use MRR.

SimonBackx:
Should I bump the migration version to v4.44 or is that not required?

daniellockyer:
@SimonBackx Yep, you'll need to bump it 🙂

daniellockyer:
@SimonBackx Do you have the associated `schema.js` change for this as well? 🙂

SimonBackx:
@daniellockyer I updated the schema.js file, and also fixed the tests 😊

daniellockyer (approved):

matthanley (approved):

allouis (approved):

PR #14449, opened 2022-04-08
Added newsletter id in post edit api

tpatel:
refs https://github.com/TryGhost/Team/issues/1502

- Support the `newsletter_id` only when sending a newsletter
- Default to the default newsletter when `newsletter_id` isn't specified


core/server/models/post.js
```diff
@@ -674,6 +674,13 @@ Post = ghostBookshelf.Model.extend({
             }
         }
 
+        // newsletter_id is read-only and should only be set using a query param when publishing/scheduling
```
matthanley, line 677: What is it that makes the newsletter_id read only? afaict you could still override it if you include it in the body

test/e2e-api/admin/posts.test.js
```diff
@@ -418,6 +418,27 @@ describe('Posts API', function () {
         res2.body.posts[0].status.should.eql('draft');
     });
 
+    it('Can\'t change the newsletter_id of a post', async function () {
```
matthanley, line 421: I don't think this test is doing what it says - it's returning a 422 because the newsletter_id 123 doesn't exist, not because updating the newsletter_id is blocked.

matthanley (changes requested):

core/server/models/post.js
tpatel: At the moment the `newsletter_id` is neither in the input nor in the output of the post API, so including it in the body will return an error.
Could you point me how we're doing it for https://github.com/TryGhost/Ghost/blob/d7a450559f843175ca78d8818eda8202d1c8521e/core/server/models/post.js#L684 ? I think it would be good that I implement a similar approach in case we choose to add `newsletter_id` to the post api.

test/e2e-api/admin/posts.test.js
tpatel: It does return a 422 because the property is blocked in the API as it isn't defined in the admin schema. I've changed to a valid id using the newsletters test fixtures.

core/server/models/post.js
matthanley: It's stripped in the API schema [here](https://github.com/TryGhost/SDK/blob/1f90c51a5b34e86fae5c1b6c3b56497f30462f63/packages/admin-api-schema/lib/canary/posts.json#L188-L190)

core/server/services/posts/posts-service.js
```diff
@@ -65,6 +76,14 @@ class PostsService {
             const sendEmail = model.wasChanged() && this.shouldSendEmail(model.get('status'), model.previous('status'));
 
             if (sendEmail) {
+                // Set the newsletter_id if it isn't passed to the API
+                if (!frame.options.newsletter_id) {
+                    const newsletters = await this.models.Newsletter.findAll({status: 'active'}, {transacting: frame.options.transacting});
```
matthanley, line 81: Super minor but I think we can add a `limit:1` to this so we don't load all of the newsletters

tpatel: I always have to look into the code to remember how to use things like `limit` with our models. Looking at https://github.com/TryGhost/Ghost/blob/a2b3568956477b995553e6662a71810e5a656952/core/server/models/base/plugins/sanitize.js#L45-L48 it seems that we don't support `limit` for the `findAll` method?

tpatel: Oh I need to switch to `findPage`, the naming does make sense 😅

matthanley (approved):

PR #14458, opened 2022-04-11
Fixed the newsletters api permissions

tpatel:
refs https://github.com/TryGhost/Team/issues/1504

- The permissions were missing in the fixture file
- This caused some Ghost installs to not have the right permissions
- This is fixed by adding the missing permissions to the fixture file and creating a migration to resolve the missing permissions


matthanley (changes requested):
The permissions have been added to the fixtures, but they've not been mapped to the Admin role

daniellockyer (approved):

matthanley (approved):

PR #14459, opened 2022-04-11
Handled default newsletter fixtures

rshbhgrg:
closes https://github.com/TryGhost/Team/issues/1491

- Added default newsletter fixture for new sites
- Handled default newsletter rename on setup
  - renames newsletter name and sender name to site title on setup

Note: This PR doesn't cover migrating default newsletter for existing sites to newsletter table, that is a separate change.

Fixture [here](https://github.com/TryGhost/Ghost/pull/14459/files#diff-396038cecd7a381616d00954ae18a655ae2a8af71ea65866bf09d2c7cc1b5235)


core/server/services/auth/setup.js
```diff
@@ -107,7 +107,9 @@ async function doSettings(data, settingsAPI) {
     return user;
 }
 
-async function doProduct(data, productsAPI) {
+// Update names for default product and newsletter to site title
+async function doProductAndNewsletter(data, api) {
```
matthanley, line 111: Curious: why combine these two concepts rather than handle them separately?

test/e2e-api/admin/members.test.js
```diff
@@ -1365,7 +1365,7 @@ describe('Members API: with multiple newsletters', function () {
     });
 
     // Edit a member
-    it('Can and and edit with custom newsletters', async function () {
+    it('Can add and edit with custom newsletters', async function () {
```
matthanley, line 1368: I believe this rename means we have redundant snapshots: https://github.com/TryGhost/Ghost/blob/c9a858d977be39ac71ebda2e489ab96ce772b596/test/e2e-api/admin/__snapshots__/members.test.js.snap#L2483

rishabhgrg: 🤔 Not sure why they didn't go away with `UPDATE_SNAPSHOTS=1`, i'll retry and see if it maybe just needs deletion of snapshot and recreate.

core/server/services/auth/setup.js
rishabhgrg: I was honestly split between doing them separate or together and tried both, and while not a strong pref but eventually kept together as it seemed to make more sense to keep related default fixture update together under a single fn. Probably could use a fn rename to `doFixtureDefaults` or something, but i am ok to split them up as well if that seems better.

test/e2e-api/admin/members.test.js
rishabhgrg: Yeah delete and update worked to get rid of dead tests, its now updated on PR :)

core/server/services/auth/setup.js
matthanley: > seemed to make more sense to keep related default fixture update together

I'm not sure they are related though apart from performing the same action on them 🤔

It's nabd, but I think as soon as we add another <X> that needs the same treatment we'll want to split them out anyway.

matthanley (approved):

daniellockyer (approved):

PR #14460, opened 2022-04-11
Added filtering for members api on newsletter subscription

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1484

While sending a post to a specific newsletter, we'll need to get list/count of members eligible to receive the post. This change enables members admin API to filter list of members on specific newsletter by their slug.

### DIFF
--- a/core/server/models/member.js
+++ b/core/server/models/member.js
@@ -30,6 +30,9 @@ const Member = ghostBookshelf.Model.extend({
         }, {
             key: 'products',
             replacement: 'products.slug'
+        }, {
+            key: 'newsletters',
+            replacement: 'newsletters.slug'
         }];
     },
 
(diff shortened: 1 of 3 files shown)


core/server/models/member.js
```diff
@@ -30,6 +30,9 @@ const Member = ghostBookshelf.Model.extend({
         }, {
             key: 'products',
             replacement: 'products.slug'
+        }, {
+            key: 'newsletters',
+            replacement: 'newsletters.slug'
```
matthanley, line 35: are we happy to use slug for this (and products) given that it'll always be `default-newsletter`?

matthanley, line 35: are we happy to use slug for this (and products) given that it'll always be `default-newsletter`?

rishabhgrg: 🤔 Not sure i get it, this change is basically to expand a filter like `newsletters:abc` to internally map to `newsletters.slug:abc` since the actual value in filter will be the slug property of newsletter.

matthanley: So my question is: are we happy that the filter will always look like `newsletters:default-newsletter` given that there's no way to change the value of `default-newsletter`?

PR #14468, opened 2022-04-12
Added the default newsletter

tpatel:
refs https://github.com/TryGhost/Team/issues/1473

- We use the title to populate the newsletter same, slug and sender name
- We use the description to populate the newsletter description
- We use the global design settings to populate the corresponding newsletter design settings

I've enabled using a for-loop in this migration because it's iterating on an array of 9 elements which seems fine.


tpatel:
I've been hitting a [weird error](https://github.com/TryGhost/Ghost/runs/5992196380?check_suite_focus=true) :

- The “Update from latest v3” action is failing with `A GhostError occurred.` and `A ProcessError occurred.`
- The “Print debug logs” action is failing with binary operator expected

I couldn't reproduce the issue locally when booting Ghost `3.42.9` and then migrating to `4.44` using `knex-migrator`.

daniellockyer:
@tpatel 

I think it's due to:

```
[10:35:53] → Ghost was able to start, but errored during boot with: `sqlite` does not support inserting default values. Specify values explicitly or use the `useNullAsDefault` config flag. (see docs http://knexjs.org/#Builder-insert).

```

tpatel:
@daniellockyer Thanks, I haven't found what exactly is trying to insert a default value, I'm gonna look into that again

core/server/data/migrations/versions/4.44/2022-04-11-12-13-add-default-newsletter.js
```diff
@@ -0,0 +1,84 @@
...
+         */
+        async function getSetting(name, defaultValue) {
+            const setting = await knex('settings')
+                .where('key', name)
+                .select(['value'])
+                .first();
+
+            if (setting) {
+                return setting.value;
+            } else if (Object.hasOwnProperty.call(defaultSettings, name)) {
```
matthanley, line 41: Minor point, but this implementation may be difficult to reason about because it can take a default value from two different sources.

core/server/data/migrations/versions/4.44/2022-04-11-12-13-add-default-newsletter.js
```diff
@@ -0,0 +1,84 @@
...
+            if (['show_badge', 'show_feature_image', 'show_header_icon', 'show_header_title'].includes(setting)) {
+                value = value === 'true';
+            }
+            newsletter[setting] = value;
+        }
+
+        logging.info('Adding the default newsletter');
+
+        logging.info(knex('newsletters').insert(newsletter).toSQL().sql);
+        logging.info(JSON.stringify(knex('newsletters').insert(newsletter).toSQL().bindings));
```
matthanley, line 76: I guess these additional log lines are left in from debugging?

tpatel: good catch 👍

core/server/data/migrations/versions/4.44/2022-04-11-12-13-add-default-newsletter.js
```diff
@@ -0,0 +1,84 @@
...
+            status: 'active',
+            visibility: 'members',
+            subscribe_on_signup: true,
+            sort_order: 0
+        };
+
+        const designSettings = ['header_image', 'show_header_icon', 'show_header_title', 'title_font_category', 'title_alignment', 'show_feature_image', 'body_font_category', 'footer_content', 'show_badge'];
+        // eslint-disable-next-line no-restricted-syntax
+        for (const setting of designSettings) {
+            let value = await getSetting('newsletter_' + setting);
```
rishabhgrg, line 66: @tpatel I am wondering if its probably better to load all the desired setting from DB with a single knex query rather than fetching each setting individually ? Something like - 
```
await knex('settings')
            .whereIn('key', allSettings)
            .select(['key', 'value'])
```

which will fetch all settings from DB with a single query and we can then just use them directly after that ?

core/server/data/migrations/versions/4.44/2022-04-11-12-13-add-default-newsletter.js
tpatel: I've rewritten everything, I think it's simpler now

core/server/data/migrations/versions/4.44/2022-04-11-12-13-add-default-newsletter.js
tpatel: I've rewritten so that it's only one sql query to get all settings.

core/server/data/migrations/versions/4.44/2022-04-11-12-13-add-default-newsletter.js
```diff
@@ -0,0 +1,87 @@
...
+            if (config.modifier) {
+                settingsMap[config.setting] = config.modifier(settingsMap[config.setting]);
+            }
+        }
+
+        const newsletter = {
+            id: ObjectId().toHexString(),
+            name: settingsMap.title,
+            description: settingsMap.description,
+            slug: slugify(settingsMap.title),
```
rishabhgrg, line 62: I think this should be `default-newsletter`, because that is the slug we are giving to default fixture as well and we don't allow slugs to be changed at the moment. Since there will be no newsletter in DB when this is run, its better to have a consistent slug for default newsletter across fixtures/migration, wdyt ?

rshbhgrg (changes requested):

core/server/data/migrations/versions/4.44/2022-04-11-12-13-add-default-newsletter.js
```diff
@@ -0,0 +1,87 @@
...
+            sender_name: settingsMap.title,
```
matthanley, line 63: I think this should probably be `null` after https://github.com/TryGhost/Ghost/pull/14479 to maintain current behaviour

tpatel: ok I'll wait for this other PR to get merged before moving forward with this one

rishabhgrg: @tpatel Merged the other PR now so this is good to be updated :)

tpatel: awesome, I've updated the slug and sender_name changes based on your PR 👍

core/server/data/migrations/versions/4.44/2022-04-11-12-13-add-default-newsletter.js
tpatel: fixed 👍

matthanley (commented):
tbh this migration feels difficult to follow/unnecessarily complex, as if it's had new bits added to it without rethinking the whole approach.

I've taken the liberty of reimplementing it [here](https://github.com/matthanley/Ghost/blob/8e1adf003d91c950ccc7e300aceccdfa94706ebb/core/server/data/migrations/versions/4.44/2022-04-13-13-00-add-default-newsletter.js) as an example of how I'd approach it. Let me know what you think!

tpatel:
I usually do prefer having the configuration separated from the code which is why it was structured that way. I also prefer using regexp as it avoids depending on lodash. I'm not attached to my implementation so I'm fine with changing it.

tpatel:
Just a note: this is only adding the default newsletter and not subscribing the existing members to this newsletter as this is part of [another task](https://github.com/TryGhost/Team/issues/1508) that is planned later in the feature development.

rshbhgrg (approved):

matthanley (approved):

matthanley:
We should hold off merging this and merge everything at once - otherwise if people change their design settings those changes will get reverted when we release newsletters

core/server/data/migrations/versions/4.45/2022-04-13-13-00-add-default-newsletter.js
```diff
@@ -0,0 +1,82 @@
...
+                newsletter[key] = value;
+            }
+        }
+
+        logging.info('Adding the default newsletter');
+        await knex('newsletters').insert(newsletter);
+    },
+    async function down() {
+        // Noop because we don't want to reset the default newsletter values
+        logging.info(`Skipping newsletter design settings backfill rollack - not needed`);
```
daniellockyer, line 79: `s/rollack/rollback/`

daniellockyer (approved):

SimonBackx:
We might want to hold off on merging this until I added the `created_at` and `updated_at` timestamps to the newsletters table

PR #14480, opened 2022-04-14
Added migration to adjust mrr in subscriptions for cancel_at_period_end

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1515

- Sets MRR to zero for all subscriptions where `cancel_at_period_end` is true
- Rollbacks to same logic as migration https://github.com/TryGhost/Ghost/pull/14465
- We need to be careful that future backfill adjustments of MRR happen after this one for the rollback to keep working


SimonBackx:
This is a 5.0 migration for the 5.0 branch

matthanley (approved):

allouis (approved):

daniellockyer (approved):

PR #14481, opened 2022-04-14
Updated calculation of MRR history to use mrr column

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1518

- Uses mrr column now
- Doesn't account for `cancel_at_period_end` until a migration fixes the MRR values: https://github.com/TryGhost/Ghost/pull/14480


SimonBackx:
The snapshot of one of the tests failed because one of the test fixtures created a subscription with a MRR set while `cancel_at_period_end` was true. Depending on the dashboardv5 flag, that could have been a non-zero MRR or not IRL... So tests are a bit hacky here.

matthanley (approved):

PR #14485, opened 2022-04-14
Fixed the newsletter_id assignement in the post API

tpatel:
refs https://github.com/TryGhost/Team/issues/1502

- Added two tests to make sure we assign a default `newsletter_id` when needed and to make sure `newsletter_id` can't be changed once an email is sent.
- Fixed the model so that we don't change `newsletter_id` if it's already defined when publishing, and so that we clear `newsletter_id` when a post is changed to draft without an email being sent.
- Fixed the post service so that a `newsletter_id` is always defined when editing a post. The alternative would have been to do an extra post fetch and write complex logic to only fetch the `newsletter_id` when we are about to send an email. The chosen solution has less code and uses an extra newsletter fetch instead of an extra post fetch, which is more performant given there could be ten of thousands of posts and it would be surprising to have more than thousands of newsletters (a more realistic figure would be less than a hundred newsletters).

### DIFF
--- a/core/server/models/post.js
+++ b/core/server/models/post.js
@@ -676,6 +676,7 @@ Post = ghostBookshelf.Model.extend({
 
         // newsletter_id is read-only and should only be set using a query param when publishing/scheduling
         if (options.newsletter_id
+            && !this.get('newsletter_id')
             && this.hasChanged('status')
             && (newStatus === 'published' || newStatus === 'scheduled')) {
             this.set('newsletter_id', options.newsletter_id);
@@ -695,6 +696,7 @@ Post = ghostBookshelf.Model.extend({
                 return self.related('email').fetch({transacting: options.transacting}).then((email) => {
                     if (!email) {
                         self.set('email_recipient_filter', 'none');
+                        self.set('newsletter_id', null);
                     }
                 });
             });
--- a/core/server/services/posts/posts-service.js
+++ b/core/server/services/posts/posts-service.js
@@ -28,6 +28,12 @@ class PostsService {
                     message: messages.invalidNewsletterId
                 });
             }
+        } else {
+            // Set the newsletter_id if it isn't passed to the API
+            const newsletters = await this.models.Newsletter.findPage({status: 'active', limit: 1, columns: ['id']}, {transacting: frame.options.transacting});
+            if (newsletters.data.length > 0) {
+                frame.options.newsletter_id = newsletters.data[0].id;
+            }
         }
 
         if (!frame.options.email_recipient_filter && frame.options.send_email_when_published) {
@@ -76,14 +82,6 @@ class PostsService {
             const sendEmail = model.wasChanged() && this.shouldSendEmail(model.get('status'), model.previous('status'));
 
             if (sendEmail) {
-                // Set the newsletter_id if it isn't passed to the API
-                if (!frame.options.newsletter_id) {
-                    const newsletters = await this.models.Newsletter.findPage({status: 'active', limit: 1, columns: ['id']}, {transacting: frame.options.transacting});
-                    if (newsletters.data.length > 0) {
-                        frame.options.newsletter_id = newsletters.data[0].id;
-                    }
-                }
-
                 let postEmail = model.relations.email;
 
                 if (!postEmail) {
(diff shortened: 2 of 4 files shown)


test/e2e-api/admin/posts.test.js
```diff
@@ -490,6 +494,131 @@ describe('Posts API', function () {
...
+            .send({posts: [updatedPost]})
+            .expect('Content-Type', /json/)
+            .expect('Cache-Control', testUtils.cacheRules.private)
+            .expect(200);
+
+        const model = await models.Post.findOne({
+            id
+        }, testUtils.context.internal);
+
+        should(model.get('newsletter_id')).not.eql(null);
```
matthanley, line 536: This is only testing that newsletter_id is set, not that it's set to the default newsletter which is what the test claims

core/server/models/post.js
```diff
@@ -676,6 +676,7 @@ Post = ghostBookshelf.Model.extend({
 
         // newsletter_id is read-only and should only be set using a query param when publishing/scheduling
         if (options.newsletter_id
+            && !this.get('newsletter_id')
```
kevinansfield, line 679: Thinking through whether this conditional is needed. If we keep it then it means the only way to change a newsletter for a scheduled post is to make the post a draft, then re-schedule. This makes sense within the existing publishing flow because we disable the recipient selection too but it might need re-visiting in the context of the publishing flow redesign.

No issue with it going in as-is but I'll ping design team to check.

core/server/services/posts/posts-service.js
```diff
@@ -28,6 +28,12 @@ class PostsService {
                     message: messages.invalidNewsletterId
                 });
             }
+        } else {
+            // Set the newsletter_id if it isn't passed to the API
+            const newsletters = await this.models.Newsletter.findPage({status: 'active', limit: 1, columns: ['id']}, {transacting: frame.options.transacting});
```
kevinansfield, line 33: If this is meant to be finding the default newsletter then I think we'll need an `order: 'sort_order DESC'` param so we get the first newsletter in the sorted list.

kevinansfield: Or are newsletters always sorted by `sort_order`?

kevinansfield: They are! Ignore this 🙂

kevinansfield (approved):
Approving on the basis that [@matthanley's comment](https://github.com/TryGhost/Ghost/pull/14485/files#r851217467) on the tests is resolved.

I've asked product/design about the desired behaviour for changing the newsletter for a scheduled post - if anything needs changing for that it can be done in a later iteration.

core/server/models/post.js
tpatel: Given [this](https://github.com/TryGhost/Ghost/blob/7a55e1a60ad6a7a1d4d451a92a5f853857cd496b/core/server/models/post.js#L687-L688) I believe we can't change the `email_recipient_filter` while a post stays on `scheduled`. So this shouldn't be an issue, right?

kevinansfield: That's the same assumption I'm looking for clarity on with regards to the updated publishing flow.

tpatel: I'm reformulating to make sure I understand:
1. Right now we need to unpublished a scheduled post in order to change the recipients, which is enforced both in the Admin and in the API.
2. There is the option to change this flow so that we can edit the recipients of a post while it's scheduled, this needs Peter's input and if we go with it we'll need to change the server & admin implementation

kevinansfield: 1. Correct.
2. Correct. We're re-working the publishing flow so all things like this are up for re-evaluation 🙂 Nothing needed in this PR so sorry for the noise, it was just seeing the conditional that triggered the thought 😅

matthanley (approved):

matthanley:
Note the tests are consistently failing on this - approval is subject to tests passing

tpatel:
I'm puzzled by the tests failing. The error is happening on the `"after all" hook for "Can upload routes.yaml"`, however there is no `after all`, only an `after`, and it only contains `return testUtils.stopGhost();`.
I can't reproduce the failing test locally.
Most likely I've introduced an issue but our test framework isn't making it easy to debug.

tpatel:
After more digging it seems like a CI-only test failure, I'm gonna leave this task on the side for Daniel to give it a look.

PR #14488, opened 2022-04-15
Stored offer_id in subscriptions

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1519

**Tests:**
- Tests whether the metadata from an offer is read correctly and stored in the database
- Test that invalid offer ids are ignored

**Changes in members repository:** 
- Compare changes here: https://github.com/TryGhost/Members/compare/%40tryghost/members-api%405.9.2...%40tryghost/members-api%405.11.1
- The `offer_id` column of subscriptions is set based on the coupon id from Stripe
- `getByStripeCouponId` method added in the offers repository (required to look up an offer from a stripe_coupon_id)
- the `members-payments` package was bumped twice (once for changes, once for undoing those changes, my bad). Nothing else has changed in that package.


SimonBackx:
I was going over all the code changes and did some testing with Stripe in test mode, and discovered a better way to link coupons from Stripe to offers in Ghost. We already store the stripe_coupon_id in the database, so it would make more sense to search for an offer in our database using the coupon id from stripe, instead of using the metadata of discounts in Stripe.

I made the wrong assumptions that subscriptions had the offer id stored in their metadata (apparently the offer id is only stored in the metadata of the checkout session, not the subscription). To fix that, we would either need a migration that updates the stripe discounts to save the offer id from ghost, or the simpler solution above (using the stripe_coupon_id). The latter one is obviously the best solution here.

I'm going to make the required changes and re request a review later.

SimonBackx:
Tests are failing in MySQL only (not in SQLite). Investigating right now.

matthanley (approved):

PR #14489, opened 2022-04-15
Filter member email recipients based on the newsletter subscriptions

tpatel:
refs https://github.com/TryGhost/Team/issues/1524

- We need to fetch the post newsletter to grab the slug as it's needed for the member NQL filter.
- We can then use the newsletter slug and append it in the existing member NQL filter.

This is the backend implementation only, there is still the need for an admin PR to update the members counts that are displayed in the Publish flow.

### DIFF
--- a/core/server/services/mega/mega.js
+++ b/core/server/services/mega/mega.js
@@ -160,7 +160,8 @@ const transformEmailRecipientFilter = (emailRecipientFilter, {errorProperty = 'e
     }
 
     if (newsletter) {
-        switch (newsletter.visibility) {
+        const visibility = newsletter.get('visibility');
+        switch (visibility) {
         case 'members':
             // No need to add a member status filter as the email is available to all members
             break;
@@ -170,7 +171,7 @@ const transformEmailRecipientFilter = (emailRecipientFilter, {errorProperty = 'e
         default:
             throw new errors.InternalServerError({
                 message: tpl(messages.newsletterVisibilityError, {
-                    value: newsletter.visibility
+                    value: visibility
                 })
             });
         }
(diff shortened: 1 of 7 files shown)


core/server/services/mega/mega.js
```diff
@@ -384,6 +385,11 @@ async function getEmailMemberRows({emailModel, memberSegment, options}) {
         filterOptions.filter = `${filterOptions.filter}+${memberSegment}`;
     }
 
+    if (labsService.isSet('multipleNewsletters') && emailModel.get('newsletter_id')) {
+        const newsletter = await emailModel.related('newsletter').fetch();
+        filterOptions.filter = `${filterOptions.filter}+newsletters:${newsletter.slug}`;
```
matthanley, line 390: While it's not used right now, I think we should also take into account the newsletter's `visibility` property here to satisfy [this requirement](https://www.notion.so/ghost/Architecture-Notes-65c8733988e946d98872bb9f4655b40f#e656ac2e3d974b2eaecc911812f0173a).

matthanley:
@tpatel you'll also need to update [the transform](https://github.com/tpatel/Ghost/blob/211497ce95c54b3c6cad65e51178f488cff3c10f/core/server/services/mega/mega.js#L130) to stop using the member's subscribed flag.

core/server/services/mega/mega.js
```diff
@@ -377,7 +413,15 @@ async function getEmailMemberRows({emailModel, memberSegment, options}) {
...
     const filterOptions = Object.assign({}, knexOptions);
 
-    const recipientFilter = transformEmailRecipientFilter(emailModel.get('recipient_filter'), {errorProperty: 'recipient_filter'});
+    let newsletter;
+    if (labsService.isSet('multipleNewsletters')) {
+        try {
+            newsletter = await emailModel.related('newsletter').fetch();
+        } catch (error) {
+            // Ignore NotFoundError
+        }
```
kevinansfield, line 422: Can you use `const newsletter = await emailModel.related('newsletter').fetch({require: false})` instead here? My understanding is that would [ignore a not-found error](https://bookshelfjs.org/api.html#Model-instance-fetch) automatically, in which case if there's some other error we won't be silently hiding it.

tpatel: neat, I didn't know about this property, thanks 👍

SimonBackx:
I'm working on some changes here with the fixtures from https://github.com/TryGhost/Ghost/pull/14563. Still some failing tests but now they should be easy to solve with these fixtures

SimonBackx:
There is still a bug somewhere when sending without the newsletter_id option

matthanley (approved):

PR #14491, opened 2022-04-15
Added newsletter from address verification

kevinansfield:
refs https://github.com/TryGhost/Team/issues/1498
refs https://github.com/TryGhost/Team/issues/584

With `membersFromAddress` being removed in favor of separate from addresses for each newsletter the verification flow needs to be moved too.

This also addresses problems with the existing verification flow where the emailed link was pointing directly at an API route resulting in poor experience when the process is unsuccessful. With this new flow there's no specific endpoint for triggering an email change, instead a verification will be triggered as needed when creating/updating a newsletter with `meta` data inserted in the response for clients to inform users as needed. The email also links to an Admin endpoint rather than an API endpoint providing a smoother confirmation flow.

- updated `newsletter` service so the index exports an instance
  - follows the more common pattern for services
  - keeps dependency injection in one place so it's not necessary to re-specify the injected dependencies everywhere the service is consumed
- added `add` and `edit` methods to `NewsletterService` class
  - accepts same arguments as `NewsletterModel.add/edit`
  - if the supplied `sender_email` attribute requires verification the attribute is stripped before saving, a verification email is triggered, and a `meta: {sent_email_verification: ['sender_email']}` property is added to the response
  - email verification contains a link to Admin, eg https://yoursite.com/ghost/#/settings/members-email/?verifyEmail=token with the token containing the id of the newsletter that's being updated, the new email value, and the changed property name
- updated `newsletters` API so it uses the `NewsletterService.add/edit` methods that contain the verification email trigger business logic
- added `PUT /newsletters/verifications/` endpoint
  - accepts `{token: 'xyz'}` in the body, this request is triggered by Admin when the confirmation link is loaded
  - extracts data from the single-use token, then uses the `id`, `property`, and `value` data to edit the newsletter completing the update of the email address

TODO:
- [x] add/edit tests
- [x] verification tests

### DIFF
--- a/core/server/api/canary/newsletters.js
+++ b/core/server/api/canary/newsletters.js
@@ -5,6 +5,7 @@ const errors = require('@tryghost/errors');
 const messages = {
     newsletterNotFound: 'Newsletter not found.'
 };
+const newslettersService = require('../../services/newsletters');
 
 module.exports = {
     docName: 'newsletters',
@@ -53,7 +54,7 @@ module.exports = {
         statusCode: 201,
         permissions: true,
         async query(frame) {
-            return models.Newsletter.add(frame.data.newsletters[0], frame.options);
+            return newslettersService.add(frame.data.newsletters[0], frame.options);
         }
     },
 
@@ -71,7 +72,19 @@ module.exports = {
         },
         permissions: true,
         async query(frame) {
-            return models.Newsletter.edit(frame.data.newsletters[0], frame.options);
+            return newslettersService.edit(frame.data.newsletters[0], frame.options);
+        }
+    },
+
+    verifyEmailUpdate: {
+        permissions: {
+            method: 'edit'
+        },
+        data: [
+            'token'
+        ],
+        async query(frame) {
+            return newslettersService.verifyAndUpdateEmail(frame.data.token);
         }
     }
 };
--- a/core/server/api/canary/newsletters.js
+++ b/core/server/api/canary/newsletters.js
@@ -76,7 +76,7 @@ module.exports = {
         }
     },
 
-    verifyEmailUpdate: {
+    verifyPropertyUpdate: {
         permissions: {
             method: 'edit'
         },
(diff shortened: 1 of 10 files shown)


ErisDS:
I have some comments on naming 😬  ref: [API design docs](https://www.notion.so/ghost/API-Core-c7f1926386304eedb3d8fd4304bfecf8)

- endpoint names should be snake_case not kebab
- endpoint names should be nouns wherever possible :) 


Suggestions: 

`PUT /newsletters/emails/`
`PUT /newsletters/verifications/`
`PUT /newsletters/from_address/`


Any of these or similar would be fine, depends on how you think about the resource that you're operating on. 

In terms of existing patterns in the API that relate to this

`router.put('/authentication/passwordreset'` is 👍 
`router.post('/authentication/invitation'`  is 👍 (although plural would prob be even better)
`router.post('/authentication/reset_all_passwords'` is 👎 

Also (I've spent way to long looking at the request and response bodies of all our endpoints recently haha) if we have no strong use-case for a response body from the endpoint, I am heavily in favour of setting `statusCode: 204` for any custom endpoint that does a thing and reports OK or errors. I can't quite tell from the PR if there's a real need for returning the edit response?

kevinansfield:
> I have some comments on naming 😬 ref: [API design docs](https://www.notion.so/ghost/API-Core-c7f1926386304eedb3d8fd4304bfecf8)

👍 I'll update to `PUT /newsletters/verifications/`. The token contains the property name and the new value so it's a fairly generic property/value verification process.

> I can't quite tell from the PR if there's a real need for returning the edit response?

The response is by the Admin as it lets it update it's internal store with the newly updated data. Clients can't read the data from the token, all they knows is that it needs to send it to the API to get the updated newsletter in response. An alternative would be returning something minimal with the `id` of the newsletter so the client can re-fetch it to get the updated version 🤔

ErisDS:
Ok, that all makes sense to me :)

matthanley (approved):

PR #14511, opened 2022-04-19
Added migration to backfill offer ids in subscriptions

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1520

- Sets the `offer_id` in the `members_stripe_customers_subscriptions` table based on the `offer_redemptions` that have the same tier and cadence
- We currently use the same subscription <-> offer linking when viewing a member
- The MySQL query is quite optimized in a single UPDATE query, but in SQLite we'll need to run (maximum) one UPDATE query for every offer (not per subscription).
- Best to merge this migration in 4.x (not in 5.0) because it is better (less error prone) to run this migration before starting to fill the offer_id field for updated migrations instead of after (https://github.com/TryGhost/Ghost/pull/14488)
- We need the SQLite migration for sites that will only migrate to MySQL at 5.0

Merge with/after https://github.com/TryGhost/Ghost/pull/14488

### DIFF
--- a/core/server/data/migrations/versions/4.45/2022-04-19-12-23-backfill-subscriptions-offers.js
+++ b/core/server/data/migrations/versions/4.45/2022-04-19-12-23-backfill-subscriptions-offers.js
@@ -30,30 +30,14 @@ module.exports = createTransactionalMigration(
             const updatedRows = result.length;
             const subscriptionsToUpdate = result;
 
-            // Group by offer ID (we make the assumption that we will have a limited amount of offers)
-            const groupedByOffers = {};
-            // eslint-disable-next-line no-restricted-syntax
-            for (const u of subscriptionsToUpdate) {
-                if (groupedByOffers[u.offer_id]) {
-                    groupedByOffers[u.offer_id].push(u.subscription_id);
-                } else {
-                    groupedByOffers[u.offer_id] = [u.subscription_id];
-                }
-            }
-
             logging.info(`Setting the offer_id for ${updatedRows} members_stripe_customers_subscriptions`);
 
             // eslint-disable-next-line no-restricted-syntax
-            for (const offerId of Object.keys(groupedByOffers)) {
-                const subscriptionIdChunks = chunk(groupedByOffers[offerId], 998);
-
+            for (const u of subscriptionsToUpdate) { 
                 // eslint-disable-next-line no-restricted-syntax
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.45/2022-04-19-12-23-backfill-subscriptions-offers.js
```diff
@@ -0,0 +1,77 @@
...
+                    // eslint-disable-next-line no-restricted-syntax
+                    await knex('members_stripe_customers_subscriptions')
+                        .update('offer_id', offerId)
+                        .whereIn('id', subscriptionIdChunk);
+                }
+            }
+        } else {
+            // Single update query
+            const query = `
+                UPDATE
```
allouis, line 61: Does this syntax work the same as `WITH`?

e.g. `WITH (${subquery}) AS c`

core/server/data/migrations/versions/4.45/2022-04-19-12-23-backfill-subscriptions-offers.js
```diff
@@ -0,0 +1,77 @@
...
+        if (DatabaseInfo.isSQLite(knex)) {
+            // Less optimized for SQLite
+            const result = await knex.raw(subquery);
+            const updatedRows = result.length;
+            const subscriptionsToUpdate = result;
+
+            // Group by offer ID (we make the assumption that we will have a limited amount of offers)
+            const groupedByOffers = {};
+            // eslint-disable-next-line no-restricted-syntax
+            for (const u of subscriptionsToUpdate) {
```
allouis, line 36: Honestly I would simplify further for SQLite
```js
for (const u of subscriptionsToUpdate) {
  await knex('members_stripe_customers_subscriptions')
    .update('offer_id', u.offer_id)
    .where('id', u.subscription_id);
}
```

allouis (approved):
Looks good - one comment on SQLite stuff

core/server/data/migrations/versions/4.45/2022-04-19-12-23-backfill-subscriptions-offers.js
SimonBackx: Okay, I'll revert it to that (had this initially but was not sure about performance) 😄

matthanley (approved):

daniellockyer (approved):

PR #14515, opened 2022-04-19
Updated members API to return subscriptions' offers from the offer_id

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1520

**Changes in members repo**

Bumped to `5.12.0`, with the following changes:

- Compare differences via https://github.com/TryGhost/Members/compare/%40tryghost/members-api%405.11.1...%40tryghost/members-api%405.12.0
- Instead of doing the matching of the offers and subscriptions by looking at the offer redemptions, we can now look at the offer_id from subscriptions.
- This also fixes an issue where we don't attach the offer object to subscriptions in the members' browse method
- Updated browse behaviour to match the read behaviour of members (product relation needs to get loaded because it is missing in member.products if the subscription is expired).

**Tests**
- Includes test to see if the API correctly returns the offer object when fetching one or multiple members
- Check if the return format is the same for the read, edit and browse members admin API endpoints (offer was missing in subscriptions)
- Snapshot files have been updated because now the offer is returned in subscriptions (content length increased)

Merge https://github.com/TryGhost/Ghost/pull/14511 before merging this PR

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "@tryghost/logging": "2.1.5",
     "@tryghost/magic-link": "1.0.21",
     "@tryghost/member-events": "0.4.1",
-    "@tryghost/members-api": "5.11.1",
+    "@tryghost/members-api": "5.12.0",
     "@tryghost/members-events-service": "0.3.3",
     "@tryghost/members-importer": "0.5.8",
     "@tryghost/members-offers": "0.11.1",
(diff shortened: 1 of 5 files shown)


matthanley (approved):
LGTM, just requested a second set of 👀 for the members API changes from @allouis

allouis (approved):

PR #14518, opened 2022-04-20
Map subscribers to newsletter

matthanley:
Closes https://github.com/TryGhost/Team/issues/1508
Refs https://github.com/TryGhost/Ghost/pull/14468

- Maps existing subscribers to the default newsletter

A note on performance:
We loop over the rows in a potentially large table (members) but I've minimised the impact by limiting the columns we fetch. The alternative is a raw SQL query like the one below: the SQL version takes ~0.9s vs ~1.1s for the migration (my laptop, ~30k members). The disadvantage of the raw SQL implementation is the approximation of the ObjectID (instead of a legit bson ID) which isn't sequential and may impact index size/performance.

```sql
insert into members_newsletters (id, member_id, newsletter_id)
    select
        substr(replace(uuid(),'-',''),1,24),
        id,
        '62595dcbfad4e031d85d166f'
    from members
    where members.subscribed=true;
```


matthanley:
Note the tests won't actually test the migration until after https://github.com/TryGhost/Ghost/pull/14468 is merged to create the default newsletter.

SimonBackx (approved):

core/server/data/migrations/versions/4.46/2022-04-20-08-39-map-subscribers-to-default-newsletter.js
```diff
@@ -0,0 +1,74 @@
...
+        if (!memberIds.length) {
+            logging.info(`No members to subscribe - skipping`);
+            return;
+        }
+
+        logging.info(`Found ${memberIds.length} members to subscribe`);
+
+        let pivotRows = [];
+        // eslint-disable-next-line no-restricted-syntax
+        for (const memberId of memberIds) {
```
daniellockyer, line 40: Could you switch to `const pivotRows = membersIds.map(m => ...` here to save the for-loop and therefore the eslint exemption?

core/server/data/migrations/versions/4.46/2022-04-20-08-39-map-subscribers-to-default-newsletter.js
```diff
@@ -0,0 +1,74 @@
...
+                newsletter_id: newsletter.id
+            });
+        }
+
+        const chunkSize = 1000;
+        const pivotChunks = chunk(pivotRows, chunkSize);
+
+        // eslint-disable-next-line no-restricted-syntax
+        for (const pivotChunk of pivotChunks) {
+            await knex('members_newsletters').insert(pivotChunk);
```
daniellockyer, line 53: As mentioned in Slack, Knex has a `batchInsert` utility that does the same thing here. No real preference, but it does save a few lines 🙂

SimonBackx (approved):

PR #14521, opened 2022-04-20
Added subscription stats API

allouis:
refs https://github.com/TryGhost/Team/issues/1505
refs https://github.com/TryGhost/Team/issues/1466

This adds a subscription stats api which returns the total count as well as the negative and positive deltas for subscriptions broken down by Tier and Cadence.

We use the same strategy as with MRR where we count backwards from the known totals.


allouis:
Closed in favour of https://github.com/TryGhost/Ghost/pull/14547

PR #14525, opened 2022-04-21
Added the newsletter API `?include=count.posts` and `?include=count.members` query options

tpatel:
refs https://github.com/TryGhost/Team/issues/1524

- This enables admins in the ghost admin to have an overview of the total posts/members associated with a newsletter.
- Follows the `?include=count.x` convention used by other resources

### DIFF
--- a/core/server/api/canary/newsletters.js
+++ b/core/server/api/canary/newsletters.js
@@ -13,6 +13,7 @@ module.exports = {
 
     browse: {
         options: [
+            'include',
             'filter',
             'fields',
             'limit',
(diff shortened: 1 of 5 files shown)


core/server/api/canary/newsletters.js
```diff
@@ -17,6 +18,13 @@ module.exports = {
             'order',
```
ErisDS, line 18: You're missing include in the permitted options

tpatel: yes that's what I've found too, thanks for looking into that 🙏

matthanley (approved):

SimonBackx (approved):

PR #14526, opened 2022-04-21
Deleted all v1/v2/v3 migrations

daniellockyer:
Please see the individual commits for context: https://github.com/TryGhost/Ghost/pull/14526/commits


core/server/data/migrations/utils.js
```diff
@@ -490,6 +490,23 @@ function addSetting({key, value, type, group}) {
...
+/**
+ * @param {number} major
+ */
+function createFinalMigration(major) {
+    return createTransactionalMigration(
+        async function up() {
+            throw new errors.InternalServerError({
+                message: `Unable to run migrations`,
+                context: `You must be on the latest v${major}.x to update across major versions - https://ghost.org/docs/woohoo/`,
+                help: `Run 'ghost update v${major}' to get the latest v${major}.x version, then run 'ghost update' to get to the latest.`
```
ErisDS, line 502: Cool, this is good enough to merge. 

I had a local list of copy checks that I've added to and put here: https://github.com/TryGhost/Toolbox/issues/307

daniellockyer: I'll update the link to https://ghost.org/docs/update/ just for now, and then we can re-assess when going through your issue

allouis (approved):

PR #14548, opened 2022-04-21
Added `show_header_name` column to newsletters table

matthanley:
refs https://github.com/TryGhost/Team/issues/1532

- Added before the migration in https://github.com/TryGhost/Ghost/pull/14468 to populate the default newsletter
- The fixture for the default newsletter has a different value than the model and schema default
- This is because by default the newsletter name is the same as the site title, and the site title is already shown


SimonBackx:
@matthanley I think you might need to update the schema hash in one of the tests + update `core/server/data/schema/schema.js` to include the new column

matthanley:
Yep! Also needs a fixture update as well :)

core/server/data/schema/fixtures/fixtures.json
```diff
@@ -24,7 +24,8 @@
             "entries": [
                 {
                     "name":             "Default Newsletter",
-                    "slug":             "default-newsletter"
+                    "slug":             "default-newsletter",
+                    "show_header_name": false
```
ErisDS, line 28: How come this is false in the fixture, but true everywhere else? Is it that the current default has it false? I assume so but if so that detail is very much a useful "why" for the commit 😬

matthanley: This is from the issue:
> default for first newsletter automatically created = off, default for any new newsletter added manually = on

This is because by default the newsletter name is the same as the site title, and the site title is already shown by default. I'll add this to the commit

SimonBackx (approved):

daniellockyer (approved):

PR #14558, opened 2022-04-22
Backfill canceled subscription events

allouis:
refs https://github.com/TryGhost/Team/issues/1455

There are a few cases for canceled subscriptions:

1. They have now expired
2. They were canceled _before_ we added cancelation events
3. They were canceled _after_ we added cancelation events

For 1 we can ignore them, because the MRR has corrected itself.

For 2 we must insert an event with a negative mrr_delta
For 3 we must update the cancellation event with the negative mrr_delta

Because we calculate the mrr_delta from the mrr column of a subscription, this migration must run before the one where we update mrr based on cancelations: https://github.com/TryGhost/Ghost/commit/95baa87f96651003a97eac768a0596c8818307f0


core/server/data/migrations/versions/5.0/2022-04-21-15-44-backfill-mrr-events-for-canceled-subscriptions.js
```diff
@@ -0,0 +1,66 @@
+const ObjectID = require('bson-objectid').default;
+const logging = require('@tryghost/logging');
+
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
```
allouis, line 6: Will change this back to createIrreversibleMigrations

core/server/data/migrations/versions/5.0/2022-04-21-15-44-backfill-mrr-events-for-canceled-subscriptions.js
```diff
@@ -0,0 +1,66 @@
...
+                    subscription_id: subscription.id,
+                    member_id: subscription.member_id,
+                    currency: subscription.plan_currency,
+                    mrr_delta: -subscription.mrr
+                })
+            };
+        }, {toUpdate: [], toInsert: []});
+
+        await knex.batchInsert('members_paid_subscription_events', toInsert);
+        for (const event of toUpdate) {
```
allouis, line 61: Not sure about this, I think there's probably some nested update query we can do - not sure if it's worth the complexity though

core/server/data/migrations/versions/5.0/2022-04-21-15-44-backfill-mrr-events-for-canceled-subscriptions.js
```diff
@@ -0,0 +1,66 @@
...
+            );
+
+        if (subscriptions.length === 0) {
+            logging.info('No canceled subscriptions found, skipping migration.');
+            return;
+        } else {
+            logging.info(`Found ${subscriptions.length} canceled subscriptions, inserting MRR events.`);
+        }
+
+        const {toUpdate, toInsert} = subscriptions.reduce((memo, subscription) => {
```
allouis, line 32: We already loop here, so if performance is a concern, I can do the DB updates in this loop and generate the inserts to be batched

core/server/data/migrations/versions/5.0/2022-04-21-15-44-backfill-mrr-events-for-canceled-subscriptions.js
```diff
@@ -0,0 +1,66 @@
...
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const subscriptions = await knex('members_stripe_customers_subscriptions')
+            .leftJoin('members_paid_subscription_events', 'members_stripe_customers_subscriptions.id', '=', 'members_paid_subscription_events.subscription_id')
+            .join('members_stripe_customers', 'members_stripe_customers_subscriptions.customer_id', '=', 'members_stripe_customers.customer_id')
+            .join('members', 'members_stripe_customers.member_id', '=', 'members.id')
+            .where('members_stripe_customers_subscriptions.cancel_at_period_end', '=', true)
+            .where('members_stripe_customers_subscriptions.mrr', '!=', 0)
+            .where(knex.raw(`(members_paid_subscription_events.type = 'canceled' OR members_paid_subscription_events.type is null)`))
```
matthanley, line 14: any reason not to use the `.where((query) => {...})` syntax for this?

allouis: Not particularly! Do you want me to change it?

core/server/data/migrations/versions/5.0/2022-04-21-15-44-backfill-mrr-events-for-canceled-subscriptions.js
```diff
@@ -0,0 +1,77 @@
...
+            logging.info(`Found ${canceledSubscriptions.length} canceled subscriptions, updating MRR events`);
+        }
+
+        const canceledEvents = await knex('members_paid_subscription_events')
+            .select(
+                'id',
+                'subscription_id'
+            )
+            .where('type', '=', 'canceled')
+            .whereIn('subscription_id', canceledSubscriptions.map(x => x.id))
```
matthanley, line 35: Is this safe given that we don't know how many entries we might have?

allouis: As long as we haven't got more than 32k we should be good!

allouis:

PR #14561, opened 2022-04-22
Added migration to adjust mrr in subscriptions for cancel_at_period_e…

SimonBackx:
…nd (#14480)

refs https://github.com/TryGhost/Team/issues/1515

- Sets MRR to zero for all subscriptions where `cancel_at_period_end` is true
- Rollbacks to same logic as migration https://github.com/TryGhost/Ghost/pull/14465
- We need to be carefull that future backfill adjustments of MRR happen after this one for the rollback to keep working


SimonBackx:
(this migration has been merged earlier but was dropped accidentally during a force push)

daniellockyer (approved):

allouis:

PR #14562, opened 2022-04-22
Use newsletter design settings

tpatel:
refs https://github.com/TryGhost/Team/issues/1550

- Switched to using the newsletter design settings over the global settings
- Only switch if the post is associated with a unarchived newsletter
- Made the `newsletter_id` property available in the Admin API Post resource
- Added the `showHeaderName` variable that can be used in the post html template

### DIFF
--- a/core/server/services/mega/post-email-serializer.js
+++ b/core/server/services/mega/post-email-serializer.js
@@ -171,40 +171,33 @@ const parseReplacements = (email) => {
 };
 
 const getTemplateSettings = async (newsletterId = null) => {
+    let newsletter;
+    if (newsletterId) {
+        newsletter = await models.Newsletter.findOne({id: newsletterId, status: 'active'});
+    } else {
+        const newsletters = await models.Newsletter.findPage({status: 'active', limit: 1});
+        newsletter = newsletters.data[0];
+    }
+
     const accentColor = settingsCache.get('accent_color');
     const adjustedAccentColor = accentColor && darkenToContrastThreshold(accentColor, '#ffffff', 2).hex();
     const adjustedAccentContrastColor = accentColor && textColorForBackgroundColor(adjustedAccentColor).hex();
 
     const templateSettings = {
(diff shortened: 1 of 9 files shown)


test/unit/server/services/mega/post-email-serializer.test.js
```diff
@@ -81,4 +84,91 @@ describe('Post Email Serializer', function () {
...
+    describe('getTemplateSettings', function () {
+        before(function () {
+            models.init();
+        });
+
+        afterEach(function () {
+            sinon.restore();
+        });
+
+        it('uses the gloal settings when no newsletter_id is defined', async function () {
```
matthanley, line 97: Emails should always have a newsletter ID now, but we should fall back to the settings for the default newsletter. We'll be dropping the global settings in 5.0.

core/server/services/mega/post-email-serializer.js
```diff
@@ -169,21 +170,30 @@ const parseReplacements = (email) => {
     return replacements;
 };
 
-const getTemplateSettings = async () => {
+const getTemplateSettings = async (newsletterId = null) => {
+    let newsletter;
+    if (newsletterId) {
+        newsletter = await models.Newsletter.findOne({id: newsletterId, status: 'active'});
```
SimonBackx, line 176:
```suggestion
        newsletter = await models.Newsletter.findOne({id: newsletterId, filter: 'status:active'});
```
I think filtering is only supported with the filter property

core/server/services/mega/post-email-serializer.js
```diff
@@ -169,21 +170,30 @@ const parseReplacements = (email) => {
...
+    } else {
+        const newsletters = await models.Newsletter.findPage({status: 'active', limit: 1});
```
SimonBackx, line 178:
```suggestion
        const newsletters = await models.Newsletter.findPage({filter: 'status:active', limit: 1});
```

PR #14568, opened 2022-04-25
Fixed created_at newsletter migration

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1551

One of the existing migrations has changed. This should be allowed because that one isn't merged in `main`. The end result and new migrations are also compatible with users that already ran the previous version of that migration (so don't have a nullable created_at, I've tested this). The end result for everyone is the same, so no rollback is required.

- Updated existing migration to insert a nullable created_at column
- Added a migration to update all the created_at values to now
- Added a migration to drop nullable
- Also includes new helper methods to set and drop nullable

### DIFF
--- a/core/server/data/migrations/versions/4.46/2022-04-13-12-02-fill-created-at-newsletters.js
+++ b/core/server/data/migrations/versions/4.46/2022-04-13-12-02-fill-created-at-newsletters.js
@@ -1,0 +1,18 @@
+const logging = require('@tryghost/logging');
+
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Setting missing created_at values for existing newsletters');
+
+        const updatedRows = await knex('newsletters')
+            .where('created_at', null)
+            .update('created_at', new Date());
+
+        logging.info(`Updated ${updatedRows} newsletters with created_at = now`);
+    },
+    async function down() {
+        // Not required: we would lose information here.
+    }
+);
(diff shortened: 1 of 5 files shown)


core/server/data/migrations/versions/4.46/2022-04-13-12-02-fill-created-at-newsletters.js
```diff
@@ -0,0 +1,18 @@
...
+
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Setting missing created_at values for existing newsletters');
+
+        const updatedRows = await knex('newsletters')
+            .where('created_at', null)
+            .update('created_at', new Date());
```
daniellockyer, line 11: I think this needs to be 

```
const currentTimestamp = knex.raw('CURRENT_TIMESTAMP');

...


    .update('created_at', currentTimestamp);
```

SimonBackx, line 11:
```suggestion

        const now = knex.raw('CURRENT_TIMESTAMP');
        const updatedRows = await knex('newsletters')
            .where('created_at', null)
            .update('created_at', now);
```

SimonBackx: I've updated it. I would expect Date objects to get mapped correctly, but I couldn't find any documentation about it

daniellockyer (approved):

matthanley (approved):

PR #14579, opened 2022-04-26
Added newsletters to subscribe events

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1478

**Changes (bumps `members-api` to 6.0.0-alpha):**
- View changes: https://github.com/TryGhost/Members/compare/%40tryghost/members-api%405.12.0...%40tryghost/members-api%406.0.0-alpha.0
- Moved all admin API members tests to enable the multiple newsletters flag
- Checks if the susbcribe events are added correctly when adding or removing newsletters
- Checks if susbcribe events are added for default newsletters
- Fixes an error that was caused by enabling the multiple newsletters flag in the tests.

**Added orderDefaultRaw in Newsletter model**

**Tests:**
- Moved all admin API members tests to enable the multiple newsletters flag
- Checks if the susbcribe events are added correctly when adding or removing newsletters
- Checks if susbcribe events are added for default newsletters


SimonBackx:
I'm stuck on the snapshots failing here. The order of the returned members is different locally than in CI, but this is in a completely unrelated test

core/server/models/newsletter.js
```diff
@@ -72,6 +72,10 @@ const Newsletter = ghostBookshelf.Model.extend({
         return query;
     }
 }, {
+    orderDefaultRaw: function (options) {
+        return 'sort_order ASC, created_at ASC, id ASC';
```
ErisDS, line 76: Strictly this shouldn't be necessary. But I'll raise a separate issue. The implementation of `findAll` should use orderDefaultOptions but it's only used in `findPage`

ErisDS: Tracked here: https://github.com/TryGhost/Toolbox/issues/310

SimonBackx: Perfect! I tried to solve it in here (https://github.com/TryGhost/Ghost/pull/14579/commits/80f35dacc02d3cd1505a1e6702efd9b5362a8bc4), but it wasn't working so I just reverted. Better to fix in a different PR indeed

SimonBackx:
@matthanley Tests are working locally, so it should be save to merge if we solve the ordering issue of members in a different PR

PR #14586, opened 2022-04-26
Removed author_id from posts

naz:
refs https://github.com/TryGhost/Toolbox/issues/230

Notes from my own testing of the migration dropping `author_id` column from the `posts` table with 10k records:
The execution time was good, and the default algorithm was `copy`. Note, running my local instance against MySQL 5.7.29
```
  (no copy version)
  knex:query alter table `posts` drop `author_id` trx6 +2ms 

  (with copy version)
  knex:query alter table `posts` drop `author_id`, algorithm=copy trx6 +2ms
``` 

When testing this change rebased over https://github.com/TryGhost/Ghost/pull/14699, the timing was the same as above.


naz:
@daniellockyer @ErisDS only the migrations commit here needs your :eyes: it's this one: ":building_construction: Removed author_id column from the posts table" ([fb493db](https://github.com/TryGhost/Ghost/pull/14586/commits/fb493db89a9cb7d3a710b17ed4713535c284616a) hash might change if I force push again). 

I'm slightly unsure about the override for "down" migration. The reason it's there is because we do pass an empty `{}` parameter to the migration utility to make it valid method call (the third parameter is required, sorta). We don't actually need to support the down migration here, but didn't want to leave a migration in half broken state anyway :sweat_smile: Lemme know what you think!

naz:
Rebased the changes over main, but seems like we had some divergence with 5.0 based on:
Screenshot from 2022-05-02 11-11-38

core/server/data/migrations/versions/5.0/2022-04-28-03-26-remove-author-id-column-from-posts-table.js
```diff
@@ -0,0 +1,7 @@
+const {createDropColumnMigration} = require('../../utils');
+
+module.exports = createDropColumnMigration('posts', 'author_id', {});
```
daniellockyer, line 3: We can't use the utility on the `posts` table because it doesn't include `algorithm=copy`, which is a performance optimization needed for large tables

PR #14588, opened 2022-04-26
Use newsletter settings in emails

tpatel:
refs https://github.com/TryGhost/Team/issues/1550

More details in individual commits

### DIFF
--- a/core/server/services/mega/mega.js
+++ b/core/server/services/mega/mega.js
@@ -56,8 +56,7 @@ const getReplyToAddress = (fromAddress, replyAddressOption) => {
 const getEmailData = async (postModel, options) => {
     let newsletter = await postModel.related('newsletter').fetch();
     if (!newsletter) {
-        const newsletters = await models.Newsletter.findPage({filter: 'status:active', limit: 1});
-        newsletter = newsletters.data[0];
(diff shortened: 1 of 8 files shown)


core/server/services/mega/mega.js
```diff
@@ -60,14 +54,29 @@ const getReplyToAddress = () => {
  * @param {ValidAPIVersion} options.apiVersion - api version to be used when serializing email data
  */
 const getEmailData = async (postModel, options) => {
-    const {subject, html, plaintext} = await postEmailSerializer.serialize(postModel, options);
+    let newsletter = await postModel.related('newsletter').fetch();
+    if (!newsletter) {
+        const newsletters = await models.Newsletter.findPage({filter: 'status:active', limit: 1});
```
matthanley, line 59: Similarly here ([ref](https://github.com/TryGhost/Ghost/pull/14588/files#r858937909))

SimonBackx (approved):
Looks good. Slightly unrelated, it looks like we don't have tests for the serialize method? Would be great if we would add e2e tests, and maybe some sort of snapshot tests for the email template so we know when they break.

tpatel:
For context, we have plans to move the email template to a proper handlebars template to solve [an existing issue](https://github.com/TryGhost/Team/issues/1379).

PR #14597, opened 2022-04-27
Added migration that backfills newsletter_id on all subscribe events

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1478

Sets the newsletter_id to the id of the default newsletter.

### DIFF
--- a/core/server/data/migrations/versions/4.46/2022-04-27-07-59-set-newsletter-id-subscribe-events.js
+++ b/core/server/data/migrations/versions/4.46/2022-04-27-07-59-set-newsletter-id-subscribe-events.js
@@ -16,7 +16,7 @@ module.exports = createTransactionalMigration(
             return;
         }
 
-        // Set susbcribe events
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/4.46/2022-04-27-07-59-set-newsletter-id-subscribe-events.js
```diff
@@ -0,0 +1,31 @@
...
+            .orderBy('created_at', 'asc')
+            .orderBy('id', 'asc')
+            .first('id');
+
+        if (!newsletter) {
+            logging.error(`Default newsletter not found - skipping`);
+            return;
+        }
+
+        // Set susbcribe events
```
SimonBackx, line 19:
```suggestion
        // Set subscribe events
```

matthanley (approved):

daniellockyer (approved):

PR #14600, opened 2022-04-27
Added uuid to newsletter schema 

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1561

With multiple newsletters, unsubscribe links will also need to have a unique reference to the newsletter that the email is for, so that we can unsubscribe members from the particular newsletter automatically when they click on the link.
As our existing pattern for members is to use UUID as the external unique reference, this change adds UUID to newsletter schema and populates the existing newsletters with a UUID value.

- adds new `uuid` column to newsletter schema
- updates newsletter model to add default uuid


core/server/data/migrations/versions/4.46/2022-04-13-12-58-fill-uuid-for-newsletters.js
```diff
@@ -0,0 +1,19 @@
...
+
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Setting missing uuid values for existing newsletters');
+        const uuidV4 = uuid.v4();
+        const updatedRows = await knex('newsletters')
+            .where('uuid', null)
+            .update('uuid', uuidV4);
```
matthanley, line 12: Doesn't this try to give all rows the same UUID, so it'll fail if there's more than one? 🤔

rishabhgrg: Eugh yea good point, updating. I had in my mind only a single newsletter case here but we should definitely not rely on that.

rishabhgrg: Updated.

matthanley (approved):

core/server/data/migrations/versions/4.46/2022-04-13-12-57-add-uuid-column-to-newsletters.js
```diff
@@ -0,0 +1,8 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('newsletters', 'uuid', {
+    type: 'string',
+    maxlength: 36,
+    nullable: true,
```
daniellockyer, line 6: Should we be dropping the `nullable` on this field after we've populated it in the migration below?

SimonBackx (approved):
I also agree that it is best to drop the nullable after the backfill migration. But apart from that, everything looks okay 👌

core/server/data/migrations/versions/4.46/2022-04-13-12-57-add-uuid-column-to-newsletters.js
rishabhgrg: Does that mean we update the `schema` to have `nullable:false` and then just add a migration to drop it after ?

daniellockyer: I believe so 🤔 https://github.com/TryGhost/Ghost/pull/14568/files#diff-18e7c5d15a806349f8a27507183762f3ef9d201415934325f2058f7d2c858c6b

daniellockyer (approved):

SimonBackx (approved):

PR #14610, opened 2022-04-27
Fixed bulk unsubscribe and updated member import tests

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1567

**Changes in members-api**
- Compare changes: https://github.com/TryGhost/Members/compare/%40tryghost/members-api%406.1.0...%40tryghost/members-api%406.2.2
- Fixed bulk unsubscribe
- Deletes the newsletter relations instead of setting subscribed to false

**Test fail fix**
refs https://github.com/TryGhost/Ghost/pull/14621
refs https://ghost.slack.com/archives/C02G9E68C/p1651126990299689?thread_ts=1651072733.859939&cid=C02G9E68C

- Events didn't always have the same created_at as created members
- This caused a test to fail randomly in the main repo

**Changes**
- Added required helpers for members-api package
- Added required helpers for members-api package
- Version bumps of other packages are only tooling related

**Tests**
- Tests if member import still works with the legacy `subscribed` flag
- Updated member importer to use multipleNewsletters flag
- Dropped legacy members tests


matthanley (approved):

SimonBackx:
Issue of the tests failing was that some dependencies were not updated together with members-api, which caused multiple package versions getting used. Which in turn made sure events weren't send correctly (because listeners and senders were attached to different objects)

PR #14623, opened 2022-04-27
Set `importer` context for members importer

aileen:
no issue
needs https://github.com/TryGhost/Members/pull/384/commits/82eb861955577c6985bf2ebff6b30d1d82f10471

Set the context to `importer` for any members importer actions which will determine the correct source of origin in the members events table


daniellockyer:
Merging in favor of https://github.com/TryGhost/Ghost/pull/14622, which will be in 5.0 anyway

PR #14624, opened 2022-04-28
Added member endpoints for managing newsletter subscriptions

rshbhgrg:
refs TryGhost/Team#1495

With multiple newsletters, members are allowed to manage their newsletter pref via email unsubscribe link with member uuid. Since Portal needs to manage member's newsletter pref via their UUID, we need new endpoints on members that allow fetch/update of newsletter subscriptions via only uuid. The endpoints return only limited data for a member that are needed for the UI.

- adds endpoint to fetch newsletter subscriptions for member via uuid
- adds endpoint to update newsletter subscriptions for member via uuid

### DIFF
--- a/core/server/services/members/middleware.js
+++ b/core/server/services/members/middleware.js
@@ -85,7 +85,7 @@ const getMemberNewsletters = async function (req, res) {
         });
         if (!memberData) {
             res.writeHead(400);
-            res.end('Email address not found.');
+            return res.end('Email address not found.');
         } else {
             const data = _.pick(memberData.toJSON(), 'uuid', 'email', 'name', 'newsletters', 'status');
             return res.json(data);
@@ -109,7 +109,7 @@ const updateMemberNewsletters = async function (req, res) {
         });
         if (!memberData) {
             res.writeHead(400);
-            res.end('Email address not found.');
+            return res.end('Email address not found.');
         }
         const options = {
             id: memberData.get('id'),
--- a/core/server/services/members/middleware.js
+++ b/core/server/services/members/middleware.js
@@ -69,6 +69,57 @@ const getOfferData = async function (req, res) {
     });
 };
 
+const getMemberNewsletters = async function (req, res) {
+    try {
+        const memberUuid = req.query.uuid;
+
+        const memberData = await membersService.api.members.get({
+            uuid: memberUuid
+        }, {
+            withRelated: ['newsletters']
+        });
+        if (!memberData) {
+            res.writeHead(400);
+            res.end('Email address not found.');
+        } else {
+            const data = _.pick(memberData.toJSON(), 'uuid', 'email', 'name', 'newsletters', 'status');
+            return res.json(data);
+        }
+    } catch (err) {
+        res.writeHead(400);
+        res.end('Failed to unsubscribe this email address');
+    }
+};
+
+const updateMemberNewsletters = async function (req, res) {
+    try {
+        const memberUuid = req.query.uuid;
+        if (!memberUuid) {
+            res.writeHead(400);
+            return res.end('Missing member uuid');
+        }
+        const data = _.pick(req.body, 'newsletters');
+        const memberData = await membersService.api.members.get({
+            uuid: memberUuid
+        });
+        if (!memberData) {
+            res.writeHead(400);
+            return res.end('Member not found');
+        }
+        const options = {
+            id: memberData.get('id'),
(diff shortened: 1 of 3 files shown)


core/server/services/members/middleware.js
```diff
@@ -69,6 +69,57 @@ const getOfferData = async function (req, res) {
     });
 };
 
+const getMemberNewsletters = async function (req, res) {
+    try {
+        const memberUuid = req.query.uuid;
```
matthanley, line 74: We should probably guard against this existing before passing it to the members service - I'm not sure what happens if we pass `{uuid: undefined}` to `members.get`

core/server/services/members/middleware.js
```diff
@@ -69,6 +69,57 @@ const getOfferData = async function (req, res) {
...
+            res.writeHead(400);
+            return res.end('Missing member uuid');
+        }
+        const data = _.pick(req.body, 'newsletters');
+        const memberData = await membersService.api.members.get({
+            uuid: memberUuid
+        });
+        if (!memberData) {
+            res.writeHead(400);
+            return res.end('Member not found');
```
matthanley, line 107: `get` and `update` should return the same error for non-existence

matthanley (changes requested):

test/e2e-frontend/members.test.js
```diff
@@ -116,6 +116,26 @@ describe('Front-end members behaviour', function () {
...
+
+        it('should error for updating member newsletters with missing uuid', async function () {
+            await request.put('/members/api/member/newsletters')
+                .expect(400);
+        });
+
+        it('should error for updating member newsletters with invalid uuid', async function () {
+            await request.put('/members/api/member/newsletters?uuid=abc')
+                .expect(400);
+        });
```
ErisDS, line 137: We are missing tests for the success cases. IMO line 59 needs to change from

`await testUtils.initFixtures('members');` 

to

`await testUtils.initFixtures('newsletters', 'members:newsletters');` 

and we should be testing that we get the data we expect?

core/server/services/members/middleware.js
ErisDS: non-existence should also be a 404

rishabhgrg: ✅

core/server/services/members/middleware.js
rishabhgrg: ✅

core/server/services/members/middleware.js
```diff
@@ -69,6 +69,57 @@ const getOfferData = async function (req, res) {
...
+        const memberUuid = req.query.uuid;
+
+        const memberData = await membersService.api.members.get({
+            uuid: memberUuid
+        }, {
+            withRelated: ['newsletters']
+        });
+        if (!memberData) {
+            res.writeHead(400);
+            res.end('Email address not found.');
```
daniellockyer, line 88: Can we `return` here? If we add more code after this block in the future, the flow will leak through and we'll get the "already sent headers" error

core/server/services/members/middleware.js
```diff
@@ -69,6 +69,62 @@ const getOfferData = async function (req, res) {
...
+            res.writeHead(400);
+            return res.end('Invalid member uuid');
+        }
+        const data = _.pick(req.body, 'newsletters');
+        const memberData = await membersService.api.members.get({
+            uuid: memberUuid
+        });
+        if (!memberData) {
+            res.writeHead(400);
+            res.end('Email address not found.');
```
daniellockyer, line 112: This needs a `return;`

core/server/services/members/middleware.js
matthanley: Unresolving for Hannah's 404 comment (in both the get and update)

test/e2e-frontend/members.test.js
rishabhgrg: @ErisDS Yea totally, updated now, thanks 🙏

core/server/services/members/middleware.js
rishabhgrg: Done, thanks.

core/server/services/members/middleware.js
rishabhgrg: Done, thanks

matthanley (approved):

core/server/services/members/middleware.js
rishabhgrg: 🙏 Done

PR #14627, opened 2022-04-28
Fixed issue with creation of multiple newsletters

matthanley:
- Newsletter defaults were a static object with the default UUID resolved once
- This caused conflicts on subsequent create requests
- Defaults are now resolved as a function so each call gets a new UUID


rshbhgrg (approved):
Good catch!

PR #14629, opened 2022-04-28
Fixed issue with new members always subscribing to defaults

SimonBackx:
no issue

The member was updated when setting the geolocation, but that also included setting subscribed to true.

members-api change: https://github.com/TryGhost/Members/commit/805c745d7bc7a1664365c59031a3ef0d4c92e0d3


matthanley (approved):
Really nice 👌🏼

PR #14639, opened 2022-05-02
🐛 [4.X] Fixed null values in settings breaking default newsletter migration

SimonBackx:
refs https://ghost.slack.com/archives/C02G9E68C/p1651484563907609

- When the site has an empty name, it is set to NULL in settings.
- The name column is not nullable in newsletters, breaking the migration in that case.
- Fixed by excluding all non-nullable columns
- Replaced JS Date object with raw SQL `CURRENT_TIMESTAMP`


daniellockyer (approved):

daniellockyer (approved):

gllmhyt:
Despite this fix, today I got this error: Message: Ghost was able to start, but errored during boot with: alter table `newsletters` modify  `created_at` datetime null default 'NULL' - Invalid default value for 'created_at'

PR #14646, opened 2022-05-02
Rewrite old version URLs to unversioned URLs with headers

ErisDS:
closes: https://github.com/TryGhost/Toolbox/issues/315

- For all the current versioned URLs, rewrite the URL as unversioned
- Add the accept-version header
- Add a deprecated header
- TODO: add a Location or Link header

### DIFF
--- a/core/server/web/api/app.js
+++ b/core/server/web/api/app.js
@@ -1,6 +1,7 @@
 const debug = require('@tryghost/debug')('web:api:default:app');
 const config = require('../../../shared/config');
 const express = require('../../../shared/express');
+const urlUtils = require('../../../shared/url-utils');
 const sentry = require('../../../shared/sentry');
 const errorHandler = require('@tryghost/mw-error-handler');
 const APIVersionCompatibilityService = require('../../services/api-version-compatibility');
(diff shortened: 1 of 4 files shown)


core/server/web/api/app.js
```diff
@@ -14,16 +15,27 @@ module.exports = function setupApiApp() {
...
+        if (version === 'canary') {
+            version = 'v4';
         }
-        res.redirect(307, versionlessURL);
+
+        // Rewrite the url
+        req.url = versionlessUrl;
+
+        // Add the accept-version header so our internal systems will act as if it was set on the request
+        req.headers['accept-version'] = `${version}.0`;
```
naz, line 32: I think we should be overriding the `accept-header` value _only if_ it's not present in the request. The clients can set a more specific minor version through the SDKs, so we'd be loosing valuable information. Not sure what the usecase is overriding it here?

core/server/web/api/app.js
```diff
@@ -14,16 +15,27 @@ module.exports = function setupApiApp() {
...
+
+        res.header('Deprecation', `version="${version}"`);
```
naz, line 34: Would be super cool to add a link to our docs here like the [Deprecation header spec](https://tools.ietf.org/id/draft-dalal-deprecation-header-03.html#documentation) for `Link` suggests. This might be more useful in the 4.x branch but also while looking at the Deprecation header spec setting `Sunset` would be awesome too. 

All of this is nice to have though. Might make someones migration experience slightly nicer :)

naz: Oh and read the above as, "what are your thoughts on these additional headers" rather than "it should be done this way, period" :smile:

core/server/web/api/app.js
ErisDS: Yeah fair, I was thinking that if you're requesting a versioned URL then `accept-version` shouldn't be set, but I guess there are change-over cases where that's not true like Zapier.

Updated.

core/server/web/api/app.js
ErisDS: Yeah I was thinking we'd add that in once we do have docs. We don't have anything to link to right now. Have added a note to https://github.com/TryGhost/Toolbox/issues/307

PR #14656, opened 2022-05-03
Dropped nullable from members_subscribe_events.newsletter_id

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1570

- Includes utils to disable foreign key checks when dropping nullable from columns
- Migration to drop nullable from members_subscribe_events.newsletter_id

### DIFF
--- a/core/server/data/migrations/utils.js
+++ b/core/server/data/migrations/utils.js
@@ -443,18 +443,28 @@ function createDropColumnMigration(table, column, columnDefinition) {
 /**
  * @param {string} table
  * @param {string} column
- *
+ * @param {Object} options
+ * @param {boolean} options.disableForeignKeyChecks Disable foreign key checks for the down operation (when dropping nullable)
  * @returns {Migration}
  */
-function createSetNullableMigration(table, column) {
-    return createNonTransactionalMigration(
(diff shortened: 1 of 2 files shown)


core/server/data/migrations/utils.js
```diff
@@ -443,33 +444,59 @@ function createDropColumnMigration(table, column, columnDefinition) {
...
+    return createTransactionalMigration(
         async function up(knex) {
-            logging.info(`Dropping nullable: ${table}.${column}`);
+            if (DatabaseInfo.isSQLite(knex)) {
+                options.disableForeignKeyChecks = false;
+            }
+            logging.info(`Dropping nullable: ${table}.${column}${options.disableForeignKeyChecks ? ' with foreign keys disabled' : ''}`);
+
+            if (options.disableForeignKeyChecks) {
+                await knex.raw('SET FOREIGN_KEY_CHECKS=0;').transacting(knex);
```
matthanley, line 492: Rolling back a transaction doesn't reset the session variable. IMO this should be in a try/finally instead of using a transaction.


image

daniellockyer (approved):

matthanley (approved):

PR #14660, opened 2022-05-03
Migrated previous email sending settings to newsletters

matthanley:
refs https://github.com/TryGhost/Team/issues/1581

- The original migration to create the default newsletter omitted the from address and reply-to settings
- `sender_reply_to` and `members_reply_address` are both enums with the same values and copy straight across
- `members_from_address` had a default value of 'noreply' as the fallback, which is remapped to NULL in the newsletters table
- We apply the change to all newsletters (there should only be one outside of alpha) which haven't already been reconfigured


allouis (approved):

SimonBackx (approved):

SimonBackx:
I'm not sure if we should release this without changing `core/server/services/members/config.js` too. Because we'll have a 'hidden' logic to fallback to a value that they can't change.

matthanley:
The additional change to remove the hidden fallback is here: https://github.com/TryGhost/Ghost/pull/14665

PR #14688, opened 2022-05-04
[5.x] Fixed newsletter header image path

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1579

- When writing to the database, the header_image is tranformed to the transformReady path
- When reading from the database, the transformReady path is transformed to an absolute path
- Includes a test when adding a newsletter

Migration:
- Updates all newsletter who have a header_image to make sure it is saved in transform ready format
- Down operation is required to work with the old model logic and transforms it back to an absolute format

4.x version: https://github.com/TryGhost/Ghost/pull/14690


matthanley:
This needs to be cherry-picked from the 4.x branch instead of being reimplemented in 5.0, otherwise knex-migrator will complain about unexpected migrations in 4.47

PR #14695, opened 2022-05-04
Assigned existing email records and related posts to the default newsletter

matthanley:
refs https://github.com/TryGhost/Team/issues/1571

- With the addition of multiple newsletters, all emails sent previously should be assigned to the default newsletter
- This will make sure that the sent count for the default newsletter displays correctly

### DIFF
--- a/core/server/data/migrations/versions/5.0/2022-05-04-15-24-map-existing-emails-to-default-newsletter.js
+++ b/core/server/data/migrations/versions/5.0/2022-05-04-15-24-map-existing-emails-to-default-newsletter.js
@@ -9,7 +9,7 @@ module.exports = createTransactionalMigration(
         // Note we intentionally use the default newsletter slug instead of the usual orderBy logic
         let newsletter = await knex('newsletters')
             .where('slug', 'default-newsletter')
-            .first('id');
+            .first('id', 'slug');
 
         if (!newsletter) {
             // Fall back to orderBy - just in case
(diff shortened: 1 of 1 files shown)


matthanley:
@daniellockyer @SimonBackx this migration is failing tests because SQLite doesn't support UPDATE with JOIN. The migration itself fixes relationships that aren't required in a development environment so we can either:
- Skip the migration for SQLite and only run it on MySQL
- Refactor the migration so it runs on SQLite and MySQL

For reference, the query generated for MySQL is:
```
update `emails` inner join `posts` on `emails`.`post_id` = `posts`.`id` set `emails`.`newsletter_id` = ?, `posts`.`newsletter_id` = ? where `emails`.`newsletter_id` is null
```

SimonBackx:
The only thing I'm thinking about is what if we decide to drop nullable on emails.newsletter_id?

matthanley:
I've updated the implementation for SQLite compatibility

core/server/data/migrations/versions/5.0/2022-05-04-15-24-map-existing-emails-to-default-newsletter.js
```diff
@@ -0,0 +1,42 @@
...
+                .orderBy('id', 'asc')
+                .first('id');
+        }
+
+        if (!newsletter) {
+            logging.error(`Newsletter not found - skipping`);
+            return;
+        }
+
+        // Set newsletter ID only on posts with related email records without a newsletter assigned
```
daniellockyer, line 30: Can we add a log line here to print out the newsletter ID we've identified and will use in the queries? Helps with debugging 😛

SimonBackx (approved):

daniellockyer (approved):

PR #14697, opened 2022-05-05
Updated email unsubscribe links to use site home

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1495

- unsubscribe urls on emails were using a reserved `/unsubscribe` route with custom handlebars template for UI
- with multiple newsletters, the unsubscribe flow is handled directly by Portal, and current unsubscribe urls are redirected to site home with action params
- updates the newly generated unsubscribe links to use site home with action params directly, moving away from `SITE_URL/unsubscribe?uuid=POST_UUID` to `SITE_URL/?action=unsubscribe&uuid=POST_UUID&newsletter=NEWSLETTER_UUID`
- also allows us to remove the special unsubscribe route going forward


rishabhgrg:
Closed for now, we'd like to keep current redirect pattern for unsubscribe links instead of directly setting them up for home.

PR #14699, opened 2022-05-05
Altered migration utility to always use `algorithm=copy` for MySQL

daniellockyer:
refs https://github.com/TryGhost/Ghost/commit/2bfd8f8b7ec1f487c92b8497955f573bbd14f422

- we've previously had issues when dropping a column on `posts`
  because it's a large table and it can take a veeeeeeery long time with
  the default SQL produced by Knex
- we found a magic incantation that makes it super speedy (context in
  commit above) => `, algorithm=copy`
- for that migration, we did it all manually but we can now change the
  utility to always append this to the generated SQL so we don't have to
  think about the specific table size when adding or dropping columns
- this changes the `addColumn` and `dropColumn` utilities to append the
  string to SQL in MySQL, or just executes the SQL for SQLite

### DIFF
--- a/core/server/data/schema/commands.js
+++ b/core/server/data/schema/commands.js
@@ -71,10 +71,24 @@ function dropNullable(tableName, column, transaction = db.knex) {
     });
 }
 
-function addColumn(tableName, column, transaction = db.knex, columnSpec) {
-    return transaction.schema.table(tableName, function (table) {
+async function addColumn(tableName, column, transaction = db.knex, columnSpec) {
+    let sql = transaction.schema.table(tableName, function (table) {
         addTableColumn(tableName, table, column, columnSpec);
-    });
+    }).toSQL()[0].sql;
+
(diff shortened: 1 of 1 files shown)


naz:
@matthanley a note on testing this change - a [perfect real world candidate](https://github.com/TryGhost/Ghost/pull/14586) is the `author_id` removal. I'll rebase that PR over this one to have a nice testing ground :ok_hand:

naz (changes requested):
There's something funky going on with SQLite - when I used these changes to run dropping of `author_id` column from `posts` table, the column did not get dropped. It does get dropped when run with the previous version of `addColumn` method.

naz:
The MySQL version did work just fine :ok_hand:

daniellockyer:
@naz Nice catch! It turns out `.toSQL()` with SQLite is funky because it does the whole table dance. I've pushed a fix, so it should be good now

naz (approved):
SQLite does work now :1st_place_medal:

PR #14712, opened 2022-05-05
Dropped `oauth` table

daniellockyer:

- we're removing the OAuth prototype and the table was never used, so we
  should be good to drop it
- this commit adds a migration to drop the table


allouis (approved):
Looks good - do you need to remove `web/oauth` too? And the oath settings?

daniellockyer:
@allouis Yep, there's a bunch of cleanup but the table is disconnnected from the rest of the implementation so we should be able to get this in first

matthanley (approved):

PR #14713, opened 2022-05-05
🏗 Dropped `webhooks.status` column

daniellockyer:

- this column is not used and I was going to add `validation` to it but
  it's better to clean it up and re-add the column if we need it again

### DIFF
--- a/core/server/data/schema/schema.js
+++ b/core/server/data/schema/schema.js
@@ -341,8 +341,6 @@ module.exports = {
         // NOTE: integration_id column needs "nullable: true" -> "nullable: false" migration (recreate table with nullable: false)
         // CASE: Ghost instances initialized pre 4.0 will have this column set to nullable: true in db schema
         integration_id: {type: 'string', maxlength: 24, nullable: false, references: 'integrations.id', cascadeDelete: true},
-        // @TODO: add isIn validation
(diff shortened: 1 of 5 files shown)


core/server/data/schema/schema.js
```diff
@@ -346,8 +346,7 @@ module.exports = {
         // NOTE: integration_id column needs "nullable: true" -> "nullable: false" migration (recreate table with nullable: false)
         // CASE: Ghost instances initialized pre 4.0 will have this column set to nullable: true in db schema
         integration_id: {type: 'string', maxlength: 24, nullable: false, references: 'integrations.id', cascadeDelete: true},
-        // @TODO: add isIn validation
-        status: {type: 'string', maxlength: 50, nullable: false, defaultTo: 'available'},
+        status: {type: 'string', maxlength: 50, nullable: false, defaultTo: 'available', validations: {isIn: [['available']]}},
```
matthanley, line 349: Why do we need the field if we don't use it?

daniellockyer: Good point - I've switched it to dropping the column

daniellockyer:
@matthanley + @SimonBackx: note that I've changed the migration to dropping the `webhooks.status` column

SimonBackx (approved):

matthanley (approved):

PR #14715, opened 2022-05-06
Fixed `last_seen` updated for suspended users

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1461

- A suspended user was able to make it through the Express middlewares to the `updateUserLastSeen` middleware, until it was halted when checking the user permissions in the API pipeline. This was only the case for session logins, not for API keys.
- For API keys, the user status is checked:
https://github.com/TryGhost/Ghost/blob/6dc3f1bf5653adf010aca06f1bad4226d551c281/core/server/services/auth/api-key/admin.js#L178-L181
- In the session middleware, the `findUserById` in `getUserForSession` didn't filter on the active status of users:
https://github.com/TryGhost/Ghost/blob/be4146e3245e0812f1a7497d9280e5064568327d/core/server/services/auth/session/middleware.js#L22-L27
- This has been fixed now by updating the sessionService's `findUserById` method.

### DIFF
--- a/core/server/services/auth/session/index.js
+++ b/core/server/services/auth/session/index.js
@@ -32,7 +32,7 @@ const sessionService = createSessionService({
     getOriginOfRequest,
     getSession: expressSession.getSession,
     findUserById({id}) {
-        return models.User.findOne({id});
+        return models.User.findOne({id, status: 'active'});
     }
 });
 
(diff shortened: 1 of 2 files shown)


matthanley:
@SimonBackx could you put a test in for this?

matthanley (approved):

PR #14733, opened 2022-05-09
Added internal frontend integration

ErisDS:
refs: https://github.com/TryGhost/Team/issues/1599

- add an internal integration for Ghost's frontend to talk to the content API
- this is so that portal and future features can access our APIs through the correct mechanism of an API key

### DIFF
--- a/core/server/data/migrations/versions/5.0/2022-05-06-13-22-add-frontend-integration.js
+++ b/core/server/data/migrations/versions/5.0/2022-05-06-13-22-add-frontend-integration.js
@@ -1,0 +1,68 @@
+const logging = require('@tryghost/logging');
+const ObjectID = require('bson-objectid');
+const security = require('@tryghost/security');
+const {createTransactionalMigration} = require('../../utils');
+
+const frontendIntegration = {
+    slug: 'ghost-internal-frontend',
+    name: 'Ghost Internal Frontend',
+    description: 'Internal frontend integration',
+    type: 'internal'
+};
+
+const addInternalIntegrationAndKey = async (knex, integration) => {
+    const message = `Adding ${integration.name} integration`;
+
+    const existing = await knex('integrations').select('id').where('slug', integration.slug).first();
+
+    if (existing && existing.id) {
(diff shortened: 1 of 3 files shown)


matthanley (approved):

core/server/data/migrations/versions/5.0/2022-05-06-13-22-add-frontend-integration.js
```diff
@@ -0,0 +1,70 @@
...
+
+    const contentKey = {
+        id: ObjectID().toHexString(),
+        type: 'content',
+        secret: security.secret.create('content'),
+        role_id: null,
+        integration_id: integration.id,
+        created_at: now,
+        created_by: 1,
+        updated_at: now,
```
daniellockyer, line 38: Is there a reason you set `updated_at/by` here but not for the integration above?

core/server/data/migrations/versions/5.0/2022-05-06-13-22-add-frontend-integration.js
```diff
@@ -0,0 +1,70 @@
...
+
+    if (existing && existing.id) {
+        logging.warn(`Skipping ${message}`);
+        return;
+    }
+
+    const now = knex.raw('CURRENT_TIMESTAMP');
+    integration.id = ObjectID().toHexString();
+    integration.created_at = now;
+    integration.created_by = 1,
```
daniellockyer, line 26: `s/,/;/g`

core/server/data/migrations/versions/5.0/2022-05-06-13-22-add-frontend-integration.js
ErisDS: copy pasta from another migration 🙈 fixed

core/server/data/migrations/versions/5.0/2022-05-06-13-22-add-frontend-integration.js
ErisDS: 👍

daniellockyer (approved):

PR #14759, opened 2022-05-10
Readded notEnoughPermission message in post model

SimonBackx:
no issue

The message was removed somewhere, but the git history seems to be modified so I couldn't find the reference to when and why it was removed (the message seems to be added and removed in the same commit).

It is still used: https://github.com/TryGhost/Ghost/blob/c6b26ba7421da68591d758b10a172908b6b01945/core/server/models/post.js#L1262-L1264

Added in: https://github.com/tryghost/Ghost/commit/e84916747248720a5726b627412f9a8ac5e56935
No reference to where it was removed. But after the same commit above, it wasn't present any longer.


matthanley (approved):

matthanley:
@SimonBackx for future reference, small cleanup commits like this can go straight into main 👍🏼

Issue #14344, opened 2022-03-20
Issue with address ‚localhost' for Product and Video card instead of __GHOST_URL__


fueko:
### Issue Summary

Product and Video card thumbnails have the address `http://localhost:2368` when exported, which causes thumbnail location issues when importing json file.

![Zrzut ekranu 2022-03-20 o 20 36 42](https://user-images.githubusercontent.com/23366922/159179658-36e0cb82-485d-41bf-b842-13fdb8341540.png)

### Steps to Reproduce

1. Add thumbnail to Product card and export json file from Labs settings.
2. Check the thumbnail address.

### Ghost Version

4.40.0

### Node.js Version

v16.13.0

### How did you install Ghost?

localhost / macOS

### Database type

SQLite3

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

ErisDS:
@naz could you take a little look at this one please?

naz:
Hey @fueko :wave:  Thanks for the report. I've tried reproducing it on the latest 4.41.0 and wasn't able to. The export file seems to have correctly put `__GHOST_URL__` in place of the url's. Example from an export of mobiledoc field for the post I did:
```
[...]

PR #14763, opened 2022-05-10
Regenerated HTML for posts containing a product or video card

daniellockyer:
refs https://github.com/TryGhost/Ghost/issues/14344

- this migration will regenerate the HTML for posts that contain a
  product or video card as we experiencing a bug in the card generation
- this migration is mostly ripped from https://github.com/TryGhost/Ghost/blob/c0d82122b0324f5f4b6bae078174d966b29c92fd/core/server/data/migrations/versions/4.0/23-regenerate-posts-html.js

### DIFF
--- a/core/server/data/migrations/versions/5.0/2022-05-21-00-00-regenerate-posts-html.js
+++ b/core/server/data/migrations/versions/5.0/2022-05-21-00-00-regenerate-posts-html.js
@@ -1,0 +1,70 @@
+const logging = require('@tryghost/logging');
+const {createIrreversibleMigration} = require('../../utils');
+const mobiledocLib = require('../../../../lib/mobiledoc');
+const htmlToPlaintext = require('../../../../../shared/html-to-plaintext');
+
+module.exports = createIrreversibleMigration(async (knex) => {
+    logging.info(`Starting regeneration of posts HTML`);
+
+    await knex.transaction(async (trx) => {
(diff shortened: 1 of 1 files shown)


core/server/data/migrations/versions/5.0/2022-05-21-00-00-regenerate-posts-html.js
```diff
@@ -0,0 +1,70 @@
...
+    logging.info(`Starting regeneration of posts HTML`);
+
+    await knex.transaction(async (trx) => {
+        // get list of posts ids, use .forUpdate to lock rows until the transaction is finished
+        const postIdRows = await knex('posts')
+            .transacting(trx)
+            .forUpdate()
+            .select('id')
+            .where('mobiledoc', 'like', '%["video%')
+            .orWhere('mobiledoc', 'like', '%["product%');
```
kevinansfield, line 16: Was the performance of this query acceptable considering it's not indexed in any way?

daniellockyer: I tested it on a DB with ~71k posts and it took 4.65s, returning 7 matching posts. I think this is good enough for me?

kevinansfield: 👍 cool

kevinansfield (approved):

matthanley (approved):

PR #14765, opened 2022-05-10
Added validation to posts status column

daniellockyer:

- this commit adds `isIn` validation to the `status` column on the
  `posts` table, with values pulled from the model: https://github.com/TryGhost/Ghost/blob/c0d82122b0324f5f4b6bae078174d966b29c92fd/core/server/models/post.js#L28
- this also adds a migration to update any posts with invalid statuses
  to `draft`

### DIFF
--- a/core/server/data/migrations/versions/5.0/2022-05-10-14-57-cleanup-invalid-posts-status.js
+++ b/core/server/data/migrations/versions/5.0/2022-05-10-14-57-cleanup-invalid-posts-status.js
@@ -1,0 +1,15 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const affectedRows = await knex('posts')
+            .update('status', 'draft')
+            .whereNotIn('status', ['published', 'draft', 'scheduled', 'sent']);
+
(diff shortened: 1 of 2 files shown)


core/server/data/migrations/versions/5.0/2022-05-10-14-57-cleanup-invalid-posts-status.js
```diff
@@ -0,0 +1,15 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const affectedRows = await knex('posts')
```
kevinansfield, line 6: Should we have a log message before the migration starts so it's clear it failed on this migration if there's an error?

daniellockyer: Done 👍🏻 

I'll also update my other migration that's merged with the same: https://github.com/TryGhost/Ghost/blob/main/core/server/data/migrations/versions/5.0/2022-05-09-14-17-cleanup-invalid-users-status.js

kevinansfield (approved):

matthanley (approved):

PR #14771, opened 2022-05-10
Fixed incorrect usage of limit config causing Ghost not to boot

aileen:
no issue

- When applying an incorrect limits config, or missing expected values, Ghost would not boot as the errors would interrupt this process, which should not happen
- This commit catches the error thrown by the limit-service on boot sequence and transforms it into a warning if it's an `IncorrectUsageError`. Other errors are handled as before
- Added a test for the limit-service service


ErisDS (approved):
This looks great 👍

PR #14778, opened 2022-05-11
Added newsletters endpoint to Content API

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1599


core/server/models/newsletter.js
```diff
@@ -38,6 +38,11 @@ const Newsletter = ghostBookshelf.Model.extend({
         return this.hasMany('Post');
     },
 
+    // Force active newsletters for content API
+    enforcedFilters: function enforcedFilters(options) {
```
ErisDS, line 42: ❤️

core/server/api/canary/newsletters-public.js
```diff
@@ -0,0 +1,29 @@
+const models = require('../../models');
+
+const allowedIncludes = ['count.posts', 'count.members'];
```
ErisDS, line 3: Do we want to expose these? I guess probably 🤔

test/e2e-api/content/newsletters.test.js
```diff
@@ -0,0 +1,43 @@
+const {agentProvider, fixtureManager, matchers} = require('../../utils/e2e-framework');
+
+const newsletterSnapshot = {
+    id: matchers.anyObjectId,
+    uuid: matchers.anyUuid,
+    name: matchers.anyString,
```
ErisDS, line 6: You only need to add matchers for data that changes. So id, uuid, and dates.

ErisDS (changes requested):

core/server/api/canary/newsletters-public.js
rishabhgrg: 🤔 I think we should probably only expose them once we have a usecase for them ? Though in that case we could also remove them from allowed includes ?

test/e2e-api/content/newsletters.test.js
rishabhgrg: Updated, thanks.

core/server/api/canary/newsletters-public.js
rishabhgrg: Removed the `includes` for now till we have a usecase for them again.

ErisDS (approved):
This is looking great now, and a nice example of how easy it is to add endpoints now!

core/server/models/newsletter.js
```diff
@@ -38,6 +38,11 @@ const Newsletter = ghostBookshelf.Model.extend({
...
+        return options.context && options.context.public ? 'status:active' : null;
```
matthanley, line 43: Question: does this evaluate as
`(options.context && options.context.public) ? 'status:active' : null;`
or
`options.context && (options.context.public ? 'status:active' : null);`
?

matthanley (approved):

core/server/models/newsletter.js
rishabhgrg: Since`&&` has a higher precedence, think it would be `(options.context && options.context.public) ? 'status:active' : null;` , but I'll update this either ways to remove any ambiguity :)

PR #14779, opened 2022-05-11
✨ Non-blocking bulk image downloader (API)

ronaldlangeveld:
👋 Hi!

This is my second PR with the same concept, but an improvement, based on @ErisDS 's feedback https://github.com/TryGhost/Ghost/pull/14659

My curiosity couldn't stop me from not attempting another version of it.

This time, it happens in the background in a non-blocking way. 

How it works:
It's added the very bottom Alpha Features in the Labs settings on Ghost Admin.

User initiates an export by clicking the "Create new export" button.
Screenshot 2022-05-11 at 08 07 12

It would then trigger the background task and user is asked to check back later.

Screenshot 2022-05-11 at 08 04 44

Since it's a background task and non-blocking the user can carry on doing other tasks on the platform, or even close the browser.  It interacts with a new table in the database to track the status. 

Once it's done, the download button will be visible.
Screenshot 2022-05-11 at 08 04 34

The images.zip inside the backups folder inside content gets overwritten every time as it's not something where "rollbacks" would be appropriate, so it doesn't make sense to need to have more than one export file stored.

The main goal of this is to compliment the existing JSON exporter, particularly when migrating self-hosted environments and perhaps when creating a custom theme for an existing publication, to be able to easily populate posts / pages with images locally. 

Since the posts/pages reference images directly something like `__GHOST_URL__/content/images/2022/05/2022-05-03-14.54.51.jpg`, the images.zip file just need to be extracted inside the content/images folder.

I tested it locally with about 50gb images from my external hard disk dumped into the content/images folder and it zipped without issues.

Wrote some basic frontend for the prototype, under this PR: https://github.com/TryGhost/Admin/pull/2375

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,6 +107,7 @@ projectFilesBackup
 /content/media/**/*
 /content/files/**/*
 /content/public/*
+/content/backups/*
 /content/adapters/storage/**/*
 /content/adapters/scheduling/**/*
 !/content/themes/casper
--- a/config.development.json
+++ b/config.development.json
@@ -1,3 +1,0 @@
-{
-    "enableDeveloperExperiments": true
-}
--- a/content/logs/README.md
+++ b/content/logs/README.md
@@ -1,3 +1,0 @@
-# Content / Logs
-
-This is the default log file location when Ghost runs in Production.
--- a/core/server/api/canary/backups.js
+++ b/core/server/api/canary/backups.js
@@ -1,10 +1,11 @@
+const backupsServices = require('../../services/backups');
+
 module.exports = {
     docName: 'backups',
     download: {
-        statusCode: 200,
         permissions: false,
-        query(frame) {
-            return 
+        query() {
+            return backupsServices.api.exporter();
         }
     }
 };
--- a/core/server/api/canary/backups.js
+++ b/core/server/api/canary/backups.js
@@ -1,4 +1,3 @@
-
 module.exports = {
     docName: 'backups',
     download: {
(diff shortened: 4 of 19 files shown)


ErisDS:
Hey @ronaldlangeveld thanks for taking a second stab at this ❤️ . I just wanted to let you know we're in a _huge_ push to get 5.0 ready 😬 and so it might take a little while to review this, but I will once the crazy stops.

ronaldlangeveld:
@ErisDS Not a problem at all. Happy to contribute 😄 .
All the best with 5.0!

ErisDS:
Hey @ronaldlangeveld 👋 
As part of shipping 5.0 we ended up adding a `ghost backup` command to Ghost CLI that does a lot of what this was intended to achieve: https://ghost.org/docs/ghost-cli/#ghost-backup

That definitely doesn't negate the usefulness of being able to download all assets direct from the UI though.

In terms of this PR, I have a couple of bits of feedback:

You've hit on something we've also hit several times internally, which is the need for a one-time-job system layered on top of our existing [jobservice](https://github.com/TryGhost/Ghost/tree/main/core/server/services/jobs), that is backed by a jobs table. I would lean towards a more abstracted implementation for this rather than introducing a table per job-type. It reduces overhead in a lot of ways, as every new table adds overhead to initialising Ghost, and that affects the time it takes to run e2e/integration tests, and that in turn affects how fast we can develop new stuff.

Short term without that piece of architecture, I'd still use the [jobservice](https://github.com/TryGhost/Ghost/tree/main/core/server/services/jobs) to schedule a job. It could at least keep whether or not there's an active backup running in memory, to prevent thrashing?

I also still think we should send an email when it's done 🤔 people will forget what they were doing otherwise I think.

PR #14780, opened 2022-05-11
Improved newsletter limit checking for archived newsletters

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1583

- Check limits when unarchiving newsletters
- Added tests for more scenarios
- When adding newsletters, the transaction is never used outside the newsletter service. So it was safe to just add the check before we start the transaction. The tests should catch this if this ever changes.
- `limit-service` was bumped to add transactions support

### DIFF
--- a/core/server/services/newsletters/service.js
+++ b/core/server/services/newsletters/service.js
@@ -112,7 +112,8 @@ class NewslettersService {
      * @public
      * @param {object} attrs model properties
      * @param {Object} [options] options
-     * @param {Object} [options] options.transacting Don't use this option outside of the service unless you also update the limit checking.
+     * @param {boolean} [options.opt_in_existing] Opt in existing members
+     * @param {Object} [options.transacting]
      * @returns {Promise<{object}>} Newsetter Model with verification metadata
      */
     async add(attrs, options = {}) {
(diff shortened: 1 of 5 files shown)


core/server/services/newsletters/service.js
```diff
@@ -112,10 +112,18 @@ class NewslettersService {
...
      * @param {object} attrs model properties
      * @param {Object} [options] options
-     * @param {Object} [options] options.transacting
+     * @param {Object} [options] options.transacting Don't use this option outside of the service unless you also update the limit checking.
      * @returns {Promise<{object}>} Newsetter Model with verification metadata
      */
     async add(attrs, options = {}) {
+        // We need to make sure we always check the limits outside of transactions
+        // Or we'll get a deadlock in SQLite
+        if (!options.transacting) {
```
allouis, line 121: IMO this is dangerous - we should have ways to skip verification.

For me it would be preferable to remove the `options.transaction` option, and usage of it.


We don't currently use it, and if we do in the future, I would much prefer it to not work/break than to bypass our limits!

SimonBackx: You are right! I'm going to see how I can make the option private or if it would be feasible to get the transaction over to the limit-service

core/server/services/newsletters/service.js
```diff
@@ -185,6 +187,10 @@ class NewslettersService {
 
         const {cleanedAttrs, emailsToVerify} = await this.prepAttrsForEmailVerification(attrs, originalNewsletter);
 
+        if (originalNewsletter.status !== 'active' && cleanedAttrs.status === 'active') {
+            await this.limitService.errorIfWouldGoOverLimit('newsletters');
```
matthanley, line 191: I think this is missing passing `options.transacting` if it's set

matthanley (changes requested):

core/server/services/newsletters/service.js
SimonBackx: True, I didn't add it here because it wasn't used. But it is in fact. I'll also update the `findOne` call to use the transaction

matthanley (approved):

allouis (approved):

PR #14787, opened 2022-05-11
Fixed filter mapping from subscribed to newsletters for bulk operations

matthanley:
refs https://github.com/TryGhost/Team/issues/1606
requires https://github.com/TryGhost/Members/pull/398/files


allouis (approved):
It's worth noting that you can add an `all` serialiser which might help in future

PR #14789, opened 2022-05-11
Updated theme card asset default setting

matthanley:
refs https://github.com/TryGhost/Team/issues/1611

- For Ghost 5.0 card assets will be included by default, including bookmark and gallery cards


ErisDS (approved):
LGTM! Can you update the breaking changes issue with this?

PR #14790, opened 2022-05-11
Allowed tiers include and column for members 

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1145

- allows members endpoint to accept `?include=tiers`
- allows members endpoint to return `tiers` data in response
- allows members endpoint to accept `tiers` data in payload


core/server/api/canary/utils/serializers/input/members.js
```diff
@@ -64,6 +77,11 @@ module.exports = {
                 }
             });
         }
+
+        if (frame.data.members[0].tiers) {
+            frame.data.members[0].products = frame.data.members[0].tiers;
```
matthanley, line 82: Does this mapping need to be applied to the edit endpoint as well?

rishabhgrg: The `edit` serializer defers the logic to `add` [here](https://github.com/TryGhost/Ghost/blob/2e8c66d93c4abdba485fa280790312f1ee82f142/core/server/api/canary/utils/serializers/input/members.js#L72), so it automatically gets applied to it :)

matthanley (approved):

PR #14791, opened 2022-05-11
Renamed lang and session_secret default settings

ErisDS:
This PR contains two commits:

1. is the migration and related data changes to rename `lang` -> `locale` and `session_secret` -> `admin_session_secret`.
2. all the related code changes to make this work

Note: I already ripped out all the backwards compat code around locale [in this commit](https://github.com/TryGhost/Ghost/commit/196b8ad0daca996d6f9b5eb6802bf6a65d636f87) because we don't need backwards compat and the changeset was getting hairy.


daniellockyer:
@ErisDS I don't think this is working as expected 🤔 

```
1 ~/code/product/Ghost λ mysql -u root -p ghost -e "select * from settings where \`key\` in ('lang', 'locale', 'session_secret', 'admin_session_secret');"
Enter password:
+--------------------------+-------+----------------+------------------------------------------------------------------+--------+-------+---------------------+------------+---------------------+------------+
| id                       | group | key            | value                                                            | type   | flags | created_at          | created_by | updated_at          | updated_by |
+--------------------------+-------+----------------+------------------------------------------------------------------+--------+-------+---------------------+------------+---------------------+------------+
| 627cd5bb9e57b862314af2c9 | site  | lang           | en                                                               | string | NULL  | 2022-05-12 10:39:07 | 1          | 2022-05-12 10:39:07 | 1          |
| 627cd5bb9e57b862314af2ba | core  | session_secret | 508689bb1f910a00faf6eab32771e508dec3c94131f9afdb86aa8bb76c63f8ba | string | NULL  | 2022-05-12 10:39:07 | 1          | 2022-05-12 10:39:07 | 1          |
+--------------------------+-------+----------------+------------------------------------------------------------------+--------+-------+---------------------+------------+---------------------+------------+

1 ~/code/product/Ghost λ knex-migrator migrate --v 5.0 --force
[2022-05-12 09:40:32] INFO Creating database backup
[2022-05-12 09:40:32] INFO Database backup written to: /Users/neo/code/product/Ghost/content/data/ghost.ghost.2022-05-12-09-40-32.json
[2022-05-12 09:40:32] INFO Running migrations.
[2022-05-12 09:40:32] INFO Updating locale with value from lang
[2022-05-12 09:40:32] INFO Deleting lang
[2022-05-12 09:40:32] INFO Updating admin_session_secret with value from session_secret
[2022-05-12 09:40:32] INFO Deleting session_secret
[2022-05-12 10:40:32] INFO Finished database migration!

0 ~/code/product/Ghost λ mysql -u root -p ghost -e "select * from settings where \`key\` in ('lang', 'locale', 'session_secret', 'admin_session_secret');"
<empty>
```

core/server/data/migrations/versions/5.0/2022-05-11-13-12-settings-rename.js
```diff
@@ -0,0 +1,43 @@
...
+                logging.warn(`Could not find setting ${renameMapping.from}, not updating ${renameMapping.to} value`);
+                continue;
+            }
+
+            const updatedValue = renameMapping.getToValue ? renameMapping.getToValue(oldSetting.value) : oldSetting.value;
+
+            logging.info(`Updating ${renameMapping.to} with value from ${renameMapping.from}`);
+            await knex('settings')
+                .where('key', renameMapping.to)
+                .update('value', updatedValue);
```
matthanley, line 31: I think this should be an insert, because afaict we'll never have a row with `key: renameMapping.to` at this point so we'd just be deleting settings 😬

core/server/data/migrations/versions/5.0/2022-05-11-13-12-settings-rename.js
```diff
@@ -0,0 +1,43 @@
...
+                .where('key', renameMapping.from)
+                .select('value')
+                .first();
+
+            if (!oldSetting) {
+                logging.warn(`Could not find setting ${renameMapping.from}, not updating ${renameMapping.to} value`);
+                continue;
+            }
+
+            const updatedValue = renameMapping.getToValue ? renameMapping.getToValue(oldSetting.value) : oldSetting.value;
```
matthanley, line 26: Given this isn't defined on any of the maps, and I don't think we have any requirement to map values as well as keys I think we can just do an `update settings set value = renameMapping.to where value = renameMapping.from` instead of doing an insert/delete?

ErisDS:
I have re-thought this migration from the ground up @daniellockyer and @matthanley and have tested all 3 cases now. 

To be clear

CASE 1: you ran `yarn dev` before `knex-migrator migrate` = default settings will have already populated the new values. Update values + delete.

CASE 2: you ran `knex-migrator migrate` first = you'll only have the old values. Update keys.

CASE 3: something went wrong and you have no matching settings. Do nothing, default settings will fix it.

This should probably be a utility!

daniellockyer (approved):

matthanley (approved):

PR #14798, opened 2022-05-12
Renamed newsletter_id and email_recipient_filter options

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1596

- Renamed `newsletter_id` to `newsletter` option, the `newsletter` option expects a slug instead of an id.
- Renamed `email_recipient_filter` to `email_segment` option
- Default `email_segment` to `all`. Ignored if no newsletter is set
- `email_segment` is ignored if no newsletter is set
- When reverting a post to a draft, both `newsletter` and `email_segment` are reset to their default values (null, all)
- Removed legacy mapping from old email_recipient_filter values 'paid' and 'free' (already a migration in place)
- Dropped legacy throwing errors when email_recipient_filter is paid or free in transformEmailRecipientFilter
- Reorganized transformEmailRecipientFilter parameters for the now required newsletter parameter
- Fixed an issue where the newsletter filter wasn't working because it wasn't in permittedoptions
- Fixed an issue where you could send to an archived newsletter
- Added an extra protection when scheduling to an active, and later archiving the newsletter
- Dropped support for `send_email_when_published` in API
- When importing posts we currently don't have a system in place to set the newsletter_id to map the `send_email_when_published` behaviour. Since this was already the case, I won't include a fix in this PR.
- Stripped `email_recipient_filter`/`email_segment` from Content API (https://ghost.slack.com/archives/C02G9E68C/p1652363211841359?thread_ts=1650623650.233229&cid=C02G9E68C)
- Updated `admin-api-schema` to 3.2.0, which includes the new email_segment property
- Contains a temporary fix for https://github.com/TryGhost/Team/issues/1626, where the `.related('newsletter').fetch` call fails when the newsletter relation is already loaded, because of the overridden `formatOnWrite` method.

Since the `email_recipient_filter` is no longer used without a newsletter, the `none` value is no longer used. A migration transforms all those values to `all`. This should be safe, because we only send an email now when newsletter_id is not null (scheduled posts should already have a newsletter_id, even if at the time of scheduling they didn't add the newsletter_id option, because at that time, we defaulted to the default newsletter).

Admin changes to make this work: https://github.com/TryGhost/Admin/pull/2380 (will get merged at the same time)


core/server/api/canary/utils/serializers/output/mappers/posts.js
```diff
@@ -57,8 +61,6 @@ module.exports = async (model, frame, options = {}) => {
         }
         date.forPost(jsonModel);
         gating.forPost(jsonModel, frame);
-
-        delete jsonModel.newsletter_id;
```
SimonBackx, line 61: Note that this is dropped because this is removed a couple of lines later in `clean.post(jsonModel, frame);` in both the content and admin api

core/server/services/posts/posts-service.js
```diff
@@ -17,69 +17,35 @@ class PostsService {
     }
 
     async editPost(frame) {
-        let model;
-
-        // Make sure the newsletter_id is matching an active newsletter
-        if (frame.options.newsletter_id) {
-            const newsletter = await this.models.Newsletter.findOne({id: frame.options.newsletter_id, filter: 'status:active'}, {transacting: frame.options.transacting});
```
matthanley, line 24: Did this ever work? 😬

matthanley (approved):

core/server/services/posts/posts-service.js
SimonBackx: No, we didn't have tests for this 😬 We even tested with archived newsletters without knowing 😅

core/server/models/post.js
```diff
@@ -675,28 +659,30 @@ Post = ghostBookshelf.Model.extend({
...
-            && this.hasChanged('status')
-            && (newStatus === 'published' || newStatus === 'scheduled')) {
-            this.set('email_recipient_filter', options.email_recipient_filter);
+            // If the `email_segment` isn't passed at the same time, reset it to be 100% sure that they can only be used together
+            this.set('email_recipient_filter', 'all');
+
+            // email_segment is read-only and should only be set using a query param when publishing/scheduling
+            // we can't set it if we don't pass newsletter
+            if (options.email_segment
+                && (options.email_segment !== 'none')) {
```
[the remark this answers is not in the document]
SimonBackx: I've added a migration to drop that value.

SimonBackx:
Note: tests fail again because `@tryghost/admin-api-schema` reverted a required change in 3.2.1

core/server/data/migrations/versions/5.0/2022-05-13-11-38-drop-none-email-recipient-filter.js
```diff
@@ -0,0 +1,25 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        // The 'none' value is no longer supported/used
+        logging.info(`Updating posts with email_recipient_filter 'none' to 'all'`);
+
+        const affectedRows = await knex('posts')
+            .update('email_recipient_filter', 'all')
```
matthanley, line 10: Is it worth also explicitly setting `newsletter_id:null` here?

core/server/services/mega/mega.js
```diff
@@ -213,10 +196,21 @@ const addEmail = async (postModel, options) => {
...
     const filterOptions = {...knexOptions, limit: 1};
 
+    // TODO: this is a hack for https://github.com/TryGhost/Team/issues/1626
+    delete postModel.relations.newsletter;
+
     const newsletter = await postModel.related('newsletter').fetch({require: true, ..._.pick(options, ['transacting'])});
 
+    if (newsletter.get('status') !== 'active') {
+        // A post might have been scheduled to an archived newsletter.
+        // Don't send it (people can't unsubscribe any longer).
```
matthanley, line 206: What happens to the post in this situation? I assume retrying the send would fail again, but is there a path to update the newsletter and resend?

core/server/services/mega/mega.js
```diff
@@ -423,11 +417,8 @@ async function getEmailMemberRows({emailModel, memberSegment, options}) {
     const knexOptions = _.pick(options, ['transacting', 'forUpdate']);
     const filterOptions = Object.assign({}, knexOptions);
 
-    let newsletter = null;
-    if (labsService.isSet('multipleNewsletters')) {
-        newsletter = await emailModel.related('newsletter').fetch(Object.assign({}, {require: false}, _.pick(options, ['transacting'])));
-    }
-    const recipientFilter = transformEmailRecipientFilter(emailModel.get('recipient_filter'), {errorProperty: 'recipient_filter'}, newsletter);
+    const newsletter = await emailModel.related('newsletter').fetch(Object.assign({}, {require: true}, _.pick(options, ['transacting'])));
```
matthanley, line 420: Does this need the same hack as above?

SimonBackx: Not that I noticed. Main problem is that I cannot reproduce it in tests 😬

SimonBackx: I'll add it to be sure + gonna continue to investigate the issue

core/server/services/mega/mega.js
SimonBackx: Well, no email will get attached to the post. So they can retry by reverting to draft and publishing again with a different newsletter

daniellockyer (approved):

core/server/services/mega/mega.js
```diff
@@ -53,8 +53,9 @@ const getReplyToAddress = (fromAddress, replyAddressOption) => {
  * @param {Object} options
  */
 const getEmailData = async (postModel, options) => {
-    let newsletter = await postModel.related('newsletter').fetch();
+    let newsletter = postModel.relations.newsletter ?? await postModel.related('newsletter').fetch();
     if (!newsletter) {
+        // The postModel doesn't have a newsletter in test emails
         newsletter = await models.Newsletter.getDefaultNewsletter();
```
SimonBackx, line 59: Permalinking this change for reference

core/server/services/mega/mega.js
```diff
@@ -53,8 +53,9 @@ const getReplyToAddress = (fromAddress, replyAddressOption) => {
  * @param {Object} options
  */
 const getEmailData = async (postModel, options) => {
-    let newsletter = await postModel.related('newsletter').fetch();
+    let newsletter = postModel.relations.newsletter ?? await postModel.related('newsletter').fetch();
     if (!newsletter) {
+        // The postModel doesn't have a newsletter in test emails
```
SimonBackx, line 58: Permalinking this change

PR #14801, opened 2022-05-12
Extended public settings to include portal settings

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1599

- adds `portal_*` and `firstpromoter_*` settings to public settings endpoint
- also adds `version` information


core/shared/settings-cache/public.js
```diff
@@ -29,5 +29,13 @@ module.exports = {
     members_support_address: 'members_support_address',
     members_enabled: 'members_enabled',
     members_invite_only: 'members_invite_only',
-    paid_members_enabled: 'paid_members_enabled'
+    paid_members_enabled: 'paid_members_enabled',
+    firstpromoter: 'firstpromoter',
+    firstpromoter_id: 'firstpromoter_id',
```
ErisDS, line 34: Do we really need both a bool and an ID? Can we just have one?

rishabhgrg: We can do with just `firstpromoter_id`, but we'll need to add a tiny logic in front somewhere that sets it to `null` when `firstpromoter` is turned off, as right now both are set independently, so its possible to have `firstpromoter` set to `false` but still an old id to present atm. 😅

rishabhgrg: Oh I could probably just update this in the new Calculated fields that we have now 💡

rishabhgrg: Updated with new CalculatedSetting `firstpromoter_account` that is `null` if integration is off or the value otherwise. ✅

core/server/api/canary/utils/serializers/output/mappers/settings.js
```diff
@@ -12,7 +12,7 @@ module.exports = (attrs, frame) => {
     //      `forSettings` step. This logic can be rewritten once we get rid of deprecated
     //      fields completely.
     if (_.isArray(attrs)) {
-        const keysToFilter = [];
+        const keysToFilter = ['firstpromoter_id', 'firstpromoter'];
```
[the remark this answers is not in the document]
rishabhgrg: Updated.

ErisDS (approved):

PR #14802, opened 2022-05-12
Removed settings that are no longer in use

ErisDS:
refs: https://github.com/TryGhost/Team/issues/1625

Removes these settings:

```
const settingsToRemove = [
    'editor_is_launch_complete',
    'members_free_price_name',
    'members_free_price_description',
    'members_paid_signup_redirect',
    'members_free_signup_redirect',
    'members_from_address',
    'members_reply_address',
    'stripe_product_name',
    'newsletter_show_badge',
    'newsletter_header_image',
    'newsletter_show_header_icon',
    'newsletter_show_header_title',
    'newsletter_title_alignment',
    'newsletter_title_font_category',
    'newsletter_show_feature_image',
    'newsletter_body_font_category',
    'newsletter_footer_content',
];
```

✅ These settings are not used in the members repo


core/server/data/migrations/versions/5.0/2022-05-11-16-36-remove-unused-settings.js
```diff
@@ -0,0 +1,53 @@
...
+// Settings that cannot be removed, but should not be used anymore
+// 'members_monthly_price_id',
+// 'members_yearly_price_id',
+// 'portal_products',
+// 'stripe_plans'
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        // eslint-disable-next-line no-restricted-syntax
+        for (const removableSetting of settingsToRemove) {
```
matthanley, line 33: Is looping the settings to log the settings individually worth the performance impact vs `knex('settings').whereIn('key', settingsToRemove).del()`?

daniellockyer: If we wanted to reduce the number of queries, we could do a `select .. where .. in` with `settingsToRemove` to fetch the settings we have, log them out as required, and then do a `DELETE .. WHERE .. IN` to clean them up. 34 queries down to 2 queries 😬

daniellockyer: This should be resolved in https://github.com/TryGhost/Ghost/pull/14802/files#diff-b32355f6f69f9ea8d7c0c1e72a2a761c934a87ef914290cf4a90eda9cc6a9f90

matthanley (approved):

daniellockyer (approved):

PR #14819, opened 2022-05-13
Excerpt Improvements

ErisDS:
refs: https://github.com/TryGhost/Team/issues/1609

I'm changing the default output that goes into the `plaintext` field here, but I don't think it warrants regeneration.

I also have a further commit that I'm unsure of, which restricts the processing to 15 child nodes: https://github.com/ErisDS/Ghost/commit/335dd36df6385bb71cba9d831156b2b3a6b1028b This would mean the plaintext format is no longer the full document, but the first 15 children, which is quite a lot of it. 

I am not sure that anyone actually has a usecase for the plaintext field beyond the excerpt, so I'm tempted to push it and see if it impacts anyone. If it doesn't, we know we can drop the plaintext field from the API (as long as excerpts still work).


ErisDS:
closing in favour of https://github.com/TryGhost/Ghost/pull/14823

PR #14821, opened 2022-05-13
Added CLI command structure

matthanley:
Added CLI commands for REPL and timetravel functionality

- Added TimeTravel command for updating test data with a date offset
- Added REPL command for access to models and knex in development
- Added pattern for creating new CLI commands, including
  - User input
  - Output
  - Validation of `NODE_ENV`
- TimeTravel command is in the main Ghost repo because it requires the schema

Reference material: https://laravel.com/docs/9.x/artisan


package.json
```diff
@@ -194,6 +194,7 @@
     "@playwright/test": "1.22.0",
     "@tryghost/express-test": "0.10.0",
     "c8": "7.11.2",
+    "cli-progress": "^3.11.0",
```
daniellockyer, line 197: Can you pin these dependencies?

core/cli/timetravel.js
```diff
@@ -0,0 +1,62 @@
...
+    column: 'created_at'
+};
+
+module.exports = class TimeTravel extends Command {
+    async handle() {
+        const datumPoint = await knex(datum.table)
+            .max(datum.column, {as: datum.column})
+            .first();
+
+        const dateOffset = Math.floor(
```
daniellockyer, line 19: We should probably handle `members_login_events` being empty, otherwise I get:

```
...
[2022-05-16 17:32:48] INFO Timetravel will use an offset of +NaN days
...
```

core/cli/repl.js
```diff
@@ -0,0 +1,21 @@
...
+        context.models = models;
+        context.m = models;
+        context.knex = knex;
+        context.k = knex;
+    }
+
+    async handle() {
+        const repl = require('repl');
+        const cli = repl.start('> ');
+        this.initializeContext(cli.context);
```
daniellockyer, line 19: Can we add maybe a one-liner detailing the variables available?

core/cli/command.js
```diff
@@ -0,0 +1,122 @@
+const cli = require('@tryghost/pretty-cli');
```
daniellockyer, line 1: I think this dependency is hoisted from gscan's dependencies and we need to add it to package.json here too

core/cli/timetravel.js
```diff
@@ -0,0 +1,87 @@
+const Command = require('./command');
+const chalk = require('chalk');
```
daniellockyer, line 2: Same with chalk 🤔

daniellockyer (approved):

PR #14833, opened 2022-05-16
Removed oauth experiment

ErisDS:
- this didn't work the way we wanted to
- removing this will free up the namespaces to start over


package.json
```diff
@@ -174,7 +174,6 @@
     "node-jose": "2.1.1",
     "oembed-parser": "1.4.9",
     "passport": "0.5.2",
```
daniellockyer, line 176: I think you can nuke `passport` too

daniellockyer (approved):

PR #14857, opened 2022-05-18
Fixed Stripe Connect

allouis:
refs https://github.com/TryGhost/Team/issues/1637

The `stripe_connect_integration_token` setting is required for the Stripe
Connect flow to work. The existing `stripe_connect_*` settings are also not able
to be edited. So we've updated the allow list to remove those, and add the one
required for the Stripe Connect flow.


test/e2e-api/admin/settings.test.js
```diff
@@ -207,6 +207,38 @@ describe('Settings API', function () {
                     });
                 });
         });
+
+        it('Can attempt to connect to stripe', async function () {
```
allouis, line 211: We can maybe improve this test by first doing a request to connect to stripe, I'm not sure how tricky it will be but I can take a look if desired

allouis: It would be improved by not have an error response, and actually setting fake stripe keys in the DB

matthanley (approved):

PR #14858, opened 2022-05-18
Fixed incorrect member count on newsletter creation

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1634

- on newsletter creation, the newsletter model with counts was loaded before the members were opted in
- reloads newsletter model after members are opted-in to get the right count on creation

### DIFF
--- a/core/server/services/newsletters/service.js
+++ b/core/server/services/newsletters/service.js
@@ -161,14 +161,14 @@ class NewslettersService {
             // subscribe members that have an existing subscription to an active newsletter
             const memberIds = await this.MemberModel.fetchAllSubscribed(_.pick(options, 'transacting'));
 
-            newsletter.meta = newsletter.meta || {};
-            newsletter.meta.opted_in_member_count = memberIds.length;
-
             if (memberIds.length) {
                 debug(`Found ${memberIds.length} members to subscribe`);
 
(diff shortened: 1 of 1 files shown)


core/server/services/newsletters/service.js
```diff
@@ -161,14 +161,14 @@ class NewslettersService {
...
             const memberIds = await this.MemberModel.fetchAllSubscribed(_.pick(options, 'transacting'));
 
-            newsletter.meta = newsletter.meta || {};
-            newsletter.meta.opted_in_member_count = memberIds.length;
-
             if (memberIds.length) {
                 debug(`Found ${memberIds.length} members to subscribe`);
 
                 await newsletter.subscribeMembersById(memberIds, options);
+                newsletter = await this.NewsletterModel.findOne({id: newsletter.id}, {...options, require: true});
```
matthanley, line 168: We now reload the newsletter model twice ([here](https://github.com/TryGhost/Ghost/pull/14858/files#diff-d61df6455ea81ae89293ed9306ecc5f763bd281ae0e4cc4c588bd3552ab35400R155)) - can we avoid that?

matthanley (commented):
This also needs a test added

rishabhgrg:
Closed in favor of https://github.com/TryGhost/Admin/pull/2402

PR #14859, opened 2022-05-18
Fixed members list filtering on tiers

rshbhgrg:
- filtering member list on `tier:tier-slug` was failing as it only worked with `member.product` column
- member table needed to expand the filter on tier instead of product with correct value


matthanley (commented):
Nice catch! Needs a test added, too 😄

matthanley (approved):

PR #14865, opened 2022-05-19
Improved member export performance

SimonBackx:


SimonBackx:
We could also use this snippet from the members-api package's bulkEdit method (maybe we can add the other selects in this query so we only need one query)
```
const memberRows = await this._Member.getFilteredCollectionQuery(filterOptions)
    .select('members.id')
    .distinct();

const memberIds = memberRows.map(row => row.id);
```

matthanley:
Superseded by https://github.com/TryGhost/Ghost/pull/14878

PR #14870, opened 2022-05-20
Added bulkDestroy column option

SimonBackx:
refs https://ghost.slack.com/archives/C02G9E68C/p1652980792270029

- When bulk unsubscribing members, the number of deleted newsletter relations are returned instead of the number of members with newsletters that were cleared
- This adds options to the bulk destroy methods.
- You can pass the column on which you want to delete rows in a bulk destroy operation via the `column` option.

This will get used in https://github.com/TryGhost/Members/pull/400

### DIFF
--- a/core/server/models/base/plugins/bulk-operations.js
+++ b/core/server/models/base/plugins/bulk-operations.js
@@ -52,9 +52,9 @@ async function editMultiple(knex, table, chunk, options) {
     await knex(table).whereIn('id', chunk).update(options.data);
 }
 
-async function delSingle(knex, table, id) {
+async function delSingle(knex, table, id, options) {
     try {
-        await knex(table).where('id', id).del();
+        await knex(table).where(options.column ?? 'id', id).del();
     } catch (err) {
         const importError = new errors.DataImportError({
             message: `Failed to remove entry from ${table}`,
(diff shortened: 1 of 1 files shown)


core/server/models/base/plugins/bulk-operations.js
```diff
@@ -52,9 +52,9 @@ async function editMultiple(knex, table, chunk, options) {
     await knex(table).whereIn('id', chunk).update(options.data);
 }
 
-async function delSingle(knex, table, id) {
+async function delSingle(knex, table, id, options) {
```
matthanley, line 55: should this be `options = {}`?

core/server/models/base/plugins/bulk-operations.js
```diff
@@ -66,8 +66,8 @@ async function delSingle(knex, table, id) {
     }
 }
 
-async function delMultiple(knex, table, chunk) {
-    await knex(table).whereIn('id', chunk).del();
+async function delMultiple(knex, table, chunk, options) {
```
matthanley, line 69: should this be `options = {}`?

core/server/models/base/plugins/bulk-operations.js
SimonBackx: I didn't add this because editSingle, editMultiple also don't have a default value. And since this is an internal method, we know that there will always be an options object passed to it because `bulkDestroy` already has the default.

PR #14872, opened 2022-05-20
Added products and Stripe data to exports

matthanley:
- The migration path from 4.x on SQLite to 5.0 on MySQL requires an export/import
- Exports don't include the Stripe info required to map members to tiers correctly on import. This change fixes that.


ErisDS (approved):

SimonBackx (approved):

ErisDS:
Closing in favour of https://github.com/TryGhost/Ghost/pull/14877 for clarity

PR #14876, opened 2022-05-20
[4.x]🐛 Fixed member exports timing out for large sites

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1641

### DIFF
--- a/core/server/services/members/exporter/query.js
+++ b/core/server/services/members/exporter/query.js
@@ -1,6 +1,7 @@
 const models = require('../../../models');
 const {knex} = require('../../../data/db');
 const moment = require('moment');
+const _ = require('lodash');
 
 module.exports = async function (options) {
     const hasFilter = options.limit !== 'all' || options.filter || options.search;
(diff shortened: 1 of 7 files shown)


core/server/services/members/exporter/query.js
```diff
@@ -0,0 +1,88 @@
...
+    if (hasFilter) {
+        // do a very minimal query, only to fetch the ids of the filtered values
+        // should be quite fast
+        options.withRelated = [];
+        options.columns = ['id'];
+
+        const page = await models.Member.findPage(options);
+        ids = page.data.map(d => d.id);
+
+        /*
```
ErisDS, line 18: Could you add a comment line explaining why this is commented out?

test/e2e-api/admin/members.test.js
```diff
@@ -1554,7 +1554,7 @@ describe('Members API', function () {
                 'content-disposition': anyString
             });
 
-        res.text.should.match(/id,email,name,note,subscribed_to_emails,complimentary_plan,stripe_customer_id,created_at,deleted_at/);
```
ErisDS, line 1557: Is there a reason to keep these 2 tests in `members.test.js` when we've got a suite in `members-exporter.test.js` now? Seems like they are duplicates?

ErisDS (approved):
I think this can be merged - I have minor cosmetic comments.

core/server/services/members/exporter/query.js
SimonBackx: Yeah, I tried last minute to improve the id fetching. But it didn't work reliably (I think filtering with search didn't work). I still think the commented code would have better performance if that is fixed because it 'escapes' from bookshelf. So I quickly commented it out before switching branches to work on the tests for https://github.com/TryGhost/Ghost/pull/14873 (switching between 4.x and 5.x takes a bit longer with the client/admin rename). Could have stashed the changes, but then I might forget about it 😬

test/e2e-api/admin/members.test.js
SimonBackx: I don't have any tests yet for the search option in the other file, so I think it is best to keep this one until we have a new one for search

SimonBackx: One of them is no search,  I just noticed. We can clean it up later :)

PR #14883, opened 2022-05-23
🐛 Fixed email analytics crashing when processing unsubscribe/complaint events

kevinansfield:
refs https://github.com/TryGhost/Team/issues/1649

- updated unsubscribe/complaint event processing methods
  - `member.subscribed` no longer exists, replaced that part of the query with a delete of member<->newsletter association rows from the `members_newsletters` pivot table
  - kept the member `updated_at` bump so we have some timestamp record of an update
  - added creation of `member_subscribe_event` records for the newsletter unsubscribes to keep stats and history in check

TODO:
- [x] create unsubscribe events
- [ ] tests

### DIFF
--- a/core/server/services/email-analytics/lib/event-processor.js
+++ b/core/server/services/email-analytics/lib/event-processor.js
@@ -131,42 +131,14 @@ class GhostEventProcessor extends EventProcessor {
     }
 
     async handleUnsubscribed(event) {
-        const memberId = await this.getMemberId(event);
-
-        if (!memberId) {
-            return false;
-        }
-
-        const subscribedNewsletterIds = await this.db.knex('members_newsletters')
-            .where('member_id', '=', memberId)
-            .pluck('newsletter_id');
-
-        await this.db.knex('members_newsletters')
-            .where('member_id', '=', memberId)
-            .del();
-
-        const nowUTC = moment.utc().toDate();
-        for (const newsletterId of subscribedNewsletterIds) {
-            await this.db.knex('members_subscribe_events').insert({
-                id: ObjectID().toHexString(),
-                member_id: memberId,
-                newsletter_id: newsletterId,
-                subscribed: false,
-                created_at: nowUTC,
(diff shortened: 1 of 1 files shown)


core/server/services/email-analytics/lib/event-processor.js
```diff
@@ -136,10 +137,30 @@ class GhostEventProcessor extends EventProcessor {
             return false;
         }
 
+        const subscribedNewsletterIds = (
```
allouis, line 140: Not a big deal - this can be slightly simplified to
```js
const subscribedNewsletterIds = await this.db.knex('members_newsletters')
    .where('member_id', '=', memberId)
    .pluck('newsletter_id');
```

core/server/services/email-analytics/lib/event-processor.js
```diff
@@ -136,10 +137,30 @@ class GhostEventProcessor extends EventProcessor {
...
+            .where('member_id', '=', memberId)
+            .del();
+
+        for (const newsletterId of subscribedNewsletterIds) {
+            await this.db.knex('members_subscribe_events').insert({
+                id: ObjectID().toHexString(),
+                member_id: memberId,
+                newsletter_id: newsletterId,
+                subscribed: false,
+                created_at: moment.utc().toDate(),
```
allouis, line 156: Will these all be exactly the same? If not, it would be nice to generate it once, and use the same timestamp for all events.

allouis (commented):
This looks good!

`handleUnsubscribed` and `handleComplained` are almost exactly the same, but we don't update the `updated_at` column for members in `handleUnsubscribed`.

Should we do that?

allouis:
@kevinansfield Ooop! Totally missed that sorry, the order of operations was different and I didn't check below!

kevinansfield:
Extracted the duplicate bodies into a utility function to avoid any confusion like that 🙂

matthanley (approved):

PR #14889, opened 2022-05-24
Return all settings in API when editing settings

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1650

Since we introduced calculated settings, we need to return all settings when editing a setting, because those calculated settings can change.

- When editing settings, return all the settings. Previously we didn't include the calculated settings (which caused issues with admin not having up to date values for those)
- Updated Stripe <head> script to be injected based on the calculated settings

Required for https://github.com/TryGhost/Admin/pull/2405

### DIFF
--- a/core/server/services/settings/settings-bread-service.js
+++ b/core/server/services/settings/settings-bread-service.js
@@ -152,9 +152,10 @@ class SettingsBREADService {
             });
         }
 
-        return this.SettingsModel.edit(filteredSettings, options).then((result) => {
-            return this._formatBrowse(_.keyBy(_.invokeMap(result, 'toJSON'), 'key'), options.context);
-        });
+        await this.SettingsModel.edit(filteredSettings, options);
+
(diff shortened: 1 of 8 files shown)


core/server/services/settings/settings-bread-service.js
```diff
@@ -152,6 +152,8 @@ class SettingsBREADService {
             });
         }
 
+        await this.SettingsModel.edit(filteredSettings, options);
```
matthanley, line 155: What's the purpose of this additional call as well as the one below?

SimonBackx: Oh snap, this is an error. I first had the browse one over here, changed it again and forgot to remove this one. Good catch!

SimonBackx: The reason for the extra browse call above is explained in the new comment line and PR description, being that we need to return all settings when editing a setting.

matthanley (approved):

PR #14915, opened 2022-05-25
Removed GA feature flags

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1616

- Removed all GA feature flags
- Removed `tweetGridCard` alpha flag
- Changes to `members-api` and `members-importer` packages: https://github.com/TryGhost/Members/compare/%40tryghost/members-api%408.1.1...%40tryghost/members-api%408.1.2


core/server/services/mega/mega.js
```diff
@@ -305,10 +304,7 @@ async function handleUnsubscribeRequest(req) {
     }
 
     try {
-        let memberData = {subscribed: false};
```
matthanley, line 308: I feel like this should be erroring - is `handleUnsubscribeRequest` still used in 5.0?

matthanley (approved):
LGTM, just not sure if we can completely remove the unsubscribe handler method now?

Issue #14882, opened 2022-05-23
Remove the bluebird dependency from everywhere 🔥 


ErisDS:
Now that there are native promises in Node.js _and_ with async/await, bluebird really is a blast from the past. However, it's embedded pretty hard across our dependencies and some of our code uses catch predicates, which means we have to be careful about the order in which we rip bluebird out, otherwise we could break our error handling in weird and unexpected ways.

Here's some notes on what we need to do, most importantly starting with finding everywhere we depend on catch predicates:

**Native:**
- Promise.all
- Promise.then
- Promise.reject
- Promise.resolve

**Bluebird**
- catch predicates -> replace with normal if statements
- Promise.promisify -> require('util').promisify
- Promise.props -> Promise.all [example](https://github.com/serverless/serverless/pull/8414/files)
- Promise.join -> Promise.all
- Promise.each -> Promise.all + array.map
- Promise.map -> Promise.all + array.map
- Promise.filter -> for and async/await
- Promise.mapSeries -> use our sequence util (which needs an implementation of reduce, but is our only usage of it)
- Promise.delay -> promisify getTimeout? Or get rid?

---

## Refactor Approach

This refactor needs to be done in two phases:

1. ✅ Remove all the catch predicates
 
 💡  The best way to find catch predicates is with this regex: `/\.catch\([^)]+,/`.

2. Remove bluebird from everywhere else:

- Ghost monorepo - bluebird is required 117 times 😱 
- [framework](https://github.com/TryGhost/framework) - bluebird required 5 times, but these are some of the most critical usages
- [SDK](https://github.com/TryGhost/SDK) - just one usage, easy win!
- [gscan](https://github.com/TryGhost/gscan) - 3 usages
- [knex-migrator](https://github.com/TryGhost/knex-migrator) - 3 usages

**Why did we have to do catch predicates first?**

Catch predicates result in error handling that assume the Promise they are handling is a Bluebird promise. If that isn't the case, Ghost will error, and quite possibly crash.

Our code coverage isn't quite good enough to be certain that all the unhappy paths are tested and we'll find these, and it's too much cognitive load for anyone to figure out if removing Bluebird is safe whilst there are still catch predicates in the codebase.

---

## How to work on this issue

The slow and steady approach to refactoring is the only thing that really works. Branches should be very short lived (<1 day if possible). Therefore, please do not attempt to do this entire refactor in a single PR. It is not possible for us to review!

[...]

PR #14939, opened 2022-05-28
Removed BB dep from url service

Dave3of5:
My first PR for the project. Started off with a small change for #14882. It's only a single file (two actually) change as was mentioned in the Issue.

I also fixed a typo in a function name.

### DIFF
--- a/ghost/core/core/server/services/url/Resources.js
+++ b/ghost/core/core/server/services/url/Resources.js
@@ -1,5 +1,4 @@
 const _ = require('lodash');
-const Promise = require('bluebird');
 const debug = require('@tryghost/debug')('services:url:resources');
 const Resource = require('./Resource');
 const config = require('../../../shared/config');
@@ -65,7 +64,7 @@ class Resources {
     /**
      * @description Helper function to initialize data fetching.
      */
-    fetchResources() {
+    async fetchResources() {
         const ops = [];
         debug('fetchResources');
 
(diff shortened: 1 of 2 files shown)


ErisDS:
Hey @Dave3of5, thank you for this PR, really appreciate you taking the time to help out. I've just updated the [original issue](https://github.com/TryGhost/Ghost/issues/14882) to be a bit clearer that we need to remove catch predicates first (and why). I'll hold onto this PR and merge it once that work is done.

Dave3of5:
@ErisDS Got it, I'll look at creating some PR's for removing the catch predicates then first.

ErisDS (approved):

ghost/core/core/server/services/url/Resources.js
```diff
@@ -95,28 +94,28 @@ class Resources {
      * 3 event listeners connected to "_onResourceUpdated" handler and it's 'tag.edited', 'tag.attached', 'tag.detached' events
      * 1 event listener connected to  "_onResourceRemoved" handler and it's 'tag.deleted' event
      */
-    initEvenListeners() {
+    async initEventListeners() {
```
daniellockyer, line 97: I think we can do without the `async` here, and the corresponding `await`s in `ghost/core/core/server/services/url/UrlService.js` - this function doesn't do anything asynchronous, or return a Promise 🙂 (an upcoming lint rule addition to Ghost would have caught this)

Dave3of5: Good spot fix in 5371dc4

daniellockyer (approved):
@ErisDS this should be fine now 🙂

PR #14947, opened 2022-05-31
fix  SQLITE_ERROR: too many terms in compound SELECT

gcxfd:
image

see the image above

### DIFF
--- a/core/server/services/mega/mega.js
+++ b/core/server/services/mega/mega.js
@@ -495,13 +495,15 @@ async function createEmailBatches({emailModel, memberRows, memberSegment, option
             });
         });
 
-        const insertQuery = db.knex('email_recipients').insert(recipientData);
+        for (let data of _.chunk(recipientData,499)) {
 
-        if (knexOptions.transacting) {
-            insertQuery.transacting(knexOptions.transacting);
+            const insertQuery = db.knex('email_recipients').insert(data);
+            if (knexOptions.transacting) {
+                insertQuery.transacting(knexOptions.transacting);
+            }
+            await insertQuery;
+
         }
-
-        await insertQuery;
 
         return batchModel.id;
     };


ErisDS:
Hi @gcxfd Thanks for taking the time to submit this fix. In its current form, there's no explanation for why there is a for loop and it's not limited to sqlite3, so I'm fairly certain someone else would just come along and remove the loop not realising why it's there.

Are you up for iterating on the fix? We have a [DatabaseInfo package](https://github.com/TryGhost/Ghost/blob/8dd009ffa03c5a0fac4d9f5ac614f54f3c27d372/core/server/services/members/service.js#L40) you can use to detect when the engine is sqlite3 🙂

markstos:
It's not clear this is the best location for a fix. The error is clearly about about a SELECT statement, but the patch changed a bulk INSERT into individual inserts.

Looking closely at the stack trace, what failed was in job-manager on line 110. That's where bulk emails are being prepared. That appears to be where the "compound SELECT" must be generated from. I started to trace it. Maybe it's part of the logic in here?

https://github.com/TryGhost/Ghost/blob/a30e42404bdf651d833e32b10027d4d43bca4bdb/core/server/services/mega/mega.js#L360

In any case, it would be useful which "compound SELECT" is being referred to.

> A compound SELECT statement is two or more SELECT statements connected by operators UNION, UNION ALL, EXCEPT, or INTERSECT. We call each individual SELECT statement within a compound SELECT a "term".

Once it's known where the code is making a "compound SELECT" with more than 500 terms, there may be a fix which benefits other databases as well, because that's a very complex SELECT statement!

-----

New take: I found this:

> older versions of sqlite (<3.7.11) don't allow multiple row insertion using INSERT INTO tablename (columnname) VALUES (value1),(value2), ... it is very easy to end up with a workaround with >500 UNION terms for a simple insertion like INSERT INTO tablename (columnname) SELECT value1 AS columnname UNION SELECT value2 UNION SELECT value3 ... UNION SELECT

So what version of SQLite is being used here?

gcxfd:
latest , see
https://stackoverflow.com/a/14263897

ErisDS:
> The Ghost package.json shows an optional dependency on sqlite3 with a version of 5.0.8. The TryGhost/node-sqlite project shows that it's using SQLite 3.38.4. I'm not sure which applies here, but both are greather than 3.17.11.

Ghost's dependency on sqlite3 is on the node package, which is a binary wrapper around sqlite3. The 5.0.8 version number references the node package, not sqlite3 directly.

The node sqlite3 package in turn should be using a very up-to-date version of sqlite - as you said 3.38.4.

> Another possibility is that the SQLite version being used is fine, but one of the translation layers in the stack is translating the multi-insert syntax into the "INSERT INTO...SELECT FROM" form that's compatible with versions before 3.7.11, which in turn causes this breakage.

I believe a further case in which the system has an earlier version of sqlite3 already installed, and that gets used instead. 

Still, 3.7.11 is a very old version of sqlite3, sqlite3 should only be used in development mode now, meaning it should be possible to update to a later version. 

Ideally, we want to keep away from having too many divergent code paths between sqlite3 and mysql. Sqlite3 really only exists for convenience of developing & testing out Ghost locally, and more for the use case of people building themes or other tools on top of Ghost, rather than those developing Ghost itself. It's not expected that every advanced feature of Ghost should function well on sqlite3 anymore, especially features that require scale like sending large volumes of emails :) 

I'm going to close this for now. If anyone discovers an upstream bug contributing to this, or has a cleaner fix they'd like to propose I'm all ears though.

PR #14953, opened 2022-06-01
🐛 Fixed Stripe Checkout for Members w/ existing subscriptions

allouis:
refs https://github.com/TryGhost/Team/issues/1526

This adds a check for existing subscriptions for a member associated with the
email addressed used for Stripe Checkout, if any are found the Checkout Session
creation fails and responds with a 403.

We've also updated the error handling for the create-stripe-checkout-session
endpoint so that it follows the existing Ghost API patterns.

### DIFF
--- a/test/e2e-api/members/create-stripe-checkout-session.test.js
+++ b/test/e2e-api/members/create-stripe-checkout-session.test.js
@@ -29,17 +29,17 @@ describe('Create Stripe Checkout Session', function () {
         const paidTier = tiers.find(tier => tier.type === 'paid');
 
         await membersAgent.post('/api/create-stripe-checkout-session/')
-              .body({
-                  customerEmail: 'paid@test.com',
-                  tierId: paidTier.id,
-                  cadence: 'month'
-              })
-              .expectStatus(403)
-              .matchBodySnapshot({
-                  errors: [{
-                      id: matchers.anyUuid,
-                      code: 'CANNOT_CHECKOUT_WITH_EXISTING_SUBSCRIPTION'
-                  }]
-              });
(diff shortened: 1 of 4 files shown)


test/e2e-api/members/create-stripe-checkout-session.test.js
```diff
@@ -0,0 +1,45 @@
+const {agentProvider, mockManager, fixtureManager, matchers} = require('../../utils/e2e-framework');
+
+let membersAgent, adminAgent, membersService;
+
+describe('Create Stripe Checkout Session', function () {
```
matthanley, line 5: Can we add a happy-path test here as well?

test/e2e-api/members/create-stripe-checkout-session.test.js
```diff
@@ -0,0 +1,45 @@
...
+                tierId: paidTier.id,
+                cadence: 'month'
+            })
+            .expectStatus(403)
+            .matchBodySnapshot({
+                errors: [{
+                    id: matchers.anyUuid,
+                    code: 'CANNOT_CHECKOUT_WITH_EXISTING_SUBSCRIPTION'
+                }]
+            });
```
matthanley, line 43: Needs a call to `matchHeaderSnapshot` here too

test/e2e-api/members/create-stripe-checkout-session.test.js
allouis: I'll take a look at that now - we'll have to mock the Stripe API but should be easy enough

matthanley (approved):

PR #14962, opened 2022-06-03
✨ Added format option to `img-url` helper

### LINKED ISSUE #14323: Resize unsplash image width query to reduce download for smaller devices
### Issue Summary

This issue is referencing the wrongfully closed issue #12245 that still persists. As the Unsplash feature is integrated by default, one would expect that the images are served responsively. The Unsplash API supports such queries, but Ghost returns the images independent of device screen size.

Given that images that are too big are impacting UX as well as SEO, this is a critical feature for users that use the Unsplash feature.

Current workaround is to download the image and upload it on ghost.

### Steps to Reproduce

1. Create article
2. Use any Unsplash image as feature image
3. Load page through mobile device or dev tools of your browser
4. Inspect the served image (should have a width of 2000)

### Ghost Version

4.39.

### Node.js Version

16.14.0

Docker Container

### Database type

MySQL 8

### Browser & OS version

_No response_

_No response_

### Code of Conduct

SimonBackx:
fixes https://github.com/TryGhost/Ghost/issues/14323

- Fixed support for resizing images from Unsplash using the `img-url` helper (previously the size property was ignored for images from Unsplash)
- Added support for `avif` file formats (supported by sharp out of the box)
- Added support for setting the format of images, with a new  `format` option:

E.g. to convert an image to webp (only works in combination with size for now, except for Unsplash where you can use it without size):
```
{{img_url @site.cover_image size="s" format="webp"}}
```

This can help improve the performance of a theme, by serving assets in `<picture>` elements with webp and fallback image formats.

Usage example:
```html
<picture>
    <source 
        srcset="{{img_url feature_image size="s" format="avif"}} 300w,
                {{img_url feature_image size="m" format="avif"}} 600w,
                {{img_url feature_image size="l" format="avif"}} 1000w,
                {{img_url feature_image size="xl" format="avif"}} 2000w"
        sizes="(min-width: 1400px) 1400px, 92vw" 
        type="image/avif"
    >
    <source 
        srcset="{{img_url feature_image size="s" format="webp"}} 300w,
                {{img_url feature_image size="m" format="webp"}} 600w,
                {{img_url feature_image size="l" format="webp"}} 1000w,
                {{img_url feature_image size="xl" format="webp"}} 2000w"
        sizes="(min-width: 1400px) 1400px, 92vw" 
        type="image/webp"
    >
    <img
        srcset="{{img_url feature_image size="s"}} 300w,
                {{img_url feature_image size="m"}} 600w,
                {{img_url feature_image size="l"}} 1000w,
                {{img_url feature_image size="xl"}} 2000w"
        sizes="(min-width: 1400px) 1400px, 92vw"
        src="{{img_url feature_image size="xl"}}"
        alt="{{#if feature_image_alt}}{{feature_image_alt}}{{else}}{{title}}{{/if}}"
    >
</picture>
```

### DIFF
--- a/ghost/core/core/frontend/helpers/img_url.js
+++ b/ghost/core/core/frontend/helpers/img_url.js
@@ -12,6 +12,7 @@ const url = require('url');
 const _ = require('lodash');
 const logging = require('@tryghost/logging');
 const tpl = require('@tryghost/tpl');
+const imageTransform = require('@tryghost/image-transform');
 
 const messages = {
     attrIsRequired: 'Attribute is required e.g. {{img_url feature_image}}'
@@ -116,8 +117,14 @@ function getUnsplashImage(imagePath, sizeOptions) {
     const parsedUrl = new URL(imagePath);
     const {requestedSize, imageSizes, requestedFormat} = sizeOptions;
 
-    if (requestedFormat) {
-        parsedUrl.searchParams.set('fm', requestedFormat);
+    if (requestedFormat) {        
+        const supportedFormats = ['avif', 'gif', 'jpg', 'png', 'webp'];
+        if (supportedFormats.includes(requestedFormat)) {
+            parsedUrl.searchParams.set('fm', requestedFormat);
+        } else if (requestedFormat === 'jpeg') {
+            // Map to alias
+            parsedUrl.searchParams.set('fm', 'jpg');
+        }
     }
 
     if (!imageSizes || !imageSizes[requestedSize]) {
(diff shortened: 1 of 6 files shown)


markstos:
Since this is targeted at Unsplash, there is a alternative that would be less work for theme developers and also Ghost. Instead of asking Unsplash for a specific format, Unsplash in turn uses the Imgix API and it supports `auto=format`.

This allows the potential that the Unsplash API might return the AVIF format if the browser supports it, which could be even smaller than `webp` while having higher visual "appeal". Here's what the Imgix API docs say:

> If [auto=format](https://docs.imgix.com/apis/rendering/auto/auto#format) is set and the browser does not support the AVIF format, we'll first fallback to WebP if it is supported. If WebP is not supported, imgix will fall back to the source image type.

https://docs.imgix.com/apis/rendering/format/fm#avi

And here's where the Unsplash API docs mention support for `auto=format`:

https://unsplash.com/documentation#dynamically-resizable-images

Related: there's an open "Idea" for Ghost to improve support for the AVIF format. Improving the Unsplash integration could a first step.

markstos:
Sounds good, @SimonBackx

ErisDS (changes requested):
Very minor thing, but the change to middleware/handle-image-sizes.js reduces the coverage down from 100%. Could you please add a test to fix this?

Otherwise, this change looks great 👍

ErisDS (approved):

PR #14969, opened 2022-06-05
removed catch predicate from integrations

Dave3of5:
Start of work removing the catch predicates for https://github.com/TryGhost/Ghost/issues/14882.


ErisDS:
Hey @Dave3of5 thank you for the PR and sorry for dropping the ball here, we got a bit of a backlog but are almost caught up again now. Have merged and this will go out in todays release. Thanks again 🙏

PR #14970, opened 2022-06-07
chore: Set permissions for GitHub actions

nathannaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>


.github/workflows/stale.yml
```diff
@@ -2,8 +2,14 @@ name: 'Close stale issues and PRs'
 on:
   schedule:
     - cron: '0 15 * * *'
+permissions:
```
daniellockyer, line 5: Is this needed? 🤔 The `stale` Action only needs access to `issues` and `pull-requests`

.github/workflows/auto-assign.yml
```diff
@@ -2,6 +2,9 @@ name: Auto Assign
 on:
   pull_request:
     types: [opened, reopened]
+permissions:
```
daniellockyer, line 5: I think we need `pull-request: write` here because `gh-auto-assign` assigns PRs to users

.github/workflows/canary.yml
```diff
@@ -2,6 +2,11 @@ name: Canary Build
 on: workflow_dispatch
 env:
   FORCE_COLOR: 1
+permissions:
+  contents: read
+
 jobs:
   canary:
+    permissions:
+      contents: none
```
daniellockyer, line 11: Why is `none` here? 🤔

daniellockyer:
Closing for now as this is stale and outdated, but I have the issue tracked here: https://github.com/TryGhost/Toolbox/issues/591

PR #14972, opened 2022-06-08
Replaced Promise.join() with .all() in user model

Emmanuel-Melon:

### DIFF
--- a/ghost/core/core/server/models/user.js
+++ b/ghost/core/core/server/models/user.js
@@ -1005,10 +1005,10 @@ User = ghostBookshelf.Model.extend({
         let ownerRole;
         let contextUser;
 
-        return Promise.join(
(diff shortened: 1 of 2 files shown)


markstos:
Why is this change considered an improvement? The docs for Promise.join say "While [.all](http://bluebirdjs.com/docs/api/all.html) is good for handling a dynamically sized list of uniform promises, Promise.join is much easier (and more performant) to use when you have a fixed amount of discrete promises that you want to coordinate concurrently."

Did you check to see which method performed better? 

If the the goal was to use a native Promise method instead of a Bluebird method, Bluebird is still required after this change is made, so there's not much benefit there.

vikaspotluri123:
@markstos see #14882 🙂

PR #14984, opened 2022-06-26
core: refactor bulk email service to allow non-mailgun senders.

markstos:
## Use Case

Ghost documentation says "More bulk mailers besides  Mailgun are planned". This PR helps along that goal with some foundational work.

Also, I signed up with Mailgun today, and there is no indication there is a free tier. The lowest
publish tier is the $35/month "Foundation" plan. I contact support today to clarify.

## Design Goal

This refactor is designed so that the Ghost does not need to support an alternate bulk email sender in the core. Alternate providers could be implemented as unsupported, third-party NPM modules.

Retrieving analytics back from the bulk email provider could be handled in the same way.

## Implementation

 - Removes references to mailgun in bulk-email-processor.js
 - Defines a bulk emailer module in configuration, defaulting to Mailgun
 - Loads the bulk mailer module based on the config

Now a new bulk email service can be defined by:

 1. Creating a bulk email module that exports `BATCH_SIZE`,
    `getInstance` and `send` compatible with Ghost's `mailgun.js`.
 2. Upload that module to NPM.
 3. Reference that module in config at `buklEmail:module`

The alternate module will likely require its own API key, which can
also be defined in the config system or setting system.

### DIFF
--- a/core/server/services/bulk-email/bulk-email-processor.js
+++ b/core/server/services/bulk-email/bulk-email-processor.js
@@ -1,11 +1,12 @@
 const _ = require('lodash');
 const Promise = require('bluebird');
+const config = require('../../../shared/config');
 const moment = require('moment-timezone');
 const errors = require('@tryghost/errors');
 const tpl = require('@tryghost/tpl');
 const logging = require('@tryghost/logging');
 const models = require('../../models');
-const mailgunProvider = require('./mailgun');
+const bulkEmailProvider = require(config.get('bulkEmail:module'));
 const sentry = require('../../../shared/sentry');
 const debug = require('@tryghost/debug')('mega');
 const postEmailSerializer = require('../mega/post-email-serializer');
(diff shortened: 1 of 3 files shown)


ErisDS:
Hey @markstos, apologies for taking a while to get back to you on this PR - I've been slowly, slowly trying to catch up with all the issues and PRs across our repos after getting quite behind! 

As you correctly identified, we have an adapter pattern that we would use to abstract out the bulk mail provider. 

I think this is quite a desirable move for a lot of self-hosters and I'd be happy to support you or anyone else that would like to pursue laying some foundation pieces for abstracting out bulk mail and/or email analytics to work with other providers.

For now, I'll close this PR - but again I'm happy to discuss it further or offer support to make progress with this.

jugglingjsons:
Hey! 
@ErisDS would it be possible to identify what would be an acceptable end goal here?
Would it include an additional adapter alongside the abstraction?
Any chance there is a shortlist of desired adapters perhaps ?

Also is there any comments over the stuff that was proposed in this PR? I would love to pick it up, but want to understand a bit on more as to what was missing from the proposed changes.

markstos:
@jugglingjsons I like to think I had the right idea in this PR, but didn't formally follow the established adapter pattern, I doubt they want to support any other mail providers internally, so it seems fair if a new PR just implements the adapter pattern, creating a place to plug external adapters into without actually shipping other adapters. People could publish other implementations to NPM on Github.  I ended up using Mailgun, so I'm not pursuing this myself.

PR #14987, opened 2022-06-27
:bug: Fixed Multiple accounts for members

### LINKED ISSUE #14840: Multiple accounts for members
### Issue Summary

When the user use the same email address to login but with different case sensitivity, multiple accounts are created.

image

### Steps to Reproduce

1. Sign in using your email (`Name@Google.com`)
2. Click on the sign up link you received
3. Logout
4. Sign in using your email (`name@google.com`)
5. Click on the sign up link you received
6. Check your members in Ghost admin and you should see multiple accounts for the same email address

### Ghost Version

4.47.1

### Node.js Version

v14.19.1

Docker

### Database type

SQLite3

### Browser & OS version

_No response_

_No response_

### Code of Conduct

ahmedelbilal:
Resolves #14840

- It's allowed to create multi accounts with the same email
- The emails are equal but have different characters cases
- e.g example@domain.com & Example@domain.com are both saved to the DB
- I fixed the issue by changing the email to lower case before saving it to the DB

### DIFF
--- a/ghost/core/core/server/web/shared/middleware/brute.js
+++ b/ghost/core/core/server/web/shared/middleware/brute.js
@@ -103,6 +103,8 @@ module.exports = {
             ignoreIP: false,
             key(_req, _res, _next) {
                 if (_req.body.email) {
+                    // Change the email to lower case to prevent multi accounts creations
+                    req.body.email = req.body.email.toLowerCase();
                     return _next(`${_req.body.email}login`);
                 }
 


ErisDS:
Hey @ahmed-sudani, thanks for taking the time to submit this PR and apologies for taking so long to review it - we got a bit behind!

Whilst this change solves the problem on the face of things, I don't think the brute force protection layer is the right place to be modifying user input.

Further still, I don't think changing the email address to be stored as lower case is necessarily the right solution as this changes the behaviour. At the moment, we store whatever the user enters and use the database to ensure uniqueness. 

As far as I understand, the reason this problem is happening is because SQLite3 isn't enforcing the unique constraint correctly. I still believe that the correct solution is to resolve this in knex, so that the `.unique()` constraint behaves consistently across all DBs. That solves this problem permanently, for all fields that should be unique, across every project that uses knex.

PR #14989, opened 2022-06-28
Add Crawl-delay to robots.txt by default

touzoku:
We have recently faced a DDoS attack on our ghost instance using a search bot amplification mechanism.
This setting is respected by some search engines such as Bing and Yandex and mitigates such attacks.
This should be a default setting, we believe.

### DIFF
--- a/core/frontend/public/robots.txt
+++ b/core/frontend/public/robots.txt
@@ -3,3 +3,4 @@ Sitemap: {{blog-url}}/sitemap.xml
 Disallow: /ghost/
 Disallow: /p/
 Disallow: /email/
+Crawl-delay: 10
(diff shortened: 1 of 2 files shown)


ErisDS:
Hey @touzoku, I'm sorry for taking so long to respond & also very sorry to hear you've had DDoS issues. I can imagine that was pretty stressful.

It's possible to override Ghost's default `robots.txt` file by adding your own file to the root of your theme, which should at least allow you to mitigate the issue.

I'm all for improving the defaults in Ghost, but before merging this PR I'd love to see some reference material for why this specific setting should be Ghost's default. Did you find any good information about how various bots respect the setting, and what the right value is to balance timely updates and mitigating attacks?

PR #15005, opened 2022-07-06
🐛 Fixed API excerpt field issue

### LINKED ISSUE #10396: `fields=excerpt` requires underlying knowledge
The `excerpt` field is a calculated field based on `plaintext` at the moment if you request e.g. `?fields=title,excerpt` you won't receive the excerpt field unless you also supply `?formats=plaintext`.

### To Reproduce

1. Visit https://demo.ghost.io/ghost/api/v2/content/posts/?key=22444f78447824223cefc48062&fields=title,excerpt
2. Notice excerpt is missing completely
3. Add `&formats=plaintext` https://demo.ghost.io/ghost/api/v2/content/posts/?key=22444f78447824223cefc48062&fields=title,excerpt&formats=plaintext
4. Now we get the `excerpt` and the `plaintext` field.

It's unexpected that I am only able to get the excerpt field whilst the `plaintext` format is also included. This logic should be "behind the scenes" much the same as the `url` field.

### Technical details:

* Ghost Version: master

ronaldlangeveld:
closes https://github.com/TryGhost/Ghost/issues/10396

This fix now allows the API user to add field `excerpt` without the need for `plaintext` as format.
Also added new tests for these functions.


core/server/api/endpoints/utils/serializers/output/utils/extra-attrs.js
```diff
@@ -3,30 +3,59 @@ const readingMinutes = require('@tryghost/helpers').utils.readingMinutes;
...
-            let plaintext = model.get('plaintext');
-
-            if (plaintext) {
-                attrs.excerpt = plaintext.substring(0, 500);
+    // This function is split up in 3 conditions for 3 different purposes:
+    // 1. Gets excerpt from post's plaintext. If custom_excerpt exists, it overrides the excerpt but the key remains excerpt.
+    if (Object.prototype.hasOwnProperty.call(frame.options, 'columns') || _.includes(frame.options.columns, 'excerpt') || _.includes(frame.options.columns, 'excerpt') && frame.options.formats && frame.options.formats.includes('plaintext')) {
+        if (_.includes(frame.options.columns, 'excerpt')) {
+            if (!attrs.custom_excerpt || attrs.custom_excerpt === null) {
+                let plaintext = model.get('plaintext');
```
naz, line 11: fyi, there's loads of whitespaces in the end of this line. we don't want any of that :upside_down_face:

test/e2e-api/content/__snapshots__/posts.test.js.snap
```diff
@@ -23,7 +23,11 @@ Object {
       "custom_excerpt": "We've crammed the most important information to help you get started with Ghost into this one post. It's your cheat-sheet to get started, and your shortcut to advanced features.",
       "custom_template": null,
       "email_subject": null,
-      "excerpt": "We've crammed the most important information to help you get started with Ghost into this one post. It's your cheat-sheet to get started, and your shortcut to advanced features.",
+      "excerpt": "Hey there, welcome to your new home on the web!
```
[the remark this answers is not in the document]
ronaldlangeveld: okay I found.. for some reason custom_excerpt isn't overriding the generated excerpt. Let me fix that.

naz:
@ronaldlangeveld found this bit of behavior strange:
Screenshot from 2022-07-06 15-05-38

above illustrates how an excerpt is generated from the post's plaintext when there's clearly a custom excerpt defined.

Screenshot from 2022-07-06 15-04-48

and this one shows how the value changes all of the sudden if we tell the API to only include an excerpt. 

in both cases an intuitive output would be the value from custom_excerpt as it always has a higher priority over the plainext.

fwiw, this specific behavior was not clearly described in the issue but should definitely be fixed with this change

core/server/api/endpoints/utils/serializers/output/utils/extra-attrs.js
ronaldlangeveld: Fixed! :)

ronaldlangeveld:
> @ronaldlangeveld found this bit of behavior strange: Screenshot from 2022-07-06 15-05-38
> 
> above illustrates how an excerpt is generated from the post's plaintext when there's clearly a custom excerpt defined.
> 
> Screenshot from 2022-07-06 15-04-48
> 
> and this one shows how the value changes all of the sudden if we tell the API to only include an excerpt.
> 
> in both cases an intuitive output would be the value from custom_excerpt as it always has a higher priority over the plainext.
> 
> fwiw, this specific behavior was not clearly described in the issue but should definitely be fixed with this change

Apologies, I think I may have ran the snapshot update a touch careless there. Fixed it.

Modified some logic in extra-attrs and now takes excerpt and custom_excerpt into account when all posts gets queried without any specific fields.

test/e2e-api/content/posts.test.js
```diff
@@ -309,4 +309,18 @@ describe('Posts Content API', function () {
 
         tiersPostData.tiers.length.should.eql(1);
     });
+
+    it('Can use post excerpt as field', async function () {
+        await agent
+            .get(`posts/?fields=excerpt&limit=1`)
```
naz, line 315: not sure about `&limit1` here. we should be just fine requesting all the posts we have. the upside of not limiting the response - we can monitor excerpt field regressions if they happen

test/unit/api/canary/utils/serializers/output/utils/extra-attrs.test.js
```diff
@@ -4,7 +4,10 @@ const extraAttrsUtil = require('../../../../../../../../core/server/api/endpoint
 
 describe('Unit: endpoints/utils/serializers/output/utils/extra-attrs', function () {
     const frame = {
-        options: {}
+        // Question: is it okay to use actual column values here that the forPost function in extra-attrs would expect?
```
naz, line 7: I don't see why not?

ronaldlangeveld: Great!

naz:
@ronaldlangeveld few minor tweaks are left with things pointed out above. otherwise it looks :ok_hand: :ok_hand: :ok_hand: 

feel free to merge it in and if anything else comes up we can deal with it separately.

One thing to note, we need to document the new 'plaintext' field and the behavior of the excerpt field. it's also a breaking change because we finally fix the piority of `custom_excerpt`>`plaintext` generate excerpt.

naz:
@daniellockyer to be on a safe side we should check the performance degradation introduced by always adding the `excerpt` and the `plaintext` fields to the post queries.

PR #15013, opened 2022-07-08
✨ Member Counter template helpers 

ronaldlangeveld:
ref https://github.com/TryGhost/Team/issues/1667

Introducing 2 new helper handlebars tags, `{{total_members}}` and `{{total_paid_members}}` ideal for Member Sites who want to display these metrics to incentivise users to upgrade.


package.json
```diff
@@ -146,6 +146,7 @@
     "got": "9.6.0",
     "gscan": "4.31.2",
     "html-to-text": "8.2.0",
+    "human-number": "^2.0.0",
```
daniellockyer, line 149: Would you be able to pin the dependency to `2.0.0`?

ronaldlangeveld: Done!

allouis (approved):
Looks good - only thing is the comments at the top of the helpers files, which are weird - but they do follow existing patterns ☺️ 

Am not fussed if they stay or go

PR #15022, opened 2022-07-13
🐛 Fixed cookies when running Ghost without SSL

allouis:
As part of the updates to auth cookies we switched to SameSite=None
which requires an SSL connection. Local development, and some
production sites do not use SSL and so the cookie is invalid and a
session is unable to be created with the browser.


kevinansfield:
@thesved Chrome is not setting a cookie because of the _combination_ of `sameSite: none` and `secure: false`. This commit fixes it by returning to the previous behaviour of `sameSite: lax` and `secure: false` for http sites.

thesved:
@kevinansfield gotcha, my bad! 😅

PR #15026, opened 2022-07-13
Added access-control-max-age to preflight requests

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1676

- Added maxAge option to content API
- Added maxAge to members API
- Added maxAge to frontend site preflights (probably not used, but it was configured, so added to be sure)
- Added config option to change default maxAge of preflight requests


allouis (changes requested):
Can we add/use values from config rather than hardcoding them? 

https://github.com/TryGhost/Ghost/blob/main/core/shared/config/defaults.json#L98-L117

allouis (approved):

PR #15029, opened 2022-07-14
🐛 Updated support email verification flow

SimonBackx:
refs https://github.com/TryGhost/Team/issues/584

The current support email verification flow uses an API endpoint as verification URL inside the emails. This is a bad pattern, and also has the side effect that it shows a JSON error if something goes wrong.

To fix this, this commit updates the whole flow to use the same pattern as newsletters:
- You can update the `members_support_address` setting directly via the edit endpoint of settings.
- Changes to that (and future 'guarded' email properties) are blocked and generate verification emails automatically.
- When an email verification has been sent, the meta property `sent_email_verification` is set.

Other changes:
- Underlying, the implementation of email verificaton has moved from the (old) members service to the settings BREAD service. This makes it easier to add extra email addresses in settings later on that are not related to 'members'.
- Now you can update the `members_support_address` by updating the settings directly, so the `updateMembersEmail` endpoint has been deprecated and is mapped to the new behaviour.
- The SingleUseTokenProvider threw a `UnauthorizedError` error if a token was expired or invalid. Those errors are caught by the admin app, and causes it to do a page reload (making the error message and modals invisible). To fix that, I've swapped it with a validation error.

Future changes:
- Existing emails that have been sent 24h before this change is applied, still use the `validateMembersEmailUpdate` API endpoint. This endpoint has not been removed for now, to not break those emails. In a future release, we should remove this.

Changes to admin: https://github.com/TryGhost/Admin/pull/2426

### DIFF
--- a/core/server/api/endpoints/settings.js
+++ b/core/server/api/endpoints/settings.js
@@ -74,6 +74,41 @@ module.exports = {
     },
 
     /**
+     * @deprecated
+     */
+    updateMembersEmail: {
+        statusCode: 204,
+        permissions: {
+            method: 'edit'
+        },
+        data: [
+            'email',
+            'type'
+        ],
+        async query(frame) {
+            const {email, type} = frame.data;
+
+            try {
+                // Mapped internally to the newer method of changing emails
+                const actionToKeyMapping = {
+                    supportAddressUpdate: 'members_support_address'
+                };
+                const edit = {
+                    key: actionToKeyMapping[type],
+                    value: email
+                };
+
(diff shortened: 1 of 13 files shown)


matthanley:
> the updateMembersEmail endpoint is no longer needed and has been removed.

This is technically a breaking change. Can we reinstate it, but map it internally using the deprecation header to a settings update to trigger the new flow?

SimonBackx:
You are right. I readded the old endpoints ✅

PR #15030, opened 2022-07-14
🐛 Fixed 'comped' flag behavior in Member API

naz:
closes https://github.com/TryGhost/Team/issues/1674
refs https://github.com/TryGhost/Members/pull/407

- The comped flag in Members API unintentionally stopped working when v3 API was dropped with the release of Ghost v5. The flag is deprecated but should be back-compatible for now - we don't want to break integrations like Zapier.
- To properly deprecate the flag we need to plan it's removal and start signalling about it through the version headers

@allouis @SimonBackx marking you for the review as this change has triggered quite a few rewrites in the member test snapshots. The most unexpected one is the format change in the `MemberPaidSubscriptionEvent` event test. The rest seems like an expected change. It's the territory I have not touched in a long while, so would be great to get 2x:eyes: over the change :pray:


SimonBackx (approved):
The issue with `MemberPaidSubscriptionEvent` was that we had two Stripe Subscriptions with the same ID. This wasn't an issue in the past, because the subscription was ignored while the `comped` flag was removed. Now that started to cause some issues internally in the second test (because the subscription already existed and didn't change, so it didn't need to create the MemberPaidSubscriptionEvent). By using a new subscription id, it now creates a new subscription internally and also a new MemberPaidSubscriptionEvent.

\+ I also bumped `package.json` to test this locally

allouis (approved):
Thanks @SimonBackx for sorting the tests!

PR #15031, opened 2022-07-14
Redirected signin magic links to referrer when present

allouis:
refs https://github.com/TryGhost/Team/issues/1174
refs https://github.com/TryGhost/Members/pull/408

This builds on the work from the Members PR above to 1. use the
referrer as a query param for magic links and 2. to redirect to that
referrer when members are signing in.

The redirect is also restricted to URI's which are on the same domain
and subdirectory as the frontend URL, this is to prevent phishing
attacks.


core/server/services/members/middleware.js
```diff
@@ -182,6 +182,17 @@ const createSessionFromMagicLink = async function (req, res, next) {
             }
         }
 
+        if (action === 'signin') {
+            const referrer = req.query.referrer;
+            const siteUrl = urlUtils.getSiteUrl();
+
+            if (referrer && referrer.startsWith(siteUrl)) {
```
allouis, line 189: I think this check could be better maybe? Need to investigate edgecases

allouis: The default is `strict-origin-when-cross-origin` which works fine with this - we can choose to explicitly set one though!

matthanley (approved):

matthanley (changes requested):
Still needs tests 😄

core/server/services/members/middleware.js
```diff
@@ -182,6 +182,17 @@ const createSessionFromMagicLink = async function (req, res, next) {
             }
         }
 
+        if (action === 'signin') {
+            const referrer = req.query.referrer;
```
matthanley, line 186: Hannah has pointed out that this should be `?r=` because we already have that pattern in place for private site logins 😃

allouis:
@matthanley Tests have been added, and the members packages are bumped in this PR

test/unit/server/services/members/middleware.test.js
```diff
@@ -147,5 +147,37 @@ describe('Members Service Middleware', function () {
...
+            // Fake token handling failure
+            membersService.ssr.exchangeTokenForSession.resolves();
+
+            // Call the middleware
+            await membersMiddleware.createSessionFromMagicLink(req, res, next);
+
+            // Check behaviour
+            next.calledOnce.should.be.false();
+            res.redirect.calledOnce.should.be.true();
+            res.redirect.firstCall.args[0].should.eql('/blah/?action=signin&r=https%3A%2F%2Fexternal.com%2Fwhatever%2F&success=true');
```
matthanley, line 180: Should we strip the `r` param if we're not redirecting?

test/unit/server/services/members/middleware.test.js
```diff
@@ -147,5 +147,37 @@ describe('Members Service Middleware', function () {
...
+            // Fake token handling failure
+            membersService.ssr.exchangeTokenForSession.resolves();
+
+            // Call the middleware
+            await membersMiddleware.createSessionFromMagicLink(req, res, next);
+
+            // Check behaviour
+            next.calledOnce.should.be.false();
+            res.redirect.calledOnce.should.be.true();
+            res.redirect.firstCall.args[0].should.eql('/blah/my-post/?success=true');
```
matthanley, line 164: I think this is missing the `action` param?

allouis: Good shout

test/unit/server/services/members/middleware.test.js
allouis: We don't strip any other params when not redirecting 🤔 I can add that though if you think it's useful

matthanley (approved):

PR #15043, opened 2022-07-18
Added report API for comments

SimonBackx:
closes https://github.com/TryGhost/Team/issues/1684

**Migrations:**
- Added report permissions (fixtures + migrations)
- Dropped reason field in reports (no textarea in reports in V1)
- Dropped nullable from comment_likes.member_id (can't be null)
- Added SET NULL/CASCADE foreign keys for comments related tables(*)

(*):
fixes https://github.com/TryGhost/Team/issues/1687
refs https://ghost.slack.com/archives/C02G9E68C/p1658217288591369

This commit adds support for `SET NULL` foreign keys in schema and migration helpers + also fixes the foreign keys for the comment_reports, comment_likes and comments tables.

- When a member is deleted, we **do** want to keep their reports (SET NULL)
- When a member is deleted, we **do not** want to keep their likes (CASCADE)
- When a member is deleted, we **do** want to keep the comments (SET NULL)

**Changes:**
- Added report API: `POST /members/api/comments/{id}}/report/`
- Sends an email to the owner when a comment is reported
- Saves a report to the database (not used for now, but might be useful later)

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.5/2022-07-18-14-33-fix-comments-on-delete-foreign-keys.js
+++ b/ghost/core/core/server/data/migrations/versions/5.5/2022-07-18-14-33-fix-comments-on-delete-foreign-keys.js
@@ -1,0 +1,119 @@
+const {addForeign, dropForeign} = require('../../../schema/commands');
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Adding on delete SET NULL for comments');
+    
+        await dropForeign({
+            fromTable: 'comments',
+            fromColumn: 'member_id',
+            toTable: 'members',
+            toColumn: 'id',
+            transaction: knex
+        });
+    
+        await addForeign({
+            fromTable: 'comments',
+            fromColumn: 'member_id',
+            toTable: 'members',
+            toColumn: 'id',
+            setNullDelete: true,
+            transaction: knex
+        });
+    
+        logging.info('Adding on delete CASCADE for comment_likes');
+    
+        await dropForeign({
+            fromTable: 'comment_likes',
+            fromColumn: 'member_id',
+            toTable: 'members',
+            toColumn: 'id',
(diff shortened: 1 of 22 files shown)


ErisDS:
I maybe missed some context for this, but I think we need to double check this:

> Dropped nullable from member_id in reports (no anonymous reports allowed)

That would mean deleting a member also deletes their reports? Or do we actually want this to be nullable for that case?

core/server/data/migrations/versions/5.5/2022-07-18-14-29-add-comment-reporting-permissions.js
```diff
@@ -0,0 +1,12 @@
+const {combineTransactionalMigrations, addPermissionWithRoles} = require('../../utils');
+
+module.exports = combineTransactionalMigrations(
```
allouis, line 3: No need for the use of combine for a single migration

SimonBackx: Fixed! 👌

allouis (approved):
All looks good - I have a minor concern about the size of the DDL migration where we add/remove a load of constraints.

I was wondering what happens if one of the operations fail, if the rollback will set it to the correct state - I think it's safe, but would like @daniellockyer's opinion on it too!

ghost/core/core/server/data/migrations/versions/5.5/2022-07-18-14-33-fix-comments-on-delete-foreign-keys.js
```diff
@@ -0,0 +1,119 @@
+const {addForeign, dropForeign} = require('../../../schema/commands');
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Adding on delete SET NULL for comments');
```
daniellockyer, line 7: Can you make this `... comments.member_id`?

daniellockyer (approved):

PR #15044, opened 2022-07-19
🐛 Fixed sending multiple support email verification emails

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1686

- When the settings are updated with the `members_support_address` key present, it would always send a verification email
- Root cause is that the service failed to check if the email was changed or not. Due to a bug it always thought the email was changed, triggering the verification flow.
- The admin app will always send all the settings keys when changing some other value. This causes a lot of email verification emails.
- Added tests and email count checks in tests


SimonBackx:
Test fail is random, retrying...

matthanley (approved):
Tested locally, lgtm 👍🏼

PR #15053, opened 2022-07-20
Added guard for `activateTheme` being called in parallel causing sync bugs

daniellockyer:
refs https://github.com/TryGhost/Team/issues/1666

- it seems like we may have a situation where `.activateTheme()` can be called simultaneously resulting in unexpected behaviour in the sync such as duplicate theme setting records
- adjusted behaviour to keep track of the currently running activation within the service and if `.activateTheme()` is called again whilst it's in progress it will wait for completion of the first sync before exiting early or continuing with a new activation

**Note:** There is a known edge-case if there are _more_ than 2 parallel `.activateTheme()` calls. We don't believe that will be an issue but calling it out in case we do still see duplicated custom setting records being created.

### DIFF
--- a/ghost/custom-theme-settings-service/lib/service.js
+++ b/ghost/custom-theme-settings-service/lib/service.js
@@ -20,6 +20,10 @@ module.exports = class CustomThemeSettingsService {
         this.activeThemeName = null;
 
         /** @private */
+        this._activatingPromise = null;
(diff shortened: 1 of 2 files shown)


daniellockyer:
Migrated from https://github.com/TryGhost/Publishing/pull/74/files

PR #15059, opened 2022-07-21
Added jobs table creation migration

naz:
refs https://github.com/TryGhost/Toolbox/issues/357

One time jobs need a storage mechanism to be run only ever once.

Field notes:
- `id`, `created_at`, `updated_at` - standard Ghost fields
- `name` - unique name of the job, could also be used with prefixing to identify certain type of job (e.g.: backup-bob-2022-10-16, backup-sam-2023-01-13 identifying backup jobs run by users)
- `status` - 'active' | 'done' | 'failed' | 'queued'  (need to identify when the job is in progress, done, added to the queue, or errored)
- `started_at` -  when the job started execution
- `finished_at` - when the job *successfully?* finished execution

Added `jobs` to the exported tables list as they are not a part of "secret" data and would identify which data transformations have potentially have been already done through one-time jobs (main usecase of this type of jobs).

:heavy_check_mark:  ~The final spec of the table structure is still in the review.~ Having the PR opened for visibility and be ready to tweak&merge in coming days.

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.5/2022-07-21-08-56-add-jobs-table.js
+++ b/ghost/core/core/server/data/migrations/versions/5.5/2022-07-21-08-56-add-jobs-table.js
@@ -1,0 +1,11 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('jobs', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    name: {type: 'string', maxlength: 191, nullable: false, unique: true},
+    status: {type: 'string', maxlength: 50, nullable: false, defaultTo: 'queued', validations: {isIn: [['started', 'finished', 'failed', 'queued']]}},
+    started_at: {type: 'dateTime', nullable: true},
+    finished_at: {type: 'dateTime', nullable: true},
+    created_at: {type: 'dateTime', nullable: false},
+    updated_at: {type: 'dateTime', nullable: true}
+});
--- a/ghost/core/core/server/data/exporter/table-lists.js
+++ b/ghost/core/core/server/data/exporter/table-lists.js
@@ -34,7 +34,8 @@ const BACKUP_TABLES = [
     'members_newsletters',
     'comments',
     'comment_likes',
-    'comment_reports'
+    'comment_reports',
+    'jobs'
 ];
 
 // NOTE: exposing only tables which are going to be included in a "default" export file
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -771,5 +771,14 @@ module.exports = {
         member_id: {type: 'string', maxlength: 24, nullable: true, unique: false, references: 'members.id', setNullDelete: true},
         created_at: {type: 'dateTime', nullable: false},
         updated_at: {type: 'dateTime', nullable: false}
+    },
+    jobs: {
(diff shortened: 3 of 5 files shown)


ghost/core/core/server/data/migrations/versions/5.5/2022-07-21-08-56-add-jobs-table.js
```diff
@@ -0,0 +1,11 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('jobs', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    name: {type: 'string', maxlength: 191, nullable: false, unique: true},
+    status: {type: 'string', maxlength: 50, nullable: false, defaultTo: 'queued', validations: {isIn: ['started', 'finished', 'failed', 'queued']}},
```
daniellockyer, line 6: Even though it's not important here, `isIn` requires the value to be an array of an array - like you have in the schema

ghost/core/core/server/data/migrations/versions/5.5/2022-07-21-08-56-add-jobs-table.js
```diff
@@ -0,0 +1,11 @@
...
+    started_at: {type: 'dateTime', nullable: true},
+    finished_at: {type: 'dateTime', nullable: true},
+    created_at: {type: 'dateTime', nullable: false},
+    updated_at: {type: 'dateTime', nullable: false}
```
daniellockyer, line 10: Is there a reason `updated_at` is `nullable: false`? If we first insert the job, I'd argue it hasn't been "updated" yet? I spoke with @ErisDS about this discrepancy on other tables but if we agree, it'd be good to avoid adding another one

naz: Oh. It is a very minor detail, but I do like this sort of correctness. It does make sense!

naz: The thing with the `updated_at` is that it's [being set by Bookshelf](https://bookshelfjs.org/api.html#Model-instance-hasTimestamps) no matter what during the insert. We could disable this default behavior and rely solely on the user data. This is a slightly bigger change that's not in the scope here :sweat_smile:

daniellockyer: 🤔 I'd still set it to non-nullable to prepare for the lovely day that we nuke Bookshelf 😈

naz: Yeah totally!

daniellockyer (approved):
I think this is fine in terms of safety

ghost/core/core/server/data/migrations/versions/5.5/2022-07-21-08-56-add-jobs-table.js
ErisDS: Sorry I didn't get to this before - but what Daniel said :D

ErisDS:
This seems good to me - but given the use-cases we have, I feel like we might need another field. Something like a nullable type column? That way backups can identify themselves as backups and we can select * from jobs where type = 'backup'. Otherwise we have to rely on a naming convention, which feels fragile. 

Some jobs might not have a type if they are true one-offs, and that's fine IMO.

PR #15088, opened 2022-07-25
Added members.last_commented_at and last_seen_at update when commenting

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1717

- Updates last_commented_at and last_seen_at (only once a day)
- Used the LastSeenAtUpdater, so we can combine updating last_commented_at and last_seen_at in one query + used same pattern
- Updated comments service to await emails in order to make E2E tests more stable (as we don't have any method to await emails and test emails otherwise). This removed the email sending logic from the `onCreated` hook of the model (having it in the service is also cleaner I think?).

### DIFF
--- a/ghost/core/core/server/services/comments/service.js
+++ b/ghost/core/core/server/services/comments/service.js
@@ -1,5 +1,7 @@
 const tpl = require('@tryghost/tpl');
 const errors = require('@tryghost/errors');
+const {MemberCommentEvent} = require('@tryghost/member-events');
+const DomainEvents = require('@tryghost/domain-events');
 
 const messages = {
     commentNotFound: 'Comment could not be found',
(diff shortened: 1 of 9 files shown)


ghost/core/core/server/services/comments/service.js
```diff
@@ -24,13 +26,25 @@ class CommentsService {
...
     }
 
+    /**
+     * Dispatch an event that we created a new comments posted by given member
+     * @param {Object} member member model that posted the comment
+     */
+    dispatchCommentEvent(member) {
+        DomainEvents.dispatch(MemberCommentEvent.create({
+            memberId: member.id, 
+            memberLastSeenAt: member.get('last_seen_at'),
```
allouis, line 43: I know this is following the pattern we've used for the `MemberPageView` event - but this seems like we've coupled the event too tightly to the current event usage 🤔 

Perhaps A MemberCommentEvent should contain data more like:
```
{
  memberId,
  commentId,
  postId
}
```

And have the listener fetch the member - or use a service which handles checking whether or not we need to update 🤷 

Obviously that comes with a tradeoff of another database read, which if we're concerned about then perhaps we should just include the entire member in the payload? 

This isn't a huge deal, but it would be great if in the future when we want to perform some action when a member creates a comment, that we don't have to come here to edit the event, we'd have all the data we'd reasonably expect OR we'd have the pattern where we can fetch the data we need.

What do you think about this?

SimonBackx: Yes, I get what you mean :) Not sure if we should pass the models or the ids though. Passing only ids is definitely cleaner...

allouis: Agreed ids is cleaner - we could also just pass JSON representation of the models 🤔

allouis: OR - we merge as is, and add an issue somewhere to look at this in future, but not sure how likely it is we'll get to that

SimonBackx: Not gonna happen if we don't fix this now 😬

SimonBackx: I'm gonna go with the IDs I think.

allouis (approved):

PR #15089, opened 2022-07-25
Allowed configurable urls for frontend apps

rshbhgrg:
- current config for frontend apps doesn't allow easy switching of CDN for urls used to load them
- this change allows pro to switch out the CDN easily, and also avoids double use of version value
- also extends config to include url needed for loading CSS for frontend app

### DIFF
--- a/ghost/core/core/frontend/helpers/comments.js
+++ b/ghost/core/core/frontend/helpers/comments.js
@@ -1,11 +1,12 @@
 const {SafeString} = require('../services/handlebars');
-const {config, urlUtils, getFrontendKey, labs, settingsCache} = require('../services/proxy');
+const {urlUtils, getFrontendKey, labs, settingsCache} = require('../services/proxy');
+const {getFrontendAppConfig, getDataAttributes} = require('../utils/frontend-apps');
 
 async function comments(options) {
     // todo: For now check on the comment id to exclude normal pages (we probably have a better way to do this)
 
     const commentId = this.comment_id;
-        
+
     if (!commentId) {
         return;
     }
@@ -21,7 +22,7 @@ async function comments(options) {
     if (commentsEnabled === 'off' || !hasAccess) {
         return;
     }
-    
+
     let colorScheme = 'auto';
     if (options.hash.color_scheme === 'dark' || options.hash.color_scheme === 'light') {
         colorScheme = options.hash.color_scheme;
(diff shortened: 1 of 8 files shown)


ghost/core/core/shared/config/defaults.json
```diff
@@ -137,7 +137,8 @@
         "version": "2.3"
     },
     "sodoSearch": {
-        "url": "https://unpkg.com/@tryghost/sodo-search@~1.0.0/umd/sodo-search.min.js",
+        "url": "https://unpkg.com/@tryghost/sodo-search@~{version}/umd/sodo-search.min.js",
+        "css": "https://unpkg.com/@tryghost/sodo-search@~{version}/umd/main.css",
         "version": "1.0.0"
```
daniellockyer, line 142: Should we start truncating this to `major.minor` like the Portal one? It'll still resolve to the latest in unpkg and it should also work in jsdelivr because the expanded URL becomes `https://cdn.jsdelivr.net/npm/@tryghost/sodo-search@1.0/umd/sodo-search.min.js`

daniellockyer (approved):

ghost/core/core/frontend/helpers/ghost_head.js
```diff
@@ -47,9 +48,24 @@ function getMembersHelper(data, frontendKey) {
...
+        ghost: urlUtils.getSiteUrl(),
+        key: frontendKey,
+        api: urlUtils.urlFor('api', {type: 'content'}, true)
+    };
+    if (colorString) {
+        attributes['accent-color'] = colorString;
+    }
+    const dataAttributes = getDataAttributes(attributes);
+
+    if (colorString) {
```
daniellockyer, line 64: Was this leftover from something? `dataAttributes` is a string at this point 🤔

ghost/core/core/frontend/helpers/comments.js
```diff
@@ -38,18 +39,20 @@ async function comments(options) {
...
+        styles: stylesUrl,
         'post-id': this.id,
         'sentry-dsn': '', /* todo: insert sentry dsn key here */
         'color-scheme': colorScheme,
         'avatar-saturation': avatarSaturation,
         'accent-color': accentColor,
-        'app-version': config.get('comments:version'),
+        'app-version': appVersion,
         'comments-enabled': commentsEnabled
     };
```
daniellockyer, line 58: Can we switch to the new util `getDataAttributes` util below? (sorry, GitHub won't let me comment on those lines 🙃)

ghost/core/core/frontend/helpers/ghost_head.js
rishabhgrg: Ugh yea thanks, it was exactly that, was supposed to be deleted! Fixed

ghost/core/core/frontend/helpers/comments.js
rishabhgrg: Done ✅

daniellockyer (approved):

PR #15093, opened 2022-07-25
Refactored Comments API 

allouis:
🍦


ghost/core/core/server/api/endpoints/comments-comments.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
         },
         permissions: true,
         query(frame) {
-            return models.Comment.findPage(frame.options);
```
ErisDS, line 37: If and when you get chance, this file should prob be called members-comments.js or comments-members.js - comments-comments is left over from when we were going to have a separate comments api, but this is now officially part of the members api

allouis: Awesome - I'll sort that out in this PR

allouis: I'm going to go with `comments-members` as that matches the pattern we have for the content api with `resource-apitype` e.g. `posts-public`

ErisDS: 👌

PR #15105, opened 2022-07-27
Fixed failing admin preview test email

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1680

- using test emails via email preview in admin were failing due to missing post data attached to them
- adds test to make sure email segment rendering doesn't crash even with missing data

### DIFF
--- a/ghost/core/core/server/services/mega/mega.js
+++ b/ghost/core/core/server/services/mega/mega.js
@@ -69,6 +69,7 @@ const getEmailData = async (postModel, options) => {
     }
 
     return {
+        post: postModel.toJSON(), // for content paywalling
         subject,
         html,
         plaintext,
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/services/mega/mega.js
```diff
@@ -87,9 +88,6 @@ const sendTestEmail = async (postModel, toEmails, memberSegment) => {
     let emailData = await getEmailData(postModel);
     emailData.subject = `[Test] ${emailData.subject}`;
 
-    if (memberSegment) {
-        emailData = postEmailSerializer.renderEmailForSegment(emailData, memberSegment);
-    }
```
rishabhgrg, line 92: Removed as this was redundant here, we do serialization for email segment again in `bulkEmailService.send` before sending out the email.

ErisDS (approved):
LGTM 👍

PR #15123, opened 2022-08-01
Comment count helper updates

allouis:
refs https://github.com/TryGhost/Team/issues/1695

This updates the comment_count helper inline with our new design, as well as fixing some issues with the comments count API and the helper being async


ghost/core/core/frontend/helpers/comment_count.js
```diff
@@ -3,23 +3,24 @@ const {labs} = require('../services/proxy');
 
 function commentCount(options) {
     return new SafeString(`
-       <span data-ghost-comment-count="${this.id}" style="display:none">
-           ${options.fn(this)}
-       </span>`
+       <script
```
allouis, line 6: Should this perhaps be a visible element like a `span` so that we can style it whilst comments are loading?

SimonBackx, line 6: Any reason why we swapped to use empty script tags? To avoid styling issues?

ghost/core/core/frontend/helpers/comment_count.js
```diff
@@ -3,23 +3,24 @@ const {labs} = require('../services/proxy');
...
+        data-ghost-comment-count="${this.id}"
+        data-ghost-comment-count-empty="${options.hash.empty}"
+        data-ghost-comment-count-singular="${options.hash.singular}"
+        data-ghost-comment-count-plural="${options.hash.plural}"
```
SimonBackx, line 10: Should we add defaults if those parameters are not provided?

allouis: That's not in the design, and I think doing so would remove our ability to just output a number - which would be useful for a usecase like: `💬 {{comment_count empty="0"}}`

ghost/core/core/frontend/helpers/comment_count.js
allouis: We didn't want to force the content to be wrapped in a tag - and a script felt like a natural thing to use here - though I think we should probably use a visual element here, so that we can have the loading state styled

SimonBackx: I think that could be handy

allouis: As discussed in notion - we're going to leave this for now

PR #15128, opened 2022-08-02
Updated admin asset serving for ember-auto-import@2 compatibility

kevinansfield:
refs https://github.com/TryGhost/Admin/pull/2252

- Admin now copies it's build output to a single env-specific directory rather than splitting html and assets
  - `core/built/admin/{development|production}/*`
- updated the admin app's `serveStatic` definition for assets and controller's html serving to reflect the new asset paths


ErisDS (approved):
This all looks great. Additionally the folder `core/server/web/admin/views/` (which has a .gitkeep) can be removed I believe?

daniellockyer (approved):

PR #15129, opened 2022-08-02
Added migration to store trial period days on tier

rshbhgrg:

- storing trial period days allows site owners to give free trial for N days to all members signing up on a tier
- the new column name `trial_period_days` is consistent with naming used by Stripe for passing trial value
- by default, all tiers have 0 or no trial period days


ghost/core/core/server/data/schema/schema.js
```diff
@@ -422,6 +422,7 @@ module.exports = {
             defaultTo: 'none',
             validations: {isIn: [['public', 'none']]}
         },
+        trial_period_days: {type: 'integer', unsigned: true, nullable: false, defaultTo: 0},
```
rishabhgrg, line 425: The naming `trial_period_days` matches naming used by Stripe for passing trial days value for subscriptions/sessions.

daniellockyer (approved):

ghost/core/core/server/data/schema/schema.js
ErisDS: @allouis qq: what do you think about matching Stripe here? I am on the fence.

ghost/core/test/e2e-api/admin/__snapshots__/members-newsletters.test.js.snap
```diff
@@ -63,7 +63,7 @@ exports[`Members API - With Newsletters - compat mode Can fetch members who are
 Object {
   "access-control-allow-origin": "http://127.0.0.1:2369",
   "cache-control": "no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0",
-  "content-length": "2056",
```
ErisDS, line 66: How to tell there is no mapper in place 😂

ErisDS (approved):
Looks good to me. Have asked for Egg's opinion on naming just to check in with his current thinking.

ghost/core/test/e2e-api/admin/__snapshots__/members-newsletters.test.js.snap
rishabhgrg: 😄 Yeah that bugged me as well. I have marked it in my todos to have a look towards the end of Free trials if time permits to sort out

ghost/core/core/server/data/schema/schema.js
allouis: Given that this table is for storing Tiers which are a Ghost concept, there's no reason to match Stripe naming IMO.

However, I don't think the name is necessarily bad, but if there is a better name we should go for it!

All I can think is `trial_period` or `trial_days` which is a lil shorter 🦐

rishabhgrg: @ErisDS Do you have any pref on the naming between these options ? :)

ErisDS: `trial_days` feels nicest to me

rishabhgrg: 👌 Yeah sounds good to me too, will update, thanks.

daniellockyer (approved):

PR #15135, opened 2022-08-03
Implemented admin auth origin check

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1694

- Added replacements option to `@tryghost/minifier` + updated documentation and name of 'options' param which was a bit confusing. 
- At compile time, we'll replace `'{{SITE_ORIGIN}}'` with the actual and JS encoded origin string.
- Block requests to the auth frame with the wrong origin, but log a warning for now to make debugging easier.

**Question:**
- Are we 100% sure that the Ghost site is always hosted on the url which is configured in the site config? Because locally, it also doesn't redirect to the configured site. Technically, it is possible to host it on a different domain, no?

### DIFF
--- a/ghost/core/core/frontend/services/admin-auth-assets/service.js
+++ b/ghost/core/core/frontend/services/admin-auth-assets/service.js
@@ -44,7 +44,7 @@ class AdminAuthAssetsService {
      */
     async minify(globs, replacements) {
         try {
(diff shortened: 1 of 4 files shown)


allouis (changes requested):
We also need to pass the origin to the `postMessage` call as the second argument, rather than `'*'`

PR #15161, opened 2022-08-03
Added migration for storing trial start & end columns on subscription

rshbhgrg:

With free trials, members can start subscriptions with a trial period. We need to store the information about trial start and end for every subscription so it can be shown on Admin/Portal for member.

- adds new `trial_start_at` column for storing trial start date on Stripe subscription. Will in most cases match the start of subscription date.
- adds new `trial_end_at` column for storing trial end date on Stripe subscription.
- wires storing trial start and end values on stripe subscription


ErisDS (changes requested):
I just remembered Ghosts internal handling for ensuring dates are stored properly in the database assumes all dates are stored in fields that end with ‘_at’

ghost/members-api/lib/repositories/member.js
```diff
@@ -797,6 +797,8 @@ module.exports = class MemberRepository {
             subscription_id: subscription.id,
             status: subscription.status,
             cancel_at_period_end: subscription.cancel_at_period_end,
+            trial_start: subscription.trial_start || null,
+            trial_end: subscription.trial_end || null,
```
ErisDS, line 801: I think these need updating?

ErisDS: See above 😬

ErisDS (changes requested):

ghost/members-api/lib/repositories/member.js
rishabhgrg: 🙈 I had wiring in another PR and missed it was also pushed here originally. Updated here now, thanks.

ErisDS (approved):
Not for this PR, but later - if we could get some test coverage on ghost/members-api/lib/repositories/member.js that would be great!

daniellockyer (approved):

PR #15164, opened 2022-08-04
Added wiring for free trial on tiers

rshbhgrg:
- updates backend to handle new `trial_days` based on flag 
- updates admin to add new field to manage trial days on tier


ghost/stripe/lib/StripeAPI.js
```diff
@@ -372,6 +373,22 @@ module.exports = class StripeAPI {
         if (options.coupon) {
             discounts = [{coupon: options.coupon}];
         }
+
```
ErisDS, line 376: IMO this change should get a unit test

ErisDS (changes requested):

ghost/stripe/lib/StripeAPI.js
rishabhgrg: @ErisDS 👍 Added in follow up commits.

ErisDS (approved):

PR #15172, opened 2022-08-04
Cache adapters

naz:
nothing to see here just a CI run

### DIFF
--- a/ghost/core/core/frontend/helpers/title.js
+++ b/ghost/core/core/frontend/helpers/title.js
@@ -1,7 +1,7 @@
 // # Title Helper
 // Usage: `{{title}}`
 //
-// Overrides the standard behaviour of `{[title}}` to ensure the content is correctly escaped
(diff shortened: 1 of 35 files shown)


ghost/core/core/frontend/helpers/title.js
```diff
@@ -1,7 +1,7 @@
 // # Title Helper
 // Usage: `{{title}}`
 //
-// Overrides the standard behaviour of `{[title}}` to ensure the content is correctly escaped
+// Overrides the standard beharior of `{[title}}` to ensure the content is correctly escaped
```
ErisDS, line 4: Seems a find and replace went wrong here?

naz: Yes it did :grimacing: https://github.com/TryGhost/Ghost/commit/fa8d94fce200c915f95dfc9c9a14b967ff7aa50c

PR #15173, opened 2022-08-04
Redacted comment html for deleted and hidden comments

allouis:
refs https://github.com/TryGhost/Team/issues/1745

I've done this at the model layer, which works for now, but isn't great
if we eventually create some kind of Admin section, in which case we'll
want to expose the html of the comment, regardless of status. This is
something we can address as/when we get to it.

### DIFF
--- a/ghost/core/core/server/api/endpoints/index.js
+++ b/ghost/core/core/server/api/endpoints/index.js
@@ -226,6 +226,6 @@ module.exports = {
     },
 
     get commentsMembers() {
-        return shared.pipeline(require('./comments-members'), localUtils, 'comments');
+        return shared.pipeline(require('./comments-members'), localUtils, 'members');
     }
 };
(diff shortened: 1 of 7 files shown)


allouis:
@ErisDS @SimonBackx What do you think of redacting this in the model layer, as opposed to the API serialiser layer (in the mapper)

ErisDS:
Curious what the reason is for doing it in the model layer rather than the mapper?

As I understand, we're trying to move away from having anything other than immutable business logic in the models...  (e.g. redact password digest from user model cos no API should output it) 

I have concerns that this could inadvertently lead to us setting the html to null in the admin API for comments when we add one... although they're not at all well founded. 

To me this feels specific to the API?

allouis:
After discussion offline, I'm updating this PR to redact the html contents in the mapper

PR #15200, opened 2022-08-09
Handled null value for trial days on tiers api

rshbhgrg:
refs https://github.com/TryGhost/Ghost/commit/e26c977c6653fb870b4e29b5e6058375fbe45f21

- allows trial days to be null in admin api schema when set via tiers api
- handles null trial days in admin and API, sets trial days as 0 for null values

### DIFF
--- a/ghost/admin/app/models/tier.js
+++ b/ghost/admin/app/models/tier.js
@@ -14,6 +14,6 @@ export default Model.extend(ValidationEngine, {
     currency: attr('string'),
     monthlyPrice: attr('number'),
     yearlyPrice: attr('number'),
-    trialDays: attr('number'),
+    trialDays: attr('number', {defaultValue: 0}),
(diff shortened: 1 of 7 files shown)


allouis (approved):
Looks good - did you consider switching to using `line_items` rather than `subscription_data` when creating a checkout session? What is the Product view on continuing to support the deprecated `trail_from_plan` feature?

rishabhgrg:
@allouis Thanks.

> did you consider switching to using line_items rather than subscription_data when creating a checkout session?

Yeah we'll most likely switch to `line_items` with `trial_days` by the end of this cycle, I still have to do some more testing with `line_items` locally but I plan to push that up this week if all good 😄 

> What is the Product view on continuing to support the deprecated trail_from_plan feature?

We are happy to discontinue `trial_from_plan` with Free trials GA, so will clean this up after final confirmation towards that

PR #15202, opened 2022-08-09
Handled setting up free trial offers

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1726

- updates offer type validation to include `trial`
- add offer duration validation which includes `trial` for free trial offers
- updates offer setup to allow new `trial` as discount type, was prev only `fixed` and `percent`
- updates offer setup to allow `amount` as free trial days value
- updates offer setup to allow `trial` as discount duration value for trial offers, was prev only `once`/`forever`/`repeating`
- free trial offers don't need a stripe coupon created for them, checkout sessions for free trial offers ignore stripe coupon and directly pass the trial days value
- trial days of an offer take precedence over trial days added as default to a tier


daniellockyer (approved):

ErisDS (approved):
I've not sense-checked every piece of logic, but it looks like all the right things in all the right places to me, and a good amount of testing 👍

PR #15206, opened 2022-08-09
Added replies/likes count and limited replies to comments

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1723

- Added count.replies to comments
- Added replies endpoint
- Limited returned replies to 3.
- Replaced likes_count with count.likes in comments
- Instead of fetching all the likes of a comment to determine the total count, we'll now use count.likes
- Instead of fetching all the likes of a comment to determine whether a member liked a comment, we'll now use count.liked (which returns the amount of likes of the current member, being 0 or 1). This is mapped to `liked` to make it more natural to work with.

The `members.test.snap` file changed because we no longer include `liked: false` if we didn't fetch the liked relation. And in the comments events of the activity feed the liked property is therefore removed.

These changes requires an update to the `bookshelf-include-count` plugin:
- Updated to also work for nested relations
- This moves the count queries from the `bookshelf-include-count` plugin to the `countRelations` method of each model.
- Updated to keep the counts after saving a model (crud.edit didn't return the counts before)


SimonBackx:
Wanted to prevent having multiple versions of the same package in the repo, thats why I not only had to bump @tryghost/bookshelf-plugins but looks like this isn't possible without publishing some new version of other packages :/ Don't know what the normal flow is for this.

allouis (approved):
Looks good to me!

Do we deffo want to remove the `liked` property unless the likes relation is requested? I'm wondering if we might want it regardless 🤷 

RE: the packages, I would only bump what you need in this PR and let renovate pick up the rest :)

SimonBackx:
I'm not 100% sure what piece of code you are referring to. I think it is possible to only ask to include 'liked' and not 'likes'

Packages: I would normally do that, but I don't like duplicate versions of the same package because they tend to unexpectedly generate hard to debug bugs because you have duplicate classes and instances (requiring @tryghost/something will return a different instance or class in different parts of the code)

SimonBackx:
I think I know what you meant with your comment on `liked`. There was a small typo in the PR comment
> The `members.test.snap` file changed because we no longer include `liked: false` if we didn't fetch the **~likes~ liked** relation. And in the comments events of the activity feed the liked property is therefore removed.

PR #15223, opened 2022-08-12
✨ Added native comments

SimonBackx:
no issue

Removes the comments lab flag (bumping it to GA).

### DIFF
--- a/ghost/core/core/shared/labs.js
+++ b/ghost/core/core/shared/labs.js
@@ -15,7 +15,8 @@ const messages = {
 
 // flags in this list always return `true`, allows quick global enable prior to full flag removal
 const GA_FEATURES = [
-    'newsletterPaywall'
+    'newsletterPaywall',
+    'comments'
 ];
 
 // NOTE: this allowlist is meant to be used to filter out any unexpected
@@ -27,7 +28,6 @@ const BETA_FEATURES = [
 const ALPHA_FEATURES = [
     'urlCache',
     'beforeAfterCard',
-    'comments',
     'freeTrial'
 ];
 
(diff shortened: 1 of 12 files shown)


ghost/core/core/shared/labs.js
```diff
@@ -15,7 +15,8 @@ const messages = {
 
 // flags in this list always return `true`, allows quick global enable prior to full flag removal
 const GA_FEATURES = [
-    'newsletterPaywall'
+    'newsletterPaywall',
+    'comments'
```
allouis, line 19: We don't need this if you've removed all the checks for the flag btw, which it _looks_ like you have?

I would suggest either only making this change, or doing a thorough search for all flag usage and removing all of it, without adding this to the `GA_FEATURES` list.

SimonBackx: You are right, was a bit scared to remove it here 😅 Since we have a monorepo now, it is so much easier to remove it everywhere at once

allouis: Swedish let's do it!!

shubhank008:
@SimonBackx @allouis I think there is a problem/bug (probably related to nginx/cache) which causes logged in users to still see the "login/signup" box on the comments box UI, even if you are logged in.

For example, I am logged into the blog (members area) yet in the comments box UI I still see login/signup box and no matter how many times I login using the comments box, I am not able to.  

I suspect maybe its related to cache but cannot figure out which api/path is comments system using to ignore cache on it.

allouis:
You might have problems with cookies being set, could be your NGINX or something else - I would suggest heading over to https://forum.ghost.org and asking for help there with details about your setup!

shubhank008:
Actually sorted it out after a few hours of trial and error, cookies were being set but by nature cookies also get cached which caused issue.  
I had to create a nginx conf rule to detect the user's login cookie being set and to not cache or show cache pages for him.

PR #15237, opened 2022-08-15
Image size in memory cache

naz:
:see_no_evil:

### DIFF
--- a/ghost/core/core/server/adapters/cache/SettingsCacheSyncInMemory.js
+++ b/ghost/core/core/server/adapters/cache/SettingsCacheSyncInMemory.js
@@ -1,0 +1,7 @@
+const Memory = require('./Memory');
+
+class SettingsCacheSyncInMemory extends Memory {
+
+}
+
+module.exports = SettingsCacheSyncInMemory;
--- a/ghost/core/core/server/adapters/cache/ImageSizesCacheSyncInMemory.js
+++ b/ghost/core/core/server/adapters/cache/ImageSizesCacheSyncInMemory.js
@@ -1,0 +1,7 @@
+const Memory = require('./Memory');
+
+class ImageSizesCacheSyncInMemory extends Memory {
+
+}
+
+module.exports = ImageSizesCacheSyncInMemory;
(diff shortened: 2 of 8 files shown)


ghost/core/core/server/adapters/cache/SettingsCacheSyncInMemory.js
```diff
@@ -0,0 +1,7 @@
+const Memory = require('./Memory');
+
+class SettingsCacheSyncInMemory extends Memory {
```
ErisDS, line 3: 🤔 why do we need specific classes here and how would I customise this as a self-hoster?

naz: Having specific classes for each adapter comes from the adapter design we came up when working on storage adapters. Otherwise the adapter cache would be using same instance of the adapter if all of them declared just "Memory".

As for configurability by the self-hoster. There's nothing to configure for the current implementation, but once the caches are swapped out to be "async" and use [the cache-manager](https://www.npmjs.com/package/cache-manager) under the hood the self-hoster can potentially pass in the parameters like ttl and cache size (if we even want to expose these as configurable options).

We haven't used the multiple adapter system outside of storage. Having to create empty classes for memory caches does feel unnecessary and confusing. 

Had a thought about the maintainability of the codebase (to have less of these empty classes) and still making the optional configuration easy. Posibly we could add another convention to our adapter cache allowing for syntax like this:

```json
        "cache": {
            "active": "Memory",
            "settings": "SettingsCacheSyncInMemory",
            "imageSizes": "ImageSizesCacheSyncInMemory"
        }
```
Then, if there are no class definitions for the `SettingsCacheSyncInMemory` and `ImageSizesCacheSyncInMemory` we'd create instances for the "Memory" class and use the name itself to store the instances in the adapter cache under different keys :thinking: 

Can make an issue for the cleanup week to rework this convention if this idea sounds good. What are your thoughts @ErisDS?

naz: Got it tracked in https://github.com/TryGhost/Toolbox/issues/384 for the cleanup week.

naz:
landed with e4d02657ee3eafee2af43192eea950b5901df7a6.

PR #15240, opened 2022-08-15
Member Attribution DB & Models

allouis:
refs https://github.com/TryGhost/Team/issues/1803
refs https://github.com/TryGhost/Team/issues/1802

I've left out the migrations here, because we are still waiting on whether or not we should use cascadeDelete. We can still use/merge this however as it's behind an alpha flag and the migrations will add the tables before we go to beta.

If you wanna use this locally, just start Ghost with a fresh DB to get the tables added.

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.10/2022-08-16-14-25-add-member-created-events-table.js
+++ b/ghost/core/core/server/data/migrations/versions/5.10/2022-08-16-14-25-add-member-created-events-table.js
@@ -1,0 +1,11 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('members_created_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    created_at: {type: 'dateTime', nullable: false},
+    member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
+    attribution_id: {type: 'string', maxlength: 24, nullable: true},
+    attribution_type: {type: 'string', maxlength: 50, nullable: true},
+    attribution_url: {type: 'string', maxlength: 2000, nullable: true},
+    source: {type: 'string', maxlength: 50, nullable: false}
+});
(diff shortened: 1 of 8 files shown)


ErisDS:
> I've left out the migrations here, because we are still waiting on whether or not we should use cascadeDelete. We can still use/merge this however as it's behind an alpha flag and the migrations will add the tables before we go to beta.

> If you wanna use this locally, just start Ghost with a fresh DB to get the tables added.

Really like this approach 👌

SimonBackx (approved):

SimonBackx (approved):

ghost/core/core/server/models/subscription-created-event.js
```diff
@@ -0,0 +1,30 @@
+const errors = require('@tryghost/errors');
+const ghostBookshelf = require('./base');
+
+const SubscriptionCreatedEvent = ghostBookshelf.Model.extend({
+    tableName: 'member_created_events',
```
SimonBackx, line 5:
```suggestion
    tableName: 'subscription_created_events',
```

ghost/core/core/server/data/migrations/versions/5.10/2022-08-16-14-25-add-member-created-events-table.js
```diff
@@ -0,0 +1,11 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('member_created_events', {
```
daniellockyer, line 3: Is there a reason we've gone for the singular version of members?

SimonBackx (approved):

ghost/core/core/server/data/migrations/versions/5.10/2022-08-16-14-25-add-member-created-events-table.js
allouis: Because the word member here is part of the event name, rather than the relation - I can rename it though.

Should we also change it to subscriptions_created_events, or leave that singular?

daniellockyer (approved):

PR #15241, opened 2022-08-16
Added expiry column for complimentary subscriptions

rshbhgrg:

- allows adding expiry date to complimentary subscriptions, effectively allowing cardless trials.
- expiry is stored in members<>products mapping table


allouis:
How does this work with [bookshelf-relations](https://github.com/tryghost/bookshelf-relations)? Is there a way to add extra attributes, or will we have to insert these rows manually?

rishabhgrg:
@allouis Good Q, i was looking into this yesterday, and the best way i found to update the `expiry_at` column on updating a member is using `updatePivot` at the time of save, have added the initial implementation for it [here](https://github.com/TryGhost/Ghost/pull/15245/files#diff-fe6ed3439b1ddd35205200c77c9f1d6a0356377dfca0359245e40554f56895f6R378-R387)

daniellockyer (approved):

ErisDS (approved):

PR #15242, opened 2022-08-16
Added member attribution script

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1804

- Adds a script that is only injected when the member attribution alpha flag is enabled
- This script builds a history and saves it in localStorage as `ghost-history` that contains something like this: 
``` json
[
    {
       "time": 1660650730,
        "path": "/about/"
   },
   {
        "time": 1660651730,
       "path": "/welcome/"
    }
]
```

- Keeps track of the time of every page visit, so we can correctly remove old items. I also considered saving the time separately and clearing the whole history when the saved time is older than 24h, but that would have the side effect that items older than 24h might leak into the history if you visit every 12 hours (to give an example). Plus, having objects in the history might make it easier to add other attributes to the items if we ever want to do that in the future. We also have access to the time between visits.
- Added `.eslintrc` configuration for this new frontend script. This makes it easier to spot errors when developing, and follow the same syntax rules as other scripts. In the future it can allow us to require an older ECMA version in the browser script. If we like this pattern, we could also use it for other frontend scripts.

### DIFF
--- a/ghost/core/core/frontend/services/member-attribution-assets/service.js
+++ b/ghost/core/core/frontend/services/member-attribution-assets/service.js
@@ -1,0 +1,83 @@
+// const debug = require('@tryghost/debug')('comments-counts-assets');
+const Minifier = require('@tryghost/minifier');
+const path = require('path');
+const fs = require('fs').promises;
+const logging = require('@tryghost/logging');
+const config = require('../../../shared/config');
+
+class MemberAttributionAssetsService {
+    constructor(options = {}) {
+        /** @private */
+        this.src = options.src || path.join(config.get('paths').assetSrc, 'member-attribution');
+        /** @private */
+        this.dest = options.dest || config.getContentPath('public');
+        /** @private */
+        this.minifier = new Minifier({src: this.src, dest: this.dest});
+    }
+
+    /**
+     * @private
+     */
+    generateGlobs() {
+        return {
+            'member-attribution.min.js': '*.js'
+        };
+    }
+
+    /**
+     * @private
+     */
+    generateReplacements() {
+        return {};
+    }
+
+    /**
+     * @private
+     * @returns {Promise<void>}
+     */
+    async minify(globs, options) {
+        try {
+            await this.minifier.minify(globs, options);
+        } catch (error) {
+            if (error.code === 'EACCES') {
+                logging.error('Ghost was not able to write member-attribution asset files due to permissions.');
+                return;
+            }
+
+            throw error;
+        }
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


ghost/core/core/frontend/web/site.js
```diff
@@ -75,6 +75,9 @@ module.exports = function setupSiteApp(routerConfig) {
     // Comment counts
     siteApp.use(mw.servePublicFile('built', 'public/comment-counts.min.js', 'application/javascript', constants.ONE_YEAR_S));
 
+    // Comment counts
```
iBotPeaches, line 78: Looks like a copy/paste. This comment doesn't match below.

SimonBackx:
```suggestion
    // Member attribution
```

ghost/core/core/frontend/src/member-attribution/member-attribution.js
```diff
@@ -0,0 +1,90 @@
+// Location where we want to store the history in localStorage
+const STORAGE_KEY = 'ghost-history';
+
+// How long before an item should expire (24h)
+const TIMEOUT = 24 * 60 * 60 * 1000;
+
+// Maximum amount of urls in the history
+const LIMIT = 15;
```
allouis, line 8: How did you come to this number?

allouis (approved):

ghost/core/core/frontend/services/member-attribution-assets/service.js
```diff
@@ -0,0 +1,83 @@
+// const debug = require('@tryghost/debug')('comments-counts-assets');
```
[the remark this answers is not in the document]
SimonBackx: I agree, I've created a new tech debt issue for this: https://github.com/TryGhost/Team/issues/1807

ghost/core/core/frontend/src/member-attribution/member-attribution.js
SimonBackx: For now, this is arbitrary. I think 15 will be more than enough, or rather too much. We can later update this number after we've come up with our Last Post Algorithm™️. I think that will give us more information to determine this number. What do you think?

allouis: Okay cool - sounds good to me, pretty easy to change in future!

My initial thoughts were that it might be too small, like you could easily visit 15 pages in under 24 hours 🤔

Thinking about it more though - the chances of none of them being a Post or a Page is small, and that's all we really care about atm

PR #15243, opened 2022-08-16
Added member attribution events and storage

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1808
refs https://github.com/TryGhost/Team/issues/1809
refs https://github.com/TryGhost/Team/issues/1820
refs https://github.com/TryGhost/Team/issues/1814

### Changes in `member-events` package

- Added MemberCreatedEvent (event, not model)
- Added SubscriptionCreatedEvent (event, not model) 

### Added `member-attribution` package (new)

- Added the AttributionBuilder class which is able to convert a url history to an attribution object (exposed as getAttribution on the service itself, which handles the dependencies)
```
[{
    "path": "/",
    "time": 123
}]
```
to
```
{
    "url": "/",
    "id": null,
    "type": "url"
}
```

- event handler listens for MemberCreatedEvent and SubscriptionCreatedEvent and creates the corresponding models in the database.

### Changes in `members-api` package

- Added urlHistory to `sendMagicLink` endpoint body + convert the urlHistory to an attribution object that is stored in the tokenData of the magic link (sent by Portal in this PR: https://github.com/TryGhost/Portal/pull/256).
- Added urlHistory to `createCheckoutSession` endpoint + convert the urlHistory to attribution keys that are saved in the Stripe Session metadata (sent by Portal in this PR: https://github.com/TryGhost/Portal/pull/256).

- Added attribution data property to member repository's create method (when a member is created)
- Dispatch MemberCreatedEvent with attribution

###  Changes in `members-stripe-service` package (`ghost/stripe`)

- Dispatch SubscriptionCreatedEvent in WebhookController on subscription checkout (with attribution from session metadata)

This PR builds on top of https://github.com/TryGhost/Ghost/pull/15240

### DIFF
--- a/ghost/member-attribution/lib/event-handler.js
+++ b/ghost/member-attribution/lib/event-handler.js
@@ -1,0 +1,39 @@
+const {MemberCreatedEvent, SubscriptionCreatedEvent} = require('@tryghost/member-events');
+
+class MemberAttributionEventHandler {
+    constructor({MemberCreatedEvent: MemberCreatedEventModel, SubscriptionCreatedEvent: SubscriptionCreatedEventModel, DomainEvents}) {
+        this._MemberCreatedEventModel = MemberCreatedEventModel;
+        this._SubscriptionCreatedEvent = SubscriptionCreatedEventModel;
+        this.DomainEvents = DomainEvents;
+    }
+
+    subscribe() {
+        this.DomainEvents.subscribe(MemberCreatedEvent, async (event) => {
+            const attribution = event.data.attribution;
+
+            await this._MemberCreatedEventModel.add({
+                member_id: event.data.memberId,
+                created_at: event.timestamp,
+                attribution_id: attribution?.id ?? null,
+                attribution_url: attribution?.url ?? null,
+                attribution_type: attribution?.type ?? null,
+                source: event.data.source
+            });
+        });
+
+        this.DomainEvents.subscribe(SubscriptionCreatedEvent, async (event) => {
+            const attribution = event.data.attribution;
+
+            await this._SubscriptionCreatedEvent.add({
+                member_id: event.data.memberId,
+                subscription_id: event.data.subscriptionId,
+                created_at: event.timestamp,
+                attribution_id: attribution?.id ?? null,
+                attribution_url: attribution?.url ?? null,
+                attribution_type: attribution?.type ?? null
+            });
+        });
+    }
+}
+
+module.exports = MemberAttributionEventHandler;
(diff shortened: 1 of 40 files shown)


ghost/member-attribution/lib/event-handler.js
```diff
@@ -0,0 +1,40 @@
...
+            const attribution = event.data.attribution;
+
+            await this._SubscriptionCreatedEvent.add({
+                member_id: event.data.memberId,
+                subscription_id: event.data.subscriptionId,
+                created_at: event.timestamp,
+                attribution_id: attribution?.id ?? null,
+                attribution_url: attribution?.url ?? null,
+                attribution_type: attribution?.type ?? null,
+                source: event.data.source
```
allouis, line 34: We don't need `source` here

ghost/member-attribution/lib/event-handler.js
```diff
@@ -0,0 +1,40 @@
+const {MemberCreatedEvent, SubscriptionCreatedEvent} = require('@tryghost/member-events');
+
+class MemberAttributionEventHandler {
```
allouis, line 3: I don't think this should be part of the Member Attribution service, there's nothing here specific to attribution 🤔 

Long term, we should be emitting events from the members package, and then listening to them and storing them just like we are here.

I would have a `MemberEventService` somewhere which literally just listens to events and stores them, and then the attribution service is just used to convert history -> attribution.

That would keep all of the event storage logic in one place.

Having said that, I don't want to block this PR on it

allouis (approved):

ghost/member-attribution/lib/event-handler.js
SimonBackx: True, I think it would be nice to have all the event handlers in one place. Initially I did have a small dependency on the attribution package because originally the urlHistory was inside the events (not the attribution object).

Just to be sure, do you mean moving this code to the existing `members-events-service`, just like the `last-seen-at-updater`?

ghost/member-attribution/lib/event-handler.js
SimonBackx: Good catch, thanks! 🙌

ghost/member-attribution/lib/event-handler.js
allouis: Yeah I think the `members-events-service` would work, but I think we can create a generic "Event storage" or "Event persistence" handler, because it's not going to be running logic, it's going to be just a series of:
```js
DomainEvents.subscribe(Event, (event) => {
  Model.add(convertEventToModelData(event));
});
```

SimonBackx: Created a cleanup issue for this: https://github.com/TryGhost/Team/issues/1821

PR #15245, opened 2022-08-17
Handled expiring complimentary subscriptions

rshbhgrg:

- adds method to get active products for a member to exclude expired comped subs
- wires member add/edit to update expiry of a comped sub
- updates admin ui to allow adding expiry

### DIFF
--- a/ghost/core/core/server/services/members/jobs/clean-expired-comped.js
+++ b/ghost/core/core/server/services/members/jobs/clean-expired-comped.js
@@ -1,0 +1,105 @@
+const {parentPort} = require('worker_threads');
+const ObjectId = require('bson-objectid').default;
+const {chunk: chunkArray} = require('lodash');
+const debug = require('@tryghost/debug')('jobs:clean-expired-comped');
+const moment = require('moment');
+
+// recurring job to clean expired complimentary subscriptions
+
+// Exit early when cancelled to prevent stalling shutdown. No cleanup needed when cancelling as everything is idempotent and will pick up
+// where it left off on next run
+function cancel() {
+    if (parentPort) {
+        parentPort.postMessage('Expired complimentary subscriptions cleanup cancelled before completion');
+        parentPort.postMessage('cancelled');
+    } else {
+        setTimeout(() => {
+            process.exit(0);
+        }, 1000);
+    }
+}
+
+if (parentPort) {
+    parentPort.once('message', (message) => {
+        if (message === 'cancel') {
+            return cancel();
+        }
+    });
+}
+
+(async () => {
+    const cleanupStartDate = new Date();
+    const db = require('../../../data/db');
+    debug(`Starting cleanup of expired comp subscriptions`);
+    const expiredCompedRows = await db.knex('members_products')
+        .where('expiry_at', '<', moment.utc().startOf('day').toISOString())
+        .select('*');
+
+    let deletedExpiredSubs = 0;
+    let updatedMembers = 0;
+
+    // Run cleanup for expired comp subscriptions
+    // Removes expired comped entries from members_products table
+    // Updates affected members status to free from comped
+    // Adds member status event for going from comped to free
+    if (expiredCompedRows?.length) {
+        const rowIds = expiredCompedRows.map(d => d.id);
+        const memberIds = expiredCompedRows.map(d => d.member_id);
+
+        // Delete all expired comped rows
+        deletedExpiredSubs = await db.knex('members_products')
+            .whereIn('id', rowIds)
+            .del();
+
+        // fetch all comped members to update
+        const membersToUpdate = await db.knex('members')
+            .whereIn('id', memberIds)
+            .andWhere('status', 'comped');
+
+        const updateMemberIds = membersToUpdate.map(d => d.id);
+
+        // Update all comped members to free
+        updatedMembers = await db.knex('members')
+            .whereIn('id', updateMemberIds)
+            .update({
+                status: 'free'
+            });
+
+        const statusEvents = membersToUpdate.map((member) => {
+            const now = db.knex.raw('CURRENT_TIMESTAMP');
+
+            return {
+                id: ObjectId().toHexString(),
+                member_id: member.id,
+                from_status: member.status,
+                to_status: 'free',
+                created_at: now
+            };
+        });
+
+        // SQLite >= 3.32.0 can support 32766 host parameters
+        // each row uses 5 variables so ⌊32766/5⌋ = 6553
+        const chunkSize = 6553;
+
+        const chunks = chunkArray(statusEvents, chunkSize);
+
+        // Adds status event for members going comped->free
+        for (const chunk of chunks) {
+            await db.knex('members_status_events').insert(chunk);
+        }
+    }
+
+    let cleanupEndDate = new Date();
+
+    debug(`Removed ${deletedExpiredSubs} expired subscriptions, updated ${updatedMembers} members in ${cleanupEndDate.valueOf() - cleanupStartDate.valueOf()}ms`);
+
+    if (parentPort) {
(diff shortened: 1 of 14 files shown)


ghost/core/core/server/services/members/jobs/clean-expired-comped.js
```diff
@@ -0,0 +1,71 @@
...
+
+if (parentPort) {
+    parentPort.once('message', (message) => {
+        if (message === 'cancel') {
+            return cancel();
+        }
+    });
+}
+
+(async () => {
```
naz, line 28: While we are at it, @rishabhgrg can you please check if we still need this self-execution function syntax. AFAIR our current range of Node versions supports top level await :thinking:

ghost/core/core/server/services/members/jobs/clean-expired-comped.js
```diff
@@ -0,0 +1,71 @@
...
+    const cleanupStartDate = new Date();
+    const db = require('../../../data/db');
+    debug(`Starting cleanup of expired comp subscriptions on ${cleanupStartDate}`);
+    const expiredCompedRows = await db.knex('members_products')
+        .where('expiry_at', '<', moment.utc().startOf('day').toISOString())
```
naz, line 33: Is the logic here to expire products from yesterday and below or should include the ones set to expire at midnight (5 minutes before the run of this job)? in other words should it be `<` or `<=` :thinking:

ghost/core/core/server/services/members/jobs/clean-expired-comped.js
```diff
@@ -0,0 +1,71 @@
...
+    // Run cleanup for expired comp subscriptions
+    // Removes expired entries and updates affected member statuses
+    if (expiredCompedRows?.length) {
+        const rowIds = expiredCompedRows.map(d => d.id);
+        const memberIds = expiredCompedRows.map(d => d.member_id);
+
+        // Delete all expired comped rows
+        deletedExpiredSubs = await db.knex('members_products')
+            .whereIn('id', rowIds)
+            .del();
```
naz, line 48: by removing these records my understanding is we don't want to know in the future if the member ever has been on the "expiring comped" plan?

ghost/core/core/server/services/members/jobs/jobs.js
```diff
@@ -0,0 +1,27 @@
...
+            const s = Math.floor(Math.random() * 60); // 0-59
+
+            // Run everyday at 12:05:X AM to clean all expired complimentary subscriptions
+            jobsService.addJob({
+                at: `${s} 5 0 * * *`,
+                job: path.resolve(__dirname, 'clean-expired-comped.js'),
+                name: 'clean-expired-comped'
+            });
+
+            hasScheduled = true;
```
naz, line 22: The method only ever runs once in the Ghost process lifecycle - during the boot during `members.init()` execution. Why do we need this protective code from running the method multiple times? If it is run multiple times does it signal something wrong is going on?

ghost/core/core/server/services/members/service.js
```diff
@@ -6,6 +6,7 @@ const db = require('../../data/db');
 const MembersConfigProvider = require('./config');
 const MembersCSVImporter = require('@tryghost/members-importer');
 const MembersStats = require('./stats/members-stats');
+const memberJobs = require('./jobs/jobs');
```
naz, line 9: I think leaving the pattern that exists in the `analytics` jobs (index.js file in the root) would be best for consistency sake. It should not be full of logic though :D

ghost/core/core/server/services/members/jobs/clean-expired-comped.js
```diff
@@ -0,0 +1,72 @@
...
+                status: 'free'
+            });
+    }
+
+    let cleanupEndDate = new Date();
+
+    debug(`Removed ${deletedExpiredSubs} expired subscriptions, updated ${updatedMembers} members in ${cleanupEndDate.valueOf() - cleanupStartDate.valueOf()}ms`);
+
+    if (parentPort) {
+        parentPort.postMessage(`Removed ${deletedExpiredSubs} expired subscriptions, updated ${updatedMembers} members in ${cleanupEndDate.valueOf() - cleanupStartDate.valueOf()}ms`);
```
naz, line 64: I'm thinking if we would end up with logs full of `Removed 0 expired subscriptions...` on majority of instances :thinking: It is probably fine in the beginning. How do you think we should handle it if that's the case?

naz (approved):
Did look through job-related part of the change. Left a few clarification questions. For the most part looks :ok_hand: :ok_hand: :ok_hand:

ghost/core/core/server/services/members/jobs/clean-expired-comped.js
rishabhgrg: > Is the logic here to expire products from yesterday and below

It should be this, basically any sub whose expiry date is already in past and doesn't include today should be expired, as the ones that are expiring midnight means they still have access to tier till today.

ghost/core/core/server/services/members/jobs/clean-expired-comped.js
rishabhgrg: Yeah don't think we have any usecase to keep that info right now, but that did strike a good point to add member status event update as part of this job, so we do have a record of member's status update in the DB when they go from comped to free. 💡 Adding that now

ghost/core/core/server/services/members/jobs/jobs.js
rishabhgrg: 🤔 So i picked that pattern from email analytics, which I get has this check because it is run either on init or sometime down the line based on if emails need to be fetched. I am hoping comp check will also move in a similar direction maybe in the cleanup, so i kept the pattern intact as there is no reason to not have it imo. 
I don't think it will be run multiple times right now so i am ok either ways on keeping or removing this check, but also i don't think there is any harm in keeping it as well :)

ghost/core/core/server/services/members/service.js
rishabhgrg: I wanted to keep away from overloading `index.js` file as we generally avoid them, no strong prefs tho so happy to change it back to index here :)

ghost/core/core/server/services/members/jobs/clean-expired-comped.js
rishabhgrg: 🤔 Yea that's a fair point, we could remove the logs for when nothing happens in future if that becomes annoying maybe and only output stuff when any changes are done ? Even with optimisations, don't think we can completely avoid not running jobs when there are no expiring subs, so there will always be some case where it runs but doesn't do anything

ghost/core/core/server/services/members/jobs/clean-expired-comped.js
rishabhgrg: From a quick check, top-level await seems to be still [only available in ES modules so we'll probably have to keep this style for now, but I'll come back to it later today to make sure that's the case.

ghost/core/core/server/services/members/jobs/clean-expired-comped.js
rishabhgrg: Done, added logic to insert member status events based on members going from comped -> free. We ofc still lose the expiry information, but don't think we need to keep that

ghost/core/core/server/services/members/service.js
rishabhgrg: Updated back to `index.js` ✔️

ErisDS (approved):
If @naz is happy, I am happy :)

ghost/core/core/server/services/members/jobs/clean-expired-comped.js
naz: oh this is a neat workaround!

PR #15249, opened 2022-08-17
Fixed odd behavior with saving complimentary members

rshbhgrg:
closes https://github.com/TryGhost/Team/issues/1715
refs https://github.com/TryGhost/Members/pull/407

- when comped boolean was brought back for backwards compatibility, it accidentally opened a path from admin to set comped stripe subscriptions on members on save, as Admin always sends back `comped` value
- since we don't allow stripe comped subscriptions anymore, this caused weird behavior with assigning the default tier on comped stripe subscription.
- this change removes the `json.comped` value to be sent up from Admin to avoid any weird comped changes to the member


rishabhgrg:
Closed with https://github.com/TryGhost/Ghost/commit/f00004822310706e000477afcffc494f4ca0e346

PR #15266, opened 2022-08-19
Added member attribution to member details page

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1817

- This PR adds the attribution property when you fetch a single member.
- Added `attribution` which returns null or the AttributionResource (same as attribution but with a title and up to date url)
- Added `subscriptions[x].attribution` that contains the attribution resource for each subscription
- Implemented these new properties in admin and make them visible (needs more design improvements, but just visible for now)

These properties are only added when the `memberAttribution` labs flag is enabled. We might swap this with an option in the future.

### DIFF
--- a/ghost/member-attribution/lib/attribution.js
+++ b/ghost/member-attribution/lib/attribution.js
@@ -1,9 +1,70 @@
 /**
- * @typedef {object} Attribution
- * @prop {string|null} [id]
- * @prop {string|null} [url]
- * @prop {string} [type]
+ * @typedef {object} AttributionResource
+ * @prop {string|null} id
+ * @prop {string|null} url
+ * @prop {'page'|'post'|'author'|'tag'|'url'} type
+ * @prop {string|null} title
  */
+
+class Attribution {
+    #urlTranslator;
+
+    /**
+     * @param {object} data
+     * @param {string|null} [data.id]
+     * @param {string|null} [data.url]
+     * @param {'page'|'post'|'author'|'tag'|'url'} [data.type]
+     */
+    constructor({id, url, type}, {urlTranslator}) {
+        /** @type {string|null} */
+        this.id = id;
+        
+        /** @type {string|null} */
+        this.url = url;
+
+        /** @type {'page'|'post'|'author'|'tag'|'url'} */
+        this.type = type;
+
+        /**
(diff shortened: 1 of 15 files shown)


ghost/member-attribution/lib/attribution.js
```diff
@@ -1,10 +1,71 @@
...
+    #urlTranslator;
+
+    /**
+     * @param {object} data
+     * @param {string|null} [data.id]
+     * @param {string|null} [data.url]
+     * @param {'page'|'post'|'author'|'tag'|'url'} [data.type]
+     */
+    constructor({id, url, type}, {urlTranslator}) {
+        /** @type {string|null} */
```
allouis, line 19: We shouldn't need these as the types can be inferred by the `@param` tags above - no harm in having them tho!

allouis (approved):

ghost/member-attribution/lib/service.js
```diff
@@ -5,11 +5,19 @@ const AttributionBuilder = require('./attribution');
...
-    constructor({MemberCreatedEvent, SubscriptionCreatedEvent, urlService, labsService}) {
+    constructor({Post, User, Tag, MemberCreatedEvent, SubscriptionCreatedEvent, urlService, labsService}) {
         const eventHandler = new MemberAttributionEventHandler({MemberCreatedEvent, SubscriptionCreatedEvent, DomainEvents, labsService});
         eventHandler.subscribe();
 
-        const urlTranslator = new UrlTranslator({urlService});
+        this.urlService = urlService;
+        this.models = {MemberCreatedEvent, SubscriptionCreatedEvent};
+
+        const urlTranslator = new UrlTranslator({
```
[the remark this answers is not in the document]
SimonBackx: Should be fixed/improved in https://github.com/TryGhost/Ghost/commit/02168b41ceb5b75005584b791b80d5e6beaff11a

ghost/member-attribution/lib/attribution.js
SimonBackx: Fixed in https://github.com/TryGhost/Ghost/commit/02168b41ceb5b75005584b791b80d5e6beaff11a

PR #15274, opened 2022-08-22
🐛 Fixed empty error csv file for member imports

rshbhgrg:
closes https://github.com/TryGhost/Team/issues/1828

- members importer was sending empty error files in case of invalid rows in CSV, hiding both error and affected rows
- fixes typo in `content` option passed to attachment (ref - https://nodemailer.com/message/attachments )

### DIFF
--- a/ghost/members-importer/lib/importer.js
+++ b/ghost/members-importer/lib/importer.js
@@ -259,6 +259,34 @@ module.exports = class MembersCSVImporter {
     }
 
     /**
+     * Send email with attached CSV containing error rows info
+     *
+     * @param {Object} config
+     * @param {String} config.emailRecipient - email recipient for error file
+     * @param {String} config.emailSubject - email subject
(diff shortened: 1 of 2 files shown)


allouis (approved):
Was a little tricky to find the fix in this changeset! I think a more minimal fix would be a one-line change to fix the `content` key, which would make seeing the fix a lot easier! Was there a particular reason you pulled this out into a new method?

rishabhgrg:
@allouis Ah yea sorry I should have flaged the change in commit as well. The reason to extract it out into a separate fn was to be able to write a unit test for it, as the current send method was inside a job which made writing the test very tricky.

allouis:
Ah okay, I think a full test would have made more sense, we can stub the `_addJob` method etc... and check the email is sent correctly, whereas this just checks the new method, but not that the method is used anywhere!

PR #15276, opened 2022-08-22
Added email alert notification fields for user

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1825

- adds 3 new columns to users table for storing email alert preferences for member signups/cancellation
- adds column for new member signup alert
- adds column for paid subscription started alert
- adds column for paid subscription canceled alert

Migrations [here](https://github.com/TryGhost/Ghost/pull/15276/commits/03cb88427c6b615e10e0de53f00642544cee64f0)


ghost/core/core/server/data/migrations/versions/5.11/2022-08-22-11-03-add-free-member-signup-notif-column-to-users.js
```diff
@@ -0,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('users', 'free_member_signup_notification', {
+    type: 'boolean',
+    nullable: false,
+    defaultTo: false
```
ErisDS, line 6: Why are we defaulting to false?

ErisDS (commented):
I have some questions :)

ghost/core/core/server/data/migrations/versions/5.11/2022-08-22-11-03-add-free-member-signup-notif-column-to-users.js
rishabhgrg: For existing sites, we want all the notifications to be switched off by default, and we'll notify via changelog if anyone wants to turn them on. So we are defaulting to false on all migrations.

For new sites, we only want the owner to have the notifications on by default during setup, everyone else will by default have them as off.

ErisDS: Am following up with @prschulz about this, I think we are going to change it to on by default.

ghost/core/core/server/data/migrations/versions/5.11/2022-08-22-11-03-add-free-member-signup-notif-column-to-users.js
```diff
@@ -0,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('users', 'free_member_signup_notification', {
```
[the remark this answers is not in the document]
daniellockyer: It seems we've allowed both in recent releases. The guidelines only say "minimal changeset" but that's for the PR as a whole.

I think the shorter ones tends to get lumped into one migration, but those with more columns get their own. I'm easy either way but we could probably combine this into one migration for now

ghost/core/core/server/data/migrations/versions/5.11/2022-08-22-11-03-add-free-member-signup-notif-column-to-users.js
rishabhgrg: Ah ok, in that case, I guess we'll still need to keep it as `false` in this migration, and then add another migration that goes over all `owner`(and `administrator` if we want them) users only and toggles them to `true`.
We do have a check in email logic to ignore non owner/admin users irrespective of their value for this column so all users having `true` value won't affect final email sent, but guess it still makes sense to set the right values for users from start.

ghost/core/core/server/data/migrations/versions/5.11/2022-08-22-11-03-add-free-member-signup-notif-column-to-users.js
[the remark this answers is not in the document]
rishabhgrg: Ah cool, i'll lump this together in that case, thanks @daniellockyer

ghost/core/core/server/data/migrations/versions/5.11/2022-08-22-11-03-add-free-member-signup-notif-column-to-users.js
ErisDS: I really think these should be true by default at the database layer. That migration is just unnecessary extra work.

rishabhgrg: Cool, will update the migrations and default values to true. 👍

ErisDS: Note: I'm still discussing with Peter and haven't got absolute confirmation.

ErisDS (approved):

daniellockyer (approved):

PR #15286, opened 2022-08-23
Implemented `context` on Actions events

daniellockyer:
refs https://github.com/TryGhost/Toolbox/issues/356

- in order to show data that we might not necessarily still have around
  (ie. when you delete a post, you might want the title), we're going to
  start utilizing the `context` column
- right now, we store the `primary_name` for deleted events, and we also
  store the `setting` `key` and `group` so we can reference it in the
  audit log

### DIFF
--- a/ghost/admin/app/helpers/audit-log-event-fetcher.js
+++ b/ghost/admin/app/helpers/audit-log-event-fetcher.js
@@ -82,6 +82,10 @@ export default class AuditLogEventFetcher extends Resource {
                 this.hasReachedEnd = true;
             }
 
+            actions.forEach((a) => {
+                a.context = JSON.parse(a.context);
+            });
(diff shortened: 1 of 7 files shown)


ghost/admin/app/helpers/audit-log-event-fetcher.js
```diff
@@ -82,6 +82,10 @@ export default class AuditLogEventFetcher extends Resource {
                 this.hasReachedEnd = true;
             }
 
+            actions.forEach((a) => {
```
daniellockyer, line 85: At this point, I should probably switch to using the model

daniellockyer: Will update when I have the chance

PR #15294, opened 2022-08-23
Added backfill migration for members created events

SimonBackx:
closes https://github.com/TryGhost/Team/issues/1836

- Uses the timestamps from the members table to determine the timestamps for the events
- Clears the table when downgrading to prevent having multiple rows for the same member

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-41-backfill-members-created-events.js
+++ b/ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-41-backfill-members-created-events.js
@@ -8,6 +8,11 @@ module.exports = createTransactionalMigration(
         const members = await knex('members')
             .select('id', 'created_at');
 
+        if (members.length === 0) {
+            logging.warn(`Skipping migration because no members found`);
+            return;
+        }
+
         const toInsert = members.map((member) => {
             return {
                 id: ObjectID().toHexString(),
--- a/ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-41-backfill-members-created-events.js
+++ b/ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-41-backfill-members-created-events.js
@@ -8,23 +8,20 @@ module.exports = createTransactionalMigration(
         const members = await knex('members')
             .select('id', 'created_at');
 
-        const toInsert = [];
-
-        // eslint-disable-next-line no-restricted-syntax
-        for (const member of members) {
-            toInsert.push({
+        const toInsert = members.map((member) => {
+            return {
                 id: ObjectID().toHexString(),
                 member_id: member.id,
                 created_at: member.created_at,
                 source: 'member'
-            });
-        }
+            };
+        });
 
         logging.info(`Inserting ${toInsert.length} members created events`);
         await knex.batchInsert('members_created_events', toInsert);
     },
     async function down(knex) {
         logging.info(`Clearing all members created events`);
-        await knex('members_created_events').truncate();
+        await knex('members_created_events').del();
     }
 );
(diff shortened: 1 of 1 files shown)


ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-41-backfill-members-created-events.js
```diff
@@ -0,0 +1,30 @@
...
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const members = await knex('members')
+            .select('id', 'created_at');
+
+        const toInsert = [];
+
+        // eslint-disable-next-line no-restricted-syntax
+        for (const member of members) {
```
allouis, line 14: @daniellockyer Are you okay with this loop?

daniellockyer: Yeah it should be ok, but would it also be better to use `.map` here?

ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-41-backfill-members-created-events.js
```diff
@@ -0,0 +1,30 @@
...
+                source: 'member'
+            });
+        }
+
+        logging.info(`Inserting ${toInsert.length} members created events`);
+        await knex.batchInsert('members_created_events', toInsert);
+    },
+    async function down(knex) {
+        logging.info(`Clearing all members created events`);
+        await knex('members_created_events').truncate();
```
allouis, line 28: We usually use `.del()` for this kind of operation - though `.truncate()` is more correct - do you know if there's a perf improvement with this? NABD either way IMO

SimonBackx: Truncate should be a little bit faster, don't think we'll notice :)

SimonBackx: But truncate doesn't support rollbacks, we might need that?

ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-41-backfill-members-created-events.js
allouis: Don't think it'll make a difference RE: perf, that's just a style thing AFAIC - what do you prefer?

ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-41-backfill-members-created-events.js
allouis: If that's the case, let's stick with `del()` for now :)

ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-41-backfill-members-created-events.js
SimonBackx: I've swapped it, `.map` does make sense here as it avoids all the pushes :)

ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-41-backfill-members-created-events.js
```diff
@@ -0,0 +1,27 @@
+const ObjectID = require('bson-objectid').default;
+const logging = require('@tryghost/logging');
+
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const members = await knex('members')
```
allouis, line 8: I'll add an early exit here in the case of no members

allouis:
@daniellockyer I think this is ready for final review

daniellockyer (approved):

PR #15296, opened 2022-08-23
Fixed Action event `resource_type` from `page` to `post`

daniellockyer:
refs https://github.com/TryGhost/Toolbox/issues/356
refs https://github.com/TryGhost/Ghost/commit/3a9016639cb58759bbe852859d64077414cc5a48

- I misunderstood the purpose of a column, and changed the values that
  are inserted into it, which broke relation includes in Bookshelf
- I've since reverted that in the commit above but this migration is to
  fixup the data that got stored in the DB
- we want to replace `resource_type` = `page` back to `post`, but then
  use the `context` column as described in the referenced commit to
  store that the type is actually a `page`, so we can link to it
  from the audit log accordingly
- I'm overwriting the `context` column without taking into account the
  current contents but that's ok because this bug existed before we
  started using `context`

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-59-fix-page-resource-type.js
+++ b/ghost/core/core/server/data/migrations/versions/5.11/2022-08-23-13-59-fix-page-resource-type.js
@@ -1,0 +1,22 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info(`Changing Action event 'page' resource_type to 'post'`);
+
+        const affectedRows = await knex('actions')
+            .update({
+                resource_type: 'post',
+                context: JSON.stringify({
+                    type: 'page'
+                })
+            })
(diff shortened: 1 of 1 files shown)


ErisDS (approved):

allouis (approved):
Where is the join based on the `resource_type` made?

This obviously fixes things, but it does leak DB implementation into our data layer, it might have been preferable to modify the code where we do the join (which has to know about DB implementation by nature) to handle joining to the correct table based on the resource type.

NABD tbh :+1:

daniellockyer:
@allouis Yeah so we use Bookshelf's `morphTo` here: https://github.com/TryGhost/Ghost/blob/5934794d86303b7f64958582165471762929f131/ghost/core/core/server/models/action.js#L20-L22 and it does some auto-magical joining underneath

The way I am approaching this is that it used to work, then I broke it (doing `include=resource` causes the endpoint to 500), so now I need to fix it, especially so I can use it in the Audit Log project. Whether or not this is a good method is another question for when we consider our use of the ORM in general

allouis:
Cool thanks!

Yeah fo' sure wasn't expecting any changes ☺️

PR #15302, opened 2022-08-24
[WIP] Handled tier expiry in members importer/exporter

rshbhgrg:
WIP - This is under active discussion and development and not final.

refs https://github.com/TryGhost/Team/issues/1827

- manages export and import of complimentary tiers

### DIFF
--- a/ghost/core/core/server/services/members/exporter/query.js
+++ b/ghost/core/core/server/services/members/exporter/query.js
@@ -42,7 +42,7 @@ module.exports = async function (options) {
             END) as subscribed
         `))
         .select(knex.raw(`
-            (SELECT GROUP_CONCAT(product_id) FROM members_products f WHERE f.member_id = members.id) as tiers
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/services/members/exporter/query.js
```diff
@@ -42,7 +42,7 @@ module.exports = async function (options) {
             END) as subscribed
         `))
         .select(knex.raw(`
-            (SELECT GROUP_CONCAT(product_id) FROM members_products f WHERE f.member_id = members.id) as tiers
+            (SELECT GROUP_CONCAT(product_id,'~~',COALESCE(expiry_at,'')) FROM members_products f WHERE f.member_id = members.id) as tiers
```
rishabhgrg, line 45: Since `members_products` column has both product id and its expiry now, this allows returning the product on member in format - `TIER_ID~~EXPIRY_AT`, where `EXPIRY_AT` is set as empty string if its null (for forever expiry).

rishabhgrg:
Closed as the importer changes for `tiers`/`products` needs a bit more thought at the product level.

PR #15303, opened 2022-08-24
Fixed missing options when creating or editing an Offer

daniellockyer:
- without this, the model doesn't have the context on who was
  adding/editing it
- this resulted in being unable to store actions for Offers because the
  `actor` is unknown
- this is the pattern we use elsewhere in the code so I've copied it
  into here


ErisDS (approved):
Yes frame.options should be passed to the model always, else we lose track of transactions and all sorts of other issues

rshbhgrg (approved):

PR #15305, opened 2022-08-24
Enabled free trials via tiers, offers and complimentary subscriptions

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1726

- allows site owners to add a default free trial period to their tiers
- allows site owners to create custom offers that allow free trials on tiers
- allows site owners to give cardless free trials to members by setting expiry on complimentary subscriptions
- also allows complimentary members to upgrade as paid member

### DIFF
--- a/ghost/core/test/unit/server/models/stripe-customer-subscription.test.js
+++ b/ghost/core/test/unit/server/models/stripe-customer-subscription.test.js
@@ -25,35 +25,7 @@ describe('Unit: models/stripe-customer-subscription', function () {
             sinon.restore();
         });
 
-        it('ignores trial details when flag is off', function () {
-            const trialStartAt = new Date();
-            const trialEndAt = new Date(Date.now() + 7 * 86400 * 1000); // trial ending 7 days from now
-            const stripeSubscription = {
-                customer_id: 'fake_customer_id',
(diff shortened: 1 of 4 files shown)


ghost/core/test/unit/server/models/stripe-customer-subscription.test.js
```diff
@@ -25,33 +25,6 @@ describe('Unit: models/stripe-customer-subscription', function () {
...
-            };
-
-            const json = serialize(stripeSubscription);
-            should.exist(json.plan);
-            should.exist(json.customer);
-            should.not.exist(json.trial_start_at);
-            should.not.exist(json.trial_end_at);
-        });
-
         it('returns trial details with flag enabled', function () {
```
ErisDS, line 28: nitpick but can we fix the name of this test and remove the following line now? Saves cleanup later, especially given that it doesn't say which flag this refers to - we'll likely lose this!

rishabhgrg: Ah yeah sure, good point! On it.

rishabhgrg: Done ✅

PR #15314, opened 2022-08-25
Added staff service to manage subscription email alerts 

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1826

- adds new staff service package that manages email alerts for staff users
- manages new free member signup as well pad sub start/cancel emails
- adds new method to user model for fetching all users that are supposed to receive alerts

### DIFF
--- a/ghost/core/core/server/models/user.js
+++ b/ghost/core/core/server/models/user.js
@@ -489,6 +489,32 @@ User = ghostBookshelf.Model.extend({
     },
 
     /**
+     * Returns users who should receive a specific type of alert
+     * @param {'free-signup'|'paid-started'|'paid-canceled'} type The type of alert to fetch users for
+     * @param {any} options
+     * @return {Promise<[Object]>} Array of users
+     */
+    getEmailAlertUsers(type, options) {
+        options = options || {};
+
+        let filter = 'status:active';
+        if (type === 'free-signup') {
+            filter += '+free_member_signup_notification:true';
+        } else if (type === 'paid-started') {
+            filter += '+paid_subscription_started_notification:true';
+        } else if (type === 'paid-canceled') {
+            filter += '+paid_subscription_canceled_notification:true';
+        }
+        return this.findAll(_.merge({filter, withRelated: ['roles']}, options)).then((users) => {
+            return users.toJSON().filter((user) => {
+                return user?.roles?.some((role) => {
+                    return ['Owner', 'Administrator'].includes(role.name);
+                });
+            });
+        });
+    },
+
+    /**
      * ### Edit
      *
      * Note: In case of login the last_seen attribute gets updated.
--- a/ghost/staff-service/lib/emails.js
+++ b/ghost/staff-service/lib/emails.js
@@ -1,0 +1,291 @@
+const {promises: fs} = require('fs');
+const path = require('path');
+const _ = require('lodash');
+const moment = require('moment');
+
+class StaffServiceEmails {
+    constructor({logging, models, mailer, settingsCache, urlUtils}) {
+        this.logging = logging;
+        this.models = models;
+        this.mailer = mailer;
+        this.settingsCache = settingsCache;
+        this.urlUtils = urlUtils;
+
+        this.Handlebars = require('handlebars');
+    }
+
+    async notifyFreeMemberSignup(member, options) {
+        const users = await this.models.User.getEmailAlertUsers('free-signup', options);
+
+        for (const user of users) {
+            const to = user.email;
+            const memberData = this.getMemberData(member);
+
+            const subject = `🥳 Free member signup: ${memberData.name}`;
+
+            const templateData = {
+                memberData,
+                siteTitle: this.settingsCache.get('title'),
+                siteUrl: this.urlUtils.getSiteUrl(),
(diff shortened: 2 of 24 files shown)


ghost/staff-service/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "lib"
+  ],
+  "devDependencies": {
+    "mocha": "10.0.0",
+    "should": "13.2.3",
+    "sinon": "14.0.0"
+  },
+  "dependencies": {
+    "lodash": "4.17.21",
+    "moment": "2.29.1"
```
SimonBackx, line 27: Not sure about the dependencies in the monorepo, but `handlebars` might be missing here (though I think it should work without it).

rishabhgrg: 🤔 Yea this is a weird one, because I don't see `handlebars` as a direct dependency for any of the packages including `core`, though `comments` service also uses it. Not sure about adding it separately here when we haven't done so far, but will double check on this for whats the expected behavior.

rishabhgrg: Pinned handlebars to `"handlebars": "4.7.7"` which seems the right version here.

SimonBackx (approved):

ghost/core/core/server/models/user.js
```diff
@@ -488,6 +488,32 @@ User = ghostBookshelf.Model.extend({
...
+
+        let filter = 'status:active';
+        if (type === 'free-signup') {
+            filter += '+free_member_signup_notification:true';
+        } else if (type === 'paid-started') {
+            filter += '+paid_subscription_started_notification:true';
+        } else if (type === 'paid-canceled') {
+            filter += '+paid_subscription_canceled_notification:true';
+        }
+        return this.findAll(_.merge({filter, withRelated: ['roles']}, options)).then((users) => {
```
allouis, line 508: This reads that the `filter` option can be overridden, which would make this method behave unusually - was this purposeful?

ghost/staff-service/lib/emails.js
```diff
@@ -0,0 +1,291 @@
...
+        let msg = Object.assign({
+            from: this.fromEmailAddress,
+            forceTextContent: true
+        }, message);
+
+        return this.mailer.send(msg);
+    }
+
+    async renderEmailTemplate(templateName, data) {
+        const htmlTemplateSource = await fs.readFile(path.join(__dirname, './email-templates/', `${templateName}.hbs`), 'utf8');
```
allouis, line 280: Can we cache this file so we don't have to read from disk each time?

ghost/staff-service/lib/emails.js
```diff
@@ -0,0 +1,291 @@
...
+                subject,
+                html,
+                text
+            });
+        }
+    }
+
+    async notifyPaidSubscriptionCanceled({member, cancellationReason, tier, subscription}, options = {}) {
+        const users = await this.models.User.getEmailAlertUsers('paid-canceled', options);
+        const subscriptionPriceData = _.get(subscription, 'items.data[0].price');
```
allouis, line 98: We support the optional chaining now which is preferred to using lodash!

allouis: Also it would have been nice to have this method define an object shape it wants, and the call site can pass that in, that way this package isn't dependant on the shape of subscriptions.

We're most likely going to be changing that soon, and we'll now need to either update this package, or map to a very nested and Stripe specific data structure

ghost/staff-service/lib/staff-service.js
```diff
@@ -0,0 +1,46 @@
+class StaffService {
+    constructor({logging, models, mailer, settingsCache, urlUtils}) {
+        /** @private */
```
allouis, line 3: This doesn't apply to the logging below, just FYI

ghost/staff-service/test/hello.test.js
```diff
@@ -0,0 +1,10 @@
+// Switch these lines once there are useful utils
+// const testUtils = require('./utils');
+require('./utils');
+
+describe('Hello world', function () {
```
allouis, line 5: Can delete this

ghost/core/core/server/models/user.js
rishabhgrg: > was this purposeful?

It was not, that's a good point. I think it should be other way around like below
```
_.merge(options, {filter, withRelated: ['roles']})
```

allouis: Yeah, perhaps even `merge({}, options, {filter, withRelated: ['roles']});` so that we don't augment the options object passed in

ghost/staff-service/lib/emails.js
rishabhgrg: > it would have been nice to have this method define an object shape it wants, and the call site can pass that in, that way this package isn't dependant on the shape of subscriptions.

Yeah i agree, I noted cleaning this up because it doesn't make sense for the this package to know the internals of subscription object, will do that 👌

ghost/staff-service/lib/emails.js
rishabhgrg: Probably yea, will have a quick look at this later today or write it up so we don't miss it.

PR #15315, opened 2022-08-25
Added email alert triggers for member signup and subscription

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1826

- fires email alert on free member creation after they finish signing up via checkout link
- triggers a paid subscription start alert when a new paid subscription is linked
- triggers paid subscription cancellation alert when an active subscription is set to cancel on period end

### DIFF
--- a/ghost/members-api/lib/MembersAPI.js
+++ b/ghost/members-api/lib/MembersAPI.js
@@ -60,6 +60,7 @@ module.exports = function MembersAPI({
     },
     stripeAPIService,
     offersAPI,
+    staffService,
     labsService,
     newslettersService,
     memberAttributionService
@@ -85,6 +86,7 @@ module.exports = function MembersAPI({
         stripeAPIService,
         tokenService,
         newslettersService,
+        staffService,
         labsService,
         productRepository,
         Member,
@@ -149,6 +151,7 @@ module.exports = function MembersAPI({
         productRepository,
         StripePrice,
         tokenService,
+        staffService,
         sendEmailWithMagicLink
     });
 
@@ -218,6 +221,12 @@ module.exports = function MembersAPI({
             return member;
         }
         const newMember = await users.create({name, email, labels, newsletters, attribution});
+
+        // Notify staff users of new free member signup
+        if (labsService.isSet('emailAlerts')) {
+            await staffService.notifyFreeMemberSignup(newMember.toJSON());
+        }
+
         await MemberLoginEvent.add({member_id: newMember.id});
         return getMemberIdentityData(email);
     }
@@ -324,7 +333,7 @@ module.exports = function MembersAPI({
         memberBREADService,
         events: eventRepository,
         productRepository,
-        
+
         // Test helpers
         getTokenDataFromMagicLinkToken
     };
--- a/ghost/members-api/lib/repositories/member.js
+++ b/ghost/members-api/lib/repositories/member.js
@@ -1203,6 +1203,22 @@ module.exports = class MemberRepository {
                     member_id: member.id,
                     from_plan: subscriptionModel.get('plan_id')
                 }, sharedOptions);
+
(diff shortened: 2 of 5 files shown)


rishabhgrg:
@allouis I haven't figured out yet how to add test for checking email trigger for paid member cancellation at period end, as we don't have a pattern right now to trigger `updateSubscription` in e2e test. I am working on figuring that out but hopefully it's not a blocker for this one.

ghost/members-api/lib/MembersAPI.js
```diff
@@ -218,6 +221,12 @@ module.exports = function MembersAPI({
             return member;
         }
         const newMember = await users.create({name, email, labels, newsletters, attribution});
+
+        // Notify staff users of new free member signup
+        if (labsService.isSet('emailAlerts')) {
+            await staffService.notifyFreeMemberSignup(newMember.toJSON());
```
allouis, line 227: I'm not sure why this is done here, rather than in the create method for member repo. I get that we don't want to send the free signup email when creating paid members, but a flag could be passed when creating a member by Stripe? 

We shouldn't be having side effects like this happening in a method with a name implying a read operation IMO

ghost/members-api/lib/repositories/member.js
```diff
@@ -932,6 +938,16 @@ module.exports = class MemberRepository {
                 mrr_delta: model.get('mrr'),
                 ...eventData
             }, options);
+
+            // Notify paid member subscription start
+            if (this._labsService.isSet('emailAlerts')) {
+                await this.staffService.notifyPaidSubscriptionStart({
+                    member: member.toJSON(),
+                    offer: offer ? this._offerRepository.toJSON(offer) : null,
```
allouis, line 946: It would be preferable to either use the `Offer` object here OR if you want a POJO, use the offer service rather than repository - it's not a massive deal as we're gonna reshuffle stuff around

allouis (commented):
It would have been really nice to use events for this, rather than method calls - what we've done is add another dependency to the members-api, which already has a _lot_! 

By using events we instead make this new feature dependent on the members service, which means it's super easy to change implementation by just changing that package/service.

This would also make it easier to test, as we can test that the events are emitted in the e2e tests where possible, but in the email alerts package we can test that the email is sent when receiving those events

ghost/members-api/lib/repositories/member.js
rishabhgrg: Yeah so the reason i went this way was I couldn't find anything directly exposing a way to go POJO from Offer object in there, as the DTO was only used on specific methods. tbh I was expecting a directly method on Offer object to convert to POJO, but yea definitely could be a better way to handle this in cleanup with reshuffle.

ghost/members-api/lib/MembersAPI.js
rishabhgrg: >  get that we don't want to send the free signup email when creating paid members, but a flag could be passed when creating a member by Stripe?

Yea agree adding a flag to `create` method from here and triggering it in there definitely makes more sense, I can clean this up today. Only reason it went in there was this was the first alert I added and this was the quickest way.

allouis: Okay cool - thanks :)

ghost/members-api/lib/repositories/member.js
rishabhgrg: > The general idea was that the Offer object is used internally, as you can read all the information you want from is, and the DTO (Data Transfer Object) is used for the API.

Ah gotcha! I can clean this up along with the reshuffle.

rishabhgrg:
> It would have been really nice to use events for this, rather than method calls - what we've done is add another dependency to the members-api, which already has a lot!

@allouis Noted, that might be better yeah long term, I'll write up a tech debt note for this, interested in how this shapes up with events. Thanks for the review.

PR #15316, opened 2022-08-25
Promoted Member Attribution to a private beta

allouis:
We promote from alpha -> beta so that we don't require the
enableDeveloperExperiments flag, and remove the toggle from the UI so
it's not possible to enable without modifying the DB/using API.

### DIFF
--- a/ghost/core/core/shared/labs.js
+++ b/ghost/core/core/shared/labs.js
@@ -24,14 +24,14 @@ const GA_FEATURES = [
 // NOTE: this allowlist is meant to be used to filter out any unexpected
 //       input for the "labs" setting value
 const BETA_FEATURES = [
-    'activitypub'
+    'activitypub',
+    'memberAttribution'
 ];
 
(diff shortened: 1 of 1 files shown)


ghost/admin/app/templates/settings/labs.hbs
```diff
@@ -226,19 +226,6 @@
...
-                            <h4 class="gh-expandable-title">Member Attribution</h4>
-                            <p class="gh-expandable-description">
-                               Track which posts & pages are driving Member Subscriptions & Paid Conversions
-                            </p>
-                        </div>
-                        <div class="for-switch">
-                           <GhFeatureFlag @flag="memberAttribution" />
-                        </div>
-                    </div>
-                </div>
```
SimonBackx, line 241: Can't we keep it here for local debugging reasons? As this is only visible when developer experiments is on

allouis: Good shout

PR #15325, opened 2022-08-27
members-api: escape HTML

ElijahPepe:
The response returned by createCheckoutSetupSession() when a valid identity exists but not a valid subscription ID is susceptible to an XSS vulnerability; the response from the body is equivalent to the request body.


SimonBackx:
@ElijahPepe Thank you for your PR. Security vulnerabilities should be reported via the steps described [here](https://ghost.org/docs/security/#reporting-vulnerabilities), to avoid disclosing them before a fix is released. We investigated your PR internally and don't think this is a security vulnerability, but we made some related changes and will merge your PR anyway. Please contact us via `security@ghost.org` if you would like to discuss this further (more privately).

PR #15331, opened 2022-08-29
🐛 Fixed newsletters not rendering with non-HTML safe chars

allouis:
refs https://github.com/TryGhost/Team/issues/1379

We have disallowed non-html content to be rendered unescaped in the
email, this fixes issues caused by including characters such as "<"
which would completely break rendering as they will "open" new tags.

### DIFF
--- a/ghost/core/core/server/services/mega/template.js
+++ b/ghost/core/core/server/services/mega/template.js
@@ -28,7 +28,7 @@ const sanitizeKeys = (obj, keys) => {
 module.exports = ({post, site, newsletter, templateSettings}) => {
     const date = new Date();
     const hasFeatureImageCaption = templateSettings.showFeatureImage && post.feature_image && post.feature_image_caption;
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/services/mega/template.js
```diff
@@ -1,15 +1,43 @@
...
+        }
+    }
+
+    return sanitized;
+};
+
 module.exports = ({post, site, newsletter, templateSettings}) => {
     const date = new Date();
     const hasFeatureImageCaption = templateSettings.showFeatureImage && post.feature_image && post.feature_image_caption;
+    const cleanPost = sanitizeKeys(post, ['title', 'excerpt', 'html']);
```
allouis, line 31: I've found we do need to sanitize the `html` property in case in includes something like `'<'`.

I think this should be fine as we don't strip any tags or attributes

kevinansfield (approved):
Added sanitisation to some additional properties and expanded the tests a bit, otherwise looks good to me 👍

PR #15332, opened 2022-08-30
🐛 Fixed removing comped subscriptions for members with active subs

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/1859

**Problem:**
When for some reason a member has an active subscription (or legacy comped subscription) for product A, and a comped subscription for product B. You cannot remove comped subscription B.

**Fixed by:**
Updating the API to allow more flexible product changes on members.
- Allow the removal of (comped) products on a member, as long as that product doesn't have a related subscription
- (still) allow the addition of comped products to a member, as long as that member doesn't have other active subscriptions. This matches the existing behaviour, but now this is only checked for added products.
- Includes tests for these edge cases

### DIFF
--- a/ghost/core/test/e2e-api/admin/members.test.js
+++ b/ghost/core/test/e2e-api/admin/members.test.js
@@ -38,6 +38,10 @@ async function getPaidProduct() {
     return await models.Product.findOne({type: 'paid'});
 }
 
+async function getOtherPaidProduct() {
+    return (await models.Product.findAll({type: 'paid'})).models[0];
+}
+
 async function getNewsletters() {
     return (await models.Newsletter.findAll({filter: 'status:active'})).models;
 }
@@ -840,7 +844,7 @@ describe('Members API', function () {
 
         const newMember = body.members[0];
 
-        await agent
+        const updatedMember = await agent
             .put(`/members/${newMember.id}/`)
             .body({members: [compedPayload]})
             .expectStatus(200)
@@ -1347,7 +1351,7 @@ describe('Members API', function () {
             mrr: 100
         });
 
-        // Save this member for the next test
+        // Save this member for the next tests
         memberWithPaidSubscription = newMember;
     });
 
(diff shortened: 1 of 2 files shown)


ghost/core/test/e2e-api/admin/members.test.js
```diff
@@ -1384,6 +1388,126 @@ describe('Members API', function () {
         should.deepEqual(memberWithPaidSubscription, readMember, 'Editing a member returns a different format than reading a member');
     });
 
+    it('Cannot add complimentary subscriptions to a member with an active subscription', async function () {
+        if (!memberWithPaidSubscription) {
+            // Previous test failed
+            this.skip();
```
allouis, line 1394: This is an interesting pattern, have you seen this used elsewhere in the tests?

I'm not against it, as it is frustrating when a single test failing causes a cascade of later ones failing.

Ultimately that's because we have tests depending on some global state outside of the test, which IMO is not good.

We _could_ fix this by making sure each test does the setup that it needs, but that adds overhead.

Using this pattern at least makes the coupling explicit 🤔 Though what would happen if we removed the previous test, it would render this test useless right?

I'd be interested in your thoughts either way on this, and if you think it's a good idea - let's talk with Hannah about sharing/using this pattern going forwards

ghost/core/test/e2e-api/admin/members.test.js
```diff
@@ -1384,6 +1388,126 @@ describe('Members API', function () {
...
+            .expectStatus(200);
+
+        const beforeMember = body2.members[0];
+        assert.equal(beforeMember.tiers.length, 2, 'The member should have two products now');
+
+        // Now try to remove only the complimentary one
+        const compedPayload = {
+            id: memberWithPaidSubscription.id,
+            // Remove all complimentary subscriptions
+            tiers: [...memberWithPaidSubscription.tiers]
```
allouis, line 1468: No need for spread operator here, right? NABD

ghost/core/test/e2e-api/admin/members.test.js
```diff
@@ -1384,6 +1388,126 @@ describe('Members API', function () {
...
+            .put(`/members/${memberWithPaidSubscription.id}/`)
+            .body({members: [compedPayload]})
+            .expectStatus(200);
+
+        const updatedMember = body.members[0];
+        assert.equal(updatedMember.status, 'paid', 'Member should still have the paid status');
+        assert.equal(updatedMember.tiers.length, 1, 'The member should have one product now');
+        assert.equal(updatedMember.tiers[0].id, memberWithPaidSubscription.tiers[0].id, 'The member should have the paid product');
+
+        /*const compedPayload = {
```
allouis, line 1481: This is the same as the previous test right? Let's remove the comment, or uncomment and remove previous test - up to you ☺️

allouis (approved):
Couple comments but otherwise 👍

ghost/core/test/e2e-api/admin/members.test.js
SimonBackx: Yeah, I introduced this pattern a couple of months ago in the same file. I didn't really think a lot about it when writing these additional tests. I don't think this pattern is ideal, but it was an effort to prevent having to repeat all the Stripe mocking in every test. But moving that repeated part to the top might be a better solution... 🤔 Or we could build the models manually without mocking Stripe...

ghost/core/test/e2e-api/admin/members.test.js
SimonBackx: Right 😅👌
```suggestion
            tiers: memberWithPaidSubscription.tiers
```

ghost/core/test/e2e-api/admin/members.test.js
SimonBackx: Copy paste issue or something 😬

ghost/core/test/e2e-api/admin/members.test.js
```diff
@@ -1384,6 +1388,126 @@ describe('Members API', function () {
...
+            id: memberWithPaidSubscription.id,
+            // Remove all paid subscriptions (= not allowed atm)
+            tiers: []
+        };
+
+        await agent
+            .put(`/members/${memberWithPaidSubscription.id}/`)
+            .body({members: [compedPayload]})
+            .expectStatus(400);*/
```
SimonBackx, line 1490:
```suggestion
```

ghost/core/test/e2e-api/admin/members.test.js
SimonBackx: Gonna move this to a tech debt issue and clean up the whole test file in one go 🧹

naz: @SimonBackx what was the outcome of the cleanup? If the `memberWithPaidSubscription` is used in a group of test, like it seems to be done here, probably best would be grouping such tests into a "describe" and having a `memberWithPaidSubscription` be set up through fixtures? 🤔

Issue #15104, opened 2022-07-27
Confusing redirection to /signin/ and /signup/ after clicking on the magic link


fueko:
### Issue Summary

The idea of the change introduced in version 5.4.0 => https://github.com/TryGhost/Ghost/commit/e60806de45ddaff66c13e5806afd21ea6425c503 is very useful for Portal, but in certain situations it creates confusion among users.

If someone signs in or registers from a custom /signup/ or /signin/ page, after clicking on the magic link, they end up on the /signup/ and /signin/ page again, which is very confusing for many people - they often re-login.

Can the new redirects be restricted to the Portal only? The best solution would be for custom forms to be excluded from the new auto redirect option. Usually they are only available on /signin/ and /signup/.

Many sites also use the option to collect email addresses to subscribe to a newsletter in which case it is also confusing to return to the previous page.

### Steps to Reproduce

Log in through the custom form, when you click on the magic link you will return to the login page. 

### Ghost Version

5.5.0

### Node.js Version

v16.13.0

### How did you install Ghost?

localhost

### Database type

SQLite3

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

ErisDS:
Hey @fueko I know we still have a few tickets open around this internally, so I'll talk to the team and see what we can do to alleviate this.

oxscience:
+1. Would love to see a solution soon :-), thanks!

ErisDS:
We recognise that this is an issue and are working on a plan for how to resolve it for all the various use cases. However, this is not being considered as a critical or blocking bug, and so will be tackled after the current inflight projects are completed in a couple of weeks time.

fueko:
Hi @ErisDS, I understand and thank you for the info! However, a really large number of buyers are asking me how to fix this problem because they don't understand why the login flow has been changed.

Maybe it would be possible to add a new attribute, e.g. `data-members-autoredirect="false"`, to disable this behavior for custom forms? Of course, this is just an idea, but it would give theme developers control over this.

Another idea is to disable this behavior for `pages`, this would not be ideal, but for a while it would avoid the current problems until a better solution is in place.

oxscience:
> We recognise that this is an issue and are working on a plan for how to resolve it for all the various use cases. However, this is not being considered as a critical or blocking bug, and so will be tackled after the current inflight projects are completed in a couple of weeks time.

[...]

PR #15335, opened 2022-08-30
✨ Added optional data-attribute to enable/disable autoredirection.

ronaldlangeveld:
refs https://github.com/TryGhost/Ghost/issues/15104 https://github.com/TryGhost/Team/issues/1800

- On custom sign up and login forms, users often wouldn't want to be redirected to that page after signing in.
- This takes a new data-attribute value (eg `data-members-autoredirect="false"`) sent from Portal into account before parsing the referrer on the magic link URL.

Related - https://github.com/TryGhost/Portal/pull/261


allouis (commented):
Nice and simple - could we just add an e2e test for it here? https://github.com/TryGhost/Ghost/blob/main/ghost/core/test/e2e-api/members/send-magic-link.test.js

allouis (approved):
Great tests!

PR #15345, opened 2022-08-31
Reset rate limiting for members magic link upon successful login

allouis:
refs https://github.com/TryGhost/Team/issues/1771

We don't have access to `req.brute.reset` due to the way the flow
works, we have one endpoint which sends an email with a magic link,
and another route which handles the login. We don't want to apply
brute force protection to both because our rate limiting is designed
for API requests not web page visits (which is how login is handled).

Because of this we require access to the underlying ExpressBrute
instance exposed by the spam-protection module, so that we can perform
the reset.

### DIFF
--- a/ghost/core/core/server/services/members/middleware.js
+++ b/ghost/core/core/server/services/members/middleware.js
@@ -156,7 +156,7 @@ const createSessionFromMagicLink = async function (req, res, next) {
 
     try {
         const member = await membersService.ssr.exchangeTokenForSession(req, res);
-        spamPrevention.membersAuth().reset(req.ip, `${member.email}login`);
+        spamPrevention.membersAuth().reset(req.ip, `${member?.email}login`);
         const subscriptions = member && member.subscriptions || [];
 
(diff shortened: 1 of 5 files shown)


ghost/core/core/server/services/members/middleware.js
```diff
@@ -156,14 +156,7 @@ const createSessionFromMagicLink = async function (req, res, next) {
...
         const member = await membersService.ssr.exchangeTokenForSession(req, res);
-        await new Promise((resolve, reject) => {
-            spamPrevention.membersAuth().reset(req.ip, `${member.email}login`, function (err) {
-                if (err) {
-                    reject(err);
-                }
-                resolve();
-            });
-        });
+        spamPrevention.membersAuth().reset(req.ip, `${member.email}login`);
```
allouis, line 159: We don't actually want to stop someone from signing in if this fails, and we also don't want to block the request on it - so running this in the background is preferred

SimonBackx (approved):

PR #15348, opened 2022-09-01
🐛 Fixed sending emails from email domain that includes www subdomain

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/1855
fixes https://github.com/TryGhost/Team/issues/1866

This commit moves all duplicate methods to get the support email address to a single location. Also methods to get the default email domain are moved.

For the location, I initially wanted to put it at the settings service. But that service doesn't feel like the right place. Instead I created a new settings helpers service. This service takes the settingsCache, urlUtils and config and calculates some special 'calculated' settings based on those:

- Support email methods
- Stripe (active) keys / stripe connected (also removed some duplicate code that calculated the keys in a couple of places)
- All the calculated settings are moved to the settings helpers

I'm not 100% confident in whether this is the right place to put the helpers. Suggestions are welcome.

### DIFF
--- a/ghost/core/core/server/services/comments/emails.js
+++ b/ghost/core/core/server/services/comments/emails.js
@@ -3,15 +3,15 @@ const path = require('path');
 const moment = require('moment');
 const htmlToPlaintext = require('@tryghost/html-to-plaintext');
 const postEmailSerializer = require('../mega/post-email-serializer');
-const settingsService = require('../settings/settings-service');
 
 class CommentsServiceEmails {
-    constructor({config, logging, models, mailer, settingsCache, urlService, urlUtils}) {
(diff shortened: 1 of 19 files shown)


allouis (approved):
I'm also not 100% sure if this is the right place for them, perhaps the `SettingsHelpers` class should live inside the settings service and be exported from it.

This is more of a code location issue rather than an architectural issue though - I think this is a good move in terms of architecture.

This isn't a huge deal to me, and am happy for it to go in as is

SimonBackx:
@allouis Yeah, my priority was to fix the bug and deduplicate the code. I think it should be straightforward to move it to somewhere else in a future change.

PR #15358, opened 2022-09-02
🐛 Fixed missing email alerts for adding new members via API

### LINKED ISSUE #15324: Adding a new member with Admin API does not trigger member email alerts
### Issue Summary

When adding a new member with the Admin API endpoint the new member is added but the notification is not sent to the staff users and all staff users have notifications enabled.

### Steps to Reproduce

Enable new signup email alerts for staff users.
Create a new member using the Admin API.
The user is created, it is listed in the members page.
Staff users do not receive email about new member.
Email sending works, other email (ex: staff user invite) are sent and received.

### Ghost Version

5.12.0

### Node.js Version

16.17.0

digital ocean + ubuntu

### Database type

MySQL 8

### Browser & OS version

_No response_

```shell
No errors
```

### Code of Conduct

rshbhgrg:
closes https://github.com/TryGhost/Team/issues/1864
closes https://github.com/TryGhost/Ghost/issues/15324

- email alerts for free member was previously setup to trigger only on magic link verification
- moves free member alert trigger to member creation, triggers for both members added via API and direct signup
- ignores email alert for members created via stripe webhook as they'll trigger the paid member alert

### DIFF
--- a/ghost/members-api/lib/repositories/member.js
+++ b/ghost/members-api/lib/repositories/member.js
@@ -215,6 +215,11 @@ module.exports = class MemberRepository {
             options = {};
         }
 
+        // Member creation is triggered via stripe webhook
+        // Used to ignore free member email alert
+        const viaStripe = options.flag === 'stripe';
+        delete options.flag;
+
         const {labels} = data;
 
         if (labels) {
@@ -291,6 +296,11 @@ module.exports = class MemberRepository {
             to_status: member.get('status'),
             ...eventData
         }, options);
+
+        // Notify free member added
+        if (['member', 'api'].includes(source) && !viaStripe && memberStatusData.status === 'free') {
+            await this.staffService.notifyFreeMemberSignup(member.toJSON());
+        }
 
         const newsletters = member.related('newsletters').models;
 
--- a/ghost/members-api/test/unit/lib/repositories/member.test.js
+++ b/ghost/members-api/test/unit/lib/repositories/member.test.js
@@ -264,4 +264,216 @@ describe('MemberRepository', function () {
             sinon.restore();
         });
     });
+
+    describe('create', function (){
+        let Member;
+        let staffService;
+        let notifySpy;
+        let MemberPaidSubscriptionEvent;
+        let StripeCustomerSubscription;
+        let MemberStatusEvent;
+        let MemberProductEvent;
+        let stripeAPIService;
+        let newslettersService;
+        let productRepository;
+        let labsService;
+
+        beforeEach(async function () {
+            notifySpy = sinon.spy();
+
+            Member = {
+                add: sinon.stub().resolves({
+                    related: () => {
+                        return {
+                            query: sinon.stub().returns({
+                                fetchOne: sinon.stub().resolves({})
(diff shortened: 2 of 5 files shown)


ghost/members-api/lib/repositories/member.js
```diff
@@ -215,6 +215,11 @@ module.exports = class MemberRepository {
             options = {};
         }
 
+        // Member creation is triggered via stripe webhook
+        // Used to ignore free member email alert
+        const viaStripe = options.flag === 'stripe';
```
allouis, line 220: This flag is further coupling Members to Stripe, it would be cool if we didn't have to do that!

In this previous comment https://github.com/TryGhost/Ghost/pull/15315#pullrequestreview-1085464683 I mentioned that it would be nice to use events for this feature, so that we're not adding more dependencies. Now that we're coupling further it seems like a good time to do that 🤔 

As it stands however we're unable to use the event, because it doesn't contain enough information for you to determine whether or not to send the email - that is due to a problem with the architecture of Members, but it shouldn't be too big to solve.

### Solution 1

We can add a `'stripe'` source to the `MemberCreatedEvent` - so that when you catch the event, you can ignore those which were caused by Stripe. This would mean updating the `create` method to somehow allow us to get that source through - perhaps by overloading the `options.context` param, I'm not sure on details - but we'd want it to be generic.

This is a big change though in terms of our data, and I'm not 100% sure we want to add that source to this event, yet at least.

### Solution 2

We can pass a list of subscriptions to the `create` method, and instead of calling `linkSubscription` in the webhook handler, we can call `linkSubscription` in the `create` method, this will mean that the event will be associated with a Member with a `status` of `'paid'` - also allowing you to ignore the event.

Then your handler would look something like this:
```js
handler(event) {
  const member = await getMember(event.data.member_id);
  if (event.data.source !== 'api' || event.data.source !== 'member') {
    return;
  }
  if (member.status === 'free') {
    sendEmail();
  }
}
```

It's a bit of a pain that we only give the `id` in the event for now, but I think we should leave that as is, for now at least.

---

What do you think of splitting this out now - does that make sense to you?

If so, what do you think of these solutions? Can you think of any other ways to reduce the coupling between these two features?

ghost/members-api/test/unit/lib/repositories/member.test.js
```diff
@@ -264,4 +264,216 @@ describe('MemberRepository', function () {
             sinon.restore();
         });
     });
+
+    describe('create', function (){
```
allouis, line 268: Nice - way to get us started with some tests for this method 💃

ghost/members-api/lib/repositories/member.js
rishabhgrg: Thanks @allouis , as discussed, will rewrite this using Soln 2 as it seems the right time to move this to events.

rishabhgrg:
Closing as will be re-written in favor of using events for sending alerts

rishabhgrg:
Follow up PR which uses events for email alerts [here](https://github.com/TryGhost/Ghost/pull/15387)

PR #15359, opened 2022-09-02
Replaced all 'bio' references with 'expertise' for member comments.

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/1772.

Large cleanup - replaced all references of 'bio' with 'expertise' on the Comments UI (https://github.com/TryGhost/comments-ui/pull/10) and API.

This is a bit of a unique one considering it contains a migrations (renamed column `bio` to `expertise`) which is suppose to be a standalone PR, in which case it's a separate [commit](https://github.com/TryGhost/Ghost/pull/15359/commits/63d54d54f30badbb888746f9c7cbff96ae5cd793). However since it's a breaking change, I kept it in one PR.


ghost/core/core/server/data/migrations/versions/5.13/2022-09-02-12-55-rename-members-bio-to-expertise.js
```diff
@@ -0,0 +1,16 @@
+const logging = require('@tryghost/logging');
+const {createNonTransactionalMigration} = require('../../utils');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        await knex.schema.table('members', (table) => {
+            table.renameColumn('bio', 'expertise');
```
allouis, line 7: @daniellockyer Is this gonna work for SQLite? We bundle > 3.25 now right?

allouis (commented):
Looks good - just a Q for Daniel RE SQLite

ghost/core/core/server/data/migrations/versions/5.13/2022-09-02-12-55-rename-members-bio-to-expertise.js
daniellockyer: We do - yes

I worry about performance though - we use `ALGORITHM=COPY` for adding and removing a column. I wonder if we should do the same here?

daniellockyer: Actually, in MySQL 8, the default is ALGORITHM=INSTANT, which is probably better

ghost/core/core/server/data/migrations/versions/5.13/2022-09-02-12-55-rename-members-bio-to-expertise.js
```diff
@@ -0,0 +1,16 @@
+const logging = require('@tryghost/logging');
+const {createNonTransactionalMigration} = require('../../utils');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
```
daniellockyer, line 5: Can you add log lines in each of the `up` and `down`, just so it's obvious we're running this migration?

ghost/core/core/server/data/migrations/versions/5.13/2022-09-02-12-55-rename-members-bio-to-expertise.js
```diff
@@ -0,0 +1,17 @@
...
+        logging.info(`Changing column bio to expertise in members table`);
+        await knex.schema.table('members', (table) => {
+            table.renameColumn('bio', 'expertise');
+        });
+    },
+    async function down(knex) {
+        await knex.schema.table('members', (table) => {
+            logging.info(`Changing column expertise to bio in members table`);
```
daniellockyer, line 13: If we could move this log line up one line, then we should be all good 👌🏻

ghost/core/core/server/data/migrations/versions/5.13/2022-09-02-12-55-rename-members-bio-to-expertise.js
```diff
@@ -0,0 +1,30 @@
+const logging = require('@tryghost/logging');
+const {createNonTransactionalMigration} = require('../../utils');
+
```
daniellockyer, line 3: Sorry, the folder path needs to be updated too - `5.13` -> `5.14`

PR #15361, opened 2022-09-03
🐛 Fixed Hidden error in "Add tier" modal

### LINKED ISSUE #15290: Hidden error in "Add tier" modal
Try to add a tier but don’t fill the tier name field. There should be a field error that name is mandatory and the button should switch to “Retry. Instead, seems like nothing changes.

image

## Steps to reproduce 

1. Visit Settings > Memberships > Premium > Add tier
2. Hit the "Add tier" button without entering a name 
3. Notice that nothing happens
4. Look in the console, see the following error appearing:

image

Very unhelpful "Uncaught undefined" 😬 

## Expected behaviour

- The title field should get a red border
- A message should appear under the field saying name is required
- The "Add tier" button should change to a red "Retry" button

### Ghost Version
5.10

h2akim:
closes #15290 

- Capture error from model errors
- Add hasValidated for name property to properly mark field as error/success
- Add property to hasValidated after each failed validation
- Wrap saving on try-catch to suppress uncaught exception (validation error).

### DIFF
--- a/ghost/admin/app/components/modal-tier.js
+++ b/ghost/admin/app/components/modal-tier.js
@@ -1,7 +1,7 @@
-import {action} from '@ember/object';
 import ModalBase from 'ghost-admin/components/modal-base';
 import TierBenefitItem from '../models/tier-benefit-item';
 import classic from 'ember-classic-decorator';
+import {action} from '@ember/object';
 import {currencies, getCurrencyOptions, getSymbol} from 'ghost-admin/utils/currency';
 import {A as emberA} from '@ember/array';
 import {htmlSafe} from '@ember/template';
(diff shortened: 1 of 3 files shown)


ghost/admin/app/components/modal-tier.js
```diff
@@ -183,10 +179,17 @@ export default class ModalTierPrice extends ModalBase {
...
+        
+        try {
+            yield this.tier.save();
+            this.hasSaved = true;
+            yield this.confirm();
+            this.send('closeModal');
+        } catch (error) {
+            if (error === undefined) {
+                // Validation error
+            }
```
kevinansfield, line 191: Could you re-throw here if it's not a validation error? Just so any unexpected server errors are still caught/tracked/displayed

h2akim: Sure. Added throw error

kevinansfield (approved):

kevinansfield:
Thanks for tackling this @h2akim 🙂

PR #15365, opened 2022-09-05
Added test that validates output HTML of email template

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1871

This commit adds a test to the serialize method of `post-emaiserializer`. It checks whether the generated email HTML is valid and standard HTML5 and that all properties are escaped.

To do this validation, I depend on the new `html-validate` dev dependency. Just parsing the HTML with a HTML parser is not enough to guarantee that the HTML is okay.

Apart from that this fixes:
- Removed the sanitizeHTML method and replaced it with normal HTML escaping. We don't want to allow any HTML in the escaped fields. Whereas `sanitizeHTML` still allows valid HTML, but we don't want that and want the same behaviour as on the site. E.g., a post with a title `All your need to know about the <br /> tag` should actually render the same title and non-html content, being `All your need to know about the &lt;br /&gt; tag`
- The file, nft and audio card didn't (always) escape the injected HTML fields (new version @tryghost/kg-default-cards) 
- `@tryghost/string` is bumped because it contains the new escapeHtml method


ghost/core/test/unit/server/services/mega/post-email-serializer.test.js
```diff
@@ -89,31 +89,36 @@ describe('Post Email Serializer', function () {
...
+                    parsedErrors.push(`${item.ruleId}: ${item.message}\n   At line ${item.line}, col ${item.column}\n   HTML-snippet:\n${html}`);
                 }
-                const start = Math.max(item.line - 4, 0);
-                const end = Math.min(item.line + 4, lines.length - 1);
-
-                const html = lines.slice(start, end).map(l => l.trim()).join('\n\t');
-                parsedErrors.push(`${item.ruleId}: ${item.message}\n At line ${item.line}, col ${item.column}\n HTML-snippet:\n\t${html}`);
             }
-            should(report.valid).eql(0, 'Expected valid HTML without warnings, got errors:\n' + parsedErrors.join('\n\n'));
+            should(report.valid).eql(true, 'Expected valid HTML without warnings, got errors:\n' + parsedErrors.join('\n\n'));
```
allouis, line 149: We should be moving toward using `assert` rather than `should` - I get that we're already using `should` here though - up to you if you wanna move everything over or leave it for now

allouis (approved):

ghost/core/test/unit/server/services/mega/post-email-serializer.test.js
SimonBackx: Lets move it all at once instead of mixing them? Maybe we need some cleanup issues to move to assert

PR #15366, opened 2022-09-05
Added Tier price and currency data to products table

allouis:
refs https://github.com/TryGhost/Team/issues/1765

In order to better handle deleted objects in Stripe we want to decouple Members from Stripe. 

These changes allow us to have the Tier concept completely independent of the Stripe tables, such that the Stripe data can be generated as/when it's needed - which will help to protect against missing data.


ghost/core/core/server/data/migrations/versions/5.14/2022-09-02-20-25-add-columns-to-products-table.js
```diff
@@ -0,0 +1,21 @@
...
+    createAddColumnMigration('products', 'monthly_price', {
+        type: 'integer',
+        unsigned: true,
+        nullable: true
+    }),
+    createAddColumnMigration('products', 'monthly_price', {
+        type: 'integer',
+        unsigned: true,
+        nullable: true
+    }),
```
SimonBackx, line 15: One should be `yearly_price` I guess?

SimonBackx (approved):

ghost/core/core/server/data/migrations/versions/5.14/2022-09-02-20-52-backfill-new-product-columns.js
```diff
@@ -0,0 +1,37 @@
...
+            logging.info(`Updating ${rows.length} Tiers with price and currency information`);
+        }
+
+        for (const row of rows) {
+            await knex('products').update(row).where('id', row.id);
+        }
+    },
+    async function down(knex) {
+        logging.info('Removing currency and price information for all tiers');
+        knex('products').update({
```
daniellockyer, line 31: We need an `await` here

allouis: very cheeky yes we do

daniellockyer (approved):

allouis:
This is on hold until I get a chance to pick it back up in the next cooldown

ghost/core/core/server/data/migrations/versions/5.19/2022-09-02-20-52-backfill-new-product-columns.js
```diff
@@ -0,0 +1,37 @@
...
+                't.id as id',
+                'mp.amount as monthly_price',
+                'yp.amount as yearly_price',
+                knex.raw('coalesce(yp.currency, mp.currency) as currency')
+            )
+            .leftJoin('stripe_prices AS mp', 't.monthly_price_id', 'mp.id')
+            .leftJoin('stripe_prices AS yp', 't.yearly_price_id', 'yp.id')
+            .where('t.type', 'paid');
+
+        if (!rows) {
```
daniellockyer, line 18: Do we need a `.length` here? Otherwise `rows` is `[]` which is still truthy

allouis: Good catch

daniellockyer (approved):

PR #15371, opened 2022-09-06
Moved (un)like endpoint code to comments service

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/1861

- Moved like and unlike endpoint handling to comments service and controller
- Moved small part of report logic to comments controller
- Added proper 401 authentication error when not authenticated as member


allouis (approved):
👍

PR #15376, opened 2022-09-06
Added property cleaning to tag relations on pages + posts

daniellockyer:
refs https://github.com/TryGhost/Ghost/pull/15375

- we currently pass all properties for the `tags` property of a `page`/`post` body down further into Ghost, which is causing issues because it's handling properties it doesn't expect
- this is showing up because it's triggering save history events for tags when a post is edited
- this commit introduces a clean util which has an allowlist of properties allows on tag relations
- this list was taken from the schema: https://github.com/TryGhost/SDK/blob/128f8fb006ca0846cad4f3702a538998f834ee2a/packages/admin-api-schema/lib/schemas/posts.json#L214-L227

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/serializers/input/pages.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/input/pages.js
@@ -5,6 +5,7 @@ const url = require('./utils/url');
 const slugFilterOrder = require('./utils/slug-filter-order');
 const localUtils = require('../../index');
 const postsMetaSchema = require('../../../../../data/schema').tables.posts_meta;
+const clean = require('./utils/clean');
 
 function removeMobiledocFormat(frame) {
     if (frame.options.formats && frame.options.formats.includes('mobiledoc')) {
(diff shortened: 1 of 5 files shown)


allouis (commented):
I think we _might_ want to pass through all valid Tag properties for now, as this is a big change to the API - but I'm happy to be wrong about this.

I'd like @ErisDS to give final approval of this, as she has much more context around how the API should work with regards to creating Tag resources as part of creating/updating a Post.

Overall this approach is great - the only thing I'm concerned about is which properties we allow through ☺️

ErisDS:
OK so technically we should pass through all attributes for a tag, as the API effectively supports creating a tag at the same time as adding/editing a post.

However, I'm happy to apply this restriction for now and see if anyone complains that the functionality breaks. If they do, we should expand the list to all fields as a high priority fix, because we _have_ introduced a breaking change. If they don't, we are in a better position in terms of locking down our API.

ErisDS (approved):

PR #15387, opened 2022-09-08
Updated member email alerts to trigger via events

rshbhgrg:
closes https://github.com/TryGhost/Team/issues/1864
closes https://github.com/TryGhost/Team/issues/1865
closes https://github.com/TryGhost/Team/issues/1867
refs https://github.com/TryGhost/Team/issues/1881

- updates sending member alert emails using events instead of adding a direct dependency which is harder to maintain long term
- using events also allows decoupling the actual email sending from the main request for an action like cancel subscription or paid subscription

### DIFF
--- a/ghost/members-api/lib/repositories/member.js
+++ b/ghost/members-api/lib/repositories/member.js
@@ -39,7 +39,6 @@ module.exports = class MemberRepository {
      * @param {any} deps.OfferRedemption
      * @param {import('../../services/stripe-api')} deps.stripeAPIService
      * @param {any} deps.labsService
-     * @param {any} deps.staffService
      * @param {any} deps.productRepository
      * @param {any} deps.offerRepository
      * @param {ITokenService} deps.tokenService
@@ -61,7 +60,6 @@ module.exports = class MemberRepository {
         productRepository,
         offerRepository,
         tokenService,
-        staffService,
         newslettersService
     }) {
         this._Member = Member;
@@ -77,7 +75,6 @@ module.exports = class MemberRepository {
         this._productRepository = productRepository;
         this._offerRepository = offerRepository;
         this.tokenService = tokenService;
-        this.staffService = staffService;
         this._newslettersService = newslettersService;
         this._labsService = labsService;
 
--- a/ghost/members-api/lib/repositories/member.js
+++ b/ghost/members-api/lib/repositories/member.js
@@ -3,7 +3,7 @@ const errors = require('@tryghost/errors');
 const logging = require('@tryghost/logging');
 const tpl = require('@tryghost/tpl');
 const DomainEvents = require('@tryghost/domain-events');
-const {MemberCreatedEvent, SubscriptionCreatedEvent, MemberSubscribeEvent} = require('@tryghost/member-events');
+const {MemberCreatedEvent, SubscriptionCreatedEvent, MemberSubscribeEvent, SubscriptionCancelledEvent} = require('@tryghost/member-events');
 const ObjectId = require('bson-objectid');
 const {NotFoundError} = require('@tryghost/errors');
 
@@ -203,6 +203,8 @@ module.exports = class MemberRepository {
      * @param {Date} [data.created_at]
      * @param {Object[]} [data.products]
      * @param {Object[]} [data.newsletters]
+     * @param {Object} [data.stripeCustomer]
+     * @param {string} [data.offerId]
      * @param {import('@tryghost/member-attribution/lib/history').Attribution} [data.attribution]
      * @param {*} options
      * @returns
@@ -212,7 +214,7 @@ module.exports = class MemberRepository {
             options = {};
         }
 
-        const {labels} = data;
+        const {labels, stripeCustomer, offerId, attribution} = data;
 
         if (labels) {
             labels.forEach((label, index) => {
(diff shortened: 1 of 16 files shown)


ghost/member-events/lib/PaidSubscriptionStartedEvent.js
```diff
@@ -0,0 +1,60 @@
+/**
```
rishabhgrg, line 1: Had 2 minds about overloading the existing SubscriptionCreatedEvent vs adding a new one which captures all the information for paid susbcription, and a new one seemed better as the two have different usecases.

ghost/member-events/lib/MemberSubscriptionCancellationEvent.js
```diff
@@ -0,0 +1,52 @@
+/**
```
rishabhgrg, line 1: Noticed while adding this, we also have [MemberPaidCancellationEvent](https://github.com/TryGhost/Ghost/blob/7084217d3df39fbf6f3cba302c63414282882873/ghost/member-events/lib/MemberPaidCancellationEvent.js) which was added for Analytics but doesn't seem to have any active usecases. Maybe a good candidate for removing in future to avoid having two similar named events

ghost/member-events/lib/PaidSubscriptionStartedEvent.js
allouis: I think using the `SubscriptionCreatedEvent` makes more sense here ☺️ We have an issue to move it here: https://github.com/TryGhost/Team/issues/1881

So would be great if you could use that instead

ghost/members-api/lib/repositories/member.js
```diff
@@ -967,15 +993,30 @@ module.exports = class MemberRepository {
...
-            }
+            const offerData = offer ? {
+                name: offer.name.value,
+                amount: offer.amount.value,
+                type: offer.type.value,
+                durationInMonths: offer.duration.value.months,
+                duration: offer.duration.value
+            } : null;
+
+            DomainEvents.dispatch(PaidSubscriptionStartedEvent.create({
```
allouis, line 1004: If we switch to use the `SubscriptionCreatedEvent` you'll notice that we just pass id's through - which I know is a bit of a PITA, but I think it's the best way to go until we've really nailed down this event pattern.

ghost/members-api/lib/repositories/member.js
```diff
@@ -1232,21 +1273,37 @@ module.exports = class MemberRepository {
...
+                };
+                const tierData = ghostProduct ? {
+                    id: ghostProduct.get('id'),
+                    name: ghostProduct.get('name')
+                } : null;
+
+                const context = options?.context || {};
+                const source = this._resolveContextSource(context);
+
+                DomainEvents.dispatch(MemberSubscriptionCancellationEvent.create({
```
allouis, line 1300: As above - do you mind if we just pass ids for now? I'm concerned about having different data shapes used throughout the app, this event pattern is still fairly new and hasn't quite settled yet.

ghost/member-events/lib/MemberSubscriptionCancellationEvent.js
allouis: How do you feel about `SubscriptionCancelledEvent` to match with `SubscriptionCreatedEvent`?

allouis (commented):
This approach looks so much better!

Few requests though:
- Reuse the existing `SubscriptionCreatedEvent`
- Pass id's in event data for now
- Rename `PaidSubscriptionCanceledEvent` to `SubscriptionCancelledEvent` for consistency with the created event

ghost/member-events/lib/PaidSubscriptionStartedEvent.js
rishabhgrg: Ah! So we'd want the `SubscriptionCreatedEvent` to be moved to where we have the new `PaidSubscriptionStartedEvent` in `linkSubscription` right for that issue? Guess that'll also need to pass down the attribution changes ? And how do you feel about overloading it with full data instead of IDs ?

ghost/member-events/lib/MemberSubscriptionCancellationEvent.js
rishabhgrg: Sounds good if we are moving to `SubscriptionCreatedEvent` above :)

ghost/members-api/lib/repositories/member.js
rishabhgrg: My concern with passing IDs only is that it's not just one data object that we are fetching, we'll need to make requests to fetch multiple bits like the associated offer/tier/member other than the subscription, which is a lot of overhead for values that we already have 🤔

rishabhgrg: Gonna switch to using IDs for now so we can come back and clean this up later with a proper data pattern

ghost/member-events/lib/PaidSubscriptionStartedEvent.js
rishabhgrg: Resolved below, gonna switch to using ids

ghost/members-api/lib/repositories/member.js
rishabhgrg: 👍

allouis (approved):

PR #15395, opened 2022-09-12
Fixed performance of column rename migration

daniellockyer:
- the migration in 5.14 renames the `bio` columns on `members` to `expertise`
- unfortunately, the Knex helper we were using does a lot of interesting things with foreign keys that are slow on bigger MySQL clusters, and that we don't need here
- this commit refactors the migration to use raw SQL if the DB is MySQL, else we use the helper because SQLite SQL might be different here

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.14/2022-09-02-12-55-rename-members-bio-to-expertise.js
+++ b/ghost/core/core/server/data/migrations/versions/5.14/2022-09-02-12-55-rename-members-bio-to-expertise.js
@@ -1,34 +1,50 @@
+const DatabaseInfo = require('@tryghost/database-info');
 const logging = require('@tryghost/logging');
 const {createNonTransactionalMigration} = require('../../utils');
 
-module.exports = createNonTransactionalMigration (
+module.exports = createNonTransactionalMigration(
     async function up(knex) {
-        // check if the column exists before trying to rename it
+        logging.info('Renaming members.bio to members.expertise');
+
         const hasBio = await knex.schema.hasColumn('members', 'bio');
         const hasExpertise = await knex.schema.hasColumn('members', 'expertise');
 
-        if (hasBio && !hasExpertise) {
-            logging.info('Renaming members.bio to members.expertise');
-            await knex.schema.table('members', (table) => {
-                table.renameColumn('bio', 'expertise');
-            }
-            );
-        } else {
-            logging.info('members.bio does not exist, skipping rename');
+        // If we don't have the `bio` column, or the `expertise` column already exists, we're
+        // not in the right state to run this migration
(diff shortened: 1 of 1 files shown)


ghost/core/core/server/data/migrations/versions/5.14/2022-09-02-12-55-rename-members-bio-to-expertise.js
```diff
@@ -1,34 +1,54 @@
...
+        logging.info('Renaming members.bio to members.expertise');
+
         const hasBio = await knex.schema.hasColumn('members', 'bio');
+        if (!hasBio) {
+            logging.warn('members.bio does not exist, skipping rename');
+            return;
+        }
+
         const hasExpertise = await knex.schema.hasColumn('members', 'expertise');
+        if (hasExpertise) {
```
allouis, line 16: Would we want to throw here? Or drop the `bio` column maybe?

Having both a `bio` and `expertise` column is a completely invalid state

ghost/core/core/server/data/migrations/versions/5.14/2022-09-02-12-55-rename-members-bio-to-expertise.js
```diff
@@ -1,34 +1,54 @@
...
 
-        if (hasBio && !hasExpertise) {
-            logging.info('Renaming members.bio to members.expertise');
-            await knex.schema.table('members', (table) => {
-                table.renameColumn('bio', 'expertise');
-            }
-            );
-        } else {
-            logging.info('members.bio does not exist, skipping rename');
+        if (DatabaseInfo.isMySQL(knex)) {
```
allouis, line 21: I was thinking we should wrap it, nice

allouis (commented):
Approach looks good, one comment regarding how we handle invalid states

ghost/core/core/server/data/migrations/versions/5.14/2022-09-02-12-55-rename-members-bio-to-expertise.js
daniellockyer: Good question... I think we should throw so as to gracefully bail out and avoid further damage

daniellockyer: That said, migrations should be idempotent so if it already exists, we shouldn't try and do anything with it

I'm thinking I should collapse it back down into one if-statement and only do the rename if it's in the valid state

daniellockyer: Ok how's this? (latest push)

allouis (approved):

PR #15397, opened 2022-09-12
Added `posts.lexical` database field

kevinansfield:
closes https://github.com/TryGhost/Team/issues/1884

- adds `post.lexical` ready for use by the lexical-powered editor re-write
- fulfils the same purpose as `posts.mobiledoc` so uses the same field properties
- added to allowed formats in Post model so it won't be included by default in API responses meaning tests/snapshots don't need updating at present


daniellockyer:
The migration name needs changing, but it LGTM otherwise

daniellockyer (approved):

PR #15406, opened 2022-09-13
Passthrough redirect service

allouis:


ghost/link-replacement/lib/link-replacement.js
```diff
@@ -1,33 +1,51 @@
...
+
         // 1. Add attribution
         // TODO: this should move the the attribution service in the future
         url.searchParams.append('rel', newsletter.get('slug') + '-newsletter');
         url.searchParams.append('attribution_id', post.id);
         url.searchParams.append('attribution_type', 'post');
 
         // 2. Add redirect for link click tracking
-        const redirect = await this.createRedirect(url.toString());
+        const redirect = await this.#linkRedirectService.addRedirect(url);
```
SimonBackx, line 44: If we add `from` as a parameter, what do you think if we would do something like this instead (doesn't have to be in this PR):
```
const redirect = await this.#linkClickTrackingService.createTrackedRedirect(url);
```
Which internally would call `await this.#linkRedirectService.addRedirect(generatedSlug, url);`

SimonBackx: Maybe we can also pass the member uuid so that it includes that in the returned `from`

SimonBackx: Essentially we could also name it `this.#linkClickTrackingService.createTrackedLink(url);` which returns a normal URL and hides the internal implementation detail that it works via redirects 🤷‍♂️

ghost/link-redirects/lib/LinkRedirectsService.js
```diff
@@ -0,0 +1,34 @@
+const LinkRedirect = require('./LinkRedirect');
+
+class LinkRedirectsService {
+    /**
+     * @param {URL} to
+     *
+     * @returns {Promise<LinkRedirect>}
+     */
+    async addRedirect(to) {
```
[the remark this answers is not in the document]
allouis: As discussed will do this in a later commit

PR #15408, opened 2022-09-14
Added post stats service to return total posts in Explore endpoint

aileen:
no issue

- The explore endpoint needs to expose the total amount of published posts
- To be more consistent, this PR creates a PostStats class which is exposed as `stats` method within the PostService; just like it's done with the MemberService
- Moved existing method to return the date of the most recently published post into the stats service
- Updated the explore service test to reflect the new return property


ghost/core/test/e2e-api/admin/explore.test.js
```diff
@@ -17,7 +17,8 @@ describe('Explore API', function () {
                 .expectStatus(200)
                 .matchBodySnapshot({
                     explore: {
-                        most_recently_published_at: anyISODateTime,
```
aileen, line 20: @daniellockyer I actually do want to return a datetime, but somehow only got a date back from the db query. I couldn't make it work. Is there a better way of doing so?

ghost/core/core/server/services/posts/stats/post-stats.js
```diff
@@ -0,0 +1,31 @@
+class PostStats {
+    /**
+     * @param {Object} config
+     * @param {Object} config.db - an instance holding knex connection to the database
+     * @param {Boolean} config.isSQLite - flag identifying if storage is connected to SQLite
+     */
+    constructor({db, isSQLite}) {
+        this._db = db;
+        this._isSQLite = isSQLite;
```
allouis, line 9: No biggie, but I would consider not passing this in, and calculating it in this module.

ghost/core/core/server/services/posts/stats/post-stats.js
```diff
@@ -0,0 +1,31 @@
+class PostStats {
+    /**
+     * @param {Object} config
+     * @param {Object} config.db - an instance holding knex connection to the database
+     * @param {Boolean} config.isSQLite - flag identifying if storage is connected to SQLite
+     */
+    constructor({db, isSQLite}) {
+        this._db = db;
```
allouis, line 8: You might consider passing a knex instance here instead - also no biggie

ghost/core/core/server/services/posts/stats/post-stats.js
```diff
@@ -0,0 +1,31 @@
...
+        this._isSQLite = isSQLite;
+    }
+
+    /**
+     * Returns the most recently `published_at` post that was published or sent
+     * via email
+     */
+    async getMostRecentlyPublishedPostDate() {
+        const result = await this._db.knex.raw('SELECT * FROM posts WHERE status="published" OR status="sent" ORDER BY published_at DESC LIMIT 1');
```
allouis, line 17: If you use knex `select|where|limit` methods rather than `raw`, I think you won't have to check for the database type as knex will handle under the hood, something like `knex('posts').where('blah', ....`

ghost/core/core/server/services/posts/stats/post-stats.js
allouis: Also don't forget you can now use private members, so no need for underscore prefixes! `this.#db = db;` etc...

ghost/core/test/e2e-api/admin/explore.test.js
allouis: I think you want to wrap the output in a `Date` object, so if you update the signature of the `getMostRecentlyPublishedPostDate(): string` to `getMostRecentlyPublishedPostDate(): Date` everything should work as expected

allouis (approved):
Looks good - few v. minor comments if you care for them

ghost/core/core/server/services/posts/stats/post-stats.js
aileen: Ah, ok. I honestly copies the members stats code here which was using exactly that approach.

allouis: S'all good - don't need to change it if you don't want!

ghost/core/core/server/services/posts/stats/post-stats.js
```diff
@@ -0,0 +1,35 @@
...
+    constructor() {
+        this.#db = db;
+    }
+
+    /**
+     * Returns the most recently `published_at` post that was published or sent
+     * via email
+     */
+    async getMostRecentlyPublishedPostDate() {
+        const [result] = await this.#db.knex.select('published_at')
```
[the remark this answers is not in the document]
aileen: Fixed that! 🙌

daniellockyer (approved):

PR #15409, opened 2022-09-14
Added `email_track_clicks` setting

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/1900
refs https://github.com/TryGhost/Team/issues/1901

- Defaults to the same value as the current email_track_opens setting for existing installations, otherwise defaults to true
- Had to use a custom migration because the `addSetting` helper doesn't support using an existing setting as current value
- Added a minimal UI to change the setting, but this still needs some design magic 🪄✨
- Link replacement is disabled if `email_track_clicks` is disabled. In the future we might consider to still do parial additions, such as source attribution and maybe redirects (to discuss).

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.15/2022-09-14-12-46-add-email-track-clicks-setting.js
+++ b/ghost/core/core/server/data/migrations/versions/5.15/2022-09-14-12-46-add-email-track-clicks-setting.js
@@ -1,0 +1,59 @@
+const logging = require('@tryghost/logging');
+const ObjectId = require('bson-objectid').default;
+const {createTransactionalMigration} = require('../../utils');
+
+const {MIGRATION_USER} = require('../../utils/constants');
+const key = 'email_track_clicks';
+const cloneValueOf = 'email_track_opens';
+const defaultValue = 'true';
+const group = 'email';
+const type = 'boolean';
+
+// Note: we cannot use the addSetting helper here because we need to base the value on the current value of email_track_opens
+module.exports = createTransactionalMigration(
+    async function up(connection) {
+        const settingExists = await connection('settings')
+            .where('key', '=', key)
+            .first();
+        if (settingExists) {
+            logging.warn(`Skipping adding setting: ${key} - setting already exists`);
+            return;
+        }
+
+        // Value is based on the email_track_opens value
+        const reuseValueOfSetting = await connection('settings')
+            .where('key', '=', cloneValueOf)
+            .first();
+
+        logging.info(`Adding setting: ${key}`);
+        const now = connection.raw('CURRENT_TIMESTAMP');
+
+        return connection('settings')
+            .insert({
+                id: ObjectId().toHexString(),
+                key,
+                value: reuseValueOfSetting ? reuseValueOfSetting.value : defaultValue,
+                group,
+                type,
+                created_at: now,
+                created_by: MIGRATION_USER,
+                updated_at: now,
+                updated_by: MIGRATION_USER
+            });
+    },
+    async function down(connection) {
+        const settingExists = await connection('settings')
+            .where('key', '=', key)
+            .first();
+
+        if (!settingExists) {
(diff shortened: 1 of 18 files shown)


ghost/core/core/server/data/migrations/versions/5.15/2022-09-14-12-46-add-email-track-clicks-setting.js
```diff
@@ -0,0 +1,59 @@
...
+        return connection('settings')
+            .insert({
+                id: ObjectId().toHexString(),
+                key,
+                value: reuseValueOfSetting ? reuseValueOfSetting.value : defaultValue,
+                group,
+                type,
+                created_at: now,
+                created_by: MIGRATION_USER,
+                updated_at: now,
```
daniellockyer, line 40: We're not very strict with doing this (yet) but can I suggest we don't add `updated_at`, `updated_by` when we insert a new entry? Strictly speaking, I'd say it's been created but not _updated_ yet

ghost/core/core/server/services/mega/post-email-serializer.js
```diff
@@ -336,8 +336,10 @@ const PostEmailSerializer = {
 
         // Also replace the links in the HTML version
         if (labs.isSet('emailClicks')) {
-            if ((!options.isBrowserPreview && !options.isTestEmail) || process.env.NODE_ENV === 'development') {
-                content.html = await linkReplacement.service.replaceLinks(content.html, newsletter, postModel);
+            if (settingsCache.get('email_track_clicks')) {
```
allouis, line 339: I think that we want this check in the link-tracking service rather than here, as we still want to be able to update links for sent emails

allouis (commented):
Maybe we should be using the helper to create the setting here, and running a separate migration which populates the setting from the track opens setting in the same version that we release this feature?

I know it's probably an edge case, but migrating here could result in sites having link tracking enabled whilst having track opens disabled at the point of upgrade.

daniellockyer:
I like @allouis' suggestion re: using the migration helper 🙂

ghost/core/core/server/services/mega/post-email-serializer.js
SimonBackx: I agree, and added it in here on purpose until we moved it to a better suitable place in the future 😊

SimonBackx:
> Maybe we should be using the helper to create the setting here, and running a separate migration which populates the setting from the track opens setting in the same version that we release this feature?
> 
> I know it's probably an edge case, but migrating here could result in sites having link tracking enabled whilst having track opens disabled at the point of upgrade.

Good catch, I didn't think about that. I'll rewrite the migration to use the existing helper, and we can later write a new migration to update the value once this feature is released.

ghost/core/core/server/data/migrations/versions/5.15/2022-09-14-12-46-add-email-track-clicks-setting.js
SimonBackx: Is it okay if I change this in the addSetting helper too:
https://github.com/SimonBackx/Ghost/blob/e259dd11b6758272c89410e586eafdb98d198f22/ghost/core/core/server/data/migrations/utils/settings.js#L34-L37

ghost/core/core/server/services/mega/post-email-serializer.js
SimonBackx: I moved it to a more permanent location in the `LinkReplacementService`. We can still decide to move it to the `LinkClickTrackingService`, but not super sure about that (and that is something we can easily change later)

ghost/core/core/server/data/migrations/versions/5.15/2022-09-14-12-46-add-email-track-clicks-setting.js
daniellockyer: Fine with me 👍🏻

daniellockyer (approved):

allouis (approved):

PR #15413, opened 2022-09-15
Handled referrer attribution capture and storage

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1906, https://github.com/TryGhost/Team/issues/1907

- updates member attribution script to capture referrer url/source/medium information and store it in the history for later usage
  - stores `refUrl`, `refSource` and `refMedium` in history
- updates attribution service to use referrer data from history, and translate it to referrer source and medium using `ReferrerTranslator`
- includes referrer source/medium/url to stripe checkout metadata, allowing storing it in event later when a member completes payment

### DIFF
--- a/ghost/core/core/frontend/src/member-attribution/member-attribution.js
+++ b/ghost/core/core/frontend/src/member-attribution/member-attribution.js
@@ -62,7 +62,7 @@ const LIMIT = 15;
             // Valid item (so all following items are also valid by definition)
             return true;
         });
-        
+
         if (firstNotExpiredIndex > 0) {
             // Remove until the first valid item
             history.splice(0, firstNotExpiredIndex);
@@ -70,6 +70,26 @@ const LIMIT = 15;
             // Not a single valid item found, remove all
             history = [];
         }
+
+        // Fetch referrer data from query params
+        let refParam;
+        let sourceParam;
+        let utmSourceParam;
+        let utmMediumParam;
+        try {
+            // Fetch source/medium from query param
+            const url = new URL(window.location.href);
+            refParam = url.searchParams.get('ref');
+            sourceParam = url.searchParams.get('source');
+            utmSourceParam = url.searchParams.get('utm_source');
+            utmMediumParam = url.searchParams.get('utm_medium');
+        } catch (e) {
+            console.error('[Member Attribution] Parsing referrer from querystring failed', e);
+        }
+
+        const refSource = refParam || sourceParam || utmSourceParam || null;
+        const refMedium = utmMediumParam || null;
+        const refUrl = window.document.referrer || null;
 
         // Do we have attributions in the query string?
         try {
@@ -80,7 +100,10 @@ const LIMIT = 15;
                 history.push({
                     time: currentTime,
                     id: params.get('attribution_id'),
-                    type: params.get('attribution_type')
+                    type: params.get('attribution_type'),
+                    refSource,
+                    refMedium,
+                    refUrl
                 });
 
                 // Remove attribution from query string
@@ -98,12 +121,23 @@ const LIMIT = 15;
         if (history.length === 0 || history[history.length - 1].path !== currentPath) {
             history.push({
                 path: currentPath,
-                time: currentTime
+                time: currentTime,
+                refSource,
+                refMedium,
+                refUrl
             });
         } else if (history.length > 0) {
             history[history.length - 1].time = currentTime;
+            // Update referrer information for same path if available (e.g. when opening a link on same path via external referrer)
+            if (refSource) {
+                history[history.length - 1].refSource = refSource;
+                history[history.length - 1].refMedium = refMedium;
+            }
+            if (refUrl) {
+                history[history.length - 1].refUrl = refUrl;
+            }
         }
-        
+
         // Restrict length
         if (history.length > LIMIT) {
             history = history.slice(-LIMIT);
--- a/ghost/member-attribution/lib/attribution.js
+++ b/ghost/member-attribution/lib/attribution.js
@@ -15,11 +15,19 @@ class Attribution {
      * @param {string|null} [data.id]
      * @param {string|null} [data.url] Relative to subdirectory
      * @param {'page'|'post'|'author'|'tag'|'url'} [data.type]
+     * @param {string|null} [data.refSource]
+     * @param {string|null} [data.refMedium]
+     * @param {URL|null} [data.refUrl]
      */
-    constructor({id, url, type}, {urlTranslator}) {
+    constructor({
+        id, url, type, refSource, refMedium, refUrl
+    }, {urlTranslator}) {
         this.id = id;
         this.url = url;
         this.type = type;
+        this.refSource = refSource;
+        this.refMedium = refMedium;
+        this.refUrl = refUrl;
 
         /**
          * @private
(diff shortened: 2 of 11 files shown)


ghost/member-attribution/lib/attribution.js
```diff
@@ -80,21 +88,52 @@ class AttributionBuilder {
...
+
+        for (const item of history) {
+            const {referrerSource, referrerMedium, referrerUrl} = this.referrerTranslator.getAttributionData(item);
+            if (referrerSource && referrerMedium) {
+                return {
+                    referrerSource,
+                    referrerMedium,
+                    referrerUrl
+                };
+            }
```
SimonBackx, line 127: I'm wondering here that when both `item.referrerSource` and `item.referrerMedium` are null/undefined, that we still get default values. So if you visit
`https://ghost.site/my-post?rel=this-is-referred`
Then later click through to a different post
`https://ghost.site/other-post`

You'll essentially lose that first `rel` because the last one already will return some source and medium defaults. Is this intended behaviour?

ghost/core/core/frontend/src/member-attribution/member-attribution.js
```diff
@@ -71,6 +71,26 @@ const LIMIT = 15;
...
+        let sourceParam;
+        let utmSourceParam;
+        let utmMediumParam;
+        try {
+            // Fetch source/medium from query param
+            const url = new URL(window.location.href);
+            refParam = url.searchParams.get('ref');
+            sourceParam = url.searchParams.get('source');
+            utmSourceParam = url.searchParams.get('utm_source');
+            utmMediumParam = url.searchParams.get('utm_medium');
```
SimonBackx, line 86: Should we remove these params from the URL afterwards?

ghost/member-attribution/lib/attribution.js
rishabhgrg: @SimonBackx Yea i see what you mean, we'll definitely update this to be more smart about how we calculate the referrer source/medium from the history, for example as you said we could maybe first check history for any entry that has `ref` and if not only then fallback to calculating them from URL. 
The current logic to calculate referrer sources just places the pieces in place for where things will happen, and we'll have more iterations on how the values are calculated from history once this is good :)

SimonBackx: Perfect! 👌 Just wanted to make sure 😊

ghost/member-attribution/lib/referrer-translator.js
```diff
@@ -0,0 +1,95 @@
...
+     *
+     * @param {Object} deps
+     * @param {Object} deps.urlUtils
+     */
+    constructor({urlUtils}) {
+        this.urlUtils = urlUtils;
+    }
+
+    getAttributionData({referrerUrl, referrerSource, referrerMedium}) {
+        return {
```
allouis, line 15: Does it make more sense to return `null` here when there is no source or medium?

That way this class is solely responsible for generating referrer data, including what is/isn't a valid referrer

allouis (commented):
Approach looks good - just one main comment about moving the logic of what is/isn't a valid referrer into a single place

ghost/core/core/frontend/src/member-attribution/member-attribution.js
rishabhgrg: 🤔 I was thinking about it too, specially after looking at the fact we remove it for attribution id/sources, but i think these params are different because they are set externally, and there might be other analytics/tools that are setup to rely on them, so imo doesn't make sense for us to remove them, wdyt ?

ghost/member-attribution/lib/attribution.js
rishabhgrg: That's good, thanks 😄

ghost/member-attribution/lib/referrer-translator.js
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     *
+     * @param {Object} deps
+     * @param {Object} deps.urlUtils
+     */
+    constructor({urlUtils}) {
+        this.urlUtils = urlUtils;
+    }
+
+    getAttributionData({referrerUrl, referrerSource, referrerMedium}) {
```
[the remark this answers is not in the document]
rishabhgrg: Yeah right, will update JSDoc types, wanted to make sure pieces are in right place before going all in with those changes :)

ghost/member-attribution/lib/referrer-translator.js
```diff
@@ -0,0 +1,95 @@
...
+ * Translates referrer data into source and medium
+ */
+class ReferrerTranslator {
+    /**
+     *
+     * @param {Object} deps
+     * @param {Object} deps.urlUtils
+     */
+    constructor({urlUtils}) {
+        this.urlUtils = urlUtils;
```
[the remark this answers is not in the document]
rishabhgrg: Yea that works as well, the reason I had left `urlUtils` was in case we'd want to extend the logic in future, will switch to simpler `adminUrl` for now and we can switch back again later. 👍

ghost/member-attribution/lib/referrer-translator.js
rishabhgrg: Yeah that makes sense, though I am wondering if this should change completely where instead of calculating referrer for each individual history item, we might as well pass in the whole history here so it can have some smart calculations on how to derive the `Source` and `Medium` based on different history items. Gonna try that, otherwise switch to returning `null` if it keeps the current check for individual history item

ghost/core/core/frontend/src/member-attribution/member-attribution.js
SimonBackx: You are right about external tools, didn't think about that 🤔

ghost/member-attribution/lib/attribution.js
rishabhgrg: This should be much better now with updated logic to calculate referrer from history :)

ghost/core/core/frontend/src/member-attribution/member-attribution.js
rishabhgrg: Will keep it in for now, and based on how it feels later on testing we can remove it if needed

ghost/member-attribution/lib/attribution.js
```diff
@@ -15,11 +15,19 @@ class Attribution {
      * @param {string|null} [data.id]
      * @param {string|null} [data.url] Relative to subdirectory
      * @param {'page'|'post'|'author'|'tag'|'url'} [data.type]
+     * @param {string|null} [data.referrerSource]
+     * @param {string|null} [data.referrerMedium]
+     * @param {string|null} [data.referrerUrl]
```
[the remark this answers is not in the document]
rishabhgrg: Done

ghost/member-attribution/lib/referrer-translator.js
```diff
@@ -0,0 +1,95 @@
...
+
+    isGhostNewsletter({referrerSource}) {
+        // if refferer source ends with -newsletter
+        if (referrerSource && referrerSource.endsWith('-newsletter')) {
+            return true;
+        }
+    }
+
+    isGhostOrgUrl(url) {
+        if (url.includes('ghost.org')) {
```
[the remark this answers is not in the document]
rishabhgrg: Updated logic to use URL checks where possible instead of string, thanks for the idea.

ghost/member-attribution/lib/referrer-translator.js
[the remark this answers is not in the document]
rishabhgrg: Updated deps to pass in `adminUrl` and `siteUrl` directly ✔️

ghost/member-attribution/lib/referrer-translator.js
rishabhgrg: @allouis Updated the logic here to pass in the `history` directly, as the logic to smartly calculate referrer information needs access to whole history so it can be looped over for different cases. Also now returns `null` instead of object with nulls when no referrer information can be created, lmk if this makes sense.

rishabhgrg:
@allouis @SimonBackx Thanks for the feedbacks. Have updated PR now with suggested changes, including logic cleanup where `ReferrerTranslator` now takes in the `history` directly instead of each history item so it can calculate the referrer information more effectively. This is missing a piece where we use a curated list of external urls for calculating source/medium in the translator, but thats incoming shortly after as a follow-up PR.

PR #15418, opened 2022-09-15
Wired up link redirects & tracking

allouis:
- Uses an in-memory repository for now whilst in development
- Updates the LinkReplacementService to choose the slug
- Exposes a `getSlug` method so we can ensure uniqueness
- Emits the RedirectEvent for use by LinkTracking

### DIFF
--- a/ghost/link-redirects/lib/LinkRedirectsService.js
+++ b/ghost/link-redirects/lib/LinkRedirectsService.js
@@ -1,20 +1,61 @@
+const crypto = require('crypto');
+const DomainEvents = require('@tryghost/domain-events');
+const RedirectEvent = require('./RedirectEvent');
 const LinkRedirect = require('./LinkRedirect');
 
+/**
+ * @typedef {object} ILinkRedirectRepository
+ * @prop {(url: URL) => Promise<LinkRedirect>} getByURL
+ * @prop {(linkRedirect: LinkRedirect) => Promise<void>} save
+ */
+
 class LinkRedirectsService {
+    /** @type ILinkRedirectRepository */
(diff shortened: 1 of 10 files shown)


ghost/link-redirects/lib/LinkRedirectsService.js
```diff
@@ -1,21 +1,62 @@
...
+     * @param {string} slug
      *
      * @returns {Promise<LinkRedirect>}
      */
-    async addRedirect(to) {
-        const from = new URL(to);
-
-        from.searchParams.set('redirected', 'true'); // Dummy for skateboard
+    async addRedirect(to, slug) {
+        const from = new URL(`r/${slug}`, this.#baseURL);
```
SimonBackx, line 51: This is not blocking at all (we can update this later), but wouldn't it be better if the `/r` prefix was part of the `getSlug`, and the `addRedirect` would take a full path instead?

That way we can also use the `LinkRedirectsService` like this later:
```
// Post changed slug (pseudo code)
const oldUrl = post.getUrl({relative: true});
await Post.edit({slug: newSlug}, {id: post.id});
const newUrl = post.getUrl({relative: false});

redirectsService.addRedirect(new URL(newURL), oldUrl);
``` 
(we an also build a completely different service for that too).

Maybe it would also make more sense to put the `slug` parameter first?

allouis: tbh I don't know if we even want a `/r` prefix - this is all just placeholder for now!


RE: slug param first - I went second because I was considering having it optional - still not sure about that 🤷

PR #15425, opened 2022-09-16
Wired up link-replacement to store links against posts in the DB

allouis:
https://github.com/TryGhost/Team/issues/1916

This uses an in-memory store for now which is essentially a noop because we're not doing anything with the data. It should be simple enough to update the repository implementation with the real DB models once they've been created & merged.

### DIFF
--- a/ghost/link-replacement/lib/PostLink.js
+++ b/ghost/link-replacement/lib/PostLink.js
@@ -1,0 +1,22 @@
+/**
+ * @typedef {import('bson-objectid').default} ObjectID
+ */
+
+class PostLink {
+    /** @type {ObjectID} */
+    post_id;
+    /** @type {ObjectID} */
+    link_id;
+
+    /**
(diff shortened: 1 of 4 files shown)


ghost/link-replacement/lib/PostLink.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * @typedef {import('bson-objectid').default} ObjectID
+ */
+
+class PostLink {
+    /** @type {ObjectID} */
+    post_id;
+    /** @type {ObjectID} */
+    link_id;
```
allouis, line 9: @SimonBackx I went for just an ID for now just to get things moving, not sure if we'll eventually want to switch to a full `LinkRedirect` object

SimonBackx: Didn't we want to put the PostLink class in the link-tracking package?

PR #15430, opened 2022-09-19
Added member labels to template options

tomwadley:
Being able to access a member's labels from the theme template allows customizing the user experience based on the type of member.

Trivial example:

```
{{#foreach @member.labels}}
    {{#match name "=" "vip"}}
        Thanks for being a VIP!
    {{/match}}
{{/foreach}}
```

I think there are many potential use cases that would be enabled by allowing labels to be used like this. For example, you could build a simple tier-like system (separate to the existing Stripe tier system) that hides certain post's content based on a member's tier label.

If we're happy exposing labels to theme templates, a useful next step could be adding label support to the `has` handlebars helper to make it easier to query whether or not a member has a given label.

### DIFF
--- a/ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js
+++ b/ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js
@@ -33,6 +33,7 @@ function updateLocalTemplateOptions(req, res, next) {
                 default_payment_card_last4: sub.default_payment_card_last4 || '****'
             });
         }),
+        labels: req.member.labels,
         paid: req.member.status !== 'free',
         status: req.member.status
     } : null;


ErisDS:
Thanks for this PR - I totally agree with this change in principle. 

However, this change is ringing an alarm bell for me because it touches the one place in the code where the theme API isn't directly pulling from the Content API and we haven't got any clear pattern or defined rules here - as you can see it's just a random made up JS object 😬 

I'm slightly concerned about expanding it to include more stuff, when it's not clear what it is and whether it should be public info. Lemme have a little think on this and a chat with the engineers who know the code best, and figure if this is is defo safe to expose.

ErisDS:
I'm sorry, I totally dropped the ball on this PR 😔. It's now absolutely ancient. I'm gonna close it, but if you're still interested in working on this, please re-raise the PR fresh and we'll figure out how to work through it. 

(Please don't comment to request re-open, comments get lost, new PRs trigger new automations and will get our attention)

PR #15432, opened 2022-09-19
Updated source grouping to use known referrers

rshbhgrg:
Instead of collecting and deciding on the Medium/Source for all of these ourselves, we will be leveraging Plausible's already well organized list [here](https://github.com/plausible/analytics/blob/master/priv/ref_inspector/referers.yml) for grouping incoming referrers.

- adds new package with known referrers list from Plausible's list
  - transforms yaml list from Plausible to JSON format, and flattened out with each domain/url as a separate key
- updates attribution logic to use known referrers list for grouping Source and Medium

### DIFF
--- a/ghost/member-attribution/lib/referrer-translator.js
+++ b/ghost/member-attribution/lib/referrer-translator.js
@@ -5,6 +5,8 @@
  * @prop {URL|null} [refUrl]
  */
 
+const knownReferrers = require('@tryghost/referrers');
+
 /**
  * Translates referrer info into Source and Medium
  */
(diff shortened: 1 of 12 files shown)


rishabhgrg:
@ErisDS The main thing I'd love to get your 👀 on here is the new `@tryghost/referrers` package and the pattern it uses to expose the JSON with list of known referrers for attribution service. Currently it's following the structure same as [Nameservers package](https://github.com/TryGhost/nameservers), but was wondering if it will make sense for it to use .js format to expose the list which is closer to pattern used for rest of monorepo packages.

PR #15436, opened 2022-09-19
Added referrer attribution columns to events table

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1931

- referrer source and medium along with url will need to be be stored in the events table along with rest of attribution data
- stores information on two event tables
  - `members_created_events` for newly created members(both free and paid)
  - `members_subscription_created_events` for new paid subscription

Note: Source and Medium are created by us and are not expected to be too long, so have 191 char limit. Referrer URL is stored as Top-level domain only

### DIFF
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -497,6 +497,9 @@ module.exports = {
             }
         },
         attribution_url: {type: 'string', maxlength: 2000, nullable: true},
+        referrer_source: {type: 'string', maxlength: 191, nullable: true},
+        referrer_medium: {type: 'string', maxlength: 191, nullable: true},
+        referrer_url: {type: 'string', maxlength: 2000, nullable: true},
         source: {
             type: 'string', maxlength: 50, nullable: false, validations: {
                 isIn: [['member', 'import', 'system', 'api', 'admin']]
@@ -634,7 +637,10 @@ module.exports = {
                 isIn: [['url', 'post', 'page', 'author', 'tag']]
             }
         },
-        attribution_url: {type: 'string', maxlength: 2000, nullable: true}
+        attribution_url: {type: 'string', maxlength: 2000, nullable: true},
+        referrer_source: {type: 'string', maxlength: 191, nullable: true},
+        referrer_medium: {type: 'string', maxlength: 191, nullable: true},
+        referrer_url: {type: 'string', maxlength: 2000, nullable: true}
     },
     offer_redemptions: {
         id: {type: 'string', maxlength: 24, nullable: false, primary: true},
--- a/ghost/core/core/server/data/migrations/versions/5.16/2022-09-19-17-44-add-referrer-columns-to-member-events-table.js
+++ b/ghost/core/core/server/data/migrations/versions/5.16/2022-09-19-17-44-add-referrer-columns-to-member-events-table.js
@@ -1,0 +1,21 @@
+const {createAddColumnMigration, combineNonTransactionalMigrations} = require('../../utils');
+
+module.exports = combineNonTransactionalMigrations(
+    createAddColumnMigration('members_created_events', 'referrer_source', {
+        type: 'string',
+        maxlength: 191,
+        nullable: true
+    }),
+
+    createAddColumnMigration('members_created_events', 'referrer_medium', {
+        type: 'string',
+        maxlength: 191,
+        nullable: true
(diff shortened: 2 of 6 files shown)


ghost/core/core/server/data/migrations/versions/5.16/2022-09-19-17-44-add-referrer-attribution-columns-to-events-table.js
```diff
@@ -0,0 +1,39 @@
...
+        maxlength: 191,
+        nullable: true
+    }),
+
+    createAddColumnMigration('members_subscription_created_events', 'attribution_ref_url', {
+        type: 'string',
+        maxlength: 2000,
+        nullable: true
+    })
+);
```
rishabhgrg, line 39: @daniellockyer Does all the columns in single migration work better here or should i break this into two separate migrations for two tables ?

ghost/core/core/server/data/schema/schema.js
```diff
@@ -497,6 +497,9 @@ module.exports = {
             }
         },
         attribution_url: {type: 'string', maxlength: 2000, nullable: true},
+        attribution_ref_source: {type: 'string', maxlength: 191, nullable: true},
+        attribution_ref_medium: {type: 'string', maxlength: 191, nullable: true},
+        attribution_ref_url: {type: 'string', maxlength: 2000, nullable: true},
```
ErisDS, line 502: What's the difference between attribution_url and attribution_ref_url?

rishabhgrg: @ErisDS `attribution_url` is the Ghost site URL from which member signed up or paid, for ex. a specific post or site home page. 
`attribution_ref_url`  captures the referrer's URL which led member to that particular post(or other) Ghost site URL, like Ghost Explore or Twitter.

ErisDS: OK - I think we might want to rename these fields to `referrer_url`, `referrer_medium` and `referrer_source` and keep `attribution` to mean our internal post attribution?

ErisDS: @allouis do you agree?

ghost/core/core/server/data/migrations/versions/5.16/2022-09-19-17-44-add-referrer-attribution-columns-to-events-table.js
daniellockyer: Although we have no strict rules here, I think it might be better to keep it to 1 migration, 1 table

Therefore, maybe let's split them apart?

ghost/core/core/server/data/schema/schema.js
rishabhgrg: Main reason for keeping the `attribution_` prefix was to have all attribution data(internal and external) grouped together for the event, but given the above confusion rename sounds good to me 👍

ghost/core/core/server/data/migrations/versions/5.16/2022-09-19-17-44-add-referrer-attribution-columns-to-events-table.js
rishabhgrg: Yea sounds good, updating

ghost/core/core/server/data/schema/schema.js
allouis: @ErisDS I do 👍

ErisDS: Also whilst we're here, before we have any "fun" the correct spelling is "referrer" with 2 R's... or 4 R's if you're a pedant 😂

rishabhgrg: Haha all done to make sure we have the right "referrer" 😄

ghost/core/core/server/data/migrations/versions/5.16/2022-09-19-17-44-add-referrer-attribution-columns-to-events-table.js
rishabhgrg: Updated now ✅

rishabhgrg:
@ErisDS @daniellockyer This is ready for review again after above changes

ErisDS (approved):
LGTM 👍

daniellockyer (approved):

PR #15438, opened 2022-09-20
🪞🪞HTTP caching improvements

naz:
refs https://github.com/TryGhost/Toolbox/issues/410

This PR is here to make tracking of HTTP cache improvements easier to port to Ghost 4.x and get review from the engineers who are maintaining HTTP cache-related features in shared caches.

@jloh @JoeeGrigg I will be adding more caching improvements here based on the doc I've shared privately. Needs your eyes to check if the changes are compatible with our caching infra 🪞🪞 The check for "user specific" request/response is here in the framework's [http-cache-utils pacakge](https://github.com/TryGhost/framework/blob/a7f15287452cb8b50e6c26389f7a44677bdb1542/packages/http-cache-utils/lib/http-cache-utils.js).

### DIFF
--- a/ghost/core/core/frontend/web/middleware/error-handler.js
+++ b/ghost/core/core/frontend/web/middleware/error-handler.js
@@ -7,7 +7,7 @@ const config = require('../../../shared/config');
 const renderer = require('../../services/rendering');
 
 // @TODO: make this properly shared code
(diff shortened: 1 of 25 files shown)


jloh:
LGTM 👍

PR #15446, opened 2022-09-20
Added links API

allouis:
This expose the /links endpoint on the Admin API, which is filterable by Post ID.

We also add in memory repository implementations, just because they exist locally. I don't think we should necessarily merge them with this PR, but it is what I was building this API against initially.

Missing tests & some small bug fixes


ghost/core/core/server/api/endpoints/links.js
```diff
@@ -0,0 +1,26 @@
...
+        options: [
+            'filter'
+        ],
+        permissions: false,
+        async query(frame) {
+            const links = await linkTrackingService.service.getLinks(frame.options);
+
+            return {
+                data: links,
+                meta: {
```
allouis, line 16: Not sure about hardcoding this here, should maybe come from the `getLinks` method?

ghost/core/core/server/api/endpoints/links.js
```diff
@@ -0,0 +1,26 @@
...
+                    pagination: {
+                        total: 1,
```
allouis, line 18: Should be links.length

ghost/core/core/server/services/link-redirection/LinkRedirectRepository.js
```diff
@@ -31,6 +31,22 @@ module.exports = class LinkRedirectRepository {
...
 
+    async getAll(options) {
+        const collection = await this.#LinkRedirect.findAll(options);
+
+        const result = [];
+
+        for (const model of collection.models) {
+            result.push(new LinkRedirect({
+                id: model.id,
+                from: new URL(model.get('from'), this.#urlUtils.getSiteUrl()),
```
allouis, line 42: This needs to correctly work with subdirectories, we need to trim the leading `/` from the from property

ghost/core/core/server/services/link-redirection/LinkRedirectRepository.js
```diff
@@ -31,6 +31,22 @@ module.exports = class LinkRedirectRepository {
...
+                to: new URL(this.#urlUtils.transformReadyToAbsolute(model.get('to')))
```
SimonBackx, line 43: `transformReadyToAbsolute` has moved to the model layer for now because the activity feed also needs to return these. Ideally the activity feed would also use the repository pattern, but with all the relations that would become difficult (because we need click event models, member models, link models and post models) unless we just ignore the whole bookshelf relations and start wiring up the relations manually using repositories. Then do a call to the LinkClickRepository to get the 'events' (and where you need to pass in a repository for members, links and posts), which returns a different kind of object that includes the member, link and posts. Curious to know your opinion on this.

allouis: Ah cool, good to know, thanks!

The plan is to still keep all of the DB code in the models for now, so that everything is in one place for pulling out/improving/replacing in future. I think the repository pattern serves us best in new code so we can easily stub and mock the data layer in development, as well as typing all of our code, we don't need to update the activity feed stuff just yet I don't think :)

ghost/core/core/server/api/endpoints/links.js
SimonBackx: Lets keep it in the endpoint implementation for now. The `getLinks` method shouldn't really know the endpoint structure I think

ghost/link-tracking/lib/LinkClickTrackingService.js
```diff
@@ -62,6 +65,45 @@ class LinkClickTrackingService {
...
+                id: link.link_id.toHexString(),
+                url: link.to.toString(),
+                post_id: postLinks.find((postLink) => {
+                    return postLink.link_id.equals(link.link_id);
+                }).post_id.toHexString(),
+                click_events: events
+            };
+
+            data.push(result);
+        }
```
SimonBackx, line 102: IMO this is where the repository pattern is lacking. We have to do `this.#linkRedirectService.getAll` effectively is returning the same database models in the implementation. And we are fetching all the events of all the links in a loop. I'm not sure why we need to return all the event when we are only going to display the total count.

SimonBackx:
I moved the in memory repositories to a different Pr for now: https://github.com/TryGhost/Ghost/pull/15455

PR #15448, opened 2022-09-21
Added post referrers stats API

SimonBackx:
closes https://github.com/TryGhost/Team/issues/1942

- Added data fixtures for referrers
- Added new endpoint to fetch referrer stats for a given post: `/stats/referrers/posts/:id`
- Added new ReferrersStatsService, responsible for calculating referrer stats

### DIFF
--- a/ghost/stats-service/lib/referrers.js
+++ b/ghost/stats-service/lib/referrers.js
@@ -51,14 +51,6 @@ class ReferrersStatsService {
 
         return [...map.values()].sort((a, b) => b.paid_conversions - a.paid_conversions);
     }
-
-    /**
-     * Return a list of all the attribution sources (with signup and conversion counts) grouped per date
-     * @returns {Promise<AttributionCountStatDate[]>}
-     */
-    async getHistory() {
-        // TODO: implement
-    }
 }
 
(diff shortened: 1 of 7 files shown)


ghost/stats-service/lib/referrers.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Return a list of all the attribution sources for a given post, with their signup and conversion counts
+     * @param {string} postId 
+     * @returns {Promise<AttributionCountStat[]>}
+     */
+    async getForPost(postId) {
+        const knex = this.knex;
+        const sigupRows = await knex('members_created_events')
```
allouis, line 17: Lil typo here - should be `signupRows`

allouis (commented):
Looks good - small comment!

Is the TODO part of another issue? Should we leave the method out all together for now?

SimonBackx:
Yup, Rish is working on the todo here: https://github.com/TryGhost/Ghost/pull/15449 We can probably delete the todo when we merge both changes

ghost/stats-service/lib/referrers.js
SimonBackx: ✅

PR #15449, opened 2022-09-21
Added new referrer history stats API

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/1939

- adds new `/stats/referrers` endpoint that returns historical data of sources by date and count of paid and free signups for each
- tests

### DIFF
--- a/ghost/core/core/server/web/api/endpoints/admin/routes.js
+++ b/ghost/core/core/server/web/api/endpoints/admin/routes.js
@@ -135,6 +135,7 @@ module.exports = function apiRoutes() {
     router.get('/stats/mrr', mw.authAdminApi, http(api.stats.mrr));
     router.get('/stats/subscriptions', mw.authAdminApi, http(api.stats.subscriptions));
     router.get('/stats/referrers/posts/:id', mw.authAdminApi, http(api.stats.postReferrers));
+    router.get('/stats/referrers/history', mw.authAdminApi, http(api.stats.referrersHistory));
 
     // ## Labels
     router.get('/labels', mw.authAdminApi, http(api.labels.browse));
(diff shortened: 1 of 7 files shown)


ghost/core/core/server/web/api/endpoints/admin/routes.js
```diff
@@ -135,6 +135,7 @@ module.exports = function apiRoutes() {
     router.get('/stats/mrr', mw.authAdminApi, http(api.stats.mrr));
     router.get('/stats/subscriptions', mw.authAdminApi, http(api.stats.subscriptions));
     router.get('/stats/referrers/posts/:id', mw.authAdminApi, http(api.stats.postReferrers));
+    router.get('/stats/referrers/history', mw.authAdminApi, http(api.stats.referrersHistory));
```
SimonBackx, line 138:
```suggestion
    router.get('/stats/referrers', mw.authAdminApi, http(api.stats.referrersHistory));
```
The mrr and member_count endpoints also return a history. And I think it is better to stay consistent here. Initially the `/stats/members_count` endpoint also was `/stats/members/count-history`. But on review at the time, that was changed to be shorter, [ref](https://ghost.slack.com/archives/C02G9E68C/p1648737467414789?thread_ts=1648644801.253699&cid=C02G9E68C)

rishabhgrg: Ah cool, I had originally kept is as `/referrers` but changed it to be more explicit later, will change it back :)

rishabhgrg: Updated

ghost/stats-service/lib/referrers.js
```diff
@@ -51,6 +53,83 @@ class ReferrersStatsService {
...
+        });
+
+        paidConversionEntries.forEach((entry) => {
+            const entryDate = moment(entry.date).format('YYYY-MM-DD');
+            const existingEntry = allEntries.find(e => e.source === entry.source && e.date === entryDate);
+
+            if (existingEntry) {
+                existingEntry.paid_conversions = entry.paid_conversions;
+            } else {
+                allEntries.push({
```
SimonBackx, line 80: We probably need to sort in the end because now the dates won't be sorted correctly any longer

rishabhgrg: @SimonBackx Good catch! 🎣 Updated and also added test for sorted dates

PR #15471, opened 2022-09-26
🐛 Fixed default content CTA message for page vs post

rshbhgrg:
closes https://github.com/TryGhost/Team/issues/1898

- the default content cta always used the terminology as `post` when showing message that users don't have access to some content
- this caused confusion when users were looking at a page and message showed "This post is for subscribers only"
- updates the message to correctly reflect `page` vs `post` on the default cta
- updates default cta to use accent color from global site property instead of local value(which is same)

### DIFF
--- a/ghost/core/core/frontend/helpers/content.js
+++ b/ghost/core/core/frontend/helpers/content.js
@@ -18,9 +18,12 @@ const createFrame = hbs.handlebars.createFrame;
 function restrictedCta(options) {
     options = options || {};
     options.data = options.data || {};
+
     _.merge(this, {
+        // @deprecated in Ghost 5.16.1 - not documented & removed from core templates
         accentColor: (options.data.site && options.data.site.accent_color)
     });
+
     const data = createFrame(options.data);
     return templates.execute('content-cta', this, {data});
 }
--- a/ghost/core/core/frontend/helpers/tpl/content-cta.hbs
+++ b/ghost/core/core/frontend/helpers/tpl/content-cta.hbs
@@ -1,6 +1,6 @@
 {{{html}}}
 <aside class="gh-post-upgrade-cta">
-    <div class="gh-post-upgrade-cta-content" style="background-color: {{accentColor}}">
+    <div class="gh-post-upgrade-cta-content" style="background-color: {{@site.accent_color}}">
         {{#has visibility="paid"}}
             <h2>This {{#is "page"}}page{{else}}post{{/is}} is for paying subscribers only</h2>
         {{/has}}
@@ -11,9 +11,9 @@
             <h2>This {{#is "page"}}page{{else}}post{{/is}} is for subscribers on the {{tiers}} only </h2>
         {{/has}}
         {{#if @member}}
-            <a class="gh-btn" data-portal="account/plans" style="color:{{accentColor}}">Upgrade your account</a>
+            <a class="gh-btn" data-portal="account/plans" style="color:{{@site.accentColor}}">Upgrade your account</a>
         {{else}}
-            <a class="gh-btn" data-portal="signup" style="color:{{accentColor}}">Subscribe now</a>
+            <a class="gh-btn" data-portal="signup" style="color:{{@site.accent_color}}">Subscribe now</a>
             <p><small>Already have an account? <a data-portal="signin">Sign in</a></small></p>
         {{/if}}
     </div>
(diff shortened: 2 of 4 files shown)


ghost/core/core/frontend/helpers/content.js
```diff
@@ -19,7 +19,8 @@ function restrictedCta(options) {
     options = options || {};
     options.data = options.data || {};
     _.merge(this, {
-        accentColor: (options.data.site && options.data.site.accent_color)
+        accentColor: (options.data.site && options.data.site.accent_color),
+        resource: options.data.root?.page ? 'page' : 'post'
```
rishabhgrg, line 23: @ErisDS Probably using `#is` helper in the CTA to render `page` vs `post` conditionally instead of this pattern is better here ? Was originally inclined towards this as it gives easy way for custom content helpers to render `page` vs `post`, but realized later the use `#is` here is more consistent without adding any new variables. Wdyt ?

ghost/core/core/frontend/helpers/tpl/content-cta.hbs
```diff
@@ -2,13 +2,13 @@
 <aside class="gh-post-upgrade-cta">
     <div class="gh-post-upgrade-cta-content" style="background-color: {{accentColor}}">
```
ErisDS, line 3: Not sure when this happened, but IMO this should always be `{{@site.accentColor}}`. It's a global property, there's no reason to extend the theme API here and a local copy?

ghost/core/core/frontend/helpers/tpl/content-cta.hbs
```diff
@@ -2,13 +2,13 @@
...
         {{#has visibility="paid"}}
-            <h2>This post is for paying subscribers only</h2>
+            <h2>This {{resource}} is for paying subscribers only</h2>
```
ErisDS, line 5: We already have the `{{#is}}` helper to determine context of whether something is a post or page. Is there a reason why we can't use the existing pattern here?

ErisDS (changes requested):

ghost/core/core/frontend/helpers/tpl/content-cta.hbs
rishabhgrg: [Updated](https://github.com/TryGhost/Ghost/pull/15471#discussion_r979900173) ✅

ghost/core/core/frontend/helpers/content.js
rishabhgrg: Updated. ✅

ghost/core/core/frontend/helpers/tpl/content-cta.hbs
rishabhgrg: Will still keep the local copy of `accentColor` property for themes as otherwise it will be a breaking change for theme devs already relying on that in custom cta helpers, and we can get rid of that in 5.0.

ghost/core/core/frontend/helpers/tpl/content-cta.hbs
```diff
@@ -1,19 +1,31 @@
 {{{html}}}
 <aside class="gh-post-upgrade-cta">
-    <div class="gh-post-upgrade-cta-content" style="background-color: {{accentColor}}">
+    <div class="gh-post-upgrade-cta-content" style="background-color: {{@site.accent_color}}">
         {{#has visibility="paid"}}
-            <h2>This post is for paying subscribers only</h2>
+            {{#is "page"}}
```
ErisDS, line 5: Is there a reason for not doing this inline? e.g. `<h2>This {{#is "page"}}page{{else}}post{{/is}} is for paying subscribers only</h2>`

rishabhgrg: Will switch to this, no reason other than this one felt cleaner to me, but no strong prefs. :)

ghost/core/core/frontend/helpers/content.js
```diff
@@ -19,7 +19,8 @@ function restrictedCta(options) {
     options = options || {};
     options.data = options.data || {};
     _.merge(this, {
-        accentColor: (options.data.site && options.data.site.accent_color)
+        accentColor: (options.data.site && options.data.site.accent_color),
```
ErisDS: Although we can't remove this line with the given changes, we should add a comment above it something like: `// @deprecated in Ghost 5.16.1 never documented & removed from core templates`

ErisDS (approved):
We need to add a deprecation notice but otherwise this is GTG 👍

PR #15475, opened 2022-09-26
[Under discussion] Updated attribution data in member api when missing type

rshbhgrg:
- previously, we hid the whole attribution data from api on missing the attribution type value, as for page/post attribution it didn't make sense to include anything
- with referrer attribution also included as part of attribution object now, we cannot set the whole object as null now as that hides the referrer information


rishabhgrg:
Closed with https://github.com/TryGhost/Ghost/pull/15499

PR #15476, opened 2022-09-27
🐛 Fixed note field keyboard save in admin members form

### LINKED ISSUE #15450: Keyboard shortcut to save does not save data in admin members form
### Issue Summary

When using the cmd+s keyboard shortcut while editing a Member _when the input field is focused_  the change not actually saved - When the page is reloaded, the data is not present.  *This only seems to happen if the input field is focused*.  If no field is focused (no green outline) then the save works as expected.

### Steps to Reproduce

1. default install of ghost 5.15
2. create a member
3. select (focus) the "Note" input and add some content
4. keeping the note field focused (green outline) use the cmd+s "save" shortcut.  The "saved" green button indicator is shown (top right of right admin panel)
5. reload the page - observe the data added to note field is not saved

note: the shortcut works if the field is not focused.

### Ghost Version

5.15

### Node.js Version

16

self hosted (EC2) on Ubuntu 22.04. followed self-host guide from ghost documentation

### Database type

MySQL 8

### Browser & OS version

Chrome (105.0.5195.125) & Mac OS (12.4)

_No response_

### Code of Conduct

AmbroziuBaban:
closes #15450

- the object property "note" was updated only on focus out
  - this is wrong
- the property should be updated on input 
  - this problem occured only for TextArea component

### DIFF
--- a/ghost/admin/app/components/gh-member-settings-form.hbs
+++ b/ghost/admin/app/components/gh-member-settings-form.hbs
@@ -57,7 +57,7 @@
                             @class="gh-member-details-textarea"
                             @tabindex="3"
                             @value={{this.scratchMember.note}}
-                            @input={{fn this.setProperty "note" this.scratchMember.note}}
+                            {{on "input" this.updateProperty}}
                             data-test-input="member-note"
                         />
                         <GhErrorMessage @errors={{this.member.errors}} @property="note" />
--- a/ghost/admin/app/components/gh-member-settings-form.hbs
+++ b/ghost/admin/app/components/gh-member-settings-form.hbs
@@ -57,7 +57,7 @@
                             @class="gh-member-details-textarea"
                             @tabindex="3"
                             @value={{this.scratchMember.note}}
-                            @focus-out={{fn this.setProperty "note" this.scratchMember.note}}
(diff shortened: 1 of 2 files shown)


AmbroziuBaban:
After the last commit, the change passes the lint test.

The current fix contains a function that doesn't need the property name or value passed as params. It extracts them from the event object which is passed by default.

I notice that the Name and Email field have the same behavior. For this 2 fields the bug passed unnoticed because the name is updated in client at the focus out.  but when the "save" request contains the old values since the object update happened after the request was mad to the server.

To reproduce this:
1. Edit the name
2. Hit keyboard shortcut for save (Ctrl + S)
3. Hard refresh the browser (Ctrl + F5)

kevinansfield (approved):

ErisDS:
Hi @AmbroziuBaban thanks so much for this PR. It has now been merged 🎉 and will appear in the next release of Ghost - usually Fridays.

I'm not sure if you found this through hacktoberfest, but I've added the accepted label to this PR to make sure it counts. We have plenty of open issues and ongoing projects for hacktoberfest and beyond. We would love to see you around again 👋

PR #15478, opened 2022-09-27
Extracted cache-control's max age to config

naz:
closes https://github.com/TryGhost/Toolbox/issues/411

@daniellockyer mostly would like to have a second pair of eyes on the new config names used in the `defaults.json` 🙏

### DIFF
--- a/ghost/core/core/shared/config/defaults.json
+++ b/ghost/core/core/shared/config/defaults.json
@@ -131,6 +131,9 @@
         },
         "cors": {
             "maxAge": 86400
+        },
+        "contentAPI": {
+            "maxAge": 0
         }
     },
     "imageOptimization": {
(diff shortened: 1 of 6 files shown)


ghost/core/core/shared/config/defaults.json
```diff
@@ -105,6 +105,9 @@
         "frontend": {
             "maxAge": 0
         },
+        "publicAsset": {
```
daniellockyer, line 108: I'd personally probably opt for plural here, idk something about `publicAsset` sounds wrong. We have plural on `customRedirects` but that's the only other example - not too fussed though

naz: Following the only example is better than introducing a new convention 👌

PR #15480, opened 2022-09-27
Simplified link tracking related tables naming

allouis:
- Removes superfluous "link" from table names
- Fixes type definititon of dropTables util
- Updates & renames models
- Noop existing migrations to avoid unnecessary work


ghost/core/core/server/data/migrations/utils/tables.js
```diff
@@ -38,7 +38,7 @@ function addTable(name, tableSpec) {
 /**
  * Creates migration which will drop a table
  *
- * @param {[string]} names  - names of the tables to drop
```
allouis, line 41: `[string]` means that we accept an array with exactly one string in, whereas `string[]` is an array of zero or more strings

SimonBackx:
What do you think about the `LinkRedirectRepository` and `LinkRedirect` names?

ghost/core/core/server/data/migrations/versions/5.16/2022-09-19-09-04-add-link-redirects-table.js
```diff
@@ -1,10 +1,4 @@
...
-    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
-    from: {type: 'string', maxlength: 2000, nullable: false},
-    to: {type: 'string', maxlength: 2000, nullable: false},
-    post_id: {type: 'string', maxlength: 24, nullable: true, unique: false, references: 'posts.id', setNullDelete: true},
-    created_at: {type: 'dateTime', nullable: false},
-    updated_at: {type: 'dateTime', nullable: true}
-});
+module.exports = {
+    async up() {},
+    async down() {}
```
SimonBackx, line 3: Maybe we need log messages here?

allouis: Good shout 👍

allouis: Channelling your inner @daniellockyer 😝

SimonBackx: Learning from experience 😅

daniellockyer: 😆

ghost/core/core/server/data/migrations/versions/5.17/2022-09-27-13-53-remove-click-tracking-tables.js
```diff
@@ -0,0 +1,3 @@
+const {dropTables} = require('../../utils');
+
+module.exports = dropTables(['link_redirects', 'members_link_click_events']);
```
daniellockyer, line 3: This produces an irreversible migration, which is only something we do in majors. We'll have to switch to some other utils, or only use the `.up` from this

ErisDS (approved):
I have no additional notes.

ghost/core/core/server/data/migrations/versions/5.17/2022-09-27-13-53-remove-click-tracking-tables.js
allouis: Aha okay!

daniellockyer (approved):

PR #15481, opened 2022-09-27
Added E2E test for whole link tracking flow

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1967

- Test if the whole flow works as expected, and works together
- We can test edge cases of all the independent parts in separate tests 
- Adds a basic helper to get an agent for the frontend (spent too much time on a better solution so I decided to keep the existing supertest agent)

### DIFF
--- a/ghost/core/test/integration/services/mega.test.js
+++ b/ghost/core/test/integration/services/mega.test.js
@@ -211,7 +211,7 @@ describe('MEGA', function () {
 
             // Check if click was tracked and associated with the correct member
             const member = await models.Member.findOne({uuid: memberUuid});
-            const clickEvent = await linkClickRepository.getAll({member_id: member.id, link_id: link.link_id.toHexString()})
+            const clickEvent = await linkClickRepository.getAll({member_id: member.id, link_id: link.link_id.toHexString()});
             assert(clickEvent.length === 1, 'Click event was not tracked');
         });
     });
--- a/ghost/core/test/integration/services/mega.test.js
+++ b/ghost/core/test/integration/services/mega.test.js
@@ -127,6 +127,9 @@ describe('MEGA', function () {
             const linkRedirectService = require('../../../core/server/services/link-redirection');
             const linkRedirectRepository = linkRedirectService.linkRedirectRepository;
 
+            const linkTrackingService = require('../../../core/server/services/link-tracking');
+            const linkClickRepository = linkTrackingService.linkClickRepository;
+
             sinon.stub(_mailgunClient, 'getInstance').returns({});
             const sendStub = sinon.stub(_mailgunClient, 'send');
 
@@ -208,12 +211,8 @@ describe('MEGA', function () {
 
             // Check if click was tracked and associated with the correct member
             const member = await models.Member.findOne({uuid: memberUuid});
-            const clickEvent = await models.MemberLinkClickEvent.findOne({
-                member_id: member.id,
-                link_id: link.link_id.toHexString()
-            });
-            assert(member, 'Invalid member');
-            assert(clickEvent, 'Click event was not tracked');
+            const clickEvent = await linkClickRepository.getAll({member_id: member.id, link_id: link.link_id.toHexString()})
+            assert(clickEvent.length === 1, 'Click event was not tracked');
         });
     });
 });
--- a/ghost/core/test/integration/services/mega.test.js
+++ b/ghost/core/test/integration/services/mega.test.js
@@ -6,9 +6,12 @@ const models = require('../../../core/server/models');
 const sinon = require('sinon');
 const assert = require('assert');
 let agent;
-const testUtils = require('../../utils');
-const supertest = require('supertest');
-const configUtils = require('../../utils/configUtils');
+
+async function sleep(ms) {
+    return new Promise((resolve) => {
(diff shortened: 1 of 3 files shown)


ghost/core/test/integration/services/mega.test.js
```diff
@@ -94,4 +101,117 @@ describe('MEGA', function () {
...
+            emailModel.get('status').should.eql('submitted');
+
+            // Get email data (first argument to sendStub call)
+            const emailData = sendStub.args[0][0];
+            const recipientData = sendStub.args[0][1];
+            const replacements = sendStub.args[0][2];
+
+            const recipient = recipientData[Object.keys(recipientData)[0]];
+
+            // Do the actual replacements for the first member, so we don't have to worry about them anymore
```
allouis, line 152: Comment confused me - this does it for all members, right?

ghost/core/test/integration/services/mega.test.js
```diff
@@ -94,4 +101,117 @@ describe('MEGA', function () {
...
+                }
+
+                // Check if the link is a tracked link
+                assert(href.includes('?m=' + memberUuid), href + ' is not tracked');
+
+                if (!firstLink) {
+                    firstLink = new URL(href);
+                }
+            }
+            const links = (await models.LinkRedirect.findAll({post_id: emailModel.get('post_id')})).models;
```
allouis, line 191: Would be cool if we can use the repository or the API for this - otherwise gonna have to update this as part of the DB rename.

allouis (commented):
Nice test! Little comment about using Repository/API over models - just so that we keep the abstraction and can change implementation without breaking tests

ghost/core/test/integration/services/mega.test.js
SimonBackx: No, it takes the first recipient (= member) of the email above, and replaces all the replacements with the replacements defined for that member

SimonBackx: - `replacements` contains the ids and syntax or all the replacements (this is acutally never send to mailgun, but we stub the mailgun client which is our package)
- `recipientData ` contains all the recipients and the replacement data for that recipient

allouis: Ahhh okay - why don't we do it "properly" and use the `index` param of `forEach` to look up the recipient?\

Not saying we should - just asking why we don't ☺️

ghost/core/test/integration/services/mega.test.js
```diff
@@ -94,4 +101,119 @@ describe('MEGA', function () {
...
+                .expect('Location', link.to.href)
+                .expect(302);
+
+            // Since this is all event based, we need to wait a coulple of ms
+            // in the future we should wait for all dispatched events to be completed.
+            await sleep(200);
+
+            // Check if click was tracked and associated with the correct member
+            const member = await models.Member.findOne({uuid: memberUuid});
+            const clickEvent = await models.MemberLinkClickEvent.findOne({
```
allouis, line 211: Sorry I missed this one, but same here!!

ghost/core/test/integration/services/mega.test.js
SimonBackx: I'm confused 🤔

SimonBackx: Can you give an example of what you mean?

ghost/core/test/integration/services/mega.test.js
SimonBackx: Sorry, also missed this one! Only sad thing is that we cannot search on uuid directly (now we search member on uuid, to convert that id back to uuid later in the repository)

allouis: Yeah it's a real shame - this is one of the downsides of having two unique identifiers - we can look at optimising it in future

ghost/core/test/integration/services/mega.test.js
allouis: Something like this:

```js
            replacements.forEach((replacement, index) => {
                 const recipient = recipientData[Object.keys(recipientData)[index]];
                 emailData[replacement.format] = emailData[replacement.format].replace(
                     replacement.match,
                     recipient[replacement.id]
                 );

                 // Also force Mailgun format
                 emailData[replacement.format] = emailData[replacement.format].replace(
                     `%recipient.${replacement.id}%`,
                     recipient[replacement.id]
                 );
             });
```

Or am I completely misunderstanding the situation here?

SimonBackx: The replacements index doesn't have anything to do with the recipients indexes 🤔 Also, it doesn't really make sense to replace a HTML string (intended as a template for multiple recipients, where the template will get replaced depending on who is receiving the email) with data from multiple recipients 🙃 
Note that `emailData[replacement.format]` is referring to the same HTML string.

Its like replacing the first {{firstname}} in the email with the name of the first recipient, and replacing the {{lastname}} in the same email with the lastname of the following recipient. Does that make it a bit more clear? 😊

allouis: Okay so I've totally misunderstood 😁

allouis (approved):

PR #15487, opened 2022-09-27
Allowed storing attribution data on member events

rshbhgrg:
refs https://github.com/TryGhost/Ghost/commit/90034577b84ad37f686c34143415c83792281800

- this change was missed in above commit and would have blocked capturing attribution data for sites without the flag

### DIFF
--- a/ghost/members-events-service/test/event-storage.test.js
+++ b/ghost/members-events-service/test/event-storage.test.js
@@ -104,10 +104,13 @@ describe('EventStorage', function () {
             sinon.assert.calledOnceWithMatch(MemberCreatedEventModel.add, {
                 member_id: '123',
                 created_at: new Date(0),
-                attribution_id: null,
-                attribution_type: null,
-                attribution_url: null,
-                source: 'test'
+                attribution_id: '123',
+                attribution_type: 'post',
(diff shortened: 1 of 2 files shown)


ghost/members-events-service/test/event-storage.test.js
```diff
@@ -178,7 +181,7 @@ describe('EventStorage', function () {
             sinon.assert.calledTwice(subscribeSpy);
         });
 
-        it('filters if disabled', function () {
+        it('works with flag disabled', function () {
```
SimonBackx, line 184: I think I would just remove this test

rishabhgrg: Ah cool will do, originally I did, but then decided to keep it since we are not bumping the flag to GA, just removing it for few cases, so this seemed like a nice check.

SimonBackx (approved):

ghost/members-events-service/test/event-storage.test.js
SimonBackx: Aha, that is true! Same for me, we can definitely keep it 👌

allouis (approved):

PR #15491, opened 2022-09-28
[WIP] Added attribution for importer and other services

rshbhgrg:


ghost/members-api/lib/repositories/member.js
```diff
@@ -337,10 +340,10 @@ module.exports = class MemberRepository {
                 }
             }
         }
-
+        const updatedAttribution = this._memberAttributionService.updateAttributionForSource(data.attribution, source);
```
allouis, line 343: I think this is happening to close to the core of the code, and would be better suited to living at the endpoint/BREAD service level

rishabhgrg: Moved to https://github.com/TryGhost/Ghost/pull/15499

PR #15492, opened 2022-09-28
Added opened, clicked and received email filtering to members

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/1993

- Allows filtering members by opened, clicked and received email
- Adds clicked_links filter relation to Member model.
- Adds emails filter relation to Member model.
- Adds opened_emails filter expansion to Member model.
- Updated GhResourceSelect to be able to only show list posts by setting the `type` attribute to `email`.
- Improved code reuse in `filter-value` component.

### DIFF
--- a/ghost/core/core/server/models/member.js
+++ b/ghost/core/core/server/models/member.js
@@ -47,9 +47,6 @@ const Member = ghostBookshelf.Model.extend({
             key: 'conversion',
             replacement: 'conversions.attribution_id'
         }, {
-            key: 'clicked_links',
-            replacement: 'clicked_links'
-        }, {
             key: 'opened_emails.post_id',
             replacement: 'emails.post_id',
             expansion: 'email_recipients.opened_at:>=emails.created_at' // this is a tiny hack for mongo-knex not supporting value based checks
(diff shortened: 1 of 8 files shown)


ghost/core/core/server/models/member.js
```diff
@@ -46,6 +46,13 @@ const Member = ghostBookshelf.Model.extend({
         }, {
             key: 'conversion',
             replacement: 'conversions.attribution_id'
+        }, {
+            key: 'clicked_links',
+            replacement: 'clicked_links'
```
ErisDS, line 51: Why are we replacing like for like here?

SimonBackx: Hmmm... Just tested it again without this rule and it works. I added this rule (which indeed is a bit weird) because it wasn't transforming the `clicked_links` to the proper name in the query (and got errors from that). Maybe a temporary issue with code that wasn't reloaded properly. Removed it now ✅

ghost/core/core/server/models/member.js
```diff
@@ -46,6 +46,13 @@ const Member = ghostBookshelf.Model.extend({
...
+        }, {
+            key: 'opened_emails.post_id',
+            replacement: 'emails.post_id',
+            expansion: 'email_recipients.opened_at:>=emails.created_at' // this is a tiny hack for mongo-knex not supporting value based checks
```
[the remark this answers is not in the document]
allouis: Don't quite understand why we need this additional constraint

allouis (approved):
Looks good - just one q

PR #15493, opened 2022-09-28
Added back button from editor to analytics

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/1992

When you edit a post via the editor, the back button on the top of the editor should go back to the analytics page and the status at the top should be removed.


SimonBackx:
@kevinansfield Before I fix tests for this, I wanted to know what you think of this approach. Currently you won't be able to go to the editor if the post also has an analytics page. So this keeps the solution simple. An alternative would be to add some parameter to the editor so it know what the previous page was, but I don't really know how to do that. What do you think about this?

PR #15495, opened 2022-09-28
Ghost Explore in Ghost

aileen:
no issue

- Added Ghost Explore screen behind alpha flag
- Moved existing `/explore` route to `/explore/connect` which we'll redirect to for outside requests
- Added iframe communication with Ghost Explore


ghost/admin/app/components/gh-explore-iframe.js
```diff
@@ -0,0 +1,45 @@
...
+            this.explore.getExploreIframe().contentWindow.postMessage({
+                query: 'getSiteData',
+                response: 'siteData'
+            }, '*');
+        }
+    }
+
+    _handleSiteDataUpdate(data) {
+        // TODO: now we received the site data, we can show the fireworks and the actual data
+        this.explore.siteData = data.siteData;
```
aileen, line 43: @peterzimon This is where we receive the site data, so we can trigger confetti and the success modal with that data.

ghost/admin/app/components/gh-explore-iframe.js
```diff
@@ -0,0 +1,50 @@
+import Component from '@ember/component';
```
kevinansfield, line 1:
```suggestion
import Component from '@glimmer/component';
```

ghost/admin/app/controllers/application.js
```diff
@@ -19,6 +20,9 @@ export default class ApplicationController extends Controller {
     @reads('config.hostSettings.billing.enabled')
         showBilling;
 
+    @reads('explore.enabled')
+        showExplore;
+
```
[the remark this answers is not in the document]
kevinansfield: Or better yet, use `{{#if (feature "exploreApp")}}...{{/if}}` in templates to match other feature flag usage.

ghost/admin/app/templates/application.hbs
```diff
@@ -14,6 +14,9 @@
             {{#if this.showBilling}}
                 <GhBillingModal @billingWindowOpen={{this.billing.billingWindowOpen}} />
             {{/if}}
+            {{#if (feature "exploreApp")}}
+                <GhExploreModal @exploreWindowOpen={{this.explore.exploreWindowOpen}} />
```
kevinansfield, line 18: Minor thing but there's no need to pass `@exploreWindowOpen` through as an argument, the component's `visibilityClass` getter can use `this.explore.exploreWindowOpen` directly and then there's no need to inject the service on the application controller.

ghost/admin/app/components/gh-nav-menu/main.js
```diff
@@ -15,6 +15,7 @@ import {task} from 'ember-concurrency';
 @tagName('')
 export default class Main extends Component.extend(ShortcutsMixin) {
     @service billing;
+    @service explore;
```
kevinansfield, line 18: This isn't used anywhere since switching to the `{{feature}}` helper 🙂

PR #15499, opened 2022-09-29
Added referrer attribution from request context

rshbhgrg:
closes https://github.com/TryGhost/Team/issues/2007

- uses request context to add referrer source and medium for a new member
- uses integration name as referrer medium if exists

### DIFF
--- a/ghost/member-attribution/lib/service.js
+++ b/ghost/member-attribution/lib/service.js
@@ -36,9 +36,7 @@ class MemberAttributionService {
             // If context has integration, set referrer medium as integration anme
             if (context?.integration?.id) {
                 try {
-                    const integration = await this.models.Integration.findOne({id: context.integration.id}, {
-                        withRelated: ['api_keys', 'webhooks']
-                    });
+                    const integration = await this.models.Integration.findOne({id: context.integration.id});
                     attribution.referrerMedium = integration?.get('name');
                 } catch (error) {
                     // return null for integration name if not found
--- a/ghost/member-attribution/lib/service.js
+++ b/ghost/member-attribution/lib/service.js
@@ -110,10 +110,6 @@ class MemberAttributionService {
      * @returns {import('./attribution').AttributionResource|null}
      */
     getEventAttribution(eventModel) {
-        if (eventModel.get('attribution_type') === null) {
-            return null;
-        }
-
         const _attribution = this.attributionBuilder.build({
             id: eventModel.get('attribution_id'),
             url: eventModel.get('attribution_url'),
@@ -123,7 +119,7 @@ class MemberAttributionService {
             referrerUrl: eventModel.get('referrer_url')
         });
 
-        if (_attribution.type !== 'url') {
+        if (_attribution.type && _attribution.type !== 'url') {
             // Find the right relation to use to fetch the resource
             const tryRelations = [
                 eventModel.related('postAttribution'),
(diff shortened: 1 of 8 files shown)


ghost/member-attribution/lib/service.js
```diff
@@ -14,6 +14,42 @@ class MemberAttributionService {
...
+                attribution.referrerSource = 'Created Manually';
+            } else if (source === 'api') {
+                attribution.referrerSource = 'Created via API';
+            }
+
+            // If context has integration, set referrer medium as integration anme
+            if (context?.integration?.id) {
+                try {
+                    const integration = await this.models.Integration.findOne({id: context.integration.id}, {
+                        withRelated: ['api_keys', 'webhooks']
```
allouis, line 40: Why are we fetching the api_keys and webhooks?

ghost/members-api/lib/services/member-bread.js
```diff
@@ -254,6 +254,10 @@ module.exports = class MemberBREADService {
         let model;
 
         try {
+            const attribution = await this.memberAttributionService.getAttributionFromContext(options?.context);
+            if (attribution) {
+                data.attribution = attribution;
```
allouis, line 259: This should be of type `Attribution` rather than `AttributionResource`

rishabhgrg: 🤔 What we pass down to the member create is actually just a plain resource object and not `Attribution` object, which is why I had added `AttributionResource` here. We don't pass down the `Attribution` object to the event, if we use `.build(...)` here would that need a `fetchResource` again down the line ?

ghost/member-attribution/lib/service.js
rishabhgrg: Yea that'll go away in final cleanup, came in with when pasting for quick hack locally.

allouis: Okay cool

ghost/members-api/lib/services/member-bread.js
allouis: Hmm, the definition for the `create` and `linkSubscription` methods both expect an Attribution object - so I think we either need to fix it here, or in there!

rishabhgrg: Ah yea so it will need an update in `create` and `linkSubscription` i think, will update.

ghost/member-attribution/lib/service.js
rishabhgrg: Updated ✅

ghost/members-api/lib/services/member-bread.js
rishabhgrg: Updated.

allouis: Thank uu!!

allouis (approved):

ghost/member-attribution/lib/service.js
```diff
@@ -63,7 +111,15 @@ class MemberAttributionService {
...
+            return {
+                id: null,
+                url: null,
+                title: null,
+                type: null,
+                referrerSource: null,
+                referrerMedium: null,
+                referrerUrl: null
+            };
         }
```
SimonBackx, line 123: In the activity feed we return this, while not in the activity feed we return `getSubscriptionCreatedAttribution`, which is 
```js
const attribution = this.attributionBuilder.build({
            id: subscriptionCreatedEvent.get('attribution_id'),
            url: subscriptionCreatedEvent.get('attribution_url'),
            type: subscriptionCreatedEvent.get('attribution_type'),
            referrerSource: subscriptionCreatedEvent.get('referrer_source'),
            referrerMedium: subscriptionCreatedEvent.get('referrer_medium'),
            referrerUrl: subscriptionCreatedEvent.get('referrer_url')
        });
        return await attribution.fetchResource();
```

rishabhgrg: Ah nice catch! I was not sure if it was a good idea to let the builder run here if type was `null`, but think the check for type in line below should cover it, pushing this, thanks.

PR #15501, opened 2022-09-29
Added email track clicks column and cleaned up frontend checks

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2008

- New column that stores email click tracking at the time it was created
- Improved frontend side checks for when to show analytics


ErisDS (approved):
LGTM, I don't know as much as @daniellockyer but afaik add column uses algorithm=copy now, and is therefore safe to do on any table.

PR #15508, opened 2022-09-30
Used the higher of click count and open count for email open count

allouis:
We process clicks much faster than we process Mailgun events which can result in a higher click rater than open rate shown on the dashboard. This ensures that the open rate will never be lower than the click rate. This is a stopgap solution until we can get click events updating the opened_at time for email_recipients


SimonBackx (approved):
Looks good! Only needs to be merged into the v5.17.1 patch branch

PR #15511, opened 2022-09-30
🐛 Fixed active state bug in sidebar nav

### LINKED ISSUE #15506: Menu items lose active state when "New ..." modal is open
### Issue Summary

The relevant item in the left-hand menu should remain bold when the "New X" modal is open

image

image

### Steps to Reproduce

1. Open Ghost admin
2. Click on tags, note that "Tags" in the left menu is bold
3. Click on "New tag" and notice that "Tags" in the left menu is no longer bold

Same for Members, Offers, etc.

### Ghost Version

5.16

### Node.js Version

16

ghost install

### Database type

MySQL 8

### Browser & OS version

_No response_

_No response_

### Code of Conduct

stephensauceda:
Should close #15506.

The sidebar items that _don't_ open the editor were missing some relative `current-when` attributes so the nested views weren't considered "active".

### DIFF
--- a/ghost/admin/tests/helpers/members.js
+++ b/ghost/admin/tests/helpers/members.js
@@ -13,3 +13,9 @@ export function disableMembers(server) {
         ? server.db.settings.update({key: 'members_signup_access'}, {value: 'none'})
         : server.create('setting', {key: 'members_signup_access', value: 'none', group: 'members'});
 }
+
+export function enablePaidMembers(server) {
+    server.db.settings.find({key: 'paid_members_enabled'})
+        ? server.db.settings.update({key: 'paid_members_enabled'}, {value: true})
+        : server.create('setting', {key: 'paid_members_enabled', value: true, group: 'members'});
+}
--- a/ghost/admin/app/components/gh-nav-menu/main.hbs
+++ b/ghost/admin/app/components/gh-nav-menu/main.hbs
@@ -87,18 +87,18 @@
                         {{/if}}
                     </li>
                     {{#if this.showTagsNavigation}}
-                        <li><LinkTo @route="tags" data-test-nav="tags">{{svg-jar "tag"}}Tags</LinkTo></li>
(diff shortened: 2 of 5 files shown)


stephensauceda:
Added some tests for tags and members. Still in the process of doing so for offers. Noticing the `Offers` menu item isn't present when any of the `acceptance/offers-test.js` tests run. Implies that `settings.paidMembersEnabled : false` in the test environment though it looks like the all the criteria for it being true are present in the fixture. New to Ember so still fumbling about.

ghost/admin/tests/helpers/members.js
```diff
@@ -13,3 +13,9 @@ export function disableMembers(server) {
         ? server.db.settings.update({key: 'members_signup_access'}, {value: 'none'})
         : server.create('setting', {key: 'members_signup_access', value: 'none', group: 'members'});
 }
+
+export function enablePaidMembers(server) {
+    server.db.settings.find({key: 'paid_members_enabled'})
+        ? server.db.settings.update({key: 'paid_members_enabled'}, {value: true})
+        : server.create('setting', {key: 'paid_members_enabled', value: true, group: 'members'});
+}
```
stephensauceda, line 21: I'm not actually sure this is totally the "right" way to go about this. But it's what I found that works, at least. Alternatives I tried:

1.  adding `paid_members_enabled` directly to `mirage/fixtures/settings.js` and then loading that in with `loadFixtures` in the test file. 
2. manually creating the setting with `this.server.create('setting', ...)` in the `beforeEach` of the `beforeEach` of the `as Owner` block of tests.

Neither worked for some reason though I don't ever that could be 💯 user error.

kevinansfield: This is the right way to do it as it keeps it explicit 👍 The other two approaches are also valid in some cases though so it's interesting they didn't work 🤔

kevinansfield (approved):

ErisDS:
Hey @stephensauceda thanks so much for this PR 🙏 It's merged now and I've added the `hacktoberfest-approved` label 🎉 . 

This fix will go out in the next release of Ghost (usually Fridays). 

We have many more issues open and more being created all the time, we'd love to see you around again soon 👋

PR #15514, opened 2022-10-02
🐛 Fixed broken `close` buttons on modals

ronaldlangeveld:
Certain modals have close buttons as `<a>` elements. Those containing `{{on "click" @close}}` and `href=""` would not take the `on click` function into account, meaning the whole page would get redirected to `"/"` (back to the root of the dashboard) as a result of the `href` instead of closing the modal.
From my understanding `{{on "click" @close}}` doesn't enforce `preventDefault` behaviour.

Reproducible example:
Go to members, apply some filters, click ⚙️ and select "Add label for selected members". The modal should open.
Click the Close (X) button top right of the modal. 

This affects more modals, this is just one example. 

Version: [5.16.2](https://github.com/TryGhost/Ghost/releases/tag/v5.16.2)
Ghost Pro
Chrome and Safari

- [x] There's a clear use-case for this code change, explained


kevinansfield (approved):
LGTM 👍 Very minor thing but the `role="button"` attributes can be removed on actual `<button>` elements because they're redundant.

PR #15522, opened 2022-10-03
🐛 Fixed redirect to signin modal not shown when logged out

### LINKED ISSUE #15291: Regression: Redirect to signin/re-authenticate modal not shown when logged out
We've had a regression with how unauthenticated requests are handled by Admin. Unauthenticated requests could occur if you've logged out in another tab and didn't realise you'd gone back to editing on a tab that looked like it was still logged in.

## Steps to reproduce 

1. Open Ghost admin and ensure you're logged in in two different tabs
2. In tab 1, open the editor
3. In tab 2, logout
4. In tab 1, attempt to save

## Previous behaviour
- whilst editing - show a re-authentication modal to allow quick sign-in and save
- whilst performing any other action - redirect to signin screen

## Current behaviour:
- all actions - show a slightly cryptic alert message `Authorization failed Unable to determine the authenticated user or integration. Check that cookies are being passed through if using session authentication.`

This is likely happening because we've moved from a `401` to a `403`. The cookie is being sent so the request is not "Unauthenticated" but it no longer matches a valid session so its "Unauthorized".

Admin isn't currently set up to recognise a 403 and trigger the auto-logout/re-authenticate behaviour. There are potential edge cases if we treat 403 and 401 the same when it's a genuine "not permitted" error but Admin shouldn't be making any such requests unless your role is changed by another user whilst you're using it.

Note: There is a skipped test for the re-authentication modal here: https://github.com/TryGhost/Ghost/blob/0695f74a653bff85d5a08ee7f97fc9a4a8ddbe63/ghost/admin/tests/acceptance/authentication-test.js#L97 which may have contributed to this regression slipping in. Ideally we'd get this test running again as well to prevent this regressing in future.

### Ghost Version
5.10

h2akim:
fixes #15291

- An attempt to improve re-authenticate modal toggle - show re-authenticate modal everytime use save (ctrl/cmd + s)
- An attempt to fix redirection when user re-login on different tab. Prevent redirection to sign-in page since the user already logged in on another tab.
- Re-enable `editor` test on `authentication-test.js`

### DIFF
--- a/ghost/admin/app/controllers/editor.js
+++ b/ghost/admin/app/controllers/editor.js
@@ -124,6 +124,7 @@ export default class EditorController extends Controller {
 
     _leaveConfirmed = false;
     _previousTagNames = null; // set by setPost and _postSaved, used in hasDirtyAttributes
+    _reAuthenticateModalToggle = false;
 
     /* computed properties ---------------------------------------------------*/
 
@@ -476,6 +477,9 @@ export default class EditorController extends Controller {
 
             post.set('statusScratch', null);
 
+            // Clear any error notification (if any)
+            this.notifications.clearAll();
+
             if (!options.silent) {
                 this._showSaveNotification(prevStatus, post.get('status'), isNew ? true : false);
             }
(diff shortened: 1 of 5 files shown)


ErisDS:
@h2akim Thanks for this PR - I notice it is still in a draft state & not ready for review, is that correct?

h2akim:
@ErisDS Yes. Correct, I'll probably submit this PR for review later.

kevinansfield (approved):

ErisDS:
Hey @h2akim thanks again, this PR is also merged and labelled now. You're really racking up those bug fixes, love it ❤️ Keep them coming!

PR #15527, opened 2022-10-04
🔒 Fixed magic link endpoint sending multiple emails

allouis:
Without validation it was possible to send a string of comma separated email addresses to the endpoint, and an email would be sent to each address, bypassing any rate limiting.

### DIFF
--- a/ghost/magic-link/package.json
+++ b/ghost/magic-link/package.json
@@ -25,6 +25,8 @@
   },
   "dependencies": {
     "@tryghost/errors": "1.2.17",
+    "@tryghost/tpl": "0.1.18",
+    "@tryghost/validator": "0.1.29",
     "jsonwebtoken": "8.5.1"
   }
 }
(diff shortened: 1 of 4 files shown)


allouis:
@daniellockyer The changes to `yarn.lock` don't look correct - what is the proper way to add dependencies to our packages? I've used `yarn add` here.

Should I also be updating these dependencies to use version `0.0.0` or is that for `core` only?

ghost/magic-link/package.json
```diff
@@ -25,6 +25,8 @@
   },
   "dependencies": {
     "@tryghost/errors": "1.2.17",
+    "@tryghost/tpl": "^0.1.18",
```
naz, line 28: What's the convention these days for intenal packages regarding versioning? Do we allow for "loose" aka "^x.x.x" or still should be doing "strict" versioning aka "x.x.x"?

daniellockyer: All dependencies that appear in package.json files in this repo should be pinned to a specific version

Even if you don't commit that, Renovate will open a PR to pin them afterwards

PR #15538, opened 2022-10-06
Swapped member limit verification trigger event

naz:
closes https://github.com/TryGhost/Toolbox/issues/399

- The MemberCreatedEvent event is more accurate representation of the limit nature - counting the number of members created. The previous MemberSubscribeEvent was slightly hacky solution because a member could be subscribed/unsubscribed multiple times and distorting the limit counts.

Yo @allouis! The part I wanted to check on here is extraction of a separate `getCreatedEvents` method in the event repository. Was thinking of using the `getSignupEvents` originally, but it had mixed returns based on the `memberAttribution` flag, so decided to go with a very explicit separate method (less probability for events to mixup in the future). Lemme know if you'd do things differently!

### DIFF
--- a/ghost/members-api/lib/repositories/event.js
+++ b/ghost/members-api/lib/repositories/event.js
@@ -267,6 +267,10 @@ module.exports = class EventRepository {
             };
         }
 
+        return this.getCreatedEvents(options, filters);
+    }
+
+    async getCreatedEvents(options = {}, filters = {}) {
         options = {
             ...options,
             withRelated: ['member', 'postAttribution', 'userAttribution', 'tagAttribution'],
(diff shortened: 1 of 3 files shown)


ghost/members-api/lib/repositories/event.js
```diff
@@ -95,9 +95,6 @@ module.exports = class EventRepository {
         if (filters['data.created_at']) {
             options.filter.push(filters['data.created_at'].replace(/data.created_at:/g, 'created_at:'));
         }
-        if (filters['data.source']) {
```
allouis, line 98: Is this definitely not used anywhere else?

allouis (approved):
Looks good to me! One comment but if you're happy den I am too

ghost/members-api/lib/repositories/event.js
naz: fair point. this parameter seems like a dynamic one from the filters. best to leave it as is :)

PR #15552, opened 2022-10-07
Removed Bluebird from all non-core packages

daniellockyer:
refs https://github.com/TryGhost/Ghost/issues/14882

- this removes Bluebird from all other packages in the monorepo apart from core
- this should make it easier to move forwards with stripping the monorepo of Bluebird in favor of native Promises


ErisDS:
epic!

Issue #15537, opened 2022-10-06
Webhook test coverage 


naz:
Webhooks have been a part of Ghost since very first versions. This area has been constantly lagging behind with solid test coverage. Having tests would add confidence to webhooks stability. Snapshot tests would be a great source of example data webhooks deliver to external integrations. 

To improve this situation, Ghost has introduced a special purpose Webhooks snapshot tests. These are snapshot-based tests (similar to the[ end-to-end API tests](https://github.com/TryGhost/Ghost/tree/a6c76dadc3c369d61e9e7fb3e65c42570efb9273/ghost/core/test/e2e-api/admin)) with one big difference - they test the outgoing request payload instead of API responses.

To understand how end-to-end testing works at Ghost have a read through [this guide](https://ghost.notion.site/End-to-end-Testing-6a2ef073b1754b18aff42e24a632a007). The concepts applied in end-to-end API tests are very similar to ones used in webhook tests.

Here is an [example of a webhook test suite](https://github.com/TryGhost/Ghost/blob/78c97d10a6f6a86c7b132534bcc71a6623688799/ghost/core/test/e2e-webhooks/posts.test.js) for for `post.published` and `post.added` events.

The above suite covers two event's which Webhook events listen to. There are 29 total events Webhooks can respond to, 27 left to cover. We need help writing tests for them!  

Below are some pointers around what you could help us with. For the most part it's covering all of the webhook events with at least one "happy path" test.

### 1. Write a test
First, find one webhook event listener to cover - search for an event that has no test coverage in the list of [all webhook event listeners](https://github.com/TryGhost/Ghost/blob/66f11d5e4548db3b3c4d9242e5e217863aac088b/ghost/core/core/server/services/webhooks/listen.js#L12-L45).

<details>
<summary>Copy of all events below for quick reference and ✅ for the ones with at least one test:</summary>

- [x] 'site.changed' (https://github.com/TryGhost/Ghost/pull/15595)
- [x] 'post.added' (https://github.com/TryGhost/Ghost/commit/b536cf5d9c93dd3dde6dca05c82d6761384c2fb3)
- [x] 'post.deleted' (https://github.com/TryGhost/Ghost/pull/15572)
- [x] 'post.edited' (https://github.com/TryGhost/Ghost/pull/15625)
- [x] 'post.published' (https://github.com/TryGhost/Ghost/commit/0f4aeaaa802738f06aed066ece39f9666f2ec1ce)
[...]

PR #15555, opened 2022-10-07
Added e2e tests for `tag.edited` webhook

dshubhadeep:
Added e2e tests for tag.edited webhook

Ref - #15537


dshubhadeep:
@daniellockyer apologies. I've updated the message.

ghost/core/test/e2e-webhooks/tags.test.js
```diff
@@ -100,4 +100,52 @@ describe('tag.* events', function () {
...
+    
+        webhookMockReceiver
+            .matchHeaderSnapshot({
+                'content-version': anyContentVersion,
+                'content-length': anyNumber,
+                'user-agent': anyGhostAgent
+            })
+            .matchBodySnapshot({
+                tag: {
+                    current: {...tagSnapshot, url: anyLocalURL},
```
naz, line 146: Imo, worth including the `url: anyLocalURL` bit in the shared `tabSnapshot` object now that it's been used in multiple places. Have you tried moving it @dshubhadeep ?

naz (changes requested):
Minor refactor request, if it's possible. The rest is 👍

ghost/core/test/e2e-webhooks/tags.test.js
dshubhadeep: Yes, it was done like that before. But tests were failing for tag.deleted webhook since the `previous` key in the snapshot didn't have the URL.
Maybe I could create a new variable something like snapshotWithUrl ?

naz: Lets keep the extraction to the [rule of 3](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)). If we have to reuse the same form 3rd time lets extract a `tagSnapshotWithURL` like you proposed or create a builder function having a "withURL: boolean" parameter. It's been used 2 times now, so it's good to go as is 👍

naz (approved):

ErisDS:
Hi @ thanks so much for this PR. Sorry for the delay, but this has now been merged 🎉 and will appear in the next release of Ghost - usually Fridays.

I'm not sure if you found this through hacktoberfest, but I've added the accepted label to this PR to make sure it counts. We have plenty of open issues and ongoing projects for hacktoberfest and beyond. We would love to see you around again 👋

PR #15560, opened 2022-10-07
🐛 Fixed member importer crash for failed imports

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2004

- for imports, members are created inside a transaction, which causes the member created events to be dispatched.
- its possible that transactions for import can be rolled back if for some reason there is an error down the line while inserting other member properties. The rollback doesn't commit the member to DB, but the event dispatched earlier will still try to create the member created event which fails due to missing member id.
- knex transactions resolve the `executionPromise` both in case of explicit commit or rollback from the user, so just the transaction end check will not be good enough to make sure the member exists in DB
- adds explicit config to knex to reject transaction in case of explicit rollback, which is then caught and event is not dispatched

### DIFF
--- a/ghost/members-api/lib/repositories/member.js
+++ b/ghost/members-api/lib/repositories/member.js
@@ -91,6 +91,19 @@ module.exports = class MemberRepository {
         });
     }
 
+    dispatchEvent(event, options) {
+        if (options?.transacting) {
+            // Only dispatch the event after the transaction has finished
+            options.transacting.executionPromise.then(async () => {
+                DomainEvents.dispatch(event);
+            }).catch(() => {
+                // catches transaction errors/rollback to not dispatch event
+            });
+        } else {
+            DomainEvents.dispatch(event);
+        }
+    }
+
     isActiveSubscriptionStatus(status) {
         return ['active', 'trialing', 'unpaid', 'past_due'].includes(status);
     }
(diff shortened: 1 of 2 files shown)


ghost/members-api/lib/repositories/member.js
```diff
@@ -337,12 +337,24 @@ module.exports = class MemberRepository {
...
+            // Only dispatch the event after the transaction has finished
+            options.transacting.executionPromise.then(async () => {
+                DomainEvents.dispatch(MemberCreatedEvent.create({
+                    memberId: member.id,
+                    attribution: data.attribution,
+                    source
+                }, eventData.created_at));
+            }).catch(() => {
+                // ignore errors for transaction rollback and don't dispatch event
+            });
```
SimonBackx, line 350: Just so I understand this correctly, what happens when we don't add an empty catch? + we don't have this catch in `linkSubscription`

SimonBackx:
We still do:
```
DomainEvents.dispatch(MemberSubscribeEvent.create({
      memberId: member.id,
      source: source
  }, eventData.created_at));
```
But I couldn't find any listeners for this event. This isn't a big problem now, but can we delete this or also wrap it with the execution promise?

SimonBackx (approved):

ghost/members-api/lib/repositories/member.js
rishabhgrg: @SimonBackx If we don't catch it then the error floats up in case of a transaction rollback and we get `Unhandled Rejection`. The catch here ensures we don't do anything wrt event if for any reason transaction is not committed, which should ideally only happen for the importer. 
I have extracted this out in a common method now and applied to all usages in member repo, so this should help us cover any suspect event dispatch.

SimonBackx (approved):

PR #15580, opened 2022-10-10
Added subscriptions table creation migration

naz:
refs https://github.com/TryGhost/Team/issues/2030

- Adds `subscriptions` table to the DB schema.
- The new table is aimed to support a native "subscription" primitive in Ghost that most resembles previously used `members_stripe_customers_subscriptions` table

WIP for the upcoming `subscriptions` table changes.

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.19/2022-10-10-06-58-add-subscriptions-table.js
+++ b/ghost/core/core/server/data/migrations/versions/5.19/2022-10-10-06-58-add-subscriptions-table.js
@@ -1,0 +1,19 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('subscriptions', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    type: {type: 'string', maxlength: 50, nullable: false},
+    status: {type: 'string', maxlength: 50, nullable: false},
+    member_id: {type: 'string', maxlength: 24, nullable: false, unique: false, references: 'members.id', cascadeDelete: true},
+    tier_id: {type: 'string', maxlength: 24, nullable: true, unique: false, references: 'products.id'},
+    cadence: {type: 'string', maxlength: 50, nullable: true},
+    currency: {type: 'string', maxlength: 50, nullable: true},
+    amount: {type: 'integer', nullable: true},
+    payment_provider: {type: 'string', maxlength: 50, nullable: true},
+    payment_subscription_url: {type: 'string', maxlength: 2000, nullable: true},
+    payment_user_url: {type: 'string', maxlength: 2000, nullable: true},
+    offer_id: {type: 'string', maxlength: 24, nullable: true, unique: false, references: 'offers.id'},
+    expires_at: {type: 'dateTime', nullable: true},
+    created_at: {type: 'dateTime', nullable: false},
+    updated_at: {type: 'dateTime', nullable: true}
+});
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -605,6 +605,43 @@ module.exports = {
         updated_at: {type: 'dateTime', nullable: true},
         updated_by: {type: 'string', maxlength: 24, nullable: true}
     },
+    subscriptions: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        type: {
+            type: 'string', maxlength: 50, nullable: false, validations: {
+                isIn: [['free', 'comped', 'paid']]
+            }
+        },
+        status: {
+            type: 'string', maxlength: 50, nullable: false, validations: {
+                isIn: [['active', 'expired', 'cancelled']]
+            }
+        },
+        member_id: {type: 'string', maxlength: 24, nullable: false, unique: false, references: 'members.id', cascadeDelete: true},
+        tier_id: {type: 'string', maxlength: 24, nullable: true, unique: false, references: 'products.id'},
+
+        // These are null if type !== 'paid'
+        cadence: {
+            type: 'string', maxlength: 50, nullable: true, validations: {
+                isIn: [['month', 'year']]
+            }
+        },
+        currency: {type: 'string', maxlength: 50, nullable: true},
+        amount: {type: 'integer', nullable: true},
+
+        // e.g. 'stripe'
+        payment_provider: {type: 'string', maxlength: 50, nullable: true},
+        // e.g. Stripe Subscription Link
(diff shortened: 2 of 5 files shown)


ghost/core/core/server/data/schema/schema.js
```diff
@@ -600,6 +600,45 @@ module.exports = {
...
+        // These are null if type !== 'paid'
+        cadence: {
+            type: 'string', maxlength: 50, nullable: true, validations: {
+                isIn: [['month', 'year']]
+            }
+        },
+        currency: {type: 'string', maxLength: 3, nullable: true},
+        amount: {type: 'integer', nullable: true},
+        
+        // e.g. 'stripe'
```
naz, line 627: I copied these example values from the spec and left them as I liked the idea of having examples right next to the column definition. How do you @daniellockyer @allouis feel about having slightly more information next to the schema definition?

ghost/core/core/server/data/schema/schema.js
```diff
@@ -600,6 +600,45 @@ module.exports = {
...
+        // e.g. Stripe Subscription Link
+        payment_subscription_url: {type: 'string', maxlength: 2000, nullable: true},
+        // e.g. Stripe Customer Link
+        payment_user_url: {type: 'string', maxlength: 2000, nullable: true},
+      
+        offer_id: {type: 'string', maxlength: 24, nullable: true, unique: false, references: 'offers.id'},
+      
+        expires_at: {type: 'dateTime', nullable: true},
+        created_at: {type: 'dateTime', nullable: false},
+        created_by: {type: 'string', maxlength: 24, nullable: false},
```
naz, line 638: @daniellockyer do we keep using the `x_by` fields or do we retire them in favor of using `actions` table? I'm thinking the latter, as there's no good reason to have redundant data like this.

ghost/core/core/server/data/schema/schema.js
```diff
@@ -600,6 +600,45 @@ module.exports = {
...
+            type: 'string', maxlength: 50, nullable: false, validations: {
+                isIn: [['free', 'comped', 'paid']]
+            }
+        },
+        status: {
+            type: 'string', maxlength: 50, nullable: false, validations: {
+                isIn: [['active', 'expired', 'cancelled']]
+            }
+        },
+        member_id: {type: 'string', maxlength: 24, nullable: false, unique: false, references: 'members.id', cascadeDelete: true},
```
naz, line 615: This column is still WIP and is pending discussion, both it's presence and the cascade delete.

ghost/core/core/server/data/schema/schema.js
```diff
@@ -600,6 +600,45 @@ module.exports = {
...
+        tier_id: {type: 'string', maxlength: 24, nullable: true, unique: false, references: 'products.id'},
```
naz, line 616: Used `tier_id` instead of `product_id` here. The thinking behind the decision is that we'd have less work to do (and less execution time) once we rename the `products` table to `tiers`

ghost/core/core/server/data/schema/schema.js
daniellockyer: They can definitely go, nuke 'em 💥

ghost/core/core/server/data/migrations/versions/5.19/2022-10-10-06-58-add-subscriptions-table.js
```diff
@@ -0,0 +1,31 @@
...
+module.exports = addTable('subscriptions', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    type: {
+        type: 'string', maxlength: 50, nullable: false, validations: {
+            isIn: [['free', 'comped', 'paid']]
+        }
+    },
+    status: {
+        type: 'string', maxlength: 50, nullable: false, validations: {
+            isIn: [['active', 'expired', 'cancelled']]
```
allouis, line 12: It's fine to leave but you don't need `validations` in the migration as they're only used by model layer

ghost/core/core/server/data/schema/schema.js
allouis: Fine by me!

allouis (approved):
Looks good

ghost/core/core/server/data/schema/schema.js
daniellockyer: Me too 🙂

ghost/core/core/server/data/migrations/versions/5.19/2022-10-10-06-58-add-subscriptions-table.js
```diff
@@ -0,0 +1,19 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('subscriptions', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    type: {type: 'string', maxlength: 50, nullable: false},
+    status: {type: 'string', maxlength: 50, nullable: false},
+    member_id: {type: 'string', maxlength: 24, nullable: false, unique: false, references: 'members.id', cascadeDelete: true},
+    tier_id: {type: 'string', maxlength: 24, nullable: true, unique: false, references: 'products.id'},
+    cadence: {type: 'string', maxlength: 50, nullable: true},
+    currency: {type: 'string', maxLength: 3, nullable: true},
```
daniellockyer, line 10: This is a bug - should be `maxlength` but it turns out we also make the same mistake elsewhere. I'm also not sure if we should be using `3` because the places where we get it right, we use `50`. The only problem is those using `maxLength` are set to 191

```
./schema.js
447:        currency: {type: 'string', maxlength: 50, nullable: true},
465:        currency: {type: 'string', maxlength: 50, nullable: true},
531:        currency: {type: 'string', maxLength: 3, nullable: false},
580:        currency: {type: 'string', maxLength: 3, nullable: false},
639:        plan_currency: {type: 'string', maxLength: 3, nullable: false}
689:        currency: {type: 'string', maxLength: 3, nullable: false},
```

@allouis What would you prefer to do here in terms of length for this new column? I can write up a tech debt issue for fixing these other ones

allouis: Eeesh! That's a tricky one!

In terms of the new column, I think 50 is what we should go for, 3 was probably a mistake in hindsight

daniellockyer: Ok cool 👍🏻 

My plan to fix `maxLength` is to switch it to `maxlength: 191`, which is effectively what it resolved to, and then we can reduce it down if we want to in the future

PR #15581, opened 2022-10-10
Added members_feedback table

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2041

### DIFF
--- a/ghost/core/core/server/data/exporter/table-lists.js
+++ b/ghost/core/core/server/data/exporter/table-lists.js
@@ -40,7 +40,8 @@ const BACKUP_TABLES = [
     'comment_reports',
     'jobs',
     'redirects',
-    'members_click_events'
+    'members_click_events',
+    'members_feedback'
 ];
 
 // NOTE: exposing only tables which are going to be included in a "default" export file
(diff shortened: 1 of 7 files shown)


lenabaidakova (approved):
🚀

ghost/core/core/server/data/migrations/versions/5.19/2022-10-10-10-04-add-members-feedback-table.js
```diff
@@ -0,0 +1,10 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('members_feedback', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    type: {type: 'string', maxlength: 50, nullable: false, defaultTo: 'positive', validations: {isIn: [['positive', 'negative']]}},
```
allouis, line 5: Just FYI the `validations` config is not used at the DB level but at the model level, so it's not necessary here.

allouis (commented):
Looks good - only thing is, do we want to edit these records, or have a new event for each vote?

SimonBackx:
> Looks good - only thing is, do we want to edit these records, or have a new event for each vote?

I think changing a vote is an edge case (and would only happen right after voting when a user makes a mistake) and it would make things a lot easier if we would not have to take that into account and just edit a vote when they change a vote instead of creating a new one.

daniellockyer (approved):

PR #15585, opened 2022-10-10
Ensured Member Newsletters relation is ordered by sort_order

allouis:
Without this we can get inconsistent ordering in the API which can cause tests to fail


allouis:
Closing because this requires us adding the `sort_order` column to the members_newsletters table

PR #15589, opened 2022-10-11
Added `feedback_enabled` to `newsletters` table

lenabaidakova:
closes TryGhost/Team#2042
- Added ability to enable audience feedback per newsletter (just on BE side).


daniellockyer:
@lenabaidakova Just one minor point: we only use emoji commits for user-facing changes (ie. features released or bugs fixed). Most users probably don't care about a new column on their DB so we tend to leave them off for migrations. You'll need to amend the commit 🙂

lenabaidakova:
> @lenabaidakova Just one minor point: we only use emoji commits for user-facing changes (ie. features released or bugs fixed). Most users probably don't care about a new column on their DB so we tend to leave them off for migrations. You'll need to amend the commit 🙂

Fixed :)

SimonBackx (approved):

PR #15596, opened 2022-10-12
Added importer for custom theme settings

### LINKED ISSUE #15542: Custom theme settings (Site Design settings) not restored from import
### Issue Summary

(Probably not too relevant, but worth mentioning just in case) I upgraded an ancient < v1 by exporting its data into a fresh v1.26.2 (I think), I then exported the data from the v1.26.2 instance and imported it into the latest version v5.17.2. All was well.

Its using the default theme casper (v5.3.1), I fiddled with the admin ui a little including changing some of the things under Site Design such as enabling Show logo in navigation, disabling Show publication cover and changing a few other things. I added a twitter url, added some primary and secondary navigation links. That was about it.

I realised that I would need to make some changes to the theme, so I installed ghost locally with ghost install local. I took an export from the other system and imported it into the local one. Everything seems to have imported with the exception of the changes made under Site Design. The navigation menus are there, the twitter url is there, the posts are there.

I took a look in the export file I can see things like:
"custom_theme_settings": [..., {"id":"633dae33554f811bb2335f53","theme":"casper","key":"show_logo_in_navigation","type":"boolean","value":"true"}, ...] 

I checked a few of the other settings that are missing too and they look correct in the export.

There were a few warnings during import for things that were ignored. My User, Newsletter, a couple of Product and 'Settings: Theme not imported, please upload in Settings - Design'. The last one is a little suspicious, but I think it's just the theme itself not being imported due to it already existing rather than the Site Design settings themselves?

On the off chance doing this from an export was wrong, I use ghost backup to create a backup and imported the data file within it to a clean local instance of ghost instead of the exported one. It has the exact same issue.

To rule out the rest of my content as a problem I setup a fresh local instance, changed the settings under Site Design. Exported the data. Setup another fresh local instance. Imported the data. Looked at the settings under Site Design. They are not set to the ones in the export which I would expect them to be.

Am I doing something wrong? Should I expect these settings to import ok?

Thanks

### Steps to Reproduce

1. Install ghost with ghost cli: ghost install local
2. Open resultant ghost in browser and setup
3. Go to settings -> Design
4. Under Homepage, disable Show publication cover
5. Save changes
6. Go to Settings -> Labs, Export
7. Install another ghost instance (I did ghost stop, removed the entire first instance) using ghost cli: ghost install local
8. Open resultant ghost in browser and setup
9. Go to Settings -> Labs, Import... Select the file exported a moment go
10. Go to Settings -> Design.
11. Under Homepage, note Show publication cover is enabled whereas it should be disabled

### Ghost Version

v5.17.2

### Node.js Version

v16.13.1

ghost cli: ghost install local

### Database type

SQLite3

### Browser & OS version

_No response_

_No response_

### Code of Conduct

illiteratewriter:
closes #15542

### DIFF
--- a/ghost/core/core/server/data/importer/importers/data/custom-theme-settings.js
+++ b/ghost/core/core/server/data/importer/importers/data/custom-theme-settings.js
@@ -3,6 +3,7 @@ const Promise = require('bluebird');
 const debug = require('@tryghost/debug')('importer:roles');
 const BaseImporter = require('./base');
 const models = require('../../../../models');
+const {activate} = require('../../../../services/themes/activate');
 
 class CustomThemeSettingsImporter extends BaseImporter {
     constructor(allDataFromFile) {
@@ -60,8 +61,17 @@ class CustomThemeSettingsImporter extends BaseImporter {
                 })
                 .reflect());
         });
+        
+        const opsPromise = Promise.all(ops);
+        
+        opsPromise.then(() => {
+            models.Settings.findOne({key: 'active_theme'})
+                .then((theme) => {
+                    activate(theme.get('value'));
+                });
+        });
 
-        return Promise.all(ops);
+        return opsPromise;
     }
 }
 module.exports = CustomThemeSettingsImporter;
--- a/ghost/core/core/server/data/importer/importers/data/custom-theme-settings.js
+++ b/ghost/core/core/server/data/importer/importers/data/custom-theme-settings.js
@@ -1,0 +1,67 @@
+const _ = require('lodash');
+const Promise = require('bluebird');
+const debug = require('@tryghost/debug')('importer:roles');
+const BaseImporter = require('./base');
+const models = require('../../../../models');
+
+class CustomThemeSettingsImporter extends BaseImporter {
+    constructor(allDataFromFile) {
+        super(allDataFromFile, {
+            modelName: 'CustomThemeSetting',
+            dataKeyToImport: 'custom_theme_settings'
+        });
+    }
+
+    beforeImport() {
+        debug('beforeImport');
+        return super.beforeImport();
+    }
+
+    doImport(options, importOptions) {
+        debug('doImport', this.modelName, this.dataToImport.length);
+
+        let ops = [];
+
+        _.each(this.dataToImport, (item) => {
+            ops.push(models.CustomThemeSetting.findOne({theme: item.theme, key: item.key}, options)
+                .then((setting) => {
+                    if (_.isObject(item.value)) {
+                        item.value = JSON.stringify(item.value);
+                    }
+                    
+                    if (setting) {
+                        setting.set('value', item.value);
+                        if (setting.hasChanged()) {
+                            return setting.save(null, options)
+                                .then((importedModel) => {
+                                    if (importOptions.returnImportedData) {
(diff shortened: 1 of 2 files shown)


ErisDS:
Hey again @illiteratewriter thanks for sending over this PR :) 

You said [on the issue here](https://github.com/TryGhost/Ghost/issues/15542#issuecomment-1275597916):

> There is one issue; even though the DB is getting updated, the UI doesn't seem to reflect the changes immediately. It is taking some time after importing to reflect the changes. If I restart the server, the changes get reflected.

This is because the custom theme settings are stored in an in-memory cache to keep Ghost performant. The handling of listing and updating settings is wrapped in a [service](https://github.com/TryGhost/Ghost/blob/main/ghost/custom-theme-settings-service/lib/service.js) so that updating models and the cache are managed at the same time.

Here in the PR you're talking to the model directly, which is what all the other importers do, so it makes sense! However, this bypasses the cache and leaves it out of date.

There are two options:

- Update the importer to use the `listSettings`  and `updateSettings` methods [exposed by the service](https://github.com/TryGhost/Ghost/blob/main/ghost/custom-theme-settings-service/lib/service.js#L86) 
- Req

 I don't have a strong sense of which is best but perhaps @kevinansfield who designed the service has an opinion.

ErisDS:
Sorry for the close - I hit an odd keyboard combo halfway through commenting 🙈

illiteratewriter:
Hey @ErisDS, thank you so much for taking the effort to explain the code to me. :)

I've used [activate()](https://github.com/TryGhost/Ghost/blob/0bfbee55236c8c1670713be22d4faf16cf272d89/ghost/core/core/server/services/themes/activate.js#L46) which is a wrapper over `activateTheme()`. Let me know your thoughts on it.

illiteratewriter:
Hey @ErisDS, thank you for walking me through the code. :)

Unfortunately, I've been unable to figure out how to pass the second parameter `theme` (checked theme output from gscan). Can you help me with that? 

Meanwhile I've found an alternate solution, but that involves the code doing a bit more steps. Let me know your thoughts.

ErisDS:
This is definitely a tricky problem - we don't import the "activeTheme" setting via import either for partly this reason (also because we don't know if the theme is present). 

So for clarity the problem is that if you are importing custom theme settings for the theme that is already active, the internal custom theme settings cache doesn't get updated and so the settings don't appear.

I assume this is not the case if you import custom theme settings for a theme that isn't active - if you go activate the theme, the settings will appear.

I the case that you're overwriting settings for the active theme, you're needing to call the higher-level `.activate()` method, because `activateTheme()` requires you to get a full theme object, and that's what `.activate()` does... totally makes sense.

I think the only caveat here then is that you should only need to call `activate()` in the case that custom theme settings were imported for the active theme, otherwise it is unnecessary work.

illiteratewriter:
@ErisDS updated with the change.

ErisDS (approved):

ErisDS:
Hey @illiteratewriter - thanks for persisting with this, it was a tricky one! This change will go out in today's release 🎉  

I know you're planning to pick up the i18n project, but another place to look at contributing would be extending our importer tests to cover the case you added here. The state of those tests isn't great so it's not gonna be the easiest, but it would be good to know we won't regress on this behaviour if we e.g. change how theme activations worked or something.

Issue #15505, opened 2022-09-29
Starting Ghost with missing active theme breaks Design Settings screen


ErisDS:
### Issue Summary

If the selected active theme does not exist in the filesystem the Design Settings screen errors and is unusable.

The problem occurs because Ghost's API is returning an unexpected state - when requesting a list of themes none of them has the active: true flag.

The preferred solution is to cater for this state in Admin somehow

e.g. have a blank-slate state for the design settings screen indicating that no theme is active and point user to the change-theme screen

### Steps to Reproduce

1. Activate a custom theme
2. Stop Ghost and delete or rename the custom theme directory
3. Start Ghost
4. Try to access the design settings screen

![image](https://user-images.githubusercontent.com/101513/193050450-fdd1b30c-f923-49a8-99a6-eedb5c5a96e6.png)

### Ghost Version

5.16

### Node.js Version

16

### How did you install Ghost?

ghost install

### Database type

MySQL 8

### Relevant log / error output

```shell
<img width="630" alt="image" src="https://user-images.githubusercontent.com/101513/193117369-f2947482-993a-440e-91c8-6f5cc11a5c37.png">
```

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

Walexero:
Do i have to have a NodeJS knowledge to work on the issue or having a plain JS knowledge woulod suffice, if it would, i would like to work on the issue.

ErisDS:
Hi @Walexero, this is a bug in our Ember.js admin client. If you understand the code enough to fix the issue, then you have enough knowledge... the only person who knows if you have enough knowledge is you!

Walexero:
Hello @ErisDS, i already pulled ghost and trying to work on it locally, i first starteed the server with `yarn dev` but just gave the home page and didn't provide the theme option, tried signing up locally but doesn't provide options beyond those provided before signing up. Then i stopped the server and ran it with `yarn dev:admin` but the server errored out with deprecation issues. I had ghost set up with the configuration guidelines on https://ghost.org/docs/install/source/[ghost setup guideline] so is there anything wrong with the setup i did?

ErisDS:
Doesn't sound like there was anything wrong with your setup. This is a bug in Ghost's admin panel, which lives at `/ghost/`.

Walexero:
do you have a way to figure this out so i can work on this open issue

ErisDS:
[...]

PR #15602, opened 2022-10-12
🐛 Fix missing active theme breaks design screen

kristophesankar:
refs #15505

When starting Ghost with a missing active theme, the design settings screen and change theme screen both end up in a broken state with the user unable to select a new theme as the active one. 

The design screen has no default (or blank) slate, and so shows a preview of an empty theme.
- First added a new default screen to serve as a placeholder for when the state contains no active theme.
- Added a check for when there was no active theme, then redirects the user to the default screen .

The change theme screen wants to set an active property on the theme that should be active in the theme list.
- Added a check to see whether there is an active theme set. If there isn't one, don't bother trying to add the active property.

### DIFF
--- a/ghost/admin/app/routes/settings/design/index.js
+++ b/ghost/admin/app/routes/settings/design/index.js
@@ -11,7 +11,7 @@ export default class SettingsDesignIndexRoute extends AuthenticatedRoute {
 
     confirmModal = null;
     hasConfirmed = false;
-    themes = this.store.findAll('theme');
+    themes = this.store.peekAll('theme');
 
     afterModel() {
         super.afterModel(...arguments);
--- a/ghost/admin/app/routes/settings/design/index.js
+++ b/ghost/admin/app/routes/settings/design/index.js
@@ -7,9 +7,19 @@ export default class SettingsDesignIndexRoute extends AuthenticatedRoute {
     @service customThemeSettings;
     @service modals;
     @service settings;
+    @service store;
 
     confirmModal = null;
     hasConfirmed = false;
+    themes = this.store.findAll('theme');
+
+    afterModel() {
+        super.afterModel(...arguments);
+        let activeTheme = this.themes.findBy('active', true)
+        if (typeof activeTheme === 'undefined') {
+           return this.transitionTo('settings.design.no-theme');
+        }
+    }
 
     @action
     willTransition(transition) {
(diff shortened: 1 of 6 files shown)


ErisDS:
This looks absolutely spot-on to me 👌 but I've requested a second pair of eyes from @kevinansfield on the ember code 🙂

ghost/admin/app/routes/settings/design/index.js
```diff
@@ -7,9 +7,19 @@ export default class SettingsDesignIndexRoute extends AuthenticatedRoute {
     @service customThemeSettings;
     @service modals;
     @service settings;
+    @service store;
 
     confirmModal = null;
     hasConfirmed = false;
+    themes = this.store.findAll('theme');
```
kevinansfield, line 14: Setting properties to promises like this in singletons (routes, controllers, services) isn't good practice because there's no control over when the class instance will be instantiated and the promise is kicked off.

If we know that themes are loaded in a higher-level route and already in the store (eg, by adding the `findAll` to the list of [promises that the `routes/settings/design.js` model hook waits for](https://github.com/TryGhost/Ghost/blob/main/ghost/admin/app/routes/settings/design.js#L19-L23)) then we can confidently can use a non-promise-returning method like `themes = this.store.peekAll('theme')` here.

kevinansfield (changes requested):

ghost/admin/app/routes/settings/design/index.js
kristophesankar: This makes a lot of sense! I've gone ahead and pushed the changes.

kevinansfield (changes requested):

ghost/admin/app/routes/settings/design/no-theme.js
```diff
@@ -0,0 +1,16 @@
+import AuthenticatedRoute from 'ghost-admin/routes/authenticated';
+import {inject as service} from '@ember/service';
+
+export default class SettingsDesignNoThemeRoute extends AuthenticatedRoute {
+    @service store;
+
+    themes = this.store.findAll('theme');
```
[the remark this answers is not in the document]
kristophesankar: Thanks for catching that. I just amended the commit to include that change.

kevinansfield (approved):

ErisDS:
Hi @kristophesankar thanks so much for this PR. It has now been merged 🎉 and will appear in the next release of Ghost - usually Fridays.

I'm not sure if you found this through hacktoberfest, but I've added the `hacktoberfest-accepted` label to this PR to make sure it counts. We have plenty of open issues and ongoing projects for hacktoberfest and beyond. We would love to see you around again 👋

PR #15604, opened 2022-10-13
Removed `bool` type from schema

daniellockyer:
refs https://github.com/TryGhost/Toolbox/issues/441

- we tend to have a mix of `bool` and `boolean` in the schema and migrations, which has become a real nit for me at this point
- we don't do any special handling between `bool` and `boolean`, it's just something we pass to Knex
- `bool` is an alias for `boolean` but `boolean` is actually documented - https://knexjs.org/guide/schema-builder.html#boolean
- this commit switches Ghost to only using `boolean` in the schema and migrations, and removes `bool` from the allowlist in tests to prevent us from adding it again in the future
- this should make absolutely no difference to the DB because both resulted in the same column

### DIFF
--- a/ghost/core/core/server/data/schema/validator.js
+++ b/ghost/core/core/server/data/schema/validator.js
@@ -63,7 +63,7 @@ function validateSchema(tableName, model, options) {
 
         // validate boolean columns
         if (Object.prototype.hasOwnProperty.call(schema[tableName][columnKey], 'type')
-            && (schema[tableName][columnKey].type === 'bool' || schema[tableName][columnKey].type === 'boolean')) {
+            && schema[tableName][columnKey].type === 'boolean') {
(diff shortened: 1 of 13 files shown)


ghost/core/core/server/data/schema/validator.js
```diff
@@ -63,7 +63,7 @@ function validateSchema(tableName, model, options) {
 
         // validate boolean columns
         if (Object.prototype.hasOwnProperty.call(schema[tableName][columnKey], 'type')
-            && (schema[tableName][columnKey].type === 'bool' || schema[tableName][columnKey].type === 'boolean')) {
```
naz, line 66: love it when smelly ifs like this disappear 🔥

naz (approved):
George Boole approves 👍

PR #15606, opened 2022-10-13
Fixed incorrect maximum length definition in schema

daniellockyer:
refs https://github.com/TryGhost/Toolbox/issues/441

- whilst reviewing another PR, I noticed we were incorrectly using `maxLength` instead of `maxlength` in the schema column definition
- it turns out we've already been doing this wrong for a while with other columns
- this key is not acted upon, so the maximum column length was not applied
- fixing up the DB to the correct maximum length is something to fix in the future but right now, the schema does not reflect the size of the column that actually got created
- the fallback when `maxlength` is not provided is currently 191 [0], so this commit switches the schema and migrations to using the correct key name and column length that they are using when applied

[0]: https://github.com/TryGhost/Ghost/blob/24670aa555e2b39d2505853c6cb19f6c976c2738/ghost/core/core/server/data/schema/commands.js#L27

### DIFF
--- a/ghost/core/test/unit/server/data/schema/schema.test.js
+++ b/ghost/core/test/unit/server/data/schema/schema.test.js
@@ -23,7 +23,6 @@ const VALID_KEYS = {
     ],
     string: [
         'maxlength',
-        'maxLength', // this is the result of a bug and should NOT be used, will be deleted soon
         'nullable',
         'primary',
         'unique',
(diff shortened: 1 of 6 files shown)


ghost/core/test/unit/server/data/schema/schema.test.js
```diff
@@ -27,7 +27,6 @@ const VALID_KEYS = {
     ],
     string: [
         'maxlength',
-        'maxLength', // this is the result of a bug and should NOT be used, will be deleted soon
```
naz, line 30: 😮😮😮 now I see how it sneaked it with the subscriptions table 🙈

daniellockyer: Actually, before yesterday we had no test for this at all 😭 I added `maxLength` so I could merge, and this cleans up the usage

But my god it's easy to miss 😅

naz (approved):
I'd also add a `@NOTE` in here or link to an issue to review scaling the 191 limit back to 50 at some point.

Issue #10360, opened 2019-01-10
URL service decoupling


gargol:
## Known Problems

1. The URL service uses the model layer directly. The decision was made before we invented API versions. The decision was made because raw knex queries are so much faster than using Bookshelf. _We couple the URL Service (frontend) with the model layer._ 

2. The URL service listens on Ghost events and sometimes we have to re-fetch the data from the database because these events are API independent (!) e.g. v2 does only show authors with posts. But the model layer doesn't know that.

3. Because of (2), we run into the problem reported [here](https://github.com/TryGhost/Ghost/issues/10356). We can't do any async operations if the Ghost events are triggered, because otherwise, the API will return 404 urls. We couple the API with the URL service.

4. URL Utils are used throughout codebase which couples it to every module in the system. (solved in https://github.com/TryGhost/Ghost/issues/10773)

## Future Architecture

Imagine you split Ghost into two repositories:

- API (models...)
- Frontend (Handlebars, Dynamic Routing ....all the frontend features....)

You can't because it's too coupled.

### Known Challenges:

- How to listen on API events? Webhooks?
- Use the Content API directly
- Can we continue using [knex_raw](e)? (described [here](https://github.com/TryGhost/Ghost/issues/10124)) or should we look for an alternative approach?

### Also to consider
Some tests were disabled due to delay introduced by additional db call in resources service and would be re-enabled after the new mechanism is in place. Related [convo](https://github.com/TryGhost/Ghost/commit/df1ba8aee1e3a9f2032d3a96b0684b8c1a3cd353#r31932511)

kirrg001:
Updated the issue description a little. More coming soon...

kirrg001:
@ErisDS 🤔 Could you pls leave a reason why you've closed this issue?

ErisDS:
All issues tagged with `routing` have been tagged with `later` and closed until we prioritise working on the Routing feature again.

PR #15613, opened 2022-10-13
Added e2e tests for `page.unpublished` webhook

iamhalldor:
Adding test coverage for webhook events as per #15537


halldorbjarni:
The test fails when run via `yarn run test:all` but `yarn test:single test/e2e-webhooks/pages.test.js` succeeds (at least locally). There seems to be some state pollution in the test harness that produces different urls for the `author url` AFAICT. 

I haven't been able to figure out where it's coming from yet but I'd much prefer fixing the interdependency between tests than allowing for both urls in the snapshot matching.

halldorbjarni:
Any idea why the urls are different @naz? The fixtures are initialized for each test class so not sure why there's a difference between running the test class vs a whole suite.

naz:
Hey @halldorbjarni 👋  The issue with a '404' URL has been haunting Ghost for a while now. For more context you can have a read through this issue - https://github.com/TryGhost/Ghost/issues/10360 (pt. 3 mentions this exact 404 issue). There's a whole family of bugs that are related to how Ghost calculates public resource URLs and this is one of it's variations. We are planning a bigger rewrite of the routing system, until then having a "centralized and documented hack" is our best solution.

There's a documented [local URL matcher](https://github.com/TryGhost/Ghost/commit/320c6e0dd3f9338081f7284d7a800c3c39d412f4) used in [post webhook tests](https://github.com/TryGhost/Ghost/blob/6b7088ba20415719e4e18df79464829125b68ee0/ghost/core/test/e2e-webhooks/posts.test.js#L18) and other places with similar behavior to the on in this PR.  My suggestion would be using the `anyLocalURL` matcher for the author's url attribute matching here. I is a hacky solution, but should unblock progress. All uses of `anyLocalURL` will be substituted with a concrete snapshot url match once we fix the root of all 404 bugs - routing 🙃 

Hope this answer helps and massive thanks for putting your time into this contribution!

halldorbjarni:
Updated! Thanks for the clarification @naz, I definitely spent some time going down the debug rabbithole thinking this was due to a dirty test database... 😅

naz (approved):
🚀 exactly what we discussed! Good job @halldorbjarni 👍

ErisDS:
Hey @halldorbjarni thanks for this PR, it's merged and should be included in today's release.

I'm looking at your bluebird PRs now :)

PR #15619, opened 2022-10-14
Added the feedback buttons in the emails

lenabaidakova:
closes TryGhost/Team#2046
closes TryGhost/Team#2045
- Added feedback buttons markup.
- Added feedback links generation.

### DIFF
--- a/ghost/audience-feedback/lib/AudienceFeedbackService.js
+++ b/ghost/audience-feedback/lib/AudienceFeedbackService.js
@@ -1,19 +1,27 @@
 class AudienceFeedbackService {
+    /** @type URL */
+    #baseURL;
     /**
-     * @param {string} siteUrl
+     * @param {object} deps
+     * @param {object} deps.config
+     * @param {URL} deps.config.baseURL
+     */
+    constructor(deps) {
+        this.#baseURL = deps.config.baseURL;
+    }
+    /**
      * @param {string} uuid
      * @param {string} postId
      * @param {0 | 1} score
      */
-    buildLink(siteUrl, uuid, postId, score) {
-        const params = new URLSearchParams({
-            action: 'feedback',
-            post: postId,
-            uuid,
-            score
-        });
+    buildLink(uuid, postId, score) {
+        const url = new URL(this.#baseURL);
+        url.searchParams.set('action', 'feedback');
+        url.searchParams.set('post', postId);
+        url.searchParams.set('uuid', uuid);
+        url.searchParams.set('score', `${score}`);
 
-        return new URL(`${siteUrl}?${params.toString()}`);
+        return url;
     }
 }
 
--- a/ghost/audience-feedback/lib/AudienceFeedbackService.js
+++ b/ghost/audience-feedback/lib/AudienceFeedbackService.js
@@ -1,7 +1,19 @@
 class AudienceFeedbackService {
-    buildLink() {
-        // todo
-        return new URL('https://example.com');
+    /**
+     * @param {string} siteUrl
+     * @param {string} uuid
+     * @param {string} postId
+     * @param {0 | 1} score
+     */
+    buildLink(siteUrl, uuid, postId, score) {
+        const params = new URLSearchParams({
+            action: 'feedback',
+            post: postId,
+            uuid,
+            score
+        });
+
+        return new URL(`${siteUrl}?${params.toString()}`);
     }
 }
 
--- a/ghost/core/core/server/services/bulk-email/bulk-email-processor.js
+++ b/ghost/core/core/server/services/bulk-email/bulk-email-processor.js
@@ -234,6 +234,9 @@ module.exports = {
                 unsubscribe_url: postEmailSerializer.createUnsubscribeUrl(recipient.member_uuid, {newsletterUuid})
             };
 
+            // create unique urls for every recipient (for example, for feedback buttons)
+            emailData = postEmailSerializer.createUserLinks(emailData, recipient.member_uuid);
+
             // computed properties on recipients - TODO: better way of handling these
             recipient.member_first_name = (recipient.member_name || '').split(' ')[0];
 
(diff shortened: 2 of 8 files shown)


ghost/audience-feedback/lib/AudienceFeedbackService.js
```diff
@@ -1,7 +1,21 @@
+const urlUtils = require('../../core/core/shared/url-utils');
```
allouis, line 1: Would be cool to pass in the necessary data as dependencies, similar to how we do here with `baseUrl` https://github.com/TryGhost/Ghost/blob/main/ghost/link-redirects/lib/LinkRedirectsService.js

ghost/audience-feedback/lib/AudienceFeedbackService.js
```diff
@@ -1,7 +1,21 @@
...
-    buildLink() {
-        // todo
-        return new URL('https://example.com');
+    /**
+     * @param {string} uuid
+     * @param {string} postId
+     * @param {0 | 1} score
+     */
+    buildLink(uuid, postId, score) {
+        const siteUrl = urlUtils.getSiteUrl();
```
allouis, line 10: RE: https://ghost.slack.com/archives/C02G9E68C/p1665739008939039 we should use `urlFor` rather than `getSiteUrl`

ghost/audience-feedback/lib/AudienceFeedbackService.js
```diff
@@ -1,7 +1,21 @@
...
+        const params = new URLSearchParams({
+            action: 'feedback',
+            post: postId,
+            uuid,
+            score
+        });
+
+        return new URL(`${siteUrl}?${params.toString()}`);
```
allouis, line 18: This is a little convoluted, I think it would read better as something like:
```js
const url = new URL(siteUrl);
url.searchParams.set('action', 'feedback');
// etc...

return url;
```

ghost/core/core/server/services/mega/feedback-buttons.js
```diff
@@ -0,0 +1,69 @@
+const {Color} = require('@tryghost/color-utils');
+
+const templateStrings = {
+    like: '%{feedback_button_like}%',
+    dislike: '%{feedback_button_dislike}%'
+};
+
+const generateLinks = (postId, uuid, html) => {
+    const audienceFeedback = require('../audience-feedback');
```
allouis, line 9: Why is this `require`d here as opposed to the top of the file?

ghost/core/core/server/services/mega/post-email-serializer.js
```diff
@@ -107,6 +108,23 @@ const PostEmailSerializer = {
...
+     * Generate personalised links for each user
+     *
+     * @param {string} memberUuid member uuid
+     * @param {Object} email
+     */
+    createUserLinks(email, memberUuid) {
+        const result = {...email};
+
+        result.html = feedbackButtons.generateLinks(result.post.id, memberUuid, result.html);
+        result.plaintext = htmlToPlaintext.email(result.html);
```
allouis, line 123: Is it necessary that we do this here?

allouis (commented):
Coupla comments but overall looks good

ghost/audience-feedback/lib/AudienceFeedbackService.js
lenabaidakova: Fixed :)

ghost/audience-feedback/lib/AudienceFeedbackService.js
lenabaidakova: Fixed :)

ghost/audience-feedback/lib/AudienceFeedbackService.js
lenabaidakova: Fixed :)

ghost/core/core/server/services/mega/feedback-buttons.js
lenabaidakova: I think, it's just a result of my experiments. All services started in `boot` file, moved it to the top :)

ghost/core/core/server/services/mega/post-email-serializer.js
lenabaidakova: I'd suggest keeping it :)  We added new links to Html and I think it's better to update the plain version

ghost/core/core/server/services/mega/template.js
```diff
@@ -70,6 +71,10 @@ table {
     width: 100%;
 }
 
+table.inline {
+  width: auto !important;
+}
+
```
rishabhgrg, line 77: @lenabaidakova Quick check, this CSS doesn't has any side-effects on existing template if the alpha flag is off right ?

ghost/core/core/server/services/bulk-email/bulk-email-processor.js
```diff
@@ -234,6 +234,9 @@ module.exports = {
                 unsubscribe_url: postEmailSerializer.createUnsubscribeUrl(recipient.member_uuid, {newsletterUuid})
             };
 
+            // create unique urls for every recipient (for example, for feedback buttons)
+            emailData = postEmailSerializer.createUserLinks(emailData, recipient.member_uuid);
```
rishabhgrg, line 238: @lenabaidakova Guess this will not leak anything out of flag due to checks later, but maybe this should be wrapped behind the feedback labs flag so we don't change any email data unless flag is enabled

ghost/core/core/server/services/mega/template.js
lenabaidakova: It shouldn't but you are right, better to isolate them =)
Moved to feedback buttons https://github.com/TryGhost/Ghost/pull/15619/commits/075e753af78efec452c5ce012c39b47f41ed21b7

ghost/core/core/server/services/bulk-email/bulk-email-processor.js
lenabaidakova: Done :)
https://github.com/TryGhost/Ghost/pull/15619/commits/204df614de6aa9a6b1cb08b26fafb547bc0b56b0

rshbhgrg (approved):

moreofmorris (approved):
Looking good and thanks for fixing the spacing issue in the emails 😊

PR #15626, opened 2022-10-15
Removed bluebird from `fixture-utils.js`

iamhalldor:
#14882 

I opted to use the in-house `sequence` function when refactoring Bluebird's `Promise.each` to avoid deadlock issues (see 734ef66). I'm not familiar enough with the database schema to know which `Promise.each` are safe to refactor to `Promise.all`.

### DIFF
--- a/ghost/core/test/utils/fixture-utils.js
+++ b/ghost/core/test/utils/fixture-utils.js
@@ -1,5 +1,4 @@
 // Utility Packages
-const Promise = require('bluebird');
 const _ = require('lodash');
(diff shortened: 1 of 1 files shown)


ErisDS:
Thanks again @halldorbjarni :) This file had had another Promise.map added since you fixed it, which I've converted and merged in along with your changes. That shows just how quickly the tech debt is growing whilst we still have bluebird in the codebase!

We really, really appreciate the help to get bluebird removed ❤️

PR #15637, opened 2022-10-17
Added conversions count and separate analytics page

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2084

- When audience feedback is enabled, we use a single 'conversions' count instead of having separate ones for signups and paid conversions.
- The analytics component is separated so we can change it without breaking the existing page.


lenabaidakova (commented):
🚀

lenabaidakova (approved):

PR #15638, opened 2022-10-17
🐛 Fixed large mailgun replacements

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2096

When generating the recipient data for emails, the email clicks implementation is resulting in a recipient variable being added called replacement_xxx once for each link containing the same UUID.

This generates a lot of unnecessary data overhead for emails, and it turns out that mailgun has a 25MB message limit. We wouldn't have come close if we only included the uuid once.


allouis (approved):
Real shame about having to use a regex

ErisDS (approved):
Tested and it is now working how I'd expect 🙌

PR #15640, opened 2022-10-17
Improved URL syncing between Admin and Explore

aileen:
no issue

Improve the route communication between Ghost Admin and Ghost Explore to reflect route changes in the URL and correctly navigate to Explore sub routes

### DIFF
--- a/ghost/admin/app/controllers/explore.js
+++ b/ghost/admin/app/controllers/explore.js
@@ -20,6 +20,7 @@ export default class ExploreController extends Controller {
     @action
     closeConnect() {
         if (this.explore.isIframeTransition) {
+            this.explore.sendRouteUpdate({path: '/explore'});
             this.router.transitionTo('/explore');
         } else {
             this.router.transitionTo('/dashboard');
@@ -43,6 +44,7 @@ export default class ExploreController extends Controller {
             // to the submit page and fetch the required site data
             setTimeout(() => {
                 this.explore.toggleExploreWindow(true);
+                this.explore.sendRouteUpdate({path: '/explore'});
                 this.router.transitionTo('explore');
             }, 500);
         } else {
--- a/ghost/admin/app/controllers/explore.js
+++ b/ghost/admin/app/controllers/explore.js
@@ -44,8 +44,6 @@ export default class ExploreController extends Controller {
             // to the submit page and fetch the required site data
             setTimeout(() => {
                 this.explore.toggleExploreWindow(true);
-                this.explore.sendRouteUpdate({path: '/explore'});
-                this.router.transitionTo('explore');
             }, 500);
         } else {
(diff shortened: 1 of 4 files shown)


ghost/admin/app/routes/explore/index.js
```diff
@@ -43,10 +43,22 @@ export default class ExploreRoute extends AuthenticatedRoute {
...
+                if (destinationUrl?.includes('/explore/submit')) {
+                    // Ensure the model is loaded before sending the controller action
+                    const model = await this.model;
+                    this.controllerFor('explore').submitExploreSite(model);
+                } else {
+                    let path = destinationUrl.replace(/explore\//, '');
+                    path = path === '/' ? '/explore/' : path;
+
+                    if (destinationUrl?.includes('/explore/about')) {
+                        window.open(`${this.explore.exploreUrl}about/`, '_blank').focus();
```
aileen, line 57: This should simulate the same behaviour as when clicking on the `about` link which opens a new tab.

ghost/admin/app/routes/explore/index.js
```diff
@@ -43,10 +43,22 @@ export default class ExploreRoute extends AuthenticatedRoute {
...
             if (destinationUrl?.includes('/explore')) {
                 isExploreTransition = true;
+                this.explore.isIframeTransition = isExploreTransition;
 
-                // Send the updated route to the iframe
-                if (transition?.to?.params?.sub) {
-                    this.explore.sendRouteUpdate({path: transition.to.params.sub});
+                if (destinationUrl?.includes('/explore/submit')) {
+                    // Ensure the model is loaded before sending the controller action
+                    const model = await this.model;
```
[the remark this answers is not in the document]
aileen: I hope and think I found a better way of handling this now... 😬

ghost/admin/app/controllers/explore.js
```diff
@@ -1,13 +1,23 @@
...
+import {isEmpty} from '@ember/utils';
 import {inject as service} from '@ember/service';
+import {task} from 'ember-concurrency';
 
 export default class ExploreController extends Controller {
     @service explore;
     @service router;
+    @service store;
+
+    integrations = null;
```
kevinansfield, line 12: This can be set directly to `this.store.peekAll('integration')`, makes it easier to reason about because you don't then have to find where it's been set.

ghost/admin/app/controllers/explore.js
```diff
@@ -1,13 +1,23 @@
...
+
+    constructor() {
+        super(...arguments);
+        this.initIntegrations.perform();
```
kevinansfield, line 16: For controllers it's usually best to avoid data loading in a constructor because we're not in control of when a controller is initialised. If you want models to be loaded it would be better to keep that in the route

PR #15641, opened 2022-10-17
Fixed PromiseObject related errors thrown from members activity list

kevinansfield:
no issue

- the `memberRecord` getter on the members activity controller was returning the direct result of `store.findRecord()` which in reality is a `PromiseObject` that is a type of proxy. Although templates handle this OK, any JS code using the object needs to know that it's not a typical type of object and has to use `.get()` for all property access to avoid errors
- switched `memberRecord` from a getter to a resource so we have more control over the returned object and loading lifecycle
  - added a `MemberFetcher` resource class that awaits the result of the store find and only then sets the value meaning we always have a fully resolved model object
  - being a resource the fetch will only occur when the `member` id property changes, unlike the getter which performed the fetch every time it was accessed


SimonBackx (approved):
Fixes the issue 👌 Not sure why the tests fail in CI, they succeed locally

PR #15642, opened 2022-10-17
Added e2e tests for `post.published.edited` webhook

kritikash18:
Added e2e tests for post.published.edited webhook

Refs https://github.com/TryGhost/Ghost/issues/15537


naz:
Hey-hey @kritikash18 👋  Could this possibly be a problem with matching the "author" propery like the discussed [here](https://github.com/TryGhost/Ghost/pull/15613#issuecomment-1280213239)? 🤔

halldorbjarni:
Hey @kritikash18, I was curious if this was the same issue that I ran into. It turns out the routing issue is present here but it's actually the reason your single test suite run works. 

`yarn test:single test/e2e-webhooks/posts.test.js` generates a `post['current']['url'] = 'http://127.0.0.1:2369/404/'` due to the routing bug mentioned above and sends it to the webhook. Luckily this still matches the regex for `anyLocalURL` perfectly. 

However, `yarn test:all` generates the correct `post['current']['url'] = http://127.0.0.1:2369/testing-post-published-edited-webhook/'` **but** here the regex `anyLocalURL` doesn't match since it doesn't support dashes. To add it you can edit the `anyLocalUrl` matcher at `Ghost/ghost/core/test/utils/e2e-framework.js` to be something like `stringMatching(/http:\/\/127.0.0.1:2369\/\[w-]+\//),`. Other url allowed symbols could be added as well for completeness.

What I'm mostly curious about is why this matching error is not singled out in the output and instead the jest diff points at **all** the matchers as a potential problem. Any idea here @naz? I'm not sure if it's a jest bug or a bug in how Ghost directly invokes some internal jest packages.

naz:
Oh hey @halldorbjarni 👋  Thanks for joining the discussion! 

>matching error is not singled out in the output

I've came across this issue few times now and I think it's worth the investigation and a separate issue in the test framework. It's super hard to figure out which parts did not match with the current output 💯

halldorbjarni:
> I've came across this issue few times now and I think it's worth the investigation and a separate issue in the test framework. 

@naz I agree. It reproduces on a minimal example.  I've opened a [bug](https://github.com/facebook/jest/issues/13480) on jest. Let's get this fixed!

kritikash18:
Updated the local URL to support - and it worked. Thanks for the pointers @naz and @halldorbjarni !

ghost/core/test/utils/e2e-framework.js
```diff
@@ -384,7 +384,7 @@ module.exports = {
         // @NOTE: hack here! it's due to https://github.com/TryGhost/Toolbox/issues/341
         //        this matcher should be removed once the issue is solved - routing is redesigned
         //        An ideal solution would be removal of this matcher altogether.
-        anyLocalURL: stringMatching(/http:\/\/127.0.0.1:2369\/\w+\//),
+        anyLocalURL: stringMatching(/http:\/\/127.0.0.1:2369\/[A-Za-z0-9_-]+\//),
```
naz, line 387: ohhh! nice catch here 🎣

naz (approved):
Good to land in the main 👍  Thanks for great work @kritikash18!

ErisDS:
@kritikash18 merged and labelled 🎉  Thanks so much for your continued contributions, please keep them coming 🙏

PR #15643, opened 2022-10-18
Added support for supplying the foreign key constraint name

daniellockyer:
- this allows us to choose the foreign key constraint name when the auto-generated one would be too long

### DIFF
--- a/ghost/core/core/server/data/schema/commands.js
+++ b/ghost/core/core/server/data/schema/commands.js
@@ -48,6 +48,10 @@ function addTableColumn(tableName, table, columnName, columnSpec = schema[tableN
         // check if table exists?
         column.references(columnSpec.references);
     }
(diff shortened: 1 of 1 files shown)


ghost/core/core/server/data/schema/commands.js
```diff
@@ -215,11 +219,12 @@ async function hasForeignSQLite({fromTable, fromColumn, toTable, toColumn, trans
  * @param {string} configuration.fromColumn - column of the table to add the foreign key to
  * @param {string} configuration.toTable - name of the table to point the foreign key to
  * @param {string} configuration.toColumn - column of the table to point the foreign key to
+ * @param {string} configuration.constraintName - name of the FK to create
```
daniellockyer, line 222: This should be optional

PR #15644, opened 2022-10-18
 Added `ghost_subscription_id` column to `members_stripe_customers_subscriptions`

daniellockyer:
refs https://github.com/TryGhost/Team/issues/2034

- this table will be used to link Stripe subscriptions to Ghost
  subscriptions via a foreign key that we add at a later point


ghost/core/core/server/data/migrations/versions/5.20/2022-10-18-10-13-add-ghost-subscription-id-column-to-members-scm.js
```diff
@@ -0,0 +1,10 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('members_stripe_customers_subscriptions', 'ghost_subscription_id', {
+    type: 'string',
+    maxlength: 24,
+    nullable: true,
+    references: 'subscriptions.id',
+    constraintName: 'mscs_ghost_subscription_id_foreign',
```
daniellockyer, line 8: Note: the original one was `members_stripe_customers_subscriptions_ghost_subscription_id_foreign` so I just abbreviated the table name

allouis (approved):

PR #15649, opened 2022-10-18
Added bulk endpoint for fixing newsletter links 

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2104

- adds new bulk edit endpoint on links that allows updating all matching links for a post and update them to a new link
  - the bulk edit for links also manages the attribution params on the new link, by appending the newsletter `ref` and post attribution if needed.
- the bulk edit endpoint also updates the `updated_at` value for the link, which will allow highlighting the `Edited` flag on UI for a link if `created_at` doesn't match `updated_at`.
- this also handles reset for link click count if edited, as we update the count method to only fetch click events that are greater than the last updated for the desired link

### DIFF
--- a/ghost/core/core/server/api/endpoints/links.js
+++ b/ghost/core/core/server/api/endpoints/links.js
@@ -21,5 +21,35 @@ module.exports = {
                 }
             };
         }
+    },
+    bulkEdit: {
+        statusCode: 200,
+        headers: {},
+        options: [
+            'filter'
+        ],
+        data: [
+            'action',
+            'meta'
+        ],
+        validation: {
+            data: {
+                action: {
+                    required: true,
+                    values: ['updateLink']
+                }
+            },
+            options: {
+                filter: {
+                    required: true
+                }
+            }
+        },
+        permissions: {
+            method: 'edit'
+        },
+        async query(frame) {
+            return await linkTrackingService.service.bulkEdit(frame.data.bulk, frame.options);
+        }
     }
 };
--- a/ghost/core/core/server/services/link-tracking/PostLinkRepository.js
+++ b/ghost/core/core/server/services/link-tracking/PostLinkRepository.js
@@ -1,4 +1,5 @@
 const {FullPostLink} = require('@tryghost/link-tracking');
+const _ = require('lodash');
 
 /**
  * @typedef {import('bson-objectid').default} ObjectID
@@ -22,8 +23,8 @@ module.exports = class PostLinkRepository {
     }
 
     /**
-     * 
-     * @param {*} options 
+     *
+     * @param {*} options
      * @returns {Promise<InstanceType<FullPostLink>[]>}
      */
     async getAll(options) {
@@ -46,6 +47,29 @@ module.exports = class PostLinkRepository {
         }
 
         return result;
+    }
+
+    async updateLinks(linkIds, updateData, options) {
+        const bulkUpdateOptions = _.pick(options, ['transacting']);
+
+        const bulkActionResult = await this.#LinkRedirect.bulkEdit(linkIds, 'redirects', {
+            ...bulkUpdateOptions,
+            data: updateData
+        });
+
+        return {
+            bulk: {
+                action: 'updateLink',
+                meta: {
+                    stats: {
+                        successful: bulkActionResult.successful,
+                        unsuccessful: bulkActionResult.unsuccessful
+                    },
+                    errors: bulkActionResult.errors,
+                    unsuccessfulData: bulkActionResult.unsuccessfulData
+                }
+            }
+        };
     }
 
     /**
--- a/ghost/link-tracking/lib/LinkClickTrackingService.js
+++ b/ghost/link-tracking/lib/LinkClickTrackingService.js
@@ -2,6 +2,10 @@ const {RedirectEvent} = require('@tryghost/link-redirects');
 const LinkClick = require('./LinkClick');
 const PostLink = require('./PostLink');
 const ObjectID = require('bson-objectid').default;
+const errors = require('@tryghost/errors');
+const nql = require('@tryghost/nql');
+const _ = require('lodash');
+const tpl = require('@tryghost/tpl');
 
 /**
  * @typedef {object} ILinkClickRepository
@@ -25,13 +29,21 @@ const ObjectID = require('bson-objectid').default;
  * @prop {(to: URL, slug: string) => Promise<ILinkRedirect>} addRedirect
  * @prop {() => Promise<string>} getSlug
  * @prop {({filter: string}) => Promise<ILinkRedirect[]>} getAll
+ * @prop {({filter: string}) => Promise<string[]>} getFilteredIds
  */
 
 /**
  * @typedef {object} IPostLinkRepository
  * @prop {(postLink: PostLink) => Promise<void>} save
  * @prop {({filter: string}) => Promise<FullPostLink[]>} getAll
+ * @prop {(linkIds: array, data, options) => Promise<FullPostLink[]>} updateLinks
  */
+
+const messages = {
+    invalidFilter: 'Invalid filter value received',
+    unsupportedBulkAction: 'Unsupported bulk action',
+    invalidRedirectUrl: 'Invalid redirect URL value'
+};
 
 class LinkClickTrackingService {
     #initialised = false;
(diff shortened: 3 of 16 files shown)


ghost/core/core/server/services/link-tracking/PostLinkRepository.js
```diff
@@ -48,6 +62,116 @@ module.exports = class PostLinkRepository {
...
+            };
+        } catch (e) {
+            throw new errors.BadRequestError({
+                message: tpl(messages.invalidFilter),
+                context: e.message
+            });
+        }
+    }
+
+    #getRedirectLinkWithAttribution({newLink, oldLink, postId}) {
```
allouis, line 97: I don't think this is the correct place to put this logic, this class should only be concerned with the persistence and retrieval of data from the DB - but the params here are very much application/feature code - and should probably live in the `@tryghost/link-tracking` package

ghost/core/core/server/services/link-tracking/PostLinkRepository.js
```diff
@@ -48,6 +62,116 @@ module.exports = class PostLinkRepository {
...
+        // append post attribution to site urls
+        const isSite = this.#urlUtils.isSiteUrl(newUrl);
+        if (isSite) {
+            newUrl.searchParams.set('attribution_type', 'post');
+            newUrl.searchParams.set('attribution_id', postId);
+        }
+        return newUrl;
+    }
+
+    async updateLinks(data, options) {
```
allouis, line 114: Same as previous comment, I think there's too much application logic in this class, and it should be pulled out to the tracking/redirect package.

This class can probably just be updated to something like:
```
/**
 * @param {PostLink[]} postLinks
 * @returns {Promise<smth>}
 */ 
async saveMany(postLinks) {
  // impl
}
```

With the fetching and mapping of values happening outside of the repository

allouis (commented):
Couple comments RE moving feature/application logic out into the package

ghost/core/core/server/api/endpoints/links.js
```diff
@@ -21,5 +21,33 @@ module.exports = {
...
+                    values: ['updateLink']
+                }
+            },
+            options: {
+                filter: {
+                    required: true
+                }
+            }
+        },
+        permissions: false,
```
ErisDS, line 48: I don't think we should be disabling perms here?

ErisDS (changes requested):

ghost/core/core/server/api/endpoints/links.js
rishabhgrg: D'oh! No we shouldn't, had copied it over from `browse` and missed! I am working on separate PR to add permissions for links endpoint and will then update here as well.

ghost/core/core/server/services/link-tracking/PostLinkRepository.js
rishabhgrg: @allouis Updated to move logic out to `LinkTrackingService`, and the `PostLinkRepository` now gets the link ids to update along with the new data so it doesn't has any logic.

ghost/core/core/server/services/link-tracking/PostLinkRepository.js
rishabhgrg: Updated to move logic out to `LinkTrackingService` and kept this to just update given linkIds with new data.

ghost/core/core/server/api/endpoints/links.js
rishabhgrg: Updated permissions and fixed for `browse` endpoint as well with a migration on `main`.

ghost/link-tracking/lib/LinkClickTrackingService.js
```diff
@@ -78,7 +97,102 @@ class LinkClickTrackingService {
...
+
+        // get new redirect link with proper attribution
+        const newRedirectUrl = this.#getRedirectLinkWithAttribution({
+            newLink: data.meta?.link?.to,
+            oldLink: redirectUrl.href,
+            postId
+        });
+
+        // fetch link rows with old redirect url to update
+        const linkRows = await this.#LinkRedirect.getFilteredCollectionQuery(updatedFilterOptions)
```
allouis, line 172: I don't think we should be using a bookshelf model here directly, instead the link redirect service/repository (probably service) would be better ☺️

ghost/core/core/server/services/link-tracking/PostLinkRepository.js
```diff
@@ -48,6 +49,29 @@ module.exports = class PostLinkRepository {
         return result;
     }
 
+    async updateLinks(linkIds, updateData, options) {
+        const bulkUpdateOptions = _.pick(options, ['transacting']);
+
+        const bulkActionResult = await this.#LinkRedirect.bulkEdit(linkIds, 'redirects', {
```
allouis, line 55: Our bulkX methods will allow some operations to fail, are we okay with a partial edit from a product POV?

ghost/link-tracking/lib/LinkClickTrackingService.js
```diff
@@ -78,7 +97,102 @@ class LinkClickTrackingService {
...
+   */
+    #parseLinkFilter(filter) {
+        // decode filter to manage any encoded uri components
+        filter = decodeURIComponent(filter);
+
+        try {
+            const filterJson = nql(filter).parse();
+            const postId = filterJson?.$and?.[0]?.post_id;
+            const redirectUrl = new URL(filterJson?.$and?.[1]?.to);
+            if (!postId || !redirectUrl) {
```
allouis, line 115: Should we be using an NQL filter if we're going to be restricting it to such a particular format? Would it be simpler to pass a `postId` and and a `redirectUrl` in the body/query? 🤔

ghost/link-tracking/lib/LinkClickTrackingService.js
```diff
@@ -78,7 +97,102 @@ class LinkClickTrackingService {
...
+    /**
+   * validate and manage the new redirect url in filter
+   * `to` url needs decoding and transformation to relative url for comparision
+   * @param {string} filter
+   * @returns {Object} parsed filter
+   * @throws {errors.BadRequestError}
+   */
+    #parseLinkFilter(filter) {
+        // decode filter to manage any encoded uri components
+        filter = decodeURIComponent(filter);
```
allouis, line 109: If we do stick with using NQL filter - IMO the decoding of URI component should happen way closer to the HTTP level, in the endpoint/controller. Not a big deal but this service shouldn't need to be concerned with the encoding of data in transit

allouis (commented):
Couple more comments! Looks nicer having this pulled out into service now though ☺️

ghost/link-tracking/lib/LinkClickTrackingService.js
rishabhgrg: The reason i'd stick to keeping the NQL filter is because that uses our standard bulk endpoint  pattern, while passing stuff in body will make it really custom. We currently we have limited examples - links and members - for this pattern, but i am planning to update the docs with it this week.

allouis: Do we have similar restrictions on the filters for the members bulk endpoints?

ghost/core/core/server/services/link-tracking/PostLinkRepository.js
rishabhgrg: Think we are ok with that atm, as ideally unlike bulk members the links don't have any dependency/checks before the bulk operation so they shouldn't fail, and if they do we can flag up on UI I guess as we have access to all links, but I'll double check on this

ghost/link-tracking/lib/LinkClickTrackingService.js
rishabhgrg: Don't think so, but the usecase here is different as in case of Members the bulk edits do add/remove label based on the filter which can be anything and isn't dependent on label being added/removed. While here if the bulk action is `updateLink` its necessary to have the old link and post id in place as you shouldn't be able to bulk update without it.

ghost/link-tracking/lib/LinkClickTrackingService.js
rishabhgrg: 🤔 I am ok to move this up to endpoint controller, I had originally thought about it as well but it felt inconsistent tbh to have bits split between controller and service as for other places, tho i didn't feel too strongly either ways. Gonna try moving this up in this case to controller, is it good if that happens in a separate change ?

ghost/link-tracking/lib/LinkClickTrackingService.js
rishabhgrg: Updated ✅

ErisDS (approved):

PR #15650, opened 2022-10-18
Added post_id filter and total to activity feed API

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2091
fixes https://github.com/TryGhost/Team/issues/2089

- Added new fixtures to make testing easier for the activity feed
- Improved E2E test coverage of activity feed with separate test file
- Added data.post_id filter to enable filtering by events related to a given post
- Fixed return types in JSDoc of test agents (TypeScript interprets these as `typeof Agent` if we don't add `InstanceType<Agent>`)
- Added total pagination metadata to activity feed API (to allow a basic type of pagination using filters)

### DIFF
--- a/ghost/core/core/server/api/endpoints/members.js
+++ b/ghost/core/core/server/api/endpoints/members.js
@@ -435,10 +435,7 @@ module.exports = {
             method: 'browse'
         },
         async query(frame) {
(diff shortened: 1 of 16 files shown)


lenabaidakova (approved):
🚀

PR #15652, opened 2022-10-18
Added tabs component

lenabaidakova:
closes TryGhost/Team#2086

### DIFF
--- a/ghost/admin/app/components/tabs/tabs.hbs
+++ b/ghost/admin/app/components/tabs/tabs.hbs
@@ -1,0 +1,30 @@
+<div ...attributes>
+    <div
+        role="tablist"
+        class="tab-list"
+    >
+        {{yield
+            (hash
+                tab=(
+                    component "tabs/tab"
+                    selectedIndex=this.selectedIndex
+                    onSelect=this.handleSelect
+                    onKeyup=this.handleKeyup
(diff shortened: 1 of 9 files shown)


ghost/admin/app/components/tabs/tabs.hbs
```diff
@@ -0,0 +1,30 @@
+<div ...attributes>
```
lenabaidakova, line 1: Wasn't added styles for this component as we don't have a predefined set of possible variants. It gives opportunity to style it for each usage

SimonBackx (approved):

moreofmorris:
@lenabaidakova Looking great! Just a few tiny things here. "conversions" should have a capital "C" and we should swap around "More like this" and "Conversions" so it reads `Sent → Opened → Clicked → More like this → Conversions`.

CleanShot 2022-10-19 at 10 40 53@2x

lenabaidakova:
> @lenabaidakova Looking great! Just a few tiny things here. "conversions" should have a capital "C" and we should swap around "More like this" and "Conversions" so it reads `Sent → Opened → Clicked → More like this → Conversions`.
> 
> CleanShot 2022-10-19 at 10 40 53@2x

Fixed :)

moreofmorris (approved):

PR #15657, opened 2022-10-19
Added sentiment ordering and include for posts

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2090

- This changes how sentiment is exposed in the API. Now it is exposed as a `sentiment` relation, directly on the model (no longer in counts). Internally we still use `count.sentiment`.
- Content API users (and themes) can include the 'sentiment' relation and order by sentiment.
- Updated Admin to use sentiment instead of count.sentiment
- Test theme: https://github.com/tryghost/casper/tree/audience-feedback


lenabaidakova (approved):
⭐

PR #15660, opened 2022-10-19
Fixed default feedback enabled when flag is disabled

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2114
fixes https://github.com/TryGhost/Team/issues/2115

When a new newsletter is created, the frontend will send feedback_enabled to true. We'll catch this in the backend and don't allow setting feedback_enabled to true when audience_feedback flag is disabled. This is also handled for editing newsletters.

To fix this in existing sites, I added a migration that disables feedback for all sites (since this is an alpha feature). Once we'll release the feature later, it will be disabled for existing newsletters, just like expected.


lenabaidakova (approved):
Thank you for fixing 🙌

daniellockyer (changes requested):
Need to change the minor version path but LGTM otherwise

daniellockyer (approved):

PR #15664, opened 2022-10-20
Added link bulk edit endpoint permissions

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2104

- adds permissions for `bulkEdit` endpoint on links
- new endpoint allows fixing newsletter links via Admin

### DIFF
--- a/ghost/core/core/server/data/schema/fixtures/fixtures.json
+++ b/ghost/core/core/server/data/schema/fixtures/fixtures.json
@@ -625,8 +625,8 @@
                     "object_type": "link"
                 },
                 {
-                    "name": "Bulk edit links",
-                    "action_type": "bulkEdit",
+                    "name": "Edit links",
+                    "action_type": "edit",
                     "object_type": "link"
                 }
             ]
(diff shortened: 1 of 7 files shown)


ghost/core/core/server/data/schema/fixtures/fixtures.json
```diff
@@ -623,6 +623,11 @@
                     "name": "Browse links",
                     "action_type": "browse",
                     "object_type": "link"
+                },
+                {
+                    "name": "Bulk edit links",
+                    "action_type": "bulkEdit",
```
allouis, line 629: Should this just be an `edit` permission?

allouis (commented):
Looks good, one question but NABD

ghost/core/core/server/data/schema/fixtures/fixtures.json
rishabhgrg: 🤔 Yea i was thinking of action type used in endpoint, but maybe makes more sense to update this to `edit` and update permissions for endpoint to use that.

rishabhgrg: Updated to `edit`

daniellockyer (approved):
Commit probably needs updating to remove `bulk` but LGTM otherwise

allouis (approved):

PR #15667, opened 2022-10-20
Added new members filters and refactored filters

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2112

- Removed a bit of duplicate code across templates and components that was used to handle filters
- Updated filter objects to contain information about the filter
- Added resource filters that are able to select a single resource, which can be used in columns
- Filters can now define columns by themselves. Not all columns already make use of this functionality, but we can move those over later (cleanup: https://github.com/TryGhost/Team/issues/2133)
- The filter definitions became quite long. We should move them to separate files in the future: https://github.com/TryGhost/Team/issues/2134
- Filters can now have custom NQL parsing
- Improved support for parsing recursive or grouped NQL queries
- Added support for filtering members by feedback


ghost/admin/app/components/members/filter.hbs
```diff
@@ -84,7 +85,7 @@
                 </button>
                 <button
                     class="gh-btn gh-btn-primary"
-                    data-test-button="members-apply-filter" type="button" {{on "click" this.applyFilter}} {{on "keyup" this.handleSubmitKeyup}}
+                    data-test-button="members-apply-filter" type="button" {{on "click" (fn this.applyFiltersPressed dd)}} {{on "keyup" this.handleSubmitKeyup}}
```
SimonBackx, line 88: When you didn't make changes to the filter, the dropdown didn't close. That is why I created a new action here that also sends a close command to the dropdown.

kevinansfield (approved):

lenabaidakova (approved):

PR #15669, opened 2022-10-21
Added filtered events tables to the analytics page

lenabaidakova:
closes TryGhost/Team#2087

### DIFF
--- a/ghost/admin/app/components/posts/analytics.hbs
+++ b/ghost/admin/app/components/posts/analytics.hbs
@@ -44,47 +44,72 @@
     <Tabs::Tabs class="gh-tabs-analytics" as |tabs|>
         {{#if this.post.hasBeenEmailed}}
             <tabs.tab>
-              <h3>Sent</h3>
-              <p>{{format-number this.post.email.emailCount}}</p>
+                <h3>Sent</h3>
+                <p>{{format-number this.post.email.emailCount}}</p>
             </tabs.tab>
 
-            <tabs.tabPanel>Sent</tabs.tabPanel>
+            <tabs.tabPanel>
+                <Posts::PostActivityFeed
+                    @postId={{this.post.id}}
+                    @eventType="sent"
+                    @linkQuery={{hash filterParam=(concat "emails.post_id:[" this.post.id "]") }}
+                    @linkText="See filtered members for sent"
+                />
+            </tabs.tabPanel>
 
             {{#if this.post.showEmailOpenAnalytics }}
                 <tabs.tab>
-                  <h3>Opened</h3>
-                  <p>{{format-number this.post.email.openedCount}} <strong>{{this.post.email.openRate}}%</strong></p>
+                    <h3>Opened</h3>
+                    <p>{{format-number this.post.email.openedCount}} <strong>{{this.post.email.openRate}}%</strong></p>
                 </tabs.tab>
 
-                <tabs.tabPanel>Opened</tabs.tabPanel>
+                <tabs.tabPanel>
+                    <Posts::PostActivityFeed
+                        @postId={{this.post.id}}
+                        @eventType="opened"
+                        @linkQuery={{hash filterParam=(concat "opened_emails.post_id:[" this.post.id "]") }}
+                        @linkText="See filtered members for opened"
+                    />
+                </tabs.tabPanel>
             {{/if}}
 
             {{#if this.post.showEmailClickAnalytics }}
                 <tabs.tab>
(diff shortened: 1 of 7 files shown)


ghost/admin/app/components/posts/analytics.hbs
```diff
@@ -44,47 +44,72 @@
...
+                <h3>Sent</h3>
+                <p>{{format-number this.post.email.emailCount}}</p>
             </tabs.tab>
 
-            <tabs.tabPanel>Sent</tabs.tabPanel>
+            <tabs.tabPanel>
+                <Posts::PostActivityFeed
+                    @postId={{this.post.id}}
+                    @eventType="sent"
+                    @linkQuery={{hash filterParam=(concat "emails.post_id:[" this.post.id "]") }}
```
SimonBackx, line 55: Note to myself: I'll need to replace the [] in my PR when the filters are updated to always be a single ID.

lenabaidakova:
Will write tests here https://github.com/TryGhost/Team/issues/2128

ghost/admin/app/helpers/activity-feed-fetcher.js
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        if (actionType === actions.showNext) {
+            this.shownEvents = this.shownEvents + this.data.length;
+        }
+
+        this.hasReachedStart = this.totalEvents === meta.pagination.total;
+        this.hasReachedEnd = this.shownEvents === this.totalEvents;
+
+        // todo: it's temporarily fix, pagination breaks if few events happen at the same time, easy to reproduce on email clicks
```
lenabaidakova, line 110: Added temporarily fix, issue https://github.com/TryGhost/Team/issues/2129

ghost/admin/app/helpers/activity-feed-fetcher.js
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        this.eventsBookmarks.push(lastEventDate);
+        this.loadEventsTask.perform({filter}, actions.showNext);
+    }
+
+    @action
+    loadPreviousPage() {
+        this.eventsBookmarks.pop();
+        let filter = `data.created_at:<'${this.eventsBookmarks[this.eventsBookmarks.length - 1]}'`;
```
SimonBackx, line 83: Do we need to push the initial `currentTime` to the bookmarks in `setup`? I would think that going to the second page, then back to the previous page would let `this.eventsBookmarks[this.eventsBookmarks.length - 1]` evaluate to undefined 🤔

SimonBackx (approved):

ghost/admin/app/helpers/activity-feed-fetcher.js
lenabaidakova: We need `currentTime` just to keep first element of every page. For example, when we go to the next page we'll have 2 elements in `eventsBookmarks`.
How they appeared:
- User open first page, we save time of first event of first page, `this.eventsBookmarks = ['firstPageEvent']`
- User open second page, we save time of first event of second page, `this.eventsBookmarks = ['firstPageEvent', 'secondPageEvent']`

When user go to the previous one, we need first element of previous page. So, in this case, we go to the first page (`this.eventsBookmarks[0]`)

Quite messy, but didn't find better solution 😅

ghost/admin/app/components/posts/analytics.hbs
```diff
@@ -112,7 +137,7 @@
 
             {{#if this.showLinks }}
                 {{#if (is-empty this.links) }}
-                    {{!-- Empty state --}}
+                {{!-- Empty state --}}
```
lenabaidakova, line 140: https://github.com/TryGhost/Team/issues/2130#issue-1417955160

ghost/admin/app/helpers/activity-feed-fetcher.js
SimonBackx: You say that `['firstPageEvent']` should be in the `this.eventsBookmarks` initially (which is indeed what I meant), but I don't see the code that does that. 
```
    async setup() {
        const currentTime = moment.utc().format('YYYY-MM-DD HH:mm:ss');
        let filter = `data.created_at:<'${currentTime}'`;

        if (this.args.named.filter) {
            filter += `+${this.args.named.filter}`;
        }

        await this.loadEventsTask.perform({filter}, actions.showNext);
    }
  ```
 This is called after the Resource is loaded, right? But nothing is pushed to `eventsBookmarks`

PR #15670, opened 2022-10-21
Tiers decoupling wire up

allouis:

https://careers.ghost.org

### DIFF
--- a/ghost/core/core/server/services/members/api.js
+++ b/ghost/core/core/server/services/members/api.js
@@ -13,6 +13,7 @@ const SingleUseTokenProvider = require('./SingleUseTokenProvider');
 const urlUtils = require('../../../shared/url-utils');
 const labsService = require('../../../shared/labs');
 const offersService = require('../offers');
+const tiersService = require('../tiers');
 const newslettersService = require('../newsletters');
 const memberAttributionService = require('../member-attribution');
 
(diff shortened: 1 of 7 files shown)


ghost/core/core/server/api/endpoints/tiers.js
```diff
@@ -47,26 +29,11 @@ module.exports = {
         data: [
             'id'
         ],
-        validation: {
```
daniellockyer, line 50: GitHub won't let me comment on the above, but we can delete the following now too:

```
options: [ 
    'include'
],
```

ghost/core/core/server/api/endpoints/tiers.js
```diff
@@ -100,23 +63,11 @@ module.exports = {
         headers: {
             cacheInvalidate: true
         },
-        validation: {
```
daniellockyer, line 103: Should we be keeping `id` as an `option` here? Or is it actually optional now?

allouis: No that's my B - will fix

ghost/payments/lib/payments.js
```diff
@@ -12,13 +14,237 @@ class PaymentsService {
...
+        /** @private */
         this.offersAPI = deps.offersAPI;
         /** @private */
         this.stripeAPIService = deps.stripeAPIService;
 
         DomainEvents.subscribe(OfferCreatedEvent, async (event) => {
             await this.getCouponForOffer(event.data.offer.id);
         });
+
+        DomainEvents.subscribe(TierCreatedEvent, async (event) => {
```
daniellockyer, line 31: This is pretty cool now

PR #15676, opened 2022-10-22
Removed bluebird from misc packages

iamhalldor:
#14882

### DIFF
--- a/ghost/core/core/server/services/auth/setup.js
+++ b/ghost/core/core/server/services/auth/setup.js
@@ -183,7 +183,7 @@ function sendWelcomeEmail(email, mailAPI) {
                     }]
                 };
 
-                return mailAPI.send(payload, {context: {internal: true}})
+                mailAPI.send(payload, {context: {internal: true}})
                     .catch((err) => {
                         err.context = tpl(messages.unableToSendWelcomeEmail);
                         logging.error(err);
--- a/ghost/core/core/server/services/auth/setup.js
+++ b/ghost/core/core/server/services/auth/setup.js
@@ -183,7 +183,7 @@ function sendWelcomeEmail(email, mailAPI) {
                     }]
                 };
 
-                mailAPI.send(payload, {context: {internal: true}})
+                return mailAPI.send(payload, {context: {internal: true}})
                     .catch((err) => {
                         err.context = tpl(messages.unableToSendWelcomeEmail);
                         logging.error(err);
--- a/ghost/api-framework/lib/headers.js
+++ b/ghost/api-framework/lib/headers.js
@@ -1,6 +1,5 @@
 const url = require('url');
 const debug = require('@tryghost/debug')('headers');
-const Promise = require('bluebird');
 const INVALIDATE_ALL = '/*';
 
 const cacheInvalidate = (result, options = {}) => {
--- a/ghost/api-framework/lib/headers.js
+++ b/ghost/api-framework/lib/headers.js
@@ -1,5 +1,6 @@
 const url = require('url');
 const debug = require('@tryghost/debug')('headers');
+const Promise = require('bluebird');
 const INVALIDATE_ALL = '/*';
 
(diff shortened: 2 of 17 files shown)


ghost/core/core/server/services/auth/setup.js
```diff
@@ -183,7 +183,7 @@ function sendWelcomeEmail(email, mailAPI) {
                     }]
                 };
 
-                mailAPI.send(payload, {context: {internal: true}})
```
halldorbjarni, line 186: A quick explainer for this change. 

There seems to be a subtle difference between bluebird and native promises with the order they process callbacks in the event loop. This is particularly important [here](https://github.com/TryGhost/Ghost/blob/main/ghost/api-framework/lib/pipeline.js#L233): 

When executing the `Authentication.Setup` API bluebird will finish processing that call and then process the mail API call **before** returning the context to the test runner. 

With native promises the `Authentication.Setup` API will finish processing and return to the test runner context which fails on assertion that mail has been sent. **After that** it will process the callbacks for the mail API and finish sending the mail. 

This is change forces us to wait for the mail API to completely finish before returning a response from the `Authentication.Setup` call. I don't know if having the mail API async from the Authentication call is necessary @ErisDS? 

If so, there are two other options: 

- to somehow make sure the mail API send call is executed before returning from the Authentication API but don't wait for a response. 
- mock all of the mail API and check it was called instead of mocking the send method inside of the mail API.

halldorbjarni: After reading [this](https://stackoverflow.com/a/36877743/3191816) it seems like we shouldn't try to say assume anything about the execution orders of independent promise chains. And since the mail API call isn't returned it starts a fully independent promise chain. Today I learnt something new... :) 

To my best knowledge the email assertion mechanism only worked "accidentally" because bluebird promise callbacks are handled in different order compared to native promises? If the mail API call needs to be a fire and forget call then I think our best option is to mock the call to the mail API call itself and assert on that.

ErisDS: This is super fun. In our newer e2e tests we have a "mockManager" which can mock the mail sending and we assert based on that:

https://github.com/TryGhost/Ghost/blob/8927432801b7541bcbc5d9cb23237c39d0972a7c/ghost/core/test/e2e-api/admin/newsletters.test.js#L38

https://github.com/TryGhost/Ghost/blob/8927432801b7541bcbc5d9cb23237c39d0972a7c/ghost/core/test/e2e-api/admin/newsletters.test.js#L231

If we can use the same tooling or build a similar one for this use case that would be great, but I think you're about to hit on more problems we have with the auth endpoints not working with the new e2e test framework due to some underlying issues with how the tmp dir is managed between tests 😬

ErisDS (changes requested):
Hey @halldorbjarni so grateful for all this work! This is quite a large changeset, and even with reviewing as fast as I can, things are gonna fall behind.

It might be an idea to roll back the harder changes here and put them into a new, small PR that we can work on and then get the simpler changes here merged?

ghost/core/core/server/services/auth/setup.js
halldorbjarni: Hey thanks for the review @ErisDS. Before I add more tooling to the mocking library above I just want to confirm the following: we want to keep the mail API a "fire and forget" call (i.e. on a new promise chain) when calling the Authentication API? 

> It might be an idea to roll back the harder changes here and put them into a new, small PR that we can work on and then get the simpler changes here merged?

Yes, I agree. I can carve out the removal of bluebird from the `api-framework` module to another PR.

ErisDS (approved):

ErisDS:
Hey @halldorbjarni thanks for splitting the fiddly bit out 👍 got this one merged & labelled now. Apologies for the delay we had a bit of a crazy week polishing some new features 🎉 

Am working through the PR list as fast as I can now and will checkout your other one ASAP

PR #15682, opened 2022-10-24
Added email sent events

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2137

For the analytics page, we need the sent events to show up immediately after sending an email. Otherwise we need to wait for emails to be marked as received (which takes too long) before being able to show them on the analytics page.

This adds the email_sent_event, which is hidden by default everywhere and used on the analytics page.


lenabaidakova (approved):
🙌

PR #15685, opened 2022-10-25
Removed bluebird from `api-framework` module

iamhalldor:
#14882

### DIFF
--- a/ghost/core/core/server/services/auth/setup.js
+++ b/ghost/core/core/server/services/auth/setup.js
@@ -183,7 +183,7 @@ function sendWelcomeEmail(email, mailAPI) {
                     }]
                 };
 
-                mailAPI.send(payload, {context: {internal: true}})
+                return mailAPI.send(payload, {context: {internal: true}})
                     .catch((err) => {
                         err.context = tpl(messages.unableToSendWelcomeEmail);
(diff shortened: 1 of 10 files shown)


ghost/core/core/server/services/auth/setup.js
```diff
@@ -183,7 +183,7 @@ function sendWelcomeEmail(email, mailAPI) {
                     }]
                 };
 
-                mailAPI.send(payload, {context: {internal: true}})
```
halldorbjarni, line 186: Waiting on confirmation from @ErisDS if we can simply make the mail API call synchronous with the rest of the Authentication API. 

If not, I'll revert this line change and make the necessary changes to the mail mocking library.

ErisDS: I think we can change this to sync and see if that creates other issues 👍

ErisDS (approved):

ErisDS:
Hey @halldorbjarni I've merged and labelled this one now. Thanks for all of your contributions 🙏 

We're down to less than 100 usages of bluebird now I think... incredible. Not sure if you're planning to keep going outside of hacktoberfest but we'd love it if you did ❤️

PR #15689, opened 2022-10-25
Fixed missing column values for default paid tiers

daniellockyer:
fixes https://github.com/TryGhost/Toolbox/issues/455 refs https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/data/migrations/versions/5.19/2022-09-02-20-52-backfill-new-product-columns.js

- the referenced migration does not handle backfilling the currency/monthly_price/yearly_price for the default paid tiers where they do not originate from Stripe
- this is causing issues in Ghost because of the missing data
- this migration backfills the columns for products where they are paid but do not currently contain values due to the bug above

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.21/2022-10-25-12-05-backfill-missed-products-columns.js
+++ b/ghost/core/core/server/data/migrations/versions/5.21/2022-10-25-12-05-backfill-missed-products-columns.js
@@ -1,0 +1,35 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info(`Fixing currency/monthly_price/yearly_price values for default paid tiers`);
+
+        const currencyUpdated = await knex('products')
+            .update('currency', 'usd')
+            .where({
+                currency: null,
+                type: 'paid'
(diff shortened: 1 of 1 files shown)


allouis (approved):

ghost/core/core/server/data/migrations/versions/5.21/2022-10-25-12-05-backfill-missed-products-columns.js
```diff
@@ -0,0 +1,35 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info(`Fixing currency/monthly_price/yearly_price values for default paid tiers`);
+
+        const currencyUpdated = await knex('products')
+            .update('currency', 'usd')
```
daniellockyer, line 9: @allouis I've inserted it as lowercase here as that's what we seem to insert in existing entries, but your original SQL was uppercase. Does it matter? I presume we handle the lowercase/uppercase appropriately

allouis: It's all good - the code handles converting them 👍

PR #15691, opened 2022-10-25
Add command to generate demo data

sam-lord:
refs: https://github.com/TryGhost/Toolbox/issues/440

New command to generate demo data, creates data for over 20 tables in Ghost, suitable for testing most features of the dashboard, as well as making guided product tours using newsletters, tiers, many posts and tags.

Still a work in progress, and the code style will be updated as we find better ways to work on this as other engineers join in. But I wanted to merge early while this is only a 1,500+ line diff.


yarn.lock
```diff
@@ -2447,6 +2447,11 @@
     minimatch "^3.1.2"
```
daniellockyer, line 2447: If you run `yarn` from the top-level repo, it should clear up the `moment` and `@tryghost/logging` changes - there's some weird yarn thing where it ignores resolutions if you run it from a workspace package

PR #15693, opened 2022-10-26
Added linking between member and subscription created events

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2160

- Adds a `batch_id` to both events that contain the same ID if they were created at the same time.
- Removes duplicate signup/conversion events using the batch_id
- Requires an update in mongo-knex to work (refs https://ghost.slack.com/archives/C02G9E68C/p1666773313272409?thread_ts=1666767872.375009&cid=C02G9E68C)
- Some dependencies needed an update to load the latest mongo-knex
- Added tiers to membersUtils, loaded on startup (we can start to use this instead of fetching it every time)


ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-04-50-member-subscription-created-request-id.js
```diff
@@ -0,0 +1,73 @@
...
+
+        // Create batches (insertBatch doesn't support the onConflict option)
+        const batchSize = 1000;
+    
+        for (let i = 0; i < rows.length; i += batchSize) { // eslint-disable-line no-restricted-syntax
+            const batch = rows.slice(i, i + batchSize);
+
+            // Update the members_created_events table using INSERT ON DUPLICATE KEY UPDATE trick
+            const response1 = await knex('members_created_events').insert(batch.map((r) => {
+                return {
```
allouis, line 39: Do we need all of the columns here, or can we just have `id` and `request_id`?

ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-04-50-member-subscription-created-request-id.js
```diff
@@ -0,0 +1,73 @@
...
+        // Attach a unique id to each row
+        for (const row of rows) { // eslint-disable-line no-restricted-syntax
+            row.request_id = ObjectId().toHexString();
+        }
+
+        // Create batches (insertBatch doesn't support the onConflict option)
+        const batchSize = 1000;
+    
+        for (let i = 0; i < rows.length; i += batchSize) { // eslint-disable-line no-restricted-syntax
+            const batch = rows.slice(i, i + batchSize);
```
allouis, line 35: We've previously used the `chunk` method from `lodash` instead of this - not a big deal tho

allouis (commented):
Looks good to me, I do wonder about the `request_id` naming though. Perhaps `transaction_id` or `group_id` or `batch_id` works better, as it is more specific to the concept of a batch of events, rather than a request 🤷

ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-04-50-member-subscription-created-request-id.js
allouis: After testing I see we do need all the columns

SimonBackx: Yup, MySQL complains that they don't have a default set if we don't send them

SimonBackx:
I like the `batch_id` naming! Wasn't a fan of `transaction_id` because that can be confused with SQL transactions

allouis:
Yeah agree RE transaction!

allouis (approved):

ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-04-50-member-subscription-created-request-id.js
```diff
@@ -0,0 +1,73 @@
+const logging = require('@tryghost/logging');
+const ObjectId = require('bson-objectid').default;
+const {createTransactionalMigration} = require('../../utils');
+const DatabaseInfo = require('@tryghost/database-info');
+
+// This migration links together members_created_events and members_subscription_created_events
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        if (DatabaseInfo.isSQLite(knex)) {
```
daniellockyer, line 10: Is there a specific reason we're skipping doing this on SQLite? Just curious

ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-04-50-member-subscription-created-request-id.js
daniellockyer: @SimonBackx Yeah I'd prefer to use `_.chunk` here if possible 😬

ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-04-50-member-subscription-created-request-id.js
```diff
@@ -0,0 +1,73 @@
...
+    
+        for (let i = 0; i < rows.length; i += batchSize) { // eslint-disable-line no-restricted-syntax
+            const batch = rows.slice(i, i + batchSize);
+
+            // Update the members_created_events table using INSERT ON DUPLICATE KEY UPDATE trick
+            const response1 = await knex('members_created_events').insert(batch.map((r) => {
+                return {
+                    id: r.m_id,
+                    batch_id: r.batch_id,
+                    member_id: r.member_id,
```
daniellockyer, line 42: Small suggestion but can you comment the fields that you were forced to include because it's an INSERT vs the ones we actually care about (`id`, `batch_id`)?

ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-04-50-member-subscription-created-request-id.js
SimonBackx: `TIMESTAMPDIFF` isn't supported on SQLite (we could probably search for a workaround), and I wasn't sure about 'conflict on insert' working properly over there. For non-production databases I couldn't really see a value in merging the events

daniellockyer (approved):

PR #15694, opened 2022-10-26
Migrated members importer to use tiers

naz:
Hey @allouis! This is more of an fyi on how I've approached structuring the migration from deprecated "getDefaultProduct" to tier's API. Let me know if this is in line with what we've discussed!

### DIFF
--- a/ghost/tiers/lib/InMemoryTierRepository.js
+++ b/ghost/tiers/lib/InMemoryTierRepository.js
@@ -17,6 +17,8 @@ class InMemoryTierRepository {
     toPrimitive(tier) {
         return {
             ...tier,
+            active: (tier.status === 'active'),
+            type: tier.type,
             id: tier.id.toHexString()
         };
     }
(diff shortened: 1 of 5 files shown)


ghost/tiers/lib/InMemoryTierRepository.js
```diff
@@ -17,6 +17,8 @@ class InMemoryTierRepository {
     toPrimitive(tier) {
         return {
             ...tier,
+            active: (tier.status === 'active'),
```
naz, line 20: This bit was confusing af to figure out. Basically anything that is a getter is not later accessible through JSON filtering. We have to do explicit mapping here to be able to filter by those values

allouis: Damn okay - we can look at using the `toJSON` method of the objects here instead, which should make things easier

allouis (approved):

PR #15695, opened 2022-10-26
Added feedback buttons to emails

lenabaidakova:
closes TryGhost/Team#2075

### DIFF
--- a/ghost/core/core/server/services/mega/feedback-buttons.js
+++ b/ghost/core/core/server/services/mega/feedback-buttons.js
@@ -18,15 +18,27 @@ const generateLinks = (postId, uuid, html) => {
         0
     );
 
-    html = html.replace(templateStrings.like, positiveLink.href);
-    html = html.replace(templateStrings.dislike, negativeLink.href);
+    html = html.replace(new RegExp(templateStrings.like, 'g'), positiveLink.href);
+    html = html.replace(new RegExp(templateStrings.dislike, 'g'), negativeLink.href);
 
     return html;
 };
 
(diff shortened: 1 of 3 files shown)


lenabaidakova:
Need to improve for Outlook https://github.com/TryGhost/Team/issues/2167

moreofmorris:
Hey @lenabaidakova, I just tested in Gmail app and this is what shown for me

CleanShot 2022-10-26 at 10 40 48@2x

I'm not sure if this helps simplify this but the more I think about it, the most we should probably just use black for the icon and text? That would massively simplify this, is that right?

moreofmorris (approved):

ghost/core/core/server/services/mega/feedback-buttons.js
```diff
@@ -18,15 +18,27 @@ const generateLinks = (postId, uuid, html) => {
...
 
 const getTemplate = (accentColor) => {
-    const likeButtonHtml = getButtonHtml(templateStrings.like, 'More like this', accentColor);
-    const dislikeButtonHtml = getButtonHtml(templateStrings.dislike, 'Less like this', accentColor);
+    const likeButtonHtml = getButtonHtml(
+        templateStrings.like,
+        'More like this',
+        accentColor,
+        'like-icon',
+        'https://static.ghost.org/v4.0.0/images/thumbs-up.png'
```
SimonBackx, line 33: I think at this point it would be better to upload them at /v5.0.0, no?

SimonBackx (approved):
Nice change! Only left a small comment

PR #15696, opened 2022-10-26
Backfilled `subscriptions` from existing subscription states

daniellockyer:
fixes https://github.com/TryGhost/Team/issues/2031, https://github.com/TryGhost/Team/issues/2032 and https://github.com/TryGhost/Team/issues/2033

- this migration backfills the `subscriptions` table from the various subscription states a member can be in

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-13-32-backfill-comped-subscriptions.js
+++ b/ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-13-32-backfill-comped-subscriptions.js
@@ -1,0 +1,42 @@
+const logging = require('@tryghost/logging');
+const ObjectId = require('bson-objectid').default;
+
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const compedMembers = await knex('members')
+            .select([
+                'members.id',
+                'members_products.product_id',
+                'members_products.expiry_at'
+            ])
+            .leftJoin('members_products', 'members.id', 'members_products.member_id')
+            .where('members.status', 'comped');
+
+        if (!compedMembers.length) {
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-13-32-backfill-subscriptions.js
```diff
@@ -0,0 +1,112 @@
...
+                if (batchResponse[0] !== 0) {
+                    logging.error(`Inserted ${batchResponse[0]} members_stripe_customers_subscriptions, expected 0`);
+                    throw new Error('Rolling back');
+                }
+            }
+        } else {
+            logging.info(`No new Ghost subscriptions need backfilling into members_stripe_customers_subscriptions`);
+        }
+    },
+    async function down(knex) {
```
daniellockyer, line 108: This needs changing/thinking about, but it's good for testing because it resets us back

ghost/core/core/server/data/migrations/versions/5.21/2022-10-26-13-32-backfill-subscriptions.js
```diff
@@ -0,0 +1,112 @@
...
+        if (!freeProduct) {
+            logging.warn(`Unable to fetch free product - skipping subscription backfill`);
+            return;
+        }
+
+        const now = knex.raw('CURRENT_TIMESTAMP');
+
+        const ghostSubscriptionIds = [];
+
+        const rowsToInsert = members.map((m) => {
```
daniellockyer, line 42: I should probably consider looking at the `members_stripe_customers_subscriptions.type` first because we a free member might have a canceled paid subscription that we should also include

daniellockyer:
Closing as the pitch has been shelved for now - migrations have been copied into the ref'd issues

PR #15697, opened 2022-10-26
added tests for custom theme settings importer

illiteratewriter:


illiteratewriter:
Hey @ErisDS I've added tests for importing custom theme settings, but I was unable to figure out a way to test the `activate` function. The current code I've written for activate is sort of incorrect. Currently the tests is relying on the fact that `return models.CustomThemeSetting.findAll({})` is taking more time than `models.Settings.findOne`, but there might be times when the other one is faster.

Can you help me write the test correctly? (i am thinking of something along the lines of [waitFor](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor) from react testing library)

naz:
Ha! You have found and interesting Bluebird bug @illiteratewriter 😄  For reasons I have not investigated, the "then" method in the test where we do the `sinon.assert.calledOnce(activateStub)` check is executed before the "then" [chain in the Importer](https://github.com/TryGhost/Ghost/blob/aba255d27656366217e5e7f37bf69f3265eaad40/ghost/core/core/server/data/importer/importers/data/custom-theme-settings.js#L69-L75) finishes. Essentially, importer's then and test then are being run in parallel 🤯 

Having `return models.CustomThemeSetting.findAll({})` gave us time to "wait" for the importer to finish (without guarantees). It's a hack for sure. We could go two ways here:
1. Add a nice comment in the test explaining why the hack is there, and link it to&in the [bluebird cleanup issue](https://github.com/TryGhost/Ghost/issues/14882), so we don't forget to remove the hack
2. Remove Bluebird from the importer and rely on native promises (async/await) instead of Bluebird's `Promise.all`

Having a "wait for" would also be sort of a hack, because ideally we should have finished execution of `doImport` in the test after it's been resolved.

naz:
[This PR](https://github.com/TryGhost/Ghost/pull/15758) where we remove bluebird opened by @sam-lord should very likely fix these nasty hacks 😊

illiteratewriter:
@naz in that case i will have this commit rebased and fixed once the PR to remove bluebird is merged

ErisDS:
Hey @illiteratewriter his PR is now pretty ancient so I'm gonna close it, but if you're still interested in working on this, please re-raise the PR fresh and we'll figure out how to work through it. Bluebird was fully removed from Ghost abut a year ago, so you shouldn't hit the same issues.

(Please don't comment to request re-open, comments get lost, new PRs trigger new automations and will get our attention)

PR #15704, opened 2022-10-27
Removed usages of products repository

naz:
refs https://github.com/TryGhost/Team/issues/2077

@allouis this is just an FYI about what about to land with serializers substitution to use Tiers package instead of productsRepository.


naz:
This changeset might get picked up again when we are back to working on members decoupling

daniellockyer:
@naz What was left to do in this PR to be able to ship it?

naz:
@daniellockyer I think fixing the snapshots where currency casing has changed back to uppercase.... but don't think that is relevant anymore as we have shipped the breaking change anyway. so would keep the lowercase change (nobody seem to have complained about it so far)

PR #15705, opened 2022-10-27
Added new setting to toggle tracking of member sources

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2168

- the new setting allows site owners to control if they want to track the sources for new member signups and subscriptions
- its switched on by default, but can be toggled off from new analytics settings page


ghost/core/core/server/data/migrations/versions/5.15/2022-09-14-12-46-add-email-track-clicks-setting.js
```diff
@@ -1,8 +1,8 @@
 const {addSetting} = require('../../utils');
 
 module.exports = addSetting({
-    key: 'email_track_clicks',
```
daniellockyer, line 4: Is there a reason you're changing the old migration? 😅

rishabhgrg: Other than the fact my 🧠 is a mess, no other reason! 😅  Updated now

daniellockyer (approved):

SimonBackx (approved):

PR #15712, opened 2022-10-27
Updated breadcrumbs for member page

lenabaidakova:
closes TryGhost/Team#2140

### DIFF
--- a/ghost/admin/app/templates/member.hbs
+++ b/ghost/admin/app/templates/member.hbs
@@ -10,10 +10,6 @@
                     <LinkTo @route="posts.analytics" @models={{this.fromAnalytics}}>
                         Analytics
                     </LinkTo>
-                    {{svg-jar "arrow-right-small"}}
-                    <LinkTo @route="members" data-test-link="members-back">
-                        Members
-                    </LinkTo>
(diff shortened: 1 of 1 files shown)


ghost/admin/app/templates/member.hbs
```diff
@@ -10,10 +10,6 @@
                     <LinkTo @route="posts.analytics" @models={{this.fromAnalytics}}>
                         Analytics
                     </LinkTo>
-                    {{svg-jar "arrow-right-small"}}
-                    <LinkTo @route="members" data-test-link="members-back">
-                        Members
-                    </LinkTo>
```
SimonBackx, line 16: I'm afraid this breaks the other logic too. When going to 'see members' clicking a member -> the breadcrumbs now should include members 😬

SimonBackx: So there are 2 possibilities:
- If coming from the members page, and didn't came from analytics to the members page, just show 'members'
- If coming directly from the analytics page, just show 'analytics' but skip members
- If coming from the analytics page, to the members page, to a specific member, we need to show both analytics and members.

See, this is getting a bit complicated. And in Ember we should somehow be able to track the previous routes. So far that implementation has been pretty hacky, so we don't have a pattern for that yet.

PR #15734, opened 2022-10-31
🐛 Add ability to cache comments count endpoint

lenabaidakova:
closes TryGhost/Team#2094

PR to comments repo https://github.com/TryGhost/comments-ui/pull/12

### DIFF
--- a/ghost/core/core/server/web/comments/routes.js
+++ b/ghost/core/core/server/web/comments/routes.js
@@ -1,19 +1,24 @@
 const express = require('../../../shared/express');
+const config = require('../../../shared/config');
 const api = require('../../api').endpoints;
 const {http} = require('@tryghost/api-framework');
+const shared = require('../shared');
 
 const bodyParser = require('body-parser');
 const membersService = require('../../../server/services/members');
 
 module.exports = function apiRoutes() {
     const router = express.Router('comment api');
-
     router.use(bodyParser.json({limit: '50mb'}));
 
     // Global handling for member session, ensures a member is logged in to the frontend
     router.use(membersService.middleware.loadMemberSession);
 
-    router.post('/counts', http(api.commentsMembers.counts));
+    const countsCache = shared.middleware.cacheControl(
+        'public',
+        {maxAge: config.get('caching:commentsCountAPI:maxAge')}
+    );
+    router.get('/counts', countsCache, http(api.commentsMembers.counts));
 
     router.get('/', http(api.commentsMembers.browse));
     router.get('/:id', http(api.commentsMembers.read));
--- a/ghost/core/core/shared/config/defaults.json
+++ b/ghost/core/core/shared/config/defaults.json
@@ -140,6 +140,9 @@
         },
         "contentAPI": {
             "maxAge": 0
+        },
+        "commentsCountAPI": {
+            "maxAge": 0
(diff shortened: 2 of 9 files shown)


ghost/core/core/shared/config/defaults.json
```diff
@@ -140,6 +140,9 @@
         },
         "contentAPI": {
             "maxAge": 0
+        },
+        "comments": {
```
naz, line 144: I do like the `contentAPI` convention used above. This is an API we are caching, so maybe `commentsAPI` would be a better fit. Also, with the discussion on the issue itself _if_ we end up only caching the comments endpoint long-term, than `commentsCountAPI` would be even better name.

ghost/core/core/server/web/comments/routes.js
```diff
@@ -1,19 +1,22 @@
...
 const {http} = require('@tryghost/api-framework');
+const shared = require('../shared');
 
 const bodyParser = require('body-parser');
 const membersService = require('../../../server/services/members');
 
 module.exports = function apiRoutes() {
     const router = express.Router('comment api');
 
+    router.use(shared.middleware.cacheControl('public', {maxAge: config.get('caching:comments:maxAge')}));
```
naz, line 13: The issue we are solving here is for one specific endpoint and not the whole API. I think we would be most safe to apply this caching rule to the counts endpoint first to unblock the issue and do more research around correct approach for the whole Comments API

naz (changes requested):
These changes look good. The controller/api level changes are good to go into main I think. The part around setting a custom cache header should to be endpoint-specific.

ghost/core/core/shared/config/defaults.json
lenabaidakova: Renamed it to `commentsCountAPI`

ghost/core/core/server/web/comments/routes.js
lenabaidakova: Added cache just of counts API

lenabaidakova:
@naz , thank you for explanation 🙌 

Fixed next issues:
- Renamed setting in config to `commentsCountAPI`
- Applied just for one count API
- Added `stale-while-revalidate` option

ghost/mw-cache-control/lib/mw-cache-control.js
```diff
@@ -12,10 +12,18 @@ const isString = require('lodash/isString');
  * @param {'public'|'private'} profile Use "private" if you do not want caching
  * @param {object} [options]
  * @param {number} [options.maxAge] The max-age in seconds to use when profile is "public"
+ * @param {number} [options.staleWhileRevalidate] The stale-while-revalidate in seconds to use when profile is "public"
```
naz, line 15: Oh exciting! Nice 😊

naz (approved):
Yup! All of this looks about right 👌  If I had to be extra nit-picky about it, would separate it into two commits and `staleWhileRevalidate` functionality there with specific explanation for whys and hows 😊

lenabaidakova:
> Yup! All of this looks about right 👌 If I had to be extra nit-picky about it, would separate it into two commits and `staleWhileRevalidate` functionality there with specific explanation for whys and hows 😊

Fixed :)

PR #15739, opened 2022-11-01
replaced Promise.map() with .all() in services/bulk-email #14882

CarolBarno:

https://careers.ghost.org

### DIFF
--- a/ghost/core/core/server/services/bulk-email/bulk-email-processor.js
+++ b/ghost/core/core/server/services/bulk-email/bulk-email-processor.js
@@ -1,5 +1,4 @@
 const _ = require('lodash');
-const Promise = require('bluebird');
 const moment = require('moment-timezone');
 const errors = require('@tryghost/errors');
 const tpl = require('@tryghost/tpl');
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/services/bulk-email/bulk-email-processor.js
```diff
@@ -80,14 +79,14 @@ module.exports = {
...
-        const batchResults = await Promise.map(batchIds, async ({id: emailBatchId, member_segment: memberSegment}) => {
+        const batchResults = await Promise.all(batchIds.map(async ({id: emailBatchId, member_segment: memberSegment}) => {
             try {
                 await this.processEmailBatch({emailBatchId, options, memberSegment});
                 return new SuccessfulBatch(emailBatchId);
             } catch (error) {
                 return new FailedBatch(emailBatchId, error);
             }
-        }, {concurrency: 10});
+        }, {concurrency: 10}));
```
daniellockyer, line 89: This would not be valid because native `Promise.all` does not take `{concurrency: X}` as a second argument

daniellockyer: Also, is this code legacy now @SimonBackx?

SimonBackx: This is indeed legacy and not used as long as emailStability is a GA flag

daniellockyer:
Hey @CarolBarno, this file no longer exists, so I'm going to close the PR 

We're really close to finalizing the Bluebird issue though, in case you want to pick up some other ones 🙂

PR #15742, opened 2022-11-01
✨ Added specific newsletter support for bulk unsubscribes

ronaldlangeveld:
ref https://github.com/TryGhost/Team/issues/2013

Added support to be able to bulk unsubscribe a selected (filtered) list on members from specific newsletters.


allouis (commented):
I think the general approach makes sense, but I think we should consider creating a `MemberNewsletter` bookshelf model 🤔 

That way we can use `findAll` rather than `bulkFind` and then we can use `bulkDelete` method to delete all the records.

ghost/admin/app/components/members/modals/bulk-unsubscribe.hbs
```diff
@@ -43,11 +43,28 @@
...
+                        </p>
+                        <OneWaySelect
+                            @value={{this.selectedNewsletterId}}
+                            @options={{this.newsletterList}}
+                            @optionLabelPath="name"
+                            @optionValuePath="value"
+                            @optionTargetPath="value"
+                            @update={{this.setSelectedNewsletter}}
+                            data-test-select="members-single-newsletter-unsubscribe"
+                        />
```
kevinansfield, line 59: Small thing but it's good to always indent html to match the conditionals, helps to keep the code understandable at a glance and avoid problems like we saw in the flag cleanup where the lack of proper indentation made it tricky to see which blocks should be cleaned up.

ghost/admin/app/components/members/modals/bulk-unsubscribe.js
```diff
@@ -19,27 +22,62 @@ export default class BulkUnsubscribeMembersModal extends Component {
...
+    }
+
+    get newsletterList() {
+        const newsletters = this.store.peekAll('newsletter');
+        const activeNewsletters = newsletters.filter(newsletter => newsletter.status !== 'archived');
+        let list = [{
+            name: 'All newsletters',
+            value: 'all'
+        }];
+        activeNewsletters.map((newsletter) => {
```
kevinansfield, line 42: If you're not using the return value of a map please use `forEach` instead.

kevinansfield (approved):
A couple of style related comments to get in the habit of doing but otherwise the Ember stuff LGTM 👍

ghost/members-api/lib/repositories/member.js
```diff
@@ -726,9 +729,21 @@ module.exports = class MemberRepository {
         const memberIds = memberRows.map(row => row.id);
 
         if (data.action === 'unsubscribe') {
-            return await this._Member.bulkDestroy(memberIds, 'members_newsletters', {column: 'member_id'});
+            const hasNewsletterSelected = (Object.prototype.hasOwnProperty.call(data, 'newsletter') && data.newsletter !== null);
+            if (hasNewsletterSelected) {
+                const newsletters = await this._MemberNewsletter.where('newsletter_id', data.newsletter).fetchAll();
```
allouis, line 734: Rather than use `reduce` below to get the correct list we should be able to make an SQL query to get the correct data.

So could do something like `this._MemberNewsletter.where().andWhereIn()` Or we can use the `getFilteredCollectionQuery` method above and pass in an NQL filter like `newsletter_id:blah+member_id:[id1,id2,id3...]`

ghost/members-api/lib/repositories/member.js
```diff
@@ -726,9 +729,21 @@ module.exports = class MemberRepository {
...
+                const toUnsubscribe = newsletters.toJSON().reduce((acc, newsletter) => {
+                    if (memberIds.includes(newsletter.member_id)) {
+                        acc.push(newsletter.id);
+                    }
+                    return acc;
+                }, []);
+                return await this._Member.bulkDestroy(toUnsubscribe, 'members_newsletters', {column: 'id'});
```
allouis, line 741: This can be written as `this._MemberNewsletter.bulkDestroy(toUnsubscribe)`

ghost/members-api/lib/repositories/member.js
```diff
@@ -726,9 +729,21 @@ module.exports = class MemberRepository {
...
+            }
+            if (!hasNewsletterSelected) {
+                return await this._Member.bulkDestroy(memberIds, 'members_newsletters', {column: 'member_id'});
```
allouis, line 744: What would be really cool is if we could update the `bulkDestroy` method to make calls like this nicer, imagine if we could instead do:
```js
this._MemberNewsletter.bulkDestroy({member_id: memberIds});


// bulk-operations.js

async bulkDestroy(data, tableName, options) {
  if (!Array.isArray(data)) {
    const [key, ...keys] = Object.keys(data);

    assert(keys.length === 0);
    return this.bulkDestroy(data[key], tableName, {column: key});
  }
  // ...
}
```

Not necessarily in scope for this PR but would be much nicer!

allouis (commented):
Looks good just a couple comments!

PR #15745, opened 2022-11-01
Remove bluebird from `server/api/endpoints`

iamhalldor:
#14882 

🚀 we're almost there!

### DIFF
--- a/ghost/core/core/server/api/endpoints/db.js
+++ b/ghost/core/core/server/api/endpoints/db.js
@@ -1,4 +1,3 @@
-const Promise = require('bluebird');
 const moment = require('moment-timezone');
 const dbBackup = require('../../data/db/backup');
 const exporter = require('../../data/exporter');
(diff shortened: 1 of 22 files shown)


ghost/core/core/server/api/endpoints/db.js
```diff
@@ -113,15 +112,15 @@ module.exports = {
 
                     return models.Post.findAll(queryOpts)
                         .then((response) => {
-                            return Promise.map(response.models, (post) => {
+                            return Promise.all(response.models.map((post) => {
                                 return models.Post.destroy(Object.assign({id: post.id}, queryOpts));
-                            }, {concurrency: 100});
```
halldorbjarni, line 118: Just highlighting the only functional change here. I don't know how important it is to keep the concurreny at 100 for post deletions. Unfortunately the history behind that change is gone due to a refactor from what I can tell...

If we need the concurrency limit I can look into promise pool libraries.

ErisDS: There's not a lot to go on, but the blame for that change is here: https://github.com/halldorbjarni/Ghost/commit/c8f2dd11bada27e977beca249bf41898daa31278

I think it's to do with ensuring we don't send too many events all at once.

halldorbjarni: @ErisDS I took a crack at writing a promise pool function myself: https://github.com/TryGhost/framework/pull/86

Seems like it would be useful to have for throttling these large async operations, there are several places in the code that use that Bluebird concurrency option.

halldorbjarni: @ErisDS thanks for merging the framework PR. Do you want me to make a PR to bump the version of the promise library in the framework repo or is this something that Ghost employees handle?

ErisDS: Sorry for the radio silence, it's been _manic_. The version of the framework library is now bumped.

halldorbjarni: No problem @ErisDS. This change should be integrated with the new promise pool library function now.

PR #15747, opened 2022-11-01
Fixed feature flags not being correct immediately after completing sign-in process

kevinansfield:
closes https://github.com/TryGhost/Team/issues/2110

- added failing test showing feature service @feature properties weren't updating after no-auth->authed updates of config service
  - shows `labs` and feature properties on the `feature` service are not reacting to changes in the `settings` service
  - removing the `@computed` on the `feature.labs` getter stops it being cached but it then fails on the `feature.testFlag` property
- removed usage of `Ember.computed` to avoid any caching behaviour and let properties act like normal getters/setters

### DIFF
--- a/ghost/admin/app/services/feature.js
+++ b/ghost/admin/app/services/feature.js
@@ -1,40 +1,46 @@
 import $ from 'jquery';
-import Ember from 'ember';
 import EmberError from '@ember/error';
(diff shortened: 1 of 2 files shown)


kevinansfield:
Closing in favour of https://github.com/TryGhost/Ghost/pull/15764

PR #15758, opened 2022-11-02
Removed Bluebird and sequence from importer

sam-lord:
refs: https://github.com/TryGhost/Toolbox/issues/432

Bluebird promises and the sequence method were definitely causing some issues with memory retention.

It seems like there's still an issue with a memory leak that prevents extremely large imports from processing fully, but this will make them easier to track down.

Currently it seems like removing the Knex transaction will prevent any memory leak. I'll do some investigation to see why that is the case.

### DIFF
--- a/ghost/core/core/server/data/importer/importers/data/custom-theme-settings.js
+++ b/ghost/core/core/server/data/importer/importers/data/custom-theme-settings.js
@@ -1,5 +1,4 @@
 const _ = require('lodash');
-const Promise = require('bluebird');
 const debug = require('@tryghost/debug')('importer:roles');
 const BaseImporter = require('./base');
 const models = require('../../../../models');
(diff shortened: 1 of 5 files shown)


sam-lord:
Whilst most of the code was transferred over to achieve the same logic, it seems like the code for tags relied on a race condition -- going through step by step with a debugger is hard because of the extensive use of Bluebird & then()-chains, but it seems like the old check `if (tag)` never returned true.

ghost/core/core/server/data/importer/importers/data/custom-theme-settings.js
```diff
@@ -18,64 +17,57 @@ class CustomThemeSettingsImporter extends BaseImporter {
...
-            models.Settings.findOne({key: 'active_theme'})
-                .then((theme) => {
-                    const currentTheme = theme.get('value');
-                    if (this.dataToImport.some(themeSetting => themeSetting.theme === currentTheme)) {
-                        activate(currentTheme);
-                    }
-                });
+        models.Settings.findOne({key: 'active_theme'}).then((theme) => {
+            const currentTheme = theme.get('value');
+            if (this.dataToImport.some(themeSetting => themeSetting.theme === currentTheme)) {
```
illiteratewriter, line 61: @sam-lord since you're using `this.dataToImport.shift()` above, this will always become falsy. Haven't checked other files. Saw this because I was working on this file on another PR.

sam-lord: Thanks for noticing this! I've fixed this by checking the condition first, then activating the theme if the condition is true after importing all of the data.

I've checked all other importers and custom-theme-settings was the only one which referenced the dataToImport array during import - so this should be fixed now.

sam-lord:
@ErisDS I've been looking at the error here on the DB tests. It's expecting 6 import errors and receiving 7. The extra error is for a duplicate tag:

```json
{
    "message": "Entry was not imported and ignored. Detected duplicated entry.",
    "help": "Tag",
    "context": "{\"id\":\"636a3f3d77eb308d952b9873\",\"name\":\"Getting Started\",\"slug\":\"getting-started\",\"description\":null,\"feature_image\":null,\"parent_id\":null,\"visibility\":\"public\",\"og_image\":null,\"og_title\":null,\"og_description\":null,\"twitter_image\":null,\"twitter_title\":null,\"twitter_description\":null,\"meta_title\":null,\"meta_description\":null,\"codeinjection_head\":null,\"codeinjection_foot\":null,\"canonical_url\":null,\"accent_color\":null,\"created_at\":\"2022-11-08 11:36:27\",\"updated_at\":\"2022-11-08 11:36:27\"}"
},
```

I haven't been able to make a change to the tag importer that has passed all tests. I believe that this additional import error should be expected but wanted to check your understanding here before modifying a test to match behaviour.

PR #15762, opened 2022-11-03
Cleaned up GA Flags

ronaldlangeveld:
ref https://github.com/TryGhost/Team/issues/2126

Cleaned up redundant GA feature flags.


ghost/core/core/server/models/member.js
```diff
@@ -154,11 +154,7 @@ const Member = ghostBookshelf.Model.extend({
             .query((qb) => {
                 // avoids bookshelf adding a `DISTINCT` to the query
                 // we know the result set will already be unique and DISTINCT hurts query performance
-                if (labs.isSet('compExpiring')) {
-                    qb.columns('products.*', 'expiry_at');
```
daniellockyer, line 158: If we're removing `compExpiring`, do we not need to keep `                    qb.columns('products.*', 'expiry_at');`?

ghost/admin/app/components/modal-tier.hbs
```diff
@@ -305,22 +250,18 @@
                                                 </span>
                                             {{/if}}
                                         {{/if}}
-                                    {{else}}
```
kevinansfield, line 308: I think the removed `{{else}}` should be the one above? This one is the opposite of the `{{#if this.stripeYearlyAmount}}` line.

Sidenote: The improper indenting here makes the conditionals harder to follow 😓

ghost/admin/app/templates/offer.hbs
```diff
@@ -47,18 +47,14 @@
...
-                        {{#if (feature 'freeTrial')}}
                         <h4 class="gh-main-section-header gh-main-section-header-with-info small bn">
                             <span>Discount information</span>
                             <span class="midgrey gh-main-section-header-info tooltip-left" data-tooltip="Once saved, this cannot be changed" >{{svg-jar "info"}}</span>
                         </h4>
-                        {{else}}
                         <h4 class="gh-main-section-header small bn">
                             Discount information <span class="midgrey"> &ndash; Once saved, this cannot be changed.</span>
                         </h4>
-                        {{/if}}
```
kevinansfield, line 59: This whole block needs removing, right now discount info is duplicated.

kevinansfield (changes requested):

PR #15771, opened 2022-11-07
reauthenticate back to the sign in modal

veriperi:
```
          // we should see a re-auth modal
            expect(findAll('.fullscreen-modal #login').length, 'modal exists').to.equal(1);
            expect(findAll('.fullscreen-modal #login').length, 'modal exists').to.equal(0);
        });

```

https://careers.ghost.org


ErisDS:
Hey @veriperi I'm not sure what you're trying to do here?

PR #15774, opened 2022-11-07
Moved all code for feedback chart to one file

lenabaidakova:
closes TryGhost/Team#2143
- All logic for feedback pie chart was spread through multiple files. It would be difficult to scale it. Now it is encapsulated in one file.


ghost/admin/app/components/posts/feedback-events-chart.js
```diff
@@ -1,22 +1,35 @@
...
+    @action
+    onTooltipInsert(node) {
+        this.tooltipNode = node;
+    }
+
+    @action
+    onMouseleave() {
+        this.tooltipNode.style.opacity = '0';
+        this.tooltipNode.style.position = 'fixed';
+        this.tooltipNode.style.left = '2000';
```
SimonBackx, line 20: Some remarks here:
- Can't we only set opacity to 0 and leave out the rest? (I think setting left to '2000' is invalid because it is lacking a unit). I think displaying the tooltip at some position outside the window can cause scrolling issues on some devices.
- Don't we need to also disable pointer events or set display to none to avoid interaction when it is hidden? Setting opacity to 0 alone doesn't disable interaction.

lenabaidakova: Changed to `display: 'none'`

SimonBackx (approved):

PR #15784, opened 2022-11-08
Cleaned up member filters

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2134
fixes https://github.com/TryGhost/Team/issues/2133

- Moved all filters to separate files to make the filter component a lot more readable and easier to maintain.
- Removed long switch style code from hbs for filter column values
- Filters for features that are disabled (such as open tracking, click tracking or member attribution) are now hidden when they are disabled
- The open rate column in the members table is now only visible if open tracking is enabled


SimonBackx:
@kevinansfield Do you think this folder structure and way of importing the filters is okay, or do you prefer a different style?

ghost/admin/app/utils/member-filters/subscription-attribution.js
```diff
@@ -0,0 +1,21 @@
+import {MATCH_RELATION_OPTIONS} from './relation-options';
+
+export const SUBSCRIPTION_ATTRIBUTION_FILTER = {
+    label: 'Subscription started on post/page', 
+    name: 'conversion', 
+    valueType: 'string', 
+    resource: 'post', 
+    feature: 'memberAttribution',
```
SimonBackx, line 8: Todo: remove this feature flag, and replace it with settings checks to enable/disable certain filters if certain features (not flags) are disabled

kevinansfield:
> @kevinansfield Do you think this folder structure and way of importing the filters is okay, or do you prefer a different style?

The "util" files are only used by the member filter component so I'd keep them colocated in the `components/members/filter` directory.

ghost/admin/app/components/members/list-item.hbs
```diff
@@ -29,15 +29,13 @@
...
-                {{/if}}
-            </LinkTo>
-        {{/if}}
+        <LinkTo @route="member" @model={{@member}} class="gh-list-data middarkgrey f8 {{unless @member.name "gh-members-list-open-rate-noname"}}" data-test-table-data="open-rate">
+            {{#if (not (is-empty @member.emailOpenRate))}}
+                <span>{{@member.emailOpenRate}}%</span>
+            {{else}}
+                <span class="midlightgrey">N/A</span>
+            {{/if}}
+        </LinkTo>
```
kevinansfield, line 38: Is the removal of the feature flag here intentional? `emailAnalytics` was a special-case feature because it wasn't a labs flag but a user-selectable option. The original idea was that all email analytics UI is hidden when email analytics is disabled but maybe that's changed now? I'm not seeing `emailAnalytics` really used anywhere else so it might just be we need another cleanup task for it.

kevinansfield (approved):

ghost/admin/app/components/members/list-item.hbs
SimonBackx: Hmmm, the `Open rate` header in the table doesn't have this check. And I was seeing this column locally so that got me confused I think. I had the impression `feature "emailAnalytics"` always returned true. Gonna check what is happening here, brb

SimonBackx: So the emailAnalytics feature isn't anymore, but it now maps to config.emailAnalytics, which is true by default:
https://github.com/SimonBackx/Ghost/blob/641c87c5d744997788e61b2b5573b4aae7ec68b3/ghost/core/core/shared/config/defaults.json#L158

I'm gonna create a cleanup issue to replace it with`settings.emailTrackOpens` || `settings.emailTrackClicks` + check if we still need it in the backend. I think the config is still used for enabling/disabling the background job

SimonBackx: I think on this place it is good to remove it, because otherwise it wouldn't add the column in the rows when it did add the column header in the table.

PR #15789, opened 2022-11-09
🐛 Fixed link click counts for duplicate links

SimonBackx:
refs https://github.com/TryGhost/Team/issues/2213

When an email contains multiple links to the same destination, that link is only shown once on the analytics page. The total displayed count is not summed correctly.


rshbhgrg (approved):
Nice catch, that could explain some of the undercounting on Admin.

PR #15792, opened 2022-11-09
🐛 Fixed offer links with an archived tier

SimonBackx:
refs https://github.com/TryGhost/Team/issues/2233

**Problem**
When a user clicks an offer link that has an archived tier, the site blocks and you are no longer able to scroll. This is because the product for that offer can't be found. This has been fixed by updating the `isActiveOffer` helper to also check for the existence of the corresponding tier.

**Solution**
- You no longer are able to create new offers if there are no active tiers
- A custom message is shown that instructs the user to create a new tier if there are not active tiers on the offers page
- Improved detection of changes in tiers by correctly reloading the members utils service after tier changes
- Portal redirects to the homepage for offers with an archived tier (same behaviour as invalid offers)
- Offers of an archived tier are no longer visible in the dashboard


rshbhgrg (approved):
Good stuff! 🎉

PR #15794, opened 2022-11-09
🐛 Fixed pasting newlines in post titles

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2193

- When pasting a title with a newline, we now trim the string and clear newslines before pasting.
- When sending the slug to the backend to generate a unique slug, we now sluggify it in the frontend before adding it to the URL to prevent issues with unsupported characters (causing possible routing problems in Pro).


kevinansfield (approved):
👍 fixes the underlying bug, I think it would still be good to apply a trim and newline removal on any pastes to the title though as it shouldn't be possible to have newlines in the title (UI doesn't allow them to be created manually and the front-end will ignore them)

SimonBackx:
@kevinansfield I added a trim and cleanup on title paste. I think we only want to do this on paste to prevent other issues (we could also do it on blur, but not 100% sure about that).

kevinansfield (approved):

PR #15796, opened 2022-11-10
🐛 Fixed visible canceled events in conversions tab on analytics page

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2238

**Issue**
When viewing the 'conversions' tab on the analytics page, you could sometimes see more listed events than the total number of conversions. This is because other subscription events are also shown in the list. E.g., if a new member became a paid subscriber that is attributed to a given post, and later that subscrption has been canceled, that canceled event would also be shown on the analytics page. This isn't really desirable.

**Fix**
Now only 'created' subscription events are shown when the activity feed is filtered by post_id. The other subtypes aren't related to that given post and should be excluded.

### DIFF
--- a/ghost/members-api/lib/repositories/event.js
+++ b/ghost/members-api/lib/repositories/event.js
@@ -170,38 +170,6 @@ module.exports = class EventRepository {
     }
 
     async getSubscriptionEvents(options = {}, filter) {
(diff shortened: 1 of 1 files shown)


lenabaidakova (approved):
:shipit:

PR #15798, opened 2022-11-10
Reduced default post relations

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2085

Don't load relations we don't need anymore for the posts table. And reload the individual post when we open the analytics page with more relations that we actually need.


lenabaidakova (approved):
🙌

PR #15800, opened 2022-11-10
Made content importer work as a job

sam-lord:
refs:  https://github.com/TryGhost/Toolbox/issues/431

Wrapped the current content importer function as a job, repurposing the members importer email to notify the user who initiates the content import when it is complete.

The API is changed a little to model after the members CSV import also - with a 201 accepted response if jobs are accepted, and a 202 if the job is processed inline and new data is created.

---

Main thing left on this PR before merging is that the db regression tests use the content import API, but would need to pass `forceInline` in order for it to have the same behaviour as before. I didn't want to expose this to the public API, but I think it would need to go there in order for those tests to work - unless we mocked out a property on the import manager 🤔


ghost/core/core/server/data/importer/import-manager.js
```diff
@@ -351,45 +335,132 @@ class ImportManager {
...
     /**
      * Import From File
      * The main method of the ImportManager, call this to kick everything off!
      * @param {File} file
      * @param {ImportOptions} importOptions to allow override of certain import features such as locking a user
-     * @returns {Promise<ImportResult[]>}
+     * @returns {Promise<Object.<string, ImportResult>>}
      */
     async importFromFile(file, importOptions = {}) {
+        if (!importOptions.forceInline && !importOptions.runningInJob) {
```
naz, line 404: I think instead of "forceInline" we could use the pattern of checking for test environment `env?.startsWith('testing')` (l[ike done here](https://github.com/TryGhost/Ghost/blob/9e35c4dd7a8c1904a5162bc064db2b3e04757ef5/ghost/core/core/server/services/members/service.js#L150)). No need to expose extra parameters in the API if it does not have a valid usecase.

sam-lord: > I think instead of "forceInline" we could use the pattern of checking for test environment `env?.startsWith('testing')` (l[ike done here](https://github.com/TryGhost/Ghost/blob/9e35c4dd7a8c1904a5162bc064db2b3e04757ef5/ghost/core/core/server/services/members/service.js#L150)). No need to expose extra parameters in the API if it does not have a valid usecase.

Totally! I think I copied the `forceInline` out of the members importer but using env sounds perfect. Anything to get the tests passing :)

PR #15807, opened 2022-11-11
Minifier Windows support

### LINKED ISSUE #14278: Cannot generate assets (cards, admin-auth, etc) on local windows install
### Issue Summary

Installed ghost ( Node.js version - found v16.14.0 | npm: '8.3.1' | Ghost-CLI version: 1.18.1
| Ghost version: 4.38.0 ) on my Windows 10 machine. ``cards.min.css`` don't seem to load on my local dev site, but it does on the editor.

Editor:
image

Published:
image

I touched NOTHING. This is a fresh install, tried six times.

Here's the Package.json file, cards should show up.

image

Additionally, when I try to go to ``http://localhost:2368/public/cards.min.css`` on my browser, I get a 404 error.

Lastly, someone reported the same issue in 2021: [in this post](https://forum.ghost.org/t/css-and-js-for-cards-are-not-being-loaded-in-local-development-mode/27410), a guy reported the same issue I'm having. Sadly, he seems to have abandoned ghost because of it, because no more posts under his username were made and he didn't bother to fill a bug report.

### Steps to Reproduce

1. Install ghost through npm locally on Windows
2. Start Ghost
3. Add cards on any blog post
4. Publish 

### Ghost Version

4.38

### Node.js Version

16.14.0

Windows 10 (npm)

### Database type

SQLite3

### Browser & OS version

Chrome, Firefox, Windows 10

```shell
{"name":"Log","hostname":"DESKTOP","pid":3668,"level":30,"req":{"meta":{"requestId":"34727ba4-f918-4ee5-aaf7-696ed49b2390","userId":null},"url":"/public/cards.min.css","method":"GET","originalUrl":"/public/cards.min.css","params":{},"headers":{"host":"localhost:2368","user-agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:97.0) Gecko/20100101 Firefox/97.0","accept":"text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8","accept-language":"en-US,en;q=0.5","accept-encoding":"gzip, deflate","connection":"keep-alive","cookie":"**REDACTED**","upgrade-insecure-requests":"1","sec-fetch-dest":"document","sec-fetch-mode":"navigate","sec-fetch-site":"none","sec-fetch-user":"?1"},"query":{}},"res":{"_headers":{"x-powered-by":"Express","cache-control":"no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0","content-type":"text/html; charset=utf-8","etag":"W/\"3207-66Cm6wMvq5dLdcDIxd1CmDjHrEc\"","vary":"Accept-Encoding","content-encoding":"gzip"},"statusCode":404,"responseTime":"40ms"},"msg":"","time":"2022-03-07T20:07:44.546Z","v":0}
```

### Code of Conduct

9larsons:
Resolves [#14278](https://github.com/TryGhost/Ghost/issues/14278).

Change Minifier to use node-glob instead of tiny-glob for file matching, as node-glob has Windows support. There was no apparent performance hit with this change.

### DIFF
--- a/ghost/core/core/server/services/mail/GhostMailer.js
+++ b/ghost/core/core/server/services/mail/GhostMailer.js
@@ -96,6 +96,7 @@ module.exports = class GhostMailer {
      * @param {string} message.subject - email subject
      * @param {string} message.html - email content
      * @param {string} message.to - email recipient address
+     * @param {string} [message.replyTo]
      * @param {string} [message.from] - sender email address
      * @param {string} [message.text] - text version of this message
      * @param {boolean} [message.forceTextContent] - maps to generateTextFromHTML nodemailer option


ErisDS (changes requested):
Hey @9larsons thanks for the PR 🙂 As it's your first I had to hit the approve button before CI ran, and as you can see there's a lint error and some unit tests are failing. 

Would you be able to take a look at resolving these issues?

9larsons:
Hm, odd. I ran those both locally and had no issues before submitting. I'll definitely take a look.

9larsons:
It seems Github's environment results in different results. Running my branch locally, I have passes for both:

image

image

Lint within Github is catching the missing semicolon while locally it is not. Any idea why the linter results would differ?

The other question is less clear... It seems the glob path is pulling the full path instead of the relative for the Minifier test failures in the Github tests. Github is running 'yarn workspaces run test:unit' vs. the developer instructions (and provided scripts) of just 'yarn test:unit'.

It seems both of these are the result of using workspaces. What should I be running to be correct?

ErisDS:
The commands are correct, the issue here is with where you are running them. 

Ghost is a monorepo with multiple packages inside `/ghost`. You're running commands inside `ghost/core` which is the main package / where the bulk of the application is, but running a `yarn` command here is scoped to that package.

The changes you've made live in `ghost/minifier`. Running the commands in that package will scope them again.

If you run the commands from the top level, they'll run across all packages.

9larsons:
Right, that makes sense. I guess I'm still thinking of the old client/server split.

I did initially try from the top level and was only able to make yarn test work when within the core package. Basically, I'm running in to a fair number of issues on Windows and I'm over time sorting them out... I will say, if you do any of this strictly on Linux, it all works great. There's ~17 failing Windows unit tests within ghost/core and I also cannot run yarn test from the parent/monorepo dir as there's a failure with another error.

I've started looking at resolving some of those failings Windows tests, but it seems like it's more trouble than just that.

At any rate, I'll get this resolved when I can. Thanks for the info.

ErisDS:
For the purposes of these changes, you only need to get it working in `ghost/minifier` - but you'll prob need to test on windows + linux (even if that's just using CI) to make sure the tests are robust across OSes.

Doesn't windows have linux built in now anyway for this exact reason? 😬

9larsons:
Yep, I did check on both Windows & Linux since it does need to work on both, but I wasn't previously at the right level for the tests to capture `ghost/minifier` as you mentioned.

Windows still makes you install Windows Subsystem for Linux. I've recently switched to that over an Ubuntu VM and it seems to work fairly well.

Frankly, I'm ready to scrap ever doing dev work on Windows 🙃.

9larsons:
Sorry for the delay, busy week. It should be good now. I actually found a minifier unit test that didn't appear to be checking what it was intending to, and updated that. I think the two main callouts are:

1) node-glob always returns a Unix filepath, so I had to manually handle that and called it out in the card-assets tests as path.join would give incorrect results for Windows users.

2) I made the decision to pare back the error handling in minifier to no longer require a return value. It's debatable that you'd want to throw an error when given a glob or filepath that yields no file(s). I opted to not because the theme engine by default passes *.js and *.css to minifier, and if for some reason a theme did not have one of those, it seems odd to error when we're just trying to be helpful in collecting all files.

9larsons:
No clue off the bat why the post test would be failing, as this didn't touch that code. 

I'm planning to get the MacBook set up tonight and finally put this to rest. I was working on this still using Win10 and WSL in parallel, and I'm learning to distrust WSL.

ghost/minifier/lib/minifier.js
```diff
@@ -69,7 +65,7 @@ class Minifier {
     }
 
     async getMatchingFiles(src) {
-        return await glob(this.getFullSrc(src));
+        return glob.sync(this.getFullSrc(src));
```
ErisDS, line 68: I know you said you didn't see a perf impact, but I am still _slightly_ nervous about this change. As far as I know the change to add this feature reduced boot performance significantly.

Tiny-glob claims to be a lot faster than node-glob but then it also claims to be have windows support (which is interesting).

Anyway all things considered I'd suggest we at least keep this async so that it doesn't block the boot process:

https://github.com/isaacs/node-glob/issues/420#issuecomment-698251894

9larsons:
I've been out most of the past week. I'll take another pass at this here in the next few days.

I previously did some local testing with tiny-glob and wasn't getting any of the expected output. Both packages make the same claim about specifying globs with a forward slash regardless of platform, though only node-glob had a successful output for me on Windows. https://github.com/terkelg/tiny-glob/issues/40 may have something to do with it. 

When you say 'the change to add this feature' are you talking about adding the minifier? Or did you do some testing with this branch and looked at boot times? From what I saw, even if it is ~350% faster as it claims, it's all in the sub-millisecond territory.

ErisDS:
By "the change to add this feature" I mean when I originally implemented the card asset pipeline. We have metrics for boot-time on Pro and they were noticeably impacted.

9larsons:
Sorry, I synced my branch and it automatically closed this off. I'll open a new PR as I'm not sure if reopening this or opening a new one is easier for you. At any rate, after more investigation I found a solution we can both be happier with.

PR #15817, opened 2022-11-14
Update the cover image in default fixtures

minimaluminium:
We're starting to ship Casper without default cover image, so that the header section will have accent color as background by default. It's only meant for new sites, and shouldn't change anything on existing sites.


ErisDS:
This is failing because we have an integrity check to ensure that accidental changes are not made to the fixtures. These changes won't apply to any existing site as there is no accompanying migration, but that is expected in this case.

I will fix the integrity check and merge.

PR #15818, opened 2022-11-14
Included e2e tests in coverage reports

naz:
closes https://github.com/TryGhost/Toolbox/issues/475


daniellockyer:
@naz It might be worth reviewing https://github.com/TryGhost/Ghost/pull/15818/files#diff-f6b7e5e019ee108f2ee81520491c526f7969ff8837f1c47b62c4a69a0215bfd9R23-R30 too

naz:
@daniellockyer the link is now broken (possibly linked to something pre-force-push) which lines did you mean exactly?

daniellockyer:
@naz Whoops! These ones: https://github.com/TryGhost/Ghost/blob/17932759d3f91d19afa219a9287c7fc72f62f03a/ghost/core/.c8rc.json#L21-L30

They currently exclude coverage for the API controllers and Express app routing

naz:
Ah yes, good spot!

naz:
I think given the commands `yarn test:unit` and `yarn test:ci` are run separately and have quite different percentages/files of coverage it would make sense to have two separate configs. Unless we'd include the unit tests run as a part of "test:ci" command? Was the reason we run these as two separate commands to get a faster feedback loop @daniellockyer?

daniellockyer:
@naz To be honest the commands are a bit of a mess developed over the years to meet local development + CI requirements - having your current implementation is probably the best right now

PR #15823, opened 2022-11-14
Limited Stripe price lookups

SimonBackx:
refs https://github.com/TryGhost/Team/issues/2262

Makes sure we only loop active Stripe prices. If we find an inactive price, we also update it in our database now after this change.

### DIFF
--- a/ghost/payments/lib/payments.js
+++ b/ghost/payments/lib/payments.js
@@ -215,7 +215,7 @@ class PaymentsService {
             currency,
             interval: cadence,
             amount,
-            active: 1
+            active: true
         }).query().select('id', 'stripe_price_id');
 
         for (const row of rows) {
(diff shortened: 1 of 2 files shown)


ghost/payments/lib/payments.js
```diff
@@ -214,8 +214,9 @@ class PaymentsService {
             stripe_product_id: product.id,
             currency,
             interval: cadence,
-            amount
-        }).query().select('stripe_price_id');
+            amount,
+            active: 1
```
rishabhgrg, line 218: @SimonBackx Gotten a bit confused here but should this be `true` or `1` here ?

SimonBackx: Wasn't sure about the knex/bookshelf boolean mapping. But if you think true also works, then it should be fine 😊
```suggestion
            active: true
```

rishabhgrg: Yea think we use `true` usually but even i am confused tbh which is right/ideal. If `active:true` works locally then lets go with that as that is a better representation

rshbhgrg (approved):
LGTM, just a minor comment about checking/changing the `active` to `true` :)

PR #15824, opened 2022-11-14
Add complaint events to activity feed

lenabaidakova:
closes TryGhost/Team#2248

- Added ability to show spam emails in member activity feed


ghost/core/core/server/services/email-analytics/lib/event-processor.js
```diff
@@ -135,7 +135,21 @@ class GhostEventProcessor extends EventProcessor {
     }
 
     async handleComplained(event) {
-        return this._unsubscribeFromNewsletters(event);
+        const emailId = await this.getEmailId(event);
+        let updateComplaint = false;
+
+        if (emailId) {
```
lenabaidakova, line 141: @allouis , do we need to hide it behind feature flag? Or it's okay to leave it as is? So, we'll collect info, but won't show it in UI.

lenabaidakova:
> This is looking good @lenabaidakova and appears to work as expected!
> 
> I was curious about the "Unsubscribed from Ghost Inc." event that appears right after. Is that something we added or is it just a natural side effect of what happens when the email gets suppressed?
> 
> CleanShot 2022-11-16 at 12 17 25@2x

@moreofmorris , we added it previously. Member is unsubscribed from newsletters when we receive spam event from MG

lenabaidakova:
https://github.com/TryGhost/Team/issues/2248#issuecomment-1318175125

Issue #14020, opened 2022-01-20
AMP YouTube video looks too small 


rubalv96:
### Issue Summary

When you insert a YouTube video into a post and view it with AMP the video looks very small. The <amp-youtube> element that Ghost generates contains fixed width and height sizes.
There is no solution to view those YouTube videos with a proper size in the AMP version.

![image](https://user-images.githubusercontent.com/36295331/150339812-3697bfb6-e909-4910-ad0e-7a7ce3cad4c3.png)

### Steps to Reproduce

1. Insert a YouTube video in a post.
2. View this post in the AMP version
3. You will see that the inserted video is displayed at a very small size.

### Ghost Version

4.3.3

### Node.js Version

v14.17.1

### How did you install Ghost?

Ubuntu 18.04.5 LTS

### Database type

SQLite3

### Browser & OS version

Google Chrome Version 96.0.4664.45 (Official Build) (64-bit)

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

ErisDS:
Thanks for the bug report. We'd love a Pr to fix this :)

denik1981:
Seems the problem is that Amp is trying to parse an already defined iframe with fixed set widths and heights. And to be non-obtrusive I believe it is extracting those and using them as their own attributes giving the wrong expected behavior of having a tiny video. 

The width and height are already present from the model and I don't think they are necessary because in non-amp pages there is a process which is sanitizing videos giving them a responsive behavior (check jquery.fitvids.js).

With my lack of knowledge about what ghost does, I cannot say very much, but I suspect the place on where to check for a fix is within the Koening editor which is IMHO rendering the HTML iframes that we are seeing in the persistent model. Check for _popupateIframe function and see if it is possible to remove the width and height attributes from there.

sahyrlelyaams:
Wanted to try my hand at fixing this...
Any pointers on how to view the post in AMP?

matthanley:
@sahyrlelyaams you can add `/amp` to the end of the URL for a published post to view the AMP rendered version.

PR #15826, opened 2022-11-15
🐛 Fixed amp-youtube being too small

jacobsimon:
## Summary

This is one potential solution to issue #14020 that I'd like to suggest. The impact of this bug is that it is difficult to view YouTube videos in blog posts via Google search, because the video is too small to view and the "full-screen" button is not clickable at this size. ([See example here](https://blog.tryexponent.com/top-aws-interview-questions/amp/))

The solution proposed here is to add CSS for `amp-youtube` to set the dimensions of the iframe instead of passing explicit height and width attributes. Instead we set the video width to the viewport width and then set the height to a standard 16:9 ratio of YouTube videos.

I acknowledge this is likely not the "best" solution and would definitely be open to other ideas! 

## Screenshots

| Before | After | 
| --- | --- |
| Screen Shot 2022-11-14 at 6 46 44 PM | Screen Shot 2022-11-14 at 6 46 37 PM

## Checklist


ErisDS:
I'm happy to merge this change as an improvement for now. It's worth noting though that Google AMP is going away, and so we will remove it in Ghost 6.0.

PR #15828, opened 2022-11-15
🐛 Fixed error when importing members (old)

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2266

- The getCreatedEvents method was removed but was still used in the verification trigger
- Added regression test that covers this in the future


SimonBackx:
Moved to https://github.com/TryGhost/Ghost/pull/15831

PR #15831, opened 2022-11-15
🐛 Fixed error when importing members

SimonBackx:
fixes TryGhost/Team#2266

The getCreatedEvents method was removed but was still used in the verification trigger


SimonBackx:
Closed in favour of https://github.com/TryGhost/Ghost/pull/15832

PR #15832, opened 2022-11-15
Removed getCreatedEvents and added verification trigger test

SimonBackx:
refs https://github.com/TryGhost/Team/issues/2266

This removes the deprecated `getCreatedEvents` method in the event repository and adds tests to the verification trigger to see if we don't break anything.

Changes extracted from https://github.com/TryGhost/Ghost/pull/15831

### DIFF
--- a/ghost/core/core/server/services/members/service.js
+++ b/ghost/core/core/server/services/members/service.js
@@ -74,6 +74,36 @@ const processImport = async (options) => {
     return result;
 };
 
+const updateVerificationTrigger = () => {
+    verificationTrigger = new VerificationTrigger({
+        apiTriggerThreshold: _.get(config.get('hostSettings'), 'emailVerification.apiThreshold'),
+        adminTriggerThreshold: _.get(config.get('hostSettings'), 'emailVerification.adminThreshold'),
+        importTriggerThreshold: _.get(config.get('hostSettings'), 'emailVerification.importThreshold'),
+        isVerified: () => config.get('hostSettings:emailVerification:verified') === true,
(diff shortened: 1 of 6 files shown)


ghost/core/core/server/services/members/service.js
```diff
@@ -74,6 +74,36 @@ const processImport = async (options) => {
     return result;
 };
 
+const updateVerificationTrigger = () => {
```
SimonBackx, line 77: This changes was required to make testing easier because  the `VerificationTrigger` is not updated after config changes.

ghost/verification-trigger/lib/verification-trigger.js
```diff
@@ -19,7 +19,7 @@ class VerificationTrigger {
      * @param {number} deps.importTriggerThreshold Threshold for triggering Import sourced verifications
      * @param {() => boolean} deps.isVerified Check Ghost config to see if we are already verified
      * @param {() => boolean} deps.isVerificationRequired Check Ghost settings to see whether verification has been requested
-     * @param {(content: {subject: string, message: string, amountTriggered: number}) => void} deps.sendVerificationEmail Sends an email to the escalation address to confirm that customer needs to be verified
+     * @param {(content: {subject: string, message: string, amountTriggered: number}) => Promise<void>} deps.sendVerificationEmail Sends an email to the escalation address to confirm that customer needs to be verified
```
SimonBackx, line 22: Should be a promise so we can properly await sending an email. If we don't want to await, that decision should be made inside the verification trigger class and not elsewhere. But for testing awaiting is easier and doesn't require a hacky timeout.

Issue #15814, opened 2022-11-12
No tier when importing on csv files


satoshinotdead:
### Issue Summary

There is no selector for tiers when importing csv previously exported by Ghost.

https://forum.ghost.org/t/select-complementary-tier-when-importing-members/26659

### Steps to Reproduce

1. Export some user with tier.
2. Import the same file previously exported.

### Ghost Version

5.22.4

### Node.js Version

16.18.0

### How did you install Ghost?

Debian Buster

### Database type

MySQL 8

### Browser & OS version

Firefox on Debian

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

PR #15835, opened 2022-11-16
Clarified "Complimentary Plan" in the importer

naz:
refs https://github.com/TryGhost/Ghost/issues/15814

- There's often a confusion around the logic of what makes "Complimentary Plan". It's especially confusing during the import process, where a bulk of members gets assigned to complimentary plan. This is a "quick and dirty" fix for the problem, something more holistic will be worked out in the future


naz:
This might get picked up at some point when we have time. Closing for now

PR #15851, opened 2022-11-19
fix: missing "the" in sentence regarding trial

neverything:

There seems to be a missing "the" in the trial message on the signup page of the portal. The senctence just sounds a bit off without it.


ErisDS:
Thank you for the suggestion 🙂

neverything:
Thank you for reviewing and merging it, my client is going to be very happy.

PR #15855, opened 2022-11-21
Added source and source type columns to emails table

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2280

- we are moving away from storing raw html and plaintext directly on email in favor of storing mobiledoc/lexical format which can be used to translate the final html for the email
- adds new columns `source` and `source_type` to store `mobiledoc`/`lexical` information for an email depending on the format used for the post
- `source_type` can be one of `html`(for existing emails) or `mobiledoc` / `lexical` (for new emails)
- adds a migration to copy existing `html` data on `emails` to `source` so the new email flow can work with it, and we can later drop the `html` column
- we default to `source_type` as `html` as existing emails have `html` stored for them


ghost/core/core/server/data/migrations/versions/5.24/2022-11-21-15-03-populate-source-column-with-html-for-emails.js
```diff
@@ -0,0 +1,16 @@
...
+        await knex('emails')
+            .update({
+                source: knex.ref('html')
+            });
+    },
+
+    async function down() {
+        // no-op: we don't want to remove the data
+    }
+);
```
SimonBackx, line 16: Maybe we need to delay this migration and only merge it when we make the flag GA, otherwise new emails will still have an empty source. Unless we alter MEGA to also set source and source_type

rishabhgrg: I was thinking of updating mega [here](https://github.com/TryGhost/Ghost/blob/955605572ef5deb0330b201b4fb2558adcd1ed4d/ghost/core/core/server/services/mega/mega.js#L233-L248) to add `source` value as the same html so we don't have to do this again.
The reason i was thinking of getting this early and not wait for GA is that it'll allow us to easily test all existing functionality using `source` / `source_type` and by the end of GA we just need to remove other columns, wdyt ?

SimonBackx: Yes, that is a good solution 👌

ghost/core/core/server/data/migrations/versions/5.24/2022-11-21-15-03-populate-source-column-with-html-for-emails.js
```diff
@@ -0,0 +1,16 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Populating source from html in emails table');
+        await knex('emails')
```
daniellockyer, line 7: Can you get the number of affected rows returned from this `.update` call, and log it out for visibility? 🙂

rishabhgrg: Done, thanks

daniellockyer (approved):

SimonBackx (approved):

PR #15857, opened 2022-11-21
Fix scriptUrl code

touzoku:
`scriptUrl` can be a boolean (if set to `false`), but the code assumes it is a string and crashes on the `String.prototype.includes()` call.

This PR aligns the code with the line below, which does `stylesUrl` correctly.

P.S. You should really start migrating to typescript to avoid such kind of bugs and make the life of OSS contributors easier. At least for the framework (storage plugins etc.) code.

### DIFF
--- a/ghost/core/core/frontend/utils/frontend-apps.js
+++ b/ghost/core/core/frontend/utils/frontend-apps.js
@@ -4,7 +4,7 @@ function getFrontendAppConfig(app) {
     const appVersion = config.get(`${app}:version`);
     let scriptUrl = config.get(`${app}:url`);
     let stylesUrl = config.get(`${app}:styles`);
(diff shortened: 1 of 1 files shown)


touzoku:
This PR fixes this issue: https://forum.ghost.org/t/is-it-possible-to-disable-jsdelivr-cdn/31761

aq1:
Why hasn't it been merged yet? I need this!

daniellockyer:
Hi @touzoku - I couldn't get your PR to solve the problem. The statement eventually evaluates to `> false?.includes('{version}')` which results in `Uncaught TypeError: false?.includes is not a function`

I've gone in a different direction to solve this [here](https://github.com/TryGhost/Ghost/pull/17114), but thank you for opening the PR 🙂

PR #15860, opened 2022-11-21
Added table to store email recipient failures

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2291

When sending out mails to individual recipients, its possible that recipient gets a temporary or permanent failure for receiving the mail. Temporary failures can generally get resolved after a bit when the recipient’s mail server accepts the email, unlike permanent failures. For both customer visibility and easier debugging on what went wrong while delivering to a particular recipient, we’ll store the permanent/temporary failure for a recipient.

- migration adds a new table that stores the failure information for the recipients

### DIFF
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -835,6 +835,24 @@ module.exports = {
             ['email_id', 'member_email']
         ]
     },
+    email_recipient_failures: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        email_id: {type: 'string', maxlength: 24, nullable: false, references: 'emails.id'},
+        member_id: {type: 'string', maxlength: 24, nullable: true},
+        email_recipient_id: {type: 'string', maxlength: 24, nullable: false, references: 'email_recipients.id'},
+        code: {type: 'integer', nullable: false, unsigned: true},
+        enhanced_code: {type: 'string', maxlength: 50, nullable: true},
+        message: {type: 'string', maxlength: 2000, nullable: false},
+        severity: {
+            type: 'string',
+            maxlength: 50,
+            nullable: false,
+            defaultTo: 'permanent',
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/data/migrations/versions/5.24/2022-11-21-17-10-add-error-recipient-failures-table.js
```diff
@@ -0,0 +1,20 @@
+const utils = require('../../utils');
+
+module.exports = utils.addTable('offers', {
```
SimonBackx, line 3:
```suggestion
module.exports = utils.addTable('email_recipient_failures', {
```

ghost/core/core/server/data/schema/schema.js
```diff
@@ -824,6 +824,24 @@ module.exports = {
             ['email_id', 'member_email']
         ]
     },
+    email_recipient_failures: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        email_id: {type: 'string', maxlength: 24, nullable: false, references: 'emails.id'},
+        member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id'},
```
SimonBackx, line 830: I'm wondering whether we should delete email_recipient_failures if a member is deleted or not 🤔 We don't delete the email_recipient because it doesn't have a foreign key

rishabhgrg: 🤔 Hmm good point, tho it might be weird to have `email_recipient_failures` to be deleted but not `email_recipients` on deleting a member

rishabhgrg: As discussed, have made the `member_id` to be nullable so we can set it to `null` if a member is deleted, and also removed the reference for `member_id`

SimonBackx (approved):

daniellockyer (approved):

SimonBackx (approved):

PR #15865, opened 2022-11-22
Added new email batch sending service

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2284


rshbhgrg (approved):
We'll need to add some unit tests for email-service eventually, but overall lgtm

PR #15877, opened 2022-11-24
Added email renderer implementation draft

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2308

- Still has some missing pieces, but mostly works.
- Uses new handlebars template for emails
- When sending emails with the new email stability flag enabled, one test email is now sent via the default smtp ghost mailer.


rshbhgrg (approved):
Overall lgtm. The test coverage is just tiny bit under, and we should break down the merge into multiple commits, specially for the changes to GhostMailer/settings as separate to the ones we are adding new for email-serivice. 😄

PR #15879, opened 2022-11-24
Added new email event processor

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2310

This moves the processing of the events from the event-processor to a new email-event-processor in the email-service package.

- The `EmailEventProcessor` only translates events from providerId/emailId to their known emailId, memberId and recipientId, and dispatches the corresponding events.
- Since `EmailEventProcessor` runs in a separate worker thread, we can't listen for the dispatched events on the main thread. To accomplish this communication, the events dispatched from the `EmailEventProcessor` class are 'posted' via the postMessage method and redispatched on the main thread.
- A new `EmailEventStorage` class reacts to the email events and stores it in the database. This code mostly corresponds to the (now deleted) subclass of the old `EmailEventProcessor`
- Updating a members last_seen_at timestamp has moved to the lastSeenAtUpdater.
- Email events no longer store `ObjectID` because these are not encodable across threads via postMessage
- Includes new E2E tests that test the storage of all supported Mailgun events. Note that in these tests we run the processing on the main thread instead of on a separate thread (couldn't do this because stubbing is not possible across threads)

There are some missing pieces that will get added in later PRs (this PR focuses on porting the existing functionality):
- Handling temporary failures/bounces
- Capturing the error messages of bounce events


ghost/job-manager/lib/job-manager.js
```diff
@@ -109,6 +113,13 @@ class JobManager {
                         finished_at: new Date()
                     });
                 }
+            } else {
+                if (typeof message === 'object' && this.#domainEvents) {
+                    // Is this an event?
+                    if (message.event) {
+                        this.#domainEvents.ee.emit(message.event.type, message.event.data);
```
allouis, line 120: `ee` is a private member, so we shouldn't be accessing it here.

I understand the reasoning behind needing to access the raw event emitter here as the js objects aren't serialized with all the info.

It would be really cool if we could either remove the private restriction from the `ee` member, ideally we wouldn't do that though because that opens the surface area of our interface a huge amount, or if we could add another couple of methods.

Perhaps something like:
```typescript
type RawEventData {
  name: string,
  data: object
}
```

```typescript
getRawEventData(event: IEvent): RawEventData
dispatchRaw(rawEvent: RawEventData): void
```

We can even skip the whole `getRawEventData` if you want, but it might be a good idea to keep the whole logic of reading `consturctor.name` in that same class.

ghost/email-events/lib/EmailBouncedEvent.js
```diff
@@ -1,7 +1,3 @@
-/**
- * @typedef {import('bson-objectid').default} ObjectID
- */
-
 module.exports = class EmailBouncedEvent {
```
allouis, line 1: I've realized that these events don't confirm to the `IEvent` interface defined by `@tryghost/domain-events` which is my :b:

I think they can go in as-is, and we can either rework the events, or rework the `IEvent` interface to be more generic during cleanup - I've added a TODO to come back to this next week either way ☺️

allouis (commented):
Just one comment about using the private `ee` member!

ghost/job-manager/lib/job-manager.js
SimonBackx: Thanks, I agree. I've created a dispatchRaw method, but didn't add a new RawEventData interface to keep it a bit more readable. The `dispatch` is updated to call the `dispatchRaw` now

allouis: Awesome okay 👍

ghost/core/core/server/services/jobs/job-service.js
```diff
@@ -15,7 +16,9 @@ const errorHandler = (error, workerMeta) => {
 };
 
 const workerMessageHandler = ({name, message}) => {
-    logging.info(`Worker for job ${name} sent a message: ${message}`);
+    if (typeof message === 'string') {
+        logging.info(`Worker for job ${name} sent a message: ${message}`);
+    }
```
SimonBackx, line 21: This was added because jobs can also post objects (used for sending domain events)

ghost/email-service/lib/email-event-processor.js
```diff
@@ -0,0 +1,182 @@
...
+        }
+        return recipient;
+    }
+
+    /**
+     * @param {EmailIdentification} emailIdentification 
+     */
+    async handleTemporaryFailed(emailIdentification) {
+        const recipient = await this.getRecipient(emailIdentification);
+        // TODO: store and emit event
```
SimonBackx, line 73: Storing error information is for a future PR

ghost/domain-events/lib/DomainEvents.js
```diff
@@ -28,7 +28,7 @@ class DomainEvents {
      * @returns {void}
      */
     static subscribe(Event, handler) {
-        DomainEvents.ee.on(Event.name, handler);
```
allouis, line 31: What's the reasoning for switching this?

Previously we were able to pass these methods around as functions individually, but with this change we can't do that without creating a new function with the `this` context bound.

I don't think we're using that feature anywhere, but curious why we'd change this

SimonBackx: Oh, I didn't know we passed it around as functions. I can't find a reference to it. I thought using 'this' was cleaner than repeating the constructor name. Also not sure if it is a good pattern to not have 'this' bound, because that requires you to know the implementation of the method and makes it more dangerous to change the implementation later. I'll revert this change for now, to be sure we don't break anything

rshbhgrg (approved):

PR #15895, opened 2022-11-29
Bumped GHA Runners to larger machines

daniellockyer:
wip


daniellockyer:
This doesn't seem to be faster than our current machines ☹️

PR #15896, opened 2022-11-29
Added wiring for mailgun provider

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2309

- adds new mailgun provider to send out batch emails


ghost/email-service/lib/mailgun-email-provider.js
```diff
@@ -0,0 +1,168 @@
...
+        try {
+            const messageData = {
+                subject,
+                html,
+                plaintext,
+                from,
+                replyTo,
+                id,
+                track_opens: !!options.openTrackingEnabled,
+                track_clicks: !!options.clickTrackingEnabled
```
SimonBackx, line 116: This should always be false because we do our own link tracking

SimonBackx: Maybe we can just set `clickTrackingEnabled` to false in the sending service. Because the email renderer is adding the click tracking

ghost/email-service/lib/mailgun-email-provider.js
```diff
@@ -0,0 +1,168 @@
...
+            };
+
+            // create recipient data for Mailgun using replacement definitions
+            const recipientData = recipients.reduce((acc, recipient) => {
+                acc[recipient.email] = this.#createRecipientData(recipient.replacements);
+                return acc;
+            }, {});
+
+            // update content to use Mailgun variable syntax for all replacements
+            ['html', 'plaintext', 'subject'].forEach((key) => {
```
SimonBackx, line 126: Cool, so we could use 'smart' variables inside subjects too 😊 Would need an update in the email renderer, but otherwise this could be supported

ghost/email-service/lib/mailgun-email-provider.js
rishabhgrg: Yeah i think this should come from outside of provider, because provider shouldn't have any of our app logic in it but works with the option provided. So as you said we can set `clickTrackingEnabled` as false in sending service which will automatically come down :)

ghost/email-service/lib/mailgun-email-provider.js
rishabhgrg: So I was confused about `subject` originally, but I saw we allow subject to be transformed [here](https://github.com/TryGhost/Ghost/blob/9fe1274fbe23ec40cb88dc9b5a01e5415fee3f4b/ghost/mailgun-client/lib/mailgun-client.js#L49) which is why I kept it here. Tho we don't add the `subject` to formats [here](https://github.com/TryGhost/Ghost/blob/65d910f43b3e2962fffe3acbd0d8c964401ffbbb/ghost/core/core/server/services/mega/post-email-serializer.js#L191) so it isn't getting used right now, but i think it can potentially be supported.

ghost/email-service/lib/batch-sending-service.js
```diff
@@ -286,6 +286,7 @@ class BatchSendingService {
         try {
             const members = await this.getBatchMembers(batch.id);
             const response = await this.#sendingService.send({
+                emailId: email.id,
```
rishabhgrg, line 289: Allows adding reference to email id in mailgun, as currently added in mailgun client [here](https://github.com/TryGhost/Ghost/blob/80decf83031acdab9916d758eaddde08840ba2e8/ghost/mailgun-client/lib/mailgun-client.js#L69-L72)

ghost/email-service/lib/batch-sending-service.js
```diff
@@ -297,18 +298,21 @@ class BatchSendingService {
 
             await batch.save({
                 status: 'submitted',
-                provider_id: response.id
+                provider_id: response.id,
+                // reset error fields when sending succeeds
+                error_status_code: null,
+                error_message: null,
+                error_data: null
```
rishabhgrg, line 305: Resets error fields for email if the mail succeeds

ghost/email-service/lib/mailgun-email-provider.js
```diff
@@ -0,0 +1,168 @@
...
+            debug(`sent message (${Date.now() - startTime}ms)`);
+            logging.info(`Sent message (${Date.now() - startTime}ms)`);
+
+            // Return mailgun provider id, trim <> from response
+            return {
+                id: response.id.trim().replace(/^<|>$/g, '')
+            };
+        } catch ({error, messageData}) {
+            // REF: possible mailgun errors https://documentation.mailgun.com/en/latest/api-intro.html#status-codes
+            let ghostError = new errors.EmailError({
```
rishabhgrg, line 149: @SimonBackx Have repurposed existing `EmailError` here for passing the error instead of creating new `EmailProviderError` as it seemed unnecessary, but i am not 100% on that, wdyt ?

ghost/email-service/lib/mailgun-email-provider.js
SimonBackx: Ah, yes, but in the past the format was never set to subject

ghost/email-service/lib/mailgun-email-provider.js
SimonBackx: Should be fine 👌

SimonBackx (approved):

ghost/email-service/lib/mailgun-email-provider.js
rishabhgrg: Yea exactly, should we remove the `subject` from this list for now then ? Don't want accidentaly coverting a {{name}} in subject to the member name i guess right

PR #15898, opened 2022-11-29
Updated the filter naming in Post Analytics

lenabaidakova:
closes TryGhost/Team#2329
- Replace 'Received' on 'Sent' in member's filter
- Moved links for feedback analytics from chart to table

### DIFF
--- a/ghost/admin/app/components/members/filter.js
+++ b/ghost/admin/app/components/members/filter.js
@@ -1,7 +1,7 @@
 import Component from '@glimmer/component';
 import moment from 'moment-timezone';
 import nql from '@tryghost/nql-lang';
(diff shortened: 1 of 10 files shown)


moreofmorris:
This is looking good so far, I think these changes make sense seeing them in practice. Curious what others will think, too!

I did notice a few bits... If there is only More Like This or only Less Like This available, then we maybe only show that one so people aren't clicking through to no members in the filtered list.
CleanShot 2022-11-30 at 10 10 38@2x

And though "Sent email" shows on the column, when I click around in the filter, for some reason it shows "Received email" again. I can't seem to reliably reproduce this, it seems if I just pick some random emails from the list.
CleanShot 2022-11-30 at 10 13 40@2x

Hope that's cool!

moreofmorris (approved):
This is looking great now, thanks @lenabaidakova!

lenabaidakova:
> This is looking good so far, I think these changes make sense seeing them in practice. Curious what others will think, too!
> 
> I did notice a few bits... If there is only More Like This or only Less Like This available, then we maybe only show that one so people aren't clicking through to no members in the filtered list. CleanShot 2022-11-30 at 10 10 38@2x
> 
> And though "Sent email" shows on the column, when I click around in the filter, for some reason it shows "Received email" again. I can't seem to reliably reproduce this, it seems if I just pick some random emails from the list. CleanShot 2022-11-30 at 10 13 40@2x
> 
> Hope that's cool!

@moreofmorris , good catch! 🚀 
- Fixed **More/Less like this** links, now should be visible if any info available
- Fixed column **Received email**

All should be good 😊

PR #15899, opened 2022-11-29
Implemented email previews and tests using new email flow

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2330

Uses new flow for previewing and testing emails (only if email stability flag is enabled)

### DIFF
--- a/ghost/email-service/lib/email-service.js
+++ b/ghost/email-service/lib/email-service.js
@@ -160,6 +160,7 @@ class EmailService {
                 exampleMember.email = member.get('email');
                 exampleMember.name = member.get('name');
             } else {
+                exampleMember.name = ''; // Force empty name to simulate name fallbacks
                 exampleMember.email = email;
             }
         }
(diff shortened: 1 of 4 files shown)


ghost/email-service/lib/email-service.js
```diff
@@ -127,14 +135,87 @@ class EmailService {
...
+     * @param {*} newsletter 
+     * @param {import('./email-renderer').Segment} segment 
+     * @param {string[]} emails
+     */
     async sendTestEmail(post, newsletter, segment, emails) {
-        // eslint-disable-next-line no-restricted-syntax
-        throw new Error('Sending a test email has not been implemented yet. Turn off the email stability flag is you need this functionality.');
+        const members = [];
+        for (const email of emails) {
+            members.push(await this.getExampleMember(email));
```
rishabhgrg, line 206: 🤔 Wondering if its a good idea to use dummy values for `name` when someone enters an email that doesn't match a valid member on site, as the current `getExampleMember` method will always attach a member name to the member data which is not ideal while testing fallbacks on the email.

SimonBackx: Good point, I'll remove the default name :)

SimonBackx: Or they could enter a member email address that doesn't have a name to test fallbacks 🤔

rshbhgrg (approved):

PR #15901, opened 2022-11-29
Added storage for email failures

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2332

Saves events in the database and collects error information.

Do note that we can emit the same events multiple times, and as a result out of order. That means we should correctly handle that a delivered event might be fired after a permanent failure. So a delivered event is ignored if the email is already marked as failed. Also delivered_at is reset to null when we receive a permanent failure.

### DIFF
--- a/ghost/email-service/lib/email-event-storage.js
+++ b/ghost/email-service/lib/email-event-storage.js
@@ -1,12 +1,15 @@
-const {EmailDeliveredEvent, EmailOpenedEvent, EmailBouncedEvent, EmailUnsubscribedEvent, SpamComplaintEvent} = require('@tryghost/email-events');
+const {EmailDeliveredEvent, EmailOpenedEvent, EmailBouncedEvent, EmailTemporaryBouncedEvent, EmailUnsubscribedEvent, SpamComplaintEvent} = require('@tryghost/email-events');
 const moment = require('moment-timezone');
+const logging = require('@tryghost/logging');
 
 class EmailEventStorage {
     #db;
     #membersRepository;
+    #models;
 
-    constructor({db, membersRepository}) {
+    constructor({db, models, membersRepository}) {
         this.#db = db;
+        this.#models = models;
         this.#membersRepository = membersRepository;
     }
 
@@ -20,7 +23,19 @@ class EmailEventStorage {
         });
 
         domainEvents.subscribe(EmailBouncedEvent, async (event) => {
-            await this.handlePermanentFailed(event);
+            try {
+                await this.handlePermanentFailed(event);
+            } catch (e) {
+                logging.error(e);
+            }
+        });
+
+        domainEvents.subscribe(EmailTemporaryBouncedEvent, async (event) => {
+            try {
+                await this.handleTemporaryFailed(event);
+            } catch (e) {
+                logging.error(e);
+            }
         });
 
         domainEvents.subscribe(EmailUnsubscribedEvent, async (event) => {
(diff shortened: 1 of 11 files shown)


ghost/email-service/lib/email-event-storage.js
```diff
@@ -1,40 +1,75 @@
...
         });
     }
 
     async handleDelivered(event) {
+        // To properly handle events that are received out of order (this happens because of polling)
+        // we only can set an email recipient to delivered if they are not already marked as failed
+        // Why handle this her?  An email can be 'delivered' and later have a delayed bounce event. So we need to prevent that delivered_at is set again.
         await this.#db.knex('email_recipients')
             .where('id', '=', event.emailRecipientId)
+            .whereNull('failed_at')
```
rishabhgrg, line 72: 🤔 What happens if we have a temporary failure, and then we later have that email delivered by mailgun on retry ? I think in that case we'll miss the delievered event which'll not be right. We could maybe update the check for both null failed_at or severity is temporary which will mean only ignore delivered after permanent failures, wdyt ?

SimonBackx: For temporary failures we don't set failed_at, so that should still work, right? 🤔

rishabhgrg: Ah right 👌

ghost/mailgun-client/lib/mailgun-client.js
```diff
@@ -172,12 +172,19 @@ module.exports = class MailgunClient {
...
+            id: event.id,
             type: event.event,
             severity: event.severity,
             recipientEmail: event.recipient,
             emailId: event['user-variables'] && event['user-variables']['email-id'],
             providerId: providerId,
-            timestamp: new Date(event.timestamp * 1000)
+            timestamp: new Date(event.timestamp * 1000),
+
+            error: event['delivery-status'] && event['delivery-status'] ? {
```
rishabhgrg, line 183: > event['delivery-status'] && event['delivery-status']

This seems like an accidental repeat, guess the second one wanted to test something like if `code` exists on `delivery-status` ?

ghost/mailgun-client/lib/mailgun-client.js
```diff
@@ -172,12 +172,19 @@ module.exports = class MailgunClient {
...
+                code: event['delivery-status'].code,
+                message: event['delivery-status'].message || event['delivery-status'].description,
```
rishabhgrg, line 185: While normally it shouldn't happen, but think we should restrict the final value here for message to `2000` chars as that's the max allowed length in our DB. Will allow us to capture the error in case mailgun returns a message/description which is longer instead of failing due to DB limit, wdyt ? 
Same thing for `enhanced_code` as well.

ghost/mailgun-client/lib/mailgun-client.js
SimonBackx: Oh good catch, that was certainly the intention

ghost/mailgun-client/lib/mailgun-client.js
SimonBackx: Cool, I'll add a length restriction

rshbhgrg (approved):

ErisDS:
> Do note that we can emit the same events multiple times, and as a result out of order. That means we should correctly handle that a delivered event might be fired after a permanent failure. So a delivered event is ignored if the email is already marked as failed. Also delivered_at is reset to null when we receive a permanent failure.

This is unexpected in my opinion. Why have we changed this logic?

IMO we should not be resetting any events as it’s all part of the trail of what happened to an email. A large point of the work being done is to improve the train and increase the data, removing part of the story seems extremely counter to that.

PR #15906, opened 2022-11-30
Removed unused members settings module

naz:
- This module is marked for deletion with a todo comment.  It also popped up as least covered place in the codebase, because it was never used.

@SimonBackx as you've added the original "todo" for removal of this module, can you please double check if I'm not missing anything with this change? The tests did pass locally, but I do not have enough context for members to know if something else to consider 🙃


SimonBackx (approved):
I couldn't find any remaining usage of it, so I think it is safe to remove 😊

naz:
Thanks Simon 🙌

PR #15908, opened 2022-11-30
Wired up storage of EmailSpamComplaintEvents

allouis:
refs https://github.com/TryGhost/Team/issues/2337

This also removes the code which unsubscribes members from newsletters because that is not the spec of the feature.

### DIFF
--- a/ghost/email-service/lib/email-event-storage.js
+++ b/ghost/email-service/lib/email-event-storage.js
@@ -13,6 +13,9 @@ class EmailEventStorage {
         this.#membersRepository = membersRepository;
     }
 
+    /**
+     * @param {import('@tryghost/domain-events')} domainEvents
+     */
     listen(domainEvents) {
         domainEvents.subscribe(EmailDeliveredEvent, async (event) => {
             try {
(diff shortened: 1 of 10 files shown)


ghost/email-service/lib/email-event-storage.js
```diff
@@ -61,7 +66,11 @@ class EmailEventStorage {
     }
 
     async handleComplained(event) {
-        return this.unsubscribeFromNewsletters(event);
+        await this.#EmailSpamComplaintEvent.add({
```
SimonBackx, line 69: Awaiting discussion about duplicate events, this might need protection to prevent adding the same event multiple times in the db

ghost/email-service/lib/email-event-storage.js
```diff
@@ -61,7 +66,11 @@ class EmailEventStorage {
     }
 
     async handleComplained(event) {
-        return this.unsubscribeFromNewsletters(event);
```
SimonBackx, line 64: Do we want a feature flag check here?

allouis: Discussed in slack: we're good to store the new data without a flag check, as it'll give us a backlog of events if release is delayed.

ghost/email-service/lib/email-event-storage.js
allouis: Going to add protection for now - good shout!

allouis:
@daniellockyer I've updated the schema and a migration to add a unique constraint to the email_spam_complaint_events table, so that we can ensure that duplicate events do not get inserted.

daniellockyer (approved):
@allouis The change is fine with me but this won't apply retroactively to developers who have already run the migration, so please keep this in mind when it comes to bugs found during development

PR #15910, opened 2022-11-30
Added E2E tests for batch sending

SimonBackx:
refs https://github.com/TryGhost/Team/issues/2339

- Includes a new pattern in the job manager that allows us to properly await jobs.
- Added new convenience mocking methods to stub settings
- Tests the main flows for bulk sending:
    - Sending in multiple batches
    - Sending to multiple segments
    - Handling a failed batch and retrying that batch
- Fixes bug in batch generation (ordering not working)

In a different PR I'll add more detailed tests.

### DIFF
--- a/ghost/core/test/e2e-api/admin/members-exporter.test.js
+++ b/ghost/core/test/e2e-api/admin/members-exporter.test.js
@@ -150,13 +150,13 @@ describe('Members API — exportCSV', function () {
             })
         });
 
-        const labelsList = labels.map(label => label.get('name')).join(',');
+        const labelsList = labels.map(label => label.get('name')).sort().join(',');
 
         await testOutput(member, (row) => {
             basicAsserts(member, row);
             should(row.subscribed_to_emails).eql('false');
             should(row.complimentary_plan).eql('');
-            should(row.labels).eql(labelsList);
+            should(row.labels.split(',').sort().join(',')).eql(labelsList);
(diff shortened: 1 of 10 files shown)


ghost/email-service/lib/mailgun-email-provider.js
```diff
@@ -144,16 +144,29 @@ class MailgunEmailProvider {
...
+                });
+            } else {
+                ghostError = new errors.EmailError({
+                    statusCode: undefined,
+                    message: e.message,
+                    errorDetails: undefined,
+                    context: e.context || 'Mailgun Error',
+                    code: 'BULK_EMAIL_SEND_FAILED'
+                });
+            }
```
SimonBackx, line 169: We need to catch code errors apart from Mailgun errors, so this change was needed to prevent that the catch would also throw

ghost/core/test/e2e-api/admin/members-exporter.test.js
```diff
@@ -150,13 +150,13 @@ describe('Members API — exportCSV', function () {
             })
         });
 
-        const labelsList = labels.map(label => label.get('name')).join(',');
+        const labelsList = labels.map(label => label.get('name')).sort().join(',');
```
SimonBackx, line 153: Just a random CI error that I fixed with this. Unrelated to other changes.

PR #15923, opened 2022-12-02
Improved email concurrency sending

SimonBackx:
no issue

- Send 10 batches at the same time
- Increase job manager concurrency to 3 (allows to send emails and have a member import at the same time).

### DIFF
--- a/ghost/job-manager/lib/job-manager.js
+++ b/ghost/job-manager/lib/job-manager.js
@@ -39,7 +39,7 @@ class JobManager {
      * @param {Object} [options.domainEvents] - domain events emitter
      */
     constructor({errorHandler, workerMessageHandler, JobModel, domainEvents}) {
-        this.queue = fastq(this, worker, 1);
+        this.queue = fastq(this, worker, 3);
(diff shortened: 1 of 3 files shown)


ghost/job-manager/lib/job-manager.js
```diff
@@ -39,7 +39,7 @@ class JobManager {
      * @param {Object} [options.domainEvents] - domain events emitter
      */
     constructor({errorHandler, workerMessageHandler, JobModel, domainEvents}) {
-        this.queue = fastq(this, worker, 1);
+        this.queue = fastq(this, worker, 3);
```
rishabhgrg, line 42: Will defer this to @naz as he has winder context on introducing such change in job manager :)

SimonBackx: @naz Reverting this for now until you are back. We need to talk about this later, as I think we need to increase this a bit. When you are importing a lot of members all the jobs stall because of the concurrency of 1.
```suggestion
        this.queue = fastq(this, worker, 1);
```

PR #15934, opened 2022-12-03
Changed tag for imported content to be internal

PaulAdamDavis:
No issue

When importing content from a JSON file in Settings > Labs, a public tag like `Import 2022-12-03 19:57` gets added to each newly imported post.

I believe this tag should not be public. It definitely serves a useful purpose but has no useful function for readers of the site and should not be shown to readers.

As an example, if the imported content has no other tags (which is common when coming from some other platforms), the import tag becomes the primary tag and is shown front & center to readers. Here's an example from Casper.

screenshot-08-09-20-03-12-2022

With this change, the tag is now internal `#Import 2022-12-03 19:57` & `hash-import 2022-12-03 19:57`. Its utility is the same, but won't clutter the front-end anymore.

### DIFF
--- a/ghost/core/core/server/api/endpoints/db.js
+++ b/ghost/core/core/server/api/endpoints/db.js
@@ -83,7 +83,7 @@ module.exports = {
         permissions: true,
         query(frame) {
             const siteTimezone = settingsCache.get('timezone');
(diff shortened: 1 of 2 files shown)


JohnONolan:
Agree 👍

PR #15937, opened 2022-12-05
Fixed suppression list handling of spam complaint events

allouis:
refs https://github.com/TryGhost/Team/issues/2351

We were storing all suppressions with a reason of bounced, rather than
spam for spam complaint events.


ErisDS:
I think this fix went out already?

allouis:
Yup I pushed these fixes to `main`, I guess it wasn't in sync and GitHub didn't autoclose!

PR #15944, opened 2022-12-05
🐛 Fixed errors of old events from deleted members

SimonBackx:
refs https://ghost.slack.com/archives/C02G9E68C/p1670215917451249

When a member is deleted, and we receive an opened event for an email to that member. We threw an uncaught Bookshelf EmptyResponse error.

- This change makes fetching the member not a requirement when handling that event in the last seen at updater.
- It also adds try catches for all event listeners in the last seen at updater


rshbhgrg (approved):
Tiny logging improvement suggestion but looks good overall, great catch! 🎣

PR #15949, opened 2022-12-06
Fixed unnecessary identity requests and 403s for non-Owner accounts

kevinansfield:
no issue

`<GhBillingIframe>` generates a request to the `/identities/` endpoint every time Admin is accessed for all users, however that endpoint is only accessible to users with the Owner role meaning we have a lot of unnecessary 403 errors in event logs and the developer console.

- added early exit when we know the logged in user doesn't have the Owner role
- removed the subscription fetching code that wasn't reachable (`token` was always `undefined`)
  - the BMA sends subscription data as soon as it's available so the extra fetch isn't necessary


aileen (approved):
LGTM!

PR #15952, opened 2022-12-06
Unsubscribed existing Members with suppressed emails

allouis:
refs https://github.com/TryGhost/Ghost/pull/15951 - This PR contains the changes for future Members

refs https://github.com/TryGhost/Team/issues/2367

We already have existing Members which have their emails suppressed that need to have their newsletter subscriptions removed.


lenabaidakova (approved):

ghost/core/core/server/data/migrations/versions/5.26/2022-12-05-09-56-update-newsletter-subscriptions.js
```diff
@@ -0,0 +1,18 @@
...
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const rows = await knex
+              .pluck('m.id')
+              .from('members AS m')
+              .innerJoin('suppressions AS s', 'm.email', '=', 's.email_address');
+
+        logging.info(`Unsubscribing ${rows.length} members from all newsletters`);
```
daniellockyer, line 11: Can we early-return here if `rows.length` == 0? Knex seems to execute the following lines with

```
[2022-12-08 05:48:57] INFO Unsubscribing 0 members from all newsletters
  knex:query delete from `members_newsletters` where 1 = ? trx5 +2ms
  knex:bindings [ 0 ] trx5 +2ms
```

which is just odd and avoidable

allouis: Sure

daniellockyer (approved):

rshbhgrg (approved):

allouis:
This is blocked on merging until we release the feature

daniellockyer (approved):

PR #15954, opened 2022-12-06
🐛 Fixed broken redemption count for offers

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2369

- offer id was not getting attached to stripe checkout metadata, causing the checkout event to not store any offer information for a subscription. This got changed in a prev refactor [here](https://github.com/TryGhost/Ghost/commit/25d8d694a051781e9f45e3542b75143462c69ec4#diff-b7dfcd660902a2a20dff7da5e886d8e10234bda4ba78228255afc8d4a8e78cf6L206)
- cleans up offer id handling for checkout session event

### DIFF
--- a/ghost/members-api/lib/repositories/member.js
+++ b/ghost/members-api/lib/repositories/member.js
@@ -914,13 +914,6 @@ module.exports = class MemberRepository {
             }
         } else if (offerId) {
             offer = await this._offerRepository.getById(offerId, {transacting: options.transacting});
-            if (offer) {
-                offerId = offer.id;
-            } else {
-                // Reset offer id to null as it is not valid
-                offerId = null;
-                logging.error(`Received an unknown offer id (${offerId}) for subscription - ${subscription.id}.`);
-            }
         }
 
         const subscriptionData = {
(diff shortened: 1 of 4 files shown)


ghost/members-api/lib/repositories/member.js
```diff
@@ -914,6 +914,13 @@ module.exports = class MemberRepository {
             }
         } else if (offerId) {
             offer = await this._offerRepository.getById(offerId, {transacting: options.transacting});
+            if (offer) {
```
allouis, line 917: I don't understand this change - is it directly related to the bug?

allouis (commented):
Looks good - just one question about a change!

ghost/members-api/lib/repositories/member.js
rishabhgrg: Its indirectly related, part of the bug wouldn't have happened if we were using the calculated offerId above in the subscription created event instead of using `data.offerId` directly, which just attaches the offer id from metadata to checkout session. Because of that we missed the offer id calculated by stripe coupon as well due to the main bug, otherwise we would have atleast captured coupon offers.
Using calculated `offerId` here ensures that we don't blindly pass in the offer id that we get in data, but only pass in the calculated valid offer id

allouis: Okay, so do we need this change to fix the bug? If we don't I'd prefer to remove it and open a separate issue for a separate bug so that we can keep this patch as minimal as possible

rishabhgrg: Cool fair, will remove it from here and we can add this in later separately.

allouis (approved):

PR #15959, opened 2022-12-07
Added Playwright "Private site" test

naz:

- Test enabling private site and checking access with a password

The test is failing due to invalid private site login response (looks like passwords do not match for some reason). Needs a bit more work.


sam-lord:
This will need a tidy-up step to remove the password (& maybe assert that it's accessible again?) to work with the staging site setup. 

Looks nice though!

naz:
landed in main as https://github.com/TryGhost/Ghost/commit/ffd07ea74435716559c29fdba3dda1cade5870e5

PR #15960, opened 2022-12-07
Added post scheduling Playwright test and loosened time restrictions

SimonBackx:

- Adds a test that schedules a post 5 seconds in the future and waits for it to be published
- Reduced the time restrictions for scheduling: 
    - The minimum time in the frontend is now 5 seconds in the future (came from 5 minutes in the future) 
    - The time picker now suggests 10 minutes in the future instead of the minimum scheduling time (came from 5 minutes) 
    - In the backend, a post will be allowed to be scheduled if it is at least 2 minutes in the past (came from 2 minutes in the future) 
    - The scheduler will publish a post if it is at least 5 minutes in the past, and maximum 5 minutes in the future (came from 2 minutes)

### DIFF
--- a/ghost/admin/app/validators/post.js
+++ b/ghost/admin/app/validators/post.js
@@ -193,7 +193,7 @@ export default BaseValidator.create({
                 model.errors.add('publishedAtBlogDate', 'Must be in the past');
                 this.invalidate();
 
-            // scheduled must be at least 2 mins in the future when first scheduling
+            // scheduled must be in the future when first scheduling
             } else if ((model.changedAttributes().status || model.changedAttributes().publishedAtUTC) && status === 'scheduled' && !isInFuture) {
                 model.errors.add('publishedAtBlogDate', 'Must be in the future');
                 this.invalidate();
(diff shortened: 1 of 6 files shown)


ghost/core/test/e2e-browser/admin/publishing.spec.js
```diff
@@ -35,7 +35,7 @@ const openPublishFlow = async (page) => {
  * @param {import('@playwright/test').Page} page
  */
 const closePublishFlow = async (page) => {
-    await page.locator('[data-test-modal="publish-flow"] [data-test-button="back-to-editor"]').click();
+    await page.locator('[data-test-button="close-publish-flow"]').click();
```
SimonBackx, line 38: The other button is not present for scheduled post, but the top left button is

ghost/admin/app/validators/post.js
```diff
@@ -195,7 +195,7 @@ export default BaseValidator.create({
 
             // scheduled must be at least 2 mins in the future when first scheduling
```
kevinansfield, line 196: Can we update/remove this comment to avoid any confusion?

SimonBackx:
```suggestion
            // scheduled must be in the future when first scheduling
```

kevinansfield (approved):

PR #15962, opened 2022-12-07
Added Playwright "Un-publish + Re-publish with send" test

kevinansfield:

- added more utils for working with the publish flows

### DIFF
--- a/ghost/core/test/e2e-browser/admin/publishing.spec.js
+++ b/ghost/core/test/e2e-browser/admin/publishing.spec.js
@@ -1,5 +1,6 @@
 const {expect, test} = require('@playwright/test');
 const {DateTime} = require('luxon');
+const {createMember} = require('../utils');
 
 /**
  * Start a post draft with a filled in title and body. We can consider to move this to utils later.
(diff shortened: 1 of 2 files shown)


ghost/core/test/e2e-browser/admin/publishing.spec.js
```diff
@@ -58,7 +66,7 @@ const publishPost = async (page, {type = 'publish', time} = {}) => {
 
     // set the publish type
     await page.locator('[data-test-setting="publish-type"] > button').click();
-    await page.locator(`[data-test-publish-type="${type}"]`).setChecked(true);
+    await page.locator(`[data-test-publish-type="${type}"]`).setChecked(true, {timeout: 50});
```
naz, line 69: @kevinansfield  The `locator(`[data-test-publish-type="${type}"]`)` doesn't really work once Mailgun is enabled. With all three publishing options enabled the selector fails to choose the right option. When Mailgun is disabled, the selector chose the only available button and moved on.

naz: I've pushed a hacky fix with "findByText" selectors for this, so I could debug Mailgun further

daniellockyer:
Going to have to close this for now because the PR is stale. Please re-open if the contents were important 🙂

PR #15970, opened 2022-12-08
Archive/unarchive tiers playwright test

ErisDS:
I'll admit, I was obsessing over details of how the utilities worked and how to do this instead of focusing on getting it done.

If someone else wants to wrap it cool, else I'll pick it up later


rishabhgrg:
@ErisDS I can take this up, to confirm tho this is just supposed to do archive/unarchive of tier in Admin right ? Checking because the test is under `porta/tiers.spec.js` but should probably be in `admin/tiers.spec.js` in that case

lenabaidakova:
Test was added here https://github.com/TryGhost/Ghost/pull/16101

PR #15973, opened 2022-12-09
🐛 Fixed 'Invalid status code: undefined' in members api

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2377

When there is an error thrown that is not a Ghost error, there is no status code in the error. Calling res.writeHead with an undefined status code, throws an error and crashes Ghost.

This change fixes that and adds logging for those errors.


lenabaidakova (approved):
:shipit:

PR #15974, opened 2022-12-09
Resubscribe member to default newsletters after removing from suppression list

lenabaidakova:
refs TryGhost/Team#2372

### DIFF
--- a/ghost/core/core/server/services/members/middleware.js
+++ b/ghost/core/core/server/services/members/middleware.js
@@ -101,7 +101,13 @@ const getMemberData = async function (req, res) {
 const deleteSuppression = async function (req, res) {
     try {
         const member = await membersService.ssr.getMemberDataFromSession(req, res);
+        const options = {
+            id: member.id,
+            withRelated: ['newsletters']
+        };
         await emailSuppressionList.removeEmail(member.email);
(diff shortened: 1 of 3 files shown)


allouis (commented):
I think that ideally we might wanna use `update({subscribed: true}` rather than fetchign the default newsletters ourselves, so that we keep the logic of getting them in the members package.

I'm also not 100% sure about if this should happen in the middleware 🤔  Did you see how the removal of newsletters was implemented https://github.com/TryGhost/Ghost/commit/9736d942e1041ffc9e588db1386b2271599151a9?

It would also be cool to get a test for this!

lenabaidakova:
> I think that ideally we might wanna use `update({subscribed: true}` rather than fetchign the default newsletters ourselves, so that we keep the logic of getting them in the members package.
> 
> I'm also not 100% sure about if this should happen in the middleware 🤔 Did you see how the removal of newsletters was implemented [9736d94](https://github.com/TryGhost/Ghost/commit/9736d942e1041ffc9e588db1386b2271599151a9)?
> 
> It would also be cool to get a test for this!

I found such comment https://github.com/TryGhost/Ghost/blob/main/ghost/members-api/lib/repositories/member.js#L521

allouis (approved):

PR #15976, opened 2022-12-09
Added try catch for errors in all domain event listeners

SimonBackx:
refs https://github.com/TryGhost/Team/issues/2370

Reduces the amount of unhandled errors that could cause a crash in Ghost


allouis (approved):
Bit of a shame to add hard dependency to this module, but we can refactor in future when we make this not a singleton

PR #15995, opened 2022-12-13
🐛 Handled unknown Mailgun events

SimonBackx:
refs https://ghost.slack.com/archives/C02G9E68C/p1670916538764019

- We receive events that don't have an emailId or providerId.
- We filter those events now and log them as an error

### DIFF
--- a/ghost/mailgun-client/lib/mailgun-client.js
+++ b/ghost/mailgun-client/lib/mailgun-client.js
@@ -125,7 +125,7 @@ module.exports = class MailgunClient {
                 value: Date.now() - startTime,
                 statusCode: 200
             });
-            let events = page?.items?.map(this.normalizeEvent) || [];
+            let events = (page?.items?.map(this.normalizeEvent) || []).filter(e => !!e);
             debug(`fetchEvents: finished fetching first page with ${events.length} events`);
 
             let eventCount = 0;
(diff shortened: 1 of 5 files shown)


ghost/mailgun-client/lib/mailgun-client.js
```diff
@@ -203,6 +203,12 @@ module.exports = class MailgunClient {
     normalizeEvent(event) {
         const providerId = event?.message?.headers['message-id'];
 
+        if (!providerId && !(event['user-variables'] && event['user-variables']['email-id'])) {
```
rishabhgrg, line 206: Double checking the use of `&&` here in this condition, what happens if we have a case where `providerId` is empty but `event['user-variables']['email-id']` exists ? If such a case does happen, we'll go ahead with this as a valid event and is that acceptable ?

SimonBackx: Yes, that is acceptable because we can still transform the providerId into an emailId later: https://github.com/TryGhost/Ghost/blob/aa693039a354a186f1ecd4a279a7b015f67bddc7/ghost/email-service/lib/email-event-processor.js#L147-L168

I don't know if we ever get those events, but we have been supporting them in the past

rishabhgrg: Ah cool, all good then 😄

rshbhgrg (approved):

PR #15996, opened 2022-12-13
Fixed email preview text with new stability flow

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2382

The preview text is getting set to subject line in the new email flow so it repeats multiple times in the inbox(subject+preview+title). This was because the new flow doesn't use the post serialisation that the old system did, causing excerpt to be empty in the email rendering.

Old system was using post serialisation here - https://github.com/TryGhost/Ghost/blob/a721e4f2d7428e59f153bbd0e06d19d0c7ef9e63/ghost/core/core/server/services/mega/post-email-serializer.js#L136-L139. 

This change adds explicit method to calculate the preview text for email in email renderer service using same logic as used in old system.

### DIFF
--- a/ghost/email-service/lib/email-renderer.js
+++ b/ghost/email-service/lib/email-renderer.js
@@ -476,6 +476,21 @@ class EmailRenderer {
         return this.#renderTemplate(data);
     }
 
+    #getPostExcerpt(postModel) {
+        let plaintext = postModel.get('plaintext');
+        let customExcerpt = postModel.get('custom_excerpt');
+
+        if (customExcerpt !== null){
+            return customExcerpt;
+        } else {
+            if (plaintext) {
+                return plaintext.substring(0, 500);
+            } else {
+                return null;
+            }
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


ghost/email-service/lib/email-renderer.js
```diff
@@ -476,6 +476,25 @@ class EmailRenderer {
...
 
+    /**
+     * Get email preheader text from post model
+     * @param {object} postModel
+     * @returns
+     */
+    #getEmailPreheader(postModel) {
+        let plaintext = postModel.get('plaintext');
+        let customExcerpt = postModel.get('custom_excerpt');
+        if (customExcerpt !== null){
```
SimonBackx, line 487: In the old system, and empty `custom_excerpt`. would get transformed into `${postModel.get('title')} – `, maybe we better change this check to accomplish the same? 

```suggestion
        if (customExcerpt) {
```

SimonBackx (approved):

PR #16021, opened 2022-12-16
🐛 Fixed invalid email getting saved for members

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2235

We found some cases which can cause a site to have member emails that have invalid characters like `member@example.com�`. This happened due to the `validator` version used by Ghost not able to catch some specific cases as invalid email, allowing members to be created with them either via Admin or Importer or direct signup. Portal UI already blocked these email as invalid. This change

- updates `@tryghost/validator` to include a latest version of email validator that catches these invalid cases
- doesn't allow member creation with invalid email like above
- doesn't allow existing member emails to be edited to invalid

### DIFF
--- a/ghost/members-api/lib/repositories/member.js
+++ b/ghost/members-api/lib/repositories/member.js
@@ -6,6 +6,7 @@ const DomainEvents = require('@tryghost/domain-events');
 const {MemberCreatedEvent, SubscriptionCreatedEvent, MemberSubscribeEvent, SubscriptionCancelledEvent} = require('@tryghost/member-events');
 const ObjectId = require('bson-objectid').default;
 const {NotFoundError} = require('@tryghost/errors');
+const validator = require('@tryghost/validator');
 
 const messages = {
     noStripeConnection: 'Cannot {action} without a Stripe Connection',
@@ -16,7 +17,8 @@ const messages = {
     subscriptionNotFound: 'Could not find Subscription {id}',
     productNotFound: 'Could not find Product {id}',
     bulkActionRequiresFilter: 'Cannot perform {action} without a filter or all=true',
-    tierArchived: 'Cannot use archived Tiers'
+    tierArchived: 'Cannot use archived Tiers',
+    invalidEmail: 'Invalid Email'
 };
 
 /**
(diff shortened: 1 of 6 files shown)


ghost/members-api/lib/repositories/member.js
```diff
@@ -439,6 +449,17 @@ module.exports = class MemberRepository {
             }
         }
 
+        // Throw error if email is invalid and it's been changed
+        if (
+            initialMember.get('email') !== memberData.email &&
```
SimonBackx: `initialMember` can be null here

ghost/members-api/lib/repositories/member.js
```diff
@@ -439,6 +449,18 @@ module.exports = class MemberRepository {
...
+            initialMember?.get('email') && memberData.email
+            && initialMember.get('email') !== memberData.email
+            && !validator.isEmail(memberData.email, {legacy: false})
+        ) {
+            throw new errors.ValidationError({
+                message: tpl(messages.invalidEmail),
+                property: 'email'
+            });
+        }
+
```
SimonBackx, line 463: Currently this won't validate the email address if we are not updating `products` or `newsletters`, because then initialMember will always be null 🤔

rishabhgrg: Yeah true, I thought about that as an acceptable edge case given it won't be hit via the normal routes end user can update a member's email. Adding an extra check to fetch member and compare the email if it's changed seems like a big change to cover for this edge case atm, as the main idea is to block new invalid emails to enter the system. Wdyt ?

SimonBackx: Yes, I agree that now is not the right time to always fetch the member. Maybe we do need a cleanup issue for this?

SimonBackx (approved):

ghost/members-api/lib/repositories/member.js
rishabhgrg: Yup, i'll add a note on original as well as to a related cleanup issue to update the email validator to use new method instead of legacy :)

PR #16041, opened 2022-12-22
feat: create environment for every PR using Uffizzi

daramayis:
This PR is a part of https://github.com/TryGhost/Ghost-CLI/issues/1700 and helps demonstrate Uffizzi integration with **Ghost**. 

This will help create preview environments on all of **Ghost** Pull Requests and help contributors including the maintainers iterate faster on their PRs. 
I have created a PR over https://github.com/daramayis/f-Ghost/pull/1#issuecomment-1357606670 to show what it looks like to have a preview environment deployed against your PR.

You can check out the preview over [here](https://app.uffizzi.com/github.com/daramayis/f-Ghost/pull/1) as well.

cc: @waveywaves

### DIFF
--- a/.github/workflows/uffizzi-build.yml
+++ b/.github/workflows/uffizzi-build.yml
@@ -1,0 +1,95 @@
+name: Build PR Image
+on:
+  pull_request:
+    types: [opened,synchronize,reopened,closed]
+
+jobs:
+
+  build-application:
+    name: Build and Push `application`
+    runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


ErisDS:
Hey @daramayis, thanks for raising a PR and demoing this. It isn't what I was expecting as the original issue was raised in Ghost-CLI, so I thought you had some system where we could test new versions of the CLI tool - which is very much an unsolved problem.

As we already have a system that deploys latest Ghost builds along with fixtures and a configuration that matches our production environments. At the moment we don't have a clear use case for these Uffizzi builds for PRs.

Thanks again 🙂

PR #16054, opened 2023-01-02
Added error handling for ENAMETOOLONG import error

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2200

When zipping a folder that contains files with UTF-8 characters in the filename, using the MacOS Archive Utility, the resulting zip will be missing some UTF-8 configuration bit. This breaks the unzipper, causing it to decode the filenames using the wrong encodign.

When the file names are long, and become longer than the length allowed by the OS, an ENAMETOOLONG error is thrown. This error is not handled by the importer, and causes the import to fail.

This adds a specific check for this error so we can show a clear error message to the user, that helps them to resolve the issue. We are currently unable to fix the issue on our side, because of a lack of well supported zip libraries for node.


ErisDS (changes requested):
Have added some notes to the issue :)

SimonBackx:
@ErisDS Could you give this another look? 😊

PR #16060, opened 2023-01-03
Improved email verification required checks

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2366
refs https://ghost.slack.com/archives/C02G9E68C/p1670232405014209

Probem described in issue.

In the old MEGA flow:
- The `email_verification_required` check is now repeated inside the job

In the new email service flow:
- The `email_verification_required` is now checked (didn't happen before)
- When generating the email batch recipients, we only include members that were created before the email was created. That way it is impossible to avoid limit checks by inserting new members between creating an email and sending an email.
- We don't need to repeat the check inside the job because of the above changes

Improved handling of large imports:
- When checking `email_verification_required`, we now also check if the import threshold is reached (a new method is introduced in vertificationTrigger specifically for this usage). If it is, we start the verification progress. This is required for long running imports that only check the verification threshold at the very end.
- This change increases the concurrency of fastq to 3 (refs https://ghost.slack.com/archives/C02G9E68C/p1670232405014209). So when running a long import, it is now possible to send emails without having to wait for the import. Above change makes sure it is not possible to get around the verification limits.

Refactoring:
- Removed the need to use `updateVerificationTrigger` by making thresholds getters instead of fixed variables.
- Improved awaiting of members import job in regression test

### DIFF
--- a/ghost/core/test/regression/api/admin/members-importer.test.js
+++ b/ghost/core/test/regression/api/admin/members-importer.test.js
@@ -8,10 +8,10 @@ const config = require('../../../../core/shared/config');
 const configUtils = require('../../../utils/configUtils');
 const settingsCache = require('../../../../core/shared/settings-cache');
 const models = require('../../../../core/server/models');
+const jobManager = require('../../../../core/server/services/jobs/job-service');
 
 const {mockManager, sleep} = require('../../../utils/e2e-framework');
 const assert = require('assert');
-const {_updateVerificationTrigger} = require('../../../../core/server/services/members');
 
 let request;
 
@@ -253,7 +253,6 @@ describe('Members Importer API', function () {
             verified: false,
             escalationAddress: 'test@example.com'
         });
-        _updateVerificationTrigger();
 
         const res = await request
             .post(localUtils.API.getApiQuery(`members/upload/`))
@@ -321,7 +320,8 @@ describe('Members Importer API', function () {
             verified: false,
             escalationAddress: 'test@example.com'
         });
-        _updateVerificationTrigger();
+
+        const awaitCompletion = jobManager.awaitCompletion('members-import');
 
         const res = await request
             .post(localUtils.API.getApiQuery(`members/upload/`))
@@ -338,7 +338,7 @@ describe('Members Importer API', function () {
         should.exist(jsonResponse.meta);
 
         // Wait for the job to finish
-        await sleep(15000);
+        await awaitCompletion;
 
         assert(!!settingsCache.get('email_verification_required'), 'Email verification should now be required');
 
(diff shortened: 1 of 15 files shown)


ghost/verification-trigger/test/verification-trigger.test.js
```diff
@@ -345,7 +403,7 @@ describe('Email verification flow', function () {
         });
 
         const trigger = new VerificationTrigger({
-            apiTriggerThreshold: Infinity,
+            getImportTriggerThreshold: () => Infinity,
```
SimonBackx, line 406: Note that the test was testing on the wrong variable here

SimonBackx:
@allouis This PR also contains the change we discussed yesterday that in the new email flow we'll only send the email to the members at the time when the email model was created

ghost/core/test/regression/api/admin/members-importer.test.js
```diff
@@ -321,7 +320,8 @@ describe('Members Importer API', function () {
             verified: false,
             escalationAddress: 'test@example.com'
         });
-        _updateVerificationTrigger();
+
+        const awaitCompletion = jobManager.awaitCompletion('members-import');
```
naz, line 324: Nice 👌

ghost/verification-trigger/package.json
```diff
@@ -7,7 +7,7 @@
   "main": "index.js",
   "scripts": {
     "dev": "echo \"Implement me!\"",
-    "test:unit": "NODE_ENV=testing c8 --all --check-coverage --reporter text --reporter cobertura mocha './test/**/*.test.js'",
+    "test:unit": "NODE_ENV=testing c8 --all --check-coverage --100 --reporter text --reporter cobertura mocha './test/**/*.test.js'",
```
naz, line 10: Yessss 💪

ghost/email-service/lib/email-service.js
```diff
@@ -73,7 +79,15 @@ class EmailService {
 
         // Check host limit for disabled emails or going over emails limit
         if (this.#limitService.isLimited('emails')) {
-            await this.#limitService.errorIfWouldGoOverLimit('emails');
+            await this.#limitService.errorIfWouldGoOverLimit('emails', {addedCount});
+        }
+
+        // Check email verification required
+        // We need to check this inside the job again
```
naz, line 86: It's a bit of an odd comment here because it's the same one as "outside" the job. Maybe worth rephrasing it in a way that explains the check here is an extra one to run the check on per-job bases - separate to the one done before scheduling jobs?

naz (approved):
This looks good 👌 Good to see the test coverage going up to 100% and some of the hacky module exports being cleaned up at the same time!

ghost/email-service/lib/email-service.js
SimonBackx: You are right, this comment doesn't make sense 🤦🏻‍♂️

ghost/verification-trigger/package.json
SimonBackx: 💃🪩

allouis (approved):

PR #16064, opened 2023-01-03
🐛 Fixed batches can have an empty "to" field

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2246

This solution adds some retries when fetching the recipients for a batch. For an unknown reason the recipients can be empty (while they aren't in the database). This should fix the issue for now until we find more information about the root cause.


rshbhgrg (approved):
Good idea and LGTM 🎉

PR #16068, opened 2023-01-04
Resolved issue where we redundantly created Customer objects in Stripe. Fixes #16057.

### LINKED ISSUE #16057: Ghost appears to redundantly create Stripe Customer objects via API
### Issue Summary

While this isn't high impact, it appears that checkouts initiated in Ghost to create new subscriptions will redundantly create customer objects in Stripe, with one copy having only a name/email address and the other copy having the actual subscription inside of it. This will pollute search results and make customer counts inaccurate.

Note here, for example, how the count of new customers is almost 2X the count of new successful payments.

スクリーンショット 2023-01-03 10 25 20

I believe I've figured out what is happening via Ghost's invocation of Stripe APIs. Here's what the history of API calls looks like for a single new subscriber:

スクリーンショット 2023-01-03 10 17 43

We see Ghost pre-creating a Customer object prior to creating a Checkout session, which is one valid way to do it. However, the call to create the checkout session (POST /v1/checkout/sessions) does not reference the created Customer object. This results in Stripe creating a second Customer "at need", per the description [here](https://stripe.com/docs/api/checkout/sessions/create)

> customer: ... If blank for Checkout Sessions in payment or subscription mode, Checkout will create a new Customer object based on information provided during the payment flow.

The fix is simple, I think: cause the /v1/checkout/sessions POST API call to contain the customer ID, which will look something like cust_... , in the request.

There's some cleanup which could be done relatively trivially by Ghost on users' behalf, or in the alternative, you could release a snippet for users to query their Stripe API, search for the empty customers that duplicate live customers, and delete the empties. (I might write a script for this.)

### Steps to Reproduce

I think this should be observable for any Ghost instance making new subscriptions against Stripe.

1. Connect Ghost to Stripe.
2. Start a new paid subscription using a new email address.
3. Wait a few seconds for bookkeeping to happen.
4. Check the associated Stripe account; you will likely see two new customer objects, one with only the name/email and the other with the name/email and associated subscription/payment method/etc.

### Ghost Version

Current version via Ghost Pro hosting

### Node.js Version

Current version via Ghost Pro hosting

Current version via Ghost Pro hosting

### Database type

MySQL 5.7

### Browser & OS version

_No response_

_No response_

### Code of Conduct

patio11:
Fixes #16057 and adds some tests.

Briefly, Ghost created two Customer objects via the Stripe API when an existing subscriber would upgrade to a paid subscription, one in an API call to create the Customer and then a second as a side effect of an API call to create a Checkout session for the user. The fix is passing the reference to the Customer object to the API call to create the Checkout session; Stripe will no long redundantly create a Customer object in this case.

This largely impacts the owner's experience of the Stripe Dashboard; it will correct their new Customer count (going forward) and make searches for users by name or email address return one responsive object which has the actual subscription in it versus returning two and forcing them to look in each to e.g. refund a transaction or similar. 

I've verified this behavior locally against the test version of the API:

スクリーンショット 2023-01-04 11 44 29

As you can see, after switching to the branch with the fix, we get the expected single user created; prior to applying the fix we get two users created per checkout.

## Code hygiene and correctness issues

* I have tested manually, as above. 
* I have verified that no files I change have issues with lint.
* I have verified that all tests pass, via yarn test. This includes tests which I added.

I have reasonably strong confidence that I have not broken pre-existing behavior, mostly as a result of adding unit tests which exercise that pre-existing behavior, in the case where we are creating the customer and the checkout session atomically (AFAIK, this happens when a user who is not subscribed to the newsletter elects a paid subscription immediately).

### DIFF
--- a/ghost/stripe/lib/StripeAPI.js
+++ b/ghost/stripe/lib/StripeAPI.js
@@ -364,8 +364,9 @@ module.exports = class StripeAPI {
      */
     async createCheckoutSession(priceId, customer, options) {
         const metadata = options.metadata || undefined;
+        const customerId = customer ? customer.id : undefined;
         const customerEmail = customer ? customer.email : options.customerEmail;
-        const customerId = customer ? customer.id : null;
+ 
         await this._rateLimitBucket.throttle();
         let discounts;
         if (options.coupon) {
@@ -387,12 +388,11 @@ module.exports = class StripeAPI {
             delete subscriptionData.trial_from_plan;
             subscriptionData.trial_period_days = options.trialDays;
         }
-        const session = await this._stripe.checkout.sessions.create({
+
+        let stripeSessionOptions = {
             payment_method_types: ['card'],
             success_url: options.successUrl || this._config.checkoutSessionSuccessUrl,
             cancel_url: options.cancelUrl || this._config.checkoutSessionCancelUrl,
-            customer: customerId,
-            customer_email: customerEmail,
             // @ts-ignore - we need to update to latest stripe library to correctly use newer features
             allow_promotion_codes: discounts ? undefined : this._config.enablePromoCodes,
             metadata,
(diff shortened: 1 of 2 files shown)


allouis:
@patio11 Thanks very much for the detailed issue/PR and adding tests!! This will go out in our next release 💃

PR #16069, opened 2023-01-04
Fixed MailgunEmailSuppressionList adding non-5xx failures to the list

allouis:
no-issue

The current implementation of the suppression list is incorrectly adding emails to the suppression list for permanent failure events which have an error code outside of the 5xx range.

### DIFF
--- a/ghost/core/core/server/services/email-suppression-list/MailgunEmailSuppressionList.js
+++ b/ghost/core/core/server/services/email-suppression-list/MailgunEmailSuppressionList.js
@@ -96,12 +96,15 @@ class MailgunEmailSuppressionList extends AbstractEmailSuppressionList {
 
     async init() {
         const handleEvent = reason => async (event) => {
-            if (reason === 'bounce') {
-                if (event.error.code < 500 || event.error.code > 599) {
-                    return;
-                }
-            }
             try {
+                if (reason === 'bounce') {
+                    if (!Number.isInteger(event.error?.code)) {
+                        return;
+                    }
+                    if (event.error.code < 500 || event.error.code > 599) {
+                        return;
+                    }
+                }
                 await this.Suppression.add({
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/services/email-suppression-list/MailgunEmailSuppressionList.js
```diff
@@ -96,6 +96,11 @@ class MailgunEmailSuppressionList extends AbstractEmailSuppressionList {
 
     async init() {
         const handleEvent = reason => async (event) => {
+            if (reason === 'bounce') {
+                if (event.error.code < 500 || event.error.code > 599) {
```
ErisDS, line 100: Do we definitely always have event.error.code?

SimonBackx (approved):

ghost/core/core/server/services/email-suppression-list/MailgunEmailSuppressionList.js
```diff
@@ -96,6 +96,11 @@ class MailgunEmailSuppressionList extends AbstractEmailSuppressionList {
...
+                    return;
+                }
+            }
```
SimonBackx, line 103: Can we wrap this inside the try catch to avoid unhandled promises too?

ghost/core/core/server/services/email-suppression-list/MailgunEmailSuppressionList.js
allouis: Ah okay, yeah we should deffo update the JSDoc to reflect reality ☺️

ghost/core/core/server/services/email-suppression-list/MailgunEmailSuppressionList.js
allouis: Sure 👍

PR #16070, opened 2023-01-04
Added migrations to drop and recreate the suppressions table

allouis:
There are currently two issues with the suppressions table:
  - We have some incorrect rows
  - We have missing UNIQUE constraints

We want to completely wipe the tables and start fresh, as well as make
sure that the UNIQUE constraints are added, so we drop the table
completely, and then re-add it, which should result in an empty
suppressions table with all expected constraints.

We're also interested in renaming the `email_address` column to `email`
this has been done in a separate commit so that we can change our minds.


ghost/core/core/server/data/migrations/versions/5.27/2023-01-04-04-12-add-suppressions-table.js
```diff
@@ -0,0 +1,9 @@
+const {addTable} = require('../../utils');
```
daniellockyer, line 1: You need to alter the timestamp on this because it's currently ordered before the drop table

allouis: 🤣 I guess minutes wasn't precise enough

daniellockyer (approved):

lenabaidakova (approved):

allouis:
Force pushed to fix integrity test

PR #16071, opened 2023-01-04
✨ Added redeemed offers filtering for members

ronaldlangeveld:
Adds the ability to filter members based on the offer(s) they redeemed when subscribing.

ref https://github.com/TryGhost/Team/issues/2011


ghost/core/test/regression/models/model_members.test.js
```diff
@@ -503,6 +505,74 @@ describe('Member Model', function run() {
...
+                offer_id: offerModel.get('id'),
+                member_id: member.get('id'),
+                created_at: new Date(),
+                subscription_id: subscription.get('id')
+            }, context);
+
+            const offerId = addRedemption.get('offer_id');
+
+            const members = await Member.findPage({filter: `offer_redemptions:${offerId}`});
+            should.equal(members.data.length, 1, 'Can search for members with offer redemptions');
```
allouis, line 574: I think we should check that the specific member you added in this test is the one that is returned, maybe update the email address to something like `offerredemptionfilteringtest@whatevber.comsdlkcjkl`

allouis (approved):
Looks good from backend side - one comment about improving test a little

ghost/core/test/regression/models/model_members.test.js
ronaldlangeveld: Done 😊

ghost/admin/app/components/gh-members-segment-select.js
```diff
@@ -120,6 +120,26 @@ export default class GhMembersSegmentSelect extends Component {
             options.push(labelsGroup);
         }
 
+        const offers = yield this.store.query('offer', {limit: 'all'});
+
+        if (offers.length > 0 && !this.args.hideOffers) {
```
kevinansfield, line 125: Can the `if (!this.args.hideOffers)` conditional be used to avoid waiting on the API call when it's not needed?

ghost/admin/app/components/gh-member-settings-form.js
```diff
@@ -23,6 +23,7 @@ export default class extends Component {
     @tracked showMemberTierModal = false;
     @tracked tiersList;
     @tracked newslettersList;
+    @tracked offersList;
```
kevinansfield, line 26: Where is this used?

ghost/admin/app/templates/offers.hbs
```diff
@@ -92,9 +92,9 @@
                                     <span class="midgrey strike ml2">{{currency-symbol offer.finalCurrency}}{{gh-price-amount offer.originalPrice}}</span>
                                 {{/if}}
                             </LinkTo>
-                            <LinkTo @route="offer" @model={{offer}} class="gh-list-data">
+                            <a href="#" {{on "click" (fn this.offerMembers offer.id)}}  class="gh-list-data pointer">
```
kevinansfield, line 95: This can stay as a `<LinkTo>` so the behaviour is kept in the template and easily visible, query params can be set like this:

```
<LinkTo @route="members" @query={{hash filter=(concat "offer_redemptions:" offer.id)}} class="gh-list-data">
```

kevinansfield (changes requested):
A few inline comments on small things plus two other areas that need addressing:

1. the `<DefaultRecipientsSelect>` component was updated to include offers but the associated `<GhMembersRecipientSelect>` component wasn't so it's possible to set a default recipient segment that isn't selectable when publishing (these components should really be refactored to share behaviour if the design allows but that's a different task)
2. there are no Admin tests 😬

PR #16075, opened 2023-01-04
Added DomainEvents.allSettled utility method

SimonBackx:
no issue

With the increased usage of DomainEvents, it gets harder to build reliable tests without having to resort to timeouts. This utility method allows us to wait for all events to be processed before continuing with the test.

This change should speed up tests and make them more reliable.

It only adds extra code when running tests and shouldn't impact production.


naz (approved):
This is a great find and the approach is similar we had in the jobs manager I believe? Noice :ok_hand:

SimonBackx:
For the job manager I'm thinking to rewrite the awaitCompletion in the future to not always require a name, so we can just await 'all' like here.

PR #16080, opened 2023-01-05
Handled EmailBounceEvent with 605 error code

allouis:
When Mailgun fails to deliver an email to an address because the
address has already bounced before, it gives us a permanent fail event
with a 605 error code rather than a 5xx one. Because we want to
"backfill" our suppressions data with previously bounced email
addresses, we want to handle this specific error code.

We may update this logic in the future based on new information from
Mailgun with respect to their 6xx error codes and the
meanings/underlying cause of theme.

This also moves the tests which check for whether or not emails are
suppressed into their own fail so that we do not pollute the event
storage tests, and adds more tests cases.


ghost/core/core/server/services/email-suppression-list/MailgunEmailSuppressionList.js
```diff
@@ -101,7 +101,7 @@ class MailgunEmailSuppressionList extends AbstractEmailSuppressionList {
                     if (!Number.isInteger(event.error?.code)) {
                         return;
                     }
-                    if (event.error.code < 500 || event.error.code > 599) {
+                    if (event.error.code < 500 || event.error.code > 599 && event.error.code !== 605) {
```
allouis, line 104: This is the actual logic change

ghost/core/test/integration/services/email-service/email-event-storage.test.js
```diff
@@ -209,9 +209,6 @@ describe('EmailEventStorage', function () {
         const providerId = emailBatch.provider_id;
         const timestamp = new Date(2000, 0, 1);
 
-        const {body: {members: [member]}} = await agent.get(`/members/${memberId}`);
```
allouis, line 212: I've removed the checks from this test in favor of moving them into their own test file below

PR #16081, opened 2023-01-05
Added visible theme errors in admin

SimonBackx:
refs https://github.com/TryGhost/Team/issues/2393

- During boot and loading the active theme, we now cache the result of the gscan validation. Cache configuration can happen in `adapters.cache.gscan`
- We now also return non-fatal errors when activating or adding a theme.
- When the `themeErrorsNotification` feature flag is on, we fetch the active theme (which includes the validation information) when loading admin
- If the currently active theme has errors, we show an error notification that can open the error modal
- Added a new endpoint: `/ghost/api/admin/themes/active/` that returns the result of the last gscan validation of the active theme. If no cache is available, it will run a new gscan validation.
- Added new permissions for the active action/endpoint (author, editor, administrator)


daniellockyer (approved):

ghost/core/core/server/api/endpoints/themes.js
```diff
@@ -15,6 +16,18 @@ module.exports = {
         }
     },
 
+    active: {
+        headers: {
+            cacheInvalidate: true
```
naz, line 21: I'm confused, why would we want to invalidate cache on a GET request 🤔

ghost/core/core/server/api/endpoints/themes.js
```diff
@@ -15,6 +16,18 @@ module.exports = {
         }
     },
 
+    active: {
```
naz, line 19: We are doing an "active" here instead of "read" because for the client it would be hard to find out which theme is currently active? We don't usually name controller functions by the "filter", so not sure if this naming is necessary?

SimonBackx: More because we only have cached data available for the active theme. But thinking about it further we could change it to a read action with a filter. It would just run a new gscan report on the theme if it is not asked.

ghost/core/core/server/services/themes/validate.js
```diff
@@ -41,16 +60,57 @@ const check = async function check(theme, isZip) {
...
 
+    // In production we don't want to show warnings
+    // Warnings are meant for developers only
+    if (config.get('env') === 'production') {
+        checkedTheme.results.warning = [];
+    }
+
+    // Cache the result
+    try {
+        await gscanCacheStore.set(themeName, checkedTheme);
```
naz, line 75: Have you looked into the size of the "checkedTheme" object? I do remember we shove quite a lot of information there, not too crazy but there might be a bit of a memory overhead added to what the instance normally uses. Did you run napkin math of how much that would be just to be sure nothing horrific is happening?

ghost/core/core/server/services/themes/validate.js
```diff
@@ -5,20 +5,39 @@ const config = require('../../../shared/config');
...
+ */
+
+/**
+ * The cache store for storing the result of the last theme validation
+ * @type {CacheStore}
+ */
+let gscanCacheStore;
+
+module.exports.init = () => {
+    gscanCacheStore = adapterManager.getAdapter('cache:gscan');
```
naz, line 33: I think this way of initializing the adapter is fine for now, given the time constraint. But we should leave a comment and open an issue to redo this initialization into calss with "cacheStore" passed in as a parameter. Validator should now know anything about adapter manager ideally.

I see we only call "init" once anyway, maybe worth passing `gscanCacheStore` as a parameter straight away.

SimonBackx: Oh good to know! I wasn't sure about this as I only found usages of `adapterManager.getAdapter('cache:*')`

ghost/core/core/server/api/endpoints/themes.js
SimonBackx: And it would throw a validation error if the requested theme is not loaded, but I think they need to be loaded and that is probably an edge case right now

ghost/core/core/server/api/endpoints/themes.js
SimonBackx: My 🅱️

PR #16094, opened 2023-01-10
🐛 Fixed staff token can't export DB

urish:
Allows staff tokens to make backups of the DB by issuing a GET request to the `/db` API endpoint. My use case is automatic daily backups (from a GitHub action).

p.s. it might be worth noting that `yarn test:all` should be run inside `ghost/core`, it took me a white to figure that out.

### DIFF
--- a/ghost/core/core/server/web/api/endpoints/admin/middleware.js
+++ b/ghost/core/core/server/web/api/endpoints/admin/middleware.js
@@ -35,7 +35,7 @@ const notImplemented = function (req, res, next) {
         schedules: ['PUT'],
         files: ['POST'],
         media: ['POST'],
-        db: ['POST'],
+        db: ['GET', 'POST'],
         settings: ['GET']
     };
 


markstos:
I agree about the note about `yarn test:all`-- as a fellow ghost user, that tripped me up too.

I can see how supporting DB exports via the API could be convenient for what you describe, but it also brings added risks because it makes it easier to ex-filtrate personally identifiable information (PII), including names and emails.

An alternate way to make daily backups is run a cron job or other daily timer which backups up the MySQL database and uploads that somewhere. Or, if you are using Cloud hosting, you have the option to trigger daily snapshots.

Finally, you can reverse engineer how the web-based backups created by admins are authenticated and use your same kind of pattern by setting those auth headers. It seems the admin cookies may expire after a few months, so if you attempt this you may need some way to refresh the admin cookie periodically.

Because this is rarely requested feature and because it comes with additional data privacy risk, I recommend against merging this PR.

ErisDS:
Hey, I'm sorry we didn't address this PR earlier. I can confirm the existing behaviour is intentional. I don't disagree with your desire for doing daily automated backups, and that we should have a way to support that, I'm not convinced this change is the right way to achieve it.

I think others use the `ghost backup` command in Ghost-CLI. 

If you're running in production, I'd recommend using a mysql-specific tool for backing up your DB, and doing something else to keep a copy of the `/content` folder. The backup command and endpoint were all designed before Ghost had several of the more advanced member-driven features, and if you're using those you'll find they only backup content, not everything.

PR #16114, opened 2023-01-13
🔥 Removed support for {{lang}} helper

naz:
refs https://github.com/TryGhost/Toolbox/issues/497

- Removes {{lang}} helper which has been completely deprecated since Ghost 5.0. When using the helper in themes with 5.x backend the output will be empty


naz:
landed in main as https://github.com/TryGhost/Ghost/commit/c2b18f0c87b659043c08c8e7b4c3c3c43fe54027

PR #16117, opened 2023-01-13
🐛 Fixed storing original files for images

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/481

This change fixes an issue when multiple images with the same name are uploaded in parallel. The current system does not guarantee that the original filename is stored under NAME+`_o`, because the upload for the original file and the resized file are happening in parallel.

Solution:
- Wait for the storage of the resized image (= the image without the _o suffix) before storing the original file.
- When that is stored, use the generated file name of the stored image to generate the filename with the _o suffix. This way, it will always match and we don't risk both files to have a different number suffix. We'll also set the `targetDir` argument when saving the file, to avoid storing the original file in a different directory (when uploading a file around midnight both files could be stored in 2023/01 and 2023/02). 

Some extra optimisations needed with this fix:
- Previously when uploading image.jpg, while it already exists, it would store two filenames on e.g., `image-3.jpg` and `image_o-3.jpg`. Note the weird positioning of `_o`. This probably caused bugs when uploading files named `image-3.jpg`, which would store the original in `image-3_o.jpg`, but this original would never be used by the handle-image-sizes middleware (it would look for `image_o-3.jpg`). This fix would solve this weird naming issue, and make it more consistent. But we need to make sure our middlewares (including handle-image-sizes) will be able to handle both file locations to remain compatible with the old format. This isn't additional work, because it would fix the old bug too.
- Prevent uploading files that end with `_o`, e.g. by automatically stripping that suffix from uploaded files. To prevent collisions.

Advantage(s):
- We keep the original file name, which is better for SEO.
- No changes required to the storage adapters.

Downside(s):
- The storage of both files wouldn't happen parallel any longer. But I expect the performance implications to be minimal.
- We'll probably need some changes to the routing to make this work (now the uploading in the endpoint is separated from the normalize middleware). Should be manageable.


ghost/core/test/e2e-api/admin/posts.test.js
```diff
@@ -108,16 +108,20 @@ describe('Posts API', function () {
     });
 
     describe('Create', function () {
-        it('Can create a post with mobiledoc', async function () {
```
SimonBackx, line 111: Remove debug

ghost/core/test/e2e-api/admin/posts.test.js
```diff
@@ -108,16 +108,20 @@ describe('Posts API', function () {
...
 
             await agent
-                .post('/posts/?formats=mobiledoc,lexical,html')
-                .body({posts: [post]})
+                .post('/posts/?formats=mobiledoc,lexical,html', {
+                    headers: {
+                        'content-type': 'application/json'
+                    }
+                })
+                .body(JSON.stringify({posts: [post]}))
```
SimonBackx, line 124: Some debugging for the express test changes that need to be removed

kevinansfield (approved):

PR #16119, opened 2023-01-13
Updated Stripe Checkout payment types support

JohnONolan:
No issue
Stripe Checkout no longer requires specifying payment types in code, this is now managed in Stripe Dashboard.
Docs: https://stripe.com/docs/payments/dashboard-payment-methods

Needs testing before release


allouis (changes requested):
We need to update our handling of the `checkout.session.completed` webhook to only run the existing code if the payment is successful, we're also going to need to add handling for the new webhook `checkout.session.async_payment_succeeded` and possibly
`checkout.session.async_payment_failed` - depending on our approach to how we work with async payments

I _think_ the simplest approach we can take is:
- Ignore `checkout.session.completed` webhooks if payment is not complete.
- Handle `checkout.session.async_payment_succeeded` webhooks the same as `checkout.session.completed` as if the payment was complete.
- Ignore `checkout.session.async_payment_failed` webhooks

Based on the docs that should work - but I haven't dived deep into this, will deffo need some testing!

JohnONolan:
@allouis would you be able to take this in a cleanup week maybe?

allouis:
Sure! I've added this to the cleanup board 💃

prschulz:
@allouis - This got closed automatically. Did you create another issue for it? Or make progress here? Trying to shape what needs to happen from here for the Core team.

allouis:
@prschulz I didn't create another issue for it - the PR itself was adding to the product backlog project - but looks like this issue https://github.com/TryGhost/Team/issues/1264 will end up covering these changes.

What needs to happen is an investigation into how to handle async payment methods, to check the approach described above will work, and then to implement those changes, or come up with a different solution based on the investigation

prschulz:
Cool beans - thank you, @allouis. I've updated the other issue [here](https://github.com/TryGhost/Team/issues/1264#issuecomment-1464057158) to make it actionable.

PR #16127, opened 2023-01-16
Added email content testing capabilities

naz:
refs https://github.com/TryGhost/Team/issues/2691

- Outgoing emails have been a weak point of Ghost's stability recently. The concept of "emailMockReceiver" similarly to "webhookMockReceiver", allows to test side-effects like outgoing emails.
- This is a first iteration which should lay groundwork for testing all outgoing emails in the future
- The change adds a new concept of "email mock receiver" which is very similar to how the "webhook mock receiver" works. The email mock receiver exposes two methods to record and verify snapshots:
- matchHTMLSnapshot - records and verifies only the HTML content of the outgoint email
- matchMetadataSnapshot - records and verifies all the non-HTML properties sent along an email content, e.g.: to address, plaintext, subject, etc.
- What's missing is matching content based on dynamic content like dates, links with JWT tokens, etc.

### DIFF
--- a/ghost/core/package.json
+++ b/ghost/core/package.json
@@ -201,8 +201,7 @@
   },
   "devDependencies": {
     "@playwright/test": "1.29.2",
-    "@tryghost/express-test": "0.11.8",
-    "@tryghost/jest-snapshot": "0.5.0",
+    "@tryghost/express-test": "0.12.0",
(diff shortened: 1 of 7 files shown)


naz:
The bump of the snapshot-manager packager  - a single instance is now used across Ghost, is causing issues. Investigating.

PR #16130, opened 2023-01-16
add webmentions feature flag

9larsons:
refs #2406
Adding the alpha feature flag so we can begin testing.


SimonBackx:
Since this is already merged in main, I'm closing this to avoid confusion! Changes of the feature flag look good 👌

PR #16137, opened 2023-01-17
Added logging to webmention endpoint

SimonBackx:
closes https://github.com/TryGhost/Team/issues/2424

Adds the new webmention endpoint, responds with the correct statuscode and empty body at `https://site.ghost/webmentions/receive`.

Also updates the head link tag to the new endpoint location. We don't use /webmention because that could conflict with post slugs.

### DIFF
--- a/ghost/core/core/frontend/helpers/ghost_head.js
+++ b/ghost/core/core/frontend/helpers/ghost_head.js
@@ -86,7 +86,7 @@ function getSearchHelper(frontendKey) {
 function getWebmentionDiscoveryLink() {
     try {
         const siteUrl = urlUtils.getSiteUrl();
-        const webmentionUrl = new URL('webmention', siteUrl);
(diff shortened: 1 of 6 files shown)


ghost/core/core/frontend/helpers/ghost_head.js
```diff
@@ -86,7 +86,7 @@ function getSearchHelper(frontendKey) {
 function getWebmentionDiscoveryLink() {
     try {
         const siteUrl = urlUtils.getSiteUrl();
-        const webmentionUrl = new URL('webmention', siteUrl);
+        const webmentionUrl = new URL('webmentions/receive', siteUrl);
```
daniellockyer, line 89: @SimonBackx I think we need a trailing slash on the end of this because Ghost adds slashes, and it'll result in a 301 otherwise

daniellockyer: Never mind, my testing was wrong

PR #16138, opened 2023-01-17
Added `Content-Version` header to all API requests

daniellockyer:
refs https://github.com/TryGhost/Team/issues/2400

- we've deemed it useful to start to return `Content-Version` for all API requests, because it becomes useful to know which version of Ghost a response has come from in logs
- this should also help us detect Admin<->Ghost API mismatches, which was the cause of a bug recently (ref'd issue)


naz (approved):
No major concerns 👌  Having this header in response for every request, would also make the behavior consistent with how webhooks send this header with every payload.

PR #16139, opened 2023-01-17
Added email snapshots for `API versioning` tests

daniellockyer:
refs https://github.com/TryGhost/Team/issues/2691

- this switches the test over to using email snapshots, which should test more of our output than manually checking certain strings exist

### DIFF
--- a/ghost/core/test/e2e-api/shared/version.test.js
+++ b/ghost/core/test/e2e-api/shared/version.test.js
@@ -10,6 +10,7 @@ const settingsMatcher = {
 describe('API Versioning', function () {
     describe('Admin API', function () {
         let agentAdminAPI;
+        let emailMockReceiver;
 
         before(async function () {
             agentAdminAPI = await agentProvider.getAdminAPIAgent();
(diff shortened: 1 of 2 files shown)


naz:
Well that was an easy migration! Nice 👌

naz (approved):

ghost/core/test/e2e-api/shared/version.test.js
```diff
@@ -129,11 +130,9 @@ describe('API Versioning', function () {
                     }]
                 });
 
-            mockManager.assert.sentEmailCount(1);
-            mockManager.assert.sentEmail({
-                subject: 'Attention required: Your Zapier integration has failed',
-                to: 'jbloggs@example.com'
-            });
+            emailMockReceiver.sentEmailCount(1);
```
naz, line 133: This bit actually makes a good point to stop using the `mockManager.assert` bit in favor of `emailMockReceiver.sentEmailCount`. I've pushed deprecation note into the mock manager to make it extra clear

daniellockyer: Cool!

PR #16146, opened 2023-01-18
Added outbound link tagging setting

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2432
    
Adds outbound_link_tagging setting (enabled by default and behind feature flag). If the feature flag is enabled, and the setting is disabled, we won't add ?ref to links in emails.
    
This includes new E2E tests for email click tracking, which were also extended to check outbound link tagging (for both MEGA and the new email stability flow).

Also fixes a test fixture for the comments_enabled setting.


ghost/core/test/e2e-api/admin/settings.test.js
```diff
@@ -7,7 +7,7 @@ const {stringMatching, anyEtag, anyUuid, anyContentLength, anyContentVersion} =
 const models = require('../../../core/server/models');
 const {anyErrorId} = matchers;
 
-const CURRENT_SETTINGS_COUNT = 69;
+const CURRENT_SETTINGS_COUNT = 71;
```
SimonBackx, line 10: +2 because comments_enabled was missing in fixtures

allouis (approved):

daniellockyer (approved):

PR #16150, opened 2023-01-19
Added mentions database table

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/2417

- started adding a schema for the mentions db table
- includes a migration and a bookshelf model


ghost/core/core/server/models/mention.js
```diff
@@ -0,0 +1,18 @@
+const ghostBookshelf = require('./base');
+
+const Mention = ghostBookshelf.Model.extend({
+    tableName: 'mentions',
+
+    targetId: function targetId() {
```
allouis, line 6: This should be renamed to `post` I think, but we can also leave the relation out of this PR so that we can get it merged and started on the repository implementation

allouis (approved):

ghost/core/core/server/data/migrations/versions/5.31/2023-01-19-07-46-add-mentions-table.js
```diff
@@ -0,0 +1,23 @@
...
+    source_site_title: {type: 'string', maxlength: 2000, nullable: true},
+    source_excerpt: {type: 'string', maxlength: 2000, nullable: true},
+    source_author: {type: 'string', maxlength: 2000, nullable: true},
+    source_featured_image: {type: 'string', maxlength: 2000, nullable: true},
+    source_favicon: {type: 'string', maxlength: 2000, nullable: true},
+    target: {type: 'string', maxlength: 2000, nullable: false},
+    resource_id: {type: 'string', maxlength: 24, nullable: true},
+    resource_type: {type: 'string', maxlength: 50, nullable: true},
+    created_at: {type: 'dateTime', nullable: false},
+    payload: {type: 'text', maxlength: 1000000000, nullable: true}
```
allouis, line 17: This seems way too big for the `payload` - we need to cap this to a reasonable size IMO something in the thousands makes more sense to me - but we should check with Nick & Peter

ghost/core/core/server/models/mention.js
```diff
@@ -0,0 +1,18 @@
...
+    }
+});
+
+const Mentions = ghostBookshelf.Collection.extend({
+    model: Mention
+});
+
+module.exports = {
+    Mention: ghostBookshelf.model('Mention', Mention),
+    Mentions: ghostBookshelf.collection('Mentions', Mentions)
```
allouis, line 17: We don't need this or the definition above

ghost/core/core/server/data/schema/schema.js
```diff
@@ -979,5 +979,20 @@ module.exports = {
...
+        source_site_title: {type: 'string', maxlength: 2000, nullable: true},
+        source_excerpt: {type: 'string', maxlength: 2000, nullable: true},
+        source_author: {type: 'string', maxlength: 2000, nullable: true},
+        source_featured_image: {type: 'string', maxlength: 2000, nullable: true},
+        source_favicon: {type: 'string', maxlength: 2000, nullable: true},
+        target: {type: 'string', maxlength: 2000, nullable: false},
+        resource_id: {type: 'string', maxlength: 24, nullable: true},
+        resource_type: {type: 'string', maxlength: 50, nullable: true},
+        created_at: {type: 'dateTime', nullable: false},
+        payload: {type: 'text', maxlength: 65535, nullable: true}
```
allouis, line 996: I'm still not convinced about the size of this 🤔 Seems massive but maybe we want that

ronaldlangeveld: For now I followed the guidelines on top of [schema.js](https://github.com/TryGhost/Ghost/blob/45d65663f4af83cb4233a85d6555075298a34863/ghost/core/core/server/data/schema/schema.js#L7).

ghost/core/core/server/data/migrations/versions/5.31/2023-01-19-07-46-add-mentions-table.js
```diff
@@ -0,0 +1,23 @@
+const {addTable} = require('../../utils');
+
+// stay below 
+const mentionsTable = addTable('mentions', {
```
SimonBackx, line 4: Any reason why we didn't go for `module.exports = addTable(...);`

ghost/core/core/server/data/migrations/versions/5.31/2023-01-19-07-46-add-mentions-table.js
```diff
@@ -0,0 +1,23 @@
...
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    source: {type: 'string', maxlength: 2000, nullable: false},
+    source_title: {type: 'string', maxlength: 2000, nullable: true},
+    source_site_title: {type: 'string', maxlength: 2000, nullable: true},
+    source_excerpt: {type: 'string', maxlength: 2000, nullable: true},
+    source_author: {type: 'string', maxlength: 2000, nullable: true},
```
SimonBackx, line 10: Maybe `2000` is a bit over the top for author? Our internal limit for author names is `191` for our own posts

SimonBackx: But not so important I think 🤷‍♂️

SimonBackx: + you followed the spec above

SimonBackx (approved):

ghost/core/core/server/data/migrations/versions/5.31/2023-01-19-07-46-add-mentions-table.js
daniellockyer: Yeah can we switch to the suggestion above? 🙂

ghost/core/core/server/data/schema/schema.js
allouis: But do we want to store it as text, or as a "large string" like 1000-2000 in size 🤔

allouis: Let's just go with this for now 👍

daniellockyer (approved):

PR #16151, opened 2023-01-19
Added MentionSendingService

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2409

The MentionSendingService listens for post changes and sends webmentions for outbound links in the post.

### DIFF
--- a/ghost/webmentions/lib/MentionSendingService.js
+++ b/ghost/webmentions/lib/MentionSendingService.js
@@ -1,0 +1,164 @@
+const errors = require('@tryghost/errors');
+const logging = require('@tryghost/logging');
+
+class MentionSendingService {
+    #discoveryService;
+    #externalRequest;
+    #getSiteUrl;
+    #getPostUrl;
+    #isEnabled;
+
+    constructor({discoveryService, externalRequest, getSiteUrl, getPostUrl, isEnabled}) {
+        this.#discoveryService = discoveryService;
+        this.#externalRequest = externalRequest;
+        this.#getSiteUrl = getSiteUrl;
+        this.#getPostUrl = getPostUrl;
+        this.#isEnabled = isEnabled;
+    }
+
+    get siteUrl() {
+        try {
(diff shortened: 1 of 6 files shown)


ghost/webmentions/test/MentionSendingService.test.js
```diff
@@ -0,0 +1,440 @@
...
+        //        source: new URL('https://example.com'),
+        //        target: new URL('https://example.com'),
+        //        endpoint: new URL('https://example.org/webmentions-test')
+        //    });
+        //    assert(scope.isDone());
+        //    assert(scope2.isDone());
+        //});
+        // TODO: also check if we don't follow private IPs after redirects
+
+        it('Does not send to private IP', async function () {
```
9larsons, line 420: Should we comment these out until the `MentionSendingService` is connected?

ghost/webmentions/lib/MentionSendingService.js
```diff
@@ -0,0 +1,163 @@
+const errors = require('@tryghost/errors');
+const logging = require('@tryghost/logging');
+
+module.exports = class MentionSendingService {
+    #discoveryService;
```
55sketch, line 5: out of interest, I've not seen the `#` pattern before, what is it?

SimonBackx: # marks a private attribute or method, refs https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

We used to also have `_` before because Node only supports this from 12.0. Sometimes we also just mark something as `@private` with JSDocs. Sometimes we cannot really use `#` because those methods are also not accessible for unit testing (haven't found a way around that yet)

ghost/webmentions/test/MentionSendingService.test.js
SimonBackx: As discussed, these tests can remain here 😊

PR #16154, opened 2023-01-19
Add mention discovery service

9larsons:
fixes https://github.com/TryGhost/Team/issues/2407

The MentionDiscoveryService fetches mentioned sites to return the webmention endpoint.

### DIFF
--- a/ghost/webmentions/lib/MentionDiscoveryService.js
+++ b/ghost/webmentions/lib/MentionDiscoveryService.js
@@ -1,0 +1,73 @@
+const cheerio = require('cheerio');
+const errors = require('@tryghost/errors');
+
+module.exports = class MentionDiscoveryService {
+    #externalRequest;
+
+    constructor({externalRequest}) {
+        this.#externalRequest = externalRequest;
+    }
+
+    /**
+     * Fetches the given URL to identify the webmention endpoint
+     * @param {URL} url 
+     * @returns {Promise<URL|null>}
+     */
+    async getEndpoint(url) {
+        try {
+            const response = await this.#externalRequest(url.href, {
+                throwHttpErrors: false,
+                followRedirects: true,
+                maxRedirects: 10,
+            });
+            if (response.statusCode === 404) {
+                throw new errors.BadRequestError({
+                    message: 'Webmention discovery service could not find target site',
+                    statusCode: response.statusCode
+                });
+            }
+            return this.getEndpointFromResponse(response);
+        } catch (error) {
+            return null;
+        }
+    }
+
+    /**
+     * @private
+     * Parses the given response for the first webmention endpoint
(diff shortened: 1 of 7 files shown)


ghost/webmentions/lib/MentionDiscoveryService.js
```diff
@@ -0,0 +1,67 @@
...
+     * Fetches the given URL to identify the webmention endpoint
+     * @param {URL} url 
+     * @returns {Promise<URL|null>}
+     */
+    async getEndpoint(url) {
+        // *TODO* see webmention spec 4.3 & 4.5
+        // -> Make sure we don't fetch locally or in LAN.
+        // -> keep in mind that a DNS domain can also point to a localhost or LAN IP
+        try {
+            const response = await fetch(url.href, {redirect: 'follow'});
```
SimonBackx, line 16: If we swap this out for `got` we'll automatically fix the private IP issue

ghost/webmentions/package.json
```diff
@@ -20,7 +20,11 @@
...
     "c8": "7.12.0",
     "mocha": "10.2.0",
+    "nock": "^13.3.0",
     "sinon": "15.0.1"
   },
-  "dependencies": {}
+  "dependencies": {
+    "cheerio": "^1.0.0-rc.12",
+    "node-fetch": "2"
+  }
```
SimonBackx, line 29: For dependencies we always pin them to a fixed number, and keep those numbers the same across all the packages. So usually I search for `cheerio":` in code to find the current version we use for `cheerio` in a different package, and then set that version in `package.json`.

In this case, nock should be `13.3.0` for example.

ghost/webmentions/package.json
```diff
@@ -20,7 +20,11 @@
   "devDependencies": {
     "c8": "7.12.0",
     "mocha": "10.2.0",
+    "nock": "^13.3.0",
     "sinon": "15.0.1"
   },
-  "dependencies": {}
+  "dependencies": {
+    "cheerio": "^1.0.0-rc.12",
+    "node-fetch": "2"
```
SimonBackx, line 28: We'll need `@tryghost/errors` here too

SimonBackx (changes requested):
Looks good! 🙌 I've left some comments for some minor things

ghost/webmentions/lib/webmentions.js
```diff
@@ -1,3 +1,4 @@
 module.exports.InMemoryMentionRepository = require('./InMemoryMentionRepository');
 module.exports.MentionsAPI = require('./MentionsAPI');
+module.exports.MentionsDiscoveryService = require('./MentionDiscoveryService')
```
SimonBackx, line 3: You have a linting error for missing ; at the end here

PR #16157, opened 2023-01-20
Optimized dynamic routing url updates

naz:
Optimizations around URL recalculation for published post updates.

- Full URL regeneration process was happening even when only unrelated to URL generation fields were updated (e.g. 'plaintext' change in post does not affect the URL of the post). Stopping the  "resource updated" event processing early circumvents full url regeneration inside of DynamicRouting, which can be quite heavy depending on routing configuration
- The URLResourceUpdatedEvent is supposed to be emmited whenever there's an update to the resource already associated with the URL and no url-affecting fields were touched.

The issue left to solve here is sitemaps update with the latest `lastmod` value whenever the post's test is modified.

### DIFF
--- a/ghost/core/core/server/services/url/Resources.js
+++ b/ghost/core/core/server/services/url/Resources.js
@@ -1,5 +1,7 @@
 const _ = require('lodash');
 const debug = require('@tryghost/debug')('services:url:resources');
+const DomainEvents = require('@tryghost/domain-events');
+const {URLResourceUpdatedEvent} = require('@tryghost/dynamic-routing-events');
 const Resource = require('./Resource');
 const config = require('../../../shared/config');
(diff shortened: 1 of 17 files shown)


ghost/core/core/server/services/url/Resources.js
```diff
@@ -297,6 +299,25 @@ class Resources {
 
         const resourceConfig = _.find(this.resourcesConfig, {type: type});
 
+        if (resourceConfig && Object.keys(model._changed).length) {
```
naz, line 302: This code is the 🥩  of the issue. It bypasses the URL update process whenever only non-essential properties were changed.

The 'updated_at' is taken into account separately, so we can emit an event about the resource update, which sitemaps service needs.

PR #16167, opened 2023-01-23
Fixed configUtils and adapter cache issues in E2E tests

SimonBackx:
no issue

There are a couple of issues with resetting the Ghost instance between E2E test files:

These issues came to the surface because of new tests written in https://github.com/TryGhost/Ghost/pull/16117

**1. configUtils.restore does not work correctly**
`config.reset()` is a callback based method. On top of that, it doesn't really work reliably (https://github.com/indexzero/nconf/issues/93)

What kinda happens, is that you first call `config.reset` but immediately after you correcty reset the config using the `config.set` calls afterwards. But since `config.reset` is async, that reset will happen after all those sets, and the end result is that it isn't reset correctly.

This mainly caused issues in the new updated images tests, which were updating the config `imageOptimization.contentImageSizes`, which is a deeply nested config value. Maybe some references to objects are reused in nconf that cause this issue?

Wrapping `config.reset()` in a promise does fix the issue.

**2. Adapters cache not reset between tests**
At the start of each test, we set `paths:contentPath` to a nice new temporary directory. But if a previous test already requests a localStorage adapter, that adapter would have been created and in the constructor `paths:contentPath` would have been passed. That same instance will be reused in the next test run. So it won't read the new config again. To fix this, we need to reset the adapter instances between E2E tests.

How was this visible? Test uploads were stored in the actual git repository, and not in a temporary directory. When writing the new image upload tests, this also resulted in unreliable test runs because some image names were already taken (from previous test runs).

image
*Test images are stored in `ghost/core/content/images` instead of a temporary directory*

**3. Old 2E2 test Ghost server not stopped**
Sometimes we still need access to the frontend test server using `getAgentsWithFrontend`. But that does start a new Ghost server which is actually listening for HTTP traffic. This could result in a fatal error in tests because the port is already in use. The issue is that old E2E tests also start a HTTP server, but they don't stop the server. When you used the old `startGhost` util, it would check if a server was already running and stop it first. The new `getAgentsWithFrontend` now also has the same functionality to fix that issue.

### DIFF
--- a/ghost/core/test/regression/site/frontend.test.js
+++ b/ghost/core/test/regression/site/frontend.test.js
@@ -211,13 +211,12 @@ describe('Frontend Routing', function () {
                 });
 
                 after(function (done) {
-                    configUtils.restore();
-
-                    testUtils.startGhost({forceStart: true})
-                        .then(function () {
-                            request = supertest.agent(config.get('url'));
-                            addPosts(done);
-                        });
+                    configUtils.restore().then(() => {
+                        return testUtils.startGhost({forceStart: true});
+                    }).then(function () {
+                        request = supertest.agent(config.get('url'));
+                        addPosts(done);
+                    });
                 });
 
                 it('should redirect without slash', function (done) {
--- a/ghost/core/test/utils/e2e-framework.js
+++ b/ghost/core/test/utils/e2e-framework.js
@@ -55,8 +55,7 @@ const startGhost = async (options = {}) => {
     // NOTE: need to pass this config to the server instance
     configUtils.set('paths:contentPath', contentFolder);
 
-    // We need to clear the adapter cache because the config changed
-    // And the LocalStorageAdapter is cached, which means that it will still try to store files to the old content path if we don't recreate all adapters
+    // Adapter cache has to be cleared to avoid reusing cached adapter instances between restarts
     adapterManager.clearCache();
 
     const defaults = {
--- a/ghost/core/test/utils/e2e-framework.js
+++ b/ghost/core/test/utils/e2e-framework.js
@@ -34,6 +34,8 @@ const db = require('./db-utils');
 // Services that need resetting
 const settingsService = require('../../core/server/services/settings/settings-service');
 const supertest = require('supertest');
+const {stopGhost} = require('./e2e-utils');
+const adapterManager = require('../../core/server/services/adapter-manager');
 
 /**
  * @param {Object} [options={}]
@@ -52,6 +54,10 @@ const startGhost = async (options = {}) => {
 
     // NOTE: need to pass this config to the server instance
     configUtils.set('paths:contentPath', contentFolder);
+
+    // We need to clear the adapter cache because the config changed
+    // And the LocalStorageAdapter is cached, which means that it will still try to store files to the old content path if we don't recreate all adapters
+    adapterManager.clearCache();
 
     const defaults = {
         backend: true,
@@ -335,6 +341,11 @@ const getAgentsWithFrontend = async () => {
         server: true
     };
     try {
+        // Possible that we still have a running Ghost server from a previous old E2E test
+        // Those tests never stopped the server in the tests manually
+        await stopGhost();
+
+        // Start a new Ghost server with real HTTP listener
         ghostServer = await startGhost(bootOptions);
         const app = ghostServer.rootApp;
 
--- a/ghost/core/test/utils/e2e-utils.js
+++ b/ghost/core/test/utils/e2e-utils.js
@@ -113,6 +113,9 @@ const restartModeGhostStart = async ({frontend, copyThemes, copySettings}) => {
 
     debug('init done');
 
+    // Adapter cache has to be cleared to avoid reusing cached adapter instances between restarts
+    adapterManager.clearCache();
+
     // Reset the settings cache
     await settingsService.init();
     debug('settings done');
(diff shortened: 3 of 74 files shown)


ghost/core/test/regression/site/frontend.test.js
```diff
@@ -211,13 +211,12 @@ describe('Frontend Routing', function () {
...
 
                 after(function (done) {
-                    configUtils.restore();
-
-                    testUtils.startGhost({forceStart: true})
-                        .then(function () {
-                            request = supertest.agent(config.get('url'));
-                            addPosts(done);
-                        });
+                    configUtils.restore().then(() => {
```
naz, line 214: We can use the async/await syntax here like in all the other changes?

ghost/core/test/utils/e2e-utils.js
```diff
@@ -167,6 +172,10 @@ const freshModeGhostStart = async (options) => {
 
     await settingsService.init();
 
+    // We need to clear the adapter cache because the config changed
+    // And the LocalStorageAdapter is cached, which means that it will still try to store files to the old content path if we don't recreate all adapters
+    adapterManager.clearCache();
```
naz, line 177: Few things here:
1. Probably best to place the clear cache call above the "settingsService.init" just in case when settings service starts using adapters - cache adapter for example. Wdyt?

2. The comment provides too many details about the "why" we are doing the "clearCache" 🙃  When looking at it out of context I would need to figure out how LocalStorageAdapter is related here specifically etc.  I think just saying that the "adapter cache has to be cleared to avoid reusing cached adapter instances between restarts" would be enough 👍

ghost/core/test/utils/urlUtils.js
```diff
@@ -53,7 +53,8 @@ const stubUrlUtilsFromConfig = () => {
 
 const restore = () => {
     defaultSandbox.restore();
-    configUtils.restore();
+    // eslint-disable-next-line no-console
+    configUtils.restore().catch(console.error);
```
naz, line 57: Was the "restore()" method actually erroring a lot? 

The change looks good, just asking out of interest. If it did what was the cause?

ghost/core/test/utils/e2e-framework.js
```diff
@@ -335,6 +341,11 @@ const getAgentsWithFrontend = async () => {
         server: true
     };
     try {
+        // Possible that we still have a running Ghost server from a previous old E2E test
+        // Those tests never stopped the server in the tests manually
+        await stopGhost();
```
naz, line 346: AFAIR, adding `stopGhost` here would introduce a significant overhead to test suite running time. We deliberately avoided using stopGhost in the e2e framework to speed the tests up. That's why `startGhost` is not an actual HTTP server, it's an express app which we [hook into with the framework](https://github.com/TryGhost/framework/blob/36215e772d3983781771f3507972a18974334894/packages/express-test/lib/request.js#L22) directly to bypass HTTP overhead. 

In which specific cases/orders are the test breaking? Is there an alternative way without forcing the `stopGhost` here?

naz (changes requested):
Most of the changes are great! I think we could benefit by splitting them into three separate commits... otherwise it's just way too much of a fix in one go. 

The configUtils.restore() wrapped in a Promise is a good change ready to be merged. So is the change with the adapter cache clearing. The "stopGhost" bit is tricky and needs a different approach.

ghost/core/test/utils/e2e-framework.js
SimonBackx: Do note that we don't stop ghost for every E2E test. This is only for E2E tests that still need the legacy Ghost HTTP instance (not the Express app) to access the frontend. And if that happens, we only stop the Ghost **HTTP** server that was created in the tests that use the old E2E framework. Also, I don't think this causes any overhead because inside `stopGhost`,  it checks for `ghostServer` to be running. So it only stops the existing (legacy!) http server if it is needed.

So to give a better overview.
- We have tests that use the legacy e2e-utils file to startGhost. These never stop Ghost, because startGhost in there stops it for them. This means that the Ghost HTTP server is left open when the last legacy test E2E test runs.
- In our new E2E test framework, we normally never start a HTTP server. Unless if we need the frontend. For some reason we can't really access the frontend using the new test agents. But that can be fixed in the future probably. As a temporary workaround we still use `supertest.agent` for that. And that requires a HTTP server. So for a minority of new tests, we still need a HTTP server (for now). In that case we need to stop existing HTTP servers from the legacy E2E tests if it is running.

Since the test ordering is very unpredictable, and the stopGhost server only stops the legacy HTTP server if it is running, I don't see a big problem in keeping this here for now until we are able to switch the new E2E tests that need the frontend server to a TestAgent that doesn't require HTTP.

ghost/core/test/utils/urlUtils.js
SimonBackx: No, I added this to prevent having uncaught promise exceptions. Just bad practice to not catch promise errors, because that could crash Node if we don't explicitly have a global uncaught promise listener.

ghost/core/test/utils/e2e-framework.js
naz: Okay, gotcha! The dots are connected now 😄 

This explanation reminded me of a [traumatic commit](https://github.com/TryGhost/Ghost/commit/26b0bbc623885f65ce770cec2a0daf5291c115d7) I've made last year. As a result of that commit we have [this issue](https://github.com/TryGhost/Toolbox/issues/471), which I think is worth referencing here. That issue mentions:
>  fix the underlying problem of state management when frontend has to be initialized for e2e tests

^ this is pretty much the problem we are solving here, right? 

A more explicit solutions would be making sure we call "ghostStop" in the "after" clause wherever we start ghost with frontend, like [done here](https://github.com/TryGhost/Ghost/commit/26b0bbc623885f65ce770cec2a0daf5291c115d7#diff-ef9a744fed6a74ceed8b2346945c7dfa8cddd752ead8f1522bb1687307b8f0e3R38). It's away to much to track and remember of other engineers to add in their tests when working on new suites. Your solutions seems more universal, given there's no noticeable overhead just calling the 'stopServer' every time. 

I think the reasoning is good behind these changes! Can't think of a better approach to fix those "dirty" frontend starts 😅

PR #16170, opened 2023-01-24
Added mentions email notifications

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/2429

- started created a new MentionsNotifications class that sends an email when a new webmention is received.
- is triggered by a an event listener


ghost/webmentions/lib/MentionsAPI.js
```diff
@@ -160,8 +162,9 @@ module.exports = class MentionsAPI {
                 sourceFavicon: metadata.favicon,
                 sourceFeaturedImage: metadata.image
             });
+            const event = MentionCreatedEvent.create({data: mention, timestamp: new Date()});
```
allouis, line 165: Can we update this to use the pattern we use in Tiers?

First we add the Event to the Tier object like this: https://github.com/TryGhost/Ghost/blob/main/ghost/tiers/lib/Tier.js#L282-L284

And then in the TierRepository, we dispatch all necessary events: https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/services/tiers/TierRepository.js#L112-L114

allouis (commented):
Looks good! One comment on the event stuff :) 

Only other thing was discussed in Slack - I think it's better to use the StaffService here, rather than create a new module for it

allouis (commented):
This looks great! Just one comment about where we do the labs check - but otherwise 👍

allouis (approved):

PR #16174, opened 2023-01-24
🐛 Fixed invalid expiry for member tier subscriptions

rshbhgrg:
refs https://github.com/TryGhost/Team/issues/2476

When upgrading from a Complimentary subscription with an expiry, to a paid Subscription of the same Tier, the Member was eventually losing access to the Tier when the complimentary subscription expires as the `expiry_at` on the mapping was not removed. This change fixes the code by setting expiry as null when a member upgrades their subscription to paid. Contains 2 migrations

- [Removed invalid expiry tier expiry date for paid members](https://github.com/TryGhost/Ghost/pull/16174/commits/a8fe6c44de09a06699966ca3ce548db487a4f38d)
- [Restored missing tier mapping for paid members](https://github.com/TryGhost/Ghost/pull/16174/commits/63a869845ff465252b049ae0101e76f2e60bc58c)

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.32/2023-01-24-08-00-fix-invalid-tier-expiry-for-paid-members.js
+++ b/ghost/core/core/server/data/migrations/versions/5.32/2023-01-24-08-00-fix-invalid-tier-expiry-for-paid-members.js
@@ -5,6 +5,8 @@ module.exports = createTransactionalMigration(
     async function up(knex) {
         logging.info('Removing expiry dates for paid members');
         try {
+            // Fetch all members with a paid status that have an expiry date
+            // Paid members should not have an expiry date
             const invalidExpiryIds = await knex('members_products')
                 .select('members_products.id')
                 .leftJoin('members', 'members_products.member_id', 'members.id')
--- a/ghost/core/core/server/data/migrations/versions/5.32/2023-01-24-08-00-fix-invalid-tier-expiry-for-paid-members.js
+++ b/ghost/core/core/server/data/migrations/versions/5.32/2023-01-24-08-00-fix-invalid-tier-expiry-for-paid-members.js
@@ -1,0 +1,42 @@
+const logging = require('@tryghost/logging');
+const {chunk: chunkArray} = require('lodash');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Removing expiry dates for paid members');
+
+        const paidMembersWithExpiry = await knex('members_products')
+            .select('members_products.id')
+            .leftJoin('members', 'members_products.member_id', 'members.id')
+            .where('members.status', '=', 'paid')
+            .whereNotNull('members_products.expiry_at');
+
+        logging.info(`Found ${paidMembersWithExpiry.length} paid members with expiry dates`);
+
+        if (paidMembersWithExpiry.length === 0) {
+            return;
+        }
+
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/data/migrations/versions/5.32/2023-01-24-08-00-fix-invalid-tier-expiry-for-paid-members.js
```diff
@@ -0,0 +1,42 @@
...
+        if (paidMembersWithExpiry.length === 0) {
+            return;
+        }
+
+        logging.info(`Removing expiry dates for ${paidMembersWithExpiry.length} paid members`);
+
+        // SQLite >= 3.32.0 can support 32766 host parameters
+        // We use one for the SET clause, and the rest can be
+        // used to populate the WHERE IN (?... clause.
+        const chunkSize = 32765;
```
allouis, line 26: We don't need chunking anymore as far as I'm aware - SQLite is only for development

ghost/core/core/server/data/migrations/versions/5.32/2023-01-24-08-00-fix-invalid-tier-expiry-for-paid-members.js
```diff
@@ -0,0 +1,42 @@
...
+
+        const paidMembersWithExpiryIds = paidMembersWithExpiry.map(row => row.id);
+
+        const chunks = chunkArray(paidMembersWithExpiryIds, chunkSize);
+
+        // eslint-disable-next-line no-restricted-syntax
+        for (const chunk of chunks) {
+            await knex('members_products')
+                .update('expiry_at', null)
+                .whereIn('id', chunk);
```
allouis, line 36: Can we instead do `whereIn('member_id', memberIds)` and then replace the join at the top with something like:
```
const memberIds = knex.select('id').from('members').where('status', 'paid').pluck('id');
```

allouis (commented):
Looks good - couple suggestions for one of the migrations!

I'm not 100% sure about the fix, I don't really understand why it fixes it - would be cool to add a test or something if possible :)

rishabhgrg:
> I'm not 100% sure about the fix, I don't really understand why it fixes it -

@allouis Yea i am looking into the best way to add a test for this one, its a bit tricky 😅 
As for the fix, the problem right now is we only set the `expiry_at` on `members_products` when it has a value, ignoring it it was empty which is the case when member upgrades it to paid. The fix updates that code to also set `expiry_at` to `null` if there is no value coming for it, so when a member upgrades to paid on same tier, that will set the `expiry_at` on the `members_products` pivot table to null.

ghost/core/core/server/data/migrations/versions/5.32/2023-01-24-08-00-fix-invalid-tier-expiry-for-paid-members.js
rishabhgrg: Ah right, updated 👍

ghost/core/core/server/data/migrations/versions/5.32/2023-01-24-08-00-fix-invalid-tier-expiry-for-paid-members.js
rishabhgrg: Yea sure, have updated with the matching change now :)

daniellockyer (approved):

PR #16176, opened 2023-01-24
🐛 Reduced concurrency when fetching Mailgun events

SimonBackx:

This change adds a small sleep in between dispatching events in the worker thread that reads the events from Mailgun. That should reduce the amount of queries we fire parallel to each other and could cause the connection pool to run out of connections.

It also reduces the amount of concurrent sending to 2 from 10. Also to make sure the connection pool doesn't run out of connections while sending emails, and to reduce the chance of new connections falling back on a (delayed) replicated database.


ghost/core/test/e2e-api/shared/__snapshots__/version.test.js.snap
```diff
@@ -813,7 +813,7 @@ generated some helpful information about the error that you can share with the
 creator of the Zapier integration below: 
 
  Integration expected Ghost version:v3.1 
-Current Ghost version:v5.30
```
naz, line 816: This test should not be using the email snapshots just yet as there's a dynamic bit with Ghost version. Which is a [documented limitation](https://ghost.notion.site/End-to-end-Testing-6a2ef073b1754b18aff42e24a632a007#3a22bcdf5ab747309d8ab1bef8b98578) atm.

ghost/email-service/test/batch-sending-service.test.js
```diff
@@ -461,7 +461,7 @@ describe('Batch Sending Service', function () {
             sinon.assert.callCount(sendBatch, 101);
             const sendBatches = sendBatch.getCalls().map(call => call.args[0].batch);
             assert.deepEqual(sendBatches, batches);
-            assert.equal(maxRunningCount, 10);
+            assert.equal(maxRunningCount, 2);
```
lschiavini, line 464: Why not use the same const as before?

const MAX_SENDING_CONCURRENCY = 2; at batch-sending-service.js

PR #16183, opened 2023-01-25
Reduced email error messages

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2487


ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/emails.js
```diff
@@ -13,5 +14,11 @@ module.exports = (model, frame) => {
         );
     });
 
+    if (!labs.isSet('emailErrors')) {
+        if (jsonModel.status === 'failed') {
+            jsonModel.status = 'submitted';
```
rishabhgrg, line 19: This would mean we'll show `submitted` for a completely failed email as well right ? 🤔

rshbhgrg (approved):
Looks good overall, bit of a concern about removing the ability to retry for real email failures but i think thats needed short term so unavoidable.

PR #16184, opened 2023-01-25
Improved email error logging

SimonBackx:
no issue

Logs errors to Sentry and adds error codes.

### DIFF
--- a/ghost/email-service/test/batch-sending-service.test.js
+++ b/ghost/email-service/test/batch-sending-service.test.js
@@ -125,8 +125,12 @@ describe('Batch Sending Service', function () {
                     status: 'pending'
                 }
             });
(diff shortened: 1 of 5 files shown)


ghost/email-service/test/batch-sending-service.test.js
```diff
@@ -664,7 +726,7 @@ describe('Batch Sending Service', function () {
             });
 
             assert.equal(result, false);
-            sinon.assert.calledTwice(errorLog);
+            sinon.assert.notCalled(errorLog);
```
SimonBackx, line 729: So in this case it is not logged because the mailgun provider would have logged the error

SimonBackx: In the past it was logged twice

PR #16185, opened 2023-01-25
Email debug updates

peterzimon:
Fixes a couple of UX issues on the email debug screen:

- shows [...] button only for errors actually longer than the available space to avoid confusion about where there's more error text
- use actual avatars instead of fake red/blue dots to make it consistent with the rest of the app
- adds click through to member details screen to easily access member data if needed
- updates text select for provider ID for easier copying
- removes unused "Download full error message" icon


ghost/admin/app/components/posts/debug.hbs
```diff
@@ -67,29 +67,37 @@
...
                             <div class="gh-email-debug-failure-codes">
                                 <span class="gh-email-debug-failure-code">Failure code: <span>{{failure.code}}</span></span>
-                                <span class="gh-email-debug-failure-code">Enhanced code: <span>{{failure.enhancedCode}}</span></span>
-                            </div>
-                            <div class="gh-email-debug-readmore-error">
-                                <input type="checkbox" id="error-{{index}}-checkbox" class="toggle-checkbox">
-                                <label for="error-{{index}}-checkbox">{{svg-jar "dotdotdot"}}</label>
-                                <span>{{failure.message}}</span>
+                                {{#if failure.enhancedCode}}
+                                    <span class="gh-email-debug-failure-code">Enhanced code: <span>{{failure.enhancedCode}}</span></span>
```
rishabhgrg, line 97: @peterzimon Don't think we need another `span` here inside the original one right ?

rshbhgrg (approved):
Left a comment for tiny change, rest LGTM

ghost/admin/app/components/posts/debug/email-error-message.js
```diff
@@ -0,0 +1,15 @@
...
+    @tracked more = false;
+
+    @action
+    setup(element) {
+        const errorContainer = element.querySelector('span');
+        if (element.offsetWidth <= errorContainer.offsetWidth) {
+            this.more = true;
+        }
+    }
+}
```
rishabhgrg, line 15: We should ideally have an extra line in the end of the files , not a big deal tho :)

ghost/admin/app/components/posts/debug.hbs
peterzimon: we kinda do to format the enhanced code value (this is valid HTML fwiw)

rishabhgrg: Aah cool, TIL 😄

PR #16187, opened 2023-01-25
Added referrals invite notification

aileen:
no issue

- Ghost users that make >= $100 MRR will see a dismissible notification that invites them to the Ghost Referral program
- Only Admin and Owner users will see it
- Only applies when Stripe is setup and connected in live mode
- By saving a `referralInviteDismissed` property to the users' `accessibility` JSON object we can determine if the notification has been dismissed and won't show it again
- Added new `gh-referral-invite` component

### DIFF
--- a/ghost/admin/app/components/gh-nav-menu/footer.hbs
+++ b/ghost/admin/app/components/gh-nav-menu/footer.hbs
@@ -1,17 +1,12 @@
-<div class="gh-nav-bottom" {{did-insert this.loadCurrentMRR}}>
+<div class="gh-nav-bottom">
     {{#if this.hasThemeErrors}}
     <button type="button" class="gh-footer-toast gh-theme-error-toast" {{on "click" (fn this.openThemeErrors null)}}>
         <span class="gh-footer-toast-title gh-notification-title">Your theme contains errors</span>
         <span class="gh-footer-toast-p">Some functionality on your site may be limited &rarr;</span>
     </button>
-    {{else if this.showReferralInvite}}
-    <a href="https://referrals.ghost.org/?ref=admin" class="gh-referral-toast" target="_blank" rel="noopener noreferrer">
-        <span class="gh-referral-toast-close" {{on "click" (fn this.dismissReferralInvite)}}>&#xd7;</span>
-        <strong>You qualify for our invite-only  referral program.</strong>
-        <p class="gh-footer-toast-p">Refer people to Ghost and earn a <strong>30% share</strong> of the subscription revenue, every single month.</p>
-        <button class="gh-btn gh-btn-black gh-referral-toast-button"><span>Find out more &rarr;</span></button>
-    </a>
     {{/if}}
+
+    <GhReferralInvite @hasThemeErrors={{this.hasThemeErrors}} />
 
     <div class="flex items-center justify-between">
         <div class="pe-all">
--- a/ghost/admin/app/components/gh-nav-menu/footer.hbs
+++ b/ghost/admin/app/components/gh-nav-menu/footer.hbs
@@ -1,4 +1,4 @@
-<div class="gh-nav-bottom">
+<div class="gh-nav-bottom" {{did-insert this.loadCurrentMRR}}>
     {{#if this.hasThemeErrors}}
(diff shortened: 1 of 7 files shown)


ghost/admin/app/components/gh-nav-menu/footer.js
```diff
@@ -35,6 +63,22 @@ export default class Footer extends Component {
...
+
+    @action
+    dismissReferralInvite(event) {
+        event.preventDefault();
+        event.stopPropagation();
+        const key = 'referralInviteDismissed';
+        const value = moment().tz(this.settings.timezone);
+        const options = {user: true};
+
+        return this.feature.update(key, value, options);
```
kevinansfield, line 79: `this.feature.update` is more of an internal implementation detail and shouldn't be used outside of the feature service.

Inside the feature service you can add

```js
@feature('referralInviteDismissed', {user: true}) referralInviteDismissed;
```

then in this code you can use that directly, e.g.

```js
if (!this.feature.referralInviteDismissed) {}`

this.feature.referralInviteDismissed = moment().tz(this.settings.timezone)
```

ghost/admin/app/components/gh-nav-menu/footer.hbs
```diff
@@ -1,9 +1,16 @@
...
+<div class="gh-nav-bottom" {{did-insert this.loadCurrentMRR}}>
     {{#if this.hasThemeErrors}}
-    <button type="button" class="gh-theme-error-toast" {{on "click" (fn this.openThemeErrors null)}}>
-        <span class="gh-notification-title">Your theme contains errors</span>
-        <span class="gh-theme-error-p">Some functionality on your site may be limited &rarr;</span>
+    <button type="button" class="gh-footer-toast gh-theme-error-toast" {{on "click" (fn this.openThemeErrors null)}}>
+        <span class="gh-footer-toast-title gh-notification-title">Your theme contains errors</span>
+        <span class="gh-footer-toast-p">Some functionality on your site may be limited &rarr;</span>
     </button>
+    {{else if (and this.showReferralInvite (not this.hasThemeErrors))}}
```
kevinansfield, line 7: This can be simplified, the `not this.hasThemeErrors` is already handled by the `if/else`

ghost/admin/app/components/gh-nav-menu/footer.hbs
```diff
@@ -1,9 +1,16 @@
...
+    <a href="https://referrals.ghost.org/?ref=admin" class="gh-referral-toast" target="_blank" rel="noopener noreferrer">
+        <span class="gh-referral-toast-close" {{on "click" (fn this.dismissReferralInvite)}}>&#xd7;</span>
+        <strong>You qualify for our invite-only  referral program.</strong>
+        <p class="gh-footer-toast-p">Refer people to Ghost and earn a <strong>30% share</strong> of the subscription revenue, every single month.</p>
+        <button class="gh-btn gh-btn-black gh-referral-toast-button"><span>Find out more &rarr;</span></button>
+    </a>
```
kevinansfield, line 13: It would be cleaner to pull this into a separate component, then you can write it as a glimmer component and use it's `constructor` to load the MRR stats rather than `{{did-insert}}` which we generally want to avoid (it's designed as a band-aid whilst migrating from old-style to new-style Ember and is [generally considered an anti-pattern](https://github.com/emberjs/ember-render-modifiers#when-to-use-these-modifiers-and-when-not-to-use-them)).

Means there's less old-style code that needs migrating later down the line and the show/not-show/dismiss logic is all self-contained rather than being mixed in with the other nav footer code 😄 (+ easier to test, easier to move when designs change)

kevinansfield (changes requested):

ghost/admin/app/components/gh-nav-menu/footer.js
```diff
@@ -11,8 +11,6 @@ import {inject as service} from '@ember/service';
 export default class Footer extends Component {
     @service session;
     @service router;
-    @service whatsNew;
-    @service feature;
```
aileen, line 15: There were not used in the footer component

kevinansfield (approved):

PR #16200, opened 2023-01-30
Added mentions permissions

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/2420

- Added user roles and permissions for the mentions API.
- We only have a `browse` function for our current use case, accessible by `administrator` and `admin integration`


ghost/core/core/server/data/schema/fixtures/fixtures.json
```diff
@@ -763,7 +768,8 @@
                     "newsletter": "all",
                     "explore": "read",
                     "comment": "all",
-                    "link": "all"
+                    "link": "all",
+                    "mention": "browse"
```
SimonBackx, line 772: We need to add the same to `"Admin Integration": {` a couple of lines further. Just like the test fixtures one

allouis (approved):

daniellockyer (approved):

PR #16201, opened 2023-01-30
Added outbound link tagging for web posts

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2433

- Moved all outbound link tagging code to separate OutboundLinkTagger
- Because a site can easily enable/disable this feature, we don't store the ?refs in the HTML but add them on the fly for now in the Content API.


SimonBackx:
Since the HTML is edited using cheerio, when outbound link tagging is enabled (behind a flag), some HTML characters are replaced, such as `&apos;`. I don't immediately think that will cause issues, but I'm open for a second opinion here 🤔

9larsons:
I can't think of a situation right now where the substitutions will cause an issue. From some quick searching, `&apos;` is not part of the HTML4 standard... and I think that's probably just fine. 

Two comments/questions:
1. It's probably worth adding tests to cover `addToHtml` in `outbound-link-tagger.test.js`. (service coverage here is not great)
2. Without knowing much about `member-attribution`, it seems odd for this functionality to live here. In a sense this is post/site attribution, as we'll be tagging any outbound links of posts and newsletters. 

Functionally it all looks good.

9larsons:
Oh and minor thing, if we're touching all this code, it's an opportunity to update the language use to use `blocklist`.

9larsons (approved):
Marking approved since I don't have any functional concerns.

SimonBackx:
1. I've added new test coverage for the addToHtml method, you were right 😊
2. I agree that member-attribution isn't the best location for this. Ideally it should be in a separate package, but I'm keeping it here for now to not make the changes too big. In a future cleanup this could be moved.
3. 100% good point. I've updated the term! Thanks for pointing this out.

PR #16209, opened 2023-01-31
Implemented awaitable DomainEvents

SimonBackx:

**Main reason for this change:**
We need a way to await DomainEvents, otherwise we risk sending too many events in one go, if we for example send events in a for loop. This is causing **issues in the email event processing**, which sends thousands of events in a loop. This causes the db connection pool to run out of connections and eventually time out. A proper implementation should wait for each event to be processed before sending the next event. With this change that becomes possible.

**Why not native EventEmitter?**
This uses a new AsyncEventEmitter, which follows the same interface as EventEmitter but is awaitable. It also only fires event listeners one after each other instead of parallel (= more predictable outcome + connection pool is limited and we don't need that concurrency). The native EventEmitter doesn't support this functionality, so custom code was required.

**Other advantages:**
- We can write more stable code (and tests!) by awaiting dispatchEvents where needed. E.g. in the members API code, we sometimes dispatch events that affect the result of the same request. (e.g. creating a member -> attribution stored -> attribution not returned in same request randomly sometimes because delay in events) That means that tests are not stable and the responses of those requests are not predictable and could be incorrect.

### DIFF
--- a/ghost/core/test/e2e-api/admin/members.test.js
+++ b/ghost/core/test/e2e-api/admin/members.test.js
@@ -746,7 +746,7 @@ describe('Members API', function () {
     it('Can add a member and trigger host email verification limits', async function () {
         configUtils.set('hostSettings:emailVerification', {
             apiThreshold: 0,
-            adminThreshold: 1,
+            adminThreshold: 2,
             importThreshold: 0,
             verified: false,
             escalationAddress: 'test@example.com'
--- a/ghost/core/test/e2e-api/admin/members.test.js
+++ b/ghost/core/test/e2e-api/admin/members.test.js
@@ -2072,7 +2072,8 @@ describe('Members API', function () {
                 email: 'member1@test.com'
             });
 
-        nock('https://api.stripe.com:443')
+        const getNock = nock('https://api.stripe.com:443')
+            .persist()
             .get(`/v1/subscriptions/${stripeSubscriptionId}`)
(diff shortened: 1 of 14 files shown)


ghost/core/test/e2e-api/admin/members.test.js
```diff
@@ -2072,7 +2072,8 @@ describe('Members API', function () {
                 email: 'member1@test.com'
             });
 
-        nock('https://api.stripe.com:443')
+        const getNock = nock('https://api.stripe.com:443')
+            .persist()
```
SimonBackx, line 2076: This getter is called multiple times, but this wasn't caught because we werent awaiting events I think

ghost/members-api/lib/repositories/member.js
```diff
@@ -841,12 +846,19 @@ module.exports = class MemberRepository {
...
+            await this._Member.transaction(async (transacting) => {
+                await this.linkSubscription(data, {
                     ...options,
                     transacting
                 });
+                // Catch updated executionPromise
+                executionPromise = transacting.executionPromise;
             });
+            if (executionPromise) {
+                // Wait for added events
```
SimonBackx, line 859: This is required because we wouldn't wait for SubscriptionCreatedEvents -> unpredictable if those are executed before we 'read' the member back in the response for the same API endpoint = unpredictable responses.

PR #16215, opened 2023-02-01
Mentions ordering

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/2498

- orders mentions by timestamp.


allouis (commented):
I don't think the `waitForOneSecond` approach is a good pattern as it's gonna slow down tests.

Instead we can either:
- Use `sinon.useFakeTimers`, this allows us to fake the time and tick it forward
- Don't use `processWebmention` instead create `Mention`s directory and save them in the repository.

allouis (approved):

PR #16216, opened 2023-02-01
Added initial basic milestone emails in-memory repository

aileen:
refs https://www.notion.so/ghost/Marketing-Milestone-email-campaigns-1d2c9dee3cfa4029863edb16092ad5c4

This adds a milestone entity and in-memory repository in a new `milestone-emails` package. This also adds a first initial definition of milestones and their types which is held in the default config to avoid DB changes when, e. g. values change.

This should get everything in place to begin with the service implementation.


aileen:
@allouis the first version of the package with basic API functionality is done. Not sure how I can get the test coverage for the index files up, but tomorrow I'll continue with work on the service itself within Ghost core 😅

ghost/milestone-emails/lib/Milestone.js
```diff
@@ -0,0 +1,229 @@
...
+function validateCurrency(type, currency) {
+    if (type === 'members') {
+        return null;
+    }
+
+    if (!currency || (currency && typeof currency !== 'string')) {
+        return 'usd';
+    }
+
+    if (currency.length > 3) {
```
allouis, line 181: I think in this case you probs wanna return `'usd'` because it's not like a title that we can just truncate - it needs to be a valid currency code

ghost/milestone-emails/lib/Milestone.js
```diff
@@ -0,0 +1,229 @@
...
+/**
+ *
+ * @param {string} name
+ * @param {number} value
+ * @param {'arr'|'members'} type
+ * @param {string|null} currency
+ *
+ * @returns {string}
+ */
+function validateName(name, value, type, currency) {
```
allouis, line 197: I don't think you need to store the `name` as `this.#name` it seems to be calculated entirely from other values so you can just add a getter for it like:
```
get name() {
  if (this.type === 'arr') {
    return `arr-${this.value}-${this.currency}`;
  }
  return `members-${this.value}`;
}
```

allouis (commented):
Looking good - there's a couple comments nothing major tho!

One thing I would say is to consider a little more about the MilestonesAPI - There's a few methods on there that you might want private.

ghost/milestone-emails/lib/Milestone.js
aileen: Yeah, you're right. Should fall back to `usd` and for now we only deal with that anyway

ghost/milestone-emails/lib/MilestonesEmailService.js
```diff
@@ -0,0 +1,149 @@
...
+     * @param {Object} deps.config
+     * @param {IQueries} deps.queries
+     * @param {string} deps.defaultCurrency
+     */
+    constructor(deps) {
+        this.#mailer = deps.mailer;
+        this.#api = deps.api;
+        this.#config = deps.config;
+        this.#queries = deps.queries;
+        this.#defaultCurrency = deps.defaultCurrency;
```
aileen, line 47: @allouis after checking, we'll have to expect to only deal with a default currency. Probably the currency as defined in the default product. Still need to figure out on the other end to grab that.

ghost/milestone-emails/lib/MilestonesEmailService.js
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     *
+     * @param {number} milestone
+     * @param {'arr'|'members'} type
+     * @param {boolean} hasMembersImported
+     *
+     * @returns {Promise<Milestone>}
+     */
+    _saveMileStoneAndSendEmail = async (milestone, type, hasMembersImported) => {
```
aileen, line 71: Not sure if we write private functions like this?

ghost/milestone-emails/lib/MilestonesEmailService.js
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        // Two cases in which we don't want to send an email
+        // 1. There has been an import of members within the last week
+        // 2. The last email has been sent less than two weeks ago
+        const shouldSendEmail = await this.#api.shouldSendEmail();
+
+        if (shouldSendEmail && !hasMembersImported) {
+            // TODO: hook up GhostMailer or use StaffService and trigger event to send email
+            // await this.#mailer.send({
```
aileen, line 88: This is a step to decide next: use GhostMailer directly or utilise the StaffService and send an event 🤔

ghost/milestone-emails/lib/MilestonesEmailService.js
```diff
@@ -0,0 +1,149 @@
...
+
+            if (milestonesForCurrency && currentARRForCurrency) {
+                // get the closest milestone we're over now
+                milestone = this._getMatchedMilestone(milestonesForCurrency.values, currentARRForCurrency.arr);
+
+                // Fetch the latest milestone for this currency
+                const latestMilestone = await this.#api.getLatestArrMilestone(this.#defaultCurrency);
+
+                if (!latestMilestone || milestone > latestMilestone.value) {
+                    return await this._saveMileStoneAndSendEmail(milestone, 'arr', hasMembersImported);
```
aileen, line 125: Not sure what what we should expect to be returned from the service. For now I returned the created milestone if there was one created.

allouis (commented):
Looks good! One thing - I don't quite know if we need both the `MilestonesAPI` and the `MilestoneEmailService` - I think we can just merge those into a single class?

Happy to be wrong on this, so let me know what you think and why you split them in that way

allouis (approved):

PR #16218, opened 2023-02-01
Added automatic database query retrying to the email service

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2512

The email sending is a crucial flow that should not be interrupted. If there are database connection issues, we should try to recover from them automatically by retrying individual database queries and transactions for a limited amount of time.

This adds a helper method to retry database operations. It limits all database queries before sending the actual email to maximum 10 minutes. Database operations that happen after sending have a higher retry time because they are more crucial to prevent data loss (e.g. saving that an email was sent).

### DIFF
--- a/ghost/email-service/lib/batch-sending-service.js
+++ b/ghost/email-service/lib/batch-sending-service.js
@@ -31,6 +31,10 @@ class BatchSendingService {
     #db;
     #sentry;
 
+    // Retry database queries happening before sending the email
+    #BEFORE_RETRY_CONFIG = {maxRetries: 10, maxTime: 10 * 60 * 1000, sleep: 2000};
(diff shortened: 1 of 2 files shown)


ghost/email-service/lib/batch-sending-service.js
```diff
@@ -316,7 +343,13 @@ class BatchSendingService {
         let succeeded = false;
 
         try {
-            const members = await this.getBatchMembers(batch.id);
+            const members = await this.retryDb(
+                async () => {
+                    return await this.getBatchMembers(batch.id);
```
SimonBackx, line 348: We also need an empty rows check here. Which is a weird database issue that somehow isn't caught and is just returning 0 rows

allouis (dismissed):

PR #16222, opened 2023-02-03
Added query caching to products repository

naz:
refs https://github.com/TryGhost/Toolbox/issues/515

- The `select * from products` was one of the top queries putting pressure on the database. Caching query results would eliminate most of this pressure because Product as a resource does not change much in the system
- This change adds caching to the "list" method only. A follow up commit will add similar aching to the "get" method.
- The best cache for Products is "Memory" one because there are usually not more than few dozen of products defined on the single instance that don't change often (usually are left as is after initial setup)
- The cache-key for the entries consists of following pattern: `{method_name}-{stringified-arguments}`
- Above caching pattern can be reused across repositories that are heavy on reads and light on wrights mostly-universally (looking at you tags!)

🙈 PR is here for progress tracking. What's left todo:
- [x] initialize products cache and pass it into members-api
- [x] check for Product writes bypassing repository
- [ ] profit big time

### DIFF
--- a/ghost/members-api/lib/repositories/product.js
+++ b/ghost/members-api/lib/repositories/product.js
@@ -89,7 +89,10 @@ class ProductRepository {
         global['con' + 'sole']['l' + 'og']('cache key: ', cacheKey);
         global['con' + 'sole']['l' + 'og']('CACHING: ', JSON.stringify(product, null, 4));
 
-        await this.#cache.set(cacheKey, product);
+        if (product) {
+            await this.#cache.set(cacheKey, product);
+        }
+
         return product;
     }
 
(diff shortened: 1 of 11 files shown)


ghost/members-api/lib/repositories/product.js
```diff
@@ -87,45 +100,47 @@ class ProductRepository {
...
+        } else if ('id' in data) {
+            product = await this._Product.findOne({id: data.id}, options);
+        } else if ('slug' in data) {
+            product = await this._Product.findOne({slug: data.slug}, options);
         }
 
-        if ('slug' in data) {
-            return await this._Product.findOne({slug: data.slug}, options);
+        if (!product) {
+            throw new NotFoundError({message: 'Missing id, slug, stripe_product_id or stripe_price_id from data'});
```
allouis, line 139: We never used to throw if it was missing, we'd return null

naz: oh!! good catch

naz: that was it :D thanks for saving loads of headache haha

allouis: 💃

naz:
Similar optimization will be done through https://github.com/TryGhost/Ghost/pull/16224 instead. It requires too much work to make the legacy ProductsRepository play ball with our test suites and there's no point investing time into code we are planning to throw away soon anyway.

PR #16223, opened 2023-02-03
Implemented mentions rate limiter

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/2419

- added basic rate limiter implementation to the mentions receiving endpoint
- still need to configure how it gets limited (time / after how many hits, etc), but it can load data from `defaults` in the meantime.
- add additional validation, eg checking if it contains real body data, eg `source` and `target`?


ghost/core/core/server/web/shared/middleware/api/spam-prevention.js
```diff
@@ -123,6 +126,34 @@ const globalReset = () => {
...
+    });
+
+    mentionsBlockInstance = mentionsBlockInstance || new ExpressBrute(store,
+        extend({
+            attachResetToRequest: false,
+            failCallback(req, res, next) {
+                return next(new errors.TooManyRequestsError({
+                    message: `Too many mention attempts`,
+                    context: tpl(messages.mentionsBlock,
+                        {rfa: spamMentionsBlock.freeRetries + 1 || 5, rfp: spamMentionsBlock.lifetime || 60 * 60})
```
allouis, line 147: These values don't seem to be used in the template

ghost/core/core/server/web/shared/middleware/api/spam-prevention.js
```diff
@@ -20,7 +20,11 @@ const messages = {
         error: 'Only {rateSigninAttempts} tries per IP address every {rateSigninPeriod} seconds.',
         context: 'Too many login attempts.'
     },
-    tooManyAttempts: 'Too many attempts.'
+    tooManyAttempts: 'Too many attempts.',
+    mentionsBlock: {
+        error: 'Only {rateMentions} mentions per IP address every {rateMentionsPeriod} seconds.',
```
allouis: These values aren't being passed in to the `tpl` function

ghost/core/core/server/web/shared/middleware/mentions-limiter.js
```diff
@@ -0,0 +1,13 @@
+const url = require('url');
```
allouis, line 1: The rest of our rate limiting middleware is inside `middleware/brute.js` - I think this should probably live there, too

ghost/core/core/server/web/shared/middleware/mentions-limiter.js
```diff
@@ -0,0 +1,13 @@
...
+const spamPrevention = require('./api/spam-prevention');
+
+module.exports = {
+    globalLimits(req, res, next) {
+        return spamPrevention.mentionsBlock().getMiddleware({
+            ignoreIP: false,
+            key(_req, _res, _next) {
+                _next(url.parse(_req.url).pathname);
```
allouis, line 9: This should be the unique key for blocks - in this case I think we just want it as a static value

ghost/core/core/shared/config/defaults.json
```diff
@@ -102,6 +102,12 @@
             "maxWait": 43200000,
             "lifetime": 43200,
             "freeRetries": 8
+        },
+        "mentions_block": {
```
allouis, line 106: RE: Naming - this block is specifically for incoming Webmentions - I think we should probably use the word webmentions rather than mentions in the naming.

ghost/core/core/shared/config/defaults.json
```diff
@@ -102,6 +102,12 @@
...
+            "minWait": 60,
+            "maxWait": 600,
+            "lifetime": 6000,
+            "freeRetries": 2
```
allouis, line 110: For the values, maybe let's try and use some values that will have little to no impact initially?

I haven't as much context as you but looking at https://github.com/AdamPflug/express-brute/blob/master/index.js maybe something like:
```
{
  minWait: 1,
  maxWait: 1,
  lifetime: 0,
  freeRetries: 100
}
```

ronaldlangeveld: Couldn't get the values that liberal as then tests cannot reproduce a block - simply resets too fast.
Made it a little stricter than your suggestion, but still way more liberal compared to what it was.

```
"webmentions_block": {
            "minWait": 10,
            "maxWait": 100,
            "lifetime": 1000,
            "freeRetries": 100
        }
```

ghost/core/core/server/web/shared/middleware/api/spam-prevention.js
```diff
@@ -123,6 +126,32 @@ const globalReset = () => {
...
+        createTable: false,
+        knex: db.knex
+    });
+
+    webMentionsBlockInstance = webMentionsBlockInstance || new ExpressBrute(store,
+        extend({
+            attachResetToRequest: false,
+            failCallback(req, res, next) {
+                return next(new errors.TooManyRequestsError({
+                    message: `Too many mention attempts`
```
allouis, line 145: This should use the message defined above!

ghost/core/core/server/web/shared/middleware/api/spam-prevention.js
```diff
@@ -123,6 +126,32 @@ const globalReset = () => {
     return globalResetInstance;
 };
 
+const webMentionsBlock = () => {
```
allouis, line 129: Minor but "webmention" should be cased as `webmention` rather than `webMention` because it's one word

allouis (changes requested):
Couple small changes otherwise :+1:

PR #16224, opened 2023-02-03
[WIP] Used TierRepository in posts serializer

allouis:
🍦 

- We shouldn't fetch data in the serializer, but we not addressing that now
- We want to cache access to Tiers, and it's easier to do that in the TierRepository
- So we update a heavy user of Tiers to use the TierRepository so it can take adv of caching


allouis:
We can pick this up in future

allouis:
Implemented in https://github.com/TryGhost/Ghost/pull/16775

PR #16231, opened 2023-02-06
Moved email event fetching to main thread

SimonBackx:

This moves the processing of Mailgun events to the main thread. By using a simple approach where we emit a start event on the worker thread (via the job manager) and listen for it on the main thread. This is needed because for now the job manager doesn't support scheduling periodic jobs on the main thread (not offloaded).

Apart from that, the email processor now uses the email event storage directy instead of emitting events (it is still emitting event for now). This makes sure we await for the event to be processed before continuing with the next event.


ghost/core/core/server/services/email-analytics/wrapper.js
```diff
@@ -0,0 +1,90 @@
...
+const debug = require('@tryghost/debug')('jobs:email-analytics:fetch-latest');
+
+class EmailAnalyticsServiceWrapper {
+    init() {
+        if (this.service) {
+            return;
+        }
+
+        const {EmailAnalyticsService} = require('@tryghost/email-analytics-service');
+        const {EmailEventStorage, EmailEventProcessor} = require('@tryghost/email-service');
```
SimonBackx, line 11: We can move both classes to the `@tryghost/email-analytics-service` package in a future cleanup if needed

allouis (approved):

PR #16234, opened 2023-02-06
moved sending mentions to job

9larsons:
refs TryGhost/Team#2523
-implemented jobsService within mention sending service
-updated tests so we don't need to sleep

### DIFF
--- a/ghost/webmentions/test/MentionSendingService.test.js
+++ b/ghost/webmentions/test/MentionSendingService.test.js
@@ -3,9 +3,21 @@ const assert = require('assert');
 const nock = require('nock');
 // non-standard to use externalRequest here, but this is required for the overrides in the libary, which we want to test for security reasons in combination with the package
 const externalRequest = require('../../core/core/server/lib/request-external.js');
+// non-standard to use jobservice here as well, but it's best to be jobbing off sending mentions
+const jobsService = require('../../core/core/server/services/jobs');
 const sinon = require('sinon');
 const logging = require('@tryghost/logging');
 const {createModel} = require('./utils/index.js');
+
+let jobService = {
+    async addJob(name, fn) {
+        jobsService.addJob({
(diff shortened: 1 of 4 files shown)


9larsons:
@allouis added you as reviewer. I don't love implementing the `jobsService` within the `webmentions` package but I'm not seeing another option... It makes the tests a bit awkward as-is.

Also with respect to the `webmentions` package unit tests, I was torn between the current implementation and moving the job to within `sendAll` (right where we're sending for each link). Not sure if you've got some thoughts there.

ghost/webmentions/test/MentionSendingService.test.js
```diff
@@ -3,10 +3,22 @@ const assert = require('assert');
...
 // non-standard to use externalRequest here, but this is required for the overrides in the libary, which we want to test for security reasons in combination with the package
 const externalRequest = require('../../core/core/server/lib/request-external.js');
+// non-standard to use jobservice here as well, but it's best to be jobbing off sending mentions
+const jobsService = require('../../core/core/server/services/jobs');
 const sinon = require('sinon');
 const logging = require('@tryghost/logging');
 const {createModel} = require('./utils/index.js');
 
+let jobService = {
+    async addJob(name, fn) {
```
allouis, line 13: I would use a fake job service here, something like:
```
const jobService = {
  async addJob(name, fn) {
    return fn();
  }
};
```

allouis: If you wanted you can also use `sinon.stub` or something to make sure that we call the `addJob` method

allouis (commented):
This looks good to me!

One comment on the tests, I don't think we should be pulling in the actual job service in the webmentions package unit tests

As for whether to wrap sendAll or send - I think sendAll is reasonable here

PR #16238, opened 2023-02-07
[WIP] Added support for updating and deleting mentions

allouis:
🍦

### DIFF
--- a/ghost/core/core/server/services/mentions/BookshelfMentionRepository.js
+++ b/ghost/core/core/server/services/mentions/BookshelfMentionRepository.js
@@ -63,9 +63,7 @@ module.exports = class BookshelfMentionRepository {
      * @returns {Promise<Page<import('@tryghost/webmentions/lib/Mention')>>}
      */
     async getPage(options) {
-        // TODO: sqlite uses 0, mysql uses false... how do we construct this filter properly?
-        const filter = `deleted:[0,false]` + (options.filter ? `+(${options.filter})` : '');
-        const page = await this.#MentionModel.findPage({...options, filter});
+        const page = await this.#MentionModel.findPage(options);
 
         return {
             data: await Promise.all(page.data.map(model => this.#modelToMention(model))),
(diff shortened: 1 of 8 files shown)


ghost/core/core/server/services/mentions/BookshelfMentionRepository.js
```diff
@@ -63,7 +63,9 @@ module.exports = class BookshelfMentionRepository {
      * @returns {Promise<Page<import('@tryghost/webmentions/lib/Mention')>>}
      */
     async getPage(options) {
-        const page = await this.#MentionModel.findPage(options);
+        // TODO: sqlite uses 0, mysql uses false... how do we construct this filter properly?
+        const filter = `deleted:[0,false]` + (options.filter ? `+(${options.filter})` : '');
```
9larsons, line 67: Not sure how we should handle `sqlite` using `0/1` vs. mysql using `true/false`.

ghost/webmentions/test/InMemoryMentionRepository.test.js
```diff
@@ -76,4 +76,35 @@ describe('InMemoryMentionRepository', function () {
         assert(pageThree.meta.pagination.prev === 2);
         assert(pageThree.meta.pagination.next === null);
     });
+
+    describe(`GetPage`, function () {
```
9larsons, line 80: This test doesn't really provide any code coverage but it confirm the behavior we want - that the front end will not receive deleted mentions.

ghost/core/core/server/services/mentions/BookshelfMentionRepository.js
allouis: `true/false` _should_ work for SQLite! I believe knex will handle the difference

allouis: I've updated this to use the `enforcedFilters` method in models - which is used as the `overrides` property in `@tryghost/nql`

PR #16241, opened 2023-02-07
Milestone emails

aileen:
refs https://www.notion.so/ghost/Marketing-Milestone-email-campaigns-1d2c9dee3cfa4029863edb16092ad5c4

Added milestone email service behind a flag. The service will currently run on boot and is meant to be scheduled soon, which should happen in the next step. For now it's protected behind the alpha flag.


ghost/milestone-emails/lib/MilestonesEmailService.js
```diff
@@ -2,24 +2,19 @@ const Milestone = require('./Milestone');
 
 /**
  * @template Model
- * @typedef {object} Mention<Model>
+ * @typedef {Object} Mention<Model>
  * @prop {Model[]} data
  */
 
 /**
- * @typedef {object} IMilestoneRepository
```
allouis, line 10: Same here!

ghost/milestone-emails/lib/MilestonesEmailService.js
```diff
@@ -115,10 +117,10 @@ module.exports = class MilestonesEmailService {
 
     /**
      *
-     * @param {Array} goalValues
+     * @param {Array.<number>} goalValues
```
allouis, line 123: I think `number[]` is preferable here ☺️

allouis (commented):
Generally we don't want to use the wrapper types like `Object` `Number` `Array` etc...

In the case of `Object` this is actually a lot looser than you think, this would allow a `Number` or an `Array` to be passed, because they are both instances of `Object`.

Instead we wanna use like `object` or `<type>[]` (e.g. `number[]` or `Thing[]`) for a list of types.

I think it's better that we move the labs check into the job that runs, because otherwise switching the labs flag will require a restart of the server for it to take effect

ghost/milestone-emails/lib/MilestonesEmailService.js
```diff
@@ -2,24 +2,19 @@ const Milestone = require('./Milestone');
 
 /**
  * @template Model
- * @typedef {object} Mention<Model>
```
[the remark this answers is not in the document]
allouis: Actually - what is this type even used for?

ghost/milestone-emails/lib/InMemoryMilestoneRepository.js
```diff
@@ -1,10 +1,9 @@
 /**
  * @typedef {import('./Milestone')} Milestone
- * @typedef {import('./MilestonesAPI').IMilestoneRepository} IMilestoneRepository
  */
 
 /**
- * @implements {IMilestoneRepository}
```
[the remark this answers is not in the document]
aileen: It errors for me as the MilestoneAPI doesn't exist anymore and we don't import it as we are in the `IMilestoneRepository`?

ghost/milestone-emails/lib/MilestonesEmailService.js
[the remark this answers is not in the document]
aileen: Not used. Removed it. Was a leftover from merging the repository into the API 😅

ghost/milestone-emails/lib/MilestonesEmailService.js
aileen: Not sure I'm following. Are you saying it should be lowercase, or that we're not using the `IMilestoneRepository`, cause I thought we need it for [here](https://github.com/TryGhost/Ghost/pull/16241/files#diff-6b0c2622bbe1fd366bbeb1262f7e2b878284878ad8aea79d15f1b063d24ebe34R27)

ghost/milestone-emails/lib/MilestonesEmailService.js
aileen: This errors, because I am returning a number. Returning the value of the first index of the filtered Array 😅

ghost/milestone-emails/lib/MilestonesEmailService.js
allouis: Oh sorry - I meant about the `object` :)

ghost/core/core/server/services/milestone-emails/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * @param {'arr'|'members'} milestoneType
+ *
+ * @returns {Promise<any>}
+ */
+module.exports = async (milestoneType) => {
+    // const labs = require('../../../shared/labs');
+
+    // if (labs.isSet('milestoneEmails')) {
```
aileen, line 11: @allouis I can't seem to access labs. It would always return `false` so not sure how to solve that.

ghost/core/core/server/services/milestone-emails/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+    const milestonesEmailService = new MilestonesEmailService({
+        mailer,
+        repository,
+        milestonesConfig, // avoid using getters and pass as JSON
+        queries
+        // TODO: do we need to check if Stripe is live enabled?
+    });
+
+    return await milestonesEmailService.checkMilestones(milestoneType);
```
aileen, line 35: Not sure if this is the right way to go, because when using it as in-memory storage it gets instantiated every time again, which doesn't make sense I think....

ghost/core/core/server/services/milestone-emails/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+        const s = Math.floor(Math.random() * 60); // 0-59 second
+        const m = Math.floor(Math.random() * 60); // 0-59 minute
+        const h = Math.floor(Math.random() * 24); // 0-23 hour
+        const wd = Math.floor(Math.random() * 7); // 0-6 weekday
+
+        jobsService.addJob({
+            at: `${s} ${m} ${h} * * ${wd}`, // Every week
+            // at: '55 * * * * *', // every minute for local development
+            job: path.resolve(__dirname, 'jobs/run-arr-milestones.js'),
```
aileen, line 56: In order to schedule it with a CRON tab I had to remove `offloaded:false` as this seems to be only working for one-off jobs, unless I'm missing something.

The only way then I could make it work was to create job files that immediately invoke, similar to how it's done for `email-analytics` and `update-check` services.

ghost/core/core/server/services/milestone-emails/index.js
aileen: Note: I also tried requiring it in the job files with the same result

allouis: Do you have `enableDeveloperExperiments` set? That's needed for alpha flags - also can you look at the `labs` setting and see what the value is

aileen: Yup, all set. None of the flags are actually available at this stage it seems 😔

ghost/core/test/e2e-server/services/milestone-emails.test.js
```diff
@@ -0,0 +1,222 @@
...
+    });
+
+    after(async function () {
+        await configUtils.restore();
+        sinon.restore();
+    });
+
+    beforeEach(function () {
+        labsStub = sinon.stub(labs, 'isSet').returns(true);
+        stripeModeStub = sinon.stub(stripeService.api, 'mode').get(() => 'live');
```
aileen, line 165: I think I messed up the DB tests with this one

ghost/core/core/server/services/milestone-emails/index.js
```diff
@@ -0,0 +1,57 @@
+// Stubbing stripe in test was causing issues. Moved it
```
allouis, line 1: Can you rename this file to `service.js` and then create an `index.js` with contents of `module.exports = require('./service.js')`

allouis (commented):
One comment den merge away!

PR #16242, opened 2023-02-08
Added `mention_notifications` column

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/2526

- created a migration for a new boolean column in users that would determine if the staff user gets an email when the publication receive a new mention.


ghost/core/core/server/models/user.js
```diff
@@ -68,7 +68,8 @@ User = ghostBookshelf.Model.extend({
             comment_notifications: true,
             free_member_signup_notification: true,
             paid_subscription_started_notification: true,
-            paid_subscription_canceled_notification: false
+            paid_subscription_canceled_notification: false,
+            mention_notifications: true
```
allouis, line 72: Have you discussed with product about what the default should be?

allouis (approved):

ghost/core/core/server/models/user.js
ronaldlangeveld: Ref https://ghost.slack.com/archives/C02G9E68C/p1675916876727609

daniellockyer (approved):

Issue #16125, opened 2023-01-16
Email recipient subsets can be chosen but it sometimes still sends to everyone when retrying


backspace:
### Issue Summary

A grassroots group I’m part of is wanting to start sending out a newsletter but we ran into the same problems helpfully documented in #15725. As an experiment, I tried creating a label so I could send to a subset of subscribers of under 100. I’ve been trying this repeatedly as I continue to try to get it sent and I’ve found in a specific circumstance that after I complete the penultimate step it says it’s going to send to all subscribers despite me having chosen the label, like this:

![ghost-subscriber-filter-broken](https://user-images.githubusercontent.com/43280/212576117-c6803b90-aa09-4745-a2b3-c9705ad81295.gif)

If I understand this correctly, it’s a UI-only problem, since the records in `email_receipients` have already been created and will be reused, so the actual subset will still be used.

Thanks for all the work on this over the years 💞

### Steps to Reproduce

1. Publish a post that includes emailing a subset of subscribers, but have the emailing fail.
2. Unpublish the post.
3. Refresh the browser.
4. Try publishing again with emailing, it will show initially that it will send only to the subset but then on the last step will show that’s emailing everyone.

### Ghost Version

5.30.0

### Node.js Version

16.19.0

### How did you install Ghost?

Docker image served by Dokku on Ubuntu 16.04

### Database type

MySQL 8

### Browser & OS version

Brave 1.41.100 (Chromium 103.0.5060.134) on macOS 13.1

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

ErisDS:
cc @55sketch

allouis:
I can reproduce this, and I can also confirm that it's just a UI problem.

I believe the offending code is here: https://github.com/TryGhost/Ghost/blob/v5.31.0/ghost/admin/app/utils/publish-options.js#L170-L172

We're not taking into account the `this.post.emailSegment` property, which seems to be fine when the post is new, because they both default to the same thing. And then when changing the segment in the publish options, we update both. But when refreshing we lose that information and end up with the default!

@kevinansfield I haven't got a lot of context here - can we just return `this.post.emailSegment` if it exists here?

kevinansfield:
Agreed, we should be using `this.post.emailSegment` in the publish options recipient filter if the post has it set.

[...]

PR #16247, opened 2023-02-08
🐛 Fixed default email recipients always being "everyone" rather than matching post visibility

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/16125

The previous fix for incorrect recipient details being shown when re-sending a failed email introduced another bug that prevented the "Match post visibility" default recipients setting from working.

- the server always sets `post.emailSegment` to `'all'` for new posts so the publish flow recipient filter logic that checked for `post.emailSegment` being present always defaulted to `'all'` rather than falling back to the selected default recipients setting
- when a post has been published but the email failed it will have its `newsletter` value set so we can use that as a check for using the `post.emailSegment` value in place of the default recipients setting

### DIFF
--- a/ghost/admin/app/utils/publish-options.js
+++ b/ghost/admin/app/utils/publish-options.js
@@ -168,7 +168,7 @@ export default class PublishOptions {
     }
 
     get recipientFilter() {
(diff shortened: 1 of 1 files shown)


flavianh:
Aren't there any unit tests in Ghost that could prevent such regressions from happening?

PR #16248, opened 2023-02-08
Added public tags (Content API) caching

naz:
- There are a lot of repeated cacheable tag-related queries coming from
{{get}} helpers in themes that can be cached.
- Having a repository layer deal with very specific type of query allows
to add extra functionality, like caching, on top of the database query
- This commit is wiring code that addds a default in-memory cache to
all db queries. Note, it lasts forever and has no "reset" listeners. The
production cache is mean to have a short time-to-live (TTL) - removes a need
to keep the cache always fresh.

It's a WIP, but sharing the direction early.

Left to do:
- [x] Feature flag for gradual rollout (will be hidden behind host settings config)
- [x] Unit tests for new modules
- [x] Redis cache configuration with TTL - find suitable implementation and prepare for Pro to use.

For review @allouis. This PR uses a combination of an established pattern repository + service wrapper initialization, but also adds a cache awareness to the repository. Wanted to get your eyes on it for a quick check if this all looks sound. Also note, the cache is only active when there's a `hostSettings` setting present - would run only on hosted environment, so we don't have to worry about self-hosted environments for the time being.

### DIFF
--- a/ghost/core/core/server/services/tags-public/service.js
+++ b/ghost/core/core/server/services/tags-public/service.js
@@ -1,0 +1,26 @@
+class TagsPublicServiceWrapper {
+    async init() {
+        if (this.api) {
+            // Already done
+            return;
+        }
+
+        // Wire up all the dependencies
+        const models = require('../../models');
+        const adapterManager = require('../adapter-manager');
+
+        const tagsCache = adapterManager.getAdapter('cache:tagsPublic');
+        const {TagsPublicRepository} = require('@tryghost/tags-public');
+
+        this.linkRedirectRepository = new TagsPublicRepository({
+            Tag: models.TagPublic,
+            cache: tagsCache
+        });
+
+        this.api = {
+            browse: this.linkRedirectRepository.getAll.bind(this.linkRedirectRepository)
+        };
+    }
+}
+
+module.exports = new TagsPublicServiceWrapper();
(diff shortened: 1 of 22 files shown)


ghost/tags-public/lib/TagsPublicRepository.js
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Gets all tags and caches the returned results
+     * @param {Object} options
+     * @returns
+     */
+    async getAll(options) {
+        let cacheKey;
+
+        if (this.#cache) {
```
naz, line 37: This guard is here to make caching optional

ghost/core/core/server/adapters/cache/TTLMemoryCache.js
```diff
@@ -0,0 +1,50 @@
+const TTLCache = require('@isaacs/ttlcache');
```
naz, line 1: I'm working on moving this cache to a separate package as it doesn't rly belong in core.

ghost/core/core/server/services/tags-public/service.js
```diff
@@ -0,0 +1,31 @@
...
+            return;
+        }
+
+        // Wire up all the dependencies
+        const models = require('../../models');
+        const adapterManager = require('../adapter-manager');
+        const config = require('../../../shared/config');
+
+        let tagsCache;
+        if (config.get('hostSettings:tagsPublicCache:enabled')) {
```
naz, line 14: Only initializes the cache for tags in hosted environment with a dedicated cache mechanism.

ghost/core/core/shared/config/defaults.json
```diff
@@ -28,6 +28,10 @@
             "active": "Memory",
             "settings": {},
             "imageSizes": {},
+            "tagsPublic": {
```
naz, line 31: This is not needed. Will remove it now!

PR #16255, opened 2023-02-09
🐛 Fixed subscriptions visible as "Active" within Ghost Admin

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2542 
fixes https://github.com/TryGhost/Team/issues/2543 
fixes https://github.com/TryGhost/Team/issues/2544

- Hides incomplete subscriptions
- Shows Past Due subscriptions
- Fixed UI issues with 3+ subscriptions
- Fixed missing complimentary subscription when one subscription was incomplete/inactive
- Fixed sending a paid subscription started email for incomplete subscriptions. This change also required us to actually send the email when the incomplete subscription eventually becomes active. So the introduction of a new `SubscriptionActivatedEvent` made sense/was required (because sending a SubscriptionCreatedEvent again would cause other issues).

### DIFF
--- a/ghost/admin/app/styles/layouts/members.css
+++ b/ghost/admin/app/styles/layouts/members.css
@@ -2298,11 +2298,6 @@ p.gh-members-import-errordetail:first-of-type {
     font-size: 1.2rem;
 }
 
-.gh-membertier-subscription span.pink {
-    color: var(--pink);
-    background: color-mod(var(--pink) a(20%));
-}
-
 .gh-cp-membertier.multiple-subs .gh-membertier-subscription {
     margin-top: 12px;
     padding-top: 12px;
--- a/ghost/admin/app/styles/layouts/members.css
+++ b/ghost/admin/app/styles/layouts/members.css
@@ -2298,6 +2298,11 @@ p.gh-members-import-errordetail:first-of-type {
     font-size: 1.2rem;
 }
 
+.gh-membertier-subscription span.pink {
+    color: var(--pink);
+    background: color-mod(var(--pink) a(20%));
+}
+
 .gh-cp-membertier.multiple-subs .gh-membertier-subscription {
     margin-top: 12px;
     padding-top: 12px;
@@ -2564,9 +2569,10 @@ p.gh-members-import-errordetail:first-of-type {
 .gh-cp-membertier-attribution .gh-tier-card-header {
     display: flex;
     align-items: flex-start;
+    position: relative;
 }
 
-.gh-cp-membertier-attribution .gh-tier-card-header:nth-of-type(2) {
+.gh-cp-membertier-attribution .gh-tier-card-header:not(:first-child) {
     margin-top: 16px;
     border-top: 1px solid var(--whitegrey);
     padding-top: 16px;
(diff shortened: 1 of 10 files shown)


ghost/admin/app/styles/layouts/members.css
```diff
@@ -2564,9 +2564,10 @@ p.gh-members-import-errordetail:first-of-type {
 .gh-cp-membertier-attribution .gh-tier-card-header {
     display: flex;
     align-items: flex-start;
+    position: relative;
 }
 
-.gh-cp-membertier-attribution .gh-tier-card-header:nth-of-type(2) {
+.gh-cp-membertier-attribution .gh-tier-card-header:not(:first-child) {
```
SimonBackx, line 2570: This change adds support for more than 2 subscriptions in a list in admin (line and margin was missing)

ghost/members-api/lib/services/member-bread.js
```diff
@@ -64,9 +64,12 @@ module.exports = class MemberBREADService {
         }
 
         const subscriptionProducts = (member.subscriptions || [])
-            .filter(sub => sub.status !== 'canceled')
+            .filter(sub => this.memberRepository.isActiveSubscriptionStatus(sub.status))
```
SimonBackx, line 67: This fixes that we also add a complimentary plan if all subscriptions are inactive, but we do have an active tier

allouis: 👍

ghost/members-api/lib/repositories/member.js
```diff
@@ -1031,6 +1031,24 @@ module.exports = class MemberRepository {
                     currency: subscriptionPriceData.currency,
                     mrr_delta: mrrDelta
                 }, options);
+
+                // Did we activate this subscription?
+                // This happens when an incomplete subscription is completed
+                // This always happens during the 3D secure flow, so it is important to catch
+                if (originalStatus !== 'active' && updatedStatus === 'active') {
```
allouis, line 1038: Do we want to be specific here and do:
```
if (originalStatus === 'incomplete' && updatedStatus === 'active') {
```

Are there other cases we want to catch? My thinking here is moving from `'trialing'` to `'active'` gonna be caught with this too

allouis (commented):
One comment - let me know what you think! Otherwise 👍

ghost/members-api/lib/repositories/member.js
SimonBackx: Luckily the `getStatus` method already converts `trailing` to `active` here. So checking for `incomplete` would also not really work unless we change that. 

Because with the change to `SubscriptionActivatedEvent`, we only send an initial email when creating the subscription in the database if  `isActiveSubscriptionStatus` is true (previously we did that always). So then this change catches all the situations where they didn't got that initial email, but now they are.

So the cases where it changes VS previously is when people canceled their subscription and then later reactivate them.

allouis: Aha - I missed that this was our internal statuses rather than Stripe - that all makes sense 👍

PR #16259, opened 2023-02-10
Optimized saving relations when order has not changed

daniellockyer:
- if the order of the relations is the same, we don't need to update the `sort_order` on the pivot, which we currently do all the time
- this saves an `update` statement for every relation item a model has


naz (approved):
Works like a charm!

SimonBackx (approved):

daniellockyer:
Closing as tests reveal a bug I'm not sure how to fix right now

PR #16260, opened 2023-02-10
🐛 Fixed storing email failures with an empty message

SimonBackx:
no issue

- When we receive an email failure with an empty message, the saving of the model would fail because of schema validation that requires strings to be non-empty.
- This adds more logging to the email analytics service to help debug future issues
- Performance improvement to storing delivered, opened and failed emails by replacing COALESCE with WHERE X IS NULL (tested and should give a decent performance boost locally).

### DIFF
--- a/ghost/email-analytics-service/lib/email-analytics-service.js
+++ b/ghost/email-analytics-service/lib/email-analytics-service.js
@@ -1,5 +1,6 @@
 const EventProcessingResult = require('./event-processing-result');
 const debug = require('@tryghost/debug')('services:email-analytics');
+const logging = require('@tryghost/logging');
 
 /**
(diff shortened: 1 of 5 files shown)


ghost/email-analytics-service/lib/email-analytics-service.js
```diff
@@ -176,9 +177,12 @@ module.exports = class EmailAnalyticsService {
     }
 
     async aggregateStats({emailIds = [], memberIds = []}) {
+        logging.info(`Aggregating email analytics for ${emailIds.length} emails`);
         for (const emailId of emailIds) {
             await this.aggregateEmailStats(emailId);
         }
+
+        logging.info(`Aggregating email analytics for ${memberIds.length} members`);
```
SimonBackx, line 185: This one is specifically taking a long time

PR #16261, opened 2023-02-10
Upgraded got package from v9.6.0 to v11.8.6

9larsons:
Refs TryGhost/Team#2459
-upgraded got from v9.6.0 to v11.8.6 to support following redirects (and other fixes)
-got v12+ requires ESM, so we do not want to upgrade further at this time
-required changes to a few libraries that use externalRequests
-mention discovery service tests updated to test for follow redirects

### DIFF
--- a/ghost/core/test/unit/server/lib/request-external.test.js
+++ b/ghost/core/test/unit/server/lib/request-external.test.js
@@ -279,10 +279,13 @@ describe('External Request', function () {
                 throw new Error('Request should have rejected with invalid url message');
             }, (err) => {
                 should.exist(err);
+                // got v11+ throws an error instead of the external requests lib
                 err.message.should.be.equal('No URL protocol specified');
             });
         });
 
+        // with got v11, for some reason this went from a GotError HTTPError with rich details to a shitty HTTPError with no data or even code
+        // TODO: maybe try a higher version? or see if this happened with v10
         it('[failure] can handle an error with statuscode not 200', function () {
             const url = 'http://nofilehere.com/files/test.txt';
             const options = {
@@ -300,7 +303,9 @@ describe('External Request', function () {
             }, (err) => {
                 requestMock.isDone().should.be.true();
                 should.exist(err);
-                err.response.statusMessage.should.be.equal('Not Found');
+                // console.log(err);
+                // TODO: no error code or status message
+                err.code.should.be.equal(`ERR_NON_2XX_3XX_RESPONSE`);
             });
         });
 
@@ -322,9 +327,11 @@ describe('External Request', function () {
             }, (err) => {
                 requestMock.isDone().should.be.true();
                 should.exist(err);
-                err.response.statusMessage.should.be.equal('Internal Server Error');
-                err.response.body.should.match(/something awful happened/);
-                err.response.body.should.match(/AWFUL_ERROR/);
+                // TODO: checking in on status code options... Got seems to throw a generic code
+                err.code.should.be.equal(`ERR_NON_2XX_3XX_RESPONSE`);
+                // err.statusMessage.should.be.equal('Internal Server Error');
+                // err.body.should.match(/something awful happened/);
+                // err.body.should.match(/AWFUL_ERROR/);
             });
         });
     });
--- a/ghost/core/test/unit/server/lib/request-external.test.js
+++ b/ghost/core/test/unit/server/lib/request-external.test.js
@@ -303,9 +303,7 @@ describe('External Request', function () {
             }, (err) => {
                 requestMock.isDone().should.be.true();
                 should.exist(err);
-                // console.log(err);
-                // TODO: no error code or status message
-                err.code.should.be.equal(`ERR_NON_2XX_3XX_RESPONSE`);
+                err.response.statusMessage.should.be.equal('Not Found');
             });
         });
 
@@ -327,11 +325,7 @@ describe('External Request', function () {
             }, (err) => {
                 requestMock.isDone().should.be.true();
                 should.exist(err);
-                // TODO: checking in on status code options... Got seems to throw a generic code
-                err.code.should.be.equal(`ERR_NON_2XX_3XX_RESPONSE`);
-                // err.statusMessage.should.be.equal('Internal Server Error');
-                // err.body.should.match(/something awful happened/);
-                // err.body.should.match(/AWFUL_ERROR/);
+                err.response.statusMessage.should.be.equal(`Internal Server Error`);
             });
         });
     });
--- a/ghost/core/test/unit/server/lib/request-external.test.js
+++ b/ghost/core/test/unit/server/lib/request-external.test.js
@@ -263,7 +263,7 @@ describe('External Request', function () {
                 throw new Error('Request should have rejected with invalid url message');
             }, (err) => {
                 should.exist(err);
-                err.message.should.be.equal('URL empty or invalid.');
+                err.message.should.be.equal('Invalid URL');
             });
         });
 
@@ -279,7 +279,7 @@ describe('External Request', function () {
                 throw new Error('Request should have rejected with invalid url message');
             }, (err) => {
                 should.exist(err);
-                err.message.should.be.equal('URL empty or invalid.');
+                err.message.should.be.equal('No URL protocol specified');
             });
         });
 
@@ -300,7 +300,7 @@ describe('External Request', function () {
             }, (err) => {
                 requestMock.isDone().should.be.true();
                 should.exist(err);
-                err.statusMessage.should.be.equal('Not Found');
+                err.response.statusMessage.should.be.equal('Not Found');
             });
         });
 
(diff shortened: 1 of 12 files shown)


9larsons:
@SimonBackx would definitely appreciate a second set of eyes. I have run tests but not functional testing to make sure stuff works. I'll aim to do some of that on Monday. I've had to update tests for features I don't know much about so some functional testing would be helpful. 

Most of the pain came from bugs introduced in the new `got` versions and refactoring the checks in the `oembed-service`.

ghost/core/test/unit/server/lib/request-external.test.js
```diff
@@ -300,7 +303,9 @@ describe('External Request', function () {
             }, (err) => {
                 requestMock.isDone().should.be.true();
                 should.exist(err);
-                err.statusMessage.should.be.equal('Not Found');
+                // console.log(err);
+                // TODO: no error code or status message
+                err.code.should.be.equal(`ERR_NON_2XX_3XX_RESPONSE`);
```
SimonBackx, line 308: According to the documentation (https://github.com/sindresorhus/got/tree/v11#gothttperror), we should be able to access the status etc via `response`

```suggestion
err.response.statusMessage.should.be.equal('Not Found');
```

Does that work? I so, I would clean this up for the other tests too and remove the comments.

ghost/oembed-service/lib/oembed-service.js
```diff
@@ -68,16 +67,7 @@ class OEmbed {
...
-            if (this.isIpOrLocalhost(url)) {
-                return this.unknownProvider(url);
-            }
-            const response = await externalRequest(url, requestConfig);
-            if (this.isIpOrLocalhost(response.url)) {
-                return this.unknownProvider(url);
-            }
-            return response;
-        };
+        this.externalRequest = externalRequest; // changed this because this is already happening in externalRequest + we can't await because we lose .json() possibility on promise
```
SimonBackx, line 70: Gonna remove my own debug comment here 😊
```suggestion
        this.externalRequest = externalRequest;
```

ghost/oembed-service/lib/oembed-service.js
```diff
@@ -184,12 +171,9 @@ class OEmbed {
...
     async fetchPageJson(url) {
-        const {body, url: pageUrl} = await this.fetchPage(
-            url,
-            {
-                json: true
-            });
-
+        const res = await this.fetchPage(url, {}); // .json() doesn't seem to parse appropriately... bug with got v10?
+        const body = JSON.parse(res.body);
+        const pageUrl = res.url;
```
SimonBackx, line 176:
```suggestion
        const res = await this.fetchPage(url, {responseType: 'json'});
        const body = res.body;
        const pageUrl = res.url;
```
Gonna check this

ghost/webmentions/lib/MentionSendingService.js
```diff
@@ -74,22 +74,24 @@ module.exports = class MentionSendingService {
 
     async send({source, target, endpoint}) {
         logging.info('[Webmention] Sending webmention from ' + source.href + ' to ' + target.href + ' via ' + endpoint.href);
+        
         const response = await this.#externalRequest.post(endpoint.href, {
-            body: {
+            json: {
```
SimonBackx, line 79: The spec explicitly requires `application/x-www-form-urlencoded`, not json

ghost/webmentions/lib/MentionSendingService.js
```diff
@@ -74,22 +74,24 @@ module.exports = class MentionSendingService {
...
                 source: source.href,
                 target: target.href,
                 source_is_ghost: true
             },
-            form: true,
             throwHttpErrors: false,
             maxRedirects: 10,
             followRedirect: true,
             methodRewriting: false, // WARNING! this setting has a different meaning in got v12!
```
SimonBackx, line 87:
```suggestion
```
Default is true in Got v11, which we want. In got v12 the default is false, but the meaning changed to match that of got v11. Very confusing but the meaning we need here is that a POST should stay a POST after a redirect

ghost/webmentions/test/MentionSendingService.test.js
```diff
@@ -352,31 +352,45 @@ describe('MentionSendingService', function () {
...
+            const target = new URL('https://target.com/target');
+            const endpoint = new URL('https://example.org/webmentions-test');
             const scope = nock('https://example.org')
                 .persist()
-                .post('/webmentions-test', `source=${encodeURIComponent('https://example.com/source')}&target=${encodeURIComponent('https://target.com/target')}&source_is_ghost=true`)
+                .post('/webmentions-test', {
+                    source: source.href,
+                    target: target.href,
+                    source_is_ghost: true
+                })
```
SimonBackx, line 364: The test was right here (see other comment)

ghost/webmentions/test/MentionSendingService.test.js
```diff
@@ -426,41 +440,6 @@ describe('MentionSendingService', function () {
...
-        //    const service = new MentionSendingService({externalRequest});
-        //    await service.send({
-        //        source: new URL('https://example.com'),
-        //        target: new URL('https://example.com'),
-        //        endpoint: new URL('https://example.org/webmentions-test')
-        //    });
-        //    assert(scope.isDone());
-        //    assert(scope2.isDone());
-        //});
-        // TODO: also check if we don't follow private IPs after redirects
```
SimonBackx, line 453: I think we can uncomment the test, because it should pass

ghost/oembed-service/lib/oembed-service.js
SimonBackx: I think that should work, because .json() automatically returns the body instead of the response

ghost/core/test/unit/server/lib/request-external.test.js
9larsons: Awesome, yes. Not sure why I wasn't seeing that last week.

ghost/webmentions/lib/MentionSendingService.js
9larsons: https://github.com/sindresorhus/got/blob/v11.8.6/readme.md#methodrewriting for ref
I saw some discussion within `got` docs/issues that `post` requests should not be redirected. Not sure if you have any background on that, but that added to the confusion for me here.

ghost/webmentions/test/MentionSendingService.test.js
9larsons: Pulling this comment down here.

I saw some discussion within got docs/issues that post requests should not be redirected. Not sure if you have any background on that, but that added to the confusion for me here.

https://github.com/sindresorhus/got/issues/216#issuecomment-237610929

This is what has me concerned. I didn't get that test to work, and I created a small local npm repo with just `got` to try testing it and wasn't able to get anywhere with successfully following a redirect. I get an `ERR_NOCK_NO_MATCH` error. I haven't tried digging through the `got` codebase yet to track down why. Hoping you might have some insight into this before going down that path.

ghost/webmentions/lib/MentionSendingService.js
SimonBackx: Correct, in a browser context a POST request generally gets rewritten as a GET. But the webmentions spec (not in a browser context) explicitly states that we should keep the POST method (otherwise it would not work since the body would get lost).

SimonBackx: Took me a while to figure it out, but there was an error in the commented test:
```
it('Can handle redirect responses', async function () {
            const scope = nock('https://example.org')
                .persist()
                .post('/webmentions-test')
                .reply(302, '', {
                    Location: 'https://example.org/webmentions-test-2'
                });
            const scope2 = nock('https://example.org')
                .persist()
                .post('/webmentions-test-2')
                .reply(201);

            const service = new MentionSendingService({externalRequest});
            const res = await service.send({
                source: new URL('https://example.com'),
                target: new URL('https://example.com'),
                endpoint: new URL('https://example.org/webmentions-test')
            });
            assert(scope.isDone());
            assert(scope2.isDone());
        });
```
The location header was not send by Nock because of this, so that caused the issue. Tests should pass with this change 😊

ghost/webmentions/test/MentionSendingService.test.js
SimonBackx: See other comment

ghost/core/test/unit/server/lib/request-external.test.js
```diff
@@ -279,10 +279,13 @@ describe('External Request', function () {
...
             }, (err) => {
                 should.exist(err);
-                err.message.should.be.equal('URL empty or invalid.');
+                // got v11+ throws an error instead of the external requests lib
+                err.message.should.be.equal('No URL protocol specified');
             });
         });
 
+        // with got v11, for some reason this went from a GotError HTTPError with rich details to a shitty HTTPError with no data or even code
+        // TODO: maybe try a higher version? or see if this happened with v10
```
SimonBackx, line 288:
```suggestion
```

SimonBackx:
@9larsons I think there is one outstanding test that needs to be readded. After that this PR is good to merge.

ghost/core/test/unit/server/lib/request-external.test.js
9larsons: 😅 I'll just remove the comment altogether. Your suggestion about changing to use the response error code works how we need.

PR #16264, opened 2023-02-13
Added verified mention column

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/2549

- Added `verified` column to mentions table
- Added a new migration
- Updated schema
- Updated integrity test
- Added verified to model and bookshelf mention repo to use the new column


SimonBackx (approved):
This probably won't make it for 5.34, so the version might need to be changed before merging

ronaldlangeveld:
> This probably won't make it for 5.34, so the version might need to be changed before merging

Awesome, thanks, will bump it up a version before merging to have it ready for the release after. 👌

ghost/webmentions/lib/Mention.js
```diff
@@ -95,6 +95,12 @@ module.exports = class Mention {
         return this.#sourceFeaturedImage;
     }
 
+    /** @type {boolean} */
+    #verified;
```
allouis, line 99: This was already implemented as part of https://github.com/TryGhost/Team/issues/2548

ghost/core/core/server/services/mentions/BookshelfMentionRepository.js
```diff
@@ -107,7 +108,8 @@ module.exports = class BookshelfMentionRepository {
             resource_id: mention.resourceId?.toHexString(),
             resource_type: mention.resourceId ? 'post' : null,
             payload: mention.payload ? JSON.stringify(mention.payload) : null,
-            deleted: Mention.isDeleted(mention)
+            deleted: Mention.isDeleted(mention),
+            verified: mention.verified || false
```
allouis, line 112: I don't think we need this check - the Mention entity always has a `boolean` for the verified property

ghost/core/core/server/services/mentions/BookshelfMentionRepository.js
```diff
@@ -54,7 +54,8 @@ module.exports = class BookshelfMentionRepository {
             sourceAuthor: model.get('source_author'),
             sourceExcerpt: model.get('source_excerpt'),
             sourceFavicon: model.get('source_favicon'),
-            sourceFeaturedImage: model.get('source_featured_image')
+            sourceFeaturedImage: model.get('source_featured_image'),
+            verified: model.get('verified')
```
allouis, line 58: This doesn't currently work - we need to work out the best way to pass the `verified` property in

allouis (commented):
I think we should remove the repository changes from this PR, 1. because migrations PR's should be as small as possible, but 2. because the Entity changes haven't been finished so we're unsure of the API we need to use!

ghost/webmentions/lib/Mention.js
ronaldlangeveld: Makes sense, thanks. Will remove

ronaldlangeveld: Oops, thanks. It wasn't yet on my branch at the time.

allouis (approved):

daniellockyer (approved):

PR #16271, opened 2023-02-14
[POC] Refactored milestones service to preserve API instance

allouis:
This splits the `initAndRun` method into an `init` and a `run` method.

The `init` method follows the same pattern as other services where it will only instantiate once, meaning that `initAndRun` can be called multiple times without blowing away any in-memory resources.


aileen:
Included in https://github.com/TryGhost/Ghost/pull/16272

PR #16273, opened 2023-02-14
Implemented email analytics retrying

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2562

New event fetching loops:
- Reworked the analytics fetching algorithm. Instead of starting again where we stopped during the last fetching minus 30 minutes, we now just continue where we stopped. But with ms precision (because no longer database dependent after first fetch), and we stop at NOW - 1 minute to reduce chance of missing events.
- Apart from that, a missing fetching loop is introduced. This fetches events that are older than 30 minutes, and just processes all events a second time to make sure we didn't skip any because of storage delays in the Mailgun API.
- A new scheduled fetching loop, that allows us to schedule between a given start/end date (currently only persisted in memory, so stops after a reboot)

UI and endpoint changes:
- New UI to show the state of the analytics 'loops'
- New endpoint to request the analytics loop status
- New endpoint to schedule analytics
- New endpoint to cancel scheduled analytics
- Some number formatting improvements, and introduction of 'opened' count in debug screen
- Live reload of data in the debug screen

Other changes:
- This also improves the support for maxEvents. We can now stop a fetching loop after x events without worrying about lost events. This is used to reduce the fetched events in the missing and scheduled event loop (e.g. when the main one is fetching lots of events, we skip the other loops).
- Prevents fetching the same events over and over again if no new events come in (because we always started at the same begin timestamp). The code increases the begin timestamp with 1 second if it is safe to do so, to prevent the API from returning the same events over and over again.
- Some optimisations in handing the processing results (less merges to reduce CPU usage in cases we have lots of events).

Testing:
- You can test with lots of events using the new mailgun mocking server (Toolbox repo `scripts/mailgun-mock-server`). This can also simulate events that are only returned after x minutes because of storage delays.

### DIFF
--- a/ghost/email-analytics-service/lib/email-analytics-service.js
+++ b/ghost/email-analytics-service/lib/email-analytics-service.js
@@ -222,7 +222,8 @@ module.exports = class EmailAnalytics {
             await this.processEventBatch(events, processingResult, fetchData);
             eventCount += events.length;
 
-            // Every 5 minutes we do an aggregation and clear the processingResult
+            // Every 5 minutes or 5000 members we do an aggregation and clear the processingResult
+            // Otherwise we need to loop a lot of members afterwards, and this takes too long without updating the stat counts in between
             if (Date.now() - lastAggregation > 5 * 60 * 1000 || processingResult.memberIds.length > 5000) {
                 // Aggregate and clear the processingResult
                 // We do this here because otherwise it could take a long time before the new events are visible in the stats
--- a/ghost/email-analytics-service/lib/email-analytics-service.js
+++ b/ghost/email-analytics-service/lib/email-analytics-service.js
@@ -1,5 +1,6 @@
 const EventProcessingResult = require('./event-processing-result');
 const logging = require('@tryghost/logging');
+const errors = require('@tryghost/errors');
 
 /**
  * @typedef {import('@tryghost/email-service').EmailEventProcessor} EmailEventProcessor
@@ -11,6 +12,10 @@ const logging = require('@tryghost/logging');
  * @property {Date} [lastStarted] Date the last fetch started on
  * @property {Date} [lastBegin] The begin time used during the last fetch
  * @property {Date} [lastEventTimestamp]
+ */
+
+/**
+ * @typedef {FetchData & {schedule: {begin: Date, end: Date}}} FetchDataScheduled
  */
 
 const TRUST_THRESHOLD_MS = 30 * 60 * 1000; // 30 minutes
@@ -33,6 +38,11 @@ module.exports = class EmailAnalytics {
     #fetchMissingData = null;
 
     /**
+     * @type {FetchDataScheduled}
+     */
+    #fetchScheduledData = null;
+
+    /**
      * @param {object} dependencies
      * @param {EmailEventProcessor} dependencies.eventProcessor
      */
@@ -42,6 +52,14 @@ module.exports = class EmailAnalytics {
         this.queries = queries;
         this.eventProcessor = eventProcessor;
         this.providers = providers;
+    }
+
+    getStatus() {
+        return {
+            latest: this.#fetchLatestData,
+            missing: this.#fetchMissingData,
+            scheduled: this.#fetchScheduledData
+        };
     }
 
     /**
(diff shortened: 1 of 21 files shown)


ghost/mailgun-client/lib/mailgun-client.js
```diff
@@ -107,13 +110,19 @@ module.exports = class MailgunClient {
         }
     }
 
+    /**
+     * Fetches events from Mailgun
+     * @param {Object} mailgunOptions
+     * @param {Function} batchHandler
+     * @param {Object} options
+     * @param {Number} options.maxEvents Not a strict maximum. We stop fetching after we reached the maximum AND received at least one event after begin (not equal) to prevent deadlocks.
+     * @returns {Promise<void>}
```
SimonBackx, line 119: The result has been removed because it didn't make much sense as we could also access the results by a batchHandler implementation. That reduces the need to concat the array on every loop iteration. It is also a lot more flexible to work with inside the batchHandler.

ghost/mailgun-client/lib/mailgun-client.js
```diff
@@ -132,16 +141,14 @@ module.exports = class MailgunClient {
...
             while (events.length !== 0) {
-                const batchResult = await batchHandler(events);
-
-                result = result.concat(batchResult);
+                await batchHandler(events);
                 eventCount += events.length;
 
-                if (eventCount >= maxEvents) {
-                    break pagesLoop;
+                if (eventCount >= maxEvents && (!beginTimestamp || !events[events.length - 1].timestamp || (events[events.length - 1].timestamp.getTime() > beginTimestamp))) {
```
SimonBackx, line 150: This fixes the deadlock issue, if maxEvents is 600, and we receive 600 events with the same timestamp as the begin timestamp, we would keep fetching them over and over again.

SimonBackx: Luckily, Mailgun timestamps have ms precision, so this shouldn't happen a lot + if it happens it should be limited in extra events

ghost/email-analytics-service/lib/email-analytics-service.js
```diff
@@ -25,72 +56,270 @@ module.exports = class EmailAnalyticsService {
         this.providers = providers;
     }
 
-    async fetchAll() {
```
SimonBackx, line 28: fetchAll was never used

ghost/email-analytics-service/lib/email-analytics-service.js
```diff
@@ -25,72 +56,270 @@ module.exports = class EmailAnalyticsService {
...
+        // Store that we started fetching
+        fetchData.running = true;
+        fetchData.lastStarted = new Date();
+        fetchData.lastBegin = begin;
+
+        let lastAggregation = Date.now();
+        let eventCount = 0;
+
+        // We keep the processing result here, so we also have a result in case of failures
+        let processingResult = new EventProcessingResult();
```
SimonBackx, line 217: The way this works has changed. In the past, we kept on creating new `EventProcessingResult`s and passing and merging them around. IMO with the number of events that caused a lot of CPU overhead (array merges + unique filters on those arrays). Instead, all events are appended to one EventProcessingResult. Apart from that `EventProcessingResult` is reset every 5 minutes after aggregating the related emails and members stats. That keeps it from becoming too big + that we have to wait a long time before we start aggregating the data (= stats become visible for the publishers).

SimonBackx: This is also the reason the processingResult is not returned any longer and why the aggregation moved to the analytics service (instead of the wrapper). Also in case of fatal errors during processing, we can still aggregate the events that we received before that error (otherwise we could have out of date aggregated stats in emails and members).

ghost/email-service/lib/email-event-processor.js
```diff
@@ -82,10 +82,8 @@ class EmailEventProcessor {
                 emailId: recipient.emailId,
                 timestamp
             });
-            await this.#eventStorage.handleOpened(event);
-
             this.#domainEvents.dispatch(event);
-            await waitForEvent(); // Avoids knex connection pool to run dry
```
SimonBackx, line 88: The waitForEvent delayed processing a batch of 300 events from 1.5s to 20s locally.

ghost/email-analytics-service/lib/email-analytics-service.js
```diff
@@ -25,67 +56,265 @@ module.exports = class EmailAnalyticsService {
...
-            return result;
+    async fetchLatest({maxEvents = Infinity} = {}) {
+        // Start where we left of, or the last stored event in the database, or start 30 minutes ago if we have nothing available
+        const begin = await this.getLastEventTimestamp();
+        const end = new Date(Date.now() - FETCH_LATEST_END_MARGIN_MS); // ALways stop at x minutes ago to give Mailgun a bit more time to stabilize storage
+
+        if (end < begin) {
+            // Skip for now
+            logging.info('[EmailAnalytics] Skipping fetchLatest because end (' + end + ') is before begin (' + begin + ')');
+            //return 0;
```
SimonBackx, line 82: Commented out for the tests. Need to find a better way to make this work with tests

SimonBackx:
Sorry to bother everyone with a review. But given the size and risks of the changes, wanted to make sure the logic is okay.

ghost/email-analytics-service/lib/email-analytics-service.js
```diff
@@ -25,67 +56,265 @@ module.exports = class EmailAnalyticsService {
...
-    async fetchLatest({maxEvents = Infinity} = {}) {
-        const result = new EventProcessingResult();
+    /**
+     * Fetches events that are older than 30 minutes, because then the 'storage' of the Mailgun API is stable. And we are sure we don't miss any events.
+     * @param {object} options
+     * @param {number} [options.maxEvents] Not a strict maximum. We stop fetching after we reached the maximum AND received at least one event after begin (not equal) to prevent deadlocks.
+     */
+    async fetchMissing({maxEvents = Infinity} = {}) {
+        // We start where we left of, or 30 minutes before the last event we received
+        const begin = this.#fetchMissingData?.lastEventTimestamp ?? new Date(Date.now() - TRUST_THRESHOLD_MS * 2);
```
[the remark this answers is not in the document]
SimonBackx: No, this comment is out of date.
```suggestion
        // We start where we left of, or 1 hour ago if we never fetched a missing event
        const begin = this.#fetchMissingData?.lastEventTimestamp ?? new Date(Date.now() - TRUST_THRESHOLD_MS * 2);
```

PR #16277, opened 2023-02-15
Milestones creation events

aileen:
no issue

- In preparation of using event emitting for Milestone achievements, we needed to add a dedicated `MilestoneCreatedEvent` to the `Milestone` entity.
- The event will be emitted using `DomainEvents` when a new milestone is saved, which will allow us to listen to these events.
- With the switch of using a `MilestoneCreatedEvent` we'll be decoupling the mailing functionality and not need `GhostMailer` as dependency in the package anymore

### DIFF
--- a/ghost/milestones/lib/InMemoryMilestoneRepository.js
+++ b/ghost/milestones/lib/InMemoryMilestoneRepository.js
@@ -13,6 +13,17 @@ module.exports = class InMemoryMilestoneRepository {
     /** @type {Object.<string, true>} */
     #ids = {};
 
+    /** @type {import('@tryghost/domain-events')} */
+    #DomainEvents;
+
+    /**
+     * @param {object} deps
+     * @param {import('@tryghost/domain-events')} deps.DomainEvents
+     */
+    constructor(deps) {
+        this.#DomainEvents = deps.DomainEvents;
+    }
+
     /**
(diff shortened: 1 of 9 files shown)


ghost/milestones/lib/InMemoryMilestoneRepository.js
```diff
@@ -27,6 +38,10 @@ module.exports = class InMemoryMilestoneRepository {
         } else {
             this.#store.push(milestone);
             this.#ids[milestone.id.toHexString()] = true;
+
+            for (const event of milestone.events) {
```
allouis, line 42: This should run outside of the `if/else` block - because we want to emit events for all saves.

Obviously ATM we only have an event when it's saving a new milestone - but I think for correctness it's better to move it :)

ghost/milestones/test/Milestone.test.js
```diff
@@ -110,5 +112,15 @@ describe('Milestone', function () {
 
             assert(milestone.name === 'members-100');
         });
+
+        it('Will create event for new milestone', async function () {
```
allouis, line 116: Might be worth adding explicit test that event isn't added for existing milestone - NABD if you don't wanna

allouis (commented):
Two minor comments - merge away after fixing

PR #16281, opened 2023-02-16
Slack notifications service for Milestones behind flag

aileen:

- Added a `slack-notifications` repository which handles sending Slack messages to a URL as defined in our Ghost(Pro) config (also includes a global switch to disable the feature if needed) and listens to `MilestoneCreatedEvents`.
- Added a `slack-notification` service which listens to the events on boot.
- In order to have access to further information such as the reason why a Milestone email hasn't been sent, or the current ARR or Member value as comparison to the achieved milestone, I added a `meta` object to the `MilestoneCreatedEvent` which then gets accessible by the event subscriber. This avoid doing further requests to the DB as we need to have this information in relation to the event occurred.


aileen:
Going to add an e2e server test as well

ghost/core/core/server/services/slack-notifications/service.js
```diff
@@ -0,0 +1,52 @@
+const DomainEvents = require('@tryghost/domain-events');
+const config = require('../../../shared/config');
+const labs = require('../../../shared/labs');
+const logging = require('@tryghost/logging');
+
+class SlackNotificationsServiceWrapper {
+    /** @type {import('@tryghost/slack-notifications/lib/SlackNotificationsService')} */
+    #api;
+
+    static create({siteUrl, isEnabled, webhookUrl}) {
```
aileen, line 10: Will add JSDoc here. Forgot!

aileen: @allouis not sure if `@returns {object} SlackNotificationsService` this is correct 😬

ghost/core/test/e2e-server/services/milestones.test.js
```diff
@@ -151,9 +151,6 @@ describe('Milestones Service', function () {
 
     beforeEach(async function () {
         sinon.createSandbox();
-        // TODO: stub out stripe mode
```
aileen, line 154: Left over todo, forgot to remove when I added the stubs for it

ghost/core/core/server/services/slack-notifications/service.js
allouis: Just `@returns {SlackNotificationsService}` ☺️

PR #16290, opened 2023-02-17
Added timeout for running milestone service

aileen:
no issue

- Instead of running milestone service directly on boot, set a random timeout of 0-4 days to run after boot
- Updated tests
- Service is still behind a beta flag

### DIFF
--- a/ghost/core/core/server/services/milestones/service.js
+++ b/ghost/core/core/server/services/milestones/service.js
@@ -1,7 +1,7 @@
 const DomainEvents = require('@tryghost/domain-events');
 const logging = require('@tryghost/logging');
 
-const JOB_TIMEOUT = 1000 * 60 * 60 * 24 * (Math.floor(Math.random() * 3)); // 0 - 4 days;
+const JOB_TIMEOUT = 1000 * 60 * 60 * 24 * (Math.floor(Math.random() * 4)); // 0 - 4 days;
 
 const getStripeLiveEnabled = () => {
     const settingsCache = require('../../../shared/settings-cache');
@@ -72,21 +72,12 @@ module.exports = {
     },
 
     /**
+     *
      * @param {number} [customTimeout]
      *
-     * @returns {Promise<object>}
+     *  @returns {Promise<object>}
      */
-    async initAndRun(customTimeout) {
-        /**
-        * @param {number} ms
-        * @returns {Promise<void>}
-        */
-        async function sleep(ms) {
-            return new Promise((resolve) => {
-                setTimeout(resolve, ms);
-            });
-        }
-
+    async scheduleRun(customTimeout) {
         const timeOut = customTimeout || JOB_TIMEOUT;
 
         const today = new Date();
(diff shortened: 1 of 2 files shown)


ghost/core/test/e2e-server/services/milestones.test.js
```diff
@@ -206,29 +225,42 @@ describe('Milestones Service', function () {
...
     it('Does not run when milestoneEmails labs flag is not set', async function () {
         mockManager.mockLabsDisabled('milestoneEmails');
 
-        const result = await milestonesService.initAndRun();
+        const resultPromise = milestonesService.initAndRun(10);
+        await clock.tickAsync(10);
+        const result = await resultPromise;
         assert(result === undefined);
+
+        assert(loggingStub.called);
```
aileen, line 249: No idea why, but the stub returns a call count much MUCH higher than I anticipate and also am receiving when running it locally 🤔

ghost/core/core/server/services/milestones/service.js
```diff
@@ -1,4 +1,7 @@
 const DomainEvents = require('@tryghost/domain-events');
+const logging = require('@tryghost/logging');
+
+const JOB_TIMEOUT = 1000 * 60 * 60 * 24 * (Math.floor(Math.random() * 3)); // 0 - 4 days;
```
allouis, line 4: This is 0-3 days, but the comment says 0-4

ghost/core/core/server/services/milestones/service.js
```diff
@@ -69,10 +72,33 @@ module.exports = {
     },
 
     /**
+     * @param {number} [customTimeout]
+     *
      * @returns {Promise<object>}
      */
-    async initAndRun() {
+    async initAndRun(customTimeout) {
```
allouis, line 79: Is this new param only for testing? Perhaps the number should come from config, or we can just use the fake clock in the tests to advance by 4 days 🤷‍♂️

ghost/core/core/server/services/milestones/service.js
```diff
@@ -69,10 +72,33 @@ module.exports = {
...
+        const today = new Date();
+        const msNow = today.getMilliseconds();
+        const newMs = msNow + timeOut;
+        const jobDate = today.setMilliseconds(newMs);
+
         await this.init();
+
+        logging.info(`Running milestone emails job on ${new Date(jobDate).toString()}`);
+
+        await sleep(timeOut);
```
allouis, line 101: I don't think we want the init method to wait for up to 3 days, I think it would be better to return from the function and have the code run in a timeout separately

allouis (approved):
Few comments but not major

ghost/core/core/server/services/milestones/service.js
aileen: Ah yeah, will update

ghost/core/core/server/services/milestones/service.js
aileen: Yes, this is just for testing purposes

aileen: Added a note. The custom timeout won't last long in the codebase, as we'll soon move to properly schedule the job. I left it in there for now 😅

ghost/core/core/server/services/milestones/service.js
aileen: Switched to use a separate fn which returns a promise with the timeout and doesn't wait for it to be resolved. This should unblock anything that would prevent `boot` from finishing.

PR #16305, opened 2023-02-21
Added `BookshelfMilestoneRepository` implementation

aileen:
needs https://github.com/TryGhost/Ghost/pull/16300

This stores the received milestones in the database.

### DIFF
--- a/ghost/core/core/server/models/milestone.js
+++ b/ghost/core/core/server/models/milestone.js
@@ -1,0 +1,9 @@
+const ghostBookshelf = require('./base');
+
+const Milestone = ghostBookshelf.Model.extend({
+    tableName: 'milestones'
+});
+
+module.exports = {
+    Milestone: ghostBookshelf.model('Milestone', Milestone)
+};
--- a/ghost/core/core/server/services/milestones/BookshelfMilestoneRepository.js
+++ b/ghost/core/core/server/services/milestones/BookshelfMilestoneRepository.js
@@ -1,0 +1,136 @@
+const {Milestone} = require('@tryghost/milestones');
+
+/**
+ * @typedef {import('@tryghost/milestones/lib/MilestonesService').IMilestoneRepository} IMilestoneRepository
+ * @typedef {import('@tryghost/milestones/lib/MilestonesService')} Milestone
+ */
+
+/**
+ * @implements {IMilestoneRepository}
+ */
+module.exports = class BookshelfMilestoneRepository {
+    /** @type {Object} */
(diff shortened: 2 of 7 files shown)


ghost/core/test/e2e-server/services/milestones.test.js
```diff
@@ -156,7 +157,8 @@ describe('Milestones Service', function () {
     beforeEach(async function () {
         loggingStub = sinon.stub(logging, 'info');
         const now = new Date();
-        clock = sinon.useFakeTimers(now.getTime());
+        const threeMonthsAgo = now.setMonth(now.getMonth() - 3);
```
allouis, line 160: The `setMonth` method mutates the `Date` instance, so after this line the `now` variable is actually three months ago. Obviously fine here but if someones changes the tests and tries to use the `now` value they'll get unexpected results!

Also the value of `new Date(threeMonthsAgo).getTime()` and `threeMonthsAgo` are the same! Date objects are freaking weird.

I would probably change this to something like:
```js
const threeMonthsAgo = new Date();
threeMonthsAgo.setMonth(threeMonthsAgo.getMonth() - 3);
clock = sinon.useFakeTimers(threeMonthsAgo.getTime());
```

ghost/core/test/e2e-server/services/milestones.test.js
```diff
@@ -11,6 +11,7 @@ const milestonesService = require('../../../core/server/services/milestones');
 let agent;
 let counter = 0;
 let membersCounter = 0;
+const inFifteenDays = 1000 * 60 * 60 * 24 * 15;
```
allouis, line 14: I don't really care, but we would generally call this `fifteenDays` without the `'in'` prefix

allouis (approved):
The repository looks good - I'm unsure why the tests have changed though they should all be independent of any storage concerns 🤔 

If possible I would try and separate out these changes into separate commits/PR - if they can't be separated I'd be interested to know what we need to change the tests, it might be a sign we're too tightly coupled

ghost/core/test/e2e-server/services/milestones.test.js
aileen: Moved test updates to https://github.com/TryGhost/Ghost/pull/16314

ghost/core/test/e2e-server/services/milestones.test.js
aileen: Moved test updates to https://github.com/TryGhost/Ghost/pull/16314

PR #16307, opened 2023-02-21
Added `MilestoneCreatedEvent` to staff service notifications

aileen:

- Added MilestoneCreatedEvent subscription to staff-service incl. first handling
- Logs information for now instead of actually sending an email

### DIFF
--- a/ghost/staff-service/lib/emails.js
+++ b/ghost/staff-service/lib/emails.js
@@ -194,6 +194,36 @@ class StaffServiceEmails {
         }
     }
 
+    async notifyMilestoneReceived({milestone}) {
+        const users = await this.models.User.getEmailAlertUsers('milestone-received');
+        for (const user of users) {
+            const to = user.email;
+            const milestoneTypePretty = milestone.type === 'arr' ? 'ARR' : 'Members';
+            const subject = `🏆 New ${milestoneTypePretty} Milestone achieved!`;
+
+            const templateData = {
+                siteTitle: this.settingsCache.get('title'),
+                siteUrl: this.urlUtils.getSiteUrl(),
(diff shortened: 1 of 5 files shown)


ghost/staff-service/lib/emails.js
```diff
@@ -190,6 +190,36 @@ class StaffServiceEmails {
         }
     }
 
+    async notifyMilestoneReceived({milestone}) {
+        const users = await this.models.User.getEmailAlertUsers('milestone-received');
+        for (const user of users) {
```
allouis, line 195: I know you're following the existing pattern here - but I think it might be wrong...

We send an email to each user one after the other, and we don't catch any errors, so if one email fails, the remaining unsent emails will not be sent.

I'm not sure if that is the behaviour you want? If it is - ignore this 😁 

I would imagine that we either want to:
1. Continue with sending the emails in series but wrap the email sending in a try/catch and log any errors, but continue to try sending to the remaining users.
2. Send the emails in parallel, and use `Promise.allSettled` to log any errors, which'd look something like:

```js
const emailPromises = [];
for (const user of users) {
  // stuff...
  emailPromises.push(this.sendMail({...}));
}

const results = Promise.allSettled(emailPromises);

for (const result of results) {
  if (result.status === 'rejected') {
    logging.warn(result.reason);
  }
}
```

allouis (commented):
Looking good! Lil comment about sending logic :)

ghost/staff-service/lib/emails.js
aileen: Moving this handling into a new PR, as we'll take some more time to create all the templates.

SimonBackx (approved):

PR #16308, opened 2023-02-21
Implemented retrying for sending email via Mailgun API

SimonBackx:
no issue

Retry sending an email up to 5 times if it failed.


ghost/email-service/lib/batch-sending-service.js
```diff
@@ -77,6 +79,13 @@ class BatchSendingService {
         if (AFTER_RETRY_CONFIG) {
             this.#AFTER_RETRY_CONFIG = AFTER_RETRY_CONFIG;
         }
+        if (MAILGUN_API_RETRY_CONFIG) {
+            this.#MAILGUN_API_RETRY_CONFIG = MAILGUN_API_RETRY_CONFIG;
+        } else {
+            if (process.env.NODE_ENV.startsWith('test')) {
+                this.#MAILGUN_API_RETRY_CONFIG = {maxRetries: 0};
```
SimonBackx, line 86: Ideally we would only reduce the sleep, but E2E tests currently depend on no retries to test other logic

PR #16316, opened 2023-02-22
Improved resilience of `adapter-cache-memory-ttl` tests

daniellockyer:
- when CI is slow, these tests can erroneously fail
- we should prefer to use `sinon.useFakeTimers` over awaiting a Promise with a timeout for better resilience against slow machines


daniellockyer:
Closing as this was becoming a dead end

PR #16318, opened 2023-02-22
Milestone emails templates and sending implementation

aileen:

- Added email template for milestones with using a configuration file for different member milestone values, as we're sending different content for each one
- Implement sending the email to users who have `milestone-notifications` enabled, currently still behind a flag

---
### Todos before merging:
- [x] Finish email template design
- [x] Ship with public beta

### DIFF
--- a/ghost/staff-service/lib/emails.js
+++ b/ghost/staff-service/lib/emails.js
@@ -187,13 +187,50 @@ class StaffServiceEmails {
      * @returns {Promise<void>}
      */
     async notifyMilestoneReceived({milestone}) {
-        const users = await this.models.User.getEmailAlertUsers('milestone-received');
+        if (!milestone?.emailSentAt || (milestone?.meta && milestone?.meta?.reason)) {
+            // Do not send an email when no email was set to be sent or a reason
+            // not to send provided
+            return;
+        }
 
-        // TODO: send email with correct templates
+        const emailPromises = [];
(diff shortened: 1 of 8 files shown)


ghost/staff-service/lib/emails.js
```diff
@@ -170,13 +170,50 @@ class StaffServiceEmails {
      * @returns {Promise<void>}
      */
     async notifyMilestoneReceived({milestone}) {
+        if (!milestone?.emailSentAt || (milestone?.meta && milestone?.meta?.reason)) {
```
JBTAdmin, line 173: Do we really need `milestone?.meta` before && operator?

ghost/staff-service/lib/emails.js
```diff
@@ -170,13 +170,50 @@ class StaffServiceEmails {
...
+            // Do not send an email when no email was set to be sent or a reason
+            // not to send provided
+            return;
+        }
+
+        const emailPromises = [];
         const users = await this.models.User.getEmailAlertUsers('milestone-received');
+        const formattedValue = this.getFormattedAmount({currency: milestone?.currency, amount: milestone.value, maximumFractionDigits: 0});
```
JBTAdmin, line 181: Don't know the whole context but for `currency: milestone?.currency` you are using optional chaining on the other hand for `amount: milestone.value` you are not. Which looks odd to me.

SimonBackx (approved):
One tiny comment, but apart from that everything looks good! 👌

PR #16322, opened 2023-02-23
Updated outbound_link_tagging setting to be based on members_track_sources

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2601

Sets the value of outbound_link_tagging to the same value of members_track_sources, so that is disabled by default for privacy sensitive sites.

Also makes sure the `outbound_link_tagging` setting is available in admin (currently excluded because it is in the analytics group)

### DIFF
--- a/ghost/admin/app/services/settings.js
+++ b/ghost/admin/app/services/settings.js
@@ -55,7 +55,7 @@ export default class SettingsService extends Service.extend(ValidationEngine) {
     _loadSettings() {
         if (!this._loadingPromise) {
             this._loadingPromise = this.store
-                .queryRecord('setting', {group: 'site,theme,private,members,portal,newsletter,email,amp,labs,slack,unsplash,views,firstpromoter,editor,comments'})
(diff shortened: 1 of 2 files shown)


ghost/admin/app/services/settings.js
```diff
@@ -55,7 +55,7 @@ export default class SettingsService extends Service.extend(ValidationEngine) {
     _loadSettings() {
         if (!this._loadingPromise) {
             this._loadingPromise = this.store
-                .queryRecord('setting', {group: 'site,theme,private,members,portal,newsletter,email,amp,labs,slack,unsplash,views,firstpromoter,editor,comments'})
+                .queryRecord('setting', {group: 'site,theme,private,members,portal,newsletter,email,amp,labs,slack,unsplash,views,firstpromoter,editor,comments,analytics'})
```
SimonBackx, line 58: Small fix (setting wasn't passed to admin because the analytics group was excluded)

ronaldlangeveld (approved):
LGTM!

daniellockyer (approved):

PR #16330, opened 2023-02-23
combined fetch for metadata and verification

9larsons:
refs TryGhost/Team#2582
-removed duplicative fetch request
-added mention as type in oembed service

Note: I'm not a huge fan of creating a new type for this in the oembed service. It was the cleanest way to utilize what we already had with the metadata request as this already pulled the `body` to scrape the metadata. 

Should we rename the `WebmentionMetadata` class to make it more clear that it's pulling the source and not just metadata?


9larsons:
There's also no unit tests for the `oembed` package, and nothing in the service covers this new type but I felt the latter was ok given that it's just an extension of the `bookmark` type that does have coverage.

I could add some unit tests if we wanted to get some more coverage with this.

SimonBackx (approved):
Nice minimal change. Good question! I don't really have a better name for the WebmentionMetaData class. Let's keep it for now 👌

SimonBackx:
> Note: I'm not a huge fan of creating a new type for this in the oembed service. It was the cleanest way to utilize what we already had with the metadata request as this already pulled the body to scrape the metadata.

I agree. I don't have a much better alternative either. Maybe we can create tech debt issue to rethink this during cleanup?

PR #16358, opened 2023-03-03
Fixed filtering in Tiers API

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/2665

Whilst writing a new e2e test for tiers, I found that passing options and filters to tiers have no effect, despite being used by some existing tests eg [ here](https://github.com/TryGhost/Ghost/blob/901485c47b92a323ed0c4d393d580cbd8ca23a30/ghost/core/test/e2e-api/admin/tiers.test.js#L116).
That created a false conception that the implementation exists.


daniellockyer:
Closing until later

PR #16375, opened 2023-03-08
:bug: Fixed maintaining archived newsletter subscriptions for members

9larsons:
refs TryGhost/Team#2698
-archived newsletters will no longer be unsubscribed when saving member


9larsons:
@allouis Any idea why the snapshot needed changing here? I don't get why my changes would result in a change in the context.

allouis:
@9larsons Looks like you've fixed a bug - we didn't always load the `initialMember` before, but now we do. That piece of code throws a NotFound error with a custom message, I'm guessing elsewhere in the flow a NotFound error is thrown with a generic message and that was what the snapshot used before!

allouis (commented):
Fix looks good! Would be good to add a couple e2e tests just to prove that we cannot subscribe and unsubscribe from archived newsletters ☺️

PR #16376, opened 2023-03-08
✨Excludes sites with `canonical_url` meta from sitemap.

ronaldlangeveld:
ref https://github.com/TryGhost/Team/issues/2531

This commit fixes the issue where non-canonical URLs are included in the XML sitemap, leading to poor SEO for our user's sites. The solution implemented is to exclude any page or post that specifies a canonical URL in its metadata from the sitemap.

To achieve this, a condition has been added to the 'addUrl' method, which checks for the existence of a canonical URL in the metadata of the resource being added to the sitemap. If a canonical URL is present, the resource is excluded from the sitemap.

With this fix, our user's sites will have better SEO and improved search engine visibility.


ghost/core/core/frontend/services/sitemap/base-generator.js
```diff
@@ -22,6 +22,13 @@ class BaseSiteMapGenerator {
         this.maxPerPage = 50000;
     }
 
+    hasCanonicalUrl(datum) {
```
55sketch, line 25: I'm a bit confused on the logic here - could this not be simplified to just `return datum.canonical_url ?? null`?

Indeed, do we even need this function? Could line 78 not read `if (node && !datum.canonical_url)`?

Am I missing something in the data structure that makes this more simplified approach not viable?

ronaldlangeveld: Absolutely, I initially had it that way, then I had the testing coverage bot shouting at me, so made it into a separate method and have that unit tested by itself.

55sketch: Ah ok gotcha, so could the function itself be leaner? the `?? null` seems superflous in this structure?

ronaldlangeveld: Simplified it into a single line. 😊

PR #16377, opened 2023-03-08
Prevented Explore iframe to be loading before needed

aileen:
no issue

- We were loading the Explore iframe together with all assets in Admin
- This change will avoid that and only start loading and rendering Explore, once it's clicked

### DIFF
--- a/ghost/admin/app/components/gh-explore-iframe.hbs
+++ b/ghost/admin/app/components/gh-explore-iframe.hbs
@@ -3,8 +3,6 @@
     class="explore-frame"
     frameborder="0"
     title="Explore"
-    {{did-insert this.setup}}
     {{did-update this.handleDarkModeChange this.feature.nightShift}}
-    {{did-insert this.attachMessageListener}}
     ...attributes
 ></iframe>
--- a/ghost/admin/app/components/gh-explore-iframe.hbs
+++ b/ghost/admin/app/components/gh-explore-iframe.hbs
@@ -3,7 +3,7 @@
     class="explore-frame"
     frameborder="0"
     title="Explore"
-    {{did-update this.handleDarkModeChange this.feature.nightShift}}
     {{did-insert this.setup}}
+    {{did-update this.handleDarkModeChange this.feature.nightShift}}
     ...attributes
 ></iframe>
--- a/ghost/admin/app/components/gh-explore-iframe.hbs
+++ b/ghost/admin/app/components/gh-explore-iframe.hbs
@@ -5,6 +5,6 @@
     title="Explore"
     {{did-insert this.setup}}
     {{did-update this.handleDarkModeChange this.feature.nightShift}}
+    {{did-insert this.attachMessageListener}}
     ...attributes
->
-</iframe>
+></iframe>
--- a/ghost/admin/app/components/gh-explore-iframe.hbs
+++ b/ghost/admin/app/components/gh-explore-iframe.hbs
@@ -4,5 +4,6 @@
     frameborder="0"
     title="Explore"
     {{did-update this.handleDarkModeChange this.feature.nightShift}}
+    {{did-insert this.setup}}
     ...attributes
 ></iframe>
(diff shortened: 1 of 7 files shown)


ghost/admin/app/components/gh-explore-iframe.hbs
```diff
@@ -1 +1,9 @@
-<iframe id="explore-frame" class="explore-frame" frameborder="0" title="Explore" ...attributes {{did-insert this.setup}}></iframe>
\ No newline at end of file
+<iframe
+    id="explore-frame"
+    class="explore-frame"
+    frameborder="0"
+    title="Explore"
+    {{did-insert this.setup}}
+    {{did-insert this.attachMessageListener}}
```
kevinansfield, line 7: [`{{did-insert}}` is an anti-pattern](https://github.com/emberjs/ember-render-modifiers#when-to-use-these-modifiers-and-when-not-to-use-them) and we shouldn't introduce any more of them. In this case its use causes potential problems because you're attaching an event listener to `window` but there's no cleanup so if the iframe gets rendered and destroyed multiple times you'll have multiple event handlers responding to the iframe message event.

Two typical routes to go when you're reaching for `{{did-insert}}`:
1. if you're using the `{{did-insert}}` modifier to get access to the element then consider adding a custom modifier, especially if the behaviour is fairly generic (e.g. `{{on-window-event "message" this.handleIframeMessage}}`)
2. if you're not using the element and you're working in a component then you can use the component's lifecycle with its `constructor()` and `willDestroy()` hooks (probably the simpler option in this case)

ghost/admin/app/components/gh-nav-menu/main.js
```diff
@@ -111,6 +112,11 @@ export default class Main extends Component.extend(ShortcutsMixin) {
         this.billing.openBillingWindow(this.router.currentURL);
     }
 
+    @action
+    toggleExploreWindow() {
+        this.explore.openExploreWindow(this.router.currentURL);
```
kevinansfield, line 117: Not directly related to this PR but I think we can drop the arguments to `openExploreWindow()` and [clean up some of the logic](https://github.com/TryGhost/Ghost/blob/main/ghost/admin/app/services/explore.js#L93-L111). The first argument passed in is used to store the `previousUrl` notionally for when the modal gets closed but this isn't a modal and doesn't have any "close modal" functionality so it's never used, same with `childRoute` as that argument is never passed in from any method calls.

aileen: I have removed the route handling logic. We should probably really tackle that same thing for billing as well, as it also isn't using a full screen modal

ghost/admin/app/services/explore.js
```diff
@@ -90,26 +88,29 @@ export default class ExploreService extends Service {
...
     }
 
+    ensureIframeIsLoaded() {
+        if (this.getExploreIframe() && !this.getExploreIframe()?.src) {
+            this.getExploreIframe().src = this.getIframeURL();
+        }
+    }
+
     getExploreIframe() {
         return document.getElementById('explore-frame');
```
aileen, line 115: @kevinansfield this works all ok (didn't implement any loading state yet), except for when Ember re-renders when in development mode. It appears the iframe is not available at the time of the re-render then and I'm unsure how to solve this without using the `did-insert` hook? Is there an example somewhere where we faced a similar issue? I couldn't find an equivalent in the glimmer docs to re-create the same livecyle hook 🤔

kevinansfield: I think it’s worth keeping the original `did-insert` behaviour to avoid too much change, my previous comment was more about adding another `did-insert` and the cleanup problems associated with that approach.

To get rid of all the `did-insert` usage I think would require a bit more of an architecture rethink. One thing that might be worth trying is switching `getIframeUrl()` to a `get iframeUrl()` and setting `src={{this.explore.iframeUrl}}` on the iframe itself so it's immediately loading with the right URL and doesn't need to be reset externally once rendered.

aileen: Ah cool. I revived the `did-insert` hook again and converted `getIframeUrl` to a getter. Everything works fine and as expected to me now. Now I only need to implement a loading screen in Explore itself

daniellockyer:
@aileen What is the status of this PR? Can we merge it, or does it need more work?

aileen:
Was waiting for a review here from @kevinansfield 😄

PR #16385, opened 2023-03-09
:bug: Fixed upper price limit in Membership Tiers

sagzy:
closes https://github.com/TryGhost/Team/issues/2287
- Stripe allows a max. amount of 999,999.99, cf. Stripe docs https://stripe.com/docs/api/payment_intents/object#payment_intent_object-amount
- The publisher now sees an error if they choose a price that is higher than the upper limit


rshbhgrg (approved):
LGTM 🎉

PR #16386, opened 2023-03-09
Fixed paid subscription graph

SimonBackx:
Description for each fix in each commit


allouis (approved):
I appreciate these changes are probably fairly tricky to test - do you have any idea _how_ tricky? Would be great to add some tests if possible!

Otherwise looks gr8!!

SimonBackx:
I could add some tests for the expected paid subscription event rows. It would only test if the SQL query work, not really the logic. Don't think there is an easy way to test the logic because then we would need to simulate lots of different Stripe subscriptions

SimonBackx:
I'm working on tests, but I'll commit them separately from this PR

PR #16390, opened 2023-03-10
Mentions email report

allouis:
This will attempt to send a mentions report email on boot and also every hour (at a random minute and second), but not doing so if either:

1. There are no new mentions since the last report
2. The last report was sent within 24 hours

All of this is behind the mentions emails labs flag, which means it can be released/tested independently of the rest of the mentions feature.

The scheduler doesn't support local timezones at the moment, so sending every day at around 4pm was difficult, and would end up error prone unless we add proper support for timezones - I've done the best I can to make sure that we send something every day if possible!

### DIFF
--- a/ghost/core/core/server/services/mentions-email-report/service.js
+++ b/ghost/core/core/server/services/mentions-email-report/service.js
@@ -1,0 +1,142 @@
+const MentionEmailReportJob = require('@tryghost/mentions-email-report');
+
+/**
+ * @typedef {import('@tryghost/mentions-email-report/lib/mentions-email-report').MentionReport} MentionReport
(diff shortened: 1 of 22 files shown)


allouis:
The first commit is from https://github.com/TryGhost/Ghost/pull/16399 and can be rebased out once that PR is merged to `main`

ghost/core/core/server/services/mentions-email-report/service.js
```diff
@@ -0,0 +1,142 @@
...
+                }
+
+                return new Date(parsedDate);
+            },
+            async setLatestReportDate(date) {
+                await models.Settings.edit({
+                    key: 'lastMentionsReportEmailTimestamp',
+                    value: date
+                });
+            }
```
SimonBackx, line 91: refs

SimonBackx (approved):

PR #16391, opened 2023-03-10
Removed Bluebird `mapSeries` usage

daniellockyer:
refs https://github.com/TryGhost/Ghost/issues/14882

- we're working on removing Bluebird and there are a few places where we can switch it to native for-loops (if we don't need the result), or the `sequence` function from `@tryghost/promise`


daniellockyer:
Closing as this isn't exactly working right now

PR #16398, opened 2023-03-13
[WIP] Redis Cache Adapter Upgrades

JoeeGrigg:
# Additions

# #Added Non Redis Cluster Support

If the constructor of the redis client isn't a cluster then it will now fall back to returning the full client when requesting the primary node.

##  Added a reset method

The reset method first gets all keys stored for the cache and then one by one deletes them from redis. I tried to get it working with a single delete command but had issues. I am not sure if that was due to the length and complexity of the keys or something else but this way seems safer.

# Known Issues

I see one main issue with the cache adapter right now and that is that the `keyPrefix` is not required. I agree that it should be an optional field in config however it causes problems when it comes to getting all keys (this includes when deleting). The problem is that without a key prefix right now we generate the `_keysPattern` as empty. This means that it will return no keys. However, if we switched it to `*` then it would return all keys on that redis instance, then causing errors when resetting if there are any keys that they don't have permission to delete. Also, every time a reset occurred it would reset all cache and not just that specific cache.

I think the solution is to make the constructor of the cache pass in a name for that cache; for example `tags-public`, `posts-public` or `image-sizes` and not have that as something that has to be passed in as config. Then, in config, we can pass the key prefix and it would all get combined into the prefix. For example `[Config Defined Key Prefix][Cache Constructor Name]:[Key]`. So, in key prefix we would define `site:[site id]:` and the constructor would be given the name `posts-public` so that would combine into `site:[site id]:posts-public`.

### DIFF
--- a/ghost/adapter-cache-redis/lib/AdapterCacheRedis.js
+++ b/ghost/adapter-cache-redis/lib/AdapterCacheRedis.js
@@ -83,6 +83,14 @@ class AdapterCacheRedis extends BaseCacheAdapter {
         });
     }
 
+    async #getKeys() {
+        const primaryNode = this.#getPrimaryRedisNode();
+        if (primaryNode === null) {
+            return [];
+        }
+        return await this.#scanNodeForKeys(primaryNode);
+    }
+
     /**
      * This is a recommended way to build cache key prefixes from
      * the cache-manager package. Might be a good contribution to make
@@ -144,12 +152,7 @@ class AdapterCacheRedis extends BaseCacheAdapter {
      */
     async keys() {
         try {
-            const primaryNode = this.#getPrimaryRedisNode();
-            if (primaryNode === null) {
-                return [];
-            }
-            const rawKeys = await this.#scanNodeForKeys(primaryNode);
-            return rawKeys.map((key) => {
+            return (await this.#getKeys()).map((key) => {
                 return this._removeKeyPrefix(key);
             });
         } catch (err) {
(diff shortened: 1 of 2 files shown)


naz:
@JoeeGrigg  I've done some local testing and here are the stats for caches with >50k records (which we can easily expect for large sites).

Run #1:
```
[ghost] Calling reset for 61234 keys with prefix site_id_here:posts-content-api:
[ghost] Call to reset took 1732.7628750801086 milliseconds
[ghost] Average per call 0.02829739809713735 milliseconds
```

Run #2:
```
[ghost] Calling reset for 69777 keys with prefix site_id_here:posts-content-api:
[ghost] Call to reset took 1767.651249885559 milliseconds
[ghost] Average per call 0.025332863979327845 milliseconds
```

Round #3
```
[ghost] Calling reset for 64500 keys with prefix site_id_here:posts-content-api: 
[ghost] Call to reset took 1381.9722909927368 milliseconds
[ghost] Average per call 0.021425926992135455 milliseconds
```

code I've used to get above stats:
```js
    async reset() {
        const keys = await this.#getKeys();
        console.log(`Calling reset for ${keys.length} keys with prefix ${this._buildKey('')}`);
        const startTime = performance.now();

        for (const key of keys) {
            await this.cache.del(key);
        }

        const endTime = performance.now();
        console.log(`Call to reset took ${endTime - startTime} milliseconds`);
        console.log(`Average per call ${(endTime - startTime) / keys.length} milliseconds`);
    }
```

and to populate the cache with large amount of data just added additional entries on each `.set` call in the cache:
```js
for (let i = 0; i < 10000; i++) {
    await this.cache.set(this._buildKey(`${key}-${i}`), value);
}
```

The `~1.6s` per reset of ~60k records isn't too bad. It seems, so the next step would be testing this on staging environment. Let's see if the timing is similar in cluster settings.

naz:
Heads up, I've added an extra commit with error handling and rebased the branch over the latest main.

PR #16399, opened 2023-03-13
Added setting to store last mention report email date

allouis:
Because there is no guarantee about a daily job running exactly once a day, we need to store the last time that the email was sent, so that we can refrain from sending one if it's been less than a day since the last.

A setting has been used for this as we don't currently have a pattern for it, we might want to consider moving this to some kind of cache based solution in future. This has been added as a core setting so that we don't expose it via the API.

### DIFF
--- a/ghost/core/core/server/data/schema/default-settings/default-settings.json
+++ b/ghost/core/core/server/data/schema/default-settings/default-settings.json
@@ -1,5 +1,9 @@
 {
     "core": {
+        "last_mentions_report_email_timestamp": {
+            "defaultValue": null,
+            "type": "datetime"
+        },
         "db_hash": {
             "defaultValue": null,
             "type": "string"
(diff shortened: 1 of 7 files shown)


ghost/core/core/server/data/schema/default-settings/default-settings.json
```diff
@@ -1,5 +1,9 @@
 {
     "core": {
+        "last_mentions_report_email_timestamp": {
+            "defaultValue": null,
+            "type": "datetime"
```
SimonBackx, line 5: I don't see any other usage of type datetime, is this supported?

SimonBackx:
Is this missing a `addSetting` migration?

ghost/core/core/server/data/schema/default-settings/default-settings.json
allouis: Good shout  - we manually parse it when we use it, I guess should change to "string"?

allouis:
@daniellockyer What's the status of the `addSetting` migrations now? Do we still need them?

rishabhgrg:
@SimonBackx I have updated this PR with setting migration, update of `dateTime` to `string` and fixing tests for new core setting.

SimonBackx (approved):

PR #16405, opened 2023-03-13
Added handling for initial and skipped Milestones

aileen:

- When milestones will be activated we would send out emails to users that are way above the achieved milestone, as we didn't record milestones before
- The plan is to implement a 0 milestone and don't send an email for achieving those and also add all achieved milestones in the first run until a first milestone is stored in the DB, then increment from there.
- This change takes care of two cases:
    1. Milestones gets enabled and runs initially. We don't want to send emails unless there's already at least one milestone achieved. For that we add a 0 milestone helper and add a `initial` reason to the meta object for the milestone event, so we can choose not to ping Slack and also disable email sending for all milestones achieved in this initial run.
    2. All achieved milestones will be stored in the DB, even when that means we skip some. This introduces the `skipped` reason which also doesn't send emails for the skipped milestones, but will do for correctly achieved milestones (always the highest one).
- Added handling for slack notifications to not attempt sending when reason is `skipped` or `initial`


SimonBackx (approved):
Makes sense 👌

PR #16431, opened 2023-03-17
Modified outbound link tagger to use ?ref=domain instead of ?ref=siteTitle

cmraible:
refs TryGhost/Team#2755

- Outbound link tagger will now append ?ref=example.com or ?ref=example.ghost.io instead of ?ref=siteTitle


rshbhgrg (commented):
LGTM 🎉

PR #16446, opened 2023-03-20
Fixed Stripe connection modal button visibility

minimaluminium:
refs https://github.com/TryGhost/Team/issues/2483

- the button is toggled depending on `stripeConnectIntegrationToken` which is a calculated value
- however, this calculated value isn't reset when Stripe is disconnected right away without closing the modal
- the reset action was already available and it's now passed to `StripeSettingsForm`, so that it can be called when Stripe is disconnected

### DIFF
--- a/ghost/admin/app/components/settings/members/stripe-settings-form.js
+++ b/ghost/admin/app/components/settings/members/stripe-settings-form.js
@@ -259,6 +259,7 @@ export default class StripeSettingsForm extends Component {
         yield this.ajax.delete(url);
         yield this.settings.reload();
 
+        this.args.reset();
         this.args.onDisconnected?.();
     }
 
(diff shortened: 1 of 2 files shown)


rshbhgrg (approved):

ghost/admin/app/components/settings/members/stripe-settings-form.js
```diff
@@ -259,6 +259,7 @@ export default class StripeSettingsForm extends Component {
         yield this.ajax.delete(url);
         yield this.settings.reload();
 
+        this.args.reset();
```
rishabhgrg, line 262: @minimaluminium Might be a good idea to use the same syntax as below that checks for availability of `reset` in args by doing `this.args.reset?.()`

PR #16447, opened 2023-03-20
Added French (fr) translations

sagzy:
no issue
- Used informal form ("tu")


ghost/i18n/lib/i18n.js
```diff
@@ -1,6 +1,6 @@
 const i18next = require('i18next');
 
-const SUPPORTED_LOCALES = ['en', 'nl'];
+const SUPPORTED_LOCALES = ['en', 'nl', 'fr'];
```
naz, line 3: let's make it a multi-line array declaration straight away. I imagine every single translation will be modifying this one line causing lots of conflicts.

daniellockyer: ➕ I would agree, nice idea

daniellockyer: + alphabetically sorted? 😬

sagzy: yes, good idea! Updated :)

PR #16458, opened 2023-03-21
Add Migrate app

PaulAdamDavis:
Adds the UI for self-serve migrations

### DIFF
--- a/ghost/admin/app/components/gh-migrate-iframe.js
+++ b/ghost/admin/app/components/gh-migrate-iframe.js
@@ -1,0 +1,77 @@
+import Component from '@glimmer/component';
+import {action} from '@ember/object';
+// import {htmlSafe} from '@ember/template';
+import {inject as service} from '@ember/service';
+
+export default class GhMigrateIframe extends Component {
+    @service migrate;
+    @service router;
+    @service feature;
+    @service notifications;
+
+    willDestroy() {
+        super.willDestroy(...arguments);
+        window.removeEventListener('message', this.handleIframeMessage);
+    }
+
+    @action
+    setup() {
+        this.migrate.getMigrateIframe().src = this.migrate.getIframeURL();
+        window.addEventListener('message', this.handleIframeMessage);
+    }
+
+    @action
+    async handleIframeMessage(event) {
+        if (this.isDestroyed || this.isDestroying) {
+            return;
+        }
+
+        // only process messages coming from the migrate iframe
+        if (event?.data && this.migrate.getIframeURL().includes(event?.origin)) {
+            if (event.data?.request === 'apiUrl') {
+                await this._handleUrlRequest();
+            }
+
+            if (event.data?.route) {
+                this._handleRouteUpdate(event.data);
+            }
+
+            if (event.data?.siteData) {
+                this._handleSiteDataUpdate(event.data);
+            }
+        }
+    }
+
+    // The iframe can send route updates to navigate to within Admin, as some routes
+    // have to be rendered within the iframe and others require to break out of it.
+    _handleRouteUpdate(data) {
+        const route = data.route;
+        this.migrate.isIframeTransition = route?.includes('/migrate');
+        this.migrate.toggleMigrateWindow(this.migrate.isIframeTransition);
+        this.router.transitionTo(route);
+    }
+
+    async _handleUrlRequest() {
+        let theToken = await this.migrate.apiToken();
+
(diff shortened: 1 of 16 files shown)


ghost/admin/app/components/gh-migrate-iframe.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @action
+    async handleIframeMessage(event) {
+        if (this.isDestroyed || this.isDestroying) {
+            return;
+        }
+
+        // only process messages coming from the migrate iframe
+        if (event?.data && this.migrate.getIframeURL().includes(event?.origin)) {
```
allouis, line 30: This should be a specific check comparing origins, probably something like:
```js
const url = new URL(this.migrate.getIframeURL());
if (event.origin === url.origin) {
  ...
}
```

ghost/admin/app/components/gh-migrate-iframe.js
```diff
@@ -0,0 +1,77 @@
...
+        this.migrate.getMigrateIframe().contentWindow.postMessage({
+            request: 'initialData',
+            response: {
+                apiUrl: this.migrate.apiUrl,
+                apiToken: theToken,
+                cookie: 'ghost-admin-api-session=s:12345678/po; Path=/ghost; Expires=Wed, 06 Sep 2023 05:18:55 GMT; HttpOnly; SameSite=Lax',
+                darkMode: this.feature.nightShift,
+                stripe: this.migrate.isStripeConnected
+            }
+        }, '*');
```
allouis, line 66: The usage of `'*'` is a security issue as it will post to all iframes, instead it should be limited to the origin that you expect, which is likely `(new URL(this.migrate.getIframeUrl()).origin`

ghost/admin/app/components/gh-migrate-iframe.js
```diff
@@ -0,0 +1,77 @@
...
+
+    async _handleUrlRequest() {
+        let theToken = await this.migrate.apiToken();
+
+        this.migrate.getMigrateIframe().contentWindow.postMessage({
+            request: 'initialData',
+            response: {
+                apiUrl: this.migrate.apiUrl,
+                apiToken: theToken,
+                cookie: 'ghost-admin-api-session=s:12345678/po; Path=/ghost; Expires=Wed, 06 Sep 2023 05:18:55 GMT; HttpOnly; SameSite=Lax',
```
allouis, line 62: What's the purpose of this out of interest?

ghost/admin/app/components/gh-migrate-iframe.js
```diff
@@ -0,0 +1,77 @@
+import Component from '@glimmer/component';
+import {action} from '@ember/object';
+// import {htmlSafe} from '@ember/template';
```
naz, line 3: This should be removed? 🙈 
```suggestion
```

ghost/admin/app/components/gh-migrate-iframe.js
allouis, line 30: Is it expected that a message can perform multiple actions? If not I would suggest returning after each one, or making them exclusive somehow to avoid any issues from running multiple!

allouis (changes requested):
Might want to consider moving the api token generation to the server to simplify things, but not a major issue!

Most important things are the origin validation for receiving and sending messages in both directions, so in the code running inside the iFrame too!

ghost/admin/app/services/migrate.js
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    async apiToken() {
+        // TODO: Getting the token can be improved
+        const ghostIntegrationsUrl = this.ghostPaths.url.api('integrations') + '?include=api_keys';
+        return this.ajax.request(ghostIntegrationsUrl).then(async (response) => {
+            const ssmIntegration = response.integrations.find(r => r.slug === 'self-serve-migration');
+
+            const key = ssmIntegration.api_keys[0].secret;
+            const [id, secret] = key.split(':');
```
naz, line 35: we are expecting that this version will be released along with the [key formatting fix](https://github.com/TryGhost/Ghost/pull/16508), so there will be nothing to update in the future. sweeet 👌

ghost/admin/app/components/gh-migrate-iframe.js
PaulAdamDavis: It's a holdover from me testing before we got the proper auth working. Will remove!

PR #16481, opened 2023-03-24
🐛 Fixed UI bug when choosing email recipients

cmraible:
- closes https://github.com/TryGhost/Team/issues/2786

- when trying to deselect all recipients, the UI would pass null as the filter to publish-options
- publish-options would then fallback to the default options, which is all recipients, and then the UI wouldn't update which made things even weirder
- we want to fallback to the default recipients when the recipientFilter is undefined (e.g. hasn't been set at all), but not when it is explicitly set to null (e.g. when the user has deselected all recipients)

### DIFF
--- a/ghost/admin/app/utils/publish-options.js
+++ b/ghost/admin/app/utils/publish-options.js
@@ -172,7 +172,11 @@ export default class PublishOptions {
     }
 
     get recipientFilter() {
-        return this.selectedRecipientFilter || (this.post.newsletter && this.post.emailSegment) || this.defaultRecipientFilter;
+        if (this.selectedRecipientFilter === undefined) {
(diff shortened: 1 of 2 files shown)


cmraible:
History for additional context —

- Issue that prompted the initial change: https://github.com/TryGhost/Ghost/issues/16125
- PR that fixed the issue but created this bug inadvertently on Jan 29: https://github.com/TryGhost/Ghost/pull/16197/files
- Another, less relevant PR that came after: https://github.com/TryGhost/Ghost/pull/16247

This PR returns to the previous method of explicitly checking for undefined vs implicitly checking falsy-ness.

rshbhgrg (approved):
Great catch, lgtm! 🎉

PR #16490, opened 2023-03-24
🐛 Fixed newsletter subscription for members not saving

rshbhgrg:
fixes https://github.com/TryGhost/Team/issues/2783 
refs https://github.com/TryGhost/Ghost/commit/cb05fae5a3d4a5a6fc716daeef831eb821daa35a

The root cause of the issue was the fact we no longer checked for lack of `newsletters` property on member data before checking its `subscribed` property which is now deprecated. This caused a cascading effect where `subscribed:false` property on a member overrides the value for `newsletters` data. The check was accidentally removed in a previous bug fix.

So for members that were not subscribed to any newsletters, saving a newsletter subscription failed as they had their `subscribed` set to `false`, and it was resetting the newsletter subscription to empty always.

### DIFF
--- a/ghost/core/test/e2e-api/admin/__snapshots__/members.test.js.snap
+++ b/ghost/core/test/e2e-api/admin/__snapshots__/members.test.js.snap
@@ -356,6 +356,149 @@ Object {
 }
 `;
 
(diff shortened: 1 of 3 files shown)


9larsons (approved):
Looks good 👍

PR #16506, opened 2023-03-27
Added newsletter color columns

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2825


allouis (changes requested):
I think some casper changes snuck in here - but otherwise 👍

PR #16461, opened 2023-03-21
Added migrations for Self-Serve Migration Integration and API key

naz:
refs https://github.com/TryGhost/Team/issues/2790
refs https://github.com/TryGhost/Ghost/commit/3747df1bc871752d91bc49f0f0f588f6857c4a2e#diff-396038cecd7a381616d00954ae18a655ae2a8af71ea65866bf09d2c7cc1b5235

- This integration will be used to perform self-serve migrations.
- The integration will be limited to these Admin API endpoints: 
```
POST /ghost/api/admin/db
POST /ghost/api/admin/db/media/inline
POST /ghost/api/admin/members/upload
GET  /ghost/api/admin/tags/:id
GET  /ghost/api/admin/tags/slug/:slug
```

PR #16508, opened 2023-03-27
Fixed self-serve api key type

naz:
refs https://github.com/TryGhost/Ghost/pull/16461

- The referenced migration had an incorrect 'type' assigned to the self-serve integration's api key. Should have been "admin" instead of "core"

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.41/2023-03-27-17-51-fix-self-serve-integration-api-key-type.js
+++ b/ghost/core/core/server/data/migrations/versions/5.41/2023-03-27-17-51-fix-self-serve-integration-api-key-type.js
@@ -1,0 +1,36 @@
+const {InternalServerError} = require('@tryghost/errors');
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = module.exports = createTransactionalMigration(
+    async function up(knex) {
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/data/migrations/versions/5.41/2023-03-27-17-51-fix-self-serve-integration-api-key-type.js
```diff
@@ -0,0 +1,36 @@
+const {InternalServerError} = require('@tryghost/errors');
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Adding Admin API key for "Self-Serve Migration Integration"');
```
kevinansfield, line 7: Does this log need updating to avoid any potential confusion?

kevinansfield (approved):

ghost/core/core/server/data/migrations/versions/5.41/2023-03-27-17-51-fix-self-serve-integration-api-key-type.js
naz: Good catch, should be "Updating" 👍

PR #16509, opened 2023-03-27
Fixed broken links with outbound link tagging

cmraible:
refs TryGhost/Team#2840

- moves the `entities.decode()` step to the `LinkReplacer` class so that it's applied to all links, not just the ones that are replaced in the email service

### DIFF
--- a/ghost/email-service/lib/email-renderer.js
+++ b/ghost/email-service/lib/email-renderer.js
@@ -8,7 +8,6 @@ const {textColorForBackgroundColor, darkenToContrastThreshold} = require('@trygh
 const {DateTime} = require('luxon');
 const htmlToPlaintext = require('@tryghost/html-to-plaintext');
 const tpl = require('@tryghost/tpl');
-const entities = require('entities');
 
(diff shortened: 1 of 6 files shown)


ghost/email-service/lib/email-renderer.js
```diff
@@ -294,7 +293,7 @@ class EmailRenderer {
         if (options.clickTrackingEnabled) {
             html = await this.#linkReplacer.replace(html, async (url) => {
                 // Decode any escaped entities in the url
-                url = new URL(entities.decode(url.toString()));
+                url = new URL(url.toString());
```
rishabhgrg, line 296: This line i guess is redundant now, as even if we remove this `url` is already of type `URL` 😄

rishabhgrg:
@cmraible I am closing this in favor of https://github.com/TryGhost/Ghost/pull/16514 - to avoid any issues on merging this because it accidentally got the Casper submodule changes as well. I have cherry-picked the commit here to the new PR directly so it shouldn't cause any issues.

PR #16534, opened 2023-03-31
Added Czech translations

tombarys:
It should be complete translation. I am good at Czech but sure nothing is perfect :). I will check the translation later in the contexts where it occurs and update it eventually too.


tombarys:
Hi, I found some texts in email being not translated but forgot how to suggest/send translations to you. Could you navigate me, please? Thanks!

PR #16537, opened 2023-04-01
Removed Bluebird from core/server/services

swoutch:
refs #14882

- bluebird is not needed anymore with promises in Node.js and async/await

------------------------------------

Hello, this is my first contribution to the project, so I apologize in advance if I did something wrong. If this is the case, don't hesitate to tell me :)


daniellockyer:
Closing as these changes are all in `main` 🙂

PR #16545, opened 2023-04-04
Added Portal signup terms settings and migration

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2885 
fixes https://github.com/TryGhost/Team/issues/2896 
fixes https://github.com/TryGhost/Team/issues/2877

This change adds the new portal_signup_terms_html setting and portal_signup_checkbox_required setting and the corresponding migration.

### DIFF
--- a/ghost/admin/mirage/fixtures/settings.js
+++ b/ghost/admin/mirage/fixtures/settings.js
@@ -52,7 +52,7 @@ export default [
     setting('theme', 'active_theme', 'Casper'),
 
     // PRIVATE
-    setting('private', 'is_private', 'false'),
+    setting('private', 'is_private', false),
     setting('private', 'password', ''),
     setting('private', 'public_hash', ''),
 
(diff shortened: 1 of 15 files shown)


ghost/admin/mirage/fixtures/settings.js
```diff
@@ -81,6 +81,8 @@ export default [
     setting('portal', 'portal_button_style', 'icon-and-text'),
     setting('portal', 'portal_button_icon', null),
     setting('portal', 'portal_button_signup_text', 'Subscribe'),
+    setting('portal', 'portal_signup_terms_html', null),
+    setting('portal', 'portal_signup_checkbox_required', 'false'),
```
allouis, line 85: I see we have other booleans as strings - but just wan to double check that's correct and they get parsed?

allouis (approved):

ghost/admin/mirage/fixtures/settings.js
SimonBackx: Yeah, doesn't make much sense. I'll try to update them all to booleans and see if the tests still work. Because that would make more sense.

SimonBackx: ✅

PR #16546, opened 2023-04-04
Fixed a few typos in the French translation

thibaultmalfoy:
I've fixed a few typos and added true (curly) apostrophes instead of straight single quotes for a better typography.

I have a doubt about translating "you" as "tu" (instead of "vous"). "Tu" can be too informal in French in some contexts. I would love to hear other opinions on this point. Thanks!


sagzy:
@thibaultmalfoy thanks for the fixes! 💯 Ok to merge @daniellockyer 

Regarding "tu" vs "vous", that's a great point. Also have a doubt. I think "vous" is more common and more professional, whereas "tu" brings more proximity with the reader. I checked out a few websites in the publishing space – most of them use the "vous" form, I could only find one using the "tu" form ([Snowball](https://www.snowball.xyz/)). So, tend to agree with using the "vous" form instead. @thibaultmalfoy would you like to give it a go in a separate PR and then we can compare the two versions?

daniellockyer:
Thanks both! Happy to receive further PRs 🙂

PR #16575, opened 2023-04-06
Added Italian locale for Portal

marcoborghesi:
I'm Italian and i've completed the translation to my language. It should be good to go!
A note. I've used* for gender fluid word where needed :)


daniellockyer:
Thanks @marcoborghesi! 🙂

PR #16588, opened 2023-04-07
Fixed layout issue caused by the new Public Preview card in Lexical

sagzy:
refs https://github.com/TryGhost/Team/issues/2914

- with mobiledoc, the Public Preview card would render a `<!--members-only-->` html comment. This comment is then transformed into either a CTA to upgrade (= the user doesn't have full access to the post), or left untouched (= the user has full access to the post)
- now with Lexical, a HTML element wrapper is necessary around the html comment: `<span class="gh-post-upgrade-cta-visibility"><!--members-only--></span>`. This change hides the extra <span> wrapper when necessary, so that the final layout is not affected


sagzy:
Update: not a good idea. Closing PR

PR #16594, opened 2023-04-08
Added Slovenian locale

martinverbic:
This should be complete for Slovenian. Will update once I see it in action!


daniellockyer:
Thanks @martinverbic! 🕺🏻

PR #16608, opened 2023-04-11
Added JSON test reports as artifacts in CI

cmraible:
refs TryGhost/Team#2833

- with this commit, unit and database tests will generate JSON test report files which will be uploaded as artifacts in CI
- the hope is we can pull the artifacts from github to do some data analysis on which tests are flaking most often and/or retrying the most
- adds mocha-multi-reporters and @mochajs/json-file-reporter to allow logging to a JSON file as well as the standard 'spec' stdout output simultaneously


daniellockyer:
@cmraible Gonna close this for now until we reassess flaky tests, but nice idea!

PR #16616, opened 2023-04-12
Added Chinese local translation (zh) for protal.json&ghost.json

AreChen:
A complete Chinese localization translation has been done!
进行了完整的中文化翻译！


AreChen:
It seems that the translation field for the DASHBOARD page has not been seen. Is it possible to perform a complete localization translation? Also, I would like to ask if the theme's translation is now included in these two files being translated currently?

PR #16621, opened 2023-04-12
Updated Slovenian locale

martinverbic:
Some changes based on initial use.


daniellockyer:
Thanks again @martinverbic 🙂

PR #16623, opened 2023-04-13
Added Chinese local translation (zh) for portal.json&ghost.json

AreChen:
Added Chinese local translation (zh) for protal.json&ghost.json


daniellockyer:
Thanks @AreChen! 🕺🏻

PR #16624, opened 2023-04-13
Updated GhContextMenu to handle freezing the SelectionList

allouis:
### <samp>🤖 Generated by Copilot at 1d69b45</samp>

Refactored the context menu component and its usage in the posts list to improve the user experience and code quality. Introduced a `state` property and a `setState` method to the `gh-context-menu` component to handle different scenarios. Used the `gh-context-menu` component in the `posts-list/context-menu` component to simplify the modal and loading logic. Added a `#frozen` property and methods to the `selection-list` utility to prevent the selection of posts from changing while the context menu is active.

### DIFF
--- a/ghost/admin/app/components/gh-context-menu.js
+++ b/ghost/admin/app/components/gh-context-menu.js
@@ -2,15 +2,67 @@ import Component from '@glimmer/component';
 import SelectionList from '../utils/selection-list';
 import {action} from '@ember/object';
 import {inject as service} from '@ember/service';
+import {task} from 'ember-concurrency';
 import {tracked} from '@glimmer/tracking';
+
+const states = [
+    'default', // default state
+    'open', // menu open
+    'modal', // modal open
+    'loading' // performing an action
+];
 
 export default class GhContextMenu extends Component {
     @service dropdown;
+    @service modals;
 
     @tracked isOpen = false;
     @tracked left = 0;
     @tracked top = 0;
     @tracked selectionList = new SelectionList();
+
+    state = states[0];
+
+    #originalConfirm = null;
+    #modal = null;
+
+    setState(state) {
+        switch (state) {
+        case this.state:
+            return;
+        case 'default':
+            this.isOpen = false;
+            this.selectionList.unfreeze();
+            this.#closeModal();
+            this.state = state;
+            return;
+        case 'open':
+            if (this.state !== 'default') {
+                return;
+            }
+            this.isOpen = true;
+            this.selectionList.freeze();
+            this.#closeModal();
+            this.state = state;
+            return;
+        case 'modal':
+            if (this.state !== 'open') {
+                return;
+            }
+            this.isOpen = false;
+            this.selectionList.freeze();
+            this.state = state;
+            return;
+        case 'loading':
+            if (this.state !== 'open' && this.state !== 'modal') {
+                return;
+            }
+            this.isOpen = false;
+            this.selectionList.freeze();
+            this.state = state;
+            return;
+        }
+    }
 
     get name() {
         return this.args.name;
@@ -36,12 +88,14 @@ export default class GhContextMenu extends Component {
 
     @action
     open() {
-        this.isOpen = true;
+        this.setState('open');
     }
 
     @action
     close() {
-        this.isOpen = false;
+        if (this.state === 'open') {
+            this.setState('default');
+        }
     }
 
     @action
@@ -65,7 +119,7 @@ export default class GhContextMenu extends Component {
             }
 
             this.open();
-        } else if (this.isOpen) {
+        } else {
             this.close();
         }
     }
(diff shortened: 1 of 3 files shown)


ghost/admin/app/components/gh-context-menu.js
```diff
@@ -3,15 +3,67 @@ import SelectionList from '../utils/selection-list';
 import {action} from '@ember/object';
 import {inject as service} from '@ember/service';
 import {tracked} from '@glimmer/tracking';
+import {task} from 'ember-concurrency';
+
+const states = [
```
allouis, line 8: We don't really use this list, but I thought it's useful to know what the states are 🦐

ghost/admin/app/components/gh-context-menu.js
```diff
@@ -36,12 +88,14 @@ export default class GhContextMenu extends Component {
...
     @action
     open() {
-        this.isOpen = true;
+        this.setState('open');
     }
 
     @action
     close() {
-        this.isOpen = false;
+        if (this.state === 'open') {
```
allouis, line 96: We add this check because you can only "close" the menu when it is open. However you can transition to the `default` state from any other state.

ghost/admin/app/components/gh-context-menu.js
```diff
@@ -74,4 +128,36 @@ export default class GhContextMenu extends Component {
     stopClicks(event) {
         event.stopPropagation();
     }
+
+    @task
+    *confirmWrapperTask(...args) {
```
allouis, line 133: This had to be a method and use instance variables rather than an inline task and context variables because for some reason Ember was not happy about inline tasks.

ghost/admin/app/components/posts-list/context-menu.js
```diff
@@ -137,6 +139,30 @@ export default class PostsContextMenu extends Component {
...
+
+    @task
+    *featurePostsTask() {
+        const updatedModels = this.selectionList.availableModels;
+        yield this.performBulkEdit('feature');
+
+        // Update the models on the client side
+        for (const post of updatedModels) {
+            post.set('featured', true);
+        }
```
SimonBackx, line 152: This was changed on main and needs to use the new form:
```suggestion
// We need to do it this way to prevent marking the model as dirty
        this.store.push({
                data: {
                    id: post.id,
                    type: 'post',
                    attributes: {
                        featured: true
                    }
                }
            });
```

Otherwise the post is marked as 'dirty' and when opening the editor, it will think there are new changes

ghost/admin/app/components/posts-list/context-menu.js
```diff
@@ -137,6 +139,30 @@ export default class PostsContextMenu extends Component {
...
+
+    @task
+    *unfeaturePostsTask() {
+        const updatedModels = this.selectionList.availableModels;
+        yield this.performBulkEdit('unfeature');
+
+        // Update the models on the client side
+        for (const post of updatedModels) {
+            post.set('featured', false);
+        }
```
SimonBackx, line 165:
```suggestion
        // Update the models on the client side
        for (const post of updatedModels) {
            // We need to do it this way to prevent marking the model as dirty
            this.store.push({
                data: {
                    id: post.id,
                    type: 'post',
                    attributes: {
                        featured: false
                    }
                }
            });
        }
```

ghost/admin/app/components/posts-list/context-menu.js
SimonBackx:
```
// Remove posts that no longer match the filter
this.updateFilteredPosts();
``` 
Is also missing at the end

ghost/admin/app/components/posts-list/context-menu.js
SimonBackx:
```
// Remove posts that no longer match the filter
this.updateFilteredPosts();
``` 
Is also missing at the end

ghost/admin/app/components/gh-context-menu.js
SimonBackx: Maybe put this in a documentation comment so that it also shows up when using this.state?

SimonBackx: And we could add JSDoc with a `'default'|'open'|'modal'|'loading'` value

ghost/admin/app/components/gh-context-menu.js
```diff
@@ -74,4 +128,36 @@ export default class GhContextMenu extends Component {
...
+        this.#modal = this.modals.open(Modal, data);
+        this.#modal.then(() => {
+            this.setState('default');
+        });
+    }
+
+    #closeModal() {
+        this.#modal?.close();
+        this.#modal = null;
+    }
```
SimonBackx, line 156: Much cleaner than passing around the close method 👌

ghost/admin/app/components/gh-context-menu.js
```diff
@@ -2,16 +2,68 @@ import Component from '@glimmer/component';
...
+            this.state = state;
+            return;
+        case 'loading':
+            if (this.state !== 'open' && this.state !== 'modal') {
+                return;
+            }
+            this.isOpen = false;
+            this.selectionList.freeze();
+            this.state = state;
+            return;
```
SimonBackx, line 63: I'm wondering if we should already close here. Because there is no indication that the action is still loading at the moment, unless we add that somewhere. But probably easier to add an indication in the context menu until the loading is done. Wdyt?

PR #16649, opened 2023-04-17
Added author & title to post revisions

mike182uk:
no issue

Added `created_by` and `title` columns to `post_revisions` table

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.45/2023-04-17-11-05-add-post-revision-author.js
+++ b/ghost/core/core/server/data/migrations/versions/5.45/2023-04-17-11-05-add-post-revision-author.js
@@ -1,7 +1,7 @@
 const {combineNonTransactionalMigrations,createAddColumnMigration} = require('../../utils');
 
 module.exports = combineNonTransactionalMigrations(
-    createAddColumnMigration('post_revisions', 'created_by', {
+    createAddColumnMigration('post_revisions', 'author_id', {
         type: 'string',
         maxlength: 2000,
         nullable: true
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/data/migrations/versions/5.45/2023-04-17-11-05-add-post-revision-author.js
```diff
@@ -0,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('post_revisions', 'created_by', {
+    type: 'string',
+    maxlength: 24,
+    nullable: true
```
mike182uk, line 6: Set this to `nullable` to not interfere with existing records

allouis: 👍

ghost/core/core/server/data/migrations/versions/5.45/2023-04-17-11-05-add-post-revision-author.js
```diff
@@ -0,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('post_revisions', 'created_by', {
```
allouis: Appreciate you following existing convention, but I think because these are not resources we should use `author_id` instead!

allouis: cc @daniellockyer wdyt?

allouis (commented):
Just one comment about column name otherwise seems 👍

ghost/core/core/server/data/migrations/versions/5.45/2023-04-17-11-05-add-post-revision-author.js
mike182uk: @allouis No problemo, thats updated 👍

PR #16654, opened 2023-04-17
Added announcement data to settings

naz:
refs https://github.com/TryGhost/Team/issues/3011

- This is a data structure needed to support Announcement Bar feature - allows to create custom site-wide announcements tailored to the audience.
- The `announcement_content` is meant to hold displayed HTML content of the announcement and will be exposed through unauthenticated Content Site API

- The `announcement_visibility` sets the target audience to display the Announcement Bart to:
  - `public` - Everyone
  - `visitors` - Logged out visitors only
  - `members` - Members only
  - `paid` - Paid members only

- The `announcement_background` sets the CSS class that should be applied to the Announcement Bar. and will be exposed through unauthenticated Content Site API. Three styles are available:
  - `accent` - matches the color of the site accent
  - `dark` - dark style
  - `light` - light style

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.45/2023-04-18-12-56-add-announcement-settings.js
+++ b/ghost/core/core/server/data/migrations/versions/5.45/2023-04-18-12-56-add-announcement-settings.js
@@ -1,0 +1,24 @@
+const {combineTransactionalMigrations, addSetting} = require('../../utils');
+
+module.exports = combineTransactionalMigrations(
+    addSetting({
+        key: 'announcement_content',
+        value: null,
+        type: 'string',
+        flags: 'PUBLIC',
+        group: 'announcement'
+    }),
+    addSetting({
+        key: 'announcement_visibility',
+        value: 'public',
+        type: 'string',
+        group: 'announcement'
+    }),
+    addSetting({
+        key: 'announcement_background',
+        value: 'dark',
+        type: 'string',
+        flags: 'PUBLIC',
+        group: 'announcement'
+    })
+);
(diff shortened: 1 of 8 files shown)


ghost/core/core/server/data/migrations/versions/5.45/2023-04-18-12-56-add-announcement-settings.js
```diff
@@ -0,0 +1,24 @@
+const {combineTransactionalMigrations, addSetting} = require('../../utils');
+
+module.exports = combineTransactionalMigrations(
+    addSetting({
+        key: 'announcement_content',
+        value: null,
+        type: 'string',
+        flags: 'PUBLIC',
```
naz, line 8: We have not added these 'PUBLIC' flags in a while. I think they are appropriate here as the field will be exposed through unauthenticated public API. But maybe we have changed thinking about the use of these that I have missed 🤔

allouis (approved):

lenabaidakova (approved):
🚀

ghost/core/core/server/data/migrations/versions/5.45/2023-04-18-12-56-add-announcement-settings.js
naz: As discussed internally, we still keep using the `flags` field as we did not have a decision on removing it. Also made a fix to the migration utility making it possible to add these fields in migrations - https://github.com/TryGhost/Ghost/commit/7969399cdf99fe5d99bccf82084f2c97656bfe3c

PR #16670, opened 2023-04-18
Removed post revision author id on user deletion

mike182uk:
no issue

When a user is deleted any post revisions created by the user are set to be owned by nobody (null) rather than deleting the post revisions associated with the user


allouis (commented):
Looks good, one small comment

PR #16673, opened 2023-04-19
Fixed erroneous timing in reported Mailgun metrics

daniellockyer:
- we have calls to the metrics library so we can measure the time it takes the Mailgun API to return a response
- however, there's a bug in the code whereby if the `batchHandler` takes a long time and then throws an error, this time will be reported to metrics
- this is misleading because it looks like Mailgun is taking a long time if the databases are slow
- this pulls the specific SDK call out into a function so it's easier to wrap with timing code

---

### <samp>🤖 Generated by Copilot at d20e4a3</samp>

Refactor `fetchEvents` method in `mailgun-client.js` to use a new helper method for Mailgun API calls and metrics. This improves code quality and maintainability.

### DIFF
--- a/ghost/mailgun-client/lib/mailgun-client.js
+++ b/ghost/mailgun-client/lib/mailgun-client.js
@@ -111,6 +111,29 @@ module.exports = class MailgunClient {
     }
 
     /**
+     * @param {import('mailgun.js').default} mailgunInstance
+     * @param {Object} mailgunConfig
+     * @param {Object} mailgunOptions
+     */
+    async getEventsFromMailgun(mailgunInstance, mailgunConfig, mailgunOptions) {
+        const startTime = Date.now();
+        try {
+            const page = await mailgunInstance.events.get(mailgunConfig.domain, mailgunOptions);
+            metrics.metric('mailgun-get-events', {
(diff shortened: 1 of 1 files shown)


ghost/mailgun-client/lib/mailgun-client.js
```diff
@@ -153,26 +173,18 @@ module.exports = class MailgunClient {
...
-                    statusCode: 200
-                });
+
                 // We need to cap events at the time we started fetching them (see comment above)
                 events = (page?.items?.map(this.normalizeEvent) || []).filter(e => !!e && e.timestamp <= startDate);
                 debug(`fetchEvents: finished fetching next page with ${events.length} events`);
             }
         } catch (error) {
-            // Log and re-throw Mailgun errors
+            // Log and re-throw errors
```
allouis, line 186: This comment is useless

allouis (approved):
Nice - this is deffo the way to do it going forward, anything that needs metrics should be isolated like this

ghost/mailgun-client/lib/mailgun-client.js
allouis: (no offence) remove it if you cba

PR #16687, opened 2023-04-20
Corrected announcement_visibility setting type

naz:
refs https://github.com/TryGhost/Ghost/pull/14264

- With a requirement change we need to transform `announcement_visibility` setting to be an "array" instead of a "string". Array structure will allow us to hold multiple filters at once giving more coverage to the audience targeting usecases.
- Example filter variations we'll support are: 
[ ] Logged out visitors
[ ] Members
[ ] Free members
[ ] Paid members

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/validators/input/settings.js
+++ b/ghost/core/core/server/api/endpoints/utils/validators/input/settings.js
@@ -22,7 +22,8 @@ module.exports = {
             const arrayTypeSettings = [
                 'notifications',
                 'navigation',
-                'secondary_navigation'
+                'secondary_navigation',
+                'announcement_visibility'
             ];
 
             const emailTypeSettings = [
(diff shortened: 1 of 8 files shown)


ghost/core/core/server/api/endpoints/utils/validators/input/settings.js
```diff
@@ -64,14 +65,26 @@ module.exports = {
...
             if (setting.key === 'announcement_visibility') {
-                const visibilityValue = setting.value;
+                // NOTE: safe to parse because of array validation up top
+                const visibilityValues = JSON.parse(setting.value);
 
-                if (!['public', 'visitors', 'members', 'paid'].includes(visibilityValue)) {
-                    const visibilityError = new ValidationError({
-                        message: tpl(messages.invalidAnnouncementVisibilityValueReceived),
-                        property: setting.key
+                // NOTE: combination of 'free_members' & 'paid_members' makes just a 'members' filter
```
lenabaidakova, line 71: If paid membership is not enabled, can we send just `free_members` filter, or should it be  `free_members` & `paid_members`?

ghost/core/core/server/data/migrations/versions/5.45/2023-04-18-12-56-add-announcement-settings.js
```diff
@@ -8,12 +8,6 @@ module.exports = combineTransactionalMigrations(
         flags: 'PUBLIC',
         group: 'announcement'
     }),
-    addSetting({
```
lenabaidakova, line 11: noob question 🐢 : are we permitted to edit previous migrations before release?

lenabaidakova (approved):
Checked it locally, looks good :)
Have a couple of questions, but they're quite general (more for future me)

ghost/core/core/server/api/endpoints/utils/validators/input/settings.js
naz: I think we can make a synthetic variable for the API Client saying just `'members'`, which the backend will transform under the hood to handle the logic. Would be much nicer for the API client to understand rather than digging into the details of "what to combine in which order". What do you think?

ghost/core/core/server/data/migrations/versions/5.45/2023-04-18-12-56-add-announcement-settings.js
naz: This goes by similar logic like "renaming the column" - [described here](https://www.notion.so/ghost/Database-migrations-eb5b78c435d741d2b34a582d57c24253?pvs=4#a1693683c4da4f42b39ca335e4d23f00). Removing this bit of migration is equivalent to "nooping" the migration.

ghost/core/core/server/api/endpoints/utils/validators/input/settings.js
lenabaidakova: It sounds good, will be easier to handle it on the client with such approach 🙌

PR #16693, opened 2023-04-21
Added `post-status` and `reason` col to `post_revisions`

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/3099

### <samp>🤖 Generated by Copilot at 83140b5</samp>

This pull request adds a new column `post_status` to the `post_revisions` table to store the status of the post revision. It also updates the schema definition, the migration file, and the integrity test to reflect the new column.

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.45/2023-04-21-08-54-add-post-revision-status.js
+++ b/ghost/core/core/server/data/migrations/versions/5.45/2023-04-21-08-54-add-post-revision-status.js
@@ -6,11 +6,8 @@ module.exports = combineNonTransactionalMigrations(
     createAddColumnMigration('post_revisions', 'post_status', {
         type: 'string',
         maxlength: 50,
-        nullable: true
-        defaultTo: 'draft',
-        validations: {
-            isIn: [['published', 'draft', 'scheduled', 'sent']]
-        }
+        nullable: true,
+        defaultTo: 'draft'
     }),
     createAddColumnMigration('post_revisions', 'reason', {
         type: 'string',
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/data/schema/schema.js
```diff
@@ -404,7 +404,8 @@ module.exports = {
         created_at_ts: {type: 'bigInteger', nullable: false},
         created_at: {type: 'dateTime', nullable: false},
         author_id: {type: 'string', maxlength: 24, nullable: true, references: 'users.id', cascadeDelete: false, constraintName: 'post_revs_author_id_foreign'},
-        title: {type: 'string', maxlength: 2000, nullable: true, validations: {isLength: {max: 255}}}
+        title: {type: 'string', maxlength: 2000, nullable: true, validations: {isLength: {max: 255}}},
+        post_status: {type: 'string', maxlength: 50, nullable: true, validations: {isIn: [['draft', 'published', 'scheduled']]}}
```
cmraible, line 408: should we add 'sent' as an available status? I think if they send as an email but don't publish, the status will be 'sent', which we prob want to indicate on the revisions as well.

cmraible: oh yea looks like you've got it in the migration, just not in the schema

ghost/core/core/server/data/migrations/versions/5.45/2023-04-21-08-54-add-post-revision-status.js
```diff
@@ -0,0 +1,15 @@
+// For information on writing migrations, see https://www.notion.so/ghost/Database-migrations-eb5b78c435d741d2b34a582d57c24253
+
+const {createAddColumnMigration} = require('../../utils');
+
+// status: {type: 'string', maxlength: 50, nullable: false, defaultTo: 'draft', validations: {isIn: [['published', 'draft', 'scheduled', 'sent']]}},
+
+module.exports = createAddColumnMigration('post_revisions', 'post_status', {
+    type: 'string',
+    maxlength: 50,
+    nullable: false,
```
cmraible, line 10: should be nullable, right?

allouis (approved):

PR #16695, opened 2023-04-21
Added feature_image to post_revisions

allouis:
This will allow us to store historical data for feature images so we can diff and restore them

---

### <samp>🤖 Generated by Copilot at 0cd8088</samp>

This pull request adds a new column `feature_image` to the `post_revisions` table to store the feature image URL of a post revision. It also updates the schema definition and the integrity test hash to reflect the database change.


ronaldlangeveld (approved):
LGTM 👍🏻

PR #16707, opened 2023-04-25
🐛 Fixed images sometimes being stored as `data:` URLs when copy/pasting from other editors

kevinansfield:
refs https://github.com/TryGhost/Team/issues/2887

Images could sometimes be pasted into the editor (noticed especially with Google Docs) with `data:` URLs rather than typical `https:` URLs. That causes problems because data URLs are large binary blobs that get stored in the `posts` table and passed through many areas of the system that doesn't expect large binary blobs, causing knock-on effects.

- added handling to our editor's image card to detect when the card is displayed in the editor with a `data:` URL and if it was then it converts it to a file and uploads it so the image can be stored and displayed the same way as any other image
 - handles uploads on both paste and opening a post in the editor that was previously saved with a `data:` URL

### DIFF
--- a/ghost/admin/lib/koenig-editor/addon/components/koenig-card-image.js
+++ b/ghost/admin/lib/koenig-editor/addon/components/koenig-card-image.js
@@ -5,12 +5,36 @@ import {
     IMAGE_MIME_TYPES
 } from 'ghost-admin/components/gh-image-uploader';
 import {action, computed, set, setProperties} from '@ember/object';
+import {fetch} from 'fetch';
 import {utils as ghostHelperUtils} from '@tryghost/helpers';
 import {isEmpty} from '@ember/utils';
 import {run} from '@ember/runloop';
 import {inject as service} from '@ember/service';
 
 const {countWords} = ghostHelperUtils;
+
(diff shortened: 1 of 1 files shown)


ghost/admin/lib/koenig-editor/addon/components/koenig-card-image.js
```diff
@@ -5,13 +5,37 @@ import {
...
 import {utils as ghostHelperUtils} from '@tryghost/helpers';
 import {isEmpty} from '@ember/utils';
 import {run} from '@ember/runloop';
 import {inject as service} from '@ember/service';
 
 const {countWords} = ghostHelperUtils;
 
+// when converting data: URIs to files we need to give them unique names so
+// we don't run into server-side race condition issues with filename deduplication
+let dataImgCount = 1;
```
kevinansfield, line 18: I thought this had been fixed but I saw it crop up again when running locally and multiple separate image uploads were instantly triggered. Unlikely to happen in real-world conditions due to various latencies with non-local requests but something to keep an eye on.

Symptom appeared to be images being saved over the top of each other so there were duplicates showing on the page.

naz (approved):
Works great!

PR #16744, opened 2023-05-05
added Arabic locale for Portal

KSAART:
I think this is perfect Arabic, however If someone can improve then I will be happy.


9larsons:
Thanks for the submission. I'm closing this for cleanup as we have these strings at this point.

PR #16745, opened 2023-05-05
Cleaned up member attribution flag

SimonBackx:
no issue

This commit removes the `memberAttribution` feature flag from the codebase. Some CSS classes are not removed as removing them and updating the associated CSS files have side effects sadly.


ghost/admin/app/components/members-activity/table-row.hbs
```diff
@@ -36,7 +36,7 @@
                         {{/if}}
                     </span>
                     {{#if event.description}}
-                        <div class="ghost-members-activity-event-description {{if (feature "memberAttribution") 'feature-memberAttribution'}}">
+                        <div class="ghost-members-activity-event-description feature-memberAttribution">
```
SimonBackx, line 39: Removing these classes and updating the associated CSS files have side effects sadly, so I'm currently going to keep these in

PR #16751, opened 2023-05-07
Updated Czech translations

Zavrod:
Just minor details


daniellockyer:
Hi @Zavrod, you seem to have recreated the files under `cz`, but the ISO code for Czech is `cs` as is currently implemented. Would you be able to make your changes on the originals?

daniellockyer:
Thanks @Zavrod! 🙂

PR #16757, opened 2023-05-08
Added support for inviting users via Admin API Integration

allouis:
Whilst Admin API Integrations had the permissions to create invites they were
blocked from doing so at the HTTP level. We've removed this restriction for
creating Invites as well as browsing Roles, because a Role ID is necessary to
create an invite. The code was also not setup to support Admin API Integrations
as it made assumptions about the existence of a User. That has been updated in
the permissions layer - so that the Invites are limited to Contributors,
Authors and Editors as well as at the email layer, which has has the copy and
from address updated to reflect the lack of a User creating the Invite.


allouis:
The email template has been duplicated and edited to support changes in content.

Originally:
```
Rediverge is using Ghost to publish things on the internet! John has invited you to join them. Please click the link below to activate your account.

...

If you have trouble activating your Rediverge account, you can reach out to John on john@ghost.org for assistance.
```

Now:
```
Rediverge is using Ghost to publish things on the internet! You have been invited to join them. Please click the link below to activate your account.

...

If you have trouble activating your Rediverge account, you can reach out on ghost@rediverge.com for assistance.
```

PR #16758, opened 2023-05-08
Added Uzbek locale

eugene0928:


daniellockyer:
Thanks @eugene0928! 🕺🏻

PR #16760, opened 2023-05-08
Update search text highlighting to escape HTML

binary-koan:
refs: https://github.com/TryGhost/Team/issues/2390

Escapes each matched and non-matched segment of the post title in the admin search field, to make sure they're displayed in plain text but still have matches highlighted.


sagzy (commented):
Looks good to me – @daniellockyer can you also give this a quick 👀 please?

daniellockyer (approved):

PR #16761, opened 2023-05-08
Use default email sender for setting change confirmations

binary-koan:
refs: https://github.com/TryGhost/Team/issues/2527

I looked through the history of these services but couldn't see a particular reason why they override the sender domain. So just removed the custom handling, which will make them use the default `noreply@<site>` or `mail.from` config

Changed emails
* Support address change confirmation
* Newsletter from address change confirmation (had the same problem)

### DIFF
--- a/ghost/core/core/server/services/settings/SettingsBREADService.js
+++ b/ghost/core/core/server/services/settings/SettingsBREADService.js
@@ -31,6 +31,7 @@ class SettingsBREADService {
         /* email verification setup */
 
         this.ghostMailer = new mail.GhostMailer();
+        this.getDefaultFromEmail = mail.utils.getDefaultFromEmail;
 
         const {transporter, getSubject, getText, getHTML, getSigninURL} = {
             transporter: {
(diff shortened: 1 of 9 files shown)


sagzy:
@rishabhgrg you have more context on this than me, could you give this a 👀?

ghost/core/core/server/services/settings/SettingsBREADService.js
```diff
@@ -360,13 +360,6 @@ class SettingsBREADService {
      * @private
      */
     async sendEmailVerificationMagicLink({email, key}) {
-        const [,toDomain] = email.split('@');
-
-        let fromEmail = `noreply@${toDomain}`;
-        if (fromEmail === email) {
-            fromEmail = `no-reply@${toDomain}`;
-        }
-
```
rishabhgrg, line 369: So the reason we added this particular logic was to avoid both `to` and `from` address to be the same value, which makes Email Providers to mark the email as spam or Gmail shows a yellow banner. So in case the to address is same as from `noreply@domain`, we change the sender to value to `no-reply` to avoid that.

binary-koan:
After discussion, this is intentional and fixing like this for gmail would break other things

PR #16767, opened 2023-05-10
✨ Implemented duplicate post functionality

mike182uk:
refs: https://github.com/TryGhost/Team/issues/3139 https://github.com/TryGhost/Team/issues/3140

- Added duplicate post functionality to post list context menu
  - Currently only a single post can be duplicated at a time
  - Currently only enabled via the `Making it rain` flag
- Added admin API endpoint to copy a post - `POST ghost/api/admin/posts/<post_id>/copy/`
- Added admin API endpoint to copy a page - `POST ghost/api/admin/pages/<page_id>/copy/`


ghost/posts-service/lib/PostsService.js
```diff
@@ -365,6 +366,58 @@ class PostsService {
...
+
+    async copyPost(frame) {
+        const existingPost = await this.models.Post.findOne({
+            id: frame.options.id,
+            status: 'all'
+        }, frame.options);
+
+        const newPostData = omit(
+            existingPost.attributes,
+            [
```
naz, line 378: I'm slightly cautious when using omit in code as it hides additional maintenance long-term. For example, when we add a new field to the `posts` table we'd have remember to exclude it here too if it's a field we don't want to copy over. On the other hand, we'd get a failure for the snapshot test and would need to update it, so there is some signal for us to check on this list in the future. What are your thoughts on this @mike182uk ?

ghost/core/test/e2e-api/admin/__snapshots__/posts.test.js.snap
```diff
@@ -731,6 +731,336 @@ Object {
...
+Object {
+  "access-control-allow-origin": "http://127.0.0.1:2369",
+  "cache-control": "no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0",
+  "content-length": "3693",
+  "content-type": "application/json; charset=utf-8",
+  "content-version": StringMatching /v\\\\d\\+\\\\\\.\\\\d\\+/,
+  "etag": StringMatching /\\(\\?:W\\\\/\\)\\?"\\(\\?:\\[ !#-\\\\x7E\\\\x80-\\\\xFF\\]\\*\\|\\\\r\\\\n\\[\\\\t \\]\\|\\\\\\\\\\.\\)\\*"/,
+  "location": StringMatching /https\\?:\\\\/\\\\/\\.\\*\\?\\\\/posts\\\\/\\[a-f0-9\\]\\{24\\}\\\\//,
+  "vary": "Accept-Version, Origin, Accept-Encoding",
+  "x-cache-invalidate": "/*",
```
naz, line 933: Seems like we still get full cache invalidation here 😬

mike182uk: Ah, I think this is from the API call in the test to first [add a post](https://github.com/TryGhost/Ghost/pull/16767/files/5a3b97986c8f6893ee200fd49847617e0839807f#diff-28524bbad2eeb2d3e802b88861c642f7c7161eccd472e081a5a2353f40dfc953R461-R476) 

[This](https://github.com/TryGhost/Ghost/pull/16767/files/5a3b97986c8f6893ee200fd49847617e0839807f#diff-e54f061c2d15ed48485978a227b90dca6e502ff43fabe311f48d3c1f581f88faR1032) will be the headers for the newly added endpoint

When testing locally, definitely not getting the cache invalidation header when copying

naz: ah... ok after debugging this for a bit figured it's the first request creating the post 😅 we don't need the explicit `cacheInvalidate = false` in the controller though! will make a comment on that separately.

mike182uk: Snap 😀

ghost/core/core/server/api/endpoints/pages.js
```diff
@@ -238,5 +238,31 @@ module.exports = {
...
+            }
+        },
+        permissions: {
+            docName: 'posts',
+            method: 'add'
+        },
+        async query(frame) {
+            return postsService.copyPost(frame)
+                .then((model) => {
+                    this.headers.cacheInvalidate = false;
```
naz, line 262: we don't need this declaration here if we also remove the `headers: {},` declaration above. in this case the code just becomes `return postsService.copyPost(frame);`

ghost/core/core/server/api/endpoints/posts.js
```diff
@@ -279,5 +279,30 @@ module.exports = {
...
+                required: true
+            }
+        },
+        permissions: {
+            method: 'add'
+        },
+        async query(frame) {
+            return postsService.copyPost(frame)
+                .then((model) => {
+                    this.headers.cacheInvalidate = false;
```
naz, line 302: same thing as [here](https://github.com/TryGhost/Ghost/pull/16767/files#r1190778503)

ghost/core/test/e2e-api/admin/pages.test.js
```diff
@@ -430,4 +430,33 @@ describe('Pages API', function () {
         res.body.should.be.empty();
         res.headers['x-cache-invalidate'].should.eql('/*');
     });
+
+    it('Can copy a page', async function () {
```
naz, line 434: We should not be making any more of these test using the old way of e2e testing - without snapshots. Lets take similar strategy here that we took with old posts tests - [posts-legacy.test.js](https://github.com/TryGhost/Ghost/blob/72ed8f56f641fc1ca81e4e363fcd282c195f95c0/ghost/core/test/e2e-api/admin/posts-legacy.test.js#L22). Lets rename this test suite to `pages-legacy.test` and than the new test in `pages.test.js` should be written with a snapshot.

ghost/api-framework/lib/headers.js
```diff
@@ -122,16 +122,26 @@ module.exports = {
         }
 
         const locationHeaderDisabled = apiConfigHeaders?.location === false;
-        const hasFrameData = frame?.method === 'add' && result[frame.docName]?.[0]?.id;
+        const hasFrameData = ['add', 'copy'].includes(frame?.method) && result[frame.docName]?.[0]?.id;
```
naz, line 125: I think a cleaner way to pass this information would be allowing for a custom option in the endpoint definition itself. Checking for method names in the framework gives a code smell tbh. 🤔 

We could introduce something like `frame.location` and determine the returned header base on the value there, this way we would not create additional coupling between framework and method names inside of endpoint definitions.

naz (changes requested):
It's a good direction and I think we are 99% there just a few minor tweaks we'd need to do. Great job Mike 🚀

ghost/api-framework/lib/headers.js
naz: We don't have to go throught all controllers, just check if the new config value is present than use it and if not leave the previous logic as is.

ghost/core/core/server/api/endpoints/pages.js
mike182uk: Fixed in https://github.com/TryGhost/Ghost/pull/16767/commits/bd8dc9037be19589336dd0bb834ab58ccfe63206

ghost/core/core/server/api/endpoints/posts.js
mike182uk: Fixed in https://github.com/TryGhost/Ghost/pull/16767/commits/bd8dc9037be19589336dd0bb834ab58ccfe63206

ghost/api-framework/lib/headers.js
mike182uk: See what you think to https://github.com/TryGhost/Ghost/pull/16767/commits/15191f86ba1e1b1c9c670505c6b3c96ebe3db51e 👍

ghost/api-framework/lib/headers.js
```diff
@@ -122,28 +122,24 @@ module.exports = {
         }
 
         const locationHeaderDisabled = apiConfigHeaders?.location === false;
-        const hasFrameData = ['add', 'copy'].includes(frame?.method) && result[frame.docName]?.[0]?.id;
+        const hasLocationResolver = apiConfigHeaders?.location?.resolve;
+        const hasFrameData = (frame?.method === 'add' || hasLocationResolver) && result[frame.docName]?.[0]?.id;
```
naz, line 126: Yessss! No additional hardcoded method in the framework layer and gives us a clear path of migration to move away from hardcoded `add` when we find appetite for it 👍

naz (approved):
The copy post feature is good for prime time :shipit:

PR #16769, opened 2023-05-10
Added @tryghost/in-memory-repository

allouis:
This is the first pass at an abstract class for our in-memory repository implementations. This should speed up local development and deduplicate code

### DIFF
--- a/ghost/in-memory-repository/src/InMemoryRepository.ts
+++ b/ghost/in-memory-repository/src/InMemoryRepository.ts
@@ -1,0 +1,80 @@
+import nql from '@tryghost/nql';
+
+type Entity<T> = {
+    id: T;
+    deleted: boolean;
+}
+
+type Order<T> = {
+    field: keyof T;
+    direction: 'asc' | 'desc';
+}
+
+type OrderOption<T extends Entity<any>> = Order<T>[];
+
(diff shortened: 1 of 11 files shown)


ghost/in-memory-repository/src/InMemoryRepository.ts
```diff
@@ -0,0 +1,80 @@
+import nql from '@tryghost/nql';
+
+type Entity<T> = {
+    id: T;
+    deleted: boolean;
```
naz, line 5: @allouis I'm a confused about this "deleted" property. Why do we need to have it around instead of removing the Entity from the store alltogether?  I'm missing some context probably, so wanted to understand what use case or reasoning led to such data structure.

ghost/in-memory-repository/src/InMemoryRepository.ts
```diff
@@ -0,0 +1,80 @@
...
+                return item;
+            });
+        } else {
+            this.store.push(entity);
+            this.ids.set(entity.id, true);
+        }
+    }
+
+    public async getById(id: string): Promise<T | null> {
+        return this.store.find(item => item.id === id) || null;
```
naz, line 44: Should we be calling toPrimitive here too, like we do in getAll ?

PR #16773, opened 2023-05-11
🐛 Fixed outbound link tagger tagging non http urls

mike182uk:
refs https://github.com/TryGhost/Team/issues/3172

The outbound link tagger was tagging non http urls (i.e `javascript:`, `mailto:`) which would prevent these urls from working as expected. This change only allows urls to be tagged if they use the `http(s)` protocol.

### DIFF
--- a/ghost/member-attribution/lib/OutboundLinkTagger.js
+++ b/ghost/member-attribution/lib/OutboundLinkTagger.js
@@ -55,6 +55,11 @@ class OutboundLinkTagger {
             return url;
         }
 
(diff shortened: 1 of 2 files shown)


allouis (approved):

allouis (commented):
I don't think this will catch protocol-less URLs e.g. `://whatever.com`? 🤔

allouis:
Though it seems that the URL object doesn't support them anyway - so I think we're good

PR #16778, opened 2023-05-11
🐛 Fixed trailing slash and space in HTML metadata elements

benjaminrancourt:
no issue
The Nu Html Checker from https://validator.w3.org/nu/ says "Trailing slash on void elements has no effect and interacts badly with unquoted attribute values."

For example: https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.merisia.ca%2F
image

---

### <samp>🤖 Generated by Copilot at 5f18867</samp>

Refactor `ghost_head` helper to generate valid HTML5 and enable webmention integration.

### DIFF
--- a/ghost/core/core/frontend/helpers/ghost_head.js
+++ b/ghost/core/core/frontend/helpers/ghost_head.js
@@ -208,15 +208,15 @@ module.exports = async function ghost_head(options) { // eslint-disable-line cam
         if (context) {
             // head is our main array that holds our meta data
             if (meta.metaDescription && meta.metaDescription.length > 0) {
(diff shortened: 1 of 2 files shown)


daniellockyer:
Hey @benjaminrancourt 🙂 The change looks good to me but would you be able to fix the tests? The snapshots can be updated with `UPDATE_SNAPSHOTS=1 yarn test:unit` within `ghost/core`, and then check + commit the changes

benjaminrancourt:
@daniellockyer The tests don't seem to running well on my computer (maybe because I'm currently using Windows?), but I updated the snapshots as requested, so it should be OK in the pipelines!

daniellockyer:
Thanks @benjaminrancourt! 🙂

PR #16796, opened 2023-05-16
Added a collections package with CRUD logic

naz:
closes https://github.com/TryGhost/Team/issues/3166

### <samp>🤖 Generated by Copilot at 7ed6756</samp>

This pull request introduces a new package called `collections` that manages everything to do with collections, which are groups of posts that can be filtered and ordered in different ways. The pull request adds the files for the package metadata, configuration, documentation, type definition, service, repository, and tests. The purpose of the pull request is to implement the core functionality and logic of the collections package, and to ensure its quality and consistency.


ghost/collections/src/CollectionsService.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    async getAll(): Promise<Collection[]> {
+        return await this.repository.getAll();
+    }
+
+    async destroy(id: string): Promise<void> {
+        const collection = await this.getById(id);
+
+        if (collection) {
```
naz, line 29: @allouis  this is another bit that is strange to implement on the service layer. I think it would be much better to have the `destroy` method baked into the repository interface instead of leaving it up to a service to a leak knowledge about "deleted" property. Wdyt?

ghost/collections/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "repository": "https://github.com/TryGhost/Ghost/tree/main/packages/collections",
+    "author": "Ghost Foundation",
+    "private": true,
+    "main": "build/index.js",
+    "types": "build/index.d.ts",
+    "scripts": {
+        "dev": "tsc --watch --preserveWatchOutput --sourceMap",
+        "build": "tsc",
+        "prepare": "tsc",
+        "test:unit": "NODE_ENV=testing c8 --src src --all --exclude 'src/Collection.ts' --exclude 'test/' --check-coverage --100 --reporter text --reporter cobertura mocha -r ts-node/register './test/**/*.test.ts'",
```
naz, line 13: @allouis I didn't find a quick a better quick way to not count the file that only includes a type definition in the test coverage, so used this `--exclude 'src/Collection.ts'` explicit way as a stopgap. have you encountered a better way in the past?

ghost/collections/src/Collection.ts
```diff
@@ -0,0 +1,15 @@
+// @NOTE: file names having only type declarations should also
+//        be uppercased
+/* eslint-disable ghost/filenames/match-regex */
```
naz, line 3: @allouis I think files only including `type` export should be allowed to be capitalized too 🤔

ghost/collections/package.json
vikaspotluri123: I ran into the same thing in simple-mfa; my solution was to have a dedicated folder for interfaces and [exclude](https://togithub.com/vikaspotluri123/simple-mfa/blob/e51a557bbe5cd95b1c5c685f8c93b21585c26025/.c8rc.json) it in my c8 config.

However, since your tests are written in typescript a possible solution is to use the types in the tests (e.g. explicitly typing a variable) - but that also depends on well ts-node and c8 work together 🙂

naz: Oh that's a good shout with a separate interfaces folder 🙌

PR #16813, opened 2023-05-17
Updated public string wrapping

eugene0928:


eugene0928:
test error in unit tests

daniellockyer:
Hey @eugene0928, thank you for the PR. After consideration, we decided to do this task internally so we could make sure they were all done in a timely manner. As such, your PR is now outdated and I'm going to close it but please consider checking out the translations we currently have 🙂

PR #16821, opened 2023-05-18
🐛 Fixed issue where single letter product slugs cause 500 error

mike182uk:
refs https://github.com/TryGhost/Team/issues/3224

When a product has a slug that is a single letter, checking if a user had access to view a post associated with that product would cause a 500 error. The underlying cause of this issue is https://github.com/TryGhost/NQL/issues/20 This fix circumvents this issue by providing a value that the nql lexer will not error out on


naz (approved):
:shipit: let's get it into main. nice find Mike!

PR #16823, opened 2023-05-18
Added forced debug output for EmailRecipients fetching

naz:
refs https://github.com/TryGhost/Team/issues/3229

- The issue we are observing that even though the returned amount of email recipients should not ever accede the max batch size (1000 in case of MailGun), there are rare glitches when this number is doubled and we fetch 2000 records instead.
- The fix takes it's best guess in de-duping data in the batch and then truncates it if the amount of records is still above the threshold. This ensures we at least end up sending the emails out to some of the recipients instead of none.
---

### <samp>🤖 Generated by Copilot at 1757e81</samp>

This pull request enhances the `BatchSendingService` class for email sending with better documentation, logging, and error handling. It also fixes a minor bug and adds a warning for large batches.

### DIFF
--- a/ghost/email-service/lib/BatchSendingService.js
+++ b/ghost/email-service/lib/BatchSendingService.js
@@ -415,7 +415,7 @@ class BatchSendingService {
             if (members.length > this.#sendingService.getMaximumRecipients()) {
                 // @NOTE the unique by member_id is a best effort to make sure we don't send the same email to the same member twice
                 logging.warn(`Email batch ${originalBatch.id} has ${members.length} members, which exceeds the maximum of ${this.#sendingService.getMaximumRecipients()}. Filtering to unique members`);
-                members = uniqBy(members, 'member_id');
+                members = uniqBy(members, 'email');
 
                 if (members.length > this.#sendingService.getMaximumRecipients()) {
                     // @NOTE this is a best effort logic to still try sending an email batch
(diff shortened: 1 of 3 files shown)


ghost/email-service/lib/BatchSendingService.js
```diff
@@ -513,6 +529,24 @@ class BatchSendingService {
...
+
+            const _debug = require('@tryghost/debug')._base;
+            try {
+                _debug.enable('ghost-query');
+                await this.#models.EmailRecipient.findAll({filter: `batch_id:${batchId}`, withRelated: ['member', 'member.stripeSubscriptions', 'member.products']});
+            } finally {
+                // @NOTE: disables all debugged modules, there's no
+                //        way to disable just one module.
+                //        The behavior rare enough and is acceptable to catch the live issue
+                _debug.disable();
```
rishabhgrg, line 545: @naz Sorry this is just probably my lack of context on this, but what's the effect of `_debug.disable()` ?

naz: Above we enable `_debug.enable('ghost-query');` debug outputs for db queries, with hopes to fetch what exact queries were run at the time when size of the batch is larger than we expect. the `_debug.disable()` statement stops those outputs after the query produced by `await this.#models.EmailRecipient.findAll()` finishes execution

rishabhgrg: Ah perfect, thanks

rshbhgrg (approved):
LGTM

PR #16826, opened 2023-05-18
Enforce explicit cacheInvalidate header

mike182uk:
refs https://github.com/TryGhost/Team/issues/3005


ghost/core/core/server/api/endpoints/actions.js
```diff
@@ -4,6 +4,9 @@ module.exports = {
     docName: 'actions',
 
     browse: {
+        headers: {
```
naz, line 7: None of the GET methods should ever reset the cache. I think it mostly makes sense to apply the header to the write methods like POST/PUT/DELETE methods.

ghost/api-framework/lib/pipeline.js
```diff
@@ -239,6 +257,9 @@ const pipeline = (apiController, apiUtils, apiType) => {
             }
 
             return Promise.resolve()
+                .then(() => {
+                    return STAGES.validation.controller(apiImpl);
```
naz, line 261: Super clean approach. Love it!

ghost/core/test/unit/api/cache-invalidation.test.js
```diff
@@ -0,0 +1,42 @@
...
+        endpointPaths.forEach((endpointPath) => {
+            const endpointConfig = require(endpointPath);
+            const ignoreKeys = ['docName'];
+
+            Object.keys(endpointConfig).forEach((key) => {
+                if (ignoreKeys.includes(key) || typeof endpointConfig[key] === 'function') {
+                    return;
+                }
+
+                assert.notEqual(
```
naz, line 34: Yup! Let's get this approach out there 👌

ghost/core/test/unit/api/cache-invalidation.test.js
```diff
@@ -0,0 +1,42 @@
...
+    before(async function () {
+        // Initialise models - Utilised by various endpoints to reference static fields (i.e models.Post.allowedFormats) when required in
+        models.init();
+
+        // Initialise settings cache - Utilised by the indentites endpoint when required in
+        await settings.init();
+    });
+
+    it('Controller actions explicitly declare cacheInvalidate header', async function () {
+        const endpointPaths = glob.sync('*.js', {
```
naz, line 19: Would be good to have an extra assertion checking if there are any endpoint paths in case we move where our controllers live

PR #16840, opened 2023-05-22
Add e2e test for update check script

mike182uk:
refs https://github.com/TryGhost/Team/issues/3234

Added an e2e for the update check script to detect potential breakages in the script due to uninitialised dependencies in the isolated execution environment


ghost/core/core/server/update-check.js
```diff
@@ -17,7 +17,7 @@ const UpdateCheckService = require('@tryghost/update-check-service');
  * @returns {Promise<any>}
  */
 module.exports = async ({rethrowErrors = false} = {}) => {
-    const allowedCheckEnvironments = ['development', 'production'];
+    const allowedCheckEnvironments = ['testing', 'testing-mysql', 'development', 'production'];
```
naz, line 20: I wonder if adding the update check on all `testing-*` environments is desired from test run speed perspective? What are the comparative execution time numbers on CI for this version vs the main?

My understanding is we'd set `process.env.UPDATE_CHECK_URL` and `process.env.FORCE_UPDATE` only in [the test](https://github.com/TryGhost/Ghost/pull/16840/files#diff-33cb15986ef23b4b684cbd19cffab0b441cee210f8d07dbb8073072e8ac264fcR29-R31) itself - why not use these two environmnental variables as a condition to initialize the update check? 😜

ghost/core/test/e2e-server/run-update-check.test.js
```diff
@@ -0,0 +1,54 @@
+const {spawn} = require('child_process');
```
naz, line 1: We haven't had this type of test in the past where we'd spawn a whole child process etc. I'd go down the path of getting the job running through Ghost-native path - JobsManager. I'm also thinking tests for jobs specifically should get their own directory?

mike182uk: Aha, why didn't I think of that 😀 I think i got hung up on the fact `run-update-check.js` is essentially a standalone file that can be executed directly. I've updated the test to make use of the job manager, and also moved the test into its own `jobs` directory 👍

ghost/core/core/server/update-check.js
mike182uk: No need for these env vars now we have switched to using the job manager directly - Instead we pass the the info in. Have made it so if `forceUpdate` is explicitly defined, the environment check is ignored.

ghost/core/test/e2e-server/jobs/update-check.test.js
```diff
@@ -0,0 +1,54 @@
...
+        if (mockUpdateServer) {
+            mockUpdateServer.close();
+        }
+    });
+
+    it('successfully executes the update checker', async function () {
+        let mockUpdateServerRequestCount = 0;
+
+        // Initialise mock update server
+        mockUpdateServer = http.createServer((req, res) => {
```
naz, line 27: I'd substitute this with `nock`, which we use everywhere else ([example](https://github.com/TryGhost/Ghost/blob/b58dac262ec7075daa3b3d948cc4860d171bb20a/ghost/members-api/test/unit/lib/geolocation.test.js#L32)) to mock external HTTP dependencies.

ghost/core/core/server/update-check.js
```diff
@@ -14,14 +14,22 @@ const UpdateCheckService = require('@tryghost/update-check-service');
...
  * @param {Object} [options]
  * @param {Boolean} [options.rethrowErrors] - if true, errors will be thrown instead of logged
+ * @param {Boolean} [options.forceUpdate] - if true, the update check will be triggered regardless of the environment or scheudle, defaults to config if no value provided
+ * @param {String} [options.updateCheckUrl] - the url to check for updates against, defaults to config if no value provided
  * @returns {Promise<any>}
  */
-module.exports = async ({rethrowErrors = false} = {}) => {
-    const allowedCheckEnvironments = ['development', 'production'];
+module.exports = async ({
+    rethrowErrors = false,
```
naz, line 22: Nice. Also gets us a step closer to passing all the configs as a parameters instead of depending on the config module 👌

naz (approved):
There is one comment around using nock instead of the manually written mock server, other than that this solves the issue enough for the time budget we have 👍

ghost/core/test/e2e-server/jobs/update-check.test.js
mike182uk: We wouldn't be able to use nock here because the job manager executes the job in a separate process

mike182uk: Comment added in https://github.com/TryGhost/Ghost/pull/16840/commits/154de6ff46a02e4039650eb1b8d2b98cf9cc096b

Issue #16628, opened 2023-04-13
i18n Milestone 3: Wrap all public-facing strings


daniellockyer:
_This issue is part of our community i18n project. The full project is documented [on the Ghost forum](https://forum.ghost.org/t/hacktoberfest-project-i18n-for-portal-member-emails-comments-ui-sodosearch/33062)._

---

✨ We have now started shipping our [i18n beta](https://forum.ghost.org/t/help-translate-ghost-beta/37461) in Ghost, and we're merging community pull requests to [add & update languages](https://github.com/TryGhost/Ghost/tree/main/ghost/i18n/locales).

However, not all public-facing strings in Portal and our member emails are currently interpolated for translations. In order to have a full set of translations, we need to ensure all strings are wrapped with the `t(..)` helper.

#### Setting up

1. Clone the Ghost repository and [set up your local environment for development](https://ghost.org/docs/install/source/)
2. Run `yarn dev --portal` to get things running at http://localhost:2368/
3. Toggle `Enable Portal translations (beta)` from http://localhost:2368/ghost/#/settings/labs
4. (optional) Change the Publication Language to make it easier to find uninterpolated strings: http://localhost:2368/ghost/#/settings/general 

#### Wrapping strings

Once you've found a public-facing string from code review or by using Portal/Ghost signup emails, you can easily wrap it so it will be picked up for translation.

1. Add `t('..')` around the string
  a. You may need to import the `t` function from the AppContext in Portal 	
2. Run `yarn translate` within `ghost/i18n` to ensure the locale JSON files include the new strings

You can see plenty of examples of how this is currently done using this search query: https://github.com/search?q=repo%3ATryGhost%2FGhost+path%3Aghost%2Fportal+%7Bt%28%27&type=code

Please take extra care to ensure you pass any variables into the `t` function, and make sure HTML escaping does not cause problems with the output.

#### Contributing

Commit the changes and submit a PR for us to review and merge 🙂 

Be sure to read the contributing guide and reference this issue on the 3rd line of the commit: https://github.com/TryGhost/Ghost/blob/main/.github/CONTRIBUTING.md

---

Thank you to all contributors on the i18n beta so far 🙏🏻 

If you speak a language that is not included in our current set of supported languages, please consider [submitting it](https://forum.ghost.org/t/help-translate-ghost-beta/37461) 🌐

eugene0928:
new language is added

sam-lord:
Hard case here:

[...]

PR #16861, opened 2023-05-24
Wrapped public-facing strings with i18next `t` function

sam-lord:
refs https://github.com/TryGhost/Ghost/issues/16628

---

### <samp>🤖 Generated by Copilot at f44b375</samp>

This pull request adds internationalization support to the portal component, which is a feature that allows Ghost members to manage their subscriptions and profiles. It uses the `t` function from the app context and the `@doist/react-interpolate` library to translate and interpolate the UI strings according to the user's language preference. It modifies several components and files in the `ghost/portal/src/components/pages/AccountHomePage` directory and adds a new dependency to the `ghost/portal/package.json` file.


ghost/portal/src/components/pages/AccountHomePage/AccountHomePage.js
```diff
@@ -38,6 +38,7 @@ export default class AccountHomePage extends React.Component {
                     onClose={() => this.context.onAction('closePopup')}
                     handleSignout={e => this.handleSignout(e)}
                     supportAddress={supportAddress}
+                    t={this.context.t}
```
daniellockyer, line 41: Maybe we can pull `t` out in `const {member, site} = this.context;` above and pass it in a bit cleaner here?

PR #16862, opened 2023-05-24
Updated member email templates to use i18n function

daniellockyer:
refs https://github.com/TryGhost/Ghost/issues/16628

---

### <samp>🤖 Generated by Copilot at b4c7f22</samp>

Refactored and localized the email templates for signing in, signing up, subscribing, and updating email address. Moved all the templates to `signin.js` and used the translation function `t` to customize and translate the email content based on the site settings.


daniellockyer:
Merged here: https://github.com/TryGhost/Ghost/commit/f2277ded405eeb215bf0f76e4232b8500dd29910

PR #16865, opened 2023-05-25
Merged similar history events

minimaluminium:
refs https://github.com/TryGhost/Team/issues/2463

- it merges the similar history events by adding additional properties to the events data
- `skip` is added to the current event when the next event is similar to it, so that the event isn't output on the frontend
- `count` is added to the last item of the similar events and it's used for outputting the number of repeated similar events

### DIFF
--- a/ghost/admin/app/helpers/history-event-fetcher.js
+++ b/ghost/admin/app/helpers/history-event-fetcher.js
@@ -91,7 +91,26 @@ export default class HistoryEventFetcher extends Resource {
                 this.hasReachedEnd = true;
             }
 
-            actions.forEach((a) => {
+            let count = 1;
+
+            actions.forEach((a, index) => {
+                const nextAction = actions[index + 1] || null;
+
+                // depending on the similarity, add additional properties to be used on the frontend for grouping
+                // skip - used for hiding the event on the frontend
+                // count - the number of similar events which is added to the last item
+                if (nextAction) {
+                    if (a.resource_id === nextAction.resource_id && a.event === nextAction.event) {
+                        a.skip = true;
+                        count += 1;
+                    } else {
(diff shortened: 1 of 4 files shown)


ghost/admin/app/helpers/history-event-fetcher.js
```diff
@@ -91,11 +91,30 @@ export default class HistoryEventFetcher extends Resource {
...
+
+            actions.reverse().forEach((a, index) => {
+                const nextAction = actions[index + 1] || null;
+
+                // depending on the similarity, add additional properties to be used on the frontend for grouping
+                // skip - used for hiding the event on the frontend
+                // count - the number of similar events which is added to the last item
+                if (nextAction || (!nextAction && actions[index - 1].skip)) {
+                    if (nextAction && a.resource_id === nextAction.resource_id && a.event === nextAction.event) {
+                        a.skip = true;
```
naz, line 104: I was gonna write a comment here about how we could use "reduce" here to make things cleaner and only keep the "viewActions" without the skip property. but this approach is more readable and is "good enough" to do what we need 👌

ghost/admin/app/helpers/history-event-fetcher.js
```diff
@@ -91,11 +91,30 @@ export default class HistoryEventFetcher extends Resource {
...
+                            count = 1;
+                        }
+                    }
+                }
+
                 a.context = JSON.parse(a.context);
             });
 
-            this.data.push(...actions);
+            this.data.push(...actions.reverse());
```
naz, line 117: Should we also handle the "similar" actions that are right no the edge of the previously inserted actions? The situation I'm thinking about here is when the last action that we have previously loaded is similar to the very first action in the just fetched actions.

PR #16867, opened 2023-05-25
Added email snapshot test for members webhooks

daniellockyer:
refs https://github.com/TryGhost/Team/issues/3305

- this helps prevent regressions in the emails that we send to members
- should be useful when I'm about to introduce i18n wrappers to the email templates

---

### <samp>🤖 Generated by Copilot at d6bc2d6</samp>

Refactored a test case for Stripe checkout webhooks to use email snapshots. Imported `regexes` module to handle token query parameters in snapshots. Improved email assertions with `emailMockReceiver` from `e2e-framework`.


daniellockyer:
Closing as there is some weird issue with HTML formatting

PR #16873, opened 2023-05-25
Collections <> Posts Admin Api

naz:
refs https://github.com/TryGhost/Team/issues/3260

### <samp>🤖 Generated by Copilot at 8f82f6a</samp>

This pull request adds support for associating posts with collections in the Ghost CMS. It introduces new classes and types for posts data, such as `PostDTO`, `PostsDataRepositoryInMemory` and `PostsDataRepositoryBookshelf`. It updates the `Collection` and `CollectionsService` classes to handle the posts array, and modifies the collections output serializer to include the posts property. It also adds and improves test cases for the collections module and the collections e2e API.


ghost/collections/src/CollectionsService.ts
```diff
@@ -1,41 +1,65 @@
...
+    /**
+     *
+     * @param collection to add tags to
+     * @param postIds
+     */
+    private async addPosts(collection: Collection, postDTOs: PostDTO[]) : Promise<Collection> {
+        const postIds = postDTOs.map(post => post.id);
+        const posts = await this.postsRepository.getBulk(postIds);
+
+        collection.posts = posts.map((post: any) => post.id);
```
allouis, line 35: I think `collection.posts` should be of type `PostDTO[]` but this sets it to `string[]`

ghost/collections/src/PostsDataRepositoryBookshelf.ts
```diff
@@ -0,0 +1,17 @@
+type PostsDataRepositoryBookshelfDeps = {
+    Post: any;
+}
+
+export class PostsDataRepositoryBookshelf {
```
allouis, line 5: We want to keep all bookshelf dependent code together in the Ghost core - so we shouldn't define this repository in the package

ghost/collections/src/Collection.ts
```diff
@@ -22,6 +23,8 @@ export class Collection {
     updatedAt: Date;
     deleted: boolean;
 
+    posts: PostDTO[];
```
allouis, line 26: I'm wondering if this is the best design type here. By requiring a `tile` and `slug` property, we need to always fetch the related data in the DB when using a SQL backed repository. 

That could well be fine if we only want to expose the title/slug from the API - but if we want to include whole post objects then we'd want to either:
1. Do another fetch of the related data to get all of it in the controller/usecase layer
2. Expand this `PostDTO` type to include all the fields which we do wanna use in the API

I think if the fieldset is small then it's okay - but if we think we're gonna wanna expose the entire Post object then we'll be bloating the Collection entity.

So TL;DR - if we just going to respond with a subset of the post fields I think this makes sense, otherwise I would do something like:

```ts
class Collection {
  // stuff...
  posts: ObjectID[]

  addPost(post: PostDTO): void {
    if (this.posts.includes(post.id)) {
      return;
    }
    if (this.type === 'automatic') {
      if (!nql(this.filter).queryJSON(post)) {
        throw new Error('Post doesn't match filter');
      }
    }
    this.posts.push(post.id);
  }
  // stuff..
}
```

This gives us the benefit of the fuller Post object for validation but keeps the exposed fields small to allow us to make fewer DB queries.

allouis (commented):
Some small comments - one larger one around the structure of the Collection entity

ghost/collections/src/PostsDataRepositoryBookshelf.ts
naz: Hm ok. So the packages would only hold the "in memory" kind of repositories and we inject any other Bookshelf/Redis/whatevs from the outside. Am I understanding it correctly?

allouis: Yeah - ideally we keep all the bookshelf related code in one package so it's easier to rip out down the line.

The idea is that the package contains the business logic, and the infrastructure implementations can be outside of it - this keeps the scope of the package narrower and doesn't require stubbed bookshelf dependencies for testing!

At the moment bookshelf repositories have been implemented in `Ghost/ghost/core/server/services/**/BlahBookshelfRepository` but we'll probably want to pull those out to `Ghost/ghost/core/server/repositories/bookshelf` or `Ghost/bookshelf-repositories` or something to keep them all together

naz: yup, makes sense. I've moved the bookshelf posts wrapper into core's codebase 👍

ghost/collections/src/Collection.ts
naz: Agreed on minimizing the the size of exposed fields 👌  

For now we don't have a concrete spec for which fields we'd want to keep around and which ones we don't. My thinking is to start from this basic structure that would be easy to filter right away with NQL and than optimize/reshuffle things as we get feedback. 

An extra thing we'd want to start keeping track straight away is order or items in the collection. So I think if we go with shrinking of the "posts" relation we'd go to something like:
```
{
  id: ObjectID, // to be able to modify the relational object individually
  post_id: ObjectID, 
  sort_order: number
}
```

ghost/collections/src/Collection.ts
```diff
@@ -4,6 +4,7 @@ const {ValidationError} = require('@tryghost/errors');
 const tpl = require('@tryghost/tpl');
 
 import ObjectID from 'bson-objectid';
+import {PostDTO} from './PostDTO';
```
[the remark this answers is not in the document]
naz: I was in decision fatigue between naming the "relation object" either a "Collect" or "Post" or "PostDTO". 

Collect - would be a completely new concept, which I'm not sure we need just yet
Post - would be confusing with actual Post models
PostDTO - seemed like a good name for a thing composed from Posts model instances (entities) and something that is transfered across to different Domain (collections)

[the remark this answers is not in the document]
allouis: FWIW When I picked this work up yesterday and started trying to get it across the line I went with `CollectionPost` that is it's a Post subset specific for Collections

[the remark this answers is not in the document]
naz: exactly what I've been trying to get to today 😆

PR #16887, opened 2023-05-29
Added Finnish locale

mrpuhis:
Complete translation to Finnish.


daniellockyer:
Thanks @mrpuhis! 🕺🏻

mrpuhis:
> Thanks @mrpuhis! 🕺🏻

Hi,

Is there some kind of waiting time after the files have been added? Im using Wave template and at least for now i only get the english texts when i change the language to fi in settings.

daniellockyer:
The translations will be available in the next release of Ghost 🙂

PR #16892, opened 2023-05-30
Added debug data dump to email service

naz:
closes https://github.com/TryGhost/Team/issues/3324

- When the recipients batch size is larger than the limit in addition to logging the error we need extra data to figure out what exactly is inside those `2000` or `3000` records causing faulty behavior.
- This change grabs all available models and dumps them into a file inside of the `content/data` folder. The code is temporary and should be removed once the problem is narrowed down

---

### <samp>🤖 Generated by Copilot at 551532f</samp>

This pull request adds a new feature to the email service module that allows dumping members data to a file when the email recipients limit is exceeded. This is useful for debugging the email service and verifying the members information. The pull request also improves the testing of the `BatchSendingService` class and passes the file path parameter to the `EmailServiceWrapper` class.


ghost/email-service/lib/BatchSendingService.js
```diff
@@ -533,6 +537,20 @@ class BatchSendingService {
...
             logging.error(`Batch ${batchId} has ${mappedMemberLikes.length} members, but the sending service only supports ${BATCH_SIZE} members per batch.`);
+
+            // @NOTE below is a throwaway code, that's why it is dirty and lives here instead of a nice module
+            if (this.#debugStorageFilePath) {
+                const fs = require('fs-extra');
+                const path = require('path');
+                const currentTimeFilename = new Date().toISOString().replace(/:/g, '-').replace(/\./gi, '-');
+                const outputFileName = `email-batch-sending-members-${currentTimeFilename}.json`;
+                const outputFilePath = path.join(this.#debugStorageFilePath, '/', outputFileName);
+                const jsonModels = models.map(m => m.toJSON());
```
naz, line 548: This bit might be flaky. I'll change to a raw "JSON.stringify" on the models instance because we don't really know the shape of this object and could expect anything to happen here

SimonBackx (approved):

naz:
landed in main as https://github.com/TryGhost/Ghost/commit/1605198a1ac5927c8b355efd3250700f7dcdf81b

PR #16894, opened 2023-05-30
Added typescript declerations to i18n

ronaldlangeveld:
refs https://github.com/TryGhost/Team/issues/3307

- added typescript declaration generation as we are building the new signup ember form in React + Typescript and we want to add support for i18n from the beginning.
- This adds a declaration to support importing the i18n js module inside typescript.

### <samp>🤖 Generated by Copilot at bd8d195</samp>

Added TypeScript support and a new namespace for the `i18n` module. This enables type checking and localization for the signup form component in the portal frontend.

### DIFF
--- a/ghost/i18n/package.json
+++ b/ghost/i18n/package.json
@@ -28,7 +28,7 @@
     "c8": "7.13.0",
     "i18next-parser": "8.0.0",
     "mocha": "10.2.0",
-    "typescript": "^5.0.4"
+    "typescript": "5.0.4"
   },
   "dependencies": {
     "i18next": "22.5.0"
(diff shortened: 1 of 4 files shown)


ghost/i18n/package.json
```diff
@@ -25,7 +27,8 @@
   "devDependencies": {
     "c8": "7.13.0",
     "i18next-parser": "8.0.0",
-    "mocha": "10.2.0"
+    "mocha": "10.2.0",
+    "typescript": "^5.0.4"
```
daniellockyer, line 31: Pinned please 🙏🏻 

```suggestion
    "typescript": "5.0.4"
```

PR #16898, opened 2023-05-31
Added storybook to signup form

binary-koan:
refs https://github.com/TryGhost/Team/issues/3299

This uses the standard Storybook layout I'm familiar with - not sure how to set up the alternative layout from Koenig, and not sure if that's something we prefer to use in general.

### DIFF
--- a/ghost/signup-form/.storybook/preview.tsx
+++ b/ghost/signup-form/.storybook/preview.tsx
@@ -1,8 +1,7 @@
 import React from 'react';
 
 import type { Preview } from "@storybook/react";
-import '../src/styles/demo.css';
-import '../src/styles/iframe.css';
+import './storybook.css';
 
 const preview: Preview = {
 	parameters: {
--- a/ghost/signup-form/.storybook/preview.tsx
+++ b/ghost/signup-form/.storybook/preview.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
 
-import '../src/styles/demo.css';
 import type { Preview } from "@storybook/react";
+import '../src/styles/demo.css';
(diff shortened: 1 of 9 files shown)


ghost/signup-form/.vscode/settings.json
```diff
@@ -0,0 +1,17 @@
...
+      "editorGroupHeader.tabsBackground": "#1b1e2b",
+      "tab.border": "#00000000",
+      "sideBarSectionHeader.border": "#00000000",
+      "terminal.border": "#a6accd"
+    },
+    "activityBar.background": "#540E48",
+    "titleBar.activeBackground": "#761365",
+    "titleBar.activeForeground": "#FEFCFE"
+  }
+}
```
SimonBackx, line 17: Should probably be added to gitignore

binary-koan: Oops, that was just from me opening the repo outside a workspace one time. I'll just delete it, shouldn't be added normally

SimonBackx (approved):
Just one comment is that I'm not sure if we should keep FormView and SuccessView + their stories inside the pages folder as they aren't a page any longer. Maybe they could also stay within the same file (FormPage) to keep it easier to edit?

binary-koan:
@SimonBackx Hmm, not sure - it also feels a bit weird moving them into the same files and exporting them as separate components purely for storybook. I would maybe just put everything under `components` rather than having a `pages` directory at all since the total number of components is still small, and figure out a better structure once it becomes hard to manage

I'll merge this as is to unblock the other tasks, but feel free to reorganise the components later!

PR #16900, opened 2023-05-31
Removed emailErrors flag

mike182uk:
refs https://github.com/TryGhost/Team/issues/3337

Removed emailErrors flag as it is no longer needed

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/emails.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/emails.js
@@ -1,4 +1,3 @@
-const config = require('../../../../../../../shared/config');
 const emailService = require('../../../../../../services/email-service');
 
 module.exports = (model, frame) => {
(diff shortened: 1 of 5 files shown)


ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/emails.js
```diff
@@ -19,7 +18,7 @@ module.exports = (model, frame) => {
         }
     }
 
-    if (!labs.isSet('emailErrors') && !!(config.get('bulkEmail') && config.get('bulkEmail').mailgun)) {
+    if (config.get('bulkEmail') && config.get('bulkEmail').mailgun) {
```
daniellockyer, line 21: @SimonBackx Correct me if I'm wrong but removing the flag would actually involve removing this whole block because we don't want to remap the status

SimonBackx: That is right, the whole block needs to be removed. Currently `labs.isSet('emailErrors')` always returns true because it is a GA flag, so this block never executes. What this block was doing, was making sure that any API users (such as admin) never received a status of 'failed', so they would never show an error message when an email failed to send.

mike182uk: Good catch @daniellockyer thanks for the heads up

Nice one for the explanation @SimonBackx 

I've updated this to remove the whole block now 👍

PR #16919, opened 2023-06-02
Updated Bahasa Indonesia translations

rkafp:
The translation for all new strings has been added.


daniellockyer:
Thanks @rkafp! 🕺🏻

PR #16924, opened 2023-06-02
Onboarding events

aileen:
### <samp>🤖 Generated by Copilot at 757ad8c</samp>

This pull request introduces a major refactoring of the segment service, which tracks analytics events for Ghost. It splits the service into two classes, `DomainEventsAnalytics` and `ModelEventsAnalytics`, that handle different types of events. It also adds unit tests for these classes and the service itself.

### DIFF
--- a/ghost/core/core/server/services/segment/DomainEventsAnalytics.js
+++ b/ghost/core/core/server/services/segment/DomainEventsAnalytics.js
@@ -9,7 +9,8 @@ const {MilestoneCreatedEvent} = require('@tryghost/milestones');
  */
 
 /**
- * @typedef {import('../../../shared/sentry')} sentry
+ * @typedef {object} IExceptionHandler
+ * @prop {(err: Error) => void} captureException
  */
 
 /**
@@ -22,7 +23,7 @@ const {MilestoneCreatedEvent} = require('@tryghost/milestones');
  * @param {logging} logging
  * @param {object} trackDefaults
  * @param {string} prefix
- * @param {sentry} sentry
+ * @param {IExceptionHandler} exceptionHandler
  * @param {DomainEvents} DomainEvents
  * @prop {} subscribeToEvents
  */
@@ -34,8 +35,8 @@ module.exports = class DomainEventsAnalytics {
     #trackDefaults;
     /** @type {string} */
     #prefix;
-    /** @type {sentry} */
-    #sentry;
+    /** @type {IExceptionHandler} */
+    #exceptionHandler;
     /** @type {logging} */
     #logging;
     /** @type {DomainEvents} */
@@ -45,7 +46,7 @@ module.exports = class DomainEventsAnalytics {
         this.#analytics = deps.analytics;
         this.#trackDefaults = deps.trackDefaults;
         this.#prefix = deps.prefix;
-        this.#sentry = deps.sentry;
+        this.#exceptionHandler = deps.exceptionHandler;
         this.#logging = deps.logging;
         this.#DomainEvents = deps.DomainEvents;
     }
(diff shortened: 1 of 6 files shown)


ghost/core/core/server/services/segment/DomainEventsAnalytics.js
```diff
@@ -0,0 +1,40 @@
...
+
+    constructor(deps) {
+        this.#analytics = deps.analytics;
+        this.#trackDefaults = deps.trackDefaults;
+        this.#prefix = deps.prefix;
+        this.#sentry = deps.sentry;
+    }
+
+    async #handleMilestoneCreatedEvent(type, event) {
+        if (type === MilestoneCreatedEvent
```
allouis, line 20: I don't think you need this check?

ghost/core/core/server/services/segment/DomainEventsAnalytics.js
```diff
@@ -0,0 +1,40 @@
+const _ = require('lodash');
+const logging = require('@tryghost/logging');
+const DomainEvents = require('@tryghost/domain-events');
```
allouis, line 3: Would be cool to dependency inject this instead, will be easier to test

ghost/core/core/server/services/segment/ModelEventsAnalytics.js
```diff
@@ -0,0 +1,63 @@
...
+        data: {key: 'key', value: 'value'}
+    }
+];
+
+module.exports = class ModelEventsAnalytics {
+    #analytics;
+    #trackDefaults;
+    #prefix;
+    #sentry;
+    #toTrack;
```
allouis, line 38: You can also do `#toTrack = [ ... ]` rather than defining and assigning, not a big deal tho

allouis (commented):
Looks good, few small comments but not a big deal!

It would be nice if we pulled this out into a separate package, but I appreciate if you don't have the time right now

ghost/core/core/server/services/segment/DomainEventsAnalytics.js
aileen: Yeah, will have to touch that again, once we might send a 'StripeLiveEnabled' event or similar

ghost/core/core/server/services/segment/DomainEventsAnalytics.js
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * @typedef {import('@tryghost/logging')} logging
+ */
+
+/**
+ * @typedef {import('analytics-node')} analytics
+ */
+
+/**
+ * @typedef {import('../../../shared/sentry')} sentry
```
aileen, line 12: @allouis Not sure if this is correct, as the sentry module doesn't have types yet... Would it be better to leave it as an `object` instead?

allouis: So generally speaking you wanna define what you _need_ rather than on what you get.

So rather than depending on concrete implementations, we would depend on an interface.

In this case I would do something like:
```
/**
 * @typedef {object} IExceptionHandler
 * @prop ((err: Error) => void) captureException
 */
```

And then the `sentry` object can satisfy that dependency.

allouis: You can even then do `this.exceptionHandler.captureException` - which decouples you from sentry completely! NABD though as sentry is deffo pretty core to Ghost atm

aileen: Cool, did that!! Thanks!

PR #16937, opened 2023-06-05
Updated Turkish translations

EchoBilisim:
missing translations are complete


daniellockyer:
Thanks as always @EchoBilisim! 🙂

Issue #16825, opened 2023-05-18
Mobile keyboard doesn't appear in the same time with the search modal


8lad:
### Issue Summary

When you use the default search on the mobile device, the keyboard doesn't appear in the same time when appears search input. Only after click on the search input the keyboard appears.

<img src="https://github.com/TryGhost/Ghost/assets/42737958/4a4a1acf-57cb-4d00-8d56-2a3ea895344d.png" width=40% height=40%>
<img src="https://github.com/TryGhost/Ghost/assets/42737958/4d6588b9-55ba-4386-9d23-afa17f486b7d.png" width=40% height=40%>

### Steps to Reproduce

1. Click on the search button with data-ghost-search
2. The search modal with the input appears
3. Click on the input
4. The keyboard appears

### Ghost Version

5.40.1

### Node.js Version

18.16.0

### How did you install Ghost?

Ubuntu

### Database type

MySQL 8

### Browser & OS version

Chrome 113.0.5672.76 & Android 12

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

daniellockyer:
Hey there, thank you so much for the bug report.

That does look like something that shouldn't happen! A PR to fix this issue would be very welcome 🙂

amrinder1189:
HI @8lad @daniellockyer , can you please tell ,me where to start or can you elaborate more, Where i can find this The lever page ??

PR #16950, opened 2023-06-07
Fixed search input focus on iOS

minimaluminium:
refs https://github.com/TryGhost/Ghost/issues/16825

- iOS is very restrictive on input focus which doesn't allow without any user interaction
- The only workaround was creating a temporary input, focus it, and remove it after moving the focus to the actual searfch input
- It also moves the inputRef to the parent component, so that it can be used in the click event handler, because the focus event only works when it's inside a function that's triggered after user interaction


naz (approved):
Hacky solution for a hacky problem 🚀

PR #16954, opened 2023-06-07
Added collections CRUD permissions

naz:
refs https://github.com/TryGhost/Team/issues/3220

- Added permissions for collection resources

---

### <samp>🤖 Generated by Copilot at 2d51619</samp>

This pull request adds new permissions for collections, a feature that allows users to group posts by custom criteria. It modifies the `fixtures.json` files in the `data/schema` and `test/utils` folders, and creates a new migration file in the `data/migrations` folder. It also updates the relevant tests in the `test` folder to match the new permissions.

### DIFF
--- a/ghost/core/core/server/api/endpoints/collections.js
+++ b/ghost/core/core/server/api/endpoints/collections.js
@@ -19,8 +19,7 @@ module.exports = {
             'page',
             'filter'
         ],
-        // @NOTE: should have permissions when moving out of Alpha
-        permissions: false,
+        permissions: true,
         query(frame) {
             return collectionsService.api.getAll(frame.options);
         }
@@ -33,8 +32,7 @@ module.exports = {
         data: [
             'id'
         ],
-        // @NOTE: should have permissions when moving out of Alpha
-        permissions: false,
+        permissions: true,
         async query(frame) {
             const model = await collectionsService.api.getById(frame.data.id);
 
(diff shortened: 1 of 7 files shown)


allouis (commented):
We've required separate permissions for adding and removing posts from collections, but I don't see them added to any of the roles. Do we need separate permissions for that, or can we use the edit permission for collections?

Also I don't get why an Author can add a collection but not edit one 🤔

naz:
We did not introduce a separate permission for adding/removing the posts from the collection, because the concept here is similar to modifying the collection itself - e.g. when modifying the `automatic` collection's `filter`. The way I approached it I've [reused the same permissions](https://github.com/TryGhost/Ghost/pull/16954/commits/749531663c6a556c96805c1a870f09673c6f659f) from "collection" level  e.g:
```js
destroyPost: {
\\...
       permissions: {
            method: 'destroy'
       },
\\...
```

> Also I don't get why an Author can add a collection but not edit one 🤔

I've copied permissions to the dot from tags, given that's what we are mimicking here. I think the thinking here was that an Author can create a record to reuse it somewhere but not be able to modify their own or someone else's records.

allouis:
Aha, I missed that you removed the custom docName for those methods - my bad! (Sidenote though, the fact that docName maps to permissions isn't particularly obvious - we should think about making that more explicit in future, e.g. something like `permissions: 'edit:posts'`)

I think these permissions are still incorrect though, both the `addPost` and `destroyPost` methods should use the `'edit'` permission action IMO, as they're editing a collection not adding or destroying one!

Hmm - that seems really weird to me, but if we wanna replicate tags permissions then okay 👍

PR #16958, opened 2023-06-07
Optimized number of database queries when inserting fixtures

daniellockyer:
refs https://github.com/TryGhost/Toolbox/issues/592

- this section of code inserts fixtures into the DB if they don't already exist
- the check for whether they already exist is done per fixture, so we have a 1:1 mapping for insertions to selections
- instead, we can load all models beforehand and check against that list
- this removes about 17k DB queries for MySQL tests

---

### <samp>🤖 Generated by Copilot at 755a5c7</samp>

Optimize fixture loading by reducing database queries. Refactor `ensureDefaultSettings` in `FixtureManager.js` to fetch and match models in memory.

### DIFF
--- a/ghost/core/core/server/data/schema/fixtures/FixtureManager.js
+++ b/ghost/core/core/server/data/schema/fixtures/FixtureManager.js
@@ -226,6 +226,8 @@ class FixtureManager {
             });
         }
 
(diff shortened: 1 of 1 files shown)


daniellockyer:
Closing for now as other targets produce better results 😀

PR #16963, opened 2023-06-07
Stripe enabled event

aileen:
### <samp>🤖 Generated by Copilot at 08fc49b</samp>

This pull request adds domain events to the `StripeService` module to communicate the status of the Stripe integration with live mode. It also updates the `SettingsBREADService` module to use the `StripeService` module to connect to Stripe when the Stripe public key setting is changed. It introduces two new files, `StripeLiveEnabledEvent.js` and `StripeLiveDisabledEvent.js`, that define the custom domain events for Stripe.


ghost/core/core/server/services/settings/SettingsBREADService.js
```diff
@@ -229,6 +229,20 @@ class SettingsBREADService {
...
 
+            // Detect if Stripe is now connected in live mode
+            const stripePublicKeySetting = result.find(setting => setting.get('key') === 'stripe_connect_publishable_key');
+            if (stripePublicKeySetting) {
+                const previous = stripePublicKeySetting.previousAttributes().value;
+                const current = stripePublicKeySetting.get('value');
+
+                if (current?.match(/pk_live/) && (!previous?.match(/pk_live/) || !previous)) {
+                    // Require the Stripe service here as it breaks existing tests otherwise
+                    const stripeService = require('../stripe');
```
aileen, line 240: @allouis requiring the stripe service here, fixes the previously failing test, but I don't think this should be the right approach. 

I was looking into calling `stripeService.connect()` directly from the `settings` API endpoint (just like we do for `disconnect()`, but I don't receive the previous state and want to avoid sending this event every time when the stripe settings are edited 🤔

allouis: @aileen Can we inject the stripe service into the constructor rather than require it here?

allouis (commented):
Can we put the check inside of the `if (stripeConnectData) {` block on line 190? Would that simplify the checks at all?

aileen:
> Can we put the check inside of the `if (stripeConnectData) {` block on line 190? Would that simplify the checks at all?

Yeah, that's actually a good point. I chose to put it after we have the model update response to be able and compare the before and after, but I don't think it's really necessary, so I moved it to the logic block on line 190 and simplified it.

allouis (approved):

PR #16966, opened 2023-06-07
Updated Italian translations

Zabriskije:
Added translation for new strings + some corrections for the old ones.


daniellockyer:
Thanks @Zabriskije! 🕺🏻

PR #16981, opened 2023-06-10
Updated Bahasa Indonesia translations

rkafp:
The translations for all new strings have been added.


daniellockyer:
Thanks @rkafp! 🙂

Issue #15007, opened 2022-07-07
Cannot get reading time from API via ?fields param


ErisDS:
### Issue Summary

When making a request to the Content API for posts and including `reading_time` as a field, it is not returned.

### Steps to Reproduce

Request the demo site with `reading_time`:

https://demo.ghost.io/ghost/api/canary/content/posts/?key=22444f78447824223cefc48062&fields=title,url,reading_time

Note that it does work if you include the `html` field, which reading time is calculated based on.

https://demo.ghost.io/ghost/api/canary/content/posts/?key=22444f78447824223cefc48062&fields=title,url,reading_time,html

### Ghost Version

5.2

### Node.js Version

14

### How did you install Ghost?

ghost cli

### Database type

MySQL 8

### Browser & OS version

N/a

### Relevant log / error output

```shell
n/a
```

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

raulmar0:
I have the exact problem. I've seen that when you request some specific fields you need to try with slightly different names than the ones marked on the fields from the GET/posts response. For example, in the GET/posts response it is marked "tags" in plural but when you add the "&fields" to the request you need to put "tag" in singular. Although I haven't found the correct name for reading_time.

devvrat011:
Sir plss assign this to me. I can fix this bug

as-ajitsingh:
Field `reading_time` doesn't work if field `html` is not present. You need to add field `html` too to get it.

PR #16983, opened 2023-06-11
Update collections via post api

allouis:
This PR adds support for adding and removing a post to/from a collection, via the Posts Admin API.

There are also some miscellaneous fixes, for the `god_mode` param and improving the e2e-api tests for posts.

Big changes:
- Post format serialization has been pulled out of `toJSON` and into serializer
  - `reading_time` is fixed, included by default and no longer relies on `html` format 
- `PostsService#editPost` now returns a DTO
- `PostsService#editPost` accepts an `eventHandler` to pass back change information
- posts endpoint/controller uses `eventHandler` to set cache invalidation header


ghost/posts-service/lib/PostsService.js
```diff
@@ -61,6 +62,46 @@ class PostsService {
             }
         }
 
+        if (this.isSet('collections') && frame.data.posts[0].collections) {
```
naz, line 65: Yup! This is pretty much the the way to do it imo. It's super verbose, but whatever. My understanding is that we are allowing to use an array of string ids or an array of objects with ids here. It would follow the same pattern we have for tags/labels, but for those resources we also allow assigning by "slug". I'm a little on the fence if we should mimic existing pattern 1:1 OR support even less syntax options at least in early stages. Not a big deal which way we go with this or stay with current approach. What is your thinking on this topic?

ghost/posts-service/lib/PostsService.js
```diff
@@ -61,6 +73,50 @@ class PostsService {
...
+                if (!collectionId) {
+                    continue;
+                }
+                const existingCollection = existingCollections.find(c => c.id === collectionId);
+                if (existingCollection) {
+                    continue;
+                }
+                const found = await this.collectionsService.getById(collectionId);
+                if (!found) {
+                    // TODO: Create collection
```
naz, line 95: Creating a collection would be a similar logic to what we do for "tags" when we pass in nonexistent ones to the post. In case of collections I don't think we should be having a create logic here, at least for now. I'd change this to "Create collection?" - as we don't know if this is what we want to have here as a feature 😅

ghost/core/core/server/api/endpoints/posts.js
```diff
@@ -21,6 +22,22 @@ const unsafeAttrs = ['status', 'authors', 'visibility'];
 
 const postsService = getPostServiceInstance();
 
+/**
+ * @param {string} event
+ */
+function getCacheHeaderFromEventString(event, dto) {
```
naz, line 28: Yeah, sick!

ghost/core/test/e2e-api/admin/utils.js
```diff
@@ -89,7 +89,8 @@ const expectedProperties = {
         'tiers',
         'newsletter',
         'count',
-        'post_revisions'
+        'post_revisions',
+        'reading_time'
```
naz, line 93: Does this also fix https://github.com/TryGhost/Ghost/issues/15007?

PR #16990, opened 2023-06-12
Add endpoint to record mail events

mike182uk:
refs https://github.com/TryGhost/Team/issues/3319


ghost/api-framework/lib/validators/input/all.js
```diff
@@ -126,7 +126,7 @@ module.exports = {
 
         // NOTE: this block should be removed completely once JSON Schema validations
         //       are introduced for all of the endpoints
-        if (!['posts', 'tags'].includes(apiConfig.docName)) {
+        if (!['posts', 'tags', 'mail_events'].includes(apiConfig.docName)) {
```
naz, line 129: Why didn't we go with the path that the comment above suggests - adding a "JSON Schema validation" for `mail_events`?

naz: Or rather "we should make it super obvious and easy to add validation schema" instead of having the hacky option as the only easy way to continue on with the work! 😂

ghost/core/test/e2e-api/admin/mail-events.test.js
```diff
@@ -0,0 +1,44 @@
...
+            ]
+        };
+
+        await agent
+            .post('/mail_events/', {
+                headers: {
+                    'content-type': 'application/json'
+                }
+            })
+            .body(payload)
```
naz, line 41: When using [e2e framework's assertions](https://www.notion.so/ghost/End-to-end-Testing-6a2ef073b1754b18aff42e24a632a007?pvs=4#b90c6286099849b493f6c7fbc11edd8e) it's important to add `matchBodySnapshot` and `matchHeaderSnapshot` in base tests like this one so we can have snapshots to track the response payload evolution over time. And catch bugs if the payload changes unintentionally.

ghost/mail-events/src/MailEventRepository.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+interface Model {
+    add(attributes: ModelAttributes): Model;
+}
+
+export default class MailEventRepository {
+    constructor(private model: Model) {}
+
+    async persist(event: MailEvent) {
```
naz, line 18: The convention we've developed over time is calling the method that saves the data "save", like [this example](https://github.com/TryGhost/Ghost/blob/005a488eedecb1f9502811e59a5e2f6e70135c8b/ghost/webmentions/lib/InMemoryMentionRepository.js#L40). We should try keeping the naming conssitent so in the future we could have for example a unified "IRepository" interface and in general to have consistent developer experience across the codebase.

ghost/core/test/e2e-api/admin/mail-events.test.js
mike182uk: 👍 Have added these assertions in

allouis (commented):
All looks good - I left one comment about the in-memory-repository, but there's no need to change it 👍 

I think we're just missing an update to `.github/dev.js` ala https://github.com/TryGhost/Ghost/commit/eee5eed6e234bd5d52de978e28d90b370887ab26

ghost/mail-events/src/InMemoryMailEventRepository.ts
```diff
@@ -0,0 +1,12 @@
+import {MailEvent} from './MailEvent';
+import {MailEventRepository} from './MailEventRepository';
+
+export class InMemoryMailEventRepository implements MailEventRepository {
```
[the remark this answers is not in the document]
mike182uk: Updated in https://github.com/TryGhost/Ghost/pull/16990/commits/512dc7dc163035f50f825f34074e0671cdc9d51d

ghost/core/core/server/data/migrations/versions/5.52/2023-06-13-12-24-add-temp-mail-events-table.js
daniellockyer, line 1: This now needs updating for 5.53

Issue #15725, opened 2022-10-30
Mailgun BATCH_SIZE with a fix value of 1000 is incompatible with new users


ltoinel:
### Issue Summary

Hi,

When you want to use Mailgun with the flex offer, they block your account to a maximum BATCH_SIZE to 10.
This value is currently to defined to 1000 and cannot be overide with settings by the end user.

in bulk-email-processor.js : 
`
module.exports = {
    BATCH_SIZE: MailgunClient.BATCH_SIZE,
`

Could be  : 
`
module.exports = {
    BATCH_SIZE: bulkEmailConfig?.mailgun?.batch_size,
`

A new setting to allow the users to set a BATCH_SIZE to 10 could be great to allow Mailgun to evaluate the mail quality before to increase to 1000 the batch size.

Thank-you ! 

### Steps to Reproduce

1. Create a ghost instance.
2. Create a Mailgun account.
3. Add 11 members to the newsletter
4. Send a newsletter
5. Mailgun will not accept more that 10 members in the recipient.

### Ghost Version

v5.19.0

### Node.js Version

v16.17.0

### How did you install Ghost?

Ghost CLI + Ghost Update

### Database type

MySQL 8

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

dest4:
The new setting is probably the best idea.

Other, probably worse, approaches:
- start with batch size of 1000 but if Mailgun account limitation is detected, retry with 10 (no config needed, but more complex to implement?).
- set the batch size to 10 by default for all users (trivial to implement, but slow delivery for big publishers?)

ltoinel:
@dest4  : Overwise use a dynamic batch size regarding your members count ? This allows you to start your blog with the mailgun restrictions.

`if (nb_members < 100) 
     batch_size =10;
else {
      batch_size =1000;
}`

ErisDS:
We are actively talking to mailgun about how to make the onboarding process work more smoothly. I've raised this to them as, as far as I know, there is no way for us to auto-detect when the account is limited.

dest4:
Knowing before trying that the account is limited is probably not possible. But once the call to Mailgun API fails, it seems to be possible to infer it.

Here are a couple of lines of log when sending fails due to account limitation

```
[2022-10-31 19:38:29] WARN The email service received an error from mailgun and was unable to send.

The email service received an error from mailgun and was unable to send.

[...]

PR #16993, opened 2023-06-12
feat: configurable mailgun bulk email batch size #15725

jakub-k-slys:

---

### <samp>🤖 Generated by Copilot at 3a2f539</samp>

This pull request adds a new setting `mailgunBatchSize` to the Ghost application, which allows the user to configure the batch size for sending emails via the mailgun integration. It also updates the email service, the mailgun client, and the email analytics provider to use the new setting value. Additionally, it adds tests, migrations, and fixtures to support the new setting.

### DIFF
--- a/ghost/admin/app/components/editor/modals/preview/email.js
+++ b/ghost/admin/app/components/editor/modals/preview/email.js
@@ -58,7 +58,7 @@ export default class ModalPostPreviewEmailComponent extends Component {
 
     get mailgunIsEnabled() {
         return this.config.mailgunIsConfigured ||
-            !!(this.settings.mailgunApiKey && this.settings.mailgunDomain && this.settings.mailgunBaseUrl);
+            !!(this.settings.mailgunApiKey && this.settings.mailgunDomain && this.settings.mailgunBaseUrl && this.settings.mailgunBatchSize);
     }
 
     get paidMembersEnabled() {
--- a/ghost/admin/app/components/settings/newsletters.hbs
+++ b/ghost/admin/app/components/settings/newsletters.hbs
@@ -115,6 +115,16 @@
                                                 data-test-mailgun-domain-input
                                             />
                                         </GhFormGroup>
+                                        <GhFormGroup class="no-margin">
+                                            <label class="fw6 f8" for="mailgun-batch-size">Mailgun batch size</label>
+                                            <input
+                                                id="mailgun-batch-size"
+                                                type="text"
+                                                class="gh-input mt1"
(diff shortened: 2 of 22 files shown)


jakub-k-slys:
Hi @SimonBackx, will you be able to review the PR and provide feedback?
CC @ltoinel @alias-rahil @KarimJedda

SimonBackx:
Hey @jakub-k-slys, thank you for your contribution! 😊 Is there any reason why you would like to change the batch size? AFAIK, Mailgun only supports up to 1.000 recipients at the same time, and I don't see a reason to reduce this number, or do you have any?

jakub-k-slys:
Hi @SimonBackx. Many thanks for getting back! When it comes to Mailgun, it offers plenty of subscription plans that current ghost codebase works well with. However, there is also Flex plan that allows to send up to 1000 emails per month without being charged and you pay for everything above 1000. The major issue with this plan is Mailgun limits the number of recipients in batches. It is limited to 9 :( and that's something I was hit by (and other fellows). I am hoping that reducing the number of recipients in batches would allow me to send newsletters again... Therefore I'd like to introduce an ability to adjust how big the batch can be. Does it make sense to you? Looking forward to hear you!

SimonBackx:
Hey @jakub-k-slys! I definitely think you should check with Mailgun support to get your limit increased. This has been raised a couple of times on the forum, and it is a limit Mailgun sometimes enforces on new accounts to prevent spam: https://forum.ghost.org/t/too-many-recipients-error-while-sending-email-newsletter-using-mailgun/10716/12

Apart from that, given the amount of users with these issues, and to make it easier to build a sending history, we could allow a similar change in Ghost, but the solution you've implemented is a little bit too complex, given the very limited use case. So if you want to make the following changes, we can get this merged:

- Switch from using a setting to [a configuration value](https://ghost.org/docs/config/). A configuration value is defined in a config file, and doesn't change after booting. I don't think we need complicated UI to change this value, since this is already a very technical thing to setup, so it shouldn't be visible for normal users (that will only complicate things during setup).
- Remove all UI changes

You can add the config under `ghost/core/core/shared/config/defaults.json`, under a new `bulkEmail` key and `batchSize` value.

You can alter `ghost/mailgun-client/lib/MailgunClient.js` to read the new config value, and replace the static variable with a getter method. You can then use the same getter in `ghost/email-service/lib/MailgunEmailProvider.js` `getMaximumRecipients`

jakub-k-slys:
@SimonBackx what you proposing sounds really great. I will try to rework the PR in my spare time according to your suggestions and will back to you. Many thanks for help! 😃

PR #17025, opened 2023-06-14
🎨 Fixed a linting error in main branch

vershwal:
I found a linting error while running `yarn lint` on the main branch.  Found this while testing one of my own PRs. 

Error:-
image
Looks like it got introduced here [917ab0a](https://github.com/TryGhost/Ghost/commit/917ab0a79e788faf5df30a89503b510e43d4dd28)

Reason:-
We are using `'plugin:ghost/test'` in the .eslintrc.js file which internally verifies that file names must follow this regex:-
`'ghost/filenames/match-regex': ['error', '^[a-z0-9-.]+$', null, true]`

Ref.: https://www.npmjs.com/package/eslint-plugin-ghost


vershwal:
After analyzing the Test Suite / Lint (pull_request) here on GitHub for this PR I figured that the error was specific to my local setup. Updating the dependencies fixed it for me locally as well. We don't need these changes. 
Thanks!!

vershwal:
After analyzing the Test Suite / Lint (pull_request) here on GitHub for this PR I figured that the error was specific to my local setup. Updating the dependencies fixed it for me locally as well. We don't need these changes.
Thanks!!

PR #17028, opened 2023-06-15
Tier One-TIme Pricing (In Progress) 

rac146:
While Monthly and Yearly subscriptions exist in Ghost, there is currently no option for one-time pricing (i.e. pay once and have unlimited access to a specific tier with no end date).

I wanted to add it for a custom site I'm working on. This is currently rough happy-path code. Most definitely not ready yet, but wanted to get attention of the team. Based on the code it looks like this has been considered/discussed, but never implemented.

The biggest challenge so far is retreating away from the stripe subscription object that comes back for yearly/monthly subscriptions. A one-time transaction will not have this information, so relevant information has to be retrieved from elsewhere.

If the Ghost team is interested in this, please let me know and i'll do what's required for a full PR.


rac146:
will re-open later

PR #17032, opened 2023-06-15
Removed Bluebird Promise.each from db.utils

vershwal:
https://github.com/TryGhost/Ghost/issues/14882
I have used the in-house sequence function for refactoring Bluebird's Promise.each
Taken some reference from here: https://github.com/TryGhost/Ghost/pull/15626


daniellockyer:
Hey @vershwal - would you be able to rebase and remove the Bluebird `Promise` import at the top? 🙂

PR #17043, opened 2023-06-16
🐛 ✨ Added filter for specific tier in admin panel

### LINKED ISSUE #16988: Admin panel filters for posts don't display posts for tiers.
### Issue Summary

An issue in two parts:
1) A bug.  If a post is set for specific tiers access, the admin panel does not show it when any of the available dropdown options that should apply (members only, paid members only) are selected.  The post only shows up when 'all access' is selected.

2) Perhaps a feature request.  The Ghost admin panel doesn't have a dropdown to allow me to see posts by specific tiers.  That'd be a logical thing to include.
image

### Steps to Reproduce

Create a paid tier 
Assign some posts to it.  (Restrict access to 'specific tier')  
Navigate to the posts page of the admin portal.
Select paid members only or members only.  The tier-restricted posts do not appear.  

I would expect tier-restricted posts to appear with these filters applied.
I think the expected UI here would also include specific tiers as an option in that dropdown.

### Ghost Version

5.50

### Node.js Version

Ghost Pro

Ghost Pro

### Database type

MySQL 8

### Browser & OS version

Chrome, Windows 10

_No response_

### Code of Conduct

vershwal:
fixes https://github.com/TryGhost/Ghost/issues/16988
I have added a filter for specific tier(s) in the admin panel. 
The part to fetch posts based on this filter was already taken care of in the API `/ghost/api/admin/posts/` with query param `filter=visibility%3Atiers`

Results after this change:-

image

image

image

image

@daniellockyer @cathysarisky Hope this helps. Let me know if anything else is needed here.


vershwal:
@daniellockyer could you please give your comments on this PR as well.

55sketch:
@vershwal I think that adding "Specific Tiers" to the dropdown creates a bit of UI confusion, as I would expect as a user to be able to actually specify which tiers once I click on this (so, a more granular filter).

I think the approach would be to make sure that posts set to "Specific Tier" are included in the filtered response for "Paid" - as far as I can tell the Specific Tiers option on posts never includes the free tier, so this catchall should cover all use cases.

vershwal:
@55sketch I think these filters are also related to how we select the access while publishing the post:-

image


While selecting the access, The two categories are different `Paid-members only`  and `Specific tier(s)` I don't think it will be fair to show both of these under `Paid-members only` filter. This will also create confusion. What do you think?

vershwal:
Although here we are not providing a more granular filter, we are at least listing all the posts that have access for `Specific tier(s)`(That the user selected while publishing the post)

55sketch:
Yes that's correct, but when you select "specific tiers" when writing a post you then choose _which_ tiers that the post is visible to. Using the same terminology in the filter UI would suggest the same flow - I select and filter by the posts matching _some array_ of specific tiers that I choose.

The specific tiers selection on the post creation page will _always_ show **paid-only** tiers. It does not show free. So including these posts in the **Paid** filter on the posts list is correct behaviour, that fact that they don't currently is a bug.

We should resolve that even if down the line we may want to explore having a more granular filter for specific tiers.

Does this all make sense?

vershwal:
Okay!! Ya, that make sense. I'll change this to show the specific tiers post under the `Paid-members only` filter.

vershwal:
Doest this look good @55sketch?

55sketch:
Yep - that's what I would have done too! :D 

Can we add a test to ensure that both types of posts are displayed when the filter is set?

vershwal:
Modified the acceptance tests to check that the correct visibility is passed in the API when `Paid members-only` filter is selected.

55sketch (approved):
LGTM!

PR #17047, opened 2023-06-16
Updated to use assert/strict everywhere

ErisDS:
refs: https://github.com/TryGhost/DevOps/issues/8

- Our eslint plugin has been updated with new rules around the node assert library
- The first rule being enabled enforces that we use assert/strict instead of just assert
- This means we don't need to use the strict version of methods, as the standard version will work that way by default
- This change preempts the update of the package, so that it will be possible to merge the changes

---

### <samp>🤖 Generated by Copilot at 25d924a</samp>

Refactor various test files to use non-deprecated `assert` methods. This improves the code quality and avoids unnecessary warnings. The files affected are mostly in the `ghost/core/test/e2e-api` and `ghost/adapter-cache-*` directories, as well as some other test files in different modules. The main change is to use `assert.equal` instead of `assert.strictEqual` where appropriate.


ghost/core/test/integration/services/email-service/email-event-storage.test.js
```diff
@@ -462,7 +462,7 @@ describe('EmailEventStorage', function () {
         // Message and code not changed
         assert.equal(permanentFailures.models[0].get('message'), 'Not delivering to previously bounced address');
         assert.equal(permanentFailures.models[0].get('code'), 605);
-        assert.equal(permanentFailures.models[0].get('enhanded_code'), null);
+        assert.equal(permanentFailures.models[0].get('enhanded_code'), undefined);
```
daniellockyer, line 465: Should this actually be `enhanced_code` anyway?

ErisDS: Yes, updated

PR #17048, opened 2023-06-16
🐛 Fixed post scheduling for sites with transferred ownership

cmraible:
refs TryGhost/Team#3494

- By default, the post scheduler runs as user_id = 1, which is the original owner of the site
- It is possible to transfer ownership to another user, then delete the original owner, so some sites don't have a user_id = 1
- With the addition of post history, we save the author_id from the contextUser, which is usually the logged in user
- However when the scheduler runs and tries to publish a post, it sets the author_id to 1, which then fails the foreign key constraint on the post_revisions table
- This commit fixes the issue by checking if the contextUser exists, and falling back to the author of the most recent post_revision if it doesn't
- This is a bit of a hack/band-aid to fix the immediate issue, but ideally this check would be made in the contextUser function, so we can always be confident the contextUser exists without firing an ad-hoc DB query
- Additionally, we should refactor to query for the current owner's user_id instead of defaulting to 1


ghost/core/core/server/models/post.js
```diff
@@ -875,7 +876,8 @@ Post = ghostBookshelf.Model.extend({
                     revision_interval_ms: POST_REVISIONS_INTERVAL_MS
                 }
             });
-            const authorId = this.contextUser(options);
+            let authorId = this.contextUser(options);
+            const authorExists = User.findOne({id: authorId});
```
daniellockyer, line 880: I think we're missing an await here

allouis (commented):
I think using the previous revisions author is weird because it's going to provide different behaviour dependent on the state of the site 

What about updating the `this.contextUser` method to:
1. Return a Promise
2. Return the Owner User id rather than `1`

🤔

cmraible:
> I think using the previous revisions author is weird because it's going to provide different behaviour dependent on the state of the site
> 
> What about updating the `this.contextUser` method to:
> 
> 1. Return a Promise
> 2. Return the Owner User id rather than `1`
> 
> 🤔

Agree with you that this would be a better solution, but I'm struggling to implement it properly...I just pushed what I have right now to this branch — this should work in theory, but it breaks a bunch of tests. I think the pattern of relying on '1' being the owner user is still pretty entrenched in the codebase, especially in the test framework. 

I'm having a hard time untangling the circular dependency on the owner user — e.g. when we create the owner user in the fixture-utils, we rely on the owner user already existing https://github.com/TryGhost/Ghost/blob/82acf85b297279639a3c14a7b215d8e6d9c0ea2c/ghost/core/test/utils/fixture-utils.js#L225

Let me know your thoughts, struggling with this one a bit.

PR #17057, opened 2023-06-19
Moved collections initialisation behind labs flag

allouis:
We ran into an issue where the large `published:true` query was affecting the boot time of large sites which had knock-on effects with availability.

---

### <samp>🤖 Generated by Copilot at fda807e</samp>

This pull request simplifies the collections service by removing the `built-in-collections.js` module and using feature flags to dynamically create default collections in `collections/index.js`. This allows for more customization and consistency of collections.


naz (approved):
lesgoooo

PR #17073, opened 2023-06-20
Removed Bluebird `Promise.map` from apps index.js file

vershwal:
Refactored code for https://github.com/TryGhost/Ghost/issues/14882
Removed Bluebird Promise.map from index.js file


daniellockyer:
Thanks again @vershwal 🙂

PR #17103, opened 2023-06-22
🐛 Fixed newsletter post duplication

mike182uk:
closes https://github.com/TryGhost/Team/issues/3521

When duplicating a newsletter (email only post) some of the fields relating to the newsletter (i.e `newsletter_id`) were being erroneously copied over. Upon publishing this made the post appear as if it was an email only post that had been sent to subscribers when it actually had not

### DIFF
--- a/ghost/posts-service/lib/PostsService.js
+++ b/ghost/posts-service/lib/PostsService.js
@@ -526,7 +526,8 @@ class PostsService {
                 'published_at',
                 'published_by',
                 'canonical_url',
-                'count__clicks'
+                'count__clicks',
+                'newsletter_id'
             ]
         );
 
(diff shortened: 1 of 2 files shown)


ghost/posts-service/lib/PostsService.js
```diff
@@ -526,7 +526,8 @@ class PostsService {
                 'published_at',
                 'published_by',
                 'canonical_url',
-                'count__clicks'
+                'count__clicks',
+                'newsletter_id'
```
naz, line 530: When we add another field to the post would it automatically be included in the copied version? Maybe worth reconsidering allowlist vs blocklist approach?

daniellockyer: 💯 - the pattern of remembering to omit fields vs allowing them when needed just leads to a world of pain

mike182uk: Alrighty then, i swapped it out to explicitly define the fields to be copied over

PR #17108, opened 2023-06-23
Added tables to persist Collections

allouis:
refs https://github.com/TryGhost/Team/issues/3427

### DIFF
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -1028,5 +1028,22 @@ module.exports = {
         message_id: {type: 'string', maxlength: 150, nullable: false},
         recipient: {type: 'string', maxlength: 191, nullable: false},
         occurred_at: {type: 'dateTime', nullable: false}
+    },
+    collections: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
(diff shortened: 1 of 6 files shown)


ghost/core/core/server/data/schema/schema.js
```diff
@@ -1021,5 +1021,22 @@ module.exports = {
         currency: {type: 'string', maxlength: 24, nullable: true},
         created_at: {type: 'dateTime', nullable: false},
         email_sent_at: {type: 'dateTime', nullable: true}
+    },
+    collections: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        title: {type: 'string', maxlength: 191, nullable: false},
+        slug: {type: 'string', maxlength: 191, nullable: false, unique: true},
+        description: {type: 'string', maxlength: 2000, nullable: true},
+        type: {type: 'string', maxlength: 50, nullable: false},
```
naz, line 1030: We'd like to have this validated straight away imo. The two allowed variables right now would be `automatic` and `manual`

allouis: Validation is in the Entity rather than the bookshelf model - I don't think there's much point duplicating that?

daniellockyer (approved):

PR #17111, opened 2023-06-23
Removed bluebird `promise.map` from db.js file

vershwal:
Refactored code for https://github.com/TryGhost/Ghost/issues/14882
Removed bluebird `promise.map` from db.js file


daniellockyer:
@vershwal This ends up changing the functionality of the code because we lose the concurrency. Instead, you can use https://github.com/TryGhost/framework/blob/main/packages/promise/lib/pool.js 🙂

vershwal:
@daniellockyer I have done the changes. Hope it is fine now!

daniellockyer:
Hey @vershwal - unfortunately a different contributor got to it first, so this is already changed on `main`. Thanks for the PR anyway 🙂

PR #17137, opened 2023-06-27
Fixed members-only content incorrectly showing in plaintext email

### LINKED ISSUE #16131: Ghost Newsletter email includes ALL text when a paid-members newsletter is sent to free members, in plaintext only
### Issue Summary

When an email is sent to all members with the access set to 'paid members only', the expected behavior is that free members would see only the preview text, followed by a CTA to subscribe.  That's exactly what happens for HTML email.  However, plain text email reveals the FULL text.  See images here: https://cathy.sarisky.link/newsletters-public-previews-and-more/ (post is not finished, but I think the images are revealing!)

### Steps to Reproduce

1- Create a post, add a public preview, put text above and below it.  Set it to paid members only.
2. Email the newsletter to both free and paid members.

Check the PLAIN TEXT of what free users receive.  (In Gmail, 'show original'.)

### Ghost Version

5.30.0

### Node.js Version

Ghost Pro.

Ghost Pro.

### Database type

MySQL 8

### Browser & OS version

Ghost Pro / tested with a Gmail and Outlook addresses

_No response_

### Code of Conduct

mike182uk:
fixes https://github.com/TryGhost/Ghost/issues/16131

Members only content was incorrectly being shown in a plaintext email due to the email `preheader` using the post model `plaintext` field directly (which contained the members-only content). This changes this behaviour so that the post html content is utilised for the `preheader` but has all members-only content (post-preview content + segmented content) removed

### DIFF
--- a/ghost/email-service/lib/EmailRenderer.js
+++ b/ghost/email-service/lib/EmailRenderer.js
@@ -278,24 +278,33 @@ class EmailRenderer {
 
                 // Remove the members-only content
                 html = html.slice(0, membersOnlyIndex);
-
-                const $ = cheerio.load(html);
-
-                $('[data-gh-segment]').get().forEach((node) => {
-                    if (node.attribs['data-gh-segment'] !== 'status:free') {
-                        $(node).remove();
-                    }
-                });
-
-                html = $.html();
             }
         }
+
+        let $ = cheerio.load(html);
+
+        // Remove parts of the HTML not applicable to the current segment - We do this
+        // before rendering the template as the preheader for the email may be generated
+        // using the HTML and we don't want to include content that should not be
+        // visible depending on the segment
+        $('[data-gh-segment]').get().forEach((node) => {
+            // TODO: replace with NQL interpretation
+            if (node.attribs['data-gh-segment'] !== segment) {
+                $(node).remove();
+            } else {
+                // Getting rid of the attribute for a cleaner html output
+                $(node).removeAttr('data-gh-segment');
+            }
+        });
+
+        html = $.html();
 
         const templateData = await this.getTemplateData({
             post,
             newsletter,
             html,
-            addPaywall
+            addPaywall,
+            segment
         });
         html = await this.renderTemplate(templateData);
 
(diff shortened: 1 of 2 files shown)


ghost/email-service/lib/EmailRenderer.js
```diff
@@ -278,6 +278,16 @@ class EmailRenderer {
...
+
+                const $ = cheerio.load(html);
+
+                $('[data-gh-segment]').get().forEach((node) => {
+                    if (node.attribs['data-gh-segment'] !== 'status:free') {
+                        $(node).remove();
+                    }
+                });
+
+                html = $.html();
```
SimonBackx, line 290: Can't we move the part from a little bit down here to above `getTemplateData` instead of repeating it twice? So that it happens for all segments instead of only free members?

\+ maybe add a comment for future reference why we need to remove the segments before we call `getTemplateData`, so that the preheader doesn't contain paid text

ghost/email-service/test/email-renderer.test.js
```diff
@@ -1057,6 +1057,40 @@ describe('Email renderer', function () {
             should($('.preheader').text()).eql('Custom excerpt');
         });
 
+        it('does not include member-only content in preheader', async function () {
+            renderedPost = '<div> Lexical Test </div> <div data-gh-segment="status:-free"> members only section</div> some text for both <!--members-only--> finishing part only for members';
```
SimonBackx, line 1061: We need to have the same test with paid member only content, but without the paywall `<!--members-only-->`, because in that case it should also hide the `<div data-gh-segment="status:-free">` part from the preheader

ghost/email-service/lib/EmailRenderer.js
```diff
@@ -715,13 +724,18 @@ class EmailRenderer {
...
         let customExcerpt = postModel.get('custom_excerpt');
         if (customExcerpt) {
             return customExcerpt;
         } else {
             if (plaintext) {
+                // If paywall is enabled, we need to get the plaintext from the html
+                // as the plaintext field on the model may contain paywalled content
+                if (addPaywall) {
+                    plaintext = htmlToPlaintext.email(html);
+                }
```
SimonBackx, line 738: I think we need to check for segment == `status:free` instead of whether there is a paywall or not, because it is also possible to have content that is only visible for paid members when there is no paywall card in the post.

ghost/email-service/test/email-renderer.test.js
SimonBackx: (you'll probably add some more tests to to get the coverage back to 100%)

ghost/email-service/lib/EmailRenderer.js
mike182uk: 👍 Updated in https://github.com/TryGhost/Ghost/pull/17137/commits/553fadb1f31277b5f5f92d1318cb9c6663707c4b

ghost/email-service/test/email-renderer.test.js
mike182uk: 👍 Updated in https://github.com/TryGhost/Ghost/pull/17137/commits/553fadb1f31277b5f5f92d1318cb9c6663707c4b

Heh the coverage had fallen due to the test modifying the global `renderedPost` var meaning subsequent tests were using this modified version which in turn meant a specific code path never got executed 😳

ghost/email-service/lib/EmailRenderer.js
mike182uk: 👍 Updated in https://github.com/TryGhost/Ghost/pull/17137/commits/553fadb1f31277b5f5f92d1318cb9c6663707c4b

SimonBackx (approved):

PR #17142, opened 2023-06-27
Fixed leaking pivot fields

mike182uk:
fixes https://github.com/TryGhost/Team/issues/2657

The `omitPivot` option does not have an affect on a models `_previousAttributes`. When we serialise a model and want to retrieve the previous attributes we need to ensure we manually remove the pivot fields

See https://github.com/bookshelf/bookshelf/blob/7704fbc5e83a35357d89458d9269a7e739964296/lib/base/model.js#L512

### DIFF
--- a/ghost/core/core/server/models/base/plugins/overrides.js
+++ b/ghost/core/core/server/models/base/plugins/overrides.js
@@ -88,7 +88,10 @@ module.exports = function (Bookshelf) {
             // CASE: get JSON of previous attrs
             if (options.previous) {
                 const clonedModel = _.cloneDeep(this);
-                clonedModel.attributes = this._previousAttributes;
+                clonedModel.attributes = _.omitBy(
+                    this._previousAttributes,
+                    (value, key) => key.startsWith('_pivot_')
+                );
 
                 if (this.relationships) {
                     this.relationships.forEach((relation) => {
--- a/ghost/core/core/server/models/base/plugins/overrides.js
+++ b/ghost/core/core/server/models/base/plugins/overrides.js
@@ -1,3 +1,4 @@
+const {PIVOT_PREFIX} = require('bookshelf/lib/constants');
 const _ = require('lodash');
 
 /**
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/models/base/plugins/overrides.js
```diff
@@ -88,7 +88,10 @@ module.exports = function (Bookshelf) {
             // CASE: get JSON of previous attrs
             if (options.previous) {
                 const clonedModel = _.cloneDeep(this);
-                clonedModel.attributes = this._previousAttributes;
+                clonedModel.attributes = _.omitBy(
+                    this._previousAttributes,
+                    (value, key) => key.startsWith('_pivot_')
```
naz, line 93: Would be best to use the constant straight from the bookshelf instead of introducing a magic string here - https://github.com/bookshelf/bookshelf/blob/7704fbc5e83a35357d89458d9269a7e739964296/lib/constants.js#L1C1-L1C1. If it's not exposed, so be it 😆  - leaving a comment for thwere the matic string comes from would help with future maintenance 👍

naz (commented):
Other than small maintainability improvement this is gooood 👌  Did not expect such an outcome for the issue tbh. This was a total surprise. I would have thought this kind of functionality would be baked into Bookshelf during the serialize execution 🤔

ghost/core/core/server/models/base/plugins/overrides.js
mike182uk: Its not exposed directly, but is accessible. What do you think of [this](https://github.com/TryGhost/Ghost/pull/17142/commits/3258050aef87b5e86348137057917a0bdad1c014)? Although we are accessing something that is technically not exported, we still have the safety of the tests failing if something changes around this so could be a good trade-off over us having to introduce the magic string ourselves

naz: That's a perfect solution imo 👌

PR #17157, opened 2023-06-29
Wired up BookshelfCollectionsRepository to CollectionsService

allouis:
image

### DIFF
--- a/ghost/core/core/server/services/collections/service.js
+++ b/ghost/core/core/server/services/collections/service.js
@@ -1,7 +1,7 @@
 const {
-    CollectionsService,
-    CollectionsRepositoryInMemory
+    CollectionsService
 } = require('@tryghost/collections');
+const BookshelfCollectionsRepository = require('./BookshelfCollectionsRepository');
 const labs = require('../../../shared/labs');
 
 class CollectionsServiceWrapper {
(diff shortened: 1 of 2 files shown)


allouis:
I've done this separately to the bulk of the changes so that we can release them separately, with the thinking that we might not necessarily want to switch the DB backed storage immediately!

ghost/core/core/server/services/collections/service.js
```diff
@@ -1,25 +1,26 @@
...
     /** @type {CollectionsService} */
     api;
 
     constructor() {
-        const postsRepository = require('./PostsRepository').getInstance();
-        const collectionsRepositoryInMemory = new CollectionsRepositoryInMemory();
         const DomainEvents = require('@tryghost/domain-events');
+        const postsRepository = require('./PostsRepository').getInstance();
+        const models = require('../../models');
+        const collectionsRepositoryInMemory = new BookshelfCollectionsRepository(models.Collection);
```
naz, line 15: `bookshelfCollectionsRepository` for a variable name now? or just a `collectionsRepository` 😛

allouis: Ha yeah 😝

naz:
Gave it a lil spin. Works great! There's no good reason to hold off the merge for this as we are nearing an alpha version and should be working with close to real world state.

PR #17187, opened 2023-07-03
Added migration and schema change for `posts_meta.hide_title_and_feature_image`

kevinansfield:
closes https://github.com/TryGhost/Team/issues/3550

We want to allow an option to hide the title and feature image on a per-page basis, to do that we need somewhere to store the setting value. The existing `posts_meta` table is the simplest candidate, especially as this is a single setting and we don't have a desire to introduce many such settings.

- added migration that adds the `hide_title_and_feature_image` column to the `posts_meta` table with a `boolean` data type and a default value of `false` (matches behaviour of all existing pages)
- updated schema file for initial database creation
- removed property from API output via serializers to keep migration PR minimal

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/pages.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/pages.js
@@ -8,5 +8,8 @@ module.exports = async (model, frame, options) => {
     delete jsonModel.email_only;
     delete jsonModel.newsletter_id;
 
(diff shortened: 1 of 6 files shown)


kevinansfield:
Full API implementation on top of this data change can be seen in https://github.com/TryGhost/Ghost/pull/17211/commits/59d522322f75f3e1fd3832220228cd4394ab1d99

ronaldlangeveld (approved):

PR #17208, opened 2023-07-05
🐛 Fixed member import newsletter subscription prefs

mike182uk:
fixes https://github.com/TryGhost/Product/issues/2947

When an import is performed and a member already exists:
- If `subscribe_to_emails` is set to `true` and the existing member already has newsletter subscriptions, these are preserved
- If `subscribe_to_emails` is set to `true` and the existing member does not already have newsletter subscriptions, no new subscriptions are added in case the member has previously unsubscribed
- If `subscribe_to_emails` is set to `false` the existing member will be unsubscribed from all newsletters regardless of their preference


cmraible (approved):
LGTM — checked out and tested locally, works as expected. Nice work!

PR #17247, opened 2023-07-07
Added built in collection fixtures

naz:
refs https://github.com/TryGhost/Arch/issues/25

- The instance should have two built-in collections "index" and "featured". These have been filled through in-memory tricks before, now they should come pre-populated through fixtures mechanism.

---

### <samp>🤖 Generated by Copilot at cfdde0a</samp>

This pull request adds two built-in collections, index and featured, to the Ghost database schema and data. It uses a migration script and a fixtures file to create the collections and removes the redundant code from the collections service. The built-in collections allow users to easily access all posts or only featured posts.


ghost/core/core/server/data/migrations/versions/5.55/2023-07-07-14-15-32-add-built-in-collections.js
```diff
@@ -0,0 +1,60 @@
...
+                slug: 'index'
+            })
+            .first();
+
+        if (existingIndexCollection) {
+            logging.warn('Index collection already exists, skipping');
+        } else {
+            await knex('collections').insert({
+                id: (new ObjectID()).toHexString(),
+                name: 'Index',
```
ErisDS, line 20: Detail to polish: we should decide if this built-in should be called "Latest" rather than Index. I think given the aim to make this easy to understand for users - it probably should be.

naz: Agreed, matching the product naming makes sense 👍 If we market it and use it as "Latest" in cards, it would be pretty confusing seeing "Index" in collections UI once we have one and figuring out what the hell this thing does 😄

naz: One thing that will be confusing though is "latest" implies there's ordering applied to the posts in the collection, which we are not forcing explicitly yet. I do think we should make the default order of posts in collections more explicit though.

ErisDS (commented):
I don't think we need to worry about sort order rn

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
```diff
@@ -0,0 +1,118 @@
...
+
+const insertPostCollections = async (knex, collectionId, postIds) => {
+    logging.warn(`Batch inserting ${postIds.length} collection posts for collection ${collectionId}`);
+
+    const collectionPosts = postIds.map((postId, index) => {
+        return {
+            id: (new ObjectID()).toHexString(),
+            collection_id: collectionId,
+            post_id: postId,
+            sort_order: index
```
naz, line 14: Hardcode to `0` - we don't need it yet.

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
```diff
@@ -0,0 +1,118 @@
...
+            })
+            .first();
+
+        if (!existingLatestCollection) {
+            logging.warn('Latest collection does not exists, skipping');
+        } else {
+            // Default sort order same as for posts defaults
+            const latestPostsRows = await knex('posts')
+                .select('id')
+                .orderBy(DEFAULT_POST_ORDER);
```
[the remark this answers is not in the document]
naz: We should not be worrying about it. The whole concept of ordering should not have entered our codebase just yet as it only creates artificial problems like this one. The ordering will be 🔥 from this migration.

ghost/core/test/utils/fixtures/fixtures.json
```diff
@@ -1123,6 +1123,28 @@
...
+                "match": "slug",
+                "relation": "posts"
+            },
+            "to": {
+                "model": "Post",
+                "match": "slug"
+            },
+            "entries": {
+                "latest": {
+                    "integrations": "integrations",
```
naz, line 1138: This notation was pretty counter intuitive to figure out... Basically we have to have a duplicate "key" and "value" here for [matching to work](https://github.com/TryGhost/Ghost/blob/57a851227cad6e30c400636daf3bed517078de75/ghost/core/core/server/data/schema/fixtures/FixtureManager.js#L43). And the `key = key === 0 && value ? value : key;` in the matcher makes the matching logic even more confusing as it kinda works for the first record if passed in as an array, but than stops working when more elements are added.

naz: Reading the tests somewhat helps to understand the logic. But initially it was pretty confusing getting this right as the expected intuitive syntax like:
```js
"latest": ["integrations", "grow", "sell", ....]
```
didn't work out of the box.

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
```diff
@@ -0,0 +1,104 @@
...
+            id: (new ObjectID()).toHexString(),
+            collection_id: collectionId,
+            post_id: postId,
+            sort_order: 0
+        };
+    });
+
+    const batches = chunk(collectionPosts, 1000);
+
+    for (const batch of batches) { // eslint-disable-line no-restricted-syntax
```
naz, line 20: This part might cause performance issues. Needs additional checks on an instance with large amount of posts.

naz:
There are flaky tests coming from `ghost/core/test/e2e-api/admin/collections.test.js` - they fail 20% of a time when run on MySQL locally. Possible reason for the failure is a race condition in the event processing inside of the collections service.

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
naz: Ran the migration on a db with ~4K records. Took `94.457ms` on my local machine. Imo we should be fine here. @daniellockyer any extra checks you'd perform for this kind of batch insert?

naz: For the scalability target of `100,000` the napkin math is 94ms x 25 = `2350ms` to run this query on local machine. If we assume the Pro environment is 10x slower (interesting metric to actually find out maybe?) that would be ~23s or running time.

daniellockyer: FWIW you don't need to loop over chunks because that's what `batchInsert` does: https://knexjs.org/guide/utility.html#batchinsert

naz: TIL! Ty

ghost/collections/src/Collection.ts
```diff
@@ -69,7 +69,7 @@ export class Collection {
     }
 
     public async edit(data: Partial<Collection>, uniqueChecker: UniqueChecker) {
-        if (this.type === 'automatic' && this.slug !== 'index' && (data.filter === null || data.filter === '')) {
```
allouis, line 72: Wondering if we should instead have:
```
if (this.slug === 'latest' || this.slug === 'featured') {
  throw new ValidationError('Cannot edit internal collections');
}
```

ghost/collections/src/CollectionsService.ts
```diff
@@ -401,7 +401,11 @@ export class CollectionsService {
             filter: `posts:${postId}`
         });
 
-        return collections.map(collection => this.toDTO(collection));
+        return collections.map(collection => this.toDTO(collection))
+            .sort((a, b) => {
```
allouis, line 405: Should this be happening at the repository level? 🤔

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
```diff
@@ -0,0 +1,99 @@
...
+            })
+            .first();
+
+        if (!existingLatestCollection) {
+            logging.warn('Latest collection does not exists, skipping');
+        } else {
+            const latestPostsRows = await knex('posts')
+                .select('id')
+                .where({
+                    type: 'post'
```
allouis, line 36: Does this need a filter for `status: 'published'`?

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
```diff
@@ -0,0 +1,99 @@
...
+            })
+            .first();
+
+        if (!existingFeaturedCollection) {
+            logging.warn('Featured collection does not exist, skipping');
+        } else {
+            const featuredPostsRows = await knex('posts')
+                .select('id')
+                .where({
+                    featured: true,
```
allouis, line 56: Does this need a filter for `status: 'published'`?

allouis (approved):
Few small comments.

One thing though - I'm not sure about the use of an empty filter for the "latest" collection 🤔 What do you think about using a filter of ~`-id:null`~ `id:-null` or something like that, which would include all of the posts, but it allows our validation to be as simple as "All automatic collections have a filter"

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
naz: nope. we would store all the posts on the db layer and only filter by published on the application layer. the fixture for "latest" collection has no filter at the moment at all.

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
naz: same answer as above for "latest" collection.

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
allouis: Why do we do that? Is it going to make our pagination more difficult?

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-15-55-add-built-in-collections.js
```diff
@@ -0,0 +1,60 @@
...
+                slug: 'featured'
+            })
+            .first();
+
+        if (existingFeaturedCollection) {
+            logging.warn('Featured collection already exists, skipping');
+        } else {
+            await knex('collections').insert({
+                id: (new ObjectID()).toHexString(),
+                title: 'Featured Posts',
```
kevinansfield, line 40: Could we standardise on not having `Posts` in the collection title? So `Latest` and `Featured` rather than `Latest` and `Featured Posts` so that when used in the UI it's consistent?

allouis (changes requested):
We need to update the collection name as Kevin has requested before merging

ghost/collections/src/CollectionsService.ts
naz: true. ordering is respositorie's concert 👍

naz: Note, when moved to the repository level there was a race condition happening in the post's event processing. So leaving this as is for current purposes, but will need to move this logic once the event processing is improved.

ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
naz: It does, but keeps the separation of logic clean (we had a very brief chat about this [here](https://ghost.slack.com/archives/C05DK3SP801/p1688723868922009))

ghost/collections/src/Collection.ts
naz: We might 🤔 As for now we don't have a requirement for the built-in collections to be "immutable". Maybe something we'd add eventually if that comes up as a feature. It looks like an obvious one to have but don't think we need it just yet.

naz:
> about the use of an empty filter for the "latest" collection 🤔 What do you think about using a filter of -id:null id:-null or something like that, which would include all of the posts, but it allows our validation to be as simple as "All automatic collections have a filter"

@allouis I feel like this is one of those that seems like a good idea but would add future questions like "why is this thing even here?" from anybody reading through the codebase. Pros: we would save an extra if statement(s). Cons: add a confusing statement, which would need extra explanation, which would not be visible in the DB schema but rather somewhere in comments inside of the collections service. 

I'd leave it as is. Unless there are more good reasons to go down this path.

allouis (approved):

daniellockyer (approved):

PR #17251, opened 2023-07-07
🐛 Fixed members unable to unsubscribe from plan if hidden in Portal 

cmraible:
refs TryGhost/Product#3563

- For a member on a paid plan, which had subsequently been hidden from portal, the member was unable to unsubscribe/change plan because the 'Change' button was hidden
- This change restores the 'Change' button for members on a paid plan, even if the plan is hidden from portal
- This change also makes some modifications to the 'Change Plan' page, like showing the current active plan even if it is hidden, and displays a message to comped members to contact support if they want to change their plan

### DIFF
--- a/apps/portal/src/components/Global.styles.js
+++ b/apps/portal/src/components/Global.styles.js
@@ -109,6 +109,12 @@ export const GlobalStyles = `
         cursor: pointer;
     }
 
+    p a {
+        font-weight: 500;
+        color: var(--brandcolor);
+        text-decoration: none;
+    }
+
     svg {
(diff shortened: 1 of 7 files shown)


cmraible:
Planning to revisit this one tomorrow (7/12)

cmraible:
Have this mostly working, except it will currently allow the paid member to switch their plan from Yearly to Monthly, even if Monthly is supposed to be hidden.

CleanShot 2023-07-13 at 21 23 23@2x
CleanShot 2023-07-13 at 21 23 30@2x

PR #17257, opened 2023-07-10
Updated German translations

sven09:
Just added all german translations. The JSON is valid. But i haven't done any further tests.

---

### <samp>🤖 Generated by Copilot at 1cac2a0</samp>

This pull request adds German localization for the Ghost membership and subscription feature. It translates all the strings in the `portal.json` file, which is used to render the portal UI and messages. This change enhances the user experience and accessibility of Ghost for German-speaking audiences.


daniellockyer:
Thanks @sven09! 🙂

PR #17326, opened 2023-07-12
Added search by email for Stripe Customer ID during member import

sagzy:
refs https://github.com/TryGhost/Product/issues/3593
- when importing members via CSV, it's now possible to add the value "auto" for the "stripe_customer_id" field. When "auto" is passed, the importer will search for a Stripe customer based on the email address provided

### DIFF
--- a/ghost/stripe/lib/StripeAPI.js
+++ b/ghost/stripe/lib/StripeAPI.js
@@ -233,14 +233,43 @@ module.exports = class StripeAPI {
         try {
             const result = await this._stripe.customers.search({
                 query: `email:"${email}"`,
-                limit: 1
+                limit: 10,
+                expand: ['data.subscriptions']
             });
+            const customers = result.data;
 
-            if (result.data.length === 0) {
+            // No customer found, return null
+            if (customers.length === 0) {
                 return;
             }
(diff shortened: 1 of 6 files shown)


ghost/stripe/lib/StripeAPI.js
```diff
@@ -222,6 +222,30 @@ module.exports = class StripeAPI {
...
+     * @see https://stripe.com/docs/api/customers/search
+     *
+     * @returns {Promise<string|null>} Stripe Customer ID, if found
+     */
+    async getCustomerIdByEmail(email) {
+        try {
+            const result = await this._stripe.customers.search({
+                query: `email:"${email}"`,
+                limit: 1
+            });
```
SimonBackx, line 237: I'm wondering if we shouldn't search more than 1 customer, and return the customer with the newest subscriptions? To avoid reusing an old customer

sagzy: Yes that makes sense, thanks! Updated in [300364f](https://github.com/TryGhost/Ghost/pull/17326/commits/300364f83606933356a91df8025e72b87ee5da41)

PR #17373, opened 2023-07-14
Retain newsletter subscriptions on suppression

mike182uk:
refs https://github.com/TryGhost/Product/issues/2610


ghost/core/core/server/data/migrations/versions/5.56/2023-07-15-10-11-12-update-members-email-disabled-field.js
```diff
@@ -0,0 +1,22 @@
...
+    async function up(knex) {
+        logging.info('Setting email_disabled to true for all members that have their email on the suppression list');
+
+        await knex('members')
+            .join('suppressions', 'members.email', 'suppressions.email')
+            .update({
+                email_disabled: true
+            });
+    },
+    async function down(knex) {
```
55sketch, line 14: is there something that would prevent a race condition here where a member on suppressions list gets set to `true` and then the general one runs and it gets set to `false`? 

I don't know the actual migration process but "up" suggests that runs first, and then "down" runs at the end, but that seems to me it would undo what was done in up?

55sketch: Ah nice, thanks for clarifying 👍

daniellockyer:
Apart from the migration versions, PR should be ok from a perf POV

daniellockyer (approved):

ghost/core/core/server/data/migrations/versions/5.56/2023-07-15-10-11-12-update-members-email-disabled-field.js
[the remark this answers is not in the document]
mike182uk: I move them back to `5.56` now `5.55` is out?

ghost/core/core/server/data/migrations/versions/5.55/2023-07-14-10-11-12-add-email-disabled-field-to-members.js
daniellockyer, line 1: @mike182uk Sorry, you'll have to move them to 5.56 again now 😅

mike182uk: 👍 https://github.com/TryGhost/Ghost/pull/17373/commits/e0f03e59d412db95f5352b0ad0eb47aee3161ba2

daniellockyer (approved):

PR #17414, opened 2023-07-19
Updated German translations

heartcrazed:
I, as a German native speaker, reviewed all the translations and added some new where some were missing. Some translations/terms were not consistent, so I smoothed these out. Hope this helps.


daniellockyer:
Thanks @heartcrazed! 🙂

PR #17415, opened 2023-07-19
✨ Added support for importing Stripe Coupons as Offers

sagzy:
closes https://github.com/TryGhost/Product/issues/3595
- when importing paid members with a coupon in Stripe, we currently search for the corresponding offer in our database and attach it to the subscription if found. However, if an offer doesn't exist in the database, we do not create one and don't attach any offer to the subscription
- with this change, we now support the creation of a new offer, based on a Stripe coupon, if it didn't exist already

### DIFF
--- a/ghost/offers/lib/application/OfferRepository.js
+++ b/ghost/offers/lib/application/OfferRepository.js
@@ -43,6 +43,14 @@ const mongoTransformer = flowRight(statusTransformer, rejectNonStatusTransformer
  * @prop {string} filter
  */
 
+/**
+ * @typedef {object} OfferAdditionalParams
+ * @prop {string} productId — the Ghost Product ID
+ * @prop {string} currency — the currency of the plan
+ * @prop {string} interval — the billing interval of the plan (month, year)
+ * @prop {boolean} active — whether the offer is active upoon creation
+ */
+
 class OfferRepository {
     /**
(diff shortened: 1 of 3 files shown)


ghost/offers/lib/application/OfferRepository.js
```diff
@@ -179,6 +187,39 @@ class OfferRepository {
...
+     */
+    async createFromCoupon(coupon, params, options) {
+        const {productId, currency, interval, active} = params;
+        const code = coupon.name && coupon.name.split(' ').map(word => word.toLowerCase()).join('-');
+
+        const data = {
+            active,
+            name: coupon.name,
+            code,
+            product_id: productId,
```
SimonBackx, line 203: I might be wrong, but I don't see the product_id and interval options in the [Stripe docs](https://stripe.com/docs/api/coupons/create)? Or maybe I'm misunderstanding something

sagzy: Yes indeed, they're not present in the Stripe Coupon (there are more data fields in the Offer object, than in the Stripe Coupon):
- Product ID is the internal Ghost Product ID
- Currency and Interval are stored outside of the Coupon in Stripe

cf. these [lines](https://github.com/TryGhost/Ghost/blob/2574409ce3fd70674f59df69c1272abfceb4b0f5/ghost/members-api/lib/repositories/MemberRepository.js#L954C1-L956C93)

PR #17453, opened 2023-07-21
Added feature to convert and open mobiledoc posts in the lexical editor

cmraible:
refs TryGhost/Product#3638

- Added `convert_to_lexical` flag to the posts/pages edit endpoint
- Added 'convertToLexical' feature flag so we can enable/disable this feature independently from the main lexical beta flag
- Modified admin posts/pages list to point to the lexical editor for _all_ posts, regardless of mobiledoc vs lexical (if the flag is on)
- Added call to edit endpoint with `convert_to_lexical` in the lexical editor admin route if the page/post is currently in mobiledoc and the flag is enabled

### DIFF
--- a/ghost/admin/app/adapters/page.js
+++ b/ghost/admin/app/adapters/page.js
@@ -11,6 +11,11 @@ export default class Page extends ApplicationAdapter {
             parsedUrl.searchParams.append('save_revision', saveRevision);
         }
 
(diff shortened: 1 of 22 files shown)


cmraible:
To do:
- [ ] Add PostsService tests
- [ ] Add Posts input serializer tests
- [ ] Duplicate for pages

PR #17495, opened 2023-07-25
Added donations API

SimonBackx:
refs https://github.com/TryGhost/Product/issues/3648

Split the method into smaller parts so we can reuse individual parts for the upcoming donation checkout session.


apps/portal/src/utils/api.js
```diff
@@ -356,7 +356,8 @@ function setupGhostApi({siteUrl = window.location.origin, apiUrl, apiKey}) {
                 identity: identity,
                 metadata: metadataObj,
                 successUrl,
-                cancelUrl
+                cancelUrl,
+                type: 'donation'
```
SimonBackx, line 360: Need to be reverted before merging

PR #17505, opened 2023-07-26
Added test email rate limiting

ronaldlangeveld:
refs https://github.com/TryGhost/Product/issues/3651

- This is a security fix that addresses an issue causing malicious users to abuse the test / preview email API endpoint.
- We have multiple procedures in place now to limit such users.
- First, we now only allow one email address to be passed into the `sendTestEmail` method. This method only have one purpose, which is to compliment the test email functionality within the Editor in Admin and therefore have no reason to send to more than one email address at a time.
- We then add an additional rate limiter to prevent a user from making multiple requests, eg via a script.
- The new imposed limit is 10 test emails per hour.

---

### <samp>🤖 Generated by Copilot at 5f4b6db</samp>

No summary available (Limit exceeded: required to process 105780 tokens, but only 50000 are allowed per call)


SimonBackx (approved):
👌

PR #17506, opened 2023-07-26
Implement member import with tier

mike182uk:
refs https://github.com/TryGhost/Product/issues/3629


ghost/members-importer/lib/MembersCSVImporter.js
```diff
@@ -33,24 +34,27 @@ module.exports = class MembersCSVImporter {
...
         this._getMembersRepository = getMembersRepository;
         this._getDefaultTier = getDefaultTier;
+        this._getTierByName = getTierByName;
         this._sendEmail = sendEmail;
         this._isSet = isSet;
         this._addJob = addJob;
         this._knex = knex;
         this._urlFor = urlFor;
         this._context = context;
+        this._tierCache = new Map();
```
allouis: I can't see that we ever clear this cache - which means we could be dealing with outdated values.

ghost/members-api/lib/repositories/MemberRepository.js
```diff
@@ -1635,4 +1639,162 @@ module.exports = class MemberRepository {
...
+     * This method should be used in-conjunction with the `linkSubscription` method to ensure that the changes
+     * made in Stripe are reflected in Ghost - This is not executed as part of this to allow for flexibility
+     *
+     * @param {Object} data
+     * @param {String} data.customer_id - Stripe customer ID
+     * @param {String} data.product_id - Ghost product ID
+     * @param {Object} options
+     * @returns {Promise<Object>}
+     */
+    async forceStripeSubscriptionToProduct(data, options) {
```
allouis, line 1668: I don't think this method belongs in here - we want to decouple the `members-api` package from Stripe completely, and this takes us in the other direction.

I would consider moving this into the import package, or a separate one - but I think importer is fine for now

allouis (commented):
Approach looks good! Would be great if we can move the new stripe code out of members-api though, that will allow us to keep members/stripe decoupled as well as not have to return the archive method

ghost/members-importer/lib/MembersCSVImporter.js
mike182uk: [It gets cleared on each execution ](https://github.com/TryGhost/Ghost/pull/17506/files#diff-a34d8cdc0a7caff73fbddfd0f4e205fb9b247d1b4e1b8b648c05f40a29812a79R121-R122) 🙂

ghost/members-api/lib/repositories/MemberRepository.js
mike182uk: Ok cool , I'll look at moving this into the `members-importer` package seeming though that this logic will probably only ever be called within the context of importing members 👍

mike182uk: I've moved this logic over to the `members-importer` package now - ended up being a little trickier than expected to get the relevant dependencies passed around so had to rejig the package a little

PR #17533, opened 2023-07-30
Updated Bulgarian translations

yovko:
I've fixed some of the previously translated strings and translated the rest.


daniellockyer:
Thanks @yovko! 🙂

PR #17536, opened 2023-07-31
Added migrations for donations API

SimonBackx:
fixes https://github.com/TryGhost/Product/issues/3655

- This adds a new table for storing donation events in the database.
- Makes it possible to store stripe_products without associating it with a tier/Ghost product (required for one time purchases).
- Updates the schema for stripe_prices to allow enum value of `donation`, which is required to query specifically on the prices used for donations (existing one_time is not enough).

Database changes are discussed in the Tech Spec: https://www.notion.so/ghost/Tech-Spec-5cd6929f7960462ebcbf198176e0d899?pvs=4#1f18cc5a38294f61a091e5be63fe1059

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.57/2023-07-26-12-44-stripe-products-nullable-product.js
+++ b/ghost/core/core/server/data/migrations/versions/5.57/2023-07-26-12-44-stripe-products-nullable-product.js
@@ -1,0 +1,4 @@
+const {createSetNullableMigration} = require('../../utils');
+
+// We need to disable foreign key checks because if MySQL is missing the STRICT_TRANS_TABLES mode, we cannot revert the migration
+module.exports = createSetNullableMigration('stripe_products', 'product_id', {disableForeignKeyChecks: true});
--- a/ghost/core/core/server/data/migrations/versions/5.57/2023-07-27-11-47-49-create-donation-events.js
+++ b/ghost/core/core/server/data/migrations/versions/5.57/2023-07-27-11-47-49-create-donation-events.js
@@ -1,0 +1,19 @@
+// For information on writing migrations, see https://www.notion.so/ghost/Database-migrations-eb5b78c435d741d2b34a582d57c24253
+
+const {addTable} = require('../../utils');
+
+module.exports = addTable('donation_payment_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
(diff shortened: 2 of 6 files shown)


ghost/core/core/server/data/migrations/versions/5.57/2023-07-27-11-47-49-create-donation-events.js
```diff
@@ -0,0 +1,19 @@
+// For information on writing migrations, see https://www.notion.so/ghost/Database-migrations-eb5b78c435d741d2b34a582d57c24253
+
+const {addTable} = require('../../utils');
+
+module.exports = addTable('donation_payment_events', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    name: {type: 'string', maxlength: 191, nullable: true},
+    email: {type: 'string', maxlength: 191, nullable: false, unique: true},
```
sagzy, line 8: Curious about `unique: true` for the email field. Why should this be unique? An user with email "abc@example.com" should be able to do multiple donations, and each donation will create a new `donation_payment_event`, right?

SimonBackx: Good catch, this was definitely not intended - copy paste mistake 🙃 I'll remove it

ghost/core/core/server/data/migrations/versions/5.57/2023-07-26-12-44-stripe-products-nullable-product.js
```diff
@@ -0,0 +1,4 @@
+const {createSetNullableMigration} = require('../../utils');
+
+// We need to disable foreign key checks because if MySQL is missing the STRICT_TRANS_TABLES mode, we cannot revert the migration
+module.exports = createSetNullableMigration('stripe_products', 'product_id', {disableForeignKeyChecks: true});
```
ErisDS, line 4: Flagging that the tech spec documents this change, but doesn't have any discussion of pros/cons and in particular if there's any fallout work that would need to be done in any other services to maintain data consistency?

SimonBackx: This was discussed and I've raised the issue again, but still think this is the best solution overall 😊

PR #17576, opened 2023-08-02
Added migrations for Tips & Donations' settings

sagzy:
refs https://github.com/TryGhost/Product/issues/3668

- `Tips and Donations` feature offers two settings: `donations_currency`, and `donations_suggested_amount`
   - `donation_currency`: the currency to be used for the donation. Defaults to `USD`, not nullable.
   - `donation_suggested_amount`: an anchor price for the donation. Defaults to `0`, not nullable.
- Both settings belong to a new group `donations`

cf. [Tech Spec](https://www.notion.so/ghost/Tech-Spec-5cd6929f7960462ebcbf198176e0d899?pvs=4#6e8b34c45f0c4c78b48c9e7725a307c8)

### DIFF
--- a/ghost/core/test/e2e-api/admin/__snapshots__/settings.test.js.snap
+++ b/ghost/core/test/e2e-api/admin/__snapshots__/settings.test.js.snap
@@ -305,10 +305,6 @@ Object {
       "value": null,
     },
     Object {
-      "key": "donations_enabled",
-      "value": false,
-    },
-    Object {
       "key": "donations_currency",
       "value": "USD",
     },
@@ -707,10 +703,6 @@ Object {
       "value": null,
     },
     Object {
-      "key": "donations_enabled",
-      "value": false,
-    },
-    Object {
       "key": "donations_currency",
       "value": "USD",
     },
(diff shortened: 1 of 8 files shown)


ghost/core/test/e2e-api/admin/__snapshots__/settings.test.js.snap
```diff
@@ -694,6 +706,18 @@ Object {
...
+      "key": "donations_enabled",
+      "value": false,
+    },
+    Object {
+      "key": "donations_currency",
+      "value": "USD",
+    },
+    Object {
+      "key": "donations_suggested_amount",
+      "value": "0",
```
sagzy, line 719: Unsure why snapshots have `"0"` (string) instead of `0` (number) for `donation_suggested_amount`

allouis: We don't cast values to the number type, not sure why https://github.com/TryGhost/Ghost/blob/81ef2ade39e57c8431585d971ba6c601278441a3/ghost/core/core/server/models/settings.js#L139-L160

allouis (approved):
We don't validate currencies for Tiers either ATM. I think this is all good!

ErisDS (approved):
LGTM 👍

PR #17588, opened 2023-08-03
Added CollectionPost model & relation to Collection

allouis:
By using the `collections_posts` table as a pivot table, I couldn't see a way to get bookshelf to *not* load the Post models for the relation. We don't actually need those Post models for our usescases, the the queries were causing issues with our database servers! Here we've added a new CollectionPost model which allows us to treat the collections_posts table as a resource ratherthan pivot, and means we can use the hasMany relation rather than belongsToMany relation. This removes all queries to the posts table when fetching collections with relations


naz (approved):
It's a super simple change but works well for our current usecases. I think this is good to go 🚀

PR #17601, opened 2023-08-06
Updated Polish translations

Kr-Fr:
Polish  translation of this file is completed.


Kr-Fr:
Is everything ok? When will the changes be published?

PR #17607, opened 2023-08-07
Added donation notifications column

SimonBackx:
fixes https://github.com/TryGhost/Product/issues/3691

When a site receives a donation, we'll send an email to all administrator. They can disable this in their profile settings. This migration adds the required column to make this possible. It defaults to true.

Link to tech spec: https://www.notion.so/ghost/Tech-Spec-5cd6929f7960462ebcbf198176e0d899?pvs=4#156834b0231d4968a203d797c6817d21


sagzy (approved):
LGTM 👍

PR #17609, opened 2023-08-07
Added index on posts table for published_at column

naz:
closes https://github.com/TryGhost/Arch/issues/18

- The prev/next helpers are slow and are causing major performance issues. The helpers are using `posts.published_at` for comparisons extensively, which causes a full table scan - bad for query performance.
- We use published_at in other queries too (like default order for queries fetching all posts), so there  might be a slight performance boost across the system with this new index.

### <samp>🤖 Generated by Copilot at 71414cb</samp>

This pull request adds an index on the `published_at` column of the `posts` table to improve query performance. It also updates the schema definition, the migration file, and the integrity test file accordingly.

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.59/2023-08-07-11-17-05-add-posts-published-at-index.js
+++ b/ghost/core/core/server/data/migrations/versions/5.59/2023-08-07-11-17-05-add-posts-published-at-index.js
@@ -1,4 +1,5 @@
 const logging = require('@tryghost/logging');
+const DatabaseInfo = require('@tryghost/database-info');
 const {createNonTransactionalMigration} = require('../../utils');
 
 const INDEX_NAME = 'posts_published_at_index';
@@ -7,7 +8,7 @@ module.exports = createNonTransactionalMigration(
     async function up(knex) {
         let hasIndex = false;
 
-        if (knex.client.config.client === 'sqlite3') {
+        if (DatabaseInfo.isSQLite(knex)) {
             const result = await knex.raw(`select * from sqlite_master where type = 'index' and tbl_name = 'posts' and name = '${INDEX_NAME}'`);
             hasIndex = (result.length !== 0);
         } else {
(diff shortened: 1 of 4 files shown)


allouis (commented):
We've used slightly different signatures between `index` and `dropIndex` (passing `string` to the first and `string[]` to the second) - which I think is a little confusing.

We also define the `INDEX_NAME` but only use it for querying - I guess this works as it's the default name produced by both databases, but I think it would be good to explicitly pass it like:
```js
table.index(['published_at'], INDEX_NAME);
// ...
table.dropIndex(['published_at'], INDEX_NAME);
```

But we'd need to update the schema for this too, to support like `index: <index_name>` as well as `index: true`

naz:
@allouis agreed on using consistent syntax. I've gone with `table.index(columns)` as that's the one that's most clearly [documented by knex](https://knexjs.org/guide/schema-builder.html#index).
```
table.index(['published_at']);
// ...
table.dropIndex(['published_at']);
```

As for passing in or even using `INDEX_NAME` inside of these methods, I'm not sure we need it. Right now the purpose of logging the index name is to assist debugging for whoever runs the migration:
```
show index from posts where Key_name = 'posts_published_at_index';
```

We don't need or never used named indexes before, so don't think we need to implement anything extra right now ☺️

ghost/core/core/server/data/migrations/versions/5.59/2023-08-07-11-17-05-add-posts-published-at-index.js
```diff
@@ -0,0 +1,50 @@
+const logging = require('@tryghost/logging');
+const {createNonTransactionalMigration} = require('../../utils');
+
+const INDEX_NAME = 'posts_published_at_index';
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        let hasIndex = false;
+
+        if (knex.client.config.client === 'sqlite3') {
```
daniellockyer, line 10: It'd be great if we could refactor to use `@tryghost/database-info` like https://github.com/TryGhost/Ghost/blob/3fa3a34187cdbad75f1831d743d4d855737d6b51/ghost/core/core/server/data/migrations/versions/4.0/29-fix-foreign-key-for-members-stripe-customers-subscriptions.js#L7

PR #17630, opened 2023-08-08
🎨 Added support for relative links in emails

SimonBackx:
fixes https://github.com/TryGhost/Product/issues/3687

After this change, relative URLs in emails will be replaced with absolute URLs using the post URL. Making relative Portal URLs possible etc.

Updates the test data generator to fix invalid URL encoding (somehow a backslash + escaped double quote was added when it wasn't required).


sagzy:
🚀

PR #17701, opened 2023-08-11
:bug: fixed fields autocomplete not working in popup

### LINKED ISSUE #16960: 1Password can't autofill Member email address
### Issue Summary

Users cannot use a saved email or password on Ghost.io sites with 1Password. I've reproduced this on Firefox and Chrome with the 1Password browser plugin, and also iOS. It probably doesn't work with any password manager, given the reason for the bug, but I've only tested 1Password.

The problem is that the login form is shown to the user in an anonymous (domain-less) iframe. You can read more about how I worked out what the problem was with Ghost in [this thread](https://1password.community/discussion/comment/686048#Comment_686048), and more about the technical specifics of the `about:srcdoc` issue in [this other thread](https://1password.community/discussion/comment/330986/#Comment_330986).

Suffice it to say, password managers are not able to determine the domain of the page hosting the form when that page is generated dynamically and set inside an anonymous iframe. Because one of the important goals of password managers is to [prevent phishing](https://blog.1password.com/1password-keeps-you-safe-by-keeping-you-in-the-loop/), they are very particular about ensuring that they only suggest an item for completion when the domain of the page matches the saved domain. Since the login forms in Ghost are all hosted inside these anonymous frames, there is no domain, just `about:srcdoc`. In the case of desktop browsers, 1Password does use the domain in the URL bar to look for possible matches. But it still won't allow you to auto-fill into an anonymous iframe for the phishing reason described above.

It seems like the solution to this will have to involve either loading actual pages with a URL into the iframe, or else dispensing with the iframe entirely and just overlaying the form with JS/CSS.

The code which generates this page [is probably here](https://github.com/TryGhost/Ghost/blob/9da246ac8448720c210e854519eae72acd8bba46/ghost/portal/src/components/Frame.js#L30), although there are a couple of places which seem to generate anonymous iframes, so I'm not exactly sure which one is used by the login form(s).

### Steps to Reproduce

1. Create a Ghost.io login in your password manager. Set the email and password. Ensure that it's tied to the domain used by your Ghost.io site.
2. Visit the Ghost.io site and attempt to auto-fill the username (email) and password.
3. Notice that you can't autofill them. On iOS, it doesn't even find the correct item.

### Ghost Version

5.50.2? whatever version is live on ghost.io

### Node.js Version

0

Hosted by ghost.io

### Database type

Other

### Browser & OS version

Mac: Chrome, Firefox; iOS: Safari

_No response_

### Code of Conduct

susannakosic:
closes https://github.com/TryGhost/Ghost/issues/16960
- replaced iframe by div

### <samp>🤖 Generated by Copilot at d6e1ccb</samp>

The pull request updates the `PopupModal` component and the portal tests to fix some signin issues and improve the test code quality. It replaces the `Frame` component with a `div` element for the signin page, and uses the `screen` object and the `popupFrame` element for querying DOM elements.

### DIFF
--- a/apps/portal/src/components/Global.styles.js
+++ b/apps/portal/src/components/Global.styles.js
@@ -38,7 +38,7 @@ export const GlobalStyles = `
 
     body {
         margin: 0px;
-        font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen, Ubuntu, Cantarell, "Open Sans", "Helvetica Neue", sans-serif;
+        font-family: Inter, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen, Ubuntu, Cantarell, "Open Sans", "Helvetica Neue", sans-serif;
         font-size: 1.6rem;
         height: 100%;
         line-height: 1.6em;
@@ -48,7 +48,7 @@ export const GlobalStyles = `
         box-sizing: border-box;
         overflow: hidden;
     }
-
+    
     button,
     button span {
         font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen, Ubuntu, Cantarell, "Open Sans", "Helvetica Neue", sans-serif;
@@ -97,9 +97,8 @@ export const GlobalStyles = `
     p {
         font-size: 15px;
         line-height: 1.5em;
-        margin-bottom: 24px;
     }
-
+    
     strong {
         font-weight: 600;
     }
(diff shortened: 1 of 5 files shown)


apps/portal/src/components/Global.styles.js
```diff
@@ -38,7 +38,7 @@ export const GlobalStyles = `
 
     body {
         margin: 0px;
```
susannakosic, line 40: Note for whoever will review this PR.
When testing signin popup, I noticed some font and paragraph changes on the page: headings font looked smaller, and there was a bigger separation between paragraphs (only when popup is open). This was caused by styles injected in the popup which were previously isolated in the iframe and now are embed in the body. I fixed the visual glitch by adding Inter font and removing parragraph bottom margin from global styles, although I am not 100% sure this was the correct approach.

PR #17789, opened 2023-08-23
Added validation for collection filters

allouis:


ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
```diff
@@ -61,24 +62,29 @@ module.exports = class BookshelfCollectionsRepository {
...
+                description: json.description,
+                filter: json.filter,
+                type: json.type,
+                featureImage: json.feature_image,
+                posts: json.collectionPosts.map(collectionPost => collectionPost.post_id),
+                createdAt: json.created_at,
+                updatedAt: json.updated_at
+            });
+        } catch (err) {
+            sentry.captureException(err);
```
naz, line 85: My understanding is that we want to surface these errors more prominently instead of throwing everything into one bag with `logger.error`. If that's the intention maybe we should create a different abstraction instead of depending on very concrete 3rd party like sentry?

allouis: `sentry.captureException` doesn't log out the event, it captures it in sentry and we can setup alerts for it! 

RE: a different abstraction - I think potentially yes, but RN I think that's premature, we currently refer to sentry by name everywhere else, and given this is part of the application codebase (i.e not business logic) we're okay to continue

I think this is missing a `logger.error` though - so I'll add that

PR #17794, opened 2023-08-23
🐛 Fixed translations with spaces missing in signup-form

SimonBackx:
refs https://ghost.slack.com/archives/C02G9E68C/p1692784819620269 
refs https://github.com/TryGhost/Product/issues/3504

- Somehow, when using i18n in TypeScript, the require will put some of the keys into 'default' and not into the root. Mainly all keys that have a space in them. Couldn't find any documentation about this
- The solution is to also add 'default' to the keys that are being used in the code

This change also fixes the translate script (wasn't updated for updated paths), includes the missing translations, and already adds comments to lookup translation strings


ghost/i18n/locales/nn/portal.json
```diff
@@ -56,7 +56,7 @@
     "Error": "Feil",
     "Expires {{expiryDate}}": "Går ut {{expiryDate}}",
     "Forever": "For evig",
-    "Free Trial - Ends {{trialEnd}}": "Gratis prøveperiode - sluttar {{trialEnd}}",
+    "Free Trial – Ends {{trialEnd}}": "",
```
SimonBackx, line 59:
```suggestion
    "Free Trial – Ends {{trialEnd}}": "Gratis prøveperiode - sluttar {{trialEnd}}",
```

PR #17813, opened 2023-08-24
Fixed flaky test behavior for bulk actions

naz:
refs https://ghost.slack.com/archives/C02G9E68C/p1692816097875899

- With introduction of extra e2e test coverage for Collections some tests started to fail at random. The root issue here was the transaction processing collections was started before the original bulk action (bulk edit, bulk publish/unpublish, etc.) was fully committed.

---

### <samp>🤖 Generated by Copilot at 3b8660b</samp>

This pull request enhances the `PostsService` class to emit domain events and return post IDs for bulk edit and destroy operations. It also updates the e2e test for the bulk edit API to ensure the collection posts are refreshed after unpublishing. These changes improve the functionality and reliability of the posts service and API.


allouis (approved):
I think it would be really good to expand on the commit message, or a comment here, explaining that it's because of the `if (!options.transaction)` blocks that mean the transaction isnt commit until everything in the method returns, and so we need to pull the domain event stuff outside of that method!

PR #17819, opened 2023-08-24
Refactored Collection and extracted Entity

allouis:
This is a small spike into making it easier to work with entities, attemping to remove the boiler plate required for getters and setters of properties, whilst retaining type information.

It also supports validation of properties, both in the default setters, and on instantiation (but that requires boilerplate in the create factory)

Custom setters will need to call `validate` manually.

---

### <samp>🤖 Generated by Copilot at 3fd22c4</samp>

This pull request refactors the `Collection` and `CollectionsService` classes to use a new abstract `Entity` class, which provides common functionality and validations for entities. It also adds a custom error for deleting built-in collections, and updates the test cases accordingly. The changes improve the code quality, consistency, and user experience of the collections feature.

### DIFF
--- a/ghost/collections/src/CollectionsService.ts
+++ b/ghost/collections/src/CollectionsService.ts
@@ -1,5 +1,4 @@
 import logging from '@tryghost/logging';
-import tpl from '@tryghost/tpl';
 import {Knex} from "knex";
 import {
     PostsBulkUnpublishedEvent,
@@ -10,24 +9,12 @@ import {
 import {Collection} from './Collection';
 import {CollectionRepository} from './CollectionRepository';
 import {CollectionPost} from './CollectionPost';
-import {MethodNotAllowedError} from '@tryghost/errors';
 import {PostDeletedEvent} from './events/PostDeletedEvent';
 import {PostAddedEvent} from './events/PostAddedEvent';
 import {PostEditedEvent} from './events/PostEditedEvent';
 import {PostsBulkDestroyedEvent} from '@tryghost/post-events';
 import {RepositoryUniqueChecker} from './RepositoryUniqueChecker';
 import {TagDeletedEvent} from './events/TagDeletedEvent';
-
-const messages = {
-    cannotDeleteBuiltInCollectionError: {
-        message: 'Cannot delete builtin collection',
-        context: 'The collection {id} is a builtin collection and cannot be deleted'
-    },
-    collectionNotFound: {
-        message: 'Collection not found',
-        context: 'Collection with id: {id} does not exist'
-    }
-};
 
 interface SlugService {
     generate(desired: string, options: {transaction: Knex.Transaction}): Promise<string>;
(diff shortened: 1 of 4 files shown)


ghost/collections/src/CollectionsService.ts
```diff
@@ -521,16 +508,7 @@ export class CollectionsService {
...
-                throw new MethodNotAllowedError({
-                    message: tpl(messages.cannotDeleteBuiltInCollectionError.message),
-                    context: tpl(messages.cannotDeleteBuiltInCollectionError.context, {
-                        id: collection.id
-                    })
-                });
-            }
-
-            collection.deleted = true;
+            collection.delete();
```
allouis, line 511: Changed to a `delete()` method which handles throwing for us

ghost/collections/src/CollectionsService.ts
```diff
@@ -292,7 +279,7 @@ export class CollectionsService {
             const collections = await this.collectionsRepository.getAll({transaction});
 
             for (const collection of collections) {
-                if (collection.includesPost(postId)) {
+                if (collection.posts.includes(postId)) {
```
allouis, line 282: No need for a specific method here

ghost/collections/src/CollectionsService.ts
```diff
@@ -10,25 +9,13 @@ import {
...
 import {CollectionPost} from './CollectionPost';
-import {MethodNotAllowedError} from '@tryghost/errors';
 import {PostDeletedEvent} from './events/PostDeletedEvent';
 import {PostAddedEvent} from './events/PostAddedEvent';
 import {PostEditedEvent} from './events/PostEditedEvent';
 import {PostsBulkDestroyedEvent} from '@tryghost/post-events';
 import {RepositoryUniqueChecker} from './RepositoryUniqueChecker';
 import {TagDeletedEvent} from './events/TagDeletedEvent';
 
-const messages = {
```
allouis, line 21: Unused

ghost/collections/src/Entity.ts
```diff
@@ -0,0 +1,104 @@
+
```
allouis, line 1: Here be dragons

naz: This would become a base for all future entities in the system, right? So, similarly to the `domain-events` package we'd probably want a `domain-entity` package?

ghost/collections/src/Entity.ts
```diff
@@ -0,0 +1,104 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    protected validations!: any;
+
+    constructor(data: Data) {
+        this._attr = data;
+        this.attr = new Proxy(this._attr, {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            set(target: any, key: any, value: unknown) {
+                if (target[key] !== value) {
+                    if (Reflect.has(target, 'updatedAt')) {
```
naz, line 19: Another one to think of is probably a property identifying if the entity has been modified or "touched" since instantiation 🤔 Is see this useful in situations where we check if we need to reset cache when the data has been modified. Wdyt?

ghost/collections/src/Entity.ts
allouis: Yeah exactly, haven't split it out yet as it was just playing around with it!

daniellockyer:
Closing for now as the PR is stale but please re-open if the contents were important 🙂

PR #17823, opened 2023-08-24
Promoted collections feature to GA

allouis:
### <samp>🤖 Generated by Copilot at 60f2073</samp>

Graduated `collections` feature from alpha to GA. This allows users to create custom content routes and templates with `ghost/core/core/shared/labs.js`.


allouis:
We don't want to put this feature GA yet

PR #17831, opened 2023-08-26
Removed Bluebird dependency

chairulakmal:
refs #14882

👋 Hi there! I'm a new contributor and I've made minimum changes to the last file requiring Bluebird and then remove it from `ghost/core`. I tried to refactor the code to operate with async/await but after realizing how raw-knex is used, I opted to instead do a one line change where Bluebird Promise.props is used. Subsequently I updated the core package.json by removing Bluebird, since it's now unnecessary for any file in ghost/core. Your feedback is much appreciated!

Thanks,
kai

- [x] The build will pass (run `yarn test:all` in core and `yarn lint`)

---

### <samp>🤖 Generated by Copilot at e625a9f</samp>

This pull request removes the `bluebird` library and updates the `raw_knex` plugin to use native promises methods. This simplifies the code and reduces the number of dependencies in the project.


chairulakmal:
Hi @daniellockyer, thanks for running the CI tests! I didn't realize there's a regression test specifically for core. Will fix it today or tomorrow.

chairulakmal:
Instead of transforming raw_knex into an asynchronous function, I chose to make one minor change to eliminate the need for Bluebird's Promise.props. This approach successfully achives the goal of eliminating Bluebird from the core. Furthermore, I verified that executing `yarn test:all` yielded comparable results to those of the existing main branch.

ghost/core/core/server/models/base/plugins/raw-knex.js
```diff
@@ -166,7 +165,7 @@ module.exports = function (Bookshelf) {
                         })();
                     });
 
-                    return Promise.props(props)
+                    return Promise.all(Object.values(props))
```
chairulakmal, line 168: Promise.all returns arrays of resolved promises, which means the original keys are lost. Need to add the original keys back to the resolved promises array from Promise.all.

daniellockyer:
Thanks @chairulakmal! 😀

PR #17856, opened 2023-08-29
Added new setting to enable Recommendations

sagzy:
closes https://github.com/TryGhost/Product/issues/3772

- Recommendations are disabled by default and will get enabled when a first recommendation is added. The "recommendations_enabled" setting will be not shown explicitly in Admin, it will rather be set implicitly based on whether there is at least one recommendation
- The "recommendations_enabled" setting belongs to a new setting group called "recommendations"

[Tech Spec](https://www.notion.so/ghost/Recommentions-Tech-Spec-598f09e4109f4eef8d9e32db7245a90c?pvs=4)


ghost/core/core/server/data/schema/default-settings/default-settings.json
```diff
@@ -565,5 +565,15 @@
             },
             "type": "number"
         }
+    },
+    "recommendations": {
+        "recommendations_enabled": {
+            "defaultValue": "true",
```
SimonBackx, line 571:
```suggestion
            "defaultValue": "false",
```

ghost/core/core/server/data/migrations/versions/5.61/2023-08-29-09-32-32-add-recommendations-enabled-setting.js
```diff
@@ -0,0 +1,8 @@
+const {addSetting} = require('../../utils');
+
+module.exports = addSetting({
+    key: 'recommendations_enabled',
+    value: 'true',
```
SimonBackx, line 5:
```suggestion
    value: 'false',
```

SimonBackx:
We probably also want to update `ghost/core/test/utils/fixtures/default-settings.json`, `ghost/core/test/utils/fixtures/default-settings-browser.json` and `ghost/core/core/shared/settings-cache/public.js`

sagzy:
> We probably also want to update `ghost/core/test/utils/fixtures/default-settings.json`, `ghost/core/test/utils/fixtures/default-settings-browser.json` and `ghost/core/core/shared/settings-cache/public.js`

Thanks! Added these as well

ghost/core/core/server/data/migrations/versions/5.61/2023-08-29-09-32-32-add-recommendations-enabled-setting.js
sagzy: Updated, following the Slack discussion [here](https://ghost.slack.com/archives/C04TMVA1D7A/p1693310637853019)

ghost/core/core/server/data/schema/default-settings/default-settings.json
sagzy: Updated, following the Slack discussion [here](https://ghost.slack.com/archives/C04TMVA1D7A/p1693310637853019)

SimonBackx (approved):

Issue #9189, opened 2017-10-29
How to remove the trailing slash on the canonical meta tag on the home page?


michlschmitt:
The meta tag rel="canonical" on the home page has a trailing slash, though the blog_url is without trailing slash? How can this be removed?

The canonical url should be http://example.com instead http://example.com/ for SEO purposes.

I use the latest Ghost version.

acburdine:
@michlschmitt hey :) Ghost has built-in middleware that automatically redirects all requests without a trailing slash to their corresponding urls that _do_ have a trailing slash. For example, if you go to `http://example.com` (without a trailing slash), you will see that the request is redirected to `http://example.com/` (with a trailing slash). Therefore, the canonical URL is correct based on the intended behavior of Ghost.

Closing this as it is intended behavior and not a bug.

MiguelCrespo:
Hi, @acburdine Is there a way to remove this trailing slash from the URIs? I don't see any benefit of having the trailing as the default because most people don't write a slash at the end of the URI neither popular websites have this structure

kirrg001:
@MiguelCrespo 

This is currently not configureable. Ghost uses for **any** url a trailing slash and will redirect none trailing slashes - originally decided [here](https://github.com/TryGhost/Ghost/issues/448). At the end, with or without trailing slash, SEO treats both as unique urls with unique content, [see](https://webmasters.googleblog.com/2010/04/to-slash-or-not-to-slash.html). That means, you as a product need to decide for a format.

Furthermore, Ghost doesn't use any extensions for e.g. pages or collections, which is a good argument for the decision being made in 2013. Also personally speaking, i really like the look of trailing slashes 😜

Feel free to create a new idea/topic in our [forum](https://forum.ghost.org) - then people can up vote for making this configureable.

codinggirl:
I need the NO TRAILING SLASH style.

danielpacuraru:
i'd like this to be configurable as i prefer the 'no trailing slash' variation

Sudospective:
As much as I respect your decision for your software, I absolutely hate trailing slashes in any website's URL, let alone my own. I feel this decision was based on preference and should be left to the preference of the user. This isn't something that will force me to use a difference CMS, I love Ghost, but it certainly left me blindsided not knowing this before I installed.

adamhenson:
> Therefore, the canonical URL is correct based on the intended behavior of Ghost.

[...]

PR #17859, opened 2023-08-29
Set trailingSlash: true for @site.url

### LINKED ISSUE #16712: Trailing slash stripped from @site.url and resulting in redirects
### Issue Summary

[Ghost's policy](https://github.com/TryGhost/Ghost/issues/9189#issuecomment-340292525) is to include trailing slashes on URLs. However, trailing slashes are [stripped](https://github.com/TryGhost/Ghost/blob/792834e742b753dce3871b128e7f1f369498f34c/ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js#L12) from `@site.url`. That causes a problem when Ghost is installed at a subdirectory, because all links to `@site.url` are redirected to the path with a trailing slash. In other words, there is at least one invalid link on every page with a link to `@site.url`. While ordinary users may not notice anything but a slight slowdown, for SEO these invalid links are pretty negative.

I believe the problem can be solved by removing `trailingSlash: false` [in `updateLocalTemplateOptions`](https://github.com/TryGhost/Ghost/blob/792834e742b753dce3871b128e7f1f369498f34c/ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js#L12), but I'm not sure what effect that would have on Ghost installations on a domain (rather than a subdirectory).

### Steps to Reproduce

1. Install Ghost to a subdirectory, setting the configured `url` value like `https://www.example.com/blog/`.
2. Use the Casper theme (or any other theme that uses `@site.url`).
3. On any page that uses default.hbs, click the logo image. Notice you are 301 redirected from `` to ``.

### Ghost Version

5.44.0

### Node.js Version

v16.20.0

Ubuntu 20.04

### Database type

MySQL 8

### Browser & OS version

Chrome 108

```shell
Not applicable
```

### Code of Conduct

dan-jensen:
[Ghost policy](https://github.com/TryGhost/Ghost/issues/9189#issuecomment-340292525) is to add a trailing slash to all URLs generated. However, it was not doing that for `@site.url`. That created a problem for Ghost installations at subdirectories, because requests to the generated URL like `example.com/blog` would be redirected by Ghost to the trailing-slash version like `example.com/blog/`. That redirect harmed SEO.

This replaces `trailingSlash: false` with `trailingSlash: true` in the `@site.url` generation. Now Ghost complies with its trailing slash policy and avoids creating an unnecessary, SEO-harming redirect. This helps Ghost installations at subdirectories, and does not affect Ghost installations at domains.

Resolves #16712

### DIFF
--- a/ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js
+++ b/ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js
@@ -9,7 +9,7 @@ function updateLocalTemplateOptions(req, res, next) {
 
     // adjust @site.url for http/https based on the incoming request
     const siteData = {
-        url: urlUtils.urlFor('home', {trailingSlash: false}, true)
+        url: urlUtils.urlFor('home', {trailingSlash: true}, true)
     };
 
     // @TODO: it would be nicer if this was proper middleware somehow...


dan-jensen:
@ronaldlangeveld I'm hoping this is an easy win, would you have any time to do the code review?

broksonic21:
We ran into this too - we could really use a site.url with trailing slash option (or default)

dan-jensen:
@daniellockyer looks like you're a backend dev (I'm not sure who else to reach out to). Could you review + merge this PR?

muratcorlu:
In an open source project that has 42k stars, someone is begging here to make their PR to be noticed since 8 months. That's really sad. :(

ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js
```diff
@@ -9,7 +9,7 @@ function updateLocalTemplateOptions(req, res, next) {
 
     // adjust @site.url for http/https based on the incoming request
     const siteData = {
-        url: urlUtils.urlFor('home', {trailingSlash: false}, true)
```
muratcorlu, line 12: Could you test this setup with storages? I think storages uses this base url to construct uploaded file urls. It will be nice to be sure this doesn't break anything (I'm not sure if we have automated tests cover those cases).

dan-jensen: @muratcorlu thanks for reviewing. That's a good question. I think a Ghost backend developer could answer that question pretty quickly – certainly much quicker than me.

I think if storages are simply appending strings to the base URL, there should be a separate PR to make that more intelligent and bug-proof. If there is also missing test coverage, it should probably be added at the same time.

cmraible:
Hi @dan-jensen, thanks for submitting, and apologies for not taking a look at this PR sooner!

My main concern here is that adding the trailing slash to `@site.url` could break existing themes that have historically depended on there _not_ being a trailing slash. It's fairly common in themes to construct links like: `href="{{@site.url}}/page/2/"` — if we add a trailing slash to `@site.url`, then this would become `href="{base_url}//page/2/" which also isn't great.

Just for my own edification, could you elaborate on how this behavior is harming SEO? It looks like redirecting to the version with a trailing slash is [recommended by Google](https://developers.google.com/search/blog/2010/04/to-slash-or-not-to-slash#your-sites-configuration-and-your-options) but I'm not an expert in SEO so happy to learn something new here!

I realize the issue isn't that we're doing the redirect, but that we're linking to the non-trailing slash version of the URL, but I don't fully understand how that is impacting SEO.

cmraible:
Also, just in the interest of offering a workaround while we sort this out, I wonder if this couldn't be solved more expediently at the theme level by adding the trailing slash manually? Are you using a default theme or something custom?

dan-jensen:
@cmraible fantastic, thanks for taking a look!

> could you elaborate on how this behavior is harming SEO?

I became aware of this problem when an Ahrefs SEO scan flagged these redirects as harming SEO on a site. The Ahrefs blog [says](https://ahrefs.com/blog/301-redirects/) Google stopped directly penalizing for redirects in 2019. So I believe the SEO penalty is because of the added delay to page load times caused by the second request.

> I wonder if this couldn't be solved more expediently at the theme level by adding the trailing slash manually

Yep, that's what we did. But now we can't update the existing theme or change themes without going back through the files and making those changes again. Plus, other people will end up going through this same experience, and they might not know how.

> if we add a trailing slash to @site.url, then this would become href="{base_url}//page/2/" which also isn't great.

Yes, but for every theme where that's true for domain/subdomain installations, the inverse (no slash instead of double-slash) is __currently__ a problem for subdirectory installations. Also, I'm not sure the core should suffer from an inconsistency because some themes do something they shouldn't. Shouldn't themes use a helper method to generate URLs, or use only paths, to avoid this exact conundrum?

> could break existing themes

Yes, that is my biggest concern too. Maybe wait for the next major release, where breaking changes are expected?

cmraible:
> Yep, that's what we did. But now we can't update the existing theme or change themes without going back through the files and making those changes again. Plus, other people will end up going through this same experience, and they might not know how.

Glad to hear you were able to work around this! I was more thinking we could update this in our official themes, so it wouldn't require any theme changes for folks who run into this in the future. If I'm understanding correctly, the issue is just where we link to the index page on e.g. the site logo.

I think this would resolve the problem for sites on subdirectory installs, while also not breaking any existing sites/themes that expect the `@site.url` to not have a trailing slash for building URLs.

@dan-jensen Does that sound reasonable? Would that fix the problem your concerned about with redirects?

dan-jensen:
@cmraible not unreasonable, but sounds like a lot of work now, and hard to remember during future theme-building.

After thinking more about this, I think the simplest solution is to tweak Ghost to stop removing trailing slashes from the user-provided `url` value in the config. Or at least stop removing them when a directory is detected (because that's where the trailing slash is critical). This way users could enter a trailing slash if appropriate, and Ghost would respect this user input as intentional/meaningful.

Respecting user input should have minimal impact on existing installations:

- For installations without a trailing slash on the `url` value (the vast majority of installations): There would be no change.
- For installations with a trailing slash (a minority): Some of these installations should not have the trailing slash, and for this minority-of-a-minority there will be some link paths containing double-slashes. However, these paths are still navigable by browsers, and a Googler says ["that’s not an issue"](https://www.searchenginejournal.com/google-answers-if-double-slash-in-url-has-seo-impact/504364/) for SEO, so there would be no meaningful change.

Could you help me figure out how to tweak Ghost to stop removing the trailing slash from `url` values?

ErisDS:
I'm sorry we didn't come back to you on this. The behaviour around trailing slashes in Ghost is very baked-in, and affects so much that any attempt to change it is hard. 

I do see the problem you're describing and it is frustrating, but it's the kind of thing where the impact and work required to get it sorted is outsized to the size of the problem.

I think the solution would involve adding some kind of url helper to compose urls in themes, rolling that out slowly, and then deprecating the old @site.url helper  - or something of that ilk.

I'm gonna close for now as this is super duper old. We can always revisit with a new plan in a different issue or PR.

PR #17931, opened 2023-09-01
Added missing translation helpers in members emails.

partnerwerk:

- [ x] There's a clear use-case for this code change, explained below
- [ x] Commit message has a short title & references relevant issues
- [ x] The build will pass (run `yarn test:all` and `yarn lint`)

---

Use case: Signup emails for members where missing translations in the very last line. While the upper part is localized nicely this line still used english language.

Thank you for your work!

Henning

### <samp>🤖 Generated by Copilot at 2ae51b0</samp>

This pull request adds or updates translations for email-related texts in the `ghost.json` files for German and English locales. It also replaces some hardcoded texts with the `t` function in the email service files to enable internationalization.


daniellockyer:
@partnerwerk I don't think we'll be able to merge this because languages where the word order is different won't make sense after this

partnerwerk:
@aniellockyer It is not me who built it like that. I just added the option for translating the two snippets which already are in the specific english word order. I think keeping one line in english while everything is translated correctly does not make sense either. Definitely the whole line could be improved. This would be an intermediate step which works well in quite a few languages. The others could be kept without translations until there is a better solution.

PR #17938, opened 2023-09-02
Fixed private mode cookie for local development

### LINKED ISSUE #17514: Make this site private - not working in Chrome and Opera
### Issue Summary

Change or delete the row 58:
versions/5.54.4/core/frontend/apps/private-blogging/lib/middleware.js

 ```
return session({
            name: 'ghost-private',
            maxAge: constants.ONE_MONTH_MS,
            signed: false,
            sameSite: 'lax'     <----- row 58: instead of 'none' or delete row
        })(req, res, next);

```
Or you can delete the row 58 because **sameSite: 'Lax' is the default value.**
You can't code 'secure' within an object - **secure: true** will not work.

Works now in Chrome and Opera.
See [https://web.dev/i18n/en/samesite-cookies-explained](https://web.dev/i18n/en/samesite-cookies-explained/#samesite=lax-by-default)
x (1) Mark cross-site cookies as Secure to allow setting them in cross-site contexts

### Steps to Reproduce

See [https://forum.ghost.org/t/make-this-site-private-not-working/39938/1](https://forum.ghost.org/t/make-this-site-private-not-working/39938/1)

### Ghost Version

5.54.4

### Node.js Version

v18.15.0

local, macos

### Database type

SQLite3

### Browser & OS version

_No response_

_No response_

### Code of Conduct

### PR DESCRIPTION
**Pull request by joe-blocher**
"Make this site private" - not working in Chrome and Opera 
Closes #17514

---

### <samp>🤖 Generated by Copilot at acad566</samp>

Fix private blogging cookie issue with Chrome 80+. Update `sameSite` and `secure` options for the cookie in `middleware.js` based on SSL configuration.

### CODE REVIEW

File: ghost/core/core/frontend/apps/private-blogging/lib/middleware.js [hunk 1]
```diff
@@ -55,7 +55,8 @@ const privateBlogging = {
             name: 'ghost-private',
             maxAge: constants.ONE_MONTH_MS,
             signed: false,
-            sameSite: 'none'
+            sameSite: urlUtils.isSSL(config.get('url')) ? 'none' : 'lax',
+            secure: urlUtils.isSSL(config.get('url'))
```
markstos, line 59: Ghost is intended to work locally, over `http`, as well in production, over `https`. 

Before, the `sameSite` attribute for the cookie was always set to `none`. 

Here's what the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies) have to say about the sameSite attribute:

> None specifies that cookies are sent on both originating and cross-site requests, but only in secure contexts (i.e., if SameSite=None then the Secure attribute must also be set).

But for the local-development case in the bug report, the `secure` flag wasn't being set and *can't* be set. 

Here, the proposed solution it continue to `sameSite: none` just as before in the https/production case, but resolves the issue by setting `sameSite: lax` for the local case. 

As @joe-blocher points out, Ghost is already using the exact same solution in `core/server/services/auth/session/express-session.js` for a similar case. 

This change LGTM. I recommend merging.

markstos (approved):

joe-blocher:
Not fixed in version 5.82.2!

kilmarnock:
I observe the error in 5.91.1. Can I help with some debug information?

PR #17943, opened 2023-09-03
deps: remove bluebird override because bluebird has been removed from Ghost software

markstos:
Work on #14882 has removed Bluebird as a dependency for Ghost components.

As a result, this toggle is not needed anymore

### DIFF
--- a/ghost/core/core/server/overrides.js
+++ b/ghost/core/core/server/overrides.js
@@ -1,10 +1,3 @@
-/**
- * If we enable bluebird debug logs we see a huge memory usage.
- * You can reproduce by removing this line and import a big database export into Ghost.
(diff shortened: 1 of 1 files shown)


FraneCaleta (approved):

daniellockyer:
Thanks for the PR @markstos! 🙂 We do need to do this, but I'd like to wait until we've fully removed Bluebird from Ghost and its dependencies (which might be tricky because Bookshelf uses it internally)

daniellockyer:
Hi @markstos, unfortunately Bookshelf still uses Bluebird underneath, and we're unlikely to be able to replace it any time soon. I'm going to close this PR for now but it's definitely on my mind to do 🙂

PR #17956, opened 2023-09-04
Add ability to specify newsletter in a custom sign-up form

mike182uk:
refs https://github.com/TryGhost/Product/issues/3809

When an input is added to a custom sign-up form with the attribute `data-members-newsletter`, the value of the input will be used as the newsletter to subscribe the member to.

### DIFF
--- a/apps/portal/src/data-attributes.js
+++ b/apps/portal/src/data-attributes.js
@@ -38,12 +38,14 @@ export function formSubmitHandler({event, form, errorEl, siteUrl, submitHandler}
         email: email,
         emailType: emailType,
         labels: labels,
-        newsletters: newsletters,
         name: name,
         autoRedirect: (autoRedirect === 'true')
     };
     if (urlHistory) {
         reqBody.urlHistory = urlHistory;
+    }
+    if (newsletterInputs.length > 0) {
+        reqBody.newsletters = newsletters;
     }
 
     fetch(`${siteUrl}/members/api/send-magic-link/`, {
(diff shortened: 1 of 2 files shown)


apps/portal/src/data-attributes.js
```diff
@@ -32,6 +38,7 @@ export function formSubmitHandler({event, form, errorEl, siteUrl, submitHandler}
         email: email,
         emailType: emailType,
         labels: labels,
+        newsletters: newsletters,
```
vershwal, line 41: Hey @mike182uk 
Can we add this `newsletters` field to `reqBody` only when `data-members-newsletter` is included? 

**Why do we need this?**
1. This way the code is not backward compatible. As we pass an empty array, it doesn't subscribe to the default newsletters. [here](https://github.com/TryGhost/Ghost/blob/main/ghost/members-api/lib/repositories/MemberRepository.js#L285) 

2. There are two cases:-

- Case 1: When the site doesn't have a custom sign-up form(Current scenario).
- Case 2: When a site has a custom sign-up form but the user explicitly unchecks all the newsletters(In this case we only do a sign-in, and don't subscribe to any newsletter).

     In both these cases, we are receiving `reqBody.newsletters === []` and the backend cannot differentiate between the two.

mike182uk: Good catch @vershwal 🙌 

I've updated that [here](https://github.com/TryGhost/Ghost/pull/17956/commits/7e8e73f9637b3a79ae12e1ca27045d5aeb555e97)

Issue #17922, opened 2023-09-01
New sign ups Free email Newsletter notifications has a wrong from email address


angelfplaza:
### Issue Summary

Explain roughly what’s wrong:
There are a lot of themes that use a non pop up subscribe option and those are the ones that are not working. 

The new sign up notification from email address is supposed to be [no-reply@myowndomain.com](mailto:no-reply@myowndomain.com). I have already set up all from email address as another address using the smtp config. That email is used for password reset and welcome emails. 
To change the newsletter email, that was also sent from the wrong email address. 
Even with that wrong, I clicked on the link to replace it and now the newsletter settings shows the right one, but the signups are still being sent from the wrong address. 

What did you expect to happen?
Both the email to change the newsletter from address and the new signups for new free subscribers should receive the confirmation email using the SMTP setup email and/or the newsletter one. Not the default no-reply@domainc.com

### Steps to Reproduce

New install. 
activate the theme Edition https://github.com/TryGhost/Edition/ 
go to Settings → Newsletter → Customise- change email- that one will be sent from the wrong email. 
Approve that change (if you managed to get the email as I did). 

Then confirm that has been changed. 
Subscribe to the free emails using a custom embebed form from the theme, not the portal one. 
The email is again sent from the wrong one. 

### Ghost Version

 5.60.0

### Node.js Version

v18.17.1

### How did you install Ghost?

ubuntu 22 following guide and ghost cli

### Database type

MySQL 8

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

markstos:
The same issue was recently reported on the Ghost forum by a different user and a different version: 5.59.4: https://forum.ghost.org/t/staff-new-sign-ups-email-notifications-has-a-wrong-from-email-address/40671

markstos:
I can confirm I'm experiencing this issue as well. I just hadn't noticed because email programs don't always display the email address these days, so the issue may be more widespread than is being reported. The earliest message with the issue is August 14th. Unfortunately, I've emptied my trash so I don't have an earlier message where I can confirm the address is set as intended, but I believe it was something other than ghost@mydomain.com before. Sorry, I don't have the version handy where I ran into the issue-- I've also upgraded since then. 

I'm not using Docker and am using Ubuntu 22, installed using the CLI.

markstos:
I'm able to confirm this is a bug in Ghost. Here's how I isolated and confirmed the issue.

[...]

PR #17966, opened 2023-09-04
email:  Use noreply@ instead of ghost@ to simplify and avoid confusion

markstos:
refs https://github.com/TryGhost/Ghost/issues/17922
refs https://forum.ghost.org/t/staff-new-sign-ups-email-notifications-has-a-wrong-from-email-address/40671

As seen in the linked issues, getting emails from ghost@ appears to be mistake because:

 1. It's different from other emails that Ghost generates, like noreply@
 2. The `ghost@` email appars that perhaps it could be replied to, but often it's set up.

This is also an opportunity to simplify the system by using `noreply@` in more cases. Further refactoring could be done simply internals now that they are the same address.

Both the contributing doc and the pull request template recommend that I use `yarn test` or `yarn test:all` to test my change locally. But neither works for me:

```
yarn run v1.22.19
$ nx run-many -t test

>  NX  The current directory isn't part of an Nx workspace.

  To create a workspace run:
  npx create-nx-workspace@latest <workspace name>


>  NX   NOTE  For more information please visit https://nx.dev/
```

```
❯ yarn test:all
yarn run v1.22.19
error Command "test:all" not found.
```


markstos:
This conflict has been resolved and this is ready for consideration again.

9larsons:
Thanks for this, @markstos — and apologies it sat so long.

Since this PR was opened, Ghost moved outbound email addressing onto the `EmailAddressService` / `getDefaultEmail()` path. The staff "from" address is now `getDefaultEmail()` — your configured `mail.from`, or noreply@<domain> by default — rather than ghost@<domain>, so the confusion this set out to fix is already resolved on main. The specific ghost@${defaultEmailDomain} fallback this PR edits was removed in [#22001](https://github.com/TryGhost/Ghost/pull/22001) (cleaning up the newEmailAddresses flag), and @tryghost/staff-service was later moved into core — so 3 of the 5 files here no longer exist at their old paths and the change no longer applies.

The one remaining spot that still hardcodes ghost@<domain> is the API-key invite reply-to in services/invites/invites.js; if that's worth tidying, it'd be best as a small fresh PR routing through 1getDefaultEmail()`.

Closing as superseded — thank you for the contribution, and for flagging the inconsistency.

PR #17970, opened 2023-09-05
Improved efficiency of Members request caching for the Posts Content API 

allouis:
At the moment we're using the entire `option` object as the cache key for requests to the posts content API. For unauthenticated requests this is fairly okay, but for requests with members authentication we are storing the same content under many keys!

The changes here are to enable custom key data generation at the endpoint layer, as well as implementing custom key data for the posts api.

We strip all options which don't have an effect on the content, e.g. `debug`. And then only include member data that has an effect on the content. This reduces the number of unique keys and so allows us to cache more data.

### DIFF
--- a/ghost/api-framework/lib/pipeline.js
+++ b/ghost/api-framework/lib/pipeline.js
@@ -1,5 +1,6 @@
 const debug = require('@tryghost/debug')('pipeline');
 const _ = require('lodash');
+const stringify = require('json-stable-stringify');
 const errors = require('@tryghost/errors');
 const {sequence} = require('@tryghost/promise');
 
@@ -229,7 +230,13 @@ const pipeline = (apiController, apiUtils, apiType) => {
             frame.docName = docName;
             frame.method = method;
 
-            let cacheKey = JSON.stringify(frame.options);
+            let cacheKeyData = frame.options;
+            if (apiImpl.generateCacheKeyData) {
+                cacheKeyData = await apiImpl.generateCacheKeyData(frame);
+            }
+
+            const cacheKey = stringify(cacheKeyData);
+
             if (apiImpl.cache) {
                 const response = await apiImpl.cache.get(cacheKey);
 
--- a/ghost/core/core/server/api/endpoints/posts-public.js
+++ b/ghost/core/core/server/api/endpoints/posts-public.js
@@ -23,6 +23,34 @@ const rejectPrivateFieldsTransformer = input => mapQuery(input, function (value,
     };
 });
 
+function generateOptionsData(frame, options) {
+    return options.reduce((memo, option) => {
+        let value = frame.options?.[option];
+        if (['include', 'fields', 'formats'].includes(option)) {
+            value = value?.split(',').sort();
+        }
+
+        if (option === 'page') {
+            value = value || 1;
+        }
+
+        return {
+            ...memo,
+            [option]: value
+        };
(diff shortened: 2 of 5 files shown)


ghost/api-framework/lib/pipeline.js
```diff
@@ -230,6 +230,9 @@ const pipeline = (apiController, apiUtils, apiType) => {
             frame.method = method;
 
             let cacheKey = JSON.stringify(frame.options);
```
naz, line 232: We can save some time doing this serialization only when needed. `JSON.stringify` is not a crazy expensive operation but who knows what's in the frame options and would add up unnecessarily 

What I mean is:
```
let cacheKey;

if (apiImpl.generateCacheKey)
   cacheKey = await apiImpl.generateCacheKey(frame);
} else {
  // default key
  cacheKey = JSON.stringify(frame.options);
}

File: ghost/core/core/server/api/endpoints/posts-public.js [hunk 2]
```diff
@@ -31,6 +31,34 @@ module.exports = {
...
+        generateCacheKey(frame) {
+            const stringify = require('json-stable-stringify');
+            const keyData = {
+                include: frame.options?.include?.split(',').sort(),
+                filter: frame.options?.filter,
+                fields: frame.options?.fields?.split(',').sort(),
+                formats: frame.options?.formats?.split(',').sort(),
+                limit: frame.options?.limit,
+                order: frame.options?.order,
+                page: frame.options?.page || 1,
```
naz, line 43: Interesting one. Should we consider setting a default for the `options.page` on some other layer to avoid the `||` condition here? 🤔

ghost/api-framework/lib/pipeline.js
allouis: Yup! And can also use the stable-json-stringify here too

ghost/core/core/server/api/endpoints/posts-public.js
allouis: Potentially yeah, but it's tricky because this gets run before most other layers.

ghost/core/core/server/api/endpoints/posts-public.js
```diff
@@ -31,6 +31,34 @@ module.exports = {
...
+                collection: frame.options?.collection
+            };
+
+            if (frame.options?.context?.member) {
+                keyData.auth = {
+                    free: frame.options?.context?.member.status === 'free',
+                    tiers: frame.options?.context?.member.products?.map((product) => {
+                        return product.slug;
+                    }).sort(),
+                    labels: frame.options?.context.member.labels?.map((label) => {
```
naz, line 54: Do labels affect content gating?

ghost/api-framework/lib/pipeline.js
naz: Agreed!

ghost/core/core/server/api/endpoints/posts-public.js
allouis: Nope! As discussed in slack - have removed :)

ghost/core/core/server/api/endpoints/posts-public.js
allouis: It doesn't really work unless we shift lots of things round, and start running serialisation before cache key generation. Leaving as-is for now

PR #17975, opened 2023-09-05
Added logic to update the recommendations_enabled setting on add/delete

sagzy:
closes https://github.com/TryGhost/Product/issues/3799

- the recommendations_enabled setting is updated when a recommendation is created or deleted. It's enabled as soon as there is at least one recommendation in the database
- the recommendations_enabled  setting exists to avoid fetching the recommendation count from the database directly in themes. The setting is cached and doesn't need a read every time from the database


ghost/recommendations/src/RecommendationService.ts
```diff
@@ -16,22 +21,40 @@ export class RecommendationService {
...
 
+    async updateRecommendationsEnabledSetting(recommendations: Recommendation[]) {
+        const expectedSetting = (recommendations.length > 0).toString();
+        const currentSetting = this.settingsService.read('recommendations_enabled');
+
+        if (currentSetting && currentSetting === expectedSetting) {
+            return;
+        }
+
+        await this.settingsService.edit([{key: 'recommendations_enabled', value: expectedSetting}], {context: {}});
```
SimonBackx, line 55: We should pass in a context of `{internal: true}` here (because we should block writes to this key if it is not an internal write), but I would abstract that away in the wrapper service instead and simplify the methods that are passed to the RecommendationService

sagzy: Thanks! Refactored in https://github.com/TryGhost/Ghost/pull/17975/commits/43d1781e7d86b2a5a50d754729208135dd805f50

PR #17983, opened 2023-09-05
🐛 Fixed incorrect rendering of links following line breaks in beta editor

kevinansfield:
closes https://github.com/TryGhost/Product/issues/3827

- links following a line break in the editor were being rendered before the line break when previewing/publishing
- bumps Koenig packages which includes relevant fix in `@tryghost/kg-lexical-html-renderer`


daniellockyer:
This was merged and released as part of 5.61.2: https://github.com/TryGhost/Ghost/releases/tag/v5.61.2

PR #17994, opened 2023-09-06
Added checks for subscription to multiple newsletter through custom sign up form

vershwal:
refs https://github.com/TryGhost/Product/issues/3810


ghost/members-api/lib/repositories/MemberRepository.js
```diff
@@ -281,7 +282,29 @@ module.exports = class MemberRepository {
             memberStatusData.status = 'comped';
         }
 
-        // Subscribe member to default newsletters
+        //checks for custom signUp forms
+        if (memberData.type === 'subscribe' && memberData.newsletters) {
+            const allValidNewsletters = await this._newslettersService.browse({limit: 'all'});
```
mike182uk, line 287: Instead of querying for all newsletters, could we just query for the specific newsletter we are looking for with a `filter` param? i.e:

```js
options.filter = 'status:active+id:<ID_OF_NEWSLETTER>'

await this._newslettersService.browse(options);

File: ghost/members-api/lib/members-api.js [hunk 2]
```diff
@@ -264,7 +264,7 @@ module.exports = function MembersAPI({
             }
         }
 
-        const newMember = await users.create({name, email, labels, newsletters, attribution, geolocation});
+        const newMember = await users.create({name, email, labels, newsletters, attribution, geolocation, type});
```
mike182uk, line 267: Im not sure about passing `type` to the repository here - `type` seems to be an API / request specific thing. Why do we need to pass this? Wouldn't `newsletters` be enough?

ghost/members-api/lib/repositories/MemberRepository.js
```diff
@@ -18,7 +18,8 @@ const messages = {
     productNotFound: 'Could not find Product {id}',
     bulkActionRequiresFilter: 'Cannot perform {action} without a filter or all=true',
     tierArchived: 'Cannot use archived Tiers',
-    invalidEmail: 'Invalid Email'
+    invalidEmail: 'Invalid Email',
+    invalidNewsLetterId: 'Cannot subscribe to invalid newsletter {id}'
```
mike182uk, line 22:
```suggestion
    invalidNewsletterId: 'Cannot subscribe to invalid newsletter {id}'
```

(`newsletter` is one word so we don't need to capitalise `Letter` 👍)

ghost/members-api/lib/repositories/MemberRepository.js
```diff
@@ -281,7 +282,29 @@ module.exports = class MemberRepository {
             memberStatusData.status = 'comped';
         }
 
-        // Subscribe member to default newsletters
+        //checks for custom signUp forms
+        if (memberData.type === 'subscribe' && memberData.newsletters) {
+            const allValidNewsletters = await this._newslettersService.browse({limit: 'all'});
+            const allValidNewslettersIds = allValidNewsletters.map(newsletter => newsletter.id);
+
+            if (memberData.newsletters) {
```
mike182uk, line 290: I think this conditional is redundant - We already check `memberData.newsletters` is `truthy` in the parent conditional

ghost/members-api/lib/repositories/MemberRepository.js
```diff
@@ -281,7 +282,29 @@ module.exports = class MemberRepository {
...
+                const subscribedNewsletterIds = memberData.newsletters.map(newsletter => newsletter.id);
+                const invalidIds = subscribedNewsletterIds.filter(id => !allValidNewslettersIds.includes(id));
+                if (invalidIds.length > 0) {
+                    throw new errors.BadRequestError({message: tpl(messages.invalidNewsLetterId, {id: invalidIds})});
+                } else if (memberData.newsletters.length > 0) {
```
mike182uk, line 295: I don't think we need the `else if` here - If `invalidIds.length > 0` then an error is thrown which short circuits the rest of the function

vershwal: Yes, makes sense!

ghost/members-api/lib/members-api.js
vershwal: Hmm...I passed `type` because I have added [these checks](https://github.com/TryGhost/Ghost/blob/637f5c927c292f30852763914e22cfa2c715f28f/ghost/members-api/lib/repositories/MemberRepository.js#L286) and we don't want this validation for the `sign-up` flow through portal.

ghost/members-api/lib/repositories/MemberRepository.js
vershwal: aaah...I was exploring this options.filter yesterday. But couldn't find a way to do IN query. 
Is there a way to make an IN query using options.filter? 
Basically this: `SELECT * FROM newsletters WHERE id IN ('asdf3456788dfgvh', 'fdghv4567dftgy');` 

Because otherwise if I make a query using 'status:active+id:<ID_OF_NEWSLETTER>' I'll have to do a separate query for all the subscribed newsletters and that's too many unnecessary db calls when we can get the data in one call.

ghost/members-ssr/lib/members-ssr.js
```diff
@@ -201,43 +201,51 @@ class MembersSSR {
...
-        }
 
-        this._setSessionCookie(req, res, member.email);
+            this._setSessionCookie(req, res, member.email);
 
-        return member;
+            return member;
+        } catch (err) {
+            if (err.errorType === 'BadRequestError') {
+                return Promise.reject(new BadRequestError({
```
vershwal, line 244: This is handled [here](https://github.com/TryGhost/Ghost/blob/637f5c927c292f30852763914e22cfa2c715f28f/ghost/core/core/server/services/members/middleware.js#L285).
The behavior is that, when an invalid id is passed, do a standard 302 redirect to the homepage, with success=false. 
This is different from the 400 error we discussed in the tech spec but this felt more standard depending on how we are handling `BadRequestError` currently. 
What are your thoughts? Does this approach look fine?

image

    
image

ghost/members-api/lib/repositories/MemberRepository.js
mike182uk: > I'll have to do a separate query for all the subscribed newsletters 

Remember for the 🛹 we are only interested in a single newsletter

ghost/members-ssr/lib/members-ssr.js
mike182uk: I'm not sure Im following - Why do we need to make a change to `exchangeTokenForSession` for the work we are doing? I would not have expected to need to touch this area of the codebase... The diff is a bit hard to grok, but it looks like you have just wrapped the contents of the function in a try..catch?

ghost/members-api/lib/members-api.js
mike182uk: Is there not another way that we can achieve this without explicitly passing `type` to the repository? Can we not just check for the presence of `memberData.newsletters` and then validate the contents?

vershwal: we have memberData.newsletters in both `signup` and `subscribe` type. Let me think on this one again.

ghost/members-api/lib/repositories/MemberRepository.js
vershwal: Simplified this for 🛹

ghost/members-api/lib/members-api.js
mike182uk: whats the problem with them being in both flows? Is it because you are only wanting to validate in one of the flows? If so, what is the harm in validating in both flows?

ghost/members-api/lib/repositories/MemberRepository.js
```diff
@@ -281,7 +282,18 @@ module.exports = class MemberRepository {
...
 
-        // Subscribe member to default newsletters
+        //checks for custom signUp forms
+        if (memberData.newsletters && memberData.newsletters.length > 0) {
+            const savedNewsletter = await this._newslettersService.browse({filter: `id:'${memberData.newsletters[0].id}'`});
+            if (savedNewsletter.length === 0) {
+                throw new errors.BadRequestError({message: tpl(messages.invalidNewsletterId, {id: memberData.newsletters[0].id})});
+            }
+            if (savedNewsletter[0].status === 'archived') {
+                memberData.newsletters = [];
```
vershwal, line 292: Doing this is important because otherwise it was still subscribing to the archived newsletter.

PR #18008, opened 2023-09-07
Optimized PostEditedEvent processing in collections

naz:

- When PostEditedEvent data contains no visible changes we can skip the matching collections update process altogether. Each call to `updatePostInMatchingCollections` creates a transaction in addition to fetching all collections. There's no need to process anything when there are no relevant changes in the post edit!

---

### <samp>🤖 Generated by Copilot at f113a92</samp>

Optimized collections service and added test for post edit events. Used `lodash` library for data comparison and added it as a dependency in `package.json`.

### DIFF
--- a/ghost/collections/src/CollectionsService.ts
+++ b/ghost/collections/src/CollectionsService.ts
@@ -1,5 +1,6 @@
 import logging from '@tryghost/logging';
 import tpl from '@tryghost/tpl';
+import isEqual from 'lodash/isEqual';
 import {Knex} from "knex";
 import {
     PostsBulkUnpublishedEvent,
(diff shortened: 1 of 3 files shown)


ghost/collections/src/CollectionsService.ts
```diff
@@ -183,6 +184,10 @@ export class CollectionsService {
         });
 
         this.DomainEvents.subscribe(PostEditedEvent, async (event: PostEditedEvent) => {
+            if(isEqual(event.data.current, event.data.previous)) {
```
allouis, line 187: Can we make this more specific? I think we just need to check `published_at`, `featured` and `tags`!

naz: The fields on the current/previous are pretty [limited already](https://github.com/TryGhost/Ghost/blob/78e2cb0c284dceca3803700fa31bff559dc41213/ghost/collections/src/events/PostEditedEvent.ts#L1-L19). Or do you mean to future proof once we add more fields to the event itself?

allouis: Yeah I think just to be future proof/more specific 🤷  NABD but will save unnecessary updates

PR #18022, opened 2023-09-08
Added logging to collections event handlers

allouis:
This will help us with debugging issues we're seeing with large sites

### <samp>🤖 Generated by Copilot at fbd7308</samp>

This change adds error handling to the `CollectionsService` class, which manages the collections of posts in Ghost. It prevents unhandled promise rejections and logs the errors when domain events affect the collections. It is part of a pull request that fixes several collections-related issues.

### DIFF
--- a/ghost/collections/src/CollectionsService.ts
+++ b/ghost/collections/src/CollectionsService.ts
@@ -177,7 +177,7 @@ export class CollectionsService {
             try {
                 await this.removePostFromAllCollections(event.id);
             } catch (err) {
-                logging.error({err});
+                logging.error({err, message: 'Error handling PostDeletedEvent'});
             }
         });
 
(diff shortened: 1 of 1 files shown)


ghost/collections/src/CollectionsService.ts
```diff
@@ -174,47 +174,83 @@ export class CollectionsService {
     subscribeToEvents() {
         this.DomainEvents.subscribe(PostDeletedEvent, async (event: PostDeletedEvent) => {
             logging.info(`PostDeletedEvent received, removing post ${event.id} from all collections`);
-            await this.removePostFromAllCollections(event.id);
+            try {
+                await this.removePostFromAllCollections(event.id);
+            } catch (err) {
+                logging.error({err});
```
naz, line 180: I think in most instances we don't use this syntax 🤔  Just the `logging.error(err)` or wrapping it in a GhostError is the way to log

PR #18030, opened 2023-09-08
Added support for selectable newsletters in custom sign up-forms

vershwal:
refs https://github.com/TryGhost/Product/issues/3836


apps/portal/src/data-attributes.js
```diff
@@ -25,7 +25,9 @@ export function formSubmitHandler({event, form, errorEl, siteUrl, submitHandler}
 
     let newsletterInputs = event.target.querySelectorAll('input[data-members-newsletter]') || [];
     for (let i = 0; i < newsletterInputs.length; ++i) {
-        newsletters.push({id: newsletterInputs[i].value});
+        if (newsletterInputs[i].checked || newsletterInputs[i].selected) {
+            newsletters.push({id: newsletterInputs[i].value});
+        }
```
mike182uk, line 30: I think we should just be able to expand the query selector instead of doing this:

```js
let newsletterInputs = event.target.querySelectorAll('input[type=hidden][data-members-newsletter], input[type=checkbox][data-members-newsletter]:checked, input[type=radio][data-members-newsletter]:checked') || [];
```

mike182uk (approved):

PR #18057, opened 2023-09-11
Fixing leaking properties from Members Admin API

naz:
closes https://github.com/TryGhost/Arch/issues/87

@allouis the main changes are in "Trimmed down Members Admin API response data" and "Added newsletter payload serializer" commits. Both are not big changes but given this is an API with almost no test coverage until now would love an extra 👀  on it!

### <samp>🤖 Generated by Copilot at d9a296e</samp>

This pull request adds newsletter support to the members API and updates the relevant tests. It allows the API to query and return members based on their newsletter preferences and tiers. It also modifies the `members.test.js` file to reflect the new `newsletters` field in the member objects.


ghost/core/test/e2e-frontend/members.test.js
```diff
@@ -488,6 +473,43 @@ describe('Front-end members behavior', function () {
...
+                memberData.should.not.have.property('id');
+                memberData.newsletters.should.have.length(1);
+
+                // @NOTE: this should be a snapshot test not code
+                Object.keys(memberData.newsletters[0]).should.have.length(4);
+                memberData.newsletters[0].should.have.properties([
+                    'id',
+                    'name',
+                    'description',
+                    'sort_order'
```
allouis, line 509: Should the frontend include `sort_order`?

allouis (commented):
Looks good to me! One Q

ghost/core/core/server/services/members/utils.js
```diff
@@ -4,6 +4,7 @@ function formatNewsletterResponse(newsletters) {
     });
 }
 
+module.exports.formatNewsletterResponse = formatNewsletterResponse;
```
naz, line 7: This is the meat of the change. We use the default newsletter formatter for the rest of the member endpoints.

ghost/core/test/e2e-frontend/members.test.js
naz: yup, needed for the frontend [here](https://github.com/TryGhost/Ghost/blob/2554c0df317f84a33238343a4ad78e46b8a307db/apps/portal/src/utils/helpers.js#L519). not sure why, but that is out of scope for what we try to achieve here.

PR #18071, opened 2023-09-12
Added newsletter in get helper

vershwal:
refs https://github.com/TryGhost/Product/issues/3859


mike182uk:
@vershwal This doesn't seem to be working when I've tried it in a template:

```
{{#get "newsletters" as |newsletter|}}
  {{#foreach newsletter}}
    {{name}}
  {{/foreach}}
{{/get}}
```

ghost/core/core/frontend/helpers/foreach.js
```diff
@@ -31,6 +31,12 @@ module.exports = function foreach(items, options) {
         if (Object.values(items).length > 0 && checks.isPost(Object.values(items)[0])) {
             visibility = visibility || 'all';
         }
+    } if (_.isArray(items) && items.length > 0 && checks.isNewsletter(items[0])) {
```
mike182uk, line 34:
```suggestion
    } 
    
    if (_.isArray(items) && items.length > 0 && checks.isNewsletter(items[0])) {
```

Having the `if` on the same line made it look like it was part of the `else if` which it is not

ghost/core/core/frontend/services/data/checks.js
```diff
@@ -3,6 +3,11 @@ function isPost(jsonData) {
         Object.prototype.hasOwnProperty.call(jsonData, 'title') && Object.prototype.hasOwnProperty.call(jsonData, 'slug');
 }
 
+function isNewsletter(jsonData) {
+    return Object.prototype.hasOwnProperty.call(jsonData, 'name') &&
+        Object.prototype.hasOwnProperty.call(jsonData, 'subscribe_on_signup') && Object.prototype.hasOwnProperty.call(jsonData, 'slug');
```
mike182uk, line 8: Maybe we could use less overlapping named fields here? How about `subscribe_on_signup`, `status` and `visibility`?

ghost/core/test/unit/frontend/helpers/foreach.test.js
```diff
@@ -282,6 +290,14 @@ describe('{{#foreach}} helper', function () {
             ]
         };
 
+        const newsletterArrayHashWithVisibility = {
```
mike182uk, line 293: How about we co-locate `newsletterObjectHashWithVisibility` and `newsletterArrayHashWithVisibility` with the test seeming though its unlikely they will be used in other tests?

ghost/core/test/unit/frontend/helpers/foreach.test.js
```diff
@@ -402,6 +402,27 @@ describe('{{#foreach}} helper', function () {
...
+                    third: {name: 'third', visibility: 'paid', slug: 'third', subscribe_on_signup: false}
+                }
+            };
+            const newsletterArrayHashWithVisibility = {
+                newsletters: [
+                    {name: 'first', visibility: 'members', slug: 'first', subscribe_on_signup: true},
+                    {name: 'second', visibility: 'members', slug: 'second', subscribe_on_signup: true},
+                    {name: 'third', visibility: 'paid', slug: 'third', subscribe_on_signup: false}
+                ]
+            };
```
mike182uk, line 419: We don't need `slug` in the test data anymore right?

ghost/core/test/unit/frontend/services/data/checks.test.js
```diff
@@ -26,6 +27,14 @@ describe('Checks', function () {
         checks.isPage({title: 'Test', show_title_and_feature_image: true}).should.eql(true);
     });
 
+    it('isNewsletter', function () {
+        checks.isNewsletter({}).should.eql(false);
+        checks.isNewsletter({name: 'Test'}).should.eql(false);
+        checks.isNewsletter({name: 'Test', slug: 'test'}).should.eql(false);
```
mike182uk, line 33: I don't think we need this assertion

mike182uk (approved):
Couple of minor tweaks @vershwal but good to 🚢

vershwal:
Done 🎉

Issue #12032, opened 2020-07-09
Stop handling SQLite too many variables error


ErisDS:
## Problem

People using SQLite3 in production sometimes fall foul of SQLite's 999 variable limit. We've been thinking about how to refactor to cope with this, and we have a few bits and pieces in the codebase, but actually SQLite fixed it in their 3.32 version: https://www.sqlite.org/releaselog/3_32_0.html

Hopefully Node's SQLite project will come bundled with 3.32 by default soon as it's already in master.

**Appetite:** 0.5 day

## Solution:

1. We should remove all workarounds that we have for this in the codebase, they are code we don't need to maintain anymore.
E.g. https://github.com/TryGhost/Ghost/commit/08f54d2f57f35d5c20bdd824ffef6eaac9bb9fe9
E.g. https://github.com/TryGhost/Ghost/commit/c7012935144db53d3c07e3079167de5278631c95
E.g. https://github.com/TryGhost/Ghost/commit/a29ac2691a1ae79e93ca65f46127e9fc378881c2
2. We should _maybe_ look at adding an error handler that catches the error and converts it to a nicely formatted Ghost error telling people to upgrade to SQLite 3.32 - but only if this can be done quickly and easily in one place.
> SQLITE_ERROR: too many SQL variables
3. Maybe mention this as a minimum requirement in our docs

naz:
> Hopefully Node's SQLite project will come bundled with 3.32 by default soon as it's already in master.

@ErisDS saw [your conversation](https://github.com/mapbox/node-sqlite3/issues/1292#issuecomment-668494366) in the thread asking for maintenance help for node-sqllte3 and having commit rights to the repo. The [5.0.1 version was merged](https://github.com/mapbox/node-sqlite3/pull/1366) on GH but never published to npm because the build failed. Would be cool if you could influence a proper release of 5.0.1 (also happy to help in any way possible myself). 

To do a proper cleanup we'd need new binaries, so that Ghost could use worker_threads instead of [current workaround](https://github.com/TryGhost/Ghost/pull/12431) with child processes.

naz:
Tried out fresh 5.0.1 build of node-sqlite3. Unfortunately it does not compile the binaries locally (my environment is Ubuntu 20.04, with Node v12, and python 3). The issue is this one described in [node-sqlite3 here](https://github.com/mapbox/node-sqlite3/issues/1415) (outdated node-gyp dependency).

[...]

PR #18081, opened 2023-09-12
Handled CollectionPost relations manually

allouis:
bookshelf-relations was generating tonnes of select queries from the posts table in order to update the relations. I spent all of yesterday trying to get it working with a hasMany relationship, but couldn't manage it, so we've instead switched to handling the relations manually. Aside from the naming of temporary variables, I quite like this approach, it isn't overly complex, and it allows for customisations in the future such as an inverted sort_order.

This logic could be pulled out into a util (not bookshelf plugin) where it could be used explicitly, but with the complexity hidden, we'll see ig.

---

### <samp>🤖 Generated by Copilot at 239f2b7</samp>

This pull request refactors the collections service and the collection model to use a separate relation model for collection-post associations. It also updates the tests and the service wrapper to match the new implementation. This is done to improve the performance and reliability of the collections feature.

### DIFF
--- a/ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
+++ b/ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
@@ -1,6 +1,7 @@
 const logger = require('@tryghost/logging');
 const Collection = require('@tryghost/collections').Collection;
 const sentry = require('../../../shared/sentry');
+const {default: ObjectID} = require('bson-objectid');
 /**
  * @typedef {import('@tryghost/collections/src/CollectionRepository')} CollectionRepository
  */
@@ -10,8 +11,10 @@ const sentry = require('../../../shared/sentry');
  */
 module.exports = class BookshelfCollectionsRepository {
     #model;
-    constructor(model) {
+    #relationModel;
+    constructor(model, relationModel) {
         this.#model = model;
+        this.#relationModel = relationModel;
     }
 
     async createTransaction(cb) {
@@ -105,6 +108,14 @@ module.exports = class BookshelfCollectionsRepository {
             await this.#model.destroy({id: collection.id});
             return;
         }
+        if (!options.transaction) {
+            return this.createTransaction((transaction) => {
+                return this.save(collection, {
+                    ...options,
+                    transaction
+                });
+            });
+        }
         const data = {
             id: collection.id,
             slug: collection.slug,
@@ -113,7 +124,6 @@ module.exports = class BookshelfCollectionsRepository {
             filter: collection.filter,
             type: collection.type,
             feature_image: collection.featureImage || null,
-            posts: collection.posts.map(postId => ({id: postId})),
             created_at: collection.createdAt,
             updated_at: collection.updatedAt
         };
@@ -122,7 +132,8 @@ module.exports = class BookshelfCollectionsRepository {
             {id: data.id},
             {
                 require: false,
-                transacting: options.transaction
+                transacting: options.transaction,
+                withRelated: ['collectionPosts']
             }
         );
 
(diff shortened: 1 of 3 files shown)


allouis:
I wanted to completely delete the `posts` relation - but it FUCKS up the tests big time. I didn't dig into it yet, but I'm guessing it's fixtures related.

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
```diff
@@ -105,6 +108,14 @@ module.exports = class BookshelfCollectionsRepository {
             await this.#model.destroy({id: collection.id});
```
naz, line 108: We should put the destroy behind a transaction too 🤔

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
```diff
@@ -122,19 +132,68 @@ module.exports = class BookshelfCollectionsRepository {
...
         );
 
         if (!existing) {
             await this.#model.add(data, {
                 transacting: options.transaction
             });
+            const thingsToInsert = collection.posts.map((postId, index) => {
+                return {
+                    id: (new ObjectID).toHexString(),
+                    sort_order: collection.type === 'manual' ? index : 0,
```
naz, line 147: Thinking out loud question, does the sort_order assignment type of logic belong to a repository or should we expect it to be handled on the entity level? I'm thinking the latter is more suitable, but then is sorting something entity should be even concerned about?

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
```diff
@@ -122,19 +132,68 @@ module.exports = class BookshelfCollectionsRepository {
...
+                    collection_id: collection.id,
+                    post_id: postId
+                };
+            });
+            if (thingsToInsert.length > 0) {
```
naz, line 152: I'd go as far as move this block higher and check for `if (collection.posts.length > 0)`

naz: Coming from a place where reading a code would be like: oh do this whole piece of logic if there's anything to it and mapping is part of this logic imo.

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
allouis: `sort_order` is a persistence concern - it's just an implementation detail so that we can get the same entity back after saving!

Maybe an automatic collection should have a `Set` of post ids instead of `Array` so that it isn't implied that it should be ordered 🤔

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
```diff
@@ -122,19 +132,68 @@ module.exports = class BookshelfCollectionsRepository {
...
+                }
+            }
+
+            const thingsToInsert = things.filter(thing => thing.id !== null);
+
+            if (thingsToInsert.length > 0) {
+                await this.#relationModel.query().insert(thingsToInsert).transacting(options.transaction);
+            }
+            if (thingsToDelete.length > 0) {
+                await this.#relationModel.query().delete().whereIn('id', thingsToDelete).transacting(options.transaction);
```
naz, line 195: Vaguely reminds of the problems we used to deal with in SQLite where the "WHERE IN (...)" was not performing due to a [query parameter limit](https://github.com/TryGhost/Ghost/issues/12032). Good those days are over 😅

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
allouis: The checks are around the calls to the database because those will throw if passed empty arrays. Else where we've filtered the arrays so I want to keep the check close to where they're used to avoid trying to insert 0 values

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
allouis: Good shout

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
```diff
@@ -122,19 +132,68 @@ module.exports = class BookshelfCollectionsRepository {
...
+                    sort_order: collection.type === 'manual' ? index : 0,
+                    collection_id: collection.id,
+                    post_id: postId
+                };
+            });
+
+            const thingsToDelete = [];
+
+            if (collection.type === 'manual') {
+                await this.#relationModel.query().delete().where('collection_id', collection.id).transacting(options.transaction);
```
naz, line 173: This is fine too. Unless we start getting manual collections with thousands of posts we should not make the logic complicated 👌

naz (approved):
LFG 🚀

PR #18096, opened 2023-09-13
✨ Added `Source` as the new default theme

cmraible:
refs TryGhost/Product#3510

- Added `TryGhost/Source` as a submodule in `ghost/core/content/themes` so `Source` will ship with Ghost (along with `Casper`)
- With this change, new installs will use `Source` as the default theme. Existing sites will have `Source` installed, but not activated, as this is a large change and we don't want to drastically change existing sites without warning. Users can upgrade to use `Source` simply by clicking 'Activate' in design settings.
- Updated protections to prevent users from uploading their own conflicting version of `Source`


.gitmodules
```diff
@@ -2,3 +2,6 @@
 	path = ghost/core/content/themes/casper
 	url = ../../TryGhost/Casper.git
 	ignore = all
+[submodule "ghost/core/content/themes/source"]
+	path = ghost/core/content/themes/source
+	url = git@github.com:TryGhost/Source.git
```
daniellockyer, line 7: Can you change this to the following:
```suggestion
	url = ../../TryGhost/Source.git
	ignore = all
```

I don't remember why, but that URL fixes some weird git issues I had before

And `ignore = all` helps stop `source` show up as unstaged in git

daniellockyer:
@cmraible We should also include the actual git submodule addition (?) in this. You'll need to initialize it locally and then `git add ghost/core/core/content/themes/source`

.gitmodules
cmraible: Updated ✅

cmraible:
@daniellockyer my bad, I thought I had done that but I think the .gitignore blocked me.

When running `git add ghost/core/content/themes/source` I was getting an error (I think from a git hook): 'Aborting commit due to submodule update. Please unstage the submodule(s) and commit again.'

Got around this by using `git commit --no-verify ...` and that seemed to work, but just wanted to give a heads up in case that borked anything else up. That was the only change in the commit, so I don't think it would have bypassed any important tests, but just wanted to double check with you.

daniellockyer:
@cmraible Yeah that's totally cool. That check is just there to stop people committing the submodule in day-to-day operations

I tend to do `CI=1 git ...` to bypass it myself 😀

PR #18114, opened 2023-09-13
Added core content API integration fixture

kevinansfield:
refs https://github.com/TryGhost/Product/issues/3874

- the new collections card needs to access the Content API rather than the Admin API in order to show the card as it will appear on the front-end but we don't have a default integration that can be fetched via the Admin API for Admin to use when fetching from the Content API
- adds a new "Ghost Core Content API" integration with the `core` type so that it can be read via the `/admin/integrations/` endpoint and used in Admin to make Content API requests

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.63/2023-09-13-13-34-11-add-ghost-core-content-integration-key.js
+++ b/ghost/core/core/server/data/migrations/versions/5.63/2023-09-13-13-34-11-add-ghost-core-content-integration-key.js
@@ -22,7 +22,7 @@ const addIntegrationContentKey = async (knex, integration) => {
 
     if (!existingIntegration) {
         throw new InternalServerError({
-            message: `Count not find "${integration.name}" integration`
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/data/migrations/versions/5.63/2023-09-13-13-34-11-add-ghost-core-content-integration-key.js
```diff
@@ -0,0 +1,87 @@
...
+const addIntegrationContentKey = async (knex, integration) => {
+    const message = `Adding "${integration.name}" integration content key`;
+
+    const existingIntegration = await knex('integrations').select('id').where({
+        slug: integration.slug
+    }).first();
+
+    if (!existingIntegration) {
+        throw new InternalServerError({
+            message: `Count not find "${integration.name}" integration`
```
55sketch, line 25: typo "Count"

9larsons (approved):

PR #18120, opened 2023-09-13
🐛 Fixed portal showing incorrect expiry date for comped subscription

mike182uk:
refs https://github.com/TryGhost/Product/issues/3875

When a member had a comped subscription, the portal was showing an incorrect expiry date. This was because the `expiry_date` was being set to the `created_at` date of the subscription, rather than the `expiry_date` of the comped subscription

### DIFF
--- a/ghost/members-api/lib/services/MemberBREADService.js
+++ b/ghost/members-api/lib/services/MemberBREADService.js
@@ -99,7 +99,7 @@ module.exports = class MemberBREADService {
                     default_payment_card_last4: '****',
                     cancel_at_period_end: false,
                     cancellation_reason: null,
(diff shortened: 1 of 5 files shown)


ghost/members-api/lib/services/MemberBREADService.js
```diff
@@ -99,7 +99,7 @@ module.exports = class MemberBREADService {
                     default_payment_card_last4: '****',
                     cancel_at_period_end: false,
                     cancellation_reason: null,
-                    current_period_end: moment(startDate).add(1, 'year'),
```
cmraible, line 102: amazing

cmraible (approved):
Nice find, lol! Tested this locally and it corrected the issue with portal showing the wrong expiry data

PR #18134, opened 2023-09-14
🏗️ Added constraint to collections_posts table

naz:

- This constraint will make sure each collection contains only one entry of the post
- This came out as a part of deeper investigation into OOM Ghost instance crashes

---

### <samp>🤖 Generated by Copilot at 0f477a0</samp>

This pull request adds a unique constraint on the `collections_posts` table to ensure that each post can only belong to one collection. It includes a new migration file, an updated schema definition, and a revised integrity test.


naz:
These changes will most likely end with next major. Tracked [here](https://github.com/TryGhost/Product/issues/3914)

PR #18140, opened 2023-09-14
Fixed transaction lock from rendering pages with collections when saving

kevinansfield:
no issue

- rendering with a collection card means the lexical renderer fetches post data, however when saving a post we're inside of a `FOR UPDATE` transaction meaning the fetches that the renderer attempts stall out due to a transaction lock
- all bookshelf events like `onSaving`, `onSaved` etc end up being wrapped up in a transaction due to the post model's custom `edit` and `add` methods which enforces a transaction around the bookshelf method meaning we can't use the hook methods to handle the render+update
- added a new `rerenderIfNeeded` method that is explicitly called once the add/edit transactions have completed, this checks if we have lexical content that hasn't been rendered then renders outside of a transaction before updating the page's html field directly via knex to avoid triggering extra bookshelf hooks/events
- updated `onSaving` method to set `html` to `null` rather than rendering when we're saving a page so the new `rerenderIfNeeded` methods conditions are met to render after the initial save

### DIFF
--- a/ghost/core/core/server/models/post.js
+++ b/ghost/core/core/server/models/post.js
@@ -1335,12 +1335,15 @@ Post = ghostBookshelf.Model.extend({
         // pages are rendered after saving so data can be fetched outside of a transaction
         if (model.get('lexical') !== null && model.get('html') === null) {
             const html = await lexicalLib.render(model.get('lexical'));
+            const plaintext = htmlToPlaintext.excerpt(html);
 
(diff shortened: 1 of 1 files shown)


kevinansfield:
Closing in favour of https://github.com/TryGhost/Ghost/pull/18148 as fixing the underlying transaction issue is preferable to this complicated workaround.

PR #18188, opened 2023-09-16
🎨Enabled more detailed error message on creating insecure passwords

qwerzl:
no issue

- [x] There's a clear use-case for this code change, explained below:

Ghost's 'Sorry, you cannot use an insecure password.' message has been quite confusing as Ghost tries to enforce a much stricter password rule than others. In my case, I have a long password which got rejected by Ghost and eventually found out it was the keyword 'ghost'. This commit points out what rule the password is conflicting with.

---

### <samp>🤖 Generated by Copilot at da4eda9</samp>

Enhanced password validation feedback in the admin app. The file `password.js` now shows more informative error messages when the password does not meet the security requirements.


markstos:
I like this change. It's important that user's understand an error message well enough to know how they might fix the related issue.

WilliamOConnell:
This tripped me up when trying to use Ghost for the first time, it kept rejecting my (very long/complex) passwords because they contained the string `ghost` and it took a lot of trial-and-error to figure out what was going on. It was a really bad first impression.

Personally I would just remove that check altogether, but this change would be a huge improvement. What needs to happen to get this merged? Looks like the tests haven't run, is there some way to re-trigger that?

markstos (approved):
All changes here are improvements. Thanks.

PR #18194, opened 2023-09-18
Optimized collection repository queries

naz:

- Creating bookshelf models for each collection_post relation created a massive overhead. On a dataset with 500k collections_posts records the timing was roughly 7s comparing to 810ms after the optimization.

---

### <samp>🤖 Generated by Copilot at 66d4782</samp>

Optimize memory and performance of collections fetching by querying post ids only. Refactor `BookshelfCollectionsRepository` to use a new private method and modify existing methods.

### DIFF
--- a/ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
+++ b/ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
@@ -66,6 +66,15 @@ module.exports = class BookshelfCollectionsRepository {
      * @returns {Promise<Array<{post_id: string}>>}
      */
     async #fetchCollectionPostIds(collectionId, options = {}) {
+        if (!options.transaction) {
+            return this.createTransaction((transaction) => {
+                return this.#fetchCollectionPostIds(collectionId, {
+                    ...options,
+                    transaction
+                });
+            });
+        }
+
         return await this.#relationModel
             .query()
             .select('post_id')
--- a/ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
+++ b/ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
@@ -86,11 +86,20 @@ module.exports = class BookshelfCollectionsRepository {
      * @param {object} [options]
      * @param {string} [options.filter]
      * @param {string} [options.order]
+     * @param {string[]} [options.withRelated]
      * @param {import('knex').Transaction} [options.transaction]
      */
     async getAll(options = {}) {
+        const filteredOptions = Object.assign({}, options);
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
```diff
@@ -53,6 +53,26 @@ module.exports = class BookshelfCollectionsRepository {
...
+     * @param {Object} options bookshelf options
+     *
+     * @returns {Promise<Array<{post_id: string}>>}
+     */
+    async #fetchCollectionPostIds(collectionId, options = {}) {
+        return await this.#relationModel
+            .query()
+            .select('post_id')
+            .where('collection_id', collectionId)
+            .transacting(options.transaction);
```
allouis, line 73: I think this needs to go in a conditional based on if `options.transaction` is present

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
```diff
@@ -44,28 +46,69 @@ module.exports = class BookshelfCollectionsRepository {
...
+     *       The method could be further optimized to fetch posts for
+     *       multiple collections at a time.
+     *
+     * @param {string} collectionId collection to fetch post ids for
+     * @param {Object} options bookshelf options
+     *
+     * @returns {Promise<Array<{post_id: string}>>}
+     */
+    async #fetchCollectionPostIds(collectionId, options = {}) {
+        if (!options.transaction) {
```
allouis, line 74: Why does this _have_ to run in a transaction?

allouis (commented):
Approach looks good - minor comment about adding transaction overhead and if it's necessary!!

The query counts have changed (exciting!!) - do those numbers make sense to you?

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
naz: Good catch! The method is used across most of the fetch/save methods in the repository. But then It should be up to the calling method to decide if to create/pass in a transaction or not. Would be too much overhead to create a transaction for each getById or getBySlug call

naz: have removed this block 👌

naz:
> The query counts have changed (exciting!!) - do those numbers make sense to you?

Yes they did! And they did check out. We basically introduce 1 extra call for each collection fetch. When it's getAll - 2 extra calls for the built-in collections. 

We will have to optimize these calls at some point, but for now it seems like "good enough" of a solution to get us going.

PR #18262, opened 2023-09-21
[WIP] Virtual latest collection

allouis:
refs https://github.com/TryGhost/Arch/issues/95

🍦 
---

### <samp>🤖 Generated by Copilot at e011338</samp>

This pull request improves the performance and consistency of the collections feature by refactoring the `Collection` and `CollectionsService` classes and updating the corresponding test file. It also introduces a special case for the `latest` collection, which fetches all posts dynamically.

### DIFF
--- a/ghost/collections/src/CollectionsService.ts
+++ b/ghost/collections/src/CollectionsService.ts
@@ -560,12 +560,20 @@ export class CollectionsService {
         });
     }
 
-    async getById(id: string, options?: {transaction: Knex.Transaction}): Promise<Collection | null> {
-        return await this.collectionsRepository.getById(id, options);
+    async getById(id: string, options?: {transaction: Knex.Transaction}): Promise<CollectionDTO | null> {
+        const collection = await this.collectionsRepository.getById(id, options);
+        if (!collection) {
+            return null;
+        }
+        return this.toDTO(collection);
     }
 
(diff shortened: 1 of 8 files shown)


ghost/collections/src/CollectionsService.ts
```diff
@@ -144,6 +144,16 @@ export class CollectionsService {
                 sort_order: index
             }))
         };
+        if (collection.slug === 'latest') {
+            const allPosts = await this.postsRepository.getAll({ // TODO Support getting just ids
```
naz, line 148: We can drop down to a pure knex query straight away here to avoid the bookshelf pitfall again, right?

allouis: Needs to be inside the `PostRepository` implementation, but yeah

ghost/collections/src/CollectionsService.ts
```diff
@@ -594,18 +609,17 @@ export class CollectionsService {
     }
     async getCollectionsForPost(postId: string): Promise<CollectionDTO[]> {
         const collections = await this.collectionsRepository.getAll({
-            filter: `posts:${postId}`
+            filter: `posts:${postId},slug:latest`
```
naz, line 612: This is because the post is always in the 'latest' colleciton.

PR #18271, opened 2023-09-21
Ensure that UpdateCheck job always completes

sam-lord:
refs: https://github.com/TryGhost/Product/issues/3782

When Ghost runs a job, all exceptions must be handled so that the job can report "done" (as in finished) - regardless of whether the job has succeeded.

This patch is one solution, the other would be to introduce semantics for "failing" a job in a way which would also remove it from the queue, allowing Ghost to shut down / the worker to be terminated.

Here's the remaining issue, whilst it seems like Ghost is shutting down happily at this point, it has moved the shutdown time from the call to stop Bree into the call to process.exit(0) from the main thread. It seems like some network resources aren't being released by Node.js, causing a delay before the application can exit. I'm hopeful that the environment of Ghost (Pro) is going to be more efficient at tidying up these resources than it would have been with the original slow-down.
If nothing else, this now feels semantically correct in that Bree thinks the job is done, kills the worker and thinks that all the resources are tidied up.

Edit: This also contains an update to the latest version of Bree @ version 9.

---

### <samp>🤖 Generated by Copilot at 956a1de</samp>

Handle update check errors in worker script. The change adds error handling to `run-update-check.js` to avoid unhandled rejections and worker crashes, and to notify the parent process of the failure.

### DIFF
--- a/ghost/core/core/boot.js
+++ b/ghost/core/core/boot.js
@@ -410,7 +410,7 @@ async function initBackgroundServices({config}) {
     }
 
     const updateCheck = require('./server/update-check');
-    updateCheck.scheduleRecurringJobs();
+    await updateCheck.scheduleRecurringJobs();
 
     const milestonesService = require('./server/services/milestones');
     milestonesService.initAndRun();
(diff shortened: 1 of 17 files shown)


daniellockyer:
Going to close this for now - we can come back to it in the future

PR #18291, opened 2023-09-22
[WIP] Collection events

allouis:
🍦 

---

### <samp>🤖 Generated by Copilot at fa7c04c</samp>

The pull request adds domain events to the collections feature of Ghost. It introduces two new classes, `CollectionPostAdded` and `CollectionPostRemoved`, that represent the events of adding or removing posts from collections. It also modifies the `Collection` class, the `BookshelfCollectionsRepository` class, and the `CollectionsServiceWrapper` class to emit and handle these events. It updates the `./src/index.ts` file to export the new classes.

### DIFF
--- a/ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
+++ b/ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
@@ -204,45 +204,39 @@ module.exports = class BookshelfCollectionsRepository {
                 transacting: options.transaction
             });
 
-            const collectionPostsRelations = collection.posts.map((postId, index) => {
-                return {
-                    id: (new ObjectID).toHexString(),
-                    sort_order: collection.type === 'manual' ? index : 0,
-                    collection_id: collection.id,
-                    post_id: postId
-                };
-            });
-
-            const collectionPostRelationsToDeleteIds = [];
-
             if (collection.type === 'manual') {
+                const collectionPostsRelations = collection.posts.map((postId, index) => {
+                    return {
+                        id: (new ObjectID).toHexString(),
+                        sort_order: index,
+                        collection_id: collection.id,
+                        post_id: postId
+                    };
+                });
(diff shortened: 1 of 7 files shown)


ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
```diff
@@ -204,45 +204,39 @@ module.exports = class BookshelfCollectionsRepository {
...
+                        post_id: event.data.post_id
+                    };
+                });
 
-            if (missingCollectionPostsRelations.length > 0) {
-                await this.#relationModel.query().insert(missingCollectionPostsRelations).transacting(options.transaction);
-            }
-            if (collectionPostRelationsToDeleteIds.length > 0) {
-                await this.#relationModel.query().delete().whereIn('id', collectionPostRelationsToDeleteIds).transacting(options.transaction);
+                if (collectionPostsToDelete.length > 0) {
```
naz, line 234: Took me a while to figure out we have swapped  order for adding/deleting 😅

ghost/core/core/server/services/collections/BookshelfCollectionsRepository.js
```diff
@@ -201,46 +204,46 @@ module.exports = class BookshelfCollectionsRepository {
...
+                    await this.#relationModel.query().delete().where('collection_id', collection.id).whereIn('post_id', collectionPostsToDelete).transacting(options.transaction);
+                }
+                if (collectionPostsToInsert.length > 0) {
+                    await this.#relationModel.query().insert(collectionPostsToInsert).transacting(options.transaction);
+                }
             }
+
+            options.transaction.executionPromise.then(() => {
+                for (const event of collection.events) {
+                    this.#DomainEvents.dispatch(event);
```
naz, line 244: Why do we need to dispatch these? Nobody's listening, right? 📞

allouis: No but I think it is the correct thing to do, for when someone does want to listen!

PR #18304, opened 2023-09-22
Added recommendation_notifications column to users table

SimonBackx:
fixes https://github.com/TryGhost/Product/issues/3939


55sketch (approved):

apps/admin-x-settings/src/components/settings/general/UserDetailModal.tsx
```diff
@@ -201,6 +202,15 @@ const EmailNotificationsInputs: React.FC<UserDetailProps> = ({user, setUserData}
                         setUserData?.({...user, mention_notifications: e.target.checked});
                     }}
                 />}
+                {hasRecommendations && <Toggle
+                    checked={user.recommendation_notifications}
+                    direction='rtl'
+                    hint='Every time another site recommends your work'
```
sagzy, line 208: Would suggest: "Every time a site recommends you". But Peter's area + need to see whether we actually want to send an email _every time_, or not. So can come back to this later.

sagzy (approved):

PR #18334, opened 2023-09-25
Upgraded to latest version of @tryghost/request

sam-lord:
refs: https://github.com/TryGhost/Product/issues/3782

Latest version of request avoids using the OS-level name resolution system. This prevents long shutdowns from occurring where the OS is blocking the process from exiting cleanly.

The new version uses `got` 13, which has many breaking changes. Some of these are resolved within @tryghost/request (like response errors), but input parameters need to be changed.

---

### <samp>🤖 Generated by Copilot at f519026</samp>

This pull request migrates the Ghost core and related modules from using the `@tryghost/request` library to the `got` library for making HTTP requests. It updates the request options, headers, and response types to match the new API and conventions of `got`. It also updates the dependency version and the test assertions accordingly.

### DIFF
--- a/ghost/core/core/server/adapters/scheduling/scheduling-default.js
+++ b/ghost/core/core/server/adapters/scheduling/scheduling-default.js
@@ -274,21 +274,21 @@ SchedulingDefault.prototype._pingUrl = function (object) {
     const maxTries = 30;
 
     const options = {
-        timeout: requestTimeout,
-        method: httpMethod.toLowerCase(),
-        retry: 0,
-        headers: {
-            'Content-Type': 'application/json'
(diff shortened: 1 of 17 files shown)


ghost/core/test/e2e-api/admin/__snapshots__/pages.test.js.snap
```diff
@@ -1698,7 +1698,7 @@ We've crammed the most important information to help you get started with Ghost
 
 
 
-25 Sep 2023
```
daniellockyer, line 1701: @sam-lord There's a hardcoded date here - can you flag with Product to fix this?

PR #18360, opened 2023-09-26
AdminX: reroute settings-x to settings

peterzimon:
refs. https://github.com/TryGhost/Product/issues/3949

- before releasing we need to reroute all settings from `settings-x` to `settings`


.github/workflows/browser-tests.yml
```diff
@@ -77,8 +77,7 @@ jobs:
 
     - name: Build Admin
       if: env.ENVIRONMENT == 'browser-tests-local'
-      working-directory: ghost/admin
-      run: yarn build:dev
+      run: yarn workspace @tryghost/admin-x-settings build && yarn workspace ghost-admin build:dev
```
daniellockyer, line 80:
```suggestion
      run: yarn nx run ghost-admin:build:dev
```

daniellockyer: This is automatically configured to build admin-x-settings if we run it through NX

ghost/core/content/themes/source
daniellockyer, line 1: This needs to not be deleted 😀

binary-koan: I specifically told git to reset that change 😠 Thanks for pointing it out!

PR #18411, opened 2023-10-02
✨ Added "startsWith" check to the match helper

ramrami:
Added "startsWith" string comparison to the match helper. 

I think this simple check will unlock and simplify a lot of use cases for theme developers:

**Group logic when dealing with similar custom settings**:
```handlebars
{{!-- Suppose layout could be: "List full", "List narrow", "Grid 2 cols", "Grid 3 cols", "Grid masonry" --}}
{{#match @custom.layout "startsWith" "Grid"}} 
    has-grid
{{/match}}
```

**Check if a URL is external**:
```handlebars
{{^match (url absolute="true") "startsWith" @site.url }} 
    is-external
{{/match}}
```

**Simulate nested tags**:
```handlebars
{{#match slug "startsWith" "hash-parent-tag-" }} 
    {{> template-parent }}
{{/match}}
```

**Help implementing nested navigation**:
```handlebars
{{#match label "startsWith" "--" }} 
    is-child-nav-item
{{/match}}
```

**Add a badge to a navigation item**:
```handlebars
{{#match label "startsWith" "[NEW]" }} 
    <span>{{t "NEW"}}</span>
{{/match}}
```

---

---

### <samp>🤖 Generated by Copilot at 082c4f9</samp>

This pull request adds support for the `startsWith` operator to the `match` template helper, which allows matching data values against string prefixes. It also adds unit tests for the new operator in the `match.test.js` file.

### DIFF
--- a/ghost/core/core/frontend/helpers/match.js
+++ b/ghost/core/core/frontend/helpers/match.js
@@ -57,6 +57,9 @@ const handleMatch = (data, operator, value) => {
     case '<=':
         result = data <= value;
         break;
+    case 'startsWith':
+        result = _.isString(data) && _.isString(value) && data.startsWith(value);
+        break;
     default:
         result = data === value;
     }
--- a/ghost/core/test/unit/frontend/helpers/match.test.js
+++ b/ghost/core/test/unit/frontend/helpers/match.test.js
@@ -246,6 +246,38 @@ describe('Match helper', function () {
             }, hash);
         });
 
+        describe('Explicit Starts With', function () {
+            runTests({
+                // Using string values
+                '{{match empty "startsWith" ""}}': 'true',
+                '{{match empty "startsWith" " "}}': 'false',
+                '{{match string "startsWith" "Hello"}}': 'true',
+                '{{match string "startsWith" "World"}}': 'false',
+                '{{match string_true "startsWith" "tr"}}': 'true',
+                '{{match string_false "startsWith" "tr"}}': 'false',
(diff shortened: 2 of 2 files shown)


ramrami:
@ErisDS Any chance we can get this to upstream?

zeevox:
Rather disappointing to see how long this pull request has been ignored. I wholeheartedly agree that this simple addition unlocks many possibilities for Ghost themes. 

For example, for a newspaper site that uses `issue-123` tags it is incredibly useful to check if it is an issue tag by comparing whether it starts with `issue-`. The best thing, for now, I have come up with (which requires an extra API call) to check whether a tag starts with `issue-` is to use the `{{#get}}` filtering.

```hbs
{{#get "tags" filter="slug:~^'issue-'+slug:{{slug}}" limit="1" as |issue|}}
```

One thing to note, for consistency it might be useful to follow precedent set by the [Content API filtering operators](https://ghost.org/docs/content-api/#operators) and use `~^` for "starts with" (it also provides `~` for "contains" and `~$` for "ends with")

Is there an associated issue for this pull request? I'd love to see this change merged and perhaps that would help generate some more attention.

ErisDS:
I'm sorry - I did think that I'd responded to this 🙈 

I have just merged https://github.com/TryGhost/Ghost/pull/23306 which has the same syntax for startsWith and endsWith operators that we use in our filtering everywhere - this was the feedback that I missed giving.

Further updates / additions to match helper are welcome!

PR #18422, opened 2023-10-02
Lexical GA

9larsons:


9larsons:
- moved the beta editor (Lexical-based editor) to the default editor; all pages and posts will now use it
- all mobiledoc posts will remain mobiledoc; if opened, they will be converted to Lexical on the fly and be opened within the new editor
- we will likely migrate all posts to lexical at some point down the road

PR #18428, opened 2023-10-02
Improve external media inliner URL handling

PaulAdamDavis:
no issue

There are URLs that cause the external media inliner to fail. This PR attempts to address those. In particular, URLs with:

- Spaces `https://example.com/file name.jpg`
- No extension `https://example.com/file-name`
- Unicode characters `https://example.com/你好.png`
- No scheme `//example.com/image.jpg`
- Query params `https://example.com/image.jpg?v=1&s=small`
- Duplicated characters `https://example.com/---1.jpg`
- Quotes `https://example.com/ghost-logo’s-cool.jpg`

... and all combinations of the above.

---

### <samp>🤖 Generated by Copilot at 81bdb08</samp>

This pull request improves the external media inliner module by using a better file type detection library, refactoring the code to use modern syntax and expose public methods, and updating the tests to use binary fixtures and cover more scenarios.


PaulAdamDavis:
@naz I've requested a PR from you! In particular, I'm looking to find out if I break any existing patterns or logging with these changes. I also made the methods public so I can test their output specifically. Hopefully _that_ doesn't cause any issues.

ghost/external-media-inliner/lib/ExternalMediaInliner.js
```diff
@@ -39,14 +40,22 @@ class ExternalMediaInliner {
      * @param {string} requestURL - url of remote media
      * @returns {Promise<Object>}
      */
-    async #getRemoteMedia(requestURL) {
+    async getRemoteMedia(requestURL) {
         // @NOTE: this is the most expensive operation in the whole inlining process
         //        we should consider caching the results to improve performance
+
+        // Enforce https
+        requestURL = requestURL.replace(/^\/\//g, 'https://');
```
naz, line 48: What's going to happen if the file is not available over https? 🤔

ghost/external-media-inliner/lib/ExternalMediaInliner.js
```diff
@@ -63,21 +72,22 @@ class ExternalMediaInliner {
...
-            .split('/')
-            .pop()
-            .split('#')[0]
-            .split('?')[0];
+        const removeExtRegExp = new RegExp(`.${extension}`, '');
+        const fileNameNoExt = path.parse(requestURL).base.replace(removeExtRegExp, '');
 
-        const extension = mime.extension(contentType) || filename.split('.').pop();
+        // CASE: Query strings _can_ form part of the unique image URL, so rather that strip them include the in the file name, and then trim to 200 chars
+        // 200 is a magic number which could be refined. Also trim leading & trailing dashes.
```
naz, line 83: Let's at least use one of the real limiting numbers here, like for windows it's 255 chars according to [this](https://stackoverflow.com/a/265782)

naz: but doesn't really matter much  😄

ghost/external-media-inliner/package.json
```diff
@@ -22,7 +22,7 @@
     "mocha": "10.2.0",
     "sinon": "15.2.0"
   },
-"dependencies": {
-    "mime-types": "2.1.35"
-}
+  "dependencies": {
+    "file-type": "16"
```
naz, line 26: should be 16.0.0 ?

ghost/external-media-inliner/lib/ExternalMediaInliner.js
```diff
@@ -63,21 +72,22 @@ class ExternalMediaInliner {
      * @param {Object} response - response from request
      * @returns {Object}
      */
-    #extractFileDataFromResponse(requestURL, response) {
-        const headers = response.headers;
-        const contentType = headers['content-type'];
+    async extractFileDataFromResponse(requestURL, response) {
+        const fileInfo = await FileType.fromBuffer(response.body);
```
naz, line 76: are there cases where this method might fail? should we still keep the content-type as a fallback?

ghost/external-media-inliner/lib/ExternalMediaInliner.js
PaulAdamDavis: I've set this to http. http>https redirects are common, so I think this is fine

ghost/external-media-inliner/lib/ExternalMediaInliner.js
PaulAdamDavis: I landed on 248. Saw 255, 257, etc with 248 being the limit for folders. Airing on the side of caution makes sense to me here. The final file URL will be different anyway, so it isn't that important.

ghost/external-media-inliner/package.json
PaulAdamDavis: `16.5.4` actually :D `yarn add -E file-type@16` was the culprit.

ghost/external-media-inliner/lib/ExternalMediaInliner.js
PaulAdamDavis: Text-based files would be missed, but as `file-type` checks for binary files ([ref](https://github.com/sindresorhus/file-type/blob/v16.5.4/readme.md?plain=1#L7)), this is a good shout. I've added the content-type / path extension fallback back in.

ghost/external-media-inliner/lib/ExternalMediaInliner.js
naz: It does seem fine, I'm just thinking of the edge case where https would be 404 for whatever bad reason 😂  Let's fix it when it breaks I guess. The web in 2023 should be all on https anyway!

ghost/external-media-inliner/lib/ExternalMediaInliner.js
naz: Cool-cool 👍  nbd anyway, just wondered if we could put a boundary that has some meaning.

ghost/external-media-inliner/lib/ExternalMediaInliner.js
```diff
@@ -63,21 +73,32 @@ class ExternalMediaInliner {
...
      */
-    #extractFileDataFromResponse(requestURL, response) {
-        const headers = response.headers;
-        const contentType = headers['content-type'];
+    async extractFileDataFromResponse(requestURL, response) {
+        let extension;
+
+        try {
+            const fileInfo = await FileType.fromBuffer(response.body);
+            extension = fileInfo.ext;
```
naz, line 81: This bit might return undefined according to docs:
> Or undefined when there is no match.

Worth putting the content-type fallback should account for that case too

PaulAdamDavis: If it returns undefined, the try/catch fails with `Cannot read properties of undefined (reading 'ext')`, so I think this already falls back to using the content-type/extension correctly.

naz: That's against of a strict "do not control the logic flow with exceptions" but well, ok 😄 maybe just leave a comment about why/when the exception is thrown haha because if someone else reads this it would be slightly weird

PaulAdamDavis: Fair comment! I've added a comment explaining it.

PR #18457, opened 2023-10-03
Added IncomingRecommendation API

sagzy:
no issue

- added `GET /incoming_recommendations/` browse endpoint to the Admin API
    - we store incoming recommendations as mentions in the database. The new endpoint reuses the Mentions API underneath to fetch verified mentions of type recommendation
    - recommendation-specific attributes are returned by the new endpoint, including calculated fields such as the "RecommendingBack" boolean
- show "Recommend back" option for sites recommending me, only if I haven't recommended the site already


ghost/core/content/themes/source
sagzy, line 1: This should not be in the PR

sagzy:
Closing as duplicate of #18478

PR #18527, opened 2023-10-06
Renamed reason to description in recommendations table

SimonBackx:
fixes https://github.com/TryGhost/Product/issues/4005

We no longer use the 'reason' of a recommendation, but allow a flexible description instead. Because this is a breaking change in the API, we do this before making this feature GA.
- Added new database utils for renaming a column
- Added new migration to rename the column
- Updated all references in code


ghost/core/core/server/data/schema/commands.js
```diff
@@ -156,6 +156,20 @@ async function dropColumn(tableName, column, transaction = db.knex, columnSpec =
     }
 }
 
+/**
+ * @param {string} tableName
+ * @param {string} from
+ * @param {string} to
+ * @param {import('knex').Knex.Transaction} [transaction]
+ */
+async function renameColumn(tableName, from, to, transaction = db.knex) {
```
daniellockyer, line 165: We need to be careful here: https://ghost.slack.com/archives/C02G9E68C/p1662985535756619 and https://github.com/TryGhost/Ghost/commit/a6991746c3aec3e925e2d50f4e88edf87c8dcc74

SimonBackx: Thanks for the heads up! I was planning to ask for your review anyway because of this. Is it okay if I implement this raw query approach in the helpers instead? I think we'll have to use this in the future migrations again, and probably good to have a helper for it

SimonBackx: I've updated the code for this change

sagzy (approved):
💯

daniellockyer (approved):

PR #18533, opened 2023-10-09
Modified getLocalTime to add timeonly option

ronaldlangeveld:
no issue

---

### <samp>🤖 Generated by Copilot at 1e288b3</samp>

Added a `format` parameter to the `getLocalTime` function and used it in the `Hint` component to show only the time part of the local time in the time zone setting. This improved the user experience of the Ghost admin panel.


daniellockyer:
Closing as it seems this has been solved in a different way

PR #18553, opened 2023-10-10
Bug fix for referral source: Return known source for referral_source …

vershwal:
Fixes https://github.com/TryGhost/Product/issues/3928


55sketch:
@vershwal LGTM!

PR #18581, opened 2023-10-11
Remove default post revisions

allouis:
### <samp>🤖 Generated by Copilot at 94f6307</samp>

This pull request adds more fields to the query objects for the editor routes to support the post preview feature, and removes some fields from the default relations for the pages and posts APIs to optimize the data fetching and performance.


ghost/admin/app/routes/editor/edit.js
```diff
@@ -31,7 +31,8 @@ export default class EditRoute extends AuthenticatedRoute {
 
         let query = {
             // eslint-disable-next-line camelcase
-            id: post_id
+            id: post_id,
+            include: 'tags,authors,authors.roles,email,tiers,newsletter,count.clicks,post_revisions,post_revisions.author'
```
allouis, line 35: This is gross tbh - it would be nice to skip setting the `include` here and just request the revisions when opening the history feature

PR #18613, opened 2023-10-13
Switched to using containerised playwright

sam-lord:
refs: https://github.com/TryGhost/DevOps/issues/78

This should reduce the time to install playwright to 0, from around 40 seconds.


daniellockyer:
Closing as we now have a different way to make these faster

PR #18643, opened 2023-10-16
🐛 Fixed active icons in dark mode (admin tab navigation, mobile)

lorumic:
no issue

- When using dark mode on mobile, some of the icons in the admin tab navigation become invisible for the active section, due to a black fill applied to them, which is not overridden.
There was a fix applied to the "dashboard" icon already - as this is the default landing one, the bug would have hardly gone unnoticed there. I used the same principle and applied it to the "posts" and "members" icons as well.

Before:

Peek 2023-10-16 13-50

After:

Peek 2023-10-16 13-51


peterzimon:
Mobile menu has been updated since then.

PR #18667, opened 2023-10-17
Added unit tests for AdminX API hooks

binary-koan:
refs https://github.com/TryGhost/Product/issues/3831

---

### <samp>🤖 Generated by Copilot at b39e813</samp>

This pull request refactors the `createMutation` and `mutate` hooks in the `admin-x-settings` app to simplify the mutation options and fetch logic. It also adds the `@testing-library/react` dependency and unit tests for the custom hooks and the query cache update functions. The goal of these changes is to improve the code quality and test coverage of the app.


binary-koan:
Can't figure out the specs here, will try again when we separate this package (smaller surface area to debug then)

PR #18668, opened 2023-10-17
🎨 Improve post title-slug sync

lorumic:
no issue

- Before, one had to manually clear the post slug from the post settings panel, and then change the title (which triggers an auto-save), to see the title change reflected in the post slug. This is annoying and cumbersome, and may lead to unwanted post slugs, so I propose an updated implementation, which essentially keeps the post slug constantly synced with the post title except in the following 2 cases:
1. The user has manually changed the slug.
2. The post has already been published.
‎ 

### DIFF
--- a/ghost/admin/app/controllers/lexical-editor.js
+++ b/ghost/admin/app/controllers/lexical-editor.js
@@ -28,8 +28,6 @@ import {mobiledocToLexical} from '@tryghost/kg-converters';
 import {inject as service} from '@ember/service';
 
 const DEFAULT_TITLE = '(Untitled)';
-// suffix that is applied to the title of a post when it has been duplicated
-const DUPLICATED_POST_TITLE_SUFFIX = '(Copy)';
 // time in ms to save after last content edit
 const AUTOSAVE_TIMEOUT = 3000;
 // time in ms to force a save if the user is continuously typing
@@ -124,6 +122,11 @@ export default class LexicalEditorController extends Controller {
      * Flag used to determine if we should return to the analytics page or to the posts/pages overview
      */
     fromAnalytics = false;
+
+    /**
+     * Flag used to determine if the user has input a custom slug manually
+     */
+    isUserCustomSlug = false;
 
     // koenig related properties
     wordCount = 0;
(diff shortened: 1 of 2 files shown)


lorumic:
> The intentional behaviour here is that as long as you haven't manually set a post slug or published your post the title+slug are kept in sync. This used to be the case but appears to have broken recently. We do not want to _always_ sync the title+slug because that causes problems when you want to manually set a slug that is different to the title.

I see your point here. However, there might be users like me who might want to keep them always synced, or at least have a less cumbersome way than the current one to sync them. Can we maybe reach a compromise? For example, a small refresh-like icon button could be added next to the slug input field. This, once clicked, generates the slug from the current title and updates it. What do you think?

kevinansfield:
Adding a button specifically for that goes against our Ghost's design ethos and over-complicates the UI.

Having the slug auto-update on title change _only_ when it hasn't been changed manually was previously working very well and has been the case for many years without issue. Restoring that behaviour would be preferable

lorumic:
Ok, I understand. I didn't know about the old behaviour that you mentioned as I started using Ghost recently, so it's always been broken for me, but I can try to restore that same behaviour based on your description of it.

markstos:
@kevinansfield Is right here that it's not desirable for the slug to auto-update after the post has been published. This could cause published URLs to break when a typo or formatting issue in the title is fixed.

lorumic:
Thanks @kevinansfield and @markstos for your comments that helped me understand how this should behave.

To recap, I updated the implementation so that post title and slug are kept synced except in the 2 cases you mentioned above:

1. The user has manually changed the slug.
2. The post has already been published.

Please check again and let me know if the current implementation looks good.

ErisDS:
Sorry, we didn't get around to reviewing this PR. In the meantime, there have been two fixes that I believe cover all the cases that this PR would have covered:

https://github.com/TryGhost/Ghost/commit/7083bd0628c473a7e4375007dfc0eea5f953a999
https://github.com/TryGhost/Ghost/commit/d732f83a9ff8032cc76f64df01e20b2a1d41cb8d

Let us know if you find further issues.

PR #18700, opened 2023-10-19
🎨 Improved headers layout in post editor and post settings side panel

### LINKED ISSUE #18690: Published post actions wrap over "add feature image"
### Issue Summary

I experienced this issue on an old Android phone - so kind of an edge case - and I was able to reproduce it easily in the device view of the Chrome dev tools.
The problem is: when editing a published post on mobile, if there is not enough space in the header, the 2 Update/Unpublish action buttons at the top wrap to a second line, due to the `flex-wrap` attribute of their parent, and cover the "Add feature image" button, making it unusable:

Peek 2023-10-18 15-31

I guess this might also be the reason why, for a draft post, `display: none` is applied to the "Preview" button when screen width is less than or equal to 500px. I don't know if a similar solution would be good in this case, though, as they might both be important, so we might not want to hide either of them. However, being able to add the feature image to a published post is important too, so hopefully a reasonable solution can be found - or maybe none at all, if it's indeed too much of an edge case.

### Steps to Reproduce

1. Open Chrome dev tools, toggle device mode.
2. Select "Dimensions: Responsive" to dynamically resize the device width.
3. Lower the width until you see the problem: the 2 action buttons wrap to a second line and cover the upper part of the editor, where the "Add feature image" button is placed.

### Ghost Version

5.69.3

### Node.js Version

v18.18.2

From source (Ubuntu)

### Database type

MySQL 5.7

### Browser & OS version

Google Chrome 118.0.5993.70, Ubuntu

_No response_

### Code of Conduct

lorumic:
- fixes #18690
- This change makes the headers for post editor main section and post settings side panel use `sticky` position instead of `absolute`/`fixed` respectively. This allows to solve the following 2 problems:
1. The published post actions wrapping over the "Add feature image" button (see #18690).
2. The header of the post settings section having incorrect width, and causing the underlying content to be visible on the right edge.

Problem 1 - before:

Peek 2023-10-19 12-19

Problem 1 - after:

Peek 2023-10-19 12-21

Problem 2 - before:

Peek 2023-10-19 12-20

Problem 2 - after:

Peek 2023-10-19 12-22

### DIFF
--- a/ghost/admin/app/styles/components/koenig.css
+++ b/ghost/admin/app/styles/components/koenig.css
@@ -15,7 +15,7 @@
 
 /* Padded container housing title + editor canvas, scrollable content */
 .gh-koenig-editor-pane {
-    padding: 11vw 92px;
+    padding: calc(11vw - 94px) 92px 11vw 92px;
 }
 
 @media (min-width: 500px) and (max-width: 960px) {
--- a/ghost/admin/app/styles/components/settings-menu.css
+++ b/ghost/admin/app/styles/components/settings-menu.css
@@ -120,10 +120,10 @@
 /* ---------------------------------------------------------- */
 
 .settings-menu-header {
-    position: fixed;
+    position: sticky;
+    top: 0;
     display: flex;
     width: 100%;
-    max-width: 364px;
     padding: 36px 24px 24px;
     justify-content: space-between;
     align-items: center;
@@ -207,7 +207,7 @@
 }
 
 .settings-menu-content {
-    padding: 92px 0 0;
+    padding: 8px 0 0;
 }
 
 .gh-post-settings {
@@ -216,7 +216,7 @@
 
 @media (max-width: 1024px) {
     .settings-menu-content {
-        padding-top: 72px;
+        padding-top: 5px;
     }
 }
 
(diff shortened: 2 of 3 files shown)


sanne-san (commented):
Thank you @lorumic for your PR! These are some great improvements, only I saw that it breaks the header height between 448-1024px. This would have to be fixed before we can merge it 🙂 

CleanShot 2023-11-21 at 15 17 46@2x

lorumic:
> Thank you @lorumic for your PR! These are some great improvements, only I saw that it breaks the header height between 448-1024px. This would have to be fixed before we can merge it 🙂

Thanks for the review @sanne-san. You're right, we need to set an appropriate padding for the header in that width range. I've just updated the PR to address this, could you please check that it now behaves as expected?

sanne-san (commented):
Thank you for the change! Having another look, I'm still seeing a variety of issues across published and draft posts unfortunately. I'm going to close this PR because the mobile header will need some more major rethinking, which I'm going to pick up. Thanks!


https://github.com/TryGhost/Ghost/assets/65487235/ab7ccaa3-7dd6-4e00-a6ea-085dd35f6187

CleanShot 2023-11-22 at 11 08 50@2x

PR #18608, opened 2023-10-13
🎨 Improved search keywords in Admin-X Settings

betschki:
Got some code for us? Awesome 🎊!

Please include a description of your change & check your PR against this list, thanks!

- [x] There's a clear use-case for this code change, explained below
- [x] Commit message has a short title & references relevant issues
- [x] The build will pass (run `yarn test:all` and `yarn lint`)

We appreciate your contribution!

---

### <samp>🤖 Generated by Copilot at b1ece9b</samp>

Moved and updated the search keywords for the settings sections from their respective component files to the `AdvancedSettings.tsx` file. This was part of a refactoring to consolidate the settings components and to improve the searchability of the settings options.

PR #18704, opened 2023-10-19
Added individual settings fields to search keywords

betschki:
The goal of this commit is to make all the individual setting fields searchable by adding them to the keywords. This should significantly improve UX when trying to look for "comments", for example, which one might not suspect to be paired with "Access" settings in memberships.


betschki:
This builds upon https://github.com/TryGhost/Ghost/pull/18608 and is related to a discussion in the Ghost forum here: https://forum.ghost.org/t/searching-the-new-ui-bug/42043/2

daniellockyer:
Thanks @betschki! 🙂

PR #18748, opened 2023-10-25
Implemented Sentry Replays in admin

cmraible:
refs TryGhost/Product#4064

- More info on Sentry Replays here: https://docs.sentry.io/product/session-replay/
- Basically it records the app's DOM in the moments before an error occurs and allows you to 'play it back' from within Sentry
- You can inspect the DOM as if you were in the browser with them when the error occurred
- Note: Text and images are obfuscated, so it visually isn't that impressive, but being able to inspect the DOM is a game changer for debugging I think


cmraible:
Tabling this for now, might come back to it at some point.

PR #18770, opened 2023-10-26
Improved Sentry error messages for SQL errors

cmraible:
no issue

- In the catch-all 'UNEXPECTED ERROR' handler, we set the error context to the full text of the failing SQL query. This text is then sent to Sentry, with the query as the event type, which messes up issue grouping because it includes all the parameters, which are dynamic.
- This (admittedly hacky) solution adds special handling for SQL query errors that aren't caught elsewhere to improve the error message that is sent to Sentry. There shouldn't be any change to end users, who will still receive a generic 500 UNEXPECTED ERROR message. 
- Ideally we should handle SQL query errors much further up the stack, but this change should at least surface these errors in a more consolidated and manageable way.

### DIFF
--- a/ghost/mw-error-handler/lib/mw-error-handler.js
+++ b/ghost/mw-error-handler/lib/mw-error-handler.js
@@ -87,6 +87,15 @@ module.exports.prepareError = (err, req, res, next) => {
                 message: err.message,
                 statusCode: err.statusCode
             });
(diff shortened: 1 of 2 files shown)


cmraible:
Closing in favor of https://github.com/TryGhost/Ghost/pull/18797

PR #18785, opened 2023-10-28
Updated translations with latest changes

honhatduy:
Add new strings


daniellockyer:
It seems we've already merged a PR to update these translations. Please re-open a PR with revised changes if you'd like to make an amendment 🙂

PR #18797, opened 2023-10-30
Improved error handling for SQL errors

cmraible:
refs TryGhost/Product#4083

- In the vast majority of cases, we shouldn't have SQL errors in our code. Due to some limitations with validating e.g. nql filters passed to the API, sometimes we don't catch these errors and they bubble up to the user.
- In these rare cases, Ghost was returning the raw SQL error from mysql which is not very user friendly and also exposes information about the database, which generally is not a good practice.
- To make things worse, Sentry was treating every instance of these errors as a unique issue, even when it was exactly the same query failing over and over.
- This change improves the error message returned from the API, and also makes sure that Sentry will group all these errors together, so we can easily see how many times they are happening and where.
- It also adds more specific context to the event that is sent to Sentry, including the mysql error number, code, and the SQL query itself.

### DIFF
--- a/ghost/core/core/shared/sentry.js
+++ b/ghost/core/core/shared/sentry.js
@@ -22,13 +22,25 @@ if (sentryConfig && !sentryConfig.disabled) {
                     event.exception.values[0].type = exception.context;
                 }
 
+                // This is a mysql2 error — add some additional context
+                if (exception.sql) {
+                    event.exception.values[0].type = `SQL Error ${exception.errno}: ${exception.sqlErrorCode}`;
+                    event.exception.values[0].value = exception.sqlMessage;
+                    event.contexts.mysql = {
+                        errno: exception.errno,
+                        code: exception.sqlErrorCode,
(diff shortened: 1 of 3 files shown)


cmraible:
API Response Before:
image


API Response After:
CleanShot 2023-10-30 at 12 06 17@2x

cmraible:
Sentry Event Before ([link](https://ghost-foundation.sentry.io/issues/4575028265/?environment=development&project=1879038&query=&statsPeriod=7d&stream_index=1)):
CleanShot 2023-10-30 at 12 07 05@2x


Sentry Event After ([link](https://ghost-foundation.sentry.io/issues/4575182808/?environment=development&project=1879038&query=&statsPeriod=7d&stream_index=0)):
CleanShot 2023-10-30 at 12 07 30@2x
CleanShot 2023-10-30 at 12 08 22@2x

mike182uk (approved):
Looks good 😎

PR #18805, opened 2023-10-31
Added indexes to `members_created_events` and  `members_subscription_created_events`

SimonBackx:
fixes https://github.com/TryGhost/Product/issues/4085

Increases the performance for the post analytics export by adding new indexes. These indexes are used when counting the amount of (paid) subscribers that were attributed to a given post. With the indexes, the time required to export 700 posts with 300k members decreases from 40s to 0.6s.

Tests show that adding these indexes should be very fast (< 1 s) if the tables contain up to 300k rows.

### DIFF
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -523,7 +523,7 @@ module.exports = {
         id: {type: 'string', maxlength: 24, nullable: false, primary: true},
         created_at: {type: 'dateTime', nullable: false},
         member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
-        attribution_id: {type: 'string', maxlength: 24, nullable: true},
+        attribution_id: {type: 'string', maxlength: 24, nullable: true, index: true},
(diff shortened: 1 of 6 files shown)


ghost/core/core/server/data/schema/schema.js
```diff
@@ -538,7 +538,10 @@ module.exports = {
                 isIn: [['member', 'import', 'system', 'api', 'admin']]
             }
         },
-        batch_id: {type: 'string', maxlength: 24, nullable: true}
+        batch_id: {type: 'string', maxlength: 24, nullable: true},
+        '@@INDEXES@@': [
```
daniellockyer, line 542: I think we should just add `index: true` to `attribution_id`. `@@INDEXES@@` also works but it's primarily for composite indexes

SimonBackx: Oh, I completely missed that this was possible. I'll update it

SimonBackx: Updated ✅

daniellockyer (approved):

PR #18884, opened 2023-11-06
Filter members by email disabled

SimonBackx:
fixes https://github.com/TryGhost/Product/issues/4108

- Updates filters behind a new alpha feature flag so you can also filter on members who have email disabled (because the email had a permanent bounce, they reported spam or the email address is invalid)
- When returning members, we now also use the email_disabled flag to set email_suppression.suppressed correctly (in case they are out of sync, which should normally never happen).

### DIFF
--- a/ghost/admin/app/components/members/filter.js
+++ b/ghost/admin/app/components/members/filter.js
@@ -178,18 +178,6 @@ export default class MembersFilter extends Component {
             return [filter];
         });
 
-        // find list of newsletters from store and add them to filter list if there are more than one newsletter
-        // it also removes the 'subscribed' filter from the list as that would unsubscribe members from all newsletters, instead replace it with a filter for each newsletter
-        // if (this.newsletters?.length > 1) {
-        //     // remove the 'subscribed' filter from the list
-        //     //availableFilters = availableFilters.filter(prop => prop.name !== 'subscribed');
-        //
-        //     // find the index of the 'basic' group and insert the 'multiple newsletters' filter after it
-        //     const indexes = availableFilters.map((obj, index) => (obj.group === 'Basic' ? index : null)).filter(i => i !== null);
-        //     const lastIndex = indexes.pop();
-        //     availableFilters.splice(lastIndex + 1, 0, ...NEWSLETTERS_FILTER(this.newsletters));
-        // }
-
         // only add the offers filter if there are any offers
         if (this.offers.length > 0) {
             availableFilters = availableFilters.concat(OFFERS_FILTER);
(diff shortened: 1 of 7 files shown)


ghost/admin/app/components/members/filter.js
```diff
@@ -1,7 +1,7 @@
 import Component from '@glimmer/component';
 import moment from 'moment-timezone';
 import nql from '@tryghost/nql-lang';
-import {AUDIENCE_FEEDBACK_FILTER, CREATED_AT_FILTER, EMAIL_CLICKED_FILTER, EMAIL_COUNT_FILTER, EMAIL_FILTER, EMAIL_OPENED_COUNT_FILTER, EMAIL_OPENED_FILTER, EMAIL_OPEN_RATE_FILTER, EMAIL_SENT_FILTER, LABEL_FILTER, LAST_SEEN_FILTER, NAME_FILTER, NEWSLETTERS_FILTER, NEXT_BILLING_DATE_FILTER, OFFERS_FILTER, PLAN_INTERVAL_FILTER, SIGNUP_ATTRIBUTION_FILTER, STATUS_FILTER, SUBSCRIBED_FILTER, SUBSCRIPTION_ATTRIBUTION_FILTER, SUBSCRIPTION_START_DATE_FILTER, SUBSCRIPTION_STATUS_FILTER, TIER_FILTER} from './filters';
+import {AUDIENCE_FEEDBACK_FILTER, CREATED_AT_FILTER, EMAIL_CLICKED_FILTER, EMAIL_COUNT_FILTER, EMAIL_FILTER, EMAIL_OPENED_COUNT_FILTER, EMAIL_OPENED_FILTER, EMAIL_OPEN_RATE_FILTER, EMAIL_SENT_FILTER, LABEL_FILTER, LAST_SEEN_FILTER, NAME_FILTER, NEWSLETTERS_FILTERS, NEXT_BILLING_DATE_FILTER, OFFERS_FILTER, PLAN_INTERVAL_FILTER, SIGNUP_ATTRIBUTION_FILTER, STATUS_FILTER, SUBSCRIBED_FILTER, SUBSCRIPTION_ATTRIBUTION_FILTER, SUBSCRIPTION_START_DATE_FILTER, SUBSCRIPTION_STATUS_FILTER, TIER_FILTER} from './filters';
```
sagzy, line 4: No biggie, only for consistency with the rest of the filters: would have kept the singular name `NEWSLETTER_FILTER`

sagzy (commented):
Nice! 🙌 Code is looking good, added one comment on naming. Also have additional feedback from doing QA, I'll post that directly into Slack

ghost/admin/app/components/members/filter.js
SimonBackx: Yeah, I wanted to make it more clear that this filter is actually multiple filters, one for each newsletter, hence the name change

PR #18885, opened 2023-11-06
Fixed email rendering bug in Android gmail client

cmraible:
refs https://github.com/TryGhost/Ghost/pull/18587/files and https://github.com/TryGhost/Ghost/pull/17475/files

- In October 2022, juice introduced a change that began inlining `width: auto;` and `height: auto` from CSS on image tags, resulting in `width="auto"` and `height="auto"` being added to image tags in our email renderer (https://github.com/Automattic/juice/commit/cb62062794e07cf4c0f30b16bd2c6b31b8144c41)
- The change in juice broke our email rendering in Outlook. The two referenced commits above were attempts to work around this new behavior, but ultimately each attempt fixed one issue while creating another issue
- This commit reverts our previous attempts to work around this new behavior in juice, and instead forks juice to revert to its original behavior, which does not inline `width` and `height` attributes if they are set to `auto`
- Long-term it would be better to make the behavior configurable in the upstream juice repo so we don't have to maintain a fork, but this should fix the email renderer in the meantime

### DIFF
--- a/ghost/core/package.json
+++ b/ghost/core/package.json
@@ -197,7 +197,7 @@
     "json-stable-stringify": "1.0.2",
     "jsonpath": "1.1.1",
     "jsonwebtoken": "8.5.1",
-    "juice": "9.1.0",
+    "juice": "cmraible/juice#fix-auto-attributes",
     "keypair": "1.0.4",
     "knex": "2.4.2",
     "knex-migrator": "5.1.5",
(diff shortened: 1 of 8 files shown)


cmraible:
Closed in favor of https://github.com/TryGhost/Ghost/commit/3a4417ebd6bcd4ed7b827dc2493a8f4273b9580b so we don't have to fork juice

PR #18915, opened 2023-11-08
Made new recommendation email more flexible

djordjevlais:
- Removes `min-height` of the card in order to make the template more flexible half-empty states look better (missing description, short description, featured image...)
- Fixes wrong variables used in Outlook-specific template


sagzy (approved):
LGTM 👍

PR #18922, opened 2023-11-09
Removed email from Mailgun's suppression list on unsubscribe

sagzy:
refs https://github.com/TryGhost/Product/issues/4075

- when a member clicks on "Unsubscribe from that list" from Apple Mail, the member's email is put into Mailgun's Unsubscribe suppression list. Ghost listens for "Unsubscribe" events from Mailgun, and unsubscribes the member from all the newsletters
- now, the member is only unsubscribed from the newsletter they unsubscribe to (not all of them)
- now, the email is also deleted from Mailgun's suppression list, so that it doesn't affect any other membership


ghost/email-service/lib/EmailEventStorage.js
```diff
@@ -128,11 +139,24 @@ class EmailEventStorage {
...
+            const existingNewsletters = member.related('newsletters');
+
+            if (!existingNewsletters) {
+                return [];
+            }
+
+            const email = await this.#models.Email.findOne({id: event.emailId});
+            const newsletterToRemove = await email.get('newsletter_id');
+
+            return existingNewsletters.toJSON().filter(newsletter => newsletter.id !== newsletterToRemove);
```
SimonBackx, line 156: I think we should still map this to ids, otherwise we do an update on a relation with lots of properties. Here or at least when using findNewslettersToKeep. It is probably also a bit more performant to use `models` instead of `toJSON` because we only need the ids later.
```suggestion
            return existingNewsletters.models.filter(newsletter => newsletter.id !== newsletterToRemove).map(n => { return {id: n.id}; });
```

ghost/email-service/lib/EmailEventStorage.js
```diff
@@ -128,11 +139,24 @@ class EmailEventStorage {
...
+                withRelated: ['newsletters']
+            });
+            const existingNewsletters = member.related('newsletters');
+
+            if (!existingNewsletters) {
+                return [];
+            }
+
+            const email = await this.#models.Email.findOne({id: event.emailId});
+            const newsletterToRemove = await email.get('newsletter_id');
```
SimonBackx, line 154:
```suggestion
            const newsletterToRemove = email.get('newsletter_id');
```

SimonBackx:
I do have a concern that popped up in my mind about the whole flow. Imagine the following situation:

- Site A sends a newsletter to person A.
- Person A unsubscribes via the mailto
- Mailgun adds it to the global unsubscribe list
- Site B sends a newsletter to person A
- Email got blocked by Mailgun
- Site A fetches unsubscribes and removes person A from Mailguns list (this can happen with a huge delay of hours because sometime we receive thousands and thousands of events that take a while before they are all processed)

ghost/email-service/lib/EmailEventStorage.js
sagzy: thanks! Updated in 35f0ab52b0bdc6db3da4e82e92a5f91338686728

ghost/email-service/lib/EmailEventStorage.js
sagzy: thanks! Updated in 35f0ab52b0bdc6db3da4e82e92a5f91338686728

PR #18935, opened 2023-11-09
🔒 Added support for logging out members on all devices

SimonBackx:
fixes https://github.com/TryGhost/Product/issues/3738
[Tech Spec ](https://www.notion.so/ghost/Member-Session-Invalidation-13254316f2244c34bcbc65c101eb5cc4)

- Adds the transient_id column to the members table. This defaults to email, to keep it backwards compatible (not logging out all existing sessions)
- Instead of using the email in the cookies, we now use the transient_id
- Updating the transient_id means invalidating all sessions of a member
- Adds an endpoint to the admin api to log out a member from all devices
- Added the `all` body property to the DELETE session endpoint in the members API. Setting it to true will sign a member out from all devices.
- Adds a UI button in Admin to sign a member out from all devices
- Portal 'sign out of all devices' will not be added for now

Related changes (added because these areas were affected by the code changes):
- Adds a serializer to member events / activity feed endpoints - all member fields were returned here, so the transient_id would also be returned - which is not needed and bloats the API response size (`transient_id` is not a secret because the cookies are signed)
- Removed `loadMemberSession` from public settings browse (not used anymore + bad pattern)

Performance tests on site with 50.000 members (on Macbook M1 Pro):
- Migrate: 6s (adding column 4s, setting to email is 1s, dropping nullable: 1s)
- Rollback: 2s


ghost/admin/app/components/members/modals/logout-member.js
```diff
@@ -0,0 +1,31 @@
...
+    get member() {
+        return this.args.data.member;
+    }
+
+    @task({drop: true})
+    *logoutMemberTask() {
+        try {
+            const url = this.ghostPaths.url.api('/members/', this.member.id, '/logout/');
+            const options = {};
+            yield this.ajax.post(url, options);
```
sagzy, line 19: Should we be using a `DELETE` request instead of `POST` for logging out, i.e. deleting a session?

ghost/core/core/server/data/schema/schema.js
```diff
@@ -417,6 +417,7 @@ module.exports = {
     members: {
         id: {type: 'string', maxlength: 24, nullable: false, primary: true},
         uuid: {type: 'string', maxlength: 36, nullable: true, unique: true, validations: {isUUID: true}},
+        transient_id: {type: 'string', maxlength: 191, nullable: false, unique: true},
```
sagzy, line 420: Wonder if we can use the `uuid` instead of the `email` as the default value for `transient_id`? Then we could keep that `maxLength` to 36 instead of 191. We can then also apply validation to the field for better consistency (`isUUID: true`)

sagzy (approved):
LGTM 👍 Added two small comments for us to discuss further

ghost/admin/app/components/members/modals/logout-member.js
SimonBackx: I'll change it to `DELETE /members/id/sessions/`

SimonBackx: ✅

SimonBackx:
Performance tests on site with 50.000 members (on Macbook M1 Pro):
- Migrate: 6s (adding column 4s, setting to email is 1s, dropping nullable: 1s)
- Rollback: 2s

ghost/core/core/server/api/endpoints/members.js
```diff
@@ -143,6 +143,31 @@ module.exports = {
...
+        ],
+        validation: {
+            options: {
+                id: {
+                    required: true
+                }
+            }
+        },
+        permissions: {
+            method: 'destroy'
```
allouis, line 162: Is this deffo the permission we want to use here?

I would expect this to be `'edit'` maybe? Though I'll be honest - that might be essentially the same as `'destroy'`

Not a big deal, but just wanted to make sure we don't want to change it, or add a new permission!

ghost/members-api/lib/repositories/MemberRepository.js
```diff
@@ -211,6 +212,16 @@ module.exports = class MemberRepository {
         }, options);
     }
 
+    async _generateTransientId() {
```
allouis, line 215: Why is this async?

allouis (commented):
Nice impl - few Q's!

ghost/core/core/server/api/endpoints/members.js
SimonBackx: I chose the most restrictive permissions (though they are indeed the same), but edit also makes sense for me. I'm not a big fan to add a new permission specifically for this

ghost/members-api/lib/repositories/MemberRepository.js
SimonBackx: I had async token generators in mind, but currently it is just implemented with a sync uuid. So my thinking was that generators are often async, so the method was also marked like that. But since it is a private method here, it doesn't really matter and I was probably overthinking 😅

PR #18940, opened 2023-11-09
Fixed a bug causing emails to be sent twice to the same member

cmraible:


daniellockyer:
Closing as this has been fixed in https://github.com/TryGhost/Ghost/commit/888acfb0c52ee08f7ca4c07ce6b6fc78fbc41e31

Issue #18949, opened 2023-11-12
Unintended behavior when entering post title using Japanese IME


t8m8:
### Issue Summary

In the Japanese (and possibly other languages) IME, enter key is pressed when converting a string, but the string is then duplicated and entered into the body of the article as well. It may be happening with IMEs in other languages.

https://github.com/TryGhost/Ghost/assets/10049972/18bad31a-e71d-4e37-bb4e-7f9d7c3a6799

### Steps to Reproduce

1. Enter a post title using Japanese IME
2. Press enter key to convert a string

It is expected that the conversion will be done correctly and the cursor will be behind the converted string.

### Ghost Version

5.72.2

### Node.js Version

18.12.1

### How did you install Ghost?

\-

### Database type

MySQL 5.7

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

PR #18950, opened 2023-11-12
🐛 Fixed the post title writing issue using Japanese IME

t8m8:
refs https://github.com/TryGhost/Ghost/issues/18949
- changed editor to control on a keyCode basis instead of key

behavior after this change:

https://github.com/TryGhost/Ghost/assets/10049972/beadc5f6-5223-4830-ac32-20ca2b7fbdc2

---

### <samp>🤖 Generated by Copilot at 9a912e7</samp>

Fixed a bug in the editor title input field that prevented keyboard navigation to the content area. Used `keyCode` instead of `key` in `gh-koenig-editor-lexical.js` to handle keyboard events consistently across browsers.

### DIFF
--- a/ghost/admin/app/components/gh-koenig-editor-lexical.js
+++ b/ghost/admin/app/components/gh-koenig-editor-lexical.js
@@ -113,16 +113,16 @@ export default class GhKoenigEditorReactComponent extends Component {
             return;
         }
 
-        const {keyCode} = event;
+        const {key} = event;
         const {value, selectionStart} = event.target;
 
         const couldLeaveTitle = !value || selectionStart === value.length;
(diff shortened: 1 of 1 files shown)


kevinansfield (changes requested):
It's preferable to use `key` rather than the deprecated key codes. Instead of this change could you try adding an early return for `event.isComposing`? E.g.

```js
if (!editorAPI || event.isComposing) {
    return;
]
```

t8m8:
@kevinansfield Thank you! I've updated to check `event.originalEvent.isComposing`

kevinansfield:
Thanks for the contribution! Tested and all seems to be working well 👌

kevinansfield (approved):

PR #18969, opened 2023-11-13
Added RewriteFrames integration to Admin Sentry

cmraible:
no issue

- Currently our stack traces in Production include the admin build version in the paths, e.g. `/admin/1633/assets` instead of `admin/assets`
- This confuses the error grouping logic in Sentry, resulting in many duplicate issues being created every time we release a new version of admin
- Ultimately, this makes it really difficult to determine if a 'New' issue in Sentry is actually new, or if it's just the first time we've seen it in this release.
- This commit adds the `RewriteFrames` integration to the Admin Sentry client, which will strip the build version from the paths in the stack traces, and allow Sentry to group issues correctly.
- With this, hopefully we will have far fewer 'New' issues created, so we can again start alerting on the 'New' condition in Sentry.

### DIFF
--- a/ghost/admin/package.json
+++ b/ghost/admin/package.json
@@ -40,7 +40,7 @@
     "@glimmer/component": "1.1.2",
     "@html-next/vertical-collection": "3.0.0",
     "@sentry/ember": "7.78.0",
-    "@sentry/integrations": "^7.80.0",
+    "@sentry/integrations": "7.80.0",
     "@tryghost/color-utils": "0.2.0",
     "@tryghost/ember-promise-modals": "2.0.1",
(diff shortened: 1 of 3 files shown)


ghost/admin/package.json
```diff
@@ -40,6 +40,7 @@
     "@glimmer/component": "1.1.2",
     "@html-next/vertical-collection": "3.0.0",
     "@sentry/ember": "7.78.0",
+    "@sentry/integrations": "^7.80.0",
```
mike182uk, line 43: Should we use a pinned version here? i.e `7.80.0`

cmraible: Good catch! Updated :)

PR #19009, opened 2023-11-16
Improved performance of redis cache purge

cmraible:
no issue

- When purging  the redis cache, we were looping over all the keys and sending a `del` command for each key.
- Redis can accept an array of keys in a single `del` command, so we can send all the keys in a single command which reduces the number of round trips to redis.
- In a basic local benchmark test, looping over 3390 keys took 55ms to purge, and sending all the keys in a single command took 21ms to purge.


cmraible:
Closing for now — added some of this logging in a separate commit last week, but that did not include the `del` optimization. Added the logging/metrics in first so we can see the actual impact of this change.

PR #19021, opened 2023-11-16
GRO-35 Updated members filter copy

djordjevlais:
fixes [GRO-35](https://linear.app/tryghost/issue/GRO-25/)


djordjevlais:
fixes GRO-35

PR #19025, opened 2023-11-16
Scoped cache purging to be more domain specific

mike182uk:
no refs

Caches were being cleared whenever the `site.changed` event was published, which happens whenever a model is added/edited/deleted. This was causing not just all caches to be purged, but also caches unrelated to the model being added/edit/deleted. As the cache system is expanded to include more domain specific caches, this approach will need to be revisited due to interdependencies between the caches, but for now this should be sufficient.

### DIFF
--- a/ghost/core/core/server/services/posts-public/service.js
+++ b/ghost/core/core/server/services/posts-public/service.js
@@ -1,3 +1,5 @@
+const logging = require('@tryghost/logging');
+
 class PostsPublicServiceWrapper {
     async init() {
         if (this.api) {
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/services/posts-public/service.js
```diff
@@ -15,8 +17,22 @@ class PostsPublicServiceWrapper {
...
+                    'post.published.edited',
+                    'post.unpublished',
+                    'page.deleted',
+                    'page.published',
+                    'page.published.edited',
+                    'page.unpublished',
+                    'post.tag.attached',
+                    'post.tag.detached',
+                    'page.tag.attached',
+                    'page.tag.detached'
```
cmraible, line 34: Should we also include the events for any includes (authors, tiers) and maybe `tag.edited` as well?

Sort of sucks that they are all stored in the same request body so we can't be more surgical, but I can't think of a way around it unless we fully rethink the approach of caching response bodies.

mike182uk: Ah yeh we should definitely clear on those events based on how we are caching the data. Have added:

- `tag.edited`
- `tag.deleted`
- `user.edited` (author)
- `user.deleted` (author)

mike182uk:
This is pretty outdated now and possibly superseded. Closing for now 👍

PR #19031, opened 2023-11-16
Removed usage of unquoted ids in filter strings

SimonBackx:
fixes [GRO-34](https://linear.app/tryghost/issue/GRO-34/)
fixes [GRO-33](https://linear.app/tryghost/issue/GRO-33/)

This adds a new way to run all tests with enforced numeric ObjectIDs. These
numeric ids cause issues if they are used withing NQL filters. So they
surface tiny bugs in our codebase.

You can run tests using this option via:
NUMERIC_IDS=1 yarn test:e2e

Removed some defensive logic that could be explained by this discovered
issue.


daniellockyer:
This PR is genius 😍

PR #19059, opened 2023-11-21
Added an email rendering test for all Koenig cards

cmraible:
refs TryGhost/Product#4125

This PR adds two new integration tests to ensure all our Koenig cards are rendered properly after going through the EmailRenderer. Although we have thorough tests for the cards themselves in the Koenig repo, the EmailRenderer does post-processing on the rendered HTML, such as inlining CSS, which can adversely impact the rendered output of our cards in email clients (usually Outlook).

Since email newsletters are a core feature of Ghost, these bugs are typically fairly urgent, and since it is email, they are also quite difficult to troubleshoot and fix. These two tests are intended to prevent bugs of this sort, which in the past have been created by seemingly harmless changes like bumping dependencies that are used in the EmailRenderer.

The idea is to create a 'Golden Post' which has at least 1 of every card from Koenig, run that post through the EmailRenderer, and take a snapshot of the rendered HTML. In the future, if we make any changes to the EmailRenderer or the Koenig cards themselves, this will trigger us to carefully consider the changes, and it provides an 'expected' output to compare our changes against.

Additionally, the second test simply checks that all cards from `kg-default-nodes` are included in the 'Golden Post'. This protects against any new cards that we will add in the future — as soon as we add them to Koenig and bump `kg-default-nodes` in Ghost, this test will fail, prompting us to add the new card to the Golden Post and update the snapshots.

We should also run the 'Golden Post' through a test in Litmus, which allows us to visually inspect the rendered email across many different email clients. Ideally we would create a process to review the output of the 'Golden Post' in Litmus whenever we update the snapshot as well.

### DIFF
--- a/ghost/core/test/integration/services/email-service/README.md
+++ b/ghost/core/test/integration/services/email-service/README.md
@@ -1,0 +1,12 @@
+# What is a golden post?
+The golden post is a single lexical post that has at least one example of every card that is available in the Koenig editor (with a few exceptions for cards that should never make it into an email). We have run into problems in the past where a small change to a particular card or to the EmailRenderer itself results in seriously mangled email rendering in one or more clients (usually Outlook).
+
+# How do I update the golden post to include a new card?
+If you're seeing a failing test like `The golden post does not contain the ${card} card`, that means that you (or someone else) has added a new node to `@tryghost/kg-default-nodes` that is not currently represented in the golden post. This test is here to trigger a review of the rendered email of the new card, to make sure it doesn't break the formatting in email clients. To update this test properly, please do the following:
+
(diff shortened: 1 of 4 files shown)


cmraible:
Planning to rerun this test as the images appear to be broken in Litmus — I think this is just a copy/paste issue but will get back to this on Monday.

cmraible:
Need to put a bit more thought into this, specifically around how to _update_ this test. The test works in its current form, but updating will be a bit of a PITA so trying to think of a better way to do this.

The golden post is currently stored on main.ghost.org but that isn't guaranteed to always exist. It's easy enough to keep the lexical string up to date since it's stored in a fixture, but the images and other attachments aren't included.

PR #19063, opened 2023-11-21
Added more Sentry logging in BatchSendingService

cmraible:
no issue

- Added more Sentry logging in BatchSendingService so we can more easily setup alerts for when something goes wrong


cmraible:
Closed in favor of https://github.com/TryGhost/Ghost/pull/19321

PR #19071, opened 2023-11-21
Fixed decimal places in currencies

ronaldlangeveld:
no issue


ronaldlangeveld:
Fixed here https://github.com/TryGhost/Ghost/pull/19125

PR #19085, opened 2023-11-22
Optimized active newsletter members query to remove subquery

daniellockyer:
Had to change branch name but it's still available here: https://github.com/TryGhost/Ghost/compare/main...v5.74.2-wip


daniellockyer:
Had to change branch name but it's still available here: https://github.com/TryGhost/Ghost/compare/main...v5.74.2-wip

PR #19094, opened 2023-11-22
Added email address alignment protections

SimonBackx:
ref GRO-54
fixes GRO-63
fixes GRO-62
fixes GRO-69

When the config `hostSettings:managedEmail:enabled` is enabled, or the new flag (`newEmailAddresses`) is enabled for self-hosters, we'll start to check the from addresses of all outgoing emails more strictly.

- Current flow: nothing changes if the managedEmail config is not set or the `newEmailAddresses` feature flag is not set
- When managedEmail is enabled: never allow to send an email from any chosen email. We always use `mail.from` for all outgoing emails. Custom addresses should be set as replyTo instead. Changing the newsletter sender_email is not allowed anymore (and ignored if it is set).
- When managedEmail is enabled with a custom sending domain: if a from address doesn't match the sending domain, we'll default to mail.from and use the original as a replyTo if appropriate and only when no other replyTo was set. A newsletter sender email addresss can only be set to an email address on this domain.
- When `newEmailAddresses` is enabled: self hosters are free to set all email addresses to whatever they want, without verification. In addition to that, we stop making up our own email addresses and send from `mail.from` by default instead of generating a `noreply`+ `@` + `sitedomain.com` address

A more in depth example of all cases can be seen in `ghost/core/test/integration/services/email-addresses.test.js`

Includes lots of new E2E tests for most new situations. Apart from that, all email snapshots are changed because the from and replyTo addresses are now included in snapshots (so we can see unexpected changes in the future).

Dropped test coverage requirement, because tests were failing coverage locally, but not in CI

Fixed settings test that set the site title to an array - bug tracked in GRO-68

### DIFF
--- a/ghost/core/core/server/services/newsletters/NewslettersService.js
+++ b/ghost/core/core/server/services/newsletters/NewslettersService.js
@@ -249,10 +249,10 @@ class NewslettersService {
         const cleanedAttrs = _.cloneDeep(attrs);
         const emailsToVerify = [];
         const emailProperties = [
-            {property: 'sender_email', type: 'from'}
+            {property: 'sender_email', type: 'from', emptyable: true, error: messages.senderEmailNotAllowed}
         ];
 
-        if (!this.emailAddressService.useNewEmailAddresses) {
+        if (!this.emailAddressService.service.useNewEmailAddresses) {
             // Validate reply_to is either newsletter or support
             if (cleanedAttrs.sender_reply_to !== undefined) {
                 if (!['newsletter', 'support'].includes(cleanedAttrs.sender_reply_to)) {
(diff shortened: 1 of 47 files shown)


ghost/core/core/server/services/newsletters/NewslettersService.js
```diff
@@ -243,14 +247,28 @@ class NewslettersService {
     async prepAttrsForEmailVerification(attrs, newsletter) {
         const cleanedAttrs = _.cloneDeep(attrs);
         const emailsToVerify = [];
+        const emailProperties = [
+            {property: 'sender_email', type: 'from'}
+            //{property: 'sender_reply_to', type: 'replyTo'}
```
sagzy, line 252: Leftover

SimonBackx: On my radar 😊

PR #19097, opened 2023-11-22
Forced use of index on `members_newsletters`

daniellockyer:
refs https://github.com/TryGhost/Product/issues/4181

- we've seen MySQL change its query planner to use a different index
  than the ideal one, resulting in drastically slower query performance
  when fetching newsletters with the member count
- this forces the use of the ideal index on MySQL
- this kept many of the Ghost team up overnight, so I hope MySQL is
  happy


naz (approved):
worked like a charm on both MySQL and SQLite 🚀

PR #19147, opened 2023-11-27
Updated support email address verification for DMARC changes

SimonBackx:
fixes GRO-71

- Current flow: unchanged
- New managed flow: verification required
- New managed flow with custom sending domain: only verification required for different domains
- Self hosters (feature flag): no verification required

### DIFF
--- a/ghost/core/core/server/services/newsletters/NewslettersService.js
+++ b/ghost/core/core/server/services/newsletters/NewslettersService.js
@@ -330,11 +330,11 @@ class NewslettersService {
             fromEmail = `no-reply@${toDomain}`;
         }
 
-        if (this.emailAddressService.useNewEmailAddresses) {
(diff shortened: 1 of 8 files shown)


sagzy (commented):
Code looks good! Did a round of QA and things work as expected.

One edge case that is not handled in the UI: what happens if the publisher clears the support email address field (= empty field)? Currently, the backend returns an error (Please enter a valid email address.) but nothing is surfaced in the UI.

I see two ways of solving this:
1. Either surface this error in the UI
2. Or, onsider empty address field = default value. Allow the publisher to clear the support email address field, and in that case, save the default value `noreply` in the DB

SimonBackx:
@sagzy Thanks! I checked but the empty email address value bug was already present (probably introduced with the new settings UI) so I'm not going to tackle that in this PR.
I created two new issues to track these: GRO-93 and GRO-92

PR #19149, opened 2023-11-27
Add Lexical support to the external-media-inliner

PaulAdamDavis:
This adds Lexical support to the external media inliner package. 

The bulk of the changes here are determining what format of the content is available, which is then used to decide what post property to change & update.

### DIFF
--- a/ghost/external-media-inliner/lib/ExternalMediaInliner.js
+++ b/ghost/external-media-inliner/lib/ExternalMediaInliner.js
@@ -261,24 +261,24 @@ class ExternalMediaInliner {
                 const mobiledocContent = post.get('mobiledoc');
                 const lexicalContent = post.get('lexical');
 
-                let theContent;
-                let contentTarget;
-
-                if (mobiledocContent) {
-                    theContent = mobiledocContent;
-                    contentTarget = 'mobiledoc';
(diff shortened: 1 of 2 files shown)


ghost/external-media-inliner/lib/ExternalMediaInliner.js
```diff
@@ -250,11 +258,27 @@ class ExternalMediaInliner {
...
         for (const post of posts) {
             try {
-                const inlinedMobiledoc = await this.inlineMobiledoc(post.get('mobiledoc'), domains);
+                const mobiledocContent = post.get('mobiledoc');
+                const lexicalContent = post.get('lexical');
+
+                let theContent;
+                let contentTarget;
+
+                if (mobiledocContent) {
```
naz, line 267: I'm not really up to date on how we store both formats - is there a possibility both would not be null? I think because `lexical` is a preferred one it would be a better practice to check if it exists first and do the manipulation on it. 

Also, if it is the case that both format can be not null on the post, we'd probably need to process both?

PaulAdamDavis: I've added support for processing both, if present. The code is now cleaner & makes more sense too!

PR #19155, opened 2023-11-28
Updated newsletter settings GRO-76 GRO-90

djordjevlais:
refs GRO-76, GRO-90

- Reused `renderReplyToEmail` and `renderSenderEmail` from `NewsletterDetailModal.tsx` like Simon suggested
- Changed Newsletter preview header in Newsletter settings to show Reply-to address instead of the “To” mock placeholder
- When no custom sending domain is set, switched “Sender email address” to disabled TextField for better visual rhythm
- Added a link to the help doc for (Pro) users without custom sending domain set (the doc itself will still need to be updated though, but Sam and the gang are already aware of this)
- When custom sending domain is set, added a hint about the default value, which IMO is clearer than using a placeholder


apps/admin-x-settings/src/components/settings/email/newsletters/NewsletterPreview.tsx
```diff
@@ -113,6 +113,7 @@ const NewsletterPreview: React.FC<{newsletter: Newsletter}> = ({newsletter}) =>
         headerTitle={headerTitle}
         senderEmail={renderSenderEmail()}
         senderName={newsletter.sender_name || title}
+        senderReplyTo={newsletter.sender_reply_to}
```
SimonBackx, line 116: This value can be `newsletter` or `support` too, so to avoid showing those values, we'll need to map it first to the right email address.

You can use the `renderReplyToEmail` method from `apps/admin-x-settings/src/components/settings/email/newsletters/NewsletterDetailModal.tsx` to do this

SimonBackx: If you implement this, maybe also try to remove the duplicate implementation of `renderSenderEmail`. And use the one from `apps/admin-x-settings/src/components/settings/email/newsletters/NewsletterDetailModal.tsx` instead (this one is more correct)

djordjevlais: Okay, I did both, I _think_ it went okay, but could use another look from you and @sagzy 🙏

PR #19156, opened 2023-11-28
Indexed over members_newsletters(newsletter_id, member_id)

allouis:
We were seeing slow queries when joining on this table, and the index speeds them up. The down migration is tricky because when we add the index MySQL can optimise away some `KEY` indexes on the `newsletter_id` column. When we then go to remove the newly created index, there is no index for the FK!

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.75/2023-11-27-15-55-add-members-newsletters-index.js
+++ b/ghost/core/core/server/data/migrations/versions/5.75/2023-11-27-15-55-add-members-newsletters-index.js
@@ -12,16 +12,17 @@ module.exports = createNonTransactionalMigration(
         try {
             await dropIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
         } catch (err) {
-            logging.error({
-                err,
-                message: 'Error dropping index over members_newsletters(newsletter_id, member_id)'
-            });
+            if (err.code === 'ER_DROP_INDEX_FK') {
+                logging.error({
+                    message: 'Error dropping index over members_newsletters(newsletter_id, member_id), re-adding index for newsletter_id'
+                });
 
-            logging.info('Creating index over members_newsletters(newsletter_id)');
-            await addIndex('members_newsletters', ['newsletter_id'], knex);
 
-            logging.info('Dropping index over members_newsletters(newsletter_id, member_id)');
-            await dropIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
+                await addIndex('members_newsletters', ['newsletter_id'], knex);
+                await dropIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
+	         }
+	         
+	         throw err;
         }
     }
 );
--- a/ghost/core/core/server/data/migrations/versions/5.75/2023-11-27-15-55-add-members-newsletters-index.js
+++ b/ghost/core/core/server/data/migrations/versions/5.75/2023-11-27-15-55-add-members-newsletters-index.js
@@ -1,19 +1,16 @@
 const logging = require('@tryghost/logging');
 const {createNonTransactionalMigration} = require('../../utils');
+const {addIndex, dropIndex} = require('../../../schema/commands');
 
 module.exports = createNonTransactionalMigration(
     async function up(knex) {
         logging.info('Adding index over members_newsletters(newsletter_id, member_id)');
-        await knex.schema.alterTable('members_newsletters', function (table) {
-            table.index(['newsletter_id', 'member_id'], 'idx_members_newsletters_newsletter_id_member_id');
-        });
+        await addIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
     },
     async function down(knex) {
         logging.info('Dropping index over members_newsletters(newsletter_id, member_id)');
         try {
-            await knex.schema.alterTable('members_newsletters', function (table) {
-                table.dropIndex('idx_members_newsletters_newsletter_id_member_id');
-            });
+            await dropIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
         } catch (err) {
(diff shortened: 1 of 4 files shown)


allouis:
God damn SQLite - will address that

daniellockyer:
@allouis We have a few utils to re-use here, as they already do some detection around missing indexes - how about:

```js
const COLUMNS = ['newsletter_id', 'member_id'];

module.exports = createNonTransactionalMigration(
    async function up(knex) {
        await commands.addIndex('members_newsletters', COLUMNS);
    },
    async function down(knex) {
        logging.info('Dropping index over members_newsletters(newsletter_id, member_id)');
        try {
            await commands.dropIndex('members_newsletters', COLUMNS);
        } catch (err) {
            logging.error({
                err,
                message: 'Error dropping index over members_newsletters(newsletter_id, member_id)'
            });

            await commands.addIndex('members_newsletters', 'newsletter_id');
            await commands.dropIndex('members_newsletters', COLUMNS);
        }
    }
);
```

daniellockyer:
IDK if that solves the SQLite issue though

ghost/core/core/server/data/schema/schema.js
```diff
@@ -936,7 +936,10 @@ module.exports = {
     members_newsletters: {
         id: {type: 'string', maxlength: 24, nullable: false, primary: true},
         member_id: {type: 'string', maxlength: 24, nullable: false, references: 'members.id', cascadeDelete: true},
-        newsletter_id: {type: 'string', maxlength: 24, nullable: false, references: 'newsletters.id', cascadeDelete: true}
+        newsletter_id: {type: 'string', maxlength: 24, nullable: false, references: 'newsletters.id', cascadeDelete: true},
+        '@@INDEXES@@': [
+            ['member_id', 'newsletter_id']
```
daniellockyer: Can we switch these around to reflect changes done to live instances?

ghost/core/core/server/data/migrations/versions/5.75/2023-11-27-15-55-add-members-newsletters-index.js
```diff
@@ -0,0 +1,27 @@
...
+            logging.error({
+                err,
+                message: 'Error dropping index over members_newsletters(newsletter_id, member_id)'
+            });
+
+            logging.info('Creating index over members_newsletters(newsletter_id)');
+            await addIndex('members_newsletters', ['newsletter_id'], knex);
+
+            logging.info('Dropping index over members_newsletters(newsletter_id, member_id)');
+            await dropIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
```
daniellockyer, line 24:
```suggestion
            if (err.code === 'ER_DROP_INDEX_FK') {
                logging.error({
                    message: 'Error dropping index over members_newsletters(newsletter_id, member_id), re-adding index for newsletter_id'
                });


                await addIndex('members_newsletters', ['newsletter_id'], knex);
                await dropIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
	         }
	         
	         throw err;
```

Both the `addIndex` and `dropIndex` utils log out so we don't need to dupe the effort

ghost/core/core/server/data/migrations/versions/5.75/2023-11-27-15-55-add-members-newsletters-index.js
```diff
@@ -0,0 +1,28 @@
...
+        } catch (err) {
+            if (err.code === 'ER_DROP_INDEX_FK') {
+                logging.error({
+                    message: 'Error dropping index over members_newsletters(newsletter_id, member_id), re-adding index for newsletter_id'
+                });
+
+
+                await addIndex('members_newsletters', ['newsletter_id'], knex);
+                await dropIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
+	         }
```
daniellockyer, line 23:
```suggestion
                await dropIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
                return;
	         }
```

ghost/core/core/server/data/migrations/versions/5.75/2023-11-27-15-55-add-members-newsletters-index.js
```diff
@@ -0,0 +1,29 @@
...
+const {createNonTransactionalMigration} = require('../../utils');
+const {addIndex, dropIndex} = require('../../../schema/commands');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        logging.info('Adding index over members_newsletters(newsletter_id, member_id)');
+        await addIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
+    },
+    async function down(knex) {
+        logging.info('Dropping index over members_newsletters(newsletter_id, member_id)');
```
daniellockyer, line 11:
```suggestion
        await addIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
    },
    async function down(knex) {
```

ghost/core/core/server/data/migrations/versions/5.75/2023-11-27-15-55-add-members-newsletters-index.js
```diff
@@ -0,0 +1,27 @@
...
+
+
+                await addIndex('members_newsletters', ['newsletter_id'], knex);
+                await dropIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
+                return;
+	         }
+	         
+	         throw err;
+        }
+    }
```
daniellockyer, line 26:
```suggestion

                await addIndex('members_newsletters', ['newsletter_id'], knex);
                await dropIndex('members_newsletters', ['newsletter_id', 'member_id'], knex);
                return;
             }
	         
             throw err;
        }
    }
```

daniellockyer (approved):

PR #19163, opened 2023-11-28
Updated Portal to use calculated support and default email addresses

sagzy:
fixes GRO-72

- added "default_email_address" and "support_email_address" to the public settings
- when available, use these addresses in Portal. Otherwise, fallback to current logic


SimonBackx (approved):
🔥

PR #19229, opened 2023-12-04
Updated logic to allow a different name for the free tier

sagzy:
fixes PROD-69

### DIFF
--- a/apps/portal/src/utils/helpers.js
+++ b/apps/portal/src/utils/helpers.js
@@ -451,6 +451,11 @@ export function getFreeProductBenefits({site}) {
 
 export function getFreeTierTitle({site}) {
     const freeProduct = getFreeProduct({site});
+
+    if (freeProduct?.name === 'Free' && hasOnlyFreeProduct({site})) {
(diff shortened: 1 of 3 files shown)


apps/portal/src/utils/helpers.js
```diff
@@ -450,11 +450,8 @@ export function getFreeProductBenefits({site}) {
 }
 
 export function getFreeTierTitle({site}) {
-    if (hasOnlyFreeProduct({site})) {
-        return 'Free membership';
-    } else {
-        return 'Free';
-    }
+    const freeProduct = getFreeProduct({site});
+    return freeProduct?.name || 'Free';
```
SimonBackx, line 454: Think we should keep this backwards compatible by also returning `Free membership` if there is only one free product, otherwise the title will change before the feature is released

sagzy: Updated in [8891e61](https://github.com/TryGhost/Ghost/pull/19229/commits/8891e61097c710530e688189f547255ef41db1a7)

PR #19234, opened 2023-12-04
Added kelonio performance testing

cmraible:
no issue

- This PR introduces mocha performance tests to Ghost. Kelonio is a 3rd party library that provides functionality for running an arbitrary function `n` times and making assertions on the mean/min/max/std.dev execution time across all iterations.
- The goal is to create a robust and repeatable method to evaluate different queries (or other arbitrary long running code) for performance. It also gives us a toolset to protect against performance regressions when we introduce new code as the test suite will fail if any of the performance targets aren't met.
- Having a repeatable test suite like this should make it easier to evaluate Ghost's performance across a variety of different environments (self hosted vs Pro vs K8s...) and a variety of use-cases (content heavy installations, subscriber heavy installations, small/medium/large sites) without having to do a bunch of manual testing.
- This PR is a prototype and isn't (yet) meant to be a useful test suite — the tests included in the `newsletters.perf.test.js` file are simply examples to demonstrate how this could work.
- To run the performance test suite, run `yarn test:perf` from within `ghost/core`. The thresholds are currently set to intentionally create a failure to demonstrate how it works when a performance constraint is not met.


cmraible:
Example test output (with a failing test):

CleanShot 2023-12-04 at 16 44 13@2x

cmraible:
@daniellockyer curious to hear if you have any thoughts about this! I know you did some similar work with the boot time performance tests, so not sure if you had anything in mind for checking performance for specific API endpoints.

cmraible:
Closing PR for now, but keeping the branch for future reference. Will look back into performance testing options in the new year (a few alternatives to kelonio are available, so it's worth trying those out)

PR #19238, opened 2023-12-05
Added portal default plan setting

SimonBackx:
fixes PROD-61

This adds a new default plan setting. It defaults to yearly, which is the current default selected interval in Portal.

Behind the new portal improvements feature flag, the default plan can be changed. It will also change automatically if the available intervals are changed.

This PR also wires up passing the new setting to the Portal preview.


sagzy (approved):
LGTM :)

PR #19243, opened 2023-12-05
Added Sentry performance monitoring to Ghost Server

cmraible:
refs ARCH-21

- We currently have NewRelic setup for a few of our largest customers for monitoring performance, but it is too expensive to enable across all sites
- Sentry has similar (but simpler) performance monitoring tools to keep track of response times that are available to us for free, but we just haven't configured them
- This PR sets up Sentry Performance monitoring for API requests so we can have one place for monitoring errors + performance so we can stay on top of response times more easily.
- Tracing is disabled by default, so there is no additional overhead unless `sentry.tracing.enabled` is set to `true` in the site's config. Additionally, `sentry.tracing.sampleRate` should be set to a decimal value between 0 and 1. This value defaults to 0 to avoid accidentally blowing through quota, and requires a value to explicitly be set in order to send the traces to Sentry.

### DIFF
--- a/ghost/core/core/app.js
+++ b/ghost/core/core/app.js
@@ -30,8 +30,9 @@ const maintenanceMiddleware = (req, res, next) => {
 const rootApp = () => {
     const app = express('root');
     app.use(sentry.requestHandler);
-    app.use(sentry.tracingHandler);
-
+    if (config.get('sentry')?.tracing?.enabled === true) {
+        app.use(sentry.tracingHandler);
+    }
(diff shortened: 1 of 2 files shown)


daniellockyer (changes requested):
I think we should make tracing opt-in via config `sentryConfig.tracing === true` or something. Tracing isn't as "cheap" in Sentry as error capturing, so we can enable it on a few select sites first

daniellockyer (approved):

PR #19245, opened 2023-12-05
Added Spanish locale for Portal

estaribel:
I made some changes to make certain phrases sound more natural (some expressions where directly translated from english and while they work I believe there are better options!). I am a spanish speaker from Spain so the dialect might be reflected in some instances –changed "acá" for "aquí", both are equally correct to my knowledge–. Cheers!


9larsons:
Closing due to age and unresponsiveness. Please open a new submission if you have translations to contribute.

PR #19272, opened 2023-12-07
Added initial permissions implementation

allouis:
This is the standard Nest approach, with a few things missing.

1. A decorator to determine which routes require which permissions
2. A "Guard" which restricts access to the route based on the metadata set by the decorator, and the authorized context
3. A Middleware (or Guard) to load the authorized context 

We are missing 3. here - because that is already handled by Ghost. 

This implementation is not complete as it doesn't take into account api keys.


ghost/core/core/server/services/auth/api-key/admin.js
```diff
@@ -60,7 +61,7 @@ const authenticate = function apiKeyAdminAuth(req, res, next) {
         }));
     }
 
-    return authenticateWithToken(req, res, next, {token, JWT_OPTIONS: JWT_OPTIONS_DEFAULTS});
```
naz, line 63: RE: JWT_OPTIONS_DEFAULTS - do we have sufficient test coverage to make sure this change doesn't break anything? I saw the `ignoreMaxAge` being passed around explicity now. 

Ideally this type of change should land in `main` and then we'd rebase over it, because we have to idea when this `arch` branch will get merged and the change would be good to validate individually, not as a part of this big refactor 🤔

allouis: It's only used in these two functions, and one of them omits the `maxAge` option. To me this is a small enough change to see that the logic is the same. I can look at adding some more tests if you disagree tho

Agree that we shouldn't put this change into `main` as part of a huge refactor, but not sure if _now_ is the right time to get it into `main`?

ghost/ghost/src/index.ts
```diff
@@ -10,6 +11,11 @@ export async function create() {
     return app;
 }
 
+export function addProvider(obj: ClassProvider | ValueProvider) {
+    AppModule.providers?.push(obj);
+    AppModule.exports?.push(obj.provide);
```
naz, line 16: Exposing as part of exports is only here to be able to configure the dependency incjection, right? As in we should not be referencing the app module directly from anywhere in the http code? 

If that's the case maybe exposing a more specific separate variable like `InjectableValues` or something more descriptive would be better 🤔  

I'm mostly guessing here and trying to understand why are we pushing straight to the `exports`

ghost/core/core/server/web/api/endpoints/admin/routes.js
```diff
@@ -329,15 +329,11 @@ module.exports = function apiRoutes() {
...
-    router.get('/snippets', mw.authAdminApi, labs.skipMiddleware('nestSnippetsAPI', http(api.snippets.browse)));
-    //router.get('/snippets/:id', mw.authAdminApi, http(api.snippets.read));
-    router.get('/snippets/:id', mw.authAdminApi, labs.skipMiddleware('nestSnippetsAPI', http(api.snippets.read)));
-    //router.post('/snippets', mw.authAdminApi, http(api.snippets.add));
-    router.post('/snippets', mw.authAdminApi, labs.skipMiddleware('nestSnippetsAPI', http(api.snippets.add)));
-    //router.put('/snippets/:id', mw.authAdminApi, http(api.snippets.edit));
-    router.put('/snippets/:id', mw.authAdminApi, labs.skipMiddleware('nestSnippetsAPI', http(api.snippets.edit)));
-    //router.del('/snippets/:id', mw.authAdminApi, http(api.snippets.destroy));
-    router.del('/snippets/:id', mw.authAdminApi, labs.skipMiddleware('nestSnippetsAPI', http(api.snippets.destroy)));
+    router.get('/snippets', labs.skipMiddleware('nestSnippetsAPI', mw.authAdminApi), labs.skipMiddleware('nestSnippetsAPI', http(api.snippets.browse)));
```
naz, line 332: oh look the `skipMiddleware` is useful! 😄

ghost/ghost/src/common/decorators/permissions.decorator.ts
```diff
@@ -0,0 +1,8 @@
+import {Reflector} from '@nestjs/core';
+
+type UserRole = 'Contributor' | 'Author' | 'Editor' | 'Admin' | 'Owner';
```
naz, line 3: Maybe add a note here where these values are coming from - the `schema.json` in the legacy codebase. Same for the rest in the file. Just so the future us know where to clean up

ghost/ghost/src/http/guards/admin-api-authentication.guard.ts
```diff
@@ -0,0 +1,100 @@
...
+    ExecutionContext,
+    Inject
+} from '@nestjs/common';
+import {Request, Response} from 'express';
+import {Actor} from '../../common/types/actor.type';
+import ObjectID from 'bson-objectid';
+
+// Here we extend the express Request interface with our new type
+declare global {
+    // eslint-disable-next-line @typescript-eslint/no-namespace
```
naz, line 13: If we are actually using the namespacing - worth relaxing the eslint rule globally? Otherwise leave a comment about why and the benefit of namespacing in this situation?

ghost/ghost/src/http/guards/admin-api-authentication.guard.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    // This is `any` because again it represents a bookshelf model
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private async setUserActor(user: any, request: Request) {
+        await user.related('roles').fetch();
+        const json = user.toJSON();
+        request.actor = {
+            // BS To work around Owner id === 1
+            id: ObjectID.createFromHexString(json.id === '1' ? 'abcd12345678901234567890' : json.id),
```
naz, line 95: oh dear lord. the `1` becomes `abcd12345678901234567890` in the new codebase 🙈

ghost/ghost/src/index.ts
allouis: This is not the "node" concept of `module.exports` - this is the Nest concept of `exports` which is a list of providers which are able to be consumed from outside of the module

ghost/ghost/src/http/guards/admin-api-authentication.guard.ts
allouis: Hahahaha - I'm thinking now we should use `'deadbeefdeadbeefdeadbeef'` 😁

allouis: Or `dead01d0dead01d0dead01d0` (dead id)

naz: `deadc0de` would also be noice! finally an acceptable way to use hexspeak 😂

ghost/ghost/src/nestjs/admin-api.module.ts
```diff
@@ -0,0 +1,35 @@
...
+        {
+            provide: 'SnippetsRepository',
+            useClass: KnexSnippetsRepository
+        },
+        SnippetsService
+    ]
+})
+export class AdminAPIModule {}
+
+/*
```
naz, line 19: This implementation is not working or is it an example equivalent implementation but without the decorator?

ghost/ghost/src/http/admin/controllers/snippets.controller.ts
```diff
@@ -7,13 +7,25 @@ import {now} from '../../../common/helpers/date.helper';
...
+import {AdminAPIAuthentication} from '../../guards/admin-api-authentication.guard';
 
 @Controller('snippets')
+@UseGuards(AdminAPIAuthentication, PermissionsGuard)
 @UseInterceptors(LocationHeaderInterceptor)
 @UseFilters(GlobalExceptionFilter)
 export class SnippetsController {
     constructor(private readonly service: SnippetsService) {}
 
+    @Roles([
```
naz, line 57: I think even nicer way could be having a controller level `@Roles` decorator and then combine that with method level decorators, to reduce the syntax. But this way is perfectly fine as it's clear straight away "who-can-do-what?"

naz:
I think I've done the best I can in glancing through all of this. It's a great example of a bridge between the legacy code and the new ways! The commends I've left are nice-to-haves, so feel free to merge at will

PR #19273, opened 2023-12-07
Nest native serialization with class-validator/class-transformer

naz:
A spike for request data validation to be used in controllers


ghost/ghost/src/http/pipes/filter.pipe.ts
```diff
@@ -0,0 +1,10 @@
+import {PipeTransform, Injectable} from '@nestjs/common';
+
+@Injectable()
+export class ParseFilterQueryPipe implements PipeTransform {
+    // @NOTE: the return type here should ideally be a typed NQL object (ask Egg why!)
```
allouis, line 5: 😂

ghost/ghost/src/http/controllers/snippet.dto.input.ts
```diff
@@ -0,0 +1,26 @@
...
+
+class SnippetDTO {
+    @IsString()
+    @IsNotEmpty()
+    @MaxLength(191)
+        name?: string;
+
+    @IsString()
+    @IsOptional()
+        lexical?: string;
```
allouis, line 12: It's annoying that `lexical?: string` literally means it's a string and it's optional, but we have to add decorators for both of these things 😕

naz: Yeah... without "IsOptional" it failed the "IsString" validation. Quite verbose this validator is

PR #19276, opened 2023-12-07
Added yearly discount indicator to monthly/yearly toggle PROD-60 PROD-222 PROD 223

djordjevlais:
refs PROD-60, PROD-222, PROD-223, PROD-89, PROD-94

- Indicator shows up in the monthly/yearly toggle if there are any paid tiers with yearly discount, and shows the highest discount in order to nudge visitors towards checking out yearly plan
- A couple of smaller portal improvements: typos, spacing, alignment


apps/portal/src/components/common/ProductsSection.js
```diff
@@ -847,9 +858,19 @@ function YearlyDiscount({discount}) {
     }
 }
 
-function ProductPriceSwitch({selectedInterval, setSelectedInterval}) {
+function ProductPriceSwitch({selectedInterval, setSelectedInterval, products}) {
```
SimonBackx, line 861: Products won't be always set, if you search on `ProductPriceSwitch`, it isn't set in the `ChangeProductSection` component. So you'll need to add it there too

djordjevlais: @SimonBackx hm but it seems `products` is already passed to `ChangeProductSection`, and when I go to “Change plans” I can see the indicator. Or where you thinking about something else?
image

apps/portal/src/components/common/ProductsSection.js
```diff
@@ -847,9 +858,19 @@ function YearlyDiscount({discount}) {
...
     const {site, t} = useContext(AppContext);
     const {portal_plans: portalPlans} = site;
+
+    // Filter only paid products
+    const paidProducts = products.filter(product => product.id !== 'free');
```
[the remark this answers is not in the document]
djordjevlais: @SimonBackx I tried this, but it complained that `amount` is undefined in case of free product, so I’ve added `?` like this `product.monthlyPrice?.amount` to check if it’s undefined. Makes sense? 🙏

apps/portal/src/components/common/ProductsSection.js
SimonBackx: I meant that here it is not passed yet: https://github.com/TryGhost/Ghost/blob/c969dd18a8edd07fc22b6ad6774a4c82dd73460d/apps/portal/src/components/common/ProductsSection.js#L1031-L1034

You'll need to pass the products parameter to the `ProductPriceSwitch` component inside the `ChangeProductSection` linked above

apps/portal/src/components/common/ProductsSection.js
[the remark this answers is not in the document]
SimonBackx: I would just keep the filter, but change it to what I suggested

PR #19315, opened 2023-12-11
Added custom Sentry integration for Knex.js

mike182uk:
no refs

Added custom Sentry integration for Knex.js to trace database queries in Sentry


cmraible:
Nice one! LGTM

PR #19319, opened 2023-12-11
Added Sentry Profiling to Ghost server

cmraible:
refs ARCH-29

- Added Sentry Profiling to collect more detailed performance data on the backend. 
- This feature is opt-in behind a config. To enable profiling, first enable tracing with `sentry.tracing.enabled: true`, then set `sentry.profiling.enabled: true` and `sentry.profiling.sampleRate` to a decimal number between 0 and 1.

### DIFF
--- a/ghost/core/package.json
+++ b/ghost/core/package.json
@@ -59,7 +59,6 @@
   "dependencies": {
     "@extractus/oembed-extractor": "3.2.1",
     "@sentry/node": "7.86.0",
-    "@sentry/profiling-node": "^1.2.6",
     "@tryghost/adapter-base-cache": "0.1.10",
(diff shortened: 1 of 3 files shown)


ghost/core/package.json
```diff
@@ -59,6 +59,7 @@
   "dependencies": {
     "@extractus/oembed-extractor": "3.2.1",
     "@sentry/node": "7.86.0",
+    "@sentry/profiling-node": "^1.2.6",
```
daniellockyer, line 62: I __think__ this contains native binaries. Sometimes these are a pain and might not be installed by Yarn (when they fail to build/download). We could move this to `optionalDependencies` and try-catch the require?

cmraible: Thanks — updated ✅

PR #19321, opened 2023-12-12
Added Sentry logging in BatchSendingService

cmraible:
refs ARCH-25

- Added a log message if the email_count on an email differs from the totalCount calculated while creating batches by more than 1%, so we can investigate further.


mike182uk (approved):
Nice one, looks good 👍

PR #19369, opened 2023-12-13
Fixed error when deleting post

kevinansfield:
closes https://github.com/TryGhost/Product/issues/4230

- deleting a post could cause React components to trigger save tasks during teardown which then threw errors because they attempt to set properties on a deleted model instance
- added checks to the `beforeSaveTask()` to abort if the post object has been deleted


danielraffel:
I seem to be able to reproduce #19563 in ghost 5.79.0

PR #19398, opened 2023-12-19
📮 Postmark integration

andreascreten:
I have been working on an integration of Postmark with Ghost. Please provide your feedback before I continue.

## To-Do

- [x] Implement analytics
- [ ] Handle Postmark errors
- [ ] Add tests


markstos:
Hello @andreascreten, I'm not a Ghost core team member, but am someone who previously submitted my own PR to allow non-mailgun senders ( https://github.com/TryGhost/Ghost/pull/14984 ), so I can give you some feedback from that perspective. (Ultimately, I ended up using Mailgun and abandoned that effort). 

I don't think there's a lot of appetite for maintaining non-Mailgun code in the core with their small team, but I think there /is/ interest in implementing an adapter pattern for bulk-email and email-analytics so other options can be plugged in and, crucially, completely maintained by third-parties outside of the core.

So I think the ideal PR implements the adapter pattern and has Mailgun as the included, default solution implementing that package. There would actually be no references to Postmark in the PR, but the PR would enable you make the swap using the config file system and third-party modules that implement bulk-sending and analytics.  The rest of your solution would in modules you maintain and possibly publish to NPM yourself. 

I see you made some progress in that direction replacing explicit references to MailGun with more generic "Bulk Email" or "Email Analytics" language when that part of the code has multiple possible implementations.  

## References

 * [Adapter Manager README](https://github.com/TryGhost/Ghost/blob/main/ghost/adapter-manager/README.md)
 * [Adapter Manager Code](https://github.com/TryGhost/Ghost/blob/98d49f57d52299a7363c1adb040abc0e273b3e88/ghost/adapter-manager/lib/AdapterManager.js)
 * [Other Ghost adapters on NPM](https://www.npmjs.com/search?q=ghost%20adapter)

What appears to be happening is that a path to where a custom solution is stored gets declared under the "adapters" section in the config file. If this is not present, the internal implementation in used.

## Admin UX

I saw your PR also took care of getting some Postmark bits to appear in the admin UI. I'm not sure the current adapter pattern covers that. What's there so far may only cover purely backend tasks, like swapping out data storage.  It's possible you need to extend the adapter system to allow modules that implement the adapter system to have bits appear in the admin UI when they are loaded.

Issue #1405, opened 2013-11-05
Static Asset Management


ErisDS:
It's time to upgrade Ghost's static asset system to something a little more heavy duty, so that we can handle more use cases, and also add proper cache control headers as discussed at length in issue #932 (closed in favour of this issue).

We need to be able to serve static assets for 
- Ghost admin
- themes 
- apps
- images

This should entirely replace or wrap our current usage of connect static.

We still need to be able to hot-swap themes.

This needs to be abstracted so that it uses local file storage by default, but so that we can provide a way for apps to override this. They may wish to use a CDN, or store files elsewhere.

Static assets need to be versioned so that the cache can be invalidated. This should be done using a hash of the file which is appended to the filename, something like (ghost.js -> ghost.aba23928.js)

We'll need a helper for outputting script and style tags for managed assets both in themes, and in the Ghost admin, something like:

* `{{asset "css/screen.css"}}` for themes, and if you're on Ghost it will look like 
* `{{asset "css/screen.css" ghost="true"}}` (or it could be {{asset "css/screen.css" "ghost"}})

It would be nice if this looked for a specific (assets/css, assets/js) structure by default, encouraging theme developers to structure their theme in a consistent way, but could be overridden.

ErisDS:
Closing this. At some point we'll probably want to add a much more extensive asset pipeline to Ghost, but what we have pretty much works for now. When it's time to add something better, we'll get it spec'd out in more detail.

smasher164:
I'm having trouble with browser caching of my assets. Because browsers don't cache files with a query string `?v=`, I think a possible fix would be to rev the file names differently. Instead of `mylogo.gif?v=1.2`, we should use `mylogo.1.2.gif`. This provides both the advantage of cache busting, as well as proxy caching of assets.

Looking through the helpers/ directory, `Ghost/core/server/helpers/utils.js` seems to be inserting the query string `<%= source %>?v=<%= version %>`. However, I'm not sure how to change it because `<%= source %>` contains the file extension.

ErisDS:
> Because browsers don't cache files with a query string

To the best of my knowledge, all modern browsers do cache files with a query string. Some network-level caches do not. Modifying the filename instead of adding a query string is a non-trivial change that would require a significant reworking of file lookup.

vvo:
:100: to what @ErisDS, browser do cache query strings. Still, not using query string is safer when behing proxies/CDNs etc.

PR #19438, opened 2024-01-04
Add theme config setting for `asset_hash`

joshperry:
The current asset hashing scheme is very naive and also pretty unflexible from a configuration standpoint. This particuler PR aims to address the latter.

Disabling the asset hashing (set to `false`) is particularly important for themes that may use a build step and can put hashes directly into the filenames themselves, effectively end-running ghost core with the implementing of content-based hashes.

The hash can also be set to a static value, so it could conceptually be a release hash which would be better than a random startup value in a lot of cases.

This is backwards compatible with the current behavior.

I also simplified a number of unit tests which were relying on an asset hash by disabling it for them.

A couple `getCompExpiry` unit tests in portal code are failing with off-by-one day errors on a date, I don't think my changes are responsible for that.

### DIFF
--- a/ghost/core/core/frontend/meta/asset-url.js
+++ b/ghost/core/core/frontend/meta/asset-url.js
@@ -44,13 +44,15 @@ function getAssetUrl(path, hasMinFile) {
     // Add the path for the requested asset
     output = urlUtils.urlJoin(output, path);
 
-    // Ensure we have an assetHash
-    // @TODO rework this!
-    if (!config.get('assetHash')) {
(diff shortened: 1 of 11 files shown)


joshperry:
Are you guys even interested in an effective asset hashing strategy? AFAICT ghost still uses a startup-random hash for assets, whether they've changed or not. I mean, maybe this project is dead and that's why you've slept on this feature... Perhaps a fork is in order to keep this codebase viable with modern frontend tech.

https://github.com/TryGhost/Ghost/issues/1405

PR #19441, opened 2024-01-04
Updated Sentry env to use PRO_ENV when available

mike182uk:
refs [ARCH-33](https://linear.app/tryghost/issue/ARCH-33/fix-sentry-environment)

To ensure that we are correctly identifying the environment that data is being sent to Sentry from, we can use the `PRO_ENV` environment variable if it is available. This will be set to `production` in production and `staging` in staging. If `PRO_ENV` is not available, we will fall back to retrieving the environment from config (`env`)


cmraible (approved):
Nice one! LGTM

PR #19445, opened 2024-01-04
Added support for using the Admin URL in Recommendations

sagzy:
fixes PROD-241
- when recommending a Ghost site, it's now possible to use either the Admin URL or the site URL

### DIFF
--- a/ghost/recommendations/src/RecommendationService.ts
+++ b/ghost/recommendations/src/RecommendationService.ts
@@ -142,17 +142,23 @@ export class RecommendationService {
     }
 
     async checkRecommendation(url: URL): Promise<Partial<RecommendationPlain>> {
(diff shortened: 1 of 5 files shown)


ghost/recommendations/src/RecommendationService.ts
```diff
@@ -142,17 +142,23 @@ export class RecommendationService {
...
+        }
+
+        const updatedUrl = metadata.url || url;
+        const existing = await this.repository.getByUrl(updatedUrl);
+
+        // If a recommendation with this URL already exists, return it
         if (existing) {
-            this._updateRecommendationMetadata(existing);
-            await this.repository.save(existing);
             return existing.plain;
```
SimonBackx, line 157: This no longer updates the existing recommendation in the database, why is that?

PR #19479, opened 2024-01-14
Added experimental filesystem-based cache adapter

daniellockyer:
refs https://www.notion.so/ghost/Filesystem-based-caches-dd7d6b6fe6924cb39a10d37f16ad3e7d
refs https://github.com/TryGhost/DevOps/issues/125

- this will serialize the cache to JSON files, which means its very
  simple versus alternatives that require a lot more setup
- this is a drop-in adapter with a full test suite

### DIFF
--- a/ghost/adapter-cache-fs/src/AdapterCacheFS.ts
+++ b/ghost/adapter-cache-fs/src/AdapterCacheFS.ts
@@ -1,0 +1,107 @@
+import crypto from 'crypto';
+import fs from 'fs-extra';
+import path from 'path';
+import Base from '@tryghost/adapter-base-cache';
+
+type AdapterOptions = {
+    cachePath?: string;
+}
+
+/**
+ * Filesystem-based cache adapter for Ghost
+ *
+ * To instantly invalidate the cache, we use a timestamp in the folder name. This means that
+ * when we want to invalidate the cache, we can just change the timestamp and we drop all knowledge
+ * of the old cache.
+ *
+ * To keep the overhead low, we only create the directory structure upon initialization, and then
+ * during the reset() method. ie. we don't do this for ever set() call. If we want to do further directory
+ * partitioning, this may need to change.
+ *
+ * It is up to the host system to clean up files periodically.
+ */
+export default class AdapterCacheFS extends Base {
+    // Keep a reset count to protect against resets occuring in the same millisecond and the cache not
+    // getting reset
+    #resetCount: number = 0;
+
+    // Initalize the timestamp with the current Date in ms
+    #timestamp: number = Date.now();
+
+    #cachePath: string;
+
+    constructor(options?: AdapterOptions) {
+        super();
+
+        this.#cachePath = options?.cachePath || '/tmp';
+
+        // Set up the initial cache folder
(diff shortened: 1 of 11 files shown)


ghost/adapter-cache-fs/src/AdapterCacheFS.ts
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * Also sets up the new cache folder so we don't have to set to do it in the .set() method.
+     */
+    async reset() {
+        this.#resetCount += 1;
+
+        const newTimestamp = Date.now();
+        this.#timestamp = newTimestamp;
+
+        await fs.mkdirp(this._getPrefixedCacheFolder(newTimestamp));
```
allouis, line 99: It may be possible to have a race condition where the timestamp and resetCount are updated, but this operation hasn't completed and so `set` will try writing to a non-existent directory

ghost/adapter-cache-fs/src/AdapterCacheFS.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    async keys(): Promise<string[]> {
```
allouis, line 102: So this is "deprecated" and only used by the settings cache, where we want to get _all_ of the values, so we do a `keys()` and then a `get(key)` on each of them; meaning filenames here will in fact be useless.

TBH settings _should_ know all of the keys, so those can probably be discerned from somewhere else in the code, rather than the cache.

ghost/adapter-cache-fs/src/AdapterCacheFS.ts
```diff
@@ -0,0 +1,107 @@
...
+ * of the old cache.
+ *
+ * To keep the overhead low, we only create the directory structure upon initialization, and then
+ * during the reset() method. ie. we don't do this for ever set() call. If we want to do further directory
+ * partitioning, this may need to change.
+ *
+ * It is up to the host system to clean up files periodically.
+ */
+export default class AdapterCacheFS extends Base {
+    // Keep a reset count to protect against resets occuring in the same millisecond and the cache not
```
allouis, line 24: Good thinking - though we wouldn't have this issue if we're to use a "random" string rather than a timestamp for the folder name? I realise we lose a simple way to see the most recent entries, but the filesystem will have that data 🤔

ghost/adapter-cache-fs/src/AdapterCacheFS.ts
```diff
@@ -0,0 +1,107 @@
...
+        this.#cachePath = options?.cachePath || '/tmp';
+
+        // Set up the initial cache folder
+        fs.mkdirpSync(this._getPrefixedCacheFolder());
+    }
+
+    _generateKey(key: string): string {
+        // md5 is not secure but it's fast and we're not using it for security
+        // the probability of a collision is low enough that we don't care
+        return crypto.createHash('md5').update(key).digest('hex');
```
allouis, line 45: We should probs do this for Redis too

allouis (commented):
Nice 'n' simple! I like it 💃 

Have you tested this on any Pro infra yet?

daniellockyer:
Although useful, closing for now

PR #19481, opened 2024-01-14
Added cache config to stats endpoints

SimonBackx:
no issue

Allows to enable cache via hostSettings.statsCache.enabled. This will need proper cache timeouts in order to function correctly.

Usage in config:
```
"hostSettings": {
        "statsCache": {
            "enabled": true
        }
    },
    "adapters": {
        "cache": {
            "Redis": {
                "host": "127.0.0.1",
                "port": 6379,
                "username": "",
                "password": "",
                "ttl": 60,
                "storeConfig": {
                    "maxRetriesPerRequest": 1,
                    "enableOfflineQueue": false,
                    "retryConnectSeconds": 60
                }
            },
            "stats": {
                "adapter": "Redis",
                "ttl": 3600,
                "refreshAheadFactor": 1,
                "keyPrefix": "site:123456:stats"
            }
        }
    },
    ```

### DIFF
--- a/ghost/core/core/server/api/endpoints/stats.js
+++ b/ghost/core/core/server/api/endpoints/stats.js
@@ -10,8 +10,14 @@ module.exports = {
             docName: 'members',
             method: 'browse'
         },
+        cache: statsService.cache,
+        generateCacheKeyData() {
+            return {
+                method: 'memberCountHistory'
+            };
+        },
         async query() {
-            return await statsService.getMemberCountHistory();
+            return await statsService.api.getMemberCountHistory();
         }
     },
     mrr: {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: ghost/core/core/server/api/endpoints/stats.js [hunk 1]
```diff
@@ -34,8 +46,12 @@ module.exports = {
             docName: 'members',
             method: 'browse'
         },
+        cache: statsService.cache,
+        generateCacheKeyData() {
+            return 'getSubscriptionCountHistory';
```
allouis, line 51: I think this should be an object

allouis (approved):

ghost/core/core/server/api/endpoints/stats.js
SimonBackx: My bad!

benjaminrancourt:
Hi @SimonBackx,

Would it be possible to also update the documentation (https://ghost.org/docs/config/#cache-adapters)?

I'm wondering if we should use the Redis cache for other services, but the current documentation only mentions the `imageSizes` cache.

https://github.com/search?q=repo%3ATryGhost%2FGhost+getAdapter%28%27cache%3A&type=code

Thanks!

PR #19486, opened 2024-01-15
Add one click deploy to readme

jbgh:
Hey!

We built this one-click deploy link for your users who may not know how to stand up their own infrastructure to easily self-host this.

After deployment users receive a link which can then be CNAME-ed to any domain.

Take it for a test run—we're eager to receive your feedback! We're more than willing to maintain this for the long haul, if this is valuable for your users.


royalfig:
This looks like a cool project, but the README isn't quite the right place to share it. Please head on over to our Forum (https://forum.ghost.org) and make a post about it there. It'll be much more visible to the community and you'll have more ability to explain your setup.

PR #19499, opened 2024-01-17
Added support for "Refresh Ahead" caching strategy

allouis:
The main changes are:
- Updating the pipeline to allow for doing a background refresh of the cache
- Remove the use of the EventAwareCacheWrapper for the posts public cache

### Background refresh

This is just an initial implementation, and tbh it doesn't sit right with me that the logic for this is in the pipeline - I think this should sit in the cache implementation itself, and then we call out to it with something like: `cache.get(key, fetchData)` and then the updates can happen internally.

The `cache-manager` project actually has a method like this called `wrap` - but every time I've used it it hangs, and debugging was a pain, so I don't really trust it.

### EventAwareCacheWrapper

This is such a small amount of logic, I don't think it's worth creating an entire wrapper for it, at least not a class based one. I would be happy to refactor this to use a `Proxy` too, so that we don't have to add methods to it each time we wanna change the underlying cache implementation.


SimonBackx:
Merged by accident 😬

PR #19505, opened 2024-01-17
🐛 Fixed searching posts in member filters

SimonBackx:
fixes PROD-201

This also increases the performance when loading the posts:
- Makes sure no relations are loaded
- Only return the fields we actually need
- Stop using limit=all, and replaced it with network based search

### DIFF
--- a/ghost/admin/app/components/gh-resource-select.js
+++ b/ghost/admin/app/components/gh-resource-select.js
@@ -11,6 +11,13 @@ import {tracked} from '@glimmer/tracking';
 
 const DEBOUNCE_MS = 200;
 
+function mapResource(resource) {
+    return {
+        id: resource.id,
+        title: resource.title
+    };
+}
+
 export default class GhResourceSelect extends Component {
     @service store;
 
@@ -130,13 +137,6 @@ export default class GhResourceSelect extends Component {
         const posts = yield this.store.query('post', {filter: 'status:published', limit: '25', fields: 'id,title'});
         const pages = yield this.store.query('page', {filter: 'status:published', limit: '25', fields: 'id,title'});
 
-        function mapResource(resource) {
-            return {
-                id: resource.id,
-                title: resource.title
-            };
-        }
-
         if (posts.length > 0) {
             options.push({
                 groupName: 'Posts',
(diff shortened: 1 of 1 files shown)


ghost/admin/app/components/gh-resource-select.js
```diff
@@ -146,4 +153,38 @@ export default class GhResourceSelect extends Component {
...
+    *fetchOptionsForSearchTask(searchTerm) {
+        // Debounce
+        yield timeout(DEBOUNCE_MS);
+
+        const options = yield [];
+
+        if (this.args.type === 'email') {
+            const posts = yield this.store.query('post', {filter: '(status:published,status:sent)+newsletter_id:-null+title:~\'' + searchTerm.replace('\'', '\\\'') + '\'', limit: '10', fields: 'id,title'});
+            options.push(...posts.map(mapResource));
+            return;
```
sagzy, line 167: Don't we want to `return options` here?

ghost/admin/app/components/gh-resource-select.js
```diff
@@ -146,4 +153,38 @@ export default class GhResourceSelect extends Component {
...
+
+        const posts = yield this.store.query('post', {filter: 'status:published+title:~\'' + searchTerm.replace('\'', '\\\'') + '\'', limit: '10', fields: 'id,title'});
+        const pages = yield this.store.query('page', {filter: 'status:published+title:~\'' + searchTerm.replace('\'', '\\\'') + '\'', limit: '10', fields: 'id,title'});
+
+        function mapResource(resource) {
+            return {
+                id: resource.id,
+                title: resource.title
+            };
+        }
```
sagzy, line 178: May want to extract `mapResource` out of the task, to avoid redefining it every time `fetchOptionsForSearchTask` is invoked? Also used by the `fetchOptionsTask` task above

sagzy (commented):
Tested locally, I cannot reproduce the bug anymore 💪 Added two comments

ghost/admin/app/components/gh-resource-select.js
SimonBackx: Oh my bad! Didn't spot this while testing

ghost/admin/app/components/gh-resource-select.js
SimonBackx: Ah did that, forgot to push that change 🙈

PR #19517, opened 2024-01-18
Added toast notifications for successful email change PROD-215

djordjevlais:
refs PROD-215

- Added toast notifications for successful sender and reply-to email address change behind the flag. The old code which uses a modal is still there, so we should make sure to remove it before the release.


apps/admin-x-settings/src/components/settings/email/newsletters/NewsletterDetailModal.tsx
```diff
@@ -539,28 +532,39 @@ const NewsletterDetailModalContent: React.FC<{newsletter: Newsletter; onlyOne: b
...
-                    title: title,
-                    prompt: prompt,
-                    cancelLabel: '',
-                    onOk: (confirmModal) => {
-                        confirmModal?.remove();
-                        modal.remove();
-                        updateRoute('newsletters');
-                    }
-                });
+                if (newEmailAddressesFlag) {
```
SimonBackx, line 551:
```suggestion
                if (newEmailAddressesFlag || isManagedEmail(config)) {
```

This flag for now won't be enabled on release and is made for self hosters. For Pro, we need to look at the config to see if the email addresses are managed (= using the new DMARC alignment setup) or not.

PR #19518, opened 2024-01-18
Used a key prefix to enable immediate cache resets

allouis:
## Why

The previous `reset()` scanned every matching key and DEL'd them one by one. It was O(n), slow against large caches, and needed cluster-aware routing to hit the right primary node. The slowness added load to the Redis cluster, and it meant that we had to limit the number of sites we could configure to use the cache.

This PR swaps SCAN-and-DEL for prefix rotation. Every key is stored under a random hash that's prepended at write time, so rotating the hash makes every entry unreachable in a single SET. Old entries age out via Redis's own eviction and TTL.

## Changes

- `prefix_hash` is lazily initialized on first read. Concurrent callers within a process share an in-flight promise; `SET NX` collapses the cross-process race so losers read back the winning value.
- `prefix_hash` reads and writes bypass cache-manager and go through the raw redis client. The meta-key holds a raw hex string, not JSON, so cache-manager's wrapper would choke on it; for reset we also want an unconditional overwrite.
- The SCAN-based helpers (`#getKeys`, `#scanNodeForKeys`, `#getPrimaryRedisNode`) and `_removeKeyPrefix` are deleted. They only existed to support the old SCAN-based `reset()`.
- The `cluster-key-slot` dependency is no longer used.

## Details

- `prefix_hash` is written without a TTL, and our Redis cluster uses `volatile-lru` eviction policy, so these keys won't be evicted.
- Every `get` and `set` now costs an extra round-trip to fetch the current `prefix_hash`. For cache hits that doubles the latency. We get O(1) resets in return, and we're accepting the cost for now. We are going to monitor Redis performance as we add new sites, and we might revisit this if it's an issue.

## Testing

The 12 integration tests from #27467 (already in main) continue to pass; they exercise the invalidation contract against real Redis.

### DIFF
--- a/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
+++ b/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
@@ -188,6 +188,10 @@ class AdapterCacheRedis extends BaseCacheAdapter {
             lookup.then((value) => {
                 clearTimeout(timer);
                 resolve(value);
+            }, () => {
+                // redis failure during lookup - treat as MISS, same as the timeout path
+                clearTimeout(timer);
+                resolve({internalKey: null, result: null});
             });
         });
     }
@@ -241,9 +245,9 @@ class AdapterCacheRedis extends BaseCacheAdapter {
      * @param {*} value
      */
     async set(key, value) {
-        const internalKey = await this._buildKey(key);
-        debug('set', internalKey);
         try {
+            const internalKey = await this._buildKey(key);
+            debug('set', internalKey);
             return await this.cache.set(internalKey, value);
         } catch (err) {
             logging.error(err);
--- a/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
+++ b/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
@@ -1,13 +1,16 @@
+const crypto = require('crypto');
 const BaseCacheAdapter = require('@tryghost/adapter-base-cache');
 const errors = require('@tryghost/errors');
 const logging = require('@tryghost/logging');
-const metrics = require('@tryghost/metrics');
 const debug = require('@tryghost/debug')('redis-cache');
 const cacheManager = require('cache-manager');
 const redisStoreFactory = require('./redis-store-factory');
-const calculateSlot = require('cluster-key-slot');
+
+const PREFIX_HASH_KEY = 'prefix_hash';
 
 class AdapterCacheRedis extends BaseCacheAdapter {
+    #prefixHashInitInFlight = null;
+
     /**
      *
      * @param {Object} config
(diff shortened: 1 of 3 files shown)


allouis (commented):
This reads the `prefix_hash` entry from Redis on every read - which may not be desired.

The thinking is that we can do an immediate reset on our side by changing/deleting that key if wanted. 

We can also look at caching it in memory, but that is something we want to get away from...

ghost/adapter-cache-redis/lib/AdapterCacheRedis.js
```diff
@@ -219,28 +241,18 @@ class AdapterCacheRedis extends BaseCacheAdapter {
         }
     }
 
+    async cyclePrefixHash() {
+        const value = crypto.randomBytes(12).toString('hex');
+        await this.cache.set(this._keyPrefix + 'prefix_hash', value, {ttl: 0});
```
allouis, line 246: Setting the TTL to 0 means that the key should not expire in Redis

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
```diff
@@ -119,11 +139,12 @@ class AdapterCacheRedis extends BaseCacheAdapter {
...
      * @param {string} key
-     * @returns {string}
+     * @returns {Promise<string>}
      */
-    _buildKey(key) {
-        if (this.keyPrefix) {
-            return `${this.keyPrefix}${key}`;
+    async _buildKey(key) {
+        const keyPrefix = await this.keyPrefix();
+        if (keyPrefix) {
```
allouis, line 146: This is always truthy now, can remove the check and alternate branch

allouis:
@coderabbitai review

kevinansfield (commented):
`_get` wraps the Redis fetch in a timeout but this introduces an extra fetch before `_get` to read the key from Redis. That key fetch isn't wrapped in the fetch timeout, should it be? I think the outcome is that the cache being down/slow can potentially cause fallback to the real API to take longer than expected

allouis:
@kevinansfield Thanks for looking! You're right, I've added an integration test to catch the key fetching timeouts, and I've updated `get` so that the timeout wraps both the key fetch and the value fetch!

kevinansfield (approved):

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
```diff
@@ -237,7 +241,7 @@ class AdapterCacheRedis extends BaseCacheAdapter {
      * @param {*} value
      */
     async set(key, value) {
-        const internalKey = this._buildKey(key);
+        const internalKey = await this._buildKey(key);
```
vershwal: We can move `_buildKey` inside the try-catch block. `_buildKey` is now async and hits Redis (via `prefixHash()` → `redisClient.get`), so it can throw. Here the `_buildKey` call sits above the try-catch, so a Redis failure would propagate uncaught to the caller.

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
```diff
@@ -164,27 +162,34 @@ class AdapterCacheRedis extends BaseCacheAdapter {
...
-            });
+            return lookup;
         }
+
+        return new Promise((resolve) => {
+            const timer = setTimeout(() => {
+                debug('get', key, 'timeout');
+                resolve({internalKey: null, result: null});
+            }, this.getTimeoutMilliseconds);
+            lookup.then((value) => {
```
vershwal, line 188: We could add a small `.catch()` here as well. Since lookup is async, if Redis fails while fetching the key or value, that rejection would currently go unhandled. The timeout still makes this behave like a cache miss, so it’s not a functional issue, but handling the rejection would avoid unnecessary `unhandledRejection` in the logs

vershwal (commented):
Overall the PR looks good. I’ve added a couple of comments, but nothing major. Let me know what you think.

daniellockyer:
😍

PR #19536, opened 2024-01-22
Added loader for preview

vershwal:
[PROD-4](https://linear.app/tryghost/issue/PROD-4/editing-offers-is-slow-and-flashy)


apps/admin-x-settings/src/components/settings/membership/portal/PortalFrame.tsx
```diff
@@ -62,7 +63,10 @@ const PortalFrame: React.FC<PortalFrameProps> = ({href, onDestroyed, selectedTab
                 src={href}
                 title="Portal Preview"
                 width="100%"
-                onLoad={() => setHasLoaded(true)}
+                onLoad={() => {
+                    setHasLoaded(true);
+                    makeVisible();
```
ronaldlangeveld, line 68: I would prefer if we not execute `makeVisible();` here as it could make the iframe visible before Portal has loaded, which would also result in flashing (ie the homepage loads before Portal loads and both would be visible).

PR #19542, opened 2024-01-22
Fixed reply-to address to stay the same after dmarc changes

sagzy:
fixes PROD-102
- after the dmarc changes, replies from members should keep going to the same custom reply-to email address

### DIFF
--- a/ghost/email-service/lib/EmailRenderer.js
+++ b/ghost/email-service/lib/EmailRenderer.js
@@ -215,11 +215,7 @@ class EmailRenderer {
         if (newsletter.get('sender_reply_to') === 'support') {
             return this.#settingsHelpers.getMembersSupportAddress();
         }
-        if (newsletter.get('sender_reply_to') === 'newsletter') {
-            if (this.#emailAddressService.managedEmailEnabled) {
-                // Don't duplicate the same replyTo addres if it already in the FROM address
-                return null;
-            }
+        if (newsletter.get('sender_reply_to') === 'newsletter' && !this.#emailAddressService.managedEmailEnabled) {
             return this.getFromAddress(post, newsletter);
         }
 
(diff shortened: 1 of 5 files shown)


ghost/email-service/lib/EmailRenderer.js
```diff
@@ -215,11 +215,7 @@ class EmailRenderer {
         if (newsletter.get('sender_reply_to') === 'support') {
             return this.#settingsHelpers.getMembersSupportAddress();
         }
-        if (newsletter.get('sender_reply_to') === 'newsletter') {
-            if (this.#emailAddressService.managedEmailEnabled) {
-                // Don't duplicate the same replyTo addres if it already in the FROM address
-                return null;
-            }
+        if (newsletter.get('sender_reply_to') === 'newsletter' && !this.#emailAddressService.managedEmailEnabled) {
```
SimonBackx, line 218: This works, but the `this.#emailAddressService.getAddress` will log an error if the replyTo address is invalid. So if it is set to `newsletter` and we continue after this if check to the `this.#emailAddressService.getAddress` part, it will log an error every time. Ideally we should set replyTo to undefined when that is the case and avoid that error log

SimonBackx: Not blocking a merge though

SimonBackx:
We'll also need to set `sender_email` to `null` when changing `sender_reply_to` when `sender_email` is not supported for sending, otherwise we won't be able to clear the replyTo address.

Ideally that check should be auto-correcting every time, so I would try to avoid using feature flag checks / config checks, and when changing `sender_reply_to` or `sender_email` actually check if EmailAddressService.getAddress({from: sender_email}) stays the same as sender_email, otherwise set it to null (because that means we cannot send from it).

ghost/email-service/lib/EmailRenderer.js
sagzy: Thanks! Addressed in a498ceba15274a08be5fc57b798d722510398e97

PR #19589, opened 2024-01-26
Added basic instrumentation to the database connection module

cmraible:
no issue

- To help debug potential causes of slow/aborted get helpers, it would be cool to get more visibility into how Ghost handles database connections, particularly if it has to spend a long time waiting to acquire a new connection from the pool.
- Under the hood, knex uses a package called tarn (https://github.com/Vincit/tarn.js/tree/3.0.2) to manage the connection pool. Tarn provides some hooks for instrumentation, so we can use those to get some basic visibility into the connection pool.
- This PR adds handling for creating, acquiring and releasing connections from Tarn's connection pool which logs some basic metrics, particularly the queue length and time it takes to acquire a connection.


cmraible:
Having a hard time figuring out why this isn't working. 

When Ghost first boots, both the messages are logged as I would expect: 
CleanShot 2024-01-25 at 17 13 01

However, I _know_ that Ghost is acquiring new connections (you can see this with `DEBUG=knex:* yarn dev`), but for whatever reason the logs aren't showing for those. It seems like for some reason the event handlers are only running once, which doesn't seem right, but I haven't been able to figure out why this is the case.

cmraible:
Lots of cleanup to do once I get this working — these logs should be behind a config, or maybe drop the log level to debug. And ideally we'd have the duration setup as a metric instead of just a plain text log. 

Focus right now is just to get it working, then I can add those parts in.

ghost/core/core/server/data/db/ConnectionPoolInstrumentation.js
```diff
@@ -0,0 +1,135 @@
...
+            numUsed: this.pool.numUsed()
+        };
+    }
+
+    handleCreateRequest(eventId) {
+        // Track when the request was submitted for calculating wait time
+        this.createStartTimes[eventId] = Date.now();
+        const poolMetrics = this.getPoolMetrics();
+        this.logging.debug(`[ConnectionPool] Creating a connection. EventID: ${eventId} Pending Creates: ${poolMetrics.numPendingCreates} Free: ${poolMetrics.numFree} Used: ${poolMetrics.numUsed}`);
+        this.metrics.metric('connection-pool-create-request', {
```
cmraible, line 27: Would it be better to put all of these under a single 'index' (e.g. `metrics-connection-pool` and then add `event: createRequest`)?

I wasn't sure if metrics have to have the same shape within a given 'index' or if they can have different properties.

daniellockyer: I believe it would be better - one index is easier to deal with in ES than multiple 🙂 AFAIK they don't have to have the same shape

cmraible: Cool cool, updated them all to `metrics-connection-pool`

PR #19614, opened 2024-01-29
Added config option to disable db backups

9larsons:
refs https://linear.app/tryghost/issue/ENG-600
- users need an option so they can perform actions like delete users without blowing up Ghost as large dbs can OOM node

### DIFF
--- a/ghost/core/core/server/data/db/backup.js
+++ b/ghost/core/core/server/data/db/backup.js
@@ -45,6 +45,12 @@ const readBackup = async (filename) => {
  * @returns {Promise<String>}
  */
 const backup = async function backup(options = {}) {
+    // do not create backup if disabled in config (this is intended for large customers who will OOM node)
+    if (config.get('disableJSBackups')) {
+        logging.info('Database backup is disabled in Ghost config');
+        return '';
+    }
+
     logging.info('Creating database backup');
 
     const filename = await exporter.fileName(options);
--- a/ghost/core/core/server/data/db/backup.js
+++ b/ghost/core/core/server/data/db/backup.js
@@ -42,13 +42,13 @@ const readBackup = async (filename) => {
  * Does an export, and stores this in a local file
  *
  * @param {Object} options
- * @returns {Promise<String>}
+ * @returns {Promise<String> | null}
  */
 const backup = async function backup(options = {}) {
     // do not create backup if disabled in config (this is intended for large customers who will OOM node)
     if (config.get('disableJSBackups')) {
         logging.info('Database backup is disabled in Ghost config');
-        return '';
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/data/db/backup.js
```diff
@@ -45,6 +45,12 @@ const readBackup = async (filename) => {
  * @returns {Promise<String>}
  */
 const backup = async function backup(options = {}) {
+    // do not create backup if disabled in config (this is intended for large customers who will OOM node)
+    if (config.get('disableJSBackups')) {
+        logging.info('Database backup is disabled in Ghost config');
+        return '';
```
allouis, line 51: Do we know for sure that consumers of this module can handle an empty string return value?

allouis (commented):
Should work - but I'd like to be sure that everywhere that calls `backup` is able to handle a `string` return - it might be better to return `null` instead, and explicitly handle it at the call site. Or if the return value isn't used anywhere, start returning a boolean representing whether a backup was made or not

ghost/core/core/server/data/db/backup.js
9larsons: Afaict the consumers don't even use the string, but it's clearly structured to return a file path in case that needs to be served to the front end. I made the return a null string because it shouldn't reject - that would likely interrupt the upstream code - and generally the return type is `Promise<String>`. 

These are the callers and the use cases I can make out:
- [Deleting a user](https://github.com/TryGhost/Ghost/blob/3c56005d44b0cbb3e6a9e72ff3cf38c40569f9c7/ghost/core/core/server/services/Users.js#L151); filename not used
- [Migration before hook](https://github.com/TryGhost/Ghost/blob/3c56005d44b0cbb3e6a9e72ff3cf38c40569f9c7/ghost/core/core/server/data/migrations/hooks/migrate/before.js#L6); filename not used but **I'm also not sure how this doesn't break migrations unless it doesn't run for large customers, because it should always fail for them?**
- [Delete all content/backup content routes](https://github.com/TryGhost/Ghost/blob/3c56005d44b0cbb3e6a9e72ff3cf38c40569f9c7/ghost/core/core/server/web/api/endpoints/admin/routes.js#L220-L224); these are only manually accessed with Ghost API requests

I don't see any issue with these use cases. I am curious about the migrations one though, which maybe you could shed some light on?

daniellockyer: @9larsons Yeah, that line is disabled on Pro 🙂

9larsons: Ah right!

allouis (approved):

PR #19627, opened 2024-01-31
Optimised reading from cache

allouis:
🍦 

This ensures that concurrent reads from the cache do not make concurrent reads from the underlying data store.

### DIFF
--- a/ghost/adapter-cache-redis/lib/AdapterCacheRedis.js
+++ b/ghost/adapter-cache-redis/lib/AdapterCacheRedis.js
@@ -82,6 +82,7 @@ class AdapterCacheRedis extends BaseCacheAdapter {
         this.refreshAheadFactor = config.refreshAheadFactor || 0;
         this.getTimeoutMilliseconds = config.getTimeoutMilliseconds || null;
         this.currentlyExecutingBackgroundRefreshes = new Set();
+        this.currentlyExecutingReads = new Map();
         this.keyPrefix = config.keyPrefix;
         this._keysPattern = config.keyPrefix ? `${config.keyPrefix}*` : '';
         this.redisClient = this.cache.store.getClient();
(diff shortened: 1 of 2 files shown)


ghost/adapter-cache-redis/lib/AdapterCacheRedis.js
```diff
@@ -6,6 +6,25 @@ const cacheManager = require('cache-manager');
...
+        }
+        if (err) {
+            _reject(err);
+        }
+        resolve = _value => _resolve(_value);
+        reject = _err => _reject(_err);
+    });
+
+    return {resolve, reject, promise};
+}
```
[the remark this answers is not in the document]
SimonBackx: Agree with Mike here. The actual inline implementation is shorter than the boilerplate version that uses the deferred method.

ghost/adapter-cache-redis/test/adapter-cache-redis.test.js
```diff
@@ -98,6 +99,40 @@ describe('Adapter Cache Redis', function () {
...
+                    getClient: sinon.stub().returns({
+                        on: sinon.stub()
+                    })
+                }
+            };
+            const cache = new RedisCache({
+                cache: redisCacheInstanceStub
+            });
+
+            const fetchData = sinon.stub().resolves('Da Value');
```
SimonBackx, line 124: I'm not sure how this test works reliably. Shouldn't we at least wait after both get calls before resoling the fetchData method?

ghost/adapter-cache-redis/test/adapter-cache-redis.test.js
```diff
@@ -98,6 +99,40 @@ describe('Adapter Cache Redis', function () {
             }
         });
 
+        it('can wait for execution in the case of duplicate reads', async function () {
```
SimonBackx, line 102: Preferable we should also add a test with a fetchData error

ghost/adapter-cache-redis/lib/AdapterCacheRedis.js
```diff
@@ -166,6 +186,26 @@ class AdapterCacheRedis extends BaseCacheAdapter {
      * @param {() => Promise<any>} [fetchData] An optional function to fetch the data, which will be used in the case of a cache MISS or a background refresh
      */
     async get(key, fetchData) {
+        const internalKey = this._buildKey(key);
+        if (this.currentlyExecutingReads.has(internalKey)) {
```
daniellockyer, line 190: I think there's possibly a change of a race condition here? Where `.has` passes, but the Promise is deleted before we get to the `.get`? Maybe?

Would this work better?

```
const currentlyExecutingPromise = this.currentlyExecutingReads.get(internalKey);
if (currentlyExecutingPromise) {
	return currentlyExecutingPromise.promise;
}
```

ghost/adapter-cache-redis/test/adapter-cache-redis.test.js
allouis: We don't need to do that because we're not `await`ing the calls to `get` - this test consistently fails without the changes and passes with them

ghost/adapter-cache-redis/lib/AdapterCacheRedis.js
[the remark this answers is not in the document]
allouis: The first way I implemented this (non split `get`/`#get`) meant that we needed to resolve/reject throughout the `get` method body and the deferred pattern worked well for that - will remove

ghost/adapter-cache-redis/test/adapter-cache-redis.test.js
SimonBackx: Okay, I'm still a bit skeptical still everything is still very synchronous and the executing already starts before awaiting the get calls, but we can just fix if it would ever start failing

PR #19649, opened 2024-02-05
CI/CD: Upgrade Node.js 18.x to 20.x

shivarm:
Upgrade Node.js 18.x to 20.x and deprecated github actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,7 @@ jobs:
           echo "$EOF" >> "$GITHUB_ENV"
 
       - name: Nx cache
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         id: cache_nx
         with:
           path: .nxcache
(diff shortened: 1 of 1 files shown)


shivam-sharma7:
@cmraible

markstos:
Why this is being done now? Only Node 18 is still officially supported: 

https://ghost.org/docs/faq/node-versions/

I see no consideration here updated updating the documentation to say that Node 20 the new official version or any notes about the process of how the need to upgrade would be communicated to all users. 

The Node 18 LTS will still receive security updates for another year.

shivam-sharma7:
let's keep as draft PR but I would say we should consider latest LTS version of Node.js

cmraible:
Hi @shivam-sharma7 , thanks for raising this! As @markstos mentioned, there is definitely more to this than just updating the tests in CI — we'll need to do some more thorough testing, update docs, etc. but thanks for bringing this up. Discussing with the team now, but I'll post an update here on v20.x when I have it 👍

PR #19658, opened 2024-02-06
🐛 Fixed members import unsubscribing members when `subscribe_to_emails` is blank

cmraible:
fixes ENG-611

- Previously, if an existing member with newsletter subscriptions was imported, and `subscribe_to_emails` was blank/empty, the member would be unsubscribed from all newsletters, which is not the expected behavior.
- This PR changes the behavior so if `subscribe_to_emails` is blank, it will not unsubscribe existing members.

### DIFF
--- a/ghost/members-csv/lib/unparse.js
+++ b/ghost/members-csv/lib/unparse.js
@@ -14,53 +14,53 @@ const DEFAULT_COLUMNS = [
     'tiers'
 ];
 
-const unparse = (members, columns = DEFAULT_COLUMNS.slice()) => {
+const unparse = (rows, columns = DEFAULT_COLUMNS.slice()) => {
     columns = columns.map((column) => {
         if (column === 'subscribed') {
             return 'subscribed_to_emails';
         }
         return column;
     });
-    const mappedMembers = members.map((member) => {
-        if (member.error && !columns.includes('error')) {
+    const mappedRows = rows.map((row) => {
+        if (row.error && !columns.includes('error')) {
             columns.push('error');
         }
 
         let labels = '';
-        if (typeof member.labels === 'string') {
-            labels = member.labels;
-        } else if (Array.isArray(member.labels)) {
-            labels = member.labels.map((l) => {
+        if (typeof row.labels === 'string') {
+            labels = row.labels;
+        } else if (Array.isArray(row.labels)) {
+            labels = row.labels.map((l) => {
                 return typeof l === 'string' ? l : l.name;
             }).join(',');
         }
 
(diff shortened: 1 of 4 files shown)


ghost/members-csv/lib/unparse.js
```diff
@@ -14,53 +14,53 @@ const DEFAULT_COLUMNS = [
     'tiers'
 ];
 
-const unparse = (members, columns = DEFAULT_COLUMNS.slice()) => {
+const unparse = (rows, columns = DEFAULT_COLUMNS.slice()) => {
```
cmraible, line 17: Nitpick but at this point these are just rows in a CSV, not members

ghost/members-csv/lib/unparse.js
```diff
@@ -14,53 +14,53 @@ const DEFAULT_COLUMNS = [
...
-            subscribed_to_emails: member.subscribed || member.subscribed_to_emails ? true : false,
-            complimentary_plan: member.comped || member.complimentary_plan,
-            stripe_customer_id: _.get(member, 'subscriptions[0].customer.id') || member.stripe_customer_id,
-            created_at: member.created_at,
-            deleted_at: member.deleted_at,
+            id: row.id,
+            email: row.email,
+            name: row.name,
+            note: row.note,
+            subscribed_to_emails: 'subscribed' in row ? row.subscribed : row.subscribed_to_emails,
```
cmraible, line 51: I'm 99% sure there's no way currently to get a value in row.subscribed, BUT on the off chance I'm missing something here, I changed this up to more explicitly check if the key exists in the row (even if the value is undefined) vs using the `||` operator.

This allows the expected value (`row.subscribed_to_emails`) to be undefined instead of coercing it to a boolean.

ghost/members-importer/test/MembersCSVImporter.test.js
```diff
@@ -221,6 +221,222 @@ describe('MembersCSVImporter', function () {
             }]);
             should.deepEqual(membersRepositoryStub.update.args[0][1].id, 'test_member_id');
         });
+
+        it('should subscribe or unsubscribe members as per the `subscribe_to_emails` column', async function () {
```
cmraible, line 225: A bit verbose and cumbersome, but this one has come up a couple times so I want to just squash it dead.

9larsons (approved):
LGTM, thanks for putting in some notes. I think my only comments here are limited to the confusing duplicative/overlapping columns and data model which is something we should revisit down the road.

cmraible:
Hard agree!

PR #19666, opened 2024-02-07
Added `pro.js` script to unauthenticated pages in admin

cmraible:
refs PA-29


cmraible:
Upon further inspection, I'm not actually sure this is necessary. Will reopen if testing in staging fails.

PR #19675, opened 2024-02-07
Fixed PostHog toolbar not working in admin

cmraible:
refs PA-32

- The PostHog toolbar relies on a value that is passed via a hash in the URL to launch successfully
- Admin overwrites the hash (since it uses hash based routing) before the toolbar has a chance to read the value
- This change checks for the hash and if it exists, it launches the toolbar using the hash value

### DIFF
--- a/ghost/admin/app/routes/home.js
+++ b/ghost/admin/app/routes/home.js
@@ -9,6 +9,11 @@ export default class HomeRoute extends Route {
     beforeModel(transition) {
         super.beforeModel(...arguments);
 
+        const toolbarJSON = new URLSearchParams(window.location.hash.substring(1)).get('__posthog');
+        if (toolbarJSON) {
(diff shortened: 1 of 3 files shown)


ghost/admin/app/routes/home.js
```diff
@@ -9,6 +9,11 @@ export default class HomeRoute extends Route {
     beforeModel(transition) {
         super.beforeModel(...arguments);
 
+        const toolbarJSON = new URLSearchParams(window.location.hash.substring(1)).get('__posthog');
+        if (toolbarJSON) {
+            window.posthog?.loadToolbar(JSON.parse(toolbarJSON));
```
cmraible, line 14: Are we okay doing calls to `window.posthog` (which is loaded in `pro.js`) in Admin?

PR #19678, opened 2024-02-08
Added re-enable of innodb redo log once DataImporter has run

kevinansfield:
no issue

- DataGenerator disables the redo log to make data imports faster but it's a persisted global config change and we were missing the re-enable query once the imports have finished
- when the redo log remains disabled an unexpected shutdown puts the database into a non-starting state with the error `Server was killed when Innodb Redo logging was disabled. Data files could be corrupt. You can try to restart the database with innodb_force_recovery=6`

### DIFF
--- a/ghost/data-generator/lib/DataGenerator.js
+++ b/ghost/data-generator/lib/DataGenerator.js
@@ -270,6 +270,13 @@ class DataGenerator {
             });
             await tableImporter.finalise();
         }
+
+        // Re-enable the redo log because it's a persisted global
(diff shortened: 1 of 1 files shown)


kevinansfield:
@SimonBackx I did look at a way of having the query run no matter how the generator script exits (e.g. Ctrl-C) but I didn't find a way of doing that with node that isn't hairy. Not sure that's really needed though? Just resetting the config after a successful run makes things safer than it was.

SimonBackx (approved):

PR #19694, opened 2024-02-14
Add the @id for json-ld

Wardormeur:
Hello,
This introduces the json-ld `@id`. The goal is to allow for extension of the schema through a different script but on the same model.
To do so, we need to be able to merge them by adding the same `@id` into both scripts as a point of reference.
This is useful if you want to add other fields, through the "ghost_header", like so
Example usage using the ghost_header, assuming the Article had the `@id` field populated :
```
<script type="application/ld+json">
    {
      "@context": "https://schema.org",
      "@id": "myarticle.com/42",
      "@type": "Article",
      "isAccessibleForFree": "False",
      "hasPart":
        {
        "@type": "WebPageElement",
        "isAccessibleForFree": "False",
        "cssSelector" : ".paywall"
        }
     }
</script>
```

I've validated this method against a modified sample on https://validator.schema.org/

- [] The build will pass (run `yarn test:all` and `yarn lint`)


ErisDS:
Hey @Wardormeur I'm sorry we didn't get around to looking at this in so long. 

I think this change is directionally correct, however if we're going to make it we'd want to do it holistically and with tests.

A good resource: https://momenticmarketing.com/blog/id-schema-for-seo-llms-knowledge-graphs#page-level-vs-site-level-scope-of-id

I'm going to close this, but if anyone is interested in contributing full @id support with tests we'd be interested to work on it together.

PR #19701, opened 2024-02-15
Updated browser tests to be less flaky

9larsons:
no refs

- Offers browser tests were subject to a race condition. I'm guessing this dates back to when we moved to Settings X (and React), as it seems the url for the offer is not present on the first render of the page - despite being returned in the `POST` request of the offer creation, the component does a `GET` on render to get the link. This is now awaited.
- The Publishing timezone test also seemed to suffer from a race condition. This is less sure of a fix as it's a much less frequent failure. The date time picker input is now validated in the test before continuing.
- Offers browser tests often timed out so the timeout has been moved to 90s for these tests.
- All tests were bumped to 75s timeout as we generally would occasionally hit the timeout.


ghost/core/test/e2e-browser/portal/offers.spec.js
```diff
@@ -33,9 +34,20 @@ test.describe('Portal', () => {
...
-            await sharedPage.waitForSelector('[data-testid="portal-popup-frame"]', {state: 'attached'});
+            try {
+                await expect(sharedPage.locator('[data-testid="portal-popup-frame"]')).toBeAttached({timeout: 1000});
+            } catch (e) {
+                await sharedPage.goto(offerLink);
+
+                // Wait for the load state to ensure the page has loaded completely
+                portalTriggerButton = sharedPage.frameLocator('[data-testid="portal-trigger-frame"]').locator('[data-testid="portal-trigger-button"]');
+                await expect(portalTriggerButton).toBeVisible();
+                await expect(sharedPage.locator('[data-testid="portal-popup-frame"]')).toBeAttached({timeout: 1000});
```
jdicke, line 49: 1000 should be extracted into a variable since you use it a couple times

ghost/core/test/e2e-browser/utils/e2e-browser-utils.js
```diff
@@ -250,64 +250,68 @@ const createTier = async (page, {name, monthlyPrice, yearlyPrice, trialDays}, en
...
-        // Selector for the elements with data-testid 'offer-item'
-        // const offerItemsSelector = '[data-testid="offer-item"]';
-        await page.getByTestId('offer-item').nth(0).click();
-        await page.getByRole('button', {name: 'Archive offer'}).click();
-
-        const confirmModal = await page.getByTestId('confirmation-modal');
-        await confirmModal.getByRole('button', {name: 'Archive'}).click();
-    }
+        // Keep offer names unique & <= 40 characters
+        offerName = `${name} (${new ObjectID().toHexString().slice(0, 40 - name.length - 3)})`;
```
jdicke, line 260: there is a potential issue with the code: it assumes that name.length is less than or equal to 37. If the length of name exceeds this limit, the generated string might be longer than 40 characters.

To make the code more robust, you can add a check to ensure that the resulting string does not exceed the desired length.

`const maxOfferNameLength = 40;

let offerName = `${name} (${new ObjectID().toHexString().slice(0, maxOfferNameLength - name.length - 3)})`;

// Ensure the offerName does not exceed the maximum length
if (offerName.length > maxOfferNameLength) {
    offerName = offerName.slice(0, maxOfferNameLength);
}`

9larsons:
It seems to be an issue with the Settings X migration and the speed of the browser runner/machine. We were missing the Offers URL in some cases, which appears to be due to a race condition with the Offers Success modal. In moving to React, this component will re-render and trigger a `GET` request that seems to (often) happen after the link is pulled from the input field that's supposed to have the link href... except it's waiting for the response to actually populate that, rather than pulling it from the `POST` request that creates the offer.

It feels odd to do a `POST` and immediately `GET` for which the response has almost identical data. It's adding extra network requests but that's a separate concern. Waiting for the field to be populated appears to have cured the flakiness of the tests.

9larsons:
We still have a flaky test with time mocking, which I'll take a look at fixing in this as well.

PR #19716, opened 2024-02-20
Added Posthog event sending to trackEvent function

55sketch:
refs PA-37

As we add Posthog to the stack we want to send the existing events we track to it, as well as opening up the method to teams to use in their initiatives.

### DIFF
--- a/ghost/admin/app/utils/analytics.js
+++ b/ghost/admin/app/utils/analytics.js
@@ -26,9 +26,9 @@ async function hashEmail(email) {
 }
 
 /**
- * Sends a tracking event to Plausible, if installed.
+ * Sends a tracking event to Plausible and Posthog, if installed.
  * 
- * By default, Plausible is not installed, in which case this function no-ops.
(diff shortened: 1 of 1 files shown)


ghost/admin/app/utils/analytics.js
```diff
@@ -34,6 +34,10 @@ export function trackEvent(eventName, props = {}) {
         (window.plausible.q = window.plausible.q || []).push(arguments);
     };
     window.plausible(eventName, {props: props});
+
+    if (window.posthog) {
```
cmraible, line 38: One thing I've just learned this week — apparently `window.posthog` [can _exist_ but not be fully loaded](https://linear.app/tryghost/issue/PA-36/windowposthogget-distinct-id-fails-even-when-windowposthog-exists). As a quick fix for the time being, I've started using `window.posthog.__loaded` to confirm it's fully loaded before calling any functions on it — we should probably do that here too!

PR #19723, opened 2024-02-21
🐛 Fixed Portal icon misaligned in Preview

ronaldlangeveld:
no issue

- because we use hidden here, it prevents the Portal iFrame from calculating the width and position correctly on the initial load.
- Changing it to `invisible` allows it to do the calculation correctly.
- Also had to reposition the loading indicator.


vershwal:
LGTM
Using 'invisible' is a good idea. Although the element is not visible to the user, it still maintain the layout of the page.

vershwal (approved):

PR #19728, opened 2024-02-21
🐛 Fixed explicit HTML entities being decoded when rendering HTML cards

kevinansfield:
closes ENG-608

- bumps Koenig rendering packages to include fix for HTML entities in HTML card content being decoded during rendering which could result in unexpected/broken output


pezi:
Almost fixed - you must update the post which is affected from the problem to get the fix working.

PR #19761, opened 2024-02-27
Added output to the get helper when the timeout is exceeded

allouis:
refs https://linear.app/tryghost/issue/ENG-670


cmraible:
Nice one, works as expected for me locally 👍

Not sure if this would be useful or not, so feel free to disregard: do you think it would make sense to decorate the span with the API options? e.g. resource, filters, limits, etc. I guess we could deduce it from the theme pretty easily, but might save us a step 🤔

cmraible (approved):

allouis:
I did initially try that, but ended up generating a bunch of invalid html, depending on the inputs to the get helper - I think for now it's best we leave this simple, and can look into improvements down the line

PR #19762, opened 2024-02-27
Ensured comment counts route doesn't load member 

allouis:
refs https://linear.app/tryghost/issue/ENG-672/

### DIFF
--- a/ghost/core/core/server/web/comments/routes.js
+++ b/ghost/core/core/server/web/comments/routes.js
@@ -11,8 +11,6 @@ module.exports = function apiRoutes() {
     const router = express.Router('comment api');
     router.use(bodyParser.json({limit: '50mb'}));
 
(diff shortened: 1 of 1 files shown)


ghost/core/core/server/web/comments/routes.js
```diff
@@ -22,14 +21,17 @@ module.exports = function apiRoutes() {
 
     router.get('/', http(api.commentsMembers.browse));
     router.get('/:id', http(api.commentsMembers.read));
```
kevinansfield, line 23: I think these use member authentication to determine if a comment has been liked?

allouis: Ugh - I think you're right, I didn't see it because it's transparently passed all the way into the model layer 😢

allouis: I've updated this to only exclude comment counts

PR #19767, opened 2024-02-28
Added support for cache invalidation headers for comments

allouis:
refs https://linear.app/tryghost/issue/ENG-676/support-cache-invalidation-headers-for-comments

This is based off of the work here https://github.com/TryGhost/Ghost/pull/19764 which gives us the ability to dynamically set headers

### DIFF
--- a/apps/comments-ui/test/utils/MockedApi.ts
+++ b/apps/comments-ui/test/utils/MockedApi.ts
@@ -166,29 +166,28 @@ export class MockedApi {
         });
 
         await page.route(`${path}/members/api/comments/*`, async (route) => {
-            if (route.request().method() === 'POST') {
-                const payload = JSON.parse(route.request().postData());
+            const payload = JSON.parse(route.request().postData());
 
-                this.#lastCommentDate = new Date();
-                this.addComment({
-                    ...payload.comments[0],
-                    member: this.member
-                });
-                return await route.fulfill({
-                    status: 200,
-                    body: JSON.stringify({
-                        comments: [
-                            this.comments[this.comments.length - 1]
-                        ]
-                    })
-                });
-            }
+            this.#lastCommentDate = new Date();
+            this.addComment({
+                ...payload.comments[0],
+                member: this.member
+            });
+            return await route.fulfill({
+                status: 200,
+                body: JSON.stringify({
+                    comments: [
+                        this.comments[this.comments.length - 1]
+                    ]
+                })
+            });
+        });
 
+        await page.route(`${path}/members/api/comments/post/*/*`, async (route) => {
             const url = new URL(route.request().url());
 
             const p = parseInt(url.searchParams.get('page') ?? '1');
             const limit = parseInt(url.searchParams.get('limit') ?? '5');
-            const order = url.searchParams.get('order') ?? '';
             const filter = url.searchParams.get('filter') ?? '';
 
             await route.fulfill({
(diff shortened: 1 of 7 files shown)


ghost/core/core/server/services/comments/CommentsController.js
```diff
@@ -35,6 +35,22 @@ module.exports = class CommentsController {
...
+                frame.options.mongoTransformer = function (query) {
+                    return {
+                        $and: [
+                            {
+                                post_id: frame.options.post_id
+                            },
+                            query
+                        ]
+                    };
+                };
```
kevinansfield, line 49: Is it worth adding a TODO comment here to clean this up when the old route is removed? Also wondering if we should do anything if the supplied filter already has a `post_id` in it 🤔

apps/comments-ui/test/utils/MockedApi.ts
```diff
@@ -166,37 +166,35 @@ export class MockedApi {
...
+                })
+            });
+        });
 
+        await page.route(`${path}/members/api/comments/post/*/*`, async (route) => {
             const url = new URL(route.request().url());
 
             const p = parseInt(url.searchParams.get('page') ?? '1');
             const limit = parseInt(url.searchParams.get('limit') ?? '5');
-            const order = url.searchParams.get('order') ?? '';
```
kevinansfield, line 191: Why was `order` removed here? I'm not saying it's wrong, just curious 😄

kevinansfield (commented):
All looks good 👍 Added a couple of v minor comments.

Would be good to add a test that hits the new endpoint.

apps/comments-ui/test/utils/MockedApi.ts
allouis: It's just not used at all by the `browseComments` method so I was getting red squiggles!

ghost/core/core/server/services/comments/CommentsController.js
kevinansfield: Reading again I'm not sure this comment makes sense as I can't see how this would really change when the old route is removed 😄

Still wonder if it's worth adding some protection against double `post_id=x AND post_id=y` in the filter but that can be a cleanup thing.

apps/comments-ui/test/utils/MockedApi.ts
allouis: I probably could/should have updated the MockedApi to use it, but was trying to keep changes minimal, I'm happy to update it though - whaddaya think?

kevinansfield: Ah, OK I see now. Fine to leave like this, especially as we'll be getting rid of the `order` param usage in the live code anyway

ghost/core/core/server/services/comments/CommentsController.js
allouis: I think we'll still need this even when the old route is removed! This is the mechanism that we limit the browse to the requested post

Yeah so, what do you think we should do? We could either error, or strip out any post_id filter operations.

I'm deffo in favour of erroring when invalid filters are used, BUT it goes against Postels law of liberal with what we accept...


I can fairly easily strip out any post_id filter operations if we think that's valuable

allouis: As discussed in slack - we're gonna leave this for now to keep changeset small

PR #19776, opened 2024-02-28
Added /comments/:id/replies/ to X-Cache-Invalidate

allouis:
refs https://linear.app/tryghost/issue/ENG-682/

This should allow us to bust both endpoints cache when write operations are made to comments.

### DIFF
--- a/ghost/core/test/e2e-api/members-comments/comments.test.js
+++ b/ghost/core/test/e2e-api/members-comments/comments.test.js
@@ -382,7 +382,9 @@ describe('Comments API', function () {
                     .matchHeaderSnapshot({
                         etag: anyEtag,
                         location: anyLocationFor('comments'),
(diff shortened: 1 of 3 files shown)


ghost/core/test/e2e-api/members-comments/comments.test.js
```diff
@@ -381,7 +381,8 @@ describe('Comments API', function () {
                     .expectStatus(201)
                     .matchHeaderSnapshot({
                         etag: anyEtag,
-                        location: anyLocationFor('comments')
+                        location: anyLocationFor('comments'),
+                        'x-cache-invalidate': matchers.anyString
```
allouis, line 385: This is annoying, and could potentially be improved.

We are using a dynamically generated comment, rather than a fixture, which means the value of teh header changes for each test run

PR #19791, opened 2024-03-04
Portal signup page spacing fixes

peterzimon:
refs. https://linear.app/tryghost/issue/DES-161/portal-signup-window-spacing-problem

The products section is returned even if there's only a free product in the publication. This results in an extra, unnecessary space between the signup form inputs and the signup button.

### DIFF
--- a/apps/portal/src/components/common/ProductsSection.js
+++ b/apps/portal/src/components/common/ProductsSection.js
@@ -978,6 +978,11 @@ function ProductsSection({onPlanSelect, products, type = null, handleChooseSignu
         className += ' gh-portal-upgrade-product';
     }
 
+    // If site doesn't have paid products then don't return an empty container
+    if (hasOnlyFreeProduct({site})) {
+        return null;
+    }
+
     let finalProduct = products.find(p => p.id === selectedProduct)?.id || products.find(p => p.type === 'paid')?.id;
     return (
         <ProductsContext.Provider value={{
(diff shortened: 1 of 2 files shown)


apps/portal/src/components/common/ProductsSection.js
```diff
@@ -978,6 +978,11 @@ function ProductsSection({onPlanSelect, products, type = null, handleChooseSignu
         className += ' gh-portal-upgrade-product';
     }
 
+    // If site doesn't have paid products then don't return an empty container
+    if (hasOnlyFreeProduct({site})) {
```
ronaldlangeveld, line 982: https://github.com/TryGhost/Ghost/blob/8e0afbd18ebc204e3968c1da44a31037a9fe9e5d/apps/portal/src/components/common/ProductsSection.js#L959-L961

We are already checking whether we have paid products or not (although with a different condition).
Should the above condition be `true`, this line will never be reached.

vershwal: Also you can always use [hasOnlyFree](https://github.com/TryGhost/Ghost/blob/8e0afbd18ebc204e3968c1da44a31037a9fe9e5d/apps/portal/src/components/common/ProductsSection.js#L667) in place of `hasOnlyFreeProduct({site})` if you still want to use it.

peterzimon: I think this is a different one as it checks which cadences are checked, while the other checks which products are checked in Portal settings. However my change could be combined with this condition like this:

```
if ((!portalPlans.includes('monthly') && !portalPlans.includes('yearly')) || hasOnlyFreeProduct({site})) {
        return null;
    }
    ```
    
 Wdyt?

PR #19799, opened 2024-03-05
Added lazy-loading of admin auth in Comments UI

kevinansfield:
closes ENG-711

When an Admin is authenticated in Comments-UI we only add moderation options to the displayed comments so we don't need to pre-emptively load the `admin-auth` iframe and make the `/ghost/api/admin/users/me/` request until some comments are actually visible.

- used `state.comments.length` property to defer rendering of the admin auth frame until comments have been fetched (after box is scrolled into view) and the count is > 0

### DIFF
--- a/apps/comments-ui/src/App.tsx
+++ b/apps/comments-ui/src/App.tsx
@@ -192,7 +192,7 @@ const App: React.FC<AppProps> = ({scriptTag}) => {
             <CommentsFrame ref={iframeRef}>
                 <ContentBox done={done} />
             </CommentsFrame>
-            {done ? <AuthFrame adminUrl={options.adminUrl} onLoad={initAdminAuth}/> : null}
+            {state.comments.length > 0 ? <AuthFrame adminUrl={options.adminUrl} onLoad={initAdminAuth}/> : null}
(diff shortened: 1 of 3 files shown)


apps/comments-ui/src/App.tsx
```diff
@@ -192,7 +192,7 @@ const App: React.FC<AppProps> = ({scriptTag}) => {
             <CommentsFrame ref={iframeRef}>
                 <ContentBox done={done} />
             </CommentsFrame>
-            <AuthFrame adminUrl={options.adminUrl} onLoad={initAdminAuth}/>
+            {done ? <AuthFrame adminUrl={options.adminUrl} onLoad={initAdminAuth}/> : null}
```
allouis: My React is a little rusty - `null` won't render anything, right?

allouis (approved):

apps/comments-ui/src/App.tsx
kevinansfield: Right, `null` in this case means render nothing.

https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null

PR #19807, opened 2024-03-06
🐛 Fixed free tier showing in the tiers-only paywall in posts

### LINKED ISSUE #19796: {{tiers}} helper has regression in 5.80 - "free" tier shown in list.
### Issue Summary

The {{tiers}} helper is including the Free tier in the list of tiers with access, even for posts set only to a paid tier.

These are posts set to a specific tier access only. Here’s 5.80 (running Source latest)
image

And here’s 5.79, also Source latest:
image

Both are on Ghost Pro - it’s just a difference between whether they’re in the early or late roll-out group, looks like!

### Steps to Reproduce

Install 5.80, hook up to Stripe, create a tier. Create a post restricted to that tier.  Browse to the post while not logged in as a user with access.

### Ghost Version

Ghost Pro

### Node.js Version

Ghost Pro

Ghost Pro

### Database type

MySQL 8

### Browser & OS version

_No response_

_No response_

### Code of Conduct

sagzy:
refs INC-36
fixes https://github.com/TryGhost/Ghost/issues/19796

- The tiers-only paywall was incorrectly rendering "Free". Example: "This post is for subscribers of the Free, Silver and Gold tiers only"
- Steps to reproduce the issue:
    1. Create a post with public visibility, publish it
    2. Then swap the visibility to specific tiers. The default selects all paid tiers. Leave it like that
    3. Update the post. The paywall show Free, even though it should be showing only the paid tiers
- This fix filters out the "free" tier when visibility is set to tiers, before updating a Post or a Page. The fix includes bulk updates from the list of Posts and Pages (right-click on a Post/Page > Change Access).


ghost/core/test/e2e-api/admin/pages-bulk.test.js
```diff
@@ -144,10 +144,12 @@ describe('Pages Bulk API', function () {
                             visibility: 'tiers',
                             tiers: [
                                 {
-                                    id: tier1.id
+                                    id: freeTier.id,
+                                    type: freeTier.get('type')
```
SimonBackx, line 148: I don't think we should add this to the API in order to make it work

ghost/core/test/e2e-api/admin/posts-bulk.test.js
```diff
@@ -179,10 +179,12 @@ describe('Posts Bulk API', function () {
                             visibility: 'tiers',
                             tiers: [
                                 {
-                                    id: tier1.id
+                                    id: freeTier.id,
+                                    type: freeTier.get('type')
```
SimonBackx, line 183: Same as above

SimonBackx (changes requested):
I think we also need a non-bulk E2E test that tries to edit a post with a free tier and check if it isn't saved (because here was the main issue).

We should make sure we don't require changes in the API. Users should still be able to set the tiers of a post by only specifying the id's.

Apart from that we need to consider to also add the fix to the output serializer so we don't return the free tier for posts that already have the free tier linked (alternative would be a migration, but that is a lot heavier).

ghost/core/test/e2e-api/admin/pages-bulk.test.js
sagzy: In Admin, when the user right-clicks on a Post and then clicks on "Change Access", we receive complete `tier` objects (so including `id`, and `type`). Our tests didn't reflect that.

Is there any other use case for the bulk edit today, where we would only receive the tier `id`?

SimonBackx: Not for the bulk API, because I think this one is undocumented, however I would think it is better that it behaves the same as the single edit API. For the 'single edit API', AFAIK your changes also require the type to be set when changing tiers, maybe that isn't covered in tests at the moment, but this should be covered. 

For the single edit API we cannot introduce a breaking change in a not major Ghost version. It is possible that a site is using the admin API directly and is only sending the id's, not the full objects.

sagzy: Fair point — the change in the PR doesn't change the single edit API, but I follow the argument of not making these two differ in requirements. Updated

ghost/core/test/e2e-api/admin/pages.test.js
```diff
@@ -285,6 +285,57 @@ describe('Pages API', function () {
...
+                        .expectStatus(200)
+                        .matchHeaderSnapshot({
+                            'content-version': anyContentVersion,
+                            etag: anyEtag,
+                            'x-cache-invalidate': anyString
+                        })
+                        .matchBodySnapshot({
+                            pages: [Object.assign({}, matchPageShallowIncludes, {
+                                published_at: null,
+                                tiers: [{id: paidTier.id, ...tierSnapshot}]
```
SimonBackx, line 333: Shouldn't this be 
```suggestion
                                tiers: [{id: paidTier.id}]
```
?

ghost/core/test/e2e-api/admin/pages.test.js
```diff
@@ -285,6 +285,57 @@ describe('Pages API', function () {
...
+                    const [pageResponse] = pageBody.pages;
+
+                    await agent
+                        .put(`/pages/${pageResponse.id}`)
+                        .body({
+                            pages: [{
+                                id: pageResponse.id,
+                                updated_at: pageResponse.updated_at,
+                                visibility: 'tiers',
+                                tiers: [freeTier, paidTier]
```
SimonBackx, line 321: We should only set the ids here
```suggestion
                                tiers: [{id: freeTier.id}, {id: paidTier.id}]
```

ghost/core/test/e2e-api/members-comments/comments.test.js
```diff
@@ -1026,7 +1026,8 @@ describe('Comments API', function () {
                 visibility: 'tiers',
                 tiers: [
                     {
-                        id: product.id
+                        id: product.id,
+                        type: product.get('type')
```
SimonBackx, line 1030: This shouldn't change

ghost/core/core/server/models/post.js
```diff
@@ -1000,6 +1000,12 @@ Post = ghostBookshelf.Model.extend({
             });
         }
 
+        // If visibility is set to specific tiers, make sure we are only saving paid tiers
+        if (this.get('visibility') === 'tiers' && this.get('tiers')) {
+            const paidTiers = this.get('tiers').filter(t => t.type === 'paid');
+            this.set('tiers', paidTiers);
+        }
```
SimonBackx, line 1007: I don't think this will work, hence why I wanted the E2E tests added. If you only specify the ids, this won't have a type property, so it won't be able to filter it. Also the frontend can pass whatever they want here - so this isn't really validating

ghost/core/test/e2e-api/admin/pages.test.js
sagzy: This is to avoid timestamp issues with the snapshots. Otherwise I get failed tests because e.g. `updated_at` are not in sync

sagzy: This is to avoid timestamps issues with the `updated_at` fields are so on

SimonBackx (approved):

PR #19820, opened 2024-03-07
Extracted Unsplash Comp into own package

ronaldlangeveld:
no issue

We currently have 3 versions of the Unsplash selector; The Admin-X version, written in Typescript + React; The Koenig version, written in Javascript + React; and the original Ember version.

This creates a standalone Unsplash-Selector package in the Ghost Apps monorepo. The plan is to only have 1 package to maintain and be able to reuse it elsewhere to avoid having to maintain separate versions of a library that does exactly the same thing everywhere.


ronaldlangeveld:
In favour of publishing it via the Koenig repo instead - https://github.com/TryGhost/Koenig/pull/1165

PR #19843, opened 2024-03-12
Switched to throwing error upon failed image processing

daniellockyer:
ref https://linear.app/tryghost/issue/ENG-740/http-500-error-when-image-processing-fails refs https://github.com/TryGhost/Ghost/commit/4aad551c72cba5e927b7ac4ee529af2f30b049e8

- upon further discussion, we've decided it's better to throw an error in this case because the uploaded image is deemed invalid and storing it on the filesystem might cause more issues with resizing/further processing in the future
- this commit implements that and alters the tests


aral190:
One requires an account to view the linear issue @daniellockyer

PR #19859, opened 2024-03-13
Optimised queries made by get helper for posts

allouis:
ref ENG-747
ref https://linear.app/tryghost/issue/ENG-747

H'okay - so what we're trying to do here is make get helper queries more cacheable. The way we're doing that is by modifying the filter used when we're trying to remove a single post from the query.

The idea is that we can remove that restriction on the filter, increase the number of posts fetched by 1 and then filter the fetched posts back down, this means that the same query, but filtering different posts, will be updated to make _exactly_ the same query, and so share a cache!

We've been purposefully restrictive in the types of filters we manipulate, so that we only deal with the simplest cases and the code is easier to understand.

### DIFF
--- a/ghost/core/core/frontend/helpers/get.js
+++ b/ghost/core/core/frontend/helpers/get.js
@@ -11,6 +11,7 @@ const Sentry = require('@sentry/node');
 
 const _ = require('lodash');
 const jsonpath = require('jsonpath');
+const nqlLang = require('@tryghost/nql-lang');
 
 const messages = {
     mustBeCalledAsBlock: 'The {\\{{helperName}}} helper must be called as a block. E.g. {{#{helperName}}}...{{/{helperName}}}',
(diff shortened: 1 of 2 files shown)


ghost/core/core/frontend/helpers/get.js
```diff
@@ -131,6 +132,53 @@ function parseOptions(globals, data, options) {
...
+                for (const query of queries) {
+                    if ('id' in query) {
+                        if ('$ne' in query.id) {
+                            // This checks that there is only one occurence of a negative id filter
+                            // So we know it's safe to use the `replace` method
+                            if (apiOptions.filter.split('id:-').length === 2) {
+                                const idToFilter = query.id.$ne;
+
+                                // The default limit is 15, the addition order is to cast apiOptions.limit to a number
+                                const limit = apiOptions.limit ? 1 + apiOptions.limit : 16;
```
kevinansfield, line 154: Where does that default limit come from?

allouis: That's a great question, I can't actually find where - but it is our default limit across our apis

allouis: Yup - get a big plus one from me on that :D

PR #19869, opened 2024-03-14
🐛 Fixed missing source + resized images producing rendered 404

daniellockyer:
fixes https://linear.app/tryghost/issue/ENG-746/http-500-responses-when-handle-image-sizes-middleware-hits-missing

- in the event a request comes in for a resized image, but the source image does not exist, we return a rendered 404 page
- we do this because we pass the NotFoundError to `next`, which skips over the static asset code where we return a plaintext 404
- also included a breaking test that ensure we go to the next middleware without an error


allouis (commented):
I think an e2e test would be really valuable for this - that will catch any cases where the error thrown for a missing file changes and doesn't have the expected `errorType`

daniellockyer:
@allouis How about https://github.com/TryGhost/Ghost/pull/19869/commits/0bed06dc204cc9dc0ea30499b86c3e2cecb749c7?

allouis (approved):

PR #19876, opened 2024-03-18
Removed usage of getters for api endpoints

allouis:
ref ENG-761
ref https://linear.app/tryghost/issue/ENG-761

By removing getters we ensure that each endpoint is only instantiated once. This _may_ have an adverse effect on boot time, so we're going to keep an eye on that, and make any necessary adjustments to the implementation.

### DIFF
--- a/ghost/core/core/server/api/endpoints/identities.js
+++ b/ghost/core/core/server/api/endpoints/identities.js
@@ -4,11 +4,16 @@ const jwt = require('jsonwebtoken');
 const jose = require('node-jose');
 const issuer = urlUtils.urlFor('admin', true);
 
-const dangerousPrivateKey = settings.get('ghost_private_key');
-const keyStore = jose.JWK.createKeyStore();
-const keyStoreReady = keyStore.add(dangerousPrivateKey, 'pem');
+let dangerousPrivateKey;
+let keyStore;
+let keyStoreReady;
 
 const getKeyID = async () => {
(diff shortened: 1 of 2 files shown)


vikaspotluri123:
Out of curiosity, is there any memory overhead with doing this? i.e. now that the `require` calls are no longer deferred, will any services that were previously not loaded now be loaded even if they're never called? I'm not sure if all services are expected to be loaded e.g. in the bootstrap, making this a non-issue 😁

allouis:
@vikaspotluri123 There is an initial memory overhead yes, but as soon as the API is accessed, we're back to the same footprint

vikaspotluri123:
Right, but I'm wondering if/how many APIs are rarely/never instantiated during a boot lifetime. E.g. webhook management might not be done often. (With the asterisk that I'm not sure how the API is accessed within the core, just thinking from an Admin/Public API aspect)

allouis:
The way the API is mounted means that - once any Admin API endpoint is mounted - _every_ Admin API endpoint file is loaded - same for the Content API - so it's not granular on a endpoint by endpoint basis

allouis:
Closing for now, in favour of https://github.com/TryGhost/Ghost/pull/19880

We have a _lot_ of circular dependencies which are hidden by the lazy-loading of files here - and cleaning it up should be part of a larger task than optimising controller instantiation.

PR #19880, opened 2024-03-18
Cached api controller pipelines

allouis:
ref ENG-761
ref https://linear.app/tryghost/issue/ENG-761

Creating these pipelines is expensive, and we don't want to do it repeatedly for the same controller. Adding caching should reduce the amount of time spent setting up pipelines for each usage of the `get` helper.

### DIFF
--- a/ghost/api-framework/lib/pipeline.js
+++ b/ghost/api-framework/lib/pipeline.js
@@ -160,6 +160,8 @@ const STAGES = {
     }
 };
 
+const controllerMap = new Map();
+
 /**
  * @description The pipeline runs the request through all stages (validation, serialisation, permissions).
  *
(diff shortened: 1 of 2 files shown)


ghost/api-framework/lib/pipeline.js
```diff
@@ -267,6 +273,10 @@ const pipeline = (apiController, apiUtils, apiType) => {
         Object.assign(obj[method], apiImpl);
         return obj;
     }, {});
+
+    controllerMap.set(apiController, result);
```
daniellockyer, line 277: Would it be worth using `docName` as the key?

allouis: `docName` isn't unique - so no! 😝

daniellockyer: wtf

allouis: One of the things `docName` is used for is to determine which permissions to use - so the `pages` endpoint has a `docName` of `'posts'` - as does the `postsPublic` endpoint

daniellockyer: unbelievable

PR #19881, opened 2024-03-18
Added missing permissions to Contributor & Editor

allouis:
ref ENG-728
ref https://linear.app/tryghost/issue/ENG-728

This is NOT a functionality change. The Post#permissible method unit tests have been updated to pass `true` as `hasUserPermission` and we can see that the permission functionality remains the same.

The permissible method of the post model is responsible for removing permission based on the data that is being modified, but the permissions module is setup to allow the permissible method to grant permission - this means that we call permissible, even if the current actor doesn't have permission, this results in code that is hard to understand and manage.

We are going to be instead returning early if an actor does not have permission, this will allow permissible method signatures to be greatly simplified (removing the need for hasUserPermission, hasApiKeyPermission & hasMemberPermission arguments).

### DIFF
--- a/ghost/core/core/server/data/schema/fixtures/fixtures.json
+++ b/ghost/core/core/server/data/schema/fixtures/fixtures.json
@@ -929,7 +929,7 @@
                     "recommendation": ["browse", "read"]
                 },
                 "Author": {
-                    "post": ["browse", "read", "add"],
+                    "post": ["browse", "read", "edit", "add", "destroy"],
                     "setting": ["browse", "read"],
                     "slug": "all",
                     "tag": ["browse", "read", "add"],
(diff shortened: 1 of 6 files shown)


ghost/core/core/server/data/schema/fixtures/fixtures.json
```diff
@@ -929,7 +929,7 @@
                     "recommendation": ["browse", "read"]
                 },
                 "Author": {
-                    "post": ["browse", "read", "add"],
```
daniellockyer, line 932: We should also update these in the test fixtures: https://github.com/TryGhost/Ghost/blob/cdf45172a7f63c0cf773abfbd31349393dda3782/ghost/core/test/utils/fixtures/fixtures.json#L1125

allouis: yes, we should!!

allouis: I've updated these, and I've updated them to be more explicit too - pulling out only the permissions we want, because there is an (unused) post:publish permission

daniellockyer (approved):

PR #19882, opened 2024-03-18
Stripe Tax: Update customer address when needed

verement:
Use the billing address entered during checkout to assess the customer's location for Stripe Tax. This is also needed to avoid an error if an existing customer lacks an address.

<hr/>

The Stripe Tax integration, when enabled, works well for new subscriptions. It does not work when an existing member tries to upgrade their subscription and Stripe’s customer record does not already have address information Stripe can use to determine a tax jurisdiction, resulting in the following error:

> **Unable to initiate checkout session**
>
> Error: Automatic tax calculation in Checkout requires a valid address on the Customer. Add a valid address to the Customer or set either `customer_update[address]` to 'auto' or `customer_update[shipping]` to 'auto' to save the address entered in Checkout to the Customer.

This change instructs Stripe to update and use the customer’s billing address entered during checkout to determine the customer’s location for Stripe Tax. It does not include the same instruction when Stripe does not already have a customer record, as that will result in a different error (Error: `customer_update` can only be used with `customer`.)

### DIFF
--- a/ghost/stripe/lib/StripeAPI.js
+++ b/ghost/stripe/lib/StripeAPI.js
@@ -488,6 +488,14 @@ module.exports = class StripeAPI {
             stripeSessionOptions.customer_email = customerEmail;
         }
 
+        // Use the billing address entered during checkout to assess the
+        // customer's location for Stripe Tax. This is also needed to avoid an
+        // error if an existing customer lacks an address. It must not be used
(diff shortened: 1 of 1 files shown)


ronaldlangeveld:
Hi @verement! Thank you so much for your time creating this PR.
A similar fix has been merged by the team and the issue should be resolved.
Take care! ☺️

PR #19887, opened 2024-03-19
Added queueing middleware to handle high request volume

9larsons:
ref https://linear.app/tryghost/issue/CFR-4/
- added requeust queueing middleware (express-queue) to handle high request volume

We've done a fair amount of investigation around improving Ghost's resiliency to high request volume. While we believe this to be partly due to database connection contention, it also seems Ghost gets overwhelmed by the requests themselves. Implementing a simple queueing system allows us a simple lever to change the volume of requests Ghost is actually ingesting at any given time and gives us options besides simply increasing database connection pool size.


ghost/core/core/server/web/parent/middleware/index.js
```diff
@@ -2,5 +2,6 @@ module.exports = {
     emitEvents: require('./emit-events'),
     ghostLocals: require('./ghost-locals'),
     logRequest: require('./log-request'),
+    queueRequest: require('./queue-request'),
```
9larsons, line 5: We use `requestQueue` and `queueRequest`. We don't have a great standard given the other mw values. Could we be consistent about using `requestQueue`?

ghost/core/core/server/web/parent/middleware/queue-request.js
```diff
@@ -0,0 +1,76 @@
+const debug = require('@tryghost/debug')('middleware:queue-request');
+const metrics = require('@tryghost/metrics');
+const path = require('node:path');
+const expressQueue = require('express-queue');
+
+const DEFAULT_QUEUE_LIMIT = 100;
```
9larsons, line 6: All our testing was with this at 20 and it seemed throughput was reasonably fine. Do we have data to know that 100 is ok? Even locally we can see ghost fall over with ~70rps.

ghost/core/core/server/web/parent/app.js
```diff
@@ -23,6 +23,13 @@ module.exports = function setupParentApp() {
     // @TODO: figure out if this is really needed everywhere? Is it not frontend only...
     parentApp.use(mw.ghostLocals);
 
+    // Enable request queuing if configured
+    const queueConfig = config.get('optimization')?.requestQueue;
```
9larsons, line 27: Is your thought to have this as our 'on' switch for testing vs. having a different build? Makes sense to me.

ghost/core/core/server/web/parent/middleware/queue-request.js
```diff
@@ -0,0 +1,76 @@
...
+
+module.exports = function queueRequest(
+    config,
+    queueFactory = expressQueue,
+    logMetric = metrics.metric.bind(metrics)
+) {
+    config = {
+        ...{
+            limit: DEFAULT_QUEUE_LIMIT
```
9larsons, line 15: `DEFAULT_QUEUE_LIMIT` is a confusing name to me because it's the active request limit. Could we change the naming to be `ACTIVE_REQUEST_LIMIT` or something similar?

ghost/core/core/server/web/parent/middleware/queue-request.js
```diff
@@ -0,0 +1,76 @@
...
+        },
+        ...config
+    };
+
+    debug('Initialising middleware with config:', config);
+
+    // @see https://github.com/alykoshin/express-queue#usage
+    const queue = queueFactory({
+        activeLimit: config.limit,
+        queuedLimit: -1 // Do not limit the number of queued requests
```
9larsons, line 25: Can we think of a reason to need a config value for `queuedLimit`? I think we're ok with it kept at -1 because other parts of Pro have rate limits 🤔.

ghost/core/core/server/web/parent/middleware/queue-request.js
```diff
@@ -0,0 +1,76 @@
...
+        logMetric('request-queue', {
+            event: 'request-completed',
+            queueLength: job.queue.getLength()
+            // @todo: What else do we want to log here?
+        });
+    });
+
+    return (req, res, next) => {
+        const ext = path.extname(req.path);
+        const staticAssetExts = ['.css', '.js', '.map', '.woff2', '.ico']; // @todo: extend this list to include other static asset types
```
9larsons, line 63: Nicely done here 🥇. We can hopefully scrape some existing logs for more data, or we can do some local testing to get more exts.

ghost/core/test/unit/server/web/parent/middleware/queue-request.test.js
```diff
@@ -0,0 +1,154 @@
+const assert = require('node:assert');
```
9larsons, line 1: Also great job on tests 🙏.

ghost/core/core/server/web/parent/middleware/queue-request.js
```diff
@@ -0,0 +1,76 @@
...
+     * - cancel - when a request is cancelled
+     * - complete - when a request has completed
+     *
+     * @see https://github.com/search?q=repo:alykoshin/mini-queue%20job._toState&type=code
+     */
+
+    queue.queue.on('queue', (job) => {
+        debug(`Request queued: ${job.data.req.path}`);
+
+        logMetric('request-queue', {
```
9larsons, line 43: I do think logging the metrics is helpful, but one of the troubles without using `GhostLogger` is that we will be changing the definition of the `res.time` field in elastic. With the current approach here we aren't using the 'total time in Ghost' as the `res.time` value, just the time spent out of the queue. In order to get a good comparison I think we want to update the `req.startTime` and `res.responseTime` fields.

9larsons (commented):
Several questions/comments on the design. Implementation looks great, as do the tests!

ghost/core/core/server/web/parent/app.js
mike182uk: Yep, means we don't have to worry about builds getting swapped out on new releases 👍

ghost/core/core/server/web/parent/middleware/index.js
mike182uk: heh, i actually opted for `queueRequest` in this situation as the convention for the other middleware was `<verb>Something` (`logRequest`, `emitEvents`) etc.

ghost/core/core/server/web/parent/middleware/queue-request.js
mike182uk: Definitely not 😄 We have a task to workout what this specific value should be, this was just a placeholder till we worked that out

ghost/core/core/server/web/parent/middleware/queue-request.js
mike182uk: Yeh I fumbled the terminology here. How about `CONCURRENCY_LIMIT`?

ghost/core/core/server/web/parent/middleware/queue-request.js
mike182uk: 🤔 Could potentially reduce memory bloat under heavy load (not having to store the reqs in the queue)

If this value is set we can provide a handler that gets executed (i.e throw a `503` or whatever)

tbh I think it should be ok as is, can always tweak if needed 👍

ghost/core/core/server/web/parent/middleware/queue-request.js
mike182uk: Was actually thinking, would it be safe to do the inverse and if there is no extension automatically queue the request? I think we pretty much use extension-less urls everywhere 🤔

ghost/core/core/server/web/parent/middleware/queue-request.js
9larsons: Works for me.

ghost/core/core/server/web/parent/middleware/queue-request.js
9larsons: `express-queue` should throw a 503 by default. Sounds good though.

ghost/core/core/server/web/parent/middleware/queue-request.js
mike182uk: 👌 Updated in https://github.com/TryGhost/Ghost/pull/19887/commits/67856d3e804bd43257a9078a801070d769670808

ghost/core/core/server/web/parent/middleware/queue-request.js
mike182uk: Discussed on call, having `queueDepth` on the req is valuable for debugging. Have moved from metric to request in https://github.com/TryGhost/Ghost/pull/19887/commits/51da221ef812d693bf534d20d2ff359ce3e82703 👍

ghost/core/core/server/web/parent/middleware/queue-request.js
9larsons: Sounds like a plan. We can audit this in testing 👍.

mike182uk: 👍 Updated in https://github.com/TryGhost/Ghost/pull/19887/commits/fae6606bec8e59d2c2bc70470058970d968e368b

daniellockyer:
@9larsons You'll also need to edit this line: https://github.com/TryGhost/Ghost/blob/cdf45172a7f63c0cf773abfbd31349393dda3782/package.json#L46

Literally don't ask me why, yarn does weird things

PR #19917, opened 2024-03-25
Removed use of subqueries in email analytics queries

kevinansfield:
closes https://linear.app/tryghost/issue/ENG-790/remove-use-of-sub-queries-in-email-analytics

Avoiding sub queries means we don't have a process tied up for longer than necessary and we can more easily see if one of the queries is non-performant.

- extracted the count queries into separate queries and used the retrieved values in the final update query
- removed a query by moving the email open rate calculation into JS as we've already fetched the necessary data before that point
- optimized calculation of `delivered_count` by switching from `IS NOT NULL` to `IS NULL` to match the typical data stored in that column so it needs to read far fewer rows from the index when counting (see below for before/after)

---

`delivered_at IS NOT NULL` vs `delivered_at IS NULL`

```
mysql> explain analyze SELECT COUNT(id) FROM email_recipients WHERE email_id = '00000000846f9cc01e7bd3cc' AND delivered_at IS NOT NULL \G;
*************************** 1. row ***************************
EXPLAIN: -> Aggregate: count(email_recipients.id)  (cost=47067 rows=1) (actual time=71.9..71.9 rows=1 loops=1)
    -> Filter: ((email_recipients.email_id = '00000000846f9cc01e7bd3cc') and (email_recipients.delivered_at is not null))  (cost=32464 rows=146030) (actual time=0.0552..69 rows=75682 loops=1)
        -> Covering index range scan on email_recipients using email_recipients_email_id_delivered_at_index over (email_id = '00000000846f9cc01e7bd3cc' AND NULL < delivered_at)  (cost=32464 rows=146030) (actual time=0.0511..43.2 rows=75682 loops=1)

mysql> explain analyze SELECT COUNT(id) FROM email_recipients WHERE email_id = '00000000846f9cc01e7bd3cc' AND delivered_at IS NULL \G;
*************************** 1. row ***************************
EXPLAIN: -> Aggregate: count(email_recipients.id)  (cost=1477 rows=1) (actual time=4.44..4.44 rows=1 loops=1)
    -> Filter: (email_recipients.delivered_at is null)  (cost=813 rows=6638) (actual time=0.23..4.23 rows=3593 loops=1)
        -> Covering index lookup on email_recipients using email_recipients_email_id_delivered_at_index (email_id='00000000846f9cc01e7bd3cc', delivered_at=NULL)  (cost=813 rows=6638) (actual time=0.229..3.9 rows=3593 loops=1)

### DIFF
--- a/ghost/core/core/server/services/email-analytics/lib/queries.js
+++ b/ghost/core/core/server/services/email-analytics/lib/queries.js
@@ -41,12 +41,14 @@ module.exports = {
     },
 
     async aggregateEmailStats(emailId) {
-        const [deliveredCount] = await db.knex('email_recipients').count('id as count').whereRaw('email_id = ? AND delivered_at IS NOT NULL', [emailId]);
+        const {totalCount} = await db.knex('emails').select(db.knex.raw('email_count as totalCount')).where('id', emailId).first() || {totalCount: 0};
+        // use IS NULL here because that will typically match far fewer rows than IS NOT NULL making the query faster
+        const [undeliveredCount] = await db.knex('email_recipients').count('id as count').whereRaw('email_id = ? AND delivered_at IS NULL', [emailId]);
         const [openedCount] = await db.knex('email_recipients').count('id as count').whereRaw('email_id = ? AND opened_at IS NOT NULL', [emailId]);
         const [failedCount] = await db.knex('email_recipients').count('id as count').whereRaw('email_id = ? AND failed_at IS NOT NULL', [emailId]);
 
         await db.knex('emails').update({
-            delivered_count: deliveredCount.count,
+            delivered_count: totalCount - undeliveredCount.count,
             opened_count: openedCount.count,
             failed_count: failedCount.count
         }).where('id', emailId);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: ghost/core/core/server/services/email-analytics/lib/queries.js [hunk 1]
```diff
@@ -41,10 +41,14 @@ module.exports = {
     },
 
     async aggregateEmailStats(emailId) {
+        const [deliveredCount] = await db.knex('email_recipients').count('id as count').whereRaw('email_id = ? AND delivered_at IS NOT NULL', [emailId]);
```
daniellockyer, line 44: Would it be more efficient to do these using `Promise.all`?

kevinansfield: Possibly but wouldn't that add extra load on the db at this point which is something we're trying to avoid?

kevinansfield: On Ghost's side we're not hugely worried about overall runtime here.

kevinansfield:
There is some potential improvement here by optimising our use of `IS NULL` vs `IS NOT NULL` based on typical usage patterns. See https://github.com/TryGhost/Ghost/pull/19918#issuecomment-2020393159

allouis (approved):

daniellockyer (approved):

PR #19918, opened 2024-03-25
 Fixed email_recipients indexes to match query usage

kevinansfield:
closes https://linear.app/tryghost/issue/ENG-791/migration-to-fix-email-recipients-indexes

Our indexes over single columns (`delivered_at`, `opened_at`, `failed_at`) were ineffective because the only time we query those is alongside `email_id` meaning we were frequently performing full table scans on very large tables during our email analytics jobs.

- added migration to add new indexes covering `email_id` and the respective columns
- added migration to drop the old indexes that weren't being used in any query plans

Local runtime with ~2M email_recipient rows:
- before: 1.7s
- after: 99ms

Explain output...

before:
```
+----+-------------+------------------+------------+-------+----------------------------------------------------------------------------------+----------------------------------------------+---------+-------+--------+----------+------------------------------------+
| id | select_type | table            | partitions | type  | possible_keys                                                                    | key                                          | key_len | ref   | rows   | filtered | Extra                              |
+----+-------------+------------------+------------+-------+----------------------------------------------------------------------------------+----------------------------------------------+---------+-------+--------+----------+------------------------------------+
|  1 | UPDATE      | emails           | NULL       | index | NULL                                                                             | PRIMARY                                      | 98      | NULL  |      1 |   100.00 | Using where                        |
|  4 | SUBQUERY    | email_recipients | NULL       | range | email_recipients_email_id_member_email_index,email_recipients_failed_at_index    | email_recipients_failed_at_index             | 6       | NULL  |   2343 |     7.76 | Using index condition; Using where |
|  3 | SUBQUERY    | email_recipients | NULL       | ref   | email_recipients_email_id_member_email_index,email_recipients_opened_at_index    | email_recipients_email_id_member_email_index | 98      | const | 159126 |    50.00 | Using where                        |
|  2 | SUBQUERY    | email_recipients | NULL       | ref   | email_recipients_email_id_member_email_index,email_recipients_delivered_at_index | email_recipients_email_id_member_email_index | 98      | const | 159126 |    50.00 | Using where                        |
+----+-------------+------------------+------------+-------+----------------------------------------------------------------------------------+----------------------------------------------+---------+-------+--------+----------+------------------------------------+
```

after:
```
+----+-------------+------------------+------------+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------+---------+------+--------+----------+--------------------------+
| id | select_type | table            | partitions | type  | possible_keys                                                                                                                                                                 | key                                          | key_len | ref  | rows   | filtered | Extra                    |
+----+-------------+------------------+------------+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------+---------+------+--------+----------+--------------------------+
|  1 | UPDATE      | emails           | NULL       | index | NULL                                                                                                                                                                          | PRIMARY                                      | 98      | NULL |      1 |   100.00 | Using where;             |
|  4 | SUBQUERY    | email_recipients | NULL       | range | email_recipients_email_id_member_email_index,email_recipients_email_id_delivered_at_index,email_recipients_email_id_opened_at_index,email_recipients_email_id_failed_at_index | email_recipients_email_id_failed_at_index    | 104     | NULL |     60 |   100.00 | Using where; Using index |
|  3 | SUBQUERY    | email_recipients | NULL       | range | email_recipients_email_id_member_email_index,email_recipients_email_id_delivered_at_index,email_recipients_email_id_opened_at_index,email_recipients_email_id_failed_at_index | email_recipients_email_id_opened_at_index    | 104     | NULL | 119496 |   100.00 | Using where; Using index |
|  2 | SUBQUERY    | email_recipients | NULL       | range | email_recipients_email_id_member_email_index,email_recipients_email_id_delivered_at_index,email_recipients_email_id_opened_at_index,email_recipients_email_id_failed_at_index | email_recipients_email_id_delivered_at_index | 104     | NULL | 146030 |   100.00 | Using where; Using index |
+----+-------------+------------------+------------+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------+---------+------+--------+----------+--------------------------+
```

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.82/2024-03-25-16-46-10-add-email-recipients-email-id-indexes.js
+++ b/ghost/core/core/server/data/migrations/versions/5.82/2024-03-25-16-46-10-add-email-recipients-email-id-indexes.js
@@ -1,0 +1,17 @@
+// For information on writing migrations, see https://www.notion.so/ghost/Database-migrations-eb5b78c435d741d2b34a582d57c24253
+
+const {createNonTransactionalMigration} = require('../../utils');
+const commands = require('../../../schema/commands');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        await commands.addIndex('email_recipients', ['email_id', 'delivered_at'], knex);
+        await commands.addIndex('email_recipients', ['email_id', 'opened_at'], knex);
+        await commands.addIndex('email_recipients', ['email_id', 'failed_at'], knex);
+    },
+    async function down(knex) {
+        await commands.dropIndex('email_recipients', ['email_id', 'delivered_at'], knex);
+        await commands.dropIndex('email_recipients', ['email_id', 'opened_at'], knex);
+        await commands.dropIndex('email_recipients', ['email_id', 'failed_at'], knex);
+    }
+);
--- a/ghost/core/core/server/data/migrations/versions/5.82/2024-03-25-16-51-29-drop-email-recipients-non-email-id-indexes.js
+++ b/ghost/core/core/server/data/migrations/versions/5.82/2024-03-25-16-51-29-drop-email-recipients-non-email-id-indexes.js
@@ -1,0 +1,17 @@
+// For information on writing migrations, see https://www.notion.so/ghost/Database-migrations-eb5b78c435d741d2b34a582d57c24253
+
+const {createNonTransactionalMigration} = require('../../utils');
+const commands = require('../../../schema/commands');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        await commands.dropIndex('email_recipients', ['delivered_at'], knex);
+        await commands.dropIndex('email_recipients', ['opened_at'], knex);
+        await commands.dropIndex('email_recipients', ['failed_at'], knex);
+    },
+    async function down(knex) {
+        await commands.addIndex('email_recipients', ['delivered_at'], knex);
+        await commands.addIndex('email_recipients', ['opened_at'], knex);
+        await commands.addIndex('email_recipients', ['failed_at'], knex);
+    }
+);
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -865,14 +865,17 @@ module.exports = {
         member_id: {type: 'string', maxlength: 24, nullable: false, index: true},
         batch_id: {type: 'string', maxlength: 24, nullable: false, references: 'email_batches.id'},
         processed_at: {type: 'dateTime', nullable: true},
-        delivered_at: {type: 'dateTime', nullable: true, index: true},
-        opened_at: {type: 'dateTime', nullable: true, index: true},
-        failed_at: {type: 'dateTime', nullable: true, index: true},
+        delivered_at: {type: 'dateTime', nullable: true},
+        opened_at: {type: 'dateTime', nullable: true},
+        failed_at: {type: 'dateTime', nullable: true},
         member_uuid: {type: 'string', maxlength: 36, nullable: false},
         member_email: {type: 'string', maxlength: 191, nullable: false},
         member_name: {type: 'string', maxlength: 191, nullable: true},
         '@@INDEXES@@': [
-            ['email_id', 'member_email']
+            ['email_id', 'member_email'],
+            ['email_id', 'delivered_at'],
+            ['email_id', 'opened_at'],
+            ['email_id', 'failed_at']
         ]
     },
     email_recipient_failures: {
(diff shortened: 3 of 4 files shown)


erik-ghost:
@kevinansfield there's a few more things I think we should get clear on:

- are we positive we're avoiding hitting caches when we're getting these timings?  Can you describe the process you're using to get the timings, and are you using SQL_NO_CACHE?  You can see some more details [here](https://forums.mysql.com/read.php?24,225286,225468#msg-225468) for what variables to look for to understand if you're hitting caches.
- adding more indexes will likely incur a penalty in performance of writing to that table.  How can we reason about what impact this change has on the performance of the writes to this table?
- although adding the indexes improves the performance, we're still hitting a lot of rows (as shown in the explain), by the _Using where_: this likely means that although it's using an index, it's still having to do a table scan for that number of rows.  Changing this so that the explain output says _Using index_ will be an improvement.
- This is likely because we're doing a IS NOT NULL, you can read more about that [here](https://dev.mysql.com/doc/refman/8.3/en/explain-output.html#jointype_ref_or_null) and you can spot that "ref" column in the explain output change from const to NULL.
- [IS NULL optimization](https://dev.mysql.com/doc/refman/8.0/en/is-null-optimization.html) can be a bit of a tricky subarea, can you include the full show create table output for this table before and after the index changes?

Thank you!

kevinansfield:
> are we positive we're avoiding hitting caches when we're getting these timings? Can you describe the process you're using to get the timings, and are you using SQL_NO_CACHE? You can see some more details [here](https://forums.mysql.com/read.php?24,225286,225468#msg-225468) for what variables to look for to understand if you're hitting caches.

There's no query cache in MySQL 8 as far as I can tell. To avoid cases of data being in ram I was restarting the server between query runs.

kevinansfield:
> adding more indexes will likely incur a penalty in performance of writing to that table. How can we reason about what impact this change has on the performance of the writes to this table?

In this case we're replacing an already existing index so we shouldn't have huge differences. I can take a look at timing a batch insert operation.

kevinansfield:
> although adding the indexes improves the performance, we're still hitting a lot of rows (as shown in the explain), by the Using where: this likely means that although it's using an index, it's still having to do a table scan for that number of rows. Changing this so that the explain output says Using index will be an improvement.

I did try looking into the `Using where;` comment earlier but everything I found was inconclusive, it seems prevailing thought was it's not very indicative of anything because it can show up simply because there's a where condition in the query ([the docs are also very ambiguous but seem to support that theory](https://dev.mysql.com/doc/refman/8.3/en/explain-output.html#jointype_ref), especially in this case where we're passing the restricted row set to the COUNT function?).

erik-ghost:
Could we use the [optimizer cost model](https://dev.mysql.com/doc/refman/8.3/en/cost-model.html) to find out what the impact of the cost of the query is to MySQL?

kevinansfield:
> Could we use the [optimizer cost model](https://dev.mysql.com/doc/refman/8.3/en/cost-model.html) to find out what the impact of the cost of the query is to MySQL?

I had used `EXPLAIN ANALYZE` on the individual selects (it doesn't work for UPDATE queries) to check and compare when I was experimenting yesterday. It appeared considerably better with the indexes in place

before:
```
mysql> explain analyze SELECT COUNT(id)FROM email_recipients WHERE email_id='00000000846f9cc01e7bd3cc' AND delivered_at IS NOT NULL \G;
*************************** 1. row ***************************
EXPLAIN: -> Aggregate: count(email_recipients.id)  (cost=88993 rows=1) (actual time=506..506 rows=1 loops=1)
    -> Filter: (email_recipients.delivered_at is not null)  (cost=81036 rows=79563) (actual time=3.08..504 rows=75682 loops=1)
        -> Index lookup on email_recipients using email_recipients_email_id_member_email_index (email_id='00000000846f9cc01e7bd3cc')  (cost=81036 rows=159126) (actual time=3.07..501 rows=79275 loops=1)
```

after:
```
mysql> explain analyze SELECT COUNT(id)FROM email_recipients WHERE email_id='00000000846f9cc01e7bd3cc' AND delivered_at IS NOT NULL \G;
*************************** 1. row ***************************
EXPLAIN: -> Aggregate: count(email_recipients.id)  (cost=47372 rows=1) (actual time=73.7..73.7 rows=1 loops=1)
    -> Filter: ((email_recipients.email_id = '00000000846f9cc01e7bd3cc') and (email_recipients.delivered_at is not null))  (cost=32769 rows=146030) (actual time=0.093..70.8 rows=75682 loops=1)
        -> Covering index range scan on email_recipients using email_recipients_email_id_delivered_at_index over (email_id = '00000000846f9cc01e7bd3cc' AND NULL < delivered_at)  (cost=32769 rows=146030) (actual time=0.0884..45.5 rows=75682 loops=1)
```

allouis (approved):

daniellockyer (approved):

PR #19920, opened 2024-03-25
Added Macedonian language (mk)

drangovski:
Added translation for Macedonian language (locales/mk)


royalfig:
@drangovski Thanks for the contribution!

PR #19924, opened 2024-03-26
Updated time formatting for drafts and scheduled posts

sanne-san:
Ref TRI-27
- Published posts now show the published date in post list, instead of updated date.
- The `gh-format-post-time` helper now has a `relative` and `absolute` and option instead of formatting being tied to `draft` and `published` state. This allows for more flexibility in how dates are displayed.
- Draft, scheduled and published posts now follow the same time formatting pattern: today, yesterday, or explicit dates if further in the past.
- Hover states for dates in the post list have been removed.
- The scheduling logic on the publish page still uses relative formatting.


kevinansfield (approved):

kevinansfield (approved):
Small test fix needed but otherwise looks great 🙌

PR #19935, opened 2024-03-27
Added SiteID to ConnectionPoolInstrumentation

ronaldlangeveld:
ref CFR-11

- added SiteID from config to the request logging to be able to track database connection timing better.


mike182uk:
This is no longer required

PR #19987, opened 2024-04-04
Added queue depth to requests

mike182uk:
refs [CFR-14](https://linear.app/tryghost/issue/CFR-14/ensure-queue-depth-is-always-set-on-req)

Added queue depth to any request that passes through the request queue middleware instead of only adding it to the request if it is queued. This makes it easier to report on the queue depth within Elastic.


ghost/core/test/unit/server/web/parent/middleware/queue-request.test.js
ronaldlangeveld, line 1: Great tests! 👌

ronaldlangeveld (approved):
LGTM and tests look great. :)

PR #20031, opened 2024-04-17
🐛 Fixed redundant member data loading for static assets

ronaldlangeveld:
refs CFR-21

Reorganised middleware execution so that member data is not redundantly loaded for static assets or the sitemap


ghost/core/core/frontend/web/middleware/members-session-handler.js
```diff
@@ -0,0 +1,25 @@
+const membersService = require('../../../server/services/members');
+
+function conditionalMemberSessionMiddleware(req, res, next) {
+    const bypassMembersSession = [
```
mike182uk, line 4: 🤔 I'm not sure I'm too keen on this -  I don't think it should be the responsibility of this middleware to know the specifics of defined routes (and this would likely turn into a maintenance burden)

Would be better to somehow inverse this - Let the route decide if it needs the member session?

Or could it be as simple as attempting to load the member session later in the request cycle (after the middleware processing the static assets has been executed)?

ronaldlangeveld: Thanks for the feedback! I would be great to rubber duck a bit around this. ☺️

My initial idea was to pass in the member session middleware passed into routes, but after investigation, I realised there may be too many variables involved in the frontend app (it's not quite like the admin/content API with each route explicitly defined).
That lead me to do it at middleware level, since the member session is already passed in globally, it would be simpler to filter out the routes by the time they are already defined, in the middleware.

Either way, I think we have some decent testing on this PR, so we can play around and see what works best in terms of maintainability and best practices.

PR #20051, opened 2024-04-18
🐛 Fixed admin crash when deleting an unsaved or imported post

cmraible:
ref https://linear.app/tryghost/issue/ENG-845/error-attempted-to-set-lexical-on-the-deleted-record
ref https://linear.app/tryghost/issue/ENG-854/🐛-deleting-imported-posts-makes-ghost-unresponsive

- When deleting a post in the editor's Post Settings Menu, if the post has unsaved changes (indicated by the `hasDirtyAttributes` property in the editor), Admin will crash because it tries to save a post revision before leaving the editor, but the post has already been deleted so saving fails.
- This can occur when editing a post and quickly deleting it from the Post Settings Menu before saving is completed.
- It can also occur when attempting to delete an imported post, as the editor will parse the `lexical` from the server and may make some minor, invisible-to-the-user changes to the `lexical` string locally (e.g. JSON formatting, or updating the JSON to use extended version of base lexical nodes), which triggers the same error.
- This fix bypasses the attempt to save a post revision when leaving the editor if the post is already deleted, which allows the transition back to the Posts route to succeed.

### DIFF
--- a/ghost/admin/app/components/gh-post-settings-menu.hbs
+++ b/ghost/admin/app/components/gh-post-settings-menu.hbs
@@ -219,7 +219,7 @@
                 </form>
                 {{#unless this.post.isNew}}
                     <div class="settings-menu-delete-button">
-                        <button type="button" class="gh-btn gh-btn-outline gh-btn-icon gh-btn-fullwidth" {{action "deletePostInternal"}}>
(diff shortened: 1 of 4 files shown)


cmraible:
Re-opening from a branch on upstream to run browser tests in CI.

PR #20069, opened 2024-04-22
Updated staff deletion logic

9larsons:
ref https://linear.app/tryghost/issue/ENG-826

- Changed staff deletion logic to do a bulk insert when adding a tag to the users' associated posts

Staff deletion logic has really poor performance at scale because we do individual updates for every post. If a user has dozens+ posts (especially in a large db with thousands of posts), this can take >60s and look like a timeout. Ultimately this should probably be a jobbed off process, but for the time being we can improve this by doing a bulk insert.

Note that this update uses the pattern for the bulk tagging of posts from the right click (bulk) actions in the posts lists in Admin. With bulk actions, **we do not trigger web hooks or the post.edited events**. We will document this and follow up on this separately.


ghost/core/core/server/services/Users.js
```diff
@@ -76,68 +77,72 @@ class Users {
...
-                await this.models.Post.edit({
-                    tags: [...post.relations.tags.models, tag]
-                }, {
-                    id: postId,
-                    context,
-                    transacting
-                });
-            }
+        // get the posts that do not have the tag already assigned
+        let usersPostIds = [];
```
allouis, line 114: v. minor, style only

It seems to me that `createdTag` is almost always gonna be true - it's like an edge case if it's not?

If that's the case, I would prefer that we make it a lil more explicit like

```js
let usersPostIds = userPosts.map(p => p.post_id);

// CASE: tag already existing and we don't wanna add it twice.
if (!createdTag) {
    const tagId = tag.get('id');
    const taggedPostIds = await this.models.Base.knex('posts_tags')
        .transacting(transacting).where('tag_id', tagId).select('post_id');

    userPostIds = userPostIds.filter(p => !taggedPostIds.includes(p));
}

allouis (commented):
Change looks good! One NABD comment about style

On the events, I do think we should keep those in, but I realise theres gonna be some overhead because we'll need the bookshelf model for the event.

This is another reason to decouple from bookshelf!

Would you mind looking into events and seeing how much/if it affects performance? We don't have to do it in the request/response cycle either - we can do it as a job outside of the current execution.

ghost/core/core/server/services/Users.js
9larsons: Ha yeah I thought about this as I was writing it and meant to come back to it. Agreed - it'll pretty much always be the case. Almost so much so that we shouldn't bother handling it. (almost)

9larsons:
I fixed the Actions table not being appropriately updated with the posts that were edited. I followed a similar pattern here as #bulkAddTags. As a note, we don't fire events for this bulk action, either.

It definitely feels like we need to scrub any instances of inserts using the `models.Base.knex()...insert(...)` because of this need, if it is a need. Otherwise we're not consistent across the product. This also feels like something that ought to be handled in the model. Here we're doing actions created in two different services (Users and Posts) that imo have no business being in the service unless services become aware of each other.

ghost/core/core/server/services/Users.js
allouis: (almost) 😝

PR #20085, opened 2024-04-24
Added Redis via Docker

mike182uk:
no refs

Redis can be utilised for various caching purposes within Ghost. This PR adds a Redis service to the docker-compose file to allow for easier local development when Redis is required

### DIFF
--- a/.github/scripts/docker-compose.yml
+++ b/.github/scripts/docker-compose.yml
@@ -22,7 +22,7 @@ services:
       interval: 1s
       retries: 120
   redis:
-    image: redis:7.2
+    image: redis:7.0
     container_name: ghost-redis
     ports:
       - "6379:6379"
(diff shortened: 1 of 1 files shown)


.github/scripts/docker-compose.yml
```diff
@@ -21,3 +21,9 @@ services:
       test: "mysql -uroot -proot ghost -e 'select 1'"
       interval: 1s
       retries: 120
+  redis:
+    image: redis:7.2
```
daniellockyer, line 25: Would you be able to use `7.0`? It's what we currently use elsewhere, so it makes sense to keep those aligned for now

mike182uk: 👍 updated in https://github.com/TryGhost/Ghost/pull/20085/commits/7b01bb8fe4f8a59ce54d8fca51fb49340a34c589

daniellockyer (approved):

PR #20110, opened 2024-05-01
Update gscan dependency to v4.43.1

sagzy:
ref https://linear.app/tryghost/issue/ENG-842/gluster-file-name-length-limit

- this version of gscan throws an error if the zip contains large filenames


sagzy:
Closing as duplicate of #20109

PR #20117, opened 2024-05-02
Added cache-control: private headers for all link redirects

cmraible:
ref https://linear.app/tryghost/issue/ENG-883/add-cache-control-private-headers-to-the-link-redirects

- Link redirects shouldn't be cached because we record which member clicked the link on the redirect
- This PR adds `cache-control: private` headers to all link redirects to prevent them from being cached


cmraible:
PR ready to go, just waiting to merge because I have a larger change in the same area going in soon and I don't want to skew my validation results by making two changes at the same time.

cmraible:
Closing for now as this would just add fuel to our redirect problem at the moment.

PR #20118, opened 2024-05-02
Added `latest_event_timestamp` to emails table

ronaldlangeveld:
ref ENG-832

- Added migrations for `latest_event_timestamp` column in emails table.
- updated schema
- updated emails model


sagzy (approved):
LGTM 👍

PR #20121, opened 2024-05-02
[WIP] HTTP Signatures

allouis:
🍦

### DIFF
--- a/ghost/ghost/src/core/activitypub/http-signature.service.ts
+++ b/ghost/ghost/src/core/activitypub/http-signature.service.ts
@@ -1,0 +1,193 @@
+import crypto from 'node:crypto';
+
+type Signature = {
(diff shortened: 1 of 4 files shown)


ghost/ghost/src/core/activitypub/http-signature.service.ts
```diff
@@ -0,0 +1,156 @@
...
+            requestMethod,
+            requestUrl,
+            requestHeaders,
+            requestBody: requestBody.toString('base64')
+        });
+        const signatureData = this.parseSignatureHeader(
+            requestHeaders.signature
+        );
+
+        if (requestMethod === 'post') {
```
sagzy, line 129: This block is not being called by the tests because of a case mismatch "post" vs "POST"

PR #20134, opened 2024-05-06
Optimised `getLastSeenEventTimestamp` function

ronaldlangeveld:
ref ENG-832

depends on https://github.com/TryGhost/Ghost/pull/20118 for migration column in emails table, will rebase accordingly.
    
- every time `handleDelivered`, `handleOpened` and `handlePermanentFailed` fires, it will also update the `latest_event_timestmap` in the email table.
- we then conditionally remove 3 email queries in favour of 1 query that will get the latest event timestamp
- added additional tests


ghost/core/core/server/services/email-analytics/lib/queries.js
```diff
@@ -14,30 +13,36 @@ module.exports = {
...
-
-        if (maxOpenedAt && !(maxOpenedAt instanceof Date)) {
-            // SQLite returns a string instead of a Date
-            maxOpenedAt = new Date(maxOpenedAt);
+        // First, try to fetch the max latest_event_timestamp
+        let {maxTimestamp} = await db.knex('emails').select(db.knex.raw('MAX(latest_event_timestamp) as maxTimestamp')).first() || {};
+
+        if (maxTimestamp) {
+            if (!(maxTimestamp instanceof Date)) {
+                maxTimestamp = new Date(maxTimestamp);
```
sagzy, line 21: Maybe worth explaining why we need to convert to a Date here?

// SQLite returns a string instead of a Date

ghost/core/core/server/services/email-analytics/lib/queries.js
```diff
@@ -14,30 +13,36 @@ module.exports = {
...
+
+            if (maxOpenedAt && !(maxOpenedAt instanceof Date)) {
+                // SQLite returns a string instead of a Date
+                maxOpenedAt = new Date(maxOpenedAt);
+            }
+
+            if (maxFailedAt && !(maxFailedAt instanceof Date)) {
+                // SQLite returns a string instead of a Date
+                maxFailedAt = new Date(maxFailedAt);
+            }
```
sagzy, line 41: I'm guessing we're keeping the old logic to smoothen the rollout, as existing emails won't have the `latest_event_timestamp` field. Shall we make a note to clean this up in a couple of weeks from now?

sagzy: You could also return early here, to separate the new logic from the old one, e.g.

```
if (maxTimestamp) {
    if (!(maxTimestamp instanceof Date)) {
        maxTimestamp = new Date(maxTimestamp);
     }
              
     return maxTimestamp;
}


// Legacy logic

...
```

sagzy (approved):
LGTM 👍

ghost/core/core/server/services/email-analytics/lib/queries.js
ronaldlangeveld: Yes correct, I'll add a note in the comments.

PR #20144, opened 2024-05-07
Added OpenTelemetry instrumentation to Ghost backend

cmraible:
This commit adds OpenTelemetry instrumentation to Ghost's backend, which allows us to view traces similar to what we see in Sentry Performance locally.

OpenTelemetry is enabled if `NODE_ENV === 'development'` or if it is explicitly enabled via config with `opentelemetry:enabled`.

It also adds a [Jaeger](https://www.jaegertracing.io/) container to Ghost's docker-compose file for viewing the traces. There's no setup required (beyond running `yarn docker:reset` to pickup the changes in the docker-compose file the first time — but this will also reset your DB so be careful). This will launch the Jaeger container, and you can view the UI to see the traces at `http://localhost:16686/search`.


ghost/core/core/shared/OpenTelemetryKnexTracing.js
daniellockyer, line 1: Does `@opentelemetry/instrumentation-knex` not work in place of this?

PR #20157, opened 2024-05-07
Handled invalid timestamp format in filters

daniellockyer:
fix https://linear.app/tryghost/issue/SLO-85/fix-http-500-on-contentposts

- in the event we give the incorrect format in a filter, MySQL will throw an error and we'll throw a HTTP 500 error
- we can capture this error and return a more useful error to the user
- ideally we'd do this in a validation step before attempting the query, but parsing this out of NQL and detecting which columns are DATETIME could be quite tricky


sagzy (approved):
LGTM! 👍

PR #20172, opened 2024-05-08
Update Istanbul timezone

muratcorlu:
Turkish time zone is set to GMT+3[^1] since 2018[^2]. Current label indicates old state.

I'm not sure if I need to create a migration file for this change.

[^1]: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
[^2]: https://en.wikipedia.org/wiki/Time_in_Turkey


vikaspotluri123:
The file looks like test fixtures, I think you need to update https://github.com/TryGhost/SDK/tree/main/packages/timezone-data

muratcorlu:
Oh, ok. I couldn't notice that. Then I'll close this and open a PR on SDK project. Thank you @vikaspotluri123

muratcorlu:
Done: https://github.com/TryGhost/SDK/pull/495

muratcorlu:
Actually, updating this fixture file stille makes sense after fixing the SDK, right?

9larsons:
Thanks for the submission(s)! I merged the SDK package and I had to make other changes in https://github.com/TryGhost/Ghost/commit/c046bbda91761551e1f7a5942df2faa85c29f241 so this change was included there. Mentioned you in the commit 👍 .

PR #20197, opened 2024-05-13
Add explicit string-width resolution to fix ESM + CJS compat issue

### LINKED ISSUE #20065: Outdated version of string-width is causing the Docker image entrypoint loading to fail.
### Issue Summary

The update to fix arm32 architecture (https://github.com/TryGhost/Ghost/commit/c2485af853f7b3b86f7f141b8496439cdc4881b4) introduce a regression in the way the Ghost package is launched. When running `node index.js` there is a module resolution error similar to this issue https://github.com/nrwl/nx/issues/19249. I managed to solve the issue locally by adding `"string-width": "4.2.3"` to the `resolutions` section.

Here is the Docker test showing the error: https://github.com/docker-library/ghost/actions/runs/8756974313/job/24034646549

I would like to contribute this fix but I am not sure which `package.json` to use because the packaged one seems to be generated from the source ones.

### Steps to Reproduce

There is probably an more streamlined reproducible but building the Docker image here then running it (https://github.com/docker-library/ghost/tree/master/5/debian)

```
docker build -t ghost .
docker run --rm ghost 
```

### Ghost Version

5.82.3

### Node.js Version

18.*, 20.*

In Docker using the Ghost CLI

### Database type

MySQL 5.7

### Browser & OS version

_No response_

_No response_

### Code of Conduct

LaurentGoderre:
Fixes #20065

### DIFF
--- a/ghost/core/package.json
+++ b/ghost/core/package.json
@@ -263,7 +263,8 @@
     "@tryghost/errors": "1.3.2",
     "@tryghost/logging": "2.4.14",
     "moment": "2.24.0",
-    "moment-timezone": "0.5.23"
+    "moment-timezone": "0.5.23",
+    "string-width": "4.2.3"
   },
   "nx": {
     "targets": {
--- a/package.json
+++ b/package.json
@@ -47,7 +47,8 @@
     "@tryghost/logging": "2.4.14",
     "jackspeak": "2.1.1",
     "moment": "2.24.0",
-    "moment-timezone": "0.5.23"
+    "moment-timezone": "0.5.23",
+    "string-width": "4.2.3"
   },
   "renovate": {
     "extends": [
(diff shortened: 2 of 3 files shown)


daniellockyer (changes requested):
Won't this set all `string-width` to the same version? We have like 3 or 4 different majors in the codebase atm and I'm concerned there are breaking changes between them (I've not checked yet)

Also, you'll need to re-run `yarn` from the top-level in order to pick this up in `yarn.lock`

daniellockyer:
Hmm, it might also need adding to `ghost/core/package.json`. Perhaps this is why the previous fix didn't work 🤔

daniellockyer:
The commit wasn't shipped in v5.82.10. We'll need to wait until tomorrow's release for that

LaurentGoderre:
@daniellockyer really? Here's what I saw

Screenshot 2024-05-16 at 9 52 23 AM

daniellockyer:
@LaurentGoderre GitHub doesn't show the correct history flow. That release was done from a branch of v5.82.9 with one commit applied on top

PR #20200, opened 2024-05-13
Added experimental headers to allow caching members content

cmraible:
ref https://linear.app/tryghost/issue/KTLO-45/deploy-members-caching-solution-to-a-single-site-to-validate-and-test

Currently we only cache publicly available content. Any content that is accessed by a logged in member is only cached for that specific member based on their cookie. As a result, almost all requests from logged in members bypass our caching layer and reach Ghost, which adds unnecessary load to Ghost and its database.

This change adds experimental headers that allow our CDN to understand which tier to cache the content against, and securely tell the CDN which tier a logged in member has access to. With these changes, we can cache the member content against the tier, rather than the individual member, which should result in a higher cache HIT ratio and reduce the load on Ghost.

For requests to the frontend of the site, Ghost will set a custom `X-Member-Cache-Tier` header to the ID of the tier of the member who is accessing the content. This tells the CDN which tier to cache the content against.

For requests to either `/members/?token=...` endpoint (the magic link endpoint) or `/members/api/member`, Ghost will set a `ghost-access` and `ghost-access-hmac` cookie with the ID of the tier of the logged in member. With these two pieces of information, our CDN can serve cached content to logged in members.

These headers are experimental, and can only be enabled via Ghost's config. To enable these headers, set `cacheMembersContent:enabled` to `true` and provide an HMAC key in `cacheMembersContent:hmacSecret`.

### DIFF
--- a/ghost/core/core/frontend/web/middleware/frontend-caching.js
+++ b/ghost/core/core/frontend/web/middleware/frontend-caching.js
@@ -1,0 +1,77 @@
+/**
+ * @file Middleware to set the appropriate cache headers on the frontend
+ */
+const config = require('../../../shared/config');
+const shared = require('../../../server/web/shared');
+const {api} = require('../../services/proxy');
+
+/**
+ * Calculate the member's active tier.
+ * @param {object} member - The member object.
+ * @param {object} freeTier - The free tier object.
+ * @returns {string|null} - The member's active tier, or null if the member has more than one active subscription.
+ */
+function calculateMemberTier(member, freeTier) {
+    const activeSubscriptions = member.subscriptions.filter(sub => sub.status === 'active');
+    if (activeSubscriptions.length === 0) {
+        return freeTier;
+    } 
+    if (activeSubscriptions.length === 1) {
+        return activeSubscriptions[0].tier;
+    }
+    return null; // More than one active subscription
+}
+
+/**
+ * @typedef {function(): Promise<object>} GetFreeTier
+ */
+
+/**
+ * Returns the frontend caching middleware.
+ * @param {GetFreeTier} [getFreeTier] - Async function that takes no arguments and resolves to the free tier object.
+ * @returns {function} Middleware function.
+ */
+const getMiddleware = async (getFreeTier = async () => {
+    const {tiers} = await api.tiers.browse();
+    return tiers.find(tier => tier.type === 'free');
+}) => {
+    const freeTier = await getFreeTier();
+    /**
+     * Middleware to set cache headers based on site configuration and request properties.
+     * @param {import('express').Request} req
+     * @param {import('express').Response} res
+     * @param {import('express').NextFunction} next
+     */
+    function setFrontendCacheHeadersMiddleware(req, res, next) {
+        // Caching member's content is an experimental feature, enabled via config
+        const shouldCacheMembersContent = config.get('cacheMembersContent:enabled');
+        // CASE: Never cache if the blog is set to private
+        // CASE: Never cache if the request is made by a member and the site is not configured to cache members content
+        if (res.isPrivateBlog || (req.member && !shouldCacheMembersContent)) {
+            return shared.middleware.cacheControl('private')(req, res, next);
+        }
+
+        // CASE: Cache member's content if this feature is enabled
+        if (req.member && shouldCacheMembersContent) {
+            // Set the 'cache-control' header to 'public'
+            const memberTier = calculateMemberTier(req.member, freeTier);
+            if (!memberTier) {
+                // Member has more than one active subscription, don't cache the content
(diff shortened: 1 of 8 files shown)


cmraible:
Currently in a custom build on staging: 2898

Staging site: https://cache-members-content-test.ghost.is/ghost/ (owner user info in 1pass)
Staging site ID: [312694](https://octo.ghostfoundation.org/site/312694)

ghost/core/core/frontend/web/middleware/frontend-caching.js
```diff
@@ -0,0 +1,95 @@
...
+    } else {
+        // Member has more than one active subscription
+        // This is rare, but it can happen. 
+        // In this case, don't bother caching the content
+        return null;
+    }
+    return memberTier;
+}
+
+const getMiddleware = (freeTier) => {
```
cmraible, line 36: Feels like there should be a better way to do this 🤔

If the member has an active subscription, we can get their tier (and its object ID) from the `req.member` or `res.locals.member` object. 

But if the member is a 'free' member, there isn't any product or subscription to grab the tier from, so we just fallback to 'free'. Somehow we need to get the freeTier into the context of the middleware. Trying to find a way to inject this into the middleware so we can test it more easily — this is the best I've come up with so far but it feels messy.

cmraible: Found a somewhat cleaner way to do this, but I'm still not super satisfied with it. Will put more thought into it, but resolving for now.

ghost/core/core/server/services/members/middleware.js
```diff
@@ -11,12 +12,50 @@ const {
...
+        if (!req.member) {
+            const accessCookie = `ghost-access=null; Max-Age=0; Path=/; HttpOnly; SameSite=Strict;`;
+            const hmacCookie = `ghost-access-hmac=null; Max-Age=0; Path=/; HttpOnly; SameSite=Strict;`;
+            const existingCookies = res.getHeader('Set-Cookie') || [];
+            const cookiesToSet = [accessCookie, hmacCookie].concat(existingCookies);
+
+            res.setHeader('Set-Cookie', cookiesToSet);
+            return;
+        }
+        const activeSubscription = req.member.subscriptions?.find(sub => sub.status === 'active');
```
cmraible, line 41: This logic is in two places, which isn't ideal. I'm surprised we don't have an established pattern for "get member's active tier" — surely we'd need this in a lot of different places.

ghost/core/core/server/web/members/app.js
```diff
@@ -45,7 +45,14 @@ module.exports = function setupMembersApp() {
     membersApp.put('/api/member/newsletters', bodyParser.json({limit: '50mb'}), middleware.updateMemberNewsletters);
 
     // Get and update member data
-    membersApp.get('/api/member', middleware.getMemberData);
+    // Caching members content is an experimental feature
+    const shouldCacheMembersContent = config.get('cacheMembersContent:enabled');
```
cmraible, line 49: 🙈 Not pretty but this should work for testing purposes. Eventually we can simplify this and remove the conditional.

ghost/core/core/frontend/web/middleware/frontend-caching.js
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * @typedef {function(): Promise<object>} GetFreeTier
+ */
+
+/**
+ * Returns the frontend caching middleware.
+ * @param {GetFreeTier} [getFreeTier] - Async function that takes no arguments and resolves to the free tier object.
+ * @returns {function} Middleware function.
+ */
+const getMiddleware = async (getFreeTier = async () => {
```
cmraible, line 34: We need a way to get the ID of the free tier in the middleware, but for the sake of unit testing I've tried to do a bit of DI here. It works, and it makes testing easier, but I wonder if there is a better way to do this that I've overlooked. Feels a bit overly complicated for what it's doing.

ghost/core/core/server/web/members/app.js
```diff
@@ -45,7 +45,14 @@ module.exports = function setupMembersApp() {
...
+    if (shouldCacheMembersContent) {
+        membersApp.get('/api/member', middleware.loadMemberSession, middleware.accessInfoSession, middleware.getMemberData);
+    } else {
+        membersApp.get('/api/member', middleware.getMemberData);
+    }
```
cmraible, line 54: 🫣 Would like to refactor this if I have time — feels like this complexity should be _in_ the middleware rather than in the app.

ErisDS (commented):
I'm trying not to go too deep on this area of the codebase (frontend middleware stacks) cos it always sucks me in - IMO this is all fine for the prototype, but we should definitely come back to this wiring & try to find better patterns, if we proceed.

PR #20207, opened 2024-05-14
Added a column disallow list in the content API posts serializer

9larsons:
ref https://linear.app/tryghost/issue/CFR-29
- Removed the mobiledoc and lexical columns from the posts input serializer, meaning they will no longer be queried for.

Get helpers are essentially a gateway to the Content API. We already strip out the mobiledoc and lexical fields in the output serializer/returned response, but this means we're passing the mobiledoc and lexical fields back from the db. This is pointless and these fields are substantial in size - by far the largest fields in the whole ghost db - leading to slowed performance.

I've updated the input serializer to strip out the lexical and mobiledoc columns so we stop doing a `select *` with every query.

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/serializers/input/posts.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/input/posts.js
@@ -149,8 +149,8 @@ module.exports = {
          */
         if (localUtils.isContentAPI(frame)) {
             // CASE: the content api endpoint for posts should not return mobiledoc or lexical
-            removeSourceFormats(frame);
-            selectAllAllowedColumns(frame);
+            removeSourceFormats(frame); // remove from the format field
+            selectAllAllowedColumns(frame); // remove from any specified column or selectRaw options
 
             setDefaultOrder(frame);
             forceVisibilityColumn(frame);
--- a/ghost/core/core/server/api/endpoints/utils/serializers/input/posts.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/input/posts.js
@@ -19,25 +19,25 @@ const messages = {
 
 // NOTE: This doesn't stop them from being FETCHED, just returned in the response. This causes
 //   the output serializer to remove them from the data object before returning.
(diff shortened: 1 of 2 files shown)


9larsons:
I'm not sure of a better way to grab the column list. Knex appears to have something we may be able to use [here](https://github.com/knex/knex/blob/176151d8048b2a7feeb89a3d649a5580786d4f4e/lib/query/querybuilder.js#L1297).

daniellockyer:
>I'm not sure of a better way to grab the column list.

IMO this is the best way - the schema file is a cheap way to lookup our DB state

ghost/core/core/server/api/endpoints/utils/serializers/input/posts.js
```diff
@@ -24,6 +35,33 @@ function removeSourceFormats(frame) {
...
+ */
+function selectAllAllowedColumns(frame) {
+    if (!frame.options.columns && !frame.options.selectRaw) {
+        // Because we're returning columns directly from the table we need to remove info columns like @@UNIQUE_CONSTRAINTS@@
+        frame.options.selectRaw = _.keys(_.omit(postsSchema, ['lexical','mobiledoc','@@UNIQUE_CONSTRAINTS@@'])).join(',');
+    } else if (frame.options.columns) {
+        frame.options.columns = frame.options.columns.filter((column) => {
+            return !['mobiledoc', 'lexical'].includes(column);
+        });
+    } else if (frame.options.selectRaw) {
```
cmraible, line 56: I don't feel strongly about this but I wonder if we should strip these columns out if they're explicitly being asked for in `selectRaw`. I don't think we do this anywhere currently, so this is more of a maintenance concern than anything, but if we have some future use-case where we need to parse the lexical to generate some calculated field, I could see this causing me an hour or two of confusion when I can't get these fields returned no matter what I do.

Not a blocker at all, just a thought

cmraible (approved):
Overall looks good to me! Feels like a big win 🎉

PR #20212, opened 2024-05-15
SLO-115 [WIP] Removed unnecessary fields from select post query

vershwal:


vershwal:
We don't need this anymore!

PR #20213, opened 2024-05-15
Fix copy in newsletter preview content

peterzimon:
no ref.

The contents of the newsletter preview referenced settings on the left but in new Settings it's on the right.


JAHOIVAHTUTUNKHAMEN:
jahovah gambino jirah

PR #20231, opened 2024-05-20
WIP first pass on ActivityPub knex repositories

aileen:
ref MOM-123

This is 100% a WIP and not ready yet. Many questions need to be answered first before proceeding

---

🚧 **WIP** 🚧

### DIFF
--- a/ghost/ghost/src/core/activitypub/actor.entity.ts
+++ b/ghost/ghost/src/core/activitypub/actor.entity.ts
@@ -16,7 +16,9 @@ type ActorData = {
     outbox: Activity[];
     inbox: Activity[];
     following: {id: URI, username?: string;}[];
-    followers: {id: URI;}[];
+    followers: {id: URI;}[],
+    featured: {id: URI;}[],
+    internal: boolean;
 };
 
 type CreateActorData = ActorData & {
@@ -82,6 +84,14 @@ export class Actor extends Entity<ActorData> {
 
     get followers() {
         return this.attr.followers;
+    }
+
+    get featured() {
+        return this.attr.featured;
+    }
+
+    get publicKey() {
+        return this.attr.publicKey;
     }
 
     async sign(request: Request, baseUrl: URL): Promise<Request> {
@@ -256,7 +266,9 @@ export class Actor extends Entity<ActorData> {
             outbox: data.outbox || [],
             inbox: data.inbox || [],
             followers: data.followers || [],
-            following: data.following || []
+            following: data.following || [],
+            featured: data.featured || [],
+            internal: data.internal || false
         });
     }
 }
--- a/ghost/ghost/src/core/activitypub/actor.entity.ts
+++ b/ghost/ghost/src/core/activitypub/actor.entity.ts
@@ -19,10 +19,12 @@ type ActorData = {
     followers: {id: URI;}[],
     featured: {id: URI;}[],
     internal: boolean;
+    updatedAt?: Date;
 };
 
 type CreateActorData = ActorData & {
-    id? : ObjectID
+    id? : ObjectID,
+    createdAt: Date;
 };
 
 export class Actor extends Entity<ActorData> {
@@ -257,6 +259,9 @@ export class Actor extends Entity<ActorData> {
                 .toString();
         }
 
+        const createdAt = validateDate(data.createdAt);
+        const updatedAt = validateDate(data.updatedAt);
+
         return new Actor({
             id: data.id instanceof ObjectID ? data.id : undefined,
             username: data.username,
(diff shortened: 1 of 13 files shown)


ghost/ghost/src/core/activitypub/actor.entity.ts
```diff
@@ -16,7 +16,9 @@ type ActorData = {
     outbox: Activity[];
     inbox: Activity[];
     following: {id: URI, username?: string;}[];
-    followers: {id: URI;}[];
+    followers: {id: URI;}[],
```
aileen, line 19: @allouis I wasn't sure if we need `followers` here. There's some funky logic with `activities` that I haven't fully gotten yet, so maybe it's wrong to add it here?

ghost/ghost/src/db/knex/activity.repository.knex.ts
```diff
@@ -0,0 +1,53 @@
...
+import {ActivityRepository} from '../../core/activitypub/activity.repository';
+import {Activity} from '../../core/activitypub/activity.entity';
+
+export class ActivityRepositoryKnex implements ActivityRepository {
+    constructor(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        @Inject('knex') private readonly knex: any
+    ) {}
+
+    #modelToActivity(model: any) {
```
aileen, line 12: @allouis I don't know if that's needed, but I expect that the repository has to return an `Actor` type, so not sure how else to serialise it...

ghost/ghost/src/db/knex/actor.repository.knex.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    #defaultActor: Actor;
+
+    constructor(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        @Inject('knex') private readonly knex: any,
+        @Inject('SettingsCache') settingsCache: SettingsCache,
+        @Inject('DomainEvents') domainEvents: DomainEvents
+    ) {
+        this.#defaultActor = Actor.create({
```
aileen, line 22: Same as above, but I was able to `create` a new actor here. Again, not sure if that is the right approach.

ghost/ghost/src/db/knex/actor.repository.knex.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    async getOne(identifier: string | ObjectID) {
+        if (identifier instanceof ObjectID) {
+            return await this.getOneById(identifier);
+        } else {
+            return await this.getOneByUsername(identifier);
+        }
+    }
+
+    async save(actor: Actor) {
```
aileen, line 85: @allouis I tried adding a follow via the frontend, and I can see that the workflow for this is via adding it as an activity first, and then only the `acceptFollow` actually add the follow to the array of the entity. But this method is only called from the `postToInbox` method, which I never managed to actually call. I think, I need some more explanation here about how it's expected to work.

ghost/ghost/src/db/knex/actor.repository.knex.ts
aileen: Also not sure, if we need a default actor, or if this is really the right place to create them?

ghost/ghost/src/core/activitypub/actor.entity.ts
```diff
@@ -84,6 +86,18 @@ export class Actor extends Entity<ActorData> {
         return this.attr.followers;
     }
 
+    get featured() {
+        return this.attr.featured;
+    }
+
+    get privateKey() {
```
allouis, line 93: I don't think we want/need to expose this

ghost/ghost/src/db/knex/activity.repository.knex.ts
allouis: The Repository deffo has to return an `Activity` type - this should be giving TS errors 🤔 

You should be able to `return new Activity(data)` here, or if you need a factory, create a new factory method and do `return Activity.create(data)`

ghost/ghost/src/db/knex/activity.repository.knex.ts
```diff
@@ -0,0 +1,53 @@
...
+            .first();
+
+        if (!row) {
+            return null;
+        }
+        return this.#modelToActivity(row);
+    }
+
+    async save(activity: Activity) {
+        const date = this.knex.raw('CURRENT_TIMESTAMP');
```
allouis, line 35: We should be reading this from the entity, rather than creating it ourself

ghost/ghost/src/db/knex/actor.repository.knex.ts
allouis: So the flow is like:

1. Alice creates Follow Activity with Bob as the `object` and sends it to Bob's Inbox
2. Bob receives the Follow Activity, and creates an Accept Activity with the Follow Activity as the `object` and sends it to Alice's Inbox 

So when you're testing follow, you need to have ngrok setup, so that the site you're following is able to send the Accept back to you, which will then get handled in the `postToInbox` method.

ghost/ghost/src/db/knex/actor.repository.knex.ts
allouis: Fixtures is the correct place for defaults, but here is good for now, we can easily make changes to the values in dev

allouis (commented):
Few comments!

ghost/ghost/src/db/knex/actor.repository.knex.ts
allouis: Although - this is going to blow away the following list each time? I think need to be careful about that

aileen:
Re-opened against `activitypub` branch → https://github.com/TryGhost/Ghost/pull/20239

PR #20243, opened 2024-05-23
Prevent extraneous subscriptions

9larsons:
ref https://linear.app/tryghost/issue/ONC-38/
- Stripe checkout webhooks should only create members for products that exist in Ghost
- Ghost makes assumptions about the default product that do not stand up to Stripe having extraneous subscription products

We found that members were being created when customers used Stripe for subscriptions that are for products unrelated to Ghost (shared Stripe account). This problem is two-fold: 1) these members have not been to/know of the Ghost site and 2) the subscription caused the member to be linked to the default subscription while using the price of the paid for subscription, causing errors in MRR reporting.

### DIFF
--- a/ghost/core/test/e2e-api/admin/__snapshots__/members.test.js.snap
+++ b/ghost/core/test/e2e-api/admin/__snapshots__/members.test.js.snap
@@ -2932,6 +2932,68 @@ Object {
 }
 `;
 
+exports[`Members API Can filter by conversion attribution 1: [body] 1`] = `
+Object {
+  "members": Array [
+    Object {
+      "avatar_image": null,
+      "comped": false,
+      "created_at": StringMatching /\\\\d\\{4\\}-\\\\d\\{2\\}-\\\\d\\{2\\}T\\\\d\\{2\\}:\\\\d\\{2\\}:\\\\d\\{2\\}\\\\\\.000Z/,
+      "email": "member1@test.com",
+      "email_count": 0,
+      "email_open_rate": null,
+      "email_opened_count": 0,
+      "email_suppression": Object {
+        "info": null,
+        "suppressed": false,
+      },
+      "geolocation": null,
+      "id": StringMatching /\\[a-f0-9\\]\\{24\\}/,
+      "labels": Any<Array>,
+      "last_seen_at": null,
+      "name": "Mr Egg",
+      "newsletters": Array [
+        Object {
+          "description": null,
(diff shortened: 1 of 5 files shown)


9larsons:
@cmraible If you want to run tests, I found the following the easiest/cleanest:
- open dashboard.stripe.com and create a new account (it can be helpful to rename it)
- `stripe login` with Stripe CLI to link to the account you made
- get a clean db `yarn knex-migrator reset && yarn knex-migrator init`
- run ghost with `yarn dev --stripe`
- connect the fresh site to Stripe w/ test mode; you'll need to link to the Account you created, so don't skip the form. I previously had to link a ton of info to Stripe, including bank account which is obnoxious.. you may need to do the same. Unless I'm missing something, their test mode is a little finicky and requires full details before being able to use everything. *note: if you do this once in Stripe, when linking Stripe to Ghost you can copy over details; I cannot seem to find this option within Stripe when creating a new Account and trying to fill details they're asking for*
- make sure your Stripe account is toggled to Test Mode
- create a product in Stripe
- create a checkout link for the product
- complete the checkout (4242 4242 4242 4242 test card)

You can use this to repro the issue on `main`, then test w/ the changes using the same setup - don't need to redo everything.
-

cmraible:
@9larsons thanks for the repro instructions! I was able to reproduce this locally on `main`, and then I ran it on this branch and the member was _not_ created, so it seems to be working as expected for me. I got a bit wrapped up in other work today so haven't had time to really review the code or consider any other edge cases, but I can at least confirm that it's working for me too! I'll plan to take a more detailed look through this tomorrow AM.

cmraible (approved):
Looks good to me, more detailed note in the Linear issue.

ErisDS:
Closing this for now as we want to come at this from a different angle later.

imox2:
We’re hitting this in production too.

Ghost (paid tiers) + another product share the same Stripe account. When the other product gets a new subscription, Ghost still links it to the member if the Stripe customer exists — causing false “paid” members and wrong MRR.

Stripe events already include `price.id` / `product.id`, and Ghost’s own `products` table stores the blog’s price IDs. This makes it easy to ignore non-Ghost subscriptions (allowlist from DB or require a metadata flag).

**Suggestion:** Add a simple toggle in the Membership settings UI — *“Only listen for my Ghost products”* — so non-technical users can enable filtering without editing code.

Happy to help with code — backend filtering and even the UI implementation if needed. Would love to see this revisited.

PR #20251, opened 2024-05-26
Added bs (Bosnian) locale

amel-s:
Comprehensive, context aware, Bosnian translations for all available strings.

Informal addressing of users (as is widely accepted in Bosnia and Herzegovina).


amel-s:
Is there anything keeping this from getting merged? Any problems at all?

PR #20269, opened 2024-05-28
Used a ~hex~ base64 encoded string for hmac secret

allouis:
We want to use a randomly generated 64 byte secret for the hmac, and utf8 encoding isn't nice to work with for this, so we're going to use a hex string and decode it into a buffer for the secret.

### DIFF
--- a/ghost/core/core/server/services/members/middleware.js
+++ b/ghost/core/core/server/services/members/middleware.js
@@ -47,7 +47,7 @@ const setAccessCookies = function setAccessCookies(member = undefined, res, free
     if (!hmacSecret) {
         return;
     }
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/services/members/middleware.js
```diff
@@ -47,12 +47,16 @@ const setAccessCookies = function setAccessCookies(member = undefined, res, free
     if (!hmacSecret) {
         return;
     }
+    const hmacSecretBuffer = Buffer.from(hmacSecret, 'hex');
+    if (hmacSecretBuffer.length === 0) {
```
allouis, line 51: I believe non-hex strings will be parsed into a zero-sized buffer

PR #20295, opened 2024-05-30
Added `name` attribute to feature flag checkboxes

kevinansfield:
no issue

- added `name` prop to `<Toggle>` component that corresponds to the checkbox's `name` attribute
- updated `<FeatureToggle>` to set the `name` prop on it's `<Toggle>` component to the feature flag that was passed in so all feature toggles get a useful name attribute without having to manually specify one


kevinansfield:
@peterzimon I've merged because it will be useful to have in before the release tomorrow and it seems like a small change. Main reason I added you for review was because it touches the AdminX design system and wondered if there was anything else that needed to be taken into consideration.

PR #20320, opened 2024-06-04
Renamed `show_subhead` to `show_subtitle`

kevinansfield:
closes https://linear.app/tryghost/issue/MOM-194

- whilst working on the feature our naming changed from "Subhead" to "Subtitle"
- this rename of the newsletter design setting column brings naming back into a consistent state before public release


sagzy (approved):
LGTM 👍

mheland:
Client had Ghost 5.83.0 install on DO Droplet, upgrade fails, possibly related?
Which intermediate version is needed to get the `show_subtitle` field inserted into DB?

```
Unable to upgrade Ghost from v5.83.0 to v5.96.0. Would you like to revert back to v5.83.0?
Message: Ghost was able to start, but errored during boot with: ALTER TABLE `newsletters` RENAME COLUMN `show_subtitle` TO `show_subhead`; - Unknown column 'show_subtitle' in 'newsletters'
```

PR #20323, opened 2024-06-04
Added custom excerpt to post revisions

kevinansfield:
closes https://linear.app/tryghost/issue/MOM-170

When the subtitle field is included in the editor it creates a disconnect with post revisions if the underlying custom excerpt data is not included so we'd like to both preview and restore the subtitle when the in-editor subtitle field is enabled.

- added `post_revisions.custom_excerpt` column to schema
- added migration to add `post_revisions.custom_excerpt` to existing databases
- added migration to populate `post_revisions.custom_excerpt` with the current `post.custom_excerpt` value from the associated record
  - ensures no data is inadvertently lost when restoring an old version
  - using current data matches what would have happened previously where custom_excerpt was never overwritten when restoring an old version
- updated post revisions handling to accept the `custom_excerpt` field
- updated Admin's revision preview and restoration to display and set the `custom_excerpt` field

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.84/2024-06-05-08-42-34-populate-post-revisions-custom-excerpt.js
+++ b/ghost/core/core/server/data/migrations/versions/5.84/2024-06-05-08-42-34-populate-post-revisions-custom-excerpt.js
@@ -1,0 +1,19 @@
+const logging = require('@tryghost/logging');
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Populating post_revisions.custom_excerpt with post.excerpt');
+
+        await knex.raw(`
+            UPDATE post_revisions
+            JOIN posts ON post_revisions.post_id = posts.id
+            SET post_revisions.custom_excerpt = posts.custom_excerpt
+        `);
+
+        logging.info('Finished populating post_revisions.custom_excerpt');
+    },
+    async function down() {
+        // Not required
+    }
+);
--- a/ghost/core/core/server/data/migrations/versions/5.84/2024-06-05-08-42-34-populate-post-revisions-custom-excerpt.js
+++ b/ghost/core/core/server/data/migrations/versions/5.84/2024-06-05-08-42-34-populate-post-revisions-custom-excerpt.js
@@ -1,15 +1,28 @@
 const logging = require('@tryghost/logging');
 const {createTransactionalMigration} = require('../../utils');
+const DatabaseInfo = require('@tryghost/database-info');
 
 module.exports = createTransactionalMigration(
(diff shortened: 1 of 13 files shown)


ghost/core/core/server/data/schema/schema.js
```diff
@@ -413,7 +413,8 @@ module.exports = {
         reason: {type: 'string', maxlength: 50, nullable: true},
         feature_image: {type: 'string', maxlength: 2000, nullable: true},
         feature_image_alt: {type: 'string', maxlength: 191, nullable: true, validations: {isLength: {max: 125}}},
-        feature_image_caption: {type: 'text', maxlength: 65535, nullable: true}
+        feature_image_caption: {type: 'text', maxlength: 65535, nullable: true},
+        custom_excerpt: {type: 'string', maxlength: 2000, nullable: true, validations: {isLength: {max: 300}}}
```
mike182uk, line 417: Any reason to not just use `string` as the type? (`posts.custom_excerpt` uses `string`)

ghost/core/core/server/data/migrations/versions/5.84/2024-06-05-08-42-34-populate-post-revisions-custom-excerpt.js
```diff
@@ -0,0 +1,19 @@
...
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Populating post_revisions.custom_excerpt with post.excerpt');
+
+        await knex.raw(`
+            UPDATE post_revisions
+            JOIN posts ON post_revisions.post_id = posts.id
+            SET post_revisions.custom_excerpt = posts.custom_excerpt
```
mike182uk, line 11: Not sure if we are too bothered about efficiency here but I think adding a check on `posts.custom_excerpt` (not null / empty) means we could skip updating when theres nothing to update? (`WHERE posts.excerpt !== '' AND posts.excerpt IS NOT NULL`)

ghost/core/core/server/data/schema/schema.js
kevinansfield: Not too sure I'm following this. `custom_excerpt` has `type: 'string'` here too - it's copied directly from the posts table in the schema.

mike182uk: Ah my bad, i commented on the wrong line - It was [here](https://github.com/TryGhost/Ghost/pull/20323/files#diff-07ed1b3460b36fa1a09f53c5b03155fab9408cc106dc1f1cf72ba3bae965f6a7R4)

kevinansfield: Ohh, good catch, thanks!

PR #20332, opened 2024-06-05
Update classnames to use `excerpt` in emails

peterzimon:
MOM-203

We'll be keeping `excerpt` as the technical name for subtitles. To avoid confusion in the future this PR changes all classnames in the email template to use `excerpt` instead of `subtitle`.


peterzimon:
@kevinansfield has already fixed this in [this PR](https://github.com/TryGhost/Ghost/pull/20334)

PR #20337, opened 2024-06-05
Fixed orphaned words in newsletter titles

kevinansfield:
ref https://linear.app/tryghost/issue/MOM-191

- added `avoidOrphanedWords` helper that replaces the last space in the supplied text with `&nbsp;` so when the text wraps you're not left with a single word on the wrapped line
- updated email template to use the helper for the title and excerpt


ErisDS:
Cleaning up older PRs 💅

PR #20357, opened 2024-06-10
Added posts.updated_at index to schema

9larsons:
ref 6dbbdff
- added index to schema


9larsons:
Can ignore the migration checklist - it was pulled in from a stale branch, fixed via rebase.

PR #20361, opened 2024-06-11
Added Greek locale for comments, portal, signup-form, ghost.

iasonastr:
We have noticed some attempts at Greek translation with no follow-up from contributors. This is a complete translation with accurate file naming conventions. We are native speakers and professionals.


iasonastr:
any ideas why this failed?

PR #20389, opened 2024-06-13
Fixed race condition when updating member's `last_seen_at` timestamp

cmraible:
ref https://linear.app/tryghost/issue/ENG-1240/race-condition-when-updating-members-last-seen-at-timestamp
    
When members click a link in an email, Ghost updates the member's `last_seen_at` timestamp, but it should only update the timestamp if the member hasn't yet been seen in the current day (based on the publication's timezone).
    
Currently there is a race condition present where multiple simultaneous requests from the same member (if e.g. an email link checker is following all links in an email) can cause the `last_seen_at` timestamp to be updated multiple times in the same day for the same member. These additional queries add a significant load on Ghost and its database, which can contribute to the exhaustion of the connection pool and eventually requests may time out.
    
The primary motivation for this change is to avoid that race condition by adding a lock to the member row, checking if `last_seen_at` has already been updated in the current day, and only updating it if it hasn't.
    
Another beneficial side-effect of this change is that it avoids locking the `labels` and `newsletters` tables, which are locked when we update the `last_seen_at` timestamp in the `members` table currently. This should improve Ghost's ability to handle a large influx of requests to redirect endpoints (confirmed with load tests), which tend to happen immediately after a publisher sends an email.

### DIFF
--- a/ghost/core/core/server/services/members-events/index.js
+++ b/ghost/core/core/server/services/members-events/index.js
@@ -1,5 +1,6 @@
 const labsService = require('../../../shared/labs');
 const DomainEvents = require('@tryghost/domain-events');
+const events = require('../../lib/common/events');
 const settingsCache = require('../../../shared/settings-cache');
 const members = require('../members');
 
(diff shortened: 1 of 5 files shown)


cmraible:
So far, have managed to fix the race condition, but may have broken a few other things — cleaning things up and adding tests now.

PR #20396, opened 2024-06-17
Newsletter captions fix

dvdwinden:
Fixes https://linear.app/tryghost/issue/DES-4/image-caption-size-in-email-newsletter.

There were no styles defined for captions for cards beyond the featured image (bookmark, gallery, video), and we had no way of targeting those captions with CSS. They are now wrapped in a div with a specific class, which allows for more selective styling, and are styled similarly to the caption of the featured image.


ghost/email-service/lib/EmailRenderer.js
```diff
@@ -419,6 +419,10 @@ class EmailRenderer {
             return {src, width, height};
         });
 
+        // Add a class to each figcaption so we can style them in the email
+        $('figure, figcaption').each((i, elem) => !!($(elem).addClass('kg-card-figcaption')));
```
kevinansfield, line 423: Why is the class also added to the `<figure>` elements?

dvdwinden:
@kevinansfield Database tests are failing (related to snapshots?) and I'm unsure what to make of it. Could you have a peek?

PR #20427, opened 2024-06-20
Added time field to slow get helper logging

mike182uk:
refs [CFR-36](https://linear.app/tryghost/issue/CFR-36/pull-out-response-time-from-ghost-logs-message-field-for-get-helper)

Added time field to slow get helper logging to make it easier to query and filter on this value in elastic without having to parse the message field


ghost/core/core/frontend/helpers/get.js
```diff
@@ -357,7 +357,9 @@ module.exports = async function get(resource, options) {
                         apiOptions,
                         returnedRows: returnedRowsCount
                     }
-                }));
+                }), {
+                    time: totalMs
+                });
```
mike182uk, line 362: I've added this as a separate arg instead of adding it to the error as the error type does not allow it (and it does not make sense to add this field to the error type). The arg should get merged into the overall log when being processed (see [here](https://github.com/TryGhost/framework/blob/main/packages/logging/lib/GhostLogger.js#L502-L512))

gilmrjc:
This changes makes sets the time property, and that changes the time stamp in the STDOUT logs.

For example:

```
2025-03-03T06:50:58.612491147Z [1970-01-01 00:00:00] WARN {{#get}} helper took 231ms to complete
```

Because the query took 213ms, the field time is parsed as a few milliseconds after the epoc time.

PR #20437, opened 2024-06-20
Added composite index to posts table for type,status

9larsons:
ref https://linear.app/tryghost/issue/CFR-35
- performance improvement intended for the content api/get helpers

The posts table is shared by posts and pages and seldom is queried for both. It makes sense to add an index on type, and from the perspective of the content API, also on status as you're almost only ever querying for published posts or published pages.


9larsons:
Index is added in a few ms for a test set of 20k posts. This may be a bit longer for the couple biggest customers - I think the largest has ~125k posts, but all in all is very quick.

mike182uk (approved):

ghost/core/core/server/data/migrations/versions/5.86/2024-06-20-17-02-15-add-posts-type-status-index.js
daniellockyer, line 1: Version needs bumping

daniellockyer (approved):

PR #20470, opened 2024-06-26
Optimized stats aggregation code for Admin dashboard

daniellockyer:
fix https://linear.app/tryghost/issue/SLO-168/rangeerror-maximum-call-stack-size-exceeded

- this code takes the API output and reduces it down to collect together stats per date
- the current code is recursive, and we've seen errors with the recursion hitting a `RangeError: Maximum call stack size exceeded` error
- as well as that, we're doing a lot of array concat'ing and cloning, which burns memory and CPU time
- instead, we can just use `.reduce`
- the new implementation is much faster than the existing one (1ms vs 85ms) and uses no recursion, so those errors should go away
- I've also verified that the output is the same between the two functions

This should prevent long recursion chains like this:
CleanShot 2024-06-26 at 14 28 01@2x

### DIFF
--- a/ghost/admin/app/utils/merge-stats-by-date.js
+++ b/ghost/admin/app/utils/merge-stats-by-date.js
@@ -1,0 +1,31 @@
+export default function mergeDates(list, entry) {
+    const [current, ...rest] = list;
+
+    if (!current) {
+        return entry ? [entry] : [];
+    }
+
+    if (!entry) {
+        return mergeDates(rest, {
+            date: current.date,
+            count: current.count,
+            positiveDelta: current.positive_delta,
+            negativeDelta: current.negative_delta,
+            signups: current.signups,
+            cancellations: current.cancellations
+        });
+    }
+
+    if (current.date === entry.date) {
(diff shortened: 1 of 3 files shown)


ghost/admin/tests/unit/utils/merge-stats-by-date.js
ErisDS, line 1: I think this file name is meant to have -test on the end of it? No idea how much it matters tho

ghost/admin/app/utils/merge-stats-by-date.js
```diff
@@ -0,0 +1,26 @@
+export default function mergeDates(list) {
```
sagzy, line 1: We may want to name this `mergeStatsByDate` too?

sagzy (approved):
LGTM! Nice fix 👍

ghost/admin/app/utils/merge-stats-by-date.js
daniellockyer: Yep! Fixed

PR #20476, opened 2024-06-27
🐛 Fixed incorrect member subscription details in Admin

allouis:
fixes https://linear.app/tryghost/issue/ENG-642

- When a subscription is in the `canceled` state the corresponding Member has no access to the Ghost site. The only time a Member will continue to have access if their subscription is due to cancel at the period end is if it is still in an active state, which is one of `active` `trialing` `unpaid` or `past_due`
- When a subscription is canceled immediately (i.e. before the end of the current billing period), we now render "Ended" without a date, because we don't store the cancellation date in the subscription object. We previously used "Ended {current_period_end}" which would sometimes lead to dates in the future
- Bonus: refactored code and added unit tests

### DIFF
--- a/ghost/admin/app/components/gh-member-settings-form.hbs
+++ b/ghost/admin/app/components/gh-member-settings-form.hbs
@@ -147,21 +147,7 @@
                                             {{/if}}
                                         {{/if}}
 
-                                        {{#if sub.isComplimentary}}
-                                            {{#if sub.compExpiry}}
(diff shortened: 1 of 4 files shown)


allouis:
@ronaldlangeveld This might be useful - I haven't tested this, and ideally we would use different properties here, e.g. `hasAccess` and switch on that instead - but I believe the fix is along these lines.

PR #20507, opened 2024-07-02
Added innodb_force_recovery config to local mysql

allouis:
ref https://dev.mysql.com/doc/refman/8.4/en/forcing-innodb-recovery.html

Docker on MacOS can sometimes force quit the container and we end up with a broken MySQL container which can't boot. This config forced MySQL to attempt to recover the DB. The value of 6 is suggested by the logs when MySQL breaks - but this will open InnoDB in a readonly mode, which isn't suitable for us.

I think we wanna have a way to run this just once, and then exit, maybe with like `yarn recover-db`


ErisDS:
Closing as we determined this wasn't the right solution. Not sure if anyone has had this problem since, but as we move the dev env more towards mysql we'll soon know if it surfaces again 😬

PR #20515, opened 2024-07-02
Added custom redirects ReDoS validation

mike182uk:
refs [ENG-709](https://linear.app/tryghost/issue/ENG-709/%F0%9F%90%9B-bad-redirects-causing-container-tear-down)

Added validation to prevent RegEx's susceptible to ReDoS from being used with custom redirects. Also moved error details out of `context` and into `errorDetails` to be consistent with error logging elsewhere as well as fix issue in admin-x where blank screen would be shown when an error occurred during redirects upload (due to logic not accounting for `context` being an object)

### DIFF
--- a/ghost/core/package.json
+++ b/ghost/core/package.json
@@ -219,7 +219,7 @@
     "node-jose": "2.2.0",
     "path-match": "1.2.4",
     "probe-image-size": "7.2.3",
-    "redos-detector": "^5.1.0",
+    "redos-detector": "5.1.0",
     "rss": "1.2.2",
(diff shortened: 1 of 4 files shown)


ghost/core/package.json
```diff
@@ -219,6 +219,7 @@
     "node-jose": "2.2.0",
     "path-match": "1.2.4",
     "probe-image-size": "7.2.3",
+    "redos-detector": "^5.1.0",
```
daniellockyer, line 222: Can you pin this dependency?

PR #20534, opened 2024-07-03
🎨 Added staff notification when a sub is canceled due to failed payments

sagzy:
ref https://linear.app/tryghost/issue/ENG-1254

- when a subscription is canceled automatically by Stripe (e.g. due to multiple failed payments), we now send a staff notification
- logic before: if a member cancels a sub in Portal, then send a staff notification
- logic now: if a subscription was active, but is now set to cancel immediately or at the end of the billing period, then send a staff notification.
- with that logic change, we now send a cancellation staff notification when:
    1. A member cancels their sub in Portal (existing)
    2. A staff member cancels a member sub in Stripe (new)
    3. A staff member cancels a member sub in Admin (new)
    4. A sub is canceled automatically by Stripe because of multiple failed payments (new)
- the copy of the staff notification email has also been updated to take into account 1) manual vs automatic cancellations, and 2) immediate vs end of billing period cancellations


sagzy:
Pending copy changes on the email notification (discussion with Peter [here](https://linear.app/tryghost/issue/ENG-1254/subscription-cancelation-due-to-failed-payment-should-send-a-staff#comment-0f34be2d))

PR #20546, opened 2024-07-04
Removed `urlCache` flag and code

daniellockyer:
- this was an experimental feature to persist the URL cache to disk so it can be read upon boot, which would save recalculating it and hopefully speed up boot times
- it was never fleshed out and the code here is a bit of a hack, so it's not really worth keeping it around
- worst case, we can revert this if we ever decide to move forwards with it


daniellockyer:
Closing as this code might actually be useful in speeding up tests for now...

PR #20570, opened 2024-07-09
Updated timezone packages

9larsons:
ref https://linear.app/tryghost/issue/ENG-1266
- Mexico changed tz to not participate in DST
- our package was a couple years behind, so we likely have fixes for other countries/regions, too


9larsons:
This didn't requiring bumping our installed versions of `moment`, even though the tz package's dependency version went up.

PR #20585, opened 2024-07-10
🔒 Anonymised the name of edited images

mike182uk:
refs [ENG-1260](https://linear.app/tryghost/issue/ENG-1260/🔒-redacting-pictures-in-pintura-leaves-easily-findable-original-image)

### DIFF
--- a/apps/admin-x-framework/src/api/images.ts
+++ b/apps/admin-x-framework/src/api/images.ts
@@ -7,13 +7,20 @@ export interface ImagesResponseType {
     }[];
 }
 
-export const useUploadImage = createMutation<ImagesResponseType, {file: File}>({
+// eslint-disable-next-line no-shadow
+export enum ImageStatus {
+    NEW = 'new',
+    EDITED = 'edited',
+}
+
+export const useUploadImage = createMutation<ImagesResponseType, {file: File, status?: ImageStatus}>({
     method: 'POST',
     path: () => '/images/upload/',
-    body: ({file}) => {
+    body: ({file, status = ImageStatus.NEW}) => {
         const formData = new FormData();
         formData.append('file', file);
         formData.append('purpose', 'image');
+        formData.append('status', status);
         return formData;
     }
 });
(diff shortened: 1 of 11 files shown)


vikaspotluri123:
Thinking from a power-user perspective, would it be possible to retain the file name, and potentially edit number? Without it, if you edit a bunch of images, you'll end up with several randomly named images without context.

e.g. `ghost-logo-{uuid}.png` or `ghost-logo-{edit_num}-{uuid}.png`

This would require having a static path for edited images (e.g. `content/images/edited`), or adding a uuid to all uploaded content.

mike182uk:
Hey @vikaspotluri123 👋 

This change is a bit of a compromise to make the image editing feature a little safer, without re-architecting how the image uploading in Ghost works. The particular use case for this is when the redact feature of the editor has been used, anonymising the name of the edited image prevents the ability to remove the numbered prefix from the image path and access the original image

vikaspotluri123:
@mike182uk understood, that's why I suggested this 😁 

> This would require having a static path for edited images (e.g. content/images/edited), or adding a uuid to all uploaded content.

mike182uk:
Not sure I'm following the power-user perspective, is this if you are self hosting and for some reason want to manually modify the contents of `content/images` on the filesystem? If so, whats the reason for needing to do this?

> would it be possible to retain the file name, and potentially edit number

Unfortuantely, based on the current implementation we wouldn't know what the current revision was to be able to include that

As for retaining the name, we would probably need to do something similar to what you are mentioning - append `uuid` on initial upload, then subsequently update this `uuid` on every revision so that you couldn't easily guess the path to the original. This would be a bigger change in Ghost with how we name incoming uploads, and would not work retroactively (i.e if i edited an existing image after this change, I wouldn't benefit from this change as the image wasn't uploaded with the `uuid`)

vikaspotluri123:
> Not sure I'm following the power-user perspective, is this if you are self hosting and for some reason want to manually modify the contents of content/images on the filesystem? If so, whats the reason for needing to do this?

It can be if I'm self-hosting, but also if I request my data from support - without having the file name for the image, if I'm looking for a specific edit (e.g. I redacted too much, go back to the previous edit) I would have to manually open every edited file to see if it's the edit you're looking for. If the file name is included in the image, the number of files to review is much smaller

> we wouldn't know what the current revision was to be able to include that

Haha, that was my guess but I figured I would include it anyway 😁 

> and would not work retroactively 

Yeah, admin would have to add extra logic to detect the older files and tell the API not to include the file name

PR #20589, opened 2024-07-10
🐛 Fixed image dimension retrieval causing Ghost requests to hang

9larsons:
ref https://linear.app/tryghost/issue/ENG-1408/
- added additional safeguards to the image size dimensions probing

For some reason that requires further investigation, the probe-image-size package was silently failing (neither resolving nor rejecting) for a particular URL. This was causing Ghost to hang on to serving the request, and after a few of these came in, ultimately caused Ghost to stop being responsive.

Rather than trying to patch a dependency, we'll wrap the call to this package and use the same timeout we pass into the package (which is ignored in this particular case) as an additional safeguard.


ErisDS (approved):
🙏 This looks great

PR #20646, opened 2024-07-23
✨ Improved performance loading posts & pages in admin

9larsons:
ref 8ea1dfb
ref https://linear.app/tryghost/issue/ONC-111

* undid the reversion for the performance improvements
* built upon new tests for the posts list functionality in admin, including right click actions
* added tests for pages view in Admin

This was reverted because it broke the Pages list view in Admin, which is a thin extension of the Posts functionality in admin (route & controller). That has been fixed and tests added.

This was originally reverted because the changes to improve loading response times broke right click (bulk) actions in the posts list. This was not caught because it turned out we had near-zero test coverage of that part of the codebase. Test coverage has been expanded for the posts list, and while not comprehensive, is a much better place for us to be in.


ghost/admin/app/controllers/posts.js
```diff
@@ -85,11 +85,8 @@ export default class PostsController extends Controller {
...
     }
 
-    get postsInfinityModel() {
-        return this.model;
-    }
-
     get totalPosts() {
+        // TODO: this might be a problem for the new infinity models
         return this.model.meta?.pagination?.total ?? 0;
     }
```
kevinansfield, line 91: This doesn't appear to be used anywhere, is it worth deleting it rather than having a comment that suggests it won't work?

9larsons: Happy to remove it - thanks for taking a pass!

PR #20647, opened 2024-07-24
Backfilled missing offer redemptions

cmraible:
ref https://linear.app/tryghost/issue/ENG-1440/backfill-offer-redemption-data-with-a-migration
    
There was a bug that caused offer redemptions to not be recorded in the database for some subscriptions that were created with an offer. However, we still have the `offer_id` attached to the subscriptions, so we are able to backfill the missing redemptions. The bug was fixed in https://github.com/TryGhost/Ghost/commit/bf895e6e99e6fe8e90f0d582fa6e38d5c4e37f5d
    
This commit only contains a migration, which queries for subscriptions that have an `offer_id` but do not have any offer redemptions recorded, and adds any missing redemptions to the `offer_redemptions` table.

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
+++ b/ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
@@ -1,0 +1,64 @@
+// For information on writing migrations, see https://www.notion.so/ghost/Database-migrations-eb5b78c435d741d2b34a582d57c24253
+
+const logging = require('@tryghost/logging');
+const DatabaseInfo = require('@tryghost/database-info');
+const {default: ObjectID} = require('bson-objectid');
+
+// For DML - data changes
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        // Backfill missing offer redemptions
+        try {
+            // Select all subscriptions that have an `offer_id` but don't have a matching row in the `offer_redemptions` table
+            logging.info('Selecting subscriptions with missing offer redemptions');
+            const result = await knex.raw(`
+                SELECT
+                    mscs.id AS subscription_id,
+                    mscs.offer_id,
+                    mscs.start_date AS created_at,
+                    m.id AS member_id
+                FROM
+                    members_stripe_customers_subscriptions mscs
+                        LEFT JOIN
+                    offer_redemptions r ON r.subscription_id = mscs.id
+                        INNER JOIN 
+                    members_stripe_customers msc ON mscs.customer_id = msc.customer_id
+                        INNER JOIN
+                    members m ON msc.member_id = m.id
+                WHERE
+                    mscs.offer_id IS NOT NULL and r.id IS NULL;
+            `);
+
+            // knex.raw() returns a different result depending on the database. We need to handle either case
+            let rows = [];
+            if (DatabaseInfo.isSQLite(knex)) {
+                rows = result;
+            } else {
+                rows = result[0];
+            }
+
+            // Do the backfil
+            if (rows && rows.length > 0) {
+                logging.info(`Backfilling ${rows.length} offer redemptions`);
+                // Generate IDs for each row
+                const offerRedemptions = rows.map((row) => {
+                    return {
+                        id: (new ObjectID()).toHexString(),
+                        ...row
+                    };
+                });
+                // Batch insert rows into the offer_redemptions table
+                await knex.batchInsert('offer_redemptions', offerRedemptions, 1000);
+            } else {
+                logging.info('No offer redemptions to backfill');
+            }
+        } catch (error) {
+            logging.error(`Error backfilling offer redemptions: ${error.message}`);
+        }
+    },
+    async function down() {
+        // We don't want to un-backfill data, so do nothing here.
+    }
+);


ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
```diff
@@ -0,0 +1,62 @@
...
+            `);
+            // knex.raw() returns a different result depending on the database. We need to handle either case
+            let rows = [];
+            // CASE: MySQL returns an array with two elements: the first element is an array of rows, the second element is metadata
+            if (result && result.length > 0 && Array.isArray(result[0])) {
+                rows = result[0];
+            // CASE: SQLite returns an array of rows
+            } else if (result && result.length > 0 && Object.prototype.hasOwnProperty.call(result[0], 'subscription_id')) {
+                rows = result;
+            }
```
cmraible, line 40: There's probably a prettier way to do this by checking the `knex.client` information, but this way doesn't make any assumptions about specific clients. It works, but I'm open to suggestions on a cleaner way to do this.

ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
```diff
@@ -0,0 +1,62 @@
...
+            if (rows && rows.length > 0) {
+                logging.info(`Backfilling ${rows.length} offer redemptions`);
+                // Generate IDs for each row
+                const offerRedemptions = rows.map((row) => {
```
ErisDS, line 44: @daniellockyer as someone who has seen a lot of these go by, do you have any concerns with this loop?

ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
cmraible: Oh awesome, thanks! Will update that shortly...

ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
```diff
@@ -0,0 +1,62 @@
...
+            logging.info('Selecting subscriptions with missing offer redemptions');
+            const result = await knex.raw(`
+                SELECT
+                    mscs.id AS subscription_id,
+                    mscs.offer_id,
+                    mscs.start_date AS created_at,
+                    m.id AS member_id
+                FROM
+                    members_stripe_customers_subscriptions mscs
+                        JOIN
```
9larsons, line 23: No changes needed, simply a note that I personally find it helpful to declare INNER JOIN.

ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
```diff
@@ -0,0 +1,62 @@
...
+                    mscs.offer_id,
+                    mscs.start_date AS created_at,
+                    m.id AS member_id
+                FROM
+                    members_stripe_customers_subscriptions mscs
+                        JOIN 
+                    members_stripe_customers msc ON mscs.customer_id = msc.customer_id
+                        JOIN
+                    members m ON msc.member_id = m.id
+                        LEFT JOIN
```
9larsons, line 27: Even though I think the query optimizer handles this, moving this join first makes sense as it is the primary and only join to filter.

ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
9larsons: I wish we had a good place to add a checkpoint for things like this. The GH migration checklist doesn't seem right as it could be noisy. Maybe we could consider adding a section like 'platform considerations'? This difference in the return for sqlite vs mysql seems like a big (common) one that we'd want to help ensure we don't miss.

ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
cmraible: Ah, nice one — TIL! 🙏🏼

ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
cmraible: Yeah — what do you think about a linting rule to warn you when using `knex.raw()`? There may be other sqlite vs mysql nuances for knex, but this seems like one that would be somewhat common. Not sure if we could detect if you've handled it correctly, but it could at least warn you and point you to `Databaseinfo.isSqlite`, and require the developer to manually acknowledge it by disabling the rule for that line?

cmraible: For now, I've [added a check to the migrations checklist](https://github.com/TryGhost/Ghost/commit/f147167a291a635f041c208da81cbc3aff5bb1b2) to at least ensure we test migrations in both DBs. This was already in the notion doc, but not in the checklist that's added to PRs with migrations.

ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
cmraible: Updated!

ghost/core/core/server/data/migrations/versions/5.89/2024-07-30-19-51-06-backfill-offer-redemptions.js
cmraible: Just for my own due diligence, I did a search and found this is a pretty common pattern in migrations, specifically for generating object IDs. I checked a couple of them, and did a search in Slack for their rollout and haven't found any drama.

Also tested on staging with 40k+ redemptions to backfill (max in production will be ~4k) and it completed in <3 seconds

PR #20670, opened 2024-07-26
🐛 Fixed Content-Type for RSS feed

### LINKED ISSUE #20634: The RSS feed endpoint returns a wrong content type
### Issue Summary

When requesting an RSS feed from a ghost blog (e.g. https://activitypub.ghost.org/rss/), the returned content-type is `content-type: text/xml; charset=utf-8` (as you can see in [this file](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/frontend/services/rss/renderer.js)).

This can result in the feed not being correctly displayed:
|How it is displayed|How it should be displayed (with an Add-on for RSS feeds)|
|-------------------|---------------------------|
|grafik|grafik|

As far as I can see, it should be `application/rss+xml`. (Sources: [[1]](https://datatracker.ietf.org/doc/id/draft-nottingham-rss-media-type-00), [[2]](https://stackoverflow.com/questions/595616/what-is-the-correct-mime-type-to-use-for-an-rss-feed)).

### Steps to Reproduce

1. Open the RSS feed URL from any ghost blog (e.g. https://activitypub.ghost.org/rss/)
2. Open the developer tools from the browser, look into the network tab.
3. (Sometimes required:) Refresh the site
4. Click on the request of the main request.
5. See in the Response Headers: `content-type: text/xml; charset=utf-8`

### Ghost Version

5.87.3

### Node.js Version

(I don't see that with docker?)

Using a docker container on a debian 12 server.

### Database type

MySQL 8

### Browser & OS version

Firefox 128.0 (64-Bit) on Windows 11 Pro

_No response_

### Code of Conduct

SteffoSpieler:
This sets the correct Content-Type (`application/rss+xml; charset=utf-8`) for the RSS Endpoint as mentioned in the issue #20634 

I hope I did everything correct here - when running `yarn dev`, it definitely sends the correct Content-Type now.

Closes #20634

### DIFF
--- a/ghost/core/core/frontend/services/rss/renderer.js
+++ b/ghost/core/core/frontend/services/rss/renderer.js
@@ -9,7 +9,7 @@ module.exports.render = function render(res, baseUrl, data) {
     return rssCache
         .getXML(baseUrl, rssData)
         .then(function then(feedXml) {
-            res.set('Content-Type', 'text/xml; charset=UTF-8');
+            res.set('Content-Type', 'application/rss+xml; charset=UTF-8');
             res.send(feedXml);
         });
 };
--- a/ghost/core/test/e2e-frontend/default_routes.test.js
+++ b/ghost/core/test/e2e-frontend/default_routes.test.js
@@ -271,7 +271,7 @@ describe('Default Frontend routing', function () {
             await request.get('/rss/')
                 .expect(200)
                 .expect('Cache-Control', testUtils.cacheRules.public)
-                .expect('Content-Type', 'text/xml; charset=utf-8')
+                .expect('Content-Type', 'application/rss+xml; charset=utf-8')
                 .expect(assertCorrectFrontendHeaders)
                 .expect((res) => {
                     res.text.should.match(/<!\[CDATA\[Start here for a quick overview of everything you need to know\]\]>/);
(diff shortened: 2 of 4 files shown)


SteffoSpieler:
```
   1) Frontend behavior tests
       extended routes.yaml (5): rss override
         serve /podcast/rss/:

      AssertionError: expected 'application/rss+xml' to equal 'application/rss+xml; charset=utf-8'
      + expected - actual

      -application/rss+xml
      +application/rss+xml; charset=utf-8
      
      at Assertion.fail (/home/runner/work/Ghost/Ghost/node_modules/should/cjs/should.js:275:17)
      at Assertion.value (/home/runner/work/Ghost/Ghost/node_modules/should/cjs/should.js:356:19)
      at /home/runner/work/Ghost/Ghost/ghost/core/test/regression/mock-express-style/api-vs-frontend.test.js:1563:61
```

Interesting. I have never changed anything regarding the `charset=utf-8` part.
Should I just update the test to check for `application/rss+xml`, or how should this be fixed?

This would be the "easiest" way:
```diff
# File ghost/core/test/regression/mock-express-style/api-vs-frontend.test.js
# Line 1559

            return localUtils.mockExpress.invoke(app, req)
                .then(function (response) {
                    response.statusCode.should.eql(200);
                    response.template.should.eql('podcast/rss');
-                    response.headers['content-type'].should.eql('application/rss+xml; charset=utf-8');
+                    response.headers['content-type'].should.eql('application/rss+xml');
                    response.body.match(/<link>/g).length.should.eql(2);
                });
```

daniellockyer:
Hi @SteffoSpieler! Thanks for the PR 🙏🏻 This should be available in the next scheduled Ghost release (Friday)

PR #20687, opened 2024-07-30
🐛 Fixed unsaved changes confirmation on Lexical schema change

ronaldlangeveld:
refs ENG-661

Fixes a long-standing issue where an outdated Lexical schema in the database triggered the unsaved changes confirmation dialog incorrectly. 
Implemented a secondary hidden Lexical instance that loads the state from the database, renders it, and uses this updated state to compare with the live editor's scratch. 
This ensures the unsaved changes prompt appears only when there are real changes from the user.

### DIFF
--- a/ghost/admin/app/components/koenig-lexical-editor.js
+++ b/ghost/admin/app/components/koenig-lexical-editor.js
@@ -682,7 +682,7 @@ export default class KoenigLexicalEditor extends Component {
                         cardConfig={cardConfig}
                         enableMultiplayer={enableMultiplayer}
                         fileUploader={{useFileUpload, fileTypes}}
-                        initialEditorState={this.args.lexical}
+                        initialEditorState={isInitInstance ? this.args.initLexicalState || this.args.lexical : this.args.lexical}
                         multiplayerUsername={multiplayerUsername}
                         multiplayerDocId={multiplayerDocId}
                         multiplayerEndpoint={multiplayerEndpoint}
--- a/ghost/admin/app/components/koenig-lexical-editor.js
+++ b/ghost/admin/app/components/koenig-lexical-editor.js
@@ -670,12 +670,6 @@ export default class KoenigLexicalEditor extends Component {
         const multiplayerUsername = this.session.user.name;
 
         const KGEditorComponent = ({isInitInstance}) => {
-            const handleInitInstance = (data) => {
-                if (this.args.initLexical && isInitInstance) {
-                    this.args.initLexical(data);
-                }
-            };
-
             return (
                 <div style={isInitInstance ? {visibility: 'hidden', position: 'absolute'} : {}}>
                     <KoenigComposer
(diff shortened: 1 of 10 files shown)


ghost/admin/app/components/koenig-lexical-editor.js
```diff
@@ -669,34 +669,53 @@ export default class KoenigLexicalEditor extends Component {
...
+                        darkMode={this.feature.nightShift}
+                        isTKEnabled={true}
+                    >
+                        <KoenigEditor
+                            editorResource={this.editorResource}
+                            cursorDidExitAtTop={this.args.cursorDidExitAtTop}
+                            placeholderText={this.args.placeholder}
+                            darkMode={this.feature.nightShift}
+                            onChange={isInitInstance ? e => handleInitInstance(e) : this.args.onChange}
+                            registerAPI={this.args.registerAPI}
```
kevinansfield, line 702: We'll want to make sure this doesn't get registered for the secondary instance otherwise we won't have a stable reference to the main editor

ghost/admin/app/components/koenig-lexical-editor.js
```diff
@@ -669,34 +669,53 @@ export default class KoenigLexicalEditor extends Component {
...
+                        />
+                        <WordCountPlugin editorResource={this.editorResource} onChange={this.args.updateWordCount} />
+                        <TKCountPlugin editorResource={this.editorResource} onChange={this.args.updatePostTkCount} />
```
kevinansfield, line 705: Same for both of these, we don't want the secondary instance interacting with any of the outside world aside from the main content onChange handler

ghost/admin/app/components/koenig-lexical-editor.js
```diff
@@ -669,34 +669,53 @@ export default class KoenigLexicalEditor extends Component {
         const multiplayerDocId = cardConfig.post.id;
         const multiplayerUsername = this.session.user.name;
 
+        const [hasInitialised, setHasInitialised] = React.useState(false);
+
+        const KGEditorComponent = ({isInitInstance}) => {
+            const handleInitInstance = (data) => {
+                if (this.args.initLexical && isInitInstance && !hasInitialised) {
+                    this.args.initLexical(data);
+                    setHasInitialised(true);
```
kevinansfield, line 678: Why do we need `hasInitialised` here? One of the issues Steve ran into is that the changes made when opening the editor come in batches so we can't guarantee that only the first `onChange` call is the one we want - if it was then we could do the same thing without having a secondary instance at all

ghost/admin/app/controllers/lexical-editor.js
```diff
@@ -1221,61 +1231,53 @@ export default class LexicalEditorController extends Controller {
...
-        // and we should consider the post still dirty.
+        // If the Adapter failed to save the post, isError will be true, and we should consider the post still dirty.
         if (post.get('isError')) {
             this._leaveModalReason = {reason: 'isError', context: post.errors.messages};
             return true;
         }
 
-        // post.tags is an array so hasDirtyAttributes doesn't pick up
-        // changes unless the array ref is changed
+        // Post tags comparison
```
kevinansfield, line 1247: Can we switch back to the comments that explain _why_ this is necessary rather than a comment that just tells you what the code is doing?

kevinansfield (approved):
Added a minor thing about code comments being replaced with less-useful comments but otherwise the functionality all looks good! 🙌

PR #20688, opened 2024-07-30
Fixed handling of single item collections

allouis:
Fedify will not use an array for the `items` key of collections when there is only a single item, which wasn't being handled in our activitypub api module.

Now we always return an array so that the components recieve consistent data.

### DIFF
--- a/apps/admin-x-activitypub/src/api/activitypub.test.ts
+++ b/apps/admin-x-activitypub/src/api/activitypub.test.ts
@@ -95,7 +95,7 @@ describe('ActivityPubAPI', function () {
             expect(actual).toEqual(expected);
         });
 
-        test('Returns an the items array when the inbox is not empty', async function () {
+        test('Returns all the items array when the inbox is not empty', async function () {
             const fakeFetch = Fetch({
                 'https://auth.api/': {
                     response: JSONResponse({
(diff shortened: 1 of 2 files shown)


apps/admin-x-activitypub/src/api/activitypub.test.ts
```diff
@@ -235,6 +278,43 @@ describe('ActivityPubAPI', function () {
...
+
+            const api = new ActivityPubAPI(
+                new URL('https://activitypub.api'),
+                new URL('https://auth.api'),
+                'index',
+                fakeFetch
+            );
+
+            const actual = await api.getFollowing();
+            const expected: Activity[] = [
```
mike182uk, line 310: Is this the correct type here?

mike182uk: Ah looked up the type, its an alias for `any` 😄 I'm assuming we have future plans to refactor this?

mike182uk (approved):

apps/admin-x-activitypub/src/api/activitypub.test.ts
allouis: Yes definitely! The main thing for these changes was to get the network out of the components so we can test

PR #20697, opened 2024-07-31
Fixed followers / following not displaying correctly in the activitypub app

mike182uk:
no refs

In a recent upgrade to fedify, the name of the key that represents the items in a collection was renamed from `items` to `orderedItems`. This change updates the API queries to use the updated key name

Related: https://github.com/TryGhost/Ghost/pull/20688

### DIFF
--- a/apps/admin-x-activitypub/src/api/activitypub.test.ts
+++ b/apps/admin-x-activitypub/src/api/activitypub.test.ts
@@ -225,7 +225,7 @@ describe('ActivityPubAPI', function () {
                 'https://activitypub.api/.ghost/activitypub/following/index': {
                     response: JSONResponse({
                         type: 'Collection',
-                        items: []
+                        orderedItems: []
                     })
                 }
             });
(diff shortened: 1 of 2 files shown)


apps/admin-x-activitypub/src/api/activitypub.test.ts
```diff
@@ -278,43 +278,6 @@ describe('ActivityPubAPI', function () {
 
             expect(actual).toEqual(expected);
         });
-
-        test('Returns an array when the items key is a single object', async function () {
```
mike182uk, line 282: Removed this test case as Fedify now returns an array when there is a single object

apps/admin-x-activitypub/src/api/activitypub.test.ts
```diff
@@ -413,43 +376,6 @@ describe('ActivityPubAPI', function () {
...
```
mike182uk, line 417: Removed this test case as Fedify now returns an array when there is a single object

PR #20701, opened 2024-08-01
Updated publishing flow

minimaluminium:
ref DES-594


ghost/admin/app/components/posts/analytics.js
```diff
@@ -302,6 +355,35 @@ export default class Analytics extends Component {
         this.mentions = yield this.store.query('mention', {limit: 5, order: 'created_at desc', filter});
     }
 
+    @task
+    *fetchPostCountTask() {
+        if (!this.post.emailOnly) {
+            const result = yield this.store.query('post', {filter: 'status:published', limit: 1});
+            let count = result.meta.pagination.total;
+
+            count += 1; // account for the new post
```
vershwal, line 364: I don't think we need to account for the new post here as the store will be updated as soon as the post is published. Please test this.

vershwal: I tested it with a new account. We don't need to do +1 here. It says 5 posts when I have 4 posts published. 
image
image

ghost/admin/app/components/posts/analytics.js
```diff
@@ -302,6 +355,35 @@ export default class Analytics extends Component {
...
+            this.postCount = count;
+        }
+    }
+
+    @task
+    *fetchPostTask() {
+        const result = yield this.store.query('post', {filter: `id:${this.post.id}`, limit: 1});
+        this.post = result.toArray()[0];
+
+        // const publishedAt = this.post.publishedAtUTC;
```
vershwal, line 375: Please remove the commented code if not needed

PR #20702, opened 2024-08-01
Added created_at index to the members_click_events table

vershwal:
Ref https://linear.app/tryghost/issue/ONC-216/improve-the-performance-of-the-membersevents-aggregated-click-event


vershwal:
We don't need this anymore!

PR #20719, opened 2024-08-05
Added configurable target delivery window for batch sending

cmraible:
ref https://linear.app/tryghost/issue/ONC-217/implement-the-deliverytime-option-in-mailgun-api-calls

Ghost experiences its highest peak load immediately after sending out a newsletter, as it recieves an influx of traffic from users clicking on the links in the email, a burst of email analytics events to process from mailgun, and an increase in organic traffic to the site's frontend as well as the admin analytics pages. The `BatchSendingService` currently sends all the batches to Mailgun as quickly as possible, which may contribute to higher peak loads.

This commit adds a `deliverytime` parameter to our API calls to Mailgun, which allows us to specify a time in the future when we want the email to be delivered. This will allow us to moderate the rate at which emails are delivered, and in turn that should moderate the peak traffic volume that Ghost receives in the first 2-3 minutes after sending an email. 

The `deliverytime` is calculated based on a configurable parameter: `bulkEmail.targetDeliveryWindow`, which specifies the maximum allowable time (in milliseconds) after the email is first sent for Ghost to instruct Mailgun to deliver the emails. Ghost will attempt to space out all the batches as evenly as possible throughout the specified window. For example, if the targetDeliveryWindow is set to `300000` (5 minutes) and there are 100 batches, Ghost will set the `deliveryTime` for each batch ~3 seconds apart.

### DIFF
--- a/ghost/email-service/lib/BatchSendingService.js
+++ b/ghost/email-service/lib/BatchSendingService.js
@@ -221,7 +221,9 @@ class BatchSendingService {
     async getBatches(email) {
         logging.info(`Getting batches for email ${email.id}`);
 
-        return await this.#models.EmailBatch.findAll({filter: 'email_id:\'' + email.id + '\''});
+        // findAll returns a bookshelf collection, we want to return a plain array to align with the createBatches method
+        const batches = await this.#models.EmailBatch.findAll({filter: 'email_id:\'' + email.id + '\''});
+        return batches.models;
     }
 
     /**
@@ -354,9 +356,16 @@ class BatchSendingService {
 
     async sendBatches({email, batches, post, newsletter}) {
         logging.info(`Sending ${batches.length} batches for email ${email.id}`);
-
+        const deadline = this.getDeliveryDeadline(email);
+        
+        if (deadline) {
+            logging.info(`Delivery deadline for email ${email.id} is ${deadline}`);
+        }
         // Reuse same HTML body if we send an email to the same segment
         const emailBodyCache = new EmailBodyCache();
+
+        // Calculate deliverytimes for the batches
+        const deliveryTimes = this.calculateDeliveryTimes(email, batches.length);
 
         // Loop batches and send them via the EmailProvider
         let succeededCount = 0;
@@ -367,7 +376,15 @@ class BatchSendingService {
         runNext = async () => {
             const batch = queue.shift();
             if (batch) {
-                if (await this.sendBatch({email, batch, post, newsletter, emailBodyCache})) {
+                const batchData = {email, batch, post, newsletter, emailBodyCache, deliveryTime: undefined};
+                // Only set a delivery time if we have a deadline and it hasn't past yet
+                if (deadline && deadline.getTime() > Date.now()) {
+                    const deliveryTime = deliveryTimes.shift();
+                    if (deliveryTime && deliveryTime >= Date.now()) {
+                        batchData.deliveryTime = deliveryTime;
+                    }
+                }
+                if (await this.sendBatch(batchData)) {
                     succeededCount += 1;
                 }
                 await runNext();
@@ -391,10 +408,10 @@ class BatchSendingService {
 
     /**
      *
-     * @param {{email: Email, batch: EmailBatch, post: Post, newsletter: Newsletter}} data
+     * @param {{email: Email, batch: EmailBatch, post: Post, newsletter: Newsletter, emailBodyCache: EmailBodyCache, deliveryTime:(Date|undefined) }} data
      * @returns {Promise<boolean>} True when succeeded, false when failed with an error
      */
-    async sendBatch({email, batch: originalBatch, post, newsletter, emailBodyCache}) {
+    async sendBatch({email, batch: originalBatch, post, newsletter, emailBodyCache, deliveryTime}) {
         logging.info(`Sending batch ${originalBatch.id} for email ${email.id}`);
 
         // Check the status of the email batch in a 'for update' transaction
@@ -440,9 +457,10 @@ class BatchSendingService {
                 }, {
                     openTrackingEnabled: !!email.get('track_opens'),
                     clickTrackingEnabled: !!email.get('track_clicks'),
+                    deliveryTime,
                     emailBodyCache
                 });
-            }, {...this.#MAILGUN_API_RETRY_CONFIG, description: `Sending email batch ${originalBatch.id}`});
+            }, {...this.#MAILGUN_API_RETRY_CONFIG, description: `Sending email batch ${originalBatch.id} ${deliveryTime ? `with delivery time ${deliveryTime}` : ''}`});
             succeeded = true;
 
             await this.retryDb(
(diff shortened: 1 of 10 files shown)


cmraible:
Todo:
- [ ] Add unit tests — the `MailgunClient.send()` function has limited test coverage, and is a bit tricky to test using the current patterns because the Mailgun endpoint accepts multipart/form-data and `nock` doesn't handle this out of the box.

- [ ] Test on a large email send with many batches — so far I've only tested this on an email with 2 recipients and a batch size of 1 and it works as expected, but there may be some surprises at scale.

- [ ] Move the inter-batch delay to config — it's currently hardcoded to 5 seconds, but ideally we'd be able to enable/disable this via config, and may want to adjust how long the delay is between each subsequent batch.

- [ ] Test with scheduled posts — it should work the same way, but just to be sure...

- [ ] Test with timezones — again, it should workfine, but just to be sure...

ghost/email-service/lib/BatchSendingService.js
```diff
@@ -360,14 +360,19 @@ class BatchSendingService {
 
         // Loop batches and send them via the EmailProvider
         let succeededCount = 0;
-        const queue = batches.slice();
+        const queue = batches.slice().map((batch, index) => {
+            return {batchIndex: index, ...batch};
+        });
```
cmraible, line 365: Since we use a `Promise.all()` to actually send all the batches with concurrency of 2 rather than a simple loop, we can't rely on any incremented counter within the loop to generate the delays — hence this initial loop to add the index to each batch, which is then used to calculate the `deliverytime` parameter below.

ghost/email-service/lib/BatchSendingService.js
```diff
@@ -360,14 +360,19 @@ class BatchSendingService {
...
 
         // Bind this
         let runNext;
+        let startTime = new Date();
         runNext = async () => {
             const batch = queue.shift();
             if (batch) {
-                if (await this.sendBatch({email, batch, post, newsletter, emailBodyCache})) {
+                const deliveryDelay = batch.batchIndex * 5000; // 5 seconds delay between each batch
```
cmraible, line 373: Reminder to self: extract the hardcoded 5000 into a configuration parameter

daniellockyer: Is 5s a suitable value? If we have 100 batches, they could be spread up to 500s (8 mins) apart. Want to make sure this won't cause more problems due to a perceived delay in deliverability 🙂

ghost/email-service/lib/SendingService.js
```diff
@@ -83,10 +83,11 @@ class SendingService {
      * @param {string|null} data.segment
      * @param {string|null} data.emailId
      * @param {MemberLike[]} data.members
+     * @param {Date} data.deliveryTime
      * @param {EmailSendingOptions} options
      * @returns {Promise<EmailProviderSuccessResponse>}
     */
-    async send({post, newsletter, segment, members, emailId}, options) {
+    async send({post, newsletter, segment, members, emailId, deliveryTime}, options) {
```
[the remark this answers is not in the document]
cmraible: Ah good shout, definitely agree

ghost/email-service/lib/BatchSendingService.js
cmraible: Found a slightly cleaner way to do this I think, using `indexOf` rather than precalculating the index

cmraible:
Quick overview of how the email sending job works, currently (before this PR):
1. The `emailJob` runs when a post is sent as an email
2. The `emailJob` creates the batches, based on some complicated segmentation logic to make sure e.g. `paid-only` content doesn't get sent to free members. Creating batches can take ~60 seconds in some cases.
3. Then the `sendBatches` method gets called, which iterates over all the batches (with some obnoxiously complicated concurrency logic — want to refactor this at some point) and sends them to mailgun to be sent immediately (without specifying `deliverytime`.

The basic gist of this change:
1. No changes to the segmentation / creating batches logic. All the magic here happens once the batches are already created.
2. A new configuration value: `bulkEmail.targetDeliveryWindow` specifies the acceptable amount of time for Ghost to take to deliver the emails. A reasonable value I've been testing with is 5 minutes. The goal is for Ghost to space out the emails as evenly as possible over those 5 minutes (but no later).
3. When the sending job first starts, it [calculates a sending deadline](https://github.com/TryGhost/Ghost/pull/20719/files#diff-ed576a9106a4363c8770f9a1dc137ecf8cfa85ca2a1748ecd1d10458bd887ca1R195-R198): `Date.now() + targetDeliveryWindow`. 
4. The `sendEmail` function then [passes this deadline to the `sendBatches` function](https://github.com/TryGhost/Ghost/pull/20719/files#diff-ed576a9106a4363c8770f9a1dc137ecf8cfa85ca2a1748ecd1d10458bd887ca1R217). 
5. `sendBatches` then iterates over all the batches, [calculating the appropriate time for the batch to be sent](https://github.com/TryGhost/Ghost/pull/20719/files#diff-ed576a9106a4363c8770f9a1dc137ecf8cfa85ca2a1748ecd1d10458bd887ca1R372-R388), based on the `deliveryTime` of the previous batch, and the number of remaining batches. **This is the meat of the PR's logic**
6. If for some reason we fall behind (e.g. mailgun is taking a long time to respond, or creating the batches takes way longer than expected, it will either [omit the delivery time](https://github.com/TryGhost/Ghost/pull/20719/files#diff-ed576a9106a4363c8770f9a1dc137ecf8cfa85ca2a1748ecd1d10458bd887ca1R379) altogether or [set the delivery time to the deadline itself](https://github.com/TryGhost/Ghost/pull/20719/files#diff-ed576a9106a4363c8770f9a1dc137ecf8cfa85ca2a1748ecd1d10458bd887ca1R384), effectively instructing mailgun to send immediately because we're past the deadline. This shouldn't happen, but is an added safety measure to make sure we don't accidentally take __too long__ to send the emails.

cmraible:
> 3. When the sending job first starts, it [calculates a sending deadline](https://github.com/TryGhost/Ghost/pull/20719/files#diff-ed576a9106a4363c8770f9a1dc137ecf8cfa85ca2a1748ecd1d10458bd887ca1R195-R198): `Date.now() + targetDeliveryWindow`.


As I write this out, I'm realizing it might make more sense to use `email.created_at + targetDeliveryWindow` rather than basing it on the current timestamp. This will effectively 'persist' the deadline, so if we get into a situation where an email fails and we're retrying, or there is some other unexpected delay, it will skip the `deliverytime` and just send the remaining batches immediately.

9larsons:
Could you add a todo item to check how this is handled during shutdown? My current understanding (I haven't looked through the batch sending code) is that we create a line for each batch and update it as needed, and we would be able to read back from this table and issue the requests to Mailgun after reboot. But... I'm curious.

ghost/email-service/test/sending-service.test.js
```diff
@@ -373,4 +391,17 @@ describe('Sending service', function () {
             sinon.assert.calledOnce(emailProvider.getMaximumRecipients);
         });
     });
+
+    describe('getTargetDeliveryWindow', function () {
+        it('returns the batch delay of the email provider', function () {
```
cmraible, line 396: Update test description

cmraible:
> Could you add a todo item to check how this is handled during shutdown/reboot? My current understanding (I haven't looked through the batch sending code) is that we create a line for each batch and update it as needed, and we would be able to read back from this table and issue the requests to Mailgun after reboot. But... I'm curious what the real answer is.

Yeah, I only tested this part locally so there may (hopefully?) be some more graceful behavior in production. Basically if the process is killed while the sending job is ongoing, the email will get stuck in a `submitting` state, and any batches not yet sent will get stuck and/or not created. 

This PR doesn't change that behavior at all (tested this on main and in this branch and it was the same behavior) so I'm going to go ahead and merge, but we def need to think about how we can recover more gracefully.

PR #20793, opened 2024-08-20
Updated tips & donations default suggested value

kevinansfield:
closes https://linear.app/tryghost/issue/PLG-156

- updated all default fixtures to use `500` ($5) as the default suggested donation value
- added migration to update existing settings using the old default of `0` to `500`
  - this is fine to apply because the feature hasn't been released so there's no explicit `0` values in the wild
- added an acceptance test for the adminx-settings tips & donations section


sagzy (approved):
LGTM 👍

PR #20813, opened 2024-08-22
Added integrity tokens for magic link email requests

sam-lord:
This should prevent untargeted attacks, since the back-end requires a token that is generated by the server before processing the request.

This is enabled as a warning in the logs only, until we're happy with the breaking change to the `send-magic-link` API.

### DIFF
--- a/ghost/core/core/server/services/members/service.js
+++ b/ghost/core/core/server/services/members/service.js
@@ -19,6 +19,7 @@ const tiersService = require('../tiers');
 const VerificationTrigger = require('@tryghost/verification-trigger');
 const DatabaseInfo = require('@tryghost/database-info');
 const settingsHelpers = require('../settings-helpers');
+const RequestIntegrityTokenProvider = require('./RequestIntegrityTokenProvider');
 
 const messages = {
     noLiveKeysInDevelopment: 'Cannot use live stripe keys in development. Please restart in production mode.',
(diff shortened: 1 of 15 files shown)


ghost/core/core/server/services/members/service.js
```diff
@@ -193,6 +194,11 @@ module.exports = {
     ssr: null,
     verificationTrigger: null,
 
+    requestIntegrityTokenProvider: new RequestIntegrityTokenProvider({
+        themeSecret: settingsCache.get('theme_session_secret'),
+        tokenDuration: 1000 * 60 * 5
```
9larsons, line 199: Are we wanting to make this last a shorter duration than the magic links to check for spam activity? Those are 24h.

9larsons (commented):
Only one small point on the duration. As these are logging generic messages, we should be able to easily track volume in elastic.

While we're pushing changes out there, do we want to add a config flag for this? Some self-hosters appear to be getting hit really hard, and even if we haven't validated this 100%, this would give them an option.

9larsons:
We briefly touched on it in Slack. We likely want to update `signup-form`'s api calls otherwise we could invalidate those using this if/when we do block requests w/o the integrity token.

https://github.com/TryGhost/Ghost/blob/main/apps/signup-form/src/utils/api.tsx

ghost/core/core/server/services/members/service.js
sam-lord: I think so, these are generated immediately before use, so we could even have them be valid for as little as 60~ seconds. There should never be a reason for a session to use an old token.

9larsons: Got it, yeah. I was thinking it was verified w/ the magic link click, not on generating the magic link, for whatever reason. Sounds good to me.

9larsons (approved):

PR #20814, opened 2024-08-22
Added honeypot field to prevent bot signup / sign-in

sam-lord:
ref KTLO-1
Should prevent untargeted attacks using headless browser bots.

### DIFF
--- a/ghost/members-api/lib/controllers/RouterController.js
+++ b/ghost/members-api/lib/controllers/RouterController.js
@@ -493,6 +493,8 @@ module.exports = class RouterController {
         }
 
         if (honeypot) {
+            logging.warn('Honeypot field filled, this is likely a bot');
+
             // Honeypot field is filled, this is a bot.
             // Pretend that the email was sent successfully.
             res.writeHead(201);
--- a/ghost/members-api/lib/controllers/RouterController.js
+++ b/ghost/members-api/lib/controllers/RouterController.js
@@ -470,7 +470,7 @@ module.exports = class RouterController {
     }
 
     async sendMagicLink(req, res) {
-        const {email, autoRedirect} = req.body;
+        const {email, honeypot, autoRedirect} = req.body;
         let {emailType, redirect} = req.body;
 
         let referer = req.get('referer');
@@ -490,6 +490,13 @@ module.exports = class RouterController {
             throw new errors.BadRequestError({
                 message: tpl(messages.emailRequired)
             });
+        }
+
+        if (honeypot) {
+            // Honeypot field is filled, this is a bot.
+            // Pretend that the email was sent successfully.
+            res.writeHead(201);
+            return res.end('Created.');
         }
 
         if (!emailType) {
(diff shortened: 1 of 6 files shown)


ghost/members-api/lib/controllers/RouterController.js
```diff
@@ -492,6 +492,13 @@ module.exports = class RouterController {
             });
         }
 
+        if (honeypot) {
```
daniellockyer, line 495: Can we log out in this block? Otherwise it's gonna be difficult to know when we're receiving honeypot'd requests

9larsons: Agreed, since the response is the same.

ghost/members-api/test/unit/lib/controllers/router.test.js
```diff
@@ -388,5 +388,52 @@ describe('RouterController', function () {
...
+                sendEmailWithMagicLinkStub = sinon.stub().resolves();
+            });
+
+            it('Sends emails when honeypot is not filled', async function () {
+                const controller = createRouterController();
+
+                req.body.honeypot = 'filled!';
+
+                await controller.sendMagicLink(req, res).should.be.fulfilled();
+                sendEmailWithMagicLinkStub.notCalled.should.be.true();
```
9larsons, line 428: Isn't this incorrect? Shouldn't we see the email stub called?

The inverse being true for the following test.

ghost/members-api/test/unit/lib/controllers/router.test.js
```diff
@@ -388,5 +388,52 @@ describe('RouterController', function () {
...
+            });
+
+            it('Does not send emails when honeypot is filled', async function () {
+                const controller = createRouterController();
+
+                await controller.sendMagicLink(req, res).should.be.fulfilled();
+                sendEmailWithMagicLinkStub.calledOnce.should.be.true();
```
9larsons, line 435: Flagging this as well. See above.

9larsons (dismissed):
Question on the functional tests and +1 to Daniel's comment on logging.

ghost/members-api/test/unit/lib/controllers/router.test.js
sam-lord: The honeypot is a fake input for a mobile phone number. The idea is that if a bot fills it then we _shouldn't_ send the email, since they've fallen into the trap :-)

ghost/members-api/lib/controllers/RouterController.js
sam-lord: Good shout, I'll add logging.

ghost/members-api/test/unit/lib/controllers/router.test.js
9larsons: Yes but I think you have the titles for the tests backwards, then? Either the content or test titles/descriptions need to swap. Functionally I think you're all good 👍.

PR #20821, opened 2024-08-23
🎨 Updated uuid to crypto.randomUUID()

stevejubx:
I noticed that Ghost only uses UUID version 4, so installing the uuid library unnecessarily increases the project size.

I think it should be replaced with crypto.randomUUID(). What do you think?


allouis:
@stevejubx Why did you close this PR?

PR #20843, opened 2024-08-28
Fixed welcome email being sent for Admin API creates

allouis:
This fix is done at the endpoint layer, which isn't ideal, but has the smallest
surface area for the change. I think we may want to move it up a layer though,
despite the extra complications. If we move the check _into_ the MembersAPI
class however, we start to run into issues with circular dependencies due to the
mess that is our dependency injection approach with the Members service

### DIFF
--- a/ghost/core/core/server/api/endpoints/members.js
+++ b/ghost/core/core/server/api/endpoints/members.js
@@ -2,6 +2,7 @@
 // as it is a getter and may change during runtime.
 const moment = require('moment-timezone');
 const errors = require('@tryghost/errors');
+const logging = require('@tryghost/logging');
 const models = require('../../models');
 const membersService = require('../../services/members');
 
@@ -11,6 +12,7 @@ const _ = require('lodash');
 
 const messages = {
     memberNotFound: 'Member not found.',
+    notSendingWelcomeEmail: 'Email verification required, welcome email is disabled',
     memberAlreadyExists: {
         message: 'Member already exists',
         context: 'Attempting to {action} member with existing email address.'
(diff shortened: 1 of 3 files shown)


allouis:
What's your initial thoughts on this @mike182uk? Do you think I should move the check up higher, or keep things simple for a fix?

mike182uk:
@allouis I don't think its too bad of an approach and understand the reason for doing so. I had a quick look and how easy it would be to inject the `verificationTrigger` into the `membersBreadApi`, and unless I'm missing something, it would be doable without circular deps, i.e swapping the initialisation stuff around [here](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/services/members/service.js#L144-L160) to be something like:

```
const verificationTrigger = initVerificationTrigger();
module.exports.verificationTrigger = verificationTrigger;

if (!membersApi) {
    membersApi = createMembersApiInstance(membersConfig, verificationTrigger);

    membersApi.bus.on('error', function (err) {
        logging.error(err);
    });
}

module.exports.ssr = MembersSSR({
    cookieSecure: urlUtils.isSSL(urlUtils.getSiteUrl()),
    cookieKeys: [settingsCache.get('theme_session_secret')],
    cookieName: 'ghost-members-ssr',
    getMembersApi: () => module.exports.api
});
```

But would mean passing `verificationTrigger` down a bunch of layers, which might not be feasible.

The only other thought is can / would `membersService.api.memberBREADService.add` be called outside the context of this endpoint? If so would the caller know they need to check verification first via this method etc.

allouis:
@mike182uk Switching the order round like that doesn't work, because the `initVerificationTrigger` function depends on the `membersApi` variable being set, so that it can read the events repository 🙃 

The BREAD service is only used from the endpoint atm, but yes you're right, with this approach if we're to use it else where the caller would need to know to add this check

mike182uk:
Ah I missed that. I think there is probably more value in resolving the issue, rather than a largish refactor to get around the circular dependency problem here, and seeming though the method is only used in this context, it is less likely to be called without the explicit verification step (and at this point there is probably a greater need to do the refactor)

PR #20863, opened 2024-08-28
Added type checks to admin-x-design-system unit tests

cmraible:
no issue

- Previously we weren't running the type checks in the `admin-x-design-system` in CI, because we only run `yarn test:unit` in CI. This adds the typechecks to the `yarn test:unit` command so CI will fail if the type checks fail.

### DIFF
--- a/apps/admin-x-framework/package.json
+++ b/apps/admin-x-framework/package.json
@@ -103,14 +103,14 @@
     "targets": {
       "dev": {
         "dependsOn": [
-          "^build",
-          "@tryghost/admin-x-design-system:build"
+          "^build"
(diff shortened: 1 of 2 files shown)


apps/admin-x-framework/package.json
```diff
@@ -109,7 +109,8 @@
       "test:unit": {
         "dependsOn": [
           "test:unit",
-          "^build"
+          "^build",
+          "@tryghost/admin-x-design-system:test:unit"
```
cmraible, line 113: For some reason tests were flaking in CI without adding this — some kind of build/race condition I think

PR #20890, opened 2024-09-02
Added sentry to support error on Portal

ronaldlangeveld:
ref PLG-204

- Added a sentry capture so we know when and if the support error component gets triggered.


cmraible:
@ronaldlangeveld just a heads up, last I checked we didn't have Sentry enabled on portal!

ronaldlangeveld:
> @ronaldlangeveld just a heads up, last I checked we didn't have Sentry enabled on portal!

Thanks for the heads up! Wasn't aware of it, since there's loads of Sentry related code inside Portal. 😅

PR #20914, opened 2024-09-03
Reimplemented email analytics prioritizing email opens

9larsons:
ref https://github.com/TryGhost/Ghost/pull/20835
- reimplemented email analytics changes that prioritized opened events over other events in order to speed up open analytics
- added db persistence to fetch missing job to ensure we re-fetch every window of events, especially important if we restart following a large email batch

We learned a few things with the previous trial run of this. Namely, that event throughput is not as high as we initially saw in the data for particularly large databases. This set of changes is more conservative, while a touch more complicated, in ensuring we capture edge cases for really large newsletter sends (100k+ members).

In general, we want to make sure we're fetching new open events at least every 5 mins, and often much faster than that, unless it's a quiet period (suggesting we haven't had a newsletter send or much outstanding event data).


9larsons:
I'm thinking we may want to include some sort of backstop for the `jobs` table data in case it isn't cleared during this migration, the migration isn't run, etc. I don't know what would be reasonable and I also don't want to obscure the existing code with weird handling of cases like this. Ideas?

cmraible:
> I don't know what would be reasonable and I also don't want to obscure the existing code with weird handling of cases like this. Ideas?

Maybe a silly idea, but could we use a new name for the job, so it wouldn't even look at the same rows if the migration doesn't run/fails for some reason? Maybe we could introduce a concept of job version, so the rows would become e.g. `email-analytics-latest-opened-v2` or something along those lines? And if we ever come across a similar challenge in the future, we just increment the version and Ghost should fallback to running the full query if it doesn't find a matching row in the table?

9larsons:
While that does work, I don't see a need to version and that makes future migrations to try and clear this out extra messy. I'd personally rather avoid that.

ghost/core/core/server/services/email-analytics/EmailAnalyticsServiceWrapper.js
```diff
@@ -100,13 +111,34 @@ class EmailAnalyticsServiceWrapper {
...
         try {
-            const c1 = await this.fetchLatest({maxEvents: Infinity});
-            const c2 = await this.fetchMissing({maxEvents: Infinity});
-
-            // Only fetch scheduled if we didn't fetch a lot of normal events
-            await this.fetchScheduled({maxEvents: 20000 - c1 - c2});
-
+            // Prioritize opens since they are the most important (only data directly displayed to users)
+            const c1 = await this.fetchLatestOpenedEvents({maxEvents: 10000});
+            if (c1 >= 10000) {
```
cmraible, line 119: Will this ever be true if maxEvents is set to 10000?

cmraible: Also just out of curiosity, how did you arrive at 10,000 as the right number for this? 2500 events per minute * 4 minutes?

ghost/core/core/server/data/migrations/versions/5.93/2024-09-03-20-09-40-null-analytics-jobs-timings.js
```diff
@@ -0,0 +1,15 @@
+// For information on writing migrations, see https://www.notion.so/ghost/Database-migrations-eb5b78c435d741d2b34a582d57c24253
```
cmraible, line 1: I think we need to change the path to the migration, since we released 5.93 today 😅 I had to change it to `/versions/5.94/asdfasdfasd` to get this to run locally

ghost/core/core/server/data/migrations/versions/5.93/2024-09-03-20-09-40-null-analytics-jobs-timings.js
```diff
@@ -0,0 +1,15 @@
...
+
+// For DML - data changes
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        await knex.raw(`
+            DELETE FROM jobs
+            WHERE name = 'email-analytics-latest-opened' OR name = 'email-analytics-latest-others' OR name = 'email-analytics-missing';
+        `);
```
cmraible, line 11: We may want to wrap this in a try/catch with some error logging, just in case there's some kind of DB error

cmraible:
I think I'm missing something here, maybe you can help me understand. I'm not sure I see how this avoids us missing events with `fetchMissing`. I see the main changes here are that we're fetching a max of 10k opened events, rather than infinity. 

As I understand this logic, if we fetch 10k opened events, we'll stop there and wait for the job to kick off again. If we end up fetching 10k opened events for a couple iterations in a row, couldn't we fall behind on `fetchMissing` again?

Again, I'm probably just confused! Maybe we can get on a call tomorrow AM to go over it 😃

ghost/core/core/server/services/email-analytics/EmailAnalyticsServiceWrapper.js
```diff
@@ -111,30 +111,33 @@ class EmailAnalyticsServiceWrapper {
...
+            }
 
             // Set limits on how much we fetch without checkings for opened events. During surge events (following newsletter send)
             //  we want to make sure we don't spend too much time collecting delivery data.
-            const c1 = await this.fetchLatestNonOpenedEvents({maxEvents: 20000});
-            if (c1 > 15000) { 
-                this.fetching = false;
-                logging.info('[EmailAnalytics] Restarting fetch due to high event count');
-                this.startFetch();
+            const c2 = await this.fetchLatestNonOpenedEvents({maxEvents: 10000 - c1});
```
cmraible, line 126: Am I understanding this correctly? If we fetch 10k opened events, we won't run `fetchLatestNonOpenedEvents` or `fetchMissing`?

9larsons: Yeah, correct. I realize this part isn't different. The key is I added db persistence to the `fetchMissing` and bumped back the fallback following boot to help account for the rollout.

Previously I had `maxEvents: Infinity` for `fetchLatestOpenedEvents`, which we could go back to, as this just keeps looping on 10k. I had it looping so that it's re-cycling around every 5 mins. That part is kinda semantics.. it's the same behavior just a bit more logging. I'm not bullish on either approach.

ghost/core/core/server/services/email-analytics/EmailAnalyticsServiceWrapper.js
9larsons: I'm being a bit conservative. The thought was the restart the job's logic _at least_ every 5 mins, so 2500*4 gives a little head room as that 2500 is variable and has an inverse relationship with the db size (email_recips table size, primarily).

I'm not following the first comment. It's >= not just >. I saw some notes previously that `maxEvents` wasn't super precise. In testing I've been seeing it fairly precise. I was going to put it at 8000 or something but that seemed very arbitrary and difficult for someone reading this to understand. Not that it's easy as is.

ghost/core/core/server/data/migrations/versions/5.93/2024-09-03-20-09-40-null-analytics-jobs-timings.js
9larsons: Added.

ghost/core/core/server/data/migrations/versions/5.93/2024-09-03-20-09-40-null-analytics-jobs-timings.js
9larsons: Updated, good catch.

cmraible (approved):

ghost/core/core/server/data/migrations/versions/5.94/2024-09-03-20-09-40-null-analytics-jobs-timings.js
```diff
@@ -0,0 +1,21 @@
...
+// For DML - data changes
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        try {
+            await knex.raw(`
+                DELETE FROM jobs
+                WHERE name = 'email-analytics-latest-opened' OR name = 'email-analytics-latest-others' OR name = 'email-analytics-missing';
+            `);
```
daniellockyer, line 14:
```suggestion
            await knex('jobs')
            	.whereIn('name', ['email-analytics-latest-opened', 'email-analytics-latest-others', 'email-analytics-missing'])
            	.del();
```

daniellockyer: I personally prefer Knex utils, as it's SQL-flavor independent

PR #20918, opened 2024-09-04
Refactored Stripe webhook controller

ronaldlangeveld:
no issue

- Moved business logic from `WebhookController` to dedicated service classes (`SubscriptionEventService`, `InvoiceEventService`, `CheckoutSessionEventService`).
- Reduced controller complexity.
- Added unit tests for individual services, increasing overall test coverage.
- Improved maintainability and scalability by isolating responsibilities in specific services, making future updates easier and safer.


sagzy (approved):
The code is looking really good! 💯 Plus the additional unit tests are going to be very useful in the future, especially as you guys work on the Stripe tax feature and so on 

On top of the green test suite, I have locally explored the use cases I could think of (paid subscription, upgrade, downgrade, manual cancellation, cancellation due to failed payments, donation), everything seems to be working fine 💪

ronaldlangeveld:
> The code is looking really good! 💯 Plus the additional unit tests are going to be very useful in the future, especially as you guys work on the Stripe tax feature and so on
> 
> On top of the green test suite, I have locally explored the use cases I could think of (paid subscription, upgrade, downgrade, manual cancellation, cancellation due to failed payments, donation), everything seems to be working fine 💪

Thanks so much for your time checking it! ☺️

ghost/stripe/lib/WebhookController.js
```diff
@@ -27,6 +27,25 @@ module.exports = class WebhookController {
...
+        // Limiting injected dependencies may cause failures when certain dependencies
+        // are not explicitly passed but are expected by the services at runtime.
+        this.initializeServices(deps);
+    }
+
+    /**
+     * Initializes event services
+     * @param {object} deps 
+     */
+    initializeServices(deps) {
```
allouis, line 44: Rather than initialise these services here, you can pass them in as dependencies - this gives you the full benefit of dependency injection!

In this case it would mean that the dependencies of the `StripeWebhookController` are just `webhookManager` and these three services! That would mean unit tests for the controller are a lot simpler, too

ghost/stripe/test/unit/lib/services/webhooks/InvoiceEventService.test.js
```diff
@@ -0,0 +1,201 @@
+const chai = require('chai');
```
allouis, line 1: As far as I know we should be using the built in `assert` library for assertions - rather than `should`/`chai`

allouis (commented):
I left some minor, non-blocking, comments ☺️ 

I have a smol question about what you mean when you refer to dynamic access to dependencies - I was imagining something like `this.deps[someDynamicVariable]` but they're all hardcoded lookups - so does that mean that the underlying values of `this.deps.<thing>` is changing?

ghost/stripe/lib/WebhookController.js
ronaldlangeveld: Thanks for the suggestion. Made this change and it works great!

ghost/stripe/test/unit/lib/services/webhooks/InvoiceEventService.test.js
ronaldlangeveld: Thanks for the heads up. Swapped them out!

PR #20919, opened 2024-09-04
ActivityPub stats, following and followers

djordjevlais:
- Moved engagement stats to a reusable component
- Fixed Following and Followers on Your Profile and moved them from modals to tabs

### DIFF
--- a/apps/admin-x-activitypub/src/components/Profile.tsx
+++ b/apps/admin-x-activitypub/src/components/Profile.tsx
@@ -1,8 +1,10 @@
 import APAvatar from './global/APAvatar';
+import ActivityItem from './activities/ActivityItem';
 import MainNavigation from './navigation/MainNavigation';
 import React, {useState} from 'react';
+import getUsername from '../utils/get-username';
 import {ActivityPubAPI} from '../api/activitypub';
-import {Heading, NoValueLabel, Tab, TabView} from '@tryghost/admin-x-design-system';
+import {Button, Heading, List, NoValueLabel, Tab, TabView} from '@tryghost/admin-x-design-system';
 import {useBrowseSite} from '@tryghost/admin-x-framework/api/site';
 import {useQuery} from '@tanstack/react-query';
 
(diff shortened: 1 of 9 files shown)


apps/admin-x-activitypub/src/components/Profile.tsx
```diff
@@ -42,9 +44,49 @@ function useFollowingCountForUser(handle: string) {
...
+    const api = new ActivityPubAPI(
+        new URL(siteUrl),
+        new URL('/ghost/api/admin/identities/', window.location.origin),
+        handle
+    );
+    return useQuery({
+        queryKey: [`followers:${handle}`],
+        async queryFn() {
+            const followerUrls = await api.getFollowers();
+            const followerActors = await Promise.all(followerUrls.map(url => api.getActor(url)));
```
allouis, line 79: This is going to create a fuck load of API queries - especially for sites with a lot a followers - this looks to me like our API isn't serving you well, and we need to fix it on the backend!

allouis (commented):
Some feedback about the react component, I think a small refactor will make it much nicer to work with!

As for the followers stuff - I think this is broken on our end - so you shouldn't have to make all of these requests

apps/admin-x-activitypub/src/components/Profile.tsx
```diff
@@ -42,9 +44,49 @@ function useFollowingCountForUser(handle: string) {
     });
 }
 
+function useFollowingForUser(handle: string) {
```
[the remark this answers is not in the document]
djordjevlais: Deal, would this mean simply moving them to a separate file and importing them, or is there more to it? (i.e. should I give it a try or leave it to you/Mike?)

apps/admin-x-activitypub/src/components/Profile.tsx
djordjevlais: @allouis TBH this one is a bit over my head 😅 But looking at [viewFollowersModal.tsx](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-activitypub/src/components/profile/ViewFollowersModal.tsx) on main, I think that’s how it functioned from the start. But at some point it started rendering an empty list in the frontend, so I thought it’d be just a quick fix so gave it a go with Cursor’s help, should have checked with you guys first 🐒  

So yeah, if it makes more sense to change the API let’s do it! One thing though: I don’t wanna create extra work for you this week since this is not Likes-related, so does it make sense to merge this part as is, and then change that later? Or just revert that commit and revisit it once you change the API?

apps/admin-x-activitypub/src/components/Profile.tsx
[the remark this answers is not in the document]
allouis: Yeah exactly! Just pulling out to a separate file so they can be reused and tested easier!

apps/admin-x-activitypub/src/components/Profile.tsx
allouis: I think revert that commit, and we will live with it being broken until it's fixed at the API layer ☺️

PR #20922, opened 2024-09-04
Added middleware to prevent other sites' content from being served

sam-lord:
ref ONC-294

### DIFF
--- a/ghost/core/core/app.js
+++ b/ghost/core/core/app.js
@@ -37,7 +37,7 @@ const siteIdMiddleware = function siteIdMiddleware(req, res, next) {
         return next();
     }
 
-    logging.info(`Mismatched site id (expected ${configSiteId}, got ${headerSiteId})`);
+    logging.warn(`Mismatched site id (expected ${configSiteId}, got ${headerSiteId})`);
 
     res.set({
         'Cache-Control': 'no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0'
--- a/ghost/core/core/app.js
+++ b/ghost/core/core/app.js
@@ -27,11 +27,29 @@ const maintenanceMiddleware = function maintenanceMiddleware(req, res, next) {
     fs.createReadStream(path.resolve(__dirname, './server/views/maintenance.html')).pipe(res);
 };
 
+// Used by Ghost (Pro) to ensure that requests cannot be served by the wrong site
+const siteIdMiddleware = function siteIdMiddleware(req, res, next) {
+    const siteId = config.get('hostSettings:siteId');
+
+    if (req.headers['x-site-id'] === siteId) {
+        return next();
+    }
+
+    res.set({
+        'Cache-Control': 'no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0'
+    });
+    res.writeHead(500);
+    res.end();
+};
+
 const rootApp = () => {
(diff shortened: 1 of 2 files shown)


ghost/core/core/app.js
```diff
@@ -27,12 +27,30 @@ const maintenanceMiddleware = function maintenanceMiddleware(req, res, next) {
     fs.createReadStream(path.resolve(__dirname, './server/views/maintenance.html')).pipe(res);
 };
 
+// Used by Ghost (Pro) to ensure that requests cannot be served by the wrong site
+const siteIdMiddleware = function siteIdMiddleware(req, res, next) {
+    const siteId = config.get('hostSettings:siteId');
+
+    if (req.headers['x-site-id'] === siteId) {
+        return next();
+    }
```
daniellockyer, line 36: I think we should cast both items to a string, so we can be sure of the types. The config is a string (although the ID is a number) but I think we pass the `x-site-id` as a number, so it's just safer to convert to the same

```suggestion
    const siteIdConfig = config.get('hostSettings:siteId');
    const siteIdHeader = req.headers['x-site-id'];

    if (`${siteIdConfig}` === `${siteIdHeader}`) {
        return next();
    }
```

ghost/core/core/app.js
```diff
@@ -27,12 +27,30 @@ const maintenanceMiddleware = function maintenanceMiddleware(req, res, next) {
...
+    res.end();
+};
+
 const rootApp = () => {
     const app = express('root');
     app.use(sentry.requestHandler);
     if (config.get('sentry')?.tracing?.enabled === true) {
         app.use(sentry.tracingHandler);
     }
+    if (config.get('hostSettings:validateSiteId')) {
```
daniellockyer, line 51: Would we be able to avoid adding another config key by checking the presence of `hostSettings:siteId`?

ghost/core/core/app.js
```diff
@@ -27,12 +28,33 @@ const maintenanceMiddleware = function maintenanceMiddleware(req, res, next) {
...
+// Used by Ghost (Pro) to ensure that requests cannot be served by the wrong site
+const siteIdMiddleware = function siteIdMiddleware(req, res, next) {
+    const configSiteId = config.get('hostSettings:siteId');
+    const headerSiteId = req.headers['x-site-id'];
+
+    if (`${configSiteId}` === `${headerSiteId}`) {
+        return next();
+    }
+
+    logging.info(`Mismatched site id (expected ${configSiteId}, got ${headerSiteId})`);
```
daniellockyer, line 40:
```suggestion
    logging.warn(`Mismatched site id (expected ${configSiteId}, got ${headerSiteId})`);
```

ghost/core/test/e2e-frontend/site_id_middleware.test.js
```diff
@@ -0,0 +1,90 @@
...
+
+    after(async function () {
+        sinon.restore();
+
+        configUtils.restore();
+
+        await testUtils.stopGhost();
+    });
+
+    describe('Using site-id middleware', function () {
```
daniellockyer, line 26: Can you add some tests that use a numerical `x-site-id` too?

mike182uk, line 26: Should we add an extra safety check test for when `hostSettings:siteId` is explicitly not set (middleware should not be executed)?

PR #20927, opened 2024-09-05
Comentários PT

chicodurden21:
Just added the portuguese translation for the comments file. I am a portuguese native, these are fine.


kevinansfield:
Thanks @chicodurden21

PR #20956, opened 2024-09-10
Added reply handling to the notifications tab of the admin-x-activitypub app

mike182uk:
refs [AP-279](https://linear.app/tryghost/issue/AP-279/handle-incoming-replies)

Added reply handling to the notifications tab of the admin-x-activitypub app so that replies to posts can be viewed by the user


apps/admin-x-activitypub/src/components/Activities.tsx
```diff
@@ -70,29 +91,79 @@ const getActorUrl = (activity: Activity): string | null => {
...
     case ACTVITY_TYPE.FOLLOW:
         return 'user-fill';
     case ACTVITY_TYPE.LIKE:
         if (activity.object) {
             return 'heart-fill';
         }
-    }    
-};
-
-const isFollower = (id: string, followerIds: string[]): boolean => {
```
mike182uk, line 82: Moved this to be scoped inside of the component so we don't have to pass `followerIds`

PR #20981, opened 2024-09-11
🌵 [SPIKE] Investigate feasibility of using prom-client for Ghost metrics

cmraible:
ref https://linear.app/tryghost/issue/ENG-1505/start-monitoring-event-loop-utilization-in-production-with
ref https://linear.app/tryghost/issue/PRO-197/node-runtime-performance-prometheus-exporter

- This spike adds a `/metrics` endpoint to Ghost that exposes all the [default metrics](https://github.com/siimon/prom-client/tree/master/lib/metrics) included in `prom-client`. The list includes some basics, like CPU and memory usage, metrics for Node's event loop, and metrics for Node's garbage collection. For many of these metrics, we currently have no visibility into them.
- It also includes some custom metrics like Ghost's connection pool utilization, request queue depth, active workers for the jobs manager, and some basic HTTP response time percentile data.
- It also adds two new services to our `docker-compose.yml` for development: prometheus, to scrape the metrics from Ghost, and Grafana to expose the metrics in a nice browser-based dashboard so you can view these metrics locally:

250rps-main

- With this foundation, we can also add our own custom metrics to our code, that will automatically be reported to prometheus and can be visualized in Grafana.
- 

# Todo
- [ ] Decouple from existing express apps and expose this on its own port, to avoid exposing metrics publicly and so Ghost will (hopefully) respond to scraping requests even if the main express apps are falling over
- [ ] Pull the prometheus (and any custom metrics) into their own package?
- [ ] Figure out why request queue unit tests are failing
- [ ] Investigate using docker compose profiles to only boot prometheus and grafana if opted in via environment variable

### DIFF
--- a/.github/scripts/docker-compose.yml
+++ b/.github/scripts/docker-compose.yml
@@ -28,5 +28,25 @@ services:
     ports:
       - "6379:6379"
     restart: always
+  prometheus:
+    image: prom/prometheus:v2.30.3
+    container_name: ghost-prometheus
+    ports:
+      - "9090:9090"
+    restart: always
+    volumes:
+      - ./prometheus/prometheus.yml:/etc/prometheus/prometheus.yml
+  grafana:
+    image: grafana/grafana:8.2.3
+    container_name: ghost-grafana
+    ports:
+      - "3000:3000"
+    restart: always
+    environment:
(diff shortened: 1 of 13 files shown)


cmraible:
Closing this PR as I'll be reimplementing this in a new branch. Keeping the branch around for now as reference.

PR #20985, opened 2024-09-12
Added experimental background job queue

9larsons:
ref https://linear.app/tryghost/issue/ENG-1556/
- added background job queue behind config flags
- when enabled, is only used for the member email analytics updates in order to speed up the parent job, and take load off of the main process that is serving requests


ghost/core/core/server/data/schema/schema.js
```diff
@@ -985,7 +985,9 @@ module.exports = {
         started_at: {type: 'dateTime', nullable: true},
         finished_at: {type: 'dateTime', nullable: true},
         created_at: {type: 'dateTime', nullable: false},
-        updated_at: {type: 'dateTime', nullable: true}
+        updated_at: {type: 'dateTime', nullable: true},
+        metadata: {type: 'text', maxlength: 1000000000, fieldtype: 'long', nullable: true},
```
cmraible, line 989: Do you have a sense of how big we'd expect this field to actually be? I wonder if allowing a maxlength this large is potentially encouraging misuse of this field?

ghost/core/core/server/data/schema/schema.js
```diff
@@ -985,7 +985,9 @@ module.exports = {
...
+        queue_entry: {type: 'integer', nullable: true, unsigned: true}
```
cmraible, line 990: Just curious: why integer instead of string?

Also a small nit: would `queue` suffice as a column name?

ghost/core/core/server/services/mentions-jobs/job-service.js
```diff
@@ -42,7 +42,7 @@ const initTestMode = () => {
     }, 5000);
 };
 
-const jobManager = new JobManager({errorHandler, workerMessageHandler, JobModel: models.Job, domainEvents});
+const jobManager = new JobManager({errorHandler, workerMessageHandler, JobModel: models.Job, domainEvents, isDuplicate: true});
```
cmraible, line 45: Not a blocker but just a note for future: it would be good to get to the bottom of why we need two instances of the job system, and ideally fix that so we don't need this anymore

ghost/core/core/server/data/schema/schema.js
9larsons: I originally had it like other text fields at 2000 chars. I _think_ that ought to be sufficient, but this is why I was holding off on the migration. We could start with 2000 and go from there?

ghost/core/core/server/data/schema/schema.js
9larsons: I felt like the schema changes were the most contentious 😂.

Right now it's being used to indicate a queue entry when `queue_entry = 1`. We could do various queues, but that would be a departure from the design of the current queue, which is a 'FIFO for all things' kind of model. I'm not sure I want to get into the complications of multiple queues without having proofed this on Pro, or seen that it's necessary.

Indexing is faster w/ integer values, I believe, and I was thinking of just making this 0/1. (I also considered adding an index to this for `queue,created_at` if we think it'll get big/slow).

I was also wondering if instead of multiple queues by string values, if we wanted to make priorities ie. when pulling jobs, `order by queue_entry,created_at desc`. 

Wdyt?

ghost/core/core/server/services/email-analytics/jobs/index.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
                 const m = Math.floor(Math.random() * 5); // 0-4
 
                 jobsService.addJob({
-                    at: `${s} ${m}/5 * * * *`,
+                    at: `${s} ${m}/1 * * * *`,
```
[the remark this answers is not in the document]
9larsons: Whoops this is from testing to make the job run faster after boot. I'll remove it.

ghost/core/core/server/services/mentions-jobs/job-service.js
9larsons: Agreed! I've got a note to look into this because this is rather clunky way of handling it that I'm also really not a fan of.

ghost/core/core/server/data/schema/schema.js
9larsons: Cool, I'll bump back. Basically, since this is a 'JSON' field I had copied what we had elsewhere. Agree it seems to be overkill and it's easier to bump up than down.

ghost/core/core/server/data/schema/schema.js
cmraible: > I felt like the schema changes were the most contentious 
Ha, yeah definitely — it's easy enough to change everything else down the line but I think it's worth putting some thought into making sure we get the schema as close to correct as we can the first time.

Yeah keeping it as an int makes sense for now I think since we're not doing multi-queues right away, but I'd like to redact my nit about the name, and keep it as `queue_entry` so that we can still use a string `queue` field in the future if we go down that route.

I think long-term we should move toward _all_ jobs in this table being in the queue, so eventually we can even remove this field altogether. But I think this makes sense given that this table is currently a bit mixed-use.

ghost/job-manager/lib/JobQueueManager.js
```diff
@@ -0,0 +1,202 @@
...
+     * 
+     * Key features:
+     * - Maintains a minimum of 500 pending tasks in the worker pool
+     * - Dynamically adjusts polling interval between 1 second and 1 minute
+     * - Uses an in-memory set to prevent duplicate job processing
+     * - Handles job execution and cleanup
+     * 
+     * @returns {Promise<void>}
+     */
+    async startQueueProcessor() {
```
cmraible, line 87: Just for the sake of readability and testability I think it might be good to break this bad boy up a bit, it's a little hard for me to follow in its current form.

ghost/core/core/server/data/schema/schema.js
9larsons: Yep, agreed. Still not 100% on how we'd want to handle scheduled jobs in that sort of setup.

ghost/core/core/server/data/schema/schema.js
9larsons: Adjusted. If I mark this resolved, does it also resolve it for you? I forget how GH handles that.

ghost/job-manager/lib/JobQueueManager.js
9larsons: I did a pretty significant refactor of that entire class to break it up a lot, and expose more so testing is easier.

9larsons:
Need to add a couple integration tests then we ought to be ok. Not sure what the failure for Ghost-CLI is.

vikaspotluri123:
It looks like there's an error that's preventing the persisted logs from being printed 🙃 [Here's](https://github.com/TryGhost/Ghost/actions/runs/10890598245/job/30219749988?pr=20985#step:11:106) the culprint:

> Ghost was able to start, but errored during boot with: Cannot find module 'workerpool'

Looking at the lockfile changes, I'm not sure if they're intentional - there are a lot of new dependencies, which I think might stem from an older version of `@tryghost/errors`
Also, the real issue is that it looks like workerpool wasn't added as a dependency?

9larsons:
Ok, clearly there's some kind of race condition with the config mocks and initializing the service. I will have to look into that when I can reprioritize this.

ghost/core/core/server/data/schema/schema.js
cmraible: Sorry, just seeing this — I think it resolves for both of us, but not sure! It's still showing unresolved for me FWIW.

9larsons:
@vikaspotluri123 Hey, it seems like the `ghost-cli` tests are hanging on restart with this. Do you have any suggestions on how to repro/troubleshoot that? Only thing I can guess is that the job queue service is hanging. It doesn't do that to me locally.

9larsons:
Note to self: need to update migration to new version before merging.

vikaspotluri123:
It took me a minute, but taking a look 👀

https://github.com/vikaspotluri123/Ghost/actions/runs/11393243539/job/31701108553?pr=14

vikaspotluri123:
```patch
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index eec2c09aeda..d68f13ce3a2 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -861,7 +861,8 @@ jobs:
 
       - name: Update from v4
         run: |
-          ghost update -f -d $V4_DIR --archive $(pwd)/ghost/core/ghost.tgz
+          ghost update -f -d $V4_DIR --archive $(pwd)/ghost/core/ghost.tgz --no-restart
+          ghost run -d $V4_DIR -D
 
       - name: Save Ghost CLI Debug Logs
         if: failure()
@@ -879,7 +880,8 @@ jobs:
         run: |
           DIR=$(mktemp -d)
           ghost install local -d $DIR
-          ghost update -d $DIR --archive $(pwd)/ghost/core/ghost.tgz
+          ghost update -d $DIR --archive $(pwd)/ghost/core/ghost.tgz --no-restart
+          ghost run -d $DIR -D
 
       - name: Print debug logs
         if: failure()
```

I hacked the job like that and found this

```
[2024-10-18 00:32:03] ERROR Unhandled rejection: insert into `jobs` (`created_at`, `finished_at`, `id`, `metadata`, `name`, `queue_entry`, `started_at`, `status`, `updated_at`) values ('2024-10-18 00:32:03', NULL, '6711ac8389d4880ed0fcad4e', NULL, 'members-migrations', NULL, NULL, 'queued', '2024-10-18 00:32:03') - SQLITE_ERROR: table jobs has no column named metadata

insert into `jobs` (`created_at`, `finished_at`, `id`, `metadata`, `name`, `queue_entry`, `started_at`, `status`, `updated_at`) values ('2024-10-18 00:32:03', NULL, '6711ac8389d4880ed0fcad4e', NULL, 'members-migrations', NULL, NULL, 'queued', '2024-10-18 00:32:03') - SQLITE_ERROR: table jobs has no column named metadata
Error Code: 
    SQLITE_ERROR
----------------------------------------
Error: insert into `jobs` (`created_at`, `finished_at`, `id`, `metadata`, `name`, `queue_entry`, `started_at`, `status`, `updated_at`) values ('2024-10-18 00:32:03', NULL, '6711ac8389d4880ed0fcad4e', NULL, 'members-migrations', NULL, NULL, 'queued', '2024-10-18 00:32:03') - SQLITE_ERROR: table jobs has no column named metadata

[2024-10-18 00:32:03] INFO Ghost URL Service Ready in 3.509s
```

9larsons:
Ah awesome, thank you. I suppose that's a good thing! I knew the migration needed updating but it's nice to know that it'll fail in CI. Should've thought of that.

cmraible (approved):
Migration looks good to me! Have some more general thoughts which I'll share in slack, but definitely nothing blocking — looking forward to trying this out in staging 🚀

PR #21025, opened 2024-09-17
Updated activitypub app to utilise new activities endpoint

mike182uk:
refs [AP-377](https://linear.app/tryghost/issue/AP-377/inbox-returning-33mb-of-data), [TryGhost/ActivityPub#40](https://github.com/TryGhost/ActivityPub/pull/40)

Updated activitypub app to utilise new activities endpoint which returns a paginated list of activities

### DIFF
--- a/apps/admin-x-activitypub/src/api/activitypub.ts
+++ b/apps/admin-x-activitypub/src/api/activitypub.ts
@@ -160,4 +160,52 @@ export class ActivityPubAPI {
         const url = new URL(`.ghost/activitypub/actions/unlike/${encodeURIComponent(id)}`, this.apiUrl);
         await this.fetchJSON(url, 'POST');
     }
+
+    get activitiesApiUrl() {
+        return new URL(`.ghost/activitypub/activities/${this.handle}`, this.apiUrl);
+    }
+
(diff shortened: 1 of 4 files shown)


apps/admin-x-activitypub/src/api/activitypub.ts
```diff
@@ -160,4 +160,52 @@ export class ActivityPubAPI {
...
+                return [];
+            }
+
+            // If the response has an items property, but it's not an array
+            // use an empty array
+            const items = Array.isArray(json.items) ? json.items : [];
+
+            // If the response has a nextCursor property, fetch the next page
+            // recursively and concatenate the results
+            if ('nextCursor' in json && typeof json.nextCursor === 'string') {
```
allouis, line 190: Will we do this dynamically eventually - or the plan is to stick with loading everything up front?

allouis (approved):

apps/admin-x-activitypub/src/api/activitypub.ts
mike182uk: Yeh it will be dynamic eventually, but for now we have no frontend pagination mechanism, so we load everything all at once (current behaviour)

PR #21055, opened 2024-09-20
✨Sodo search translation helper added, strings wrapped, ready for translator work!

cathysarisky:
- [X] The build will pass (run `yarn test:all` and `yarn lint`)

Translation (t) helper added to sodo-search.
Ghost head tweaked to include data-locale.
All (I hope) strings in sodo-search wrapped in the t helper.

Possibly poor-quality French translation strings added.


cathysarisky:
OK, so it won't quite pass automated tests, because the snapshot needs to be updated to include a locale in this line:
`    <script defer src=\\"https://cdn.jsdelivr.net/ghost/sodo-search@~[[VERSION]]/umd/sodo-search.min.js\\" data-key=\\"xyz\\" data-styles=\\"https://cdn.jsdelivr.net/ghost/sodo-search@~[[VERSION]]/umd/main.css\\" data-sodo-search=\\"http://localhost:65530/site/\\" crossorigin=\\"anonymous\\"></script>`

I don't know enough about Ghost's testing setup to know whether I'm supposed to edit the snapshot file directly, or if there's a command that generates those, so I'm going to leave it alone and hope someone will help me out here! :)

vikaspotluri123:
If you run `UPDATE_SNAPSHOT=1 yarn test` that should update your snapshots!

ghost/i18n/package.json
```diff
@@ -13,11 +13,11 @@
     "lint:code": "eslint *.js lib/ --ext .js --cache",
     "lint": "yarn lint:code && yarn lint:test",
     "lint:test": "eslint -c test/.eslintrc.js test/ --ext .js --cache",
-    "translate": "yarn translate:ghost && yarn translate:portal && yarn translate:signup-form && yarn translate:comments && node generate-context.js",
-    "translate:ghost": "NAMESPACE=ghost i18next '../core/core/{frontend,server,shared}/**/*.{js,jsx}'",
+    "translate": "yarn translate:ghost && yarn translate:portal && yarn translate:signup-form && yarn translate:comments && yarn translate:search && node generate-context.js",    "translate:ghost": "NAMESPACE=ghost i18next '../core/core/{frontend,server,shared}/**/*.{js,jsx}'",
```
vikaspotluri123, line 16:
```suggestion
    "translate": "yarn translate:ghost && yarn translate:portal && yarn translate:signup-form && yarn translate:comments && yarn translate:search && node generate-context.js",
    "translate:ghost": "NAMESPACE=ghost i18next '../core/core/{frontend,server,shared}/**/*.{js,jsx}'",
```

ghost/core/core/frontend/helpers/ghost_head.js
```diff
@@ -86,7 +86,8 @@ function getSearchHelper(frontendKey) {
     const attrs = {
         key: frontendKey,
         styles: stylesUrl,
-        'sodo-search': adminUrl
+        'sodo-search': adminUrl,
+        locale: settingsCache.get('lang') || 'en'
```
ErisDS, line 90: This key needs to be `locale` rather than `lang` please.

ErisDS (changes requested):
I gave this a quick spin and I believe I've found one thing that needs updating.

Let me know if you don't have time this week and I'll update it. 

Other than that this is pure gold! Thank you so much 🙏

ghost/core/core/frontend/helpers/ghost_head.js
cathysarisky: Yeah, I think I can change one word today. Maybe while the tea is steeping. ;)

cathysarisky:
Does the version of search also need to be bumped, analogous to portal?

ErisDS (approved):

ErisDS:
It will need bumping yes, but shipping and bumping the packages is something we need to handle as part of the release process, rather than something you're meant to do :)

PR #21062, opened 2024-09-21
🇧🇬 Bulgarian translation updates (Sep 2024)

yovko:
A needed update of the Bulgarian version. Fixing some errors. Made some improvements. And some new strings were translated.


StanTraykov:
Hi! I am not the original BG translator (I think it may have been an AI-assisted translation) but recently contributed some fixes (correcting mistakes like "провете" > "проверете", etc.). I also added a translation for the comments UI and just now for the search UI (search just became possible; PR is not merged yet, it's at #21102).

If you're actively using Ghost in BG, maybe we could coordinate/discuss the translation? There are some remaining issues with the original one but I haven't had the time to look at it more carefully. I just fixed what seemed critical: stuff that I often see. (Happy to replace any of the strings added by me, if there's a better fit.)

I noticed some issues with this PR:
- it reintroduces the misspelling of проверете as "провете"
- it spells "преди X дни" (часа, минути, etc.) beginning with a lowercase letter. This is fine by itself, of cоurse, but inconsistent with the rest, see e.g. line 43 ("One day ago", which this PR translates with an initial capital letter)
- it uses the imperative (заповедна форма) for some commands instead of just naming the action (what corresponds to the infinitive in other languages, more or less): "покажи" instead of "показване". I consider naming/infinitive to be correct as it's what used for UI elements in other languages I sometimes translate to/from (German). But both styles are definitely to be found in Bulgarian language software. Googling it, I found [this post about it](https://bglocalize.wordpress.com/%D0%BF%D0%BE%D0%B2%D0%B5%D0%BB%D0%B8%D1%82%D0%B5%D0%BB%D0%BD%D0%BE-%D0%BD%D0%B0%D0%BA%D0%BB%D0%BE%D0%BD%D0%B5%D0%BD%D0%B8%D0%B5-%D0%B2-%D0%BF%D0%BE%D1%82%D1%80%D0%B5%D0%B1%D0%B8%D1%82%D0%B5%D0%BB/). Do you have strong opinions on the matter?

yovko:
Hi, @StanTraykov!

First of all, I'm glad that someone else has committed to translating Ghost into Bulgarian - if only for the opportunity to discuss some of the nuances such as those described above. It's good to label yourself on the forum - or at least that was part of the original etiquette we were asked to follow at the beginning - I wonder if it's still relevant. But at the very least, it's useful for not getting double work.

Otherwise, on your notes...

- "провете" is an error on my part - of course, there's nothing to discuss
- "преди X дни" is because of the context in which it appears - but I'll capitalize it back (reserving the right in the future to raise this again for discussion if necessary)
- regarding the imperative vs naming, I don't have a strong opinion here...

Overall, I follow the principle that initially, it's better to have some translation than none, and then "on second reading", whatever isn't okay - to be additionally improved.

I'll change my PR accordingly...

StanTraykov:
Just had a more thorough look now, @yovko. I think the bulk of the changes is an improvement (incl. to some of my more awkward strings) & also adding some missing ones, so I hope they marge it. I visited the forum some time ago but haven't really read much. Should we discuss there? I'm https://forum.ghost.org/u/stanislav/.

Right now, action buttons/links are a mix of imperative / descriptive naming. I plan to change all to descriptive at some point for consistency, unless there are objections. E.g. if it's "скриване" and "скриване на коментара" then it should also be "показване" not "покажи". (Alternatively all can be imperative, but I think this should be reserved for actual prompts/instructions to the user, and descriptive is better.)

yovko:
Well... I think when discussing a specific case, it's probably still better to be under the specific PR, because it's clearer somehow.

ronaldlangeveld:
A huge thank you for your Bulgarian Ghost translations! 🙏🏻 We’ve recently added missing i18n strings for search and Portal 🔍, so you might find a few new strings needing translations. Please report any gaps, and we’ll fix them promptly! 🎉

StanTraykov:
@ronaldlangeveld Thanks! The search translation is also done, in a separate PR: #21102

Issue #19839, opened 2024-03-12
Error updating to Ghost 5.80 from 5.23, blog down, please help


DigitalLeaves:
### Issue Summary

Hello, I just updated my Ghost blog from 5.23 to 5.80.
After the update, I tried to restart Ghost, but it fails with this message:

Ghost instance is not running! Starting...
✖ Starting Ghost
A GhostError occurred.

```
Message: Ghost was able to start, but errored during boot with: insert into `collections_posts` (`collection_id`, `id`, `post_id`, `sort_order`) select '65f02f0dafb7eeb96c4c90a6' as `collection_id`, '65f02f0dafb7eeb96c4c90a8' as `id`, '637a812a81cfb759fad27265' as `post_id`, 0 as `sort_order` union all select '65f02f0dafb7eeb96c4c90a6' as `collection_id`,
... (many more rows) ...
, '65314decb0c970c6a6eca545' as `post_id`, 0 as `sort_order` - SQLITE_ERROR: too many terms in compound SELECT
Context: [object Object]
Help: Error occurred while executing the following migration: 2023-07-10-05-16-55-add-built-in-collection-posts.js
```

Debug Information:
    OS: Debian GNU/Linux, v11
    Node Version: v18.19.1
    Ghost Version: 5.80.2
    Ghost-CLI Version: 1.25.3
    Environment: development
    Command: 'ghost restart'
    mysql Ver 8.0.27 for Linux on x86_64 (MySQL Community Server - GPL)

### Steps to Reproduce

Start with an install of v5.23 or below 5.80
Apparently the number of posts can be a cause, so get more than 500 posts.
Run ghost update and ghost restart

### Ghost Version

5.80.2

### Node.js Version

v18.19.1

### How did you install Ghost?

Command line

### Database type

MySQL 5.7

### Browser & OS version

Chrome, MacOS

### Relevant log / error output

```shell
Ghost instance is not running! Starting...
✖ Starting Ghost
A GhostError occurred.
 
[...]

PR #21063, opened 2024-09-21
🐛 Fixed migrations for SQLite database users (#19839)

markstos:
SQLite has limit of 500 items in a compound select statement.

This limit could be hit when a complex select statement was being generated
as part of a batch insert statement.

Lowering the batch size will have minimal impact on migration performance
while improving SQLite compatibility.

One of these bulk inserts is confirmed to be affected through the linked issue.
I did confirm if the other two cases would trigger it, but the change won't hurt there
either.

Ref: https://www.sqlite.org/limits.html
Issue: #19839

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
+++ b/ghost/core/core/server/data/migrations/versions/5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js
@@ -14,7 +14,7 @@ const insertPostCollections = async (knex, collectionId, postIds) => {
         };
     });
 
-    await knex.batchInsert('collections_posts', collectionPosts, 1000);
+    await knex.batchInsert('collections_posts', collectionPosts, 100);
 };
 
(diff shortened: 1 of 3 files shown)


stromfeldt:
Just thought I'd add that I ran into this problem when updating a local instance from 5.53.3 to 5.96.0, with about 600 posts. Changing the value in .../5.55/2023-07-10-05-16-55-add-built-in-collection-posts.js from 1000 to 450 didn't work, but when I changed it to 100, as @DigitalLeave suggested, it then worked. Supposing I'm correctly following what I did.

allouis:
The default for `chunkSize` for `batchInsert` is 1000 - but this PR only updates values where they've been set explicitly.

It seems like 450 isn't low enough to fix the problem either!

I'm going to close this PR as it doesn't seem particularly active - if someone wants to fix it, then please check out the original issue where I'll leave some more information.

allouis:
@markstos I wasn't sure if you have permissions to reopen this - so I've done it for you!

markstos:
Commit pushed to lower batch size to 100.

PR #21093, opened 2024-09-24
Updated Gaelic portal / search translations

angeidheal:
Updated the Gaelic translation with the new portal and search strings.


cathysarisky (approved):
This passes my heavily-AI assisted review.  (Not a Gaelic speaker.)

cathysarisky:
> @cathysarisky 😂 Out of curiosity, which AI did you use? I haven't really used an AI before, but I'd be interested in seeing how it does with Gaelic.

Mostly Google Translate.  I would have used DeepL, but it doesn't seem to speak Gaelic either!   I'm mostly checking that these translations don't say anything rude or utterly nonsensical, making them better than the nothing we currently have!  Any detailed tweaks for style will clearly need to come from other Gaelic speakers, but getting them available in Ghost seems like a good way to get more people looking at them and hopefully improving them. :)

angeidheal:
> Mostly Google Translate. I would have used DeepL, but it doesn't seem to speak Gaelic either! I'm mostly checking that these translations don't say anything rude or utterly nonsensical, making them better than the nothing we currently have! Any detailed tweaks for style will clearly need to come from other Gaelic speakers, but getting them available in Ghost seems like a good way to get more people looking at them and hopefully improving them. :)

Ah, OK. GT isn't the best with Gaelic TBH, although it has gotten better. I'm one of the Editors / Contributors to the Gaelic translation of Wordpress, so I've tried to keep the Ghost translation consistent with it. Agreed, sometimes the best way is to get it out and in front of users - I actually found a typo I'd missed in my original translation while I was adding these new strings.

cathysarisky:
> GT isn't the best with Gaelic TBH

I'd welcome a better recommendation if you have one!  I wouldn't use GT to produce translations, but I'm hoping it is at least adequate to check for bad behavior. :)

ronaldlangeveld:
Thank you for your translation work on Ghost! 🙏🏻 We’ve recently added more search and Portal i18n strings 🔍. Some new strings may need attention, so give them a look in Gaelic. If anything is missing, let us know and we’ll address it right away! 🎉

PR #21099, opened 2024-09-24
Updated search to be dynamic in admin-x-activitypub app

mike182uk:
refs [AP-352](https://linear.app/tryghost/issue/AP-352/search-for-mastodon-usernames-in-ghost-admin)

Updated the search functionality in the admin-x-activitypub app to be dynamic and utilise the search API endpoint provided by the activitypub service


apps/admin-x-activitypub/package.json
```diff
@@ -70,5 +70,8 @@
         ]
       }
     }
+  },
+  "dependencies": {
+    "use-debounce": "10.0.3"
   }
```
mike182uk, line 76: @allouis @djordjevlais Sus that this is the only non dev dependency 🤔 Have I misunderstood how we are using this here (i.e we just stick everything in `devDependencies`)?

apps/admin-x-activitypub/src/components/inbox/FollowSiteModal.tsx
```diff
@@ -1,28 +1,10 @@
...
+
 import NiceModal from '@ebay/nice-modal-react';
-import {ActivityPubAPI} from '../../api/activitypub';
 import {Modal, TextField, showToast} from '@tryghost/admin-x-design-system';
-import {useBrowseSite} from '@tryghost/admin-x-framework/api/site';
-import {useMutation} from '@tanstack/react-query';
 import {useRouting} from '@tryghost/admin-x-framework/routing';
-import {useState} from 'react';
 
-function useFollow(handle: string, onSuccess: () => void, onError: () => void) {
```
mike182uk, line 9: Moved this out to `hooks/useActivityPubQueries.ts` so it can also be utilised by `components/global/FollowButton.tsx`

apps/admin-x-activitypub/src/components/search/ProfileSearchResultModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+                    <div className='col-[3/4] flex items-center justify-end space-x-6 px-8'>
+                        <Button icon='close' size='sm' unstyled onClick={() => modal.remove()}/>
+                    </div>
+                </div>
+            </div>
+            <div className='mt-10 w-auto'>
+                <div className='mx-auto max-w-[580px] pb-16'>
+                    <div className='z-0 flex w-full flex-col items-center'>
+                        <div className='mx-auto w-full max-w-[560px]'>
+                            <div className='h-[200px] w-full rounded-lg bg-gradient-to-tr from-grey-200 to-grey-100'>
```
mike182uk, line 44: @djordjevlais There might be a few too many divs nested here 😅 Think this can probably be refactored

apps/admin-x-activitypub/package.json
allouis: Yeah kinda weird, but I guess because we build a bundle with all the deps, we technically only needed them during build time which will have devDeps. Happy for them to be moved, I think it's weird.

mike182uk: 👍 I've moved anything thats strictly not development related (i.e testing shtuff) to `dependencies`

PR #21100, opened 2024-09-24
Update pt-BR translation of search.json

Lekler:
Following commit #21055, the following updates were made:

Added translations for terms related to authors, posts, search, and tags.


cathysarisky (approved):
Thank you for submitting these!

An AI-assisted check says these are lots better than blanks. :) I think they're good to merge unless someone who speaks Brazilian Portuguese wants to suggest improvements.

PR #21102, opened 2024-09-24
add Bulgarian translation for search

StanTraykov:
Add initial translation for Bulgarian for the search interface.


StanTraykov:
@cmraible 👋 Just added a Bulgarian translation for the search interface.

yovko:
I would suggest to translate tags as "етикети" ("тагове" is not yet an accepted word in the official dictionary, if I'm not mistaken - and perhaps it's better not to impose it) :-)

StanTraykov:
@yovko thanks, changed it to етикети to sound more natural (or maybe old-fashioned to the new generation? I don't know :D).

ronaldlangeveld:
Thank you for submitting translations to Ghost! 🙏🏻 We really appreciate every string you’ve added. There might be a few new strings that need translation in Bulgarian. If you spot any untranslated or missing strings, let us know, and we’ll fix them ASAP! 🎉

PR #21130, opened 2024-09-26
Admin X activitypub app independent release

allouis:
ref https://linear.app/tryghost/issue/AP-438

We want to have continuous deployment (or thereabouts) for the activitypub frontend, and to do that we're going to load it from the jsdelivr cdn, and deploy via npm package updates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -965,3 +965,63 @@ jobs:
           inputs: '{"version":"canary","environment":"staging","version_extra":"${{ needs.job_setup.outputs.branch_name }}"}'
           wait-for-completion-timeout: 25m
           wait-for-completion-interval: 30s
+
+  publish_admin_x_activitypub:
+    needs: [
+      job_setup,
+      job_lint,
+      job_unit-tests
+    ]
+    name: Publish @tryghost/admin-x-activitypub
+    runs-on: ubuntu-latest
+    if: always() && needs.job_setup.result == 'success' && needs.job_lint.result == 'success' && needs.job_unit-tests.result == 'success' && needs.job_setup.outputs.is_main == 'true'
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18.12.1'
+
+      - name: Restore caches
+        uses: ./.github/actions/restore-cache
+        env:
+          DEPENDENCY_CACHE_KEY: ${{ needs.job_setup.outputs.dependency_cache_key }}
+
+      - name: Build the package
+        run: yarn run nx build @tryghost/admin-x-activitypub
+
+      - name: Check if version changed
+        id: version_check
+        working-directory: apps/admin-x-activitypub
+        run: |
+          CURRENT_VERSION=$(cat package.json | jq -r .version)
+          PUBLISHED_VERSION=$(npm show @tryghost/admin-x-activitypub version || echo "0.0.0")
+          echo "Current version: $CURRENT_VERSION"
+          echo "Published version: $PUBLISHED_VERSION"
+          if [ "$CURRENT_VERSION" = "$PUBLISHED_VERSION" ]; then
+            echo "Version is unchanged."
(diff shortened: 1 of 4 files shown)


ghost/admin/lib/asset-delivery/index.js
```diff
@@ -43,6 +43,8 @@ module.exports = {
                 for (const [key, value] of Object.entries(this.packageConfig)) {
                     console.log(`Asset-Delivery: ${key} = ${value}`);
                 }
+
+                this.packageConfig[`adminXActivitypubCustomUrl`] = 'https://cdn.jsdelivr.net/npm/@tryghost/admin-x-activitypub@0.0.2/dist/admin-x-activitypub.js'
```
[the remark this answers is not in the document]
kevinansfield: Normal practice is to [set it Ghost's default config](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/shared/config/defaults.json#L183-L203) with a `minor` version listed so patches are automatically picked up immediately without having to ship Admin but you can still ship a minor to delay it being picked up before the next Ghost release.

What was the reason for trying to hardcode it in Admin? If you do that you'd need to rebuild and ship Ghost each time you want to bump the activity pub package.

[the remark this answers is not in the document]
kevinansfield: > I tried added a new config property to config/environment.js but I get build errors, and I'm also not certain on where the environment variables are set for production builds.

What build error did you get? It should be possible to add a key to the `ENV` object without any issues 🤔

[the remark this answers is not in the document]
allouis: RE: The version - that's the plan! This was just my WIP to get things working.

So does the Admin have access to the ghost config at build time? I didn't realise if that's the case!

The reason for hardcoding it was to just get things working and prove the concept

.github/workflows/ci.yml
```diff
@@ -965,3 +965,59 @@ jobs:
...
+        uses: actions/cache@v4
+        with:
+          path: |
+            ${{ github.workspace }}/apps/admin-x-activitypub/node_modules
+          key: admin-x-activitypub-${{ github.sha }}
+          restore-keys: |
+            admin-x-activitypub-
+
+      - name: Install dependencies
+        run: yarn install --prefer-offline --frozen-lockfile
```
daniellockyer, line 997: You should be able to re-use our cache here: https://github.com/TryGhost/Ghost/blob/5579f03fd5fcc13c308f8c57c2e448dc10ad8e07/.github/workflows/ci.yml#L261-L264

It includes the `node_modules` for all packages but it's much faster than a `yarn install`

.github/workflows/ci.yml
```diff
@@ -965,3 +965,59 @@ jobs:
...
+            echo "Version has changed."
+            echo "version_changed=true" >> $GITHUB_ENV
+          fi
+
+      - name: Publish to npm
+        if: env.version_changed == 'true'
+        working-directory: apps/admin-x-activitypub
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npm publish --dry-run --access public
```
daniellockyer, line 1023: One thing we might want to do after this is to auto-purge the jsDelivr cache

allouis: Yeah, I was thinking this too

ghost/admin/lib/asset-delivery/index.js
```diff
@@ -43,6 +43,8 @@ module.exports = {
                 for (const [key, value] of Object.entries(this.packageConfig)) {
                     console.log(`Asset-Delivery: ${key} = ${value}`);
                 }
+
+                this.packageConfig[`adminXActivitypubCustomUrl`] = 'https://cdn.jsdelivr.net/npm/@tryghost/admin-x-activitypub@0.x/dist/admin-x-activitypub.js'
```
daniellockyer, line 47: Does this work? 🤔 I thought it had to be `...@~0.0/..`

allouis: It is working - the docs say it should be `@0` or `@0.0` though - I'm gonna change it to that

PR #21144, opened 2024-09-26
🐛🌐 add missing language codes from i18n.js, regenerate strings.

cathysarisky:
Apparently several PRs for new languages were merged without the corresponding additions to i18n.js, which has caused new strings not to be generated for those languages, and missing translations in the frontend, even though we had them in the .json file. 

I've added language codes for all languages that had at least some strings translated in i18n/locales, and have run yarn translate to create missing strings and files for these languages.

New language codes for:
* Arabic
* Swiss German
* Greek
* Estonian
* Macedonian
* Serbian (Cyrillic)
* Thai


ghost/i18n/lib/i18n.js
```diff
@@ -2,15 +2,19 @@ const i18next = require('i18next');
...
 const SUPPORTED_LOCALES = [
     'af', // Afrikaans
+    'ar', // Arabic
     'bg', // Bulgarian
     'bs', // Bosnian
     'ca', // Catalan
     'cs', // Czech
     'da', // Danish
     'de', // German
+    'de-CH', // Swiss German
```
JohnONolan, line 12: not 100% sure - but do these need to be all lowercase? (also ln46)

cathysarisky: Looks to me like we're following https://en.wikipedia.org/wiki/IETF_language_tag.  Whether that was an explicit decision between multiple conventions or not, I don't know.

The codes that are new to the file are consistent with pt-BR (Portuguese as spoken in Brazil), which has been there for quite a while.  While we could decide to change conventions, that would require changing names of directories, and some sort of migration strategy for sites with pt-BR, which has been there in that form for quite some time.

So, no, they don't, and let's not make work for ourselves. :)  Re sr-Cyrl (and zh-Hant, not new), that's correct.  Script designators are Titlecase, while country designators are UPPERCASE, and language codes are lowercase.

package.json
```diff
@@ -124,5 +124,6 @@
     "rimraf": "5.0.10",
     "ts-node": "10.9.2",
     "typescript": "5.4.5"
-  }
+  },
+  "packageManager": "yarn@1.22.21+sha1.1959a18351b811cdeedbd484a8f86c3cc3bbaf72"
```
ErisDS, line 128: Heyo @cathysarisky do you know where this has snuck in from? Can we remove it for now?

If it's something super useful for a common tool we can discuss and add separately :)

cathysarisky: I apologize. That snuck past me.  Will fix!

PR #21151, opened 2024-09-28
✨Add content_api_key helper

cathysarisky:
This very small helper adds {{contentkey}} to the collection of handlebars helpers available to theme creators.  
This will make it easier for themes to access the content API key, without either requiring the user to get it from the integrations page and input it on the theme setting page or resorting to JavaScript to read it from one of the built-in script tag attributes -- both ugly workarounds.


cathysarisky:
> @cathysarisky Have you tested a theme with this helper with `gscan`? Also could you expand on the use case? Helpers are meant to provide a means to interact with the content API w/o needing to do so directly.

Ohh... I bet I need to patch gscan, too.  Will check, thank you. 

I've had occasions where I wanted to access the API with something other than a built-in script, such as a 'dashboard' (with extra user data over in AWS) that needed to get the URLs of featured images given the post name.  

Currently, some paid themes that use infinite scroll or client-side filtering of posts require the user to input their content API key as a theme setting.  This will give them access to it directly in the theme instead.  I've got a theme customization project to do this week where the client wants to be able to filter by multiple tags and authors.  I definitely need to talk to the content API for that one, and its silly that the client needs to copy the API key from one place to another, or else I need to pull it out of the <script> tag from a built-in helper.

I'm also laying groundwork for being able to load alternate versions of the comments/portal/search code (beyond what is supported for self-hosters in config.[*].json), or any other external integration that needs access to the content API.

cathysarisky:
> @cathysarisky Have you tested a theme with this helper with `gscan`?

No user-visible errors on uploading - apparently the check Ghost does on theme upload doesn't include checking for non-existent helpers.

Passes gscan.

cathysarisky:
I've added a test, and verified that gscan doesn't need updating.  Thanks for your comments @9larsons !

9larsons (approved):
LGTM!

thimiraonline:
How to use this and what are the uses of the helper?

cathysarisky:
> How to use this and what are the uses of the helper?

I wrote a teeny blog post about it: https://www.spectralwebservices.com/blog/introducing-a-new-helper/

But basically, you put {{content_api_key}} in your .hbs code and it writes out the content API key.  Useful if your theme has javascript that needs to use the api key.  

I'd imagine using it for loading your own copy of portal or search or another integration, in combination with #21229.

PR #21152, opened 2024-09-28
🌐⬅️add RTL to sodo-search & improved tests

cathysarisky:
🌐Patch sodo-search to work with RTL languages also. -- Cathy Sarisky

Add dir prop, calculated by i18next from language (using the dir function)
Tweak a few styles to use me/ms/pe/ps instead of mr/ml/pr/pl

Suggestions for a better way to get rtl into the iframe html welcome.  This is what I came up with.

Added updated test that checks that stemming works in English, and added tests for partial and full-word searching with RTL content.


9larsons:
LGTM! Could you add some tests? Or I can, but we won't want to merge this until we have something covering it.

cathysarisky:
> LGTM! Could you add some tests? Or I can, but we won't want to merge this until we have something covering it.

Happy to, but I could use a little help thinking through that.  Sorry, I'm a bit of a noob. :)

PR #21164, opened 2024-09-30
[DRAFT] Translate Portal error/status messages (at least the common ones?)

cathysarisky:
First draft (working) of an idea for how to translate API responses that end up being member-facing in Portal.  I don't love it.

### DIFF
--- a/ghost/i18n/locales/de/search.json
+++ b/ghost/i18n/locales/de/search.json
@@ -1,9 +1,9 @@
 {
-    "Authors": "",
-    "Cancel": "",
-    "No matches found": "",
-    "Posts": "",
-    "Search posts, tags and authors": "",
-    "Show more results": "",
-    "Tags": ""
+    "Authors": "Autor*innen",
+    "Cancel": "Abbrechen",
+    "No matches found": "Keine Treffer gefunden",
+    "Posts": "Posts",
+    "Search posts, tags and authors": "Posts, Tags und Autor*innen durchsuchen",
+    "Show more results": "Mehr Ergebnisse anzeigen",
+    "Tags": "Tags"
 }


apps/portal/src/utils/errors-for-translation.js
9larsons, line 1: How did you generate this list? Looks like Portal also has its own strings like [here](https://github.com/TryGhost/Ghost/blob/80794486a92b908d7437ed87de99b9b41d4fcf85/apps/portal/src/utils/api.js#L44), for example.

apps/portal/src/utils/errors.js
9larsons, line 1: Having to pass in `t` every time this is called is laborious. I think we'd want to store the language in global context and then pull it into this, which might require a bit of reworking, but it'd be far cleaner. (I know this is a draft, more of a bigger picture note 🙂).

9larsons, line 1: This class is also often used with the other method, `fromApiResponse`. We'd similarly need to update that. 

The more I'm looking at this the more we may want to rework how errors are handled within Portal. It'd be a larger task though worthier of the effort if we're about to spend the resources getting translations for them all.

apps/portal/src/utils/errors-for-translation.js
cathysarisky: Those are strings that I found were showing up directly from the API.  It's a dumb hack to get the 'yarn translate' to pick them up.  It's possibly not complete, but at least contains the worst offenders.  [At the point that the site is totally broken, I don't care if they're getting English, but "hey, you need to subscribe, not sign in" is something that users trigger all the time, so we should get it right...]

apps/portal/src/utils/errors.js
cathysarisky: Yeah, I was struggling to get context into utility functions without passing it... I'm not as proficient with React as I'd like to be, and it's hard dropping into the middle of a bunch of code this big and written with a mix of styles...

9larsons: Portal is... challenging to work with 😓. The main issue is this is just a class and we would need to use React hooks to be able to import the app context/state. I can handle that refactoring if we want to keep going down this path. I wish it was easier to use forked branches to collaborate.

cathysarisky:
Yep, that's where I got stuck! I've just sent you an invite to collaborate
on the fork, if that works.  I'd be grateful for the help.

apps/portal/src/utils/errors.js
cathysarisky: Yep, that's where I got stuck! I've just sent you an invite to collaborate
on the fork, if that works.  I'd be grateful for the help.

PR #21165, opened 2024-09-30
Add and adjust various German translations in `portal.json`

lsinger:

This PR adds and adjusts various German translations in `portal.json`. I'm a native speaker and have made adjustments to the best of my knowledge. As in #21163, the gender variations are a bit opinionated and I'm happy to discuss.


cathysarisky (approved):
from thewolfofallstreetz (Ghost forum): the (green) translations look good to me.

Only minor comment on 2: “(save {{highestYearlyDiscount}}%)”: “(jetzt {{highestYearlyDiscount}}% sparen)”

This string will show the highest discount of all the offered packages. So the discount will be “save up to” x%, but it could also be less.

To correctly express this in German you could say the following: (save up to {{highestYearlyDiscount}}%)": “(jetzt bis zu {{highestYearlyDiscount}}% sparen)”

My response: The string actually shows the price difference between monthly and yearly. So if you pick yearly, you’ll actually get the discount compared to monthly. I think "up to" is probably not needed here.

@lsinger, thanks for all your work to polish these! :)

PR #21166, opened 2024-09-30
[WIP] 🐛 Fix origin attribution in Tinybird analytics hits

alejandromav:
## Description
The attribution of the traffic source in the Tinybird Analytics data project was not correctly implemented, as reported by @ErisDS. We were only counting the first hit, a not all the hist within the same session. Here's an example:

**Before**
image

image

**After**
image

image

This change is tested in `all_top_sources.test`, and `all_top_sources.test.result` was updated accordingly.

## Technical details

We implemented this change in the attribution of the source of traffic in the `analytics_sources_mv` materialized view, while keeping the raw hist intact in the `analytics_hits` data source. Keeping these inmutable events as they are, and implemented logic in the materialized is considered a good practice. In the future, these attribution logic can be changed and apply it retroactively.

We also fixed a corner case in the `top_browsers` endpoint. If a user leaves a page open in a browser tab, the session expires after 30 minutes and then the users clicks an internal link, a new session_id will be generated. But since the first referral would be the same site, this session was not being taken into account. It's now fixed as well.


ErisDS:
I reordered things a bit so it was clearer what fixed what (I already merged your data changes and the filter tests) and then recreated this PR here; https://github.com/TryGhost/Ghost/pull/21187

PR #21169, opened 2024-10-01
🐛 Fixed errors creating new posts after a post access change via context menu

kevinansfield:
ref https://linear.app/tryghost/issue/ONC-323

After changing a post's access via the posts list context menu, creating new posts or members would not work correctly.

- the issue stemmed from `this.post.set('currentState.parentState.isNew', false);` that was called when changing a post's access level, after that all Ember Data models created from the store would have `isNew: false` causing Ember Data to attempt a PUT request to update the not-yet-created model rather than a POST request to create it
- we were only using a real post model instance in order to run validations against the post access level settings but we can do that just as easily by creating a new object and injecting our validation mixin


sagzy (approved):
LGTM 👍

PR #21190, opened 2024-10-02
🌐 Portal translations - make additional errors translatable.

cathysarisky:
Expose (some) Portal error strings for translations

💩This is a somewhat hacky (but test-passing and individual inspection passing) solution to the way Portal handles errors.  Or rather, the half-dozen ways Portal handles errors.

Passing 't' around with context and state, and occasionally recreating it from the site object.  Yes, I am also somewhat horrified, but a better implementation will need a major rewrite of Portal.

Addresses errors in both the popover React app and in the data-attributes.

There are probably more.  Since Portal exposes raw API responses in some places, it's hard to enumerate everything that /might/ end up being client-facing, but at least I've gotten the ones that I've commonly seen.

Improvements very welcome.


apps/portal/src/utils/errors.js
9larsons, line 1: I'm surprised this doesn't throw a linter error. In general there's a few whitespace changes that could be avoided. Not a big deal, just something to look out for down the road to help keep the history clean.

apps/portal/src/utils/errors.js
```diff
@@ -9,24 +9,110 @@ export class HumanReadableError extends Error {
...
+}
 
-    /**
-     * Only output the message of an error if it is a human readable error and should be exposed to the user.
-     * Otherwise it returns a default generic message.
-     */
-    static getMessageFromError(error, defaultMessage) {
-        if (error instanceof HumanReadableError) {
-            return error.message;
+// changing this - will export this helper.  It's not helping me to have it stuck to HumanReadableError...
```
9larsons, line 26: This could likely be cleaned up.

apps/portal/src/utils/errors.js
```diff
@@ -9,24 +9,110 @@ export class HumanReadableError extends Error {
...
+    } else {
+        // use the default message if there's no error message
+        // if we don't have a default message either, fall back to a generic message plus the actual error.
+        return alreadyTranslatedDefaultMessage || t('An error occurred:') + ' ' + error.toString();
     }
 }
+
+/* This fails linting, because it's only here so that these strings can get picked up for translation */
+// eslint-disable-next-line no-unused-vars
+const myfakeFunction = () => {
```
9larsons, line 101: Could you explain this? I'm not familiar with the parser so I'm just curious why this is necessary.

9larsons (changes requested):
I think you know what's coming 🙂. Could we get unit tests over the new content in `errors.js`?

apps/portal/src/utils/errors.js
cathysarisky: That makes two of us. The problem is that the parser seems to only pick up wrapped strings - it doesn't understand incoming variable responses from the API.  There's probably a more elegant way to do this, although I'm not sure what it is... guess I'd better dig into the parser documentation, huh?

apps/portal/src/utils/errors.js
cathysarisky: Yeah, sorry. Bit of a note to myself, followed by a now-inaccurate comment.  Will fix!

apps/portal/src/utils/errors.js
cathysarisky: The parser is pretty limited.  It finds things wrapped with t(' something ').  https://github.com/i18next/i18next-parser?tab=readme-ov-file#caveats 

I've adjusted how I do it, because it removes the need to maintain strings in two places.  I think it's better?

9larsons (approved):

9larsons:
Looks great, thanks! I'd like to revisit some time why we need to load the special messages. We don't need to worry about that here.

PR #21200, opened 2024-10-03
ActivityPub inbox layout and improved profiles

djordjevlais:
- Improved inbox layout for the home screen, optimized for long-form content
- Added `Show all` button when profile details are too long


apps/admin-x-activitypub/src/components/search/ProfileSearchResultModal.tsx
```diff
@@ -31,6 +31,21 @@ const ProfileSearchResultModal: React.FC<ProfileSearchResultModalProps> = ({
         .filter(attachment => attachment.type === 'PropertyValue');
     const posts = profile.posts; // @TODO: Do any filtering / manipulation here
 
+    const [expanded, setExpanded] = useState(false);
```
mike182uk, line 202: I think we would be better calling this `isExpanded` / `setIsExpanded` for consistency

PR #21201, opened 2024-10-03
Added Greek, Indian(Hindi), Malay, Dutch, Portuguese, Russian. Spanish , Thai to search bar

russkiimike:
Quickly created some translations with open ai gpt 4 mini  just for the search bars so translation quality is subjective, some of the languages that I know natively look good (Russian and Greek) to me. Any corrections are welcome :)


cathysarisky (changes requested):
We cannot merge a PR with inaccurate language codes.   Well, we "can", but then they'll be in the wrong folder, and they won't get picked up by the translation package for the language intended.  Here's a list:
https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes

Also note that we have Portuguese split into pt and pt-BR, so you may want to confirm which one you intended.

russkiimike:
I fixed the the codes and translations but the commit didnt show up in this thread ?

9larsons:
Hey @russkiimike it looks fine. Could you please remove the translations you are not a fluent speaker of, though? Ghost has quite the international user base and we'd prefer to source from native speakers. I'll update to the latest on main as we just pulled in a lot of translations.

fochoa8:
> Hey @russkiimike it looks fine. Could you please remove the translations you are not a fluent speaker of, though? Ghost has quite the international user base and we'd prefer to source from native speakers. I'll update to the latest on main as we just pulled in a lot of translations.

I got someone who can translate it to Portuguese. Russian, and Bengali. I give them props. Or get them to post by themselves.

cathysarisky (changes requested):
Renaming locales (pt -> pt-PT) will break every existing Ghost install that uses 'pt' as their locale.  Is there some compelling reason for this?

If it were going to happen, you'd have to update the locales file, too, but I think the better answer is: don't.

fochoa8:
> Renaming locales (pt -> pt-PT) will break every existing Ghost install that uses 'pt' as their locale. Is there some compelling reason for this?
> 
> If it were going to happen, you'd have to update the locales file, too, but I think the better answer is: don't.

I meant pt-br, sorry.

sagzy:
Hello @russkiimike, thanks for the contribution! 🙏 

As per @9larsons and @cathysarisky's reviews, could you please:
- open one PR per language
- only translate languages that you speak fluently (no AI-generated translations)
- not rename any locale files

Closing this PR in the meantime.

russkiimike:
Hi replying late, apologies! Will redo as per @sagzy request :)

PR #20631, opened 2024-07-19
Added hindi language translation

meSingh:
Added translations for Hindi language.

PR #21208, opened 2024-10-03
🌐 Hindi translations - original PR from meSingh, thank you! #20631

cathysarisky:
no ref

Hindi translations - original PR from meSingh, thank you! #20631


9larsons:
The problem seems to be that our test command runs `yarn translate`. I think this is a good thing, except that we're not catching it until it runs on CI, because it doesn't fail locally. Essentially, it's too easy to commit files from this repo and not have run `yarn test` or `yarn translate`. Probably less of an issue once we stop updating this frequently?

PR #21229, opened 2024-10-05
🎨 Added an "exclude" attribute to {{ghost_head}}

cathysarisky:
no ref

{{ghost_head}} is huge, and some power-users and theme creators want the ability to customize what it contains.  This PR makes it easier for a theme to write custom schema, or to load a custom version of portal/comments/search/etc, or to minimize load times by not loading scripts where they aren't needed, in a theme-specific way.  

Because ghost_head is controlled at the theme level, this gives folks in managed hosting the new ability to load a different version of the included app scripts (by preventing ghost_head from writing them and adding them in manually).

Usage example: ` {{ghost_head exclude="search,portal"}}  `
[lots more examples at https://ghost.slack.com/canvas/C018EKC56JF ]

(empty array)
	No changes to current behavior

search
	The built-in sodo-search script
	Includes adding the click event listener on buttons, generating the search index, and the UI.

portal
	The portal script
	Handles sign-in and sign-up, payments, tips, memberships, etc, and all the portal data-attributes.

announcement
	The announcement bar javascript
	If you'd like to use the announcement bar admin settings but not have it [mess up your CLS metric](https://www.spectralwebservices.com/blog/announcement-bar-a-review/), this is for you.

metadata
	Skips HTML tags for meta description, favicon, canonical url, robots, referrer
	Important for SEO

schema
	The LD+JSON schema
	Important for SEO

card_assets
	Loads cards.min.css and .js
	Needed on any page with a post body, unless your theme replaces them all. Assets can also be selectively loaded with the [card_assets override](https://ghost.org/docs/themes/content/?ref=spectralwebservices.com#editor-cards)

comment_counts
	Loads the comment_counts helper
	Needed if the page is using {{comments}} or data-ghost-comment-count attribute

social_data
	Produces the og: and twitter: attributes for social media sharing and previews
	Required for good social media cards

cta_styles
	Removes the call to action (CTA) styles
	Used for member signup and CTA cards - may be overwritten by your theme already


cathysarisky:
Yes, it's a whole bunch of 'how to break your site if you don't know what you're doing' options, but it also removes some limitations, especially for managed hosting users!

ghost/core/core/frontend/helpers/ghost_head.js
```diff
@@ -212,6 +220,37 @@ module.exports = async function ghost_head(options) { // eslint-disable-line cam
     const favicon = blogIcon.getIconUrl();
     const iconType = blogIcon.getIconType(favicon);
 
+    // loop over the head_excludes array and set the isUnwanted object to true for each item
+    if (options.data.config?.head_excludes) {
+        options.data.config.head_excludes.forEach((item) => {
+            isUnwanted[item] = true;
+        });
+    }
```
vikaspotluri123, line 228:
```suggestion
    const excludeList = new Set(options.data.config?.head_excludes ?? [])
```

cathysarisky: Thank you so much for your help here and on Slack!  Think I've got it all fixed up now!

cathysarisky:
Hey @JERXCHRS  - all the examples are over on my blog. https://www.spectralwebservices.com/blog/adding-new-features-to-ghost-head/  -- but I need to update, as exclude doesn't currently support a space between options.  (Needs to be exclude="portal,search" ... that's on the fix list!)

Right, all this does is remove the listed item, not alter it.  (More customizable schema would be awesome.  This gets you to a place where you can write your own without having weird duplications/contradictions, but it's probably not a full and easy solution to your problem.)  It is /technically/ possible to write schema in handlebars (I've done it), but yeah, it sucks.

PR #21238, opened 2024-10-06
🌐 RTL support for Portal

cathysarisky:
A draft of RTL support for Portal.


apps/portal/src/components/Frame.styles.js
```diff
@@ -748,16 +779,16 @@ const FrameStyles = `
...
-.mr3 { margin-right: 12px; }
-.mr4 { margin-right: 16px; }
-.mr5 { margin-right: 20px; }
-.mr6 { margin-right: 24px; }
-.mr7 { margin-right: 28px; }
-.mr8 { margin-right: 32px; }
-.mr9 { margin-right: 36px; }
-.mr10 { margin-right: 40px; }
+.mr1 { margin-inline-end: 4px; }
+.mr2 { margin-inline-endt: 8px; }
```
sagzy, line 783: Typo: `margin-inline-endt:` (extra `t`)

sagzy (commented):
Hey @cathysarisky, really cool! Left two comments

apps/portal/src/components/Frame.styles.js
cathysarisky: embarrassing! Thank you!

apps/portal/src/components/Frame.styles.js
```diff
@@ -183,7 +183,7 @@ const FrameStyles = `
 .gh-portal-btn-logout {
     position: absolute;
     top: 22px;
-    left: 24px;
+    right: 24px;
```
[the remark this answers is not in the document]
cathysarisky: agreed, that's an error.  (Although I don't actually see an impact of that change, it might be important at some screen size.)  Fixing...

sagzy (approved):
LGTM 👍 

Pending design review from @peterzimon

dvdwinden (changes requested):
Hi @cathysarisky, 

Thank you for submitting your PR! Much appreciated.

I reviewed it from a design point-of-view, and the CSS looks great. I did find two small issues when testing it.

First: when going to **Email Preferences**, the question mark for _Not receiving emails?_ sits in the wrong place. I believe this is a common issue when implementing RTL support, and I wonder if there's anything we can do to mitigate it?

image


Second: in the main **Account** modal, below the headline, there's a sentence about your current membership. On the second line of that sentence, the period is in the wrong place — similar to the question mark mentioned above. 

image

Could you have a look at these?

cathysarisky:
Thanks @dvdwinden for the review!  I think the period and question mark's positions not actually a problem.  They are in the "right place" -- it's just that your text isn't RTL (because it is English), so it looks weird.  

I'm not sure which language you have set.  (Maybe you forgot to turn on the portal translations in settings > labs?)  If you switch to Arabic, you should see the 'Not receiving emails?' rendered as: 
"Not receiving emails?": "لم تستلم رسالة بريدية؟",
Happily, the punctuation is wrapped in the string, so we get the backward question mark in the right place without portal doing extra work.
image

I switched over to Urdu for the other string, since translations are more complete there:
image

dvdwinden (approved):
@cathysarisky Excellent, thank you for clarifying — this looks great!

PR #21251, opened 2024-10-08
🐛 fixed routing error when no recommendations

### LINKED ISSUE #20771: Opening the recommendation popup only works with enabled subscription
### Issue Summary

Trying to open the recommendation pop up (https://example.org/#/portal/recommendations) - e.g. through the URL, or through the "View all" button on the homepage - doesn't do anything when `Subscription access` is disabled / set to "nobody".

### Steps to Reproduce

1. Set the Subscription access to `Nobody` (https://example.org/ghost/#/settings/members)
2. Enable the `publication info sidebar` at https://example.org/ghost/#/settings/design/edit - in the Tab Homepage.
3. Recommend a site (https://example.org/ghost/#/settings/recommendations)
4. See the recommendation panel. (https://example.org)
5. Try to click on "View all".

### Ghost Version

5.89.1

### Node.js Version

v18.20.4

Via Docker Compose

### Database type

MySQL 8

### Browser & OS version

Firefox 129.0.1 (64-Bit) on Windows 11 Pro

_No response_

### Code of Conduct

cathysarisky:
no ref.

The default routing for the portal app is a signup page.  When a site has no recommendations, the result of triggering that page (by following a link that goes there) is a prompt to sign up, or a notification that one cannot sign up, if membership is disabled.

This patch adds a "No recommendations" message, which will be shown if a user follows a recommendations link (#/portal/recommendations) on a site without recommendations.  While we shouldn't end up there very often, it'll make a lot more sense when it does! 

Correction: This fix does _not_ close #20771 - that's a separate bug!


apps/portal/src/App.js
```diff
@@ -854,7 +853,7 @@ export default class App extends React.Component {
             return {
                 page: 'supportError'
             };
-        } else if (path === 'recommendations' && hasRecommendations({site})) {
```
mike182uk, line 857: Ah, this the culprit for no the modal not triggering detailed in the referenced issue?

cathysarisky: I thought so, but on further investigation, I think he was reporting a separate bug in ghost_head that caused no loading of portal. (Separate PR filed.)

apps/portal/src/components/pages/RecommendationsPage.js
```diff
@@ -315,17 +315,29 @@ const RecommendationsPage = () => {
...
+    const heading = pageData && pageData.signup ? t('Welcome to {{siteTitle}}', {siteTitle: title, interpolation: {escapeValue: false}}) : t('Recommendations');
+
+    /* Possible cases: 
+    - no recommendations found - subhead says no recommendations are available.
+    - recommendations found - show generic message
+    - recommendations found and user just signed up - show specific message
+    */
+   
+    let subheading;
+    if (recommendationsEnabled && recommendations && recommendations.length > 0) {
```
mike182uk, line 331: Sorry, one more thing @cathysarisky would we be better initialising recommendations as `[]` instead of `null`? Then we can just check on `.length` to know if there are any recommendations or not 🤔 (not sure if it explicitly needs to be `null`)

cathysarisky: It was that way when I got to it, but I agree, it'd be a little simpler if it were initialized as [].  I can push that change over?

cathysarisky: Oh, no, hang on.  Line 318. We're checking if recommendtions is null and serving the loading page if it is.  So we do need to distinguish between null/uninitialized vs empty array because there are no recs, I think.

mike182uk:
Thanks for this @cathysarisky

fochoa8:
This is what I use in PHP to filter variables, if anyone can translate it, it would be a game changer.

image

image

image

https://github.com/user-attachments/assets/2a6303d0-08d9-4f63-833a-c2be025efab7

Thanks,
Fernando Ochoa Olivares.

MIT Researcher.

PR #21267, opened 2024-10-09
🌐 New translations and bug fix on /month and /year

cathysarisky:
No issue

Wrapped a missing /month and /year string.  Should work with and without trials now.

Added a few additional translations for Japanese and French.

Adjusted German - some strings were too long to fit in the layout.  Changed (the German equivalents of) "Start a X day free trial" to "Select", because the German is just not going to fit on the button.


mike182uk:
Thanks for this @cathysarisky looks great 👌 

Have merged in and should go out in the next release

PR #21249, opened 2024-10-07
Staff 2-factor-auth

kevinansfield:

PR #21284, opened 2024-10-10
Fixed members/signin_urls endpoint to take admin api key

### LINKED ISSUE #16748: admin/members/:id/signin_url won't take Owner's staff access token
### Issue Summary

I'm trying to make a fetch request to admin/members/:id/signin_url 

The [commit](https://github.com/TryGhost/Ghost/commit/b0ff1e7cacf619deed1d4f07c3cb0e7731707597) (Hi @naz !) that added the functionality seems to suggest this is possible. 

I'm converting the owner's staff access token into a JWT successfully, based on it working elsewhere on the admin API. 

I've tried on Ghost Pro (but it was a site where I'd switched owners, so could be an edge case).  I've also tried on a production 5.13 self-hosted site, and on a 5.47 local install (Windows).

Here's what I'm getting back (fresh 5.47 install, but all cases are the same):
Request to:  http://localhost:2368/ghost/api/canary/admin/members/6456ef85849b402bb40c61f4/signin_urls/

```
  errors: [
    {
      message: 'Permission error, cannot read member_signin_url.',
      context: 'You do not have permission to read member_signin_urls',
      type: 'NoPermissionError',
      details: null,
      property: null,
      help: null,
      code: null,
      id: '714c58e0-ec6e-11ed-94c9-6ba7df6bb971',
      ghostErrorCode: null
    }
  ]
```
The endpoint works fine if I pass in the site owner's cookie.

### Steps to Reproduce

```
const session = require('express-session')
const express = require("express");
const fetch = require('node-fetch')
const app = express();
app.use(express.json());
app.use(express.urlencoded({ extended: false }));
const jwt = require('jsonwebtoken')
const GhostAdminAPI = require('@tryghost/admin-api');
const GHOST_URL = 'http://localhost:2368'
let GHOST_API_KEY = '6456efec849b402bb40c61fc:07df42c75551444cb88af8601acbdcc4cc0d8ea983392ddb44dbb1cebb87a32b'; //local install, not worried about leakage.
let [GHOST_ID, GHOST_SECRET] = GHOST_API_KEY.split(':')

const api = new GhostAdminAPI({
  url: GHOST_URL,
  key: GHOST_API_KEY,
  version: "v5.0",
});

let GHOST_TOKEN = jwt.sign({}, Buffer.from(GHOST_SECRET, 'hex'), { // eslint-disable-line no-undef
    keyid: GHOST_ID,
    algorithm: 'HS256',
    expiresIn: '5m',
    audience: '/canary/admin'
});
console.log(GHOST_TOKEN)

async function getMember(email) {
  let result = await api.members.browse({email:email, limit: 1})
  console.log('getMember result is', result[0].id)
  return result[0].id;
}

async function getAuthUrl(id) {
  // does not work
  let result = await fetch(GHOST_URL + '/ghost/api/canary/admin/members/' + id + '/signin_urls/', {
  method: 'GET'  ,
  headers: {
      'Authorization': `Ghost ${GHOST_TOKEN}`,
      'Content-type': 'application/json'}
    }   )
  let json= await result.json()
  console.log('result is', json)
  return result;
}


async function getMemberUrl(id) {
  // works fine
  console.log('getMemberUrl request to: ',  GHOST_URL + '/ghost/api/canary/admin/members/' + id )
  let result = await fetch( GHOST_URL + '/ghost/api/canary/admin/members/' + id , {
  method: 'GET'  ,
  headers: {
      'Authorization': 'Ghost ' + GHOST_TOKEN}
  })
  let json= await result.json()
  console.log('result is', json)
  return result;
}

app.get("/", async (req, res) => { 
    res.send(`<center style="font-size:160%"> <p>This is Home Page </p>
    ${await getMemberUrl(await getMember('catsarisky@gmail.com'))}
    ${await getAuthUrl(await getMember('catsarisky@gmail.com'))}

    `);
  }
);

app.listen(3002);
```

### Ghost Version

5.47

### Node.js Version

16

Ghost Pro, also local install on Windows, also 5.13 production

### Database type

MySQL 8

### Browser & OS version

_No response_

_No response_

### Code of Conduct

cathysarisky:
closes #16748 

The members/:member_id/signin_urls endpoint currently only does cookie-based authentication.  When #21249 is merged, turning on 2FA is going to break any 3rd party processes that use it (including my social sign-in offering).

This patch gives admin API keys 'read' permission on this endpoint, and enables 3rd party processes to handle user logins the right way, instead of via a staff member's email/password.

Migration included.  Feedback appreciated.

I have the wrong name on my migration.  I can see it doesn't follow the naming convention, but I'm not sure how the names are generated.


ghost/core/test/regression/api/admin/members-signin-url.test.js
```diff
@@ -99,4 +99,60 @@ describe('Members Sigin URL API', function () {
...
+        before(async function () {
+            await localUtils.startGhost();
+            request = supertest.agent(config.get('url'));
+            await testUtils.initFixtures('members', 'api_keys');
+
+            const admin = await testUtils.createUser({
+                user: testUtils.DataGenerator.forKnex.createUser({email: 'admin+1@ghost.org'}),
+                role: testUtils.DataGenerator.Content.roles[0].name
+            });
+            request.user = admin;
```
9larsons, line 113: This doesn't appear necessary. If authenticated via key, no user should be present (I think?).

9larsons (changes requested):
Just the one comment. Tested this locally and looks good!

ghost/core/test/regression/api/admin/members-signin-url.test.js
9larsons: One side note, idk why these are regression tests. That's not a you problem, that's a problem with our patterns an lack of consistency in core :/.

cathysarisky: IDK either.  I'm honestly not convinced I'm even actually testing much of anything, given that we have fake-fixtures for testing, and I had to change them manually.  But hey, we have tests and they pass, so... yeah?

9larsons: The tests are good 👍. I checked the middleware and it's authenticating via API key as we'd like to see. We're not so worried about testing the migration as the underlying functionality. Agreed we could probably do more to cover that particular piece, but we instead mitigate migration troubles with extra review and a 'simple as possible' approach for them.

9larsons (approved):
LGTM. I've asked @mike182uk to also take a pass just because of the migration.

ghost/core/core/server/data/migrations/versions/5.97/2023-10-10-01-02-03-add-signin-urls-permissions.js
```diff
@@ -0,0 +1,8 @@
+const {combineTransactionalMigrations, addPermissionToRole} = require('../../utils');
+
+module.exports = combineTransactionalMigrations(
+    addPermissionToRole({
+        permission: 'Read member signin urls',
+        role: 'Admin Integration'
+    })
+);
```
mike182uk, line 8:
```suggestion
module.exports = addPermissionToRole({
    permission: 'Read member signin urls',
    role: 'Admin Integration'
});
```

I don't think we need the `combineTransactionalMigrations` as there is only 1 transactional migration here (`addPermissionToRole` creates a `createTransactionalMigration` under the hood)

PR #21287, opened 2024-10-13
Added post-checkout hook for submodule updates

cmraible:
no issue

- Currently to work on Ghost locally, you have to use the `--recurse-submodules` flag when first checking it out with `git clone ....`.
- Adding `git submodule update --init --recursive` to a post-checkout hook accomplishes the same thing, without the user having to know the exact correct command to run when cloning.
- This allows us to use a simpler `git clone ...` command.


cmraible (commented):
Need to test this on a fresh clone when I'm not on airplane wifi.

cmraible:
Realized we can do this in the `prepare` script (along with installing git hooks), which accomplishes the same thing essentially. This approach didn't seem to work as I expected it to — it worked when changing branches, but not immediately after a fresh clone.

PR #21314, opened 2024-10-16
Added Turkish local for portal

xyxforyou:
I jast want to change my theme to Turkish language


ghost/i18n/locales/context.json
```diff
@@ -170,7 +170,7 @@
     "Report this comment?": "Title of the modal to report a comment",
     "Retry": "When something goes wrong, this link allows people to re-attempt the same action",
     "Save": "A button to save",
-    "Search posts, tags and authors": "",
+    "Search posts, tags and authors": "Gönderileri, etiketleri ve yazarları arayın",
```
cathysarisky, line 173: translations don't belong in context.json. Please revert this change
```suggestion
    "Search posts, tags and authors": "",
```

cathysarisky (changes requested):
Hello xyzforyou!  Thanks for this effort.

You've accidentally committed a folder called 'en copy', and you have folders called tr and TR.   

What we need is a folder named 'tr', containing the five files (comments, ghost, portal, search, signup-form).   We already have Turkish translations in the system, so you should start from there and look for what needs to be updated, rather than starting from nothing.

You may find it easier to start a new pull request that has the intended changes (created by editing the existing files in locales/pr), rather than trying to fix the issues with this one.  Either way, hit the 'request review' button or tag me when you've got things fixed and I'll be back to check it over, ok?

Also, a note: this won't change your theme.  It'll change the membership pop-overs (Portal), the search section, the sign-up card, and the comments.

allouis:
@xyxforyou Do you intend to continue work on this PR?

allouis:
Closing due to inactivity

PR #21329, opened 2024-10-16
Added @tryghost/metrics-server as a dependency

cmraible:
no issue

- I apparently never added `@tryghost/metrics-server` as a dependency to `ghost/core/package.json`. It worked in most cases as a 'Ghost dependency' — yarn installs all node_modules in a flat structure, so even though it wasn't a dependency in package.json, it still resolved to the correct package, as long as the typescript packages were all built first.
- This passed CI because we explicitly run `ts:build` on all packages before running tests, and it worked in production because we build the TS packages as part of the docker build. However, when trying to run tests locally, it would sometimes fail unless you explicitly ran `nx run-many -t build:ts` at the top level before running the tests.
- Adding it as a dependency in `package.json` fixes this problem.


cmraible:
Somehow branch got messed up, closing in favor of https://github.com/TryGhost/Ghost/pull/21330

PR #21331, opened 2024-10-16
✨New helper: content_api_url

cathysarisky:
This adds a content_api_url helper, returning the url for Ghost's content api.  Takes a false argument if a relative URL is wanted.

Why? Because I just realized I don't have access to the content API's url in handlebars. :)  This addition will facilitate 3rd party integrations built into themes.

Companion PR for gscan: https://github.com/TryGhost/gscan/pull/558


ghost/core/core/frontend/helpers/content_api_path.js
```diff
@@ -0,0 +1,29 @@
...
+function getAbsoluteOption(options) {
+    const absoluteOption = options && options.hash && options.hash.absolute;
+    if (absoluteOption === undefined || absoluteOption === 'true' || absoluteOption === true || absoluteOption === null) {
+        return true;
+    } else {
+        return false;
+    }
+}
+
+module.exports.async = true;
```
allouis, line 29: I don't think this needs to be an async helper! It will be much more performant without it!

allouis: You're also missing a newline at the bottom of this file - I would recommend installing an [EditorConfig](https://editorconfig.org/) plugin for your editor

allouis (commented):
Few minor comments but otherwise it looks good!

PR #21334, opened 2024-10-16
Website Schema.org Object should have a "name" attribute 

ltoinel:
The Website Schema.org should have a name attribute on the root Website object. 
This value is used by Google Search for the name display on to top of the content title.

Documentation : 
* https://developers.google.com/search/docs/appearance/site-names?hl=fr#website

On my website, Google always use the "domain" to present my site and don't use the og:site_name as documented by google.
The missing "name" attribute on the website is maybe the source of the problem ...


allouis:
@ltoinel Can you please fix the unit tests? You should be able to do this by running `UPDATE_SNAPSHOT=1 yarn test:unit` in the repo and commit the changes!

ghost/core/test/unit/frontend/meta/schema.test.js
```diff
@@ -487,6 +487,7 @@ describe('getSchema', function () {
                 height: 500
             },
             mainEntityOfPage: 'http://mysite.com/post/my-post-slug/',
+            name: "Site Title",
```
ltoinel, line 490:
```suggestion
            name: 'Site Title',
```

ltoinel:
Tests fixed

PR #21336, opened 2024-10-17
Added migration for new settings keys

aileen:
This adds a new migration needed for custom fonts with `body_font` and `heading_font` as new possible site settings in our default settings.


kevinansfield (approved):

ErisDS (approved):
LGTM 👍

PR #21359, opened 2024-10-22
MRR/Members chart axis fix

dvdwinden:
ref https://linear.app/ghost/issue/DES-869/dashboard-mrr-member-chart-axis-is-making-flat-growth-look-like

A slight decline/increase on the MRR/members chart was shown as a very dramatic shift. These changes will make the chart appear more gradual when the changes are small. The chart will effectively "zoom in" on the range where the values fluctuate, making small changes more visible without appearing overly dramatic.

I've tested it with hardcoded values, as well as values inserted via the API, and the charts look a lot better.

**Before**
1 - chart before
2 - chart before

**After (same two values)**
3 - chart after
4 - chart after

**Gradual decline**
5 - chart gradual
6 - chart gradual

**More pronounced decline**
7 - chart dramatic
8 - chart dramatic
9 - chart dramatic

### DIFF
--- a/ghost/admin/app/components/dashboard/charts/anchor-attribution.js
+++ b/ghost/admin/app/components/dashboard/charts/anchor-attribution.js
@@ -816,20 +816,14 @@ export default class Anchor extends Component {
     }
 
     getYAxisMin() {
-        if (this.selectedChartDisplay !== 'mrr') {
-            return 0;
-        }
         const data = this.chartData.datasets[0].data;
         const min = Math.min(...data);
         return Math.floor(min * 0.95); // Start y-axis at 95% of the minimum value
     }
 
     getYAxisMax() {
-        if (this.selectedChartDisplay !== 'mrr') {
-            return null; // Let Chart.js decide the max for other chart types
-        }
         const data = this.chartData.datasets[0].data;
         const max = Math.max(...data);
-        return Math.ceil(max * 1.05); // End y-axis at 105% of the maximum value
+        return Math.ceil(max * 1.05); // End y-axis at 105% of the maximum value for all chart types
     }
 }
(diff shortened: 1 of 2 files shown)


ghost/admin/app/components/dashboard/charts/anchor-attribution.js
```diff
@@ -811,4 +814,22 @@ export default class Anchor extends Component {
 
         return Math.round((to - from) / from * 100);
     }
+
+    getYAxisMin() {
+        if (this.selectedChartDisplay !== 'mrr') {
+            return 0;
+        }
+        const data = this.chartData.datasets[0].data;
```
vershwal, line 822: We should add null or undefined checks for this.chartData and this.chartData.datasets[0] to prevent potential runtime errors when the data is unavailable. Something like this:-

```
if (!this.chartData || !this.chartData.datasets.length) {
        return null;
    }
```

ghost/admin/app/components/dashboard/charts/anchor-attribution.js
```diff
@@ -811,4 +814,22 @@ export default class Anchor extends Component {
 
         return Math.round((to - from) / from * 100);
     }
+
+    getYAxisMin() {
+        if (this.selectedChartDisplay !== 'mrr') {
```
vershwal, line 819: Do we want this change only for mrr charts? Can you please confirm?

cathysarisky:
As a recovering scientist, these charts have always bugged me, because the Y axis really ought to start at zero.  This is at least better! :)  

I'd vote for starting at zero, if that were an option.  And yes, that means that changes from 50,000 to 50,001 will look flat.  Because functionally, they ARE flat.  Trying to make changes that are too small to see into big news (or even small news) is really just inflating the noise.  If I can't see changes in the 90 day view, perhaps I'm  just looking at noisy data... :)

PR #21388, opened 2024-10-24
Changed SSO adapter to automatically verify sessions

sam-lord:
ref ENG-1680

SSO is a different flow that wouldn't need the extra email verification flow


kevinansfield (commented):
Do we have an e2e test for SSO? Would be nice to see a behaviour test that failed before this change was in place

PR #21393, opened 2024-10-24
Updated ActivityPub collection retrieval to accommodate pagination

mike182uk:
refs [AP-526](https://linear.app/ghost/issue/AP-526/implement-pagination-for-fedify-collections)

Updated followers, following, outbox and liked collection retrieval to accommodate pagination

### DIFF
--- a/apps/admin-x-activitypub/src/hooks/useActivityPubQueries.ts
+++ b/apps/admin-x-activitypub/src/hooks/useActivityPubQueries.ts
@@ -193,17 +193,6 @@ export function useFollowersForUser(handle: string) {
     });
 }
 
-export function useFollowersExpandedForUser(handle: string) {
-    return useQuery({
-        queryKey: [`followers_expanded:${handle}`],
-        async queryFn() {
-            const siteUrl = await getSiteUrl();
-            const api = createActivityPubAPI(handle, siteUrl);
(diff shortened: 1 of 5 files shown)


apps/admin-x-activitypub/src/hooks/useActivityPubQueries.ts
```diff
@@ -193,17 +193,6 @@ export function useFollowersForUser(handle: string) {
     });
 }
 
-export function useFollowersExpandedForUser(handle: string) {
```
mike182uk, line 196: No longer need as the followers collection now returns expanded items

allouis (commented):
I think we can generalise the recursion here so we simplify the code? Also - do we want to be traversing the entire collection still? Is it worthwhile to only fetch first N?

mike182uk:
> I think we can generalise the recursion here so we simplify the code? 

Yeh this could be DRY'd up, but i wasn't sure on the longevity of the usage of these endpoints in the client 🤔 (i.e are we likely to move away for usage of the AP endpoints directly)

> Also - do we want to be traversing the entire collection still? Is it worthwhile to only fetch first N?

This retains the existing behaviour without needing to change anything else in the application (i.e the call sites). The places where these fn's are used (the profile page for the site user) don't yet have a way of traversing pagination

PR #21395, opened 2024-10-24
AP design bugs

djordjevlais:
- Fixed links in profile description
- Stripped post content
- Fixed grey bg in Avatars
- Installed `clsx`


apps/admin-x-activitypub/src/components/Activities.tsx
```diff
@@ -11,6 +11,7 @@ import MainNavigation from './navigation/MainNavigation';
 import ViewProfileModal from './global/ViewProfileModal';
 
 import getUsername from '../utils/get-username';
+import stripHtml from '../utils/strip-html';
```
mike182uk, line 14: How come we need to do this if we are sanitizing on the backend?

djordjevlais: @mike182uk  We have two different use cases:
1. Keeping HTML tags and classes and then rendering that HTML nicely formatted (e.g. post content when shown in Feed layout or when shown in the drawer)
2. Stripping all of the formatting and showing just plain text (e.g. post content when shown in Inbox layout, screenshot attached)

Is there a better way to achieve this?

image

PR #21400, opened 2024-10-24
Fixed dev container build caching issues

cmraible:
no issue

- With image names specified for each `compose.yml` file, Dev Containers were being reused when they shouldn't have been: if you opened the `main` branch in a dev container, then tried to open a PR in a dev container, it would just re-use the original `main` container rather than rebuilding a new one for the PR.


cmraible:
Turns out the `yarn test:browser` command in the root of the repo does work locally, it's just got a lot of output before the tests start to run 🤦🏼‍♂️ — closing this.

PR #21402, opened 2024-10-24
Fixed browser tests giving false passing results in CI

cmraible:
no issue

- Browser tests were failing in CI but giving a passing result nonetheless
- This commit switches back to using the old method of running browser tests, using the version of the apps in the CDN rather than the local versions, because the command using the local versions wasn't exiting with the right code to trigger a failure in CI.
- This commit also fixes a few browser tests that were consistently failing, since we weren't catching them in CI.


cmraible:
This works, but ideally we'd still be using the local versions of the apps. Trying out one more option before going forward with this one.

PR #21417, opened 2024-10-25
🐛 fix search state bug & missing last result

### LINKED ISSUE #21343: Quirky search behavior
### Issue Summary

I'm seeing some not-quite-right behavior from the sodo-search package, but having trouble coming up with a simple repro case.

Here's what I've got:
Run a search for 'fitness' at https://goodvibeswellness.com/ 
Scroll down.  There are posts that actually match 'fitness', but the ui seems to get stuck with a bunch of posts that just match 'F' at the top.
I can reproduce it for this site's content (when exported and imported locally), using sodo-search 1.3, and also 1.1.  (That probably rules out the CJK changes being to blame.)
I cannot reproduce it with another site running search at 1.1, nor on another site running search @1.3.

Leaving this here in case anyone else is seeing something similar.

image

### Steps to Reproduce

see above

### Ghost Version

search 1.1-1.3

### Node.js Version

varies

local & ghost pro

### Database type

MySQL 8

### Browser & OS version

chrome, windows 10

_No response_

### Code of Conduct

cathysarisky:
closes #21343 

This fixes an issue in sodo-search where early results (for example, matching the first letter typed in the search box) would get "stuck" in the display, due to failure to update state (or actually, lacking state).

Converted PostItems to a component, and gave paginatedPosts state.

I also fixed an undocumented bug where the last search result didn't appear, due to an error in with slice's register.

### DIFF
--- a/apps/sodo-search/src/components/PopupModal.js
+++ b/apps/sodo-search/src/components/PopupModal.js
@@ -378,27 +378,29 @@ function ShowMoreButton({posts, maxPosts, setMaxPosts}) {
 function PostResults({posts, selectedResult, setSelectedResult}) {
     const {t} = useContext(AppContext);
     const [maxPosts, setMaxPosts] = useState(DEFAULT_MAX_POSTS);
+    const [paginatedPosts, setPaginatedPosts] = useState([]);
     useEffect(() => {
         setMaxPosts(DEFAULT_MAX_POSTS);
     }, [posts]);
-
+    useEffect(() => {
+        setPaginatedPosts(posts?.slice(0, maxPosts + 1));
+    }, [maxPosts, posts]);
     if (!posts?.length) {
         return null;
     }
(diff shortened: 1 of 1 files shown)


kevinansfield (commented):
@cathysarisky the use of state and `useEffect` added here is a [React anti-pattern](https://react.dev/learn/you-might-not-need-an-effect). Preferred approach is to use `posts` to build `paginatedPosts` directly in the first render rather than going the round-about way of using an effect to update state and causing a second render 🙂 

If building the list directly in the first render is causing problems then we have a different issue somewhere that needs addressing.

kevinansfield:
Ugh, sorry, I'd missed how we're using `maxPosts` state here and how `posts` changing resets the number of posts shown - this is the problem with `useEffect` for things like this, it makes the flow much harder to follow!

Seems like there's a refactor waiting here somewhere but your fix looks good for now 🙏

PR #21420, opened 2024-10-26
Pass data-locale into Portal

cathysarisky:
no issue

This change adds the ability to pass a desired language/locale into portal, as a data-locale tag added to the script that loads it.  

Falls back to the previous behavior, loading from the site settings, if no data-locale is present.  (And then to 'en', if necessary.)  No change in expected behavior if data-locale is not passed.

This provides an option to run portal in a language other than the sitewide language.

(ref: https://app.slack.com/client/T025584C4/C018EKC56JF)


apps/portal/src/App.js
```diff
@@ -551,7 +552,7 @@ export default class App extends React.Component {
     /** Fetch site and member session data with Ghost Apis  */
     async fetchApiData() {
         const {siteUrl, customSiteUrl, apiUrl, apiKey} = this.props;
-
+
```
cathysarisky, line 555:
```suggestion

```

PR #21433, opened 2024-10-28
🌐✨i18n for newsletter strings

cathysarisky:
no issue

This PR adds the ability to translate the strings that appear in the newsletter as boilerplate text, using i18next.

Variables are in single mustaches ( `{date}` ) in the translation strings (rather than `{{date}}`), because these strings occur both the email template.hbs and also .js files.  That necessitated a separate namespace.

This PR also includes changes to the newsletter button ("more like this", "less like this", "comment") that were previously delivered on desktop as images that included the text.  @sanne-san provided a rework that removed text-as-image from the desktop buttons, and allows more shared code between the two layouts, along with making the buttons translatable.

Example usage - handlebars
```
<h3 class="latest-posts-header">{{t 'Keep reading'}}</h3>

{{{t 'By {authors}' authors=post.authors }}} 
```
(NOTE: triple { required because of possible & )

Example usage - javascript
```
                getValue: (member) => {
                    if (member.status === 'comped') {
                        return t('complimentary');
                    }
                    if (this.isMemberTrialing(member)) {
                        return t('trialing');
                    }
                    // other possible statuses: t('free'), t('paid') //
                    return t(member.status);
                }
```


ghost/email-service/i18n-setup.js
vikaspotluri123, line 1: From what I've seen in the past, you want to limit the files in the package root - it might make sense to move this to the lib folder

ghost/email-service/i18n-setup.js
```diff
@@ -0,0 +1,15 @@
+const i18nLib = require('../i18n');
```
vikaspotluri123, line 1: I _think_ you don't want to import files across package boundaries - you would need to add the package as a dependency and import from it

ghost/email-service/i18n-setup.js
```diff
@@ -0,0 +1,15 @@
...
+const settingsCache = require('../core/core/shared/settings-cache');
```
vikaspotluri123, line 2: You might need to make a part of the email renderer since I think the `settingsCache` needs to be provided as a dependency rather than imported

vikaspotluri123:
Definitely take what I said with a grain of salt! Someone should confirm my observations 🙈

package.json
```diff
@@ -128,5 +128,6 @@
     "rimraf": "5.0.10",
     "ts-node": "10.9.2",
     "typescript": "5.4.5"
-  }
+  },
+  "packageManager": "yarn@1.22.21+sha1.1959a18351b811cdeedbd484a8f86c3cc3bbaf72"
```
cathysarisky, line 132:
```suggestion
```

package.json
```diff
@@ -128,5 +128,6 @@
     "rimraf": "5.0.10",
     "ts-node": "10.9.2",
     "typescript": "5.4.5"
-  }
+  },
```
cathysarisky, line 131:
```suggestion
  }
```

cathysarisky:
@sanne-san , if you have time to look at this, I'd appreciate it! :)

9larsons (approved):

PR #21440, opened 2024-10-28
eksik çeviriler yapıldı

echobilisim3421:


ghost/i18n/locales/tr/portal.json
```diff
@@ -95,8 +95,8 @@
     "In the event a permanent failure is received when attempting to send a newsletter, emails will be disabled on the account.": "Haber bülteni göndermeye çalışırken kalıcı bir hata alınması durumunda, hesapta e-postalar devre dışı bırakılır.",
     "In your email client add {{senderEmail}} to your contacts list. This signals to your mail provider that emails sent from this address should be trusted.": "E-posta istemcinizde {{senderEmail}} adresini kişi listenize ekleyin. Bu, posta sağlayıcınıza bu adresten gönderilen e-postaların güvenilir olması gerektiğini bildirir.",
     "Invalid email address": "Geçersiz e-posta adresi",
-    "Jamie Larson": "Ahmet Yılmaz",
-    "jamie@example.com": "ahmet@example.com",
+    "Jamie Larson": "Ad Soyad",
```
cathysarisky, line 98: Could you please provide some context for this change?

cathysarisky (approved):
Looks good to my AI helper.  Thanks for adding these Turkish translations!

ghost/i18n/locales/tr/portal.json
ErisDS: I double checked, and the original translation is more inline with our intention (a pretend, generic name) and the latter literally means "First name Surname". Therefore I think this update is incorrect.

ghost/i18n/locales/tr/portal.json
```diff
@@ -95,8 +95,8 @@
...
+    "jamie@example.com": "kullanıcıadı@example.com",
```
ErisDS, line 99: Similarly, this update moves away from the original intent as I believe it says "username@example.com"

ErisDS (changes requested):
Thanks so much for the additional translations! Two of the changed translations around our placeholder name and email address have moved away from the intention, whereas the original translations seem more in-line with what's expected.

Please could you either revert these, or else provide an explanation for why your change is more correct?

9larsons:
Closing as duplicate of https://github.com/TryGhost/Ghost/pull/21671.

ghost/i18n/locales/tr/portal.json
echobilisim3421: I'm sure everyone will use this, and no one would want someone else's name and surname to appear on their site. For this reason, when a name and surname are entered, the registered user will see their own name and surname. Entering a name and surname is the correct approach.

ghost/i18n/locales/tr/portal.json
echobilisim3421: Usually, the username from the email is displayed. Therefore, it is specified to enter their own email username. However, in the current translation, someone has written their own name, and it doesn't look appropriate on the site. Based on this, I wanted to clarify the correct approach.

ghost/i18n/locales/tr/portal.json
cathysarisky: You've misunderstood what this string is used for.  It's a placeholder.  It goes into the subscribe box until the user types their name.  Once they type their name, that's what appears there.  When a name and surname are entered, the user will indeed see their own name in the subscription box.  But when they first load the page, they should see an 'example' name.  Here's how it looks in English.  Notice the names are greyed out:
image
Then, after I type in my information, here's how it looks:
image

echobilisim3421: I am fully aware of what it is, but the current translation is incorrect. Therefore, I made the necessary correction.

PR #21453, opened 2024-10-29
Moved admin controls to a new button outside of the context menu

sanne-san:
REF PLG-239
- Added a new button for admin controls outside of the context menu
- The context menu now only shows actions if there are any to show
- This change in functionality is behind a feature flag


sanne-san:
Canceled due to design changes.

PR #21461, opened 2024-10-30
Removed the typescript command from browser tests

cmraible:
no issue

- Running browser tests from the root of the repo had a lot of extra output in CI because the NX daemon is disabled in CI, and it was using nx to watch & rebuild the typescript packages.
- Since the nx command for running browser tests already has the `build:ts` command as a dependency, we don't have to run it explicitly in `dev.js`, so I removed the command to watch & rebuild typescript.
- This should clean up the CI output a lot to make it easier to quickly see which browser test failed


cmraible:
Reopening on upstream

PR #21470, opened 2024-10-30
Added UUID as data attribute to posts in all views

djordjevlais:
ref https://linear.app/ghost/issue/AP-404/add-uuid-as-a-data-attribute-in-the-dom-for-easier-db-lookup

- This will allow us to find posts in the DB

### DIFF
--- a/apps/admin-x-activitypub/src/components/feed/FeedItem.tsx
+++ b/apps/admin-x-activitypub/src/components/feed/FeedItem.tsx
@@ -302,7 +302,7 @@ const FeedItem: React.FC<FeedItemProps> = ({actor, object, layout, type, comment
         return (
             <>
                 {object && (
-                    <div className={`group/article relative cursor-pointer pt-6`} onClick={onClick}>
+                    <div className={`group/article relative cursor-pointer pt-6`} data-uuid={object.id} onClick={onClick}>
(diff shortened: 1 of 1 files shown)


allouis (approved):

apps/admin-x-activitypub/src/components/feed/FeedItem.tsx
```diff
@@ -302,7 +302,7 @@ const FeedItem: React.FC<FeedItemProps> = ({actor, object, layout, type, comment
         return (
             <>
                 {object && (
-                    <div className={`group/article relative cursor-pointer pt-6`} onClick={onClick}>
+                    <div className={`group/article relative cursor-pointer pt-6`} data-uuid={object.id} onClick={onClick}>
```
mike182uk, line 305: Just an observation - `object.id` is a `URL` and not a `UUID`. To avoid ambiguity we could call this `data-object-id` instead? or am I being too pedantic 😅 @allouis @djordjevlais

mike182uk: Have changed this to `data-object-id` 👍

allouis: 😁 Sure, we can call it that! I didn't wanna block on it tho

PR #21490, opened 2024-10-31
[WIP] Theme i18n

cathysarisky:
Don't look at this one. I replaced it with a better one!
https://github.com/TryGhost/Ghost/pull/23161


ghost/core/core/frontend/helpers/t.js
```diff
@@ -26,5 +26,9 @@ module.exports = function t(text, options = {}) {
         }
     }
 
-    return themeI18n.t(text, bindings);
+    let result = themeI18n.t(text, bindings);
+
+    // The helper should always return a string, not a SafeString
+    // HTML escaping is handled by the template engine based on whether {{ or {{{ was used
+    return result;
```
cathysarisky, line 33:
```suggestion
    // The helper should always return a string, not a SafeString
    // HTML escaping is handled by the template engine based on whether {{ or {{{ was used
    return themeI18n.t(text, bindings);
```

PR #21498, opened 2024-11-01
Local French additions and improvements

bst1n:
It's difficult out of context to know whether to use an infinitive, _Ajouter votre commentaire_ or an imperative, _Ajoutez votre commentaire_. English doesn't make the difference, but French does. By default, an infinitive is preferable because it remains a suggestion, whereas an imperative gives an order. For example, I think that _Ajoutez votre commentaire_ is not relevant, because you don't necessarily have to add a comment, but you can, so you should say _Ajouter votre commentaire_. I've tried to be as precise as possible, but you may need to make some adjustments with the context.

I find that the word _membre_ is too literal a translation of the English language and includes too many innuendos such as "member of a community", "member of a clan", "member of a club", and frankly, it's not relevant to many blogs. The word _abonné_ is more relevant, because it doesn't imply a social affiliation. As for the discussion, I've even taken the liberty of writing _discussion entre lecteurs_, because that's essentially what it is. It's not a discussion between members or subscribers, it's a discussion between _readers_.

Some placeholders are completely American (Jamie Larson), so I've replaced him with a French one (Jean Martin).

Complimentary is a word that I didn't understand in the context of Ghost. I have the impression that it means _offert_, but I don't understand when it's used.


cathysarisky:
Complimentary here means that the member (reader!) has access to a paid tier, but isn't actually paying.  It means that you (the site owner) have given them a paid access level that they aren't actually paying for.

The placeholder name appears in several other files also.  It's totally fine to make it French, but we should probably do it everywhere.

cathysarisky (approved):
Thank you for these!

ghost/i18n/locales/fr/comments.json
```diff
@@ -33,11 +33,11 @@
...
     "Hide": "Masquer",
     "Hide comment": "Masquer le commentaire",
-    "Jamie Larson": "Jamie Larson",
-    "Join the discussion": "Rejoignez la discussion",
+    "Jamie Larson": "Jean Martin",
+    "Join the discussion": "Rejoindre la discussion",
     "Just now": "À l'instant",
     "Local resident": "Résident local",
-    "Member discussion": "Discussion entre membres",
+    "Member discussion": "Discussion entre lecteurs",
```
sagzy, line 40: I'd stick to `abonnés` here, as we usually don't use the term readers in English either, only members. The word `abonné` works for both free and paid members

sagzy (changes requested):
Hey @bst1n, thanks for the PR! All great changes 💯 

I have added one comment on the term "lecteur" versus "abonné"

ghost/i18n/locales/fr/comments.json
sagzy: Agree that the initial "membres" does not work well

bst1n: Okay

sagzy (approved):

PR #21509, opened 2024-11-04
Updated the local state after following an account

allouis:
refs https://linear.app/ghost/issue/AP-523

We want to preempt the Accept activity from our Follows, so we make the assumption that it's succeeded. What this means is that we have to update our `following`, `followingCount` as well as the fetched profile to set the `isFollowing` property. This gives a more fluid user experience when following accounts and keeps our state up to date.

Accounts where the follow request has to be accepted manually, are a little trickier as we don't currently have easy access to the "requested but not accepted state"

### DIFF
--- a/apps/admin-x-activitypub/src/api/activitypub.ts
+++ b/apps/admin-x-activitypub/src/api/activitypub.ts
@@ -234,9 +234,10 @@ export class ActivityPubAPI {
         };
     }
 
-    async follow(username: string): Promise<void> {
+    async follow(username: string): Promise<Actor> {
         const url = new URL(`.ghost/activitypub/actions/follow/${username}`, this.apiUrl);
-        await this.fetchJSON(url, 'POST');
+        const json = await this.fetchJSON(url, 'POST');
+        return json as Actor;
     }
 
     async getActor(url: string): Promise<Actor> {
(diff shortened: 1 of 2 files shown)


allouis:
@mike182uk Nice catch - but I _think_ this approach is preferable, so that state management is in a centralised place - I'm going to look at updating the existing code you have to use this approach

mike182uk:
> Hm, i think this might conflict with how we are updating this state elsewhere / cause double renders. For example:
> 
> On the search page, when we click a profile and load it in the profile viewer, if we click the `Follow` in the profile viewer, this currently propagates the change back to the search results:
> 
> * `useSearchForUser` hook returns a method for updating the search results state [here](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-activitypub/src/hooks/useActivityPubQueries.ts#L263)
> * `Search` component makes use of this hook [here](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-activitypub/src/components/Search.tsx#L84), passes the `updateSearchResult` fn to a `SearchResult` [here](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-activitypub/src/components/Search.tsx#L153), and then the `SearchResult` component sets up `onFollow`, `onUnfollow` fns [here](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-activitypub/src/components/Search.tsx#L34-L46)
> * `SearchResult` renders a `FollowButton` [here](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-activitypub/src/components/Search.tsx#L62-L69) and passes on the `onFollow` and `unFollow` props
> * `FollowButton` makes use of these props [here](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-activitypub/src/components/global/FollowButton.tsx#L28-L48) and executes them after the mutation has been made

Discussed on Slack with @allouis - This actually won't be a problem because the post-mutation does not update the search results 🙌

mike182uk (approved):

PR #21515, opened 2024-11-04
Updated timestamp format for comments

sanne-san:
ref https://linear.app/ghost/issue/PLG-235

- Any comments older than yesterday are now shown with the date instead of relative time
- Comments from the current year are now shown with just the month and day


apps/comments-ui/src/utils/helpers.ts
```diff
@@ -13,52 +13,53 @@ export function formatRelativeTime(dateString: string, t: TranslationFunction):
...
 
-    // Diff in hours
-    diff = diff / 60;
-    if (diff < 24) {
-        if (Math.floor(diff) === 1) {
-            return t(`One hour ago`);
-        }
-        return t('{{amount}} hrs ago', {amount: Math.floor(diff)});
+    // If we're on the same calendar day, show relative time
+    let diff = Math.round((now.getTime() - date.getTime()) / 1000);
```
kevinansfield, line 43: Can we change this variable name to `secondDiff`, and likewise below instead of redefining `diff` create new variables like `minuteDiff` and `hourDiff`? That should make it a bit easier to understand the code without having to read backwards and keep track of the changing variable meaning

apps/comments-ui/src/utils/helpers.test.ts
```diff
@@ -29,22 +29,19 @@ describe('Date/Time Formatting', () => {
...
-        it('handles just now', () => {
-            expect(helpers.formatRelativeTime('2024-02-15T14:59:57.000Z', t)).toBe('Just now');
+        beforeEach(() => {
+            clock = sinon.useFakeTimers(new Date('2024-02-15T15:00:00.000Z'));
         });
 
-        it('handles seconds ago', () => {
-            expect(helpers.formatRelativeTime('2024-02-15T14:59:30.000Z', t)).toBe('30 seconds ago');
+        it('handles just now', () => {
+            expect(helpers.formatRelativeTime('2024-02-15T14:59:57.000Z', t)).toBe('Just now');
```
kevinansfield, line 42: If we're adding multiple `expects` to each test it's good practice to provide an indication of what's under test, otherwise when there's a failure you need to find the line number it failed on and cross-reference because all you see is `expected "30 seconds ago" to equal "Just now"` with no indication of which of the three separate `expect` calls resulted in "Just now".

`expect()` accepts a second argument that adds a hint to the test output, e.g. 
```
expect(helpers.formatRelativeTime('2024-02-15T14:59:57.000Z', t), '2024-02-15T14:59:57.000Z').toBe('Just now');
```
which would output something like `2024-02-15T14:59:57.000Z: expected "5 mins ago" to be "Just now"` when failing so you can see exactly which `expect` failed.

It may be easier/cleaner to write a helper function to cover this, e.g:

```
function testFormatRelativeTime(dateStr, expected) {
    expect(helpers.formatRelativeTime(dateStr, t), dateStr).toBe(expected);
}
```

so the test becomes

```
    testFormatRelativeTime('2024-02-15T14:59:57.000Z', 'Just now');
    testFormatRelativeTime('2024-02-15T14:59:30.000Z', 'Just now');
    testFormatRelativeTime('2024-02-15T14:59:01.000Z', 'Just now');
```

PR #21516, opened 2024-11-04
add newsletter & search czech translation

kocour:
I've added missing newsletter and search strings to Czech translation.


cathysarisky (approved):
Looks good to my AI helper!

PR #21518, opened 2024-11-04
Change interpolation string for t helper

cathysarisky:
Change interpolation string for translation helpers from {{ }} to { }.  (Per discussion with Hannah, here: https://ghost.slack.com/archives/C018EKC56JF/p1730728246971699?thread_ts=1730560172.481589&cid=C018EKC56JF .)

**Context for the future:**
The 't' helper works in both javascript and handlebars.  
Syntax for javascript:
`t('some string {variable} is {another}', {variable: value1, another: value2}`
Syntax for handlebars:
`{{t 'some string {variable} is {another}' variable=value1 another=value2}}`

Using the 't' helper in handlebars means really really wanting the {substitutions} to be single bracketed, because it's really messy otherwise what's handlebars and what's interpolation.

Themes currently use single { } for variables.  We don't want to break existing themes.

We've been using double {{ }} for variables in portal, ghost, search, comments.  But then in newsletter, we had a mix of .hbs and .js, so switched to single { }.  Which meant things were inconsistent.

So, this beast of a patch converts all variables to using single brackets. 

I ran a script that went through all the existing translations and rewrote all the keys (and values)..  New changes should get added normally, but changing keys (even the variables within them) is bad, so it needed a one-time rewrite.


ErisDS:
Closing in favour of https://github.com/TryGhost/Ghost/pull/23330

PR #21521, opened 2024-11-04
AP drawer improvements

djordjevlais:
- Made `Articles` in drawer view wider
- Added `Actor` info to `Articles` in drawer view
- Removed modals we’re no longer using
- Updated `RoutingProvider` so it can work without any modals passed to it


apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
```diff
@@ -303,21 +330,45 @@ const ArticleModal: React.FC<ArticleModalProps> = ({
...
+                                <ArticleBody
+                                    excerpt={object?.preview?.content}
+                                    heading={object.name}
+                                    html={object.content}
+                                    image={object?.image}
+                                />
+                            </>
+                        )}
+
+                        {object.type === 'Article' && (
```
mike182uk, line 356: There is already a block of for `object.type === 'Article'`. Are we not able to move this into that?

djordjevlais: Good catch, changing it now

PR #21561, opened 2024-11-07
Make i18n more tolerant of missing files

cathysarisky:
This fix adds an extra fallback to 'en' when a locale folder is missing one or more translation files, and a test for the fallback.

I had to do a little bit of a refactor to extract a testable function.


9larsons:
@cathysarisky Hey, could you provide some background on the behavior that this is fixing? e.g. if I remove a translation file (`nl/search.json`) and run `yarn translate` in i18n, it'll build that file anyways. What would cause a missing file?

cathysarisky:
> @cathysarisky Hey, could you provide some background on the behavior that this is fixing? e.g. if I remove a translation file (`nl/search.json`) and run `yarn translate` in i18n, it'll build that file anyways. What would cause a missing file?

@9larsons, hey! I was trying to cover some edge cases.  A missing file causes ghost to throw an error and not start, which seemed like a big problem and well beyond what should happen.  (Missing translation file = not serious.  Not starting = serious.)  Agreed that the normal CI process ought to prevent a missing file, but it adds a little robustness for self hosters who might be hacking at the package or similar.

PR #21564, opened 2024-11-07
Removed customFonts feature flag

minimaluminium:
ref DES-949


ghost/core/test/unit/frontend/helpers/body_class.test.js
```diff
@@ -217,8 +215,6 @@ describe('{{body_class}} helper', function () {
         });
 
         it('does not include custom font classes when custom fonts are not enabled', function () {
```
aileen, line 217: This test is fully redundant now, so I'll remove it.

PR #21572, opened 2024-11-07
Removed nodemon in favor of node --watch

cmraible:
no issue

- We had reintroduced nodemon in https://github.com/TryGhost/Ghost/commit/af0f26c75f415632ae801f7da0da0e27e8eadde0#diff-bf18f8caf848e17b35e266db04bcaeaad05a3e5d069846615d2b1260482396e1 for the docker setup, but it has since caused some issues with the `yarn dev` script.
- In particular, it was causing a restart while migrations were running in development, which left a migration lock on and prevented Ghost from starting.
- This commit removes nodemon and replaces it with node --watch, which we had been using in the past without issues.

### DIFF
--- a/ghost/core/nodemon.json
+++ b/ghost/core/nodemon.json
@@ -1,3 +1,0 @@
-{
-    "ignore": ["*.test.js", "**/content/public/**", "**/core/built/**", ".yarncache/**"]
-}
--- a/ghost/core/package.json
+++ b/ghost/core/package.json
@@ -21,7 +21,7 @@
   "license": "MIT",
   "scripts": {
     "archive": "npm pack",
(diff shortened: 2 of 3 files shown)


cathysarisky:
Hey @cmraible !  My dev setup is now spamming "Daemon is not running. The watch command is not supported without the Nx Daemon." when I try to start ghost with 'yarn dev'.  

Is there something I can do to get it to behave, while not running in docker? :)  Thanks!

cmraible:
Hi @cathysarisky! Thanks for flagging this. I _think_ if you run `yarn nx reset` that should fix the issue for you. Let me know if that helps!

cathysarisky:
Thanks, that's fixed it! :)

PR #21575, opened 2024-11-07
Moved newsletter date localisation behind labs flag

ErisDS:
ref https://github.com/TryGhost/Ghost/pull/21573

- We discussed this, and figured we should get this behind the flag for now, even though it's pulled
- This gives us the ability to do a fast revert if we have to later, which is the point of having the labs "GA FEATURES" array
- Note this removes the labsEnabled variable in the tests to make it clearer if the flag is off or on (it was off!)


9larsons:
Yeah this is quick and dirty but as you said, very easy to revert and looks sound.

PR #21600, opened 2024-11-12
AP Inbox view improvements

djordjevlais:
- Improved the layout, spacing and typography of Inbox view
- Removed attachment counters for images in Inbox view
- Added a very very small variant of `APAvatar`

### DIFF
--- a/apps/admin-x-activitypub/src/components/Inbox.tsx
+++ b/apps/admin-x-activitypub/src/components/Inbox.tsx
@@ -76,7 +76,7 @@ const Inbox: React.FC<InboxProps> = ({}) => {
                         <>
                             <div className={`mx-auto flex items-start gap-8`}>
                                 <div className='flex w-full min-w-0 items-start'>
-                                    <ul className={`mx-auto flex w-full flex-col ${layout === 'inbox' ? '3xl:max-w-[800px]' : 'max-w-[500px]'}`}>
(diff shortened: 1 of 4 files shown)


apps/admin-x-activitypub/src/components/Inbox.tsx
```diff
@@ -74,9 +74,9 @@ const Inbox: React.FC<InboxProps> = ({}) => {
                         </div>
                     ) : activities.length > 0 ? (
                         <>
-                            <div className={`mx-auto flex items-start ${layout === 'inbox' ? 'max-w-6xl gap-14' : 'gap-8'}`}>
+                            <div className={`mx-auto flex items-start gap-8`}>
                                 <div className='flex w-full min-w-0 items-start'>
-                                    <ul className={`mx-auto flex ${layout === 'inbox' ? 'w-full max-w-full' : 'max-w-[500px]'} flex-col`}>
+                                    <ul className={`mx-auto flex w-full flex-col ${layout === 'inbox' ? 'xxxl:max-w-[800px]' : 'max-w-[500px]'}`}>
```
peterzimon, line 79: @djordjevlais `xxxl:` here is not the standard used in Tailwind CSS. It should be `3xl` — not sure if it's possible to start a definition with a number though

peterzimon (dismissed):

apps/admin-x-activitypub/src/components/Inbox.tsx
djordjevlais: Oh good point - it works when you wrap it like this `’3xl’`. I’ll push it in a sec 👍

PR #21622, opened 2024-11-14
Added "time to acquire connection" prometheus metric

cmraible:
ref https://linear.app/ghost/issue/ENG-1769/improve-pool-utilization-metric

- The current connection pool metrics collect a point in time sample of the number of active and used connections. With a scraping interval of 15s, these metrics aren't sensitive enough to rapid changes in the pool. This commit adds a metric to track the time to acquire a connection from the pool, and reports percentiles to prometheus.
- This should provide a much more responsive and valuable metric for identifying when there is contention in the database pool — if a query is waiting more than a few milliseconds for a connection to free up, that indicates there is contention in the database pool.
- This commit also makes the `prometheus.instrumentKnex()` method idempotent so it can safely be called on the `knex` instance more than once, and it moves the call to `prometheus.instrumentKnex()` to after the settings service is initialized in the boot process, because the `knex.pool` is torn down and recreated during the settings initialization, which removes the event listeners on the pool used to collect the time to acquire connections.

### DIFF
--- a/.github/scripts/grafana/dashboards/main-dashboard.json
+++ b/.github/scripts/grafana/dashboards/main-dashboard.json
@@ -3,7 +3,10 @@
     "list": [
       {
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
         "enable": true,
         "hide": true,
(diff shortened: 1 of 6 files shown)


cmraible:
Going to break this PR up a bit

PR #21643, opened 2024-11-18
Updated analytics button logic for disabled newsletter emails

9larsons:
ref https://linear.app/ghost/issue/ONC-594

We had a check to prevent showing the Analytics page link for email-only posts (newsletters) if newsletters were disabled. I don't see a good reason to remove this - users then have to re-enable newsletters just to see the analytics.


9larsons:
I'm going to merge this noting that we need to come to a design decision around what expected behavior is with the newsletter toggle.

If off, do we want to prevent showing any record of emails/email data? Or should we show historical data for when it was enabled? This would require changes in the analytics page as it stands.

PR #21650, opened 2024-11-19
Reorganized docker related files

cmraible:
# Proposal

## Description
This PR is a proposal to move the `docker-compose.yml` file, and all other docker related files into a new `docker` directory at the top level of the repo.

## What's the problem?
Currently our `docker-compose.yml` and all related files are nested inside the `.github/scripts` directory. There's no functional problem with this, but I'd argue that it's a confusing location. If you're new to the repository and you're trying to find the docker files, it's probably one of the last places you'd look for them — you only know to look there if you already know that's where they are. 

## What's the proposed solution?
We create a new `docker` directory at the top level of the repo, which will contain the `docker-compose.yml` file(s), the Dockerfile, and any supporting configuration files needed by the services, nested within their own directories.

### Pros:
- More conventional location makes the docker files much easier to find
- The path to any file is shorter, easier to remember and type
- Room for growth — if we add more services that need supporting configuration files or assets, we can add those to this directory (appropriately nested, like we have for `docker/prometheus` and `docker/grafana`)

### Cons:
- Currently the only non-hidden top-level directories are `apps` and `ghost`, which both contain actual source code. This would break that unwritten convention
- `docker` falls between `apps` and `ghost` alphabetically, so it would be between these two directories in most IDEs

### Alternative solutions:
- Keep these files in `.github/scripts`
- Keep the main files (`Dockerfile` and any `compose.yml` files in the root), and keep supporting files in a hidden dir (i.e. `.config`, `.infra`, `.services`, etc)
- Instead of `docker`, name the new directory `.docker` which would hoist it to the top (although this would also hide the files by default in most operating systems so they wouldn't appear in `ls` or Finder)
- Instead of `docker`, name the new directory `_docker` which would hoist it to the top, but not hide them.


cmraible:
Based on the feedback, planning to change this up _a bit_:

- Keep the `docker-compose.yml` file at the top level of the repo
- Move all the other docker files into `.docker`

PR #21677, opened 2024-11-21
chore: migrate IFrame component to functional component

crutchcorn:

    - Neither of these commands ran any tasks and `npx nx graph` doesn't show any task graphs either, I wonder if Nx is misconfigured.

---------------------

While looking at the codebase, I found:

- An unmemoized `head` element that, on re-render, would wreck havoc on performance thanks to reconstructing all of the CSS styles
- A class component that referenced migration to a functional component
- Missing property typings for the `iframe` component

I figured I'd see if I could help with these issues, so I took a stab at it. I loaded the signup page and did a comparison against both the production tab and the development tab without my changes and it seemed to work fine.

> I still worry about some of the uncleaned up effects. I know the comment mentions that the `iframe` being removed will cleanup the effects, but I'm not entirely certain of this. After all, we're passing an arrow function which the browser's GC might not handle properly.
>
> This isn't a _huge_ deal given the scope of this component, but it's a minimal refactor to handle in a more 'tidy' manner. I could do this refactor as a follow-up if requested.

### DIFF
--- a/apps/signup-form/src/components/IFrame.tsx
+++ b/apps/signup-form/src/components/IFrame.tsx
@@ -1,70 +1,58 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
-import {Component} from 'react';
+import React from 'react';
 import {createPortal} from 'react-dom';
 
-/**
- * This is still a class component because it causes issues with the behaviour (DOM recreation and layout glitches) if we switch to a functional component. Feel free to refactor.
- */
-export default class IFrame extends Component<any> {
-    node: any;
-    iframeHtml: any;
-    iframeHead: any;
-    iframeRoot: any;
+type IFrameProps = React.PropsWithChildren<{
+    style: React.CSSProperties
+    head: React.ReactNode;
+    title: string;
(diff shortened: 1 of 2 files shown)


apps/signup-form/src/components/IFrame.tsx
```diff
@@ -1,70 +1,58 @@
...
+
+    // TODO: Migrate to callback ref when React 19 cleanup refs are available: https://react.dev/blog/2024/04/25/react-19#cleanup-functions-for-refs
+    const node = React.useRef<HTMLIFrameElement>(null);
+    React.useEffect(() => {
+        function setupFrameBaseStyle() {
+            if (node.current?.contentDocument) {
+                const iframeRootLocal = node.current.contentDocument.body;
+                // This is safe because of batched updates, new to React 18
+                setIframeHead(node.current.contentDocument.head);
+                setIframeRoot(iframeRootLocal);
```
crutchcorn, line 23: No need to recall `this.forceUpdate` anymore, since the batched updates will trigger a re-render, updating the `iframeHead` and `iframeRoot` usages in the JSX

Issue #21439, opened 2024-10-28
Drop target for uploads (content and theme) is broken.


cathysarisky:
### Issue Summary

(Not a new issue - has been bugging me for months!)

The 'drop target' for drag and dropping a theme or upload file doesn't work right on Windows 10, Chrome browser.  Occurs on local installs AND on Ghost Pro.

Image: Shows hovering over the 'drop area', but the cursor still shows 'copy', and if I drop it here, it'll download it (NOT upload it to Ghost).
![image](https://github.com/user-attachments/assets/e4837bf4-580d-4cf7-9366-fbd6f262f128)

Image:  The very edge of the drop area (marked in yellow) is actually droppable.  /IF/ I can get my cursor over this sliver between the dotted and solid borders, dropping works.  The big red x area is not.
![image](https://github.com/user-attachments/assets/2cf3ea75-b33a-40a8-ba43-f781f3a7362f)

Chrome is up to date:
Version 130.0.6723.70 (Official Build) (64-bit)

**UPDATE:**  See https://github.com/TryGhost/Ghost/pull/21687 for more discussion .  I think this is not actually a problem with the drop area, but with re-rendering.  (Holding the mouse still for a long period seems to stop the rerenders, and targeting the edge was probably causing me to do that?  I think?)

### Steps to Reproduce

Use Chrome.  (Possibly use Chrome on Windows - not sure.)
Attempt to drag and drop a file into the drop area for theme upload or import upload.

I expect the entire box to be a valid drop target.

### Ghost Version

5.98 (and multiple versions before)

### Node.js Version

18

### How did you install Ghost?

various, including on Ghost Pro

### Database type

MySQL 8

### Browser & OS version

Chrome Version 130.0.6723.70 (Official Build) (64-bit)

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

aarifkhan7:
Hey @cathysarisky,

I'm a beginner interested in contributing to open source, and this issue seems like a good starting point for me. Could you please assign it to me?

Thanks!

cathysarisky:
> I'm a beginner interested in contributing to open source, and this issue seems like a good starting point for me. Could you please assign it to me?

Sorry, I don't have write-permissions on this repo so can't help you there.  The Ghost dev team has said previously that they don't assign issues to new contributors, but you're welcome to put in a PR for it.  

[...]

PR #21681, opened 2024-11-21
fix: make drop target for uploads functional across entire range

harshul786:
### Description of Change  
This PR addresses the issue [[#21439](https://github.com/TryGhost/Ghost/issues/21439)](https://github.com/TryGhost/Ghost/issues/21439), where the drop target for uploads was inconsistent and did not cover the full range of the drop area.  

**Changes made:**  
- Added `w-full` and `h-full` classes to the drop component of the `FileUpload` drag-and-drop.  
- Removed unnecessary padding that created non-functional areas within the drop target.  

### Use Case  
The change ensures that the entire designated area for uploads functions correctly as a drop target, improving the user experience and fixing the reported bug.


cathysarisky:
I'm sorry to report that this does not solve the problem.

9larsons:
Thanks for the submission. This seems to foul up the visual indicator that you're over the drop. We're working on a similar fix that's for more than just the migrations dropdown, so I'm going to close this out, but it should be resolved fairly soon.

PR #21687, opened 2024-11-21
🐛 fix Windows bug with the admin file upload widget

### LINKED ISSUE #21439: Drop target for uploads (content and theme) is broken.

cathysarisky:
closes #21439

On Windows 10/Chrome (but maybe nowhere else?), attempting to drag a file into any of the drop targets in the admin panel resulted in flickering behavior, and generally dropping didn't actually trigger the upload.

I thought originally it was a problem with the size of the drop target, but it actually appears to be a rerender bug.  In brief, handleDragging and handleStopDragging were firing repeatedly, and each fire triggered a rerender, that added or removed a div from the file upload widget.  

I suspect some browser-specific difference in how drag events fire is to blame. 

This PR moves the logic to change the classes applied to the div, rather than changing whether the div is present.

I have manually tested with Windows 10 in the users import, theme import, and content import widgets.  Styles are preserved (although I think they could be improved, as the grey outline is really faint) and uploading now works consistently, instead of mostly triggering display of the raw file most of the time.


apps/admin-x-design-system/src/global/form/FileUpload.tsx
```diff
@@ -51,7 +51,7 @@ const FileUpload: React.FC<FileUploadProps> = ({id, onUpload, children, style, u
 
     return (
         <label className={clsx('relative', className)} htmlFor={id} style={style} onDragEnter={handleDragging} onDragLeave={handleStopDragging} onDragOver={handleDragging} onDrop={handleDrop} {...props}>
-            {isDragging && <div className={clsx('absolute inset-1 rounded border-2 border-dashed border-grey-400/25', dragIndicatorClassName)} />}
+            <div className={clsx({'absolute inset-1 rounded': true, 'border-2 border-dashed border-grey-400/25': isDragging, dragIndicatorClassName: isDragging})} />
```
vikaspotluri123, line 54:
```suggestion
            <div className={clsx({'absolute inset-1 rounded': true, 'border-2 border-dashed border-grey-400/25': isDragging, [dragIndicatorClassName]: isDragging})} />
```

I might be misreading this but it looks like `dragIndicatorClassName` got unintentionally converted from a variable to a literal

cathysarisky: I think we're getting into trouble here because it can be undefined.

vikaspotluri123: _sigh_ You can do this in that case:

```suggestion
            <div className={clsx({'absolute inset-1 rounded': true, 'border-2 border-dashed border-grey-400/25': isDragging, ...dragIndicatorClassName && {[dragIndicatorClassName]: isDragging}})} />
```

vikaspotluri123: Or... something a lot simpler. Apparently clsx supports multiple arguments.

```suggestion
            <div className={clsx({'absolute inset-1 rounded': true, 'border-2 border-dashed border-grey-400/25': isDragging}, isDragging && dragIndicatorClassName)} />
```

cathysarisky: I think one of us is confused (very likely me).  I'm seeing dragIndicatorClassName getting passed in like:
```
                                <ImageUpload
                                    deleteButtonClassName='md:invisible absolute pr-3 -right-2 -top-2 flex h-8 w-10 cursor-pointer items-center justify-end rounded-full bg-[rgba(0,0,0,0.75)] text-white group-hover:!visible'
                                    deleteButtonContent={<Icon colorClass='text-white' name='trash' size='sm' />}
                                    editButtonClassName='md:invisible absolute right-[22px] -top-2 flex h-8 w-8 cursor-pointer items-center justify-center text-white group-hover:!visible z-20'
                                    fileUploadClassName='rounded-full bg-black flex items-center justify-center opacity-80 transition hover:opacity-100 -ml-2 cursor-pointer h-[80px] w-[80px]'
                                    fileUploadProps={{dragIndicatorClassName: 'rounded-full'}}
                                    ```

Comment (vikaspotluri123) [reply apps/admin-x-design-system/src/global/form/FileUpload.tsx]: What you're doing in [7bd16a1](https://github.com/TryGhost/Ghost/pull/21687/commits/7bd16a13c0f1252db0848e1ca856497ada086e8a) is functionally doing the same thing my suggestion is, just a bit more verbose!

Before the commit, you were doing something like this (rewritten to be more obvious):

```jsx
function component(props) {
	const dragIndicatorClassName = props.dragIndicatorClassName;
    const class = clsx({'dragIndicatorClassName': true});
	// reality: no quotes, js implicitly adds quotes for properties (and Python implicitly expects a variable 🙃)
	// expected:          [dragIndicatorClassName] to tell JS it's a variable (as you did)
	return <div className={classes}>Hallo</div>
}

```

With your commit, you should be able to keep destructuring `dragIndicatorClassName` since you also fixed that ^

cathysarisky: Thanks, @vikaspotluri123 !  I did some tidying up of the mess I made trying to fix my error, and I think it's a bit cleaner now.  Manual testing says it's still fixed, on windows/chrome, and tests are passing. :)  I'm going to mark this resolved, unless you see anything else amiss!

vikaspotluri123: It's still perfectly fine, but you should be able to micro-optimize it by using a string instead of array!

`isDragging && [dragIndicatorClassName]`
-->
`isDragging && dragIndicatorClassName`

9larsons:
LGTM. I don't notice a diff on Mac browsers and the performance improvements are a nice bonus. Thanks!

ronaldlangeveld:
Hey @cathysarisky @vikaspotluri123  👋

First off, thank you for the initial fix—it addressed the core issue well!

However, we noticed a side effect that impacted the button’s hover functionality and caused some Admin-X Settings tests to fail. The issues are directly related to this fix. 🙈

Expected behaviour:

https://github.com/user-attachments/assets/eb8abfdc-2618-4b32-bf2a-b6dc7c3da565

Current side effect:

https://github.com/user-attachments/assets/9836bc4d-10bf-4842-89c1-184e07486e34

Upon investigation, it seems the `<div>` element used for the fix is positioned on top of the actual button, which is likely causing the hover issue.

To address this, I made a small adjustment by adding a z-index to the` <div>` element. This fix resolves the broken tests and restores the expected hover behaviour.
Here’s the PR for reference:
[Proposed Fix](https://github.com/TryGhost/Ghost/commit/205e155f05a8ab8656deeb2d08cc5d00f464e8e4)

Unfortunately, I don’t currently have access to a Windows machine to verify that this change doesn’t undo the original fix. Would you mind taking a look and testing it on Windows?

Let me know your thoughts or if there’s anything else you’d like me to adjust.

Thanks again for your help! 🙏🏻

cc @9larsons

PR #21693, opened 2024-11-21
Added prometheus metrics for email analytics event processing

cmraible:
ref https://linear.app/ghost/issue/ENG-1508/add-custom-metrics-for-email-analytics-jobs

- Added two new metrics for the email analytics jobs:
    - `email_analytics_events_processed` — a counter incremented each time an event is processed. Sometimes the event has already been processed in the past, so this doens't always result in a new event being stored in the DB.
    - `email_analytics_events_stored` — a counter incremented each time an event is stored in the DB. For example, if an email is opened 3 times by the same recipient, this counter will only be incremented once.
- The metrics also have a label for the event type, so we can split out opened events from delivered events.


cmraible:
Closing to re-open on upstream branch

PR #21694, opened 2024-11-21
Added email analytics throughput metrics

cmraible:
ref https://linear.app/ghost/issue/ENG-1508/add-custom-metrics-for-email-analytics-jobs

- With the experimental job queue, we're using email analytics as our initial validation test case. We're hoping to see an improvement in Ghost's throughput for ingesting email events. However, we don't currently collect this data point, so it's kind of impossible to tell right now if the job queue is making things better or not.
- This PR fixes that by adding two new prometheus metrics:
    - `email_analytics_events_processed` — a counter incremented each time an event is processed. Sometimes the event has already been processed in the past, so this doens't always result in a new event being stored in the DB.
    - `email_analytics_events_stored` — a counter incremented each time an event is stored in the DB. For example, if an email is opened 3 times by the same recipient, this counter will only be incremented once.
- The metrics also have a label for the event type, so we can split out opened events from delivered events. We can use the `rate()` function in grafana to then get an `events ingested per second` metric, and compare sites with/without the job queue enabled.

### DIFF
--- a/ghost/core/core/server/services/email-analytics/EmailAnalyticsServiceWrapper.js
+++ b/ghost/core/core/server/services/email-analytics/EmailAnalyticsServiceWrapper.js
@@ -32,7 +32,8 @@ class EmailAnalyticsServiceWrapper {
                 EmailRecipientFailure,
                 EmailSpamComplaintEvent
             },
-            emailSuppressionList
+            emailSuppressionList,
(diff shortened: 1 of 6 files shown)


Ghost-Slimer:
Deployed to **staging** with ID: `3150`

PR #21723, opened 2024-11-26
✨Comment importer endpoint

cathysarisky:
WIP. 
This PR provides an endpoint that allows adding a comment to a post, using an Admin API key and specifying a member (by ID, UUID, or email). It supports one level of nesting, matching the behavior of the member/comments endpoint.  

The intent is that it can be used to get comments into Ghost from other sources, including the case of a server move / changing hosts, but this is hopefully flexible enough to be used for importers from other platforms.

Note:  This is not a full importer (nor is there an exporter), but the endpoint that such things could be built upon!

This should be the minimum set of _core_ changes required for a transfer of comments from one Ghost install to another, enabling future additions to the migration tools, importers, etc.  

TODO:  handle in_reply_to.


cathysarisky:
Closing in favor of https://github.com/TryGhost/Ghost/pull/23894

PR #21725, opened 2024-11-26
Bypass 2FA for accounts which haven't yet logged in

sam-lord:
ref ENG-1681


sam-lord:
Schema changes noted here are just a comment addition, so not actually a reason to change this to a minor release

PR #21740, opened 2024-11-27
Extracted duplicate comments test dbFns to shared test util

kevinansfield:
no issue

- we have a `dbFns` object in our e2e test files to facilitate setup for each test
- we have two different APIs that work with comments and so we have two separate tests with the same `dbFns`
- extracted the functions object to a new test utils directory so they can be shared across any test that requires comments db setup and allows the utils to be updated in a single place


ErisDS:
Cleaning up older PRs 💅

PR #21742, opened 2024-11-27
slice theme helper

muratcorlu:
`slice` helper is a very useful helper for theme developers to do stuff while generating their HTML.

most common requirement is having hierarchical navigations. Currenytly many theme developers fixes this with client-side JS which is not a convenient solution.

## Example, partial navigation menu:

Using a part of navigation in a specific place. If a secondary navigation item label starts with `@ `, then list them in a specific place.

```html
<h2>Find me on other places:</h2>
<ul>
  {{#foreach @site.secondary_navigation}}
    {{#match (slice label start=0 end=2) "@ " }}
      <li><a href="{{url}}">{{ slice label start=2 }}</a></li>
    {{/match}}
  {{#foreach}}
</ul>
```

## Example, hierarchical menu

This notation suggests to have a hierarchical menu by navigation labels like below:

```
Home
Services [
Software Development
Marketing
Design
] Contact
```

```html
<ul>
  {{#foreach @site.navigation}}
    {{#match (slice label start=-1) "[" }}
        <li><a href="{{url}}">{{ slice label start=0 end=-2 }}</a></li>
        <ul>
    {{else}}
        {{#match (slice label start=0 end=1) "]" }}
            </ul>
            <li><a href="{{url}}">{{ slice label start=2 }}</a></li>
        {{else}}
             <li><a href="{{url}}">{{ label }}</a></li>
        {{/match}
    {{/match}}
  {{#foreach}}
</ul>
```

There can be also other smart use cases with a naming convention used with internal tags.

I hope you consider my proposal.

Thanks in advance!


cathysarisky:
I think you'll also need to add it to the list of helpers.  See this recent PR for an example:
https://github.com/TryGhost/Ghost/pull/21151/files

muratcorlu:
Thank you @cathysarisky . I updated helper test file as well.

ErisDS:
Sorry we didn't get to this sooner.

I believe the main use case for this helper has been achieved by adding [string comparison](https://docs.ghost.org/themes/helpers/functional/match#string-comparisons) operators to the match helper using the same syntax as [NQL](https://docs.ghost.org/content-api/filtering#operators):

- `~` contains
- `~^` starts with
- `~$` ends with 

If I'm wrong feel free to reopen and we can discuss further.

PR #21756, opened 2024-11-28
Ensured replies remain visible when parent comment is being edited

sanne-san:
REF https://linear.app/ghost/issue/PLG-274/prevent-replies-from-being-hidden-when-editing-parent-comment
- Currently, replies are hidden when the parent comment is being edited. This PR ensures that replies remain visible when the parent comment is being edited.
- To achieve this, the `CommentComponent` now checks if the parent comment is being edited. If so, the comment content is swapped by the EditForm, while the comment's avatar, header, menu, and replies remain visible.
- The Form component now only renders the FormEditor. A FormWrapper component is used to wrap the avatar and comment header.
- This Form component is used in the EditForm without the FormWrapper, as it is already wrapped in the CommentComponent.
- The ReplyForm and the MainForm use the FormWrapper.
- The Avatar component now also accepts a `member` prop, which is used to display the avatar image. This is because it's no longer used inside the Form component.

### DIFF
--- a/apps/comments-ui/test/e2e/options.test.ts
+++ b/apps/comments-ui/test/e2e/options.test.ts
@@ -193,7 +193,15 @@ test.describe('Options', async () => {
 
         test('Uses 100 avatarSaturation', async ({page}) => {
             const mockedApi = new MockedApi({});
-            mockedApi.addComment();
+            mockedApi.addComment({
+                member: {
+                    id: 'test-id',
+                    uuid: 'test-uuid',
+                    name: 'Test User',
+                    avatar: '',
+                    expertise: ''
+                }
+            });
 
             const {frame} = await initialize({
                 mockedApi,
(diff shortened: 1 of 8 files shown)


apps/comments-ui/test/e2e/options.test.ts
```diff
@@ -193,7 +193,15 @@ test.describe('Options', async () => {
...
-            mockedApi.addComment();
+            mockedApi.addComment({
+                member: {
+                    id: 'test-id',
+                    uuid: 'test-uuid',
+                    name: 'Test User',
+                    avatar: '',
+                    expertise: ''
+                }
+            });
```
kevinansfield, line 204: Why was this needed?

sanne-san: This test was failing, and I didn't understand why – I hadn't touched the avatarSaturation, the only thing that changed is that the Avatar component isn't passed in the Form anymore but in FormWrapper. Cursor suggested this because in FormWrapper, I'm passing member directly as a prop to Avatar, while the test is expecting to find an Avatar that's rendered for a comment.

PR #21776, opened 2024-12-02
Fixed last reply cuts off drop menu

ronaldlangeveld:
ref PLG-273

- Resolves an issue where the dropdown menu was being cut off when the last comment was a reply. 
- Adjusted CSS to ensure the dropdown is fully visible in these cases.
- Added more tests to cover cases.


kevinansfield:
Closing in favour of https://github.com/TryGhost/Ghost/pull/21842

PR #21781, opened 2024-12-02
Added highlight animation to replied-to comments

sanne-san:
REF https://linear.app/ghost/issue/PLG-284/add-highlight-to-parent-comment-when-clicking-replied-to-ref
- When clicking on a replied-to reference, you scroll up to the parent comment. To guide the eye, the parent comment is highlighted with a yellow background.


apps/comments-ui/src/components/content/Comment.tsx
```diff
@@ -298,6 +298,32 @@ const CommentHeader: React.FC<CommentHeaderProps> = ({comment, className = ''})
 
         const element = (e.target as HTMLElement).ownerDocument.getElementById(comment.in_reply_to_id);
         if (element) {
+            const contentElement = element.querySelector('[data-testid="comment-content"]');
```
kevinansfield, line 301: We should avoid using test id's in our actual code. If there are elements that need to be selected I think a custom data property would be better.

apps/comments-ui/src/components/content/Comment.tsx
```diff
@@ -298,6 +298,32 @@ const CommentHeader: React.FC<CommentHeaderProps> = ({comment, className = ''})
...
+                    p.appendChild(mark);
+
+                    // Remove the mark and restore original structure after animation
+                    mark.addEventListener('animationend', () => {
+                        while (mark.firstChild) {
+                            p.appendChild(mark.firstChild);
+                        }
+                        mark.remove();
+                    }, {once: true});
+                });
```
kevinansfield, line 324: Manipulating the DOM of a React-rendered component directly is generally a bad pattern because it can be wiped away at any point if something triggers a React render.

The likelihood of an unexpected render may be low enough in this case to not be too much of an issue but it would be good to not introduce discouraged patterns.

kevinansfield: Initial thought is adding a `highlightComment` action that changes a `commentToHighlight` property on our context to the comment's `id` value. Then the `Comment` component can change it's rendering to include the highlight classes.

ronaldlangeveld: Certainly unlikely to have an unexpected render in this use case, but I think worth implementing a React implementation. I'll take a look at this. :)

apps/comments-ui/src/components/content/Comment.tsx
ronaldlangeveld: Removed this with my changes.

apps/comments-ui/test/e2e/actions.test.ts
```diff
@@ -200,21 +200,57 @@ test.describe('Actions', async () => {
...
+        mockedApi.addComment({
+            html: '<p>This is comment 1</p>',
+            replies: [
+                mockedApi.buildReply({
+                    html: '<p>This is a reply to 1</p>'
+                })
+            ]
+        });
+
+        const {frame} = await testReplyToReply(page);
```
kevinansfield, line 232: Why is this needed here? We've already tested that you can reply to a reply in the tests above, I think we can just set up the data that we need in the mockedApi instance so it's easier to see what the expected state is for the test and avoid the extra/unnecessary expectations and waits.

apps/comments-ui/src/AppContext.ts
```diff
@@ -82,6 +83,7 @@ export type EditableAppContext = {
     labs: LabsContextType,
     order: string,
     adminApi: AdminApi | null,
+    replyToHighlight: string | null,
```
kevinansfield, line 86: Can we generalise to `commentToHighlight`? I don't think there's anything specific about replies in the code  other than _right-now_ that's our only use-case, seems like the code could be simplified and made more re-usable in future by allowing any comment to be highlighted.

apps/comments-ui/src/components/content/Comment.tsx
```diff
@@ -309,8 +309,14 @@ const CommentHeader: React.FC<CommentHeaderProps> = ({comment, className = ''})
...
         const element = (e.target as HTMLElement).ownerDocument.getElementById(comment.in_reply_to_id);
-        if (element) {
+        if (element && comment.parent_id) {
+            dispatchAction('setHighlightedRepliedToComment', {
+                replyToId: comment.in_reply_to_id
+            });
             element.scrollIntoView({behavior: 'smooth', block: 'center'});
+            setTimeout(() => {
+                dispatchAction('removeHighlightedRepliedToComment', null);
+            }, 2500);
```
kevinansfield, line 319: Feels like we're spreading knowledge of the styling behaviour around too much. Can we look at alternatives? Two ideas:
1. move the timeout and state update into `setHighlightedRepliedToComment` action so it's not necessary to remember to unset it each time the action is called
2. listen to the animation event somehow to remove the state when the animation finishes - not sure how feasible that will be based on React's rendering approach

apps/comments-ui/src/actions.ts
```diff
@@ -449,7 +466,9 @@ export const Actions = {
     loadMoreReplies,
     updateMember,
     setOrder,
-    openCommentForm
+    openCommentForm,
+    setHighlightedRepliedToComment,
+    removeHighlightedRepliedToComment
```
kevinansfield, line 471: Can we change the action names to hide the internal implementation a bit? `dispatchAction('highlightComment', commentId)` reads a lot nicer 😄

apps/comments-ui/src/components/content/Comment.tsx
ronaldlangeveld: > For 1, I couldn't find an optimal way nor an existing pattern of modifying the state within the action, since it `returns` with the state changes.

I think adding an event listener to an animation would once again bring too much start / end knowledge in, which we want to avoid. 🤔 

Instead created a little reusable hook that will take care of the state before and after the timeout and handle all that for us.

apps/comments-ui/src/utils/hooks.ts
```diff
@@ -57,3 +57,18 @@ export function useRelativeTime(dateString: string) {
...
+        dispatchAction('highlightComment', {commentId});
+
+        // clears the highlight after 2.5 seconds
+        setTimeout(() => {
+            dispatchAction('highlightComment', {commentId: null});
+        }, 2500);
+    };
+
+    return highlightComment;
+};
```
kevinansfield, line 74: Can the `setTimeout` live directly inside the `highlightComment` action to avoid this extra hook? That wouldn't be too much different to how we're planning to make actions update state immediately then re-update once the API request finishes

apps/comments-ui/src/components/content/Comment.tsx
ronaldlangeveld: 💡

kevinansfield: > For 1, I couldn't find an optimal way nor an existing pattern of modifying the state within the action, since it returns with the state changes.

That's fine? We return immediately with the new state but before the return we start a timeout that kicks off a later state change. Or am I oversimplifying? That was my intention with the likes etc, where we update state immediately after kicking off a background task that makes the API request and does a final state update.

apps/comments-ui/src/actions.ts
```diff
@@ -408,6 +408,34 @@ async function openCommentForm({data: newForm, api, state}: {data: OpenCommentFo
...
+}: {
+    data: { commentId: string | null };
+    state: EditableAppContext;
+    setState: (newState: Partial<EditableAppContext>) => void;
+}) {
+    const newState = {
+        ...state,
+        commentIdToHighlight: commentId
+    };
+    setState(newState);
```
kevinansfield, line 424: Can we just return this state so the action doesn't deviate too much from the other actions?

apps/comments-ui/src/actions.ts
```diff
@@ -408,6 +408,34 @@ async function openCommentForm({data: newForm, api, state}: {data: OpenCommentFo
...
+
+    // If `commentId` is not null, schedule the clear action
+    if (commentId) {
+        setTimeout(() => {
+            setState({
+                ...state,
+                commentIdToHighlight: null
+            });
```
kevinansfield, line 432: How about calling another action here rather than setting state directly?

apps/comments-ui/src/components/content/Comment.tsx
ronaldlangeveld: The problem I’ve encountered is that invoking an action within another action (as a background task) doesn’t update the state as intended since it bypasses the dispatcher. We cannot use the dispatcher (dispatchAction) within actions because it’s a hook that’s only available within React components. This seems to be a design limitation in how actions were designed.

That’s why my initial approaches involved either dispatching the setting/unsetting of the highlight within the React component or using a reusable custom hook helper to handle it. I think this is a reasonable approach since hooks are designed for managing UI updates, which is exactly what we’re doing here.

I agree with your concern about keeping the logic within the actions. However, given our constraints, handling the timeout logic in the component or a custom hook seemed to be the most practical / clearest solution for this case.

I think there's definitely room for finding a better pattern when we start working on Likes for example.

kevinansfield: Would it make sense to pass `dispatchAction` in as an argument to `ActionHandler` when called from `dispatchAction`? That would let us trigger actions from within actions 🤔 Although maybe allowing direct state updates inside actions is also useful for other use-cases.

kevinansfield: The reason I like the idea of calling actions from within actions is that it allows us to keep specific behaviours contained and re-usable, e.g. `highlightComment` and `removeHighlightComment` - although maybe not the best example usage 😅

PR #21783, opened 2024-12-02
Added index for redirects lookup

9larsons:
ref https://linear.app/ghost/issue/ENG-1811/
- added first migration for truncating column length; required in order to be able to add an index
- added second migration for creating the index

After sending a newsletter, Ghost can struggle with the amount of incoming traffic (particularly link scanners) with particularly large pools of recipients. Part of this is link lookup, which an index has shown to help with when added manually on the db. Given that, this change is to make it native.

I ran a stats service query to gather data on usage, as well as did code review, and I've not seen any use of this table that isn't using the 8 digit slug, i.e. /r/12345678, or 11 characters total. I made this 191 in case that use case changes, while still allowing us the value of an index.

Note: SQLite does not support altering columns to my knowledge, so we have to create a temp table with the new column data to push back in. This should be ok because use of SQLite should be limited. In the past, I've seen migrations skipped for SQLite and I'd rather not do that here if able.

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.102/2024-12-02-17-32-40-alter-length-redirects-from.js
+++ b/ghost/core/core/server/data/migrations/versions/5.102/2024-12-02-17-32-40-alter-length-redirects-from.js
@@ -10,20 +10,28 @@ module.exports = createNonTransactionalMigration(
         // SQLite doesn't support altering column length
         if (DatabaseInfo.isSQLite(knex)) {
             logging.info('Creating temporary redirects table for SQLite');
-            await knex.schema.createTable('redirects_temp', function (table) {
-                table.string('id').primary();
-                table.string('from', 191).notNullable();
-                table.string('to', 2000).notNullable();
-                table.string('post_id', 24).nullable();
-                table.dateTime('created_at').notNullable();
-                table.dateTime('updated_at').nullable();
-            });
+            try {
+                await knex.schema.dropTableIfExists('redirects_temp');
+                await knex.schema.createTable('redirects_temp', function (table) {
+                    table.string('id').primary();
+                    table.string('from', 191).notNullable();
+                    table.string('to', 2000).notNullable();
+                    table.string('post_id', 24).nullable();
+                    table.dateTime('created_at').notNullable();
+                    table.dateTime('updated_at').nullable();
+                });
 
-            logging.info('Inserting data into temporary redirects table');
-            await knex.raw(`
-                INSERT INTO redirects_temp (id, from, to, post_id, created_at, updated_at)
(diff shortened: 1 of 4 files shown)


9larsons:
Note, knex has a [limit](https://github.com/TryGhost/knex-migrator/issues/51) on the character allowed (191), and MySQL has some limits as well.

ghost/core/core/server/data/migrations/versions/5.102/2024-12-02-17-32-40-alter-length-redirects-from.js
```diff
@@ -0,0 +1,79 @@
...
+                await knex.schema.dropTableIfExists('redirects_temp');
+                throw error; // Re-throw the error to ensure the migration fails
+            }
+
+            logging.info('Dropping original redirects table');
+            await knex.schema.dropTable('redirects');
+
+            logging.info('Renaming temporary redirects table to redirects');
+            await knex.schema.renameTable('redirects_temp', 'redirects');
+        } else {
```
daniellockyer, line 41: I thought Knex does this for us, for SQLite? 🤔

daniellockyer: It seems like it does anyway:

```
[2024-12-03 06:47:19] INFO Altering length of redirects.from
  knex:query BEGIN; trx6 +1ms
  knex:query SELECT type, sql FROM sqlite_master WHERE (type='table' OR (type='index' AND sql IS NOT NULL)) AND tbl_name='redirects' trx6 +0ms
  knex:query COMMIT; trx6 +0ms
  knex:query PRAGMA foreign_keys trx6 +1ms
  knex:query BEGIN; trx7 +0ms
  knex:query CREATE TABLE `_knex_temp_alter017` (`id` varchar(24) NOT NULL, `from` varchar(191) NOT NULL, `to` varchar(2000) NOT NULL, `post_id` varchar(24) NULL, `created_at` datetime NOT NULL, `updated_at` datetime NULL, FOREIGN KEY (`post_id`) REFERENCES `posts` (`id`) ON DELETE SET NULL, PRIMARY KEY (`id`)) trx7 +1ms
  knex:query INSERT INTO "_knex_temp_alter017" SELECT * FROM "redirects"; trx7 +0ms
  knex:query DROP TABLE "redirects" trx7 +0ms
  knex:query ALTER TABLE "_knex_temp_alter017" RENAME TO "redirects" trx7 +1ms
  knex:query COMMIT; trx7 +2ms
```

allouis (commented):
Other than the comment by @daniellockyer this looks good. If we can rely on `knex` to handle the table rename dance for us then we should!

Issue #21414, opened 2024-10-25
Bulk adding tags can create "duplicate" `posts_tags` rows


brvnnghi:
### Issue Summary

Tag count doubled when adding an existing tag in a post again via Context Menu.

The correct behaviour would be, if a tag is already existed in a post, it should not be tagged again using "Add a Tag" in Context Menu.

![img 2](https://github.com/user-attachments/assets/f7cbd31a-3126-4821-9829-e04aa688a1d4)
![img 3](https://github.com/user-attachments/assets/b3bd7656-f6c2-4d90-8090-e05fa220e219)
![img 4](https://github.com/user-attachments/assets/ba97e17c-a6ef-4762-91ad-e84858a40f1a)

### Steps to Reproduce

1. Create new tag
2. Tag inside a post
3. Tag again using Context Menu ("Add a tag")
4. See tag count = 2 when there's only 1 post

### Ghost Version

5.94

### Node.js Version

22

### How did you install Ghost?

Linux / Local

### Database type

MySQL 8

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

kevinansfield:
Thank you for the report.

The problem code is [located here](https://github.com/TryGhost/Ghost/blob/main/ghost/posts-service/lib/PostsService.js#L321-L373). PRs are welcome 🙂 

Aside from the double counting the duplicate post_id+tag_id rows in the `posts_tags` table does not appear to cause any other issues. Deleting the attached tag from the UI correctly deletes both rows.

If you're aware of any other problems this causes please elaborate.

As for the fix there are two approaches:

1. Add a check to the bulk edit feature to skip inserting duplicate rows
2. Add a unique index across the `posts_tags.{post_id,tag_id}` columns so data integrity is ensured at the database level (this will likely also require approach 1)

Navarjun:
Hello there
I will work on this issue, and create a PR soon.

Please let me know if there's something I should watch out for.

Navarjun:
@kevinansfield I found 2 links while trying to do the DB migration: 
1. Ancient: https://ghost.org/changelog/database-migrations/
2. In the docs, but not detailed enough:  https://ghost.org/docs/migration/custom/

Regardless, here's where I am:
I can't find the databaseVersion key in defaultSettings.json. So, I am assuming that was removed some time ago.
[...]

Issue #19871, opened 2024-03-14
Internal tags not all displayed in admin panel


cathysarisky:
### Issue Summary

It is possible to create multiple internal tags with the same name, via minor unintended abuse of the Admin API.  These tags get a -2 appended to them, for example hash-firsttag if accidentally duplicated becomes hash-firsttag-2 (and then hash-firsttag-2-2 if one makes the same API error repeatedly).

Visiting /ghost/#/tags/ shows only the original hash-firsttag tag, with the posts that have that tag listed.  It does not display the additional -2 and -2-2 versions of the tag, making it difficult to see the problem and delete the extra tags.  Given that themes use internal tags for a variety of unexpected purposes, this can cause breakages in weird places.

### Steps to Reproduce

see above.

### Ghost Version

Ghost Pro

### Node.js Version

Ghost Pro

### How did you install Ghost?

Ghost Pro

### Database type

MySQL 8

### Browser & OS version

Chrome, windows 10, recent.

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

mitulagr2:
I could work on this

PR #21787, opened 2024-12-03
🐛 Bulk adding tags created "duplicate" posts_tags rows from editor posts page

Navarjun:
https://github.com/TryGhost/Ghost/issues/21414
The commit includes a migration which adds a unique index to posts_tags table and handles the related error when bulk adding posts.

The DB errors of duplicate insertions (for unique index) are ignored because the tags user wants to add are added/exist. This respects the editor's expectation as posts have the tags as the end result.


Navarjun:
@kevinansfield @ErisDS @cathysarisky I have created a PR for [this bug](https://github.com/TryGhost/Ghost/issues/21414).
Please take a look and let me know if any changes are required.

I did 2 things:
1. Created a migration to add a unique index to post_id and tag_id
2. Updated the #bulkAddTags table to avoid duplicates without throwing an error. Essentially, it will ignore any conflicts because of duplication.

Please merge it soon, if you find it worthy.

ErisDS (changes requested):
Heya, thanks for putting this together 🙏 

Irreversible migrations can only be run in a major update, as they're effectively a breaking change. I would actually expect that to be enforced, and for CI to fail on this PR 🤔 Seems we're missing a check.

It would also be necessary to write a migration that cleans up any existing duplications, otherwise this migration will fail for anyone who has already run into this.

These migrations would need to live in `ghost/core/core/server/data/migrations/versions/6.0` and be held back on a branch until such time as we cut the major.

Not sure if there's maybe an alternative fix for the short term?

Especially considering this is not only related to https://github.com/TryGhost/Ghost/issues/21414 but also to https://github.com/TryGhost/Ghost/issues/19871, which needs a UI-based fix I think.

PR #21788, opened 2024-12-03
Added loading state when ordering comments

ronaldlangeveld:
PLG-280

- Added a loading state implementation when changing the ordering of comments.
- This improves the overall UX particularly with slower connections.
- Due to the nature of how comments and ordering are handled, we approached it with a simple state that determines whether it's done loading or not around the API query.


apps/comments-ui/src/actions.ts
```diff
@@ -23,18 +23,27 @@ async function loadMoreComments({state, api, options, order}: {state: EditableAp
 }
 
 async function setOrder({state, data: {order}, options, api}: {state: EditableAppContext, data: {order: string}, options: CommentsOptions, api: GhostApi}) {
-    let data;
+    state.commentsIsLoading = true;
```
kevinansfield, line 26: This feels like a React anti-pattern, should we be updating a context-provided state object directly?

PR #21801, opened 2024-12-04
Removed link from replied-to ref when deleted/hidden

kevinansfield:
closes https://linear.app/ghost/issue/PLG-294

- making the `replied to: [removed]` text a link was a bit confusing because clicking it does nothing
- if the replied-to comment doesn't exist (e.g. hidden/deleted and not returned in API response) or has been unpublished we replace the `<a>` with a `<span>` to remove the link behaviour


sanne-san (approved):
Pushed a small change that removes the hover effect for non-links, but otherwise LGTM!

PR #21806, opened 2024-12-04
Modified `lastSeenAt` updater to use job queue for member link click events

9larsons:
ref https://linear.app/ghost/issue/ENG-1815/create-job-queue-entry-for-lastseenatupdater
ref https://linear.app/ghost/issue/ENG-1816/implement-domainevents-in-jobqueuemanager

This change requires a few pieces that I'll likely split out into other PRs to isolate the changes. This is a single PR spike for demonstration purposes.

The `lastSeenAt` update for members after clicking an email link is not a time-sensitive operation. Given that the period following the sending a newsletter is already one of incredibly high activity for Ghost, it makes sense to shift this workload to the job queue where it can be processed in turn with other time-insensitive operations.

Ultimately, we've got a few new patterns here that I'd like to get some feedback on.
- This approach firms up the groundwork laid in the EmailAnalyticsJob-related updates that we introduced the job queue for, where **service wrappers w/in `ghost/core` are the ones listening to `DomainEvents` and submit the jobs to the `JobManager`.**
- Jobs have a new return pattern, if needed, to be able to submit events upon completion (we have two flavors in Ghost, `DomainEvents` and 'Node events', e.g. `site.changed`). In general, the return value is otherwise disregarded, so this is only important if needed to signal other activities like chaining job submissions. In pseudocode typing: `result = {result: <any>, eventData: <domainEvents || nodeEvents>}`
- Jobs should be contained within the `ghost/core/core/services/<service>/jobs` directories.
- Jobs may require new `Events` to be created and subscribed to. These can be created within the appropriate events package, e.g. `members-events`. I don't see a need to create a job-specific bundle of events.
- Jobs should only interact with `knex`/the DB directly, not using models or services. This reduces the overhead and paves the way for creation of standalone services.

Note: I'm not sure if we want to maintain two code paths for 'job queue enabled' vs 'job queue disabled'. Either we should let the `JobManager` route it to the job queue or inline runner on its own, or simply widespread enable the job queue.

### DIFF
--- a/ghost/core/core/server/services/jobs/job-service.js
+++ b/ghost/core/core/server/services/jobs/job-service.js
@@ -8,6 +8,7 @@ const logging = require('@tryghost/logging');
 const models = require('../../models');
 const sentry = require('../../../shared/sentry');
 const domainEvents = require('@tryghost/domain-events');
+const eventEmitter = require('../../lib/common/events');
 const config = require('../../../shared/config');
 const prometheusClient = require('../../../shared/prometheus-client');
 const errorHandler = (error, workerMeta) => {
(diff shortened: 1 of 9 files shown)


9larsons:
@ErisDS Is there a more correct name for the events lib we have that uses Node's event module? We have both that and DomainEvents, and it'd be nice to have a better way to semantically separate them. It's a bit confusing as is.

PR #21810, opened 2024-12-05
Add saveRaw method to LocalFilesStorage and LocalMediaStorage classes

PaulAdamDavis:
ref https://linear.app/ghost/issue/CON-3/external-media-inliner-should-inline-more-file-types

The external media inliner requests each file as a buffer. Saving this works well for images, but not media or files as the `saveRaw` method is not available on those storage adapters.

This PR adds the same `saveRaw` method from `LocalImagesStorage` to two related classes, which enables saving buffers for all other supported file types

### DIFF
--- a/ghost/core/core/server/adapters/storage/LocalFilesStorage.js
+++ b/ghost/core/core/server/adapters/storage/LocalFilesStorage.js
@@ -1,16 +1,50 @@
 // # Local File System Storage module
 // The (default) module for storing media, using the local file system
+const fs = require('fs-extra');
+const path = require('path');
 const config = require('../../../shared/config');
+
 const constants = require('@tryghost/constants');
+const urlUtils = require('../../../shared/url-utils');
 const LocalStorageBase = require('./LocalStorageBase');
+
+const messages = {
+    notFound: 'File not found',
+    notFoundWithRef: 'File not found: {file}',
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/adapters/storage/LocalFilesStorage.js
```diff
@@ -1,17 +1,51 @@
...
+const urlUtils = require('../../../shared/url-utils');
 const LocalStorageBase = require('./LocalStorageBase');
 
+const messages = {
+    notFound: 'File not found',
+    notFoundWithRef: 'File not found: {file}',
+    cannotRead: 'Could not read File: {file}'
+};
+
 class LocalFilesStorage extends LocalStorageBase {
```
daniellockyer, line 17: Given that `LocalFilesStorage`, `LocalMediaStorage` and `LocalImagesStorage` are all __extremely__ similar now, we could probably move `saveRaw` up to `LocalStorageBase` and DRY up the implementation of `saveRaw`? 🙂

PaulAdamDavis: Solid point. Done!

daniellockyer (approved):
LGTM (have not tested intended functionality though)

PR #21816, opened 2024-12-05
Removed Beta flag for Custom Fonts

peterzimon:
ref https://linear.app/ghost/issue/DES-986/move-feature-to-ga-after-two-weeks-of-wait

Next week we're rolling out Custom fonts for customers. This commit removes the beta flag for this feature.


ghost/core/test/unit/frontend/helpers/body_class.test.js
```diff
@@ -216,17 +215,6 @@ describe('{{body_class}} helper', function () {
             rendered.string.should.equal('post-template tag-foo tag-bar gh-font-heading-space-grotesk gh-font-body-noto-sans');
         });
 
-        it('does not include custom font classes when custom fonts are not enabled', function () {
```
9larsons, line 219: Would this test still be worth keeping but just removing the labs flag?

apps/admin-x-settings/src/components/settings/site/designAndBranding/ThemeSettings.tsx
```diff
@@ -70,11 +68,9 @@ const ThemeSettings: React.FC<ThemeSettingsProps> = ({sections, updateSetting})
 
                             // hides typography related theme settings from official themes
                             // should be removed once we remove the settings from the themes in 6.0
-                            if (hasCustomFonts) {
-                                const hidingSettings = themeSettingsMap[activeThemeName];
-                                if (hidingSettings && hidingSettings.includes(setting.key) && activeThemeAuthor === 'Ghost Foundation' && supportsCustomFonts) {
-                                    spaceClass += ' hidden';
-                                }
+                            const hidingSettings = themeSettingsMap[activeThemeName];
+                            if (hidingSettings && hidingSettings.includes(setting.key) && activeThemeAuthor === 'Ghost Foundation' && supportsCustomFonts) {
```
9larsons, line 72: So we should be hiding the block if the theme doesn't support it, based on [this](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-settings/src/hooks/useCustomFonts.tsx#L7) line. Then it makes sense to not have an error message.

9larsons: Actually, this block is confusing to me. Why are we only hiding the settings for official themes?

ghost/core/test/unit/frontend/helpers/ghost_head.test.js
```diff
@@ -1173,30 +1170,7 @@ describe('{{ghost_head}} helper', function () {
             }));
         });
 
-        it('does not include custom font when not set', async function () {
```
9larsons, line 1176: Same for this one.

9larsons (commented):
Few questions.

apps/admin-x-settings/src/components/settings/site/designAndBranding/ThemeSettings.tsx
minimaluminium: We have font-related custom font settings on almost all our official themes. We want to hide them when official themes support the new custom fonts feature, because it'd be confusing for users to have two font-related settings — one in the brand settings, and the other one in the theme settings.

PR #21292, opened 2024-10-14
[WIP] Upgrade Yarn to Yarn 4.5.0

ronaldlangeveld:
Have some free time during a 6 hour layover in Dubai, en route from Scotland to Korea.
Decided to give upgrading yarn a stab.

PR #21818, opened 2024-12-06
Yarn 4 upgrade fixes

crutchcorn:
This PR gets Yarn 4 (Berry) into a working state, based off of the work done by @ronaldlangeveld in #21292

It does this by:

- Adding missing packages
- Resolving things manually as-needed
- Moved things from the Ghost ESLint plugin to the root deps temporarily while I wait for some feedback from @joshuakgoldberg
- Resolving merge conflicts
- Upgrading Yarn to fix issues with lockfile generation
- Clearing the lockfile to fix issues with lockfile


ghost/announcement-bar-settings/package.json
```diff
@@ -21,6 +21,5 @@
     "c8": "8.0.1",
     "mocha": "10.2.0",
     "sinon": "15.2.0"
-  },
-  "dependencies": {}
+  }
```
JoshuaKGoldberg, line 24: This gave me an idea 😄 thank you!

https://github.com/JoshuaKGoldberg/eslint-plugin-package-json/issues/683

package.json
```diff
@@ -116,17 +122,27 @@
   },
   "devDependencies": {
     "@actions/core": "1.11.1",
+    "@kapouer/eslint-plugin-no-return-in-loop": "1.0.0",
+    "@typescript-eslint/eslint-plugin": "6.9.1",
+    "@typescript-eslint/parser": "6.9.1",
```
JoshuaKGoldberg, line 127: `eslint-plugin-ghost` has specific versions of tseslint packages, and is actually on the latest 8.17.0:

https://github.com/TryGhost/eslint-plugin-ghost/blob/0bc3217ebe1409c12496860bbef37c6d4bb3eabf/package.json#L36-L37

Having a different version here will give you two versions. https://github.com/eslint/eslint/discussions/15681 might be the issue?

I'm not super confident this is what you're seeing and haven't set all this stuff up locally to try. If this isn't the right fix, let me know and I can poke deeper.

crutchcorn: Ahh, I see the confusion this is my bad.

A few things to note:

1) The `TryGhost/Ghost` repo is using an older version of `eslint-plugin-ghost` that includes these specific versions
2) I'm hoping to remove the need for the "install in the monorepo" `peerDeps`-style dep sharing.

Instead, I'd love to `npm i eslint-plugin-ghost` and have it handle the versions of `@typescript-eslint` and such for us

JoshuaKGoldberg: Alternative, are there blockers to just using flat config? 😁

JoshuaKGoldberg: Anyway, what folks normally do is `require.resolve("@typescript-eslint/eslint-plugin)` inside the plugin. That way it resolves to the specific string path within your `node_modules/eslint-plugin/ghost/node_modules/` (if that's not deduplicated to something higher up). I'll try that out now...

JoshuaKGoldberg: I, ah, don't know why this isn't working: https://github.com/JoshuaKGoldberg/eslint-plugin-ghost/tree/require-resolve.

Steps to repro what I thought would fix things:

1. Check out that branch locally
2. Link the `Ghost` repo to that branch with `"eslint-plugin-ghost": "file:../eslint-plugin-ghost"` in the `package.json`'s `dependencies` 
3. Remove all` parser: '@typescript-eslint/parser'` lines from ESLint config files in the `Ghost` repo

I do see console logs from the local package. But:

```plaintext
 Oops! Something went wrong! :(
      
      ESLint: 8.44.0
      
      ESLint couldn't find the plugin "@typescript-eslint/eslint-plugin".
      
      (The package "@typescript-eslint/eslint-plugin" was not found when loaded as a Node module from the directory "/Users/josh/repos/Ghost/ghost/nql-filter-expansions".)
      
      It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:
      
          npm install @typescript-eslint/eslint-plugin@latest --save-dev
      
      The plugin "@typescript-eslint/eslint-plugin" was referenced from the config file in ".eslintrc.js » plugin:ghost/ts".
      
      If you still can't figure out the problem, please stop by https://eslint.org/chat/help to chat with the team.
```

21/95 targets fail in `yarn lint`. 🤔

ErisDS:
Thanks for doing this! We're not quite ready for yarn 4 yet (or sure if we want yarn v4 or pnpm) - and there's lots of other tooling that would have to be updated to match (like Ghost-CLI) in order to change it, however once we get chance to focus on updating this will be an invaluable reference.

PR #21822, opened 2024-12-06
Update comments.json

echobilisim3421:
Added missing translations


ghost/i18n/locales/tr/comments.json
```diff
@@ -29,7 +29,7 @@
     "Founder @ Acme Inc": "Kurucu @ Firma Ünvan",
     "Full-time parent": "Tam zamanlı ebeveyn",
     "Head of Marketing at Acme, Inc": "Firma, Ünvan'da Pazarlama Müdürü",
-    "Hidden for members": "",
+    "Hidden for members": "Üyeler için gizli",
     "Hide": "Gizle",
     "Hide comment": "Yorumu gizle",
     "Jamie Larson": "Ad Soyad",
```
cathysarisky, line 35: We had a discussion on another PR (although maybe it was for portal) about how this should be     "Jamie Larson": "Ahmet Yılmaz",
    "jamie@example.com": "ahmet@example.com",
    
    Would you like to update this so that the name placeholder is consistent between portal and comments?

cathysarisky (approved):

ghost/i18n/locales/tr/comments.json
cathysarisky, line 35:
```suggestion
    "Jamie Larson": "Ahmet Yılmaz",
```

cathysarisky (approved):

ghost/i18n/locales/tr/comments.json
echobilisim3421: Yes, the translation written there is incorrect. It wouldn't be Ahmet Yılmaz. It would be more logical to write your username instead.

echobilisim3421: > We had a discussion on another PR (although maybe it was for portal) about how this should be "Jamie Larson": "Ahmet Yılmaz" Would you like to update this so that the name placeholder is consistent between portal and comments?

Even though I said the translation was wrong, you changed it again. I don't want someone else's name and surname to be written on my site, I'm sure there are many people who think like me, as I said, it is more appropriate to write Name Surname instead of someone's name and surname, accordingly, the person who registers will write their name and surname.

PR #21830, opened 2024-12-08
Update search.json

alexrodba:
Added missing translations

### DIFF
--- a/ghost/i18n/locales/ca/search.json
+++ b/ghost/i18n/locales/ca/search.json
@@ -1,9 +1,9 @@
 {
-    "Authors": "",
-    "Cancel": "",
-    "No matches found": "",
-    "Posts": "",
-    "Search posts, tags and authors": "",
-    "Show more results": "",
-    "Tags": ""
+    "Authors": "Autors",
+    "Cancel": "Cancel·lar",
+    "No matches found": "No s'ha trobat cap resultat",
+    "Posts": "Articles",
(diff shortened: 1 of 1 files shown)


cathysarisky:
Question:  ChatGPT suggests that the verb tenses should be imperative rather than infinitive ( cerca vs cercar , cancel-la vs cancel-lar), since these are UI actions or buttons.  I know that's broadly true for some related languages, but I don't know how Catalan does it.  A quick reply that you've considered it and infinitive is correct would be appreciated.

cathysarisky:
Hi @Zagur ! Do you have a moment to review these Catalan refinements?

alexrodba:
> Question: ChatGPT suggests that the verb tenses should be imperative rather than infinitive ( cerca vs cercar , cancel-la vs cancel-lar), since these are UI actions or buttons. I know that's broadly true for some related languages, but I don't know how Catalan does it. A quick reply that you've considered it and infinitive is correct would be appreciated.

Yeah, it's a country-wide issue. According to Softcatalà [seems like Imperative is preferred although infinitive is correct](https://www.softcatala.org/forum/viewtopic.php?f=4&t=269). I like infinitive more because it doesn't offer the tu vs vostè debate (Cercar is more inclusive than "cerca" or "cerqueu") but I know we're just a minority thinking like that.

I'll adapt to infinitive for the time being, and leave this discussion for future debate if need be.

cathysarisky (approved):

PR #21842, opened 2024-12-09
Fixed dropdown menu being cut off in comments-ui

kevinansfield:
closes https://linear.app/ghost/issue/PLG-273

- removed previous fix which only worked on last comment but not last reply
  - keeping track of last comment/reply spread a lot of domain knowledge around for a UI-only concern and wouldn't scale if we have other dropdowns in the future
- added `useOutOfViewportClasses` hook
  - accepts an object with top/bottom/right/left containing default and outOfViewPort classes
  - applies the correct classes using the DOM rather than React so that we avoid re-renders and associated flickering or broken rendering


sanne-san (approved):
LGTM! 🙌

PR #21857, opened 2024-12-10
Changed location of built public content to temporary dir

mike182uk:
refs [ONC-662](https://linear.app/ghost/issue/ONC-662/fix-file-write-issues-in-ghost-application-related-to-asset-generation)

Changed location of built public content to temporary dir to circumvent file writing issues in some environments (i.e gluster). This is a stop-gap measure until we get to refactoring the generation of the built public content.

### DIFF
--- a/ghost/core/core/shared/config/helpers.js
+++ b/ghost/core/core/shared/config/helpers.js
@@ -1,3 +1,5 @@
+const crypto = require('crypto');
+const os = require('os');
 const path = require('path');
 const {URL} = require('url');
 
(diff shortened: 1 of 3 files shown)


ghost/core/core/shared/config/helpers.js
```diff
@@ -64,6 +66,28 @@ const isPrivacyDisabled = function isPrivacyDisabled(privacyFlag) {
...
+ */
+function getProcessTmpDirPath(config) {
+    // Memoize the computed path to avoid re-computing it on each call - The
+    // value should not change during the lifetime of the process.
+    if (processTmpDirPath === null) {
+        const siteHash = crypto.createHash('md5')
+            .update(config.getSiteUrl())
+            .digest('hex');
+
+        processTmpDirPath = path.join(os.tmpdir(), `ghost_${siteHash}_${process.pid}`);
```
daniellockyer, line 85: Is it worth just simplifying this to use use a randomly generated UUID?

mike182uk: Ah yeh ofc, since adding the memoization, its fine to use something random here 👍

mike182uk: Updated in https://github.com/TryGhost/Ghost/pull/21857/commits/0e4fde065ad65035d691697ec7d42ed75bd3f4d9

PR #21861, opened 2024-12-11
Updated Comments likes UI changes to be instant

ronaldlangeveld:
ref PLG-288

- Implemented instant UI updates for comment likes/unlikes using optimistic rendering.
- Enhanced error handling reverts state on API failure, ensuring consistency. 
- Added new test helper to mock failed API requests, needed to test revert state handling.


apps/comments-ui/src/actions.ts
```diff
@@ -186,9 +186,29 @@ async function showComment({state, api, data: comment}: {state: EditableAppConte
...
+        await api.comments.like({comment: {id: comment.id}});
+        return {
+            commentLikeLoadingId: null
+        };
+    } catch (err) {
+        // if error we revert the state
+        return {
+            comments: commentsState,
+            commentLikeLoadingId: null
+        };
```
kevinansfield, line 202: This seems dangerous as other state changes could have happened if there was a long delay before the endpoint eventually errored. I would expect a rollback to only occur for the liked status of the comment under change, would that make things considerably more difficult?

apps/comments-ui/src/actions.ts
```diff
@@ -222,19 +242,33 @@ async function likeComment({state, api, data: comment}: {state: EditableAppConte
...
-    await api.comments.report({comment});
+async function sendUnlikeToApi({api, data: comment}: {api: GhostApi, data: {id: string, commentsState: Comment[]}}) {
+    const commentsState = comment.commentsState;
 
-    return {};
+    try {
+        await api.comments.unlike({comment: {id: comment.id}});
+        return {
+            commentUnlikeLoading: null
+        };
```
kevinansfield, line 257: Do we need this? It feels clunky.

Can we keep `likeComment` as it was where it still awaits the API response but first dispatches an action that does an immediate state update? That way the component that dispatches `likeComment` can await as normal and manage the button disabling with a typical wrapper around an async action but the state still updates immediately in the background to show the like/unlike status

kevinansfield: Pseudo-code:

```
async function likeComment(id) {
    dispatchAction('updateCommentLikeState', {id, liked: true});
    try {
        await api.likeComment(id);
    } catch () {
        dispatchAction('updateCommentLikeState', {id, liked: false});
    }
}
```

kevinansfield (approved):

PR #21867, opened 2024-12-11
Create verified session for reset password

vershwal:
Fixes https://linear.app/ghost/issue/ENG-1855/fix-reset-password-flow-in-case-of-2fa

- User is getting stuck on reset password page and they receive a verify token on the mail and nothing happens on the UI. 
- This is happening because the session is not verified in the case of the reset password flow. The backend is sending a code to verify the session, but we don't have the UI in place where the user can enter the code.
- Previously, the user would be redirected to the dashboard once they clicked on "Save new password."
- Now when 2FA is on, we should directly redirect to the dashboard as the reset password link was sent to the email address anyway. This is why we are creating an already verified session in this case.


ghost/admin/app/authenticators/cookie.js
```diff
@@ -33,6 +33,9 @@ export default Authenticator.extend({
         }
 
         const data = {username: identification, password};
+        if(key) {
```
sam-lord, line 36: With the above, if it was `skipEmailVerification`, this could be:

```
if (skipEmailVerification) {
    data.skipEmailVerification = true;
}
```

ghost/core/core/server/api/endpoints/session.js
```diff
@@ -33,6 +34,8 @@ const controller = {
         return models.User.getByEmail(object.username).then((user) => {
             if (user && !user.hasLoggedIn()) {
                 skipVerification = true;
+            } else if (frame.data.key === 'password_reset') {
```
sam-lord, line 37: Should make this simpler too -- instead of caring about a string value, just saying `} else if (frame.data.skipEmailVerification) {`

sam-lord (commented):
I think this is a really logical approach, and it's just the naming which is making it unclear what the usage of `key` is until you've searched the whole project for where it's used.

PR #21878, opened 2024-12-12
Adding Posts analytics React app

peterzimon:
ref https://linear.app/ghost/issue/DES-1021/create-posts-app

Part of establishing React patterns in Ghost is to build a well-defined and fairly self-encapsulated app through which we can test assumptions and define best practices. Our guinea pig is Post analytics for this purpose. This PR creates a new React app (posts) using Shade (the new design system).


apps/admin-x-framework/package.json
```diff
@@ -111,7 +111,8 @@
         "dependsOn": [
           "test:unit",
           "^build",
-          "@tryghost/admin-x-design-system:test:unit"
+          "@tryghost/admin-x-design-system:test:unit",
+          "@tryghost/shade:test:unit"
```
allouis, line 115: This means that if the shade tests fail the admin-x-framework tests will fail too, is that what you want?

apps/admin-x-framework/src/test/render-shade.tsx
```diff
@@ -0,0 +1,62 @@
...
+                    },
+                    onDelete: () => {},
+                    onInvalidate: () => {},
+                    onUpdate: () => {}
+                }}
+                {...props}
+            />
+        </React.StrictMode>
+    );
+}
```
allouis, line 62: Missing newline here - do you have the EditorConfig plugin installed in your editor?

apps/admin-x-framework/vite.config.ts
```diff
@@ -10,6 +11,11 @@ export default (function viteConfig() {
         plugins: [
             react()
         ],
+        resolve: {
+            alias: {
+                '@': path.resolve(__dirname, './src')
```
allouis, line 16: I don't see this alias used anywhere, can we revert the changes in this file?

allouis (commented):
Left a few comments - overall looks good but there are a lot of files with missing newlines at the end, which I think they should have - EditorConfig should be enforcing that for us.

Also the change to vite config in the admin-x-framework package, I'm not sure I fully understand that

apps/shade/src/components/layout/heading.tsx
```diff
@@ -0,0 +1,60 @@
+import {cn} from '@/lib/utils';
```
allouis, line 1: Oh I see the prefix is used here - this is a new pattern, am interested in discussing it further!

apps/shade/package.json
```diff
@@ -67,13 +67,14 @@
     "@ebay/nice-modal-react": "1.2.13",
     "@radix-ui/react-avatar": "1.1.0",
     "@radix-ui/react-checkbox": "1.1.1",
+    "@radix-ui/react-dropdown-menu": "^2.1.3",
```
daniellockyer, line 70:
```suggestion
    "@radix-ui/react-dropdown-menu": "2.1.3",
```

apps/shade/package.json
```diff
@@ -67,13 +67,14 @@
...
     "@radix-ui/react-form": "0.0.3",
     "@radix-ui/react-popover": "1.1.1",
     "@radix-ui/react-radio-group": "1.2.0",
     "@radix-ui/react-separator": "1.1.0",
     "@radix-ui/react-slot": "^1.1.0",
     "@radix-ui/react-switch": "1.1.0",
-    "@radix-ui/react-tabs": "1.1.0",
+    "@radix-ui/react-tabs": "^1.1.2",
```
daniellockyer, line 77:
```suggestion
    "@radix-ui/react-tabs": "1.1.2",
```

apps/admin-x-framework/package.json
peterzimon: Hm... Maybe not yet. We just added the current design system to it to catch DS bugs, but maybe we can leave Shade out until its first release. Good point.

apps/admin-x-framework/src/test/render-shade.tsx
peterzimon: Good point, thx, fixed

Issue #21886, opened 2024-12-15
Cache control directives stale-while-revalidate and stale-if-error for frontend do not get sent even when Ghost config is set


JamesMarino:
### Issue Summary

Cache control directive `stale-while-revalidate` and `stale-if-error` for frontend do not get sent as header directives even when passed in configuration values for `caching.frontend.staleWhileRevalidate` and `caching.frontend.staleIfError`.

Please note code has not been allowed specifically for `staleIfError` cache control directive config but should be included due to its usage in conjunction with `staleWhileRevalidate`. Please see MDN docs here on [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) for further information.

Initial changes have been made here it seems https://github.com/TryGhost/Ghost/pull/15734 and I can submit a fix shortly as a PR.

Also, documentation has also not been updated for [caching](https://ghost.org/docs/config/#caching), whereabouts can changes to documentation be submitted?

### Steps to Reproduce

1. Initial Setup of the Ghost site locally 
```sh
yarn setup
```

2. Run the site with a custom `maxAge` to ensure environment variables are being passed through and with a custom `staleWhileRevalidate` value to show that this directive is not being received in the `Cache-Control` header 
```sh
export caching__frontend__maxAge=1800
export caching__frontend__staleWhileRevalidate=604800
yarn dev
```

3. Send off a request to the home page and initial post to show it's not being sent
```sh
curl -s -i http://localhost:2368/ | grep -i "Cache-Control"

# Output:
# Cache-Control: public, max-age=1800
```

4. Try another post page as well
```sh
curl -s -i http://localhost:2368/coming-soon/ | grep -i "Cache-Control"

# Output:
# Cache-Control: public, max-age=1800
```

### Ghost Version

v5.105.0

### Node.js Version

v18.20.4

### How did you install Ghost?

Local Development on MacOS 15.1.1

### Database type

MySQL 5.7

### Code of Conduct

- [X] I agree to be friendly and polite to people in this repository

JamesMarino:
No worries @ErisDS - thanks for the feedback.

I thought it would be a good idea to bring the caching config inline with the [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) HTTP spec which seems to be implemented and respected by various CDN providers.

[...]

PR #21887, opened 2024-12-16
Added stale-if-error and fixed bug for frontend cache staleWhileRevaidate config (#21886)

JamesMarino:
This PR fixes the `staleWhileRevaidate` config not being passed into the `cacheControl()` middleware if it is set via the frontend caching config. It also adds the `staleIfError` configuration to allow the `stale-if-error` `Cache-Control` header directive.

Tests have also been added for varying cases of the directives including when the Ghost config is not set.

Documentation from https://ghost.org/docs/config/#caching will also need to be updated. Happy to update this given the source.

- [X] There's a clear use-case for this code change, explained above

Closes issue https://github.com/TryGhost/Ghost/issues/21886

### DIFF
--- a/ghost/core/core/frontend/web/middleware/frontend-caching.js
+++ b/ghost/core/core/frontend/web/middleware/frontend-caching.js
@@ -15,7 +15,7 @@ function calculateMemberTier(member, freeTier) {
     const activeSubscriptions = member.subscriptions.filter(sub => sub.status === 'active');
     if (activeSubscriptions.length === 0) {
         return freeTier;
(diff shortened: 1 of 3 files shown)


ErisDS:
Sorry we didn't get around to reviewing this before we made a bit of a structural change to the codebase (all the packages were merged back into ghost/core) and today I'm having a bit of an old-PR clean up to try to make the list more maintainable.

This change looks valuable though, so if you'd like to re-open it I'll prioritise reviewing it :)

PR #21891, opened 2024-12-16
Used theme version for checking custom fonts support

minimaluminium:
ref DES-1011


ErisDS:
Cleaning up older PRs 💅

PR #21895, opened 2024-12-17
Fixed layout shift issue when Portal popup appears

minimaluminium:
ref DES-547

- when Portal popup is opened and the browser scroll bar is visible, it used to make layout shift, because we were hiding the scrollbar
- now it applies right margin to body element and the trigger button by calculating the scrollbar width only when the browser scroll bar is visible
- it also preservers the current right margin for those elements and makes the calculation based on that

### DIFF
--- a/apps/portal/src/App.js
+++ b/apps/portal/src/App.js
@@ -87,7 +87,11 @@ export default class App extends React.Component {
                 } else {
                     /** When the modal is hidden, reset overflow property for body */
                     window.document.body.style.overflow = this.bodyScroll || '';
-                    window.document.body.style.marginRight = this.bodyMargin || '';
+                    if (!this.bodyMargin || this.bodyMargin === '0px') {
+                        window.document.body.style.marginRight = '';
+                    } else {
+                        window.document.body.style.marginRight = this.bodyMargin
+                    }
                 }
             } catch (e) {
                 /** Ignore any errors for scroll handling */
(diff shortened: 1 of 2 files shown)


apps/portal/src/App.js
```diff
@@ -115,6 +124,27 @@ export default class App extends React.Component {
...
+
+        // Create an inner div
+        const inner = document.createElement('div');
+        outer.appendChild(inner);
+
+        // Calculate the width difference
+        const scrollbarWidth = outer.offsetWidth - inner.offsetWidth;
+
+        // Clean up
+        outer.parentNode.removeChild(outer);
```
sagzy, line 143: Small improvement idea: use `clientWidth` instead of creating an inner element:

```
    ...
    document.body.appendChild(div);
    
    const scrollbarWidth = div.offsetWidth - div.clientWidth;
    
    document.body.removeChild(div);
    ...
 ```

sagzy (approved):
LGTM! Added a comment for a tiny optimisation, but feel free to ignore :)

stromfeldt:
Thanks for this fix! I [mentioned the issue a couple of years ago on the forum](https://forum.ghost.org/t/opening-portal-causes-content-to-jerk/30212) (along with my piss-poor fix), but it apparently couldn't be reproduced for some reason. Your code looks far superior than mine, so I'll remove my code from the sites I've inserted it into as soon as the fix goes live. :)

minimaluminium:
@stromfeldt Your suggested solution is very simple and nice, however, the only thing with it was we still wanted to lock the vertical scrolling when the popup is open.

apps/portal/src/App.js
minimaluminium: Great idea! Added the optimization [here](https://github.com/TryGhost/Ghost/pull/21895/commits/fb2bdac616ea34ed7fbee69278068c29750eb03a#diff-e1eb9ae4190f8d5a9a3d7401f75a79cc08e8d12e7e175c9104b91a648aa845f9R132).

PR #21903, opened 2024-12-17
Filtered notifications that show in `admin-x-activitypub`

mike182uk:
refs [AP-627](https://linear.app/ghost/issue/AP-627/sanitising-note-content)

Adding client side filtering of the notifications in `admin-x-activitypub` - This is a stop-gap until we have a dedicated endpoint for returning notifications. Filtering includes:

- Do not show our own likes
- Only show `like` notifications when it is on our own posts
- Only show replies to our own posts

### DIFF
--- a/apps/admin-x-activitypub/src/components/Activities.tsx
+++ b/apps/admin-x-activitypub/src/components/Activities.tsx
@@ -224,6 +224,17 @@ const Activities: React.FC<ActivitiesProps> = ({}) => {
                 }
 
                 return true;
+            })
+            // Remove our own create activities
+            .filter((activity) => {
+                if (
+                    activity.type === ACTIVITY_TYPE.CREATE &&
+                    activity.actor?.id === userProfile?.id
+                ) {
+                    return false;
+                }
+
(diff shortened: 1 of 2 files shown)


apps/admin-x-activitypub/src/components/Activities.tsx
```diff
@@ -164,19 +176,55 @@ const Activities: React.FC<ActivitiesProps> = ({}) => {
...
     const maxAvatars = 5;
 
+    const {data: userProfile, isLoading: isLoadingProfile} = useUserDataForUser(user) as {data: ActorProperties | null, isLoading: boolean};
+
     const {getActivitiesQuery} = useActivitiesForUser({
         handle: user,
         includeOwn: true,
         includeReplies: true,
         filter: {
-            type: ['Follow', 'Like', `Create:Note:isReplyToOwn`]
```
mike182uk, line 172: `isReplyToOwn` doesn't appear to be working correctly, and will be removed when we switch to a dedicated notifications endpoint

PR #21904, opened 2024-12-17
🌐 Added and corrected strings in the Norwegian i18n locale files.

knutid:
no issue
- added missing strings
- corrected one string


cathysarisky:
Really important context missing from your PR main message and buried in a comment (https://github.com/TryGhost/Ghost/pull/21904/commits/95e98fc616a5265f553de7b69e36b057c79cd5ea) :  You're changing the locale code from no to nb.  

**The proposed change is going to break translations for all sites that currently use the 'no' locale.**  

Additionally, if the change is going to be made, the supported_locales variable of i18n.js must be updated also.  (Otherwise it won't work for EITHER locale code.)

It sounds like a switch to 'nb' is the right thing to do, but it needs to be done with more care than the current PR.  sym-linking might be appropriate, but I'm not sure if the current build step for translation files can actually handle it.

cathysarisky (changes requested):
Really important context missing from your PR main message and buried in a comment (https://github.com/TryGhost/Ghost/commit/95e98fc616a5265f553de7b69e36b057c79cd5ea) : You're changing the locale code from no to nb.

The proposed change is going to break translations for all sites that currently use the 'no' locale.

Additionally, if the change is going to be made, the supported_locales variable of i18n.js must be updated also. (Otherwise it won't work for EITHER locale code.)

It sounds like a switch to 'nb' is the right thing to do, but it needs to be done with more care than the current PR. sym-linking might be appropriate, but I'm not sure if the current build step for translation files can actually handle it.

cmraible:
@cathysarisky nice catch! I haven't tried this so we'd need to test, but it looks like i18next supports a [fallbackLng parameter](https://www.i18next.com/principles/fallback#language-fallback) which we might be able to use here. If I'm reading the docs right, basically we'd set `nb` as the fallback for `no`, then remove `no`.

cathysarisky:
Hey @xqus , would you like to review these proposed translations?

ghost/i18n/locales/nb/comments.json
```diff
@@ -29,24 +29,24 @@
...
     "Just now": "Akkurat nå",
-    "Load more ({{amount}})": "",
+    "Load more ({{amount}})": "Last flere ({{amount}})",
     "Local resident": "Lokal innbygger",
     "Member discussion": "Medlemsdiskusjon",
     "Name": "Navn",
-    "Neurosurgeon": "Neurokirurg",
-    "Once deleted, this comment can’t be recovered.": "",
+    "Neurosurgeon": "Nevrokirurg",
+    "Once deleted, this comment can’t be recovered.": "Når slettet, kan ikke kommentaren gjenopprettes.",
```
xqus, line 43: Maybe change to:  `Når denne kommentaren er slettet kan den ikke gjennopprettes.`.
This is a more natural way to say it in Norwegian.

xqus (commented):
Translations looks good. 👍 
One minor comment, but the original suggestion also works.

cathysarisky:
There are two things going on here.  One: Fixes to language.  Those will be mergeable fairly quickly.  Two: The change to locale.  That's going to require someone to work through making fallback strings work so that we don't break anything for existing users.  (Alternative: write a database migration that switches anyone set to 'no' to 'nb' ?)  I'm not going to get a chance to do that for a while -- overbooked with paying work.  @knutid, if adding that is within your skill set, please have at it!  Otherwise, you're waiting for someone to pick it up and figure it out.  So... it might be best to split the locale change out from the translation adjustments.  That'll get the translation adjustments in sooner, and decouples the issue of what the locale name is from the changes wanted in it.

knutid:
My javascript-foo is totally missing. So, I'll look into splitting, @cathysarisky.

cathysarisky:
Since this PR has a review already, we could go ahead and merge it if  the no->nb change were removed.  Or one of you could review #22194 , and we could merge that one.  As a non-Norwegian speaker, I will get out of the way and let the native speakers decide which way to proceed.

knutid:
Hi, @cathysarisky.

Close this PR, #21904, and move forward with PR #22194 and my comments, tnx!

9larsons:
Closing, thanks!

PR #21917, opened 2024-12-18
Fixed intermittent error when posting a note in `admin-x-activitypub`

mike182uk:
refs [AP-629](https://linear.app/ghost/issue/AP-629/notes-say-they-error-but-post-correctly)

Fixed intermittent error when posting a note in `admin-x-activitypub`

The error was intermittent due to it only occurring when a specific set of steps occurred, and the query cache being periodically cleared.

The error itself was due to incorrectly expecting the `outbox:${handle}` query to be an array when it was in fact an object.

This PR also resolves and issue where the reply count for new notes would display `NaN` (because the `replyCount` property was not present on newly created notes)

### DIFF
--- a/apps/admin-x-activitypub/src/components/modals/NewPostModal.tsx
+++ b/apps/admin-x-activitypub/src/components/modals/NewPostModal.tsx
@@ -17,23 +17,28 @@ const NewPostModal = NiceModal.create(() => {
 
     const handlePost = async () => {
         const trimmedContent = content.trim();
+
         if (!trimmedContent) {
             return;
         }
 
-        await noteMutation.mutate({content: trimmedContent}, {
+        noteMutation.mutate({content: trimmedContent}, {
             onSuccess() {
                 showToast({
                     message: 'Note posted',
                     type: 'success'
                 });
+
                 modal.remove();
             },
-            onError() {
+            onError(error) {
                 showToast({
                     message: 'An error occurred while posting your note.',
                     type: 'error'
                 });
+
+                // eslint-disable-next-line no-console
+                console.error(error);
             }
         });
     };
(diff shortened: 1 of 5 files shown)


apps/admin-x-activitypub/src/components/modals/NewPostModal.tsx
```diff
@@ -17,23 +17,28 @@ const NewPostModal = NiceModal.create(() => {
 
     const handlePost = async () => {
         const trimmedContent = content.trim();
+
         if (!trimmedContent) {
             return;
         }
 
-        await noteMutation.mutate({content: trimmedContent}, {
+        noteMutation.mutate({content: trimmedContent}, {
```
mike182uk, line 25: No need to await here, success / error criteria is handled by mutation callbacks

apps/admin-x-activitypub/src/hooks/useActivityPubQueries.ts
```diff
@@ -488,7 +488,8 @@ export function useReplyMutationForUser(handle: string) {
         async mutationFn({id, content}: {id: string, content: string}) {
             const siteUrl = await getSiteUrl();
             const api = createActivityPubAPI(handle, siteUrl);
-            return await api.reply(id, content) as Activity;
+
+            return api.reply(id, content);
```
mike182uk, line 492: No need to await this

apps/admin-x-activitypub/src/hooks/useActivityPubQueries.ts
```diff
@@ -500,15 +501,27 @@ export function useNoteMutationForUser(handle: string) {
         async mutationFn({content}: {content: string}) {
             const siteUrl = await getSiteUrl();
             const api = createActivityPubAPI(handle, siteUrl);
-            return await api.note(content) as Activity;
+
+            return api.note(content);
```
mike182uk, line 505: No need to await this

PR #21922, opened 2024-12-19
Removed i18n toggle from labs UI

ErisDS:
ref https://github.com/TryGhost/Ghost/commit/bb9a69edfede00cb30cfcbc9213489ca48101cf6

- We promoted i18n to GA several weeks ago now, and it's going fine
- Removing the UI first to reduce confusion
- This is before clearning up all the other references to the flag


ErisDS:
Doh, can't do browser tests on a fork

PR #21936, opened 2024-12-19
Do not inject portal if portal.url is set to 'false' in config file.

ltoinel:
Original PR comment: 
If portal URL is set to false in the config as described in the Ghost documentation, we have to return nothing.
https://ghost.org/docs/config/

"The script can be relocated by changing the URL, or disabled entirely by setting "url": false."

Cathy's addition, upon 'adopting' this PR:
Tests added, confirming the behavior is as expected, including if the url is set to string false, boolean false, null, etc.

While I'd generally encourage self hosters who don't want the portal code to load to just turn off memberships, this leaves a route for a site to use the api functionality behind members, donations, etc, without including the portal loading code.

Likely to be a common want?  No.  Minimal impact on code maintainability?  I think so.


cathysarisky (commented):
This change makes a lot of sense to me.  If the scriptUrl isn't set or is set to false, there's no reason to continue with the function.  

It also brings the behavior in line with what the docs have said for a long time. :)

cathysarisky (approved):

cathysarisky:
Hey @ltoinel , I'm not sure if you're still interested in pursuing this change or not, but I was able to snag a moment of dev attention for it.  To be mergeable, it's going to need one or more tests added, showing that it results in the desired behavior.  You might take a look at the tests for the ghost_head excludes behavior here as a starting point: https://github.com/TryGhost/Ghost/blob/2345623bb7a639b832175d1132e73bf8014557f2/ghost/core/test/unit/frontend/helpers/ghost_head.test.js#L1583

cathysarisky:
Hey @ltoinel !  Since you didn't reply, I've pushed some tests that would cover the change in your PR, in hopes of getting this functionality ready to merge!

PR #21945, opened 2024-12-23
📦 externalized portal locales

vikaspotluri123:
no issue

All locales were previously included in the main bundle, even though at most 1 locale is used.
By separating out the locales, we can reduce the bundle size by ~60%:

|                     | raw             | gzip            |
| ------------------- | --------------- | --------------- |
| Before              | 1,542.72 kB     | 365.86 kB       |
| After               | 546.38 kB       | 156.29 kB       |
| Savings             | 996.34 kB (65%) | 209.57 kB (57%) |
| After (+1 locale)   | 564.48 kB       | 162.54 kB       |
| Savings (+1 locale) | 978.24 kB (63%) | 203.32 kB (56%) |

 - Added an async i18n factory that fetches locales on-demand
   - Implemented in a separate browser entrypoint for tree shaking
 - Created an i18n singleton/cache to avoid downloading the same locale multiple times
 - Allowed the locale fetch root to be overridden via `data-locale-root`
 - Added `vite-plugin-ghost-i18n` (plugin.js) in `@tryghost/i18n` to:
   - Emit each portal locale as a build asset
   - Minify the locales (using `JSON.stringify`)
   - Prevent rewriting `import.meta.url`Got some code for us? Awesome 🎊!

## Backwards Compatibility

The new plugin (`vite-plugin-ghost-i18n`) supports incremental adoption, based on `stability`. Since this hasn't been tested at-scale, Portal's stability is currently `opt-in`.

| Stability Mode | Main Bundle Type           | Secondary Bundle Name    | Secondary Bundle Type |
| -------------- | -------------------------- | ------------------------ | --------------------- |
| opt-in         | No change (inline locales) | `{ns}-no-locales.min.js` | Async locales         |
| opt-out        | Async locales              | `{ns}-locales.min.js`    | Inline locales        |
| solo           | Async locales              | N/A                      | N/A                   |

### DIFF
--- a/apps/portal/src/App.js
+++ b/apps/portal/src/App.js
@@ -1,3 +1,4 @@
+// @ts-check
 import React from 'react';
 import * as Sentry from '@sentry/react';
 import TriggerButton from './components/TriggerButton';
@@ -15,7 +16,7 @@ import './App.css';
 import {hasRecommendations, allowCompMemberUpgrade, createPopupNotification, hasAvailablePrices, getCurrencySymbol, getFirstpromoterId, getPriceIdFromPageQuery, getProductCadenceFromPrice, getProductFromId, getQueryPrice, getSiteDomain, isActiveOffer, isComplimentaryMember, isInviteOnly, isPaidMember, isRecentMember, isSentryEventAllowed, removePortalLinkFromUrl} from './utils/helpers';
 import {handleDataAttributes} from './data-attributes';
 
-import i18nLib from '@tryghost/i18n';
+import {createAsyncInstance} from '@tryghost/i18n/browser.mjs';
 
 const DEV_MODE_DATA = {
     showPopup: true,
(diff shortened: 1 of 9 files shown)


apps/portal/src/App.js
```diff
@@ -191,10 +192,18 @@ export default class App extends React.Component {
...
             // Fetch data from API, links, preview, dev sources
             const {site, member, page, showPopup, popupNotification, lastPage, pageQuery, pageData} = await this.fetchData();
-            const i18nLanguage = this.props.siteI18nEnabled ? this.props.locale || site.locale || 'en' : 'en';
-            const i18n = i18nLib(i18nLanguage, 'portal');
+
+            const localeWithSite = this.props.siteI18nEnabled && !this.props.locale && site.locale;
+
+            if (localeWithSite && baseLocale !== localeWithSite) {
+                i18n.changeLanguage(localeWithSite);
+            }
```
cathysarisky, line 206: If I've understood correctly, this will cause the site locale setting to override what might be passed in as a prop.  Is there a compelling reason for doing so?  I suggest that if a locale is passed in as a prop, it should dominate over the site-wide setting, which allows for loading portal in a language other than the sitewide locale for some fraction of the site.  (While that's not full multi-language support, it's a good start.)

vikaspotluri123: `localeWithSite` takes the data attribute into consideration🙂 

```javascript
const localeWithSite = this.props.siteI18nEnabled && !this.props.locale && site.locale;
//                                                   ^ Passing a locale will make result `false`
```

cathysarisky: Ah! Sorry, I missed that! :)

ErisDS:
I'm going through and having a big purge of old PRs to try to get the list down to active WIP.  I'd love to get this change in though. Will you reopen it when you have a sec to get it done?

PR #22006, opened 2025-01-15
Fixed initializing ActivityPub webhooks when enabled after boot

daniellockyer:
ref https://linear.app/ghost/issue/AP-644/fix-enabling-activitypub-service-when-flag-is-enabled

- right now, we only set up the ActivityPub webhooks if the service is initialized at boot
- so if the flag is enabled after boot, the webhooks aren't setup
- this changes that to allow initializing the service either at boot, or when the labs setting is changed (and ActivityPub flag is enabled)

## Summary by CodeRabbit

- **New Features**
	- Enhanced ActivityPub service initialization with improved state management
	- Added reactive initialization based on lab settings changes

- **Refactor**
	- Simplified initialization logic for ActivityPub service
	- Improved dependency and settings handling


daniellockyer:
@coderabbitai

PR #22036, opened 2025-01-21
Fixed reading progress indicator for very short articles

djordjevlais:
ref https://linear.app/ghost/issue/AP-653/scroll-percentage-remains-at-0percent-when-no-content-to-scroll

- When an entire article fits into the viewport height, we used to show`0%` in the reading progress indicator. Now we check if that's the case, and then show `100%` if it is.

### DIFF
--- a/apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
+++ b/apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
@@ -48,6 +48,7 @@ const ArticleBody: React.FC<{
     fontFamily: SelectOption;
     onHeadingsExtracted?: (headings: TOCItem[]) => void;
     onIframeLoad?: (iframe: HTMLIFrameElement) => void;
+    onLoadingChange?: (isLoading: boolean) => void;
 }> = ({
     heading,
     image,
@@ -57,7 +58,8 @@ const ArticleBody: React.FC<{
     lineHeight,
     fontFamily,
     onHeadingsExtracted,
-    onIframeLoad
+    onIframeLoad,
+    onLoadingChange
 }) => {
     const site = useBrowseSite();
     const siteData = site.data?.site;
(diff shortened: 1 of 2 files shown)


apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
```diff
@@ -527,17 +533,29 @@ const ArticleModal: React.FC<ArticleModalProps> = ({
...
         const handleScroll = () => {
-            if (!container || !article) {
+            if (!container || !article || isLoading) {
                 return;
             }
 
             const articleRect = article.getBoundingClientRect();
             const containerRect = container.getBoundingClientRect();
+
+            // Calculate if content is shorter than viewport
```
mike182uk, line 550: I don't think you need to comment this, the code is self explanatory 👍

apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
```diff
@@ -527,17 +533,29 @@ const ArticleModal: React.FC<ArticleModalProps> = ({
...
+            const isContentShorterThanViewport = articleRect.height <= containerRect.height;
+
+            if (isContentShorterThanViewport) {
+                setReadingProgress(100);
+                return;
+            }
+
+            // Existing progress calculation for longer content
```
mike182uk, line 558: I don't think you need to comment this, the code is self explanatory 👍

apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
```diff
@@ -547,9 +565,26 @@ const ArticleModal: React.FC<ArticleModalProps> = ({
             setReadingProgress(progress);
         };
 
-        container?.addEventListener('scroll', handleScroll);
-        return () => container?.removeEventListener('scroll', handleScroll);
-    }, []);
+        // Only set up observers and listeners once content is loaded
+        if (!isLoading) {
```
mike182uk, line 569: Could maybe return early here instead so we don't have another level of nesting:

```
if (isLoading) {
  return;
}

// ...
```

mike182uk (approved):

PR #22055, opened 2025-01-27
🔒 Made names for uploaded files more secure and resilient to filesystems' limits

sagzy:
ref https://linear.app/ghost/issue/ENG-1260
ref https://linear.app/ghost/issue/ENG-1859

- names for uploaded files now include a 16-character alphanumeric hash, that gets re-generated on every upload. This prevents original files to be found, e.g. when redacting sensitive content in an image in the editor
- if the filename length is higher than what most filesystem accept (255 bytes), we truncate the base of the filename, while keeping space for the unique hash and the file extension. Example to illustrate (non-accurate):
    - input: `long-name-for-an-image-imagine-255bytes-here-blablablablabla-and-a-bit-more.png`
    - output: `long-name-for-an-image-1a2b3c4d5e6f7g890.png`

### DIFF
--- a/ghost/core/core/server/adapters/storage/LocalStorageBase.js
+++ b/ghost/core/core/server/adapters/storage/LocalStorageBase.js
@@ -48,18 +48,15 @@ class LocalStorageBase extends StorageBase {
      * @returns {Promise<String>}
      */
     async save(file, targetDir) {
-        let targetFilename;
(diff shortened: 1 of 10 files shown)


Ghost-Slimer:
Deployed to **staging** with ID: `3218`

PR #22068, opened 2025-01-29

cmraible:

- We want to be able to disable the `email_track_clicks` setting forcibly via config, such that it is turned off and the user is unable to change it in the UI.
- This commit passes a `settingsOverrides` object into the settings cache, which is passed in from the `hostSettings:settingsOverride` configuration key, and overrides both `settingsCache.get()` and `settingsCache.getAll()` to override the value of the setting that's returned with whatever is defined via configuration.

### DIFF
--- a/apps/admin-x-framework/src/api/settings.ts
+++ b/apps/admin-x-framework/src/api/settings.ts
@@ -8,6 +8,7 @@ export type SettingValue = string | boolean | null;
 export type Setting = {
     key: string;
     value: SettingValue;
+    override?: boolean;
 }
 
 export type SettingsResponseMeta = Meta & {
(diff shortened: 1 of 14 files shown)


cmraible:
High level approach here is to override the `settingsCache.get()` and `settingsCache.getAll()` methods to use the value from config, rather than from the database. Rather than building in a specific exception `email_track_clicks`, this is a generalizable pattern that can be used to override any setting.

Only the `value` of the setting can be overridden, and if it _is_ overridden, it will add an `override: true` key to the returned setting. This will be used on the frontend to disable the control from being modified.

cmraible:
Closing in favor of https://github.com/TryGhost/Ghost/pull/22088

PR #22084, opened 2025-01-30
Removed lodash from data-manipulation and sanitize model plugins

daniellockyer:
- only some small micro-optimizations but we can remove lodash from a sanitize plugin method (needs an empty object fallback because Object.keys is less forgiving than _.keys and there are some tests which don't fully initialize the model)
- also converts the data-manipulation plugins into for-loops because there are more native than using lodash, and it means we can remove lodash from this file too
- lodash is a bit of a perf burn on extreme hotpaths, so it's good to clean it up where it's not needed
- adds a bit of JSDoc to help with editor autocomplete


daniellockyer:
@coderabbitai review

PR #22088, opened 2025-01-30

cmraible:

- We want to be able to disable the email_track_clicks setting forcibly via config, such that it is turned off and the user is unable to change it in the UI.
- This commit passes a settingsOverrides object into the settings cache, which is passed in from the hostSettings:settingsOverride configuration key, and overrides both settingsCache.get() and settingsCache.getAll() to override the value of the setting that's returned with whatever is defined via configuration.


Ghost-Slimer:
Deployed to **staging** with ID: `3220`

cmraible:
Tested and working on staging  — going to break this PR into one for backend and one for frontend now.

PR #22089, opened 2025-01-31

cmraible:

- We want to have the ability to turn off click tracking for sites that are adversely impacted by massive bursts of traffic from email link checkers, but we don't currently have a pattern for this.
- This commit introduces a new configuration parameters `hostSettings:settingsOverrides`, which accepts key/value pairs of settings keys -> values. The value passed in here will override whatever value is set for the associated setting key in the database
- It also adds an `is_read_only: true` property to any setting that is overridden, which is included in the /api/admin/settings endpoint. This value can be used by the frontend to disable the control to prevent a user from trying to change the value.
- The value in the database is preserved, as the override is implemented in the settings cache `get()` and `getAll()` methods. 
- This commit only includes the backend changes — another commit will follow to allow disabling the 'Newsletter clicks' toggle in Admin's settings.

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/settings.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/settings.js
@@ -3,7 +3,7 @@ const _ = require('lodash');
 module.exports = (attrs) => {
     if (Array.isArray(attrs)) {
         return attrs.map((setting) => {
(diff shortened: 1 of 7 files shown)


cmraible:
This PR was originally drafted as https://github.com/TryGhost/Ghost/pull/22088, and includes only the backend changes. The frontend changes to admin-x-settings will be committed separately.

PR #22094, opened 2025-01-31
prevent accented urls from showing up in site metadata.

### LINKED ISSUE #21999: {{ghost_head}} and accents
### Issue Summary

My client has moved from Ghost fairly recently.  From his old import, he has some links like
/tag/sécurité incendie/page/5/

Note the accent (and also spaces).  Ghost actually deals with this, serving up the tags page for securite-indendie instead.  However, when it writes out the metadata `<link rel="prev" href` (and also "next"), it uses the unadjusted form of the tag.

The behavior may be exacerbating search engine crawlers trying to visit the wrong form of the URL.

### Steps to Reproduce

Add some random accents and spaces to an existing tag.  For example:
https://www.spectralwebservices.com/topic/advanc%C3%A9d%20topics/ (should be advanced-topics)

### Ghost Version

5.106.1

### Node.js Version

Ghost pro, also 20.

Ghost Pro, or ghost-cli install (Ubuntu)

### Database type

MySQL 8

### Browser & OS version

Chrome, Windows 10

_No response_

### Code of Conduct

cathysarisky:
Closes #21999 

If for whatever reason a site has incoming links (or internal links) that include accents in the urls, such as /tag/nèws, Ghost currently provides canonical and pagination links that preserve the incorrect formatting, leading to duplicate canonical links and an SEO mess.

(Interestingly, data retrieval works just fine, because the slugs are later parsed with security.safe.string.)

This PR runs urls through the same process, so that the generated urls in the metadata will be correct, even if the incoming urls are wèird.

Update:  I also fixed the canonical-url tests.  They were... weird, because the stubs used characters that aren't actually valid URLs?  A close look at what I did would be appreciated, because changing failing tests to make them pass should be a little suspect, right? 👀


cathysarisky:
note to myself: should redirect these requests in the url handler.

ErisDS:
Cleaning up older PRs 💅 I also think we agreed on a different approach for this?

PR #22102, opened 2025-02-03
🐛 Sodo Search: Made posts with hashtags (#) and underscores (_) searchable

dkalpakchi:
no issue
Hash symbols (#) or underscores (_) were not searchable via Sodo Search, this PR fixes that.


dkalpakchi:
Hey @cathysarisky! Thanks for bringing this to my attention! This is really weird, because I ran 'yarn lint' locally and it didn't produce any warnings or errors, so I didn't even look at what happened in the PR after that. I guess I should have :)

cathysarisky (approved):

cathysarisky:
OK, so I had time to build and play with this.  I'm of mixed mind on what the 'right' behavior is.
Current behavior:  Searching for #fun  or fun yields posts with the word #fun or fun in the title.
New behavior:  Searching for fun yields only the post containing fun (not #fun).  Searching for #fun yields only the post containing #fun (not fun).

I'm not sure this is actually an improvement, in terms of user experience.   As a user, if I searched for a #hashtag, I'd probably be OK with getting results without it, and if I searched for  a word (without #), I'd definitely be interested in things with the #word also...

cathysarisky (changes requested):
Please see recent comment.  There's nothing wrong with the PR exactly, but I'm not sure that the behavior being created is actually an improvement.  This could use some more discussion.

dkalpakchi:
Hi @cathysarisky! Sorry for taking so long to respond. Thank you for your reflection! I somewhat agree with your assessment, it's probably too restrictive to not return "fun" when searching for "#fun". At the same time if I were searching for "#fun" I would definitely want it top of the list, which is in the current implementation not a guarantee, from what I can gather. Not sure either what the best approach is here. I guess given that it was not a reported bug by anyone, and given that there is no clear consensus that this is an improvement, we can just close this PR without merging?

cathysarisky:
I don't know off the top of my head if there's a way to get search to try for #fun before falling back on fun, but I /do/ know that it's based on flexsearch, so a look at that package's documentation may be informative...

dkalpakchi:
OK, I can have a look at flexsearch and see if I could find how to do that using that library. If not, then I'll just close this PR.

On another point, I see that merging is blocked, because I have a merge commit in my PR, does it mean that this PR isn't mergeable anyway? :)

cathysarisky:
Github changed how things display recently.  I think you can switch to the 'classic' merge experience and then hit the 'merge' button.  At least, that worked for me the other day.  I'm seeing that there are no conflicts (in the 'classic' view), so I don't think there's anything wrong here.

ErisDS:
The latest flexsearch looks much more customisable. I think there's probably a way to make this kind of thing more configurable, or at least make it easy to modify and build a custom version of Ghost's search if you want minor modifications like this.

For now, I don't think this version of the fix is mergeable, not last because it's built on top of the code causing #22764 that definitely needs to go away.

PR #22112, opened 2025-02-05
Removed autosaving of tags from gh-post-settings-menu

allouis:
ref https://linear.app/ghost/issue/ONC-756
ref d91869e6406b7b1f0790cb34c92d2b8c992f3465

Rather than forcing a save on every tag change, we want to defer to the autosave functionality which will debounce saves, reducing the number of API calls when modifying tags on a post. This does mean that changing tags for existing posts will require an explicit save, as the autosave functionality does not run for published posts.

### DIFF
--- a/ghost/admin/app/components/gh-post-settings-menu.js
+++ b/ghost/admin/app/components/gh-post-settings-menu.js
@@ -625,7 +625,6 @@ export default class GhPostSettingsMenu extends Component {
     saveTags() {
         this.saveTagsTask.perform().catch((error) => {
             this.showError(error);
-            this.post.rollbackAttributes();
         });
     }
 
--- a/ghost/admin/app/components/gh-post-settings-menu.js
+++ b/ghost/admin/app/components/gh-post-settings-menu.js
@@ -622,6 +622,14 @@ export default class GhPostSettingsMenu extends Component {
     }
 
     @action
+    saveTags() {
+        this.saveTagsTask.perform().catch((error) => {
+            this.showError(error);
+            this.post.rollbackAttributes();
+        });
+    }
+
+    @action
     savePost() {
         this.savePostTask.perform().catch((error) => {
             this.showError(error);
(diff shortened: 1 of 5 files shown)


allouis:
I'm interested to know how to test this better, I went to add tests for the post-settings-menu component, but they are unit tests and completely commented out, which a note to migrate them to integration tests - shall I add an integration test for this?

I'm pretty confident I've use the right one, but I'm a little bit curious about the difference between the `_autosaveTask` and the `autosaveTask` in the lexical-editor, the `autosaveTask` seems to do an immediate save - and doesn't have any of the debouncing OR the checks for `this._canAutosave` built in.

ghost/admin/app/components/gh-psm-tags-input.js
```diff
@@ -13,6 +13,8 @@ export default class GhPsmTagsInput extends Component {
     // public attrs
     post = null;
 
+    currentTags = null;
```
kevinansfield, line 16: What the reasoning behind adding this local state/indirection? Generally it's always preferable to work directly with model state rather than duplicating it across internal component state.

kevinansfield:
> I'm a little bit curious about the difference between the _autosaveTask and the autosaveTask in the lexical-editor, the autosaveTask seems to do an immediate save - and doesn't have any of the debouncing OR the checks for this._canAutosave built in.

`_autosaveTask` is an internal task used within the editor controller, called when model changes are automatically detected or by other methods that should trigger the same "content changed" behaviour, it's the task that handles the 3sec debounced background save.

`autosaveTask` is a public task intended to be called when we want to save something as a background save rather than an explicit user-initiated save. Typically any calls to that are wrapped in a check for the post being a draft so we're not auto-updating published posts unexpectedly.

Background saves are "quiet", in that they won't trigger post-saved popup notifications or error alerts, they also don't allow the post status to be changed as background saves should never publish a post.

I expect it would be good to rename `autosaveTask` to `backgroundSaveTask` and put the is-draft check inside it.

ghost/admin/app/components/gh-psm-tags-input.js
allouis: Thanks, addressed!

ghost/admin/app/components/gh-post-settings-menu.js
```diff
@@ -621,6 +621,14 @@ export default class GhPostSettingsMenu extends Component {
         });
     }
 
+    @action
+    saveTags() {
+        this.saveTagsTask.perform().catch((error) => {
+            this.showError(error);
+            this.post.rollbackAttributes();
```
kevinansfield, line 628: I wonder if we should skip that call - `tags` is a relationship rather than an attribute and doesn't get rolled back with `rollbackAttributes()` 🤔

allouis: Ah okay, yeah I'll remove that

allouis, line 628:
```suggestion
```

allouis: Done!

PR #22126, opened 2025-02-06
Removed stdout from logging

rmgpinto:
ref https://linear.app/ghost/issue/PRO-1359

- Load tests showed a performance improvement of 35% by turning off stdout logging. The usage of PrettyStream causes write performance on high traffic conditions to degrade.

### DIFF
--- a/ghost/core/core/shared/config/env/config.production.json
+++ b/ghost/core/core/shared/config/env/config.production.json
@@ -16,6 +16,6 @@
         "rotation": {
             "enabled": true
         },
-        "transports": ["file", "stdout"]
+        "transports": ["file"]
     }
 }


daniellockyer (approved):
LGTM.

Can you prefix the commit with a relevant [gitmoji](https://gitmoji.dev/) before merging, so it shows up in release notes? Maybe 🔧? (be sure to use an actual emoji and not the shortcode)

ErisDS:
Can I go one step further? Here on the Ghost repo, the main line of the commit message is the release notes, so can we make it a little clearer like:
🔧 Removed stdout from default production logging options

It's always the little changes that get the most feedback 🤣

markstos:
Did you evaluate the inverse: Disabling file-based logging, but leaving only STDOUT?

For self-hosting, I recommend `stdout` logging, which goes into the system journal, which then handles the disk space management-- no log file rotation required.

daniellockyer:
@markstos Unfortunately, `stdout` logging is very slow and we're seeing it being a real performance bottleneck during high traffic spikes. File-based logging is a _lot_ faster (35% requests/second boost).

Anyone self-hosting is free to override this default 🙂

daniellockyer:
For stdout, we apply some styles to the output (JSON formatting, date/time etc) which have a big performance cost (I'm working on improving the performance of that) but also, stdout logging is synchronous whilst writing to file descriptor, but file-based logging is buffered. 

Either way, we've found file-based to be 10x faster than stdout logging

PR #22128, opened 2025-02-06
Refactored `extra-attrs` output serializer util

daniellockyer:
- after discussing on a code review, we agreed the existing code was very verbose and it was hard to reason about
- this is because of the incredibly long if-statements we have lingering around
- the general approach here was to extract all the includes into variables at the top, replace them in the code and then simplify the logic
- whilst doing this, I wrote some more tests to cover the other cases
- this commit also switches to assert, as it's neater to write the assertions in tests

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/serializers/output/utils/extra-attrs.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/output/utils/extra-attrs.js
@@ -8,31 +8,32 @@ const readingMinutes = require('@tryghost/helpers').utils.readingMinutes;
  * @returns {void} - modifies attrs
  */
 module.exports.forPost = (options, model, attrs) => {
-    // This function is split up in 3 conditions for 3 different purposes:
+    const columnsIncludesCustomExcerpt = options.columns?.includes('custom_excerpt');
+    const columnsIncludesExcerpt = options.columns?.includes('excerpt');
+    const columnsIncludesPlaintext = options.columns?.includes('plaintext');
+    const columnsIncludesReadingTime = options.columns?.includes('reading_time');
+    const formatsIncludesPlaintext = options.formats?.includes('plaintext');
+
     // 1. Gets excerpt from post's plaintext. If custom_excerpt exists, it overrides the excerpt but the key remains excerpt.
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/api/endpoints/utils/serializers/output/utils/extra-attrs.js
```diff
@@ -8,31 +8,35 @@ const readingMinutes = require('@tryghost/helpers').utils.readingMinutes;
...
-                let plaintext = model.get('plaintext');
-                if (plaintext) {
-                    attrs.excerpt = plaintext.substring(0, 500);
-                } else {
-                    attrs.excerpt = null;
-                }
-                if (!options.columns.includes('custom_excerpt')) {
-                    delete attrs.custom_excerpt;
-                }
+    if (columnsIncludesExcerpt || (columnsIncludesExcerpt && formatsIncludesPlaintext)) {
```
kevinansfield, line 18: Can this just be `if (columnsIncludesExcerpt)`? The second conditional would never be reached because `columnsIncludesExcerpt` will already have returned `true`.

ghost/core/core/server/api/endpoints/utils/serializers/output/utils/extra-attrs.js
```diff
@@ -8,31 +8,35 @@ const readingMinutes = require('@tryghost/helpers').utils.readingMinutes;
...
-                if (!options.columns?.includes('custom_excerpt')) {
-                    delete attrs.custom_excerpt;
-                }
+                attrs.excerpt = null;
+            }
+        } else {
+            attrs.excerpt = attrs.custom_excerpt;
+            if (!columnsIncludesCustomExcerpt) {
+                delete attrs.custom_excerpt;
             }
```
kevinansfield, line 30: Looks like this is an unnecessary duplicate as it's already run for either condition later 🙂

kevinansfield: Lol, the AI rabbit got there too. I'll be quiet for now 🤐

daniellockyer: My new best friend @coderabbitai also made this suggestion 😀

daniellockyer:
@coderabbitai Look at this PR again please

PR #22129, opened 2025-02-06
Added nice transition when stats counter changes

djordjevlais:
ref https://linear.app/ghost/issue/AP-706/add-repost-count-so-that-the-ui-can-display-the-count

- When you repost (or derepost) a post, the stats counter will now nicely animate into a larger (or smaller) number, with only the relevant digits animating and the others staying in place. We’re using a reusable hook for this.
- Updated stat counter numbers to also change the color when the post is liked or reposted by the currently logged in user

### DIFF
--- a/apps/admin-x-activitypub/src/components/feed/FeedItemStats.tsx
+++ b/apps/admin-x-activitypub/src/components/feed/FeedItemStats.tsx
@@ -1,4 +1,4 @@
-import React, {useState} from 'react';
+import React, {useEffect, useState} from 'react';
 import {Button} from '@tryghost/admin-x-design-system';
 import {ObjectProperties} from '@tryghost/admin-x-framework/api/activitypub';
 import {useAnimatedCounter} from '../../hooks/useAnimatedCounter';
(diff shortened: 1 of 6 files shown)


apps/admin-x-activitypub/src/hooks/useAnimatedCounter.tsx
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }, [value, digits]);
+
+    const updateValue = (delta: number) => {
+        setValue(prev => prev + delta);
+        setIsDecrementing(delta < 0);
+    };
+
+    return {
+        displayValue: (
```
mike182uk, line 45: I wonder if just `Counter` would be a better name for this? (Uppercased to indicate its a `component`)

apps/admin-x-activitypub/src/components/feed/FeedItemStats.tsx
```diff
@@ -28,6 +29,12 @@ const FeedItemStats: React.FC<FeedItemStatsProps> = ({
     const unlikeMutation = useUnlikeMutationForUser('index');
     const repostMutation = useRepostMutationForUser('index');
     const derepostMutation = useDerepostMutationForUser('index');
+    const {
+        displayValue: repostCountDisplay,
+        currentValue: repostCountNumber,
```
mike182uk, line 34: How about aliasing `repostCount` thats passed in to `initialRepostCount`, then aliasing `currentValue` here to `repostCount`?

apps/admin-x-activitypub/src/hooks/useAnimatedCounter.tsx
peterzimon: +1 on this

apps/admin-x-activitypub/src/components/feed/FeedItemStats.tsx
mike182uk: Updated in https://github.com/TryGhost/Ghost/pull/22129/commits/12b9675cf6eab4609e5dc550959f42c809aabeff

PR #22153, opened 2025-02-10
🌵 SPIKE: Replace dev.js script with declarative NX target

cmraible:
WIP

- This is an experiment that uses `nx` to run all the various commands that are run in the `.github/scripts/dev.js` script
- Theoretically this method could accomplish the same thing as the `.github/scripts/dev.js` script but more efficiently (since it will take full advantage of the `nx` cache and graph) and it's (arguably) a bit "cleaner"
- The original motivation for this was to lean more heavily on `nx` for file watching because it seems to be more compatible with docker compose's `--watch` flag than the native `node --watch` or `tsc --watch` are


cmraible:
No intention to merge this, closing but keeping the branch around for posterity

PR #22164, opened 2025-02-11
 Refactored CI workflows for publishing packages

daniellockyer:
- majority of the changes here are just DRYing up the implementation so we don't have the same code split across many different jobs


daniellockyer:
@coderabbitai review please

PR #22168, opened 2025-02-11
WIP: Refactor to move gating and extra-attrs logic to PostsService

kevinansfield:
ref https://linear.app/ghost/issue/PLG-352

- updated `/p/` endpoint to accept an optional `?memberStatus` query parameter with three allowed values: `anonymous/free/paid`
  - using a query parameter allows for post previews to be opened in a new tab and still display segmented views
- when a `memberStatus` query param is provided:
  - `frame.apiType` is set to `content` which triggers the content-gating paths of our output serializer's post mapper
  - where needed, `frame.original.context.member` is set to a minimal member object with a `status` property to allow content-gating NQL queries to match
  - the combination of both switches our preview API output to match what the Content API does when rendering posts with a particular logged-in member status
- if no `memberStatus` query param is provided the existing behaviour is preserved, where we render as we do for Admin API requests, showing content for all segments simultaneously


ghost/core/core/server/api/endpoints/previews.js
```diff
@@ -2,11 +2,39 @@ const tpl = require('@tryghost/tpl');
...
     postNotFound: 'Post not found.'
 };
 
+// Simulate serving content as different member states by setting the minimal
+// member context needed for content gating to function
+const _addMemberContextToFrame = (frame) => {
+    if (!frame?.options?.memberStatus) {
+        return;
+    }
+
```
ronaldlangeveld, line 17: Would it make things safer if we add member validation here, eg checking that `frame.options.memberStatus` has one of `ALLOWED_MEMBER_STATUSES` and perhaps throw an error or default? In case

kevinansfield: Yeah, we could potentially add some extra safety here. Although in theory it should never be possible as the [API filters the memberStatus option](https://github.com/TryGhost/Ghost/pull/22168/files#diff-f5713173a4138c9fb6f492e66ff27cad61e8250ecce7f13e7b7e33f39cec3411R59-R60) before this code is reached.

ErisDS:
Cleaning up older PRs 💅

PR #22169, opened 2025-02-12
:art: Added classes to tags helper output

brandishwar:
no issue

- Far easier to customize the tags helper output when the parts have CSS classes
- Avoids the need to write custom code to customize how tags look - unless you really want to

### DIFF
--- a/ghost/core/core/frontend/helpers/tags.js
+++ b/ghost/core/core/frontend/helpers/tags.js
@@ -6,10 +6,18 @@
 //
 // Note that the standard {{#each tags}} implementation is unaffected by this helper
 const {urlService} = require('../services/proxy');
-const {SafeString, escapeExpression, templates} = require('../services/handlebars');
-
-const isString = require('lodash/isString');
+const {SafeString, escapeExpression} = require('../services/handlebars');
+const _ = require('lodash');
+const isString = _.isString;
 const ghostHelperUtils = require('@tryghost/helpers').utils;
+
+const tagLink = _.template('<a class="<%= tag_class %>" href="<%= url %>"><%= text %></a>');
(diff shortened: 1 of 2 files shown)


cathysarisky (changes requested):
Hey @brandishwar ! Welcome to the Ghost repo! :)  

I don't work for Ghost.org - I'm a community contributor like you.

I think that improving the output of {{tags}} with some useful classes is a great idea.  However, I think the execution needs work.  My big concern:  You're changing the current behavior, in a way that's absolutely going to break some themes.  In particular, currently the output of {{tags}} may safely be used as a plain string, in places where HTML tags are not appropriate.   

Here's what I suggest instead (but again, I'm not the core team):

- Retain the existing behavior as much as possible.  If your changes require changing the results of an existing test, carefully ask why, and what might break.
- New span wrapping should be controlled by an argument.  (wrapAll ?)  
- There should be tests for the new functionality (set to true, not set, set to false)

It's probably safe to add a new class to existing tags in the presence of autolink, but new html is definitely going to mess something up for someone.

brandishwar:
> Here's what I suggest instead (but again, I'm not the core team):
> 
> * Retain the existing behavior as much as possible.  If your changes require changing the results of an existing test, carefully ask why, and what might break.
> * New span wrapping should be controlled by an argument.  (wrapAll ?)
> * There should be tests for the new functionality (set to true, not set, set to false)

And one scenario where someone would want to keep the plaintext (with autolink Off) is... metadata beyond what sites are supported by default in Ghost. Probably other scenarios for it that I'm not considering.

Wrapping this behind a new parameter I think is the best way to go - Off by default, of course. So I'll withdraw this PR while I get that sorted.

cathysarisky:
Sounds like a plan!

PR #22177, opened 2025-02-13
Improved docker development setup

cmraible:
no issue

This commit is a fairly significant rework of the docker development setup, with the goal of improving the performance of the Ghost container and generally improving the DX of the docker development setup.

**Enumerating `node_modules` directories as volume exclusions for the bind mount**
- The main source of performance overhead with the docker setup is the bind mount, which mounts all the code into the container. Especially on macOS, the bind mount introduces quite a bit of file system overhead. Currently the bind mount includes all the `node_modules` directories, which adds up to many thousands of files in the bind mount. This change attempts to reduce this overhead by mounting in named volumes for each `node_modules` directory in the monorepo, which effectively tells the main bind mount to ignore these directories, thus reducing the filesystem overhead.

**Removing the `yarn install` from the entrypoint script and the `yarn setup:docker` step**
- This change has two more benefits: since all the `node_modules` are installed into named volumes, we can now be confident that our local `node_modules` won't conflict with the `node_modules` in the container. Therefore, we can install `node_modules` at build time in the Dockerfile instead of in the entrypoint script, and we can also get rid of the `yarn setup:docker` setup command, which nukes your local `node_modules` before building the container.

**Build optimization - enumerating all `package.json` files**
- In a typical node app's Dockerfile, you copy the `package.json` and `yarn.lock` files into the container first, then run `yarn install`, then copy the rest of the code in. This takes advantage of Docker's build cache to avoid having to reinstall `node_modules` on every build if you've only changed code files, but not any dependencies. 
- This is more challenging to do with the monorepo since we have many `package.json` files. This commit brute forces our way through that by copying every single `package.json` file in the Dockerfile before installing dependencies and copying the code in. The result is that Docker builds become much faster, after the initial full build.

**Additional scripts**
- Since enumerating all the `node_modules` directories and `package.json` files is tedious to do manually, this commit also includes two new scripts to automate this work. Ideally we'll eventually integrate these scripts with Slimer, so when you add a new package to the monorepo it automatically updates the `Dockerfile` and `compose.yml`, but for now we can manage that ourselves since it's a fairly rare occurrence.


.docker/Dockerfile
```diff
@@ -28,14 +29,136 @@ RUN curl -s https://packages.stripe.dev/api/security/keypair/stripe-cli-gpg/publ
...
-ENV YARN_CACHE_FOLDER=$WORKDIR/.yarncache
 
+# Disable NX rejection of unknown local cache
+ENV NX_REJECT_UNKNOWN_LOCAL_CACHE=false
+
+# Copy the package.json and yarn.lock files
+COPY package.json yarn.lock ./
+
+# Copy all package.json files
+COPY apps/admin-x-activitypub/package.json apps/admin-x-activitypub/package.json
```
cmraible, line 49: I _think_ we should be able to get all these into a single COPY statement so we don't have 100+ layers 😂

PR #22179, opened 2025-02-13
:art: Added option to tags helper to wrap output in HTML with CSS classes

brandishwar:
:art: Added option to tags helper to wrap output in HTML with CSS classes

no issue
- Far easier to customize the tags helper output when the parts have CSS classes
- Avoids the need to write custom code to customize how tags look - unless you really want to


ErisDS:
Sorry we didn't get around to reviewing this. Today I'm having a bit of an old-PR clean up to try to make the list more maintainable.

Generally when suggesting changes to the theme helpers it's super helpful for us trying to review them if you include full examples of the use case you're trying to solve in the PR description. 

If you're still keen to make this change, feel free to reopen the PR with more details and we'll review it :)

PR #22187, opened 2025-02-13
Enabled Job Queue by default

cmraible:
ref https://linear.app/ghost/issue/ENG-2001/enable-the-job-queue-in-defaultsjson-once-its-rolled-out-to-all-of-pro

WIP - raising this PR so we can merge when ready, but currently monitoring the impact of the job queue on the database.

### DIFF
--- a/ghost/core/core/shared/config/defaults.json
+++ b/ghost/core/core/shared/config/defaults.json
@@ -268,5 +268,19 @@
         "batchSize": 1000,
         "captureLinkClickBadMemberUuid": false
     },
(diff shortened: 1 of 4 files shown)


cmraible:
Test is failing in CI and in Docker, but not locally on macOS. Seems to be some kind of difference in sqlite3 behavior depending on the OS?

cmraible:
We've shifted directions here a bit, and instead of enabling the job queue 100% by default, we're likely going to remove it completely.

PR #22193, opened 2025-02-14
🐛 Fixed subscription attribution not getting set for upgrading members

9larsons:
ref https://linear.app/ghost/issue/ENG-1561/

When updating existing members' subscriptions (e.g. free member upgrading to paid), the attribution was not being set on the subscription. This should resolve that by adding a code path to the `checkout.session.completed` to update the already-linked subscription.

### DIFF
--- a/ghost/core/core/server/models/subscription-created-event.js
+++ b/ghost/core/core/server/models/subscription-created-event.js
@@ -3,6 +3,7 @@ const ghostBookshelf = require('./base');
 
 const SubscriptionCreatedEvent = ghostBookshelf.Model.extend({
     tableName: 'members_subscription_created_events',
+    hasTimestamps: false, // if true (default), requires an updated_at column and schema changes
 
     member() {
         return this.belongsTo('Member', 'member_id', 'id');
--- a/ghost/members-api/lib/repositories/MemberRepository.js
+++ b/ghost/members-api/lib/repositories/MemberRepository.js
@@ -3,7 +3,7 @@ const errors = require('@tryghost/errors');
 const logging = require('@tryghost/logging');
 const tpl = require('@tryghost/tpl');
 const DomainEvents = require('@tryghost/domain-events');
-const {SubscriptionActivatedEvent, MemberCreatedEvent, SubscriptionCreatedEvent, MemberSubscribeEvent, SubscriptionCancelledEvent, OfferRedemptionEvent} = require('@tryghost/member-events');
+const {SubscriptionActivatedEvent, MemberCreatedEvent, SubscriptionCreatedEvent, SubscriptionAttributionEvent, MemberSubscribeEvent, SubscriptionCancelledEvent, OfferRedemptionEvent} = require('@tryghost/member-events');
 const ObjectId = require('bson-objectid').default;
 const {NotFoundError} = require('@tryghost/errors');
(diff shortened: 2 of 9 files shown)


cmraible:
I haven't really looked at the code here yet, but it doesn't seem to have fixed the problem 100% of the time. 

First I reproduced the issue in the ticket on `main` locally — create a free member, sign in as free member, upgrade to a paid plan, then see attribution info is missing. I tried this multiple times and it worked correctly 1 out of 5 times, so there is some flakiness.

Then I checked this PR out locally and ran through the same steps. It worked the first time, but the second time I tried it did not work. Trying a few more times to see if I can figure out what specifically is different/flaky 🤔

9larsons:
This ref might be helpful for the model: https://www.notion.so/ghost/Member-Attribution-Technical-Specification-e246832e676443e9829ab99ae659be75

We use the `members_created_events` for member attribution and `members_subscription_created_events` for the subscription attribution.

ghost/members-api/lib/repositories/MemberRepository.js
```diff
@@ -885,6 +885,18 @@ module.exports = class MemberRepository {
         return subscription;
     }
 
+    /**
+     * @param {string} subscriptionId - The stripe subscription id
+     * @param {import('@tryghost/member-attribution/lib/Attribution').AttributionResource} attribution
+     * @returns {Promise<void>}
+     */
+    async updateSubscriptionAttribution(subscriptionId, attribution) {
+        const event = SubscriptionAttributionEvent.create({
```
cmraible, line 894: Running this locally, I'm seeing that `subscriptionId` here is sometimes (always?) `null`:
image

Wondering if this is some kind of race condition depending on the order the webhooks come in?

ghost/members-events-service/lib/EventStorage.js
```diff
@@ -55,6 +55,28 @@ class EventStorage {
...
             });
         });
+
+        domainEvents.subscribe(SubscriptionAttributionEvent, async (event) => {
+            let attribution = event.data.attribution;
+
+            const subscriptionCreatedEvent = await this.models.SubscriptionCreatedEvent
+                .findOne({subscription_id: event.data.subscriptionId}, {require: false, withRelated: []});
+
+            if (!subscriptionCreatedEvent) {
```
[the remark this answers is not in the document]
cmraible: Scratch this, in light of the `browse` vs `read` bug, I'm not sure this is true.

ghost/core/core/server/models/subscription-created-event.js
```diff
@@ -3,6 +3,7 @@ const ghostBookshelf = require('./base');
 
 const SubscriptionCreatedEvent = ghostBookshelf.Model.extend({
     tableName: 'members_subscription_created_events',
+    hasTimestamps: false, // if true (default), requires an updated_at column and schema changes
```
cmraible, line 6: I'm not sure I understand why this wasn't a problem before? Is it just that we've never updated a `SubscriptionCreatedEvent` before?

cmraible (commented):
Sorry for the noise on an old PR - Linear seemed to think I was actively reviewing this, so just submitting a review to clear it from my queue.

9larsons:
- [ ]

PR #22211, opened 2025-02-17
Added attribution to members `browse` endpoint

cmraible:
ref https://linear.app/ghost/issue/ENG-2023/member-attribution-data-is-included-in-read-endpoint-but-not-browse

WIP


cmraible:
This PR adds attribution data to the members returned by the members `browse` endpoint. This was originally omitted for some reason, and that may have been a _good_ reason, so I want to be careful with this. My main concern is that this will impact the performance of the members `browse` endpoint, which is an issue we've run into the past for sites with large members tables. I want to performance test this in staging with a large site before moving forward with it.

PR #22219, opened 2025-02-18
🐛 Fixed flaky missing subscription attribution

9larsons:
ref https://linear.app/ghost/issue/ENG-1561/


9larsons:
We made some underlying changes to how `MemberRepository.linkSubscription` handles `attribution`. This tag has 13 callers, only two of which pass in `attribution`:

1. `CheckoutSessionEventService.handleSubscriptionEvent`; this is called by the `checkout.session.completed` webhook, type `subscription`. In order to call `linkSubscription` we need to have an existing member.

2. `MemberRepository.create`; this would be the case for a member created via Stripe checkout, which ought to be fairly rare. I think it's appropriate to skip this case as attribution is passed, but we have no reason to believe it would be set, as attribution is _only_ coming from `urlHistory` at this point (passed by Portal).

The behavior would change for the other callers because we are now looking to the `subscription` as a backup for attribution (`subscription.metadata.*`).

So for test cases, we ought to be able to:
- Call into `CheckoutSessionEventService.handleSubscriptionEvent` where we provide an empty `attribution` object directly, and pass along the attribution data in `subscription.metadata.*`.
- The other case we're aware of is out of sync webhooks, where `customer.subscription.created` is called before `checkout.session.completed`, creating a subscription without attribution. This should no longer be the case after these changes. We can test it against `main` to see that it fails, while on the branch it passes.

Ultimately, we're changing a base assumption that now the `session.subscription` always has the `attribution` data (if present from `urlHistory`). It would be helpful to:
- Make the same change in [stripe-mocker](https://github.com/TryGhost/Ghost/blob/265f8f650da5a52d2e80679ea422c7baf0bfddc9/ghost/core/test/utils/stripe-mocker.js#L217)
- Create an e2e test truly bypasses the mocks; e.g. an existing member subscribes via Portal and we check the subscription data in the db

ghost/core/test/e2e-api/members/webhooks.test.js
```diff
@@ -2155,6 +2170,237 @@ describe('Members API', function () {
...
+            await assertMemberEvents({
+                eventType: 'SubscriptionCreatedEvent',
+                memberId: member.id,
+                asserts: [
+                    {
+                        member_id: member.id,
+                        subscription_id: subscriptionModel.id,
+
+                        // Defaults if attribution is not set
+                        attribution_id: attribution?.id ?? null,
```
cmraible, line 2275: Since we know attribution is set in this test, I think it would be clearer to remove the optional chaining and `?? null` here.

ghost/core/test/e2e-api/members/webhooks.test.js
```diff
@@ -2155,6 +2170,237 @@ describe('Members API', function () {
...
+            // subscription created event should have attribution
+            await assertMemberEvents({
+                eventType: 'SubscriptionCreatedEvent',
+                memberId: member.id,
+                asserts: [
+                    {
+                        member_id: member.id,
+                        subscription_id: subscriptionModel.id,
+
+                        // Defaults if attribution is not set
```
cmraible, line 2375: Same comment re: optional chaining

ghost/core/test/e2e-api/members/webhooks.test.js
```diff
@@ -2155,6 +2170,237 @@ describe('Members API', function () {
...
+            let webhookSignature = stripe.webhooks.generateTestHeaderString({
+                payload: webhookPayload,
+                secret: process.env.WEBHOOK_SECRET
+            });
+
+            await membersAgent.post('/webhooks/stripe/')
+                .body(webhookPayload)
+                .header('content-type', 'application/json')
+                .header('stripe-signature', webhookSignature)
+                .expectStatus(200);
```
cmraible, line 2255: Not a blocker, but it might be cool to extract this into a helper function so we can more easily test other scenarios. 

i.e. `testStripeWebhook({eventType, existingMember})`

ghost/core/test/e2e-api/members/webhooks.test.js
```diff
@@ -2155,6 +2170,237 @@ describe('Members API', function () {
...
+
+                        // Defaults if attribution is not set
+                        attribution_id: null,
+                        attribution_url: null,
+                        attribution_type: null,
+                        source: 'member'
+                    }
+                ]
+            });
+        });
```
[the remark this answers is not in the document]
9larsons: Arguably no. I'm fine removing it, comes off a little cleaner. We don't have clear outlines of what we expect for all these workflows.

ghost/core/test/e2e-api/members/webhooks.test.js
9larsons: It yields undefined which fouls up when checking against null.

ghost/core/test/e2e-api/members/webhooks.test.js
9larsons: Agreed but I'll leave that for another time. This whole test set should probably be refactored and we should cover more webhooks.

cmraible (approved):
LGTM!

PR #22234, opened 2025-02-19
Integrating React Router in ActivityPub

peterzimon:
ref AP-701

- ATM we're using our in-house routing solution in ActivityPub. On the mid/long term (ie. Public Beta / GA) it's beneficial to switch to an existing third-party solution which handles certain use cases that we'd need out of the box. One of the most popular of these tools is React Router, for which we've already put together a spike in the Posts app prototype. This PR is replacing our in-house router with React Router.


apps/admin-x-activitypub/src/routes.ts
```diff
@@ -0,0 +1,43 @@
...
+        path: 'notifications',
+        Component: Notifications
+    },
+    {
+        path: 'profile',
+        Component: Profile
+    }
+];
+
+export const ROUTE_TITLES: Record<string, string> = {
```
allouis, line 36: It would be cool if we can define this inline with the `routes` so we know that there is one for each path

apps/admin-x-activitypub/src/components/layout/Sidebar/Sidebar.tsx
```diff
@@ -2,38 +2,35 @@ import * as React from 'react';
...
+    const location = useLocation();
+    const navigate = useNavigate();
 
     return (
         <div className='sticky top-[102px] flex min-h-[calc(100vh-102px-32px)] w-[294px] flex-col border-l border-gray-200 dark:border-gray-950'>
             <div className='flex grow flex-col justify-between'>
                 <div className='flex w-full flex-col items-start gap-8 pl-4 pt-4'>
                     <div className='flex w-full flex-col gap-px'>
-                        <SidebarButton active={route === 'inbox'} onClick={() => updateRoute('inbox')}>
+                        <SidebarButton active={location.pathname === '/inbox' || location.pathname === '/'} onClick={() => navigate('/inbox')}>
```
allouis, line 16: Would love to see this combined somehow - or wrapped into SidebarButton so it's just like 
`<SidebarButton route-"/inbox">` that would be cool

allouis (commented):
Looks good, a few minor comments

apps/admin-x-activitypub/src/components/layout/Sidebar/Sidebar.tsx
peterzimon: Nice. I changed it as you recommended

PR #22237, opened 2025-02-19
Update comments.json

noezdev:
added missing translations phrases for german (du)


kevinansfield:
Thanks for the contribution @noezdev! In this case I'm going to close the PR in favour of [this more complete German translation](https://github.com/TryGhost/Ghost/pull/22241). If you have a moment to review that and provide any recommendations on the provided translations that would be greatly appreciated.

PR #22241, opened 2025-02-19
🌐 Updated German translations

JTBrinkmann:
Added missing german translations.

I couldn't get unit tests and the linter running, but this shouldn't be related to the PR itself.


ghost/i18n/locales/de/newsletter.json
```diff
@@ -1,24 +1,24 @@
...
+    "Manage subscription": "Abos verwalten",
+    "Member since": "Mitglied seit",
+    "More like this": "Mehr hiervon",
+    "Name": "Name",
+    "paid": "zahlendes",
+    "Subscription details": "Abo Details",
+    "trialing": "Probe-",
+    "Unsubscribe": "Abbestellen",
+    "View in browser": "Im Browser ansehen",
+    "You are receiving this because you are a <strong>%%{status}%% subscriber</strong> to {site}.": "Du erhälst diese Nachricht weil du <strong>%%{status}%% Mitglied</strong> von {site} bist.",
```
cathysarisky, line 18: AI says this should be erhältst ? Correct?

ghost/i18n/locales/de/newsletter.json
```diff
@@ -1,24 +1,24 @@
...
+    "free": "kostenfreies",
+    "Keep reading": "Weiterlesen",
+    "Less like this": "Weniger hiervon",
+    "Manage subscription": "Abos verwalten",
+    "Member since": "Mitglied seit",
+    "More like this": "Mehr hiervon",
+    "Name": "Name",
+    "paid": "zahlendes",
+    "Subscription details": "Abo Details",
+    "trialing": "Probe-",
```
cathysarisky, line 15: The present set of translation strings for free/paid/trialing/complimentary get substituted in for %%status%%.  That's going to result in a space after the hyphen for trialing.  Adequate, or do you want to pull the word Mitglied into the free/trialing/etc phrases and remove it from the full-sentence translation?

cathysarisky (changes requested):
Just a few minor comments, but generally LGTM!

ghost/i18n/locales/de/comments.json
```diff
@@ -60,7 +60,7 @@
     "Show comment": "Kommentar anzeigen",
     "Sign in": "Einloggen",
     "Sign up now": "Jetzt registrieren",
-    "Sort by": "",
+    "Sort by": "Sortiere nach",
```
cathysarisky, line 63: we got a second PR today for this string that has: Sortieren nach .  Who is right? :)  (This English speaker has no idea!)

noezdev: Both should fit.

ghost/i18n/locales/de/newsletter.json
noezdev: Yes, seems to be a typo "erhältst" is correct.

ghost/i18n/locales/de/newsletter.json
JTBrinkmann: You are right, that is suboptimal. The issue is that "trialing subscriber" would be translated as a compound noun in German. I could not come up with an adjective that would fit as a separate word. 

**Ideally**, instead of inserting an adjective as placeholder, it would be changed to "free_subscriber", "paid_subscriber", etc, would should also work with other languages.

Here's an alternative, **though hacky** solution:
```json
        "complimentary": "kostenfreies Mitglied",
        "free": "kostenfreies Mitglied",
        "paid": "zahlendes Mitglied",
        "trialing": "Probemitglied",
        "You are receiving this because you are a <strong>%%{status}%% subscriber</strong> to {site}.": "Du erhälst diese Nachricht weil du <strong>%%{status}%%</strong> von {site} bist.",
 ```
 But I worried that that's not future proof, as "free" etc would be used in other contexts.

ghost/i18n/locales/de/comments.json
JTBrinkmann: I agree with noezdev. Without context, I can imagine both options, but looking at the actual comment UI, I feel like "Sortiere" is more appropriate.

ghost/i18n/locales/de/newsletter.json
cathysarisky: I'm not well positioned to evaluate how bad the wording is in German, as a non-speaker.  Your ideal solution: Redoing the placeholder strings isn't something we can do just for German -- it'd require changes in the portal code, and then all the translations would need to be updated for the new strings (likely to require individual handling).  It's a good idea, but not a quick idea.  I think your 'hacky' solution would be fine, if you prefer it over the 'Probeabo Mitglied' option, and in fact, I was going to suggest it.  The strings are only used in the newsletter, and the 'status' substitutions only occur in this one place.

cathysarisky (approved):

PR #22245, opened 2025-02-19
Improved stripe listen command in dev.js script

cmraible:
no issue

Problem:
- Running `yarn dev` with `--stripe` runs the `stripe listen ...` command in a subprocess to forward webhook events to your local instance of Ghost in development.
- Currently, you have to run `stripe login` manually before this will work. If you fail to do so, the `yarn dev` script simply hangs indefinitely without any logging to indicate why.
- Running `stripe login` manually is also a non-starter in Docker since you'd have to do this for each new container instance, which isn't practical.

Solution:
- Luckily the `stripe listen ...` command accepts an `--api-key` argument, which expects your stripe secret key. This change updates the `dev.js` script to read your stripe API key from the `STRIPE_SECRET_KEY` environment variable if it's present. As long as `STRIPE_SECRET_KEY` exists, you won't need to run `stripe login` anymore.
- It also adds a timeout of 5 seconds to the initial `stripe listen` command and will exit the process if this step fails, indicating with a clear log message that you need to either set the `STRIPE_SECRET_KEY` variable or run `stripe login`
- I also added a bunch of debug statements to the dev script which were useful in debugging this, and may prove useful for similar issues in the future, so I'll leave them in there.

### DIFF
--- a/.github/scripts/dev.js
+++ b/.github/scripts/dev.js
@@ -205,11 +205,12 @@ async function handleStripe() {
 
         let stripeSecret;
         const stripeSecretKey = process.env.STRIPE_SECRET_KEY;
+        const apiKeyFlag = stripeSecretKey ? `--api-key ${stripeSecretKey}` : '';
         try {
-            const stripeListenCommand = `stripe listen --print-secret ${stripeSecretKey ? `--api-key ${stripeSecretKey}` : ''}`;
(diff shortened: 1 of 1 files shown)


cmraible:
Weirdly this PR merged but didn't close. Closing now.

PR #22252, opened 2025-02-20
🐛 Fixed missing attribution in new member and new subscription staff emails

9larsons:
ref https://linear.app/ghost/issue/ENG-2030/

We could experience a race condition on upgraded members where the SubscriptionActivatedEvent would look to the SubscriptionCreatedEvent for subscription attribution, but since both are emitted at the same time, it was possible for the Created event to not exist.

This commit passes the attribution directly to the SubscriptionActivatedEvent when creating both events in `linkSubscription`, thereby avoiding the race condition.

We maintain the fallback behavior to looking to the SubscriptionCreatedEvent in the StaffService email code to cover the other flows.

### DIFF
--- a/ghost/members-api/lib/repositories/MemberRepository.js
+++ b/ghost/members-api/lib/repositories/MemberRepository.js
@@ -1213,6 +1213,7 @@ module.exports = class MemberRepository {
                     memberId: member.id,
                     subscriptionId: subscriptionModel.get('id'),
                     offerId: offerId,
+                    attribution: attribution,
                     batchId: options.batch_id
                 });
                 this.dispatchEvent(activatedEvent, options);
--- a/ghost/core/core/server/services/staff/index.js
+++ b/ghost/core/core/server/services/staff/index.js
@@ -1,6 +1,6 @@
 const DomainEvents = require('@tryghost/domain-events');
 const labs = require('../../../shared/labs');
-const sentry = require('../../../shared/sentry');
+
 class StaffServiceWrapper {
     init() {
         if (this.api) {
@@ -30,8 +30,7 @@ class StaffServiceWrapper {
             blogIcon,
             DomainEvents,
             memberAttributionService: memberAttribution.service,
-            labs,
-            sentry
+            labs
         });
 
         this.api.subscribeEvents();
--- a/ghost/core/core/server/services/staff/index.js
+++ b/ghost/core/core/server/services/staff/index.js
@@ -1,6 +1,6 @@
 const DomainEvents = require('@tryghost/domain-events');
 const labs = require('../../../shared/labs');
-
+const sentry = require('../../../shared/sentry');
 class StaffServiceWrapper {
     init() {
         if (this.api) {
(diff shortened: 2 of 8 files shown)


ghost/members-api/lib/repositories/MemberRepository.js
```diff
@@ -1213,6 +1213,7 @@ module.exports = class MemberRepository {
                     memberId: member.id,
                     subscriptionId: subscriptionModel.get('id'),
                     offerId: offerId,
+                    attribution: attribution,
```
cmraible, line 1216: I think we should do the same thing in the other path the emits the `SubscriptionActivatedEvent` [here](https://github.com/TryGhost/Ghost/blob/20e7d692afdf235fc0a34a6634364b64de435f67/ghost/members-api/lib/repositories/MemberRepository.js#L1128-L1135) unless there's a good reason not to that I'm overlooking

cmraible: Even if it's not needed functionally, I think it would be good just for the consistency

9larsons: I'm hesitant to do that one as we - in theory? - already have a `SubscriptionCreatedEvent` with attribution and shouldn't hit a race condition, so we should let the code 'work as designed'. I think it's ok to do it where we have because we know we're creating the `SubscriptionCreatedEvent` with the same exact data we pass to the Activated event.

If we think the design was bad from the start, then I'm open to it, but I don't think we need to touch that branch here.

cmraible: Yeah okay that's fair. I _do_ think the design was bad from the start, but I don't think there's appetite to really redesign it right now, as much as I'd like to. 

IMO the subscription attribution should either be directly in the subscription model itself, or in its own table directly related to the subscription, rather than attached to the event. But that's a pretty major rework and we want to get this fixed ASAP so this'll do for now 👍

cmraible:
LGTM!

The only other thing I'd consider adding is some kind of log or Sentry message every time we send a "New paid member" email that doesn't have attribution, right before the template is rendered [here](https://github.com/TryGhost/Ghost/blob/2cc1e28eca6079d9b02c11653099ec9d3b2990b2/ghost/staff-service/lib/StaffServiceEmails.js#L117).

We won't always have attribution so this isn't necessarily wrong, but it might be good to log these events so we can confirm that we're only omitting the attribution when there truly isn't any. I think if we just log "No attribution in new paid member email for subscription ${subscriptionModelId}" that should be enough to investigate. WDYT?

cmraible:
The only valid case I'm aware of is if the subscription is created outside of Ghost for an existing member, i.e. if the site owner manually creates a subscription on behalf of a member in their stripe dashboard. We'd still receive the webhooks but no attribution data.

cmraible (approved):
LGTM! I've tested the paid flow 10+ times and it's worked every time, and same for the "new free member" flow

PR #22258, opened 2025-02-21
Upgraded NX to version 20

cmraible:
ref https://nx.dev/blog/announcing-nx-20

- NX version 20 has a new database backed cache, which is supposedly a bit faster.
- I'm also hopeful this might help with some flakiness with running NX in Docker, but it's a good idea to stay up to date regardless.

### DIFF
--- a/.github/scripts/dev.js
+++ b/.github/scripts/dev.js
@@ -316,7 +316,6 @@ async function handleStripe() {
     debug('at least one command provided');
 
     debug('resetting nx');
-    process.env.NX_DISABLE_DB = "true";
     await exec("yarn nx reset --onlyDaemon");
     debug('nx reset');
     await exec("yarn nx daemon --start");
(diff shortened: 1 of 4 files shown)


.github/scripts/dev.js
```diff
@@ -316,7 +316,6 @@ async function handleStripe() {
     debug('at least one command provided');
 
     debug('resetting nx');
-    process.env.NX_DISABLE_DB = "true";
```
cmraible, line 319: With this line, NX kept reverting to the "legacy" cache and logging a warning. I haven't seen any issue with removing this

.github/workflows/ci.yml
```diff
@@ -17,7 +17,6 @@ env:
     ~/.cache/ms-playwright/
   CACHED_BUILD_PATHS: |
     ${{ github.workspace }}/ghost/*/build
-  NX_REJECT_UNKNOWN_LOCAL_CACHE: 0
```
cmraible, line 20: This option is no longer needed with the database backed cache, see https://nx.dev/deprecated/legacy-cache

cmraible:
Some crazy changes were made by the nx migrate process to the root package.json, trying to piece together what actually changed 🤔 looks like it mostly just moved stuff around

PR #22260, opened 2025-02-22
WIP: initial screen reader improvements for comments-ui

cathysarisky:
Adding missing roles and labels for comments-ui, hiding avatars, etc.

WIP, more to come, and I need an actual screen reader user to validate.


ErisDS:
Cleaning up older PRs 💅

PR #22311, opened 2025-03-02
🎨 Added ability to vertically resize the Spam Filters text area

evanwon:
- [X] The build will pass (run `yarn test:all` and `yarn lint`)

In Ghost Admin, the Spam Filters text area was not resizable, which made it difficult to work with for sites with more than 4 rows of blocked domains. This change makes the text area vertically resizable so, if desired, users can see more domains at the same time.

### Before
The Spam Filters text area is not resizable, which makes it difficult to work with longer lists of spam domains.
image

### After
Here's what the Spam Filters text area now looks like by default in Firefox - same size as before, but note the resize indicator in the bottom right corner.
image

And here's what it looks like when you manually resize the Spam Filters text area (note that it can only be resized vertically, which I think makes the most sense for this use case).
image

### DIFF
--- a/apps/admin-x-settings/src/components/settings/advanced/SpamFilters.tsx
+++ b/apps/admin-x-settings/src/components/settings/advanced/SpamFilters.tsx
@@ -67,6 +67,7 @@ const SpamFilters: React.FC<{ keywords: string[] }> = ({keywords}) => {
                     error={!!errors.blockedEmailDomains}
                     hint={errors.blockedEmailDomains || hint}
                     placeholder={`spam.xyz\njunk.com`}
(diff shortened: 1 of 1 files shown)


sagzy:
Hello @evanwon,

Thanks for the PR! Makes sense :) Merging

evanwon:
@sagzy Just to check, anything else needed for the pending workflow?

sagzy:
> @sagzy Just to check, anything else needed for the pending workflow?

@evanwon oh my bad! I enabled auto-merge last week and did not double check whether it actually got merged in. Merged now (for real!), thanks again for the contribution 🙏

PR #22320, opened 2025-03-03
🐛 Fixed email domain blocklist not being checked when a member updates their email address

ronaldlangeveld:
ref https://linear.app/ghost/issue/ONC-797/spam-filter-not-applied-when-updating-an-email-address

With this change, we extend the email domain blocklist to member's email update feature (previously: only on signups)

- Added logic in `MemberController.js` to validate email domains against a list of blocked domains from`settingsCache (all_blocked_email_domains)` before allowing updates.
- Enhanced error messaging in `api.js` and `actions.js` in Portal to return a specific error ("Memberships from this email domain are currently restricted.") when a blocked domain error from API is detected.
- Added a new E2E test to verify that changing an email to a blocked domain is prevented (status 400 with appropriate error message) and that changing an email to a allowed domain is still allowed (status 201)
- Added the new error message to localisation files (portal.json) across multiple languages for consistent user feedback.

### DIFF
--- a/apps/portal/src/actions.js
+++ b/apps/portal/src/actions.js
@@ -483,11 +483,13 @@ async function updateProfile({data, state, api}) {
         const action = emailUpdate.success ? 'updateProfile:success' : 'updateProfile:failed';
         const status = emailUpdate.success ? 'success' : 'error';
         let message = '';
-        if (emailUpdate.error.message === 'Memberships from this email domain are currently restricted.') {
-            message = chooseBestErrorMessage(emailUpdate.error, t('Memberships from this email domain are currently restricted.'), t);
+
+        if (emailUpdate.error) {
+            message = chooseBestErrorMessage(emailUpdate.error, t('Failed to send verification email'), t);
         } else {
-            message = emailUpdate.success ? t('Check your inbox to verify email update') : t('Failed to send verification email');
+            message = t('Check your inbox to verify email update');
         }
+
         return {
             action,
             ...(emailUpdate.success ? {page: 'accountHome'} : {}),
--- a/apps/portal/src/actions.js
+++ b/apps/portal/src/actions.js
@@ -458,8 +458,8 @@ async function updateProfile({data, state, api}) {
                 })
             };
         }
-        const message = !dataUpdate.success ? t('Failed to update account data') : t('Failed to send verification email');
 
+        const message = !dataUpdate.success ? t('Failed to update account data') : t('Failed to send verification email');
         return {
             action: 'updateProfile:failed',
             ...(dataUpdate.success ? {member: dataUpdate.member} : {}),
@@ -482,7 +482,12 @@ async function updateProfile({data, state, api}) {
     } else if (emailUpdate) {
         const action = emailUpdate.success ? 'updateProfile:success' : 'updateProfile:failed';
         const status = emailUpdate.success ? 'success' : 'error';
-        const message = !emailUpdate.success ? t('Failed to send verification email') : t('Check your inbox to verify email update');
+        let message = '';
+        if (emailUpdate.error.message === 'Memberships from this email domain are currently restricted.') {
+            message = chooseBestErrorMessage(emailUpdate.error, t('Memberships from this email domain are currently restricted.'), t);
+        } else {
+            message = emailUpdate.success ? t('Check your inbox to verify email update') : t('Failed to send verification email');
+        }
         return {
             action,
             ...(emailUpdate.success ? {page: 'accountHome'} : {}),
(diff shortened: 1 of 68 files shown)


ghost/members-api/lib/controllers/MemberController.js
```diff
@@ -41,6 +49,13 @@ module.exports = class MemberController {
             return res.end('No Permission.');
         }
 
+        const blockedEmailDomains = this._settingsCache.get('all_blocked_email_domains');
+        const emailDomain = req.body.email.split('@')[1]?.toLowerCase();
+        if (emailDomain && blockedEmailDomains.includes(emailDomain)) {
+            res.writeHead(400);
```
ronaldlangeveld, line 55: I tried adding but for some reason it completely crashes Express / Node and I couldn't get to the bottom of why it's happening.

```
                message: tpl(messages.blockedEmailDomain)
            });
```

ghost/core/test/e2e-api/members/send-magic-link.test.js
```diff
@@ -376,6 +376,19 @@ describe('sendMagicLink', function () {
...
+
+        it('blocks changing email to a blocked domain', async function () {
+            settingsCache.set('all_blocked_email_domains', {value: ['blocked-domain-setting.com']});
+            const email = 'hello@cool-domain-setting.com';
+            await membersService.api.members.create({email, name: 'Member Test'});
+
+            await membersAgent.post('/api/member/email/')
+                .body({
+                    email: 'hello@blocked-domain-setting.com',
+                    identity: '12345678'
```
ronaldlangeveld, line 388: There's currently update member update email tests, since our E2E test suite doesn't have a way of populating / mocking the identity field.

sagzy: Nice! I think the opposite test would also be useful: 'Allows changing email to a non-blocked email domain'

ghost/members-api/lib/controllers/MemberController.js
sagzy: A `throw` would need a `try/catch` block within the method (cf. `throw` example a couple of lines below in the method).

I think your approach with writing the response with a 400 directly in that case is good 👍

ghost/members-api/lib/controllers/MemberController.js
```diff
@@ -1,4 +1,9 @@
 const errors = require('@tryghost/errors');
+const tpl = require('@tryghost/tpl');
+
+const messages = {
+    blockedEmailDomain: 'Signups from this email domain are currently restricted.'
```
sagzy, line 5: Would rename the error message to e.g. `Memberships from this email domain are currently restricted.`

sagzy (commented):
Hey @ronaldlangeveld, LGTM on general, left two comments :)

apps/portal/src/actions.js
```diff
@@ -482,7 +482,12 @@ async function updateProfile({data, state, api}) {
     } else if (emailUpdate) {
         const action = emailUpdate.success ? 'updateProfile:success' : 'updateProfile:failed';
         const status = emailUpdate.success ? 'success' : 'error';
-        const message = !emailUpdate.success ? t('Failed to send verification email') : t('Check your inbox to verify email update');
+        let message = '';
+        if (emailUpdate.error.message === 'Memberships from this email domain are currently restricted.') {
+            message = chooseBestErrorMessage(emailUpdate.error, t('Memberships from this email domain are currently restricted.'), t);
+        } else {
+            message = emailUpdate.success ? t('Check your inbox to verify email update') : t('Failed to send verification email');
+        }
```
sagzy, line 490: I'd suggest not matching for `emailUpdate.error.message === 'Memberships from this email domain are currently restricted.'` here, as this limits our ability to have more error messages in the future.

Instead I'd recommend doing something like:

```
if (emailUpdate.error) {
    message = chooseBestErrorMessage(emailUpdate.error, t('Failed to send verification email'), t);
} else {
   message = t('Check your inbox to verify email update');
}
```

and then add the string `"Memberships from this email domain are currently restricted."` under `setupSpecialMessages` [here](https://github.com/TryGhost/Ghost/blob/a5f8860bcc141a7fa8415013ff2768e6f058e0eb/apps/portal/src/utils/errors.js#L43C1-L65C7)

ghost/core/test/e2e-api/members/send-magic-link.test.js
sagzy: Added in a588ec6b8859ea48ab9fda2f90dc5e676364a8b7

apps/portal/src/actions.js
sagzy: Fixed in 8bd4ee18d9caa85bf5f6ed5ddb7a577311833b86

Ghost-Slimer:
Deployed to **staging** with ID: `3254`

PR #22330, opened 2025-03-03
Added post delete mutation for posts in the `admin-x-activitypub` app

mike182uk:
refs [AP-806](https://linear.app/ghost/issue/AP-806/wire-up-the-delete-buttons-to-the-deleteid-api)

Added post delete mutation for posts in the `admin-x-activitypub` app


apps/admin-x-activitypub/src/hooks/use-activity-pub-queries.ts
```diff
@@ -815,3 +821,93 @@ export function useInboxForUser(options: {enabled: boolean}) {
...
+                            ...page,
+                            posts: page.posts.filter((item: Activity) => item.id !== id)
+                        };
+                    })
+                };
+            });
+
+            // Update the thread cache
+            const threadQueryKey = QUERY_KEYS.thread(null);
+            const previousThreads = queryClient.getQueryData<{posts: Activity[]}>(threadQueryKey);
```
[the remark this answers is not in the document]
mike182uk: Post can appear in multiple threads which is why we can't target a single thread here

apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
```diff
@@ -858,6 +865,7 @@ const ArticleModal: React.FC<ArticleModalProps> = ({
                                     <>
                                         <FeedItem
                                             actor={item.actor}
+                                            allowDelete={false}
```
allouis, line 868: Why is this hardcoded to false?

apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
```diff
@@ -878,7 +886,8 @@ const ArticleModal: React.FC<ArticleModalProps> = ({
                             {object.type === 'Note' && (
                                 <FeedItem
                                     actor={actor}
-                                    commentCount={object.replyCount ?? 0}
+                                    allowDelete={false}
```
allouis, line 889: Same here

apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
```diff
@@ -452,7 +455,11 @@ const ArticleModal: React.FC<ArticleModalProps> = ({
 
         // Update the replyCount on the current activity loaded in the modal
         // This is used for when we navigate via the history
-        object.replyCount = (object.replyCount ?? 0) + 1;
+        setReplyCount(replyCount + 1);
```
allouis, line 458: I think it is preferable to pass a function when you are modifying state based on the current state - this can avoid things getting out of sync e.g.

```ts
const [state, setState] = useState(0);
setState(state + 1);
setState(state + 1);
setState(state + 1);
```

On the next render `state` will have a value of `1` NOT `3` - this is because the setStates are called synchronously and then the render happens after the batch

In this case I don't think it causes a bug, but I do think it's good practice to instead pass a function for increment/decrement like below

```ts
setReplyCount((count) => count + 1);
setReplyCount((count) => count - 1);
```

apps/admin-x-activitypub/src/hooks/use-activity-pub-queries.ts
```diff
@@ -815,3 +829,247 @@ export function useInboxForUser(options: {enabled: boolean}) {
 
     return {inboxQuery, updateInboxActivity};
 }
+
+export function useDeleteMutationForUser(handle: string) {
```
allouis, line 833: Damn this query cache is painful 😢

apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
mike182uk: Same as https://github.com/TryGhost/Ghost/pull/22330#discussion_r1980925389

apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
mike182uk: Nice one, thanks for the nudge

mike182uk: updated in https://github.com/TryGhost/Ghost/pull/22330/commits/4700cf870bd09a527abf1c32ad73cd22da227977

allouis (approved):

PR #22356, opened 2025-03-04
Added delete action and dialog to ActivityPub feed item

minimaluminium:
ref AP-817, AP-818, AP-819 

- added a new component AlertDialog to Shade
- replaced the old Menu component with Popover from Shade for the feed item menu
- added delete option to the menu
- added a confirmation dialog to the delete action


apps/admin-x-activitypub/src/components/feed/FeedItem.tsx
```diff
@@ -250,8 +238,8 @@ const FeedItem: React.FC<FeedItemProps> = ({actor, object, layout, type, comment
     // });
 
     const UserMenuTrigger = (
-        <Button
-            className={`transition-color relative z-[9998] flex h-[34px] w-[34px] items-center justify-center ${layout === 'feed' ? 'rounded-full' : 'rounded-md'} bg-white hover:bg-gray-100 dark:bg-black dark:hover:bg-gray-950 ${(layout === 'feed' || layout === 'modal') && 'ml-auto'}`}
+        <ButtonX
```
peterzimon, line 241: We should replace and use the `Button` component from Shade instead of the old system. What's the reason to keep the old one here?

apps/admin-x-activitypub/src/components/feed/FeedItemMenu.tsx
```diff
@@ -0,0 +1,90 @@
...
+    Button,
+    Popover,
+    PopoverClose,
+    PopoverContent,
+    PopoverTrigger,
+    buttonVariants
+} from '@tryghost/shade';
+
+interface FeedItemMenuProps {
+    trigger: React.ReactNode;
```
peterzimon, line 20: I really like this pattern of passing the trigger to the dialog! What do you think about instead of passing on a ReactNode component as a prop, so the menu could be used like this:

```
<FeedItemMenu onCopyLink={...} onDelete={...}>
   <Button>Open menu</Button>
</FeedItemMenu>
```

Does this make sense to you?

peterzimon (changes requested):

peterzimon (approved):

PR #22358, opened 2025-03-04
Deprecated `models.init` pattern for initializing models

daniellockyer:
refs https://ghost.slack.com/archives/C02G9E68C/p1740997729615089 ref https://linear.app/ghost/issue/ENG-2071/remove-modelsinit-pattern

- `models.init` will dynamically require all the files in the folder and re-export them
- this is useful in that you can add a model, but the downside is that it breaks all editor autocomplete because we don't know the exports until runtime
- more generally, this pattern is super annoying and we always have to remember to do `models.init` in unit tests
- to fix that, we can deprecate the use of this (there are some other places outside of this codebase we need to remove it from too) and explicit export all the files
- this means you have to add your new model to this file, but that's better than not having any types available

### DIFF
--- a/ghost/core/core/boot.js
+++ b/ghost/core/core/boot.js
@@ -88,6 +88,12 @@ async function initCore({ghostServer, config, frontend}) {
     require('./shared/url-utils');
     debug('End: Load urlUtils');
 
+    // Models are the heart of Ghost - we want to explicitly load them here so their require time
+    // isn't hidden in other steps
+    debug('Begin: models');
+    require('./server/models');
+    debug('End: models');
+
     // Settings are a core concept we use settings to store key-value pairs used in critical pathways as well as public data like the site title
     debug('Begin: settings');
     const settings = require('./server/services/settings/settings-service');
--- a/ghost/core/core/boot.js
+++ b/ghost/core/core/boot.js
@@ -88,12 +88,6 @@ async function initCore({ghostServer, config, frontend}) {
     require('./shared/url-utils');
     debug('End: Load urlUtils');
 
-    // Models are the heart of Ghost - this is a syncronous operation
-    debug('Begin: models');
-    const models = require('./server/models');
-    models.init();
-    debug('End: models');
-
     // Settings are a core concept we use settings to store key-value pairs used in critical pathways as well as public data like the site title
     debug('Begin: settings');
     const settings = require('./server/services/settings/settings-service');
--- a/ghost/core/core/server/models/index.js
+++ b/ghost/core/core/server/models/index.js
@@ -1,10 +1,4 @@
-/**
- * Dependencies
- */
-
-const _ = require('lodash');
-const debug = require('@tryghost/debug')('models');
-const glob = require('glob');
+/* eslint-disable max-lines */
 
 // enable event listeners
 require('./base/listeners');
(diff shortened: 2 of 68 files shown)


ghost/core/core/server/models/index.js
```diff
@@ -1,35 +1,87 @@
...
+    // `base` file does not export a Base model
+    Base: require('./base'),
 
-function init() {
-    const baseNow = Date.now();
-    exports.Base = require('./base');
-    debug(`${Date.now() - baseNow}ms - Base.js require`);
-
-    let modelsFiles = glob.sync('!(index).js', {cwd: __dirname});
-    modelsFiles.forEach((model) => {
```
vikaspotluri123, line 23: Does it make sense to add this to a test? e.g. that all `ghost/core/core/server/models/*.js` are available/in the require cache?

daniellockyer: Good shout! I'll add that once I figure out this browser test failure 😆

ghost/core/core/server/models/index.js
```diff
@@ -1,35 +1,87 @@
...
+    ...require('./stripe-customer-subscription'),
+    ...require('./stripe-price'),
+    ...require('./stripe-product'),
+    ...require('./subscription-created-event'),
+    ...require('./suppression'),
+    ...require('./tag-public'),
+    ...require('./tag'),
+    ...require('./user'),
+    ...require('./webhook')
+};
```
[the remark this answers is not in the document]
vikaspotluri123: If we want this check to be run as part of the test suite, it would need to follow the existing unit tests (written in javascript, using BDD)

ghost/core/core/server/models/index.js
vikaspotluri123: The rabbit gave a pretty good starting point 🐇🧙‍♂️ https://github.com/TryGhost/Ghost/pull/22358#discussion_r1991955011

ghost/core/core/server/models/index.js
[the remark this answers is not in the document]
daniellockyer: 😆

ghost/core/core/boot.js
```diff
@@ -88,12 +88,6 @@ async function initCore({ghostServer, config, frontend}) {
     require('./shared/url-utils');
     debug('End: Load urlUtils');
 
-    // Models are the heart of Ghost - this is a syncronous operation
-    debug('Begin: models');
-    const models = require('./server/models');
-    models.init();
-    debug('End: models');
-
```
ErisDS, line 96: I'm curious about this part.

I believe that if we don't do the require in the boot file here, it'll get offloaded to whatever file requires it first, which will be the settings-service below. 

This will "hide" the work to build the require cache for models, which I think is fairly hefty?

daniellockyer: Technically yes. I removed it because there was no more "work" to do, ie. I'm removing the init function.

I can leave a models require in, indifferent either way

ErisDS: Requiring is heavy/slow though right? So feels right to me to have it wrapped in debugs and done in boot, if that's ok?

ErisDS (changes requested):

ghost/core/core/boot.js
daniellockyer: Fixed!

daniellockyer:
Sadly, I don't have the time to find the cause of the failure. Perhaps one day 🙂

PR #22363, opened 2025-03-04
Update portal.json

Nasnl:
made example consistent with other translation files


cathysarisky (approved):
thanks for the localization! :)

PR #22364, opened 2025-03-04
Update ghost.json

Nasnl:
Small improvements in Dutch (switched from Flemish expressions to formal Dutch).


cathysarisky:
Hello and thank you! @Nasnl !

@VirtualRoy or @dvdwinden , do you have a moment to review this small Dutch PR?  When we get to tweaking words, I always like to have a native speaker look it over, instead of me and the AI! :) Thanks!

ChatGPT is flagging the comma removal as incorrect, but I trust your Dutch more than it, for sure! :)

VirtualRoy:
> Hello and thank you! @Nasnl !
> 
> @VirtualRoy or @dvdwinden , do you have a moment to review this small Dutch PR? When we get to tweaking words, I always like to have a native speaker look it over, instead of me and the AI! :) Thanks!
> 
> ChatGPT is flagging the comma removal as incorrect, but I trust your Dutch more than it, for sure! :)

Hi @cathysarisky!

Regarding the switch from "kan" → "kunt" and "zal" → "zult", both forms are correct, but the updated version feels slightly more formal. It’s more a matter of preference rather than correctness.

As for the removed comma, I don’t think it was necessary to remove it. In Dutch, commas are often used to make sentences easier to read and avoid ambiguity. While Nasnl’s version isn’t incorrect, I personally prefer the original structure as it flows better.

Hope this helps!

Nasnl:
Hi @VirtualRoy and @cathysarisky,

@VirtualRoy is right that all forms mentioned of the verb are possible in Dutch, and the original was not wrong, so it's more a subtle change than a correction. Yet, I plead for this small improvement for the following reason (I asked ChatGPT to formulate it short and powerful for me; you can check this independently from my opinion).
-------------
In transactional emails the standard, grammatically correct form is "je kunt". This is because:

Standard Conjugation:
In Dutch, the second person singular of "kunnen" in a normal subject–verb order should include the -t ending. Thus, when “je” is the subject preceding the verb, “je kunt” is correct.

Context and Formality:
Transactional emails, even if written in a relatively informal tone by using “je” instead of “u”, benefit from adhering to standard grammar rules. “Je kunt” conveys professionalism and clarity.

Colloquial Usage:
While “je kan” is often heard in informal speech or certain dialects, it is not considered the standard form in written communications. For transactional emails aimed at a broad audience, using the standard form is advisable.
----------

So, while even me would sometimes *say* 'je kan', I would never write it because it doesn't seem professional. For Ghost, even in informal settings, you would not want to run the risk that the owner looks less professional.

The commas in the translations are used as they would be used in English writing, but not as in Dutch writing. In Dutch, commas are used less than English because the grammatic construction of sentences requires them to be used less, and the used translations are an example of that. Having said that, this is an even more subtle twist, so I'm not going to try to convince you, for running the risk of being seen as a 'taal-nazi' (that's Dutch for someone who goes too far in being right on language topics). ;-) 

Thanks for considering guys! Either way, it's good.

VirtualRoy:
These PR's should be so fun to audit, respect to you @cathysarisky!

I should clarify that I agree with @Nasnl on "kan" → "kunt", "zal" → "zult". It's an improvement.

cathysarisky:
Sometimes it is a little tedious, but I love that we can have Ghost in all the languages! :) 

So it sounds like there's good agreement on the verb changes.  What shall we do about the commas?

Nasnl:
@dvdwinden Yup, completely agree; that is a better translation and it delivers a better intent of the original message!

dvdwinden:
@Nasnl Great! Feel free to make the change, and then it can be merged 👏

Thank you all for contributing with this level of detail.

Nasnl:
@dvdwinden Thanks! Done. Making things better step by step.

dvdwinden (approved):

PR #22405, opened 2025-03-06
Added Captcha setting to Ghost Admin

sam-lord:
ref BAE-400

This is only enabled when the lab setting is at the moment.

We should be able configure the keys from the front end too for self-hosters, we can add that next.


Ghost-Slimer:
Deployed to **staging** with ID: `3266`

PR #22419, opened 2025-03-10
🐛 Fixed slug regeneration on whitespace change

ronaldlangeveld:
ref https://linear.app/ghost/issue/ONC-794/support-escalation-re-2-being-appended-to-slug-even-when-not-duplicate

- Added trimmed title check in `generateSlugTask` to compare `currentTitle` and `newTitle`.
- Prevents slug updates for whitespace-only edits, avoiding unnecessary calls to
  `/ghost/api/admin/slugs/post/:slug/`.
- The API endpoint generates unique slugs but doesn't consider the current post,
  so even unchanged titles trigger `-<number>` suffixes (e.g., `title-2`) if called.
- Updated tests to verify `generateSlug` isn’t called, preventing duplicate slug suffixes.

### DIFF
--- a/ghost/admin/app/controllers/lexical-editor.js
+++ b/ghost/admin/app/controllers/lexical-editor.js
@@ -952,6 +952,17 @@ export default class LexicalEditorController extends Controller {
             return;
         }
 
+        // Prevent slug regeneration when only whitespace changes are made to the title.
+        // This check compares the trimmed versions of the current and new titles to ensure
+        // that differences in leading or trailing spaces do not trigger unnecessary slug updates.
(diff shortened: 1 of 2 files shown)


9larsons (approved):
LGTM to me, didn't think about whitespace changes when this was changed. Even if it's not what they're seeing it's a good case to handle 👍.

ronaldlangeveld:
> LGTM to me, didn't think about whitespace changes when this was changed. Even if it's not what they're seeing it's a good case to handle 👍.

Thanks for the review and merge! 😊

PR #22427, opened 2025-03-11
Added onboarding to ActivityPub

minimaluminium:
ref AP-857

- hides Explore when the ActivityPub feature flag is on
- adds static steps for the onboarding experience


peterzimon (changes requested):
@minimaluminium Super cool! Could we put what's in the ActivityPub app (so everything except the main menu changes) behind a soft feature flag `onboarding`?

peterzimon (approved):

Issue #22417, opened 2025-03-08
Amazon link cards are not displayed correctly in articles


hshiozawa:
### Issue Summary

When embedding an Amazon link in an article, the link card does not display properly.

- The image does not appear.
- The title and description contain identical content.

### Steps to Reproduce

- Create a new article.
- Insert an Amazon link (e.g., https://www.amazon.com/Harry-Potter-Sorcerers-Stone-Book/dp/B017V4IMVQ/).
- Screenshot is shown below.

![Image](https://github.com/user-attachments/assets/6b0fb163-bc3c-422d-a116-ff48b79d4c5d)

### Ghost Version

Ghost v5.110.4

### Node.js Version

v20.18.1

### How did you install Ghost?

Install from Source

### Database type

MySQL 8

### Browser & OS version

Chrome 133.0.6943.142, macOS 15.3.1

### Relevant log / error output

```shell

```

### Code of Conduct

- [x] I agree to be friendly and polite to people in this repository

hshiozawa:
# Background
- I am using Ghost (Pro), but Amazon link cards are not working, which negatively impacts the article layout.
- I checked [this discussion](https://forum.ghost.org/t/images-not-visible-in-amazon-link-cards/25994/5).  I would like to contribute to fixing this issue.
- I built and debugged Ghost on my local machine. It seems that using 'metascraper-amazon' allows it to work correctly.
  - The key point is to add metascraper-amazon at the beginning of [this package list](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/services/oembed/OEmbedService.js#L289-L299). It does not work correctly if added at the end.

![Image](https://github.com/user-attachments/assets/3c1615c7-d295-4f48-b035-c550aecc9ca8)

cathysarisky:
If you have a fix, please open a PR! :)

hshiozawa:
I created a PR #22431.

cathysarisky:
Possibly related? https://github.com/TryGhost/Ghost/pull/22451

hshiozawa:
ping

hshiozawa:
This issue has not been resolved yet.

amyanger:
> This issue has not been resolved yet.

I had a go :)

hshiozawa:
This issue has not been resolved yet.

PR #22431, opened 2025-03-11
Fix amazon link card break

hshiozawa:
Fix amazon link card (#22417)

--
Checks.


9larsons:
Thanks for the PR. In testing, I haven't been able to get this to successfully work for products on Amazon. I also don't see substantive updates to this package, suggesting it's not really maintained while the root issue here is that Amazon often changes how they do this. 

Can you provide me a snippet of this working for you? I'm inclined to close this out and leave it as-is, as this seems like a lateral move.

CleanShot 2025-03-12 at 07 16 15

hshiozawa:
Thank you for testing.

When I checked in my local environment, it seemed to take a long time to connect to Amazon, so I experimented by extending the fetchPage timeout to about 10 seconds. (Could the request be timing out?)

I plan to attach a video later showing it working properly.

hshiozawa:
I don’t think the failure of the Amazon embed you mentioned is caused by my fix (metascraper-amazon). I believe this issue (request timeout) would also occur even before metascraper-amazon was introduced (i.e., on the main branch). Am I wrong?

metascraper-amazon is just a metadata parser, and the fetch request is actually generated by OEmbedService.
[OEmbedService.js#L486](https://github.com/TryGhost/Ghost/blob/72199cdbc9ded4d71d4448e046949ae0266b5abd/ghost/core/core/server/services/oembed/OEmbedService.js#L486)

hshiozawa:
Just adding a quick note to link this PR to the related issue:

https://github.com/TryGhost/Ghost/issues/22590  

I'd be happy to update this PR if needed to help move it forward.

ErisDS:
Sorry that this fix wasn't mergeable, I'm cleaning up some older PRs, but feel free to re-submit with a more maintainable solution :)

PR #22435, opened 2025-03-11
Put new replies on top when they are added in a thread in ActivityPub

minimaluminium:
ref AP-874

- currently, new replies are added to the bottom of the thread when they're added
- this adds a new parameter `addToTop` to `addToThread()` function, so that new replies are placed on top
- it still maintains the default old to new order when the thread is opened later

### DIFF
--- a/apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
+++ b/apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
@@ -448,7 +448,7 @@ const ArticleModal: React.FC<ArticleModalProps> = ({
         // Add the new reply to the thread
         activity.object.authored = true;
         activity.id = activity.object.id;
-        addToThread(activity);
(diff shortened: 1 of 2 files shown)


apps/admin-x-activitypub/src/components/feed/ArticleModal.tsx
```diff
@@ -448,7 +448,7 @@ const ArticleModal: React.FC<ArticleModalProps> = ({
         // Add the new reply to the thread
         activity.object.authored = true;
         activity.id = activity.object.id;
-        addToThread(activity);
+        addToThread(activity, true);
```
mike182uk, line 451: `addToThread` is only ever used here - Maybe adding to the top of the thread should always be the default behaviour rather than us having conditional branching?

minimaluminium: Good point!

Issue #22445, opened 2025-03-11
Add user role to for expanded editor permissions


9larsons:
### Issue Summary

Currently only admins (or owners) can manage members and moderate comments, with the only next-most-privileged role being the editor. It would be beneficial to add another level between that can do editing, member management, comment moderation, etc. that doesn't give full control over the site and settings.

### Steps to Reproduce

f

### Ghost Version

current

### Node.js Version

20

### How did you install Ghost?

local

### Database type

MySQL 5.7

### Relevant log / error output

```shell

```

### Code of Conduct

- [x] I agree to be friendly and polite to people in this repository

PR #22450, opened 2025-03-11
Migration for super editor

cathysarisky:
Migrations and tests related to migrations, for the super user feature

Ref: https://github.com/TryGhost/Ghost/issues/22445


ghost/core/core/server/data/migrations/versions/5.112/2025-03-07-13-00-00-add-comments-permissions-to-super-editor.js
9larsons, line 1: I think it might be preferable to do all permissions in one go. It's too bad we don't have a bulk insert for `addPermissionsToRole`... though we'd need to do some validation that probably isn't worth it.

9larsons (commented):
Looks good! I'm going to test migrations before merging.

I think we can remove the Collections permissions, fwiw. I'm going to check with Daniel and Fabien on that as we stripped out Collections code recently, and this might be a holdover.

ghost/core/core/server/data/migrations/versions/5.112/2025-03-07-13-00-00-add-comments-permissions-to-super-editor.js
cathysarisky: Sure. Doesn't matter at all if they happen at once or separately.

9larsons (approved):
All good here - one final thing. I've tested the migrations on both MySQL and SQLite. I'm going to test once more real quick with staging, so it'll be a little bit while I make a build from this.

I've bumped the migrations version because we're releasing 5.112 shortly, so we need to make sure this points to 5.113.

PR #22454, opened 2025-03-12
Wired up custom endpoints for posts and liked posts on Profile view

vershwal:


apps/admin-x-activitypub/src/api/activitypub.test.ts
```diff
@@ -1255,8 +957,16 @@ describe('ActivityPubAPI', function () {
         });
     });
 
-    describe('getFeed', function () {
-        test('It returns an array of posts in a feed', async function () {
+    // Use the helper function to test each endpoint that returns a paginated list of posts
+    testPaginatedApiEndpoint('getFeed', '.ghost/activitypub/feed');
+    testPaginatedApiEndpoint('getInbox', '.ghost/activitypub/inbox');
+    testPaginatedApiEndpoint('getPostsByAccount', '.ghost/activitypub/posts');
+    testPaginatedApiEndpoint('getPostsLikedByAccount', '.ghost/activitypub/posts/liked');
```
mike182uk, line 964: 🤔 I'm not sure about this, if one of the tests fail, how will i know which endpoint it has failed for?

apps/admin-x-activitypub/src/api/activitypub.test.ts
```diff
@@ -1255,8 +957,16 @@ describe('ActivityPubAPI', function () {
         });
     });
 
-    describe('getFeed', function () {
-        test('It returns an array of posts in a feed', async function () {
+    // Use the helper function to test each endpoint that returns a paginated list of posts
```
mike182uk, line 960: no need for comment 🙂

apps/admin-x-activitypub/src/api/activitypub.test.ts
```diff
@@ -1287,7 +997,7 @@ describe('ActivityPubAPI', function () {
                 fakeFetch
             );
 
-            const actual = await api.getFeed();
+            const actual = await (api[methodName as keyof typeof api] as (next?: string) => Promise<GetFeedResponse>)();
```
mike182uk, line 1000: Is `GetFeedResponse` common across all of the paginated endpoints? This type was specific for the return value of the `feed` endpoint

apps/admin-x-activitypub/src/hooks/use-activity-pub-queries.ts
```diff
@@ -79,39 +79,11 @@ const QUERY_KEYS = {
         return ['thread', id];
     },
     feed: ['feed'],
-    inbox: ['inbox']
+    inbox: ['inbox'],
+    postsByAccount: ['posts_by_account'],
+    postsLikedByAccount: ['posts_liked_by_account']
```
mike182uk, line 84: How about just `account_posts` and `account_likes`? (similar to `account_follows` above)

apps/admin-x-activitypub/src/hooks/use-activity-pub-queries.ts
```diff
@@ -955,6 +927,29 @@ export function useAccountFollowsForUser(handle: string, type: AccountFollowsTyp
     });
 }
 
+function updateActivityInCache(queryClient: QueryClient, queryKey: unknown[], id: string, updated: Partial<Activity>) {
```
mike182uk, line 930: I recently added a method to update an activity, can that be used instead of this or does this do something different?

apps/admin-x-activitypub/src/api/activitypub.test.ts
vershwal: Right, I reused it to make it work initially. But need to clean this.. the profile has exact same response. So I'll rename this to something like `PaginatedPostsResponse` and reuse it.

apps/admin-x-activitypub/src/api/activitypub.test.ts
vershwal: I have removed the refactoring from these tests for now. I'll do the refactoring in a separate PR

PR #22458, opened 2025-03-12
🌐 Added missing Swedish portal translation

julianhessel:
This is the correct swedish translation for this string.

### DIFF
--- a/ghost/i18n/locales/sv/portal.json
+++ b/ghost/i18n/locales/sv/portal.json
@@ -179,7 +179,7 @@
     "Too many different sign-in attempts, try again in {{number}} days": "För många olika inloggningsförsök, testa igen om {{number}} dagar.",
     "Too many different sign-in attempts, try again in {{number}} hours": "För många olika inloggningsförsök, testa igen om {{number}} timmar.",
     "Too many different sign-in attempts, try again in {{number}} minutes": "För många olika inloggningsförsök, testa igen om {{number}} minuter.",
-    "Too many sign-up attempts, try again later": "För många olika inloggningsförsök, prova igen senare.",
+    "Too many sign-up attempts, try again later": "För många olika registreringsförsök, prova igen senare.",
     "Try free for {{amount}} days, then {{originalPrice}}.": "Prova gratis i {{amount}} dagar, sen betalar du {{originalPrice}}.",
     "Unable to initiate checkout session": "Kan inte initiera utcheckningssession",
     "Unlock access to all newsletters by becoming a paid subscriber.": "Få tillgång till alla nyhetsbrev genom att bli betalande prenumerant",
--- a/ghost/i18n/locales/sv/portal.json
+++ b/ghost/i18n/locales/sv/portal.json
@@ -179,7 +179,7 @@
     "Too many different sign-in attempts, try again in {{number}} days": "För många olika inloggningsförsök, testa igen om {{number}} dagar.",
     "Too many different sign-in attempts, try again in {{number}} hours": "För många olika inloggningsförsök, testa igen om {{number}} timmar.",
     "Too many different sign-in attempts, try again in {{number}} minutes": "För många olika inloggningsförsök, testa igen om {{number}} minuter.",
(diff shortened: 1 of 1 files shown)


julianhessel:
> **Actionable comments posted: 0**
> 
> 🧹 Nitpick comments (1)
> > ghost/i18n/locales/sv/portal.json (1)> `182-182`: **Translation Terminology Consistency**
> > > The key `"Too many sign-up attempts, try again later"` now translates to `"För många olika inloggningsförsök, prova igen senare."` However, the Swedish term _"inloggningsförsök"_ typically refers to "login attempts" rather than "sign-up attempts." In Swedish, "sign-up" is often translated as "registrering" so you might consider using _"registreringsförsök"_ to better capture the intended meaning.
> > > Please verify that this terminology aligns with your overall localization strategy for sign-up related messages. If a different term is more appropriate, you might update the translation accordingly.
> 
> 📜 Review details
> **Configuration used: CodeRabbit UI** **Review profile: CHILL** **Plan: Pro**
> 
> 📥 Commits
> 📒 Files selected for processing (1)


The term sign-up is indeed "registrering" in swedish, but this string is actually shown when a user is trying to sign-in rather then sign-up when the console sends error 429.

cathysarisky:
Hey @julianhessel :  So weirdly, I'm not sure where this string is coming from.  I definitely set it up for Portal translation about 5 months ago along with all the others, but I went back to see if I could figure out if it was sign-in or sign-up (since we'd like to get that right for ALL languages), and I can't find it.  So, to clarify:  Are you seeing this string when trying to sign in, not sign up?  And this is within portal, in the (black) popover error message? I'd like to replicate it and see if I can track down which of several different error flows is creating the wrong message.  But first I've got to replicate it.

julianhessel:
Hey @cathysarisky!
Sorry for the slow reply and thanks for the work with translations overall!
I see that the messages have been updated since the start of this issue in the live version of Ghost. Right now they are being translated, but not the same string is chosen in the custom **sign-in** and built in portal **sign-in**:

Currently the string that seems to be viewed when error 429 gets posted on the **sign-in** form for my custom sign-in to be translated from the string: "Failed to send magic link email"
Skärmavbild 2025-03-14 kl  13 52 26

While on the built-in portal that I still have enabled but hidden it seems to show the string translated "Failed to log in, please try again"
Skärmavbild 2025-03-14 kl  13 53 06

It would probably be best for user understanding if one of the strings for the timeout would be used:
Skärmavbild 2025-03-14 kl  14 34 55

Let me know if you need more details!

cathysarisky:
A little context:  The Portal receives an API error message, and attempts to match it to one of the existing translation strings.  If the server doesn't return the timeout message, then Portal can't display it.  

If I understand your screenshots correctly, you're getting different but reasonably correct error messages for attempts to sign in with the custom form and the portal, right, and neither is the string you're trying to change in this PR?  Or have I misunderstood?

julianhessel:
That's correct. The thing is that when I was in contact with Ghost support this week, I had this string show up when error 429 was shown in console:
errormsg
Which is why I was asked to contribute to the translation of this string to have it display correctly in Swedish as this translation was missing from the .json file.
Since then, something must have been updated, cause now it's the strings above that are shown instead.
And as you mention, these messages are somewhat correct and translated now, although It may still prompt users to spam the "log-in" button as there is no mention of "try again later" in these strings sadly.

cathysarisky:
So I think it makes sense to add a string that is the _correct_ translation, and then if the underlying message from the server is wrong, that's the place it should be fixed, not by adding an incorrect translation to compensate.  I took a look for the incorrect string, and wasn't immediately able to find it, but if you can reproduce that screenshot and provide the steps to do so, please open an issue!

julianhessel:
Yeah, as the string is no longer used "incorrectly" in the portal. I changed the translation to mean sign-up instead of sign-in as intended since I can no longer reproduce the string showing up during sign-in.

cathysarisky (approved):
an AI review shows no problems

PR #22534, opened 2025-03-17
feat: change add tags to edit tags

### LINKED ISSUE #21414: Bulk adding tags can create "duplicate" `posts_tags` rows

im-adithya:
Fixes https://github.com/TryGhost/Ghost/issues/21414

## Description

As per @ErisDS' suggestion in https://github.com/TryGhost/Ghost/pull/21787, I tried to make some changes in the UI. I thought editing/updating tags is better than adding tags so I modified the add tags modal to do that instead.

## Recording

https://github.com/user-attachments/assets/5686222f-7ef1-4603-8797-2457be5e6241

## Checklist

### DIFF
--- a/ghost/admin/app/components/posts-list/context-menu.hbs
+++ b/ghost/admin/app/components/posts-list/context-menu.hbs
@@ -46,8 +46,8 @@
     {{/if}}
 
     <li>
-        <button class="mr2" type="button" {{on "click" this.addTagToPosts}} data-test-button="add-tag">
-            <span>{{svg-jar "tag"}}Add a tag</span>
+        <button class="mr2" type="button" {{on "click" this.updatePostTags}} data-test-button="edit-tags">
+            <span>{{svg-jar "tag"}}Edit tags</span>
         </button>
     </li>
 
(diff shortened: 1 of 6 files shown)


im-adithya:
@kevinansfield I need input on how to proceed with multiple post selection while adding/removing tags. Currently I show the union of all the posts selected, and the updates are currently processed as follows:
- add tags to all posts if created newly
- remove tags from all posts where it exists if a tag is removed

im-adithya:
Sorry for that 😬 

I originally thought there was no option to select multiple posts and call "Add a tag" for all at once. So I thought if it's just for a single post, we better call it Edit instead of Adding similar to how we do it in the sidebar* to avoid confusion.



*referring to this:
Screenshot 2025-03-18 at 2 28 22 PM

kevinansfield:
Closing this PR because the approach isn't workable (see https://github.com/TryGhost/Ghost/pull/22534#issuecomment-2732146296) and doesn't resolve the underlying data consistency problem.

PR #22577, opened 2025-03-19
Added browser tests covering the editor's basic saving behavior

cmraible:
ref https://linear.app/ghost/issue/ONC-833/support-escalation-re-issue-with-saving-articles-urgent

- We recently had a bug that caused the editor to fail to save if only the body was filled, and the title was left blank. With a blank title, the slug wasn't generating, which caused the initial `POST` request to fail with a validation error.
- We've had similar issues in the past and added some admin acceptance tests, but they didn't and wouldn't catch a case like this, where the bug is caused by an interaction between the editor, admin, and the Admin API.
- This commit adds a new browser test file for the editor and two basic test cases as a starting point. It uses Playwright's "Page Object Model" concept to make the tests easier to read and write, and to keep the test's as DRY as we can.
- It also adds a missing test in the Admin Acceptance tests, although this test wouldn't have caught this bug

### DIFF
--- a/ghost/admin/tests/acceptance/editor/lexical-test.js
+++ b/ghost/admin/tests/acceptance/editor/lexical-test.js
@@ -3,6 +3,7 @@ import {blur, click, currentURL, fillIn, find, waitFor, waitUntil} from '@ember/
 import {enableLabsFlag} from '../../helpers/labs-flag';
 import {expect} from 'chai';
 import {invalidateSession} from 'ember-simple-auth/test-support';
+import {pasteInEditor} from '../../helpers/editor';
 import {setupApplicationTest} from 'ember-mocha';
 import {setupMirage} from 'ember-cli-mirage/test-support';
 import {visit} from '../../helpers/visit';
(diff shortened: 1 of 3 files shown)


ErisDS:
Curious about this one - did the editor fail, or did the API fail to do what was expected?

Would you mind taking a moment to outline the exact failure mechanism we're trying to defend against?

cmraible:
The mechanism of the bug was:

* The ember controller (not the lexical editor itself) failed to generate a slug if the title was empty. Normally it would generate `untitled-{x}` as the slug if the title was missing
* When the body of the post (or anything in the sidebar) was changed, it would trigger a `POST /posts` request to save the draft
* The backend would (correctly) respond with a validation error, complaining that there is no slug, and the save would fail.

There's definitely a missing unit test (or 3 or 4) for the slug-generating behavior, which I'll add to this PR as well, and should be sufficient to protect against this exact failure mode in the future.

This browser/e2e test is intended as more of a smoke test / safety blanket to protect against any other bugs of this general breed, where a frontend bug can have unintended cascading failures ultimately resulting in posts not saving.

I have a todo to look into the admin test more. The way it's currently written, it doesn't catch this bug because (I think) it's using a mock API that bypasses the validation, so it just assumes the backend saves successfully — hence the browser test.

cmraible:
Taking a fresh look at this, I think maybe these browser tests are a bit heavy handed. Closing for now, to re-evaluate if this can be accomplished with admin tests.

PR #22603, opened 2025-03-24
Added the read:identity permission to Administrators

allouis:
ref https://linear.app/ghost/issue/AP-974

Right now it is only possible for Owners to access the identity token endpoint. But we want to open up ActivityPub to Administrators.


cathysarisky:
I wonder whether it makes sense for the 'super editor' role to also have this ability -- it seems like another spot where you might want a staff member to be able to do social media, without giving them full admin...

sagzy:
> I wonder whether it makes sense for the 'super editor' role to also have this ability -- it seems like another spot where you might want a staff member to be able to do social media, without giving them full admin...

@cathysarisky great shout! We're actively thinking of opening it up to more staff roles - starting with admins for now

Ghost-Slimer:
Deployed to **staging** with ID: `3294`

PR #22614, opened 2025-03-24
Rebuilt web analytics pipes

9larsons:
ref https://linear.app/ghost/issue/ANAL-96
- re-enabled tinybird CI tests needing to pass to merge
- bumped to version 4 for data files and api endpoints (and pointed admin to them)
- rebuilt data pipeline
- fixed data discrepancies, especially when filtering

This was a big one. We had a lot of data validation issues that needing resolving, which ultimately required reworking the data pipeilne. CI wasn't deploying the beta build because of bad config (and failing tests, which were also fixed, and whose disabling led to us missing how invalid the data was).

This will likely have some follow up changes to the data pipeline. The priority was to get accurate data, not worrying too much about performance. In brief:
- The hits include is now a materialized view, and all existing MVs were removed. All API endpoints now use `mv_hits`.
- KPIs should always be pulling **session** data, e.g. if you're filtering by visits to `/`, you should see the data for the entire session that includes the page hit to `/`; 10 pageviews, 47 minutes, etc. 
- Re: performance, we should introduce MVs for the session-level summary data used by the KPIs API.


9larsons:
Tinybird seems to have a bug, likely with versioned data files, when trying to delete them. It attempts the deletion of `mv_pages` but when it can't find it, it fatally errors. It may need to look for `mv_pages__v3` as we were trying to delete version 3.

PR #22618, opened 2025-03-25
🐛 Fixed incorrect slug increments for post title changes

allouis:
ref https://linear.app/ghost/issue/ONC-794

When checking slug availability without specifying the ID of the specific resource
being modified, the system can't distinguish between a new resource and an existing
one. This causes false "slug already in use" results when a client checks availability
of a slug that's already assigned to the resource being updated.

For example, if post ID "123" has slug "snowman" and the client checks if the slug
"snowman" is available (without specifying it's for post "123"), the system returns
"snowman-2" unnecessarily. This commonly happens during title edits that would
generate the same slug, like adding a trailing space.

This allows clients to ask "is the slug 'snowman' available for resource ID
'123'?" which handles the case where the resource already owns that slug.

This is a non-breaking change - the existing API continues to work at both the
HTTP and method call levels, with the modelId being optional.

By passing the post ID when generating a slug, the system can now properly
determine if the slug is already assigned to the post being edited. This
correctly handles cases like whitespace-only title changes where the same
slug should be maintained rather than being incremented unnecessarily.

### DIFF
--- a/ghost/admin/app/services/slug-generator.js
+++ b/ghost/admin/app/services/slug-generator.js
@@ -10,7 +10,7 @@ export default class SlugGeneratorService extends Service {
     @service ghostPaths;
     @service ajax;
 
-    generateSlug(slugType, textToSlugify) {
+    generateSlug(slugType, textToSlugify, modelId) {
         let url;
 
         if (!textToSlugify) {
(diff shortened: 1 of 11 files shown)


ghost/admin/app/services/slug-generator.js
```diff
@@ -10,15 +10,20 @@ export default class SlugGeneratorService extends Service {
...
             return resolve('');
         }
 
         // We already do a partial slugify at the client side to prevent issues with Pro returning a 404 page because of invalid (encoded) characters (a newline, %0A, for example)
-        url = this.get('ghostPaths.url').api('slugs', slugType, encodeURIComponent(slugify(textToSlugify)));
+        let name = encodeURIComponent(slugify(textToSlugify));
+        if (modelId) {
+            url = this.get('ghostPaths.url').api('slugs', slugType, name, modelId);
+        } else {
+            url = this.get('ghostPaths.url').api('slugs', slugType, name);
```
ronaldlangeveld, line 25: Are their reproducible scenarios where `modelId` may not exist given the changes?

ronaldlangeveld (approved):
One small comment, but overall, LGTM.

ghost/admin/app/services/slug-generator.js
allouis: Yeah, this service is used for generating slugs for tags and users too!

cmraible:
Blowing my mind a bit, I was so tunnel-visioned on the frontend that I hadn't considered this option at all 👀

allouis:
I've added a test to ensure that the id is passed to the `generateSlug` method when it exists as discussed in the Clean Code session.

PR #22641, opened 2025-03-25
Moved web analytics to traffic analytics and reset to version 0

cmraible:
closes https://linear.app/ghost/issue/ANAL-158/auto-update-token-api-permissions
closes https://linear.app/ghost/issue/ANAL-156/reset-tinybird-workspace-to-v0

This is a pretty big PR because there's really no sane way to break this up. It covers the following:
- Moves the `ghost/web-analytics` folder to `ghost/traffic-analytics`, to align with Slack/Linear and other internal verbiage
- Makes `ghost/traffic-analytics` a javascript package, so we can add JS code and/or package.json scripts for common operations
- Resets the version on all pipes and datasources to `0` so we have a cleanpoint going into the private beta wrt change management
- Adds the stats_page token to all pipes, to automatically add permissions so we don't have to do this manually after each new version is deployed
- Updates the CI jobs to use the new `ghost/traffic-analytics` directory


cmraible:
Holding off on merging this for now. Ideally we'll be able to find a way to break this up into multiple commits, because it's a big PR that feels kind of dangerous to merge right before launching the private beta.

PR #22604, opened 2025-03-24
Dockerhub

ErisDS:
I'm assuming this breaks on forks...

PR #22707, opened 2025-03-30
Added login to dockerhub to CI pipeline

ErisDS:
- We ran into an error: Unable to find image ‘mysql:8.0’ locall
docker: Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit. See ‘docker run --help’.
- It seems we hit our unauthenticated rate limit for dockerhub
- This adds a login, so that we get the authenticated rate limit which is much much higher
- It only attempts the login, if we're not on a fork PR, as that will cause the PR to error out completely because GitHub does not allow forks to access secrets

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -520,6 +520,13 @@ jobs:
         run: sudo service mysql stop
         if: matrix.env.DB == 'mysql8'
 
+      - name: Login to Docker Hub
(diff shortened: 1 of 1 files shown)


ErisDS:
This was a test in order to demonstrate **this** change doesn't break fork PRs, unlike the attempt in https://github.com/TryGhost/Ghost/pull/22604.


Closing in favour of [https://github.com/TryGhost/Ghost/pull/22708 (origin PR)]

PR #22710, opened 2025-03-31
Fixed two small bugs related to link click ordering

allouis:
ref https://linear.app/ghost/issue/ONC-845

### DIFF
--- a/ghost/admin/app/components/posts/analytics.js
+++ b/ghost/admin/app/components/posts/analytics.js
@@ -249,7 +249,11 @@ export default class Analytics extends Component {
             return acc;
         }, {});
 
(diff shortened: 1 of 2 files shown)


ghost/admin/app/components/posts/analytics.js
```diff
@@ -201,39 +201,19 @@ export default class Analytics extends Component {
     }
 
     updateLinkData(linksData) {
-        let updatedLinks;
-        if (this.links?.length) {
-            updatedLinks = this.links.map((link) => {
```
allouis, line 206: @kevinansfield Do you know of any reason why this shouldn't be removed - why we would need to update the links "in place"?

PR #22721, opened 2025-03-31
Unified members/paid-members enabled checks in Admin

kevinansfield:
no issue

- we had a confusing mix of manual underlying access setting checks, direct `settings.membersEnabled/paidMembersEnabled` checks, and `membersUtils.membersEnabled/paidMembersEnabled` checks
  - the `membersUtils.xEnabled` checks were originally in place to remove the repeated direct checks of `settings.membersSignupAccess !== 'none'` type checks that were spread around the codebase but the refactor to switch to those utils was never fully completed
  - we've later added API-level `membersEnabled` and `paidMembersEnabled` settings which `membersUtils.xEnabled` checks were switched to but we also started using the underlying settings directly in different parts of the codebase meaning we ended up with three different approaches
  - the confusion around and slightly different behaviour between these different methods meant our tests were also difficult to determine as setup helpers were doing different things that meant code paths using each different approach were not necessarily doing what was intended
- unified to always use `settings.membersEnabled` and `settings.paidMembersEnabled`
  - removed the duplicate properties from `membersUtils`
  - switched all uses of `membersUtils` and manual checks over to `settings` checks
  - updated test utils to update all impacted settings when enabling/disabling members
- fixed tests that were incorrectly testing against members disabled
- added missing `/api/admin/stats/mrr/` mocked endpoint that was causing dashboard tests to fail now that we are correctly testing against the members enabled state


ErisDS:
Cleaning up older PRs 💅

PR #22724, opened 2025-03-31
Cleaned up obsolete Tinybird pipes and datasources

cmraible:
ref https://linear.app/ghost/issue/ANAL-186/clean-up-pipes-and-data-sources-that-are-no-longer-relevant

- We currently have a lot of cruft in our Tinybird workspaces, with obsolete versions of the current live APIs, and some completely unused pipes and materialized views that have stuck around for one reason or another.
- This commit adds a custom deploy script which cleans up all these unused pipes and datasources, and also deletes the unused files from the repo so we can iterate from a clean start going forward.


9larsons (approved):
LGTM, I'll have to try the script to get a better feel. The fact it's hard-coded to v7 right now is yet another point of maintenance for bumping, though I'm not super concerned about that atm.

PR #22726, opened 2025-03-31
Fixed stats page filtering by source

9larsons:
ref https://linear.app/ghost/issue/ANAL-182
- moved all endpoints to use the same include file for filtering sessions
- all endpoints used filtered sessions instead of directly querying hits
- re-enabled source filter on frontend
- added test coverage for dual filters + snapshots of expected output from hits mv and the sessions pipe (which will be a view eventually)

Filtering by source was inconsistent across our various components, which is why we disabled it. This set of changes unifies our filtering logic across all endpoints and adds more coverage to raise a flag when changes are disruptive.


ghost/admin/app/styles/layouts/stats.css
```diff
@@ -302,14 +302,6 @@ a.gh-stats-data-label:hover {
     line-height: 1.3em;
 }
 
-.gh-stats-bar-text-nolink {
```
cmraible, line 305: lol, this is the first I saw how you disabled the source filtering 😂

9larsons: Pairs well with the cut-release-from-unreleased-branch workflow.

ghost/web-analytics/scripts/exec_test.sh
```diff
@@ -15,8 +15,8 @@ check_sum() {
     local file=$1
     local expected_count=$2
 
-    # Only perform the check if the file starts with "all_"
-    if [[ ! $(basename "$file") =~ ^all_ ]]; then
+    # Skip check if file starts with "all_" AND contains "timezone" in the name // TODO: we should handle this for all tests and in a better way
+    if [[ ! $(basename "$file") =~ ^all_ ]] || [[ $(basename "$file") =~ timezone ]]; then
```
cmraible, line 19: Nice for a temp solution 👍🏼

ghost/web-analytics/includes/filtered_sessions.incl
```diff
@@ -0,0 +1,36 @@
+NODE query_filters
+DESCRIPTION >
+    Get sessions that match the filter criteria
+
+SQL >
+    %
+    select distinct session_id
+    from mv_hits
```
cmraible, line 8: Should this be `_mv_hits`?

9larsons: Yep! Good catch, thanks. This is a good reason to move away from the current naming convention.

cmraible:
Just a heads up: I merged the [`.tinyenv` change](https://github.com/TryGhost/Ghost/pull/22726/commits/c148f4e369bbd3cfac6f86b0bb271f831fb5d4c0) in a separate PR and rebased, so that diff is gone.

I also fixed the timezone test warnings by changing the names of those files, so I also [removed the change to `exec_test`](https://github.com/TryGhost/Ghost/pull/22726/commits/11ca8d51cb11dd8301cdf55f9cc8ed9543f377fe) to keep this PR more focused

PR #22733, opened 2025-04-01
🌵 Spike: run Tinybird tests in mocha

cmraible:
ref https://linear.app/ghost/issue/ANAL-192/write-tests-in-jsmocha

- Our current Tinybird tests are written in bash — this spike attempts to adapt our existing Tinybird tests to a more convention JS/mocha setup

### DIFF
--- a/ghost/core/package.json
+++ b/ghost/core/package.json
@@ -34,6 +34,7 @@
     "test:integration": "yarn test:base './test/integration' --timeout=10000",
     "test:e2e": "yarn test:base ./test/e2e-* --timeout=15000",
     "test:regression": "yarn test:base './test/regression' --timeout=60000",
+    "test:tinybird": "yarn test:base './test/tinybird' --timeout=60000",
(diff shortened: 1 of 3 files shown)


cmraible:
Little spike here. Seems like it's achievable, although the Tinybird CLI is a bit buggy when invoking it through child processes like this. The `tb datasources append` command is totally ignoring the currently selected branch, and is just appending data to the `main` branch of the staging workspace, which obviously is not great. 

Thinking it might be better to use the Tinybird APIs directly, rather than wrapping the CLI in this way.

BUT this only took an hour or two, so with a little more time I'm pretty sure this is a workable idea.

cmraible:
Closing this PR — while this does seem doable, it currently would take a significant amount of work to wrap the Tinybird CLI fully, and then ongoing maintenance to keep up with updates to the CLI. For now at least, I'll plan to keep making iterative improvements to the bash based fixture tests, rather than fully reworking the test framework to use Mocha/JS

PR #22738, opened 2025-04-01
Unified the Endpoints Used to Fetch Account Details for Profile

vershwal:
Ref https://linear.app/ghost/issue/AP-1019/

- Removed `GET '/.ghost/activitypub/profile/:handle'` and related stuff from UI. 
- Using `GET /.ghost/activitypub/account/@username@domain'` to get profile for both the view


apps/admin-x-activitypub/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/admin-x-activitypub",
-  "version": "0.6.29",
+  "version": "0.6.33",
```
mike182uk, line 3: Why are we jumping 4 versions?

apps/admin-x-activitypub/src/api/activitypub.ts
```diff
@@ -374,12 +361,9 @@ export class ActivityPubAPI {
         return json as Thread;
     }
 
-    get accountApiUrl() {
-        return new URL(`.ghost/activitypub/account/${this.handle}`, this.apiUrl);
-    }
-
-    async getAccount(): Promise<GetAccountResponse> {
-        const json = await this.fetchJSON(this.accountApiUrl);
+    async getAccount(profileHandle: string): Promise<GetAccountResponse> {
```
mike182uk, line 364: maybe just `handle` instead of `profileHandle`?

apps/admin-x-activitypub/src/hooks/use-activity-pub-queries.ts
```diff
@@ -1152,14 +1139,20 @@ export function useNoteMutationForUser(handle: string, actorProps?: ActorPropert
...
 }
 
-export function useAccountForUser(handle: string) {
+export function useAccountForUser(handle: string, profileHandle: string) {
+    const queryClient = useQueryClient();
+
+    // Invalidate the cache when profileHandle changes
+    React.useEffect(() => {
+        queryClient.invalidateQueries({queryKey: QUERY_KEYS.account(profileHandle)});
+    }, [profileHandle, queryClient]);
```
mike182uk, line 1148: This seems 🐟 but I can't reason about why. Why do we need this?

apps/admin-x-activitypub/src/components/layout/Onboarding/Step1.tsx
```diff
@@ -8,7 +8,7 @@ import {useBrowseUsers} from '@tryghost/admin-x-framework/api/users';
 import {useNavigate} from '@tryghost/admin-x-framework';
 
 const Step1: React.FC = () => {
-    const {data: account} = useAccountForUser('index');
+    const {data: account} = useAccountForUser('index', 'index');
```
mike182uk, line 11: 🤔 I mentioned this in the backend PR, but it might be nicer to assume the current user if this param isn't passed? That way we are not hardcoding `index` for the 2nd param?

apps/admin-x-activitypub/package.json
vershwal: No that's because my editor does this automatically when I use vs code for commits. Will change before merging

mike182uk: oh, the pre-commit hook 😅

apps/admin-x-activitypub/src/hooks/use-activity-pub-queries.ts
vershwal: You were right at pointing this out. Not needed. Removed it. 

I have also updated the cache to use `QUERY-Key.account` in place of `QUERY-Key.profile`, as we have removed the hook updating `QUERY-Key.profile`

PR #22747, opened 2025-04-01
Formatted all datafiles and added scripts to make it repeatable

cmraible:
ref https://linear.app/ghost/issue/ANAL-186/clean-up-pipes-and-data-sources-that-are-no-longer-relevant

- Currently we don't have any kind of linting for our Tinybird datafiles, which can make it somewhat difficult to understand what's changing in a commit at a glance, due to extraneous formatting changes.
- This commit updates all the datafiles using `tb fmt` to format them. There shouldn't be any functional changes to the pipes or datasources, only the formatting of the datafiles themselves.
- This commit also introduces two new scripts: `format_datafiles`, which checks the syntax of the files and force formats them using `tb fmt ... --yes`, and `check_datafiles`, which just checks the syntax and formatting, without updating anything (for use in CI).


cmraible:
This is taking a lot longer than it should for not much added value — closing for now, may revisit in the future.

PR #22752, opened 2025-04-02
Fixed Tinybird endpoints with pathname and source filters

cmraible:
ref https://linear.app/ghost/issue/ANAL-194/automated-tests-for-all-queries-including-every-filter-plus-one-other

### DIFF
--- a/ghost/web-analytics/tests/filter_source_bing_top_devices.test.result
+++ b/ghost/web-analytics/tests/filter_source_bing_top_devices.test.result
@@ -1,0 +1,2 @@
+"device","visits","pageviews"
+"desktop",2,2
--- a/ghost/web-analytics/tests/filter_source_bing_top_devices.test.result
+++ b/ghost/web-analytics/tests/filter_source_bing_top_devices.test.result
@@ -1,2 +1,2 @@
 "device","visits","pageviews"
-"desktop",2,2
+"desktop",2,5
--- a/ghost/web-analytics/tests/filter_source_bing_top_locations.test.result
+++ b/ghost/web-analytics/tests/filter_source_bing_top_locations.test.result
@@ -1,0 +1,3 @@
+"location","visits","pageviews"
+"DE",1,1
+"GB",1,1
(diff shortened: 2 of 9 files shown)


ghost/web-analytics/tests/filter_source_bing_top_devices.test.result
```diff
@@ -0,0 +1,2 @@
+"device","visits","pageviews"
+"desktop",2,2
```
cmraible, line 2: I don't think this is correct 🤔 this should be 2 visits, 5 pageviews

ghost/web-analytics/tests/filter_source_bing_top_locations.test.result
```diff
@@ -0,0 +1,3 @@
+"location","visits","pageviews"
+"DE",1,1
+"GB",1,1
```
cmraible, line 3: Same here, this should be 1,3 and 1,2, respectively

ghost/web-analytics/tests/filter_source_bing_top_os.test.result
```diff
@@ -0,0 +1,2 @@
+"os","visits","pageviews"
+"windows",2,2
```
cmraible, line 2: And here, this should be 2,5

cmraible:
I think I found an issue while adding these tests for the source filter.

The results for the new tests I'm adding (top_os, top_locations, top_devices) all show only 2 pageviews, when there are actually 5 pageviews that should match the criteria. Upon checking the other `filter_source_bing` tests, I found that the existing results for top_browsers has the same problem.

I've just updated the test results for these tests to what they actually should be, so tests should fail here, and then I'll try to fix it without breaking any of the existing tests.

cmraible:
I found a similar issue while adding the `filter_pathname_about_top_sources` tests, which is currently overcounting pageviews. Rather than merging a test that I know is incorrect, I've combined all the known-incorrect tests into this PR, and manually updated the test result files with the expected results. This way we can iterate on the pipes to get this group of tests to pass, and we should be set from there 🎉

cmraible:
Pausing work on this for the moment — I think the path forward here is to remove the pageviews from all the pipes, which will resolve these failing tests and also simplify our existing pipes — pending feedback from Peter and Hannah before moving forward with that plan.

PR #22756, opened 2025-04-02
Separate Inbox and Feed UI and data

peterzimon:
ref AP-1006

- The feed and inbox list UI logic and data fetching is combined in a single component ATM, which makes it hard to reuse, e.g. in testing and developing the new routing based components. This PR separates the UI components from data fetching and merges the new route based Inbox with the old Inbox component so that it's easier to release the router updates.


allouis (approved):
Bootiful 🍑

PR #22757, opened 2025-04-02
[WIP] Added Captcha settings for self-hosters

sam-lord:
ref BAE-403

Issues with this PR currently:

* No design review, copy needs addressing for the new fields in Ghost Admin
* The change to using `hostSettings:captcha` for siteKey & secretKey needs to be actioned for Ghost (Pro)


ErisDS:
No longer needed

PR #22759, opened 2025-04-02
Send owner user email to self-service migrations

PaulAdamDavis:
no ref

- Sends the owner email to the self-service migration tool, so that we can ensure all users have a great experience of migrating


ghost/admin/app/services/migrate.js
```diff
@@ -63,6 +64,19 @@ export default class MigrateService extends Service {
         });
     }
 
+    async ownerUser() {
```
ErisDS, line 67: You have the correct pattern here, but this function already exists in the billing service `getOwnerUser()`. 

We can add `@service billing` to line 10 of `gh-migrate-iframe` and use that.

Or, you could go one better and move that function out of the billing service, into a new "owner" service that's used in both places - which could be more easily tested.

ErisDS (changes requested):
Ideally, there'd be at least one test added for this change to prove it works and we get an email address

ErisDS (approved):

PR #22765, opened 2025-04-02
Fixed dropdowns not closing on web preview

sanne-san:
No ref
- The share dropdown and the preview-segment dropdown were not closing when clicking inside the iframe of the post preview. This was fixed by adding an event listener to the iframe's document to close the dropdowns when clicking inside the iframe.


sanne-san:
Closing this in preference of a [more reusable approach](https://github.com/TryGhost/Ghost/pull/22766).

PR #22778, opened 2025-04-02
Fixed Tinybird `exec_test` script yielding false passing results in CI

cmraible:
no issue

Previously, test failures in parallel execution weren't being properly 
detected due to complex pipe handling with tee and grep, which could 
result in SIGPIPE (141) issues. This meant the CI could pass even when 
tests failed.

The change simplifies failure detection by:
- Removing complex pipe handling that caused SIGPIPE issues
- Directly capturing parallel's exit code
- Running all tests to completion to show all failures
- Properly propagating non-zero exit codes to CI

This ensures the CI job will now fail if any test fails, while still 
showing output from all test failures for easier debugging.


cmraible:
@coderabbitai pause

cmraible:
Fixed in a different PR

PR #22779, opened 2025-04-02
Fixed test failure exit code handling in Tinybird `exec_test` script

cmraible:
no issue

The `exec_test` script was exiting with code 141 when a test failed, which CI wasn't interpreting as a failure, and was therefore showing passing results. This commit aims to fix that by modifying the script to exit with code 1 if any tests fail.

The previous implementation used process substitution with tee and PIPESTATUS
to detect test failures, but this was unreliable because:
1. Process substitution runs in a subshell, so its exit code wasn't properly
   propagated to the parent shell
2. The SIGPIPE (141) errors occurred when the grep process in the subshell
   terminated before the tee process finished writing

Changed to use a temporary file instead, which:
1. Ensures all test output is captured before checking for failures
2. Avoids subshell exit code propagation issues
3. Eliminates the SIGPIPE errors by removing the process substitution
4. Makes the failure detection more reliable by checking the output file
   after all tests have completed

### DIFF
--- a/ghost/web-analytics/scripts/exec_test.sh
+++ b/ghost/web-analytics/scripts/exec_test.sh
@@ -159,11 +159,16 @@ else
     # If no test name provided, run all tests in parallel
     echo "Running tests in parallel using $jobs workers"
     # Use parallel to run the tests, maintaining output order
+    output_file=$(mktemp)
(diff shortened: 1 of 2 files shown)


cmraible:
Confirmed that CI goes red if a test fails now.

PR #22784, opened 2025-04-03
Improved the reader customizer in ActivityPub

minimaluminium:
ref AP-630

- added new background color option to the reader customizer
- removed line spacing option for simplicity
- improved the overall design of the customizer
- isolated table of contents and customizer logics into their own components


peterzimon (approved):
@minimaluminium Fixed a couple of minor things, but overall looks really good

PR #22796, opened 2025-04-03
Add support for Greek characters on search

miagg:
This will include the Greek unicode block so that we can search using Greek characters.

Screenshot 2025-04-03 at 5 20 48 PM


miagg:
Hey @cathysarisky, great news!
Usually the only normalizing needed for Greek searching is removing accents (for example one can search for 'καιρος' instead of 'καιρός'.
In your example it is not required to normalize greek letters to matching letters in other charset (for example greek Μ with greek Μ) usually when we need that kind of normalization we use a search engine such as algolia or meilisearch that adds fuzzy search as well.

PR #22820, opened 2025-04-05
Modified for custom front end website 

aidabo:
1, modify mail templates in ghost/core/core/server/service/mail for custom front end website
2. add video format .mov of iphone


aidabo:
OK

aidabo:
all right

PR #22874, opened 2025-04-13
🐛 search fix for encoding related to cjk

### LINKED ISSUE #22764: 🐛 Search should match results with accents if term does not include them
### Issue Summary

If a post title of page include accents, the search results should include those in the results, whether there are accents or not

### Steps to Reproduce

1. Create a post with the title Renée Zellweger (note the accented é)
2. Search for this using renee as the term
3. Observe no results

### Ghost Version

1.108.1

### Node.js Version

18

Ghost(Pro)

### Database type

MySQL 8

### Browser & OS version

All

```shell

```

### Code of Conduct

cathysarisky:
closes https://github.com/TryGhost/Ghost/issues/22764

removes the need for https://github.com/TryGhost/Ghost/pull/22796 (recently closed)

Please see https://github.com/TryGhost/Ghost/issues/22764 for more discussion

Heads-up:  This PR includes a bump to flexsearch, so that we can use some newer functionality with charsets.

Huge shout-out to @shincurry , part of whose approach in the discussion above I adopted for this PR.

This approach allows using mixed CJK and other characters. It uses the current 'Charset.Default' encoder used by Flexsearch, amended with code that runs during the 'finalize' step to split CJK characters.

Because we retain Charset.Default, we get the built-in handling of accents.

(Details above the process here: https://github.com/nextapps-de/flexsearch/blob/master/doc/encoder.md#encoder-processing-workflow )

The "Charset.CJK" offered by Flexsearch is not suitable, as it splits /all/ characters.

This PR switches to detecting CJK character, rather than trying to enumerate and detect all non-CJK characters. It should be easier to maintain, going forward.

Switched to using codepoints instead of regex because it was 100-200ms faster to load search on a test site with 3000 posts.


shincurry:
**isCJK** function, if a larger range of codepoints is put first, can it effectively improve performance? In other words, commonly used characters range first. 🤔

cathysarisky:
> If a larger range of codepoints is put first, can it effectively improve performance? In other words, commonly used characters range first. 🤔

I would have guessed that we had the three common ones first, and then the less common/supplements later, but I don't know CJK languages well enough.  

> Codepoint ranges are ordered, can binary search be applied? Although the total number of ranges is small.

That's an interesting question.  But maybe we can get this fix (that is not noticeably slower than current, fixes some languages being unsearchable, and restores behavior for languages with accents that should be normalized) through, and then follow it with performance tweaks in a second PR?

shincurry:
Yes, this PR should be merged as soon as possible.
I just have some ideas, write down to see your opinions.

shincurry:
> That's an interesting question. But maybe we can get this fix (that is not noticeably slower than current, fixes some languages being unsearchable, and restores behavior for languages with accents that should be normalized) through, and then follow it with performance tweaks in a second PR?

Yes, this PR should be merged as soon as possible.
I just have some ideas, write down to see your opinions.

PR #22875, opened 2025-04-13
Add lang stemming when available

cathysarisky:
no issue

This is a continuation on https://github.com/TryGhost/Ghost/pull/22874 's update to flexsearch 0.8, which adds some new functionality!

This PR adds using the locale (if en, fr, or de) to provide language-specific stemming.  This should provide some (modest) improvement in search quality for those languages.

**Example**:
Data:
Post title 1 contains:  "This informational meeting must not be missed"
Post title 2 contains:  "This information is important"

Current:  
Searching 'informational' returns title 1 only.  Searching 'information' returns both titles.

New: 
'en' locale:  searching 'informational' returns both titles.  (-al is in the list of suffixes removed in English)
'de' locale:  no change - only post title 1 returned.   (-al is not in the list of suffixes removed in German)

Aside:  These are the only three languages with support built into Flexsearch, but we could add additional ones if someone wants to write one and contribute it!

### Request for help
I am stuck on getting these language imports to work on CI, which is apparently node version specific.  (See https://github.com/nextapps-de/flexsearch/issues/461 .)  Since the CI runs an older version of node, I switched over to using a require for the language presets.  Which causes CI to pass, but makes the search not actually work!  So... I could use a hint here.   I am not as expert with vite or CI as maybe I need to be here...   and the ugly workaround of switching between import and require depending on environment wasn't straightforward, as I'd need to switch a whole bunch of stuff over to async.

My current thought (but I'm suspecting this is a huge code smell) would be to just use vite to copy the three files needed out of node_modules and into the search package.  So clearly, I need a better idea.


9larsons:
Sorry for the delay on this. I'm not certain there's anything (clean) we can do besides update our Node version. I'll see where we're at with 20.19 as that's the v20 LTS.

PR #22887, opened 2025-04-14
Updated user modal layout to single column

sanne-san:
Ref https://linear.app/ghost/issue/PLG-402/switch-from-2-column-to-1-column-layout
- The current modal layout is not scalable in that it can handle the number of new input fields without becoming off balance. Rather than shuffling fields around to make it fit, a single column layout would be cleaner, simpler, and more scalable.


Ghost-Slimer:
Deployed to **staging** with ID: `3332`

PR #22912, opened 2025-04-16
🐛 Fixed if all comments are hidden, no comments are visible

### LINKED ISSUE #22331: if all comments are hidden, no comments (not even hidden ones) are visible
### Issue Summary

If all comments have been hidden by a site admin, the hidden comments do not appear.  Making a new visible comment restores access to the hidden comments.

### Steps to Reproduce

Make some comments
As a site admin, hide all the comments on a post.
Reload.  Observe that the hidden comments are not displayed to the admin user.

### Ghost Version

5.110.2

### Node.js Version

20

local dev

### Database type

SQLite3

### Browser & OS version

windows 11, chrome

```shell

```

### Code of Conduct

djskinner:
closes https://github.com/TryGhost/Ghost/issues/22331

load the auth frame even if there are no visible comments because we can't check for hidden comments without having the auth frame loaded

The fix is based on the suggestion in https://github.com/TryGhost/Ghost/issues/22331#issuecomment-2710516576

Change covered with automated tests and confirmed manually in dev environment.

### DIFF
--- a/apps/comments-ui/src/App.tsx
+++ b/apps/comments-ui/src/App.tsx
@@ -224,7 +224,7 @@ const App: React.FC<AppProps> = ({scriptTag}) => {
             <CommentsFrame ref={iframeRef}>
                 <ContentBox done={done} />
             </CommentsFrame>
-            {state.comments.length > 0 ? <AuthFrame adminUrl={options.adminUrl} onLoad={initAdminAuth}/> : null}
+            {state.initStatus === 'success' && <AuthFrame adminUrl={options.adminUrl} onLoad={initAdminAuth}/>}
             <PopupBox />
         </AppContext.Provider>
     );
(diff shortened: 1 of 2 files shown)


apps/comments-ui/src/App.tsx
```diff
@@ -224,7 +224,7 @@ const App: React.FC<AppProps> = ({scriptTag}) => {
             <CommentsFrame ref={iframeRef}>
                 <ContentBox done={done} />
             </CommentsFrame>
-            {state.comments.length > 0 ? <AuthFrame adminUrl={options.adminUrl} onLoad={initAdminAuth}/> : null}
+            {state.initStatus === 'success' && <AuthFrame adminUrl={options.adminUrl} onLoad={initAdminAuth}/>}
```
djskinner, line 227: I first tried always rendering `<AuthFrame` but that lead to a stuck loading state. Perhaps a more involved change is required to avoid this conditional render entirely.

kaueburiti, line 227: Maybe you could reuse the `done` constant to avoid logic duplication, what do you think?

djskinner: 👍  thanks for the suggestion, I've made that change

9larsons:
Closing in lieu of #22915 since there's so much overlap. Thanks for taking the time on this - will ref you on the other PR.

PR #22915, opened 2025-04-16
🐛 Show hidden comments to admin when there are just hidden comments

### LINKED ISSUE #22331: if all comments are hidden, no comments (not even hidden ones) are visible

kaueburiti:
## Context
When the admin hides all comments, it becomes impossible to see the hidden comments, so you can't unhide or even delete them.

fixes https://github.com/TryGhost/Ghost/issues/22331

## The Solution
As described [here](https://github.com/TryGhost/Ghost/issues/22331#issuecomment-2710516576) by @cathysarisky, the major problem was the AuthFrame not being loaded.

But things get a bit complex when you notice that even after loading it, hidden comments still don't show. We got stuck in the loading state. The reason is that because of the way we update the states inside `initSetup`, we can never change our `done` to success, leaving us in an eternal loading state.

So, besides loading the AuthFrame, the major work was handling the state update on `initSetup` to avoid unnecessary re-renders.

## Screenshots
| Admin Perspective | Member Perspective |
| --- | --- |
CleanShot 2025-04-16 at 07 18 26@2x | CleanShot 2025-04-16 at 07 19 23@2x |

### DIFF
--- a/apps/comments-ui/package.json
+++ b/apps/comments-ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/comments-ui",
-  "version": "1.5.5",
+  "version": "1.5.6",
   "license": "MIT",
   "repository": "https://github.com/TryGhost/Ghost",
   "author": "Ghost Foundation",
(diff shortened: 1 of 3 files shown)


9larsons:
@kaueburiti Thanks for this, and sorry for the delay!

PR #22968, opened 2025-04-21
Added X property to author api endpoint

ronaldlangeveld:
ref https://linear.app/ghost/issue/PLG-415/support-using-x-in-place-of-twitter

- Using the serializer, I mapped an X property to have the same value as twitter in order to expose it so that it can be used as a condition (if statement) inside templates using handlebars, eg `{{#if x}}`.
- This change is reliant on https://github.com/TryGhost/SDK/pull/512


ghost/core/core/server/api/endpoints/utils/serializers/output/utils/clean.js
```diff
@@ -68,6 +68,9 @@ const author = (attrs, frame) => {
         if (attrs.location === '') {
             attrs.location = null;
         }
+
+        // if a twitter username is provided, use it as the x username to ensure backwards compatibility
+        attrs.x = attrs.twitter || null;
```
ErisDS, line 73: Although this pattern is not the greatest, the intention of the `clean.js` utility is for cleaning & removing attributes. There is an `extra-attrs.js` utility intended for adding attributes.

ghost/core/test/unit/frontend/helpers/social_url.test.js
```diff
@@ -1,11 +1,11 @@
-const should = require('should');
 const handlebars = require('../../../../core/frontend/services/theme-engine/engine').handlebars;
 const helpers = require('../../../../core/frontend/services/helpers');
 const social_url = require('../../../../core/frontend/helpers/social_url');
 
 const socialData = {
     facebook: 'testuser-fb',
     twitter: 'testuser-tw',
+    x: 'testuser-tw',
```
ErisDS, line 8: Whilst adding the new property here is correct with the new shape of the data, supporting x in social_urls will require additional changes to the helper (you'll need to map x to twitter) - I _think_ it makes sense to do those updates separately and focus in this PR on getting the API changes complete.

ErisDS (changes requested):

ghost/core/test/unit/frontend/helpers/social_url.test.js
ronaldlangeveld: I'll add these changes (tests) in a separate PR.

However, we can avoid making any changes to the `social_urls` template helper by creating an alias in the `@tryghost/social-urls` library in the SDK doc as I've done in this [PR](https://github.com/TryGhost/SDK/pull/512/files).

All that was needed for it work was to have `x` as a property in the user / author api and having `@tryghost/social-urls` export a function called `x`.

ghost/core/core/frontend/services/theme-engine/middleware/update-global-template-options.js
```diff
@@ -16,6 +15,13 @@ function getSiteData() {
         siteData.signup_url = '#/portal';
     }
 
+    // theme only property added to @site
```
ErisDS, line 18: To me this is a discrete change again. I'd make it in a separate PR. Probably first as it's the simplest.

PR #23011, opened 2025-04-24
Changed ghost-stats script to use site_uuid

cmraible:
WIP


cmraible:
Punting on this til next week

PR #23071, opened 2025-04-28
Added a `post_uuid` filter for all Tinybird endpoints

cmraible:
ref https://linear.app/ghost/issue/PROD-1541/add-a-filter-by-post-uuid-to-all-our-pipes

The events that we're sending from the tracker script pass along the full `pathname` of the pageview. However, in the Admin views where we pull the data, we're filtering on the `pathname` using `pathname=/{slug}/`. This isn't always the actual path to the post on the site's frontend, depending on their routing setup, so in some cases this returns the incorrect data.

We do store the `post_uuid` for any pageviews, so we should filter by that parameter instead of the `pathname`. Right now we don't have those filters built out in the Tinybird pipes, which is what this commit adds.


ghost/web-analytics/pipes/api_kpis.pipe
```diff
@@ -152,9 +153,9 @@ SQL >
...
                 a.date as date,
                 coalesce(b.visits, 0) as visits,
-                {% if defined(pathname) %}coalesce(c.pageviews, 0){% else %}coalesce(b.pageviews, 0){% end %} as pageviews,
+                {% if defined(pathname) or defined(post_uuid) %}coalesce(c.pageviews, 0){% else %}coalesce(b.pageviews, 0){% end %} as pageviews,
                 coalesce(b.bounce_rate, 0) as bounce_rate,
                 coalesce(b.avg_session_sec, 0) as avg_session_sec
             from timeseries a
             left join data b on a.date = b.date
-            {% if defined(pathname) %}left join pathname_pageviews c on a.date = c.date{% end %}
+            {% if defined(pathname) or defined(post_uuid) %}left join pathname_pageviews c on a.date = c.date{% end %}
```
cmraible, line 161: This feels kind of weird still, but I don't think adding the `post_uuid` filter makes it any worse?

PR #23078, opened 2025-04-29
Added docker compose service for shade's storybook server

cmraible:
no issue

There's currently no good way to run the shade design system in docker compose. Running it manually within a shell both doesn't work and also feels clumsy, given that this is exactly the type of thing docker compose is built for.

This commit does two things to fix this:
1. Updates the host that the storybook dev server runs on, so the service will be accessible from outside the container
2. Creates a new service for shade's storybook server, behind the `storybook` profile

The syntax in the `compose.yml` file looks scary, but it's actually a built-in YAML feature called "anchors and aliases". Here's what's going on:
- The `x-ghost-base` block is ignored by docker compose because of the leading `x`
- The `&ghost-base` creates an anchor and assigns it the `ghost-base` alias. 
- Any service that uses the same base configuration uses `<<: *ghost-base`, which applies the `ghost-base` template and merges in any configuration that follows.

At first glance, this feels like a lot of additional complexity in the `compose.yml` file for not a lot of gain. However, this creates a reusable pattern for breaking apart the `dev.js` script and starting to run each of the subprocesses it manages in separate docker compose services. We can expand this basic pattern to run admin, portal, comments, etc. all in their own services, but using the same base image definition. 

A few other things worth noting:
- This commit adds `image: ghost-dev:latest` to all services that use the same template, to prevent docker compose from trying to rebuild the same container for each service. The `build` block is only used once on the `ghost` service itself to describe how to build the `ghost-dev` image, and other services will reuse the same image, rather than building it from scratch. 
- the `depends_on` blocks were deliberately left out of the `&ghost-base` template because some services may not actually require the database or redis to be healthy before starting.


cmraible:
Closing in favor of https://github.com/TryGhost/Ghost/pull/23114, which includes this change in it.

PR #23098, opened 2025-04-29
Updated stats content to use names instead of paths

9larsons:
ref https://linear.app/ghost/issue/PROD-12/
- created stats api endpoint in admin-x-framework to be used by stats views
- added new /top-content endpoint that pulls data from tinybird and ghost
- added tinybird client to stats service for fetching tinybird data
- added names of content to the content block in the stats>web view
- moved config use to backend from frontend; reduced query parameters passed/exposed in request

UX can be improved by displaying the content name (post title, page name, author's name from author page, 'Home' page) in the Content block in the Web tab of Stats. This provided some new challenges that we'll need to make more use of, specifically meshing Tinybird data with Ghost data.

The Tinybird pipes/endpoints were updated previously to expose the post_uuid for the top_pages endpoint. This is now requested by Ghost (the backend) instead of the frontend, where the post_uuids are looked up in the database for the post title. Where we don't have a post_uuid, we use the resource lookup provided by the urlService.


ghost/core/core/server/api/endpoints/stats.js
```diff
@@ -103,6 +103,33 @@ const controller = {
...
+            cacheInvalidate: false
+        },
+        options: [
+            'siteUuid',
+            'dateFrom',
+            'dateTo',
+            'timezone',
+            'memberStatus',
+            'tbVersion'
+        ],
```
cmraible, line 118: Just a heads up, I think Pro blocks some or all of the query params

ghost/core/core/server/services/stats/TopPagesStatsService.js
```diff
@@ -0,0 +1,207 @@
...
+     * @param {Object} options
+     * @param {string} [options.siteUuid] - The site UUID 
+     * @param {string} [options.dateFrom] - Start date in YYYY-MM-DD format
+     * @param {string} [options.dateTo] - End date in YYYY-MM-DD format
+     * @param {string} [options.timezone] - Timezone for the query
+     * @param {string} [options.memberStatus] - Member status filter
+     * @param {string} [options.tbVersion] - Tinybird version for API URL
+     * @returns {Promise<Object>} The enriched top pages data
+     */
+    async getTopPages(options = {}) {
```
cmraible, line 23: Can we break this into smaller pieces and unit test it?

cmraible:
How does this endpoint behave if Traffic Analytics isn't enabled?

ghost/core/core/server/services/stats/TopPagesStatsService.js
```diff
@@ -0,0 +1,207 @@
...
+        // Use the correct request module
+        const externalRequest = require('../../lib/request-external');
+        
+        const statsConfig = config.get('tinybird:stats');
+        const localEnabled = statsConfig?.local?.enabled ?? false;
+        const endpoint = localEnabled ? statsConfig.local.endpoint : statsConfig.endpoint;
+        const token = localEnabled ? statsConfig.local.token : statsConfig.token;
+
+        // Use tbVersion if provided for constructing the URL
+        const pipeUrl = (options.tbVersion && !localEnabled) ?
```
cmraible, line 36: Not a blocker, this works for this PR, but eventually I think we'll want a dedicated service for fetching data from Tinybird to keep all this logic and URLs in one place

apps/stats/src/views/Stats/Web.tsx
```diff
@@ -1,17 +1,103 @@
...
+    site_uuid: string;
+    date_from: string;
+    date_to: string;
+    timezone: string;
+    member_status: string;
+    tb_version?: number;
+}
+
+// Custom hook to fetch data from the Ghost API instead of Tinybird directly
+const useGhostTopPagesQuery = (params: TopPageQueryParams) => {
```
cmraible, line 33: Feels like getting data from the Admin API should have a dedicated service/hook, no? A little surprised we don't have that yet 🤔 (not a blocker, was just thinking this through for my own change and turned to this PR as a reference)

ghost/core/core/server/api/endpoints/stats.js
ErisDS: Defo double check, but I believe the API allows any query params, however we have a standard of using snake_case for all names in the API, including query params.

9larsons: Noted, I'll add that to the changes I've got staged. I've got a very significant refactor of the above + realized I didn't push the urlUtils changes.

ghost/core/core/server/services/stats/TopPagesStatsService.js
9larsons: Yep, on it 👍

PR #23103, opened 2025-04-30
Added nx reset to docker entrypoint

cmraible:
no refs

The NX Daemon gets into a bad state fairly often when running in docker compose, causing annoying errors. The immediate solution to this is to run `yarn nx reset` to restart the daemon and clear the nx cache.

Luckily this command is pretty quick to run, so until we're able to fully fix the NX Daemon issues, this should hopefully reduce the error rate in the meantime.


ErisDS (approved):
Merge it!

PR #23107, opened 2025-04-30
Added delay before purging jsDelivr cache

sagzy:
ref https://linear.app/ghost/issue/PROD-1573

- last week, purging jsDelivr cache after publishing a new NPM version of admin-x-activitypub was unreliable:
   - we released a new version (0.6.68) of our JS package, admin-x-activitypub, last Thursday 24/04 around ~17:45 UTC. The cache was purged immediately after and Mike could see the latest version on his production site 
   - but, on the next day, Mike opened his production site again and was served a previous version. Same browser / same location (no VPN)

- release process:
    1. We bump the version of a package in a commit
    2. Once the commit is merged to our main branch, CI checks whether a new version has been created, and if yes, publishes a new version to npm
    3. Once a new version has been published, CI immediately clears jsDelivr cache

- maintainer of jsDelivr suggested adding one minute delay to avoid this edge case: npm registry takes at least a few seconds to start reporting the new version after it's published and jsDelivr caches the version list for 30 seconds

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1332,7 +1332,13 @@ jobs:
         if: steps.version_check.outputs.version_changed == 'true'
         run: echo "${{ steps.cdn_paths.outputs.cdn_paths }}"
 
-      - name: Purge jsdelivr cache
+      - name: Wait before purging jsDelivr cache
+        if: steps.version_check.outputs.version_changed == 'true'
+        run: |
+          echo "Purging jsDelivr cache immediately after publishing a new version on NPM is unreliable. Waiting 1 minute before purging cache..."
+          sleep 60
+
+      - name: Purge jsDelivr cache
         if: steps.version_check.outputs.version_changed == 'true'
         uses: gacts/purge-jsdelivr-cache@v1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1333,7 +1333,7 @@ jobs:
         run: echo "${{ steps.cdn_paths.outputs.cdn_paths }}"
 
       - name: Wait before purging jsDelivr cache
-        if: steps.version_check.outputs.version_changed == 'true'
+        if: steps.version_check.outputs.version_changed == 'true' && matrix.package_name == '@tryghost/admin-x-activitypub'
         run: |
           echo "Purging jsDelivr cache immediately after publishing a new version on NPM is unreliable. Waiting 1 minute before purging cache..."
           sleep 60


ErisDS (changes requested):
Adding a minute to the build time really isn't ideal. Feels a bit off to do that to fix something that happened once in all the time we've used jsDelivr.

Would you mind explaining why we need to purge the jsDelivr cache?

ErisDS:
From reading the issue, maybe it's happened more than once? Sorry if I missed that!

sagzy:
@ErisDS Two occurrences that we've noticed so far, but in general it's about being sure that a change we expect to be in production is really in production.

Use case for ActivityPub: we ship changes and communicate them in the weekly newsletter; by the time the newsletter is sent, we'd like to be sure the change is in production for all users.

Our process: ship a new version of the AP frontend, then purge jsDelivr cache to make sure all users are being served the latest version (in theory)

What's happening sometimes (2 times so far): we _think_ that we've shipped a new version to all users, but jsDelivr is still serving an old version; making our release pipeline unreliable/unpredictable.

—

**This change should not add one minute delay to all CI builds**: the delay is only added to commits on main that contains a version bump for either ActivityPub frontend, Portal, Sodo Search or Comments. All other CI runs should be unaffected by the change.

Is that more acceptable?

ErisDS:
Yeah I realise it's only added to CI runs that affect the apps, but it still feels like the wrong fix to me.

I apologise if I'm being dumb, but I don't understand why the jsDelivr cache needs purging at all - so I'm trying to see if there's a way to delete that requirement by fixing something else?

ErisDS:
FWIW short term, I think you can add `&& matrix.package_name == '@tryghost/admin-x-activitypub'` to the step that waits 1min, so that you can see if this actually fixes the issue you're seeing in activitypub?

Then we can take our time to figure out what the right solution is

sagzy:
@ErisDS We load `admin-x-activitypub` inside Ghost Admin by requesting `https://cdn.jsdelivr.net/ghost/admin-x-activitypub@0/dist/admin-x-activitypub.js` (note the `@0` version range, not a specific pinned version).

When a new version of `admin-x-activitypub` is published on 0.x.x, we can expect it to be served to users instantly given that it's within the version range requested in the code. This strategy helps us ship changes in `admin-x-activitypub` to production without having to wait for a Ghost rollout.

Now by default, jsDelivr caches assets for up to 7 days. So unless we actively ask jsDelivr to purge their cache, they won't be serving the latest version instantly to all production users.

sagzy:
@ErisDS I have restricted the wait step for the `admin-x-activitypub` package as suggested - do you feel good about moving forward with the test with that restriction in place?

sagzy:
> @ErisDS I have restricted the wait step for the `admin-x-activitypub` package as suggested - do you feel good about moving forward with the test with that restriction in place?

Oups missed that you're off this week! Based on your previous message, I will assume you're in board with the test as long as it's bounded to `admin-x-activitypub` only — if I misread, then will revert the change once you're back

sagzy:
@ErisDS Ah actually can't merge as PR is marked as "changes requested". Will wait until you're back :)

ErisDS (approved):
By all means merge this change - sorry that requiring changes meant you couldn't!!

I think the next step here is to have a think about these requirements:

The lack of proper versioning results in the need to purge the asset in jsDelivr, but CDNs like jsDelivr aren't really designed to be purged... the fact they offer this to us is above-and-beyond IMO.

So the question is - what is a change that we could make to simplify this and remove the need to purge the cache? 

We don't need an answer right away, but I'm curious if we can think of a better way.

sagzy:
> By all means merge this change - sorry that requiring changes meant you couldn't!!
> 
> I think the next step here is to have a think about these requirements:
> 
> The lack of proper versioning results in the need to purge the asset in jsDelivr, but CDNs like jsDelivr aren't really designed to be purged... the fact they offer this to us is above-and-beyond IMO.
> 
> So the question is - what is a change that we could make to simplify this and remove the need to purge the cache?
> 
> We don't need an answer right away, but I'm curious if we can think of a better way.

Once we achieve daily / more frequent Ghost rollouts, I think we could pin versions / don't need purge caches anymore. This is currently mostly a workaround to not wait for a week before frontend changes to production - once we ship more frequently, I can imagine that we get rid of that patch release + purge cache hack?

PR #23111, opened 2025-04-30
Improved the design of notifications in ActivityPub

minimaluminium:
ref PROD-708

- added user avatars with contextual icons on singular notifications to help users 
  quickly identify both the person and type of interaction
- implemented larger icons for grouped notifications to improve scanability 
  when multiple notifications are bundled together
- emphasized notification content (such as reply text) to focus user attention 
  on the most relevant information rather than secondary elements
- created visual distinction between posts and notes to help users understand 
  different content types
- added timestamps to show when each notification occurred


apps/admin-x-activitypub/src/views/Notifications/Notifications.tsx
```diff
@@ -303,36 +296,49 @@ const Notifications: React.FC = () => {
...
+                                                <div className='text-gray-700 dark:text-gray-600'>
                                                     {isLoading ?
                                                         <>
                                                             <Skeleton />
                                                             <Skeleton className='w-full max-w-60' />
                                                         </> :
-                                                        <NotificationGroupDescription group={group} />
+                                                        <div className='flex items-center gap-1'>
+                                                            <span><NotificationGroupDescription group={group} /></span>
+                                                            <span className='mt-px text-[8px]'>•</span>
```
peterzimon, line 309: I would probably use a unicode character or icon instead of "•"

PR #23121, opened 2025-04-30
🐛Fixed mis-encoded apostrophes in emailed paywall

cathysarisky:
When I wrapped the emailed paywall a few weeks ago, I missed a spot that needed {{{ }}} to preserve the encoding of apostophes.  This two-character PR fixes that.


ErisDS (approved):
Would obviously be great to do some testing here, but I have no idea how big the lift is.

I think you have more context for the 2 char change than anyone, and I'm happy for you to make and merge fixes to changes you made  :)

PR #23147, opened 2025-05-01
Added growth data in post analytics

9larsons:


cmraible:
Closing in favor of https://github.com/TryGhost/Ghost/pull/23149, which is basically the same thing, just at a new endpoint

PR #23177, opened 2025-05-05
Allowed cors for comments dev

9larsons:
no ref

This should avoid cors errors when doing dev locally using `yarn dev --comments` in Ghost.


cathysarisky:
THANK YOU!  That was annoying me last week, but not enough to open an issue on it.

PR #23182, opened 2025-05-06
🐛 Fixed CTA for public preview card not showing on post previews

### LINKED ISSUE #23181: Content CTA does not display on preview when it should
### Issue Summary

Previewing a post that has a preview line with the previewer set to 'free member' and the post set to either members only or paying members only results in the text being correctly truncated, but no content CTA being shown.

Upon publication, the content CTA correctly appears.

The problem is present in both source and a custom theme, so does not appear to be theme-related.

### Steps to Reproduce

Create a post draft, including a paywall line.
Set the post draft to members only.
Click preview. (Web, not email.)

### Ghost Version

5.119.1 & recent

### Node.js Version

20

local install & ghost pro both

### Database type

MySQL 8

### Browser & OS version

not relevant

```shell

```

### Code of Conduct

cathysarisky:
closes https://github.com/TryGhost/Ghost/issues/23181
closes https://linear.app/ghost/issue/ONC-921/oss-issue-content-cta-does-not-display-on-preview-when-it-should
refs https://linear.app/ghost/issue/ONC-930/support-escalation-re-freepaywalls-on-web-versions-of-email-only-posts

When a post is set to members only or paid only and there is a public preview card present, the "This post is for members only" CTA wasn't appearing in the post preview, even though the private content was correctly omitted. The CTA was correctly being displayed on the actual post on the frontend, however.

This pattern of setting `post.access` in the frontend controller route was previously used in all of the `entry`, `preview` and `email` routers. This [commit](https://github.com/TryGhost/Ghost/commit/fa91c6c9545a9107ba9b938f545c674ff70a462c#diff-42f164398878ef21b6e9ab8e68f273a167fe1f5e85f4b3fe0036d92ab8128e94) from 2020 switches this pattern to calculate the value of `post.access` in the content API rather than the frontend, and this same `post.access = !!post.html` line was removed from the `entry` controller at that time. However, the change wasn't carried over to the `email` or `preview` controllers.

This commit carries the pattern over to the `preview` controller, and adds a test to ensure the CTA is rendered in the preview route. I'll follow up with another PR to do the same for the `email` controller.

### DIFF
--- a/ghost/core/core/frontend/services/routing/controllers/email-post.js
+++ b/ghost/core/core/frontend/services/routing/controllers/email-post.js
@@ -51,7 +51,7 @@ module.exports = function emailPostController(req, res, next) {
                 return urlUtils.redirect301(res, routerManager.getUrlByResourceId(post.id, {withSubdirectory: true}));
             }
 
-            // post.access = !!post.html;
+            post.access = !!post.html;
 
             return renderer.renderEntry(req, res)(post);
         })
--- a/ghost/core/core/frontend/services/routing/controllers/email-post.js
+++ b/ghost/core/core/frontend/services/routing/controllers/email-post.js
@@ -51,7 +51,7 @@ module.exports = function emailPostController(req, res, next) {
                 return urlUtils.redirect301(res, routerManager.getUrlByResourceId(post.id, {withSubdirectory: true}));
             }
 
-            post.access = !!post.html;
+            // post.access = !!post.html;
 
             return renderer.renderEntry(req, res)(post);
         })
--- a/ghost/core/core/frontend/services/routing/controllers/previews.js
+++ b/ghost/core/core/frontend/services/routing/controllers/previews.js
@@ -59,13 +59,6 @@ module.exports = function previewController(req, res, next) {
                 return urlUtils.redirect301(res, urlUtils.urlJoin('/email', post.uuid, '/'));
             }
 
-            // Set access based on member status and post visibility
-            const member = req.query?.member_status && req.query.member_status !== 'anonymous' 
-                ? {status: req.query.member_status} 
-                : null;
-            // we literally don't even need this, because we get post.access from the API anyway.
-            //post.access = checkPostAccess(post, member);
-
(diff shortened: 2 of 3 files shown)


cmraible:
@cathysarisky thanks for raising the issue, and for taking the time to work on a fix!

I think you've found the correct root cause here — `post.access` effectively always being set to true in the preview route — but I don't think this solution is _quite_ right.

Admittedly, this really isn't well documented, but the first signal to me that something's not quite right is that this PR requires code from `core/server` within `core/frontend`. You'll definitely see other instances of this in the codebase as we haven't ever fully decoupled the two, but our _intent_ is for the frontend code to be decoupled from the server code as much as possible. This is probably best documented in the `bridge.js` file: https://github.com/TryGhost/Ghost/blob/8b72ea4b341990b09ebc7a54aa9cea47cb86f7d4/ghost/core/core/bridge.js#L1-L11, although that's admittedly not super discoverable unless you're actively looking for it.

Our pie-in-the-sky long term vision is that the frontend might one day be a completely separate app, which would be a client of the server/content API. If there were no other way to do this, I wouldn't block this PR on it, but it is an initial indication to me that something's not quite right.

We also have an informal policy of adding tests for any bug fixes — I totally trust you that this fixes the bug, but without a test it's really easy for someone to come along in the future, remove this code and not realize they've broken anything.

I'm context loaded on this problem, as we've had a customer raise a similar issue (basically the same thing, but in the `/email/` route). I'd like to get a fix merged for this today so we can be sure it gets into this week's release. Let me know if you have the capacity and desire to collab on this one today, I'd be happy to work through it with you. Otherwise I'm happy to take this one off your plate if that's better for you — just didn't want to steal your thunder 😀

cmraible (changes requested):
Two things I'd like to see here before merging, more details in a separate comment above:
1. We should have at least one test for this
2. If possible, we should avoid requiring code in `core/server` from `core/frontend`

cmraible:
Just converting to a draft so Coderabbit calms down

cmraible (approved):

PR #23277, opened 2025-05-09
Added NodeJS version to GHA cache keys

cmraible:
no refs

TL;DR: we run some of our test suites against multiple versions of Node, but they are currently sharing the same caches. Dependencies that are built for one version may not work for the other version, causing flaky behavior in CI. This commit adds the NodeJS version to the cache keys that we use in CI, so that each version will maintain its own cache. This should hopefully reduce flakiness in our CI workflows.

Here's a recent error from comments-ui unit tests that failed this morning:

```
#
  # Fatal error in , line 0
  # unreachable code
  #
  #
  #
  #FailureMessage Object: 0x7ffef9597580
 ```

It's followed by a huge unintelligible (to me) native stack trace, but I'm told that "This is almost certainly caused by loading precompiled code (e.g. cached .node modules or bytecode caches) that were compiled with a different version of Node.js or V8." 

This seems to track, as clearing the caches is our go-to for fixing these flaky issues we've been seeing..

### DIFF
--- a/.github/actions/restore-cache/action.yml
+++ b/.github/actions/restore-cache/action.yml
@@ -18,8 +18,14 @@ runs:
         path: ${{ env.CACHED_BUILD_PATHS }}
         key: ${{ github.sha }}
 
-    - name: Check if caches are restored
-      uses: actions/github-script@v6
+    - name: Warn if caches are not restored
+      shell: bash
       if: steps.dep-cache.outputs.cache-hit != 'true' || steps.build-cache.outputs.cache-hit != 'true'
-      with:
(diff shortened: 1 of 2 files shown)


daniellockyer:
I'd be surprised if this helped - Node.js provides the [Node-API](https://nodejs.org/api/n-api.html) for binary dependencies and this is stable across different Node versions. So binaries built for one Node-API can be run on any Node.js version that supports that Node-API version

Def strange that we're seeing an increase in this recently

Not a blocking comment but just my 2c

cmraible:
Ah okay that makes sense. I still wonder if the NX cache accounts for all that when caching build outputs 🤔

PR #23280, opened 2025-05-12
Added `newsletters.button_corners` column migration

kevinansfield:
ref https://linear.app/ghost/issue/PROD-1688

- updated schema and added migration for new `button_corners` setting on the `newsletters` table
  - validation ensures we only support the three known corner types: `square`, `rounded`, and `pill`
  - `rounded` is set as the default to match the button style used in our pre-customisation template
- updated prototype newsletter design settings to avoid errors as this is already tied to the API
- updated prototype newsletter preview to match the finalised setting value naming


lsinger (approved):
LGTM! 🚢

PR #23301, opened 2025-05-12
Added alpha newsletter design screen with button corners setting

kevinansfield:
ref https://linear.app/ghost/issue/PROD-1688

- added duplicate of the newsletter detail modal so we can move settings over from the prototype modal once they are ready
- copied prototype button content from the prototype preview content to the non-prototype preview content to show impact of button corners setting
- updated `EmailRenderer` to add `buttonBorderRadius` to the data passed through to the handlebars template
- updated handlebars template styles partial to use `buttonBorderRadius` in all button related styles
- we had some inconsistent button radius' with some at 5px and others at 6px. When the `emailCustomizationAlpha` flag is enabled we now standardize on 6px


kevinansfield:
Closing as this was merged in https://github.com/TryGhost/Ghost/pull/23307

PR #23322, opened 2025-05-13
Added `newsletters.button_style` column

kevinansfield:
ref https://linear.app/ghost/issue/PROD-1717

- added `newsletters.button_style` column ready for controlling button styles in emails, with two options of `fill` (default, current production design) and `outline` which is a new option


lsinger (commented):
I was able to check most of the items from the checklist. The testing ones I think we're discussing on Slack. 

Not sure about the filename for the migration. I didn't find specific guidelines, but judging from what we've done in the past, for adding a column we've often used the format `YYYY-MM-dd-HH-mm-ss-add-column-name-to-table-name.js`, so this migration should probably be called `2025-05-13-17-36-56-add-button-style-to-newsletters.js`. What do you think @kevinansfield?

lsinger (approved):
@kevinansfield as discussed on our call, please just use a recent convention that feels good for the filename. 🙂

ErisDS (approved):

ErisDS:
My 2c here is as long as we use hyphens, similar language (e.g. add and not adding or added) and the right date format it's fine.

Uppercase, snake case, spaces etc all cause real issues :D

PR #23325, opened 2025-05-14
Specified why we need to test on staging

allouis:
In an engineering meeting on 4th December 2024 we discussed that performance on our local machines is not comparable to a production environment, the outcome of this meeting was an agreement that we would test migrations on staging (or production-like environment) in order to understand the performance qualities.


.github/workflows/migration-review.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
             ### General requirements
 
-            - [ ]  :warning: Tested on the staging database servers
+            - [ ]  :warning: Tested performance on the staging database servers
```
lsinger, line 33: Maybe adding the why directly in the checklist helps to provide context in the future? 

```suggestion
            - [ ]  :warning: Tested performance on staging database servers, as performance on local machines is not comparable to a production environment
```

lsinger: @allouis I do wonder whether / where we have documented the _how_ here, and if there's maybe a link or reference we could provide to get to that documentation easily from the checklist.

lsinger: From our discussion on a call, I took these notes -- maybe some of this is already documented, and some of it still needs to be documented. Either way it would be nice to have a pointer in the checklist item(s) to where to look this up. 

When should we test on staging servers? 
=> When the migration does something potentially costly in a potentially large table. 

What are potentially large or difficult tables? 
- `email_recipients`
- `posts`

What are potentially costly operations? 
DDL / changing data structures:
- adding an index
- dropping columns
- maybe: adding a default value to a previously nullable column

DML / changing data:
- calculating data
- loops (and potentially nested loops)

The distinction between MySQL vs. SQLite might be necessary to test when ...
- renaming columns
- changing data involving timestamps or booleans
- adding lots of data (SQLite might have low limits)

What we didn't have a good insight on yet was _how_ to test on staging servers, which I think we should write up somewhere.

ErisDS (approved):

ErisDS:
Cleaning up older PRs 💅 

(we should reopen and merge this tomorrow or something)

PR #23349, opened 2025-05-14
Fixed errors from #23330, restore i18n strings, fix test

cathysarisky:
no -submitted- issue

#23330 was missing a correction to i18n.test.js that checked for matching {{ and }} in translations. With the change to single { around variables, the test was passing incorrectly for mis-matched single {.

When I corrected the test, it became apparent that something in the merge/update/rebase to create #23330 had resulted in missing { around some interpolated variables.

I restored the test for paired (now single) braces and fixed the missing { around some variables.

#23330 also accidentally wiped out a handful of strings in each language (due to changes in keys resulting in missing braces).  This PR restores those translations.

In the future, we should also test that expected variables (and no unexpected variables) are present in translator submissions. We haven't been doing that, and turning it on results in localized test failures. I've left that test commented out until we have a better plan for enabling it.

How this PR actually got created (files changed doesn't do it justice):
-- grab the pre-23330 strings for all languages for the portal  and ghost namespaces.
-- Re-run the 'magic' script that rewrites all their variables to be single { wrapped.
-- grab the newsletter strings (all language) and re-combine them with the rewritten ghost namespace strings. 
-- Combine all the portal.json and ghost.json files with the brace-fixing part of https://github.com/TryGhost/Ghost/pull/23346 . 
-- Run yarn translate and pray.  (results in sorting of strings into the same order they were before this PR, removing most changes except the needed ones)


ghost/i18n/migration.js
daniellockyer, line 1: Is this file needed? 👀

daniellockyer (approved):
LGTM except the one bonus file (commented)

ghost/i18n/migration.js
cathysarisky: uh. uh. oops. Fixing!

PR #23352, opened 2025-05-15
Adjusted styles for buttons across the email template

sanne-san:
Ref https://linear.app/ghost/issue/PROD-1688/implement-button-corners-setting
- The button, product and header cards now use the same button styling.
- This should make it easier to maintain and update the email templates and ensures consistent styling.
- All these changes are behind the `emailCustomizationAlpha` feature flag.


ghost/core/core/server/services/email-service/email-templates/partials/styles.hbs
```diff
@@ -1568,6 +1727,48 @@ img.kg-cta-image {
...
+    box-sizing: border-box;
+    width: auto;
+}
+table.btn td {
+    padding: 8px 16px;
+    background-color: #ffffff;
+    border-radius: 5px;
+    {{#hasFeature "emailCustomizationAlpha"}}
+    border-radius: {{buttonBorderRadius}};
+    {{/hasFeature}}
```
kevinansfield, line 1742: We're already in the "emailCustomizationAlpha" conditional here so we don't need the extra one 🙂 

```suggestion
    border-radius: {{buttonBorderRadius}};
```

ghost/core/core/server/services/email-service/email-templates/partials/styles.hbs
```diff
@@ -1568,6 +1727,48 @@ img.kg-cta-image {
...
+    text-align: center;
+    font-family: -apple-system, BlinkMacSystemFont, Roboto, Helvetica, Arial, sans-serif;
+}
+
+table.btn a {
+    border-radius: 6px;
+    {{#hasFeature "emailCustomizationAlpha"}}
+    border-radius: {{buttonBorderRadius}};
+    {{/hasFeature}}
```
kevinansfield, line 1751: Same comment as above

```suggestion
    border-radius: {{buttonBorderRadius}};
```

ghost/core/core/server/services/email-service/email-templates/partials/styles.hbs
```diff
@@ -1753,9 +1951,15 @@ img.kg-cta-image {
...
+    table.body table.btn {
+        width: 100% !important;
+    }
+    {{else}}
     table.body .btn table {
         width: 100% !important;
     }
+    {{/hasFeature}}
 
     table.body .btn a {
```
kevinansfield, line 1961: Should these styles also live in the `/* --- BUTTONS --- */` section?

sanne-san: Not sure – there's currently one long section at the bottom of the file that includes all the mobile styles. So unless we're going to unpick all of that, I'd personally prefer to keep it that way.

PR #23370, opened 2025-05-16
Improved admin 2FA browser tests

ibalosh:
I have decided to refactor one of the admin browser tests - two factor authentication.

The reason for this is based on an observation that the browser tests can be improved in my opinion. 

**I see couple of reasons for improvement**

* locators are hard coded in each test, which makes them brittle to changes
* there are better choices for locating objects on the page, that will make the tests more long lasting
* tests could use a bit better arrange act assert pattern, to make them more readable
* by better choice of names, tests can better act as documentation of features

If tests are more resilient on refactoring, developers will be more motivated on updating them if they fail. 
Less false positives, more motivation for the updates.

**The updates made to the test**

* clean up code a bit 
* have better choices in terms of locators
* arrange code in more AAA pattern fashion
* introduce [page objects ](https://playwright.dev/docs/pom) 

This should allow better readability of the code, allow you to reuse pages with their methods, without the need of hard coding locators, or needing to update them in multiple places.

In my opinion, this should make it easier to update, and add new tests. I have only refactored one file, to see if you like this direction, and see the value too in the update.


cmraible:
Hi @ibalosh thanks for this, looks really good! I'm just rebasing and kicking off CI, if everything checks out I'm aiming to merge this today

ibalosh:
thanks @cmraible !

PR #23375, opened 2025-05-16
🌵pnpm?

cmraible:
just for fun


apps/admin-x-settings/package.json
```diff
@@ -40,12 +40,20 @@
     "@tryghost/limit-service": "1.2.17",
     "@tryghost/nql": "0.12.7",
     "@tryghost/timezone-data": "0.4.7",
+    "@tanstack/react-query": "4.36.1",
+    "@ebay/nice-modal-react": "1.2.13",
+    "react-hot-toast": "2.5.2",
+    "@dnd-kit/sortable": "7.0.2",
+    "react-select": "5.10.1",
+    "clsx": "2.1.1",
```
ErisDS, line 48: At the bare minimum, would be really good to pull out all the missing dependency fixes and merge those!

cmraible: Yep that's the plan! We'll need to do it for pnpm or yarn4, so trying to get the ball rolling on the stuff we can do in advance 👌

vikaspotluri123:
One thing to consider if this becomes mainlined is that the CLI currently installs Ghost using a bundled yarn 1.x dependency. If the core package manager switches to Yarn4 or pnpm, the CLI should probably do the same

cmraible:
Closing this as we don't have immediate plans to actually implement this. We can still use it as a reference for updating dependencies and taking baby steps in this direction though.

PR #23380, opened 2025-05-17
🌐✨Allow customization of locale for transactional email sending requested via Portal

cathysarisky:
This PR adds the ability to pass a locale to to the magic link endpoint, resulting in users receiving sign-in/subscribe/sign-up emails in whatever locale is passed, overriding the sitewide default.

The Portal app can already be customized for language, accepting a data-locale attribute in the script tag and using it to overwrite sitewide locale throughout Portal.  This PR causes Portal to also pass the locale to the magic link endpoint, and sets up the magic link endpoint to receive it and use it for sending localized text.

As data-attributes do not currently take a locale argument, I haven't updated that code to pass an alternative locale.  Will save that as a separate piece of work.

Extra context:  Inability to localize email for multiple languages is the biggest blocker to running one Ghost install for sites with multiple languages in use.   (Web issues can largely be handled with some theme trickery, or a client-side translation integration.)  This PR addresses the transactional email piece of the problem.


cathysarisky:
in draft mode while I investigate an error from merging.

cathysarisky:
Closing per Slack discussion.

PR #23384, opened 2025-05-18
Updated dependencies for posts app

cmraible:
no refs

This commit adds dependencies that were being used in the posts app, but weren't declared as dependencies. It also removes a few packages that were declared, but not used.


cmraible:
Closing to reopen on upstream so it will deploy to staging

PR #23385, opened 2025-05-18
✅ Tests for valid i18n json files.

cathysarisky:
Goal: prevent commits of translation strings with added/missing variables, while allowing enough flexibility for cases where the translation really can't use all the available variables.

I'm introducing a utility that generates & updates todos for i18n.  The i18n-todos file contains a list of problems to be ignored in testing.  I'm using it right now to cause tests to pass while we wait for translators to sort out the problems, but it'll also be useful for cases where a variable might be missing for good reasons (because we need a less than exact translation to fit a UI element, for example).

I added a test for context.json not having blank values, because that's been an ongoing problem.  The flow here is a little weird - now 'yarn test' in the i18n package runs the tests, runs yarn translate (which generates any new strings and populates context.json), then runs tests again.  This avoids running string generation if something is wrong with the initial tests, while also picking up new problems following string generation.  But it does look a little weird. 🙄


vikaspotluri123:
> I'm introducing a utility that generates & updates todos for i18n.

I like the utility, but I think it should be removed before merging, since it's hopefully going to be used once 😀 

1: re: todos - does it make sense to revert the translations you detected as referencing added/undefined variables? When they get resolved at runtime, the sentence might not make sense (e.g. I'm assuming `{amount} off` becomes ` off`). I'm less sure for unused variables since there's a possibility for that to be intentional

2: nit - I don't think "todo" is the best name since the longer-term goal is to use it as a "I did this on purpose" file

> I added a test for context.json not having blank values, because that's been an ongoing problem. 

It might make sense to do this in a separate PR, mostly because of what I'm about to suggest 😆 

  * I've been thinking _against_ updating all translations in a PR - since we support so many locales, adding the untranslated key to each locale adds a lot of overhead to reviewing (GH UI isn't great, and lots of changes).
     * Alternative: have GitHub Actions detect changes to `en.json`, updates all the translations, and PRs (automerge)/commits directly to main
     * This requires `en.json` to be in sync, so there might need to be a separate target that devs need to run🤔 


3: Failing tests don't report all errors. It might be more contextual if we use the eslint reporter and drop mocha entirely. Since it's catching issues right now, I can explore this approach in a follow-up PR.

ghost/i18n/package.json
```diff
@@ -9,7 +9,8 @@
   "scripts": {
     "dev": "echo \"Implement me!\"",
     "test:base": "NODE_ENV=testing c8 --include index.js --include lib --check-coverage --100  --reporter text --reporter cobertura -- mocha --reporter dot './test/**/*.test.js'",
-    "test": "yarn test:base && yarn translate",
+    "test": "yarn test:base && yarn translate && yarn test:base",
```
vikaspotluri123, line 12: Unrelated: do you know why we run `yarn translate` as part of the test target? This doesn't trigger CI to fail

ghost/i18n/test/i18n-todos.json
```diff
@@ -0,0 +1,199 @@
...
+            {
+                "file": "da/portal.json",
+                "key": "Try free for {amount} days, then {originalPrice}.",
+                "comment": "Auto-generated: missingVariable detected in translation"
+            },
+            {
+                "file": "de/portal.json",
+                "key": "Start {amount}-day free trial",
+                "comment": "translated text doesn't fit on the button",
+                "reference": "https://github.com/TryGhost/Ghost/pull/123456-fake-pr"
```
vikaspotluri123, line 70: 🙌🏾

ghost/i18n/package.json
cathysarisky: I /think/ I remember that the behavior on CI is slightly different, and failOnUpdate ends up set.

ghost/i18n/test/i18n-todos.json
cathysarisky: Oops, I could probably tidy up the faked reference I was using for manual testing, huh?

cathysarisky:
> I like the utility, but I think it should be removed before merging, since it's hopefully going to be used once 😀

I can see needing it again - either to generate new items because we decided deliberately not to use a variable, or to clean up the todos (or whatever we're calling it) after some tidying up of the translations.  CI doesn't run it, but I think it's useful as a tool, and I'd rather keep the tools where I need them than trying to remember where I put it locally... 🤷‍♀️ 

> 2: nit - I don't think "todo" is the best name since the longer-term goal is to use it as a "I did this on purpose" file

Agreed.  "Overrides" ?

> 1: re: todos - does it make sense to revert the translations you detected as referencing added/undefined variables? When they get resolved at runtime, the sentence might not make sense (e.g. I'm assuming {amount} off becomes  off). I'm less sure for unused variables since there's a possibility for that to be intentional

Some unused vars are definitely intentional.  (German is too damn long to fit on the button in a few spots.)  In most cases, reverting will mean a blank value, which will be rendered as English.  Better or worse?  Depends on the specific case.  Another option if translators don't pick them up would be to do an AI-generated fix.  I don't love that options either.

> 3: Failing tests don't report all errors. It might be more contextual if we use the eslint reporter and drop mocha entirely. Since it's catching issues right now, I can explore this approach in a follow-up PR.
Good point.  It'd be more useful to have a list of all the problems, rather than just the first one.

PR #23386, opened 2025-05-18
Updated dependencies for posts app

cmraible:
no refs

This commit adds dependencies that were being used in the posts app, but weren't declared as dependencies. It also removes a few packages that were declared, but not used.

There shouldn't be any functionality or dependency changes here, since the `yarn.lock` didn't actually change. When upgrading to yarn 4 or pnpm, they won't allow packages with missing dependencies, so doing this for all packages in the monorepo is the first step toward upgrading to a more modern package manager.

### DIFF
--- a/apps/posts/package.json
+++ b/apps/posts/package.json
@@ -26,12 +26,23 @@
         "preview": "vite preview"
     },
     "devDependencies": {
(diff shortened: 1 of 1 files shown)


Ghost-Slimer:
Deployed to **staging** with ID: `3416`

PR #23389, opened 2025-05-18
Updated dependencies for stats apps

cmraible:
no refs

Added dependencies that are used in the stats app, but not declared in the package.json. Also removed a few dependencies that aren't actually used. Also moves some dependencies from `devDependencies` to `dependencies`, since they are used in the apps source code.

Modern package managers like pnpm and yarn 4 don't allow undeclared dependencies, so this is the first step in migrating to a more modern package manager, plus it's just good practice in general.


Ghost-Slimer:
Deployed to **staging** with ID: `3418`

PR #23431, opened 2025-05-20
🌐 Updated Bulgarian translations

yovko:
Translation of some newly introduced strings


cathysarisky:
Thanks!  As usual, it'll be a bit before these are live on your site - I'm planning to rebuild & rerelease the comments app at the end of this week so that these can be included.

yovko:
Thanks!

PR #23439, opened 2025-05-20
Added config option to disable stripe donations

ErisDS:
ref https://linear.app/ghost/issue/ONC-911/support-escalation-turn-off-donations-entirely

- When you connect stripe, the tips and donations feature is automatically enabled
- In some contexts (e.g. political) having this available is not desirable
- This adds a config-only option to disable the feature as the case is rare and doesn't warrant a setting in UI
- Instead we hide the UI if this config flag is disabled

### DIFF
--- a/apps/admin-x-framework/src/test/responses/settings.json
+++ b/apps/admin-x-framework/src/test/responses/settings.json
@@ -309,6 +309,10 @@
             "value": true
         },
         {
+            "key": "donations_enabled",
+            "value": true
+        },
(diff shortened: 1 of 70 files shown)


cathysarisky:
Hey @coderabbitai , please please please learn that it is NOT necessary to identify translation files (in ghost/i18n/locales/*/*.json) that have a blank value for the translation.  This is totally normal, and you don't need to spam the PR about it.

PR #23453, opened 2025-05-21
Added caddy to docker compose `split` profile

cmraible:
no refs

In production, events sent to Tinybird are sent via the site's own domain at the `/.ghost/analytics` path. This commit adds Caddy to the `compose.yml` file, which will allow us to mimic this setup locally, by forwarding requests to `/.ghost/analytics` to the analytics service, rather than to Ghost itself. 

This commit doesn't add this functionality straight away, but simply adds a basic Caddy configuration that we can build on. When using the `split` profile (which will become the default soon), running `docker compose up` will include this Caddy service, which does a few things:
- Serves Ghost locally at `http://localhost:80`
- Serves Ghost locally using TLS at `https://localhost:443`
     - Your browser will throw a warning the first time you access it because Caddy's CA isn't trusted. While the server is running, you can run `sudo ./.docker/caddy/trust_caddy_ca.sh` to trust Caddy's CA on your host machine, which will remove these warnings
- Additionally serves Ghost at `https://site.ghost` and admin at `https://admin.ghost`
     - You'll need to add these to your `/etc/hosts`, and update your `config.local.json` in Ghost for this to work fully


cmraible:
Pros:
- SSL out of the box (sort of)
- Simpler configuration than [nginx](https://github.com/TryGhost/Ghost/pull/23449)

Cons:
- Requires some manual setup to use custom hostnames and trust Caddy's CA

PR #23479, opened 2025-05-21
RFC: Added true e2e test suite

cmraible:
This PR creates a brand new e2e test suite from scratch. The `apps/e2e/README.md` file included in this PR is the best place to start to get acquainted with:

- What is this new test suite?
- Why do we need _another_ test suite?
- How this test suite works
- How to use this test suite


ibalosh (commented):
took a quick glance, just shared couple of ideas for future tests, nice that you are exploring switching these to typescript! it's a more natural place for them.

One thing that caught little bit my eye is, to be cautious to and not overuse expectations in page objects.
Although it's there in playwright docs too, in it's original idea Page Object model is mostly for describing pages (it's sections and properties with locators etc).

Having expectations outside of POM, makes better separation of concerns, and easier reading of the tests and their intent. Through time it can get also difficult the manage. It also kinda breaks the ideal test -> that has 1 assert (not a very strict rule to obey, and hard to satisfy, but still one to strive for)

apps/e2e/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Managing State
+One of the key considerations in building a test suite such as this is how we manage state, which includes the MySQL database state and increasingly state that is kept in other services, i.e. Tinybird for analytics data. One approach we could use here is to hook into Ghost itself to i.e. truncate database tables and load new fixtures. As much as possible, we should avoid doing this, because it couples our tests too closely with the application code. Another, IMO better approach is to directly manage the state in whatever system it is stored in from the test suite itself. For example, rather than hooking into Ghost to truncate tables and load fixtures, we should build the infrastructure to allow the test suite to talk directly to MySQL and change the state itself.
```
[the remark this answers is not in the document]
cmraible: I like this a lot, thanks for the suggestion! Agree this will probably be a smoother route, especially for the "test in production" piece of this.

e2e/src/page-objects/LoginPage.ts
```diff
@@ -0,0 +1,41 @@
...
+        this.page = page;
+        this.adminURL = adminURL;
+        this.emailInput = page.getByRole('textbox', {name: 'Email address'});
+        this.passwordInput = page.getByRole('textbox', {name: 'Password'});
+        this.signInButton = page.getByRole('button', {name: 'Sign in →'});
+        this.defaultUsername = defaultUsername;
+        this.defaultPassword = defaultPassword;
+    }
+
+    async goto() {
```
ibalosh, line 22: I might be suggesting this too early,  but.. figured to share it :) 

My assumption is that you will probably have this kind of methods across bunch of page objects, which makes them suited for a base/parent admin page and inheritance. Things like page url, visit to it, with default waits, maybe other helper methods like page resize, scroll to the end etc. They can be easily reused across pageobjects,

ErisDS:
I would like to see this merged! However, I have additional requirements:
- This suite must be run in CI
- It should be required to pass for all changes
- It must work on PRs from forks (so mocked mode?)

Thoughts?

cmraible:
> I would like to see this merged! However, I have additional requirements:
> 
> - This suite must be run in CI
> - It should be required to pass for all changes
> - It must work on PRs from forks (so mocked mode?)

No disagreement, makes sense. Just thinking outloud: the main challenge to meeting those criteria is that these tests will depend, at least in part, on using docker compose in CI — ideally for the whole setup including Ghost, but we could also use local Ghost + backing services in compose setup. This setup would be a lighter lift initially, since we aren't currently building the docker image in CI. Does a hybrid approach sound okay to you, for the sake of getting this test suite merged and operational ASAP?

My ideal "dream state" for our CI is that we replace the current `setup` step with a `build` step, which builds the docker image, then all the downstream jobs can run their tests against the built image, with whatever backing services are needed running in compose. However, doing that means pretty much a full rewrite of our CI processes, which I'd love to do but probably is out of scope for the sake of this specific change.

cmraible:
> I would like to see this merged! However, I have additional requirements:
> 
> This suite must be run in CI
> It should be required to pass for all changes
> It must work on PRs from forks (so mocked mode?)

@ErisDS No disagreement, makes sense. Just thinking outloud: the main challenge to meeting those criteria is that these tests will depend, at least in part, on using docker compose in CI — ideally for the whole setup including Ghost, but we could also use local Ghost + backing services in compose setup. This setup would be a lighter lift initially, since we aren't currently building the docker image in CI. Does a hybrid approach sound okay to you, for the sake of getting this test suite merged and operational ASAP?

My ideal "dream state" for our CI is that we replace the current setup step with a build step, which builds the docker image, then all the downstream jobs can run their tests against the built image, with whatever backing services are needed running in compose. However, doing that means pretty much a full rewrite of our CI processes, which I'd love to do but probably is out of scope for the sake of this specific change.

ErisDS:
I am in complete agreement on both the dream end state and intermediate steps 🙌

e2e/src/page-objects/LoginPage.ts
cmraible: Yes! This is definitely the dream, totally agree that we should do this from the start, so as we build out the test suite, we'll develop a full set of standardized Page Objects with the same core properties & methods

cmraible:
I've let this PR go stale 🙈

I think the right path forward here is to use this PR as a reference, but start from scratch with a super minimal implementation:
- e2e package created at the top level, same as this
- enough of a docker compose setup to boot Ghost and run the tests locally and in CI
- a simple playwright test that just visits the site's frontend and asserts that it renders

That will give us the foundation of a test suite that we can run locally and in CI, that runs on every PR and must pass before allowing the PR to merge. We can then incrementally add tests and complexity (i.e. run tinybird local) as needed for the tests we plan to write.

cmraible:
Closing in favor of https://github.com/TryGhost/Ghost/pull/23941

PR #23480, opened 2025-05-21
Updated ghost stats script to be more testable

9larsons:
no ref

This file was largely testing copies of its own logic rather than the file itself. While I'm not a huge fan of the class-based approach, it's functional and makes it MUCH easier to mock. The tests are now actually calling the code, and the browser-related pieces have been separated into their own service for testing separation.


9larsons:
Tested locally, seems to still work 👍

PR #23491, opened 2025-05-22
Added navigation support for mentioned users in ActivityPub notification

minimaluminium:
ref PROD-1863

- Implemented profile link handling for mentioned users in reply and mention notifications
- Redirected user mention links to open within the admin interface instead of new tabs


vershwal (approved):
Good one @minimaluminium!

PR #23502, opened 2025-05-22
Debugging mysql pull rate limits in docker hub

cmraible:
Trying to figure out why we're hitting rate limits in CI when pulling the mysql image, despite authenticating with docker hub

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -504,6 +504,39 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Show DOCKER_CONFIG on runner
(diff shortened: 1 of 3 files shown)


cmraible:
The issue seems to be that the mysql action itself runs in Docker, not directly on the runner, so the docker credentials from the previous "login to docker" step aren't available to the action, thus it's not really using the username & token. 

We could add our own docker login step to the MySQL action manually, which I've tested and it works, but this is potentially less secure -- the token is saved in plain text in the docker container. That's arguably fine, since the container is destroyed when the run completes, and assuming it's a read only token (which we should double check), the risk is pretty low even if it is somehow compromised.

Alternatively -- it seems a bit weird that this action runs in docker to begin with, just to pull _another_ docker image and run it. I'm sure there's probably a reason we're doing it this way instead of using the pre installed MySQL on the runner itself, or using GHA "services" to run it, but those are both worth exploring as well, and might offer some performance benefits over running MySQL in docker-in-docker.

PR #23519, opened 2025-05-26
Added migrations for newsletter customization columns

kevinansfield:
closes https://linear.app/ghost/issue/PROD-1875

- dropped `newsletters.border_color` column
  - column was left over from previous email customization work that was never used
  - we're not planning to use the field in current work so we're removing it to keep things clean
  - cleaned up unused border color code
- renamed `newsletters.title_color` to `newsletters.post_title_color`
  - to avoid confusion between post/section title colors we've decided to change `title_color` to `post_title_color` so it's more explicit
  - matching pattern will be used for soon-to-be-added `section_title_color` column
- added color-related columns to newsletters table
  - adds color-related email customization columns to `newsletters` table ready for implementation
  - updated `header_color` usage in newsletter design prototype to match final naming of `header_background_color`


lsinger (commented):
@kevinansfield I'm still seeing a few references to `border_color` or `borderColor` in a few places. Most I think are unrelated, but I'm not sure about e.g. the one in `apps/admin-x-settings/src/components/settings/email/newsletters/NewsletterPreviewContent.tsx:40` or `ghost/admin/app/models/newsletter.js:36`. Are those indeed unrelated to this property / column?

lsinger:
> [Here's a link](https://github.com/TryGhost/Ghost/blob/126d2af9dc61be18acdbe75b2503298d742dd363/ghost/admin/app/models/newsletter.js#L36) to the `newsletter.js` reference on this PR's branch -- it's still there?

[The next line](https://github.com/TryGhost/Ghost/blob/126d2af9dc61be18acdbe75b2503298d742dd363/ghost/admin/app/models/newsletter.js#L37) has a `titleColor` -- does that need renaming to `postTitleColor`?

lsinger (approved):
Apart from the two lines `ghost/admin/app/models/newsletter.js` that I mentioned I'm not sure about this looks good to go. 🙂 

Feel free to merge once you've confirmed those two lines (`borderColor ` and `titleColor` mentions) are either not a problem or are updated.

ronaldlangeveld:
LGTM. 

Was wondering why there's UI changes in a migration PR, but makes sense in this case since it would otherwise break existing code. 👌🏻

lsinger:
> LGTM.

@ronaldlangeveld thank you for taking a look. To be explicit, did you confirm that the two lines mentioning `borderColor` and `titleColor` in `newsletter.js` have nothing to do with the fields being deleted and changed in this PR?

PR #23526, opened 2025-05-26
SPIKE - Extracted Koenig button renderer into Ghost

kevinansfield:
ref https://linear.app/ghost/issue/PROD-1784

- added `ghost/core/server/services/koenig` directory
- copied button-related renderer files over from Koenig and converted from ES6 modules to CJS
- updated call to our lexical renderer to pass our custom renderers object as an option
- NOTE: currently adds "CUSTOM" to text of all buttons to provide a visual indication of the custom renderer being used


kevinansfield:
Closing as we're moving forward with implementation based on this spike outcome

PR #23568, opened 2025-05-28
🐛 Fix interpolation of some strings in Portal

cathysarisky:
no issue 

The Interpolate component used in some spots of Portal used SYNTAX_I18NEXT from @doist/react-interpolate, which expected double {{ instead of single.  Since the default is single braces and we're now using single braces, I just needed to remove the syntax line.

TODO:  Failure of these tests to fail with the switch from {{ to { went in suggests we might need some more test coverage.


ErisDS (approved):
LGTM!

cathysarisky:
closing in favor of other PR, due to random git-wizardry required.

PR #23570, opened 2025-05-28
🐛 Fix portal interpolation errors.

cathysarisky:
no ref

The Interpolate component used in some spots of Portal used SYNTAX_I18NEXT from @doist/react-interpolate, which expected double {{ instead of single. Since the default is single braces and we're now using single braces, I just needed to remove the syntax line.


vikaspotluri123:
Forum ref: https://forum.ghost.org/t/amount-appear-in-the-pricing-box/57810

PR #23585, opened 2025-05-29
✨ Changed staff token perms to match user auth

ErisDS:
closes https://linear.app/ghost/issue/ENG-2148/improve-staff-token-api-access

- Staff tokens were added years ago, but they were given the same permissions as integrations
- Integrations are restricted to "safe" actions and stable endpoints at the moment
- However, staff tokens are intended to give much the same access as if you login with your username and password
- There are two notable exceptions: Transfering Ownership & Deleting all Content. These cannot be done with a token
- This change uses a permission error instead of not implemented error (that needs to change for integrations too)
- It also introduces a full suite of middleware for the "notImplemented" middleware that seems to have flown under the radar
- There are also extensive e2e api tests, because this behaviour is critical and depends on middleware both up and down the stack working well too


ErisDS:
@coderabbitai your commit didn't apply properly

ErisDS:
@coderabbitai you've identified the problem, but not made a commit or suggested a change that I can accept. Can you take steps to actually fix it?

PR #23590, opened 2025-05-29
Bump `eslint-plugin-ghost` from 3.4.0 -> 3.4.3

allouis:
ref https://linear.app/ghost/issue/PROD-1899

We've updated the package to no longer have a git dependency, which we saw cause some errors in CI. This should have no functional changes, but will avoid running into these errors in future.


ErisDS:
Renovate has already done this! 😂

allouis:
Lol ffs!

PR #23610, opened 2025-06-01
🎨 Fixed infinite wait when auth code email fails to send in Direct mode

### LINKED ISSUE #23050: Regression: Admin login/session cookies broken in ghost:5.118.1-alpine (works in 5.117.0-alpine and earlier)
### Issue Summary

After upgrading from `ghost:5.117.0-alpine` to `ghost:5.118.x-alpine`, admin login is broken: no cookies are set in the browser, and Ghost logs show 403 errors with "Unable to determine the authenticated user or integration. Check that cookies are being passed through if using session authentication."
Rolling back to `ghost:5.117.0-alpine` or earlier resolves the issue.

### Steps to Reproduce

1. Deploy Ghost using Docker with the ghost:5.118.1-alpine image.
2. Use a standard nginx reverse proxy ([as per this tutorial)](https://community.hetzner.com/tutorials/ghost-cms-on-arm64-debian).
3. Try to log in to /ghost/#/signin from a new browser or incognito window.
4. Observe that no cookies are set and login fails with a 403.
5. Revert Docker with ghost:5.117.0-alpine
6. Try to log in to /ghost/#/signin from a new browser or incognito window. Login works successfully. 
7. Redeploy with the ghost:5.118.1-alpine image and try to login again. Login fails as described in (4)

If using the same browser that did the initial set-up, then the cookie is already set and login can be completed successfully. 

### Ghost Version

5.118.1

### Node.js Version

v18.20.8

docker-alpine in a Hetzner VPS (Arm64 CAX11)

### Database type

MySQL 8

### Browser & OS version

Tested on MacOS Sequoia: Zen (Firefox) / Chromium / Safari; Android 15: Iceraven (Firefox), Chrome

```shell
[2025-04-26 21:06:34] ERROR "GET /ghost/api/admin/users/me/?include=roles" 403 6ms

Authorization failed

"Unable to determine the authenticated user or integration. Check that cookies are being passed through if using session authentication."

Error ID:
    55d2b170-22e2-11f0-90d2-31b40eaea8c0

----------------------------------------

NoPermissionError: Authorization failed
    at authorizeAdminApi (/var/lib/ghost/versions/5.118.1/core/server/services/auth/authorize.js:33:25)
    at Layer.handle [as handle_request] (/var/lib/ghost/versions/5.118.1/node_modules/express/lib/router/layer.js:95:5)
    at next (/var/lib/ghost/versions/5.118.1/node_modules/express/lib/router/route.js:149:13)
    at authenticate (/var/lib/ghost/versions/5.118.1/core/server/services/auth/session/middleware.js:55:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

### Code of Conduct

PaulBernhardt:
fixes https://github.com/TryGhost/Ghost/issues/23050

When email is configured to use Direct mode, it can take over 15 minutes for the email to actually fail to send. When a user attempts to sign in on a new device in this situation, the login screen will appear to be stuck with no feedback.

This is especially problematic because the Docker installation sets up Direct mode for you automatically, and most hosts / ISPs block outbound port 25. This likely means almost any self hoster that has not taken the step of explicitly setting up email will unknowingly be unable to log in to from a new device, and will have difficulty even discovering why.

This change adds a timeout in Direct mode to the sendAuthCodeToUser function. If the email has not been sent after 15 seconds, an error is thrown, which will then surface the 'Failed to send email. Please check your site configuration and try again.', which will will then lead them to either setting up email correctly or disabling the email 2FA.

### DIFF
--- a/ghost/core/core/server/services/auth/session/session-service.js
+++ b/ghost/core/core/server/services/auth/session/session-service.js
@@ -275,11 +275,19 @@ module.exports = function createSessionService({
         });
 
         try {
-            await mailer.send({
+            let sendPromise = mailer.send({
                 to: recipient,
                 subject: `${token} is your Ghost sign in verification code`,
                 html: email
             });
+            // In Direct mode, it can take 15+ minutes for the mailer to actually fail to send.
+            // Compromise by giving up after 15 seconds.
+            if (mailer.state?.usingDirect) {
+                sendPromise = Promise.race([sendPromise, new Promise((_resolve, reject) => {
+                    setTimeout(reject, 15000);
+                })]);
+            }
+            await sendPromise;
         } catch (error) {
             throw new errors.EmailError({
                 ...error,
--- a/ghost/core/test/unit/server/services/auth/session/session-service.test.js
+++ b/ghost/core/test/unit/server/services/auth/session/session-service.test.js
@@ -13,6 +13,10 @@ describe('SessionService', function () {
         getSettingsCache.withArgs('admin_session_secret').returns('secret-key');
     });
 
+    afterEach(function () {
+        sinon.restore();
+    });
+
     it('Returns the user for the id stored on the session', async function () {
         const getSession = async (req) => {
             if (req.session) {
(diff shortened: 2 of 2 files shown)


ghost/core/core/server/services/auth/session/session-service.js
```diff
@@ -275,11 +275,19 @@ module.exports = function createSessionService({
...
+            let sendPromise = mailer.send({
                 to: recipient,
                 subject: `${token} is your Ghost sign in verification code`,
                 html: email
             });
+            // In Direct mode, it can take 15+ minutes for the mailer to actually fail to send.
+            // Compromise by giving up after 15 seconds.
+            if (mailer.state?.usingDirect) {
+                sendPromise = Promise.race([sendPromise, new Promise((_resolve, reject) => {
+                    setTimeout(reject, 15000);
```
sam-lord, line 287: Would be ideal if this timeout was moved into a constant. I don't think we need to worry about making it configurable at the moment, but just to have somewhere to name it and comment the "why" for the value.

FWIW I think we could take this down to 10~ seconds. Ultimately by cutting this down to anything under 15 minutes, we're guaranteeing that we're only making a single attempt. After that, we should be more concerned with the user wondering why it's taking so long to complete their action.

sam-lord (changes requested):
I like the approach you've taken + the tests look good.

I think that these changes would be better scoped to the GhostMailer class (`ghost/core/core/server/services/mail/GhostMailer.js`), so that _all_ transactional emails are given the same timeout. 

I'd imagine this would be converted to something similar, like this (GhostMailer / line 130 onwards):
```js
        if (this.state.usingDirect) {
            // Prevent the UI from hanging for an unreasonable amount of time
            const timeout = 10_000;
            const timeoutPromise = new Promise((resolve, reject) => {
                setTimeout(() => reject(createMailError({
                    message: tpl(messages.reason, {reason: 'Timeout while sending email'})
                })), timeout);
            });

            // Assume email failed to send if it takes longer than the timeout
            const response = await Promise.race([timeoutPromise, this.sendMail(messageToSend)]);
            return this.handleDirectTransportResponse(response);
        }

        const response = await this.sendMail(messageToSend);
        return response;
```

9larsons:
@PaulBernhardt 👋 I'm going to close this out because it's stale. Please feel free to re-open and address Sam's comments.

PR #23634, opened 2025-06-03
Added background color setting behind alpha flag

ronaldlangeveld:
ref https://linear.app/ghost/issue/PROD-1692/implement-background-color-setting

- Implemented background colour picker in `NewsletterDetailModalLabs.tsx` under `emailCustomizationAlpha` flag
- Updated `EmailRenderer.js` to include `backgroundColor` in design options when emailCustomizationAlpha is enabled
- Modified `styles.hbs` to apply custom background colours to various email elements with conditional dark/light contrast support
- Adjusted text colours for better contrast based on background colour
- Updated snapshot tests in email-previews.test.js.snap and cards.test.js.snap to reflect new styles


ghost/core/core/server/services/email-service/email-templates/partials/styles.hbs
```diff
@@ -335,6 +349,48 @@ h6 + .kg-paywall .kg-paywall-hr td {
...
+
+/* Exclude CTA cards with colored backgrounds from custom text color, but allow transparent ones */
+.post-content-row .kg-cta-bg-white .kg-cta-text p,
+.post-content-row .kg-cta-bg-grey .kg-cta-text p,
+.post-content-row .kg-cta-bg-blue .kg-cta-text p,
+.post-content-row .kg-cta-bg-green .kg-cta-text p,
+.post-content-row .kg-cta-bg-yellow .kg-cta-text p,
+.post-content-row .kg-cta-bg-red .kg-cta-text p,
+.post-content-row .kg-cta-bg-pink .kg-cta-text p,
+.post-content-row .kg-cta-bg-purple .kg-cta-text p {
```
kevinansfield, line 389: I wonder if we could simplify this with `.post-content-row [class~="kg-cta-bg"] .kg-cta-text p {` 🤔 

Juice should inline properly still I think so we don't lose Outlook support

PR #23675, opened 2025-06-04
Improved performance for members subscribe events

9larsons:
ref https://linear.app/ghost/issue/PROD-1789/
- Created migration to add performance indexes for members_subscribe_events to optimize queries.
- Added index for email_disabled field in members table to improve filtering efficiency.
- Rebuild `subscriber-stats` endpoint query.

A few MySQL indices were added to improve the performance of querying the `members_subscribe_events` table for the purposes of the `subscriber-stats` endpoint, which could take several seconds for data for particularly large sites. The query was also rebuilt to make better use of the new indices, as well as more efficiently filter, as instead of looking for 'active' members we strip out 'inactive' members, a much smaller subset that is usually a few percentage of total members.

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/5.123/2025-06-04-17-48-47-add-members-email-disabled-index.js
+++ b/ghost/core/core/server/data/migrations/versions/5.123/2025-06-04-17-48-47-add-members-email-disabled-index.js
@@ -1,0 +1,3 @@
+const {createAddIndexMigration} = require('../../utils');
+
+module.exports = createAddIndexMigration('members', ['email_disabled']); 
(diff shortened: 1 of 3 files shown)


cmraible:
Generally everything looks good to me, although I haven't run through the full checklist yet. Few questions, but nothing blocking:
- Do the query changes need the indices, or could we merge them separately? Mainly just thinking in terms of being able to measure the impact of the indices, which is harder if we're changing the queries at the same time.
- Do we know what the max length of the `members_subscribe_events` table is on Pro? I couldn't find anything in metabase, but I'm imagining this could be in the ~millions of rows territory for larger sites.

ghost/core/core/server/data/migrations/versions/5.123/2025-06-04-17-48-47-add-members-email-disabled-index.js
```diff
@@ -0,0 +1,3 @@
+const {createAddIndexMigration} = require('../../utils');
+
+module.exports = createAddIndexMigration('members', ['email_disabled']);
```
cmraible, line 3: Wonder if there would be any benefit to making this `['id', 'email_disabled']` instead of just the `email_disabled` field alone?

The primary use-case for the `email_disabled` field seems to usually include a join on the member id (although you've avoided that in these queries) — adding the ID would theoretically make it a covering index and speed up those joins.

cmraible:
Maybe I'm being too paranoid, but I'd want to:
- Run a stats service query to find the biggest current `members_subscribe_events` table in production
- Replicate that in staging using the data generator
- Check how long adding the index takes

I think it's really unlikely it would take more than 30 seconds, but I think it's worth doing our due diligence on

PR #23698, opened 2025-06-05
Patched `esm` to fix node v22 compatibility

kevinansfield:
no issue

- `esm` is unmaintained and compatibility is broken in node v22 which has an API change
- created a patch using code from https://github.com/wallabyjs/esm
- added `patch-package` dependency with a `postinstall` package step to locally patch the file after installing Ghost which makes a working `esm` version available to dependencies/sub-dependencies without needing to fork multiple packages to change dependencies/require statements


kevinansfield:
Closing as this was merged in https://github.com/TryGhost/Ghost/commit/ae780ded33e1cfca69ec89576968b50e6f4cc386

PR #23739, opened 2025-06-09
Updated eslint config for TailwindCSS

peterzimon:
Problem: classname ordering is getting messed up in monorepos using eslint-plugin-tailwindcss. This is especially a problem when you want to update classname ordering automatically on save because the wrong order would be saved which fails tests on CI.

According to this issue (https://github.com/francoismassart/eslint-plugin-tailwindcss/issues/241) using absolute URL for the tailwind config file could help with the issue.


peterzimon:
The suggested fix didn't solve the problem. The issue turned out to be a config problem which was resolved via https://github.com/TryGhost/Ghost/commit/804639aa41f0dcb3f6c05c2b8e87569ebf887614

PR #23760, opened 2025-06-10
Added `site_uuid` to the site Admin API endpoint

cmraible:
ref https://linear.app/ghost/issue/PROD-197/ghost-site-uuid-implementation

The `site_uuid` is an identifier for a Ghost site that unique identifies the site, which allows for integrations with multi-tenant services, such as Tinybird for traffic analytics. The `site_uuid` is intended to public, since it will naturally be exposed on the frontend in network requests. To provide access to this value in admin, this commit adds the `site_uuid` setting value to the `/site` endpoint.


cmraible:
Note: snapshot tests failing currently because the migration hasn't been merged yet. Will need to rebase after migration merges, at which point these tests should pass.

PR #23763, opened 2025-06-10
Fixed CTA Visibility on transparent background

ronaldlangeveld:
ref https://linear.app/ghost/issue/PROD-1985/cta-buttons-invisible-with-no-bg-color-dark-global-bg-outline-buttons

- Added `backgroundIsDark` check in `EmailRenderer` to detect dark newsletter backgrounds based on `background_color`.
- Updated `call-to-action-renderer` to convert black buttons (`black`, `#000000`, `#000`) to white when CTA background is transparent and newsletter background is dark, ensuring visibility.
- Added tests.
- Implemented colour logic in `call-to-action-renderer` instead of `styles.hbs` to use knowledge about button and card colours, to have better control of the CTA rendering in emails.

### DIFF
--- a/ghost/core/core/server/services/email-service/EmailRenderer.js
+++ b/ghost/core/core/server/services/email-service/EmailRenderer.js
@@ -1146,7 +1146,7 @@ class EmailRenderer {
         const hasAnyEmailCustomization = labs.isSet('emailCustomization') || labs.isSet('emailCustomizationAlpha');
 
         const backgroundColor = this.#getBackgroundColor(newsletter);
-        const backgroundIsDark = textColorForBackgroundColor(backgroundColor).hex().toLowerCase() === '#ffffff';
+        const backgroundIsDark = this.#checkIfBackgroundIsDark(newsletter);
         const postTitleColor = this.#getPostTitleColor(newsletter, accentColor);
         const titleWeight = this.#getTitleWeight(newsletter);
         const titleStrongWeight = this.#getTitleStrongWeight(titleWeight);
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/services/email-service/EmailRenderer.js
```diff
@@ -367,6 +367,7 @@ class EmailRenderer {
                 ...renderOptions.design,
                 ...betaDesignOptions,
                 backgroundColor: newsletter?.get('background_color'),
+                backgroundIsDark: this.#checkIfBackgroundIsDark(newsletter),
```
kevinansfield, line 367: Would be good to update the `backgroundIsDark` variable used inside `getTemplateData()` to call the same function

ghost/core/core/server/services/koenig/node-renderers/call-to-action-renderer.js
```diff
@@ -87,6 +87,17 @@ function emailCTATemplate(dataset, options = {}) {
...
+    if (options.feature?.emailCustomizationAlpha) {
+        const isTransparentCTA = dataset.backgroundColor === 'none';
+        const isDarkBackground = options.design?.backgroundIsDark;
+        const isBlackButton = dataset.buttonColor === 'black' || dataset.buttonColor === '#000000' || dataset.buttonColor === '#000';
+
+        if (isTransparentCTA && isDarkBackground && isBlackButton) {
+            buttonStyle = `color: #000000;`;
+            dataset.buttonColor = 'white';
+        }
+    }
```
kevinansfield, line 99: @sanne-san this currently applies to both outline and filled buttons, should it only apply when `buttonStyle` is `'outline'` as per the issue?

Examples of filled buttons without this change:
Screenshot 2025-06-10 at 11 21 18
Screenshot 2025-06-10 at 11 23 20

ronaldlangeveld: Yeah, sorry I should have mentioned - that's an edge case I found outside of just the outline.

PR #23767, opened 2025-06-10
Updated reply chain handling to support `hasMore`

allouis:
The API has been updated to provide a `hasMore` property rather than `next` for ancestors and reply chains, as this better reflects the capabilities of the API


allouis:
https://github.com/TryGhost/ActivityPub/pull/808

PR #23797, opened 2025-06-11
Added Top Sources to top content component in Posts > Growth

9larsons:
ref https://linear.app/ghost/issue/PROD-1817/


9larsons:
At the current moment, MRR is showing but there's a few issues:
- Sources data isn't updating/changing with lookback changes 
- Loading new data seems to re-render the top level components sometimes
- MRR accuracy may be in question
- Default currency is used; note that this will likely be incorrect for sites w/ multiple currencies, but that's not something we fully support

9larsons:
I think this should take care of it.

PR #23818, opened 2025-06-12
🌐 Added i18n support for meta title

SebastianSchroeder:
Enables missing i18n support for the meta title when pagination information is added.

This is the first i18n in the frontend module, I am not sure if referencing the i18n in the server module is a good way to go but I didn't know how to achieve it otherwise. As there are no tests yet I didn't add any new ones.


cathysarisky:
Hey @SebastianSchroeder  - I just stumbled on this older closed PR, and wanted to make sure you knew that this should work:
```
<title>{{meta_title page=(t "Page %")}}</title>
```

So, a localized theme can handle this title, without needing to make changes on the server.

PR #23873, opened 2025-06-16
Added share modal in Analytics

peterzimon:
ref https://linear.app/ghost/issue/PROD-1856/post-analytics-page-share-modal

- Post share dialog was missing in traffic and post analytics. This PR replicates the existing dialog (with some minor size adjustments) in Shade as a feature component


9larsons:
The display is a bit diff but I think I've got wired up what's needed.

The excerpt isn't truncated is the main piece.

CleanShot 2025-06-17 at 10 22 35

PR #23894, opened 2025-06-17
Added Admin API endpoint for creating comments with member impersonation

55sketch:
ref CON-301

This change allows comment creation via Admin API, opening up the possibility to import comments from other platforms by enabling external tools to create comments on behalf of existing members.


ghost/core/core/server/api/endpoints/comments.js
```diff
@@ -12,6 +13,25 @@ function handleCacheHeaders(model, frame) {
     }
 }
 
+function validateCommentData(data) {
+    if (!data.post_id && !data.parent_id) {
+        throw new errors.ValidationError({
+            message: 'Either post_id (for top-level comments) or parent_id (for replies) must be provided'
+        });
+    }
+}
```
cmraible, line 22: Could we do this validation in the [validation block below](https://github.com/TryGhost/Ghost/pull/23894/files#diff-2a9d263484f99bedd3fe75769f95dc0b103eb5cfbf102a32bbd0ad0271a64735R107-R114), same way it's done for `member_id` and `html`? This works but just for the sake of consistency 👌

ghost/core/core/server/api/endpoints/comments.js
```diff
@@ -67,6 +87,42 @@ const controller = {
...
+            }
+        },
+        permissions: true,
+        async query(frame) {
+            const data = frame.data.comments[0];
+            
+            validateCommentData(data);
+            setupMemberContext(frame, data.member_id);
+            
+            const result = await commentsService.controller.add(frame);
```
cmraible: Not necessarily a blocker, but I think this would be simpler to just call the commentsService functions directly, rather than going through the controller — i.e. instead of `commentsService.controller.add()`, just call `commentsService.commentOnPost()`. 

This would remove the need for the `setupMemberContext` function

ghost/core/core/server/api/endpoints/comments.js
```diff
@@ -67,6 +87,42 @@ const controller = {
...
+            data: {
+                member_id: {
+                    required: true
+                },
+                html: {
+                    required: true
+                }
+            }
+        },
+        permissions: true,
```
cmraible, line 124: Would you want to limit permissions on this endpoint to a certain role, i.e. Owner or Admin? I'm not sure, just something to consider 😃

ghost/core/core/server/api/endpoints/comments.js
55sketch: I think this was done because they aren't both required but one must be present, not sure if the other way allows for that but will check

ghost/core/core/server/api/endpoints/comments.js
55sketch: Interesting. Does the integrations API key have the same role as owner or does it have its own level?

ghost/core/core/server/api/endpoints/comments.js
55sketch: Will look at this 👍

ghost/core/core/server/api/endpoints/comments.js
55sketch: When I wasn't supporting replies and just took the post_id then that was how it was being done

cmraible: Oh I see! I didn't fully process the `&&`

cmraible:
Something else worth confirming: does this send any webhooks or email notifications for comments when it's used by an integration? I'd assume in most cases we wouldn't want that, but that behavior might be automatic, in which case we'd need to carve out some kind of exception.

ghost/core/core/server/api/endpoints/comments.js
cmraible: Did a little bit of digging here — I actually think `permissions: true` should be fine as you have it. Integrations API keys are treated as their own thing. I think you'll also need to add the comments endpoint to the admin middleware allowlist for integration tokens: https://github.com/TryGhost/Ghost/blob/c82b7ed0216229a36cb0a4e101559636c681a3a2/ghost/core/core/server/web/api/endpoints/admin/middleware.js#L44-L70

55sketch: Ah nice - yeah the allowlist change is already in the PR 🙂

ghost/core/core/server/api/endpoints/comments.js
```diff
@@ -12,6 +13,33 @@ function handleCacheHeaders(model, frame) {
     }
 }
 
+function validateCommentData(data) {
+    if (!data.post_id && !data.parent_id) {
+        throw new errors.ValidationError({
+            message: 'Either post_id (for top-level comments) or parent_id (for replies) must be provided'
+        });
+    }
+}
```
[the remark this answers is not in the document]
55sketch: this is an hallucination, that validation function does not exist in CommentsController

55sketch:
> Something else worth confirming: does this send any webhooks or email notifications for comments when it's used by an integration? I'd assume in most cases we wouldn't want that, but that behavior might be automatic, in which case we'd need to carve out some kind of exception.

@cmraible not sure why I couldn't reply to this directly ¯\_(ツ)_/¯

We set the context to 'internal' so it doesn't kick off emails. I've also added a test to ensure this

ghost/core/core/server/api/endpoints/comments.js
55sketch: This change has been made in the latest version 👍

ghost/core/core/server/services/comments/CommentsController.js
cmraible, line 1: Looking good! I think all the changes in this file are not needed now, right?

ghost/core/core/server/services/comments/CommentsService.js
```diff
@@ -216,8 +216,9 @@ class CommentsService {
      * @param {string} member - The ID of the Member to comment as
      * @param {string} comment - The HTML content of the Comment
      * @param {any} options
+     * @param {Date} createdAt - Optional custom created_at timestamp
```
cmraible:
```suggestion
     * @param {Date} [createdAt] - Optional custom created_at timestamp
```

Adding the brackets will tell JSDoc that this parameter is optional, else we'll get linter errors if we try to call this without createdAt

ghost/core/core/server/services/comments/CommentsService.js
```diff
@@ -267,8 +274,9 @@ class CommentsService {
...
```
cmraible:
```suggestion
     * @param {Date} [createdAt] - Optional custom created_at timestamp
```

ghost/core/core/server/services/comments/CommentsService.js
55sketch: oh nice! TIL

ghost/core/core/server/services/comments/CommentsController.js
55sketch: How come? Would we not still want to validate the created_at value? We don't validate inside `commentOnPost` or `replyToComment` as far as I can tell.

(FWIW this change was added in response to a coderabbit review comment, so I'm not married to it - that being said it seems to make sense to have it)

cmraible: Since we changed the endpoint to call the CommentsService methods directly for creating the comment, we're not using the `CommentsController.add()` anymore to create the comment — and the validation is handled up here: https://github.com/TryGhost/Ghost/pull/23894/files#diff-2a9d263484f99bedd3fe75769f95dc0b103eb5cfbf102a32bbd0ad0271a64735R24-R41.

This `CommentsController.add()` is only used in the `comments-members.js` endpoint, and I don't think we want members to be able to specify the timestamp of their comments. Not a blocker, just not a necessary change at the moment.

cmraible (approved):
LGTM

cmraible (approved):

PR #23902, opened 2025-06-17
Added `trafficAnalyticsTracking` flag

cmraible:
ref https://linear.app/ghost/issue/PROD-2075/add-flag-to-enable-tracking-only

For testing event ingestion, it can sometimes be useful to enable event tracking on a site without enabling the full traffic analytics feature set. This new flag will do just that — it enables the `ghost-stats.js` script, without enabling the full suite of features that is gated by the `trafficAnalytics` flag.


cmraible:
Closing to see if a new branch will fix the caching...

PR #23944, opened 2025-06-19
Fixed notification test failing

ErisDS:
- This is failing because the code is in correctly creating a  comparison between 6.0.0 and 6.0.0-alpha.0
- Used semver tools and simplified logic to ensure that version comparisons are more robust


ErisDS:
Closing in favour of https://github.com/TryGhost/Ghost/pull/23955

PR #23949, opened 2025-06-19
Expanded test coverage for node renderers

kevinansfield:
no issue

When node renderers were moved over from Koenig, the tests were set up with minimal cases to ensure the relevant renderer was running. This was fine because there were no substantial changes to renderers and tests for edge-cases still existed in Koenig. Now that we're planning on deleting the now-duplicated renderers from Koenig we'd like to ensure we're not dropping test coverage.

- fixed issue with `assertPrettifiesTo()` test helper dropping comments
- moved renderer tests over from Koenig, updated to work with Ghost's render and test setup
- updated callout renderer to use a string template in place of manual DOM building


ErisDS:
Cleaning up older PRs 💅

Issue #23924, opened 2025-06-19
🔥 Breaking Changes for 6.0


ErisDS:
We're working towards releasing Ghost 6.0. As with all major version bumps, there will be some breaking changes, especially around the API. 

This issue is a placeholder where we'll keep track of all the breaking changes we are making. The intention is to have one place to point to - but also make this transparent for the community.

## Breaking changes 
This list isn't final until 6.0 is shipped.

1. Dropping support for Node.js v18
v18 is EOL now, and no longer supported upstream, we're updating to reflect this
2. Dropping support for Node.js v20 
Node.js v22 includes some important new features that we want to use to make building Ghost easier - including TypeScript and ESM support. Maintaining backwards compatibility with v20 in this case doesn't feel super valuable given the advantages of being v22 only and the timing of the 6.0 release means we get to call it out loud as a breaking change.  
3. Return max 100 results from Ghost's APIs (deprecating `?limit=all`)
Having the ability to return all results made sense when Ghost was a small blogging tool. Now it has significant performance issues. `?limit=all` will be deprecated, and if used will only return max 100 results. Setting limit to a number higher than 100 will also result in a max of 100 results.
4. Removing AMP
Our AMP feature has been deprecated for quite some time. It will be removed in 6.0
5. Removing `GET /ghost/api/admin/session/` endpoint - this is an unused endpoint. Use `GET /ghost/api/admin/users/me/` instead.
6. Removing `created_by` & `updated_by` from everywhere
These fields were deprecated a while ago and are not used in any meaningful way in the codebase. Audit logging is done via the `actions` system which replaced usage of these fields
7. Ensuring that all users have an `ObjectID` for their ID. This means migrating any existing users who had an ID of `1` to an `ObjectID`, and changing the system so that the first user created no longer has a hardcoded ID of `1`. This allows us to remove codepaths that had to check whether the ID was `1` or an `ObjectID` (and allow us to properly type this field) as well as remove any assumptions that the owner ID is always `1`.  **It is no longer safe to assume the user with an ID of `1` is the default/owner user**
8. Stopped serving files without extensions from the theme root folder & correctly 404 for missing theme assets.

cathysarisky:
@ErisDS , it might be kind to self hosters to update what's here: https://ghost.org/docs/faq/node-versions/

A version of node that'll be discontinued this summer probably shouldn't be the _recommended_ version today?  Are there any blockers to recommending 22 now, while still on Ghost 5.x?

[...]

PR #23981, opened 2025-06-20
🔥 Removed AMP app connection to core app, Added redirectAmpUrls middleware

troyciesco:
ref https://github.com/TryGhost/Ghost/issues/23924 
closes https://linear.app/ghost/issue/PROD-2115/unload-amp-app-from-express

- First step for removing AMP support in Ghost 6.0 is no longer loading the AMP app
- The AMP app was responsible for redirecting /amp routes, so redirectAmpUrls was added as a middleware to take over that task
- redirectAmpUrls was added as part of the pretty-urls middleware sequence


ghost/core/core/frontend/helpers/ghost_head.js
```diff
@@ -304,94 +299,92 @@ module.exports = async function ghost_head(options) { // eslint-disable-line cam
         head.push('<link rel="alternate" type="application/rss+xml" title="' +
             escapeExpression(meta.site.title) + '" href="' +
             escapeExpression(meta.rssUrl) + '">');
-        // no code injection for amp context!!!
-        if (!_.includes(context, 'amp')) {
```
troyciesco, line 308: The diff for this file looks weird, but I just removed the `if` here since we don't need to check for amp anymore, and let eslint fix the indenting

ghost/core/core/frontend/meta/canonical-url.js
```diff
@@ -12,13 +12,7 @@ function getCanonicalUrl(data) {
         return data.tag.canonical_url;
     }
 
-    let url = urlUtils.urlJoin(urlUtils.urlFor('home', true), getUrl(data, false));
-
-    if (url.indexOf('/amp/')) {
```
troyciesco, line 17: i'm _pretty_ sure we don't need to do this here anymore. This function gets called by get-meta.js, and even when i visit an amp route the data passed to getCanonicalUrl doesn't have /amp in it

ghost/core/core/frontend/meta/url.js
```diff
@@ -2,16 +2,6 @@ const urlUtils = require('../../shared/url-utils');
 const urlService = require('../../server/services/url');
 const {checks} = require('../services/data');
 
-// This cleans the url from any `/amp` postfixes, so we'll never
-// output a url with `/amp` in the end, except for the needed `amphtml`
-// canonical link, which is rendered by `getAmpUrl`.
-function sanitizeAmpUrl(url) {
```
troyciesco, line 8: Same thing here as in canonical-url.js in that i don't think we need it anymore. /amp is removed before this function gets called anywhere

ghost/core/core/server/web/shared/middleware/sanitize-amp-urls.js
```diff
@@ -0,0 +1,36 @@
...
+ *   /welcome/amp/?q=1  -> /welcome/?q=1
+ *
+ * @param {import('express').Request} req - Express request object
+ * @param {import('express').Response} res - Express response object  
+ * @param {import('express').NextFunction} next - Express next function
+ * @returns {void}
+ */
+function sanitizeAmpUrls(req, res, next) {
+    const ampPattern = /\/amp\/?$/i;
+    const url = new URL(req.url, 'http://localhost');
```
troyciesco: localhost looks weird to me but i don't need a host, just wanted to be able to get the pathname and the query string easily

ghost/core/test/unit/frontend/helpers/__snapshots__/ghost_head.test.js.snap
```diff
@@ -306,7 +305,11 @@ Object {
     </script>
 
     <meta name=\\"generator\\" content=\\"Ghost 0.3\\">
-    <link rel=\\"alternate\\" type=\\"application/rss+xml\\" title=\\"Ghost\\" href=\\"http://localhost:65530/rss/\\">",
+    <link rel=\\"alternate\\" type=\\"application/rss+xml\\" title=\\"Ghost\\" href=\\"http://localhost:65530/rss/\\">
+    
+    <script defer src=\\"https://cdn.jsdelivr.net/ghost/sodo-search@~[[VERSION]]/umd/sodo-search.min.js\\" data-key=\\"xyz\\" data-styles=\\"https://cdn.jsdelivr.net/ghost/sodo-search@~[[VERSION]]/umd/main.css\\" data-sodo-search=\\"http://127.0.0.1:2369/\\" data-locale=\\"en\\" crossorigin=\\"anonymous\\"></script><style>:root {--ghost-accent-color: #123456;}</style>
```
troyciesco: @ErisDS and @jloh only tagging you here because you mentioned sodo-search today and i'm not sure what this is and if it makes sense the regenerating a snapshot would add it 😅

ghost/core/core/shared/config/overrides.json
```diff
@@ -12,10 +12,7 @@
...
     },
     "apps": {
-        "internal": [
-            "private-blogging",
-            "amp"
-        ]
+        "internal": ["private-blogging"]
     },
     "slugs": {
         "protected": ["ghost", "rss", "amp"]
```
troyciesco, line 18: i think it makes sense to leave amp as a protected slug, otherwise there's a lot more edge cases we'd need to check (like tags)

ghost/core/test/unit/frontend/helpers/__snapshots__/ghost_head.test.js.snap
troyciesco: discussed elsewhere, it makes sense that these show up after updating ghost_head.js to no longer check for amp when doing code injection. ill update the snapshot to remove unneeded tests.

ghost/core/test/unit/frontend/helpers/__snapshots__/ghost_head.test.js.snap
```diff
@@ -1,114 +1,10 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`{{ghost_head}} helper CAPTCHA does not return CAPTCHA script when disabled 1 1`] = `
```
troyciesco, line 3: this diff ended up super ugly but i removed a few amp-related tests and it cleaned up other old tests that aren't used anymore as well

ghost/core/core/server/web/shared/middleware/sanitize-amp-urls.js
lsinger: As long as one doesn't rely on making actual HTTP requests to a domain, `example.com` is the best placeholder to use.

ghost/core/core/server/web/shared/middleware/sanitize-amp-urls.js
```diff
@@ -0,0 +1,36 @@
...
+ * Example:
+ *   /welcome/amp/      -> /welcome/
+ *   /welcome/amp/?q=1  -> /welcome/?q=1
+ *
+ * @param {import('express').Request} req - Express request object
+ * @param {import('express').Response} res - Express response object  
+ * @param {import('express').NextFunction} next - Express next function
+ * @returns {void}
+ */
+function sanitizeAmpUrls(req, res, next) {
```
kevinansfield: Would naming this `redirectAmpUrls()` better match its behaviour?

ghost/core/test/e2e-frontend/default_routes.test.js
```diff
@@ -185,53 +185,7 @@ describe('Default Frontend routing', function () {
...
-                    .expect(404)
-                    .expect(/Page not found/)
-                    .expect(assertCorrectFrontendHeaders);
-            });
-        });
-
+        // AMP is no longer supported as of v6.0, so we only check the case in which it's disabled
         describe('AMP Disabled', function () {
             beforeEach(function () {
                 sinon.stub(settingsCache, 'get').callsFake(function (key, options) {
```
kevinansfield: I wonder if it's worth removing this stub of the `'amp'` setting entirely to avoid any confusion about an "amp" setting possibly existing?

ghost/core/test/legacy/site/frontend.test.js
```diff
@@ -227,20 +211,7 @@ describe('Frontend Routing', function () {
...
-                            .expect('Cache-Control', testUtils.cacheRules.noCache)
-                            .expect(404)
-                            .expect(/Post not found/)
-                            .end(doEnd(done));
-                    });
-                });
+                // AMP is no longer supported as of v6.0, so we only check the case in which it's disabled
                 describe('amp disabled', function (){
                     beforeEach(function () {
                         sinon.stub(settingsCache, 'get').withArgs('amp').returns(false);
```
kevinansfield: Similar to comment above about removing any stubbing of 'amp' settings.

ghost/core/test/e2e-frontend/default_routes.test.js
kevinansfield: 🤦 of course this setting still exists at the moment which is why it's here. Ignore this comment other than it being a reminder for later cleanup 😄

ghost/core/core/server/web/shared/middleware/sanitize-amp-urls.js
troyciesco: oooh yeah i like that better, then it doesn't look like i was testing locally and forgot to change it or something 😅

ghost/core/core/server/web/shared/middleware/sanitize-amp-urls.js
troyciesco: yeah that makes sense, i'll change it!

ghost/core/test/e2e-frontend/default_routes.test.js
troyciesco: actually you're right - the setting will still exist after this PR, but it's not needed anymore for the tests here and in frontend.test.js, so i'll remove

PR #24005, opened 2025-06-21
Updated final content API tests to latest framework

ErisDS:
ref https://www.notion.so/ghost/End-to-end-Testing-6a2ef073b1754b18aff42e24a632a007

- Author and Tag tests were still using the old framework for some reason
- Uses more minimal assertions in favour of snapshots but keeps custom assertions for important logic
- This should make it easier to migrate in the legacy author and tag tests
- Slowly consolidating all our API testing


ErisDS:
More to do here than I realised. Will come back to it later.

Issue #22077, opened 2025-01-29
`members.uuid` is marked as `nullable` in database schema


nikwen:
### Issue Summary

https://github.com/TryGhost/Ghost/pull/11341 introduced UUIDs for members.

It introduced a migration to backfill UUIDs for existing members, and introduced defaults so that each new member would have a UUID.

It seems to me like the [database schema](https://github.com/TryGhost/Ghost/blob/048f17757ce7f08f0e7ab7135d1ac5b59b395db8/ghost/core/core/server/data/schema/schema.js#L428) should be adjusted to mark the `members.uuid` column as `nullable: false`.

When https://github.com/TryGhost/Ghost/pull/14600 introduced UUIDs for newsletters, it included a [migration](https://github.com/TryGhost/Ghost/commit/756f86dbdc1e67cc4eb1158e484a944ae3c1f076#diff-2885b5f206c932aa709bb4c0b5f48832bc2e3fbfbd71da6c82c1f37f841d53cd) to drop `nullable` from the new `newsletters.uuid` column. A similar migration could probably be created for the members table.

### Steps to Reproduce

1. Look at the schema here: https://github.com/TryGhost/Ghost/blob/048f17757ce7f08f0e7ab7135d1ac5b59b395db8/ghost/core/core/server/data/schema/schema.js#L428

### Ghost Version

5.108.1

### Node.js Version

20.18.1

### How did you install Ghost?

Irrelevant for the issue

### Database type

MySQL 5.7

### Relevant log / error output

```shell

```

### Code of Conduct

- [x] I agree to be friendly and polite to people in this repository

sagzy:
Great find, and very valid! Thank you for opening the issue 🙌

PriyaDharsini2006:
can i solve the issue , i will raise the pr immediately !!

PR #24015, opened 2025-06-23
Added migration to drop nullable on members UUID

ErisDS:
closes https://github.com/TryGhost/Ghost/issues/22077
closes https://linear.app/ghost/issue/PROD-1611/breaking-change-schema-oss-issue-membersuuid-is-marked-as-nullable-in 
ref https://github.com/TryGhost/Ghost/pull/11341/commits/4095178243638636d19eb1ef0a42738098f99d29
ref https://github.com/TryGhost/Ghost/commit/12f569ebf9aa864deac18b57eb7c77eaccce8dc7

- The migration to add a UUID was originally added in Ghost 3.1
- Therefore, we're going to rerun it first, although we expect this to be a no-op in everything other than absolute edge cases
- Verified that there are no instances of null member uuids anywhere on Pro
- This then sets nullable to false in the DB after
- The net result of this change is that the schema matches the requirements / expectations about the world

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/6.0/2025-06-23-09-49-25-add-missing-member-uuids.js
+++ b/ghost/core/core/server/data/migrations/versions/6.0/2025-06-23-09-49-25-add-missing-member-uuids.js
@@ -1,0 +1,22 @@
+// For information on writing migrations, see https://www.notion.so/ghost/Database-migrations-eb5b78c435d741d2b34a582d57c24253
+
+const logging = require('@tryghost/logging');
+const uuid = require('uuid');
+
+// For DML - data changes
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const membersWithoutUUID = await knex.select('id').from('members').whereNull('uuid');
+
+        logging.info(`Adding uuid field value to ${membersWithoutUUID.length} members.`);
+
+        // eslint-disable-next-line no-restricted-syntax
+        for (const member of membersWithoutUUID) {
+            await knex('members').update('uuid', uuid.v4()).where('id', member.id);
+        }
+    },
+    // down is a no-op
+    async function down() {}
+);
--- a/ghost/core/core/server/data/migrations/versions/6.0/2025-06-23-10-03-26-members-nullable-uuid.js
+++ b/ghost/core/core/server/data/migrations/versions/6.0/2025-06-23-10-03-26-members-nullable-uuid.js
@@ -1,0 +1,5 @@
+const {createDropNullableMigration} = require('../../utils');
+
+// Running drop nullable migration without foreign key checks disabled
+// This is because uuid isn't used as a foreign key anywhere in the schema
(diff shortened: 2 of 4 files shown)


ghost/core/core/server/data/migrations/versions/6.0/2025-06-23-09-49-25-add-missing-member-uuids.js
```diff
@@ -0,0 +1,22 @@
...
+const {createTransactionalMigration} = require('../../utils');
+
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        const membersWithoutUUID = await knex.select('id').from('members').whereNull('uuid');
+
+        logging.info(`Adding uuid field value to ${membersWithoutUUID.length} members.`);
+
+        // eslint-disable-next-line no-restricted-syntax
+        for (const member of membersWithoutUUID) {
```
allouis, line 16: Do we have an idea of how many members this will be? It _should_ be zero as far as I understand - but it would be good to check this as we're doing a loop here and making a query for each item

Alternatively we could create the uuids up front and batch insert them

ghost/core/core/server/data/migrations/versions/6.0/2025-06-23-10-03-26-members-nullable-uuid.js
```diff
@@ -0,0 +1,4 @@
+const {createDropNullableMigration} = require('../../utils');
+
+// We need to disable foreign key checks because if MySQL is missing the STRICT_TRANS_TABLES mode, we cannot revert the migration
+module.exports = createDropNullableMigration('members', 'uuid', {disableForeignKeyChecks: true});
```
allouis: The `disableForeignKeyChecks` option only applies to the `up` no the `down` - so I think the comment is incorrect.

The `uuid` column isn't used as a foreign key anywhere, so do we need this flag at all?

As far as I know the 6.0 migrations are irreversible, so we don't need to provide a `down` or have the ability to revert this migration 🤔 Maybe we want to be explicit with that?

```ts
const migration = createDropNullableMigration('members', 'uuid');

module.exports = {
  ...migration,
  config: {
    ...migration.config,
    reversible: false,
  },
  async down() {
    throw new Error('Irreversible migration');
  }
};
```

allouis (commented):
Approach looks good, I left two comments, neither of them are big deals though IMO

I think we're both pretty confident that the loop is going to be essentially a noop - so I'm not too concerned about that, though I would like to know if we need to do it at all?

The `disableForeignKeyChecks` flag/comment confused me a bit - I think it would be good to correct the comment/understand why we're doing what we're doing!

ghost/core/core/server/data/migrations/versions/6.0/2025-06-23-09-49-25-add-missing-member-uuids.js
ErisDS: On pro it's zero - this is only here for absolute edge cases in the wider world, and shouldn't cause any trouble

PR #24017, opened 2025-06-23
Updated Admin tests to fail if any limit=all request gets through to API

kevinansfield:
ref https://linear.app/ghost/issue/PROD-1605/

- Ghost 6.0 will introduce a breaking change where the API `limit` parameter will be capped to 100 results
- we're updating all of our own apps to restrict requests to `limit=100` as a max and have no usage of `limit=all`
- this change ensures our Admin tests will fail if anything slips through and reaches our mocked API

### DIFF
--- a/ghost/admin/mirage/routes-test.js
+++ b/ghost/admin/mirage/routes-test.js
@@ -77,6 +77,41 @@ export default function () {
         db.tags.all().remove();
     }, 204);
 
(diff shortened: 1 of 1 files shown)


kevinansfield:
Already merged in 8bf0cce3601b55d831d347bd628fb6fb1be1dd08

PR #24031, opened 2025-06-24
Removed redundant `internalUser` from events bookshelf plugin

mike182uk:
ref https://linear.app/ghost/issue/PROD-2130

Removed redundant `internalUser` from events bookshelf plugin and inlined the value where it is being used (which is only in the events plugin). This is to prevent further usage of this value where the owner user ID is assumed to be `1`


mike182uk:
Closing in favour of https://github.com/TryGhost/Ghost/pull/24088

PR #24033, opened 2025-06-24
Added migration for migrating users with hardcoded ID of `1` to an `ObjectID`

mike182uk:
ref https://linear.app/ghost/issue/PROD-2127

Added migration for migrating user with hardcoded ID of `1` to an `ObjectID`. We want to perform this migration on existing installations so we that we do not need to add codebase checks in the future for if a user id is `1` or an `ObjectID`


ghost/core/core/server/data/migrations/versions/6.0/2025-06-24-09-19-42-use-object-id-for-owner-user.js
```diff
@@ -0,0 +1,167 @@
...
+            slug,
+            email
+        });
+
+        logging.info(`Restored owner user slug and email: ${slug}, ${email}`);
+
+        logging.info(`Successfully updated owner user ID from ${LEGACY_OWNER_USER_ID} to ${newId}`);
+    },
+
+    async function down(knex) {
```
mike182uk: Since this is being done as part of a major version, do we even need this? (i.e we can't rollback to the previous state anyway)

ghost/core/core/server/data/migrations/versions/6.0/2025-06-24-09-19-42-use-object-id-for-owner-user.js
```diff
@@ -0,0 +1,167 @@
...
+        // without violating foreign key constraints. We also need to use a
+        // temporary email and slug to avoid duplicate constraint errors
+        const clonedOwnerUser = {
+            ...currentOwnerUser,
+            id: newId,
+            slug: temporarySlug,
+            email: temporaryEmail
+        };
+
+        await knex('users').insert(clonedOwnerUser);
```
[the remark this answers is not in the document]
mike182uk: Yep this is highly unlikely 👍

ghost/core/core/server/data/migrations/versions/6.0/2025-06-24-09-19-42-use-object-id-for-owner-user.js
ErisDS: No, there's no expectation that we have a down function

ghost/core/core/server/data/migrations/versions/6.0/2025-06-24-09-19-42-use-object-id-for-owner-user.js
```diff
@@ -0,0 +1,196 @@
...
+            return;
+        }
+
+        const isOwner = await knex('roles_users')
+            .where('user_id', LEGACY_OWNER_USER_ID)
+            .where('role_id', ownerRole.id)
+            .first();
+
+        if (!isOwner) {
+            logging.warn(`User with ID ${LEGACY_OWNER_USER_ID} is not the owner, skipping migration`);
```
ErisDS: If the ownership has been transferred, there may still be a user with ID 1 in the system, which I think we would still want to clean up?

mike182uk: 👍 I've updated this now to remove the explicit owner checks and just migrate any user with the hardcoded ID of `1`

ErisDS (approved):

PR #24038, opened 2025-06-24
Added missing models to action log

ErisDS:
ref https://linear.app/ghost/issue/PROD-1597/breaking-change-schema-properly-remove-x-by-columns-in-favour-of-using

- These models should all appear in the action log / history

TODO:
- [x] verify what happens on member import
- [x] fix the output for invites

### DIFF
--- a/ghost/core/core/server/models/base/plugins/actions.js
+++ b/ghost/core/core/server/models/base/plugins/actions.js
@@ -107,11 +107,14 @@ module.exports = function (Bookshelf) {
                 }
             }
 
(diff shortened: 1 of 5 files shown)


ErisDS:
CleanShot 2025-06-24 at 11 32 31

This is working, invites appear a little weird so far

ErisDS:
Working and slightly better UI:

CleanShot 2025-06-26 at 21 11 22@2x

ErisDS:
P.S. verified member imports don't show up at all yet. That's something to fix later :)

PR #24048, opened 2025-06-24
Added site_uuid to content API settings endpoint

ErisDS:
ref https://linear.app/ghost/issue/PROD-197/ghost-site-uuid-implementation

- site_uuid is a public, read-only setting
- it should be included in the "public" api/content/settings endpoint as well as the api/admin/site endpoint
- this makes it available everywhere the getPublic() method is used
- the main reason for making this change is to make this available in the explore ping service


cmraible (approved):
LGTM!

PR #24055, opened 2025-06-24
🔥 Removed AMP from core

troyciesco:
ref https://github.com/TryGhost/Ghost/issues/23924
closes https://linear.app/ghost/issue/PROD-2116/completely-remove-amp-app-from-core

- deletes the entire AMP app and its associated tests
- no longer includes AMP as a possible context type and removes any checks for it
- removes other tests related to AMP posts
- ignores AMP as a valid setting in SettingsImporter
- removes `amperize` from `package.json`


ghost/core/test/integration/importer/v2.test.js
```diff
@@ -954,7 +954,7 @@ describe('Importer', function () {
             const exportData = exportedBodyV2().db[0];
 
             exportData.data.settings[0] = testUtils.DataGenerator.forKnex.createSetting({
-                key: 'amp',
+                key: 'portal_button',
```
troyciesco, line 911: switched this to portal_button so the test still covered everything

ghost/core/test/integration/importer/v2.test.js
```diff
@@ -966,27 +966,27 @@ describe('Importer', function () {
...
                     imported.problems.length.should.eql(0);
-                    return models.Settings.findOne(_.merge({key: 'amp'}, testUtils.context.internal));
+                    return models.Settings.findOne(_.merge({key: 'portal_button'}, testUtils.context.internal));
                 })
                 .then(function (result) {
                     result.attributes.value.should.eql(true);
                     return models.Settings.findOne(_.merge({key: 'is_private'}, testUtils.context.internal));
                 })
-                .then((result) => {
+                .then(function (result) {
```
troyciesco, line 929: the chain of `then` is a bit hard to read but it's the patterned used through the whole file. i did tidy these to not use arrow functions anymore, since the rest of the file doesn't either

kevinansfield (approved):

PR #24061, opened 2025-06-24
Added Content API endpoint to index posts for search

sagzy:

- added `GET /ghost/api/content/posts/compact` endpoint that returns a maximum of 10k posts, with limited fields 
- this endpoint will be used to index posts for search, in Sodo Search and Admin


ghost/core/core/server/api/endpoints/posts-public.js
```diff
@@ -176,6 +181,41 @@ const controller = {
...
+                    fields: BROWSE_COMPACT_DEFAULTS.fields
+                },
+                auth: generateAuthData(frame),
+                method: 'browseCompact'
+            };
+        },
+        options: [
+            'filter',
+            'order'
+        ],
```
sagzy, line 208: Drop these options

ghost/core/core/server/api/endpoints/posts-public.js
```diff
@@ -12,6 +12,11 @@ const messages = {
     postNotFound: 'Post not found.'
 };
 
+const BROWSE_COMPACT_DEFAULTS = {
+    limit: 10000,
```
sagzy, line 16: Overridable by config for Pro

PR #24075, opened 2025-06-25
Removed 4.x migrations, added final error migration 

ErisDS:
ref https://github.com/TryGhost/Ghost/commit/145dc4651aa9de71d23b25c27a60acc2c10990db  
ref https://github.com/TryGhost/Ghost/commit/b788604e15f676957f8cb603978f26a30bc80b1e  
ref https://ghost.org/docs/update-major-version/. 

- In order to upgrade to 6.x you need to be on: a) the very latest 4.x
- b) any 5.x version
- We clean out migrations from the previous migration as it allows us to
keep the codebase a little cleaner
- And means that we don't have to maintain unmanageable levels of
backwards compatibility across migration utils etc
- We add a "final migration" to the very last minor of 4.x which ensures
that attempting to update to 6.0 from a too-early version will error
with a clear message telling you what steps to take instead.
@[ErisDS](https://github.com/TryGhost/Ghost/commits?author=ErisDS)
ErisDS authored yesterday


Ghost-Slimer:
Deployed to **staging** with ID: `3566`

PR #24088, opened 2025-06-25
🎨 Ensured owner user is created with an `ObjectID`

mike182uk:
ref https://linear.app/ghost/issue/PROD-1594

This changeset updates the system to ensure that the owner user is created with an `ObjectID` instead of being hardcoded to `1`, and there are no assumptions made that the owner user is `1`:

- A simple placeholder replacement system has been introduced into the fixtures JSON to allow dynamic values to be used at runtime (for the generation of the owner user id)
  - I did toy with a few different ideas (swapping the format to a `js` func instead etc). but this snowballed into much bigger changes that did not warrant the complexity involved - i.e we don't want to be maintaining 2 different formats for the fixtures
  - The lazy initialisation of the fixtures may seem a bit strange at first, but this is to get around the fact that `models` may not always be initialised at the time they are used, depending on the environment and order of loading (i.e in the test environment, the fixture manager may be loaded before the model initialisation / the Knex migrator hooks do not get executed). This lazy initialisation means we can keep the ID generation logic for the user with the appropriate model, and we do not have to do `models.init` from within the fixture manager
- The `contextUser` property provided by the `user-type` bookshelf plugin has been updated to be async, and instead of returning hardcoded `1`, it now resolves the owner user ID from the db
  - The resolved value is cached on first retrieval and cleared when an ownership transfer occurs
  - All the call sites of this property have also been updated to be async (`post` model, `author` relation model)
- The `UpdateCheckService` has been updated to not assume that the first retrieved user is the owner user, and instead analyse the roles of those users to determine which is the owner user
- The test suite has been updated to use a fixed owner user ID that is an `ObjectID` (`5951f5fc0000000000000000`)
- The data generation tooling has been updated to retrieve the owner user ID from the db rather than assuming that it is `1`


ghost/core/core/server/models/base/bookshelf.js
```diff
@@ -54,7 +54,14 @@ ghostBookshelf.plugin(require('./plugins/bulk-operations'));
...
 
-ghostBookshelf.plugin(require('./plugins/user-type'));
+ghostBookshelf.plugin(require('./plugins/user-type'), {
+    resolveIntegrationUserId: function resolveIntegrationUserId(options) {
+        return ghostBookshelf.model('User').getOwnerId(options);
+    },
+    resolveInternalUserId: function resolveInternalUserId(options) {
+        return ghostBookshelf.model('User').getOwnerId(options);
+    }
+});
```
ErisDS, line 64: DI FTW

ghost/core/core/server/models/user.js
```diff
@@ -774,6 +774,35 @@ User = ghostBookshelf.Model.extend({
...
+        set(value) {
+            this.value = value;
+        },
+        get() {
+            return this.value;
+        },
+        clear() {
+            this.value = null;
+        }
+    },
```
ErisDS, line 788: Really like this idea. Feels like once this is done, this will need a bit of documentation in the [codex](https://www.notion.so/ghost/Codex-188dc2b333c94ab08849e00016389d86) & a message to all the devs about this change.

ghost/core/test/utils/fixtures/constants.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+    OWNER_USER_ID: '5951f5fc0000000000000000'
```
ErisDS: This is definitely a non-blocking concern, but I am wondering if we need to add this new system here. 

In the old test framework, we pull the hardcoded IDs out like this:

https://github.com/TryGhost/Ghost/blob/main/ghost/core/test/e2e-api/admin/actions.test.js#L96

In the new test framework we have a slightly nicer helper system for it:

https://github.com/TryGhost/Ghost/blob/main/ghost/core/test/e2e-api/admin/api-tokens.test.js#L163

What you've done here is clean and there's nothing wrong with it - but the question on my mind is are we going to pull out every hard-coded ID in the data-genrator and use it consistently, or is this just for the owner ID? In the latter case, I'd just have fixtures/owner-id.js rather than fixtures/constants.js.

ErisDS (approved):
All the key changes look great. I have one small question about the new constants system in the test fixtures :)

ghost/core/test/utils/fixtures/constants.js
mike182uk: Yeh i get ya

Unfortunately we can't use the e2e fixture manager for this all the time as the value is used in non e2e tests also

How do you feel about https://github.com/TryGhost/Ghost/pull/24088/commits/e6a55e1a2081796f5fe2aad5b0e7500c90131005 ? Removes the constant and uses dynamic references to owner user fixture everywhere its needed via either the e2e fixture manager, or the data generator fixtures themselves (which are used to create the fixtures) (the browser tests already do this is  some spots). This means we can get rid of the constants file and not tempt anybody else put things in there 😅

PR #24095, opened 2025-06-25
🐛 Add missing pagination for staff invitations in admin settings

jonatansberg:
The existing admin UI only handled pagination for users but not invites. 

This PR updates the `useBrowseInvites` hook to mirror the pagination handling in `useBrowseUsers`, enabling the same behaviour for invites. The staff user list component was also updated to accommodate these changes.

### Motivation 

My first naive attempt at solving this was to just bump the page size to 100. In many ways that would probably be enough, but it would still leave the bug in place. On the off chance that some large media conglomerate wants to set up Ghost and invite the entire org fixing things properly seemed like the better way to go. 

Since `limit=all` seems to be going away in Ghost 6 I opted for just doing things in the same way for invites as for the other users. To me this seemed like the better trade-off compared to removing or ignoring the limit parameter on the backend which would make things less consistent across endpoints. 

### Testing
I have added a Playwright test that verifies that the pagination works and updated the moved backend responses.

Fixes https://github.com/TryGhost/Ghost/issues/23946

---

### DIFF
--- a/apps/admin-x-framework/src/api/invites.ts
+++ b/apps/admin-x-framework/src/api/invites.ts
@@ -1,4 +1,6 @@
-import {Meta, createMutation, createQuery} from '../utils/api/hooks';
+import {InfiniteData} from '@tanstack/react-query';
+import {Meta, createMutation, createInfiniteQuery} from '../utils/api/hooks';
+import {deleteFromQueryCache, insertToQueryCache} from '../utils/api/updateQueries';
 
 export interface UserInvite {
     created_at: string;
@@ -18,10 +20,26 @@ export interface InvitesResponseType {
 
 const dataType = 'InvitesResponseType';
 
-export const useBrowseInvites = createQuery<InvitesResponseType>({
+export const useBrowseInvites = createInfiniteQuery<InvitesResponseType & {isEnd: boolean}>({
     dataType,
     path: '/invites/',
-    permissions: ['Owner', 'Administrator']
+    permissions: ['Owner', 'Administrator'],
+    defaultSearchParams: {limit: '100', include: 'roles'},
+    defaultNextPageParams: (lastPage, otherParams) => ({
+        ...otherParams,
+        page: (lastPage.meta?.pagination.next || 1).toString()
+    }),
+    returnData: (originalData) => {
+        const {pages} = originalData as InfiniteData<InvitesResponseType>;
+        const invites = pages.flatMap(page => page.invites);
+        const meta = pages[pages.length - 1].meta;
+
+        return {
+            invites,
+            meta,
+            isEnd: meta ? meta.pagination.pages === meta.pagination.page : true
+        };
+    }
 });
 
 export const useAddInvite = createMutation<InvitesResponseType, {email: string, roleId: string}>({
@@ -40,13 +58,7 @@ export const useAddInvite = createMutation<InvitesResponseType, {email: string, 
         dataType,
         emberUpdateType: 'createOrUpdate',
         // Assume that all invite queries should include this new one
-        update: (newData, currentData) => (currentData && {
-            ...(currentData as InvitesResponseType),
-            invites: [
-                ...((currentData as InvitesResponseType).invites),
-                ...newData.invites
-            ]
-        })
+        update: insertToQueryCache('invites')
     }
 });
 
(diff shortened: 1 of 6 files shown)


cathysarisky:
Hey @jonatansberg !  I'm triggering the test workflow on this, in case there's anything actionable, while you wait on dev team review.

kevinansfield:
@jonatansberg would you mind rebasing this on top of `main`? Thankyou 🙏

kevinansfield:
Hey @jonatansberg 👋 

Overall approach here looks good 🙌 

> I have't written any test for this (yet) as there weren't any obvious place to do so. Adding a Playwright test seemed a bit heavy handed.

First thing will be making sure the existing playwright tests continue to pass - currently they are failing. The failing invite tests make sense I think because the mocked API requests haven't been updated to match the `createInfiniteQuery` request but it's not clear why the theme tests are also failing 😕 As for testing the "Load more" behaviour, a Playwright test would be the go-to.

> Should I bump the versions of packages somewhere?

No, that is handled as a separate shipping step after the change has been merged 🙂

kevinansfield (changes requested):
Left feedback in previous comment https://github.com/TryGhost/Ghost/pull/24095#issuecomment-3150394996

jonatansberg:
@kevinansfield Thank you for the review! I couldn't get the Playwright tests to run locally before submitting this PR, and after that I've been offline and on vacation. I'll have another look in the coming days :)

> As for testing the "Load more" behaviour, a Playwright test would be the go-to.

Just so that I understand, you think a Playwright test would be called for in this case? If so I can go ahead and add that.

apps/admin-x-framework/src/api/invites.ts
```diff
@@ -41,10 +58,16 @@ export const useAddInvite = createMutation<InvitesResponseType, {email: string,
...
+            ...(currentData),
+            pages: [
+                ...(currentData as {pages: InvitesResponseType[]}).pages.slice(0, -1),
+                {
+                    ...(currentData as {pages: InvitesResponseType[]}).pages[((currentData as {pages: InvitesResponseType[]}).pages).length - 1],
+                    invites: [
+                        ...(currentData as {pages: InvitesResponseType[]}).pages[((currentData as {pages: InvitesResponseType[]}).pages).length - 1].invites,
+                        ...newData.invites
+                    ]
+                }
```
kevinansfield: Could this use our [`updateQueryCache` helper](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-framework/src/utils/api/updateQueries.ts#L30-L60)?

apps/admin-x-framework/src/api/invites.ts
```diff
@@ -57,8 +80,12 @@ export const useDeleteInvite = createMutation<unknown, string>({
         dataType,
         emberUpdateType: 'delete',
         update: (_, currentData, id) => ({
-            ...(currentData as InvitesResponseType),
-            invites: (currentData as InvitesResponseType).invites.filter(invite => invite.id !== id)
+            ...(currentData as {pages: InvitesResponseType[]}),
+            pages: (currentData as {pages: InvitesResponseType[]}).pages.map((page: InvitesResponseType) => ({
+                ...page,
+                invites: page.invites.filter(invite => invite.id !== id)
+            }))
```
kevinansfield: Similar to the above comment, we have a [`deleteFromQueryCache` helper](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-framework/src/utils/api/updateQueries.ts#L62-L86) that looks like it could be used here.

I could be wrong as I'm not particularly context loaded on any Admin-X stuff, it's just that the manual updates stuck out when comparing against other infinite query uses.

apps/admin-x-framework/src/api/invites.ts
jonatansberg: Oh, yes please! I really don't like the current soup that's in there. I was just about to write in Slack and ask about this since the above code is way too frail. 

Personally, I'd rather not even bother with optimistic updates here at all given that we anyways block the list with the open modal. Unfortunately removing the optimistic update breaks the current Playwright testing approach that mocks the backend, and addressing that one way or another would probably balloon this even further.

jonatansberg:
@kevinansfield The code review bot has identified a valid issue with the query helpers in that they don't update the totals correctly. I'll treat that as a followup so that we can get this merged.

kevinansfield (approved):

PR #24107, opened 2025-06-26
Added 'social_web' setting to enable/disable ActivityPub once in GA

sagzy:

- ActivityPub is going to be enabled by default in Ghost 6.0
- Publishers can disable it via a setting in Admin
- this commits adds the database migration, and a follow-up commit will add the UI


ErisDS (approved):
LGTM 👍

PR #24115, opened 2025-06-26
Renamed Inbox and Feed to Reader and Notes in ActivityPub

minimaluminium:
no issues

- Changed "Inbox" to "Reader" and "Feed" to "Notes" throughout the app
- Modified corresponding URLs to reflect new naming convention


apps/admin-x-framework/src/api/activitypub.ts
```diff
@@ -112,7 +112,7 @@ export const useBrowseInboxForUser = createQueryWithId<InboxResponseData>({
     headers: {
         Accept: 'application/activity+json'
     },
-    path: id => `/inbox/${id}`
```
allouis, line 115: I think this whole file can/should be deleted - we shouldn't be using any of these requests in our app

PR #24118, opened 2025-06-26
🔥 Enforced max `?limit` of 100 in API requests

kevinansfield:
ref https://linear.app/ghost/issue/PROD-1605/

- adds `max-limit-cap` middleware that rewrites `?limit` parameters to a capped limit, including `?limit=all`
  - has an exceptions list for endpoints that still require a larger limit param
  - has config escape valves for exceptional use-cases after 6.0 upgrade
  - by rewriting the `query.limit` on the request it allows everything lower in the stack to behave as if the user requested the capped limit, avoiding need to adjust code anywhere else
- sets default capped limit to 100
- applies the middleware to Admin API, Content API, and Comments API

### DIFF
--- a/ghost/core/core/server/web/shared/middleware/max-limit-cap.js
+++ b/ghost/core/core/server/web/shared/middleware/max-limit-cap.js
@@ -6,14 +6,15 @@ const config = require('../../../../shared/config');
 // After Ghost 6.x we only allow a max limit of 100. This middleware enforces
 // that limit by rewriting the limit parameter before it reaches any API code.
 
-// Temporary exceptions to the max limit rule
-const EXCEPTION_ENDPOINTS = [
-    '/ghost/api/admin/posts/export/',
-    '/ghost/api/admin/emails/' // /:id/batches/ and /:id/recipient-failures/
-];
-
-const allowLimitAll = config.get('optimization:allowLimitAll') || false;
-const maxLimit = config.get('optimization:maxLimit') || 100;
+const limitConfig = {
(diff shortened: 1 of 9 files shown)


sagzy (approved):

ghost/core/core/server/web/shared/middleware/max-limit-cap.js
```diff
@@ -0,0 +1,55 @@
...
+];
+
+function maxLimitCap(req, res, next) {
+    // TODO: How expensive is reading from config? Maybe move outside of the function
+    const allowLimitAll = config.get('optimization:allowLimitAll') || false;
+    const maxLimit = config.get('optimization:maxLimit') || 100;
+
+    const limit = req.query.limit;
+
+    if (!limit) {
```
ErisDS, line 22: If not removing the config reads, maybe move them below this line 🤔 

But I would probably move them up. We can override the config in testing with them outside the function so seems worth it.

kevinansfield: 👍 I've moved them up. I had it that way before but tests were failing and I thought it was because the module-level variables weren't being overridden after the `require` but they seem to be passing now so maybe it was from something else not being right at the time 🤔

kevinansfield: No, I was right that we couldn't override in tests when they were module-scope.  Added a workaround in https://github.com/TryGhost/Ghost/pull/24118/commits/cd0a383aa1cc97de3fd82f34c73560336d44124d which feels _ok_ although maybe a bit magical 😅

PR #24119, opened 2025-06-26
Update self-service migration token expiration time to 24 hours

PaulAdamDavis:
ref https://linear.app/ghost/issue/CON-307/

- The current token expiration time of 15 minutes only allows enough time to process uploaded files, with no external network requests
- Increasing the expiration time to 24 hours will give self-service migrations plenty of time to perform external requests and scrape information that's only available on the source website.
- It also allows the service to queue migrations so we can run at a set concurrency to limit total requests to any specific platform.


ErisDS (approved):
Needs a proper commit message with an explanation for why this is changing, but otherwise this is good to merge.

PR #24133, opened 2025-06-27
Added JWT authentication for Tinybird API endpoints

cmraible:
ref https://linear.app/ghost/issue/PROD-48/generate-scoped-jwts-for-reading-api-data-from-tinybird

Right now our Tinybird implementation uses static tokens. This commits enables scoped JWT tokens for authenticating with Tinybird's API endpoints, which is a more secure means of authentication. It adds a Tinybird service to begin consolidating all our Tinybird configuration and functionality, wires it up to the existing Tinybird client in the stats service, and adds a `/tinybird/token` endpoint for the frontend to grab a refreshed token when needed.

This is designed to be backwards compatible with the existing stats tokens for now — if the configuration to generate JWT tokens is in place, Ghost will use JWTs — otherwise it will fallback to the static tokens if configured.


ghost/core/core/server/api/endpoints/tinybird.js
```diff
@@ -0,0 +1,23 @@
+const TinybirdServiceWrapper = require('../../services/tinybird');
+
+/** @type {import('@tryghost/api-framework').Controller} */
+const controller = {
+    docName: 'tinybird',
+
+    read: {
```
cmraible: Changing this to `token` to represent what the endpoint is for. The root `read` endpoint may be used for future use-cases, i.e. full tinybird config

PR #24162, opened 2025-06-27
Added search index endpoint for posts in Admin

sagzy:

- added `GET /ghost/api/admin/search-index/posts` Admin API endpoint 
- this endpoint returns up to 10k posts, sorted by most recently updated and with only the necessary fields for search in Admin or the editor
- wired up the new endpoint with the basic search and flex search providers in Admin

### DIFF
--- a/ghost/admin/app/services/search-provider-basic.js
+++ b/ghost/admin/app/services/search-provider-basic.js
@@ -40,6 +40,7 @@ export default class SearchProviderBasicService extends Service {
     @service ajax;
     @service notifications;
     @service store;
+    @service ghostPaths;
 
     content = [];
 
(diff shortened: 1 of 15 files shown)


ghost/admin/app/services/search-provider-basic.js
```diff
@@ -85,9 +86,15 @@ export default class SearchProviderBasicService extends Service {
...
-        const query = {fields: searchable.fields, limit: maxSearchableLimit};
+        let url;
+        let query;
+        if (searchable.model === 'post') {
+            url = this.ghostPaths.url.api(`search-index/${pluralize(searchable.model)}`);
+            query = {};
+        } else {
+            url = `${this.store.adapterFor(searchable.model).urlForQuery({}, searchable.model)}/`;
+            query = {fields: searchable.fields, limit: '10000'};
+        }
```
sagzy, line 97: Temporary branching - will get rid of it once we have all new search index endpoints

PR #24169, opened 2025-06-27
increase users.bio length soft limit

troyciesco:
closes https://linear.app/ghost/issue/PROD-2204/increase-soft-limit-on-usersbio

description tbd


troyciesco:
Closing this PR, since it's not a breaking change it doesn't have to be part of the 6.x work - i'll open one based on main

PR #24170, opened 2025-06-27
increase length for feature_image_alt columns

troyciesco:
closes https://linear.app/ghost/issue/PROD-2203/increase-column-length-on-posts-metafeature-image-alt-and-post

- increases the length of feature_image_alt columns in posts_meta and post_revisions


ghost/core/core/server/data/migrations/versions/6.0/2025-06-30-14-00-00-update-feature-image-alt-length.js
```diff
@@ -0,0 +1,26 @@
+const logging = require('@tryghost/logging');
+const {createNonTransactionalMigration} = require('../../utils');
+const DatabaseInfo = require('@tryghost/database-info');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        if (DatabaseInfo.isSQLite(knex)) {
+            logging.warn('Skipping migration for SQLite3');
+            return;
+        }
```
kevinansfield: _Very_ minor thing but seeing as I had to look up why we skip for SQLite I think we should start commenting these blocks with the reason for skipping. In this case it seems it's because SQLite doesn't have fixed-length varchar columns?

kevinansfield (approved):

ghost/core/core/server/data/migrations/versions/6.0/2025-06-30-14-00-00-update-feature-image-alt-length.js
troyciesco: @kevinansfield yep i think that's why (i based this on an older increased column length migration)

could i even just add it to the log message, or do we try to keep those short?
i.e. `logging.warn('Skipping migration for SQLite3 (fixed-length VARCHAR columns are not enforced)');`

i can probably improve the down logging too by switching to:
`logging.warn('Reverting posts_meta.feature_image_alt column length increase is not supported (major version migrations are not reversible)');`
`logging.warn('Reverting post_revisions.feature_image_alt column length increase is not supported (major version migrations are not reversible)');`

ErisDS:
Closed and reopened because I added branch protection rules that blocked the PR and couldn't find another way to get it to refresh the settings

PR #24177, opened 2025-06-28
Refactored TinybirdService so it always creates an instance

cmraible:
Part of the purpose of this service is to consolidate all of the "is tinybird enabled?" logic into a single place. Currently this is checked via configuration in a few places, and ideally this logic would all be in one place. 

This commit refactors the Tinybird service so that `getInstance()` always returns an instance, but it may be in a disabled state if the configuration is not correct. This way, consumers can use it as :

```
const tinybirdService = getInstance();
if (tinybirdService.isEnabled) {
......
}
```

This is much cleaner than checking the configuration values directly in multiple places in the codebase. It also removes the, IMO, overly complex wrapper service pattern, in favor of a simpler singleton pattern that maintains the lazy initialization.


ErisDS (changes requested):
I feel like this might be trying to solve too many problems at once.

1. The "is tinybird enabled?" problem
- The plan for how to solve this is using calculated settings, which we'll still need, as this doesn't yet take into account the web_analytics setting as far as I can tell, and is missing limits. The plan right now is to get this all the way done for activitypub, and then roll it to analytics. It's definitely not the right pattern going forward, but I'd like to solve them the same way and then figure out the right pattern next.
2. The "overly complex wrapper service pattern" problem
- I agree we need a new solution, and this looks like an interesting solution. However, I'd like to tackle this holistically, which to me would look like 
    - discovering and documenting all our requirements
    - planning a solution
    - trying it on a few services
    - getting agreement from the team and roll it out
3. I think this might be a provider, not a service... :D 

I'd really like to avoid further fragmenting the service patterns without a solid plan.

cmraible:
Yeah all fair points — I don't have strong feelings about this, I'm happy to close this as a failed experiment. Definitely agree re: not further fragmenting the service patterns. 

> I think this might be a provider, not a service

I think you're spot on here — part of my cognitive dissonance around the wrapper pattern is that this isn't really a service at all — it's not _doing_ anything, it's just an interface to Tinybird. 

Maybe this whole thing should just be a couple utility functions that we export from `core/server/data/tinybird`, and we treat it sort of like we treat the database connection — no service, no lazy init — just require this object if you need to interact with Tinybird. 

Or maybe it shouldn't be in Ghost at all, and instead it should just be `@tryghost/tinybird-client` or some such, that we just `require()` and pass the workspace Id and token 🤔

cmraible:
> The plan for how to solve this is using calculated settings, which we'll still need, as this doesn't yet take into account the web_analytics setting as far as I can tell, and is missing limits.

Totally fair, although I was thinking about this differently — in my mind there are two separate concerns here:
1.  Does this Ghost instance use Tinybird, and if so what's the config (endpoints, tokens, etc)?
2. Is web analytics enabled on this site?

(2) requires (1), but (1) doesn't necessarily imply (2). Using your example from the other day, if we one day add `clicks` to Tinybird, I wouldn't think that would be coupled with the web analytics features. Whereas with ActivityPub it's more 1:1 — I can't think of any scenario where a site with activitypub disabled would ever need the activitypub service. 

I'm just sharing how I've been thinking about it, not arguing for any modifications to the plan. It's not something we need to figure out right now since Tinybird & web analytics are basically 1:1 for the time being.

ErisDS:
> Totally fair, although I was thinking about this differently — in my mind there are two separate concerns here:
> 
> Does this Ghost instance use Tinybird, and if so what's the config (endpoints, tokens, etc)?
> Is web analytics enabled on this site?
> (2) requires (1), but (1) doesn't necessarily imply (2).

This makes a tonne of sense! 

>  I'm happy to close this as a failed experiment. [...] part of my cognitive dissonance around the wrapper pattern is that this isn't really a service at all

I don't think you're wrong about any of this at all. I think we should revisit all of this once we're over the hump and into cleanup.

PR #24180, opened 2025-06-29
Fixed staff tokens not giving user permissions

ErisDS:
ref https://github.com/TryGhost/Ghost/pull/2336
ref https://github.com/TryGhost/Ghost/pull/23585/files#diff-3d98e1fd51a0472fe91d09aa0749a46250023b50945524e5e0815f24ed6522a6R25-R36

A long time ago now, we added staff tokens - these are tokens you can get from your user profile, and use to authenticate with the Admin API as if you were logged in with your username and password.

Given the intention of staff tokens is that access is granted as if the user tied to the token was logged in normally - the permissions were intended to be the same.

However, that was never implemented. Staff tokens were adding the user and the api_key to req, but there was no handling to make sure that resulted in the correct logic.

A while ago I started adding testing for staff tokens [[ref](https://github.com/TryGhost/Ghost/pull/23360)] 

And then I added some logic to bypass our weird 'NotImplemented' middleware [[ref](https://github.com/TryGhost/Ghost/pull/23585/files#diff-3d98e1fd51a0472fe91d09aa0749a46250023b50945524e5e0815f24ed6522a6R25-R36)]

NOTE: I left 2 things unsupported, delete all content and transfer ownership, because it felt wrong to suddenly give existing tokens access to dangerous operations.

I thought I'd cracked it - but it turned out that wasn't the case as [reported here](https://github.com/TryGhost/Ghost-CLI/issues/1952#issuecomment-2927594223).

Although the request is getting passed to our canThis logic now to determine permissions - that is STILL making the wrong choice. 

The logic in canThis states if an api_key is present, ignore the user permissions and use the api_key permissions. That was because it was implemented in a world where you could not have both.

So this PR updates that logic. If a user and an api key are present, we use the user permissions and ignore the api_key.

Note: this goes hand in hand with a change to make sure our AdminAPIAgent properly resets authentication when using a new type [[ref](https://github.com/TryGhost/Ghost/commit/8ae5aa0100d04006b629cc0f7fbe28166595ac89)]. This wasn't working properly and was making the tests not behave as expected / masking some broken cases.

Note 2: This changeset is a bunch of tests and the minimal amount of change to the permission system as possible to achieve the logical change. There's obviously a heap more that could be done to make this more readable.

### DIFF
--- a/ghost/core/core/server/services/permissions/can-this.js
+++ b/ghost/core/core/server/services/permissions/can-this.js
@@ -81,9 +81,16 @@ class CanThisResult {
                     // Check api key permissions if they were passed
                     hasApiKeyPermission = true;
                     if (!_.isNull(apiKeyPermissions)) {
-                        // api key request have no user, but we want the user permissions checks to pass
-                        hasUserPermission = true;
-                        hasApiKeyPermission = _.some(apiKeyPermissions, checkPermission);
+                        if (loadedPermissions.user) {
+                            // Staff API key scenario: both user and API key present
+                            // Use USER permissions and ignore API key permissions
+                            hasApiKeyPermission = true; // Allow API key check to pass
+                        } else {
+                            // Traditional API key scenario: API key only, no user
+                            // Use API key permissions as before
+                            hasUserPermission = true;
+                            hasApiKeyPermission = _.some(apiKeyPermissions, checkPermission);
+                        }
                     }
 
                     // Offer a chance for the TargetModel to override the results
--- a/ghost/core/test/e2e-api/admin/api-tokens.test.js
+++ b/ghost/core/test/e2e-api/admin/api-tokens.test.js
@@ -137,7 +137,7 @@ describe('Admin API', function () {
 
     // Make sure that staff tokens are blocked from certain endpoints
     describe('Staff Token Blocklist', function () {
-        describe('DELETE /db endpoint', function () {
+        describe('DELETE /db endpoint (delete all content)', function () {
             it('Owner staff token should be blocked', async function () {
                 await agent.useStaffTokenForOwner();
                 await agent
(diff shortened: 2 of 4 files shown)


allouis (commented):
This code achieves what what we want it to, but I think it makes permissions even more complicated/difficult to understand.

The problem here is that we're loading the incorrect permissions for the ApiKey - this is handled in the `providers.js` file, in the `apiKey` method.

If we instead update this method to load the permissions for the related user, if there is one. Then we correctly give the ApiKey the same permissions as the User.

I _think_ this could be as simple as adding `'user', 'user.roles' & 'user.roles.permissions'` to the `withRelated` and handling them as they're handled in the `user` method

You mention excluding two permissions in the PR, but I think the loading of permissions is a better place to do that than the NotImplemented middleware.

ErisDS:
> The problem here is that we're loading the incorrect permissions for the ApiKey - this is handled in the providers.js file, in the apiKey method.

I agree that's ONE of the problems... there are many 😂 

> I think this could be as simple as adding 'user', 'user.roles' & 'user.roles.permissions' to the withRelated and handling them as they're handled in the user method

I feel like that's a more dangerous change to make, because of the potential for impact on the DB. 

Whilst the code here isn't clean or making any effort to improve things, if we're confident it does that it's supposed to, then it feels like the right change to make right now and solve this specific problem without causing regressions or performance issues. 

> You mention excluding two permissions in the PR, but I think the loading of permissions is a better place to do that than the NotImplemented middleware.

NotImplemented middleware needs to die, it's a total hack, but right now I HAVE to explicitly forward the staff token through it else it gets the same treatment as other tokens. 

Seeing as the intention of the staff token rules is the same as what's in NotImplemented: "You're meant to have permission to do this, but we don't trust this endpoint so are blocking it for now"... I figured I'd put it in the same place for consistency, rather than inventing another new way to do permissions.

Note: In 6.x I am going to change the NotImplemented middleware to be permission middleware and throw a 403 rather than 501, as I feel that makes more sense: https://linear.app/ghost/issue/PROD-2219/breaking-change-change-notimplemented-middleware-to-throw-403

Again, I want to do more to clean it up, but that should be non-breaking once we are throwing a 403.

PR #24190, opened 2025-06-30
Updated analytics tests to work with stubbed responses

ibalosh:
* extracted route regexp creation from mockApi function, to make it easier to understand that these are used to determin admin, activity-pub or tinybird API route
* created factories for fixture responses to use with custom data for tests
* removed usage of real tinybird for now, to get these tests running
* reused e2e page-objects added at root folder, which can be used by all tests that use playwright
* added couple of more tests


apps/stats/test/utils/tinybird-helpers.ts
```diff
@@ -91,3 +98,25 @@ export async function addAnalyticsEvent(eventProps: AnalyticsEventProps) {
         throw new Error(`Request failed (${response.status}): ${errorMessage}`);
     }
 }
+
+export function mockedRequests(siteUuid: string) {
```
9larsons, line 22: I like pulling this out from the test files, but I'd move it to a more generic utils file than tinybird-helpers.

ibalosh: agreed

ibalosh:
seems my seed data for tinybird got commited by accident, how do I reset it?

ibalosh:
> seems my seed data for tinybird got commited by accident, how do I reset it?

I reverted the tinybird data to one from main. @9larsons , I think we should `.gitignore` seeded data, and just generate it locally? Currently it creates a confusion, since any change to data will generate new data there and can result in accidental push to the repository.

cmraible:
> I reverted the tinybird data to one from main. @9larsons , I think we should .gitignore seeded data, and just generate it locally? Currently it creates a confusion, since any change to data will generate new data there and can result in accidental push to the repository.

I'm not sure I understand this. The seeded data referenced is the main data fixture we use for testing our Tinybird datafiles — the datasources, materialized views and endpoints that make up the Tinybird "schema". These fixtures are loaded into Tinybird before running the tests in https://github.com/TryGhost/Ghost/blob/7ddcf60b3eb6ba20c24eb5f9293b3a27bc9a6852/ghost/core/core/server/data/tinybird/tests, and therefore need to stay in sync with the tests. I think they should remain in source control, but let me know if I'm misunderstanding something!

Docs on Tinybird fixtures and testing for reference: https://www.tinybird.co/docs/forward/test-and-deploy/test-your-project

cmraible:
@ibalosh I'm not able to get the acceptance tests to pass locally — is there any prerequisite setup to get things working? 

I've just checked out the branch, run `yarn` to install dependencies, run `yarn build` to build all the dependencies, then run `yarn test:acceptance` from `apps/stats`. Am I missing something, or are these tests not yet expected to pass?

cmraible:
I'm also curious about the future of this test suite generally — I think we could make an argument that the e2e test suite we're working on replaces these tests, so maybe we don't need these at all? I know we've debated a bit on this approach (stubbing dependencies) vs full e2e tests, but I'm not sure I'm clear on where we landed.

@ibalosh do you see this test suite, with the stubbed responses, as adding value _in addition to_ the true e2e test suite, or do you think the e2e test suite will make these acceptance tests obsolete?

cmraible:
Last thing to mention: if we are going to keep this test suite around, I think we need to add it to CI in this PR, as soon as we get them working to enforce that these tests will always pass when we make changes.

ibalosh:
hey @cmraible , 
all valid points, let me try to answer to them one by one:

**Tinybird**:  I forgot about the tests we have in tinybird. The problem I see with that is related to running yarn command from package json: `yarn reset:data:analytics`. It will break the consistency, since if you run that, the fixture folder will get some 30mb of fixtures. It is very easy to commit and push these by accident (that command is great for checking more data locally).

Due that, I think it makes more sense to .gitignore fixture, generate the data on the fly and run the tests then, it's a 10kb of fixtures anyways. After all, this is the process we plan to use for other tests too, create data, test against it. I just wanted to mention this one, no change needed. I am not sure if it affects the tests in any other way in future.

**e2e tests**: Only reason I changed these was to see how stubbing works with mockApi, and fix the tests we already had here, and to use the shared pages from the root e2e test suite. They did not work, and I wanted to ditch the temporary api call workaround I added here. 

I am still 100% for true-ish e2e tests we plan, with factories. These tests will still work then too, we will just change the mocks, to factories, and use from then on the APIs. Let's catchup on todays call and decided, whether to keep them around until we switch, or just delete the acceptance tests here in stats app for now.

Meanwhile, I fixed these tests, they should work for you now, without setup (forgot about env variables I did not remove from tests)

9larsons:
re: fixtures with the analytics generator script, the right solution here is to update it to use the events API instead of depending on it to load the ndjson file. We can't add that ndjson to gitignore because it's actually necessary for our Tinybird tests that run on CI+locally (and yes, I've accidentally committed it a few times already).

Moving the script over to events is on my todo list, I've just been focused on the factories instead.

9larsons:
As for acceptance vs e2e, I thought we found value in both. We aren't going to test every case of state management in e2e tests, for instance, though we could more easily do it with acceptance.

I know this is something we've not settled on as a group. My vote would be to fix the acceptance tests (thank you for doing that) and let's discuss today as you also mentioned. I don't see e2e as a complete replacement, though I do see e2e as the most important atm.

PR #24194, opened 2025-06-30
Replaced loadedPermissions with role

allouis:
ref https://linear.app/tryghost/issue/ENG-728
ref https://github.com/TryGhost/Ghost/pull/19896

The loadedPermissions were only used to check roles or whether an api key was used. The later is already exposed on the `context` object, so we just need to pass the role. This allows us to internally change how permissions are loaded without having downstream effects, it also moves us closer to pulling roles out of the db and into hardcoded strings.

---

This is a reopening of https://github.com/TryGhost/Ghost/pull/19896

### DIFF
--- a/ghost/core/core/server/models/comment.js
+++ b/ghost/core/core/server/models/comment.js
@@ -157,7 +157,7 @@ const Comment = ghostBookshelf.Model.extend({
         return softDelete();
     },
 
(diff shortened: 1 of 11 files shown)


allouis:
Will maybe pick this up when I've got some downtime in future, but this reopenign was to check the state of the rebase

PR #24203, opened 2025-06-30
Fixed initial 403 errors in Tinybird API requests in posts and stats apps

cmraible:
The first requests directly to the Tinybird APIs after a page load are returning a 403 in some cases, because the requests aren't waiting for the token to load from the API. This commit adds a custom hook to `admin-x-framework` that wraps Tinybird's `useQuery()` hook for making requests to Tinybird, and includes the logic to delay the request until the token is returned from Ghost's API.

### DIFF
--- a/apps/admin-x-framework/package.json
+++ b/apps/admin-x-framework/package.json
@@ -92,6 +92,7 @@
     "dependencies": {
         "@sentry/react": "7.120.3",
         "@tanstack/react-query": "4.36.1",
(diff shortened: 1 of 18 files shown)


9larsons (approved):
Made a small adjustment to `useTinybirdQuery` and adjusted `useActiveVisitors` to use the hook, which was adjusted in your earlier commit to main.

cmraible:
Tested on staging, no 403s 🎉

PR #24204, opened 2025-06-30
Removed the tracker token from the ghost-stats script

cmraible:
ref https://linear.app/ghost/issue/PROD-1961/remove-the-tracker-token-from-ghost-config-and-tracker-script

Since the prototype of Ghost's web analytics features, Ghost has been passing this tracker token along with every tracker request, which allows it to authenticate with Tinybird's events API. Now that the analytics service itself has this token and is using it to authenticate directly with Tinybird's API, we can remove it from Ghost, which keeps the token more secure.

This removes all references to the tracker token from the `ghost-stats.js` script and from `ghost_head`, which removes all dependencies on the `tinybird:tracker:token` configuration.


cmraible:
Closing in favor of https://github.com/TryGhost/Ghost/pull/24211

PR #24215, opened 2025-07-01
Automated local dev setup for traffic analytics

cmraible:
Running Ghost with traffic analytics features locally currently takes quite a bit of manual setup, to get the Tinybird local container running locally and to configure Ghost correctly to talk to it, with all the right tokens. 

This PR attempts to simplify this by creating a wrapper bash script that you can run locally, which will:
- Start tinybird-local in the background using docker compose
- Run an initial deployment to create the datasources, pipes, and critically _tokens_
- Grabs the admin token, tracker token, and workspace ID from tinybird-local
- Runs the `analytics-service` docker container, passing it the appropriate tracker token
- Sourcing the right environment variables to configure Ghost correctly

The main idea here is to make getting setup with traffic analytics locally as painless as possible — this script currently has 3 dependencies:
- docker — this is required for tinybird-local regardless of the scripting
- `jq` — for parsing the tokens from the API. We could eliminate this by rewriting part or all of this in JS
- the tinybird CLI installed locally — we could either include the installation in the script itself, or find a way to create + run these bits in a docker container


cmraible:
Draft: seems to be a bit finicky right now, but I think it should be possible to work out the kinks. This will be nice, not only for improved DX in general, but it will also be more or less required to get E2E tests working against tinybird-local.

cmraible:
Closed in favor of https://github.com/TryGhost/Ghost/pull/24295

PR #24228, opened 2025-07-01
🐛 Fixed x username validation failing on www and http

troyciesco:
fixes https://linear.app/ghost/issue/PROD-1536/unable-to-save-x-handle-with-www-in-the-the-handle-name

- the check on X usernames makes sure it's not a url first, but it was incorrectly throwing an error on valid usernames like `wwwUsername` and `httpUsername`
- this is a minimal fix to make sure those valid usernames pass

### DIFF
--- a/apps/admin-x-settings/src/utils/socialUrls/twitter.ts
+++ b/apps/admin-x-settings/src/utils/socialUrls/twitter.ts
@@ -15,13 +15,10 @@ export function validateTwitterUrl(newUrl: string) {
             username = username.slice(1);
         }
 
-        // check if username starts with http:, https:, or www. and show error if so
-        if (username.match(/^(http:|www\.|https:)|(\/)/) || !username.match(/^[a-z\d_]{1,15}$/mi)) {
-            const message = !username.match(/^[a-z\d_]{1,15}$/mi)
-                ? 'Your Username is not a valid Twitter Username'
-                : 'The URL must be in a format like https://x.com/yourUsername';
-            throw new Error(message);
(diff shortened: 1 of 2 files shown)


kevinansfield (approved):
This looks like a reasonable minimal fix 👍 

However, looking at [~Twitter~X's help docs](https://help.x.com/en/managing-your-account/x-username-rules) shows that periods aren't a valid username character. I wonder if we'd be better off changing our validation to match theirs? It would have caught the `www.` situation as well as other invalid usernames.

> Your username cannot be longer than 15 characters. Your name can be longer (50 characters) or shorter than 4 characters, but usernames are kept shorter for the sake of ease.
> A username can only contain **alphanumeric characters** (letters A-Z, numbers 0-9) with the exception of **underscores**, as noted above. Check to make sure your desired username doesn't contain any symbols, dashes, or spaces.

troyciesco:
> I wonder if we'd be better off changing our validation to match theirs? It would have caught the `www.` situation as well as other invalid usernames.

oh wait, yeah, we don't need the http/www check there at all lol. i don't think the url error message can ever even get reached in this part:
```
        if (username.match(/^(http|www)|(\/)/) || !username.match(/^[a-z\d._]{1,15}$/mi)) {
            const message = !username.match(/^[a-z\d._]{1,15}$/mi)
                ? 'Your Username is not a valid Twitter Username'
                : 'The URL must be in a format like https://x.com/yourUsername';
            throw new Error(message);
```

simplified it, and removed the `.` from the regex. thanks!

PR #24229, opened 2025-07-01
🎨 Added DID as a valid Bluesky username pattern

troyciesco:
closes https://linear.app/ghost/issue/PROD-1870/allow-perma-links-in-validation-for-bluesky
reported here https://forum.ghost.org/t/bluesky-entry-on-profile-social-links-blocks-perma-links/57712

- In addition to regular handles, Bluesky also has the concept of Decentralized Identifiers (DID) as noted in the [Bluesky docs](https://docs.bsky.app/docs/advanced-guides/resolving-identities).
- This update makes it so staff users can set their Bluesky username on Ghost to a DID
- also includes tidying specifically impacted by these changes

### DIFF
--- a/apps/admin-x-settings/src/utils/socialUrls/bluesky.ts
+++ b/apps/admin-x-settings/src/utils/socialUrls/bluesky.ts
@@ -1,8 +1,21 @@
 import validator from 'validator';
 
+function isValidBlueskyUsername(username: string): boolean {
+    const validUsernamePatterns = [
+        // DID username: did:plc: + 24 chars
+        /^did:plc:[a-zA-Z0-9._]{24}$/, 
+        // Regular username: max 15 chars
+        /^[a-zA-Z0-9._]{1,15}$/,
+        // Domain username: requires dot, max 191 chars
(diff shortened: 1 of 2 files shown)


kevinansfield (approved):
Looks good 🙌  CodeRabbit does raise a good point about the DID usernames needing to be lowercase so updating the regex sounds like a plan. Maybe we could also do a lowercase conversion if it otherwise matches?

troyciesco:
> Looks good 🙌 CodeRabbit does raise a good point about the DID usernames needing to be lowercase so updating the regex sounds like a plan. Maybe we could also do a lowercase conversion if it otherwise matches?

@kevinansfield oh yeah good point by you and coderabbit - i'd started from the suggested regex in the [atproto docs](https://atproto.com/specs/did#at-protocol-did-identifier-syntax) and modified to match bluesky's did:plc: version, but can take it even farther.

updated that, and added [formatUsername](https://github.com/TryGhost/Ghost/pull/24229/files#diff-b188ac7b6b37dd763818aeabc0353d2b293ebadfecbc42221b120492f1edab72R22) to take care of that where it's needed 👍

PR #24230, opened 2025-07-01
Fixed ordering of search indexes

ErisDS:

- When rebasing the new 6.x branch on top of these changes we started getting failures due to content length
- This makes some sense, but what was odd was that we couldn't see exactly what was changing
- This was due to the matchers in the tests being overly specific and using anyString.
- The intention of snapshot testing is to leave as many values in their natural form as possible, and only use matchers for values that will reasonably change: IDs, dates, versions etc
  - Note: I have seen Claude implement snapshot testing this way and also nearly committed it cos it looks right at first glance!
- Once the matchers were modified to minimal versions, it became clear that the API result wasn't deterministic. By that I mean it didn't repeatedly return the exact same response - the order kept changing
- This is because we don't store milliseconds in our DB, we were only ordering by updated_at, and our tests create tag fixtures way too fast
- So there are 3 separate fixes in this changeset
   - Added "id DESC" to all the sort orders, so we get a deterministic order if updated_at is the same
   - Changed our fixtures to add a second to the date used for creating tags, so each tag is 1 second apart, else the tags are still in a nondeterministic order
   - Updated snapshot matchers as mentioned above


ErisDS:
Cleaning up older PRs 💅

@sagzy something for the next couple of weeks?

sagzy:
Yes! Still on my todo list, hoping to get to it finally this week

PR #24246, opened 2025-07-02
Improved unit test coverage for Stats app

9larsons:
no ref
- added unit tests for utils and hooks
- added msw implementation that will likely be pulled out to be extended from framework

The goal here is to add some behavior testing for the various hooks. Eventually, we'll want to rebuild the endpoints and hook structure such that we likely may not need custom hooks at all for each page, unless it needs to happen on the frontend. A good example of a custom hook is the `usePostSuccessModal` which encapsulates a lot of frontend behavior.

`useFeatureFlag` and `withFeatureFlag` may get removed, or be refactored to framework. For now the tests are being skipped for `useFeatureFlag` because of the need to refactor the use of context, which is difficult to test and can't be done so with `msw`.


apps/posts/test/unit/hooks/usePostFeedback.test.tsx
```diff
@@ -0,0 +1,283 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import {beforeEach, describe, expect, it, vi} from 'vitest';
+import {renderHook} from '@testing-library/react';
+import {usePostFeedback} from '@src/hooks/usePostFeedback';
+
+// Mock the dependencies
+vi.mock('@tryghost/admin-x-framework/api/feedback');
+
+const mockGetPostFeedback = vi.mocked(await import('@tryghost/admin-x-framework/api/feedback')).getPostFeedback;
```
cmraible:
```suggestion
const getPostFeedbackApiStub = vi.mocked(await import('@tryghost/admin-x-framework/api/feedback')).getPostFeedback;
```

I'm the worst offender here and haven't been using these terms correctly myself, but I think since this is just simulating the return value from the API, and not making any assertions _itself_, this is technically a stub, not a mock.

apps/posts/test/unit/hooks/usePostSuccessModal.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+                filter: 'status:[published,sent]',
+                limit: '1',
+                fields: 'id'
+            },
+            enabled: false
+        });
+    });
+
+    it('creates modal props when post data is available', () => {
+        const mockPost = {
```
cmraible: Similarly, this isn't really a "mock", maybe we could call this `regularPost` or something along those lines?

apps/posts/test/unit/hooks/usePostSuccessModal.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+        const {result} = renderHook(() => usePostSuccessModal());
+
+        result.current.closeModal();
+
+        expect(result.current.isModalOpen).toBe(false);
+        expect(result.current.postCount).toBe(null);
+    });
+
+    it('handles email-only posts', () => {
+        const mockPost = {
```
cmraible: Same comment — maybe we could call this `emailOnlyPost`

apps/posts/test/unit/hooks/usePostSuccessModal.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+            error: null
+        } as any);
+
+        const {result} = renderHook(() => usePostSuccessModal());
+
+        expect(result.current.post?.email_only).toBe(true);
+    });
+
+    it('handles multiple authors', () => {
+        const mockPost = {
```
cmraible: `postWithMultipleAuthors`?

apps/posts/test/unit/hooks/usePostSuccessModal.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+            error: null
+        } as any);
+
+        const {result} = renderHook(() => usePostSuccessModal());
+
+        expect(result.current.post?.authors).toHaveLength(3);
+    });
+
+    it('handles posts without authors', () => {
+        const mockPost = {
```
cmraible: `postWithoutAuthors`?

apps/posts/test/unit/utils/constants.test.ts
```diff
@@ -0,0 +1,148 @@
...
+            expect(STATS_RANGES.LAST_12_MONTHS.value).toBeLessThan(STATS_RANGES.ALL_TIME.value);
+        });
+
+        it('is read-only (as const)', () => {
+            // TypeScript prevents modification at compile time
+            // Runtime immutability depends on the implementation
+            expect(typeof STATS_RANGES.TODAY.value).toBe('number');
+            expect(STATS_RANGES.TODAY.value).toBe(1);
+        });
+    });
```
cmraible: I'm not sure if we need all/any of these — there isn't really any functionality here, except for maybe the last one

cmraible (commented):
Left some comments! I'm admittedly _mostly_ just parroting feedback that @ibalosh gave me though! 

No blockers from my POV, but overall I'd love to find a better approach to testing the hooks — in general I think these tests are quite closely coupled with the implementation, and therefore making _any_ kind of change to the hooks themselves will require updating some or potentially many of these tests, even if there's no change in behavior. TBH I don't have a better suggestion, maybe @ibalosh would?

apps/posts/test/unit/utils/constants.test.ts
ibalosh: I agree. We should not test these, it seems to be too much, none of them I think, since:

- these are pretty much tests against plain objects, it's obvious what input and outputs are, pretty much we just copy it to the test (I feel it's the same risk to miss spelling in test and constant itself)
- look more like an implementation, rather than behavior we want to verify


Maybe I could get bit more context about them in the view? I would say it would be probably be better by verifying unit of behavior - like render in component, and then seeing that they are all there, with proper names?

Do we have a component where we could verify these or we could indirectly test them by playwright test that will click through filters. The closer we can get to behavior, with speed rather than implementation the better.

cmraible (approved):

9larsons:
@ibalosh @cmraible I've added you both - I think it's easiest to look at these in the codebase itself/in isolation vs. the diff view.

apps/posts/test/utils/msw-helpers.ts
```diff
@@ -0,0 +1,295 @@
...
+    // Stable APIs (auto-included with defaults unless overridden)
+    site?: Record<string, unknown>;
+    config?: Record<string, unknown>;
+    settings?: Array<Record<string, unknown>>;
+    tinybirdToken?: string;
+    
+    // Pattern 2: Escape hatch for everything complex
+    customHandlers?: any[];
+};
+
```
cmraible, line 174: The type safety this adds is awesome 🙌

cmraible:
Overall I think this is a big step in the right direction, and using MSW this way feels a lot cleaner and more sane than all the `vi.mock()` calls 🙌

A general note on the API of the `mockServer`:

As a reader reading these tests, I want to quickly understand exactly which endpoints are mocked for this test, and what response to expect. By assigning names to the endpoints (i.e. `posts`, `newsletterBasicStats`, etc, it adds a layer of translation: "which endpoint does `newsletterBasicStats` mock again?"

I'd almost prefer the more verbose, explicit syntax that's used for defining the `customHandlers`, i.e. instead of:

```
mockServer.setup({
    posts: [mockData.post({id: testPostId})]
});
```

it would be:

```
mockServer.setup([
    {method: 'get', path: '/api/admin/posts', response: [mockData.post({id: testPostId})]}
]);
```

Just food for thought, I'm not strongly opposed and generally think this moves our unit tests in the right direction 👌

apps/posts/src/hooks/usePostReferrers.ts
```diff
@@ -1,30 +1,7 @@
-import moment from 'moment';
 import {getSymbol} from '@tryghost/admin-x-framework';
 import {useMemo} from 'react';
 import {useMrrHistory, usePostGrowthStats as usePostGrowthStatsAPI, usePostReferrers as usePostReferrersAPI} from '@tryghost/admin-x-framework/api/stats';
 
-// Helper function to convert range to date parameters
-export const getRangeDates = (rangeInDays: number) => {
```
ibalosh, line 7: 👍

apps/posts/src/hooks/useEditLinks.ts
```diff
@@ -1,3 +1,6 @@
+// TODO: Remove this file entirely. This hook is just a trivial wrapper that renames properties.
```
ibalosh, line 1: 👍 I agree, I would remove it right away. Since this is a wrapper for a rename, that would indicate original names probably needs adjustment in future. This `{mutateAsync: editLinks, isLoading: isEditLinksLoading}` is confusing and I am not sure here what the intention is, since it's far from original code, and I am wondering if it's needed, and if it is, whether it should be modified here.

One way to do it as a quick workaround, not ideal, but still better in my opinion - would be to move the name transformation close to original codebase/domain. 

That would mean `apps/admin-x-framework/src/api/links.ts`. In my opinion, you should not worry about names in your current codebase in `apps/posts`, they should be already correct. It's not their domain to worry about, framework codebase should do it, unless there is a specific reason to modify it in apps? 

You could also use namespace kind of style on import, so you could have instead of:

`import {useBulkEditLinks as useEditLinksApi} from '@tryghost/admin-x-framework/api/links';`

something along the lines:

```javascript
import * as api from '@tryghost/admin-x-framework/api/links';
api.useBulkEditLinks();
```

it's clear, easy to read. But this one depends on choices before.

apps/posts/test/unit/hooks/useEditLinks.test.tsx
```diff
@@ -1,169 +1,27 @@
...
-        });
-
-        await waitFor(() => {
-            expect(result.current.isEditLinksLoading).toBe(false);
-        });
+    it('exposes the expected interface', () => {
+        const {result} = renderHook(() => useEditLinks());
+        
+        expect(typeof result.current.editLinks).toBe('function');
+        expect(typeof result.current.isEditLinksLoading).toBe('boolean');
```
ibalosh, line 25: I am wondering if we need this test at all. Type should be checkable with typescript, tests wise, it's an implementation detail.

apps/posts/test/unit/hooks/useFeatureFlag.test.tsx
```diff
@@ -1,120 +1,184 @@
...
+ * 1. Accept vi.mock for context dependencies (MSW for HTTP, vi.mock for React contexts)
+ * 2. Refactor useFeatureFlag to use useBrowseSettings directly instead of useGlobalData
+ *    This would make it work perfectly with MSW since it only needs HTTP calls
+ * 3. Create a minimal PostAnalyticsContext provider wrapper for tests
+ * 
+ * CURRENT STATUS: Tests skipped since this code is not currently in use
+ * When this feature is needed, choose solution #2 (refactor) for better architecture
+ */
+
+describe.skip('useFeatureFlag', () => {
```
ibalosh, line 24: Let's catchup on call about this one, currently it works, but the tricky way we setup the side effect in function (global data) with mock, we will figure out a better way.

apps/posts/test/unit/hooks/usePostNewsletterStats.test.tsx
```diff
@@ -1,60 +1,60 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import {beforeEach, describe, expect, it, vi} from 'vitest';
-import {createTestWrapper, mockApiHook, setupUniversalMocks} from '../../utils/test-helpers';
+import {beforeEach, describe, expect, it} from 'vitest';
+import {createTestWrapper, mockData, mockServer} from '../../utils/msw-helpers';
```
ibalosh, line 3: as I was looking through this test, I started wondering about `usePostNewsletterStats.ts`.

And saw these functions, which might be decoupled out of hook to test easier as unit tests, since that hook seems big:

* `const stats = useMemo(() ...` 
* feedbackStats
* newsletterStatsResponse
* averages

just a thought

ibalosh (approved):
Hey @9larsons 

this definitely looks better, I shared some ideas, since I think we should further:

* slim down hooks, and decouple code from there once they get big
* see if we can test elsewhere (component renders, playwright if needed)
* figure out simpler mocking, a lot going on now

I like the formatting of the tests. I added some comments on separating concerns bit better too, we can add this one in now, and catchup more on the calls.

PR #24247, opened 2025-07-02
Fixed infinite loop when `ui60` flag is enabled without analytics

kevinansfield:
no issue

- the flag-on and no-analytics state resulted in an infinite home->stats-x->home redirect loop when opening Admin
- added a check to the home redirect using the same conditional as stats-x to avoid the redirect to stats-x when we'd get redirected back


9larsons (approved):
That'll do it 👍. I'll be looking at removing config requirements today/tomorrow to further simplify this.

PR #24250, opened 2025-07-02
Removed trafficAnalyticsAlpha flag

9larsons:
ref https://linear.app/ghost/issue/ENG-2441/
- moved appropriate trafficAnalyticsAlpha-gated functionality to settings.webAnalytics if not already
- moved appropriate settings to settings.membersTrackSources instead of trafficAnalytics (do not fetch web analytics data [tinybird])


ghost/admin/app/components/posts-list/list-item-analytics.hbs
```diff
@@ -165,7 +165,7 @@
         <div class="gh-post-list-metrics-container">
 
             {{!-- Visitor count column (only show for published posts when traffic analytics is enabled AND web analytics is enabled) --}}
-            {{#if (or this.settings.webAnalytics false)}}
+            {{#if (and this.settings.webAnalytics (feature 'trafficAnalytics'))}}
```
peterzimon: @9larsons — I think we don't need to check for `(feature 'trafficAnalytics')` here, since we use `list-item-analytics.hbs` in `list.hbs` with the same condition: https://github.com/TryGhost/Ghost/blob/rm-trafficAnalytics/ghost/admin/app/components/posts-list/list.hbs#L2

peterzimon (changes requested):
Please see inline comment

ghost/admin/app/components/posts-list/list-item-analytics.hbs
9larsons: That's fair, although it'll get caught with cleanup. I'll remove and merge - thanks for taking a look!

peterzimon (approved):

PR #24254, opened 2025-07-02
Cleaned up Admin's home routing

9larsons:
ref https://linear.app/ghost/issue/PROD-2165/
- we no longer need to direct to Social/activitypub
- we should only be directing to Analytics, but need to be behind ui flag for the moment
- cleaned up what appears to be dead code for onboarding, which is in the analytics handlebars component now

### DIFF
--- a/ghost/admin/tests/acceptance/setup-test.js
+++ b/ghost/admin/tests/acceptance/setup-test.js
@@ -173,4 +173,18 @@ describe('Acceptance: Setup', function () {
                 .to.have.string('Access Denied from url: unknown.com. Please use the url configured in config.js.');
         });
     });
+
+    describe('?firstStart=true', function () {
+        beforeEach(async function () {
+            let role = this.server.create('role', {name: 'Owner'});
+            this.server.create('user', {roles: [role], slug: 'owner'});
+
+            await authenticateSession();
+        });
+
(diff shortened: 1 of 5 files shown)


9larsons:
Behavior is currently:
- with flag: takes a fresh site to onboarding, when completing or skipping takes you to analytics (analytics is selected in the sidebar)
- without flag: takes a fresh site to onboarding w/ analytics selected in the sidebar (misleading 👎), after completing/skipping onboarding takes you to dashboard

9larsons:
This needed to be handled at the application level. Ember appears to treat routes as singletons, so the fact we hit it at all when unauthenticated was problematic.

kevinansfield (approved):

ghost/admin/tests/acceptance/setup-test.js
```diff
@@ -173,18 +173,4 @@ describe('Acceptance: Setup', function () {
...
-            this.server.create('user', {roles: [role], slug: 'owner'});
-
-            await authenticateSession();
-        });
-
-        it('transitions to dashboard', async function () {
-            await visit('/?firstStart=true');
-            expect(currentURL()).to.equal('/dashboard');
-        });
-    });
```
kevinansfield: Can we bring this test back now that the query param handling is re-added?

PR #24258, opened 2025-07-03
Consolidated duplicate reposts in feed in ActivityPub

minimaluminium:
ref PROD-709

- Eliminated multiple instances of the same post from different reposters
- Displayed single post with "Person and X others reposted" label
- Reduced feed clutter by combining identical reposted content


apps/admin-x-activitypub/src/views/Feed/components/FeedList.tsx
```diff
@@ -34,6 +34,49 @@ const FeedList:React.FC<FeedListProps> = ({
     const loadMoreRef = useRef<HTMLDivElement | null>(null);
     const endLoadMoreRef = useRef<HTMLDivElement | null>(null);
 
+    // Group activities by their original post content
+    type GroupedActivity = { mainActivity: Activity; reposters: { actor: Activity['actor']; activity: Activity }[] };
```
vershwal, line 38: We have done this for feed(Notes), do we also want the same for Reader(inbox)?

image

apps/admin-x-activitypub/src/views/Feed/components/FeedList.tsx
```diff
@@ -34,6 +34,49 @@ const FeedList:React.FC<FeedListProps> = ({
...
+    const consolidatedActivities = (Object.values(groupedActivities) as GroupedActivity[])
+        .map((group: GroupedActivity) => ({
+            mainActivity: group.mainActivity,
+            reposts: group.reposters.map(r => r.activity),
+            mostRecentDate: Math.max(
+                new Date(group.mainActivity.object.published || 0).getTime(),
+                ...group.reposters.map(r => new Date(r.activity.object.published || 0).getTime())
+            )
+        }))
+        .sort((a, b) => b.mostRecentDate - a.mostRecentDate);
```
vershwal, line 78: This sorting is redundant as `r.activity.object.published` is same as `group.mainActivity.object.published`.  `r.activity.object.published` is not the date/time at which the post was reposted but the date at which the original post was created. 
I think the feed will be sorted correctly even if we remove this sorting because we receive the notes in order(recent first) from backend.

PR #24259, opened 2025-07-03
🎨 Added support for "break" to the #foreach helper

ramrami:
Allows to stop subsequent iterations from running by simply using `{{@break}}`. 
Useful when checking for a condition that should match only once. 
Note that the current iteration is fully rendered, no matter where `@break` is positioned.

Usage:
```hbs
{{#foreach tags}}
    {{#match slug "~^" "news-" }}
        {{!-- Do stuff --}}
        {{#get "posts" filter="tags:{{slug}}"}}{{/get}}

        {{!-- No need to continue further. Stop the loop. --}}
        {{@break}}
    {{/match}}
{{/foreach}}
```

### DIFF
--- a/ghost/core/core/frontend/helpers/foreach.js
+++ b/ghost/core/core/frontend/helpers/foreach.js
@@ -50,6 +50,7 @@ module.exports = function foreach(items, options) {
     let output = '';
     let frame;
     let contextPath;
+    let shouldBreak = false;
 
     limit = parseInt(limit, 10) || length;
     from = parseInt(from, 10) || 1;
@@ -67,6 +68,9 @@ module.exports = function foreach(items, options) {
 
     if (data) {
         frame = createFrame(data);
+        frame.break = () => {
+            shouldBreak = true;
+        };
     }
 
     function execIteration(field, index, last) {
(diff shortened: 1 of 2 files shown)


ErisDS:
Struggling to fully understand the use case for this one.

Can we go back a step and have a chat about the problem you're trying to solve? Thanks!

ramrami:
I'm a theme developer, and sometimes I'm required to build features that can be toggled on the post level. Since Ghost doesn't support custom post data, and clients usually don't want to touch Code Injection and JavaScript code, the next best option is to use internal tags with a prefix.

This works great using `#foreach` and `#match` until someone adds multiple tags with the same prefix when they shouldn't and breaks the layout. 

Currently, I'm working around this using some CSS tricks, but ideally, I would love to avoid generating that extra HTML in the first place, especially when `#get` is used in the iteration.

Potential solutions were adding a "filter" or "break" feature to `#foreach`, I've found "break" to be easier to implement.

cathysarisky:
Just to add my two cents here:  I have run into situations several times where this would have been an easy fix to a snarly problem where you need one element from a group that meets some criterion.  

Having said that, what would be even more useful to me would be an iterator I could increment.  That would solve this problem AND the problem where I need exactly eight total posts, selected across several get requests.

PR #24261, opened 2025-07-03
Updated 'social_web_enabled' setting calculation

sagzy:

Social web feature is available for users that match all of these criteria:
1) their social web setting is toggled on in Ghost Admin (on by default in Ghost 6.0)
2) their social web labs' flag is toggled on in Ghost Admin (opt-in beta in Ghost 5.0, on-by-default GA in Ghost 6.0) 
3) their site is not hosted on localhost, an IP address, or a subdirectory
4) (Ghost (Pro) only) their access to social web is not limited by config 

Implementation notes:
- given 4), the boot order of Limit Service has changed: it's now initiated before the Settings Service, so that calculated settings such as `social_web_enabled` can use it
- in `@tryghost/limit-service@1.4.0`, we have introduced a new sync method `isDisabled()` to read whether a flag is disabled by config or not. This can be used in replacement of the existing async method `checkWouldGoOverLimit()`, which would require making an async call (and updating CalculatedField / Cache Manager / Settings Service to support async calls!)

### DIFF
--- a/ghost/core/core/boot.js
+++ b/ghost/core/core/boot.js
@@ -94,8 +94,14 @@ async function initCore({ghostServer, config, frontend}) {
     models.init();
     debug('End: models');
 
+    // Limit service is booted before settings, so that limits are available for calculated settings
+    debug('Begin: limits');
+    const limits = require('./server/services/limits');
+    await limits.init();
+    debug('End: limits');
+
     // Settings are a core concept we use settings to store key-value pairs used in critical pathways as well as public data like the site title
-    debug('Begin: settings');
+    debug('Begin: settings-service');
     const settings = require('./server/services/settings/settings-service');
     await settings.init();
     await settings.syncEmailSettings(config.get('hostSettings:emailVerification:verified'));
@@ -312,7 +318,6 @@ async function initServices() {
     const xmlrpc = require('./server/services/xmlrpc');
     const slack = require('./server/services/slack');
     const webhooks = require('./server/services/webhooks');
-    const limits = require('./server/services/limits');
     const scheduling = require('./server/adapters/scheduling');
     const comments = require('./server/services/comments');
     const staffService = require('./server/services/staff');
@@ -337,10 +342,6 @@ async function initServices() {
     const explorePingService = require('./server/services/explore-ping');
 
     const urlUtils = require('./shared/url-utils');
-
-    // NOTE: limits service has to be initialized first
-    // in case it limits initialization of any other service (e.g. webhooks)
-    await limits.init();
 
     // NOTE: Members service depends on these
     //       so they are initialized before it.
--- a/ghost/core/core/boot.js
+++ b/ghost/core/core/boot.js
@@ -101,7 +101,7 @@ async function initCore({ghostServer, config, frontend}) {
     debug('End: limits');
 
     // Settings are a core concept we use settings to store key-value pairs used in critical pathways as well as public data like the site title
-    debug('Begin: settings-service');
+    debug('Begin: settings');
     const settings = require('./server/services/settings/settings-service');
     await settings.init();
     await settings.syncEmailSettings(config.get('hostSettings:emailVerification:verified'));
--- a/ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
+++ b/ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
@@ -240,12 +240,7 @@ class SettingsHelpers {
         }
 
         // Ghost (Pro) limits
-        //
-        // Implementation note: we're only checking that the limit flag 'limitSocialWeb' exists and not awaiting for "this.limitService.checkWouldGoOverLimit('limitSocialWeb')"
-        // Reason:
-        // - calculated settings currently don't support async method calls
-        // - 'limitSocialWeb' config either exists with disabled:true (for sites that don't have access to social web), or doesn't exist at all
-        if (this.limitService.isLimited('limitSocialWeb')) {
+        if (this.limitService.isDisabled('limitSocialWeb')) {
             logging.warn('Social web is not available for Ghost (Pro) sites without a custom domain, or hosted on a subdirectory');
             return false;
         }
--- a/ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
+++ b/ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
@@ -4,6 +4,7 @@ const errors = require('@tryghost/errors');
 const EmailAddressParser = require('../email-address/EmailAddressParser');
 const logging = require('@tryghost/logging');
 const crypto = require('crypto');
+const debug = require('@tryghost/debug')('services:settings-helpers');
 
 const messages = {
     incorrectKeyType: 'type must be one of "direct" or "connect".'
@@ -229,26 +230,26 @@ class SettingsHelpers {
     isSocialWebEnabled() {
         // UI setting
         if (this.settingsCache.get('social_web') !== true) {
-            logging.warn('Social web is disabled in settings');
+            debug('Social web is disabled in settings');
             return false;
         }
 
         // Labs setting
         if (!this.labs.isSet('ActivityPub')) {
-            logging.warn('Social web is disabled in labs');
+            debug('Social web is disabled in labs');
             return false;
         }
 
         // Ghost (Pro) limits
         if (this.limitService.isDisabled('limitSocialWeb')) {
-            logging.warn('Social web is not available for Ghost (Pro) sites without a custom domain, or hosted on a subdirectory');
+            debug('Social web is not available for Ghost (Pro) sites without a custom domain, or hosted on a subdirectory');
             return false;
         }
 
         // Social web (ActivityPub) currently does not support Ghost sites hosted on a subdirectory, e.g. https://example.com/blog/
         const subdirectory = this.urlUtils.getSubdir();
         if (subdirectory) {
-            logging.warn('Social web is not available for Ghost sites hosted on a subdirectory');
+            debug('Social web is not available for Ghost sites hosted on a subdirectory');
             return false;
         }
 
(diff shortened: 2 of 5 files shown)


ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
```diff
@@ -220,16 +222,46 @@ class SettingsHelpers {
...
+        if (!this.labs.isSet('ActivityPub')) {
+            logging.warn('Social web is disabled in labs');
+            return false;
+        }
+
+        // Ghost (Pro) limits
+        if (this.limitService.limits.limitSocialWeb?.disabled) {
+            logging.warn('Social web is not available for Ghost (Pro) sites without a custom domain, or hosted on a subdirectory');
+            return false;
+        }
```
sagzy, line 247: Note for reviewer: we're not using `await this.limitService.checkWouldGoOverLimit('limitSocialWeb')` here, as we don't want to make the calculated setting async

We may want to introduce a sync `isDisabled` method into the LimitService, but this method wouldn't be available for all types of Limit (e.g. works for FlagLimit but doesn't for MaxLimit)

ghost/core/core/boot.js
```diff
@@ -94,8 +94,14 @@ async function initCore({ghostServer, config, frontend}) {
     models.init();
     debug('End: models');
 
+    // Limit service is booted before settings, so that limits are available for calculated settings
+    debug('Begin: limits');
+    const limits = require('./server/services/limits');
+    await limits.init();
+    debug('End: limits');
```
sagzy, line 101: Note for reviewer: as we're reading from the Limit Service in the calculated setting, we need to init the limit service earlier (before settings) during boot

ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
ErisDS: > as we don't want to make the calculated setting async

Can you elaborate on why this is please?

sagzy: Yes!

- `FlagLimit` has essentially only a `disabled` boolean flag, and `errorIfWouldGoOverLimit` doesn't actually need to be async ([code](https://github.com/TryGhost/SDK/blob/00b1c6b33d2d5f7ea0c7c6fb94ebf38eca548005/packages/limit-service/lib/limit.js#L293C1-L297C6))

- but the limit service only exposes async methods for compatibility with all types of limits ([checkWouldGoOverLimit](https://github.com/TryGhost/SDK/blob/00b1c6b33d2d5f7ea0c7c6fb94ebf38eca548005/packages/limit-service/lib/LimitService.js#L79))

- if we make the calculated setting `isSocialWebEnabled` async, then we'll also need to update
    - the `getSetting()` method in `CalculatedField` ([code](https://github.com/TryGhost/Ghost/blob/0db5fc52e4d8a72a77e29934075a1211bbef70e9/ghost/core/core/server/services/settings/settings-service.js#L54))
    - the CacheManager code ([code](https://github.com/TryGhost/Ghost/blob/0db5fc52e4d8a72a77e29934075a1211bbef70e9/ghost/core/core/shared/settings-cache/CacheManager.js#L97C1-L102C6))
    - have Setting Service init method wait for all calculated fields to be populated ([code](https://github.com/TryGhost/Ghost/blob/5d5a7cfdde7eb47ae5b1424937f15b2d05cf3e65/ghost/core/core/server/services/settings/settings-service.js#L72C1-L80C7))
    

So doable all in all, but feels like adding unnecessary complexity for something that can be solved with a sync read of the `disabled` field directly

ErisDS: Ok, that makes sense. 

Still, I feel like checking the internals is code smell. 

One way around this might be to use `isLimited()` rather than the internals, because I believe in this case the limit is only ever `disabled: true` or not present at all. 

https://github.com/TryGhost/SDK/blob/main/packages/limit-service/lib/LimitService.js#L68

sagzy: Ah, if we can safely make that assumption, this simplifies things! @aileen tagging you here as well for visibility, I'll make the change in the meantime

ghost/core/test/unit/server/services/settings-helpers/settings-helpers.test.js
```diff
@@ -244,11 +247,99 @@ describe('Settings Helpers', function () {
...
+        it('returns false if the site is hosted on a localhost or IP address in production', function () {
+            urlUtils.getSiteUrl.returns('http://localhost:2368');
+            sinon.stub(process.env, 'NODE_ENV').value('production');
+
+            const settingsHelpers = new SettingsHelpers({settingsCache, config, urlUtils, labs, limitService});
+            const isEnabled = settingsHelpers.isSocialWebEnabled();
+
+            should.equal(isEnabled, false);
+            sinon.assert.calledWith(loggingSpy, 'Social web is not available for localhost or IPs addresses in production');
+        });
```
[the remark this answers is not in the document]
sagzy: Wouldn't `sinon.restore()` in the `afterEach()` block in line 283 be sufficient?

ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
sagzy: Updated in e6e720e9d9e1bc352601388093569efd99780be9

aileen: The short answer is: for a `Flag` limit, yes, this would work. 

However, this would make it impossible to set it as disabled, as we wouldn't be able to check the actual value. The sheer presence of the limit implies that the limit should be enforced. My only concern is that within Ghost, it's not clear that we're dealing with a `Flag` limit here, where this method would work, and we're introducing a pattern that might be used for other limits where it shouldn't 🤔

sagzy: Another solution can be:
- introduce a `isDisabled()` sync method on the Limit Service
- method would return a boolean, reading the `disabled` field, if used for a limit that support it (e.g. Flag)
- method would throw a `IncorrectUsageError`, if used for a limit that doesn't support it (e.g. Max)

sagzy: What do you think of that?

aileen: That would make sense, especially since we're going to have two more limits of that kind in the codebase soon. 

The `errorIfWouldGoOverLimit` is not very useful for a `Flag` limit, and it's strange that it's `async` as it shouldn't need to be. We need to be cautious for the `email` limit, which is a hybrid and _can_ be a numeric or a flag limit, and had to revert those changes. I stumbled over that limit and had failing regression tests, forcing me to revert my changes, when I tried to introduce a change for the `Flag` limit behaviour.

sagzy: Nice that failing tests caught it!

Ok I'll go ahead and open a PR on the limit service with that change then and will request a review :)

sagzy: Updated now with the new `.isDisabled()` sync check, introduced in `@tryghost/limit-service@1.4.0`

ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
```diff
@@ -220,16 +222,46 @@ class SettingsHelpers {
...
+
+        // Ghost (Pro) limits
+        if (this.limitService.isDisabled('limitSocialWeb')) {
+            logging.warn('Social web is not available for Ghost (Pro) sites without a custom domain, or hosted on a subdirectory');
+            return false;
+        }
+
+        // Social web (ActivityPub) currently does not support Ghost sites hosted on a subdirectory, e.g. https://example.com/blog/
+        const subdirectory = this.urlUtils.getSubdir();
+        if (subdirectory) {
```
aileen, line 251: We currently have two sites with a subdirectory that have ActivityPub enabled in labs and are on the grandfathered list for that reason. Those two sites won't have the limit applied because they are on that grandfathered list. 

I wonder if we 
- Have to disable the setting for those sites
- Remove them from the grandfather list
- Communicate the reason why we disabled it with them

Probably something to discuss in Slack.

aileen (approved):
I have added a question, but otherwise LGTM and very readable and maintainable.

ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
```diff
@@ -220,16 +222,46 @@ class SettingsHelpers {
...
+        // Self-hosters cannot connect to production ActivityPub servers from localhost or IPs addresses
+        const siteUrl = new URL(this.urlUtils.getSiteUrl());
+        const isLocalhost = siteUrl.hostname === 'localhost' || siteUrl.hostname === '127.0.0.1' || siteUrl.hostname === '::1';
+        const isIP = net.isIP(siteUrl.hostname);
+        if (process.env.NODE_ENV === 'production' && (isLocalhost || isIP)) {
+            logging.warn('Social web is not available for localhost or IPs addresses in production');
+            return false;
+        }
+
+        return true;
```
ErisDS, line 265: I really like the structure of the logic here 👍 

However, I feel like having WARN level logging every time the setting is evaluated is overkill. Using our `debug()` tool would be great here instead of logging and we can trust the UI to communicate with the user.

ErisDS (commented):
The logic all LGTM but I feel the warning logs are unnecessary.

ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
sagzy: Updated in 9b8297a4901b68c301625995e5fe7f7c7a6afffb

ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
sagzy: Discussed in Slack, resolving

PR #24266, opened 2025-07-03
🎨 Added more social platforms to sameAs generation

troyciesco:
closes https://linear.app/ghost/issue/ENG-2445/add-additional-social-urls-to-sameas-for-schema-validation
reported here https://forum.ghost.org/t/loving-the-new-added-social-links-but/57018

- Adds the new social platforms (i.e. other than Facebook and X) to sameAs generation for schema.org validation
- This is for the primary authors on posts as well as on individual author pages
- Minor tidying, and added tests.

output from schema.org for a post:
Screenshot 2025-07-07 at 9 18 36 AM

output from schema.org for an author page:
Screenshot 2025-07-07 at 9 18 16 AM


cathysarisky:
Ooooh, excited to see this coming through. I was just thinking the other day that someone needed to do it.  😄 

I notice that the schema currently only write out the primary author, and that pattern is retained here.  Would it make sense to write out an array of authors?  I guess maybe that's a separate PR, though.

Oh, and hey, nice to meet you! 👋

troyciesco:
> I notice that the schema currently only write out the primary author, and that pattern is retained here. Would it make sense to write out an array of authors? I guess maybe that's a separate PR, though.

@cathysarisky I wondered that too! I think you're right, it does make sense. I'm going to make a follow up ticket for it.
 
> Oh, and hey, nice to meet you! 👋

Nice to meet you too! 👋

ghost/core/core/frontend/meta/schema.js
```diff
@@ -48,31 +48,33 @@ function trimSchema(schema) {
...
-        if (data.author.twitter) {
-            sameAs.push(socialUrls.twitter(data.author.twitter));
-        }
+    if (author.website) {
+        sameAs.push(escapeExpression(author.website));
     }
 
+    SOCIAL_PLATFORMS.forEach((platform) => {
+        if (author[platform] && typeof socialUrls[platform] === 'function') {
+            sameAs.push(socialUrls[platform](author[platform]));
```
kevinansfield: I know it wasn't used before but it feels odd that we're not using the same `escapeExpression()` call here as we use for the website URL

troyciesco: yeah good call! added it + a test

kevinansfield (approved):

PR #24272, opened 2025-07-04
🐛Fixed internal search bug that makes highlighted text unclickable.

### LINKED ISSUE #24269: Internal search - highlighted portion is not clickable (Chrome bug, not on Firefox)
### Issue Summary

Image

Using the internal search and starting the search with a term that exactly matches a post title results in yellow-highlighted text.  The yellow-highlighted text is not clickable.  The grey area around it is.

The structure of the page in this region is: 

Image

Nothing in the span is clickable.  Text outside the span within the li is clickable.

Possibly a problem with event listener propagation/bubbling?

Expected behavior:  Clicking anywhere within a search result should load that result.

### Steps to Reproduce

Use Chrome on Windows.  (Not sure if the behavior is present on Mac.)

Go to /ghost.  
Identify a post title.  
Click the internal search icon, and type the start of the post name into the search box.  
Attempt to click the yellow-highlighted text.

Expected behavior: Clicking takes me to the search result
Actual behavior:  The yellow portion does not respond to clicks.  Only the surrounding grey portion does.

### Ghost Version

Ghost Pro (and self hosted, and local)

### Node.js Version

Ghost Pro

Ghost Pro

### Database type

MySQL 8

### Browser & OS version

Chrome, recent version, Windows 11.  The problem is not present in Firefox.

```shell
No console errors are present.

```

### Code of Conduct

cathysarisky:
closes https://github.com/TryGhost/Ghost/issues/24269

Adding a pointer-events: none to prevent the span of highlighted text from interfering with click events on the containing li.

Note: This is a Chrome-specific bug. It isn't present on Firefox.  Observed on Windows 11.

Review question: I've had past trouble with off-target issues when adjusting ember css. Is there a better selector to add this change to, ensuring that it only impacts the search box, and not other uses of power select?

(Sorry for the duplicate PR - I needed to open it from the Ghost repo rather than my branch so that browser tests could run.)

### DIFF
--- a/ghost/admin/app/styles/components/power-select.css
+++ b/ghost/admin/app/styles/components/power-select.css
@@ -210,6 +210,7 @@
     background: #fff6b8;
     border-radius: 1px;
     color: color-mod(var(--darkgrey) l(-10%));
+    pointer-events: none;
 }
 
 .ember-power-select-group .ember-power-select-option[aria-current="true"] {
--- a/ghost/admin/tests/acceptance/search-test.js
+++ b/ghost/admin/tests/acceptance/search-test.js
@@ -145,6 +145,19 @@ suites.forEach((suite) => {
             expect(currentURL(), 'url after selecting post').to.equal(`/editor/post/${firstPost.id}`);
         });
 
+        it('navigates to editor when highlighted text is clicked', async function () {
+            await visit('/dashboard');
+            await click('[data-test-button="search"]');
+            await typeInSearch('first post');
+            
+            // Find the highlighted text span and click on it specifically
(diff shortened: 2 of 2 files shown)


allouis (approved):

kevinansfield:
> Review question: I've had past trouble with off-target issues when adjusting ember css. Is there a better selector to add this change to, ensuring that it only impacts the search box, and not other uses of power select?

In this case I think your selector will be fine, the `.highlight` class usage comes from `{{highlighted-text}}` helper which is only used inside the search input. _If_ we started using it in other power select based components we'd run into the same issue so fixing it in one place makes sense 👍

kevinansfield (approved):

PR #24274, opened 2025-07-04
Added missing TR translations for Comments

muratcorlu:
After seeing @cathysarisky 's call for improving translations, I noticed some missing ones in Turkish translation.


cathysarisky:
Hi @EchoBilisim , do you have time to review these Turkish translations?  Also, I think it'd be great if you and @muratcorlu could have some discussion somewhere about whether the Turkish translations should be using the formal or informal form, and then we could standardize throughout.  Thanks!

muratcorlu:
#24277 is more complete over this. Let's focus on that PR. I'm closing this.

PR #24276, opened 2025-07-04
fixed unused translation ignores not being reported

vikaspotluri123:
- the previous implementation had early returns to avoid unnecessary analysis
- however, this means there's no way for us to know if an ignored rule still applies
- removed the early returns
- added a new `reportTranslationError` method to only report an error if it's not ignoredGot some code for us? Awesome 🎊!

### DIFF
--- a/ghost/i18n/test/i18n.lint.js
+++ b/ghost/i18n/test/i18n.lint.js
@@ -158,13 +158,19 @@ class LinterContext {
     }
 
     /**
-     * @param {GhostI18nLintRule} rule
-     * @param {string} key
-     */
-    isIgnored(rule, key) {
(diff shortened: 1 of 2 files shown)


ghost/i18n/test/i18n.lint.js
```diff
@@ -418,28 +425,23 @@ function analyzeSingleTranslation(key, translated, context) {
 
     const {errors, variables: used} = parseTranslationString(translated);
 
-    if (reportInvalidTranslations) {
+    {
```
vikaspotluri123, line 428: This block needs to be retained to avoid redeclaration of const variables. All of the other cases are nested under an if/for bock

cathysarisky (approved):
LGTM. Thanks for fixing the bug!

PR #24279, opened 2025-07-05
Update e2e tests

ibalosh:
ref [23941](https://github.com/TryGhost/Ghost/pull/23941)

* Simplified the initial e2e test and corrected some of the logic there
* reason for the change is to make sure we use the initial e2e test suite correctly - like separating domains (asserting in tests, navigating through complex logic in UI through page objects)


ibalosh:
Hey @cmraible , I made some small changes to e2e structure, lmk if this works:

- removed things we don't need yet, we can add them as we go if we need them
- structured the code slightly different, with helpers, tests, lmk what you think
- updated readme

e2e/package.json
```diff
@@ -12,9 +12,8 @@
     "build:ts": "tsc",
     "prepare": "tsc",
     "test": "playwright test",
-    "test:frontend": "playwright test test/frontend",
     "test:types": "tsc --noEmit",
-    "lint": "eslint src/ --ext .ts --cache"
+    "lint": "eslint helpers --ext .ts --cache"
```
cmraible, line 16: Not a blocker for this PR, but this reminded me: should we also lint our tests?

In our CI workflow, we run `yarn lint` for all the packages in the monorepo, so anything we add to this `lint` script will be run on every PR. We have a `tests/.eslintrc.js` file but as of yet we're not using it.

e2e/playwright.config.mjs
```diff
@@ -5,15 +5,14 @@ const config = {
...
     reporter: process.env.CI ? [['list', {printSteps: true}], ['html']] : [['list', {printSteps: true}]],
     use: {
         baseURL: process.env.GHOST_BASE_URL || 'http://localhost:2368',
         trace: 'retain-on-failure',
-        browserName: 'chromium',
-        headless: !process.env.PLAYWRIGHT_DEBUG
+        browserName: 'chromium'
     },
-    testDir: './test'
+    testDir: './tests'
```
cmraible, line 15: So simple, but this makes the package's intent make so much more sense 🙏

cmraible (approved):
Thanks @ibalosh, this all looks great! I really like the rename from `test` to `tests` — it's so simple but somehow that makes it so much more clear that these are "tests in the e2e test suite" rather than "tests _for_ the e2e test suite". And totally agree that `helpers` makes for sense than `src` in this context  🙌 

I left one comment/question on linting, but I don't think that needs to be covered in this PR. 

My only other note is that the PR title / commit message should follow the guidelines in `CONTRIBUTING.md` (i.e. past tense, a bit more descriptive) but I'm sure you were probably going to do before merging anyways — flagging it just in case.

Overall LGTM!

e2e/package.json
ibalosh: I think that we should. It's still code, and should follow conventions. Lint rules might be lighter than for the other code, but still super valuable (I use it a lot of time when I am lazy formatting code :) )

e2e/playwright.config.mjs
ibalosh: yes!

ibalosh:
> Thanks @ibalosh, this all looks great! I really like the rename from `test` to `tests` — it's so simple but somehow that makes it so much more clear that these are "tests in the e2e test suite" rather than "tests _for_ the e2e test suite". And totally agree that `helpers` makes for sense than `src` in this context 🙌
> 
> I left one comment/question on linting, but I don't think that needs to be covered in this PR.
> 
> My only other note is that the PR title / commit message should follow the guidelines in `CONTRIBUTING.md` (i.e. past tense, a bit more descriptive) but I'm sure you were probably going to do before merging anyways — flagging it just in case.
> 
> Overall LGTM!

thanks @cmraible ! thats exactly what I am aiming for on these initial PR's - simple changes, to align better, that we can start using across codebase. We can go big as we make progress

Yeah, commit message will be updated on the merge - thanks for reminder though, appreciated!

PR #24285, opened 2025-07-07
Added yarn resolution for ember-cli-htmlbars

kevinansfield:
no issue

- we have multiple versions of `ember-cli-htmlbars` used across our dependencies which resulted in package resolution warnings during build
- added yarn resolution for our latest-used version so we're not pulling in multiple of the same dependency
- there are no breaking changes across versions of `ember-cli-htmlbars` other than dropping old node version support meaning forcing a newer version resolution has low risk


kevinansfield:
Closing for now pending further investigation into the source of Admin build warnings about dependency resolution.

PR #24286, opened 2025-07-07
Updated tags input to using client-side filtering when all tags are loaded

kevinansfield:
no issue

- if all tags have been loaded from the server we can rely on our faster client-side filtering behaviour rather than using server-side filtering
  - for most sites this means client-side search will always be used because >100 tags isn't that common
- updated `<GhTagsTokenInput>` to disable server-side search when:
  - all tags were loaded in the first populate-dropdown request
  - all tags have subsequently been loaded by scrolling through every page

### DIFF
--- a/ghost/admin/app/components/gh-tags-token-input.hbs
+++ b/ghost/admin/app/components/gh-tags-token-input.hbs
@@ -9,9 +9,11 @@
     @optionsComponent={{component "power-select-vertical-collection-options" lastReached=(perform this.loadMoreTagsTask)}}
     @registerAPI={{this.registerPowerSelectAPI}}
     @renderInPlace={{@renderInPlace}}
-    @search={{perform this.searchTagsTask}}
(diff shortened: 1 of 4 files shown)


ghost/admin/app/components/gh-tags-token-input.hbs
```diff
@@ -9,9 +9,9 @@
     @optionsComponent={{component "power-select-vertical-collection-options" lastReached=(perform this.loadMoreTagsTask)}}
     @registerAPI={{this.registerPowerSelectAPI}}
     @renderInPlace={{@renderInPlace}}
-    @search={{perform this.searchTagsTask}}
+    @search={{if this.useServerSideSearch (perform this.searchTagsTask) null}}
```
kevinansfield, line 13: Providing `null` here means `<GhTokenInput>` falls back to it's client-side filtering

allouis (approved):

PR #24290, opened 2025-07-08
🎨 Added searchable dropdown with custom option for publication language

betschki:
Replaced text input with SelectWithOther component to prevent invalid locale entries. Users can select from 64 predefined locales or enter custom BCP 47 compliant codes via "Other" option.

This addresses the issue where users are entering invalid values like "English" instead of proper locale codes.


betschki:
Fixing linting and tests with CodeRabbit's comment

betschki:
A few screenshots for reference:

This is how it looks like if a "custom locale" is entered. It would also show up like this if a user has set a "custom locale" in Ghost before this potential merge.
image

This is the dropdown (searchable):
image

Selecting "Other" from the list switches from the dropdown to a (validated) text input field:
image

The input field is validated against BCP 47:
image

An empty string is also not possible:
image

apps/admin-x-settings/src/utils/localeValidation.ts
```diff
@@ -0,0 +1,149 @@
...
+        'en-gb-oed', 'i-ami', 'i-bnn', 'i-default', 'i-enochian', 'i-hak',
+        'i-klingon', 'i-lux', 'i-mingo', 'i-navajo', 'i-pwn', 'i-tao',
+        'i-tay', 'i-tsu', 'sgn-be-fr', 'sgn-be-nl', 'sgn-ch-de'
+    ];
+    
+    // Grandfathered regular tags (exact matches, case insensitive)
+    const grandfatheredRegular = [
+        'art-lojban', 'cel-gaulish', 'no-bok', 'no-nyn', 'zh-guoyu',
+        'zh-hakka', 'zh-min', 'zh-min-nan', 'zh-xiang'
+    ];
```
kevinansfield, line 26: What is the source of these grandfathered tags? I think it would be useful to add a bit more explanation to the comments to help future code explorers

apps/admin-x-design-system/src/global/form/SelectWithOther.tsx
```diff
@@ -0,0 +1,156 @@
...
+    const isCustomValue = selectedValue &&
+        !options.some(opt => opt.value === selectedValue) &&
+        selectedValue !== otherOption.value;
+
+    // Initialize state based on current value
+    useEffect(() => {
+        if (isCustomValue) {
+            setIsOtherSelected(true);
+        }
+    }, [isCustomValue]);
```
kevinansfield, line 50: Using an effect to set state like this is a bit of a React anti-pattern. Perhaps we can move the derived `isCustomValue` calculation above the `useState` hooks and switch to `useState(isCustomValue)` for that initial setup?

kevinansfield (changes requested):
A couple of minor things added in comments above but otherwise this looks+works great! 🙌

betschki:
Thank you @kevinansfield! Added documentation for the language tags and implemented your suggestion

apps/admin-x-design-system/src/global/form/SelectWithOther.tsx
```diff
@@ -97,8 +91,13 @@ const SelectWithOther: React.FC<SelectWithOtherProps> = ({
         return hasOtherOption ? options : [...options, otherOption];
     }, [options, otherOption]);
 
+    // Check if current value is custom (for determining display mode)
+    const isCurrentlyCustomValue = selectedValue &&
+        !options.some(opt => opt.value === selectedValue) &&
+        selectedValue !== otherOption.value;
+
```
kevinansfield: Is this duplication needed? Unless I'm missing something it's the same calculation as `isInitiallyCustomValue` so both will have exactly the same value in any call of the component function.

betschki: Fixed! :+1:

kevinansfield:
@betschki thanks for making those changes 🙌  Everything looks good to merge now but the branch has conflicts that need resolving

betschki:
@kevinansfield well, this just got messy, trying to clean things up a bit.

I have opened a new branch with all the feedback implemented: https://github.com/TryGhost/Ghost/pull/24673
To confirm I ran `git diff de389f5e47..publication-language-select-other`, which returned no result. `de389f5e47` being the commit that removed the duplication you found in https://github.com/TryGhost/Ghost/pull/24290#pullrequestreview-3116026024

PR #24295, opened 2025-07-08
Added tinybird-local and mysql services to e2e test suite

cmraible:
ref https://linear.app/ghost/issue/PROD-2250/setup-e2e-tests-in-ci-to-work-with-tinybird-local-container-with

We're working toward adding end-to-end tests for traffic analytics, and that means we'll need to have a Tinybird instance running _somewhere_ that we can send test events to, and configure Ghost & Admin to use the same instance when making requests to our Tinybird endpoints. This commit sets up a tinybird local container in the e2e test job, and adds a script that sets all the required configuration for Ghost to use this local instance of Tinybird. Using Tinybird local has a few benefits over using a staging workspace:
- we can test out changes to our Tinybird schema, which will be deployed to the tinybird local container, then run our e2e tests against the updated schema. This would be impractical if using a staging workspace.
- Requests should be a lot faster, since there should be minimal network latency over the local Docker network versus making external API requests to Tinybird cloud.

We'll also need to generate test data in Ghost's primary relational database for things like posts and members. The e2e test suite has been using `sqlite3` simply because it was easier to setup initially. Really though it would be a more accurate representation of production to run these tests using MySQL as the database, hence adding the mysql service to the e2e test job. This will also allow us to access Ghost's database directly from our tests, so we can insert/reset data in Ghost's DB without piercing Ghost's black box.


.github/workflows/ci.yml
```diff
@@ -827,7 +857,7 @@ jobs:
 
       - name: Start Ghost
         working-directory: ghost/core
-        run: DEBUG=* yarn dev &
```
cmraible, line 830: Removed debug logging now that these tests are passing consistently. Removed the `--watch` included in `yarn dev` since it adds unnecessary overhead since we're not changing files in CI.

.github/workflows/ci.yml
```diff
@@ -818,6 +838,16 @@ jobs:
...
 
+      - name: Install Tinybird CLI
+        run: curl -fsSL https://tinybird.co/install.sh | sh
+
+      - name: Deploy Tinybird Schema
+        working-directory: ghost/core/core/server/data/tinybird
+        run: tb build
+
+      - name: Get Tinybird configuration
+        run: ghost/core/core/server/data/tinybird/scripts/configure-ghost.sh
```
cmraible: Adds tinybird configuration to Ghost by exporting environment variables

ghost/core/core/server/data/tinybird/scripts/configure-ghost.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Get workspace admin token
+## This token is available in the Tinybird Local instance without authentication
+WORKSPACE_ADMIN_TOKEN=$(curl -s http://localhost:7181/tokens | jq -r ".admin_token")
+export TB_TOKEN="$WORKSPACE_ADMIN_TOKEN"
+
+# Get the admin token from the Tinybird CLI
+## This is different from the workspace admin token
+ADMIN_TOKEN=$(tb token ls --match "admin token" | grep "token:" | awk -F'token: ' '{print $2}')
+export TB_ADMIN_TOKEN="$ADMIN_TOKEN"
```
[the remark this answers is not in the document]
cmraible: --json isn't a valid option in the latest Tinybird CLI version.

cmraible:
You can see the correct environment variables for Ghost's config being set here: https://github.com/TryGhost/Ghost/actions/runs/16160890696/job/45612362625?pr=24295#step:12:15

9larsons (approved):

PR #24306, opened 2025-07-09
Added stats CTA for trial/starter plan customers

peterzimon:
ref https://linear.app/ghost/issue/PROD-2099/add-cta-for-starter-plan-trial-plan-customers
ref https://linear.app/ghost/issue/PROD-52/self-hoster-ctas

Customers on trial or on the Starter plan will have no access to traffic analytics. For them we're going to show a CTA on the Analytics overview to upgrade their plan.


apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
```diff
@@ -68,18 +83,27 @@ const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
                             handleToggleChange('web_analytics', e);
                         }}
                     />
-                    {ui60 && !isWebAnalyticsConfigured ?
-                        <div className='mb-5 rounded-md border border-grey-200 bg-grey-50 px-4 py-2.5'>
-                            <span className='flex items-start gap-2'>
-                                {/* <Icon className='-mt-px text-black' name='info-fill' size={24} /> */}
-                                <span>
-                            Web analytics in Ghost is powered by <a className='text-green underline' href="https://tinybird.co" rel="noopener noreferrer" target='_blank'>Tinybird</a> and requires configuration to start collecting data. <a className='text-green underline' href="https://ghost.org/docs/" rel="noopener noreferrer" target='_blank'>Get started &rarr;</a>
+                    {(
```
9larsons, line 86: Removed ui60 flag - that seems to only control if we show hints, and didn't seem necessary to hide the config/cta boxes considering it's still behind the beta flag.

apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx
```diff
@@ -4,15 +4,21 @@ import React, {useEffect, useState} from 'react';
 import {HostLimitError, useLimiter} from '../../../../hooks/useLimiter';
 import {List} from '@tryghost/admin-x-design-system';
 
-const features = [{
+type Feature = {
```
9larsons, line 7: I had to add types to satisfy the typescript compiler. We aren't currently using this (I removed the limit from the flags, as it didn't make a whole lot of sense to me), so this limiter is here but not limiting anything in this page. It felt worth keeping here as an aid to future flags.

Personally I'm not sure why we'd hide beta features behind limits. I suppose it's extra protections, although it makes testing the various flag combinations difficult.

apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
```diff
@@ -68,18 +83,27 @@ const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
...
+                        isWebAnalyticsLimited ? (
+                            <div className='mb-5 rounded-md border border-grey-200 bg-grey-50 px-4 py-2.5'>
+                                <span className='flex items-start gap-2'>
+                                    <span>
+                                        Get the full picture of what&apos;s working with detailed, cookie-free traffic analytics. <span className='text-green underline' onClick={() => updateRoute({route: '/pro', isExternal: true})}>Upgrade now &rarr;</span>
+                                    </span>
+                                </span>
+                            </div>
+                        ) : !isWebAnalyticsConfigured ? (
```
9larsons, line 95: It seems fine to display one message OR the other, and was cleaner to read for me.

apps/stats/src/hooks/useLimiter.ts
```diff
@@ -0,0 +1,32 @@
+import {useGlobalData} from '@src/providers/GlobalDataProvider';
```
9larsons, line 1: This is a pretty basic hook and the one in Settings has a few extra layers (context) that is specific to that app. I think we could possibly create a simple version like this in framework and then have apps like Settings extend... or we can just use this. I didn't feel strongly about moving everything about.

apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx
cmraible: Agree that I can't think of a reason we would want to hide a feature that is already in the PrivateFeatures list. That said, this change feels a little out of place in this PR — took me a few minutes to reason about what was happening here, since it's unrelated to the CTA change. Not saying you should remove it now, just mentioning it for the future 😃

apps/admin-x-settings/test/acceptance/membership/analytics.test.ts
```diff
@@ -439,10 +439,230 @@ test.describe('Analytics settings', async () => {
...
+                    ...responseFixtures.settings,
+                    settings: [
+                        ...responseFixtures.settings.settings,
+                        {key: 'web_analytics', value: false},
+                        {key: 'web_analytics_enabled', value: false},
+                        {key: 'web_analytics_configured', value: true}
+                    ]
+                }
+            }
+        }});
```
cmraible: I'd love to find a way to clean this up — it's pretty hard to read this at a glance and understand what all the values are set to. I know this is because the mockApi function is a mess in its own right, but maybe we could consolidate some of this into a helper function just for this file, since we're setting mostly the same things in this file. i.e. something like:

```
setupApi({
    "labs": {
        "trafficAnalytics": true,
        "ui60": true
    },
    "limitAnalytics": true,
    "settings": {
        "web_analytics": false,
        "web_analytics_enabled": false,
        "web_analytics_configured": true
    }
});
```

apps/admin-x-settings/test/acceptance/membership/analytics.test.ts
```diff
@@ -439,10 +439,230 @@ test.describe('Analytics settings', async () => {
...
+        // Click should trigger navigation
+        await upgradeLink.click();
+        
+        // Wait for navigation
+        await page.waitForURL(url => url.toString() !== currentUrl, {timeout: 1000}).catch(() => {
+            // Navigation might be blocked in test environment
+        });
+        
+        // In a real environment, this would navigate to /pro
+        // The click handler is set up correctly with updateRoute({route: '/pro', isExternal: true})
```
cmraible: Bit confused — So this is just testing that we navigate ~somewhere else~?

apps/stats/src/views/Stats/Overview/components/OverviewKPIs.tsx
```diff
@@ -138,12 +141,18 @@ const OverviewKPIs:React.FC<OverviewKPIsProps> = ({
     }
 
     let cols = 'lg:grid-cols-3';
-    if ((appSettings?.analytics.webAnalytics && !appSettings?.paidMembersEnabled) ||
-        (!appSettings?.analytics.webAnalytics && appSettings?.paidMembersEnabled)) {
+    
+    // Calculate number of cards being displayed
+    const hasWebAnalytics = appSettings?.analytics.webAnalytics && !showTrialCTA;
```
cmraible: I forget already  — `appSettings.analytics.webAnalytics` is the "One Setting", or no? Do we need the `&& !showTrialCTA` here?

apps/stats/src/views/Stats/Overview/components/OverviewKPIs.tsx
```diff
@@ -138,12 +141,18 @@ const OverviewKPIs:React.FC<OverviewKPIsProps> = ({
...
+    const hasTrialCTA = showTrialCTA;
+    const hasMembers = true; // Always shown
+    const hasMRR = appSettings?.paidMembersEnabled && !showTrialCTA;
```
cmraible: We don't show MRR if the user is on a trial plan? I thought we were only hiding specifically web analytics if `limitAnalytics` is set, but maybe I missed that somewhere.

apps/stats/src/views/Stats/Overview/components/OverviewKPIs.tsx
```diff
@@ -126,6 +127,8 @@ const OverviewKPIs:React.FC<OverviewKPIsProps> = ({
     const navigate = useNavigate();
     const {range} = useGlobalData();
     const {appSettings} = useAppContext();
+    const limiter = useLimiter();
+    const showTrialCTA = limiter.isLimited('limitAnalytics');
```
cmraible: It's not really a trial CTA, right? Maybe `showUpgradeCTA` would make this clearer (unless I'm wrong about the limit being applied to starter plans)

cmraible (commented):
I left a few questions, mainly because I'm not 100% clear on the requirements.

apps/stats/src/views/Stats/Overview/components/OverviewKPIs.tsx
9larsons: Agreed - @peterzimon you started with trial here, does this make sense as upgrade?

apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx
9larsons: I had to change it a little bit because otherwise the typescript wouldn't compile. I agree it feels a bit out of place... alternatively I could leave the limit, but that doesn't feel right either since I don't totally agree with them being there and it makes it challenging to test this when the flag is tied to the limit.

apps/stats/src/views/Stats/Overview/components/OverviewKPIs.tsx
9larsons: See above.

apps/stats/src/views/Stats/Overview/components/OverviewKPIs.tsx
9larsons: Yeah this is a mess, and there was some stuff that Zimo had behind the cta before that I didn't quite understand. I redid this so going to mark this resolved, and the next one, and please take another pass.

apps/stats/src/views/Stats/Overview/components/OverviewKPIs.tsx
```diff
@@ -136,22 +139,27 @@ const OverviewKPIs:React.FC<OverviewKPIsProps> = ({
             </EmptyCard>
         );
     }
-
+
```
9larsons, line 142: This is the new block, @cmraible - hopefully this makes more sense, it's the best I could come up with at this point in the day.

9larsons: The added confusion is that we should display the CTA if the limit is present, and otherwise fall back to displaying the Visit Counts component if enabled/configured.

apps/stats/src/views/Stats/Overview/components/OverviewKPIs.tsx
9larsons: Actually, I updated this to upgrade, it just makes more sense in the rest of the logic.

apps/admin-x-settings/test/acceptance/membership/analytics.test.ts
9larsons: I couldn't find a good way to test this. I changed it to just perform the click and not error. We send the external navigate to the parent app and there isn't a current pattern for handling the external navigate. I can look into this more Monday if you want to make this a more comprehensive test.

apps/admin-x-settings/test/acceptance/membership/analytics.test.ts
9larsons: Good call - it's pretty awful. Cleaned up significantly.

apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
peterzimon: @9larsons — the intention of the original copy ("Web analytics is available on the Publisher plan and above. Upgrade now ->") here is not necessarily to drive conversion but to explain why the toggle is off. I'd suggest to use it.

9larsons: Updated.

apps/admin-x-settings/test/acceptance/membership/analytics.test.ts
```diff
@@ -117,33 +167,14 @@ test.describe('Analytics settings', async () => {
...
-            }
-        }});
+        await mockApi({page, requests: createMockApiConfig({
+            trafficAnalytics: true,
+            analyticsSettings: [
+                {key: 'web_analytics', value: true},
+                {key: 'web_analytics_enabled', value: true},
+                {key: 'web_analytics_configured', value: true}
+            ]
+        })});
```
cmraible, line 177: Soooooo much better 🙌

apps/stats/src/views/Stats/Overview/components/OverviewKPIs.tsx
```diff
@@ -136,22 +139,26 @@ const OverviewKPIs:React.FC<OverviewKPIsProps> = ({
...
-    if ((appSettings?.analytics.webAnalytics && !appSettings?.paidMembersEnabled) ||
-        (!appSettings?.analytics.webAnalytics && appSettings?.paidMembersEnabled)) {
+    if (cardCount === 2) {
         cols = 'lg:grid-cols-2';
-    }
-
-    if (!appSettings?.analytics.webAnalytics && !appSettings?.paidMembersEnabled) {
+    } else if (cardCount === 1) {
         cols = 'lg:grid-cols-1';
     }
```
cmraible, line 156: If we're happy with how this is working right now, I think we should leave as is until we have better tests to refactor it. That said, here's I think a cleaner / more explicit way to do this:

```
// Calculate number of cards being displayed
    const showWebAnalytics = appSettings?.analytics.webAnalytics;
    const showUpgradeCTA = isWebAnalyticsLimited && !showWebAnalytics;

    const showWebAnalyticsCard = (showWebAnalytics || showUpgradeCTA)
    const showMembersCard = true; // Always shown
    const showMRRCard = appSettings?.paidMembersEnabled;
    
    // Determine number of columns to display, 1, 2, or 3
    const cardCount = [showWebAnalyticsCard, showMembersCard, showMRRCard].filter(Boolean).length;
    const cols = `lg:grid-cols-${cardCount}`;
```

cmraible:
Not a blocker, but the cursor should change when hovering over the "Upgrade now" link in Settings > Analytics. Personally fine with merging and fixing this in a follow up, but should be a quick fix either way. 

Screenshot 2025-07-14 at 09 59 44

apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx
cmraible: Yeah no worries — it's a little thing that's not really important. Just to clarify what I meant though:

> alternatively I could leave the limit, but that doesn't feel right either since I don't totally agree with them being there and it makes it challenging to test this when the flag is tied to the limit.

I would have left the limit in this PR to keep it focused, then came back and just reverted the whole commit that added the limit in a follow up PR. Re: testing it locally, I probably would have made this change locally if it made things easier, but not included the change in the git diff to distract from the intent of this commit.

cmraible (approved):
Left a Tiny nitpick comment here, but I'm happy for that to be cleaned up in a follow up PR. Looks good to merge from my POV 👍

PR #24311, opened 2025-07-09
Wired up calculated field for Web Analytics setting

9larsons:
ref https://linear.app/ghost/issue/PROD-2162/
- limit service, config, flag, and setting are all now routed through a SettingsHelper tag as a calculated field
- frontend now disables the setting when config or limit service are disabling it
- labs setting disables the functionality in the background AND hides the setting in the frontend

### DIFF
--- a/apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
+++ b/apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import TopLevelGroup from '../../TopLevelGroup';
 import useFeatureFlag from '../../../hooks/useFeatureFlag';
 import useSettingGroup from '../../../hooks/useSettingGroup';
-import {Button, Separator, SettingGroupContent, Toggle, withErrorBoundary} from '@tryghost/admin-x-design-system';
+import {Button, Icon, Separator, SettingGroupContent, Toggle, withErrorBoundary} from '@tryghost/admin-x-design-system';
 import {getSettingValues, isSettingReadOnly} from '@tryghost/admin-x-framework/api/settings';
 import {usePostsExports} from '@tryghost/admin-x-framework/api/posts';
 
@@ -61,14 +61,25 @@ const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
                         direction='rtl'
                         disabled={!isWebAnalyticsEnabled}
                         gap='gap-0'
-                        hint={isWebAnalyticsEnabled ? (ui60 ? 'Cookie-free, first party traffic analytics for your site' : undefined) : 'Web analytics requires additional setup in your configuration'}
+                        hint={ui60 && 'Cookie-free, first party traffic analytics for your site'}
                         label='Web analytics'
                         labelClasses='py-4 w-full'
                         onChange={(e) => {
                             handleToggleChange('web_analytics', e);
                         }}
                     />
-                    <Separator className="border-grey-200 dark:border-grey-900" />
+                    {ui60 && !isWebAnalyticsEnabled ?
+                        <div className='mb-5 rounded-md border border-grey-200 bg-grey-50 px-4 py-2.5'>
+                            <span className='flex items-start gap-2'>
+                                <Icon className='-mt-px text-black' name='info-fill' size={24} />
+                                <span>
+                            Web analytics in Ghost is powered by <a className='text-green underline' href="https://tinybird.co" rel="noopener noreferrer" target='_blank'>Tinybird</a> and requires configuration to start collecting data. <a className='text-green underline' href="https://ghost.org/docs/" rel="noopener noreferrer" target='_blank'>Get started &rarr;</a>
+                                </span>
+                            </span>
+                        </div>
+                        :
+                        <Separator className="border-grey-200 dark:border-grey-900" />
+                    }
                 </>
             )}
             <Toggle
--- a/apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
+++ b/apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
@@ -1,16 +1,12 @@
-import React, {useEffect, useState} from 'react';
+import React from 'react';
 import TopLevelGroup from '../../TopLevelGroup';
 import useFeatureFlag from '../../../hooks/useFeatureFlag';
 import useSettingGroup from '../../../hooks/useSettingGroup';
 import {Button, Separator, SettingGroupContent, Toggle, withErrorBoundary} from '@tryghost/admin-x-design-system';
-import {HostLimitError, useLimiter} from '../../../hooks/useLimiter';
 import {getSettingValues, isSettingReadOnly} from '@tryghost/admin-x-framework/api/settings';
 import {usePostsExports} from '@tryghost/admin-x-framework/api/posts';
 
 const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
-    const [canHaveWebAnalytics, setCanHaveWebAnalytics] = useState(true);
-
-    const limiter = useLimiter();
     const {
         localSettings,
         isEditing,
@@ -21,23 +17,13 @@ const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
         handleEditingChange
     } = useSettingGroup();
 
-    const [trackEmailOpens, trackEmailClicks, trackMemberSources, outboundLinkTagging, webAnalytics] = getSettingValues(localSettings, [
-        'email_track_opens', 'email_track_clicks', 'members_track_sources', 'outbound_link_tagging', 'web_analytics'
+    const [trackEmailOpens, trackEmailClicks, trackMemberSources, outboundLinkTagging, webAnalyticsUserSetting, isWebAnalyticsEnabled] = getSettingValues(localSettings, [
+        'email_track_opens', 'email_track_clicks', 'members_track_sources', 'outbound_link_tagging', 'web_analytics', 'web_analytics_enabled'
     ]) as boolean[];
 
-    const hasTrafficAnalytics = useFeatureFlag('trafficAnalytics');
+    const taBetaFlagEnabled = useFeatureFlag('trafficAnalytics');
     const ui60 = useFeatureFlag('ui60');
     const isEmailTrackClicksReadOnly = isSettingReadOnly(localSettings, 'email_track_clicks');
-
-    useEffect(() => {
-        if (limiter?.isLimited('limitAnalytics')) {
-            limiter.errorIfWouldGoOverLimit('limitAnalytics').catch((error) => {
-                if (error instanceof HostLimitError) {
-                    setCanHaveWebAnalytics(false);
-                }
-            });
-        }
-    }, [limiter]);
 
     const handleToggleChange = (key: string, e: React.ChangeEvent<HTMLInputElement>) => {
         updateSetting(key, e.target.checked);
@@ -68,13 +54,14 @@ const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
 
     const inputs = (
         <SettingGroupContent className="analytics-settings !gap-y-0" columns={1}>
-            {hasTrafficAnalytics && canHaveWebAnalytics && (
+            {taBetaFlagEnabled && (
                 <>
                     <Toggle
-                        checked={webAnalytics}
+                        checked={isWebAnalyticsEnabled && webAnalyticsUserSetting}
                         direction='rtl'
+                        disabled={!isWebAnalyticsEnabled}
                         gap='gap-0'
-                        hint={ui60 && 'Cookie-free, first party traffic analytics for your site'}
+                        hint={isWebAnalyticsEnabled ? (ui60 ? 'Cookie-free, first party traffic analytics for your site' : undefined) : 'Web analytics requires additional setup in your configuration'}
                         label='Web analytics'
                         labelClasses='py-4 w-full'
                         onChange={(e) => {
--- a/apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
+++ b/apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import TopLevelGroup from '../../TopLevelGroup';
 import useFeatureFlag from '../../../hooks/useFeatureFlag';
 import useSettingGroup from '../../../hooks/useSettingGroup';
-import {Button, Icon, Separator, SettingGroupContent, Toggle, withErrorBoundary} from '@tryghost/admin-x-design-system';
+import {Button, Separator, SettingGroupContent, Toggle, withErrorBoundary} from '@tryghost/admin-x-design-system';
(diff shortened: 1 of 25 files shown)


apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
```diff
@@ -1,16 +1,12 @@
...
 import useSettingGroup from '../../../hooks/useSettingGroup';
 import {Button, Separator, SettingGroupContent, Toggle, withErrorBoundary} from '@tryghost/admin-x-design-system';
-import {HostLimitError, useLimiter} from '../../../hooks/useLimiter';
 import {getSettingValues, isSettingReadOnly} from '@tryghost/admin-x-framework/api/settings';
 import {usePostsExports} from '@tryghost/admin-x-framework/api/posts';
 
 const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
-    const [canHaveWebAnalytics, setCanHaveWebAnalytics] = useState(true);
-
-    const limiter = useLimiter();
```
9larsons, line 13: Limiter isn't needed because it's baked into `SettingsHelper` in the backend. We will, however, need this to show the CTA in the `posts`/`stats` apps that will be a follow-up to this.

apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
```diff
@@ -21,24 +17,14 @@ const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
         handleEditingChange
     } = useSettingGroup();
 
-    const [trackEmailOpens, trackEmailClicks, trackMemberSources, outboundLinkTagging, webAnalytics] = getSettingValues(localSettings, [
-        'email_track_opens', 'email_track_clicks', 'members_track_sources', 'outbound_link_tagging', 'web_analytics'
+    const [trackEmailOpens, trackEmailClicks, trackMemberSources, outboundLinkTagging, webAnalyticsUserSetting, isWebAnalyticsEnabled] = getSettingValues(localSettings, [
```
9larsons: We've settled on `isSettingEnabled` as a naming convention for the calculated setting. Because it's really confusing here that we have `webAnalytics` and `isWebAnalyticsEnabled`, I've called it `webAnalyticsUserSetting`. Open to other ideas here, or I can keep it the same for consistency.

apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
```diff
@@ -21,24 +17,14 @@ const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
...
+        'email_track_opens', 'email_track_clicks', 'members_track_sources', 'outbound_link_tagging', 'web_analytics', 'web_analytics_enabled'
     ]) as boolean[];
 
-    const hasTrafficAnalytics = useFeatureFlag('trafficAnalytics');
+    const taBetaFlagEnabled = useFeatureFlag('trafficAnalytics');
```
9larsons, line 24: We still show/hide the setting based on the beta flag. I renamed this to be a bit clearer, so we don't have three kinds of 'analytics' flags kicking about.

apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
```diff
@@ -68,13 +54,14 @@ const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
...
+            {taBetaFlagEnabled && (
                 <>
                     <Toggle
-                        checked={webAnalytics}
+                        checked={isWebAnalyticsEnabled && webAnalyticsUserSetting}
                         direction='rtl'
+                        disabled={!isWebAnalyticsEnabled}
                         gap='gap-0'
-                        hint={ui60 && 'Cookie-free, first party traffic analytics for your site'}
+                        hint={isWebAnalyticsEnabled ? (ui60 ? 'Cookie-free, first party traffic analytics for your site' : undefined) : 'Web analytics requires additional setup in your configuration'}
```
9larsons: Zimo needs to update the copy here yet.

ghost/admin/app/components/admin-x/posts.js
```diff
@@ -23,7 +23,7 @@ export default class Posts extends AdminXComponent {
                     emailTrackClicks: this.settings.emailTrackClicks,
                     membersTrackSources: this.settings.membersTrackSources,
                     outboundLinkTagging: this.settings.outboundLinkTagging,
-                    webAnalytics: this.feature.trafficAnalytics && (this.settings.webAnalytics ?? false)
+                    webAnalytics: this.settings.webAnalyticsEnabled // use the computed setting
```
9larsons, line 26: @cmraible This might be the biggest question... we already have `webAnalytics` used across the app, and in order to not add another 15 or so files to this PR, I didn't update the naming convention. From the app perspective, this is the flag we care about, again it's more a question of consistency in using `Enabled` for the computed setting.

As mentioned above, we'll use the limit service to know whether or not we should show CTA, even though that's baked in to this computed setting.

ghost/admin/app/models/setting.js
```diff
@@ -53,14 +53,16 @@ export default Model.extend(ValidationEngine, {
...
     announcementVisibility: attr('json-string'),
+
     /**
      * Analytics settings
      */
     emailTrackOpens: attr('boolean'),
     emailTrackClicks: attr('boolean'),
     outboundLinkTagging: attr('boolean'),
     membersTrackSources: attr('boolean'),
-    webAnalytics: attr('boolean'),
```
9larsons, line 63: I didn't think it was relevant to expose the non-computed setting to the app, as that's the purpose of the computed setting. To me it makes more sense to use the limit service in-line when necessary.

ghost/core/core/server/services/settings-helpers/SettingsHelpers.js
```diff
@@ -265,8 +265,68 @@ class SettingsHelpers {
...
+        
+        // First requirement: tinybird:tracker:endpoint is always required
+        if (!tinybirdConfig || !tinybirdConfig.tracker?.endpoint) {
+            debug('Web analytics is not available without tinybird:tracker:endpoint');
+            return false;
+        }
+
+        // Second requirement: Either JWT config OR local stats config
+        const hasJwtConfig = !!(tinybirdConfig.workspaceId && tinybirdConfig.adminToken);
+        const hasLocalConfig = !!(tinybirdConfig.stats?.local?.enabled);
```
9larsons, line 334: I didn't want to add extra confusion here about the local config that's required. I figured we can update the `TinybirdService` and readme to document that. To keep this simple, all I'm checking for is that the user is intending to use local for the stats portion.

This does mean we need the tracking endpoint defined even if we're not using that part of the functionality which I'd be open to changing this logic to say if you're using local, we don't check for anything else in this block.

apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
cmraible: I like `webAnalyticsUserSetting` in this setting, I think it makes it more clear which is which 👍

cmraible: `trackWebAnalytics` might also make sense in this context, to align with the other user configurable settings. No strong feelings either way, just thinking out loud.

apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
cmraible: Nit: personally I'd prefer a more verbose, explicit name, i.e. `trafficAnalyticsLabsFlagIsEnabled`, especially since it's not going to be a beta flag in 6.0

apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
```diff
@@ -68,13 +54,14 @@ const Analytics: React.FC<{ keywords: string[] }> = ({keywords}) => {
...
     const inputs = (
         <SettingGroupContent className="analytics-settings !gap-y-0" columns={1}>
-            {hasTrafficAnalytics && canHaveWebAnalytics && (
+            {taBetaFlagEnabled && (
                 <>
                     <Toggle
-                        checked={webAnalytics}
+                        checked={isWebAnalyticsEnabled && webAnalyticsUserSetting}
                         direction='rtl'
+                        disabled={!isWebAnalyticsEnabled}
```
cmraible: There's a bug here caused by circular logic — as soon as I disable the toggle in the UI, `isWebAnalyticsEnabled` also becomes false, which disables the toggle and makes it impossible to re-enable. 

What's the intent of disabling the toggle? I'd think that if it's rendered at all, it should always be enabled — but I might be overlooking something.

apps/admin-x-settings/test/acceptance/membership/analytics.test.ts
```diff
@@ -20,25 +44,34 @@ test.describe('Analytics settings', async () => {
 
         await expect(section).toBeVisible();
 
+        await expect(section.getByLabel('Web analytics')).toBeChecked();
         await expect(section.getByLabel('Newsletter opens')).toBeChecked();
         await expect(section.getByLabel('Newsletter clicks')).toBeChecked();
         await expect(section.getByLabel('Member sources')).toBeChecked();
         await expect(section.getByLabel('Outbound link tagging')).toBeChecked();
 
+        await section.getByLabel(/Web analytics/).uncheck();
```
cmraible, line 54: I think we're missing a test for the inverse case, where the toggle starts unchecked() and we toggle it on.

ghost/admin/app/routes/posts.js
```diff
@@ -180,7 +180,7 @@ export default class PostsRoute extends AuthenticatedRoute {
         const promises = [];
         
         // Fetch visitor counts if web analytics is enabled
-        if (this.settings.webAnalytics && this.feature.trafficAnalytics) {
+        if (this.settings.webAnalyticsEnabled) {
```
cmraible, line 183: 😍 so much cleaner

cmraible (changes requested):
There's a bug [here](https://github.com/TryGhost/Ghost/pull/24311/files#r2196142933) that makes it impossible to re-enable the "Web Analytics" setting in the UI after toggling it off.

We also need to change `ghost_head` so it doesn't render the `ghost-stats.js` script if the calculated setting evaluates to false — right now if you disable the setting, it hides the UI but it doesn't stop tracking events.

ghost/admin/app/components/admin-x/posts.js
cmraible: Yeah I think this is fine — from the perspective of the posts app, it just needs a single boolean so IMO it's fine to drop the `Enabled` in this context. If we find it's confusing, we can always rename it in a follow up PR.

apps/admin-x-settings/src/components/settings/membership/Analytics.tsx
9larsons: My bad! I tested this before writing tests and changing logic slightly and didn't notice that regression.

apps/admin-x-settings/test/acceptance/membership/analytics.test.ts
9larsons: Added, as well as some test updates covering ui60.

9larsons:
Good catch! Updated.

peterzimon (approved):
Looking good from a design POV

ghost/core/core/server/services/tinybird/TinybirdService.js
```diff
@@ -44,6 +44,33 @@ const jwt = require('jsonwebtoken');
...
+    
+    if (!hasJwtConfig && !hasLocalConfig) {
+        return {
+            isValid: false,
+            reason: 'Requires either (workspaceId + adminToken) or stats.local.enabled'
+        };
+    }
+
+    return {isValid: true};
+}
```
cmraible: So nice to have this all in one place finally ❤️

ghost/core/core/frontend/helpers/ghost_head.js
```diff
@@ -362,9 +362,8 @@ module.exports = async function ghost_head(options) { // eslint-disable-line cam
             if (!_.isEmpty(tagCodeInjection)) {
                 head.push(tagCodeInjection);
             }
-            const isTbTrackingEnabled = labs.isSet('trafficAnalytics');
-            const hasTbConfig = config.get('tinybird') && config.get('tinybird:tracker');
-            if (isTbTrackingEnabled && hasTbConfig) {
+            // Use settingsHelpers to check if web analytics is enabled (includes all necessary checks)
+            if (settingsHelpers.isWebAnalyticsEnabled()) {
```
cmraible, line 366: 👌

ghost/core/core/server/services/settings/settings-service.js
```diff
@@ -109,6 +109,10 @@ module.exports = {
         // Social web (ActivityPub)
         fields.push(new CalculatedField({key: 'social_web_enabled', type: 'boolean', group: 'social_web', fn: settingsHelpers.isSocialWebEnabled.bind(settingsHelpers), dependents: ['social_web', 'labs']}));
 
+        // Web analytics
+        fields.push(new CalculatedField({key: 'web_analytics_enabled', type: 'boolean', group: 'analytics', fn: settingsHelpers.isWebAnalyticsEnabled.bind(settingsHelpers), dependents: ['web_analytics', 'labs']}));
+        fields.push(new CalculatedField({key: 'web_analytics_configured', type: 'boolean', group: 'analytics', fn: settingsHelpers.isWebAnalyticsConfigured.bind(settingsHelpers), dependents: ['web_analytics']}));
```
cmraible, line 114: Nice, I like this idea a lot. Even though it's technically "config" and not a "setting", I like how this keeps all the logic on the backend, to give the frontend a simple boolean 🙌

ghost/core/core/server/services/tinybird/TinybirdService.js
```diff
@@ -166,6 +193,15 @@ class TinybirdService {
...
     }
+
+    /**
+     * Validates tinybird configuration
+     * @param {TinybirdConfig} tinybirdConfig - The tinybird configuration object
+     * @returns {{isValid: boolean, reason?: string}} Validation result with optional reason for failure
+     */
+    static validateConfig(tinybirdConfig) {
+        return validateTinybirdConfig(tinybirdConfig);
+    }
```
cmraible: This strikes me as redundant, and I don't think this is used anywhere? I wouldn't be against renaming `validateTinybirdConfig` to just `validateConfig` since it should be clear from the context (TInybirdService), but I think we should pick one.

cmraible (approved):
Nice work! Looks good to me. Left a few comments but nothing major.

ghost/core/core/server/services/tinybird/TinybirdService.js
9larsons: Oh yeah, this was missed in my cleanup. Originally I was thinking of validating config here, but then we don't want a dependency between services. I'll remove it.

ghost/core/core/server/services/settings/settings-service.js
9larsons: The annoying part is we still need separate checks with the CTA display (plan limits) and config, but this does clean it up a bit, and importantly has a single flag for 'is this thing on?'.

PR #24317, opened 2025-07-10
Introduce post and admin analytics pages with supporting locators

ibalosh:
* added post analytics admin page per discussion yesterday with Steve
* added more admin analytics pages, to support later development and get initial sense how do we want to structure this
* added locators to support these page objects
* the updates were made to support the e2e test cases we are going to add for analytics 

note @9larsons , I added couple of locators and sections, let me know if you need more and if this is what you had in mind. Some of pages are copy of ones in stats. At some point I think we should remove them from there, and use only these as single source of truth. That way we could use them in all micro projects if needed, and that makes sense later on.

Some conventions regarding structure, naming, sections, etc will change along the way. I will refactor as we add more things, since I don't have yet the clear picture.

Let me know what you think?


cmraible:
Sorry, didn't have a chance to look at this today but will definitely give this a 👀 on Monday

cmraible:
@ibalosh All these changes LGTM! I've run the e2e tests locally and they're passing for me.

Edit: I've fixed the CI issue on `main` and rebased — they should pass now.

cmraible (approved):
LGTM!

cmraible:
Page objects all look like a great start for the tests we want to write. 

For the analytics overview page test -- is that actually reaching the analytics overview page? I don't see where we've handled admin auth yet, so I think it is likely reaching the login screen, and passing because it's just checking for the body tag. I don't think this test is the main point of this PR so not required IMO, just wanted to flag that.

ibalosh:
> Page objects all look like a great start for the tests we want to write.
> 
> For the analytics overview page test -- is that actually reaching the analytics overview page? I don't see where we've handled admin auth yet, so I think it is likely reaching the login screen, and passing because it's just checking for the body tag. I don't think this test is the main point of this PR so not required IMO, just wanted to flag that.

@cmraible you are absolutely right, that test needs a login and it passes only because it's a body check, added it just a "placeholder" for future tests, so not really a valid one. I'll remove that one for now.

PR #24321, opened 2025-07-10
 Added server-side limit validations for custom theme installations

aileen:
ref https://linear.app/ghost/issue/BAE-330/ghost-implement-server-side-limit-validations

The new `customThemes` limit will include a subset of official themes in its allowlist. This required updating the theme installation logic to properly validate both TryGhost and third-party repositories against the configured limits. Enhanced the limit validation in `installer.js` to check ALL theme installations (including TryGhost repos) when the `customThemes` allowlist contains only one theme. The logic uses `limitService.errorIfWouldGoOverLimit()` to enforce restrictions based on the allowlist configuration.

The `mockLimitService` in our test suite had critical issues that prevented proper testing. It was using `sinon.stub(limitService)`, which didn't properly stub the methods, causing tests to fail as soon as the `mockLimitService` was imported. Additionally, the global `restore()` function called `sinon.restore() `, which destroyed ALL stubs across the test suite, breaking unrelated tests that created their own stubs.

Fixed this by making `mockLimitService` stub individual methods instead of the entire object. Created a new `restoreLimitService()` function that only cleans up limit-related stubs and properly removes mocked limits from `limitService.limits`, preventing test contamination without affecting other tests' stubs.

### DIFF
--- a/ghost/core/core/server/services/themes/installer.js
+++ b/ghost/core/core/server/services/themes/installer.js
@@ -19,9 +19,18 @@ const messages = {
 const installFromGithub = async (ref) => {
     const [org, repo] = ref.toLowerCase().split('/');
 
-    //TODO: move the organization check to config
-    if (limitService.isLimited('customThemes') && org.toLowerCase() !== 'tryghost') {
-        await limitService.errorIfWouldGoOverLimit('customThemes', {value: repo.toLowerCase()});
+    if (limitService.isLimited('customThemes')) {
+        // The custom theme limit might consist of only one single theme, so we can't rely on
+        // the org alone to determine if the request is allowed or not.
+        const noOtherThemesAllowed = limitService.limits.customThemes?.allowlist?.length === 1;
+        //TODO: move the organization check to config
+        const isNotOfficialThemeRequest = org.toLowerCase() !== 'tryghost';
+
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/services/themes/installer.js
```diff
@@ -62,6 +71,11 @@ const installFromGithub = async (ref) => {
             }));
         }
 
+        if (e instanceof errors.HostLimitError) {
```
sam-lord, line 74: Is this just being more explicit about what happens with host errors?

Returning a rejected promise and rethrowing the error are the same in an async context, so I don't believe this block does anything functionally different

sam-lord (approved):
Just a quick comment to check my understanding, but overall looks good 👍

ghost/core/core/server/services/themes/installer.js
aileen: > Is this just being more explicit about what happens with host errors?

Yes, exactly. When I tested it out, I'd get a 500 error otherwise, and therefore added this extra check.

aileen: Maybe I'll test it again to verify, before I merge this

PR #24322, opened 2025-07-10
Fixed white list text in CTA cards with dark newsletter background

kevinansfield:
ref https://linear.app/ghost/issue/PROD-2265/

- giving specific colors to lists to fix mismatched bullet colors in Outlook meant that lists in CTA cards were no longer using the correct color
- added explicit overrides for fixed-color CTA backgrounds to ensure lists and list items inherit the color specific for the card rather than color used for general text
- added `ctaBgColors` array to our template data properties so we can loop over them to keep the template easier to manage

### DIFF
--- a/ghost/core/core/server/services/email-service/email-templates/partials/styles.hbs
+++ b/ghost/core/core/server/services/email-service/email-templates/partials/styles.hbs
@@ -381,6 +381,17 @@ h6 + .kg-paywall .kg-paywall-hr td {
 }
 
 /* Exclude CTA cards with colored backgrounds from custom text color, but allow transparent ones */
+{{#hasFeature "emailCustomization"}}
+{{#each ctaBgColors}}
+.post-content-row .kg-cta-bg-{{this}} .kg-cta-text p,
+.post-content-row .kg-cta-bg-{{this}} .kg-cta-text ul,
+.post-content-row .kg-cta-bg-{{this}} .kg-cta-text ul li,
+.post-content-row .kg-cta-bg-{{this}} .kg-cta-text ol,
+.post-content-row .kg-cta-bg-{{this}} .kg-cta-text ol li {
+    color: inherit !important;
+}
+{{/each}}
+{{else}}
 .post-content-row .kg-cta-bg-grey .kg-cta-text p,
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/services/email-service/email-templates/partials/styles.hbs
```diff
@@ -381,6 +381,17 @@ h6 + .kg-paywall .kg-paywall-hr td {
...
 /* Exclude CTA cards with colored backgrounds from custom text color, but allow transparent ones */
+{{#hasFeature "emailCustomization"}}
+{{#each ctaBgColors}}
+.post-content-row .kg-cta-bg-{{this}} .kg-cta-text p,
+.post-content-row .kg-cta-bg-{{this}} .kg-cta-text ul,
+.post-content-row .kg-cta-bg-{{this}} .kg-cta-text ul li,
+.post-content-row .kg-cta-bg-{{this}} .kg-cta-text ol,
+.post-content-row .kg-cta-bg-{{this}} .kg-cta-text ol li {
+    color: inherit !important;
+}
```
kevinansfield, line 392: Ideally we should be relying on the cascade and inheritance to avoid the need for specific overrides like this but we have too many specific styles to make that feasible without a big rework of the stylesheet.

I've kept the change as minimal as possible so we can fix this bug then we can tackle a larger refactor later if there's appetite.

troyciesco: i know this hasFeature block will go away soon plus you want to limit changes so i won't call it out as a blocker, but L395-401 could use ctaBgColors too

just to show i'm paying attention 😉

troyciesco (approved):

ghost/core/core/server/services/email-service/email-templates/partials/styles.hbs
kevinansfield: With the flags we tend to keep the existing code completely untouched (even if it's non-optimal) so there's no chance of accidentally including production-altering changes.

Once we've verified the fix is good lines 395-401 will disappear completely 😄

PR #24326, opened 2025-07-10
🐛 Fixed white lists in CTA cards when newsletter has dark background

kevinansfield:
ref https://linear.app/ghost/issue/PROD-2265/
ref https://github.com/TryGhost/Ghost/pull/24322

- removed feature flag around the style fixes
  - giving specific colors to lists to fix mismatched bullet colors in Outlook meant that lists in CTA cards were no longer using the correct color
  - added explicit overrides for fixed-color CTA backgrounds to ensure lists and list items inherit the color specific for the card rather than color used for general text
  - added `ctaBgColors` array to our template data properties so we can loop over them to keep the template easier to manage


cathysarisky:
Thanks for fixing my bug! :)

ErisDS (approved):

PR #24327, opened 2025-07-10
Cleaned up tinybird config, added validation

9larsons:
ref https://linear.app/ghost/issue/PROD-2207/
- removed `config.stats` checks from Ember frontend
- simplified config & validation; removed `local` option, removed `token` (only using JWT)
- changed `tinybird.stats.id` to `tinybird.stats.overrideSiteUuid` and renamed `statsConfig.id` to `statsConfig.siteUuid` for more clarity (id is a bit ambiguous)

The goal here is to provide a more streamlined and intuitive behavior and configuration. The frontend shouldn't have checks for config on whether or not to display things; we've added frontend checks for config and disable the pieces specific to config if it isn't present, and display a help message (https://github.com/TryGhost/Ghost/commit/4e37c0f64d607ab075910dda789d3b382d00f3d0).

Additionally, we've made a lot of changes over time to where we store tokens, site uuids, and general use of config that has not been cleaned up. I've added validation to the Tinybird service so that we have an actual validator with specific error messages (as long as you read the backend logs). We no longer need the `local` block as we can use comments with `jsonc` for our local config, and there's no good reason to avoid using the JWT path for validation, which dramatically streamlines setup.

We could go further by setting defaults to the endpoints and making those override-only. This already got quite large and covers the majority of QoL changes.


ghost/admin/app/routes/posts/analytics.js
```diff
@@ -57,7 +57,7 @@ export default class Analytics extends AuthenticatedRoute {
             }
 
             // This ensures that we don't load this page if the stats config is not set
-            if (!(this.config.stats && this.feature.trafficAnalytics)) {
+            if (!this.feature.trafficAnalytics) {
```
cmraible:
```suggestion
            if (!(this.feature.trafficAnalytics || this.feature.ui60)) {
```

Just future proofing for 6.0 — I think what you have is correct for now, but in its current implementation this would redirect us if trafficAnalytics is disabled, but ui60 is enabled — in which case we'd want to still display this, but just hide the traffic analytics parts.

ghost/admin/app/routes/posts-x.js
```diff
@@ -26,7 +26,7 @@ export default class PostsXRoute extends AuthenticatedRoute {
         }
 
         // This ensures that we don't load this page if the stats config is not set
-        if (!(this.config.stats && this.feature.trafficAnalytics)) {
+        if (!this.feature.trafficAnalytics) {
```
cmraible:
```suggestion
        if (!(this.feature.trafficAnalytics || this.feature.ui60)) {
```
Same here

ghost/admin/app/routes/stats-x.js
```diff
@@ -15,7 +15,7 @@ export default class StatsXRoute extends AuthenticatedRoute {
         }
 
         // This ensures that we don't load this page if the stats config is not set
-        if (!(this.config.stats && this.feature.trafficAnalytics)) {
+        if (!this.feature.trafficAnalytics) {
```
cmraible:
```suggestion
        if (!(this.feature.trafficAnalytics || this.feature.ui60)) {
```

And here

ghost/admin/app/routes/stats.js
```diff
@@ -25,7 +25,7 @@ export default class StatsRoute extends AuthenticatedRoute {
         }
 
         // This ensures that we don't load this page if the stats config is not set
-        if (!(this.config.stats && this.feature.trafficAnalytics)) {
+        if (!this.feature.trafficAnalytics) {
```
cmraible:
```suggestion
        if (!(this.feature.trafficAnalytics || this.feature.ui60)) {
```
And here

ghost/core/core/frontend/helpers/ghost_head.js
```diff
@@ -158,15 +158,12 @@ function getTinybirdTrackerScript(dataRoot) {
...
-    const env = config.get('env');
-
     const statsConfig = config.get('tinybird:tracker');
     const localConfig = config.get('tinybird:tracker:local');
     const localEnabled = localConfig?.enabled ?? false;
 
     const endpoint = localEnabled ? localConfig.endpoint : statsConfig.endpoint;
-    const token = localEnabled ? localConfig.token : statsConfig.token;
-    const datasource = localEnabled ? localConfig.datasource : statsConfig.datasource;
+    const datasource = 'analytics_events';
```
cmraible: Do we need the localConfig at all anymore? Doesn't need to be in this PR if you'd rather keep it separate, but I was hoping we could get rid of this complexity as well. I don't personally use it anymore since we have the `jsonc` config, but if others are then that's fine.

ghost/core/test/unit/server/services/tinybird/TinybirdService.test.js
```diff
@@ -161,16 +176,259 @@ describe('TinybirdService', function () {
...
 
-        it('should return the stats token if jwt is not enabled and local is not enabled', function () {
-            tinybirdConfig.workspaceId = null;
-            tinybirdConfig.adminToken = null;
-            tinybirdConfig.stats = {
-                token: 'stats-token'
+        // Removed test for deprecated stats.token configuration
+        // The validator now requires either valid remote config (workspaceId + adminToken + tracker.endpoint)
+        // or valid local config (stats.local.enabled or tracker.local.enabled with token and endpoint)
+
```
cmraible: I'd either delete this comment or move it to the PR description — seems like something the LLM is trying to communicate as a comment, which makes sense in the scope of this PR but doesn't really make sense in the codebase after its merged. I'm constantly having to remove these myself 😂

cmraible (commented):
Curious how you'd feel about removing the local config concept entirely? 

I don't think it's strictly necessary for this PR, but it would simplify a lot of this logic, and also simplify setting up your local development environment.

I don't personally use the local config block, esp. now that we have `.jsonc` to make it easy to toggle between different configs, it feels like a bit of an antipattern to me.

ghost/admin/app/routes/posts/analytics.js
9larsons: The ui60 flag is a bit unclear to me - I thought it was being used more specifically for re-orienting the nav (as it was originally called `updatedMainNav` and Settings. I'm fine adding these, but both are going away with 6.0, so I don't think it ultimately matters much.

ghost/core/test/unit/server/services/tinybird/TinybirdService.test.js
9larsons: Lol yeah, fair.

ghost/core/core/frontend/helpers/ghost_head.js
9larsons: Fair point, we can scrap it for the tracker.

ghost/admin/app/routes/posts/analytics.js
cmraible: Yeah fair enough — I guess I'm not any clearer on it, I just thought it was for all the things that are changing in 6.0 — you might be right that it was intended just for the nav

ghost/admin/app/components/posts/analytics.hbs
```diff
@@ -73,7 +73,7 @@
                             <li>
                                 <a class="view-browser" href="{{this.post.url}}" target="_blank" rel="noopener noreferrer">View in browser</a>
                             </li>
-                            {{#if (and (gh-user-can-admin this.session.user) this.config.stats (feature "trafficAnalytics"))}}
+                            {{#if (and (gh-user-can-admin this.session.user) (feature "trafficAnalytics"))}}
```
cmraible, line 76: If I load the analytics page without anything in my `tinybird` configuration, the stats app fails to load due to an empty/missing token:

Image

cmraible (changes requested):
Sorry I missed this on my first pass — I just found that the whole stats app crashes if you try to load it without Tinybird configured — a problem which isn't caused by this PR (the stats app shouldn't assume Tinybird is configured), but a problem that this PR reveals.

There's so many different combinations of config and flags and all involved here — given that this PR spans both the frontend apps some minor but potentially risky backend changes, I wonder if we should hold off on this until we have some e2e smoke tests in place. 

WDYT?

cmraible (changes requested):
Already talked about this in a call — opting to break this up into smaller chunks, and deprioritize the config changes for the time being.

PR #24350, opened 2025-07-14
Wired up Network (ActivityPub) setting in Admin

sagzy:

The Network toggle is disabled if:
1. (Ghost (Pro) only) the feature is disabled by config
2. The site is hosted on a subdirectory, localhost or an IP address in production

### DIFF
--- a/apps/admin-x-settings/src/components/settings/growth/Network.tsx
+++ b/apps/admin-x-settings/src/components/settings/growth/Network.tsx
@@ -1,22 +1,69 @@
-import React from 'react';
+import React, {useEffect, useState} from 'react';
 import TopLevelGroup from '../../TopLevelGroup';
+import validator from 'validator';
 import {Icon, SettingGroupContent, Toggle, withErrorBoundary} from '@tryghost/admin-x-design-system';
+import {Setting, getSettingValues, useEditSettings} from '@tryghost/admin-x-framework/api/settings';
+import {getGhostPaths} from '@tryghost/admin-x-framework/helpers';
+import {useGlobalData} from '../../providers/GlobalDataProvider';
+import {useHandleError} from '@tryghost/admin-x-framework/hooks';
+import {useLimiter} from '../../../hooks/useLimiter';
 
 const Network: React.FC<{ keywords: string[] }> = ({keywords}) => {
-    const enableToggle = (
-        <>
-            <Toggle
-                checked={true}
-                direction='rtl'
-                onChange={() => {}}
-            />
-        </>
-    );
+    const {settings} = useGlobalData();
+    const {mutateAsync: editSettings} = useEditSettings();
+    const handleError = useHandleError();
+    const limiter = useLimiter();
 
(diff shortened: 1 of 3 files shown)


apps/admin-x-settings/src/components/settings/growth/Network.tsx
```diff
@@ -1,22 +1,59 @@
...
+    const isHostedOnIP = isProduction && validator.isIP(window.location.hostname);
+
+    const isDisabled = isDisabledByConfig || isHostedOnSubdirectory || isHostedOnLocalhost || isHostedOnIP;
 
-    const domainIssue = true;
+    // The Network toggle is displayed as checked if:
+    // 1. The setting value is true, and
+    // 2. The toggle is not disabled
+    const [socialWebSetting] = getSettingValues<string>(settings, ['social_web']);
+    const isChecked = !!socialWebSetting && !isDisabled;
```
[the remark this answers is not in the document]
sagzy: This is a design concern - we don't want to display a checked toggle + a disabled state. It's either:
- disabled and toggled off
- enabled and can be toggled on or off

apps/admin-x-settings/src/components/settings/growth/Network.tsx
```diff
@@ -1,22 +1,69 @@
...
+    const handleError = useHandleError();
+    const limiter = useLimiter();
+
+    // The Network toggle is disabled in Admin settings if:
+    // 1. (Ghost (Pro) only) the feature is disabled by config
+    // 2. The site is hosted on a subdirectory, localhost or an IP address in production
+    const [isDisabledByConfig, setIsDisabledByConfig] = useState(false);
+    useEffect(() => {
+        async function checkLimiter() {
+            if (limiter?.isLimited('limitSocialWeb') && await limiter.checkWouldGoOverLimit('limitSocialWeb')) {
```
ErisDS: Curious why we chose to use `checkWouldGoOverLimit()` here instead of `isDisabled()`?

ErisDS (commented):
Left one question, but otherwise LGTM!

ErisDS (approved):

peterzimon (approved):

apps/admin-x-settings/src/components/settings/growth/Network.tsx
sagzy: Yep great question! 

I originally started with `isDisabled()` but noticed a race condition where the toggle would be displayed as toggled on when it shouldn't. The reason: `isDisabled()` returns `undefined` while the limiter is loading, then later returns the actual value of the flag and the toggle wouldn't re-render to reflect the change. 

I've now fixed that by adding a `key` prop to the Toggle component (which forces React to unmount/remount the component when `isChecked` value changes) and restored the `isDisabled()` code in [e92478c](https://github.com/TryGhost/Ghost/commit/e92478c14734e8a7f9b345f630174fee09ddd0f0) - which is much simpler!

PR #24353, opened 2025-07-14
🎨  Updated Facebook validation with more specific regex

troyciesco:
closes https://linear.app/ghost/issue/ENG-2454/update-facebook-validation-with-more-specific-regex

- improves the regex for usernames based on fb docs (https://www.facebook.com/help/105399436216001/)
- makes sure different formats work, like regular usernames, groups, and pages
- adds a bit of a safety net so that currently-invalid usernames don't crash the page - users should have a chance to update
- make sure it works for both staff facebook links and the site facebook link

### DIFF
--- a/apps/admin-x-settings/test/acceptance/general/socialAccounts.test.ts
+++ b/apps/admin-x-settings/test/acceptance/general/socialAccounts.test.ts
@@ -1,3 +1,4 @@
+import {FB_ERRORS} from '../../../src/utils/socialUrls/facebook';
 import {expect, test} from '@playwright/test';
 import {globalDataRequests} from '../../utils/acceptance';
 import {mockApi, testUrlValidation, updatedSettingsResponse} from '@tryghost/admin-x-framework/test/acceptance';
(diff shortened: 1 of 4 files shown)


apps/admin-x-settings/test/acceptance/general/socialAccounts.test.ts
```diff
@@ -64,26 +65,22 @@ test.describe('Social account settings', async () => {
...
-            'https://www.facebook.com/page/*(&*(%%))'
+            'page/fb123',
+            '',
+            FB_ERRORS.INVALID_USERNAME
         );
 
         await testUrlValidation(
             facebookInput,
             'facebook.com/pages/some-facebook-page/857469375913?ref=ts',
-            'https://www.facebook.com/pages/some-facebook-page/857469375913?ref=ts'
```
troyciesco, line 86: i made this invalid because we error on query params for others, like linkedin. idk if it's useful to allow though? i have no idea if any of these social platforms have tracking analytics based on params like this

kevinansfield: From what I've been able to find, Facebook strips any query params and doesn't provide any insight into where traffic arriving at your Facebook page comes from so I think this is the right call 👍 

The only way you can track traffic to Facebook pages is when using a link shortener/redirect type service but then we'd reject those addresses anyway because they wouldn't match a Facebook URL.

troyciesco:
Closing this for now. iirc as to why this went stale, there's a lot to consider about what to do with pre-existing invalid links. For example, if you tried to change something on your profile you'd get a save error - but the social link is on a different tab and might not be related to what you were doing, and it's not clear as a user what you'd need to fix. So we'd probably need some sort of modal popup and have logic to figure out if an _unchanged_ invalid fb link was the reason the save failed.

The validation for a fb link in general is also rare enough of an issue that deciding how to handle it is better suited if we ever clean up social links overall, like in https://linear.app/ghost/issue/BER-2516 (perhaps some time after ember to react is complete)

PR #24354, opened 2025-07-14
Fixed Admin stats-x redirect conditionals

kevinansfield:
no issue

- in `home` we need to use the same conditional for the stats-x redirect as we use to redirect away from `stats-x` otherwise we can end up in an infinite loop
- in `setup.done` there were two different redirects to `stats-x`, removing the old `ui60` flag conditional means we don't have unexpected redirects to `stats-x` when we don't know if it's actually accessible


9larsons (approved):
Also passing for me locally.

PR #24355, opened 2025-07-14
Wired up Explore settings

sagzy:
ref https://linear.app/ghost/issue/PROD-1771

- Publishers can join the Ghost Explore directory by enabling the 'explore_ping' setting
- Additionally, they can choose to share their member count and other growth data by enabling 'explore_ping_growth' setting

### DIFF
--- a/apps/admin-x-framework/src/api/settings.ts
+++ b/apps/admin-x-framework/src/api/settings.ts
@@ -31,7 +31,7 @@ export const useBrowseSettings = createQuery<SettingsResponseType>({
     dataType,
     path: '/settings/',
     defaultSearchParams: {
-        group: 'site,theme,private,members,portal,newsletter,email,amp,labs,slack,unsplash,views,firstpromoter,editor,comments,analytics,announcement,pintura,donations,security,social_web'
+        group: 'site,theme,private,members,portal,newsletter,email,amp,labs,slack,unsplash,views,firstpromoter,editor,comments,analytics,announcement,pintura,donations,security,social_web,explore'
     }
 });
 
--- a/apps/admin-x-settings/test/acceptance/growth/explore.test.ts
+++ b/apps/admin-x-settings/test/acceptance/growth/explore.test.ts
@@ -1,0 +1,96 @@
+import {expect, test} from '@playwright/test';
+import {globalDataRequests} from '../../utils/acceptance';
+import {mockApi, responseFixtures, toggleLabsFlag, updatedSettingsResponse} from '@tryghost/admin-x-framework/test/acceptance';
+
+test.describe('Ghost Explore', () => {
+    test.beforeEach(async () => {
+        toggleLabsFlag('ui60', true);
+    });
+
+    test('can join Ghost Explore', async ({page}) => {
+        const {lastApiRequests} = await mockApi({page, requests: {
+            ...globalDataRequests,
+            browseSettings: {
+                ...globalDataRequests.browseSettings,
+                response: {
+                    ...responseFixtures.settings,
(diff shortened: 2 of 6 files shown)


ghost/core/core/shared/settings-cache/CacheManager.js
```diff
@@ -64,6 +64,8 @@ const _ = require('lodash');
  * @property {boolean|null} social_web_enabled - Whether social web is enabled
  * @property {boolean|null} web_analytics_enabled - Whether web analytics is enabled
  * @property {boolean|null} web_analytics_configured - Whether web analytics is configured
+ * @property {boolean|null} explore_ping - Whether Explore is enabled
+ * @property {boolean|null} explore_ping_growth - Whether growth data is shared to Explore
```
ErisDS: This change is incorrect, the settings are not exposed as part of the public settings. This is true for the previous additions of social_web and web_analytics too 🙈

apps/admin-x-settings/test/acceptance/growth/explore.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        await expect(growthDataToggle).toBeVisible();
+        await expect(growthDataToggle).not.toBeChecked();
+
+        // Turn off growth data sharing
+        await growthDataToggle.click();
+
+        // Verify the API call to disable explore_ping_growth
+        expect(lastApiRequests.editSettings?.body).toEqual({
+            settings: [{key: 'explore_ping_growth', value: true}]
+        });
```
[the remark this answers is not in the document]
ErisDS: @sagzy I think the rabbit is right 🥕

apps/admin-x-framework/src/api/settings.ts
```diff
@@ -31,7 +31,7 @@ export const useBrowseSettings = createQuery<SettingsResponseType>({
     dataType,
     path: '/settings/',
     defaultSearchParams: {
-        group: 'site,theme,private,members,portal,newsletter,email,amp,labs,slack,unsplash,views,firstpromoter,editor,comments,analytics,announcement,pintura,donations,security,social_web'
+        group: 'site,theme,private,members,portal,newsletter,email,amp,labs,slack,unsplash,views,firstpromoter,editor,comments,analytics,announcement,pintura,donations,security,social_web,explore'
```
ErisDS, line 34: Afaik, the settings API returns all settings by default now and has for years, making this redundant. However, I think this change is fine for now and this would be something to tackle in cleanup (I've found a LOT to do with settings)

ErisDS (approved):
Left a few notes, but all the changes are non-functional so not going to put request changes on as I think this is easy to fix up and merge without requiring re-review!

peterzimon (approved):

apps/admin-x-framework/src/api/settings.ts
sagzy: It is a weird pattern indeed, and something that bugged me earlier. Without this change, the two new `explore_ping` and `explore_ping_growth` settings were not getting returned by the API to Admin Settings. Agree that we should address this properly during cleanup!

ghost/core/core/shared/settings-cache/CacheManager.js
sagzy: I misread the purpose of this code, my bad! I'll clean up the explore settings in this PR and will do a tiny PR to clean up the other 3

sagzy: Fixed in 0514c298b12407c0b418b25872e0394787aab41e

PR #24358, opened 2025-07-15
Fixed e2e tests failing on forks

cmraible:
The e2e test suite should be able to run and pass when running on a PR raised from a fork — this is one of the main requirements of the test suite, since not being able to run our `e2e-browser` test suite from a fork (due to a similar issue with Stripe credentials) has been a pain for so long. 

I inadvertently broke this requirement when adding our Docker hub credentials to the e2e test job for pulling the `mysql` and `tinybird` images — now when running from a fork, the e2e tests will always fail.

This PR fixes this by not providing credentials for the Docker pulls at all. This opens up the potential to hit rate limits more frequently, but for the time being this is preferable to not allowing any PRs raised from a fork to pass CI. Planning to come up with a better solution here, but this should unblock PRs from forks for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -803,9 +803,7 @@ jobs:
         image: tinybirdco/tinybird-local:latest
         ports:
           - 7181:7181
-        credentials:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


ErisDS (approved):

ErisDS:
This is approved as the next best step, but it will likely get rate limited.

It seems the most common workaround is to use a mirror, but not sure we'd find one for tinybird. 

e.g. https://cloud.google.com/artifact-registry/docs/pull-cached-dockerhub-images

The alternative is to use a read-only key and store it in vars, not secrets.

PR #24360, opened 2025-07-15
Added logic for new settings to explore ping

ErisDS:
ref https://linear.app/ghost/issue/PROD-1771/push-support-for-ghost-explore

- This wires up the explore service to honour its new settings
- This protects the whole ping with the main setting
- And partially builds the payload, based on the growth setting which is opt in
- We only even fetch the data if we need it, to reduce unnecessary load

### DIFF
--- a/ghost/core/core/server/services/explore-ping/ExplorePingService.js
+++ b/ghost/core/core/server/services/explore-ping/ExplorePingService.js
@@ -54,15 +54,17 @@ module.exports = class ExplorePingService {
             payload.posts_first = null;
         }
 
-        try {
-            const totalMembers = await this.members.stats.getTotalMembers();
-            payload.members_total = totalMembers;
-        } catch (err) {
-            this.logging.warn('Failed to fetch member statistics', {
-                error: err.message,
-                context: 'explore-ping-service'
-            });
-            payload.members_total = null;
+        if (this.settingsCache.get('explore_ping_growth')) {
+            try {
+                const totalMembers = await this.members.stats.getTotalMembers();
+                payload.members_total = totalMembers;
+            } catch (err) {
+                this.logging.warn('Failed to fetch member statistics', {
+                    error: err.message,
+                    context: 'explore-ping-service'
+                });
+                payload.members_total = null;
+            }
         }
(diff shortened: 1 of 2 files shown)


ghost/core/test/unit/server/services/explore-ping/ExplorePingService.test.js
```diff
@@ -97,7 +98,34 @@ describe('ExplorePingService', function () {
             assert.equal(payload.posts_last, null);
         });
 
-        it('returns null for members_total if no members data available', async function () {
+        it('returns undefined for members_total when members data is null', async function () {
+            membersStub.stats.getTotalMembers.resolves(null);
+
+            const payload = await explorePingService.constructPayload();
+            assert.equal(payload.members_total, undefined);
+        });
```
sagzy, line 108: Should this test be "does not include members_total"? i.e. returns undefined regardless of what member stats returns, as we're not going to fetch member data without the growth flag

ghost/core/core/server/services/explore-ping/ExplorePingService.js
```diff
@@ -1,3 +1,5 @@
+const {ghostStatsInstance} = require('../../../frontend/src/ghost-stats/ghost-stats');
```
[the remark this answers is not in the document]
sagzy: +1, caught by CI failure too

ghost/core/core/server/services/explore-ping/ExplorePingService.js
```diff
@@ -28,38 +28,39 @@ module.exports = class ExplorePingService {
...
-        } catch (err) {
-            this.logging.warn('Failed to fetch member statistics', {
-                error: err.message,
-                context: 'explore-ping-service'
-            });
-            // Continue without member statistics
-        }
-
-        return {
+        const payload = {
```
sagzy: Out of curiosity: do we want to fail the ping entirely if post stats errors? I would have imagined that we'd fallback to null values here if needed, and still send the rest of the payload (similar to member stats)

sagzy (approved):

PR #24364, opened 2025-07-15
Disabled settings editing for Stripe Connect when limit is enabled

sam-lord:
ref [BAE-330](https://linear.app/ghost/issue/BAE-330/ghost-implement-server-side-limit-validations)

Prevent setting up Stripe Connect if the limit is active.

### DIFF
--- a/ghost/core/core/server/services/settings/SettingsBREADService.js
+++ b/ghost/core/core/server/services/settings/SettingsBREADService.js
@@ -11,8 +11,7 @@ const EMAIL_KEYS = ['members_support_address'];
 const messages = {
     problemFindingSetting: 'Problem finding setting: {key}',
     accessCoreSettingFromExtReq: 'Attempted to access core setting from external request',
-    invalidEmail: 'Invalid email address',
-    stripeConnectDisabled: 'Stripe Connect is disabled on this plan'
+    invalidEmail: 'Invalid email address'
 };
 
 class SettingsBREADService {
@@ -197,9 +196,11 @@ class SettingsBREADService {
         }
 
         if (stripeConnectData) {
-            if (this.limitsService.isDisabled('limitStripeConnect')) {
+            try {
+                await this.limitsService.errorIfWouldGoOverLimit('limitStripeConnect');
+            } catch (error) {
                 throw new NoPermissionError({
-                    message: tpl(messages.stripeConnectDisabled)
+                    message: error.message
                 });
             }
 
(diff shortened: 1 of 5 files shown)


ghost/core/test/e2e-api/admin/settings.test.js
```diff
@@ -344,6 +344,43 @@ describe('Settings API', function () {
 
             sinon.assert.calledOnce(loggingStub);
         });
+
+        it('cannot edit Stripe settings when Stripe Connect limit is enabled', async function () {
```
lsinger, line 384: I'd be a bit more comfortable if we were also testing the positive case (limit not enabled => can edit settings), just to make sure this isn't something that always fails, anyway, because there's an issue in the test setup that we're not seeing. Or do we have that test somewhere else possibly?

ghost/core/core/server/services/settings/SettingsBREADService.js
```diff
@@ -194,6 +197,12 @@ class SettingsBREADService {
         }
 
         if (stripeConnectData) {
+            if (this.limitsService.isDisabled('limitStripeConnect')) {
+                throw new NoPermissionError({
+                    message: tpl(messages.stripeConnectDisabled)
+                });
+            }
```
aileen: I should've mentioned that earlier, but the limit service provides you with the correct error message and type. There's no clean method to get it, particularly while staying in sync and not switching to async code. But all you'd need to change is using this line instead of throwing the error:

`await limitService.errorIfWouldGoOverLimit('limitStripeConnect');`

This will throw the correct `HostLimitError` with the error message defined on the limit in our DB.

sam-lord: Ooh, I'll have a quick go at that

ghost/core/core/server/services/settings/SettingsBREADService.js
```diff
@@ -194,6 +196,8 @@ class SettingsBREADService {
         }
 
         if (stripeConnectData) {
+            await this.limitsService.errorIfWouldGoOverLimit('limitStripeConnect');
```
aileen, line 199: Does this work when no `limitStripeConnect` limit is present in the config?

aileen: I have tested this out locally and can confirm that it works as it should.

aileen (approved):
LGTM

PR #24367, opened 2025-07-15
Removed duplicate global-data request

ibalosh:
* removed duplicated global-data-request, so that we import it from single place - framework

Currently we are importing same HTTP request stub data from two different places, settings and framework.
By using single import path, we are avoiding fixtures/stubs drift, by having to maintain them in 2 places.


cmraible (approved):
Makes sense to me:
- Low / no risk change — only test files changed
- Each admin-x app can benefit from any of the endpoints/fixtures we add to this object, which makes setting up new admin-x-apps a lot less tedious
- Reduces the potential for drift, as you mentioned
- Each app can still use custom fixtures for specific scenarios as needed

LGTM 👍

PR #24370, opened 2025-07-15
Render block rate limit errors

sagzy:
ref https://linear.app/ghost/issue/PROD-2261

- the infrastructure will return HTTP 429/empty body when a user is rate-limited, and HTTP 403/empty body when a user is blocked
- added support for non-json error responses
- added toast error for rate-limits, when the user is performing an action (e.g. liking a post)
- todo: add general error screen for rate-limits and blocks when reloading screens

### DIFF
--- a/apps/admin-x-activitypub/src/api/activitypub.ts
+++ b/apps/admin-x-activitypub/src/api/activitypub.ts
@@ -230,17 +230,25 @@ export class ActivityPubAPI {
             return null;
         }
 
-        const json = await response.json();
-
         if (!response.ok) {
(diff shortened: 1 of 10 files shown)


sagzy (commented):
One side thought: we could be using an `Error` class instead of the current `ApiError` object:

```
export class ApiError extends Error {...}
```

and then we can get rid of `isApiError()` and simply do `if (error instanceof ApiError) {...}` in feed, inbox, etc.

—

But this extends the scope of this change, so let's merge as is for now and improve this later :)

PR #24375, opened 2025-07-15
Added small cache TTLs to frontend and content API

ErisDS:
ref https://linear.app/ghost/issue/PROD-1591/breaking-change-update-our-caching-defaults-dont-default-to-assuming

- The zero cache TTLs on public content were added in a time when headless was all the rage, and so it made sense for this to be the default
- However, for most people these TTLs are too low.
- Given that headless is a more technical setup, it seems reasonable to expect that people choosing headless can configure their cache defaults themselves
- And that we should optimise for the standard usecase where having a small cache TTL makes sense to mitigate load spikes


kevinansfield (approved):

ErisDS:
Closing in favour of  https://github.com/TryGhost/Ghost/pull/24415

PR #24378, opened 2025-07-15
🐛 Fixed missing `lexical` field in post/page webhook payloads

kevinansfield:
no issue

- webhook serializer has an explicit list of formats to include when building payloads but it was missing the `lexical` source format despite including the deprecated `mobiledoc` source format


kevinansfield:
Closing in favor of removing the `mobiledoc` field instead. Related PRs to come...

PR #24379, opened 2025-07-15
Fixed Ember Analytics routing bug with beta flags

9larsons:
no ref
- removed infinite loop when missing config
- fixed logging in routing to Analytics when `trafficAnalytics` flag OR `ui60` flag is enabled; previously required both which was incorrect

We should be routing users to Analytics when either the beta Analytics flag is enabled or the ui60 flag is enabled. We will be removing the config checks from Ember/Admin in another PR - this one needed to be removed for the regression test to pass.

### DIFF
--- a/ghost/admin/tests/acceptance/signin-test.js
+++ b/ghost/admin/tests/acceptance/signin-test.js
@@ -1,3 +1,4 @@
+import config from 'ghost-admin/config/environment';
 import {Response} from 'miragejs';
 import {authenticateSession, invalidateSession} from 'ember-simple-auth/test-support';
 import {
@@ -6,6 +7,7 @@ import {
     it
 } from 'mocha';
 import {click, currentURL, fillIn, find, findAll} from '@ember/test-helpers';
+import {enableLabsFlag} from '../helpers/labs-flag';
 import {expect} from 'chai';
 import {setupApplicationTest} from 'ember-mocha';
 import {setupMirage} from 'ember-cli-mirage/test-support';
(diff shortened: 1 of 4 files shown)


ghost/admin/tests/acceptance/signin-test.js
```diff
@@ -94,5 +96,33 @@ describe('Acceptance: Signin', function () {
             await click('[data-test-button="sign-in"]');
             expect(currentURL(), 'currentURL').to.equal('/dashboard');
         });
+
+        it('submits successfully with traffic analytics enabled', async function () {
```
cmraible, line 100:
```suggestion
        it('redirects to analytics after successful signin if analytics is enabled', async function () {
```

cmraible:
Cool, just want to make sure I'm testing the right things here. 

> removed infinite loop when missing config

Scenario: `trafficAnalytics` labs flag true, `ui60` labs flag false, and no Tinybird config
- On `main` Admin completely fails to load — I get stuck on the Ghost logo, which freezes, and then eventually Chrome kills the tab. 
- On this branch running with the same config, Ghost loads immediately and goes to the analytics app (it crashes, which we know about and I'm working on — but this is a step in the right direction)

Scenario: `trafficanalytics` labs flag false, `ui60` labs flag true, and no Tinybird config
- On `main`, the same thing happens — stuck on Ghost logo, Chrome kills tab
- On this branch, same thing happens — Ghost loads to the analytics app (which then crashes)


Is that the intent of this PR? Or is there more / something different?

9larsons:
> Cool, just want to make sure I'm testing the right things here.
> 
> > removed infinite loop when missing config
> 
> Scenario: `trafficAnalytics` labs flag true, `ui60` labs flag false, and no Tinybird config
> 
> * On `main` Admin completely fails to load — I get stuck on the Ghost logo, which freezes, and then eventually Chrome kills the tab.
> * On this branch running with the same config, Ghost loads immediately and goes to the analytics app (it crashes, which we know about and I'm working on — but this is a step in the right direction)

My intent here was to prevent Ember from failing, while the Stats crash is its own issue that you're working on, yes.

Without `config` we don't see the listing in the nav (which we're about to change) but we do see the Stats app attempt to load and crash. That seems right to me, as the subsequent changes will patch those two pieces.

cmraible (approved):
LGTM

PR #24380, opened 2025-07-15
Fixed Tinybird token fetching failing without Tinybird config

cmraible:
ref https://linear.app/ghost/issue/PROD-2309/the-analytics-tab-does-not-load-with-ui60-enabled-and-no-tinybird

Since the inception of the analytics app (`apps/stats`), we've always gated loading the app based on whether or not the `tinybird:stats` configuration block is provided. As we move toward 6.0, now the Analytics app is going to replace the dashboard for all users, regardless of whether they have traffic analytics enabled.

Currently when the `ui60` flag is enabled but there is no `tinybird` configuration available, the analytics app throws an error and fails to load because the request to fetch a Tinybird token fails. This PR modifies the logic in the GlobalDataProvider to first check if the `tinybird` configuration is set, and only if it is, it tries to fetch a token. Since the token is fetched in a react hook (`useTinybirdToken`), we can't conditionally call the hook — hence modifying the hook itself to accept a boolean `enabled` parameter — if this is set to false, it will immediately return undefined, rather than trying to fetch a token and failing.

This fixes the error thrown when loading the app by only trying to fetch a token when a token should be available, so the analytics app will load. We already have much of the UI logic to hide the web analytics features from the UI if config isn't set, so this almost entirely fixes the app when we're in the `ui60` but no `tinybird` config state.


apps/stats/src/providers/GlobalDataProvider.tsx
```diff
@@ -38,13 +38,16 @@ const GlobalDataProvider = ({children}: { children: ReactNode }) => {
     const settings = useBrowseSettings();
     const site = useBrowseSite();
     const config = useBrowseConfig() as unknown as { data: Config & { config: { stats?: StatsConfig } } | null, isLoading: boolean, error: Error | null };
```
9larsons, line 40: Looks good overall. This type is horrendous (my fault, not yours).

I can probably clean that up when making the frontend config changes/cleanup we discussed (possibly) doing next week.

9larsons (approved):

apps/stats/src/providers/GlobalDataProvider.tsx
cmraible: Sounds good!

PR #24381, opened 2025-07-15
Removed config.stats requirement from admin routing logic

9larsons:
ref https://linear.app/ghost/issue/PROD-2312
- posts/stats apps contain logic for showing/hiding Tinybird-based metrics

While this was necessary early on, we shouldn't put routing logic behind config - each app has its own logic for displaying Tinybird-based components that depend on the config.

Note: currently the apps crash as they expect `config`. That is being fixed separately. ref https://github.com/TryGhost/Ghost/pull/24380.


ghost/admin/tests/acceptance/analytics-navigation-test.js
```diff
@@ -0,0 +1,344 @@
...
+import {setupMirage} from 'ember-cli-mirage/test-support';
+import {visit} from '../helpers/visit';
+
+describe('Acceptance: Analytics Navigation', function () {
+    let hooks = setupApplicationTest();
+    setupMirage(hooks);
+
+    async function createUserAndSignIn(server, {role = 'Administrator', email = 'admin@example.com'} = {}) {
+        let roleObj = server.create('role', {name: role});
+        server.create('user', {
```
ibalosh: I would separate creating from singing in. It reads better, separates responsibilities. 
Just like update role is separate. At one point these might get extracted further. When functionality is separated it's easier to see things to refactor later on

ghost/admin/tests/acceptance/analytics-navigation-test.js
```diff
@@ -0,0 +1,304 @@
...
+        // Disable all flags by default
+        disableLabsFlag(this.server, 'trafficAnalytics');
+        disableLabsFlag(this.server, 'ui60');
+    });
+
+    afterEach(function () {
+        cleanupMockAnalyticsApps();
+    });
+
+    describe('Stats-X route (/analytics)', function () {
```
ibalosh, line 109: nit: separating tests which are under each describe to separate group of tests per file, in maybe `routes` folder could read better?

ghost/admin/tests/acceptance/analytics-navigation-test.js
```diff
@@ -0,0 +1,304 @@
...
+            expect(currentURL()).to.equal('/dashboard');
+        });
+
+        it('redirects contributors to posts', async function () {
+            enableLabsFlag(this.server, 'trafficAnalytics');
+
+            updateUserRole(this.server, 'Contributor');
+
+            await visit('/analytics');
+            expect(currentURL()).to.equal('/posts');
```
ibalosh, line 135: Are these just redirection checks? 

If so, wouldn't it make more sense to check it on lower level, like API level whether redirection happened. Not 100% sure of the behavior checked, since I see only URL checked whether it's the same.

Wouldn't it be better to check if certain element is on page, like text or something to make sure that flag worked or not?

Just thinking out loud what test is doing whether it can bring more value.

ghost/admin/tests/acceptance/analytics-navigation-test.js
```diff
@@ -0,0 +1,304 @@
...
+            expect(analyticsLink).to.exist;
+            expect(analyticsLink.textContent).to.contain('Analytics');
+        });
+
+        it('hides Analytics link when trafficAnalytics is disabled', async function () {
+            enableLabsFlag(this.server, 'ui60');
+
+            await visit('/dashboard');
+            
+            let analyticsLink = findAnalyticsNavLink();
```
ibalosh, line 220: I think these could be a one liner expect

ibalosh (approved):
overall these look good, to not block release will approve them.
I will do one more check in the morning

cmraible:
Nice, checked through all the cases and it's working exactly how I'd expect it 🙌 (except for locations tab, but that's tracked separately). 

Just going to take a quick pass at the code and tests now

cmraible:
Found one more case that we need to handle: `ui60` enabled, `trafficAnalytics` disabled. Doesn't necessarily need to be included in this PR, but we'll need to cover it when we remove the `trafficAnalytics` flag. 

Currently with only `ui60` enabled, publishing a post w/ email redirects to the _old_ post analytics

ghost/admin/app/routes/stats.js
```diff
@@ -24,8 +24,8 @@ export default class StatsRoute extends AuthenticatedRoute {
             return this.transitionTo('site');
         }
 
-        // This ensures that we don't load this page if the stats config is not set
-        if (!(this.config.stats && this.feature.trafficAnalytics)) {
+        // ensure we don't load the app if flags aren't enabled
+        if (!this.feature.trafficAnalytics || !this.feature.ui60) {
```
cmraible: Just curious, what even is this route? AFAICT it's not used anywhere. Doesn't need to be addressed here, but unless I'm missing something I think we can delete this file

ghost/admin/tests/acceptance/analytics-navigation-test.js
```diff
@@ -0,0 +1,304 @@
...
+            let post = this.server.create('post', {
+                status: 'published',
+                title: 'Test Post for Analytics'
+            });
+
+            await visit('/posts');
+            await clickPost(post.id);
+            if (await clickPostAnalytics()) {
+                expect(currentURL()).to.equal(`/posts/analytics/beta/${post.id}`);
+            }
```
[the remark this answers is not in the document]
cmraible: This seems worth fixing to me

ghost/admin/tests/acceptance/analytics-navigation-test.js
```diff
@@ -0,0 +1,304 @@
...
+            let post = this.server.create('post', {
+                status: 'published',
+                title: 'Test Post for Analytics'
+            });
+
+            await visit('/posts');
+            await clickPost(post.id);
+            if (await clickPostAnalytics()) {
+                expect(currentURL()).to.equal(`/posts/analytics/${post.id}`);
+            }
```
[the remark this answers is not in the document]
cmraible: This one too

ghost/admin/tests/acceptance/analytics-navigation-test.js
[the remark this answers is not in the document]
cmraible: No I didn't mean for you to fix it!

ghost/admin/app/routes/stats.js
9larsons: Good catch, I didn't realize this was still hanging around (or that we made a new route at all..).

cmraible (approved):
LGTM! Only thing I'd recommend fixing before merging is the two coderabbit suggestions. Few other comments but nothing that needs to be addressed in this PR.

9larsons:
> Found one more case that we need to handle: `ui60` enabled, `trafficAnalytics` disabled. Doesn't necessarily need to be included in this PR, but we'll need to cover it when we remove the `trafficAnalytics` flag.
> 
> Currently with only `ui60` enabled, publishing a post w/ email redirects to the _old_ post analytics

Fixed. This was a good one to tackle here.

PR #24382, opened 2025-07-15
Added fallback error page for unhandled errors in stats app

cmraible:
ref https://linear.app/ghost/issue/PROD-2313/fix-fallback-error-page-for-unhandled-errors-in-the-stats-app

There is currently a bug in the stats app that throws an error from the GlobalDataProvider, before the app has fully rendered. This error is preventing the whole app from loading, and instead it is falling back to this very unfriendly error page:

image

The fix for that error is irrelevant to this PR — it will be fixed in a separate PR. This PR is strictly to add a fallback error page for any unhandled errors in the stats app, so no user ever sees the abomination above. For now this adds the same exact fallback error handler that is used for all the `admin-x-` apps, defined in the Ember admin-x-component file.

image


apps/stats/src/providers/GlobalDataProvider.tsx
```diff
@@ -59,6 +59,9 @@ const GlobalDataProvider = ({children}: { children: ReactNode }) => {
         title: site.data?.site.title
     };
 
+    // Uncomment this line to test the error boundary
+    // throw new Error('test');
```
cmraible: You can uncomment this error line when running locally to trigger the error page, e.g. for testing the design. We should remove this before merging, just here to make testing easier.

apps/stats/src/components/StatsErrorPage.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+interface StatsErrorPageProps {
+    error?: Error;
+}
+
+const StatsErrorPage: React.FC<StatsErrorPageProps> = ({error}) => {
+    console.error(error);
```
cmraible: Included this here to demonstrate that we have the full error object available in this component, so we could potentially provide more information to the user than just "Loading interrupted". We should remove the console.error before merging.

apps/stats/src/App.tsx
```diff
@@ -1,10 +1,35 @@
...
+        super(props);
+        this.state = {hasError: false};
+    }
+
+    static getDerivedStateFromError(error: Error) {
+        return {hasError: true, error};
+    }
+
+    componentDidCatch(error: unknown, errorInfo: React.ErrorInfo) {
+        console.error('Stats app error:', error, errorInfo);
```
cmraible: We might want to add a Sentry capture here?

peterzimon (approved):

apps/stats/src/App.tsx
```diff
@@ -1,4 +1,6 @@
 import GlobalDataProvider from './providers/GlobalDataProvider';
+import React from 'react';
```
9larsons: This shouldn't be necessary.

9larsons (approved):

apps/stats/src/App.tsx
cmraible: Good catch, thanks — fixed.

PR #24385, opened 2025-07-16
Removed `ui60` flag for Settings <> Admin navigation

peterzimon:
ref https://linear.app/ghost/issue/PROD-2317/remove-settings-related-ui60-flag

- Depending on the state of the `ui60` flag we redirected users to different pages when quitting Settings. Also we had different main navigation menus in Admin depending on the `ui60` flag. This PR removes the flag related to these behaviors and defaults to correct behavior in 6.0.

### DIFF
--- a/apps/admin-x-settings/src/MainContent.tsx
+++ b/apps/admin-x-settings/src/MainContent.tsx
@@ -33,7 +33,7 @@ const MainContent: React.FC = () => {
         const handleKeyDown = (event: KeyboardEvent) => {
             if (event.key === 'Escape') {
                 confirmIfDirty(isDirty, () => {
-                    navigateAway('/analytics');
+                    navigateAway('/');
                 });
             }
         };
--- a/apps/admin-x-settings/src/MainContent.tsx
+++ b/apps/admin-x-settings/src/MainContent.tsx
@@ -5,7 +5,6 @@ import Users from './components/settings/general/Users';
 import {Heading, confirmIfDirty, topLevelBackdropClasses, useGlobalDirtyState} from '@tryghost/admin-x-design-system';
 import {ReactNode, useEffect} from 'react';
 import {canAccessSettings, isEditorUser} from '@tryghost/admin-x-framework/api/users';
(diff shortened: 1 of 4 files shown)


apps/admin-x-settings/src/MainContent.tsx
```diff
@@ -36,15 +33,7 @@ const MainContent: React.FC = () => {
...
-                    if (config.labs.ui60) {
-                        navigateAway('/analytics');
-                    } else {
-                        if (hasActivityPub && hasAdminAccess(currentUser)) {
-                            navigateAway('/activitypub');
-                        } else {
-                            navigateAway('/dashboard');
-                        }
-                    }
+                    navigateAway('/analytics');
```
sagzy: I think we can now change this to `navigateAway('/')` instead. That way, Ghost Admin renders the default sidebar menu (which should be analytics in 6.x for most users, but might be something else if traffic analytics is not available for some reason, e.g. to self-hosters on the legacy self-hosting setup)

apps/admin-x-settings/src/components/ExitSettingsButton.tsx
```diff
@@ -1,20 +1,11 @@
...
-    const {settings, currentUser, config} = useGlobalData();
-    const [hasActivityPub] = getSettingValues(settings, ['social_web_enabled']) as [boolean];
 
     const navigateAway = () => {
-        if (config.labs.ui60) {
-            window.location.hash = '/analytics';
-        } else {
-            window.location.hash = (hasActivityPub && hasAdminAccess(currentUser)) ? '/activitypub' : '/dashboard';
-        }
+        window.location.hash = '/analytics';
```
sagzy: Same here

apps/admin-x-settings/src/MainContent.tsx
peterzimon: Great point!

9larsons (approved):
I think this is fine. We have more places to remove ui60, and we're going to end up merging more in, but I can clean that up.

PR #24401, opened 2025-07-16
Removed trafficAnalytics flag from apps/

9larsons:
- removed the trafficAnalytics flag and behavior from Settings
- updated readme in stats app


cmraible:
LGTM except for a couple of the browser tests — looks like they are expecting the redirect to the /dashboard route, but now they're getting /analytics (which is correct — just need to update the assertions I think)

cmraible:
Rebasing now that 6.x branch has all tests passing 🤞

cmraible (approved):
LGTM! Tests are all passing now that we've fixed them on `6.x`

PR #24404, opened 2025-07-17
6.x Theme Asset Behaviour Changes

ErisDS:
closes: https://linear.app/ghost/issue/PROD-2214/return-404-from-frontends-assets-folder

These two commits go hand in hand.

If I try to add fallthrough:false to our theme assets to solve the above issue - the entire theme fails to render.

That's cos every single request goes through that middleware!

So, the [first commit](https://github.com/TryGhost/Ghost/commit/072e9d99db9bf190ec548a796bd13fc39f746a69) adds a line to correctly skip the middleware if the file doesn't have an extension,
This is potentially breaking if anyone was using this to render a file from their theme without an extension, but feels low-risk

The second commit then adds the fallthrough rule so we can not redirect missing assets and end on a 404 HTML page, but rather error immediately - with an exception for robots.txt

### DIFF
--- a/ghost/core/test/unit/frontend/web/middleware/static-theme.test.js
+++ b/ghost/core/test/unit/frontend/web/middleware/static-theme.test.js
@@ -13,13 +13,17 @@ describe('staticTheme', function () {
 
     beforeEach(function () {
         req = {};
-        res = {};
+        res = {
+            setHeader: sinon.stub()
+        };
 
         activeThemeStub = sinon.stub(themeEngine, 'getActive').returns({
             path: 'my/fake/path'
         });
 
-        expressStaticStub = sinon.spy(express, 'static');
+        expressStaticStub = sinon.stub(express, 'static').returns(function (_req, _res, _next) {
+            _next();
+        });
     });
 
     afterEach(function () {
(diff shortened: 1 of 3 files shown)


ghost/core/test/unit/frontend/web/middleware/static-theme.test.js
```diff
@@ -225,4 +229,170 @@ describe('staticTheme', function () {
...
+            staticTheme()(req, res, function next() {
+                activeThemeStub.called.should.be.false();
+                expressStaticStub.called.should.be.false();
+
+                done();
+            });
+        });
+
+        it('should NOT skip for file with extension even without slash', function (done) {
+            req.path = '/robots.txt';
```
kevinansfield: Should this be testing for something other than `robots.txt` considering that's a special-case?

kevinansfield (approved):

ghost/core/test/unit/frontend/web/middleware/static-theme.test.js
ErisDS: Good shout

kevinansfield (approved):

PR #24405, opened 2025-07-17
Removed trafficAnalytics and ui60 flags from Ember

9larsons:
ref https://linear.app/ghost/issue/PROD-2321
- removed trafficAnalytics flag from Ember routing
- removed remnants of ui60 flag from Ember routing
- removed any pathing to /dashboard, but left the route accessible and mounted
- updated all associated tests


cmraible:
The ember admin tests are all timing out. It's difficult to reason about which change in this PR is causing them to fail. I'm not against doing all of these changes in a single PR, but I'd recommend going about it in smaller, more incremental steps.

Now that we have a clean starting point with `6.x`, we know that any tests that are failing here are directly caused by changes in this PR. The first [commit](https://github.com/TryGhost/Ghost/pull/24405/commits/94e248010518ac6bffbc89e16947f7fc1d87a03c) here changes 22 files, which makes it a bit of a wild goose chase to try to figure out which changes are causing the tests to act up. I'd recommend:
- Making the smallest code change you can find — i.e. just make the change to the `ghost/admin/app/components/editor/modals/publish-flow.js` file, which seems to be fairly self-contained
- Running all of the tests, and updating any that need to be fixed for that change
- Committing the result once all tests are passing
- Then picking off the next smallest code change and starting again

I think one of two things will happen:
- You'll come across the failing tests one-by-one, and they'll be easy to fix because you know exactly what changed
- or you'll find that only one of these changes is causing all of the tests to fail — but since you know exactly what that change is, it should be a lot easier to figure out why all of the tests are suddenly failing, and to ultimately fix them

PR #24409, opened 2025-07-17
Removed failing Ember tests related to Dashboard routing

9larsons:
no ref

This is removing the tests that no longer apply and are preventing us from being green on CI. We will follow up with changes that remove the flags/Dashboard routing and properly fix the state of things.


cmraible:
LGTM, will approve once tests pass. These are test cases needed for 5.x but they don't make sense for 6.x so agree that just removing them is the right thing here.

cmraible (approved):

PR #24419, opened 2025-07-17
Removed beta flags from new Posts Analytics routing

9larsons:
- removed conditional routing to/from the posts-x (new analytics) app
- removed routing to the Ember posts analytics page; this route is still present and will be cleaned up separately
- removed a couple references to trafficAnalytics/ui60/posts-x in the Ember analytics files
- removed unnecessary test cases for routing conditions
- updated existing tests

The intent is to remove the `trafficAnalytics` and `ui60` flags from the code base. We will be updating the actual routes in a separate PR; the intended end-state for this PR is that we remove any conditionals that direct to the old/new Analytics, and only point to the new.

There's still some references to the old (Ember) analytics in various areas of the code base (e.g. Dashboard, Debug page) that we will review separately.

### DIFF
--- a/ghost/admin/app/components/posts/analytics.hbs
+++ b/ghost/admin/app/components/posts/analytics.hbs
@@ -73,11 +73,6 @@
                             <li>
                                 <a class="view-browser" href="{{this.post.url}}" target="_blank" rel="noopener noreferrer">View in browser</a>
                             </li>
-                            {{#if (and (gh-user-can-admin this.session.user) (or (feature "trafficAnalytics") (feature "ui60")))}}
-                                <li>
-                                    <LinkTo class="edit-post" @route="posts-x" @model={{this.post.id}}>Switch to Analytics (beta)</LinkTo>
-                                </li>
-                            {{/if}}
                             <li>
                                 <button
                                     type="button"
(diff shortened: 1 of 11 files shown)


ghost/admin/app/components/posts/analytics.hbs
```diff
@@ -73,11 +73,6 @@
                             <li>
                                 <a class="view-browser" href="{{this.post.url}}" target="_blank" rel="noopener noreferrer">View in browser</a>
                             </li>
-                            {{#if (and (gh-user-can-admin this.session.user) (or (feature "trafficAnalytics") (feature "ui60")))}}
-                                <li>
-                                    <LinkTo class="edit-post" @route="posts-x" @model={{this.post.id}}>Switch to Analytics (beta)</LinkTo>
-                                </li>
-                            {{/if}}
```
9larsons, line 80: This is optional. It felt worth cleaning up here, although we're not fully cleaning up the analytics route. If wanted, we can ditch this part from this set of changes.

ghost/admin/app/routes/posts/analytics.js
```diff
@@ -47,20 +47,6 @@ export default class Analytics extends AuthenticatedRoute {
         if (user.isContributor && !post.isDraft) {
             return this.replaceWith(returnRoute);
         }
-
```
9larsons, line 50: This was dead code that just led to confusion when reading through the code base. I think it's best to remove it here as it references the flags and the posts-x route.

cmraible (approved):
Nice one, LGTM!

PR #24422, opened 2025-07-17
Cleaned up Analytics routing logic

9larsons:
- removed `trafficAnalytics` and `ui60` flags from Ember routing logic; this means we are using Analytics in lieu of Dashboard and Dashboard is unlisted
- adjusted Ember routing to use `home` as the default logic where possible when navigating away from a route (easier to reason about and maintain)
- updated Ember acceptance tests; removed flag use and added Analytics mocks where necessary
- Dashboard is still directly navigable to

The purpose here is to:
- make Analytics the default for Admin users in lieu of Dashboard, regardless of flags
- clean up routing logic to make it easier to reason about


ghost/admin/app/routes/home.js
```diff
@@ -16,10 +16,12 @@ export default class HomeRoute extends AuthenticatedRoute {
...
 
-        if ((this.feature.ui60 || this.feature.trafficAnalytics) && this.session.user?.isAdmin) {
+        if (this.session.user?.isAdmin) {
             this.router.transitionTo('stats-x');
+        } else if (this.session.user?.isContributor) {
+            this.router.transitionTo('posts');
         } else {
-            this.router.transitionTo('dashboard');
+            this.router.transitionTo('site');
         }
```
9larsons, line 25: This is the core routing logic and I've tried to point everything here when we're forcing users away from a route, e.g. they don't have access as an Author.

ghost/admin/app/routes/stats-x.js
```diff
@@ -1,20 +1,13 @@
...
-        if (this.session.user.isContributor) {
-            return this.transitionTo('posts');
-        } else if (!this.session.user.isAdmin) {
-            return this.transitionTo('site');
-        }
 
-        if (!this.feature.trafficAnalytics && !this.feature.ui60) {
+        if (!this.session.user.isAdmin) {
             return this.transitionTo('home');
         }
```
9larsons, line 12: It's much cleaner to not replicate `home` logic in our various routes, and instead punt back to the `home` route where appropriate. In this case, only Admins should see Analytics.

ghost/admin/tests/acceptance/dashboard-test.js
```diff
@@ -76,10 +76,5 @@ describe('Acceptance: Dashboard', function () {
             await visit('/dashboard');
```
9larsons, line 76: Dashboard is still navigable to, but in general we shouldn't point to it. There's some various cleanup to do here that we can follow up with in a change that more fully retires Dashboard.

ghost/admin/app/routes/mentions.js
```diff
@@ -19,17 +19,17 @@ export default class MentionsRoute extends AuthenticatedRoute {
...
-        if (!this.feature.webmentions) {
-            if (this.config?.labs?.ui60) {
-                return this.transitionTo('stats-x');
-            } else {
-                return this.transitionTo('dashboard');
-            }
+        
+        // Check feature flag only when user exists (implicitly requires authentication)
+        // This follows the same pattern as other routes like tags, admin, etc.
+        if (this.session.user && !this.feature.webmentions) {
```
9larsons, line 31: This looks confusing, because it isn't very intuitive. We end up with a bit of a race condition with AuthenticatedRoute trying to redirect if unauthenticated and a transition to `home`, which is also an AuthenticatedRoute, because features aren't gated by session while auth is. It so happens to be that most of our other checks are looking at session/user.

I think a better solution on the whole would be to bake features into the Route, OR simply put the checks in the `afterModel` which doesn't have this collision due to our custom auth implementation. It's not as performant as we'd be loading the model even if we don't intend to display it, but it would be much easier to understand.

Regardless, we don't touch Admin routing too much, so I don't think we need to worry about this. Just wanted to call it out.

9larsons:
Closing in lieu of https://github.com/TryGhost/Ghost/pull/24427/files.

It seems the routing to `home` changes have some unexpected consequences in Ember tests that do not show locally. The other PR doesn't make those changes and seems to pass CI fine.

PR #24426, opened 2025-07-17
Revert "Added small cache TTLs to frontend and content API (#24415)"

cmraible:
This commit caused the browser tests to start failing — there are a few browser tests where we will make a change to something in admin, then check if that change is reflected in portal. With the TTL, the browser is caching the responses on the frontend, so portal isn't updating immediately and the tests are failing. 

This just reverts the commit to get everything passing on `6.x` again — we can squash these two next time we rebase, and reintroduce the TTL commit with a fix for the browser tests included.


cmraible:
Github seems to be confused with the checks on this branch — not sure how to get around that, so I'm going to try pushing this same branch to a new PR and try again. The browser tests did pass, but the checks aren't all showing in the PR UI.

PR #24428, opened 2025-07-17
Revert "Added small cache TTLs to frontend and content API (#24415)"

cmraible:
Reverts 0ba4826646a3f7af0da9cd5b18597d47ef840d48

The referenced commit introduced a TTL cache to some frontend and content API requests, which has broken browser tests on the `6.x` branch. For now I'm reverting this commit to get browser tests passing on the `6.x` branch again — we can squash this revert commit and the original commit next time we rebase, and reintroduce the cache TTL commit with a fix for the browser tests to keep the history clean.


ErisDS:
I've deleted this now

PR #24440, opened 2025-07-17
Removed ui60 and trafficAnalytics flags from Ghost backend

9larsons:
- removed flag from labs
- removed trafficAnalytics from settingsHelper (calculated web_analytics_enabled setting)
- removed mention of flags from docs
- updated posts list analytics to not be dependent on trafficAnalytics flag
- removed ui60 from Settings

This ended up being a bit of a grab bag at the end. The Ember posts list content is dependent on the SettingsHelper update here, so it makes reasonable sense to roll into this one PR.

ui60 was still hanging around in Settings and we want to remove that before it causes more conflicts.


cmraible:
I think we may as well remove the ui60 flag from the UI while we're at it: https://github.com/TryGhost/Ghost/blob/054812a62c68d3ca52bebb59cc1286dce778af0c/apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx#L26-L30

apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx
```diff
@@ -23,10 +23,6 @@ const features: Feature[] = [{
     title: 'Import Member Tier',
     description: 'Enables tier to be specified when importing members',
     flag: 'importMemberTier'
-}, {
```
cmraible, line 26: 🎉

cmraible (approved):
LGTM

PR #24454, opened 2025-07-20
✨Updated theme i18n to use i18next

cathysarisky:
- i18n support was added to themes a _really_ long time ago, and at the time there weren't great libraries
- Since then, i18n has been added to more parts of Ghost, but leveraging the awesome i18next library
- This PR brings theme i18n into line with the rest of Ghost by swapping out the custom code with i18next 
- The change exists behind a labs flag, with duplicated logic to make it easier to clean up later

---

> [!NOTE]
> Introduces i18next-powered theme translations gated by a new `labs` flag, with runtime reinitialization and comprehensive tests.
> 
> - **Feature Flag & Settings**
>   - Add private labs flag `themeTranslation` and expose toggle in `apps/admin-x-settings`.
>   - Update config snapshot to include `themeTranslation`.
> - **Theme i18n Integration**
>   - `core/frontend/helpers/t`: switch between legacy `themeI18n` and new `themeI18next` based on `labs.isSet('themeTranslation')`, with lazy init.
>   - `theme-engine/active`: `initI18n` initializes either `i18n` or new `i18next` service per flag.
>   - `handlebars` service: export `themeI18next` alongside existing `themeI18n`.
> - **Bridge & Events**
>   - `core/bridge`: track `themeTranslation` state; on `settings.labs.edited` or locale change, re-init active theme i18n.
> - **New i18next Theme Service**
>   - Add `theme-engine/i18next/ThemeI18n` and index: loads theme locale JSON from `content/themes/<theme>/locales`, falls back to `en` or key, supports interpolation, and exposes `t`.
> - **Core i18n Library Enhancements**
>   - Extend `@tryghost/i18n` to support `ns: 'theme'`, load theme resources from a directory, disable HTML escaping for theme interpolation, and retain existing namespaces.
>   - Minor cleanup: comment out obsolete labs listener in server i18n.
> - **Tests**
>   - Add unit tests for new `t` helper path and `ThemeI18n` behavior, plus theme resource handling in `@tryghost/i18n`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit edd29e6b1e9eb101aab9cc50d4199f2bce6d45f0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


cathysarisky:
@coderabbitai please no more reviews

muratcorlu:
@JohnONolan Hi John! Let me try my chance starting from this PR. Are we able to get some attention to this? A very valuable effort is waiting here since a long time.

ghost/core/core/frontend/helpers/t.js
```diff
@@ -26,5 +30,27 @@ module.exports = function t(text, options = {}) {
...
 
-    return themeI18n.t(text, bindings);
+    if (labs.isSet('themeTranslation')) {
+        // Use the new translation package when feature flag is enabled
+        // eslint-disable-next-line no-console
+        console.log('======themeI18next', themeI18next);
+        if (!themeI18next._i18n) {
+            themeI18next.init({
+                activeTheme: settingsCache.get('active_theme'),
+                locale: config.get('locale')
```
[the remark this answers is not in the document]
kevinansfield: @cathysarisky is there a reason `config` is used here in place of `settingsCache`?

ghost/core/core/frontend/helpers/t.js
```diff
@@ -26,5 +30,27 @@ module.exports = function t(text, options = {}) {
...
+            });
+        }
+        return themeI18next.t(text, bindings);
+    } else {
+        // Use the existing translation package when feature flag is disabled
+        // eslint-disable-next-line no-console
+        console.log('======themeI18n', themeI18n);
```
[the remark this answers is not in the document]
kevinansfield: If these are purely development debug logs I think it would be preferable to use the `debug` utility, otherwise we'll be creating a lot of log noise. Unless this was just some missed cleanup, in which case lets remove it 🙂

ghost/core/core/shared/labs.js
```diff
@@ -34,7 +34,8 @@ const GA_FEATURES = [
 const PUBLIC_BETA_FEATURES = [
     'superEditors',
     'editorExcerpt',
-    'additionalPaymentMethods'
+    'additionalPaymentMethods',
+    'themeTranslation'
```
kevinansfield: We don't tend to put features directly into public beta, there's a surprising number of people who will check everything available without knowing why or what they should be checking.

Would rolling this out as a private beta (hidden but can still be enabled on a per-site basis) first be problematic? It would let us test and gain confidence in the feature design + code before making it more widely available.

ghost/i18n/lib/i18n.js
```diff
@@ -85,18 +88,95 @@ function generateResources(locales, ns) {
...
+                        res = require(enPath);
+                    } else {
+                        res = {};
+                    }
+                } catch (enErr) {
+                    res = {};
+                }
+            } else {
+                res = {};
+            }
```
kevinansfield, line 151: Would it be helpful to log that the requested locale wasn't found or do you think that will be too noisy based on the number of untranslated themes?

cathysarisky: This runs when the locale is switched or theme is switched, IIRC. So it wouldn't be super noisy from a self-hosting perspective, but I don't know what the situation looks like for ops - perhaps debug would be a good choice here.

ghost/core/core/frontend/helpers/t.js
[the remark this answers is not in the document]
cathysarisky: I think it's occasionally useful (had trouble with an earlier version of this PR not running init at all the right times and only the right times), so I'll move it debug.

ghost/core/core/frontend/helpers/t.js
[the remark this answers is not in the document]
cathysarisky: Hmm. Honestly, it's been long enough that I've lost the context on this.  I'll switch over to settingsCache and see if anything breaks. :)

ghost/core/core/shared/labs.js
cathysarisky: No objections.  Self-hosters can enable it, right?  Would someone on Pro who wanted to test it just email support?

ghost/core/core/bridge.js
```diff
@@ -30,14 +31,33 @@ const messages = {
...
         });
+        
+        events.on('settings.labs.edited', () => {
+            const currentThemeTranslationState = labs.isSet('themeTranslation');
+            
+            if (currentThemeTranslationState !== this.previousThemeTranslationState) {
+                debug('themeTranslation flag changed from %s to %s', 
+                    this.previousThemeTranslationState, currentThemeTranslationState);
+                const locale = settingsCache.get('locale');
+                this.getActiveTheme().initI18n({locale});
```
kevinansfield, line 55: I think there may be a stale data condition here because this only re-init's the current i18n service meaning the other keeps hold of stale data. E.g.

1. Flag OFF → `themeI18n` initialized with Theme A
2. Flag ON → `themeI18next` initialized with Theme A (but `themeI18n` also keeps the Theme A data around)
3. Theme changes to Theme B → only `themeI18next` updated
4. Flag OFF → `themeI18n.t()` called, but it still has Theme A translations

Definitely edge-case though. Fix would be removing the conditional in `active.js` `init18n()` so that it always initialises both i18n libraries.

cathysarisky: I feel like I'm missing something here. In step 4, switching the flag off triggers this.getActiveTheme().initI18n({locale}), which I thought put themeI18n.t into the right state.

Happy to take the suggested fix, just wanting to understand a little better why it's needed.

kevinansfield (approved):
Nice work! 🙌 I took the liberty of pushing some `e2e-frontend` tests for theme translation as I was using those as an extra verification step.

There's one comment about a potential edge-case but I didn't make any changes as I wondered if you knew any reasons for/against the proposed fix - if you think its fixable then an e2e test to poke the problem and prove the fix would be good. This isn't a blocker however, it's very much an edge case.

Otherwise approved ✅ Thanks for working on this and addressing the previous feedback ❤️

ghost/core/core/bridge.js
kevinansfield: 🤦 You're absolutely right! 😄 Too late and over-thinking, the event means this is firing once we've already switched. Thanks for the quick clarification

PR #24455, opened 2025-07-20
Re-added social accounts back to explore payload

ErisDS:
closes https://linear.app/ghost/issue/PROD-1771/push-support-for-ghost-explore

- The payload for explore was simplified because most properties could be gotten from the public site endpoint
- But social accounts  are not available there yet
- Adding these back to explore, in lieu of adding them to the site endpoint, maybe when we add the other socials
- For now, this seems like a simpler change as it feels weird adding "twitter" to an endpont and documenting it, but we haven't entirely decided what to do about that because x is a weird property name

### DIFF
--- a/ghost/core/test/unit/server/services/explore-ping/ExplorePingService.test.js
+++ b/ghost/core/test/unit/server/services/explore-ping/ExplorePingService.test.js
@@ -23,8 +23,8 @@ describe('ExplorePingService', function () {
         settingsCacheStub.get.withArgs('active_theme').returns('alto');
         settingsCacheStub.get.withArgs('explore_ping').returns(true);
         settingsCacheStub.get.withArgs('explore_ping_growth').returns(false);
-        settingsCacheStub.get.withArgs('facebook').returns('https://facebook.com/example');
-        settingsCacheStub.get.withArgs('twitter').returns('https://twitter.com/example');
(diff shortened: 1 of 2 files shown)


ghost/core/test/unit/server/services/explore-ping/ExplorePingService.test.js
```diff
@@ -23,6 +23,8 @@ describe('ExplorePingService', function () {
         settingsCacheStub.get.withArgs('active_theme').returns('alto');
         settingsCacheStub.get.withArgs('explore_ping').returns(true);
         settingsCacheStub.get.withArgs('explore_ping_growth').returns(false);
+        settingsCacheStub.get.withArgs('facebook').returns('https://facebook.com/example');
+        settingsCacheStub.get.withArgs('twitter').returns('https://twitter.com/example');
```
sagzy: In settings table, we don't store the entire Twitter/Facebook URLs, only the usernames:
- Facebook: `example`
- Twitter: `@example`

So think we either want to:
1. Update the tests to reflect that we're returning usernames only, not full URLs
2. Or, transform the usernames into full URLs before calling the Explore service

ErisDS: Have changed the tests.

sagzy (approved):

Issue #24458, opened 2025-07-21
🐛 Search functionality in Settings sometimes returns empty nav items and content


niranjan-uma-shankar:
### Issue Summary

Some search terms lead to empty nav items and empty content. Here are a few examples:

#### Example 1

Search for "analytics" leads to an empty nav item under "General settings"
<img width="50%" alt="Image" src="https://github.com/user-attachments/assets/461e67b6-fc64-42d1-8a09-c738b47ec166" />

#### Example 2

Search for "payment" or "one time" leads to an empty nav item and content under "Email newsletter".

<img width="1210" height="901" alt="Image" src="https://github.com/user-attachments/assets/7648aadf-532a-4f81-ac9c-80779a8d313b" />

<img width="1038" height="678" alt="Image" src="https://github.com/user-attachments/assets/e5ed3a89-03ad-4a58-9edd-5c600ed2097d" />

#### Example 3

Search "membership" leads to an empty nav item under "General settings"

<img width="1073" height="362" alt="Image" src="https://github.com/user-attachments/assets/1eb222ac-09ee-4ed4-9d49-f4890f2818f1" />

#### Example 4

Search for "delete" and "site" lead to empty nav item under "Advanced".

<img width="1040" height="314" alt="Image" src="https://github.com/user-attachments/assets/58836a25-f2f5-46b4-9813-a09be704feb4" />

<img width="1081" height="797" alt="Image" src="https://github.com/user-attachments/assets/d4bf071b-324e-407f-abe3-964b9f5598d1" />

### Steps to Reproduce

1. Visit `/settings`
2. Try the search terms given in the "Issue summary" section and verify that you see empty nav items/content, as shown in the screenshots.

### Ghost Version

5.130.1

### Node.js Version

22.13.1

### How did you install Ghost?

MacOS Sequoia 15.5 & Ghost Pro

### Database type

MySQL 8

### Browser & OS version

Chrome

### Relevant log / error output

```shell

```

### Code of Conduct

- [x] I agree to be friendly and polite to people in this repository

niranjan-uma-shankar:
Fix: https://github.com/TryGhost/Ghost/pull/24461.

niranjan-uma-shankar:
> Our bot has automatically marked this issue as stale because there has not been any activity here in some time.
> The issue will be closed soon if there are no further updates, however we ask that you do not post comments to keep the issue open if you are not actively working on a PR.

[...]

PR #24299, opened 2025-07-09
Updated settings navigation structure for 6.0

peterzimon:
ref https://linear.app/ghost/issue/PROD-2209/settings-ui-changes

- Settings navigation for 6.0 was not reflecting the updated structure of the Admin

PR #24390, opened 2025-07-16
Removed `ui60` flag from Settings

peterzimon:
ref https://linear.app/ghost/issue/PROD-2317/remove-settings-related-ui60-flag

- Removed `ui60` flag for settings sidebar navigation and search, content area and related tests. This PR doesn't remove the flag itself as it's still used in the Admin.

PR #24459, opened 2025-07-21
Fix: "General settings" search keyword mismatch

niranjan-uma-shankar:
This PR fixes a bug where searching for "analytics" in the `/settings` route returns an empty result in the sidebar navigation.

g1

The issue is tracked here: https://github.com/TryGhost/Ghost/issues/24458.

As part of the upcoming 6.0 release, "Analytics" was moved under "General settings" in [PR #24299](https://github.com/TryGhost/Ghost/pull/24299). However, the `<SettingNavSection>` for "General settings" was still using  `generalSearchKeywords`, instead of the new `generalSearchKeywords5x` for 5.x installs.

This PR updates the visibility logic to use the correct keyword list, ensuring that searches like "analytics" correctly show the associated section. 

### Testing Instructions

1. Navigate to `/settings`
2. Search "analytics" and verify that you don't see any empty results; "General settings" should not show up in the results.

Screenshot 2025-07-21 at 4 07 19 PM

### DIFF
--- a/apps/admin-x-settings/src/components/Sidebar.tsx
+++ b/apps/admin-x-settings/src/components/Sidebar.tsx
@@ -201,6 +201,7 @@ const Sidebar: React.FC = () => {
                     {hasStripeEnabled && <NavItem icon='discount' keywords={growthSearchKeywords.offers} navid='offers' title="Offers" onClick={handleSectionClick} />}
                 </SettingNavSection>
 
+<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


niranjan-uma-shankar:
With the recent merge of https://github.com/TryGhost/Ghost/pull/24390, this PR is no longer necessary as the bug described gets fixed. Closing this PR.

PR #24460, opened 2025-07-21
🎨 Added support for "nql" to the #match helper

ramrami:
When passing an array or an object to the match helper and no operator, we now accept an NQL expression as value.
This allows more flexibility in the questions asked about the data while using a familiar query syntax.
Additionally, when the data is an array, we add `@matches` to the current context, containing the list of matched items.

**Examples:**
```hbs
{{match tags "slug:hash-related"}}

{{match post "tags:hash-related"}}

{{match post "authors.name:~'John'"}}

{{match post "authors:john+primary_tag:news"}}

{{match post "authors:[john,jane]"}}

{{match @site "comments_enabled:true+comments_access:'all'"}}

{{match @custom "list_layout:['Grid','Masonry']"}}
```

**Examples with `@matches`:**
```hbs
{{#match tags "slug:~'2025'"}}
    {{#foreach @matches visibility="all"}}
        <div>{{name}}</div>
    {{/foreach}}
{{/match}}

{{#match posts "featured:true"}}
    {{#foreach @matches limit="2"}}
        <div>{{title}}</div>
    {{/foreach}}
{{/match}}

```

**Breaking change**:
There is a tiny chance of breaking someone's code: matching an array or an object against a string used to always returns 'false', now if the string is a valid NQL expression and the object/array contains the queried data, it will return "true".

**Comparison with `#has`:**

```hbs
{{!-- post has a particular tag --}}
{{#has tag="photo"}}{{/has}}
{{#match . "tags.name:photo"}}{{/match}}

{{!-- or --}}
{{#has slug="welcome" tag="photo, video, audio" author="Joanna Bloggs"}}{{/has}}
{{#match . "slug:welcome, tags.name:[photo, video, audio], authors.name:'Joanna Bloggs'"}}{{/match}}

{{!-- and --}}
{{#has tag="photo"}}
    {{#has tag="panorama"}}
    {{/has}}
{{/has}}
{{#match . "tags.name:photo+tags.name:panorama"}}{{/match}}

{{!-- counting (currently not supported by nql) --}}
{{#has tag="count:>1"}}{{/has}}
{{#match tags.length ">" 1}}{{/match}}

{{!-- Any or all --}}
{{!-- 
Limitations: 
- we can't query multiple contexts simultaneously, 
- and as far as I know we can't look for non-empty strings using nql, only non-null 
--}}
{{#has any="@site.facebook, @site.twitter"}}{{/has}}
{{#match @site "facebook:-null, twitter:-null"}}{{/match}}

```

### DIFF
--- a/ghost/core/core/frontend/helpers/match.js
+++ b/ghost/core/core/frontend/helpers/match.js
@@ -1,9 +1,11 @@
-const {SafeString} = require('../services/handlebars');
+const {hbs, SafeString} = require('../services/handlebars');
 
 const logging = require('@tryghost/logging');
 const tpl = require('@tryghost/tpl');
+const nql = require('@tryghost/nql');
 
 const _ = require('lodash');
+const {handlebars: {createFrame}} = hbs;
 
 const messages = {
     invalidAttribute: 'Invalid or no attribute given to match helper'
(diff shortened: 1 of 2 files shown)


ErisDS:
This change looks _amazing_ 🎉 
I need to sit and get my head around exactly what it's doing and review it fully, but I don't anticipate requesting changes from a brief look.

My only mild concern is performance, but it shouldn't be an issue 🤔 

If it's not, we could add this to foreach as well 💥

PR #24462, opened 2025-07-21
Change authentication for self-service migrations to send API key instead of JWT

PaulAdamDavis:
ref https://linear.app/ghost/issue/CON-372/change-ghost-self-service-migrations-app-auth

- The self-service migrations app is currently authenticated with a JWT, but the expire time for these is capped at 5 minutes in the [Admin API key auth service](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/services/auth/api-key/admin.js)
- I wasn't aware of this in #24119, making that change ineffective
- The API key used to make the JWT is already available to privileged users of Ghost Admin
- This commit changes the migrate service to send the API key rather than a JWT (and handling that in the self-service app) yields the same result, allowing migrations more time to process, improving the end result
- It also removes the now-unused `jose` dependency

### DIFF
--- a/ghost/admin/app/services/migrate.js
+++ b/ghost/admin/app/services/migrate.js
@@ -1,6 +1,5 @@
 import Service, {inject as service} from '@ember/service';
 import config from 'ghost-admin/config/environment';
-import {SignJWT} from 'jose';
 import {tracked} from '@glimmer/tracking';
 
 export default class MigrateService extends Service {
--- a/ghost/admin/app/services/migrate.js
+++ b/ghost/admin/app/services/migrate.js
@@ -77,13 +77,13 @@ export default class MigrateService extends Service {
         return url;
     }
 
-    // Sends a route update to a child route in the BMA, because we can't control
+    // Sends a route update to a child route in the migrate app, because we can't control
     // navigating to it otherwise
     sendRouteUpdate(route) {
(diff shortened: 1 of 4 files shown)


troyciesco (approved):

allouis (commented):
Calling `postMessage` with `'*'` as the origin makes me feel a little uncomfortable, especially when sending an API key

I see we're passing `'*'` when updating the route info, but I can't see where this is happening when sending the key

We should restrict the `postMessage` call to the origin of the migrate service, this ensures that data can only ever be sent there, and not to a different site

There's a small note about this in the MDN docs here https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#targetorigin

PaulAdamDavis:
@allouis Good call. I have updated the `sendRouteUpdate` method.

The payload that contains the API key is sent to the iframe here https://github.com/TryGhost/Ghost/blob/main/ghost/admin/app/components/gh-migrate-iframe.js#L65 and is untouched in this PR. The origin is `migrateUrl` in the service.

These are the only 2 places where data is sent. On the receiving end, we check the origin here https://github.com/TryGhost/Ghost/blob/main/ghost/admin/app/components/gh-migrate-iframe.js#L30-L31

allouis (approved):

PR #24463, opened 2025-07-21
Fixed pages showing in top posts when no views present

9larsons:
ref https://linear.app/ghost/issue/PROD-2299/
- filtered by post type = 'post' in fallback for `top-posts-views` endpoint

When we received no data from Tinybird (or insufficient to fill 5 entries), we had fallback behavior that still pulled from the posts table to make sure we always returned up to 5 entries. This fallback behavior did not use the post.type filter like the Tinybird query, so we ended up with an edge case where we could display Pages in the Top Posts component.


9larsons:
@cmraible Would you mind taking a review pass? I know I could DRY up the tests. Imo the whole stats endpoints need a redesign (which I'm going to write up a pitch for) and I'm not particularly worried about them being clean vs. easily digestible changes. Open to changing that though.

cmraible:
LGTM pending the cutover from the `6.x` branch to `main` 👍

cmraible (changes requested):
No actual changes to request, just blocking this until `6.x` is merged to `main`. Overall LGTM, will approve as soon as the code freeze is lifted.

cmraible (approved):
LGTM!

PR #24469, opened 2025-07-21
Updated new Post Analytics to be the default for the posts/analytics route

9larsons:
ref https://linear.app/ghost/issue/PROD-2326/
- removed Ember analytics by removing `posts.analytics` route
- updated Ember references to `posts.analytics` and pointed to `posts.posts-x` (this prevents Ember crashes)
- updated Admin router to use `posts.posts-x` for `/posts/analytics/` (removed `beta/`)
- updated Posts and Stats apps to remove `/beta/`; moved to constants held by admin-x-framework

This set of changes is moving the new `posts` app to be the default for Post Analytics. This was previously at the `/posts/analytics/beta/:id` path, and is now at `/posts/analytics/:id`, replacing the Ember view with the new React app. The `/debug` and `/mentions` paths are still available and will load their respective Ember views, so it's not as seamless as it once was, though we intend to move more and more into the `posts` app.

We had a lot of hard-coded paths in the `posts` and `stats` app, which have been moved to constants within `admin-x-framework` to make future adjustments easier.

### DIFF
--- a/apps/admin-x-framework/src/utils/app-routes.ts
+++ b/apps/admin-x-framework/src/utils/app-routes.ts
@@ -1,0 +1,40 @@
+// Posts analytics routes
+// These work for both internal navigation (within Posts app) and cross-app navigation
+export const POSTS_ANALYTICS_ROUTES = {
+    OVERVIEW: (postId: string) => `/analytics/${postId}`,
+    WEB: (postId: string) => `/analytics/${postId}/web`,
+    NEWSLETTER: (postId: string) => `/analytics/${postId}/newsletter`,
+    GROWTH: (postId: string) => `/analytics/${postId}/growth`
+} as const;
+
+// Stats routes
+export const STATS_ROUTES = {
+    OVERVIEW: '/analytics',
+    WEB: '/analytics/web/',
+    LOCATIONS: '/analytics/locations/',
+    GROWTH: '/analytics/growth/',
+    NEWSLETTERS: '/analytics/newsletters/'
+} as const;
+
+// Helper to build full cross-app navigation paths
+// When navigating from Stats to Posts, prepend /posts
+export const buildCrossAppPostsRoute = (route: string) => `/posts${route}`;
+
+// Test routes - These are the hash-based routes used in Ember admin for tests
+export const TEST_ROUTES = {
+    // Stats test routes
+    STATS: {
+        OVERVIEW: '/ghost/#/analytics',
+        WEB: '/ghost/#/analytics/web',
+        LOCATIONS: '/ghost/#/analytics/locations',
+        GROWTH: '/ghost/#/analytics/growth',
+        NEWSLETTERS: '/ghost/#/analytics/newsletters'
+    },
+    // Posts analytics test routes  
+    POSTS_ANALYTICS: {
+        OVERVIEW: (postId: string) => `/ghost/#/posts/analytics/${postId}`,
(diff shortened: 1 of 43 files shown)


apps/posts/src/routes.tsx
```diff
@@ -23,7 +23,7 @@ export const routes: RouteObject[] = [
             {
 
                 // Post Analytics
-                path: 'analytics/beta/:postId',
+                path: 'analytics/:postId',
```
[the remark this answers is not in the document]
9larsons: We will handle this is a separate commit.

apps/admin-x-framework/src/utils/app-routes.ts
```diff
@@ -0,0 +1,43 @@
+// Posts analytics routes
+// These work for both internal navigation (within Posts app) and cross-app navigation
+export const POSTS_ANALYTICS_ROUTES = {
```
9larsons, line 3: I added these to a framework constants file because we reference the apps from each other, but didn't want to create an import dependency between each of them. Ideally anything shared goes to framework.

apps/stats/src/views/Stats/Newsletters/Newsletters.tsx
```diff
@@ -73,7 +73,7 @@ const NewsletterTableRows: React.FC<{
                                 <div className='group/link inline-flex items-center gap-2'>
                                     {post.post_id ?
                                         <Button className='h-auto whitespace-normal p-0 text-left hover:!underline' title="View post analytics" variant='link' onClick={() => {
-                                            navigate(`/posts/analytics/beta/${post.post_id}/`, {crossApp: true});
+                                            navigate(buildCrossAppPostsRoute(POSTS_ANALYTICS_ROUTES.OVERVIEW(post.post_id)), {crossApp: true});
                                         }}>
```
9larsons, line 77: This might feel a little awkward, but I didn't have a better idea on how to handle prefixing with `posts` given how the app routers are set up to use a base route.

apps/admin-x-framework/src/utils/app-routes.ts
```diff
@@ -0,0 +1,43 @@
...
+
+// Helper to build full cross-app navigation paths
+// When navigating from Stats to Posts, prepend /posts
+export const buildCrossAppPostsRoute = (route: string) => {
+    const normalizedRoute = route.startsWith('/') ? route : `/${route}`;
+    return `/posts${normalizedRoute}`;
+};
+
+// Test routes - These are the hash-based routes used in Ember admin for tests
+export const TEST_ROUTES = {
```
9larsons, line 27: I added these full routes for the test `Page` objects. I am open to removing this if it feels like stretching this PR out a bit too much. It felt a bit better to me than updating the routes with `/beta` in them.

cmraible (changes requested):
**I think this PR combines two good changes that would be easier to review and validate separately:**

1. Route migration: from the `/beta` routes to the now default paths
2. Route centralization: extracting the hardcoded paths to shared constants in framework

Both improvements make sense, but combining them this way makes it harder to verify that the route behavior changes work as expected, and validate that the constant extraction didn't introduce errors

Could we split this into:
- First PR: Just the beta -> default route behavior changes. 
- Follow-up PR: route constants refactor

This would make each change easier to review thoroughly (and more quickly) and give both the attention they deserve.

PR #24471, opened 2025-07-22
Added demo data override for posts and stats apps

cmraible:
ref https://linear.app/ghost/issue/PROD-2301/generate-fake-data-for-the-marketing-site-for-screenshots

❌ Not planning to merge this, this is just some hacky stuff we can deploy to staging on a test site.

The goal here is to enable a "demo data mode" for the posts and stats apps. There are fundamentally two different approaches to doing this, as I see it:
- Seeding a site's actual databases (MySQL + Tinybird) with generated raw data
- Intercepting the API requests and returning fake aggregated data

This PR opts for the latter approach, as it offers a few benefits over the former (albeit with tradeoffs):
- Data freshness — we can generate the data on demand, so it will always be fresh
- Convenience — the fake data mode can be enabled/disabled all in the browser, without any ad-hoc scripting or knowledge of code.
- Simplicity — generating realistic looking aggregated data (as returned by the API endpoints) feels like a lighter lift than generating all the data and relationships required to return the same aggregated data

How it works:
- The "fake data" mode is enabled by setting the `ghost-fake-data` key in localStorage to `'true'`. There's no UI toggle (yet) — for now this can be accomplished in the browser console with `localStorage.setItem('ghost-fake-data', 'true');`
- When "fake data" is enabled, we intercept API requests to Ghost's Admin API and Tinybird's endpoints at the framework level, and inject fake data into the responses. Doing this at the framework level avoids making any code changes to the apps themselves, makes the same pattern available in all the other admin-x-* apps, and just generally feels cleaner than overriding the responses in ~15 custom hooks individually
- Each endpoint that we choose to intercept uses a generator function for the response data, so we can dynamically generate the fake data with timestamps that make sense, some randomness, etc.

This isn't fully implemented for all required endpoints yet, but I wanted to share for early feedback on the approach before going too much further with this.

### DIFF
--- a/apps/admin-x-framework/src/hooks/useFakeData.ts
+++ b/apps/admin-x-framework/src/hooks/useFakeData.ts
@@ -1,0 +1,42 @@
+import {useMemo, useState} from 'react';
+import {createFakeDataProvider} from '../utils/fake-data';
+import type {FakeDataConfig} from '../providers/FrameworkProvider';
+
+/**
+ * Reads the fake data preference from localStorage with SSR safety
+ */
+function getFakeDataEnabled(): boolean {
+    try {
+        return localStorage.getItem('ghost-fake-data') === 'true';
+    } catch {
+        // Fallback for SSR or when localStorage is unavailable
+        return false;
+    }
+}
+
+/**
+ * Creates a complete fake data configuration object
+ */
+function createFakeDataConfig(enabled: boolean): FakeDataConfig | undefined {
+    if (!enabled) {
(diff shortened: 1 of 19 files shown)


9larsons:
I like the flow & approach 👍. I think my only comment/concern is that we're stubbing with mock data in two separate places instead of one (fetchApi and useTinybirdQuery), which is more of a 'problem' with our querying two sources than anything here.

9larsons (approved):

PR #24473, opened 2025-07-22
Updated post list analytics metric conditions

peterzimon:
ref https://linear.app/ghost/issue/ENG-2448/update-post-list-metric-conditions

Currently post list metrics (web visitors, no. of members etc.) are displayed with conditions that are not appropriate and inconsistent with the new Traffic Analytics.

Here are the issues that should be fixed:

**Web visitors**

- Should be shown if web analytics is turned on, regardless if there's data
- Current condition: it's only shown if there is visitor data

**Member data**

- Should be shown if member tracking is turned on, regardless if there's data
- Current condition: it's show only if there's member related data

CleanShot 2025-07-22 at 11 54 07@2x


9larsons (approved):
Definitely a fan of simplifying and making this more consistent & intuitive.

PR #24474, opened 2025-07-22
Updated hover style for links in Settings CTAs

peterzimon:
ref https://linear.app/ghost/issue/PROD-2303/update-hover-style-for-upgrade-now-link-in-settings-cta-text

- Links is Analytics setting related CTAs were misformatted: on the upgrade link the cursor was not `pointer`; on the self-hoster CTAs there were unnecessary underlines.


9larsons (approved):
Much nicer now 😄

PR #24476, opened 2025-07-22
Refined loading indicator for post stats overview

peterzimon:
ref https://linear.app/ghost/issue/PROD-2174/investigate-web-analytics-flash-on-email-only-posts

- The current loading indicator setup of the post analytics overview shows all sections with separate, dedicated loading indicators. However it's only possible to determine which sections should be shown only once the post data is loaded. For email-only posts this could result in showing the web section for a splitsecond and then immediately removing it, causing an unwanted flash on the UI. To overcome this problem, this PR adds a loading indicator that's shown on the whole page until the post data is loaded, resulting in showing only relevant sections without the flash.


9larsons (approved):
Simple & effective 👍

PR #24479, opened 2025-07-22
Updated new Post Analytics to be the default for posts/analytics route

9larsons:
ref https://linear.app/ghost/issue/PROD-2326/
- removed Ember analytics by removing posts.analytics route
- updated Ember references to posts.analytics and pointed to posts.posts-x (this prevents Ember crashes)
- updated Admin router to use posts.posts-x for /posts/analytics/ (removed beta/)
- updated Posts and Stats apps to remove /beta/

This set of changes is moving the new posts app to be the default for Post Analytics. This was previously at the /posts/analytics/beta/:id path, and is now at /posts/analytics/:id, replacing the Ember view with the new React app. The /debug and /mentions paths are still available and will load their respective Ember views, so it's not as seamless as it once was, though we intend to move more and more into the posts app.

### DIFF
--- a/ghost/admin/app/routes/lexical-editor.js
+++ b/ghost/admin/app/routes/lexical-editor.js
@@ -17,7 +17,7 @@ export default AuthenticatedRoute.extend({
     },
 
     setupController(controller, model, transition) {
-        if ((transition.from?.name === 'posts.analytics' || transition.from?.name?.startsWith('posts-x')) && transition.to?.name !== 'lexical-editor.new') {
+        if ((transition.from?.name === 'posts-x' || transition.from?.name?.startsWith('posts-x')) && transition.to?.name !== 'lexical-editor.new') {
             // Extract the analytics path from window.location.href to preserve the exact tab
             let currentUrl = window.location.href;
             // Convert editor URL back to analytics URL and extract just the hash portion
--- a/ghost/admin/app/routes/lexical-editor.js
+++ b/ghost/admin/app/routes/lexical-editor.js
@@ -17,7 +17,7 @@ export default AuthenticatedRoute.extend({
     },
 
     setupController(controller, model, transition) {
-        if ((transition.from?.name === 'posts-x' || transition.from?.name?.startsWith('posts-x')) && transition.to?.name !== 'lexical-editor.new') {
+        if (transition.from?.name?.startsWith('posts-x') && transition.to?.name !== 'lexical-editor.new') {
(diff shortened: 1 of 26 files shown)


cmraible:
@9larsons This is so much easier to review, thanks for splitting the changes up! Code looks good to me, just doing a quick QA pass. 

I'm getting a 500 error at the `/post/:post_id/mentions` route locally — are you seeing that too? Might just be something with my local env, so I'm deploying this to staging to test there.

Screenshot 2025-07-22 at 11 13 48@2x

ghost/admin/app/routes/lexical-editor.js
```diff
@@ -17,7 +17,7 @@ export default AuthenticatedRoute.extend({
     },
 
     setupController(controller, model, transition) {
-        if ((transition.from?.name === 'posts.analytics' || transition.from?.name?.startsWith('posts-x')) && transition.to?.name !== 'lexical-editor.new') {
+        if ((transition.from?.name === 'posts-x' || transition.from?.name?.startsWith('posts-x')) && transition.to?.name !== 'lexical-editor.new') {
```
cmraible: Non blocking: we should be able to remove this, as the `startsWith` should suffice.

cmraible (changes requested):
Only blocker is that the `/posts/:post_id/mentions` route is broken for me — everything else looks good, nice work!

ghost/admin/app/routes/lexical-editor.js
9larsons: Good catch, changed that.

9larsons:
> @9larsons This is so much easier to review, thanks for splitting the changes up! Code looks good to me, just doing a quick QA pass.
> 
> I'm getting a 500 error at the `/post/:post_id/mentions` route locally — are you seeing that too? Might just be something with my local env, so I'm deploying this to staging to test there.
> 
> Screenshot 2025-07-22 at 11 13 48@2x

Yep, I see that too.. for whatever reason this change wasn't necessary/fouling up tests in the previous PR. I think it was just a miss there and then the test failure is still stumping me a bit (as it why it didn't fail before but does now). In general, Ember routing is a bit messy w/r to services and auth, so I've added a check for the session.user and redirected to signin if that's not present. We do this elsewhere, and it's the best way to prevent services instantiating with empty data (which shouldn't happen, but is a gap in our auth implementation).

cmraible (approved):
Nice one!

PR #24481, opened 2025-07-22
Updated activitypub client cache purging strategy

mike182uk:
ref https://linear.app/ghost/issue/PROD-2329

Updated the activitypub client cache purging strategy so that we are always purging the latest major version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1144,7 +1144,7 @@ jobs:
         include:
           - package_name: '@tryghost/admin-x-activitypub'
             package_path: 'apps/admin-x-activitypub'
-            cdn_paths: 'https://cdn.jsdelivr.net/ghost/admin-x-activitypub@1/dist/admin-x-activitypub.js'
+            cdn_paths: 'https://cdn.jsdelivr.net/ghost/admin-x-activitypub@CURRENT_MAJOR/dist/admin-x-activitypub.js'
           - package_name: '@tryghost/portal'
             package_path: 'apps/portal'
             cdn_paths: 'https://cdn.jsdelivr.net/ghost/portal@~CURRENT_MINOR/umd/portal.min.js'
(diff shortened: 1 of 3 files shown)


sagzy (approved):

.github/workflows/ci.yml
```diff
@@ -1144,7 +1144,7 @@ jobs:
         include:
           - package_name: '@tryghost/admin-x-activitypub'
             package_path: 'apps/admin-x-activitypub'
-            cdn_paths: 'https://cdn.jsdelivr.net/ghost/admin-x-activitypub@0/dist/admin-x-activitypub.js'
+            cdn_paths: 'https://cdn.jsdelivr.net/ghost/admin-x-activitypub@1/dist/admin-x-activitypub.js'
```
allouis: Nice - maybe it would be good if we can just clear the cache of everything so we don't have to remember to bump this - but not sure if we can

ghost/admin/lib/asset-delivery/index.js
```diff
@@ -44,7 +44,7 @@ module.exports = {
                     console.log(`Asset-Delivery: ${key} = ${value}`);
                 }
 
-                this.packageConfig[`adminXActivitypubCustomUrl`] = 'https://cdn.jsdelivr.net/ghost/admin-x-activitypub@0/dist/admin-x-activitypub.js';
+                this.packageConfig[`adminXActivitypubCustomUrl`] = 'https://cdn.jsdelivr.net/ghost/admin-x-activitypub@1/dist/admin-x-activitypub.js';
```
allouis: This doesn't actually do anything because we have the `adminXActivitypubRemoteConfigUrl` set which takes precedence.

I think we should probably get rid of this line to avoid confusion

.github/workflows/ci.yml
mike182uk: Yeh we won't be able to clear the cache for everything 🤔

Would just clearing `@latest` work for us here?

ghost/admin/lib/asset-delivery/index.js
mike182uk: Nice one, removed in https://github.com/TryGhost/Ghost/pull/24481/commits/531833ab173b31efea2dc94f9a0f93568f7ed97a

.github/workflows/ci.yml
allouis: Hmm, I think that just clears the cache on the `@latest` url :/ I guess we'll have to update this manually

mike182uk: But that should be enough right? if we go from 1 to 2, we don't need to be purging 1 anymore as all changes going forward will be on 2?

mike182uk: Discussed with @allouis on a call - We have added a new `CURRENT_MAJOR` placeholder that works in the same way as the `CURRENT_MINOR` placeholder. This will allow us to always reference the latest MAJOR version to clear the cache for (which is what we want). My suggestion for using `@latest` would be the literal latest version, not just the major

PR #24482, opened 2025-07-22
Wired up member count in Explore

sagzy:
ref https://linear.app/ghost/issue/PROD-1771

- if growth data sharing is enabled, show a preview with the site's members count
- member count above 1000 are rendered are abbreviated, e.g. `1.2k` for `1210` members

CleanShot 2025-07-22 at 19 23 19@2x

### DIFF
--- a/apps/admin-x-settings/package.json
+++ b/apps/admin-x-settings/package.json
@@ -49,6 +49,7 @@
     "@testing-library/react": "14.3.1",
     "@tryghost/admin-x-design-system": "0.0.0",
     "@tryghost/admin-x-framework": "0.0.0",
+    "@tryghost/shade": "0.0.0",
     "@tryghost/custom-fonts": "1.0.2",
     "@types/react": "18.3.23",
     "@types/react-dom": "18.3.7",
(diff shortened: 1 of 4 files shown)


apps/admin-x-settings/package.json
```diff
@@ -49,6 +49,7 @@
     "@testing-library/react": "14.3.1",
     "@tryghost/admin-x-design-system": "0.0.0",
     "@tryghost/admin-x-framework": "0.0.0",
+    "@tryghost/shade": "0.0.0",
```
sagzy, line 52: @peterzimon Adding shade as dependency to Admin Settings, as we eventually want to replace the old design system here too. Lmk if you have any concerns here!

sagzy: The only use case for now is the `abbreviateNumber` helper, that we can get from elsewhere too - just thought it's good to introduce shade into settings anyways

peterzimon: Sounds good 👍

peterzimon (changes requested):
@sagzy can you please add a condition to show "1 member" instead of "1 members" if that's the case:

CleanShot 2025-07-22 at 19 59 10@2x

peterzimon (approved):

PR #24488, opened 2025-07-23
Created data factory framework for E2E test package

9larsons:
ref https://linear.app/ghost/issue/PROD-2327/

This PR introduces a new plugin-based test data factory for our E2E test suite. I've started with demonstrating two very basic implementations:
- Ghost plugin with Post factory
- Tinybird plugin with PageHit factory

The core concept is to have a plugin-based architecture that allows a custom mix of factories that the test writer can choose without needing to boot absolutely everything, while also providing a simple interface for test writing. We'll demonstrate the use with some simplistic tests for the Analytics apps, which is why I've started with Posts and Page Hits.

The goal for all of this is simple tests. See the `examples.test.ts` file, where use is as simple as:
```
test('create a post with views', async ({ghost,tinybird}) => {
    const post = await ghost.createPublishedPost();
    const hits = await tinybird.createPageHitsForPost(post, 100, {member_status: 'free'});   
    // navigate and assert
});
```
And the framework cleans up the created data after each test.

Major decision points:
- [ ] Plugin-based design; this helps isolate dependencies and allows for varied compositions (e.g. testing Analytics should look different from testing ActivityPub)
- [ ] Factories are simple and only handle `create()` and cleanup tracking; it's an open question if they should handle convenience functions or if those should live in the plugin, which has cross-factory knowledge (allowing `createPostWithTags`). This allows them to be (safely) tightly coupled to their own schema/type.
- [ ] `Factory`, `BasePlugin` classes help define behavior/structure
- [ ] `DataFactory` is the coordinator for the various plugins
- [ ] `test-fixtures` demonstrates how to easily integrate with Playwright/test runners to manage instantiation and cleanup

We have some serious concerns related to parallelization and shared Ghost dbs, which is not designed to be multi-site (multi-tenant) like Tinybird, which we can look in to later. Through use of `test.concurrent` we can be clever in our test design to allow for tests to run successfully both in parallel and in isolation.

### DIFF
--- a/e2e/data-factory/base-factory.ts
+++ b/e2e/data-factory/base-factory.ts
@@ -1,15 +1,28 @@
+import type {PersistenceAdapter} from './persistence/types';
+
 export abstract class Factory<TOptions = unknown, TResult = unknown> {
     abstract name: string;
+    abstract entityType: string; // Used for persistence routing
+    
+    private persistence?: PersistenceAdapter;
+    private trackingEnabled = true;
+    protected createdEntities = new Map<string, TResult>();
     
     /**
      * Set up any necessary resources (database connections, etc.)
+     * Override this method if your factory needs initialization
      */
-    abstract setup(): Promise<void>;
+    async setup(): Promise<void> {
+        // Default implementation - no setup needed
+    }
     
     /**
-     * Clean up resources and created data
+     * Factory-specific cleanup (override if needed)
+     * Called after entities are cleaned up
      */
-    abstract destroy(): Promise<void>;
+    async destroy(): Promise<void> {
+        // Default implementation - no cleanup needed
+    }
     
     /**
      * Build an object without persisting it
@@ -17,7 +30,73 @@ export abstract class Factory<TOptions = unknown, TResult = unknown> {
     abstract build(options?: TOptions): TResult | Promise<TResult>;
     
     /**
+     * Inject persistence adapter
+     */
+    setPersistence(adapter: PersistenceAdapter): void {
+        this.persistence = adapter;
+    }
+    
+    /**
+     * Control entity tracking for cleanup
+     */
+    setTracking(enabled: boolean): void {
+        this.trackingEnabled = enabled;
+    }
+    
+    /**
      * Build and persist an object
      */
-    abstract create(options?: TOptions): Promise<TResult>;
+    async create(options?: TOptions): Promise<TResult> {
+        if (!this.persistence) {
+            throw new Error(`No persistence adapter configured for ${this.name} factory`);
+        }
+        
+        // Build the entity
+        const entity = await Promise.resolve(this.build(options));
+        
+        // Persist using the adapter
+        const persisted = await this.persistence.insert(this.entityType, entity);
+        
+        // Track for cleanup if enabled
+        if (this.trackingEnabled) {
+            const id = this.extractId(persisted);
+            if (id) {
+                this.createdEntities.set(id, persisted);
+            }
+        }
+        
+        return persisted;
+    }
+    
+    /**
+     * Clean up all created entities
+     */
+    async cleanup(): Promise<void> {
+        if (!this.persistence || this.createdEntities.size === 0) {
+            return;
+        }
+        
+        const ids = Array.from(this.createdEntities.keys());
+        await this.persistence.deleteMany(this.entityType, ids);
+        this.createdEntities.clear();
+        
+        // Call factory-specific cleanup
+        await this.destroy();
+    }
+    
+    /**
+     * Extract ID from entity (override if needed)
+     */
+    protected extractId(entity: TResult): string | undefined {
+        const e = entity as Record<string, unknown>;
+        const id = e.id || e.uuid;
+        return typeof id === 'string' ? id : undefined;
+    }
+    
+    /**
+     * Get all created entities
+     */
+    getCreatedEntities(): TResult[] {
+        return Array.from(this.createdEntities.values());
+    }
 }
--- a/e2e/data-factory/base-factory.ts
+++ b/e2e/data-factory/base-factory.ts
@@ -1,0 +1,27 @@
+export abstract class Factory {
+    abstract name: string;
+    
+    abstract setup(): Promise<void>;
+    abstract destroy(): Promise<void>;
+    
+    protected generateId(): string {
+        return Date.now().toString(16) + Math.random().toString(16).substring(2, 10);
+    }
+    
+    protected generateUuid(): string {
+        return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
+            const r = Math.random() * 16 | 0;
+            const v = c === 'x' ? r : (r & 0x3 | 0x8);
+            return v.toString(16);
+        });
+    }
+    
+    protected generateSlug(text: string): string {
+        return text.toLowerCase()
+            .trim()
+            .replace(/[^a-z0-9\s-]/g, '')
+            .replace(/\s+/g, '-')
+            .replace(/-+/g, '-')
+            .replace(/^-+|-+$/g, '');
+    }
+}
--- a/e2e/data-factory/base-factory.ts
+++ b/e2e/data-factory/base-factory.ts
@@ -1,3 +1,5 @@
+import {v4 as uuid} from 'uuid';
+
 export abstract class Factory {
     abstract name: string;
     
@@ -9,11 +11,7 @@ export abstract class Factory {
     }
     
     protected generateUuid(): string {
-        return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
-            const r = Math.random() * 16 | 0;
-            const v = c === 'x' ? r : (r & 0x3 | 0x8);
-            return v.toString(16);
-        });
+        return uuid();
     }
     
     protected generateSlug(text: string): string {
--- a/e2e/data-factory/base-factory.ts
+++ b/e2e/data-factory/base-factory.ts
@@ -1,25 +1,23 @@
-import {v4 as uuid} from 'uuid';
-
-export abstract class Factory {
+export abstract class Factory<TOptions = unknown, TResult = unknown> {
     abstract name: string;
     
+    /**
+     * Set up any necessary resources (database connections, etc.)
+     */
     abstract setup(): Promise<void>;
+    
+    /**
+     * Clean up resources and created data
+     */
     abstract destroy(): Promise<void>;
     
-    protected generateId(): string {
-        return Date.now().toString(16) + Math.random().toString(16).substring(2, 10);
-    }
+    /**
+     * Build an object without persisting it
+     */
+    abstract build(options?: TOptions): TResult | Promise<TResult>;
     
-    protected generateUuid(): string {
-        return uuid();
-    }
-    
-    protected generateSlug(text: string): string {
-        return text.toLowerCase()
-            .trim()
-            .replace(/[^a-z0-9\s-]/g, '')
-            .replace(/\s+/g, '-')
-            .replace(/-+/g, '-')
-            .replace(/^-+|-+$/g, '');
-    }
+    /**
+     * Build and persist an object
+     */
+    abstract create(options?: TOptions): Promise<TResult>;
 }
--- a/e2e/data-factory/data-factory.ts
+++ b/e2e/data-factory/data-factory.ts
@@ -1,0 +1,98 @@
+import type {DataFactoryPlugin} from './plugins/base-plugin';
+import {GhostPlugin} from './plugins/ghost/ghost-plugin';
+import {TinybirdPlugin} from './plugins/tinybird/tinybird-plugin';
+import {getSiteUuid} from './plugins/ghost/database';
+
+export interface DataFactoryOptions {
+    plugins?: DataFactoryPlugin[];
+}
+
+/**
+ * Data factory that coordinates plugins for creating test data.
+ * Each test gets its own instance for proper isolation.
+ */
+export class DataFactory {
+    private plugins = new Map<string, DataFactoryPlugin>();
+    private isDestroyed = false;
+    
+    constructor(options: DataFactoryOptions = {}) {
+        const plugins = options.plugins ?? [
+            new GhostPlugin(),
+            new TinybirdPlugin()
+        ];
+        
+        // Register all plugins
+        plugins.forEach(plugin => this.register(plugin));
+    }
+    
+    register<T extends DataFactoryPlugin>(plugin: T): T {
+        this.plugins.set(plugin.name, plugin);
+        return plugin;
+    }
+    
+    getPlugin<T extends DataFactoryPlugin>(name: string): T | undefined {
+        return this.plugins.get(name) as T;
+    }
+    
+    async initialize(): Promise<void> {
+        // Setup all plugins
+        for (const plugin of this.plugins.values()) {
+            await plugin.setup();
+        }
+        
+        // Handle cross-plugin initialization
+        await this.initializeCrossPluginDependencies();
+    }
+    
+    /**
+     * Initialize Tinybird with Ghost's site UUID if both plugins are present
+     */
+    private async initializeCrossPluginDependencies(): Promise<void> {
+        const ghost = this.getPlugin<GhostPlugin>('ghost');
+        const tinybird = this.getPlugin<TinybirdPlugin>('tinybird');
+        
+        // Initialize Tinybird with Ghost's site UUID if both plugins are present
+        if (ghost && tinybird && !tinybird.isInitialized()) {
+            const siteUuid = await getSiteUuid(ghost.getDatabase());
+            if (siteUuid) {
+                await tinybird.initializeWithSiteUuid(siteUuid);
+            }
+        }
+    }
+    
+    // *
+    // Convenience getters for tests, allow for direct access to plugins
+    //  e.g. test({ghost}) => ghost.createPost()
+    //   vs. test({factory}) => factory.ghost.createPost()
+    // *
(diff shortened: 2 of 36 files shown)


e2e/data-factory/tests/example.test.ts
```diff
@@ -0,0 +1,71 @@
+import {test, expect} from '../test-fixtures';
+
+test.describe('Data Factory Examples', () => {
+    test('create a simple published post', async ({factory}) => {
```
cmraible, line 4: praise: passing the factory directly to the tests via the Playwright fixtures is really nice ❤️ I love that it's just a single import, and you've got all your factories at your disposal.

e2e/data-factory/plugins/ghost/posts/post-factory.ts
```diff
@@ -0,0 +1,102 @@
...
+        
+        // Merge with user options, handling special cases
+        const post: PostResult = {
+            ...defaults,
+            ...options,
+            // Handle published_at logic - if status is published but no published_at is set, use current time
+            published_at: options.status === 'published' && !options.published_at ? now : (options.published_at || defaults.published_at)
+        } as PostResult;
+        
+        await this.db('posts').insert(post);
```
cmraible: suggestion: one thing I feel quite strongly about is that we should decouple the _creation_ of the object from the _persistence_ of the object. In other popular factory packages like Factory Bot, each factory has a `build()` method, which generates an instance of the object in memory, and a `create()` method, which calls `build()` and then saves it.

An example of where this could be useful:
- When testing an API, I might want to create a factory for the JSON payload that the `POST /admin/api/posts` endpoint expects to create a new post. I don't need to save that object to a DB or anything, I just want a convenient way to generate it in memory. In this case, I could just call `build()` instead of `create()`.
- When testing Stripe webhooks, I might want to create a factory for a `checkout.session.completed` webhook payload from Stripe. I wouldn't want to save that anywhere, but I'd want to use a factory so I don't have to specify all the default fields of a giant JSON object

e2e/data-factory/plugins/ghost/ghost-plugin.ts
```diff
@@ -0,0 +1,128 @@
...
+    private db: Knex;
+    private postFactory: PostFactory;
+    
+    constructor(options: GhostPluginOptions = {}) {
+        super();
+        // All Ghost factories share this database connection
+        this.db = options.database ?? createDatabase();
+        
+        // Initialize factories with shared database
+        this.postFactory = new PostFactory(this.db);
```
cmraible: suggestion: I think the dependency is backwards here. I would expect the Plugin to be injected into the Factory, and provide CRUD methods like `insert()`, `delete()`, `update()`, etc. Then the Factory would use the Plugin for all data persistence operations, instead of e.g. using knex directly

This would give us better separation of concerns - the Factory doesn't need to care _how_ it's data is persisted, it just calls the Plugin's methods and the Plugin takes care of persistence. And _that_ gives us more flexibility: if we want to switch to using the Admin API instead of knex, we just create a new Plugin and pass it to the existing PostFactory.

e2e/data-factory/data-factory.ts
```diff
@@ -0,0 +1,98 @@
...
+        // Destroy all plugins in reverse order
+        const plugins = Array.from(this.plugins.values()).reverse();
+        for (const plugin of plugins) {
+            await plugin.destroy();
+        }
+        
+        this.plugins.clear();
+        this.isDestroyed = true;
+    }
+}
```
cmraible, line 98: question: what's driving the requirement of having a top-level wrapper class around the plugins?

I worry that this could get complicated really quickly — if we add Stripe, Mailgun, ActivityPub, etc plugins in the future, I'm concerned this class could become a bit tangled with complex interdependencies.

e2e/data-factory/base-factory.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+    
+    protected generateSlug(text: string): string {
+        return text.toLowerCase()
+            .trim()
+            .replace(/[^a-z0-9\s-]/g, '')
+            .replace(/\s+/g, '-')
+            .replace(/-+/g, '-')
+            .replace(/^-+|-+$/g, '');
+    }
```
cmraible, line 101: suggestion: I would expect the base Factory class to include all of the required methods that a concrete Factory should have, i.e. `build()`, `create()`.

Otherwise we could end up with different APIs for each factory, like `postFactory.create()` and `memberFactory.insert()`. This would also give us some more type safety, to i.e. ensure all factories implement the required functions

cmraible: suggestion: I think we'll end up with _a bunch_ of these, and they might make more sense as utils, rather than methods on the base class.

e2e/data-factory/data-factory.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export interface DataFactoryOptions {
+    plugins?: DataFactoryPlugin[];
+}
+
+/**
+ * Data factory that coordinates plugins for creating test data.
+ * Each test gets its own instance for proper isolation.
+ */
+export class DataFactory {
```
cmraible, line 14: suggestion: based on the name of this class, I expected it to be a subclass of Factory, but it seems to be a completely different thing. Maybe we could call this `PluginRegistry` or `PluginProvider` or something to make it more clear that this isn't a "Factory" in the same way that PostFactory is.

e2e/data-factory/data-factory.ts
```diff
@@ -0,0 +1,98 @@
...
+        const tinybird = this.getPlugin<TinybirdPlugin>('tinybird');
+        
+        // Initialize Tinybird with Ghost's site UUID if both plugins are present
+        if (ghost && tinybird && !tinybird.isInitialized()) {
+            const siteUuid = await getSiteUuid(ghost.getDatabase());
+            if (siteUuid) {
+                await tinybird.initializeWithSiteUuid(siteUuid);
+            }
+        }
+    }
```
cmraible, line 61: thought: maybe this is necessary now since we don't have a "GhostFactory" to spin up disposable Ghost instances yet, but I think ultimately our tests should _tell_ Ghost and Tinybird which `site_uuid` to use, rather than requesting it from Ghost. That would eliminate the need for any cross plugin dependencies, at least in this case.

cmraible (commented):
Disclaimer: I haven't reviewed everything in depth yet, but I've been working my way through this for the last couple hours, which I think is long enough to form some high-level opinions.

There's a lot of really good stuff here. I _really_ like the idea of passing the factories into the tests via playwright fixtures, rather than having to import them all in every test file. And the example tests all look really good to me — easy to read, understand, and presumably write 😄

I've kept my comments pretty high-level for the most part to try to keep the focus on the overall structure and class hierarchy — I think we should all align on these big important parts before getting too into the weeds on style, variable naming, etc. I've added the comments in context so we can thread on them, but they're mostly pretty high level suggestions. (_Okay I made one naming suggestion but only because I found the current name to be misleading and it had me really confused 😂_)

One high-level comment here that didn't seem to fit in any specific files:

**Directory Structure & Config Duplication**
Since this is all added to `e2e/data-factory`, it duplicates the TypeScript and Playwright config files that already exist in the `e2e` package. I don't have super strong feelings about where this code lives, but I do feel strongly that we should only have one `tsconfig.json` and one `playwright.config.mjs` in the `e2e` package. I'd initially expected this whole thing to be a subdirectory under `helpers` but I'm not married to that. Alternatively, maybe the "framework" bits of this PR could stay in the `data-factory`, and all the concrete, Ghost specific implementation files could live in `helpers/factories`, or something along those lines.

---

Re: breaking this PR into more bite-sized chunks for easier review: I wonder if it would be easier to align on the overall structure and class hierarchy if we first focused on the "framework" bits of this PR — just the `base-factory`, `base-plugin`, `data-factory` parts with maybe a few very generic usage examples, without any of the Ghost/Tinybird/Playwright specific implementation pieces. Those are IMO the parts that I'd love for us to get aligned on: what _is_ a factory? how do I define a new factory? how does a factory interact with its persistence layer? what is a plugin? etc. Once we nail those parts, it _should_ be pretty straightforward to create the Ghost/Tinybird specific implementations of them.

Not required, just a suggestion that I think might enable a more productive discussion that's more focused on the important patterns we're establishing here, and less so on the Ghost/Tinybird/Playwright implementation details.

e2e/data-factory/base-factory.ts
9larsons: Yep, good call. Agreed. I used to have a different class that did that scaffolding.

9larsons: That's fair. I think shuffling that to utils makes sense, and would be clearer in the future state where there's more.

e2e/data-factory/data-factory.ts
9larsons: Those are absolutely more descriptive of its implementation. I was trying to just provide a simple name for ingestion from the test side because `PluginRegistry` doesn't really do it for me, ha. Agreed `DataFactory` could use a better name.. lemme think on that.

e2e/data-factory/data-factory.ts
9larsons: Yep, very good point. I can imagine we could end up with co-dependent services, though if it can be avoided we should. This was the solution in the short term as I don't recall if we can supply Ghost with a uuid? Iirc we can via config. You did that work so you're more familiar, I'll have to take another look.

e2e/data-factory/data-factory.ts
9larsons: The main purpose of this was to:
- provide a centralized setup and teardown - see the `test-fixtures` file. The getters for each factory are for convenience in the test writing, so instead of needing to do `factory.ghost.createPost` you can just do `ghost.createPost`.
- handle cross-factory dependencies if/when we have them. The goal would be to avoid these (supply siteUuid to both Ghost and Tinybird, as mentioned, vs coordinating), yet I can imagine some will need to talk with each other.

In terms of concerns over independencies, do you have anything particular in mind? I think we can't avoid needing to coordinate plugins and we need a place to do that centrally such that the test writer doesn't need to worry about it.

e2e/data-factory/plugins/ghost/ghost-plugin.ts
9larsons: I agree with separating out the persistence layer, however I don't think we need to invert the structure. At most, another plugin(?).

And maybe we're mixing intents with the labels of Plugin v Factory. So architecturally I would like to see:
- Ghost (Plugin/Factory)
- - inits PostFactory
- - -> create() uses GhostDbPlugin for persistence
- - inits TagFactory
- - -> create() uses GhostDbPlugin for persistence

And it may make sense for the top-level Ghost to inject that into the factory. That way we have one class to instantiate that has the implementation details necessary - whatever boot needs to happen, whatever sequencing/ordering necessary, etc.

e2e/data-factory/plugins/ghost/posts/post-factory.ts
9larsons: I definitely agree - you've brought this up before and it's an idea I also like. This is easy enough to bolt on and I was trying to keep this slim in some areas, hard to strike that balance.

With the goal of the most basic e2e tests, I didn't include it because we need persistence for our Analytics tests.

See the above thread about the persistence. I'm happy to add this, we'll just need to figure out how we want to handle the persistence 'plugin'.

ibalosh (changes requested):
Hey @9larsons 

A lot of great ideas here, and it closely aligns to the discussions we had before. 
I like the whole concept, but there are some things we could improve.

I took a look at the current PR and will keep comment on high level for now. There seem to be too many things in one PR that we need to align on, and it might be better to split the PR up. It's tricky to grasp them all at once, and provide the best feedback.

We could get PR decision maybe first on something like how to create data (maybe single source db). If that works and we settled for architecture, we add additional persistence layer (TinyBird). Or both, but still keep PR small enough to focus on one aspect. In this PR we have concepts of plugins, factories, db's tinybird, knex, tests, cleanup orchestrator..

I like the concept of persistence, and it's exactly what I had in mind too, love it. I am not a fan of plugins though and the entity registry and I am not sure we should have these. I was thinking of having something like persistence, factories, orchestrator (setup, cleanup) etc.

Plugin is a wider term and it seem to add more complexity and make things coupled. It seem to also have mixed responsibilities, like database configs, getSiteUuid, createDatabase connections, which all seem to be related to persistence and we might want to group these utils differently. At the same time, it has factories too, which doesn't seem like the best place for them. I would keep things as simple as possible for now and separated, just like you did with persistency minus entity registry - which seem to be coupling db and api a bit with entity metadata. 

I think code might need bit of moving here and there, and bit more decoupling, overall concepts are here though.

I took pieces from code, and moved them around a bit and played with this locally, a bit simpler path I was thinking about. 
I added here the bare minimum:

**persistence folder**, just like you did pretty much the same:

```typescript
//persistence/adapter.ts
export interface PersistenceAdapter {
  create<T>(entity: string, data: Partial<T>): Promise<T>;
  update<T>(entity: string, id: string, data: Partial<T>): Promise<T>;
  destroy(entity: string, id: string): Promise<void>;
  destroy(entity: string, ids: string[]): Promise<void>;
  find<T>(entity: string, id: string): Promise<T | null>;
  find<T>(entity: string, query?: Record<string, unknown>): Promise<T[]>;
}
```

```typescript
//persistence/adapters/knex.ts
import {PersistenceAdapter} from './adapter'

export class KnexDbAdapter implements PersistenceAdapter {
  constructor(private readonly config: object) {
    console.log('Initializing KnexDbAdapter');
  }

  async create<T>(entity: string, data: Partial<T>): Promise<T> {}
  async update<T>(entity: string, id: string, data: Partial<T>): Promise<T> {}
  async destroy(entity: string, id: string): Promise<void> {}
}
```

Keeping the routes map simple:

```typescript
//persistence/adapters/tinybird.ts
import {PersistenceAdapter} from './adapter'

class TinyBirdApiAdapter implements PersistenceAdapter {
  private apiRoutesMap = {
    post: '/api/posts/',
    page-hit: '/api/page-hit',
  };

  constructor(private readonly config: object) {
    console.log('Initializing TinyBirdApiAdapter');
  }

  async create<T>(entity: string, data: Partial<T>): Promise<T> {
    const endpoint = this.apiRoutesMap[entity];
    if (!endpoint) {  throw new Error(`Unsupported entity: ${entity}`);}
     ...
  }

  // update, destroy similarly...
}
```

We could go even further, and have custom handlers for each endpoint, but that's probably an overkill. 
For nested routes, we will need to extend it with some form of resolver though.

In factories, I like the most of it, but it seems to be doing too much like cleanups, and even has the power to cleanup anything. One option would be have that done by handler in charge of cleanups, and keep factory job and footprint smaller?

**factories folder:**

```typescript
// factories/factory
import {PersistenceAdapter} from "../persistence/index";
import {User} from "./types";

// parent factory
export class Factory<T> {
  constructor(
    private adapter: PersistenceAdapter,
    private readonly defaults: Partial<T>
  ) {}

  async create(entity: string, overrides: Partial<T> = {}): Promise<T> {
    return this.adapter.create(entity,{ ...this.defaults, ...overrides });
  }
}

// extend it to be more custom and add defaults.
// factories/user.ts
export class UserFactory extends Factory<User> {
  constructor(
    this.defaults = ...
  )
}
```

```typescript
//factories/types.ts
export type User = {
  id: string;
  name: string;
  password: string;
}

type Post = {
  id: string;
  body: string;
  title: string;
}
```

Gluing it all together with something simple as below:

```typescript
import {User, Factory} from './factories';
import {KnexDbAdapter,TinyBirdApiAdapter} from "./persistence";

const dbAdapter = new KnexDbAdapter({host: "localhost", port: 3306});
const apiAdapter =  new TinyBirdApiAdapter({host: "localhost", port: 3000});

export const factories = {
  userFactory: new UserFactory<User>(dbAdapter),
  postFactory: new PostFactory(apiAdapter),
};

const postData = { body: 'text', title: 'test' };
const userData = { name: 'Igor', password: 'test' };

factories.userFactory.create(userData);
factories.postFactory.create(postData);
```

I think plugins/factories kinda mix factories, persistence (crud), utils (like HttpClient) and maybe bit more at the moment. Factories main purpose is as their name implies, creating data, (could do though update, destroy too for us, if we really need to). Cleanup for factories, other than maybe themselves seems like bit too much responsibility. Base factory does cleanup, setup, and some other things too. 

One possibility is for factories to not clean up themselves, they could participate in cleanup by registering what they created.

We could have something like: `dataRegistry.register(testId, userFactory);`,  so that we can use that registry to cleanup later. That way we still keep factories cleaner. And then do something like (or do it with factory destroy):

```typescript
import {testDataRegistry} from '../registry/test-data-registry';

export async function cleanupAllData(testId: string) {
  const entries = testDataRegistry.getAll(testId); 

  for (const entry of entries) {
    await entry.destroy();
  }

  testDataRegistry.clearAll(testId);
}
```

Even better, if the factories are not involved in registry for cleanup, and we just use same persistence adapter as we did for that factory. 

That way we have decoupled:
* persistence
* factories for creating
* data registry for handling cleanups

This also makes it cleaner, and simpler if we later decide to use different approach for cleanup (db's on demand, rolling back transactions, etc), we just update the handler for cleanups.

I agree with Chris regarding folders, and I think we should have a single source of truth for things like playwright config for example, in e2e root.

Folder structure wise, I would go for something like this on top level, just thought of some possible files that we could have:

```
helpers/
  factories/
  persistence/
  environment/
   - check-availability.ts 
  registry
   - test-data-registry.ts
  setup
   - cleanup-data.ts
   - generate-data.ts 
  config/...
  utils/
   - http-client
   - ...
  pages/..
tests  
```

in tests:

```typescript
test('create a post with views', async ({factories}) => {
    // we do not care if it's tinybird, knex, or prism - those are implementation details
    // factory name is descriptive enough to keep create method name consistent, with less to maintain
    const post = await factories.postFactory.create({publisher: true});
    const hits = await factories.pageHitsFactory.create(post, 100, {member_status: 'free'});   
    ...
});
```

Just some thoughts of options, overall great work! It has more than I thought we would have in initial PR 🙌

PR #24492, opened 2025-07-24
🐛 Require newsletter_id when email_segments is used

### LINKED ISSUE #24490: Silent failure when sending emails for posts created by API
### Issue Summary

I created some scheduled posts using the API. Now after one them attempted to send, there’s silent failure with no specific error message in the UI or in the logs.

To start with, here’s what a normal failure-to-send result looks like in the admin UI:

Image

That’s not what happened here. Instead, if you roll over the post in the post list, it says “Sent to members”. But that’s not the normal UX for success either. The normal UX for Success is the message “Sent to 62 members”.

Opening up the post, it reports the same message with the number of users missing:

Image

It shows up in the “Recent Posts” area, but there is no “Sent” or Open Rate" UI even though it was an email-only post:

Image

And this curious: If using the Post filters for “Drafts”, “Scheduled” or “Published”, it doesn’t show up in any category.

I also cannot find anything specific error in the logs around the time email should be sent. Later there’s this generic error message which is not helpful:

```
[2025-07-19 07:07:06] ERROR Internal Server Error
[2025-07-19 07:07:06] ERROR [EmailAnalytics] Error while fetching
[2025-07-19 07:07:06] ERROR Internal Server Error
[2025-07-19 07:07:06] INFO [EmailAnalytics] No new events found
[2025-07-19 07:07:06] ERROR Error while fetching email analytics Internal Server Error
[2025-07-19 07:07:06] ERROR Internal Server Error
```

I suspect the issue here is that Ghost expects every email to have email analytics available, but because of the zombie state of this post, it does not have any and this case was not accounted for, resulting in Internal Server Error.

I have about a dozen more scheduled posts created via API the same way, so I’m concerned that without understanding the root cause of this better that none of my posts scheduled for the rest of the year will work.

It seems to me that in one or places there was a data validation failure and when the bad data was attempted to be used, the post got put into an invalid state.

Because there are more scheduled posts that may have this problem, it’s possible that I could provide some kind of database-level export of the tables/rows involved if someone wants to take a look.

I checked on the Mailgun time around the time of the scheduled post and see nothing there.

When I tried duplicating the post and sending today, it worked. That confirms that the flow between Ghost and Mailgun is working fine, and if there was a bad data issue, the duplicate post feature somehow resolved it.

I tried upgrading to 5.130.1, the latest release, and the weird Admin area UX remains: Still showing as “sent” without a particular number of users, but not showing as “Published”. It’s a Heisenpost!

### Steps to Reproduce

Assuming I'm right about what the cause, the steps to reproduce are:

1. Ask for what you need from my database in terms of posts scheduled like this but yet sent.
2. Receive data you need.
3. Reproduce on a second instance. 

### Ghost Version

5.127.2

### Node.js Version

20

container on Fedora Linux 41.

### Database type

MySQL 8

### Browser & OS version

_No response_

```shell

```

### Code of Conduct

markstos:
This keeps the app from getting to an invalid state where it
appears to set to send to email but it's not actually going anyway.

Fixes #24490

## QA Log

- Added automated test
- Confirmed it failed
- Added code
- Confirmed test passed

### DIFF
--- a/ghost/core/core/server/api/endpoints/posts.js
+++ b/ghost/core/core/server/api/endpoints/posts.js
@@ -156,7 +156,9 @@ const controller = {
         options: [
             'include',
             'formats',
-            'source'
+            'source',
+            'email_segment',
+            'newsletter'
         ],
         validation: {
             options: {
--- a/ghost/core/core/server/api/endpoints/utils/validators/input/posts.js
+++ b/ghost/core/core/server/api/endpoints/utils/validators/input/posts.js
@@ -52,15 +52,26 @@ const validateSingleContentSource = async function (frame) {
     }
 };
 
+const validateEmailSegmentWithNewsletter = async function (frame) {
+    if (frame.options.email_segment && !frame.options.newsletter) {
+        return Promise.reject(new ValidationError({
+            message: 'The email_segment parameter requires the newsletter parameter to be set',
+            property: 'email_segment'
+        }));
+    }
+};
+
 module.exports = {
     async add(apiConfig, frame) {
         await jsonSchema.validate(...arguments);
         await validateVisibility(frame);
         await validateSingleContentSource(frame);
+        await validateEmailSegmentWithNewsletter(frame);
     },
     async edit(apiConfig, frame) {
         await jsonSchema.validate(...arguments);
         await validateVisibility(frame);
         await validateSingleContentSource(frame);
+        await validateEmailSegmentWithNewsletter(frame);
     }
 };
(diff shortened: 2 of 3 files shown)


markstos:
@kevinansfield Turns out there was already a similar test that says:

...
  // We also test whether email_segment is ignored if no newsletter is sent                                                        
  1319  ┆   ┆   ┆   .put(localUtils.API.getApiQuery('posts/' + id + '/?email_segment=status:-free')) 
...

After re-reading Kevin's comment in #24490, this was the wrong approach here.

PR #24522, opened 2025-07-28
Fixed duplicate tags and missing post count in tags list

troyciesco:
closes https://linear.app/ghost/issue/ENG-2478

- fixes an issue where a copy of a tag would get added to the list on /tags on save. Ember already updates items automatically, this change makes it so only new tags are added on save
- also fixes an issue where a new tag would say "posts" instead of "0 posts"

to see the issue:
- on `main`, edit a tag and then click back to /tags. You'll see the tag listed twice. It'll increase every time you edit that tag.
- create a tag, notice it just says "posts"

to see the fix:
- on this branch, create and edit tags. if you have 100+ tags, make some at the beginning and end of the list.
- tags get created and updated correctly, and only show in the list once.

### DIFF
--- a/ghost/admin/app/components/tags/list-item.hbs
+++ b/ghost/admin/app/components/tags/list-item.hbs
@@ -20,7 +20,7 @@
         </LinkTo>
     {{else}}
         <LinkTo @route="tag" @model={{@tag}} class="gh-list-data gh-tag-list-posts-count gh-list-cellwidth-10" title="Edit tag">
-            <span class="nowrap f8 midlightgrey">{{gh-pluralize @tag.count.posts "post"}}</span>
+            <span class="nowrap f8 midlightgrey">0 posts</span>
         </LinkTo>
     {{/if}}
 
--- a/ghost/admin/app/components/tags/list-item.hbs
+++ b/ghost/admin/app/components/tags/list-item.hbs
@@ -20,7 +20,7 @@
         </LinkTo>
     {{else}}
         <LinkTo @route="tag" @model={{@tag}} class="gh-list-data gh-tag-list-posts-count gh-list-cellwidth-10" title="Edit tag">
-            <span class="nowrap f8 midlightgrey">0 posts</span>
+            <span class="nowrap f8 midlightgrey" data-test-tag-count>0 posts</span>
         </LinkTo>
     {{/if}}
 
--- a/ghost/admin/app/controllers/tag.js
+++ b/ghost/admin/app/controllers/tag.js
@@ -45,10 +45,12 @@ export default class TagController extends Controller {
             if (tag.get('errors').length !== 0) {
                 return;
             }
+            const wasNew = tag.isNew;
             yield tag.save();
 
-            this.tagsManager.tagsScreenInfinityModel?.pushObjects([tag]);
-
+            if (wasNew) {
+                this.tagsManager.tagsScreenInfinityModel?.pushObjects([tag]);
+            }
             // replace 'new' route with 'tag' route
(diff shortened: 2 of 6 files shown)


ghost/admin/app/components/tags/list-item.hbs
```diff
@@ -20,7 +20,7 @@
         </LinkTo>
     {{else}}
         <LinkTo @route="tag" @model={{@tag}} class="gh-list-data gh-tag-list-posts-count gh-list-cellwidth-10" title="Edit tag">
-            <span class="nowrap f8 midlightgrey">{{gh-pluralize @tag.count.posts "post"}}</span>
+            <span class="nowrap f8 midlightgrey">0 posts</span>
```
troyciesco: Could also do `{{gh-pluralize (or @tag.count.posts 0) "post"}}` but since we have the `if` check on L17 I thought putting it manually might make more sense. 

Fixes the below issue (ignore the slug not matching, i just didn't update it):

Screenshot 2025-07-28 at 3 10 17 PM

ghost/admin/app/controllers/tag.js
```diff
@@ -45,10 +45,12 @@ export default class TagController extends Controller {
             if (tag.get('errors').length !== 0) {
                 return;
             }
+            const wasNew = tag.isNew;
             yield tag.save();
```
troyciesco, line 49: i think once `tag.save()` runs then `isNew` is changed to false, thus the `wasNew` variable.

ghost/admin/app/components/tags/list-item.hbs
kevinansfield: Seems like the right fix for me 👍

ghost/admin/app/controllers/tags.js
```diff
@@ -16,9 +16,18 @@ export default class TagsController extends Controller {
...
+        // this makes sure each tag only shows up once
+        const tagMap = new Map();
+        
+        this.tags.forEach((tag) => {
+            if (!tag.isNew && !tag.isDestroyed && !tag.isDestroying && !tag.isDeleted && (!this.type || tag.visibility === this.type)) {
+                tagMap.set(tag.id, tag);
+            }
         });
+        
+        return [...tagMap.values()];
```
kevinansfield: This seems unnecessary now with the fix for the incorrect push onto the array. Was there something else that prompted this to be included or was it from an earlier fix attempt?

troyciesco: Ah my comment isn't clear enough on what this is doing then - the `wasNew` fix works for making sure new ones get added and  edited ones don't, but if a user has 100+ tags, they can end up with a duplicate if the new one is created before its page has been fetched.

In this gif, the fix to `filteredTags` is _not_ implemented:
1. user has 100+ tags, and starts on the /tags page
2. user doesn't scroll (so doesn't fetch past the first page), adds a new tag, goes back to /tags
3. the tag will show up twice - basically the user had fetched page 1, added the new tag, and then fetched page 2, which now also had the new tag  
dup-tag


In this gif, fix is implemented:
dup-tag-fix-100-plus

def open to better ways to handle this scenario if there is one!

kevinansfield: Ahhh, gotcha. OK, so this change makes sense now 👌 Maybe we can update the comment to be even more explicit?

```
// new tags are preemptively added to the client-side tagsScreenInfinityModel,
// but if the new tag is included in a later pagination request it will end up duplicated
// this makes sure each tag only shows up once
```

kevinansfield (approved):

PR #24526, opened 2025-07-29
🐛 Fixed duplicate tag in list after tag edit

kevinansfield:
closes https://linear.app/ghost/issue/ENG-2478

- adds conditional so we only push a tag onto the tags list after save when it's a new tag, for edits we can rely on Ember auto-updating


kevinansfield:
Accidentally duplicated existing PR, closing in favor of https://github.com/TryGhost/Ghost/pull/24522

PR #24530, opened 2025-07-29
Fixed order of averages on Newsletters tab

peterzimon:
ref https://linear.app/ghost/issue/ENG-2483/analytics-newsletters-graph-shows-data-from-newest-to-oldest-instead

- The order of newsletters on the Newsletters / Avg. open rate and Avg. click rate tabs were by date descending


9larsons (approved):
Changed sort to do it on the backend instead of front 👍 .

PR #24534, opened 2025-07-29
🐛 Fixed remote posts not loading

vershwal:
ref https://linear.app/ghost/issue/PROD-2350

- If a post is not present on our db, `getReplies` API will fail to load it, this is only the case with remote posts
- Calling `getPost` API will fetch the post from network in the BE and save it in the db
- Recalling `getReplies` to load the post in UI

### DIFF
--- a/apps/admin-x-activitypub/package.json
+++ b/apps/admin-x-activitypub/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/admin-x-activitypub",
-  "version": "0.9.11",
+  "version": "0.9.12",
   "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


apps/admin-x-activitypub/src/views/Profile/components/Posts.tsx
```diff
@@ -82,9 +82,13 @@ const Posts: React.FC<PostsProps> = ({
                             type={activity.type}
                             onClick={() => {
                                 if (activity.object.type === 'Note') {
-                                    navigate(`/notes/${encodeURIComponent(activity.object.id)}`);
+                                    navigate(`/notes/${encodeURIComponent(activity.object.id)}`, {
+                                        state: {profilePost: true}
```
vershwal: Introduced this because we navigate to `/notes/{post_apId}` from multiple components like notes feed, reader feed etc

allouis (approved):

PR #24537, opened 2025-07-29
Fixed member growth source of Direct showing as a link

9larsons:
ref https://linear.app/ghost/issue/PROD-2336/
- 'Direct' should not link; we handle this in other components, and needed to add handling here to prevent dead/broken links

__Testing__
- [x] Publish a post and sign up a new member through direct navigation. Go to Post Analytics > Growth an ensure Direct is not a link.


cmraible (approved):
Nice one, looks good to me! Def an opportunity to refactor, but I appreciate that you just fixed the bug for now — can always refactor it later 👍🎉

PR #24538, opened 2025-07-29
Prevented stats script from loading within iframes

9larsons:
ref https://linear.app/ghost/issue/PROD-2338/

We previously had handling for post preview, but navigation past that first view was still loading ghost_head, and any query param/middleware solutions were ephemeral as well.

It's common practice to prevent loading of stats scripts within iframes for a variety of reasons, and this is a far simpler and more robust implementation.

__Testing__
- [ ] Clicking around in Admin > Site does not make a network request to web_analytics. 
- [ ] Click around in Editor > Preview does not make a network request to web_analytics.

### DIFF
--- a/ghost/core/core/frontend/helpers/ghost_head.js
+++ b/ghost/core/core/frontend/helpers/ghost_head.js
@@ -151,6 +151,11 @@ function getWebmentionDiscoveryLink() {
 }
 
 function getTinybirdTrackerScript(dataRoot) {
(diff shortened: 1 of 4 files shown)


cmraible (changes requested):
I like the approach of using the iframe, but I think we should keep the preview handling as well. With this change, it's possible to preview a post from within the editor, and pop the preview into a new tab, which then bypasses the iframe check.

cmraible (approved):
LGTM!

PR #24542, opened 2025-07-29
Fixed config handling crashing stats/posts apps when empty strings

9larsons:
ref https://linear.app/ghost/issue/PROD-2355/
- log console.warn instead of fatal error when config is inappropriate to fetch token (empty string)

The frontend apps require workspaceId and adminToken to be set in Ghost's config:tinybird block. When set to an empty string, they were causing the app to crash due to the response from the API. This has been updated to handle it more gracefully by logging a console warn statement and not crash the apps. The user can still see a CTA in the Settings > Analytics block that their config is incorrect.

__Testing__
- [ ] Analytics loads web data with a valid config.
- [ ] Analytics loads w/ missing config; missing web data, shows console.warn, shows CTA in Settings > Analytics.
- [ ] Analytics loads w/ empty string config (workspaceId, adminToken, + various combinations); missing web data, shows console.warn, shows CTA in Settings > Analytics.

### DIFF
--- a/apps/admin-x-framework/src/hooks/useTinybirdToken.ts
+++ b/apps/admin-x-framework/src/hooks/useTinybirdToken.ts
@@ -11,16 +11,23 @@ export interface UseTinybirdTokenOptions {
     enabled?: boolean;
 }
 
+// Track if we've already logged the warning to avoid spamming the console
+let hasLoggedConfigWarning = false;
+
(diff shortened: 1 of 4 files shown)


cmraible (approved):
LGTM, ran through all the different config possibilities and it's working for me 👍

PR #24546, opened 2025-07-29
Updated 6.0 update CTA

peterzimon:
ref https://github.com/TryGhost/Ghost/commit/5de425517d15e97af63077c13f8e270dabca20b6

- This is a refinement of the commit above. It adds a link to the 6.0 landing page and updates the copy

### DIFF
--- a/ghost/admin/app/styles/components/notifications.css
+++ b/ghost/admin/app/styles/components/notifications.css
@@ -345,7 +345,7 @@
     display: flex;
     justify-content: center;
     align-items: center;
-    padding: 12px;
+    padding: 14px;
     background: var(--black);
     color: #fff;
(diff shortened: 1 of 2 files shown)


kevinansfield (approved):

kevinansfield (changes requested):
Coderabbit's suggestion about using `<button>` instead of `<a>` is accurate. If we want to stick with `<a>` I would suggest adding the `event.preventDefault()` to `openUpdateTab()` so there are no unexpected redirects inside Admin when the link is clicked

peterzimon:
Closing this one after realizing this needs to go to 5.x instead of 6.
New PR created here: https://github.com/TryGhost/Ghost/pull/24563

PR #24547, opened 2025-07-29
Fixed ?ref being dropped by stats script attribution with direct navigation

9larsons:
ref https://linear.app/ghost/issue/PROD-2354/
ref https://github.com/TryGhost/Ghost/commit/9e048cc4c62489dad616598a10b4284763afe100
- direct navigation to a ghost site with a ?ref, ?source, ?utm_source query param is now appropriately associated with the referrer (previously mistreated as direct)

When we removed the referrer fallback in the above commit within our Tinybird pipes, we inadvertently removed fallback behavior previously in the `ghost-stats` script that is critical for handling `?ref` and related url params, whose behavior should match the `member-attribution` script.

`member-attribution` has [fallback behavior](https://github.com/TryGhost/Ghost/blob/c1465feb65863102e20876b9ae5502763d3161a7/ghost/core/core/frontend/src/member-attribution/member-attribution.js#L97) that ensures the `referrer.url` field is always populated - this is critical for the `referrer-parser` lib to [conduct](https://github.com/TryGhost/SDK/blob/b5c29510af8c1af9a11f593f91f5834ebf821594/packages/referrer-parser/lib/ReferrerParser.ts#L60) its parsing. While this isn't particularly intuitive, a fuller redesign of all this is out of scope, as the `member-attribution` parsing logic was mere centralized in TryGhost/SDK to be used by `ghost-stats`, and a more appropriate solution would be to rebuild it such that it makes sense to be used in multiple areas of the code base.

In short, we need to ensure we set `referrer.url`, and we should be using `getReferrer` for this. This is a difficult piece to catch in tests because the code is now split (`ghost-stats` script in TryGhost/Ghost, `referrer-parser` lib in TryGhost/SDK, and the traffic analytics/page hit parsing logic in TryGhost/TrafficAnalytics. We need true e2e tests to have good coverage over this part of the code base (which is a WIP).

__Testing__
- [ ] Execute page hits including `?ref=google` and related sources. Ensure those show up in Analytics > Web > Sources.
- [ ] We'll want to move this build onto staging and test site newsletters, which is difficult to do locally w/ localhost. Send a newsletter from Site 1 that links to Site 2. Click and follow the traffic.

### DIFF
--- a/ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
+++ b/ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
@@ -1,5 +1,5 @@
 import { getCountryForTimezone } from 'countries-and-timezones';
-import { parseReferrer } from '../utils/url-attribution';
+import { getReferrer, parseReferrer } from '../utils/url-attribution';
 import { processPayload } from '../utils/privacy';
 import { BrowserService } from './browser-service';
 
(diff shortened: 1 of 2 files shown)


cmraible (approved):
Tested a newsletter and a direct link from another site, both worked 👍 

Agree that this could use a refactor but this is a nice minimal, low risk change to fix the bug for now. Tests look good and should cover us for now until we get some real e2e tests cooked up here 🚀

PR #24553, opened 2025-07-29
Fixed handling for 'Today' lookback with Analytics kpis

9larsons:
ref https://linear.app/ghost/issue/PROD-2243/
- api_kpis wasn't appropriately handling timezones, and now returns localized data that doesn't allow gaps
- formatDisplayDate was assuming UTC returned data when parsing timestamps, which was incorrectly adjusting the data returns from the API

Ideally we would handle UTC data across the board, but we need a serious rethink of how we handle startDate and endDate. This appears to handle the case of 'Today' in Analytics > Overview, Analytics > Web, and Post Analytics > Overview/Web.

### DIFF
--- a/apps/shade/src/lib/utils.ts
+++ b/apps/shade/src/lib/utils.ts
@@ -160,21 +160,40 @@ export const formatQueryDate = (date: Moment) => {
 
 // Format date for UI, result is in the formate of `12 Jun 2025`
 export const formatDisplayDate = (dateString: string): string => {
+    // Check if this is a datetime string (contains time) or just a date
+    const hasTime = dateString.includes(':');
+    
     const date = new Date(dateString);
     const today = new Date();
-    const isToday = date.toISOString().slice(0, 10) === today.toISOString().slice(0, 10);
-    const isCurrentYear = date.getUTCFullYear() === today.getUTCFullYear();
-
-    const months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
-    const day = date.getUTCDate();
-    const month = months[date.getUTCMonth()];
-    const year = date.getUTCFullYear();
-
-    if (isToday) {
-        return `${day} ${month}`;
+    
+    let day, month, year, isToday, isCurrentYear;
+    
+    if (hasTime && !dateString.includes('T') && !dateString.includes('Z')) {
+        // This is a localized datetime string like "2025-07-29 19:00:00"
+        // Use local date methods to avoid timezone conversion
+        day = date.getDate();
+        month = date.getMonth();
+        year = date.getFullYear();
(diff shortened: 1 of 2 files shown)


cmraible:
Okay some thoughts:
- This does seem to fix the weird axis in the Post Analytics > Overview chart. It's no longer spilling into tomorrow for me, and the data all seems to be correct

But I'm afraid (not sure) that this will break things in other places, just because I don't _really_ understand what all we're changing here. It feels like we're fixing two things here:
- The wonky x-axis in the Post Analytics > Overview
- The off-by-one-day issue in the other KPI Views

I'd love to figure out exactly which parts of this PR fix the first issue and merge that, then tackle the second issue separately, just to make it easier to reason about. This does seem to fix a thing, but I'm having a hard time understanding _why_ it fixes it, which makes me nervous

9larsons:
Agreed. I also felt like this was two things and could've said that above, though it feels a bit wrong to push `main` into a knowingly bad state, so I combined these. We can split them to aid, although I think it's fine to also look with just temporary commit reversion. 

`api_kpis` - I feel it's reasonably easy to look at this in isolation. We're fixing this to actually handle TZs in the lookup in the 24h lookback, and we can simply look at the reply from the endpoint to reason about that change. It may be easier to instead look at the frontend changes first because of this.

apps/shade/src/lib/utils.ts
```diff
@@ -160,21 +160,40 @@ export const formatQueryDate = (date: Moment) => {
...
+    const hasTime = dateString.includes(':');
+    
     const date = new Date(dateString);
     const today = new Date();
-    const isToday = date.toISOString().slice(0, 10) === today.toISOString().slice(0, 10);
-    const isCurrentYear = date.getUTCFullYear() === today.getUTCFullYear();
+    
+    let day, month, year, isToday, isCurrentYear;
+    
+    if (hasTime && !dateString.includes('T') && !dateString.includes('Z')) {
```
9larsons, line 171: Date functions treat `2025-07-29 19:00:00` as a localized datetime, meanwhile the old version of this treated everything as a UTC date.

This change ensures we treat a localized datetime accordingly.

apps/shade/src/lib/utils.ts
```diff
@@ -160,21 +160,40 @@ export const formatQueryDate = (date: Moment) => {
...
+        // This is a localized datetime string like "2025-07-29 19:00:00"
+        // Use local date methods to avoid timezone conversion
+        day = date.getDate();
+        month = date.getMonth();
+        year = date.getFullYear();
+        isToday = date.toDateString() === today.toDateString();
+        isCurrentYear = year === today.getFullYear();
+    } else {
+        // This is either a date-only string or an ISO format with timezone
```
9larsons, line 180: This block is adding to confusion. It's really not necessary _for this change_ - it's just future proofing it against ingesting datetimes **with** timezone codes specified.

PR #24555, opened 2025-07-30
Fixed handling for 'Today' lookback with Analytics kpis

9larsons:
ref https://linear.app/ghost/issue/PROD-2243/
- formatDisplayDate was assuming UTC returned data when parsing timestamps, which was incorrectly adjusting the data returns from the API

Ideally we would handle UTC data across the board, but we need a serious rethink of how we handle startDate and endDate. This appears to handle the case of 'Today' in Analytics > Overview, Analytics > Web, and Post Analytics > Overview/Web.

### DIFF
--- a/apps/stats/src/views/Stats/Growth/components/GrowthKPIs.tsx
+++ b/apps/stats/src/views/Stats/Growth/components/GrowthKPIs.tsx
@@ -234,7 +234,7 @@ const GrowthKPIs: React.FC<{
                 const todayData = subscriptionData.find(item => item.date === today);
 
                 return [{
-                    date: formatDisplayDateWithRange(new Date(today), range),
+                    date: formatDisplayDateWithRange(today, range),
                     new: todayData?.signups || 0,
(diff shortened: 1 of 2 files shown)


cmraible (changes requested):
For some reason the app is crashing on Analytics > Growth > Paid members for me on this PR branch, but not on main.

https://github.com/user-attachments/assets/00723170-9acf-440a-a1b2-0df226199c95

apps/stats/src/views/Stats/Growth/components/GrowthKPIs.tsx
```diff
@@ -234,7 +234,7 @@ const GrowthKPIs: React.FC<{
                 const todayData = subscriptionData.find(item => item.date === today);
 
                 return [{
-                    date: formatDisplayDateWithRange(new Date(today), range),
+                    date: formatDisplayDateWithRange(today, range),
```
9larsons, line 237: We should be calling this with a string, not a Date object.

cmraible (approved):
LGTM

PR #24556, opened 2025-07-30
Fixed domain blocking UI update and labeling in ActivityPub

minimaluminium:
ref PROD-2348

- Corrected query key for domain block/unblock mutations to properly update UI
- Fixed UI not reflecting blocking changes despite successful backend operations
- Changed profile menu item from "Unblock user" to "Unblock domain" when only domain is blocked

### DIFF
--- a/apps/admin-x-activitypub/src/views/Profile/components/ProfileMenu.tsx
+++ b/apps/admin-x-activitypub/src/views/Profile/components/ProfileMenu.tsx
@@ -128,7 +128,9 @@ const ProfileMenu: React.FC<ProfileMenuProps> = ({
                                 }
                                 setDialogOpen(true);
                             }}>
-                                {isBlocked || isDomainBlocked ? 'Unblock' : 'Block'} user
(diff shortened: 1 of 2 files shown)


apps/admin-x-activitypub/src/views/Profile/components/ProfileMenu.tsx
```diff
@@ -128,7 +128,9 @@ const ProfileMenu: React.FC<ProfileMenuProps> = ({
                                 }
                                 setDialogOpen(true);
                             }}>
-                                {isBlocked || isDomainBlocked ? 'Unblock' : 'Block'} user
+                                {isBlocked || isDomainBlocked ? 'Unblock' : 'Block'}
+                                {` `}
+                                {!isBlocked && isDomainBlocked ? 'domain' : 'user'}
```
vershwal: I think we can make it more readable by using something like:-
```
{isBlocked ? 'Unblock user' : isDomainBlocked ? 'Unblock domain' : 'Block user'}
```

vershwal (approved):
Looks good, just one small comment about making the condition more readable

PR #24558, opened 2025-07-30
Fixed data inconsistency in Post Analytics > Overview tab

cmraible:
ref https://linear.app/ghost/issue/PROD-2243/inconsistency-in-the-post-analytics-overview-data-compared-to-post

On the Post Analytics > Overview tab, the top level metric shown for unique visitors and the sum of the actual data in the chart can show different values. The metric and the chart are using separate requests to the same API endpoint with different `date_from` parameters, resulting in slightly different numbers, i.e. it's possible to see "1 unique visitor" but with 0 for every data point on the chart:

image

This fixes that by removing the duplicate API call, so the data for the metric and the chart use the same exact underlying data.

### DIFF
--- a/apps/posts/src/views/PostAnalytics/Overview/Overview.tsx
+++ b/apps/posts/src/views/PostAnalytics/Overview/Overview.tsx
@@ -18,7 +18,6 @@ const Overview: React.FC = () => {
     const navigate = useNavigate();
     const {statsConfig, isLoading: isConfigLoading, post, isPostLoading, postId} = useGlobalData();
     const {totals, isLoading: isTotalsLoading, currencySymbol} = usePostReferrers(postId);
-    const {startDate, endDate, timezone} = getRangeDates(STATS_RANGES.ALL_TIME.value);
     const {appSettings} = useAppContext();
     const {emailTrackClicks: emailTrackClicksEnabled, emailTrackOpens: emailTrackOpensEnabled} = appSettings?.analytics || {};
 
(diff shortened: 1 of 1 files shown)


apps/posts/src/views/PostAnalytics/Overview/Overview.tsx
```diff
@@ -73,28 +52,22 @@ const Overview: React.FC = () => {
         return baseParams;
     }, [isPostLoading, post, statsConfig?.id, chartStartDate, chartEndDate, chartTimezone]);
 
-    const {data, loading: tbLoading} = useTinybirdQuery({
-        endpoint: 'api_kpis',
-        statsConfig: statsConfig || {id: ''},
-        params: params
-    });
-
```
cmraible, line 81: TL;DR: the "Unique Visitors" metric was using the value from this API call, which used the all time data, but the chart filtered from the `post.published_at` timestamp. 

Since some "unique visitors" can be attributed to the day before the `post.published_at` timestamp, this resulted in different data for the metric than the chart.

cmraible:
This fixes the inconsistency on the Post Analytics > Overview tab itself, and between the Overview <> Web Traffic tabs. It does not fix the fact that a "unique visitor" can be attributed to the day before a post is published, since we use the first pageview of the visit to define its timestamp. This is a stickier issue to solve, but this at least makes it less obvious to a user.

9larsons (approved):

PR #24560, opened 2025-07-30
Updated links in Admin for Analytics and Network

peterzimon:
ref https://linear.app/ghost/issue/PROD-2086/update-the-dashboard-to-native-analytics

- The link for the "Understanding analytics in Ghost" banner on the Analytics overview was pointing to the old URL.
- The link for "Get started" for self-hosters in Settings / Analytics was pointing to docs home instead of the specific article.
- The "Help" link in Settings / Network was incorrectly pointing to ghost.org/docs

### DIFF
--- a/apps/stats/src/views/Stats/Overview/Overview.tsx
+++ b/apps/stats/src/views/Stats/Overview/Overview.tsx
@@ -221,7 +221,7 @@ const Overview: React.FC = () => {
                     <HelpCard
                         description='Find out how to review the performance of your content and get the most out of post analytics in Ghost.'
                         title='Understanding analytics in Ghost'
-                        url='https://ghost.org/help/post-analytics/'>
+                        url='https://ghost.org/help/native-analytics'>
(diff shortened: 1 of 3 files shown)


apps/stats/src/views/Stats/Overview/Overview.tsx
```diff
@@ -221,7 +221,7 @@ const Overview: React.FC = () => {
                     <HelpCard
                         description='Find out how to review the performance of your content and get the most out of post analytics in Ghost.'
                         title='Understanding analytics in Ghost'
-                        url='https://ghost.org/help/post-analytics/'>
+                        url='https://ghost.org/help/native-analytics'>
```
minimaluminium, line 224: Looks like this link doesn't exist yet. I wonder if it's intended.

peterzimon: Thank you, yes, the page is created yet but it's coming.

minimaluminium (approved):

Issue #24536, opened 2025-07-29
🐛 Various inconsistent and outdated mappings in timezone data


niranjan-uma-shankar:
### Issue Summary

The mapping between IANA timezone identifiers and city labels contains several inconsistencies. The issues include incorrect timezone mappings and GMT offsets for some cities, and the usage of outdated names for some cities.

### 1. Update city names

* **Issue**: `Ulaan Bataar` is incorrectly written as two words (source: [wikipedia](https://en.wikipedia.org/wiki/Ulaanbaatar)).
**Suggested Fix**: Update to `Ulaanbataar`.

* **Issue**:  In `Chennai, Calcutta, Mumbai, New Delhi`, all cities use their modern names except `Calcutta`, which is the former name for the city.
**Suggested Fix**: Update `Calcutta` to `Kolkata` for consistency.

* **Issue**: Rangoon is the former name for the city.
**Suggested Fix**: Update `Rangoon` to `Yangon`.

### 2. Incorrect offset for Kazakhstan cities

**Issue**
* [Almaty](https://github.com/TryGhost/SDK/blob/b5c29510af8c1af9a11f593f91f5834ebf821594/packages/timezone-data/src/index.ts#L196) and [Astana](https://github.com/TryGhost/SDK/blob/main/packages/timezone-data/src/index.ts#L200), both in Kazhakstan, are currently listed under GMT+6 as separate entries.

As of March 1, 2024, Kazakhstan transitioned to a single GMT+5 timezone. From [this source](https://www.independent.co.uk/asia/central-asia/kazakhstan-astana-single-time-zone-b2505299.html):
> Starting the night of 29 February leading into 1 March, the cities of Astana, Almaty, and Shymkent, along with the Abai, Akmola, Almaty, Zhambyl, Zhetisu, Karagandy, Kostanai, Pavlodar, North Kazakhstan, Turkistan, Ulytau, and East Kazakhstan regions adjusted their local time by setting the clocks back one hour.

**Suggested Fix**

* Update both `Almaty` and `Astana` to GMT+5.
* Combine them into a single entry under the same timezone identifier, since they are cities belonging to the same country and timezone.

### 3. Incorrect offset for `Novosibirsk`

**Issue**
[...]

PR #24562, opened 2025-07-30
Update tests for the timezone-data package

niranjan-uma-shankar:
This PR updates the timezone data test and fixtures for the timezone data, which is needed for the changes made to the `timezone-data` package in https://github.com/TryGhost/SDK/pull/585.

Ref: https://github.com/TryGhost/Ghost/issues/24536

**Note**:

https://github.com/TryGhost/SDK/pull/585 will need to be shipped for the tests here to pass. In the meantime, run the following test locally to verify they pass:

```
cd ghost/admin/tests 
yarn test 1 --filter="config-manager"
```


9larsons:
@niranjan-uma-shankar Hey there - I've merged your SDK bump, thanks! This file is actually not referenced as a fixture (anymore) so I'm going to instead close this and create a PR to remove the file to prevent future confusion.

PR #24570, opened 2025-07-31
Fixed top sources not deduplicating free->paid conversions

9larsons:
ref https://linear.app/ghost/issue/ENG-2496/analytics-growth-top-sources-sources-free-members-is-not-showing
- in other places (post stats) we 'deduplicate' member counts where a free signup also includes a paid conversion in the same time frame; this was not handled in Sources
- added unit test coverage to describe cases/behavior

We've defined member attribution in the Analytics apps as only counting a free signup if the user didn't also sign up as a paid member in the same timeframe. This makes it a bit easier to reason about your total member count when looking at the free/paid split.

__Testing__
- [ ] Publish a post.
- [ ] Sign up a free member on it.
- [ ] Check Analytics > Growth to see the free signup.
- [ ] With the same post, upgrade to paid (may need to enable Stripe). 
- [ ] Check Analytics > Growth to see the free signup disappear and paid signup appear.

### DIFF
--- a/ghost/core/core/server/services/stats/ReferrersStatsService.js
+++ b/ghost/core/core/server/services/stats/ReferrersStatsService.js
@@ -250,51 +250,6 @@ class ReferrersStatsService {
         return rows;
     }
 
-    /**
-     * Fetch paid conversion sources with date range
-     * @param {string} startDate 
-     * @param {string} endDate 
-     * @returns {Promise<PaidConversionsCountStatDate[]>}
-     **/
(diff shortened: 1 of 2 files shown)


cmraible:
I've reproduced the bug on main, then checked out this branch and see that it is correct now. Just doing a quick pass on the code now.

cmraible (approved):
This is a tricky one, nicely done! LGTM

PR #24573, opened 2025-07-31
Fixed upgrade popup being rendered one newsletter too early

sagzy:
ref https://linear.app/ghost/issue/BAE-494

- Previously: given a Ghost site with e.g. a limit of 3 max. newsletters, a popup to upgrade would appear in the middle of creating newsletter 3
- Now: the popup to upgrade only appears when attempting to add newsletter 4


aileen (approved):
Looks good! Just needs the tests to pass now.

PR #24574, opened 2025-07-31
Fixed stats endpoints inconsistently handling timezone

9larsons:
ref https://linear.app/ghost/issue/PROD-2373/
- more declarative about passing timezone with date_from and date_to in stats endpoint options
- centralized date calculations and date filter application to knex query objects; this allows us to cleanly handle timezone adjustments if someone is on the day before/after UTC's current date

__The Problem__
We do not consistently handle date translations w/r to timezones. We pass the `date_from`, `date_to`, and `timezone` options to Tinybird, which does handle timezone translations, and passes the data back in a localized format. We __do not__ currently do that with Ghost data.

__Solution__
For the moment, we'll centralize utils for translating `date_from` and `date_to` to UTC dates based on the provided `timezone`, and clean up the endpoints by using a centralized filter, ensuring we're applying the date filtering in the same way everywhere.

Ideally, we'll set standards around endpoints always being in UTC, and letting the caller handle the conversion, which feels more appropriate (the client knows its timezone, why does the server need to know it?).

### DIFF
--- a/ghost/core/core/server/api/endpoints/stats.js
+++ b/ghost/core/core/server/api/endpoints/stats.js
@@ -533,12 +533,7 @@ const controller = {
             };
         },
         async query(frame) {
-            return await statsService.api.getTopSourcesWithRange(
-                frame.options.date_from, 
-                frame.options.date_to, 
-                frame.options.order || 'free_members desc',
-                frame.options.limit || 50
-            );
+            return await statsService.api.getTopSourcesWithRange(frame.options);
         }
     }
 
--- a/ghost/core/core/server/services/stats/ReferrersStatsService.js
+++ b/ghost/core/core/server/services/stats/ReferrersStatsService.js
@@ -1,5 +1,8 @@
 const moment = require('moment');
 
+// Import centralized date utilities
+const {getDateBoundaries, applyDateFilter} = require('./utils/date-utils');
+
 // Source normalization mapping - consolidated from frontend apps
 const SOURCE_NORMALIZATION_MAP = new Map([
     // Social Media Consolidation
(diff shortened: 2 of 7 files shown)


ghost/core/core/server/api/endpoints/stats.js
```diff
@@ -533,12 +533,7 @@ const controller = {
...
         },
         async query(frame) {
-            return await statsService.api.getTopSourcesWithRange(
-                frame.options.date_from, 
-                frame.options.date_to, 
-                frame.options.order || 'free_members desc',
-                frame.options.limit || 50
-            );
+            return await statsService.api.getTopSourcesWithRange(frame.options);
         }
```
9larsons, line 537: I collapsed to options here to make it more flexible. This was a bit of an oddball in this way.

ghost/core/core/server/services/stats/utils/date-utils.js
```diff
@@ -0,0 +1,37 @@
+const moment = require('moment-timezone');
```
9larsons, line 1: These utils were pulled out of PostsStatsService so that we could have centralized handling for all stats queries.

getDateBoundaries does the date/datetime -> UTC handling and always does start of/end of day.
applyDateFilter is a helper for `knex` query objects to clean up the queries.

ghost/core/core/server/services/stats/StatsService.js
```diff
@@ -123,6 +123,7 @@ class StatsService {
      * @param {number} [options.limit=20] - Max number of results to return
```
9larsons, line 123: This file is just cleaning up and making sure we're always passing timezone with date_from/date_to. (well... it's not ensuring anything, just making it more clear 😓).

ghost/core/core/server/services/stats/ReferrersStatsService.js
```diff
@@ -252,45 +255,44 @@ class ReferrersStatsService {
...
     /**
      * Fetch deduplicated member counts by source with date range
      * Returns both free signups (excluding those who converted) and paid conversions
-     * @param {string} startDate 
-     * @param {string} endDate 
+     * @param {Object} options
      * @returns {Promise<{source: string, signups: number, paid_conversions: number}[]>}
      **/
-    async fetchMemberCountsBySource(startDate, endDate) {
+    async fetchMemberCountsBySource(options) {
```
9larsons, line 293: This is the only place we use dateTime, but we were already using startOf and endOf, so I don't see this being all that relevant. I think it's just untidy naming conventions.

ghost/core/core/server/services/stats/utils/date-utils.js
```diff
@@ -0,0 +1,37 @@
...
+ * Get processed date boundaries with timezone support
+ * @param {Object} options - Options containing date and timezone info
+ * @param {string} [options.date_from] - Start date in YYYY-MM-DD format
+ * @param {string} [options.date_to] - End date in YYYY-MM-DD format
+ * @param {string} [options.timezone='UTC'] - Timezone for date interpretation
+ * @returns {{dateFrom: string|null, dateTo: string|null}} Processed dates in ISO format
+ */
+function getDateBoundaries(options) {
+    const timezone = options.timezone || 'UTC';
+    const dateFrom = options.date_from ? moment.tz(options.date_from, timezone).utc().startOf('day').toISOString() : null;
```
cmraible: This isn't quite right — we want to get the start of the day in the timezone provided, converted to UTC.

This currently gives us the start of the day in UTC, which may cut off some parts of the day from the local timezone

```suggestion
    const dateFrom = options.date_from ? moment.tz(options.date_from, timezone).startOf('day').utc().toISOString() : null;
```

Same change for the dateTo line below this.

cmraible (changes requested):
The timezone conversion isn't quite right, left a comment explaining

cmraible (approved):
LGTM!

PR #24589, opened 2025-08-02
Fix CTA card visibility preservation during HTML import via Admin API

### LINKED ISSUE #24588: CTA card cannot be imported as html with visibility settings intact
### Issue Summary

I'm trying to create a new post with the Ghost Admin API.  I have a CTA card, that should only be available to free members.
Here's the start of the html I'm passing in:
(removing the comment wrapping because otherwise github removes it upon publication, but it is wrapped like an html comment, honest!)
!--kg-gated-block:begin nonMember:false memberSegment:"status:free" --

`
<div class="kg-card kg-cta-card kg-cta-bg-grey kg-cta-minimal    " data-layout="minimal">
            
            <div class="kg-cta-content">
                
                
                    <div class="kg-cta-content-inner">
                    
                        <div class="kg-cta-text">
                            <p><span style="white-space: pre-wrap;">Thanks for reading. Paying subscribers receive access to each article on Fridays. Free subscribers receive access on the following Monday. Want faster access? Upgrade now!</span></p>
                        </div>
                    
                    
                        <a href="#/portal/account/plans" class="kg-cta-button " style="background-color: #000000; color: #ffffff;">
                            Upgrade your plan
                        </a>
                        
                    </div>
                
            </div>
        </div>
`
!--kg-gated-block:end--

(I got this HTML by inspecting what came from the Admin API so I'm pretty sure I've got it right!)

The CTA card is created correctly, based on my viewing in the editor, EXCEPT that all the visibility toggles are 'on', although they should not be.

### Steps to Reproduce

Here's the function I'm using:
`async function createNewsletterDraft(html, dateRange, options = {}) {
  const api = getAdminAPI();
  
  const title = options.title || `Newsletter - ${dateRange.after} to ${dateRange.before}`;
  const tags = options.tags || ['#newsletter-draft'];
  
  try {  
    let result;
    
      // Create new draft
      const post = {
        title: title,
        html: html,
        status: 'draft',
        tags: tags,
        custom_excerpt: `Newsletter covering ${dateRange.after} to ${dateRange.before}`
      };
      
      result = await api.posts.add(post, { source: 'html' });

    
    return result;
  } catch (error) {
    console.error('Error creating newsletter draft:', error);
    throw error;
  }
}`

### Ghost Version

latest / Ghost Pro

### Node.js Version

Ghost Pro

Ghost Pro

### Database type

MySQL 5.7

### Browser & OS version

_No response_

```shell
No errors thrown by my function. Just not getting the expected result.
```

### Code of Conduct

amyanger:
Fixes #24588

When importing HTML content with CTA cards wrapped in gated blocks via the Admin API using `{source: 'html'}`, the visibility settings were being lost during HTML-to-Lexical conversion. This caused all CTA visibility toggles to show as "on" (visible to everyone) instead of preserving the intended restrictions like "free members only".

Root cause: The `@tryghost/kg-html-to-lexical` converter strips HTML comments where gated block visibility settings are stored (e.g., `<\!--kg-gated-block:begin nonMember:false memberSegment:"status:free"-->`).

Solution:
- Created `gated-block-visibility.js` utility to extract visibility settings from gated block comments before HTML-to-Lexical conversion
- Modified posts and pages serializers to preserve visibility during conversion
- Added comprehensive tests covering various scenarios

The fix ensures that CTA cards (and other supported cards) maintain their visibility restrictions when imported via Admin API, resolving the issue where users would see incorrect "visible to everyone" settings instead of their intended member segment targeting.

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/serializers/input/pages.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/input/pages.js
@@ -11,6 +11,7 @@ const postsSchema = require('../../../../../data/schema').tables.posts;
 const clean = require('./utils/clean');
 const lexical = require('../../../../../lib/lexical');
 const sentry = require('../../../../../../shared/sentry');
+const {preserveGatedBlockVisibility} = require('./utils/gated-block-visibility');
 
 const messages = {
     failedHtmlToMobiledoc: 'Failed to convert HTML to Mobiledoc',
@@ -190,7 +191,9 @@ module.exports = {
                 }
 
                 try {
-                    frame.data.pages[0].lexical = JSON.stringify(lexical.htmlToLexicalConverter(html));
+                    const lexicalDoc = lexical.htmlToLexicalConverter(html);
+                    const lexicalWithVisibility = preserveGatedBlockVisibility(html, lexicalDoc);
+                    frame.data.pages[0].lexical = JSON.stringify(lexicalWithVisibility);
                 } catch (err) {
                     sentry.captureException(err);
                     throw new ValidationError({
--- a/ghost/core/core/server/api/endpoints/utils/serializers/input/posts.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/input/posts.js
@@ -11,6 +11,7 @@ const postsSchema = require('../../../../../data/schema').tables.posts;
 const clean = require('./utils/clean');
 const lexical = require('../../../../../lib/lexical');
 const sentry = require('../../../../../../shared/sentry');
+const {preserveGatedBlockVisibility} = require('./utils/gated-block-visibility');
 
 const messages = {
     failedHtmlToMobiledoc: 'Failed to convert HTML to Mobiledoc',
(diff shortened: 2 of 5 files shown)


kevinansfield:
@amyanger thank you for looking into this. Unfortunately I think there's been a bit of confusion around where the gated blocks handling should occur as this is a Koenig responsibility rather than a Ghost one.

Each node has a parser in the [`kg-default-nodes` package](https://github.com/TryGhost/Koenig/tree/main/packages/kg-default-nodes). Parsers for the two cards that support visibility can be found in [html-parser.js](https://github.com/TryGhost/Koenig/blob/main/packages/kg-default-nodes/lib/nodes/html/html-parser.js) and [calltoaction-parser.js](https://github.com/TryGhost/Koenig/blob/main/packages/kg-default-nodes/lib/nodes/call-to-action/calltoaction-parser.js). Any utility to extract the visibility properties of a node when parsing from HTML should occur in that package.

I'm going to close this PR to keep the PR list clean as it's not mergeable as-is. Please feel free to open a PR on the Koenig repo if you decide to tackle this again 🙏

PR #24593, opened 2025-08-03
Fixed members header disappearing during list load

kevinansfield:
closes https://linear.app/ghost/issue/PROD-2391/

- the members list uses `<GhLoadingSpinner />` which creates an element that covers it's parent in order to position in the middle, however it also has a white background so will cover anything that isn't sitting above it in the stacking context, this was resulting in the search and filter disappearing each time you made a change whilst the list refreshes
- added `z-index` to the header content so it has it's own stacking context and sits above the loading spinner

### DIFF
--- a/ghost/admin/app/styles/layouts/members.css
+++ b/ghost/admin/app/styles/layouts/members.css
@@ -98,6 +98,10 @@
     margin: 0 -48px;
 }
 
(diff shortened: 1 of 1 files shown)


9larsons (approved):
Works nicely with 2M members 👍 .

PR #24599, opened 2025-08-04
Fixed posts list newsletter analytics hover logic

9larsons:
ref https://linear.app/ghost/issue/PROD-2386/post-newsletter-analytics-hover-doesnt-respect-one-flag-disabled
- updated show/hide logic to respect settings for clicks/opens
 - added fallback to sent when both clicks+opens disabled

 W/r to the issue, this fixes the logic in the Posts List component. I'll tackle the Top Posts updates separately.

 The logic is as follows, for posts with email data:
 - Show clicks column if click tracking enabled
 - Show open column is open tracking enabled
 - Show sent column if open tracking and click tracking are disabled

 Tooltip behavior should respect the settings as well:
 - Sent always shown
 - Clicks shown if click tracking enabled
 - Opens shown if open tracking enabled

 The idea here is that Sent data should always be shown, and the other two should only be shown if the corresponding setting is enabled. There's some slight differences here with number of columns vs. Top Posts which is why I'm separating it out, as that component has less horizontal space to work with.

### DIFF
--- a/ghost/admin/tests/acceptance/content-test.js
+++ b/ghost/admin/tests/acceptance/content-test.js
@@ -1080,7 +1080,7 @@ describe('Acceptance: Posts / Pages', function () {
                 
                 // Verify sent column appears with proper formatting
                 expect(find('[data-test-analytics-sent]'), 'sent column').to.exist;
-                expect(find('[data-test-analytics-sent] .gh-content-email-stats-value').textContent.trim()).to.equal('15,000');
+                expect(find('[data-test-analytics-sent] .gh-content-email-stats-value').textContent.trim()).to.equal('15k');
                 expect(find('[data-test-analytics-opens]'), 'opens column when disabled').to.not.exist;
                 expect(find('[data-test-analytics-clicks]'), 'clicks column when disabled').to.not.exist;
             });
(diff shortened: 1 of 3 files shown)


9larsons:
@cmraible FYI I worked for a bit on adding more complete tests. Getting Mirage to appropriately mock the settings turns out to be quite difficult (at least it has been stumping me), so I added the minimum cases that only depend on the post model data and moved on.

ghost/admin/tests/acceptance/content-test.js
```diff
@@ -1002,6 +1002,89 @@ describe('Acceptance: Posts / Pages', function () {
...
+                    // Override computed properties for testing
+                    showEmailOpenAnalytics: false,
+                    showEmailClickAnalytics: false
+                });
+
+                await visit('/posts');
+                
+                // Verify sent column appears with proper formatting
+                expect(find('[data-test-analytics-sent]'), 'sent column').to.exist;
+                expect(find('[data-test-analytics-sent] .gh-content-email-stats-value').textContent.trim()).to.equal('15,000');
```
cmraible: I think this should be `15k` per @peterzimon but I'm not 100% sure that's what he meant by "make sure it's formatted with thousand indicators".

peterzimon: You're right, we should abbreviate it, e.g. `15k` — then users can see the exact number in the tooltip and in the post analytics page.

PR #24601, opened 2025-08-04
Fixed newsletter metric display in Analytics > Top Posts

9larsons:
ref https://linear.app/ghost/issue/PROD-2386/post-newsletter-analytics-hover-doesnt-respect-one-flag-disabled ref https://github.com/TryGhost/Ghost/pull/24599
- updated display logic in Top Posts

Building on the other PR (see ref), this updates the Newsletter analytics column in the Top Posts component. Logic is very similar to the Posts List column, except there's one fewer column in Top Posts, as there's less horizontal space here.

Column display:
- Hidden if no post email data
- Show open rate if open rate tracking enabled
- Show click rate if open rate tracking disabled *and* click rate tracking enabled
- Show sent count if open rate tracking and click rate tracking are disabled

Tooltip:
- Always display Sent count
- Display click count if click tracking enabled
- Display open rate if open rate enabled

### DIFF
--- a/apps/stats/src/views/Stats/Overview/components/TopPosts.tsx
+++ b/apps/stats/src/views/Stats/Overview/components/TopPosts.tsx
@@ -180,7 +180,7 @@ const TopPosts: React.FC<TopPostsProps> = ({
                                                                 return (
                                                                     <>
                                                                         <LucideIcon.Send className='text-muted-foreground group-hover:text-foreground' size={16} strokeWidth={1.5} />
-                                                                        {formatNumber(post.sent_count || 0)}
+                                                                        {abbreviateNumber(post.sent_count || 0)}
                                                                     </>
                                                                 );
                                                             }
--- a/apps/stats/src/views/Stats/Overview/components/TopPosts.tsx
+++ b/apps/stats/src/views/Stats/Overview/components/TopPosts.tsx
@@ -63,7 +63,8 @@ const TopPosts: React.FC<TopPostsProps> = ({
 
     // Show open rate if newsletters are enabled and email tracking is enabled
     const showWebAnalytics = appSettings?.analytics.webAnalytics;
-    const showOpenRate = appSettings?.newslettersEnabled && appSettings?.analytics.emailTrackOpens;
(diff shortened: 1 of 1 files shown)


cmraible:
suggestion (maybe bug?) - non-blocking: in a perfect world, I think we would not show the clicks in the tooltip if the _post was sent_ while clicks were disabled. The post in the attached screenshot was sent when clicks were disabled, and it shows 0 clicks, instead of hiding the datapoint. This would (I think) also require some backend changes, to pull in the `track_opens` and `track_clicks` value from the `emails` table. I wouldn't block this PR on it, since I think this is better than the current behavior, but just wanted to flag it in case we collectively feel that's important to get right. 

Screenshot 2025-08-04 at 18 40 29

apps/stats/src/views/Stats/Overview/components/TopPosts.tsx
```diff
@@ -126,35 +127,64 @@ const TopPosts: React.FC<TopPostsProps> = ({
...
                                                     navigate(`/posts/analytics/${post.post_id}/newsletter`, {crossApp: true});
                                                 }}>
                                                     <PostListTooltip
                                                         className={`${!appSettings?.analytics.membersTrackSources ? 'left-auto right-0 translate-x-0' : ''}`}
                                                         metrics={[
+                                                            // Always show sent
                                                             {
                                                                 icon: <LucideIcon.Send className='shrink-0 text-muted-foreground' size={16} strokeWidth={1.5} />,
                                                                 label: 'Sent',
                                                                 metric: formatNumber(post.sent_count || 0)
```
cmraible, line 142: Same comment here, not sure if we want to abbreviate the number here, i.e. `15k`

9larsons:
@cmraible Good thought - updated to use abbreviateNumber. 

As for the email data, I think that's a reasonable but separate issue/fix. [PROD-2400: Respect `track_opens` and `track_clicks` when displaying open and click data](https://linear.app/ghost/issue/PROD-2400/respect-track-opens-and-track-clicks-when-displaying-open-and-click)

PR #24614, opened 2025-08-06
Added factory frame to E2E package

9larsons:
ref https://linear.app/ghost/issue/PROD-2327/create-e2e-test-data-factory
- added factory base class
- added post factory
- added persistence adapter class
- added knex adapter
- added sample test cases for functional post factory

This PR adds the basic building blocks for data factories to be used for the e2e package, and possibly extended to other areas. There's a few key design decisions:
1. `Factory`s should be scoped to a single model, and only be responsible for creation
2. We do not need to handle tracking created data; the intent is a 'clean slate' for each test (which is a requirement for parallelism)
3. Persistence should be optional. `build` and `create` specify creating in memory and persisting an object.
4. Persistence should be decoupled from factories; use of an adapter pattern allows for easier refactoring should we need to (e.g. move from direct db access to API).
5. `Factory` and `PersistenceAdapter` need to be generalizable for all future data sources.

The overarching goal of this is to enable __easy test writing__.

We will further refine this by adding additional factories and data sources in future PRs, and we may go down the road of wrapping `PostFactory`, `UserFactory`, etc. in a Ghost handler/coordinator, which is straightforward change with this structure.


e2e/data-factory/factories/posts/post-factory.ts
```diff
@@ -0,0 +1,96 @@
...
+    name = 'post'; // Factory identifier
+    entityType = 'posts'; // Entity name (for adapter; currently the database table)
+    
+    build(options: Partial<Post> = {}): Post {
+        const now = new Date();
+        const title = options.title || faker.lorem.sentence();
+        const content = faker.lorem.paragraphs(3);
+        
+        // Generate mobiledoc format
+        const mobiledoc = {
```
9larsons, line 52: Arguably this should build Lexical. Not sure we care until we work on editor- or content-related tests.

cmraible (changes requested):
This looks really good to me! I think this is a really solid foundation to start building on, and I don't have any notes on the factory code itself. 

My only note: I think we should use the post-factory in a real test, that:
1) Creates a published post, using the factory
2) Visits the Ghost frontend and asserts the post exists

Two main reasons for this:
1) I think having tests for our tests is confusing — I think we should delete the `e2e/data-factory/tests` directory
2) I want to make sure we don't have any kind of split brain problem — I'm a little confused about why we needed to change the database client, where we didn't before — just want to make sure the factories are using the same database as Ghost in CI, basically.

Other than that, good to merge from my POV!

9larsons:
@cmraible It seems we need some more available within Ghost to be able to do this with a browser test. We see a 404 when trying to load it on the frontend. I believe it's the same issue we see when trying to load the post - we don't have an author. 

I would rather ignore doing that here, because we would have to introduce another relation/factory. Alternatively, we could log in to Admin and view the Posts List, where it does show, but then we have setup/auth to deal with, which feels like another distraction. 

On the whole I agree w/ the suggestion, but for this set, I'm happy w/ simply testing the factory works w/ itself.

cmraible:
Okay I had a little play with adding a test locally, and found a potential concern: visiting the post is 404'ing because of the way Ghost caches routes — it calculates the routes for each post on boot, then it updates the routes based on internal events. So when the Ghost server is running, and we insert a post into the DB, the routes don't exist.

I'm wondering if it would make more sense to use the API rather than using the DB directly. I was ambivalent on DB vs API before, but this feels like a good reason to use the API (where possible).

e2e/data-factory/factories/posts/post-factory.ts
```diff
@@ -0,0 +1,96 @@
...
+            cards: [],
+            markups: [],
+            sections: [[1, 'p', [[0, [], 0, content]]]],
+            ghostVersion: '5.0'
+        };
+        
+        // Create defaults object
+        const defaults: Post = {
+            id: generateId(),
+            uuid: generateUuid(),
```
ibalosh, line 64: Any reason for not using `faker.string.uuid()`, instead of custom generator? Since we already use `faker` anyways, that is one custom util method less we need to worry about, by using built-in.

e2e/data-factory/factories/posts/post-factory.ts
```diff
@@ -0,0 +1,96 @@
+import {Factory} from '../factory';
```
ibalosh, line 1: `nit:` What was the decision behind adding folder for post factory, do we plan to group more post related factories here?

e2e/data-factory/factories/factory.ts
```diff
@@ -0,0 +1,29 @@
+import type {PersistenceAdapter} from '../persistence/adapter';
+
+export abstract class Factory<TOptions extends Record<string, unknown> = Record<string, unknown>, TResult = TOptions> {
+    abstract name: string;
```
ibalosh, line 4: I think we should make identification automatic, like [here](https://github.com/TryGhost/Ghost/pull/24604/files#diff-e25bd2f5a415c5d42043c5d43c3e686c8166cd55fd624ab4217af8a7423ae3c5) , with `return this.constructor.name;`.

It's one less thing to think about when creating new factories. The less manual work we need to do, the better.

e2e/data-factory/factories/factory.ts
```diff
@@ -0,0 +1,29 @@
...
+    abstract entityType: string;
```
ibalosh, line 5: `nit nit:` entityType does work and it's correct, I am wondering though.., since entityType is related to adapter, can we use a more descriptive name to describe that, maybe something like `adapterTarget` ? 

Since we need a comment in `PostFactory` to explain it, thats a sign we should update it, if we can. (if we have to have a comment, we should add it on parrent class I think, so it's easy to find and read on each new factory we create)

maybe it could be even in that class something like:

```typescript
  persiscence = {
        adapter?: PersistenceAdapter,
        entity: string
    };
```

e2e/data-factory/factories/factory.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Build and persist an entity to the database
+     */
+    async create(options?: Partial<TOptions>): Promise<TResult> {
+        if (!this.adapter) {
+            throw new Error('Cannot create without a persistence adapter. Use build() for in-memory objects.');
+        }
+        const data = this.build(options);
+        return this.adapter.insert(this.entityType, data) as Promise<TResult>;
```
ibalosh: missing `await` here?  

(return await this.adapter.insert(this.entityType, data) as Promise<TResult>;)

e2e/data-factory/factories/factory.ts
```diff
@@ -0,0 +1,29 @@
+import type {PersistenceAdapter} from '../persistence/adapter';
+
+export abstract class Factory<TOptions extends Record<string, unknown> = Record<string, unknown>, TResult = TOptions> {
+    abstract name: string;
+    abstract entityType: string;
+    
+    protected adapter?: PersistenceAdapter;
+
+    setAdapter(adapter: PersistenceAdapter): void {
```
ibalosh: `nit:`

We could make this a constructor, which makes it one liner to use later, since we will have bunch of factories:

```typescript
   constructor(adapter?: PersistenceAdapter) {
        this.persistence = adapter;
    }
```

and then just do:

```typescript
postFactory = new PostFactory(adapter);
```

e2e/data-factory/factories/posts/post-factory.ts
```diff
@@ -0,0 +1,96 @@
...
+            newsletter_id: null,
+            show_title_and_feature_image: true
+        };
+        
+        // Merge with user options, handling special cases
+        const post: Post = {
+            ...defaults,
+            ...options,
+            // Handle published_at logic - if status is published but no published_at is set, use current time
+            published_at: options.status === 'published' && !options.published_at ? now : (options.published_at || defaults.published_at)
```
ibalosh: `nit:` could we make this simpler? It's hard to understand, plus we should be able to rewrite it in a way that comment is not needed.

e2e/data-factory/tests/post-factory.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        const post = postFactory.build({
+            title: 'Test Post',
+            status: 'draft'
+        });
+        
+        expect(post.id).toBeTruthy();
+        expect(post.title).toBe('Test Post');
+        expect(post.status).toBe('draft');
+        expect(post.slug).toBeTruthy();
+        expect(post.html).toBeTruthy();
```
ibalosh, line 57: I think we don't need to test these, we will exercise the tests anyways in e2e tests. We could leave the happy path one, but not really needed - we will duplicate most of what we will do in e2e test

ibalosh (changes requested):
Hey @9larsons added some small suggestions to check out, mostly nitpicks, other than couple of things that should be checked out, this looks great!

ibalosh:
> Okay I had a little play with adding a test locally, and found a potential concern: visiting the post is 404'ing because of the way Ghost caches routes — it calculates the routes for each post on boot, then it updates the routes based on internal events. So when the Ghost server is running, and we insert a post into the DB, the routes don't exist.
> 
> I'm wondering if it would make more sense to use the API rather than using the DB directly. I was ambivalent on DB vs API before, but this feels like a good reason to use the API (where possible).

@cmraible the more high level we can make it(as long as it's not more work for us) the better , +1 for API as long as it's not going to be much slower, still we can support our parallelization plan and it's not more tedious to do cleanups

e2e/data-factory/factories/posts/post-factory.ts
9larsons: Originally this had types separated. It's a fair point now - I'll remove it and bump it up a level.

e2e/data-factory/factories/factory.ts
cmraible: Nice, I like this too

cmraible:
> the more high level we can make it(as long as it's not more work for us) the better , +1 for API as long as it's not going to be much slower, still we can support our parallelization plan and it's not more tedious to do cleanups

I think as long as we're not inserting tons of data (i.e. hundreds/thousands of posts), the API won't be a major performance drawback — although I agree this is the biggest tradeoff of using the API vs the DB directly.

On the other hand:
- Using the API is more "end-to-end" — that's how posts are created IRL (via the UI, but ultimately using the API). I think I had this backwards initially: using the API is actually _less_ coupled to implementation details than using the database. The 404 error we see here is a good example of this. 
- Using the API simplifies the logic in the factory — i.e. we don't need to reinvent the `published_at` logic, because Ghost does that internally
- It shouldn't impact parallelization — we just target a different Ghost instance by url/port, instead of targeting a different database entirely
- It shouldn't impact clean up — this will still need to happen at the DB level regardless

To be clear, I expect there will be cases where we _need_ to directly insert data into the DB, so there's no wasted effort here — but I think there's a lot to be gained from using the API by default, and falling back to direct DB when necessary.

ibalosh:
> > the more high level we can make it(as long as it's not more work for us) the better , +1 for API as long as it's not going to be much slower, still we can support our parallelization plan and it's not more tedious to do cleanups
> 
> I think as long as we're not inserting tons of data (i.e. hundreds/thousands of posts), the API won't be a major performance drawback — although I agree this is the biggest tradeoff of using the API vs the DB directly.
> 
> On the other hand:
> 
> * Using the API is more "end-to-end" — that's how posts are created IRL (via the UI, but ultimately using the API). I think I had this backwards initially: using the API is actually _less_ coupled to implementation details than using the database. The 404 error we see here is a good example of this.
> * Using the API simplifies the logic in the factory — i.e. we don't need to reinvent the `published_at` logic, because Ghost does that internally
> * It shouldn't impact parallelization — we just target a different Ghost instance by url/port, instead of targeting a different database entirely
> * It shouldn't impact clean up — this will still need to happen at the DB level regardless
> 
> To be clear, I expect there will be cases where we _need_ to directly insert data into the DB, so there's no wasted effort here — but I think there's a lot to be gained from using the API by default, and falling back to direct DB when necessary.

agreed 100% @cmraible

9larsons:
I'm absolutely fine with using the API where we can (and posts is a place where we can). However, it does make it more difficult to provide a slim proof-of-concept like we have here, and we would still need the KnexAdapter for some use cases. Primarily, we need to know the Admin API keys in order to use the API and coordinate that piece.

I'd argue we don't need to block this based on the API use. If wanted, we can pare it down to a simpler use case (just demonstrate creating a post works), and we can follow up with a separate PR to implement the API handling.

cmraible (approved):
Good to merge from my POV! I think this is a solid starting point that we can iterate on as we start to write tests. 

Thanks for leading the charge here @9larsons, nice work!

9larsons:
@ibalosh I've changed the suggestions you had (all good ones, thanks!). I simply resolved the convos instead of responding to each. Mind taking a pass and approving when you feel it's good?

e2e/data-factory/README.md
```diff
@@ -0,0 +1,119 @@
...
+│   ├── factory.ts         # Base factory class
+│   └── posts/
+│       └── post-factory.ts # Post factory implementation
+├── persistence/
+│   ├── adapter.ts         # Persistence interface
+│   └── adapters/
+│       └── knex.ts        # Knex SQL adapter
+├── index.ts               # Main exports
+├── utils.ts               # Utility functions
+└── test/
```
ibalosh: **nit** folder is `tests`

ibalosh (approved):
I could add  one nitpick (like I would love if we ditch [should in tests](https://github.com/spotify/should-up) :) ) , but it would not make sense to fiddle with that here.

Ready for merge in my opinion too, nice one @9larsons ! 👌

9larsons:
Those tests will be scrapped, they are just there for proof of concept. Fixed the readme and I'll merge 👍 .

PR #24622, opened 2025-08-07
Fix broken analytics tests

ibalosh:
recreated to make this [PR which was a fork before pass](https://github.com/TryGhost/Ghost/pull/24190).

* extracted route regexp creation from mockApi function, to make it easier to understand that these are used to determin admin, activity-pub or tinybird API route
* created factories for fixture responses to use with custom data for tests
* removed usage of real tinybird for now, to get these tests running
* reused e2e page-objects added at root folder, which can be used by all tests that use playwright
* added couple of more tests


ibalosh:
hey @cmraible @9larsons , recreating [this PR](https://github.com/TryGhost/Ghost/pull/24190) just to make it pass (so it's not a fork, so CI likes it).

You guys ok, if we merge in this one? I know that we still need to align on acceptance (and I am for moving into another direction with them, to get more value and less tech debt), but the reason why I think we should merge this one:

* it adds more flexibility with option to mock tinybird API with playwright, which we might still need for some cases later on
* it is cleaner than current main
  * removes some number of files compared to original code on main branch  
  * removes temporary API helpers I used before
  * tests will pass now
* it does very basic rendering testing for three of our analytics pages, until we have integration tests, so we can use these

cmraible (changes requested):
Generally I have no objection to merging this, but I think we're missing two key things:
1. Documentation on how to run these tests locally in the README.md. I have gotten them to run, but they are failing for me locally. I imagine I'm missing some required setup, but it's not clear what that setup is.

2. We should run these tests in CI, otherwise we'll continue to break them accidentally since we're never really enforcing that they pass. 

I'm happy to help with the latter, but I'll need to get them working locally myself first 😃

ibalosh:
hey @cmraible 

let's catchup today later on call, there should not be any setup to run them locally. 

I will check it with fresh copy today and see what's going on. (I did, could not reproduce the issue locally, even on a fresh setup, `yarn test:acceptance` for this branch worked like charm)

ibalosh:
**note:** we postponed this for now

PR #24654, opened 2025-08-12
Updated `esm` patch for Node >= 22.18.x support

kevinansfield:
no issue

- Node 22.18.0 has changed it's `internalModuleStat` args again so our `esm` patch needed updating so it only changes the expected arguments for the Node versions where the function has two arguments
- https://github.com/wallabyjs/esm/commit/728da8e19a18ca080778dd6b5f5a4d6250bce0ee


sam-lord (approved):
Tested on node 22.16.1, 22.17.1, and 22.18.0, approved :)

PR #24663, opened 2025-08-13
Admin analytics empty state tests and poms

ibalosh:
Initial empty states analytics tests and helpers 

* config, user setup, is in utils
* added user setup in utils to create user, and store session cookies, to avoid signup multiple times
  * create user by API first in db
    * added a simple user factory for building, will switch that code to use our factories, once we add them
  * once user is created playwright setup code is executed only on first run to store the cookie for logged in user
  * this is an initial way we can test with logged in user admin, we can adjust later
  * used factories for user
* added analytics base page objects 
* added helper locators on components to support them
* added tests for empty states

**note:** sorry for bit longer PR, but most of it is just simple page objects, locators, and very basic tests


e2e/helpers/utils/app-config.ts
```diff
@@ -0,0 +1,15 @@
+import dotenv from 'dotenv';
+// load environment variables from .env file
+dotenv.config();
+
+// Simple config object with just the values
+export const appConfig = {
+    baseURL: process.env.GHOST_BASE_URL || 'http://localhost:2368',
+
+    auth: {
+        email: process.env.E2E_ACCOUNT_USERNAME || '',
```
cmraible: suggestion: I think it'd be nice to set a default value here (or somewhere else). I tried running the tests without these environment variables, and I got a "Your password must be at least 10 characters long" error. It's nice to have the option of an override, but ideally I think these should run successfully with zero config needed.

cmraible:
I'm struggling to get these passing locally 🤔 

The initial authenticate test passes, but then all the rest of the tests fail with a 403 when trying to access admin. 

I just did:
- `yarn docker:reset` to clear the database
- `yarn dev` to run Ghost (with analytics and tinybird already configured and running)
- `cd e2e && yarn test`

Any ideas? I wonder if this is related to the [wait](https://github.com/TryGhost/Ghost/blob/2d6cee844026a43d3e1169108abc29128f3adf23/e2e/tests/auth.setup.ts#L31) you mentioned in the description. We can talk more and hopefully troubleshoot together tomorrow 😃 

Screenshot 2025-08-18 at 17 42 17

e2e/tests/admin/analytics/growt.test.ts
```diff
@@ -0,0 +1,37 @@
+import {test, expect} from '@playwright/test';
```
cmraible: nit: the filename has a typo 😄

e2e/helpers/utils/app-config.ts
ibalosh: good catch! you are right, I added defaults here, and removed the need in GH actions, and just use defaults

e2e/tests/admin/analytics/growt.test.ts
ibalosh: oops :) , fixed

ibalosh:
hey @9larsons and @cmraible , this should be good to go, passed more than 10 times in CI, lmk if you would like something adjusted in the suite

cmraible:
Sounds good, I'll do a final pass on this today now that we've got everything working!

e2e/helpers/pages/admin/analytics/post-analytics/PostAnalyticsPage.ts
```diff
@@ -0,0 +1,68 @@
...
+        this.overviewButton = this.page.getByRole('button', {name: 'Overview'});
+        this.webTrafficButton = this.page.getByRole('button', {name: 'Web traffic'});
+        this.growthButton = this.page.getByRole('button', {name: 'Growth'});
+
+        this.growthSection = new GrowthSection(page);
+        this.webPerformanceSection = new WebPerformanceSection(page);
+    }
+
+    async waitForPageLoad() {
+        await this.webPerformanceSection.viewMoreButton.waitFor({state: 'visible'});
```
cmraible: just a question: why wait for the "View More" button specifically? Only asking because the view more button only shows on hover, else it's set to `opacity: 0`. It seems like `waitFor({state: 'visible'})` doesn't care that it's opacity is zero since this is working, but I wonder if there might be an earlier event we could trigger this on.

e2e/tests/admin/analytics/post-analytics/growth.test.ts
```diff
@@ -0,0 +1,41 @@
...
+test.describe('Ghost Admin - Post Analytics - Growth', () => {
+    let postAnalyticsPageGrowthPage: PostAnalyticsGrowthPage;
+
+    test.beforeEach(async ({page}) => {
+        const analyticsOverviewPage = new AnalyticsOverviewPage(page);
+        await analyticsOverviewPage.goto();
+        await analyticsOverviewPage.latestPost.viewAnalytics();
+
+        const postAnalyticsPage = new PostAnalyticsPage(page);
+        // TODO: check post analytics component, we shouldn't need to wait on page load to be able to click growth link
```
cmraible, line 15: note: I think what might be happening here is that we're switching from the `stats` app to the `posts` app for the first time. The bundle for each react app is lazy loaded when first switching to the corresponding Ember route. Definitely room to improve that load time potentially. Mainly sharing in case this helps find a way to avoid the `waitForPageLoad()`.

e2e/tests/admin/analytics/post-analytics/overview.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    PostAnalyticsGrowthPage,
+    PostAnalyticsWebTrafficPage
+} from '../../../../helpers/pages/admin';
+
+test.describe('Ghost Admin - Post Analytics - Overview', () => {
+    test.beforeEach(async ({page}) => {
+        const analyticsOverviewPage = new AnalyticsOverviewPage(page);
+        await analyticsOverviewPage.goto();
+
+        await analyticsOverviewPage.latestPost.viewAnalytics();
```
cmraible: nit (non-blocking): I think we're mixing patterns a little bit. Here we have a method for clicking a button, while in most places the pattern is like below, `postAnalyticsPage.growthSection.viewMoreButton.click()`, which I _think_ I prefer as it's more explicit. By naming this method it makes it seem important, so I wonder what it's doing — but really it's just clicking a button.

e2e/tests/admin/analytics/post-analytics/overview.test.ts
```diff
@@ -0,0 +1,46 @@
...
+        await expect(postAnalyticsPage.growthSection.card).toContainText('0');
+    });
+
+    test('empty page - overview - web performance - view more', async ({page}) => {
+        const postAnalyticsPage = new PostAnalyticsPage(page);
+        await postAnalyticsPage.growthSection.viewMoreButton.click();
+
+        const postAnalyticsWebTrafficPage = new PostAnalyticsWebTrafficPage(page);
+        await expect(postAnalyticsWebTrafficPage.body).toContainText('No sources data available');
+    });
```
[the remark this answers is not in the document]
cmraible: I agree with the rabbit, it looks like the title was updated to web performance, but the implementation of the test is the same as the growth test below this.

e2e/tests/admin/analytics/growth.test.ts
```diff
@@ -0,0 +1,37 @@
...
+        await expect(growthPage.topContent.contentCard).toContainText('No conversions');
+    });
+
+    test('empty top content card - sources', async () => {
+        await growthPage.topContent.sourcesButton.click();
+
+        await expect(growthPage.topContent.contentCard).toContainText('How readers found your site in the last 30 days');
+        await expect(growthPage.topContent.contentCard).toContainText('No conversions');
+    });
+});
```
cmraible, line 37: 😍 this whole file is so nice

e2e/tests/admin/analytics/location.test.ts
```diff
@@ -0,0 +1,11 @@
...
+import {AnalyticsLocationsPage} from '../../../helpers/pages/admin';
+
+test.describe('Ghost Admin - Locations', () => {
+    test('empty sources card', async ({page}) => {
+        const locationsPage = new AnalyticsLocationsPage(page);
+        await locationsPage.goto();
+
+        await expect(locationsPage.visitorsCard).toContainText('No visitors');
+    });
+});
```
cmraible, line 11: so good ❤️

e2e/tests/admin/analytics/overview.test.ts
```diff
@@ -0,0 +1,50 @@
+import {test, expect} from '@playwright/test';
+import {
+    AnalyticsOverviewPage,
+    AnalyticsWebTrafficPage,
+    AnalyticsGrowthPage, LoginPage
```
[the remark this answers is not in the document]
cmraible: non-blocking, unrelated: do we not have linting enabled for our tests? Usually this would be flagged by the linter anywhere in our source code, I wonder if we have things setup incorrectly in this package 🤔

e2e/tests/admin/analytics/overview.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        expect(membersCount).toContain('0');
+    });
+
+    test('top posts', async ({page}) => {
+        const analyticsOverviewPage = new AnalyticsOverviewPage(page);
+        await analyticsOverviewPage.goto();
+
+        await expect(analyticsOverviewPage.topPosts.post).toBeVisible();
+        await expect(analyticsOverviewPage.topPosts.post).toContainText('Unique visitors0');
+        await expect(analyticsOverviewPage.topPosts.post).toContainText('Free0');
```
cmraible: nit (non-blocking): not sure if I like combining the label and value assertions into one here, vs explicitly asserting on the count as in the test above this. It's fairly harmless here, but I can see someone (like myself 😄 ) copy/pasting this for another test some time in the future and then we have `.toContainText('Unique visitors12,452');` which is harder on the eyes

e2e/.env.example
```diff
@@ -0,0 +1,2 @@
+E2E_ACCOUNT_USERNAME=example
```
cmraible: nit: I think this should be email to make it clear what kind of value is expected.

```suggestion
E2E_ACCOUNT_EMAIL=example
```

e2e/package.json
```diff
@@ -11,10 +11,13 @@
...
-    "test": "playwright test",
-    "test:factory": "playwright test data-factory",
+    "test": "playwright test --project=main",
+    "test:parallel": "playwright test --project=main --workers=8",
+    "test:factory": "playwright test --project=factories",
     "test:types": "tsc --noEmit",
-    "lint": "eslint helpers --ext .ts --cache"
+    "test:setauth": "npx playwright test --project=setup",
+    "lint": "eslint helpers --ext .ts --cache",
+    "setup:user": "ts-node helpers/utils/setup-user.ts"
```
cmraible: non-blocking: are `test:setauth` and `setup:user` still necessary? I think these are a holdover from before we had this part automated in the tests themselves.

e2e/package.json
```diff
@@ -26,7 +29,10 @@
     "knex": "3.1.0",
     "mysql2": "3.14.3",
     "ts-node": "10.9.2",
-    "typescript": "5.8.3"
+    "typescript": "5.8.3",
+    "dotenv": "^16.6.1"
```
cmraible: nit: we generally pin our dependencies, which I think is enforced by renovate. it's harmless, but if we merge this as-is, renovate will just immediately raise another PR to pin this dependency.

e2e/tests/auth.setup.ts
```diff
@@ -0,0 +1,43 @@
+import {test as setup, expect} from '@playwright/test';
+import {AnalyticsOverviewPage, LoginPage} from '../helpers/pages/admin';
+import * as path from 'node:path';
+import {existsSync, readFileSync} from 'fs';
+import {appConfig, setupUser} from '../helpers/utils';
+
+const authFile = path.join(__dirname, '../playwright/.auth/user.json');
```
cmraible: note (non-blocking): this pattern works for now, but I think this will break when we move to fully isolated parallelism (i.e. nuking the database between tests). I think we may want to switch to using [fixtures](https://playwright.dev/docs/test-fixtures#execution-order) before too long

cmraible (approved):
Overall looking really good! Love how clean the tests are, they're so easy to read. 

Left a bunch of comments but mostly small stuff and some questions - nothing that I think is blocking.

e2e/tests/admin/analytics/post-analytics/overview.test.ts
ibalosh: Agreed. I thought I will need to wait there for something on a page, so I wrapped it in a method. 
Reverted back to a button. 

If we would need a wrapper, we can go by something like `viewMoreButtonClick()`, to keep things consistent.

e2e/tests/admin/analytics/overview.test.ts
ibalosh: agreed, added locators to make this one look good too :) it kinda worked, but if order changes of elements, my approach would fail

e2e/package.json
ibalosh: yep, removed both, tx!

e2e/package.json
ibalosh: gotcha, yep, makes sense, I adjusted the versions. I think too that pinning dependency is better

e2e/tests/admin/analytics/post-analytics/overview.test.ts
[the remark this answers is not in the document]
ibalosh: 👍 updated the test

e2e/tests/admin/analytics/overview.test.ts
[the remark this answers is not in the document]
ibalosh: good point, linter was just checking helpers 🙈 updated to check e2e folder, we can narrow it down if we will need it later

e2e/tests/auth.setup.ts
ibalosh: Fair point, and something to keep in mind when we switch. I'll leave it as is for now, but def will update along the way 👍.

e2e/helpers/pages/admin/analytics/post-analytics/PostAnalyticsPage.ts
ibalosh: Hey @cmraible , that is correct, per [their docs](https://playwright.dev/docs/actionability#visible) Playwright counts as invisible only `display:none` and zero size elements. 

I do find this a bit counter intuitive, updated it to just wait for performance graph to be visible, which seems to work also reliably.

e2e/tests/admin/analytics/post-analytics/growth.test.ts
ibalosh: Ah got it, that makes sense. Thanks for details, did not realize we lazy load the react apps.

I could not figure out what's causing it though :(, but wait for element to be visible is decent way to check for now (I changed it to wait performance which makes more sense.

It only happens when I click from Post analytics overview navigation link like `growth`, and I guess it might be for other elements in that navigation bar, which use routes. For now, just added in the wait workaround

ibalosh:
> Overall looking really good! Love how clean the tests are, they're so easy to read.
> 
> Left a bunch of comments but mostly small stuff and some questions - nothing that I think is blocking.

thanks @cmraible ! Implemented most of the suggestions 👌

ibalosh:
@9larsons @cmraible if no objections, plan to release this on Monday morning (making sure it doesn't break, ran it ton of times)

e2e/tests/auth.setup.ts
cmraible: 🤙

PR #24665, opened 2025-08-13
🐛 Fixed issue with header images in Outlook

troyciesco:
ref https://linear.app/ghost/issue/PROD-2437

- When using a header card with an image in an email, the images weren't rendering in Outlook
- This PR adds Outlook-specific styling so those images will render
- Tried it out with both an img tag and vml. vml was more reliable in being able to create properly-sized images, both standalone and background.

[This blog post](https://www.hteumeuleu.com/2021/background-properties-in-vml/) is a good primer on vml.

below shows the different header variations as they appear in Outlook after the fix, with the feature flag on. 

Note: I'm not sure what's causing the lack of space between some of the cards, but it seems to be only on Outlook desktop. I'm going to check and make sure there's not a missing closing tag somewhere or something - but if it's not that, I might address this in a follow up. but I'd imagine it's unlikely an email would be designed anything like this anyway.

Screenshot 2025-08-18 at 10 40 08 AM

This is with the feature flag off:
Screenshot 2025-08-18 at 11 54 25 AM

### DIFF
--- a/apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx
+++ b/apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx
@@ -27,6 +27,11 @@ const features: Feature[] = [{
     title: 'Explore',
     description: 'Enables keeping in touch with the new Explore API',
     flag: 'explore'
+},
+{
+    title: 'Outlook header card image',
+    description: 'Enables header card images styled for Outlook',
+    flag: 'emailHeaderCardOutlook'
 }];
 
 const AlphaFeatures: React.FC = () => {
--- a/ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
+++ b/ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
@@ -125,7 +125,7 @@ function emailTemplate(nodeData, options) {
                     <!--[if mso]>
                     <td class="kg-header-card-content" style="${hasImageNoSplit ? 'padding: 0;' : 'padding: 40px;'}${hasContainAndSplit ? 'padding-top: 0;' : ''}">
                     <![endif]-->
-                    <!--[if !mso]><!-- -->
+                    <!--[if !mso]><!-->
                     <td class="kg-header-card-content" style="${hasContainAndSplit ? 'padding-top: 0;' : ''}">
                     <!--<![endif]-->
                     ${hasImageNoSplit ? `
--- a/ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
+++ b/ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
@@ -86,7 +86,9 @@ function emailTemplate(nodeData, options) {
     });
 
     const hasDarkBg = nodeData.textColor?.toLowerCase() === '#ffffff';
-
+    const hasContainAndSplit = nodeData.backgroundSize === 'contain' && nodeData.layout === 'split';
+    const hasImageNoSplit = nodeData.backgroundImageSrc && nodeData.layout !== 'split';
+    
     return (
         `
         <div class="kg-header-card kg-v2${hasDarkBg ? ' kg-header-card-dark-bg' : 'kg-header-card-light-bg'}" style="color:${nodeData.textColor}; ${alignment} ${backgroundImageStyle} ${backgroundAccent}">
@@ -119,19 +121,21 @@ function emailTemplate(nodeData, options) {
             ` : ''}
             <table border="0" cellpadding="0" cellspacing="0" width="100%" style="color:${nodeData.textColor}; ${alignment} ${backgroundImageStyle} ${backgroundAccent}">
                 <tr>
-                <!--[if mso]>
-                <td class="kg-header-card-content" style="${nodeData.layout !== 'split' ? 'padding: 0;' : 'padding: 40px;'}${nodeData.layout === 'split' && nodeData.backgroundSize === 'contain' ? 'padding-top: 0;' : ''}">
-                <![endif]-->
-                <!--[if !mso]><!-- -->
-                <td class="kg-header-card-content" style="${nodeData.layout === 'split' && nodeData.backgroundSize === 'contain' ? 'padding-top: 0;' : ''}">
-                <!--<![endif]-->
-                ${labs.isSet('emailHeaderCardOutlook') && nodeData.layout !== 'split' && nodeData.backgroundImageSrc ? `
-                <!--[if mso]>
-                <v:rect xmlns:v="urn:schemas-microsoft-com:vml" fill="true" stroke="false" style="width:600px;">
-                    <v:fill src="${nodeData.backgroundImageSrc}" color="${nodeData.backgroundColor}" type="frame" aspect="atleast" focusposition="0.5,0.5" />
-                    <v:textbox inset="30pt,30pt,30pt,30pt" style="mso-fit-shape-to-text:true;">
-                <![endif]-->
-                ` : ''}
+                    ${labs.isSet('emailHeaderCardOutlook') ? `
+                    <!--[if mso]>
+                    <td class="kg-header-card-content" style="${nodeData.layout !== 'split' ? 'padding: 0;' : 'padding: 40px;'}${hasContainAndSplit ? 'padding-top: 0;' : ''}">
+                    <![endif]-->
+                    <!--[if !mso]><!-- -->
+                    <td class="kg-header-card-content" style="${hasContainAndSplit ? 'padding-top: 0;' : ''}">
+                    <!--<![endif]-->
+                    ${hasImageNoSplit ? `
+                    <!--[if mso]>
+                    <v:rect xmlns:v="urn:schemas-microsoft-com:vml" fill="true" stroke="false" style="width:600px;">
+                        <v:fill src="${nodeData.backgroundImageSrc}" color="${nodeData.backgroundColor}" type="frame" aspect="atleast" focusposition="0.5,0.5" />
+                        <v:textbox inset="30pt,30pt,30pt,30pt" style="mso-fit-shape-to-text:true;">
+                    <![endif]-->
+                    ` : ''}
+                    ` : `<td class="kg-header-card-content" style="${hasContainAndSplit ? 'padding-top: 0;' : ''}">`}
                         <table border="0" cellpadding="0" cellspacing="0" width="100%">
                             <tr>
                                 <td align="${nodeData.alignment}">
@@ -151,15 +155,13 @@ function emailTemplate(nodeData, options) {
                                 ` : ''}
                             </tr>
                         </table>
-                ${labs.isSet('emailHeaderCardOutlook') && nodeData.layout !== 'split' && nodeData.backgroundImageSrc ? `
+                ${labs.isSet('emailHeaderCardOutlook') && hasImageNoSplit ? `
                 <!--[if mso]>
                     </v:textbox>
                 </v:rect>
                 <![endif]-->
                 ` : ''}
-                    <!--[if mso]></td><![endif]-->
-                    <!--[if !mso]><!-- --></td><!--<![endif]-->
-
+                    </td>
                 </tr>
             </table>
         </div>
--- a/ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
+++ b/ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
@@ -175,14 +175,14 @@ function emailTemplate(nodeData, options) {
                 <table border="0" cellpadding="0" cellspacing="0" width="100%">
                     <tr>
                         <td background="${nodeData.backgroundImageSrc}" style="${splitImageStyle}" class="kg-header-card-image" bgcolor="${nodeData.backgroundColor}" align="center">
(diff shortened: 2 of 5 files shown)


ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
```diff
@@ -92,12 +93,30 @@ function emailTemplate(nodeData, options) {
...
             ` : ''}
             <table border="0" cellpadding="0" cellspacing="0" width="100%" style="color:${nodeData.textColor}; ${alignment} ${backgroundImageStyle} ${backgroundAccent}">
                 <tr>
+                ${labs.isSet('emailHeaderCardOutlook') && nodeData.layout !== 'split' && nodeData.backgroundImageSrc ? `
+                <!--[if mso]>
+                <v:rect xmlns:v="urn:schemas-microsoft-com:vml" fill="true" stroke="false" style="width:600px;">
+                    <v:fill src="${nodeData.backgroundImageSrc}" color="${nodeData.backgroundColor}" type="frame" aspect="atleast" focusposition="0.5,0.5" />
+                    <v:textbox inset="0,24pt,0,24pt" style="mso-fit-shape-to-text:true;">
+                <![endif]-->
+                ` : ''}
```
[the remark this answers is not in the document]
troyciesco: there's padding in .kg-header-card-content that has to be accounted for

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
```diff
@@ -85,20 +86,56 @@ function emailTemplate(nodeData, options) {
...
 
     const hasDarkBg = nodeData.textColor?.toLowerCase() === '#ffffff';
-
+    const hasContainAndSplit = nodeData.backgroundSize === 'contain' && nodeData.layout === 'split';
+    const hasImageNoSplit = nodeData.backgroundImageSrc && nodeData.layout !== 'split';
+    
     return (
         `
-        <div class="kg-header-card kg-v2${hasDarkBg ? ' kg-header-card-dark-bg' : 'kg-header-card-light-bg'}" style="color:${nodeData.textColor}; ${alignment} ${backgroundImageStyle} ${backgroundAccent}">
+        <div class="kg-header-card kg-v2 ${hasDarkBg ? 'kg-header-card-dark-bg' : 'kg-header-card-light-bg'}" style="color:${nodeData.textColor}; ${alignment} ${backgroundImageStyle} ${backgroundAccent}">
```
troyciesco, line 167: stumbled on this by accident - the lack of space made light background header cards have way more padding. moved the space outside since the ternary will always result in a value.

before:
Screenshot 2025-08-18 at 10 15 59 AM

after:
Screenshot 2025-08-18 at 10 16 17 AM

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
```diff
@@ -85,20 +86,56 @@ function emailTemplate(nodeData, options) {
...
+                    <td class="kg-header-card-content" style="${hasContainAndSplit ? 'padding-top: 0;' : ''}">
+                    <!--<![endif]-->
+                    ${hasImageNoSplit ? `
+                    <!--[if mso]>
+                    <v:rect xmlns:v="urn:schemas-microsoft-com:vml" fill="true" stroke="false" style="width:600px;">
+                        <v:fill src="${nodeData.backgroundImageSrc}" color="${nodeData.backgroundColor}" type="frame" aspect="atleast" focusposition="0.5,0.5" />
+                        <v:textbox inset="30pt,30pt,30pt,30pt" style="mso-fit-shape-to-text:true;">
+                    <![endif]-->
+                    ` : ''}
+                    ` : `<td class="kg-header-card-content" style="${hasContainAndSplit ? 'padding-top: 0;' : ''}">`}
```
troyciesco: this is redundant with L129 right now, but it'll be removed when the labs flag is removed after testing

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
kevinansfield: Good catch!

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
```diff
@@ -85,20 +86,56 @@ function emailTemplate(nodeData, options) {
...
                 </table>
             ` : ''}
             <table border="0" cellpadding="0" cellspacing="0" width="100%" style="color:${nodeData.textColor}; ${alignment} ${backgroundImageStyle} ${backgroundAccent}">
                 <tr>
-                    <td class="kg-header-card-content" style="${nodeData.layout === 'split' && nodeData.backgroundSize === 'contain' ? 'padding-top: 0;' : ''}">
+                    ${labs.isSet('emailHeaderCardOutlook') ? `
+                    <!--[if mso]>
+                    <td class="kg-header-card-content" style="${hasImageNoSplit ? 'padding: 0;' : 'padding: 40px;'}${hasContainAndSplit ? 'padding-top: 0;' : ''}">
+                    <![endif]-->
+                    <!--[if !mso]><!-- -->
```
kevinansfield: Is that extra comment just a typo or does it trigger some random rendering change in Outlook?

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
```diff
@@ -85,20 +86,56 @@ function emailTemplate(nodeData, options) {
...
+                    <td class="kg-header-card-content" style="${hasContainAndSplit ? 'padding-top: 0;' : ''}">
+                    <!--<![endif]-->
```
kevinansfield: Similar question here, why is the `endif` wrapped in another comment?

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
```diff
@@ -85,20 +86,56 @@ function emailTemplate(nodeData, options) {
...
+                                <v:fill type="frame"
+                                        aspect="atmost"
+                                        size="225pt,120pt"
+                                        src="${nodeData.backgroundImageSrc}"
+                                        color="${nodeData.backgroundColor}" />` : `
+                                
+                                <v:rect xmlns:v="urn:schemas-microsoft-com:vml" fill="true" stroke="false" 
+                                style="width:600px;height:320px;">
+                                    <v:fill type="frame" aspect="atleast" src="${nodeData.backgroundImageSrc}" color="${nodeData.backgroundColor}" />
+                                    `}
```
kevinansfield: It took me a while to parse this ternary. Maybe it would help if the formatting was normalised a bit more so there's no unnecessary line breaks and the indentation is better? If that doesn't work perhaps we can extract a function or predefined variable to make it easier to read

kevinansfield (changes requested):
Added a couple of questions for minor but slightly odd things with comments. I'd like to see the nested ternary formatting improved before we merge so it's easier to reason about when it comes to removing the flag and any future edits.

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
troyciesco: using it in conjunction with the outlook-specific code on L125-127, it basically creates an if/else block: https://www.emailonacid.com/blog/article/email-development/conditional-css-code

from that blog: "“if !mso” is declaring, “if the client is not Microsoft Outlook”. You’ll also need to append your initial  and preface your closing <![endif]--> tag with <!-- in order for Outlook to acknowledge this code."
```
<!--[if !mso]><!-->
  <p>This is a secret message for everyone who doesn’t use Outlook!</p>
<!--<![endif]-->
```

though i see compared to that example i have an extra `-->`, that must have been an autocomplete thing lol i can test and remove that bit to match the example

troyciesco: and the reason for the if/else in the first place is to remove the padding on outlook if there's an image, because the inset on L135 handles that spacing (the padding just ends up being a white box, the inset properly uses the image as a background)

apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx
```diff
@@ -27,6 +27,11 @@ const features: Feature[] = [{
     title: 'Explore',
     description: 'Enables keeping in touch with the new Explore API',
     flag: 'explore'
+},
+{
+    title: 'Outlook header card image',
+    description: 'Enables header card images styled for Outlook',
+    flag: 'emailHeaderCardOutlook'
 }];
```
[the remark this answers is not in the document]
troyciesco: There's a `<!--[if !mso]>` too, which would make for 5 openings and 5 closings, please double-check

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
```diff
@@ -88,17 +170,19 @@ function emailTemplate(nodeData, options) {
...
     return (
         `
-        <div class="kg-header-card kg-v2${hasDarkBg ? ' kg-header-card-dark-bg' : 'kg-header-card-light-bg'}" style="color:${nodeData.textColor}; ${alignment} ${backgroundImageStyle} ${backgroundAccent}">
+        <div class="kg-header-card kg-v2 ${hasDarkBg ? 'kg-header-card-dark-bg' : 'kg-header-card-light-bg'}" style="color:${nodeData.textColor}; ${alignment} ${backgroundImageStyle} ${backgroundAccent}">
             ${nodeData.layout === 'split' && nodeData.backgroundImageSrc ? `
                 <table border="0" cellpadding="0" cellspacing="0" width="100%">
                     <tr>
-                        <td background="${nodeData.backgroundImageSrc}" style="${splitImageStyle}" class="kg-header-card-image"></td>
+                        <td background="${nodeData.backgroundImageSrc}" style="${splitImageStyle}" class="kg-header-card-image" bgcolor="${nodeData.backgroundColor}" align="center">
+                            ${generateMSOSplitHeaderImage(nodeData) /* mso-only img, no shared markup */}
```
troyciesco, line 172: normally i'd destructure nodeData but 
- I do think the prefix is useful in the rest of `emailTemplate`
- I could destructure just to pass to these generate functions, but I didn't want to have some instances of things like `nodeData.backgroundImageSrc`, and others just `backgroundImageSrc` in `emailTemplate`
- Since it's getting called within this string, having `({backgroundImageSrc: nodeData.backgroundImageSrc...` and so on got pretty hard to read and format

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
```diff
@@ -68,6 +69,81 @@ function cardTemplate(nodeData, options = {}) {
...
+        return '';
+    }
+
+    // spacing here maintains proper markup indentation
+    return `
+                <!--[if mso]>
+                    </v:textbox>
+                </v:rect>
+                <![endif]-->
+                `;
```
kevinansfield: Not a huge deal but we should optimise for developer readability rather than output readability.

The output doesn't impact tests because all comparisons go through prettier first and it doesn't really impact Litmus/Parcel testing because they both have code formatters available when editing in-app. The only place output formatting becomes visible to developers is in snapshots but they are more for detecting changes across commits rather than verifying behaviour. Quite a few of our cards use the `html` or `oneline` tagged template literals which collapse everything into a single line so the indentation in source code is purely for source-code readability.

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
```diff
@@ -68,6 +69,81 @@ function cardTemplate(nodeData, options = {}) {
...
+            `;
+    }
+}
+
+function generateMSOContentWrapper(nodeData) {
+    const {backgroundImageSrc, backgroundColor} = nodeData;
+    const hasContainAndSplit = nodeData.backgroundSize === 'contain' && nodeData.layout === 'split';
+    const hasImageNoSplit = nodeData.backgroundImageSrc && nodeData.layout !== 'split';
+
+    if (!labs.isSet('emailHeaderCardOutlook')) {
```
kevinansfield, line 107: This direct usage of `labs` is an interesting wrinkle that's come up since moving renderers from Koenig to Ghost 🤔

Previously we always used `options.labs.x` (`options` being the second arg to the renderer function) because the editor was separate from Ghost and so any labs flags had to be passed down through the editor/renderer stack. If we start using `labs` directly in renderers it will mean copy operations performed in the editor will never use labs features when building HTML for the clipboard even when they "enabled" because the editor runs as a separate client app with the renderers bundled, unless we find some other way to initialise the renderers when used inside the editor.

I think it's fine to leave as-is for now because this is a very short-term flag but we should decide which way we want to standardise when it comes to unifying Ghost/Koenig.

kevinansfield (approved):

ghost/core/core/server/services/koenig/node-renderers/header-v2-renderer.js
troyciesco: Ah you're right - while working on this the snapshot was updating and I'd thought the indentations were causing changes, but it might have just been new lines. Makes sense that this self-contained function wouldn't have knowledge of where it's used (and thus where to indent) anyway. Fixed to match patterns in the rest of the file!

PR #24666, opened 2025-08-13
Updated error handler middleware to return plain text errors for static files

cmraible:
ref https://linear.app/ghost/issue/ONC-1095/investigate-failed-to-lookup-view-error-404-in-views-directory

When serving 404 errors for missing asset files, we should serve plain text error pages instead of full HTML. This prevents recursion issues if the full HTML 404 page references the missing asset, and avoids rendering errors if the theme engine hasn't been initialized yet (i.e. immediately after boot).

This identifies static file errors in the frontend's error handler middleware based on whether the request has a file extension or if the error already has the `STATIC_FILE_NOT_FOUND` error code attached to it, and returns a plain text error before reaching the template rendering logic.

### DIFF
--- a/ghost/core/core/frontend/web/middleware/error-handler.js
+++ b/ghost/core/core/frontend/web/middleware/error-handler.js
@@ -1,6 +1,8 @@
 const hbs = require('express-hbs');
 const _ = require('lodash');
+const path = require('path');
 const tpl = require('@tryghost/tpl');
+const errors = require('@tryghost/errors');
 const sentry = require('../../../shared/sentry');
 
 const config = require('../../../shared/config');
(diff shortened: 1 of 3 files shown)


ghost/core/core/frontend/web/middleware/error-handler.js
```diff
@@ -34,13 +38,47 @@ const errorFallbackMessage = err => `<h1>${tpl(messages.oopsErrorTemplateHasErro
...
+            err.code = 'STATIC_FILE_NOT_FOUND';
+        }
+
+        const message = err.message || tpl(messages.unexpectedError);
+        const statusCode = err.statusCode || 500;
+
+        // Send a simple plain text error response
+        res.status(statusCode);
+        res.type('text/plain');
+        return res.send(message);
```
cmraible, line 81: We could alternatively call `return next(err)` here to fallback to express's default error handler, but this still returns HTML (although un-styled).

Note: this also changes the behavior for missing image files - see `test/e2e-frontend/static-files.test.js` below

nly380928-ux:
> ref https://linear.app/ghost/issue/ONC-1095/investigate-failed-to-lookup-view-error-404-in-views-directory
> 
> 
> 
> When serving 404 errors for missing asset files, we should serve plain text error pages instead of full HTML. This prevents recursion issues if the full HTML 404 page references the missing asset, and avoids rendering errors if the theme engine hasn't been initialized yet (i.e. immediately after boot).
> 
> 
> 
> This identifies static file errors in the frontend's error handler middleware based on whether the request has a file extension or if the error already has the `STATIC_FILE_NOT_FOUND` error code attached to it, and returns a plain text error before reaching the template rendering logic.

kevinansfield (approved):

PR #24671, opened 2025-08-14
Added support for top-level `yarn test:browser --ui`

kevinansfield:
no issue

- previously it wasn't possible to pass additional arguments through to `yarn test:browser` meaning playwright args like `--ui` and `--headed` were unusable
- by splitting the args across a `--` separator we can separate Ghost app commands from the pass through args, then use those pass through args in the internal call to `ghost:test:browser`
- we can't use a `--` separator directly on the command line because `yarn` will strip it out when building the script command so we also need a trailing `--` in the package.json script so the command now works like `yarn test:browser --ui` as if you were calling the Ghost test script or playwright directly

### DIFF
--- a/.github/scripts/dev.js
+++ b/.github/scripts/dev.js
@@ -41,9 +41,18 @@ const availableAppFlags = {
     https: 'Serve apps using HTTPS',
     offline: 'Run in offline mode (no Stripe webhooks will be forwarded)'
 }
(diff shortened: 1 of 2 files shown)


cmraible (approved):
LGTM! I just added the browser-tests label to be sure this doesn't break in CI.

Sorry for the inconvenience!

PR #24681, opened 2025-08-15
Improved development setup for analytics with docker

cmraible:
ref https://linear.app/ghost/issue/PROD-2328/orchestrate-tinybird-and-ghost-boot-sequencesetup-in-e2e-package

- Added tinybird-local container to compose
- Added analytics service to compose
- Added tinybird CLI service to compose
- Added entrypoint scripts to get tokens from tinybird-local container and configure ghost and analytics service with them

This unlocks a zero config setup to run traffic analytics with Ghost locally in development. This works with the `analytics` and `split` COMPOSE_PROFILES enabled

---

**How it works**
With the `analytics` profile enabled, running `docker compose up` will:
- start the `tinybird-local` container
- once `tinybird-local` is healthy, the `tb-cli` service will:
   - run `tb --local build` to deploy tinybird files
   - fetch the workspace ID, admin token, and tracker token, saving them to a `.env` file in the `.../server/data/tinybird` directory
- start the analytics service, grabbing the tracker token from the `.env` file
- start ghost, grabbing the workspace ID and admin token from the `.env` file


compose.yml
```diff
@@ -152,6 +169,63 @@ services:
         - ping
       interval: 1s
       retries: 120
+
+  analytics:
+    profiles: [ analytics ]
+    image: ghost/traffic-analytics:1
+    command: ["node", "--enable-source-maps", "dist/server.js"]
+    entrypoint: [ "/app/entrypoint.sh" ]
```
ibalosh, line 176: I am wondering, any reason that we separate `entrypoint` - the default executable for the container from the Dockerfile? My worry is that with this approach, it gets little bit harder to read the original Dockerfile, for example for analytics, since entrypoint is here, and the docker file is elsewhere.

Unless we need very flexible entrypoints for containers, we probably don't need this level of decoupling?

To keep dockerfile clean, we could do something like, and just keep them together:

```
COPY entrypoint.sh /usr/local/bin/
RUN chmod +x /usr/local/bin/entrypoint.sh
ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
```

It would make running separate containers, and isolation bit simpler.

If we need choices in entrypoint execution, we can go with something like in examples [here](https://www.docker.com/blog/docker-best-practices-choosing-between-run-cmd-and-entrypoint/) in Dockerfile:

```
ENTRYPOINT ["httpd"]
 
# provides default parameters, can be overridden at runtime
CMD ["-D", "FOREGROUND"]
```

.docker/tb-cli/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# Install dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    curl \
+    jq \
+    ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /home/tinybird
+
+RUN uv tool install tinybird@0.0.1.dev285 --python 3.13 --force
```
ibalosh, line 15: **note, not related to review**: This will install specific tinybird CLI version right? I am wondering @9larsons and @cmraible if we should lock CLI version in Github actions CI workflow too, so it' doesn't break by accident by version bump

ibalosh (approved):
Hey @cmraible 

this looks great to me, nice work! My only question is related to design choice I noticed in couple of places, regarding the entrypoint being decoupled from dockerfile files.

I left a comment related to that.

compose.yml
cmraible: This is a good thought for the `tb-cli` and the `ghost` service — since those Dockerfiles are co-located in the repo, we could add the entrypoint in the Dockerfile directly and remove it from the `compose.yml`.

For the analytics service though, the Dockerfile isn't built in this repo (we pull it from docker hub instead) and the entrypoint script has logic that is only relevant in the context of this `compose.yml` file. There may be a better way to handle this though 🤔

.docker/tb-cli/Dockerfile
cmraible: Yep! I agree, esp. after the little incident we had recently. This implementation is sort of a proof of concept, and if we find that it's working reliably we can update CI to use the same method, instead of using the latest version as we are now.

compose.yml
```diff
@@ -55,6 +55,12 @@ services:
         condition: service_healthy
       redis:
         condition: service_healthy
+      tinybird-local:
+        condition: service_healthy
+        required: false
+      analytics:
+        condition: service_healthy
+        required: false
```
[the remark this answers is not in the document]
cmraible: Those are valid fields

compose.yml
cmraible: I've done this for the `tb-cli` service. I think we probably should also do it for the `ghost` service, but that feels out of place in this PR, so will tackle that in a follow up.

ibalosh: agreed 💯

ibalosh:
re-requested the review, until we figure out why sometimes it seems to not work locally

9larsons:
ref https://ghost.slack.com/archives/C07HTEJMR2R/p1755560308554199

Igor may need to rebuild the tb-cli container per Chris.

For myself, I found that Safari doesn't always treat *.localhost as loopback addresses, so we end up with mixed origins (localhost on 2368, ghost.localhost on 80, with the analytics request from 2368->80). We also can't nav to `ghost.localhost` like you can on other browsers.

For Chrome, works great!

cmraible:
I've just pushed a [commit](https://github.com/TryGhost/Ghost/pull/24681/commits/b914c16859eac96573ca8231150eb55f618ed952) that should make this work out of the box in Safari, as well as any other major browser.

9larsons (approved):
I like it - it cleans up some of the complexity in the Docker setup in lieu of a small addition to the app. Works great, nice work!

ibalosh (approved):

PR #24690, opened 2025-08-18
Fixed card asset loading with absolute URLs in ActivityPub

minimaluminium:
ref https://linear.app/ghost/issue/PROD-2440/oss-issue-cards-display-incorrectly-in-the-activitypub-reader

- Changed card asset JS and CSS loading to use absolute URLs in reader
- Resolved issue where assets failed to load when frontend and admin were on different domains


sagzy (approved):
LGTM! One tiny piece of feedback of the PR description/commit message: using the URL of the Linear issue instead of the issue ID is preferable, as we can then just click on a link to open additional context

PR #24695, opened 2025-08-18
Fixed members KPI chart missing comped members

9larsons:
ref https://linear.app/ghost/issue/PROD-2465/
- fixed chart data points including comped members
- fixed calculation missing comped members in net change (%)

Users reported comped members missing from the paid member counts. We always returned comped members from the endpoint so this change only requires some simple adjustments to the frontend. This makes the behavior consistent with the old dashboard.

Notes for testing:
- Create/generate some members
- Ensure comped counts are included in the chart (view the members_count request in the Network tab of browser dev tools)
- Ensure the % change makes sense

### DIFF
--- a/apps/stats/src/hooks/useGrowthStats.ts
+++ b/apps/stats/src/hooks/useGrowthStats.ts
@@ -168,7 +168,7 @@ const formatChartData = (memberData: MemberStatusItem[], mrrData: MrrHistoryItem
         const free = lastMemberItem?.free ?? 0;
         const paid = lastMemberItem?.paid ?? 0;
         const comped = lastMemberItem?.comped ?? 0;
-        const paidTotal = paid + comped;  // Include comped in paid total for chart
+        const paidTotal = paid + comped;
         const value = free + paidTotal;
         const mrr = lastMrrItem?.mrr ?? 0;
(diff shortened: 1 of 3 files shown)


cmraible (changes requested):
This fixes the under counting on the "Paid members" tab, but now the "Total members" tab is over counting.

To reproduce:
1. Start with an empty site
2. Setup stripe / paid members
3. Create a single paid member and a single comped member
4. Load admin and checkout Analytics > Growth > Total Members. The final datapoint from today shows 3 instead of 2

Image

cmraible (changes requested):
Ugh, sorry I missed this the first time. There's a similar bug in the Overview tab, Members chart, i.e. it shows 3 when it should show 2. I think we need to make some adjustments in the Overview component [here](https://github.com/TryGhost/Ghost/blob/fix-member-ct-comped/apps/stats/src/views/Stats/Overview/Overview.tsx#L141-L146).

One other thing (non-blocking): I'm a little surprised these changes didn't break any of the tests in https://github.com/TryGhost/Ghost/blob/fix-member-ct-comped/apps/stats/test/unit/hooks/useGrowthStats.test.tsx.

cmraible (approved):
Tested locally and it's all working for me 👍

PR #24697, opened 2025-08-18
Updated staff/author api tests to use snapshots

troyciesco:
ref https://linear.app/ghost/issue/PROD-608

- Updates author content api and user (staff) admin api to use snapshots for tests


troyciesco:
Closing this in favor of doing one PR for each

PR #24702, opened 2025-08-19
Updated Content API author endpoint tests to use snapshots

troyciesco:
ref https://linear.app/ghost/issue/PROD-608

- Updates the content API author endpoints to use snapshots for testing
- Also moves away from using `should` and uses `assert` instead
- Maintains useful assertions in addition to the snapshots

### DIFF
--- a/ghost/core/test/e2e-api/content/authors.test.js
+++ b/ghost/core/test/e2e-api/content/authors.test.js
@@ -41,8 +41,8 @@ describe('Authors Content API', function () {
                 
                 // Verify URL structure
                 const urlParts = new URL(body.authors[0].url);
-                assert.equal(urlParts.protocol, 'http:');
-                assert.equal(urlParts.host, '127.0.0.1:2369');
+                assert.ok(['http:', 'https:'].includes(urlParts.protocol));
+                assert.equal(urlParts.pathname, `/author/${body.authors[0].slug}/`);
             });
     });
     it('Can request authors including post count', async function () {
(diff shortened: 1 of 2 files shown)


ghost/core/test/e2e-api/content/authors.test.js
```diff
@@ -1,119 +1,125 @@
...
+            .matchBodySnapshot({
+                authors: new Array(3).fill(authorMatcherWithCount)
+            })
+            .expect(cacheInvalidateHeaderNotSet())
+            .expect(({body}) => {
+                const {authors} = body;
+                
+                // Verify response structure
+                localUtils.API.checkResponse(body, 'authors');
+                localUtils.API.checkResponse(authors[0], 'author', ['count', 'url'], null, null);
```
troyciesco: Left these at first to comment that I'm going to remove them lol following the [e2e testing codex](https://ghost.notion.site/End-to-end-Testing-6a2ef073b1754b18aff42e24a632a007), my understanding is that checks using localUtils can be removed but some of the others below (body length, ordering, etc) can still be useful.

ghost/core/test/e2e-api/content/authors.test.js
```diff
@@ -1,119 +1,118 @@
...
-            .expect(200);
-
-        should.not.exist(res.headers['x-cache-invalidate']);
-        const jsonResponse = res.body;
-        should.exist(jsonResponse.authors);
-        localUtils.API.checkResponse(jsonResponse, 'authors');
-        jsonResponse.authors.should.have.length(3);
-
-        // We don't expose the email address, status and other attrs.
-        localUtils.API.checkResponse(jsonResponse.authors[0], 'author', ['url'], null, null);
```
kevinansfield, line 41: I think it would be nice to capture more of this in explicit assertions. Snapshots are nice for being notified that something in API output has changed but they don't assert that new snapshot data no longer matches our agreed rules such as not exposing particular data.

If there are no explicit assertions it's very easy to do an `UPDATE_SNAPSHOT=true yarn test....` and commit, unknowingly exposing data on the public API we've decided should never be exposed.

We _could_ do this with matchers but we don't have any pattern for that in our existing tests, I think instead we tend to have some explicit assertions on the response body.

troyciesco: Gotcha, yeah that does make sense. I'd removed it based on this line in the codex: "The goal of the new framework is to be a single point of access for everything you need for testing, without having to require various utilities files. No more testUtils, utils, localUtils, etc."

Maybe I could add that localUtils back for now though and find/create a follow up somewhere to make something that could work for all tests and be added to that codex as a recommendation? 

I imagine it'd be something like `const authorMatcher = createMatcher('author')`, where createMatcher only allows the fields that checkResponse is currently allowing. 
Then (if i'm understanding right) the tests will be cleaner and mostly rely on the snapshots, and adding a new fields requires deliberate steps of both 1) updating the allowlist and 2) updating the snapshot

kevinansfield: Yeah, that sounds like a more holistic approach! I was thinking more along the lines of some explicit assertions in this one test.

I think one problem is that jest matchers don't have a way of asserting that a property _doesn't_ exist in a snapshot, it would need to be a separate step. At least it seemed that way when I did a very quick investigation earlier, I could well have missed something.

Either way, sounds like something to raise in our engineering call to see if the consensus is that we should do something or that it doesn't matter and a mismatched snapshot is enough of a signal.

kevinansfield (approved):

PR #24704, opened 2025-08-20
Make CI execution time for tests more consistent, by using dockerized playwright instead of installing dependencies

ibalosh:
I have noticed slow `setup playwright` step, couple of time last week that could take 10 minutes, and it seems it was around before too, I noticed it here, in this [CI run](https://github.com/TryGhost/Ghost/actions/runs/17151367111/job/48658277972) or this [CI run](https://github.com/TryGhost/Ghost/actions/runs/16915051583/job/47926926233).

Docker pull is a slower action, but it is fairly fast, and seem to take similar time, even when cache is used for `setup playwright` since runtime dependencies need to be installed even when cache is used - which I checked [here](https://github.com/microsoft/playwright/issues/7249) and is how we used caching.

With docker pull, times seem constant.  What are your thoughts about this? I would love to hear your opinions

**note:** I intentionally did not add something like docker:test to projects to run these, to keep change minimal for now, if we like this approach.

### DIFF
--- a/.github/actions/docker-playwright/action.yml
+++ b/.github/actions/docker-playwright/action.yml
@@ -1,5 +1,5 @@
-name: "Run E2E Tests"
-description: "Run E2E tests using Playwright in Docker"
+name: "Run Playwright Tests"
+description: "Run Playwright tests"
 
 inputs:
   ghost_base_url:
@@ -7,7 +7,7 @@ inputs:
     required: false
     default: "http://localhost:2368"
   test_command:
-    description: "Test command to run"
+    description: "Command to execute tests"
     required: true
     default: "yarn test:e2e"
 
@@ -18,8 +18,8 @@ runs:
       id: playwright-version
       shell: bash
       run: echo "version=$(node -p "require('@playwright/test/package.json').version")" >> $GITHUB_OUTPUT
-      
-    - name: Run E2E tests
+
+    - name: Run tests
       shell: bash
       run: |
         docker run --rm \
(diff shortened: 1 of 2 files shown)


cmraible (changes requested):
This is really nice! Just two issues that I think we need to resolve before this is ready to merge:

1. I've tried this out on the existing `e2e-browser` tests (these don't run on PRs by default, so I added the `browser-tests` label to the PR to try them) and they've failed. We run these tests in kind of a janky way using the script in `.github/scripts/dev.js`, so I'm not surprised it doesn't work out of the box with the Docker image. For now it might make the most sense to continue using the `setup-playwright` action for that test suite, and use the docker setup for all the others.

2. I think we will encounter Docker hub's rate limiting for pulling images pretty quickly with this, which is something we've encountered in the past. We get around this in the e2e test suite by [configuring a mirror](https://github.com/TryGhost/Ghost/blob/ghost-docker-build-ci/.github/workflows/ci.yml#L819-L823), which seems to be working well. I think we should add that step before pulling the image, and potentially take this opportunity to make that block of code to configure the mirror a re-usable action while we're at it.

ibalosh:
hey @cmraible , thanks for checking this out!

I will put this one on hold and close it for now. I updated it to pass for for browser-tests, but since msn container is not under docker registry mirror but their own one, we might hit the limit. From what I found so far, current approach is used.

If we noticed more frequently that setup playwright slowdown, I will re-open and check this out.

cmraible:
Interesting! FWIW we can add multiple mirrors and it will try them in order for each image until it finds one — so if microsoft is using a different mirror that's not necessarily a blocker (assuming the mirror's rate limits are reasonable).

ibalosh:
hm @cmraible , lets talk today once you are online, I would love this one in our CI, if we get pulls to be consistent

ibalosh:
will reopen this, and take another look with mirrors

PR #24707, opened 2025-08-20
Migrate tags list to React (part 1)

jonatansberg:
This is the first in a series of PRs to migrate the various tag related screens from Ember to React. The new tags list is available behind the `tagsX` feature flag. 

Once the new Playwright test setup is merged I'll follow up with a separate PR for tests so that we can go ahead and remove the feature flag for the list view.

The new tags list screen uses the Shade design system components and should be more or less a 1:1 port of the old Ember version. There are some minor cosmetic differences and an improved responsive layout. 

### DIFF
--- a/apps/admin-x-framework/src/api/tags.ts
+++ b/apps/admin-x-framework/src/api/tags.ts
@@ -1,8 +1,9 @@
+import {InfiniteData} from '@tanstack/react-query';
 import {
     Meta,
-    createQuery,
     createQueryWithId,
-    createMutation
+    createMutation,
+    createInfiniteQuery
 } from '../utils/api/hooks';
 
 export type Tag = {
@@ -42,9 +43,24 @@ export interface TagsResponseType {
 
 const dataType = 'TagsResponseType';
 
-const useBrowseTagsQuery = createQuery<TagsResponseType>({
+const useBrowseTagsQuery = createInfiniteQuery<TagsResponseType>({
     dataType,
-    path: '/tags/'
+    path: '/tags/',
+    defaultNextPageParams: (lastPage, otherParams) => ({
+        ...otherParams,
+        page: (lastPage.meta?.pagination.next || 1).toString()
+    }),
+    returnData: (originalData) => {
+        const {pages} = originalData as InfiniteData<TagsResponseType>;
+        const tags = pages.flatMap(page => page.tags);
+        const meta = pages[pages.length - 1].meta;
+
+        return {
+            tags,
+            meta,
+            isEnd: meta ? meta.pagination.pages === meta.pagination.page : true
+        };
+    }
 });
 
 export const useBrowseTags = ({
--- a/apps/admin-x-framework/src/api/tags.ts
+++ b/apps/admin-x-framework/src/api/tags.ts
@@ -46,10 +46,12 @@ const dataType = 'TagsResponseType';
 const useBrowseTagsQuery = createInfiniteQuery<TagsResponseType>({
     dataType,
     path: '/tags/',
-    defaultNextPageParams: (lastPage, otherParams) => ({
-        ...otherParams,
-        page: (lastPage.meta?.pagination.next || 1).toString()
-    }),
+    defaultNextPageParams: (lastPage, otherParams) => (lastPage.meta?.pagination.next
+        ? {
+            ...otherParams,
+            page: (lastPage.meta?.pagination.next || 1).toString()
+        }
+        : undefined),
     returnData: (originalData) => {
         const {pages} = originalData as InfiniteData<TagsResponseType>;
         const tags = pages.flatMap(page => page.tags);
--- a/apps/admin-x-framework/src/api/tags.ts
+++ b/apps/admin-x-framework/src/api/tags.ts
@@ -1,0 +1,79 @@
+import {
+    Meta,
+    createQuery,
+    createQueryWithId,
+    createMutation
+} from '../utils/api/hooks';
+
+export type Tag = {
+    id: string;
+    name: string;
+    slug: string;
+    url: string;
+    description: string;
+    visibility: 'public' | 'internal';
+    count?: {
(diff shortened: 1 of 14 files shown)


apps/admin-x-framework/src/api/tags.ts
```diff
@@ -0,0 +1,94 @@
...
+    const filterString = Object.entries(filter)
+        .map(([key, value]) => `${key}:${value}`)
+        .join(',');
+    return useBrowseTagsQuery({
+        ...args,
+        searchParams: {
+            limit: '100',
+            order: 'name asc',
+            include: 'count.posts',
+            filter: filterString,
```
kevinansfield, line 83: use NQL package here as discussed

kevinansfield (approved):

apps/posts/src/views/Tags/assets/icons/tags-placeholder.svg
peterzimon, line 1: Let's try to stick to Lucide Icons wherever it's possible. In this case https://lucide.dev/icons/tag works fine.

jonatansberg: Cool! Will do.

apps/posts/src/views/Tags/components/TagsList.tsx
```diff
@@ -0,0 +1,145 @@
...
+                                            href={`#/posts?tag=${item.slug}`}
+                                        >
+                                            {formatNumber(item.count?.posts)}{' '}
+                                            posts
+                                        </a>
+                                    ) : (
+                                        <span className="text-muted-foreground">
+                                            0 posts
+                                        </span>
+                                    )}
```
jonatansberg: @peterzimon Whats the stance re pluralization? 

The old Ember app used an infliction library. I'd typically rather use a translation library since e.g. message-format has great pluralization support. In the React admin we have nether of those AFAIK.

apps/posts/src/views/Tags/assets/icons/tags-placeholder.svg
jonatansberg: Seems like there is a [tags](https://lucide.dev/icons/tags) icon as well The new version of that is very similar to the old one. We are on an older version of Lucide tho that has a different design for the same icon.

apps/admin-x-framework/src/api/tags.ts
jonatansberg: Hmm. Seems like there isn't any serializer in that package, only a parser, like you said. I think we can go ahead and merge this and then I'll revisit this once I tackle the more sophisticated filters.

apps/posts/src/views/Tags/assets/icons/tags-placeholder.svg
jonatansberg: Screenshot 2025-08-26 at 16 39 01

This is what the current (i.e. the old Lucide) tags icon looks like.

apps/admin-x-framework/src/api/tags.ts
kevinansfield: It may well be that expecting the `filter` arg here to be anything other than a string is premature optimisation. NQL is meant to be a human-constructible query language so in most places when we want to pass a filter to the API we either hardcode it or construct the string ourselves based on the use-case requirements. Feels like only allowing a string at the framework API layer here would better match the real API.

apps/posts/src/views/Tags/assets/icons/tags-placeholder.svg
peterzimon: @jonatansberg In this context, let's use the [EmptyIndicator](https://github.com/TryGhost/Ghost/blob/main/apps/shade/src/components/ui/empty-indicator.tsx) component from Shade

apps/posts/src/views/Tags/components/TagsList.tsx
peterzimon: Good point. The Admin is English only ATM and there are no concrete plans for the foreseeable future to translate it, that's why we just went with dumb conditions. Feel free to use a translation library however, that's def. more future proof.

apps/posts/src/views/Tags/assets/icons/tags-placeholder.svg
jonatansberg: I'l make the switch! I must have missed that when looking through the existing components. I'll take the lack of comments re the icon as a silent approval unless you say something else :)

apps/posts/src/views/Tags/components/TagsList.tsx
jonatansberg: In this case it would end up saying `1 posts` which is incorrect, but consistent with the rest of the new admin screens. I can add an inline conditional to fix that if you prefer. 

We could definitely look at adding a translation lib as well. If we want to have translation support, then the rewrite provides a good opportunity to introduce that. Maybe that's something worth discussing in a slightly wider context first tho?

apps/posts/src/views/Tags/assets/icons/tags-placeholder.svg
jonatansberg: @peterzimon Actually the EmptyIndicator isn't a perfect fit. It requires a title, a description, actions and an icon. Here there is no description. 

I could update the EmptyState to make the description optional, or we could add one, or we keep it as it is. What would you prefer? I don't have a strong preference.

peterzimon: Let's make the description optional 👍

apps/posts/src/views/Tags/components/TagsList.tsx
peterzimon: From a design POV we should optimize to show correct pluralization, so `1 post` is strongly preferred. Translation support is not a priority now, so let's only introduce new tools if they are easy to add/maintain and makes our lives easier in general.

peterzimon (approved):

PR #24709, opened 2025-08-20
Added `membersSigninOTC` labs flag

kevinansfield:
closes https://linear.app/ghost/issue/PROD-2478/

- flag for gating logic of alpha development for a one-time-code flow alongside magic links for members signin

PR #24711, opened 2025-08-20
Added otc property to /send-magic-link/ POST data in Portal's signin

kevinansfield:
closes https://linear.app/ghost/issue/PROD-2479/
requires https://github.com/TryGhost/Ghost/pull/24709

The OTC flow for member sign-in will be opt-in on a per-form basis to avoid confusion with existing custom sign-in flows that don't display a code input after triggering the magic link. This adds an `otc: true` attribute to the data Portal sends to the `/send-magic-link/` endpoint so that Ghost knows to generate and add a code to the email.

- updated `ghost_head` helper to add a `data-members-signin-otc` attribute to the portal script tag as that's how we pass config from Ghost to Portal
  - Portal does read the `/api/content/site/` endpoint but we don't expose labs flags there
- set up `labs` property in Portal's state object so we have a pattern for using flags in Portal and assigned the `membersSigninOTC` property based on the associated data attribute
- updated Portal's `signin` action to read the new labs state and pass an `otc` property to the `sendMagicLink` API method when the flag is enabled
- updated Portal's `sendMagicLink` API method to pass the `otc` property through as a data attribute in the POST
  - when set to `undefined` as is the case when the flag is disabled then the property isn't added to the POST data

### DIFF
--- a/apps/portal/src/App.js
+++ b/apps/portal/src/App.js
@@ -58,7 +58,8 @@ export default class App extends React.Component {
             lastPage: null,
             customSiteUrl: props.customSiteUrl,
             locale: props.locale,
-            scrollbarWidth: 0
+            scrollbarWidth: 0,
+            labs: props.labs || {}
         };
     }
 
--- a/apps/portal/src/actions.js
+++ b/apps/portal/src/actions.js
@@ -79,11 +79,13 @@ async function signout({api, state}) {
 }
 
 async function signin({data, api, state}) {
-    const {t} = state;
+    const {t, labs} = state;
+
+    const otc = labs.membersSigninOTC ? true : undefined;
 
     try {
         const integrityToken = await api.member.getIntegrityToken();
-        await api.member.sendMagicLink({...data, emailType: 'signin', integrityToken});
+        await api.member.sendMagicLink({...data, emailType: 'signin', integrityToken, otc});
         return {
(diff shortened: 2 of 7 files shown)


troyciesco:
@kevinansfield I might just need to take a look at this with fresh eyes tomorrow, plus for #24709 to be merged, but I wasn't seeing the `data-members-signin-otc` switch to true in the head even after setting the lab (I pulled down from the other PR), restarting Ghost, etc. But i might be misunderstanding where this is at!

kevinansfield:
@troyciesco that's strange 🤔 Commits from both PRs are required for it to function. When the flag is enabled, on front-end requests you should see the attribute on the `<script>` tag that loads portal. Also, if you wanted to inspect the `/send-magic-link/` API request you have to remember to start the dev server with `yarn dev --portal` otherwise you'll get the CDN version of portal instead of the local one.

Screenshot 2025-08-21 at 12 22 55

troyciesco:
> @troyciesco that's strange 🤔 Commits from both PRs are required for it to function.

Weird, I'd brought the changes from the other PR down locally, which is how i could set the flag, but I must have missed something. In any event, now that main is updated with that PR i can see the data attr just fine. And thanks for the tip on `yarn dev --portal`!

troyciesco (approved):

PR #24715, opened 2025-08-20
Removed additional listeners for page_hit requests

9larsons:
ref https://linear.app/ghost/issue/PROD-2483/
- removed listeners for SPA navigation and hashchange

Theme's using SPA navigation does not play well with our implementation of the ghost-stats script. We're removing these listeners to prevent bad data from being submitted. The initial page hit will still register, so overall visitor data will at least be accurate, if not views.

We may follow this up with changes to make it easier to support SPA-like frontend themes.

### DIFF
--- a/ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
+++ b/ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
@@ -164,7 +164,7 @@ export class GhostStats {
         const referrerData = parseReferrer(location?.href);
         referrerData.url = getReferrer(location?.href) || referrerData.url; // ensure the referrer.url is set for parsing
 
-        // Wait a bit for SPA routers
+        // Debounce tracking to avoid duplicates and ensure page has settled
         this.browser.setTimeout(() => {
             this.trackEvent('page_hit', {
                 'user-agent': navigator?.userAgent,
@@ -181,13 +181,6 @@ export class GhostStats {
         if (this.isListenersAttached) {
             return;
         }
-
-        // Track history navigation
-        this.browser.addEventListener('window', 'hashchange', () => this.trackPageHit());
-        
-        // Handle SPA navigation
-        this.browser.wrapHistoryMethod('pushState', () => this.trackPageHit());
-        this.browser.addEventListener('window', 'popstate', () => this.trackPageHit());
 
         // Handle visibility changes for prerendering
         if (this.browser.getVisibilityState() !== 'hidden') {
--- a/ghost/core/test/unit/frontend/public/ghost-stats.test.js
+++ b/ghost/core/test/unit/frontend/public/ghost-stats.test.js
@@ -404,13 +404,6 @@ describe('ghost-stats.js', function () {
             ghostStats.initConfig();
         });
 
-        it('should setup navigation tracking', function () {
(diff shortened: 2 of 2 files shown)


ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
```diff
@@ -182,13 +182,6 @@ export class GhostStats {
             return;
         }
 
-        // Track history navigation
-        this.browser.addEventListener('window', 'hashchange', () => this.trackPageHit());
-        
-        // Handle SPA navigation
-        this.browser.wrapHistoryMethod('pushState', () => this.trackPageHit());
-        this.browser.addEventListener('window', 'popstate', () => this.trackPageHit());
```
cmraible, line 190: problem: I think we should keep the 'popstate' event listener, as this captures users navigating with the browser's back/forward buttons. however, if we keep it here then we'll still get the mismatched events when anyone in a SPA-like theme uses the back/forward buttons.

cmraible: suggestion: instead of removing all these event listeners, we keep keep track of whether `pushState` has been called - if it has, we ignore all future events. This way we can still respond to normal browser navigation with back/forward, but ignore any navigation within a SPA-like theme.

in other words:
- add a `this.hasPushStateBeenCalled = false` property to the class
- change the `wrapHistoryMethod` to set `this.hasPushStateBeenCalled = true` instead of calling trackPageHit.
- in `trackPageHit`, add an early exit if `hasPushStateBeenCalled === true`

cmraible (changes requested):
I didn't realize that we were explicitly listening to the `popstate` event for browser back/forward navigation, and that by removing that listener, it would prevent sending events on back/forward even on a regular theme. While not a huge problem — I'm sure it's a small fraction of total events we'd miss — ideally I think we'd still send events when navigating back/forward.

I left a suggestion for how we might be able to handle that, but I'm open to other thoughts too. It's a bit trickier than I first imagined 😅 Let me know what you think!

ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
9larsons: It's a fair concern. I don't think we actually see `popstate` with forward/back buttons though. Despite what AI and even the MDN [docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event) say, I don't see the requests firing (on main) when using forward/back on Chrome, Firefox, or Safari. I can even set the breakpoint in the browser dev tools on popstate and don't see a thing. 

Afaict, this is often manually triggered by SPA routers via `pushState` or hash changes (which do trigger the breakpoint). I think including hash changes is a mistake from the get-go (and luckily we don't use them much). 

Google Analytics seems to call this out specifically for SPAs and it seems most analytics implementations take the approach that only a full page load = a page hit.

cmraible: Huh, you're right - we already don't catch the back/forward navigation. I've seen AI hallucinate before, but never seen the MDN docs corroborate the hallucination :lolsob:

cmraible (approved):
Nice work! LGTM

PR #24716, opened 2025-08-20
Updated user admin API tests with snapshots

troyciesco:
closes https://linear.app/ghost/issue/PROD-608

- Updates the users admin API e2e tests to use snapshots, and cleans up a few of the tests


ghost/core/test/utils/fixtures/data-generator.js
```diff
@@ -1661,12 +1661,12 @@ DataGenerator.forKnex = (function () {
             author_id: _.find(DataGenerator.Content.users, {email: 'jbOgendAth@example.com'}).id,
             sort_order: 1
         },
-        {
-            id: ObjectId().toHexString(),
-            post_id: DataGenerator.Content.posts[4].id,
-            author_id: DataGenerator.Content.users[0].id,
-            sort_order: 0
-        },
+        // {
```
troyciesco: Note to self: the transfer posts to owner on deleting an author test was failing because the owner was _also_ an author on this post. So removing a post will fix it but I imagine it'll break tests elsewhere (i see legacy failing currently) so I'll have to find a better solution for that.

ghost/core/test/e2e-api/admin/users.test.js
```diff
@@ -1,492 +1,586 @@
...
-            .expect('Cache-Control', testUtils.cacheRules.private)
-            .expect(200);
-
-        should.exist(res.body.apiKey);
-        should.exist(res.body.apiKey.id);
-        should.exist(res.body.apiKey.secret);
-        should.not.exist(res.headers['x-cache-invalidate']);
-
-        should(res.body.id).not.be.equal(id);
-        should(res.body.secret).not.be.equal(secret);
```
troyciesco, line 490: This may not have been working as intended before, unless this was meant to be some sort of check to make sure the fields showed up in the correct place.

L583-584 are updated below to 1) make sure the token id is the _same_ and 2) that the secret is different. This seems to be the expected behavior in [refreshSecret](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/models/api-key.js#L61)

ghost/core/test/e2e-api/admin/users.test.js
```diff
@@ -1,492 +1,600 @@
...
+            .expect(({body}) => {
+                assert.equal(body.users[0].roles[0].name, expectedRoleName);
+            });
     });
 
     it('Can edit user with empty roles data and does not change the role', async function () {
-        const res = await request.put(localUtils.API.getApiQuery('users/me?include=roles'))
-            .set('Origin', config.get('url'))
-            .send({
+        await agent.put('users/me?include=roles')
```
kevinansfield: This is missing a trailing slash so we'll have a hidden redirect here. I guess the agent just follows it considering tests are passing but adding the trailing slash means we aren't introducing subtle behaviour that's not explicitly tested/called out.

ghost/core/test/e2e-api/admin/users.test.js
```diff
@@ -1,492 +1,600 @@
...
+            });
     });
 
     it('Cannot edit user with invalid roles data', async function () {
-        const userId = testUtils.getExistingData().users[1].id;
-        const res = await request.put(localUtils.API.getApiQuery(`users/${userId}?include=roles`))
-            .set('Origin', config.get('url'))
-            .send({
+        const userId = fixtureManager.get('users', 1).id;
+        await agent.put(`users/${userId}?include=roles`)
```
kevinansfield: No trailing slash (see earlier comment)

ghost/core/test/e2e-api/admin/users.test.js
```diff
@@ -1,492 +1,600 @@
...
+            });
     });
 
     it('Can edit user roles by name', async function () {
-        const userId = testUtils.getExistingData().users[1].id;
-        const res = await request.put(localUtils.API.getApiQuery(`users/${userId}?include=roles`))
-            .set('Origin', config.get('url'))
-            .send({
+        const userId = fixtureManager.get('users', 1).id;
+        await agent.put(`users/${userId}?include=roles`)
```
kevinansfield: No trailing slash (see earlier comment)

ghost/core/test/e2e-api/admin/users.test.js
kevinansfield: Yeah, those two `res.body.{id, secret}` checks were completely redundant considering those keys don't exist on the response 😅

kevinansfield (approved):
A few minor fixes for trailing-slash consistency but otherwise all looks good! 🙌

PR #24717, opened 2025-08-20
Added derive method for OTC

troyciesco:
ref https://linear.app/ghost/issue/PROD-2480

- adds a method on `SingleUseTokenProvider` that can derive an OTC (one-time code) based on a token


ghost/core/core/server/services/members/SingleUseTokenProvider.js
```diff
@@ -8,12 +10,16 @@ class SingleUseTokenProvider {
...
      * @param {number} dependencies.maxUsageCount - How many times a token can be used.
+     * @param {import('./MembersConfigProvider')} dependencies.membersConfig - Members config provider for auth secrets.
      */
-    constructor({SingleUseTokenModel, validityPeriod, validityPeriodAfterUsage, maxUsageCount}) {
+    constructor({SingleUseTokenModel, validityPeriod, validityPeriodAfterUsage, maxUsageCount, membersConfig}) {
         this.model = SingleUseTokenModel;
         this.validityPeriod = validityPeriod;
         this.validityPeriodAfterUsage = validityPeriodAfterUsage;
         this.maxUsageCount = maxUsageCount;
+        this.membersConfig = membersConfig;
```
troyciesco: This feels like possibly not the exact correct way to do this, but since this is a stand-in secret anyway should be ok for proof of concept

ghost/core/core/server/services/members/SingleUseTokenProvider.js
```diff
@@ -106,6 +112,54 @@ class SingleUseTokenProvider {
...
+     * @private
+     * @method deriveCounter
+     * Derives a counter from token ID and value using HMAC
+     *
+     * @param {string} tokenId
+     * @param {string} tokenValue
+     * @returns {number}
+     */
+    deriveCounter(tokenId, tokenValue) {
+        const secret = this.membersConfig.getAuthSecret();
```
troyciesco: ideally in real life this secret should be different than the one on L143 since they're doing two different things

ghost/core/core/server/services/members/SingleUseTokenProvider.js
kevinansfield: I think I'd prefer a plain `authSecret` prop rather than passing in a whole object. 

`SingleUseTokenProvider` lives inside the members service but it's otherwise agnostic and has no members-specific knowledge, it could just as easily be lifted out then used for single-use tokens in other areas of the app. Passing in a members-specific object complicates the knowledge needed to use the class and ties us to the object shape of an unrelated area of the code.

ghost/core/core/server/services/members/SingleUseTokenProvider.js
kevinansfield: Why does this need to use a secret? `hotp` already uses a secret so using a secret here as well doesn't buy us anything extra even if it was a different one, all we need as a counter is a deterministic hash derived from the values.

troyciesco: I think I was just overthinking it because the token and its id were involved, and I'd been referencing RequestIntegrityTokenProvider.js as well (but that's for actually creating a token). But from doing more reading this morning ([1](https://stackoverflow.com/questions/61282085/hotp-counter-value-security), [2](https://crypto.stackexchange.com/questions/14976/counter-to-indicate-hotp-count)) I think I was being overzealous. Can fix!

ghost/core/core/server/services/members/SingleUseTokenProvider.js
troyciesco: fixed!

(fwiw, also marked as optional for now bc i haven't updated any other spots this provider is used)

ghost/core/core/server/services/members/SingleUseTokenProvider.js
```diff
@@ -106,6 +112,44 @@ class SingleUseTokenProvider {
...
+    /**
+     * @private
+     * @method deriveCounter
+     * Derives a counter from a token ID and value
+     *
+     * @param {string} tokenId
+     * @param {string} tokenValue
+     * @returns {number}
+     */
+    deriveCounter(tokenId, tokenValue) {
```
troyciesco, line 123: this is useful as a separate function so we can use it in `verifyOTC` later

ghost/core/core/server/services/members/SingleUseTokenProvider.js
```diff
@@ -106,6 +112,46 @@ class SingleUseTokenProvider {
...
+
+    /**
+     * @method deriveOTC
+     * Derives an OTC (one-time code) from a token ID and value
+     *
+     * @param {string} tokenId - Token ID
+     * @param {string} tokenValue - Token value
+     * @returns {string} The generated one-time code
+     */
+    deriveOTC(tokenId, tokenValue) {
```
troyciesco, line 137: Went with these params instead of a token object because when we create `verifyOTC` later (example below) the otplib library takes a token as an argument, but in our case that's the otc, so i thought a token param might be a bit confusing - and I wanted to use consistent-looking signatures in all of these functions.
```
    verifyOTC(tokenId, tokenValue, otc) {
        const counter = this.deriveCounter(tokenId, tokenValue);
        return hotp.verify({token: otc, secret: this.secret, counter});
    }
```

Could also just do `id` and `value` but i thought `tokenId` and `tokenValue` were a smidge more descriptive

ghost/core/test/unit/server/services/members/SingleUseTokenProvider.test.js
```diff
@@ -0,0 +1,86 @@
...
+
+        it('should generate different codes for same token value with different ids', function () {
+            const token1 = {id: 'id-1', token: 'same-value'};
+            const token2 = {id: 'id-2', token: 'same-value'};
+            
+            const code1 = tokenProvider.deriveOTC(token1.id, token1.token);
+            const code2 = tokenProvider.deriveOTC(token2.id, token2.token);
+            
+            assert.notEqual(code1, code2);
+        });
```
troyciesco, line 84: Technically these scenarios are (practically) impossible as far as the token fields go, but since deriveOTC knows nothing about the token model I figured these were useful.

ghost/core/core/server/services/members/SingleUseTokenProvider.js
```diff
@@ -106,6 +110,46 @@ class SingleUseTokenProvider {
...
+        if (!this.secret) {
+            throw new ValidationError({
+                message: 'Cannot derive OTC: secret not configured'
+            });
+        }
+
+        if (!tokenId || !tokenValue) {
+            throw new ValidationError({
+                message: 'Cannot derive OTC: tokenId and tokenValue are required'
+            });
```
kevinansfield, line 147: Not essential and probably not really needed in this case but it's good practice to add a `code` to any custom errors. Having a non-changing `code` value means we can change the message text without breaking any error tracking or consumer logic that follows different paths depending on error.

kevinansfield (approved):

PR #24718, opened 2025-08-21
Added admin apps to compose split profile

cmraible:
The current "default" way to run Ghost with docker compose is to use the `ghost` profile, which runs `yarn dev` inside a single container. The `yarn dev` script then orchestrates all the dev servers that need to run, which isn't very "dockerish" and creates some annoying dependencies on `nx watch`, which doesn't play well in Docker.

The `split` profile is intended to eventually replace this `ghost` profile. This profile runs ghost's backend and the admin dev server in separate containers, but currently it does not build any of the admin react apps.

This PR adds the admin react apps in their own service, with a wrapper script that uses `chokidar` to watch & rebuild all the apps on file changes. It's admittedly a little ad-hoc, and this is likely a temporary solution that we can hopefully get rid of as we make progress on migrating away from Ember, but it should well enough for now.

Ultimately once we have the `split` profile at parity with the `ghost` profile, we can remove the `node_modules` volumes and the bind mount, and switch back to using docker compose's `develop:watch:` spec to sync changes into the containers.


9larsons (approved):
No major concerns here as it feels like a consistent WIP and a step in the right direction.

In terms of parity between `ghost` and `split` profiles, I think the biggest gap is that we need a solution for the frontend apps (portal, announcement bar, comments, etc). Currently these only work with the `ghost` profile

cmraible:
> Currently these only work with the `ghost` profile

Yep 100%! That _should_ be an easier problem to solve than the admin apps though, which is why I tried to tackle that first.

PR #24722, opened 2025-08-21
🐛 Fixed accent color styles disappearing in Admin design preview

fueko:
When using the `{{ghost_head}}` helper with the exclude option (for example excluding `{{ghost_head exclude="comment_counts"}}`), the accent color style can disappear in the Admin Preview. 

The issue occurs because the accent style is currently appended to the last style or script element in the head output. In the Admin Preview environment, script tags are sanitized or removed, and if the accent style was attached to one of them, it gets lost. This leads to the accent color not appearing in Admin, while still working correctly on the frontend. The inconsistency is confusing for theme developers and customers testing themes inside Admin.

Fix:

The injection logic has been adjusted so that the accent color is only appended to the last style element in the head output. If no style element exists, it will be added as a separate entry. This ensures that the accent color is never tied to a script element that could be removed in Admin Preview.


9larsons:
Hey, thanks for the PR and apologies for the delays. I wasn't able to reproduce or locate the failure mode this PR guards against: the current Admin preview paths don't strip or sanitize <script> tags in a way that would take the adjacent accent-color <style> with them, and we couldn't find an issue or report describing the loss in practice. The change also reorders {{ghost_head}} output, which breaks the existing snapshot tests. 

I'm going to close this - if you can provide repro steps, please re-open and resolve the tests and we can go from there.

fueko:
Hi! @9larsons ! Thank you for looking into this. I checked again and I’m still seeing the issue on my end. I tested it thoroughly and here’s what happens:

The accent color in the admin appears as the default, despite a custom color being set, when this exclusion set is used:

{{ghost_head exclude="card_assets,comment_counts"}}

Zrzut ekranu 2026-06-10 o 23 39 07

For a single exclusion {{ghost_head exclude="card_assets"}}, the accent color is displayed correctly:

Zrzut ekranu 2026-06-10 o 23 43 08

9larsons:
@fueko Thanks for the extra details! I'll take another look.

9larsons:
Thanks! Confirmed the root cause: in the design-preview path, the announcement-bar helper could return undefined, which got pushed into the `ghost_head` array and the accent-color block's findLastIndex then threw on it, so the catch silently dropped the accent style, code injection, and custom fonts from <head>. It's intermittent because it only surfaces when nothing else with a `</script>/</style>` tag follows the announcement bar (e.g. card assets or member-attribution excluded), which is why it slipped through. A bit squirrely to find.

fueko:
Thanks @9larsons for tracking this down and for the detailed root-cause analysis. I really appreciate the investigation and fix!

PR #24726, opened 2025-08-21
Updated member attribution to use session instead of local storage 

9larsons:
ref https://linear.app/ghost/issue/PROD-1976/
- changed from localStorage to sessionStorage
- added missing /apps/ packages to shipping workflow

In an effort to be more GDPR compliant - and to avoid pop-ups out of the box - we're changing the attribution script to use session storage.

This should have minimal impact as we send the attribution object on the signup request from Portal/signup form. So even though the user needs to check their email to verify their email and click the magic link, the magic link contains the attribution data in the JWT, rather than pulling the data after the magic link click. This would otherwise be problematic as magic links open in a new tab (ie new session).

There will be a brief cutover as we didn't choose to copy-forward or add handling for the local storage values. Given that we only have 24h lookback anyways, moving to session storage without addtl handling seems acceptable.

__Release Changes__
We [introduced](https://github.com/TryGhost/Ghost/pull/22127) a new script for shipping the /apps/ a while back. Other apps were added later, but incrementally, instead of adding all of them once the test was proven to work. The `ci.yml` and `release-apps.js` changes incorporate the missing apps.

__Testing__
- [ ] Ensure page views are saving data to session storage, NOT local storage
- [ ] Ensure attribution is working appropriately (sign up a member on a post or homepage, visit Admin > Members > member to confirm)

### DIFF
--- a/.github/scripts/release-apps.js
+++ b/.github/scripts/release-apps.js
@@ -10,7 +10,8 @@ const CONFIG_KEYS = {
     '@tryghost/portal': 'portal',
     '@tryghost/sodo-search': 'sodoSearch',
     '@tryghost/comments-ui': 'comments',
-    '@tryghost/announcement-bar': 'announcementBar'
+    '@tryghost/announcement-bar': 'announcementBar',
+    '@tryghost/signup-form': 'signupForm'
 };
 
 const CURRENT_DIR = process.cwd();
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1156,6 +1156,12 @@ jobs:
           - package_name: '@tryghost/comments-ui'
             package_path: 'apps/comments-ui'
             cdn_paths: 'https://cdn.jsdelivr.net/ghost/comments-ui@~CURRENT_MINOR/umd/comments-ui.min.js'
+          - package_name: '@tryghost/signup-form'
+            package_path: 'apps/signup-form'
+            cdn_paths: 'https://cdn.jsdelivr.net/ghost/signup-form@~CURRENT_MINOR/umd/signup-form.min.js'
+          - package_name: '@tryghost/announcement-bar'
+            package_path: 'apps/announcement-bar'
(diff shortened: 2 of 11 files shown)


.github/scripts/release-apps.js
```diff
@@ -10,7 +10,8 @@ const CONFIG_KEYS = {
     '@tryghost/portal': 'portal',
     '@tryghost/sodo-search': 'sodoSearch',
     '@tryghost/comments-ui': 'comments',
-    '@tryghost/announcement-bar': 'announcementBar'
+    '@tryghost/announcement-bar': 'announcementBar',
+    '@tryghost/signup-form': 'signupForm'
 };
```
9larsons, line 15: Signup form was missing here. I added announcementBar in a previous PR, but didn't actually realize it was missing/needed to be added to the ci.yml file, too. (below)

.github/workflows/ci.yml
```diff
@@ -1156,6 +1156,12 @@ jobs:
           - package_name: '@tryghost/comments-ui'
             package_path: 'apps/comments-ui'
             cdn_paths: 'https://cdn.jsdelivr.net/ghost/comments-ui@~CURRENT_MINOR/umd/comments-ui.min.js'
+          - package_name: '@tryghost/signup-form'
+            package_path: 'apps/signup-form'
+            cdn_paths: 'https://cdn.jsdelivr.net/ghost/signup-form@~CURRENT_MINOR/umd/signup-form.min.js'
+          - package_name: '@tryghost/announcement-bar'
+            package_path: 'apps/announcement-bar'
+            cdn_paths: 'https://cdn.jsdelivr.net/ghost/announcement-bar@~CURRENT_MINOR/umd/announcement-bar.min.js'
```
9larsons, line 1164: The ci step checks the package.json version for diffs, then handles publishing to nom if it has changed. There's also a step to clear the jsdelivr cache.

cmraible (approved):

PR #24731, opened 2025-08-22
Updated compose to be compatible with locally running analytics service

cmraible:
Building on b16e0a1405fe35df74d6bbe7cfccbcbd3665fb2b, this PR makes it possible to easily run Ghost using your local clone of the Analytics service running in its own docker compose project.

Related PR in the analytics service: https://github.com/TryGhost/TrafficAnalytics/pull/408

Key changes:
- Switched to using a shared named volume for configuration files, so docker services running in other projects can access the same shared config. This allows the analytics service, which runs in a separate docker compose project, to access the shared configuration without hardcoding any assumptions about where the two projects are relative to one another in the filesystem.
- Allowed overriding the proxy target that caddy uses for the `/.ghost/analytics/*` routes, so we can set it to the locally running analytics service's hostname. This falls back to the bundled `analytics` service if not provided.

### DIFF
--- a/.docker/analytics/entrypoint.sh
+++ b/.docker/analytics/entrypoint.sh
@@ -26,16 +26,16 @@ trap _term TERM INT
 
 # Set the TINYBIRD_TRACKER_TOKEN environment variable from the .env file
 # This file is created by the `tb-cli` service and mounted into the Analytics service container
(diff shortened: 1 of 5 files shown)


9larsons (approved):
Works great! 🚀

PR #24746, opened 2025-08-25
Added card assets endpoint in the admin

minimaluminium:
ref https://linear.app/ghost/issue/PROD-2481/oss-issue-cardsmincss-and-js-problems-for-self-hosters

- Added `/ghost/cards/` route for theme-independent CSS/JS card assets
- Updated ActivityPub reader to use admin card assets instead of theme-dependent ones
- Ensured all card styles and functionality available regardless of theme settings

### DIFF
--- a/apps/admin-x-activitypub/src/views/Inbox/components/Reader.tsx
+++ b/apps/admin-x-activitypub/src/views/Inbox/components/Reader.tsx
@@ -20,7 +20,6 @@ import {Activity} from '@src/api/activitypub';
 import {handleProfileClick} from '@src/utils/handle-profile-click';
 import {isPendingActivity} from '../../../utils/pending-activity';
 import {openLinksInNewTab} from '@src/utils/content-formatters';
-import {useBrowseSite} from '@tryghost/admin-x-framework/api/site';
 import {useDebounce} from 'use-debounce';
 import {useNavigate} from '@tryghost/admin-x-framework';
 
@@ -45,7 +44,6 @@ const ArticleBody: React.FC<{
     onIframeLoad?: (iframe: HTMLIFrameElement) => void;
     onLoadingChange?: (isLoading: boolean) => void;
     isPopoverOpen: boolean;
-    siteUrl?: string;
 }> = ({
     postUrl,
     heading,
@@ -59,8 +57,7 @@ const ArticleBody: React.FC<{
     onHeadingsExtracted,
     onIframeLoad,
     onLoadingChange,
-    isPopoverOpen,
-    siteUrl
+    isPopoverOpen
 }) => {
     const iframeRef = useRef<HTMLIFrameElement>(null);
     const [isLoading, setIsLoading] = useState(true);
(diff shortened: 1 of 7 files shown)


apps/admin-x-activitypub/src/views/Inbox/components/Reader.tsx
```diff
@@ -154,12 +151,12 @@ const ArticleBody: React.FC<{
                     waitForImages();
 
                     const script = document.createElement('script');
-                    script.src = '${siteUrl ? `${siteUrl.replace(/\/$/, '')}/public/cards.min.js` : '/public/cards.min.js'}';
+                    script.src = 'cards/admin-cards.min.js';
```
kevinansfield, line 154: This stuck out initially because it makes the script URL dependent on the currently-viewed page, _but_ I think this will be fine as long as we're using `#` URLs for the SPA

ghost/core/core/server/web/admin/card-assets.js
```diff
@@ -0,0 +1,168 @@
...
+            logging.error('Failed to load admin card assets:', error);
+            // Don't throw - allow admin to work without card assets
+        }
+    }
+
+    async ensureLoaded() {
+        if (!this.ready) {
+            await this.load();
+        }
+    }
```
kevinansfield, line 135: If we have multiple requests come in simultaneously we don't have any protection against multiple builds happening at the same time, which will likely error because later builds will be trying to write to files that already exist.

A pattern we've used elsewhere is storing the long-running promise to a variable so we can await the same variable for all requests that come in whilst that single build is running.

kevinansfield (changes requested):
Code looks good barring the potential issue with multiple simultaneous builds but we're completely missing tests for the new admin app endpoint. It would be great to see the tests added along with some tests for the "still running/ready" behaviour.

apps/admin-x-activitypub/src/views/Inbox/components/Reader.tsx
minimaluminium: Switched to absolute URLs in https://github.com/TryGhost/Ghost/pull/24746/commits/3d0d206df1bef0b296f3919d809ab6a74600bf39

ghost/core/core/server/web/admin/card-assets.js
minimaluminium: Added a multiple connection protection in https://github.com/TryGhost/Ghost/pull/24746/commits/60a8690e3f9ed772c64defcc3b1de69c6d22fb5e

apps/admin-x-activitypub/src/views/Inbox/components/Reader.tsx
```diff
@@ -154,12 +151,12 @@ const ArticleBody: React.FC<{
...
                     const script = document.createElement('script');
-                    script.src = '${siteUrl ? `${siteUrl.replace(/\/$/, '')}/public/cards.min.js` : '/public/cards.min.js'}';
+                    script.src = '/ghost/cards/admin-cards.min.js';
                     document.head.appendChild(script);
 
                     const link = document.createElement('link');
                     link.rel = 'stylesheet';
-                    link.href = '${siteUrl ? `${siteUrl.replace(/\/$/, '')}/public/cards.min.css` : '/public/cards.min.css'}';
+                    link.href = '/ghost/cards/admin-cards.min.css';
                     document.head.appendChild(link);
```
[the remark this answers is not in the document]
kevinansfield: Looks like the fix for this was reverted?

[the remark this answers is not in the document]
minimaluminium: I misunderstood the original comment—just reverted it back to the version that supports subdirectories.

PR #24748, opened 2025-08-25
Updated post factory to use API adapter in E2E suite

9larsons:
ref https://linear.app/ghost/issue/PROD-2470/
- introduced api adapter
- added Ghost api adapter wrapper
- updated post factory to use API; we needed to do this instead of directly inserting via knex because of Ghost's caching behavior
- added new sample tests

The post factory isn't terribly useful in the current state due to Ghost's caching and services, so we need to move to an API-based factory. This PR adds the scaffolding for generic http-based APIs.

We authenticate and pass the cookies w/ the request within the Playwright page. Each runner will need to authenticate before proceeding, though this is relatively simple and likely to be required for any factories.

I expect we'll need to tweak these down the road slightly. For now, this is getting us off the ground.


e2e/data-factory/index.ts
```diff
@@ -6,7 +6,43 @@ export * from './factories/user-factory';
...
+}
+
+/**
+ * Create a new PostFactory with API persistence
+ * Each factory gets its own authenticated client for test isolation
+ * 
+ * @param options - Optional overrides for Ghost instance and credentials
+ * @returns PostFactory ready to use with the specified Ghost backend
+ */
+export async function createPostFactory(options?: FactoryOptions): Promise<PostFactory> {
```
9larsons: I don't love this but we need some kind of pattern to use for pointing a given Ghost factory at a given Ghost instance for parallelization. I'm certain we'll change this, that's simply my reasoning for including this.

We need the API client in order to authenticate since we need a session. This is one reason why I think we may ultimately want a GhostFactory sort of implementation; we can supply the url and apply it to all Ghost-related factories rather than treating them all separately.

e2e/data-factory/index.ts
```diff
@@ -1,12 +1,17 @@
...
 export {KnexPersistenceAdapter} from './persistence/adapters/knex';
+export {ApiPersistenceAdapter} from './persistence/adapters/api';
+export {GhostApiAdapter} from './persistence/adapters/ghost-api';
 export type {PersistenceAdapter} from './persistence/adapter';
 
 // Utilities
 export {generateId, generateUuid, generateSlug} from './utils';
+
+// Factory Setup Helpers
+export {createPostFactory} from './setup';
```
9larsons, line 18: This will need a better pattern as we add more Ghost factories. I think we may need/want a central coordinator as we will need to specify the host if we're running several in parallel.

cmraible: I think we could potentially use Playwright's API testing functionality to remove/reduce the setup required. Ref: https://playwright.dev/docs/api-testing#using-request-context 

If we did that, (assuming I'm reading the docs right), we could piggy-back on the auth setup that we already do here (https://github.com/TryGhost/Ghost/blob/ghost-docker-build-ci/e2e/tests/auth.setup.ts#L26-L43) in our setup test.

cmraible (changes requested):
See [comment](https://github.com/TryGhost/Ghost/pull/24748/files#r2299159738)

I think this will eliminate a lot of the boilerplate, and also hopefully address the "Authorization failed" errors we're seeing in the tests in CI.

Other than that, I think the general shape of this is right, nice job!

e2e/data-factory/tests/post-factory.test.ts
```diff
@@ -1,90 +0,0 @@
-import {test, expect} from '@playwright/test';
```
9larsons, line 1: These are the old tests that we can scrap.

e2e/data-factory/index.ts
9larsons: Awesome, this worked great. I had previously implemented reading the auth file which worked well but felt a bit janky - having this baked in w/ context is ideal.

e2e/data-factory/setup/index.ts
```diff
@@ -0,0 +1,29 @@
...
+    
+    // Create context with stored auth state from Playwright setup
+    const context = options?.context || await request.newContext({
+        baseURL,
+        storageState: appConfig.auth.storageFile
+    });
+    
+    const adapter = new GhostApiAdapter(context, 'posts');
+    return new PostFactory(adapter);
+}
```
cmraible: **suggestion (blocking)**: rather than passing an `options` object and manually creating a new request context, I think we should pass the `page` object, and piggyback on the `page.request` context, which will already have the `storageState` and `baseURL` configured appropriately for the current page.

When we switch to real parallelism, each playwright worker will have its own instance of Ghost, and the shared `auth.storageFile` pattern will break. But if we use the request context from the page itself, it will already be configured with the right `baseURL` and `storageState` for the current worker.

```suggestion
export async function createPostFactory(page): Promise<PostFactory> {
    const requestContext = page.request
    const adapter = new GhostApiAdapter(requestContext, 'posts');
    return new PostFactory(adapter);
}
```

e2e/tests/post-factory.test.ts
```diff
@@ -0,0 +1,54 @@
+import {test, expect} from '@playwright/test';
+import {createPostFactory} from '../data-factory';
+import type {PostFactory} from '../data-factory';
+
+test.describe('Post Factory API Integration', () => {
+    let postFactory: PostFactory;
+
+    test.beforeAll(async () => {
+        postFactory = await createPostFactory();
+    });
```
cmraible: Expanding on my [comment above](https://github.com/TryGhost/Ghost/pull/24748/files#r2301631022), this would then become: 

```suggestion
    test.beforeEach(async ({page}) => {
        postFactory = await createPostFactory(page);
    });
```

e2e/data-factory/persistence/adapters/ghost-api.ts
```diff
@@ -0,0 +1,39 @@
...
+                endpoint: '/ghost/api/admin/posts?formats=mobiledoc,lexical,html',
+                wrapRequest: (data: EntityData) => ({posts: [data]}),
+                extractResponse: (response: GhostApiResponse) => {
+                    if (response.posts && Array.isArray(response.posts)) {
+                        return response.posts[0];
+                    }
+                    return response;
+                }
+            }
+        };
```
cmraible: **suggestion (blocking)**: including the configs for each endpoint violates the single responsibility principle IMO - now this class is both an API adapter and also a configuration registry. I think it would be cleaner to accept this configuration object into the constructor — this way the `createPostFactory()` function is the only thing that needs to change when updating how the post factory works. 

With that, we _might_ not even need this extra layer of abstraction between api / ghost-api. Or, maybe this becomes `GhostAdminApiAdapter`, and adds a reasonable default implementation for `wrapRequest`, `extractResponse`, and maybe handles the `/ghost/api/admin` endpoint prefix.

e2e/data-factory/persistence/adapters/api.ts
```diff
@@ -0,0 +1,86 @@
...
+            throw new Error(`Failed to create ${entityType}: ${response.status()}`);
+        }
+        
+        const body = await response.json() as TResponse;
+        return this.extractResponse(body) as T;
+    }
+    
+    async findById<T>(entityType: string, id: string): Promise<T | null> {
+        const baseEndpoint = this.endpoint.split('?')[0];
+        const queryParams = this.endpoint.includes('?') ? '?' + this.endpoint.split('?')[1] : '';
```
cmraible: **nit (non-blocking)**: rather than parsing these from the configured endpoint, I think it would be cleaner to configure the `endpoint` and `defaultQueryParams` separately to begin with, to avoid this kind of ad-hoc parsing.

e2e/data-factory/persistence/adapters/api.ts
```diff
@@ -0,0 +1,86 @@
...
+        const response = await this.context.get(`${baseEndpoint}/${id}${queryParams}`);
+        
+        if (response.status() === 404) {
+            return null;
```
cmraible: **suggestion (non-blocking):** IMO we should explicitly throw an error here, so this method either returns a post successfully, or throws an error (and thus fails the test, unless the error is specifically handled). 

By returning `null` in the 404 case, I'm afraid we'll end up with a bunch of optional chaining or conditional blocks, which could be avoided by throwing a "Not found" error here instead.
```
const post = await postFactory.findById(id);
expect(post?.status).toEqual('published');
```

cmraible (changes requested):

e2e/data-factory/persistence/adapters/api.ts
9larsons: Agreed, the current implementation wasn't very robust.

e2e/data-factory/persistence/adapters/api.ts
9larsons: Yep good call, agreed.

e2e/data-factory/persistence/adapters/ghost-api.ts
9larsons: Yep, also agreed. I tried a couple options and I think the `GhostAdminApiAdapter` feels good for the moment. I like the idea of removing it completely to remove a layer, but I don't like that we'd be repeating the `wrapRequest` & `extractResponse` handling.

e2e/data-factory/persistence/adapters/knex.ts
```diff
@@ -57,12 +47,16 @@ export class KnexPersistenceAdapter implements PersistenceAdapter {
...
         const result = await this.db(entityType)
             .where('id', id)
             .first();
         
-        return result || null;
+        if (!result) {
+            throw new Error(`${entityType} with id ${id} not found`);
+        }
+        
+        return result;
```
cmraible, line 59: Nice, good call on updating the knex adapter too!

cmraible (approved):
Nice one! Thanks for implementing the feedback, looks really good to me!

e2e/data-factory/persistence/adapters/api.ts
```diff
@@ -0,0 +1,89 @@
...
+ */
+export class ApiPersistenceAdapter<TRequest = unknown, TResponse = unknown> implements PersistenceAdapter {
+    protected apiContext: APIRequestContext;
+    protected endpoint: string;
+    protected queryParams: Record<string, string>;
+    protected wrapRequest: (data: TRequest) => unknown;
+    protected extractResponse: (response: TResponse) => unknown;
+    
+    constructor(options: ApiAdapterOptions<TRequest, TResponse>) {
+        this.apiContext = options.context.request;
```
ibalosh: hey @9larsons , I know that context is passed, based on recommendation by @cmraible to auth with API, which I think it's a nice idea, it auto sets up auth and uses it in API 🙌

My only concern here is that we are coupling it to playwright. I think it would be better that api adapter doesn't  know about browser context, nor api request context. If we need too, we can use api request context instead of bigger browser context. 

Adapters though just do crud requests, transform data in our case, give back responses, with httpClient we passed to it. 

To keep them abstract as we did so far, we could do provide a minimal interface to hide that complexity and remove coupling from specific API implementation and we would not import playwright for adapters.

Check this out, if you add the following to `api.ts` to expose just as much as you need:

```typescript
export interface HttpResponse {
    ok(): boolean;
    status(): number;
    json(): Promise<unknown>;
}

// Generic client that works with any response type that has the required methods
export interface HttpClient<TResponse extends HttpResponse = HttpResponse> {
    get(url: string, options?: {headers?: Record<string, string>}): Promise<TResponse>;
    post(url: string, options?: {data?: unknown, headers?: Record<string, string>}): Promise<TResponse>;
    put(url: string, options?: {data?: unknown, headers?: Record<string, string>}): Promise<TResponse>;
    delete(url: string, options?: {headers?: Record<string, string>}): Promise<TResponse>;
}

interface ApiAdapterOptions<TRequest = unknown, TResponse = unknown> {
    httpClient: HttpClient;
    ...
}


export class ApiPersistenceAdapter<TRequest = unknown, TResponse = unknown> implements PersistenceAdapter {
    **protected apiContext: HttpClient;**

    constructor(options: ApiAdapterOptions<TRequest, TResponse>) {
        **this.apiContext = options.httpClient;** // we can call it now this.httpClient
    }
}

export class GhostAdminApiAdapter extends ApiPersistenceAdapter<unknown, GhostApiResponse> {
    constructor(httpClient: HttpClient, resourcePath: string, queryParams?: Record<string, string>) {
      ....
        super({
            httpClient,
            ....
        });
    }
}
```

**setup/index.ts**

```typescript
export function createPostFactory(page: Page): PostFactory {
    const adapter = new GhostAdminApiAdapter(
        page.context().request,
        'posts',
        {formats: 'mobiledoc,lexical,html'}
    );
    return new PostFactory(adapter);
}
```

With this change, api is just doing api stuff, not knowing that playwright will do bit more.

Or we just use api request context.

e2e/data-factory/persistence/adapters/ghost-api.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+/**
+ * Ghost Admin API adapter that handles Ghost's API formatting conventions
+ * All Ghost Admin API endpoints follow the pattern:
+ * - Endpoint: /ghost/api/admin/{resource}
+ * - Request wrapper: { [resource]: [data] }
+ * - Response extractor: response[resource][0]
+ */
+export class GhostAdminApiAdapter extends ApiPersistenceAdapter<unknown, GhostApiResponse> {
```
ibalosh, line 14: **nit:** what is the reasoning behind having separate GhostAdminApiAdapter and base ApiAdapter, seems we don't need it, maybe too early to split it out?

e2e/data-factory/persistence/adapters/api.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+    
+    protected buildUrl(path?: string): string {
+        const url = path ? `${this.endpoint}/${path}` : this.endpoint;
+        const params = new URLSearchParams(this.queryParams);
+        const queryString = params.toString();
+        return queryString ? `${url}?${queryString}` : url;
+    }
+    
+    async insert<T>(entityType: string, data: T): Promise<T> {
```
ibalosh, line 53: If I understand correctly the new API implementation is using unique adapter per endpoint? 

It can work too, just seems to work now differently then original concept, where based on `entityType` we would insert, remove etc. This way it works differently than Knex implementation too, and differently than the original api adapter interface is was intended.

This is ok too, but the `entityType`  has no function here now, it just shows up in errors, while the execution is actually dependent on construction where we set things up. 

I guess if we move to separate api per every factory, we might not need the entityType at all?

e2e/data-factory/persistence/adapters/ghost-api.ts
```diff
@@ -0,0 +1,31 @@
...
+    constructor(context: BrowserContext, resourcePath: string, queryParams?: Record<string, string>) {
+        // Extract the resource name from the path (e.g., 'posts' from 'posts' or 'posts/1234')
+        const resource = resourcePath.split('/')[0];
+        
+        super({
+            context,
+            endpoint: `/ghost/api/admin/${resourcePath}`,
+            queryParams: queryParams || {},
+            wrapRequest: data => ({[resource]: [data]}),
```
ibalosh: **question/nit:** would this work for all Ghost APIs we plan to use (the format of requests)? Just want to make sure we keep object as open for extension, if that is not the case

ibalosh: I think we should use a different name here, we are not wrapping request, but transforming data?

e2e/data-factory/persistence/adapters/ghost-api.ts
```diff
@@ -0,0 +1,31 @@
...
+            extractResponse: (response: GhostApiResponse) => {
```
ibalosh: I would also use transformResponse here

ibalosh (dismissed):

e2e/data-factory/persistence/adapters/ghost-api.ts
ibalosh: hey fellas, I checked the PR, overall I like the direction. My biggest question/concern is this one - Move to separate instance of adapter per factory, if I understood the new implementation correctly. I like this idea.

It removed the need for mapper for each factory in adapter, but I just wanted to make sure we use same logic for all adapters, or whether we are fine with mismatch between Knex and Api?

Seems we use a different one now for Knex and for Api (one dependent on entityType, another not from what I see).

With the new approach, if you look at PersistenceAdapter interface, this new approach is not as clear. One would think that `insert<T>(entityType: string, data: T): Promise<T>;` inserts data for specific entityType, which is not the case. 

What actually happens is that on construction, we decide that, and then do the action. So we kind of use maping on Knex side in simplified way (entityType) to allow , and not using it on Api side.

I am fine with not using mapper, it's complex, and this removes a lot of it, just want to make sure that methods have same expectations, on adapters.

e2e/data-factory/persistence/adapters/ghost-api.ts
9larsons: The idea is that we're going to use ApiAdapter as a base for other APIs, like Tinybird.

e2e/data-factory/persistence/adapters/api.ts
9larsons: I'm not entirely following... I think I'm missing some context for these comments. I'm going to merge what we have and we can revisit as we expand.

PR #24751, opened 2025-08-26
Added vscode configuration for eslint

cmraible:
Configuring VSCode to fix linting errors on save is a common pain point for new contributors just getting started with the codebase. This configuration, combined with the ESLint extension have always worked for me, so I figure we should add these to source control so new contributors don't have to figure this out for themselves anymore.


cmraible:
Closing for lack of interest, and I'm not using VSCode at the moment. Happy to reopen if anyone finds this useful.

PR #24754, opened 2025-08-26
Display OTC in member signin emails when requested

kevinansfield:
closes https://linear.app/ghost/issue/PROD-2490

- updated signin handler in members API controller to pass the `otc: true` option when the API request to send the magic link `otc: true` in it's body params
- added OTC generation and display in MagicLink.js
  - added wrapper methods around `SingleUseTokenProvider` for fetching `id` and `otc` from a `token` value following the `getDataFromToken(token)` pattern
  - updated default HTML and text methods to include OTC and related text
- added `SingleUseTokenProvider.getIdFromToken` method
  - allows for lookup of the token id for OTC generation when only the token value is known
- updated sign-in subject to include the words "verification token" to help trigger automatic 2FA code fill on Apple devices
- updated sign-in email text to include OTC when requested
- updated sign-in email HTML to include OTC when requested


ghost/core/core/server/services/members/emails/signin.js
```diff
@@ -117,7 +121,21 @@ module.exports = ({t, siteTitle, email, url, accentColor = '#15212A', siteDomain
...
+                          <table border="0" cellpadding="0" cellspacing="0" class="btn btn-primary" style="border-collapse: separate; mso-table-lspace: 0pt; mso-table-rspace: 0pt; width: 100%; box-sizing: border-box;">
+                            <tbody>
+                              <tr>
+                                <td style="padding: 16px; background-color: #F4F5F6; border-radius: 8px; text-align: center; vertical-align: middle;" valign="middle">
+                                  <h2 style="text-align: center; vertical-align: center; letter-spacing: 5px; font-size: 24px; color: #15212A; font-weight: 600; line-height: 24px; margin: 0;">${otc}</h2>
+                                </td>
+                              </tr>
+                            </tbody>
+                          </table>
+                          <p style="font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'; font-size: 16px; color: #3A464C; font-weight: normal; margin: 0; line-height: 24px; margin-bottom: 32px;">${t('Or use this link to securely sign in:')}</p>
```
kevinansfield: Needs some spacing between table and next paragraph.

ghost/core/core/server/services/members/members-api/controllers/RouterController.js
```diff
@@ -679,7 +680,15 @@ module.exports = class RouterController {
...
-        const {emailType} = req.body;
+        const {emailType, otc} = req.body;
+
+        const options = {
+            otc: false
+        };
+
+        if (labs.isSet('membersSigninOTC') && (otc === true || otc === 'true')) {
+            options.otc = true;
+        }
```
kevinansfield: Not keen on using `options` here as it creates some inconsistency with other methods, also change naming to `includeOTC` when value is a boolean rather than an actual OTC value.

ghost/core/core/server/services/lib/magic-link/MagicLink.js
```diff
@@ -112,6 +129,38 @@ class MagicLink {
         return this.getSigninURL(token, type, options.referrer);
     }
 
+    /**
+     * getIdFromToken
+     *
+     * @param {Token} token - The token to get the id from
+     * @returns {Promise<string|null>} id - The id of the token
+     */
+    async getIdFromToken(token) {
```
troyciesco, line 138: this and `getOTCFromToken` could be private I think? I'll leave them for now, not sure if they need to be public for a future ticket

troyciesco (approved):

ghost/core/core/server/services/lib/magic-link/MagicLink.js
```diff
@@ -146,26 +201,38 @@
...
+function defaultGetHTML(url, type, email, otc) {
     let msg = 'sign in';
     if (type === 'signup') {
         msg = 'confirm your email address';
     }
+
+    if (otc) {
+        return `Enter the code ${otc} or <a href="${url}">click here to ${msg}</a> This msg was sent to ${email}`;
+    }
+
```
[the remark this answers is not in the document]
troyciesco: @kevinansfield I'm pretty sure this is ok, i think the email address gets validated before this (and actually, prob wouldn't even send lol) and the other two values are generated by the server, but just calling it out.

(@coderabbitai don't learn from my comment above, bc i still want you to call out stuff like this)

PR #24755, opened 2025-08-26
Added OTC input to portal

troyciesco:
closes https://linear.app/ghost/issue/PROD-2494

- If the `membersSigninOTC` labs flag is on and `sendMagicLink` returns an `otc_ref`, the portal will display an input to enter an OTC
- The input is currently stubbed out to the point where submitting the form will just log the token id and otc to the console
- If the labs flag is off, the existing sign in flow still occurs


apps/portal/src/App.js
```diff
@@ -978,6 +978,8 @@ export default class App extends React.Component {
             t,
             dir,
             scrollbarWidth,
+            labs,
```
troyciesco, line 981: I think it makes sense to add labs here so we can get the value from the context

apps/portal/src/utils/api.js
```diff
@@ -298,7 +298,12 @@ function setupGhostApi({siteUrl = window.location.origin, apiUrl, apiKey}) {
             });
 
             if (res.ok) {
-                return 'Success';
+                try {
+                    const data = await res.json();
+                    return data;
+                } catch (e) {
+                    return 'Success';
```
troyciesco: gonna come back to this because returning "Success" on an error is weird - probably need to reference the flag and/or otc

apps/portal/src/components/pages/MagicLinkPage.js
```diff
@@ -81,12 +97,100 @@ export default class MagicLinkPage extends React.Component {
...
+        }
+
+        // @TODO: action implementation TBD
+        const isRunning = (action === 'verifyOTC:running');
+        const isError = (action === 'verifyOTC:failed');
+
+        return (
+            <section>
+                <div className='gh-portal-section'>
+                    <InputField
```
troyciesco, line 158: i started with `InputForm` but I realized that not a lot of props pass through that component (i.e. maxlength) so I switched to `InputField` and added the props manually.

Since we'll change the design of this before GA i figured it made sense to just go with this for now.

apps/portal/src/components/pages/MagicLinkPage.test.js
```diff
@@ -1,32 +1,354 @@
...
+        });
+
+        test('calls close popup action when close button clicked', () => {
+            const {getByRole, mockOnActionFn} = setupTest();
+            const closeBtn = getByRole('button', {name: 'Close'});
+            
+            fireEvent.click(closeBtn);
+            
+            expect(mockOnActionFn).toHaveBeenCalledWith('closePopup');
+        });
```
troyciesco, line 89: These were the two original tests in the file, figured they were small enough that i could use the new `setupTest`

apps/portal/src/tests/SigninFlow.test.js
```diff
@@ -335,3 +337,158 @@ describe('Signin', () => {
         });
     });
 });
+
+describe('OTC Integration Flow', () => {
```
troyciesco, line 341: Since there were more tests in this file, I didn't make everything as fully repeatable as I did in MagicLinkPage.test.js, I just did it for the OTC ones. I can follow up with a cleaned up file here.

apps/portal/src/components/pages/MagicLinkPage.js
```diff
@@ -81,12 +98,100 @@ export default class MagicLinkPage extends React.Component {
...
+        }, () => {
+            // eslint-disable-next-line no-unused-vars
+            const {otc, errors} = this.state;
+            const {otcRef} = this.context;
+            const hasFormErrors = (errors && Object.values(errors).filter(d => !!d).length > 0);
+            if (!hasFormErrors && otcRef) {
+                // @TODO: replace with verifyOTC action
+                // For now, just log for development
+                // eslint-disable-next-line no-console
+                console.log('otc_ref and otc submitted');
```
troyciesco, line 129: @kevinansfield the rabbit complained about this line logging the token and otc, so i switched to this from `console.log(`token: ${otcRef} otc: ${otc}`);`, but that's what i used for testing

apps/portal/src/components/pages/MagicLinkPage.js
kevinansfield: I've [updated `InputField` so it adds the attributes needed for Apple's auto-fill to work](https://github.com/TryGhost/Ghost/pull/24755/commits/6077821692d26824e5433cdd6f898fdbe5d14ffe). Tested in Safari and it worked with our current OTC email content 👌

kevinansfield: ~Design-wise it will only be CSS that changes, we'll still have the same input field underneath~

Ignore that, I hadn't checked what ShadCN was doing. It creates a secondary set of individual inputs 😅

kevinansfield (approved):

PR #24759, opened 2025-08-27
Downgraded nx to 20.8.0

cmraible:
ref https://github.com/nrwl/nx/issues/32524

Version 20.12.0 has been deleted from npm. Downgrading to the latest compatible version available on npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,10 +88,12 @@
     "inquirer": "8.2.7",
     "jsonc-parser": "3.3.1",
     "lint-staged": "15.5.2",
-    "nx": "20.12.0",
     "patch-package": "8.0.0",
     "postinstall-postinstall": "2.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -88,10 +88,12 @@
...
     "jsonc-parser": "3.3.1",
     "lint-staged": "15.5.2",
-    "nx": "20.12.0",
     "patch-package": "8.0.0",
     "postinstall-postinstall": "2.1.0",
     "rimraf": "5.0.10",
     "typescript": "5.8.3"
+  },
+  "dependencies": {
+    "nx": "20.8.0"
```
tadhglewis, line 97: @cmraible should this _downgrade_ also be removing it as a dev dependency? seems odd

cmraible: Yep good catch - thanks!

PR #24761, opened 2025-08-27
Upgrade rewire package to resolve conflicting eslint versions

jonatansberg:
The previous version of `rewire` had a pinned dependency on an older version of `eslint` that would conflict with the version from the root `package.json`. This would sometimes result in unexpected linter behavior, like different linting rules being applied depending on what file you open first.

Updating the rewire package resolved the conflict.


ErisDS (approved):
Great find!

PR #24763, opened 2025-08-27
Extracted test setup in magic-link unit test

kevinansfield:
no issue

- removed some oft-repeated setup to clean up in preparation for adding more unit tests


troyciesco (approved):
it looks like there's some linting to fix, but looks good otherwise!

PR #24766, opened 2025-08-27
Fixed subscriber count history in Analytics > Newsletters

9larsons:
ref https://linear.app/ghost/issue/PROD-2525/
- aggregation type was incorrect for summarizing subscriber data ('exact' vs 'sum')
- updated /stats/subscriber-count API to return absolute values, not deltas; this is more useful for our use case
- renamed return to 'values' instead of 'deltas' to correspond with API changes

Root problem: we were using the 'exact' aggregation when calling `sanitizeChartData` which was an oversight and incorrect when using delta counts. In general, this function is doing too much, but that's a problem to sort out at a later date.

Trying to handle cumulative values was pushing the boundaries of `sanitizeChartData`, which is shared by several other charts. In the spirit of limiting the blast radius of this change, I adjusted the API rather than the util used by most of the front end, given we don't have good test coverage.

The API now returns absolute values for each day. Since we don't display a +/- chart, we don't need individual deltas. This is easier to reason about and takes the load off the frontend from doing additional processing; it just needs to pick the date it wants to display.

Note: We still pull deltas on the backend because that's how the data is stored.

__Testing__
1. `yarn docker:reset`
2. `yarn dev` and set up Ghost
3. Kill process, `yarn reset:data:analytics`, then run Ghost again
4. `yarn tb` in a separate container; make sure config is appropriately set (workspaceId, token, stats.id = `mock_site_uuid`)

This should get you data in Analytics, and the generated data should be enough to generate for the past 12mo which is enough to see the original issue. You may need to do some jockeying with `config` to make it work, or manually load the generated `analytics_events.ndjson` in to Tinybird. The key thing to validate is that timeframes make sense and that data points line up, e.g. 'Last 3 months' summaries for m-1 and m-2 line up with the same for 'YTD' and 'Last 12 months'.

### DIFF
--- a/ghost/core/test/unit/server/services/stats/posts.test.js
+++ b/ghost/core/test/unit/server/services/stats/posts.test.js
@@ -1851,11 +1851,9 @@ describe('PostsStatsService', function () {
             await _createMember('member3');
 
             // Current state
-            await db('members_newsletters').insert([
-                {id: 'mn1', member_id: 'member1', newsletter_id: newsletterId},
-                {id: 'mn2', member_id: 'member2', newsletter_id: newsletterId},
-                {id: 'mn3', member_id: 'member3', newsletter_id: newsletterId}
-            ]);
+            await _createMemberNewsletterSubscription('member1', newsletterId);
+            await _createMemberNewsletterSubscription('member2', newsletterId);
+            await _createMemberNewsletterSubscription('member3', newsletterId);
 
             // Events spanning multiple days
             await _createNewsletterSubscription(newsletterId, 'member1', true, new Date('2024-01-01T00:00:00Z'));
--- a/ghost/core/test/unit/server/services/stats/posts.test.js
+++ b/ghost/core/test/unit/server/services/stats/posts.test.js
@@ -82,7 +82,7 @@ describe('PostsStatsService', function () {
             attribution_url: `/${postId.replace('post', 'post-')}/`,
             referrer_source: referrerSource,
             referrer_url: referrerSource ? `https://${referrerSource}` : null,
-            created_at: moment(createdAt).utc().format('YYYY-MM-DD HH:mm:ss'),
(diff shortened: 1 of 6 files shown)


ghost/core/test/unit/server/services/stats/posts.test.js
```diff
@@ -1587,4 +1622,257 @@ describe('PostsStatsService', function () {
...
+
+            // Current subscriptions (total = 2)
+            await _createMemberNewsletterSubscription('member1', newsletterId);
+            await _createMemberNewsletterSubscription('member2', newsletterId);
+
+            // Subscribe/unsubscribe events
+            await _createNewsletterSubscription(newsletterId, 'member1', true, new Date('2024-01-01T00:00:00Z'));
+            await _createNewsletterSubscription(newsletterId, 'member2', true, new Date('2024-01-02T00:00:00Z'));
+            await _createNewsletterSubscription(newsletterId, 'member3', false, new Date('2024-01-03T00:00:00Z'));
+            await _createNewsletterSubscription(newsletterId, 'member3', true, new Date('2024-01-03T00:00:00Z'));
```
cmraible: I'm a little confused on the cases here — I don't think these events add up to the current subscriptions above. This looks like it should be a total of 3 subscriptions by the end state, not 2

9larsons: Member 3 is someone who subscribed and unsubscribed. Members 1 and 2 are current subscriptions. The `members_newsletters` table only contains current subscribers. So the unsubscribe from member3 would cause them to drop from that table. 

Is that what you mean needs clarification?

cmraible: Oh gotcha — I was just reading them in order, and it looks like member3 unsubscribes, then subscribes, but that makes sense.

cmraible (approved):
Nice one! Thanks for adding the tests 🎉

PR #24767, opened 2025-08-27
Added "all" compose profile for convenience

cmraible:
When using compose profiles, it can be annoying to specify all running profiles when running `docker compose down`. For example, if you have the analytics and split profiles running, you need to run `docker compose --profile analytics --profile split down` to stop all the running containers. This can also create issues when running `yarn docker:clean` or `yarn docker:build`, as any leftover containers will cause `yarn docker:clean` to fail.

This adds a new "all" profile to all services, which can be used to target every service in the compose file at once - e.g. to stop all running services, `docker compose --profile all down`. It also updates some of the package.json scripts to use the "all" profile where it makes sense - building, resetting, and taking all services down.

This gives us more succinct commands and more predictable behavior, while still allowing you to more granularly control which services you want to target by dropping down to raw `docker compose` commands.


9larsons (approved):
Yes please 🙏

PR #24777, opened 2025-08-28
Fixed logic for sending otc_ref

troyciesco:
no issue

- small cleanup to make sure that on sendMagicLink, otc_ref is only sent in the response if both 1) the labs flag is on and 2) the original request body included `otc: true`


troyciesco:
Closed in favor of other naming logic

PR #24778, opened 2025-08-28
Added verifyOTC to SingleUseTokenProvider

troyciesco:
closes https://linear.app/ghost/issue/PROD-2498

- Adds the `verifyOTC` method to `SingleUseTokenProvider`
- `verifyOTC` checks for required fields, finds the token, derives the counter, then uses `hotp.verify` to make sure the OTC is correct

### DIFF
--- a/ghost/core/core/server/services/members/SingleUseTokenProvider.js
+++ b/ghost/core/core/server/services/members/SingleUseTokenProvider.js
@@ -152,6 +152,34 @@ class SingleUseTokenProvider {
     }
 
     /**
+     * @method verifyOTC
+     * Verifies an OTC (one-time code) against a token ID by looking up the token and performing verification
+     *
+     * @param {string} tokenId - Token ID
+     * @param {string} otc - The one-time code to verify
+     * @returns {Promise<boolean>} Returns true if the OTC is valid, false otherwise
+     */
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/services/members/SingleUseTokenProvider.js
```diff
@@ -151,6 +151,34 @@ class SingleUseTokenProvider {
...
 
+    /**
+     * @method verifyOTC
+     * Verifies an OTC (one-time code) against a token ID by looking up the token and performing verification
+     *
+     * @param {string} tokenId - Token ID
+     * @param {string} otc - The one-time code to verify
+     * @returns {Promise<boolean>} Returns true if the OTC is valid, false otherwise
+     */
+    async verifyOTC(tokenId, otc) {
```
kevinansfield: I wonder if it makes sense to use `otcRef` here instead of `tokenId` and keep `token.id/uuid` as an implementation detail of this method?

kevinansfield (approved):
Aside from the naming switch to better match the current code this is good to merge 🙌

ghost/core/core/server/services/members/SingleUseTokenProvider.js
troyciesco: agree!

PR #24779, opened 2025-08-28
Added OTC verification endpoint and sign in redirect params

troyciesco:
closes https://linear.app/ghost/issue/PROD-2497

- adds an endpoint that verifies an OTC and returns a magic link with the token and a hash of the otc. This allows the caller to use the magic link to sign in
- when a user routes to the magic link, the otcVerificationHash is checked - the otc and token are verified, and a timestamp is checked so that the otcVerificationHash is only valid for 5 minutes 
- this pattern allows us to use the same logic for OTC as signing in with a magic link, but with a way to tweak a few things specifically for OTC later, such as handling usage limits, brute force checks, etc.


ghost/core/core/server/services/members/otc-hash-utils.js
```diff
@@ -0,0 +1,46 @@
+const crypto = require('crypto');
```
troyciesco: def not married to this file name. it also kinda felt more like a helper function than something that belonged on `SingleUseTokenProvider` (and def not `RouterController`) though I could be convinced otherwise

ghost/core/core/server/web/members/app.js
```diff
@@ -85,6 +85,14 @@ module.exports = function setupMembersApp() {
             return membersService.api.middleware.sendMagicLink(req, res, next);
         }
     );
+    membersApp.post(
+        '/api/verify-otc', 
+        bodyParser.json({limit: '5mb'}),
```
kevinansfield: Does this need to be 5mb? I would have thought the [default of 100kb](https://expressjs.com/en/resources/middleware/body-parser.html#limit) would be enough for this endpoint as it only receives the 6-digit OTC and a reference?

ghost/core/core/server/services/members/members-api/controllers/RouterController.js
```diff
@@ -649,6 +651,96 @@ module.exports = class RouterController {
...
+        const siteUrl = this._urlUtils.urlFor({relativeUrl: '/members/'}, true);
+        const redirectUrl = new URL(siteUrl);
+
+        // required params for magic link compatibility
+        redirectUrl.searchParams.set('token', token);
+        redirectUrl.searchParams.set('otc_verification', otcVerificationHash);
+
+        // basic URL - frontend will add referrer and action
+        return redirectUrl.toString();
+    }
```
kevinansfield: We need an `?r=` query param on this URL too, otherwise you won't get redirected back to the page you logged in on. I think we should adapt the [existing getSigninURL method](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/services/members/MembersConfigProvider.js#L85-L94) to have an optional `otcVerification` argument that adds the `otc_verification` param and use that so we aren't duplicating signin URL building. We'll also need to track the referrer value through the system so we can redirect correctly.

kevinansfield: I've made the change to use `magicLinkService.getSigninURL()`. For the referrer I think we should look at that in a different PR, this one has already grown quite large 😬

ghost/core/core/server/services/members/members-api/controllers/RouterController.js
```diff
@@ -649,6 +651,85 @@ module.exports = class RouterController {
...
+
+        const tokenProvider = this._magicLinkService.tokenProvider;
+        // Validate OTC format using provider (fallback to 6 digits if unavailable)
+        const isFormatValid = tokenProvider?.isOTCFormatValid?.(otc) ?? /^\d{6}$/.test(otc);
+
+        if (!isFormatValid) {
+            throw new errors.BadRequestError({
+                message: tpl(messages.invalidCode)
+            });
+        }
```
kevinansfield: This feels a bit unnecessary because we'll fail with the same error later on if the OTC doesn't match what is expected. Shall we remove to reduce the amount of code or was there a particular reason this was added?

ghost/core/core/server/services/members/members-api/controllers/RouterController.js
```diff
@@ -649,6 +651,96 @@ module.exports = class RouterController {
...
 
+    async verifyOTC(req, res) {
+        const {otc, otcRef} = req.body;
+
+        if (!otc || !otcRef) {
+            throw new errors.BadRequestError({
+                message: tpl(messages.badRequest),
+                context: 'otc and otc_ref are required'
+            });
+        }
```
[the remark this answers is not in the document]
kevinansfield: We only want to support `otcRef` in the request body. The inconsistent error message has been fixed.

ghost/core/core/server/web/members/app.js
troyciesco: oh nope, i just didn't realize there was a default and i mimicked the one with the lowest limit in this file. fixed!

ghost/core/core/server/services/members/members-api/members-api.js
```diff
@@ -235,12 +235,12 @@ module.exports = function MembersAPI({
         });
     }
 
-    async function getTokenDataFromMagicLinkToken(token) {
-        return await magicLinkService.getDataFromToken(token);
+    async function getTokenDataFromMagicLinkToken(token, otcVerification) {
+        return await magicLinkService.getDataFromToken(token, otcVerification);
     }
```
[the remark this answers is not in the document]
kevinansfield: `otcVerification` is optional and dependent on the signin/signup flow that's being used so doesn't need to be updated everywhere

ghost/core/core/server/services/members/members-api/controllers/RouterController.js
```diff
@@ -649,6 +651,85 @@ module.exports = class RouterController {
...
+                }));
+            }
+
+            const tokenValue = await tokenProvider.getTokenByRef(otcRef);
+            if (!tokenValue) {
+                res.writeHead(400, {'Content-Type': 'application/json'});
+                return res.end(JSON.stringify({
+                    valid: false,
+                    message: tpl(messages.invalidCode)
+                }));
```
kevinansfield: We should be following the newer pattern of `throw new errors.BadRequestError(...` rather than writing to the response directly. The direct `res.*` approach is a very old hangover from when `members-api` was a separate app from Ghost

ghost/core/core/server/services/members/members-api/controllers/RouterController.js
troyciesco: i'd gone back and forth on errors, and originally had a specific message if an OTC was too short - but then kind of leaned towards the more generic message (similar to a vague "username or password is incorrect"). 

I was also trying to catch something like this before doing the further verification processing, but it's probably more than necessary.

even besides the fact that the specific length check probably isn't necessary, looking at this now i think it's in the wrong place anyway - verifyOTC in SingleUseTokenProvider would make more sense.

i'm gonna test the error flow again locally (using the otc-portal-submit branch) just to double-check but otherwise will update!

ghost/core/test/unit/server/services/members/members-api/controllers/RouterController.test.js
```diff
@@ -1097,4 +1097,313 @@ describe('RouterController', function () {
...
+                mockMagicLinkService.tokenProvider.verifyOTC.resolves(true);
+                mockMagicLinkService.tokenProvider.getTokenByRef.resolves(OTC_TEST_CONSTANTS.TOKEN_VALUE);
+                mockSettingsCache.get.withArgs('members_email_auth_secret').returns(OTC_TEST_CONSTANTS.VALID_SECRET);
+            });
+
+            it('should return success response with valid inputs', async function () {
+                // use a fake time so we get a stable OTC verification hash for our mocked token/secret
+                sinon.useFakeTimers(new Date('2021-01-01'));
+                const TEST_OTC_VERIFICATION_HASH = '1609459200:c3784e7545cd61c87b34b9bd6d7b840c1225679c74fbc04bc07302a7a1c6aed4';
+
```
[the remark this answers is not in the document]
kevinansfield: This is already handled by the `sinon.restore()` call in the top-level `afterEach` hook

ghost/core/test/unit/server/services/members/members-api/controllers/RouterController.test.js
```diff
@@ -1097,4 +1097,313 @@ describe('RouterController', function () {
...
+            it('should throw BadRequestError when otc is missing', async function () {
+                req.body.otc = undefined;
+
+                try {
+                    await routerController.verifyOTC(req, res);
+                    assert.fail('Expected BadRequestError to be thrown');
+                } catch (error) {
+                    assert(error instanceof errors.BadRequestError);
+                    assert.equal(error.message, OTC_TEST_CONSTANTS.ERROR_MESSAGES.BAD_REQUEST);
+                    assert.equal(error.context, OTC_TEST_CONSTANTS.ERROR_MESSAGES.OTC_REQUIRED);
```
kevinansfield: We shouldn't ever be testing exact error message content unless they contain dynamic data. If we're expecting a specific error of type `x` then the error should have a unique `code` set on it and tested against.

ghost/core/core/server/services/members/members-api/controllers/RouterController.js
```diff
@@ -649,6 +651,85 @@ module.exports = class RouterController {
...
+            }
+
+            const otcVerificationHash = await this._createHashFromOTCAndToken(otc, tokenValue);
```
kevinansfield: Having business logic for creating an otc verification hash inside the RouterController feels wrong when it's the tokenProvider that ultimately validates it. Can we move the creation to `SingleUseTokenProvider` so the creation+validation both live in the same place? That would also mean the odd floating util for hash creation can live in that class too

ghost/core/test/unit/server/services/members/members-api/controllers/RouterController.test.js
```diff
@@ -1097,4 +1097,276 @@ describe('RouterController', function () {
...
+                    // reset response spies between attempts
+                    res.writeHead.resetHistory();
+                    res.end.resetHistory();
+
+                    await routerController.verifyOTC(req, res);
+
+                    sinon.assert.calledWith(res.writeHead, 400, {'Content-Type': 'application/json'});
+                    sinon.assert.calledWith(res.end, JSON.stringify({
+                        valid: false,
+                        message: OTC_TEST_CONSTANTS.ERROR_MESSAGES.INVALID_CODE
```
troyciesco: just noting pretty sure this will change when we update to not use res.write, but it's passing for now

ghost/core/core/server/services/members/members-api/controllers/RouterController.js
```diff
@@ -649,6 +652,70 @@ module.exports = class RouterController {
...
+        // TODO: obtain and pass through referrer here
+        const redirectUrl = this._magicLinkService.getSigninURL(tokenValue, 'signin', null, otcVerificationHash);
+        if (!redirectUrl) {
+            throw new errors.BadRequestError({
+                message: tpl(messages.failedToVerifyCode),
+                code: 'OTC_VERIFICATION_FAILED'
+            });
+        }
+
+        return res.json({redirectUrl});
```
troyciesco, line 716: this included `valid: true, success: true, message: 'OTC verification successful'` before, which i think are all unnecessary? at least it didn't seem like any of that information was used as part of building the magic link. obvs could easily add back if i'm wrong though

ghost/core/core/server/services/members/members-api/controllers/RouterController.js
```diff
@@ -25,9 +25,31 @@ const messages = {
...
+    if (redirect) {
+        try {
+            return new URL(redirect).href;
+        } catch (e) {
+            logging.warn(e);
+        }
+    }
+
+    return req.get('referer') || null;
+}
```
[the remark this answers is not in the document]
kevinansfield: None of those suggested conditions apply or are valid, the function is working as intended

ghost/core/core/server/web/members/app.js
```diff
@@ -81,10 +80,20 @@ module.exports = function setupMembersApp() {
...
     );
+    membersApp.post(
+        '/api/verify-otc',
+        bodyParser.json(),
+        middleware.verifyIntegrityToken,
+        // NOTE: this is wrapped in a function to ensure we always go via the getter
+        function lazyVerifyOTCMw(req, res, next) {
+            return membersService.api.middleware.verifyOTC(req, res, next);
+        }
+    );
```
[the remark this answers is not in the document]
kevinansfield: This will be added in a later PR

kevinansfield (approved):

PR #24780, opened 2025-08-28
Added actions to submit OTC from portal and sign in

troyciesco:
closes https://linear.app/ghost/issue/PROD-2501

- adds the `verifyOTC` action to the portal
- `verifyOTC` sends the otc and the otcRef to the api, and if successful, the api responds with a magic link
- portal redirects to the link immediately, signing the user in


ghost/core/core/server/services/members/members-api/controllers/RouterController.js
```diff
@@ -689,7 +689,8 @@ module.exports = class RouterController {
 
             const otcVerificationHash = await this._createHashFromOTCAndToken(otc, tokenValue);
             // TODO: obtain and pass through referrer here
-            const redirectUrl = this._magicLinkService.getSigninURL(tokenValue, 'signin', null, otcVerificationHash);
+            let referrer = req.get('referer');
```
troyciesco: Left the TODO here and copied what `sendMagicLink` does just to see it working on the frontend first (it did), will cleanup before merge

apps/portal/src/actions.js
```diff
@@ -116,6 +116,34 @@ async function signin({data, api, state}) {
...
 }
 
+async function verifyOTC({data, api, state}) {
+    const {t} = state;
+
+    try {
+        const response = await api.member.verifyOTC(data);
+
+        if (response.redirectUrl) {
+            return window.location.assign(response.redirectUrl);
```
troyciesco, line 126: switched to this, similar to data-attributes.js, so mocking was doable in the tests

kevinansfield (approved):

PR #24805, opened 2025-09-03
Added automatic environment setup for e2e tests

cmraible:
Currently we need to manually setup a Ghost instance to run the e2e test suite against, reset the database, etc.

This adds the required global setup to create multiple isolated, disposable, parallel instances of Ghost using testcontainers.

The core challenge in using testcontainers here is that the global setup doesn't run in the same process as the playwright workers, which means we lose the reference to the testcontainers we create between the global setup and the tests themselves, and the testcontainers library doesn't support i.e. `new GenericContainer().fromContainerId()`. This is why I had to drop down to use `dockerode` (a node wrapper around the Docker API, which is what testcontainers uses under the hood) for any Docker functionality after the global setup. 

- [x] Get all e2e tests passing locally
- [x] Fix container cleanup
- [x] Setup docker build in CI


e2e/playwright.config.mjs
```diff
@@ -2,45 +2,50 @@ import dotenv from 'dotenv';
...
 /** @type {import('@playwright/test').PlaywrightTestConfig} */
 const config = {
     timeout: process.env.CI ? 60 * 1000 : 30 * 1000,
     expect: {
         timeout: process.env.CI ? 30 * 1000 : 10 * 1000
     },
-    retries: 1, // Retries open the door to flaky tests. If the test needs retries, it's not a good test or the app is broken.
-    workers: 1, // One worker for now in the interest of stability. Parallelism leads to flaky tests when not done carefully.
+    retries: 0, // Retries open the door to flaky tests. If the test needs retries, it's not a good test or the app is broken.
+    workers: process.env.CI ? 2 : 10, // Enable parallelism with 4 workers locally, 2 in CI
```
9larsons: Comment doesn't match - using 10 locally

e2e/helpers/utils/setup-user.ts
```diff
@@ -1,7 +1,8 @@
 import {appConfig} from './app-config';
 import {UserFactory, User} from '../../data-factory/factories/user-factory';
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const logging = require('@tryghost/logging');
+import debug from 'debug';
+
+const log = debug('e2e:setup-user');
```
9larsons: Can you explain the move from tryghost/logging to debug?

e2e/helpers/fixtures/ghost-instance.ts
```diff
@@ -0,0 +1,124 @@
...
+        const loginPage = new LoginPage(page);
+        await loginPage.waitForLoginPageAfterUserCreated();
+        await loginPage.signIn(appConfig.auth.email, appConfig.auth.password);
+        
+        // Wait for successful login and navigate to admin to establish proper origin
+        const analyticsPage = new AnalyticsOverviewPage(page);
+        await analyticsPage.header.waitFor({ state: 'visible' });
+        
+        // Ensure we're on the Ghost admin page to set correct origin for future requests
+        await page.goto('/ghost/#/dashboard');
```
9larsons: This is confusing. We're going to analytics but then going to dashboard? Which should be removed, although I believe it redirects back.

e2e/playwright.config.mjs
cmraible: Ah yeah good catch. I'll just remove the comment as it's pretty redundant

e2e/helpers/fixtures/ghost-instance.ts
cmraible: Hm yeah that is confusing, I'll take a closer look at this bit

e2e/helpers/utils/setup-user.ts
cmraible: Yeah good call out — I'll probably revert this change here. Ultimately I think we'll want to have both - @tryghost/logging for info/error level logs, and use debug for more detailed debug output.

cmraible:
Closing in favor of: https://github.com/TryGhost/Ghost/pull/24862

The new approach uses Docker Compose to drastically simplify the implementation

PR #24810, opened 2025-09-03
Improved docker build caching for admin & microfrontends

cmraible:
With the current Dockerfile, we aren't able to cache any of the admin microfrontend builds, so we rebuild _all_ of them whenever _any_ file in the repo has changed. This PR uses multi-stage builds to improve our caching, so we only rebuild packages when their own code has changed.

Each package gets its own "builder" stage, which copies only the files it needs to build itself. Then in the final development image, we copy all of the code from the whole repo, then copy the build outputs from each builder stage into their proper locations.

In a best case scenario where you're only changing code in `ghost/core` and not touching any of the microfrontends, this reduces the warm build time by upwards of 90 seconds.

The main benefit here is in CI: now if you only change code in `ghost/core` without changing dependencies, all of the package builds should be fully cached, saving up to 3 minutes on the build time. This is less impactful for local development, since the code is mounted into the container and we rebuild on file changes _within_ the container.

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -81,9 +81,113 @@ COPY patches patches
 RUN yarn install --frozen-lockfile --prefer-offline
 
 # --------------------
+# Shade Builder
+# --------------------
+FROM development-base AS shade-builder
+WORKDIR /home/ghost
+COPY apps/shade apps/shade
+RUN cd apps/shade && yarn build
+
+# --------------------
+# Admin-x-design-system Builder
+# --------------------
+FROM development-base AS admin-x-design-system-builder
+WORKDIR /home/ghost
+COPY apps/admin-x-design-system apps/admin-x-design-system
+RUN cd apps/admin-x-design-system && yarn build
+
+# --------------------
+# Admin-x-framework Builder
+# --------------------
+FROM development-base AS admin-x-framework-builder
+WORKDIR /home/ghost
+COPY apps/admin-x-framework apps/admin-x-framework
+COPY --from=shade-builder /home/ghost/apps/shade/es apps/shade/es
+COPY --from=shade-builder /home/ghost/apps/shade/types apps/shade/types
+COPY --from=admin-x-design-system-builder /home/ghost/apps/admin-x-design-system/es apps/admin-x-design-system/es
+COPY --from=admin-x-design-system-builder /home/ghost/apps/admin-x-design-system/types apps/admin-x-design-system/types
+RUN cd apps/admin-x-framework && yarn build
+
+# --------------------
+# Stats Builder
+# --------------------
+FROM development-base AS stats-builder
+WORKDIR /home/ghost
+COPY apps/stats apps/stats
+COPY --from=shade-builder /home/ghost/apps/shade apps/shade
+COPY --from=admin-x-design-system-builder /home/ghost/apps/admin-x-design-system/es apps/admin-x-design-system/es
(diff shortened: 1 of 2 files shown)


.docker/Dockerfile
```diff
@@ -80,10 +80,114 @@ COPY patches patches
 
 RUN yarn install --frozen-lockfile --prefer-offline
 
+# --------------------
```
cmraible, line 83: Note: the order of these builder stages matters, since e.g. the microfrontends depend on the outputs from the shade and admin-x-framework packages.

.dockerignore
```diff
@@ -27,3 +27,5 @@ compose.yml
 .docker
 !.docker/**/*.entrypoint.sh
 !.docker/**/*entrypoint.sh
+
+ghost/core/core/built/admin
```
cmraible, line 31: Prevents any locally built admin assets from being copied into the container in the `COPY . . ` step — else we get errors when trying to copy these files in on line 191 since the files already exist.

.docker/Dockerfile
```diff
@@ -80,10 +80,114 @@ COPY patches patches
...
+# Admin-x-framework Builder
+# --------------------
+FROM development-base AS admin-x-framework-builder
+WORKDIR /home/ghost
+COPY apps/admin-x-framework apps/admin-x-framework
+COPY --from=shade-builder /home/ghost/apps/shade/es apps/shade/es
+COPY --from=shade-builder /home/ghost/apps/shade/types apps/shade/types
+COPY --from=admin-x-design-system-builder /home/ghost/apps/admin-x-design-system/es apps/admin-x-design-system/es
+COPY --from=admin-x-design-system-builder /home/ghost/apps/admin-x-design-system/types apps/admin-x-design-system/types
```
9larsons, line 108: This struck me as confusing but I'll follow up with Jonatan and Zimo in our Ember->React project. I don't think our framework should be depending on the design system(s). It's a fairly limited requirement.

.docker/Dockerfile
```diff
@@ -80,10 +80,114 @@ COPY patches patches
...
+RUN cd apps/admin-x-framework && yarn build
+
+# --------------------
+# Stats Builder
+# --------------------
+FROM development-base AS stats-builder
+WORKDIR /home/ghost
+COPY apps/stats apps/stats
+COPY --from=shade-builder /home/ghost/apps/shade apps/shade
```
9larsons, line 117: Why are we copying in the whole app here (posts + stats + AP builders) instead of the `/shade/es` and `/shade/types` dir like the framework builder does?

9larsons (approved):
No need to request changes, just curious about a couple items that might lead to a very minor performance improvement. This should be a nice improvement, nice work!

.docker/Dockerfile
cmraible: Ah good spot! All of the apps that use `@tryghost/shade` depend on a few src code files directly:
- [postcss.config.cjs](https://github.com/TryGhost/Ghost/blob/improve-docker-build-caching/apps/stats/postcss.config.cjs) 
- [styles.css](https://github.com/TryGhost/Ghost/blob/improve-docker-build-caching/apps/stats/src/styles/index.css)

and a few other files, so it seemed cleaner to just copy the whole directory than to add another `COPY` step for each file

.docker/Dockerfile
cmraible: Yeah I agree, this is definitely a smell - if anything I'd expect the dependency to be flipped, but ideally the framework and design system would be fully decoupled

PR #24812, opened 2025-09-03
Added utmTracking feature flag

9larsons:
ref https://linear.app/ghost/issue/PROD-2553/
- added `utmTracking` flag for utm-related beta features

We are adding this flag to put UI changes behind.


cmraible (changes requested):
All looks good, except I think we want this displayed under "Private features" in Settings > Labs so we can easily toggle it on/off in development. This whole panel is hidden in production, so live sites won't be able to turn it on/off themselves.

9larsons:
> All looks good, except I think we want this displayed under "Private features" in Settings > Labs so we can easily toggle it on/off in development. This whole panel is hidden in production, so live sites won't be able to turn it on/off themselves.

I'm confused. Isn't it?

CleanShot 2025-09-03 at 14 50 50@2x

cmraible (approved):
Not sure what I was looking at earlier, but you're totally right — my bad! Nice one 👍

PR #24813, opened 2025-09-03
Added new UTM data pipes

9larsons:
ref https://linear.app/ghost/issue/PROD-2562/
- added new pipes for utm data: api_top_utm_terms, api_top_utm_campaigns, api_top_utm_sources, api_top_utm_mediums
- new endpoints return fixed data based on the source values in the underlying data

It was proving both difficult and burdensome to update the api_top_sources pipe to parameterize the return value. Because of our need to group the data based on the column, we can't return it all in one query.

The new endpoints are basically clones of `api_top_sources` except that we're creating mock data based on the underlying source value for the moment. This is because we haven't wired up the data to start ingesting the `utm_{x}` data fields, and we'll come through later to do that.


cmraible (commented):
I'm not sure I understand the shape of this response.

As I understand it, these new UTM fields will be new fields that we'll aggregate on (i.e. they'll be in the group by instead of top sources), and they won't necessarily be a 1:1 relationship with the source.

Example: `source` might be "google", `medium` would be "search", but we might have 12 different values for `term`. Are you proposing that we would return 12 different rows for `source = google`, and aggregate that on the client depending on the current filters?

9larsons:
We discussed - after trying to parameterize the source/'return' column, it proved easiest to simply create new endpoints. I reverted the previous changes and added four new endpoints that return fixed data in the form of `utm_source,visits`, just like we do `source, visits` for `api_top_sources`.

cmraible (changes requested):
LGTM overall, nice work! Two small things:
- I think we'll want an endpoint for `utm_contents` following this same pattern 

Image

- I'm not necessarily against keeping the .md docs in the codebase, but if we do I think they should be nested under `..../core/server/data/tinybird/docs/` or something along those lines (or just removed, either way)

cmraible (approved):
Nice work! This will be really helpful I think 👌

PR #24816, opened 2025-09-04
Add E2E tests for new tag list

jonatansberg:
These PR adds some basic tests for the new tags screen to the new E2E playwright setup. 

There are still some mocking in here since the test are not yet isolated and run against the development env locally. I'd love to get rid of that once we have the testing env sorted!


jonatansberg:
Seems like the feature flag isnt enabled when you run the tests in CI... Thats why thet are failing.

e2e/helpers/pages/admin/tags/TagsPage.ts
```diff
@@ -0,0 +1,79 @@
...
+                            limit: 100,
+                            pages: 1,
+                            total: tags.meta?.pagination?.total ?? tags.tags?.length ?? 0,
+                            ...tags.meta?.pagination
+                        }
+                    },
+                    tags: tags.tags ?? []
+                })
+            });
+        });
```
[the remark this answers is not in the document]
jonatansberg: Works as is

e2e/helpers/pages/admin/tags/TagsPage.ts
```diff
@@ -0,0 +1,79 @@
...
+
+// XXX: Remove these types and the mockTagsResponse method once we have proper
+// test isolation and no longer need to mock responses.
+interface Tag {
+    id: string;
+    name: string;
+    slug: string;
+    url: string;
+    description: string;
+}
```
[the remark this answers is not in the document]
jonatansberg: Wont do. I dont want to add a dependency on the admin-x-framework at this point

e2e/helpers/pages/admin/tags/TagsPage.ts
```diff
@@ -0,0 +1,78 @@
...
+    readonly newTagButton: Locator;
+
+    constructor(page: Page) {
+        super(page);
+
+        this.pageUrl = '/ghost/#/tags';
+        this.pageContent = page.getByTestId('tags-page');
+        this.tagList = page.getByTestId('tags-list');
+
+        // XXX: This is the only way I have found to reliably selecting the table rows
```
9larsons: Can't we just add a `data-testid="tag-row"` to the `TableRow` component?

e2e/helpers/pages/admin/tags/TagsPage.ts
```diff
@@ -0,0 +1,78 @@
...
+        this.activeTab = this.tabs.locator('[data-state="active"]');
+        this.newTagButton = page.getByRole('link', {name: 'New tag'});
+    }
+
+    async selectTab(tabText: string) {
+        const tab = this.tabs.getByRole('link', {name: tabText});
+        await tab.click();
+    }
+
+    // XXX: Remove once we have proper test isolation and don't need mocking
```
9larsons: I think if you rebase you can take another shot at this - the test isolation changes just got merged.

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        }));
+
+        await tagsPage.goto();
+
+        await expect(
+            tagsPage.pageContent.getByRole('heading', {
+                name: 'Start organizing your content'
+            })
+        ).toBeVisible();
+        await expect(
```
9larsons: We should make this a locator in the page object.

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        ).toBeVisible();
+
+        await expect(tagsPage.tagList).not.toBeVisible();
+    });
+
+    test('lists public and internal tags separately', async ({page}) => {
+        const tagsPage = new TagsPage(page);
+
+        await tagsPage.mockTagsResponse(async request => ({
+            tags: request.url().includes('internal')
```
9larsons, line 28: We should build a tag factory. This is a good solution for the moment, and we can come back through and do it with the new patterns 👍.

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,227 @@
...
+                    }
+                ]
+        }));
+
+        await tagsPage.goto();
+
+        // Default to public tags
+        await expect(tagsPage.activeTab).toHaveText('Public tags');
+
+        await expect(tagsPage.tagList).toBeVisible();
```
9larsons: I'd defer to Igor as he has strong opinions on assertions and what they should look like. I think we can skip visibility checks if we're also checking content below to reduce verbosity.

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        await expect(tagsPage.tagListRow).toHaveCount(3);
+    });
+
+    test('create new tag', async ({page}) => {
+        const tagsPage = new TagsPage(page);
+        await tagsPage.goto();
+
+        await tagsPage.newTagButton.click();
+
+        await expect(page.url()).toContain('/ghost/#/tags/new');
```
9larsons: We want a new page object for this and we can assert the url with that.

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,227 @@
...
+                },
+                tags: Array.from({length: pageParam < pages ? pageSize : pageSize / 2}, (_, i) => ({
+                    id: `${i + offset}`,
+                    name: `Tag ${i + offset}`,
+                    slug: `tag-${i + offset}`,
+                    url: `https://example.com/tag-${i + offset}`,
+                    description: `Tag ${i + offset} description`
+                }))
+            };
+        });
```
9larsons, line 163: Ideally we can cut out all of this with a PageFactory builder method.

9larsons (changes requested):
Sorry for the delay on reviewing this. All in all, I'm stoked that these work! We're very close to being able to hit the ground running with this package.

I don't think these tests are critical to merge as it's more of a spike. Given that, I'd like to:
1) Rebase to pull in test isolation changes
2) Lean more on locators and the Page object(s)

I feel less strongly about the need for these to happen before merging, but I do feel strongly that they should happen:
- (Create and) Utilize a TagFactory
- Talk w/ Igor about assertions; I suspect he'll want to revise here, but I think this is a great basis for discussion on approach for us all to align on

e2e/helpers/pages/admin/tags/TagsPage.ts
jonatansberg: I had that at one point, but I ran in to some kind of issue... Coming back to this, I can't see why that wouldn't work tho 🤔

e2e/helpers/pages/admin/tags/TagsPage.ts
jonatansberg: I think it probably makes more sense to do this alongside introducing a factory, no?

e2e/tests/admin/tags/list.test.ts
jonatansberg: That seems reasonable. I'll drop that assertion.

e2e/tests/admin/tags/list.test.ts
jonatansberg: I'm having a second look at this, and I'm leaning towards that maybe they shouldn't be in the page object? 

Using page objects makes a lot of sense for things that we want to reuse across different tests. In this case we're only referencing these elements in this specific test case, meaning moving them to the page object would just add indirection.

You could argue that the increased consistency of having all selectors in page objects trumps the indirection in this case. I'm not sure I agree with that tho. While I'd be the first to agree that there is value in consistency, from my POV this is still primarily an assertion tho that just happened to leverage a selector 🤷  

Curious to hear your take on this @ibalosh!

apps/posts/src/views/Tags/components/TagsHeader.tsx
```diff
@@ -13,7 +13,7 @@ const TagsHeader: React.FC<TagsHeaderProps> = ({currentTab}) => {
             <Header.Title>Tags</Header.Title>
 
             <Header.Nav>
-                <PageMenu defaultValue={currentTab}>
+                <PageMenu data-testid="tags-header-tabs" defaultValue={currentTab}>
```
ibalosh, line 16: I am not fully sold here on passing data-testid down to a component as property. If I read the codebase correctly, we pick it up by spread operator (...props) and it could go to multiple places later, so this can easily break and point to incorrect pieces of code as we do changes. Or is there a restriction of not adding these in shade components?

I think it's better to put it on exact piece of code where we plan to use it, otherwise it will become hard to see where the issue is.

e2e/tests/admin/tags/list.test.ts
ibalosh: There are multiple reasons why I think it is better to have them in POM. 

**1. Better separation of concerns and cheap to write**

We already have tagsPage page object here, which we are using, so it's cheap and easy to hide implementation details like specific locators. Test do not care about locators - they care mostly about behaviour, which is better described by not seeing locators at all in test itself. 

It's a nice separation of concerns - you POM fellas handle the locators, actions, descriptions of the page details and I as a test will handle describing behaviour with help of POMs, asserions, utils etc. Also test suites will grow, and I don't really want to update hard coded locators (maybe I will need to wait for visibility of one of these later, so I will actually reuse it, or something else?)

**2. Easier to read** 

Since we anyways have everything in page objects, call itself will be shorter, more descriptive, and therefore easier to read.

```typescript
const tagsPage = new TagsPage(page);
await tagsPage.mockTagsResponse(async () => ({tags: []}));

await tagsPage.goto();

await expect(tagsPage.pageHeading)).toBeVisible();
await expect(tagsPage.createNewTagLink)).toBeVisible();
await expect(tagsPage.tagList).not.toBeVisible();
```

If we did not use POM at all in this test, clutters POM and it's expensive to write, yeah, maybe. But POMs are dirt cheap POJOS, no brainers to write, hide page specific details and makes AI more consistent in it's code generating.

I could argue also that they make migration to other framework easier, but that would be a stretch here :) playwright is here to stay and we do use bit more of it's insides in tests themselves (though I did do this form of migration once)

e2e/tests/admin/tags/list.test.ts
ibalosh: Agreed. It also introduces new concept into codebase, which bypasses `mockApi` that we used before.
I do like this approach since it's right there when you are reading the tests, but it is something we should catchup on and sync on it.

(that will also help figuring out where do we want to go with mocking if we would continue to use it in playwright)

e2e/tests/admin/tags/list.test.ts
ibalosh: agreed

e2e/helpers/pages/admin/tags/TagsPage.ts
```diff
@@ -0,0 +1,78 @@
+import {Locator, Page, Request} from '@playwright/test';
+import {AdminPage} from '../AdminPage';
+
+// XXX: Remove these types and the mockTagsResponse method once we have proper
```
ibalosh: what is the convention for the todo comments like the XXX here, for things that need to be checked later. From what I can see on other places, `// TODO` was used? 

Just want to make sure we use similar concepts, to find these

e2e/helpers/pages/admin/tags/TagsPage.ts
```diff
@@ -0,0 +1,78 @@
...
+        this.newTagButton = page.getByRole('link', {name: 'New tag'});
+    }
+
+    async selectTab(tabText: string) {
+        const tab = this.tabs.getByRole('link', {name: tabText});
+        await tab.click();
+    }
+
+    // XXX: Remove once we have proper test isolation and don't need mocking
+    async mockTagsResponse(handler: (request: Request) => Promise<Partial<PaginatedResponse>>) {
```
ibalosh: Hm, not a fan of this one since it breaks single responsibility rule, even though it's temporary. Could we extract it, move to separate helpers/mocks folder or something similar.

That way at least all related data to it is there, once we decide to ditch it. It keeps everything clean, and less chance to trip up AI when it writes new page objects.

e2e/helpers/pages/admin/tags/TagsPage.ts
```diff
@@ -0,0 +1,78 @@
...
+    constructor(page: Page) {
+        super(page);
+
+        this.pageUrl = '/ghost/#/tags';
+        this.pageContent = page.getByTestId('tags-page');
+        this.tagList = page.getByTestId('tags-list');
+
+        // XXX: This is the only way I have found to reliably selecting the table rows 
+        // without also targeting the header or any of the placeholder or spacer rows.
+        this.tagListRow = this.tagList.locator('tr:not([aria-hidden]):not(:first-child)');
```
ibalosh: `nit`: ouch, could you share details why this one needs to be tricky? (just curious) . I am wondering if multiple aria attributes or other attributes on tr elements would help to locate it better

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,227 @@
...
+test.describe('Ghost Admin - Tags', () => {
+    // XXX: Amend the settings response to enable tagsX feature flag for all tests
+    // until we have a proper way to enable and disable feature flags in tests
+    test.beforeEach(async ({page}) => {
+        await page.route('/ghost/api/admin/settings/*', async (route) => {
+            const response = await route.fetch();
+            const json = await response.json();
+            
+            // Override the labs field to enable tagsX feature flag
+            if (json.settings) {
```
ibalosh: could you share why there is a conditional here and second one after it inside the condition? 
It makes test harder to read and understand, with conditional logic before each test case. It should be avoided if avoidable. If not, it should be better described, since these also act partially as documentation.

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,227 @@
...
+                    slug: 'tag-3',
+                    url: 'https://example.com/tag-3',
+                    description: 'Tag 3 description'
+                }
+            ]
+        }));
+
+        await tagsPage.goto();
+
+        await expect(tagsPage.tagList).toBeVisible();
```
ibalosh: `nit`: we don't need this one, since it will fail anyways after, when doing next asserts

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,227 @@
...
+
+        await expect(tagsPage.tagListRow.getByRole('link', {name: 'Tag 1'})).toBeVisible();
```
ibalosh: move to POMs -> methods

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        await tagsPage.tagListRow.last().scrollIntoViewIfNeeded();
+
+        // Wait for third page to load
+        await expect(tagsPage.tagListRow.getByRole('link', {name: 'Tag 41', exact: true})).toBeVisible();
+        
+        await tagsPage.tagListRow.last().scrollIntoViewIfNeeded();
+        await tagsPage.tagListRow.last().scrollIntoViewIfNeeded();
+        
+        // Verify last tag is visible
+        await expect(tagsPage.tagListRow.getByRole('link', {name: 'Tag 50', exact: true})).toBeVisible();
```
ibalosh: move these to POMs, to not mix responsibilities

ibalosh (changes requested):
Hey fellas,

I went through the PR to check bit more things. I think it needs cleanup, to get to our goal of nice tests that we and AI will like too, when it starts generating new ones.

Shared some of my thoughts in there.

e2e/tests/admin/tags/list.test.ts
ibalosh: Also I am bit confused, on one side we mock in page object, while here we mock outside of page object. I think we need to keep the logic consistent, since it's easy to diverge too far.

To keep it simple I move move mocks to single place, mentioned in one of comments below. It will also keep everything easier to read, including comments.

apps/posts/src/views/Tags/components/TagsHeader.tsx
jonatansberg: Where would you suggest the testid goes instead? 

If it should go on the menu container, but not all menu containers everywhere, this would be the way to do it. The PageMenu component is typed as a React element wrapping a `div` so I don't think it's too confusing. Either way, I'm perfectly happy to delegate the responsibility of ensuring the testid ends up in the right place to the component. 

We could of course have a more generic testid on the PageMenu that would always be there. Ideally I'd rather not have the testids at all tho and instead leverage role and text selectors 🤔

e2e/tests/admin/tags/list.test.ts
jonatansberg: For me, this has nothing to do with performance. It's a matter of co-location versus reuse. 

In my past experience, blindly shoving everything into POMs creates a lot of indirection. This absolutely makes the test read nicer when you look at them, but when they break, you need to go look somewhere else to figure out why. This makes it more annoying to reason about the tests when they eventually fail. 

When the locator is plainly visible in the assertion, if it would ever fail, you can see that straight from the log. With POMs, you then need to go and pull up the failing test in the trace viewer or run the tests locally. If you do, then you need to wait for them to fail. Then maybe you can figure it out by looking in the more verbose trace viewer logs, or you need to go command clicking through one or possibly several steps of indirection in your editor, at which point you no longer have the context of the failing tests. 

For me, that extra hassle makes a lot of sense if you want to reuse your locators. I'm not sure about the trade-offs otherwise. 

Either way, this is not a hill I want to die on, so if you feel strongly about this, I'm happy to move it :)

e2e/helpers/pages/admin/tags/TagsPage.ts
jonatansberg: For me, there is a small semantic difference between `XXX` and `TODO`. I read `XXX` as "this is smelly", i.e. a callout of tech debt that should be addressed, whereas a TODO is more of a note to future readers that there the author intended to do more but was unable to for whatever reason. 

We can turn these in to TODOs if you think that makes more sense...

e2e/helpers/pages/admin/tags/TagsPage.ts
jonatansberg: I guess we can turn it into a separate helper since it's not strictly a mock... 🤔

e2e/tests/admin/tags/list.test.ts
jonatansberg: The conditionals guard against API changes and/or accidentally targeting other requests than the intended settings request. I can change to use optional chaining to flatten it our a bit.

The feature flag request override, which isn't strictly a mock either, was added after the other stuff was done as a way to forcefully enable the feature flag in CI. That's why they're in different places. 

I think we should find a completely different way to manage feature flags longer term. Either we always enable all feature flags (this is something we do elsewhere, and we discussed doing the same here while you were away) or we add something to the fixtures that lets us manipulate the feature flags in an appropriate lifecycle hook or something like that.

e2e/helpers/pages/admin/tags/TagsPage.ts
ibalosh: Makes sense. Both work, but since it’s a new term in the repo, I’d raise it on the eng call for consensus.

e2e/helpers/pages/admin/tags/TagsPage.ts
ibalosh: agreed, separate helper will do

(will bring this up on next catchup we have regarding the tests that need mocks/stubs, since this one clearly shows old [mockApi](https://github.com/TryGhost/Ghost/blob/5df8a3c6f64f021e00a90711b61386fc9f16c13f/apps/admin-x-framework/src/test/acceptance.ts#L211) call is limited, lacks flexibility

e2e/tests/admin/tags/list.test.ts
ibalosh: Interesting, for now though, let's move them to POM.

e2e/helpers/pages/admin/tags/TagsPage.ts
jonatansberg: I rephrased as todos for now

jonatansberg:
@ibalosh @9larsons I've addressed your feedback. I stopped short of introducing new POMs and a tags factory as that seemed to snowball the scope of this PR even further. I can pick those things up as separate followup PRs.

I'm also not sure about 0b056ca96d4f3e85fedd5df8bffb043d14868bc0. When pulling that out into a separate helper it felt strange to keep it next to the tests. At the same time, the other helper clearly does not belong in the shared utils. To top it off I also kinda want to throw both of them away as soon as we have a better solutions in place.

apps/posts/src/views/Tags/components/TagsHeader.tsx
ibalosh: Directly inside components, like [here](https://github.com/TryGhost/Ghost/blob/6c60ba4476d713f6850d6aa499bbea8f49b31884/apps/shade/src/components/ui/pagemenu.tsx#L123) ?

I do get it that you would need to add it then to all of them, by one of methods you just suggested, but at least it's deterministic. Not saying current one is bad, just worried this will get hard to debug if `...props` gets used more within components.

I completely agree that testId is a workaround here, and role and selectors would be way better, if there is a choice to use them. Maybe divs we already have with `role` and `aria` if other element is not applicable?

jonatansberg: I ended up adding the testid as there was no good way to target the menu without resorting to CSS selectors. 

I'm not too keen on editing the component directly in Shade. Since Shade is in turn just leveraging shadcn, any changes that we introduce makes it harder to reconcile what we have with the upstream version in case we want to (or need to) update the component in the future.

ibalosh:
> @ibalosh @9larsons I've addressed your feedback. I stopped short of introducing new POMs and a tags factory as that seemed to snowball the scope of this PR even further. I can pick those things up as separate followup PRs.
> 
> I'm also not sure about [0b056ca](https://github.com/TryGhost/Ghost/commit/0b056ca96d4f3e85fedd5df8bffb043d14868bc0). When pulling that out into a separate helper it felt strange to keep it next to the tests. At the same time, the other helper clearly does not belong in the shared utils. To top it off I also kinda want to throw both of them away as soon as we have a better solutions in place.

yeah, agreed, I am not sure either yet. We can keep it, I don't have strong feelings about it and might be a premature optimization to solve now. Will think about it as we add more tests.

e2e/tests/admin/tags/list.test.ts
ibalosh: also, I don't think we need this test at all, it's just an implementation detail. If you click the button, create a new tag, you already tested the url, which is not important fot the test.

ibalosh:
```suggestion
expect(page.url()).toContain('/ghost/#/tags/new');
```

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,193 @@
...
+                    description: `Tag ${i + offset} description`
+                }))
+            };
+        });
+
+        await tagsPage.goto();
+
+        // Verify first page loads
+        await expect(tagsPage.getRowByTitle('Tag 1')).toBeVisible();
+        await expect(await tagsPage.tagListRow.count()).toBeGreaterThan(10);
```
ibalosh:
```suggestion
        expect(await tagsPage.tagListRow.count()).toBeGreaterThan(10);
```

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,193 @@
...
+        await expect(await tagsPage.tagListRow.count()).toBeLessThan(40);
```
ibalosh:
```suggestion
        expect(await tagsPage.tagListRow.count()).toBeLessThan(40);
```

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+        // Scroll again to trigger loading of third page
+        await tagsPage.tagListRow.last().scrollIntoViewIfNeeded();
+        await tagsPage.tagListRow.last().scrollIntoViewIfNeeded();
+
+        // Wait for third page to load
+        await expect(tagsPage.getRowByTitle('Tag 41')).toBeVisible();
+        
+        await tagsPage.tagListRow.last().scrollIntoViewIfNeeded();
+        await tagsPage.tagListRow.last().scrollIntoViewIfNeeded();
```
ibalosh, line 191: `nit:` is the second scroll needed, and if so, this action could be within POM

e2e/helpers/pages/admin/tags/TagsPage.ts
```diff
@@ -0,0 +1,44 @@
...
+        this.pageContent = page.getByTestId('tags-page');
+        this.tagList = page.getByTestId('tags-list');
+        this.tagListRow = this.tagList.getByTestId('tag-list-row');
+
+        this.tabs = page.getByTestId('tags-header-tabs');
+        this.activeTab = this.tabs.locator('[data-state="active"]');
+        this.newTagButton = page.getByRole('link', {name: 'New tag'});
+
+        this.emptyStateTitle = this.pageContent.getByRole('heading', {name: 'Start organizing your content'});
+        this.emptyStateAction = this.pageContent.getByRole('link', {name: 'Create a new tag'});
```
ibalosh, line 31: since we use button in other places we should here too.

```suggestion
this.createNewTagButton = this.pageContent.getByRole('link', {name: 'Create a new tag'});
```

ibalosh (approved):
I added couple of comments here and there, but overall I think it's good. We can cleanup further as we add more and introduce tags factory.

One thing though, before you merge it, make sure it passes - I did not get the tests to pass on my machine.
Ready to approve, but did not yet, just to make sure it passes first, once it does, ready to approve @jonatansberg

e2e/helpers/pages/admin/tags/TagsPage.ts
jonatansberg: We're already using the `newTagButton` in the test that I think you are referring to? This is a different button that only shows in the empty state. We could remove it, though I don't think it adds much value.

e2e/tests/admin/tags/list.test.ts
jonatansberg: Yeah, the second scroll is needed since we're rendering items using a virtual list. So as we scroll the last item into view, more items will render below it. So we need to scroll twice to get to the bottom. 

The plan is to pull the virtualization out into the Shade eventually, and I think we can iterate on the tests when we do that. Especially since this doesn't really belong in the tag list page object anyways.

e2e/tests/admin/tags/list.test.ts
jonatansberg: Having some kind of assertion that the URL changes is actually relevant in this case, since we're navigating across the React-Ember border at this point. Thats why we want to make sure that we're actually navigated away from the list page.

jonatansberg: I fixed the assertion for now

e2e/helpers/pages/admin/tags/TagsPage.ts
ibalosh: no, `newTagButton` is the button with label `New Tag` on it. I was thinking of this other one, it's still a button, so I think we should be name it that way (I admit a used weird name). 

We could add a section in page objects, and then have something new `tagsPage.emtptyList.newTagButton.click();` or something similar, but thought thats an overkill this eary. I just went for the easiest change to get to consistence in locators type namings.

e2e/tests/admin/tags/list.test.ts
ibalosh: I see, got it, but wouldn't we actually test anyways that scenario by click the button, creating a tag later on with that type of test?

Is it as temp solution for react-ember border?

jonatansberg: Yeah, exactly. We already have tests for the Ember screen in the old setup. Once we've migrated the edit screen, we'll add the appropriate tests here.

e2e/helpers/pages/admin/tags/TagsPage.ts
jonatansberg: Ah, okay, I see. 

Even though perhaps it's more consistent I also find it a bit confusing. I wouldn't know how to tell them apart without looking at both the locator definition and the UI. The `emptyStateAction` name more clearly conveys the semantic meaning of that button to me and it also avoids that conflict. 

I guess you could call it the `emptyStateActionButton`, but that just gets wordy 😅

jonatansberg:
@9larsons Are you happy with the changes? I would need your approval in order to merge this since you requested changes :)

e2e/tests/admin/tags/list.test.ts
ibalosh: makes sense, tx!

9larsons:
Sorry for the delay. Igor is working on a tags factory (https://github.com/TryGhost/Ghost/pull/25019#pullrequestreview-3289044523) and I'll work on getting a Settings factory/endpoint for toggling labs and other settings, and then we can come back and rework this to ditch the mocks.

e2e/tests/admin/tags/list.test.ts
```diff
@@ -0,0 +1,196 @@
...
+import {TagsPage} from '../../../helpers/pages/admin';
+import {overrideFeatureFlags} from '../../../helpers/utils';
+import {mockTagsResponse} from './helpers/mock-tags-response';
+
+test.describe('Ghost Admin - Tags', () => {
+    test.beforeEach(async ({page}) => {
+        await overrideFeatureFlags(page, {tagsX: true});
+    });
+
+    test('shows empty state when no tags exist', async ({page}) => {
```
9larsons, line 11: This part is goofy - we seed the db with a tag by default. I haven't decided if we should clear the db of 'default entries'. That feels likely, as we're not testing the 'clean state' of Ghost but rather the functionality and behavior.

9larsons (approved):
As noted above, we're going to end up changing a lot of these tests and I didn't mean to hold up the posts app changes in here. We'll come back around once the Tags factory and Settings factory PRs are merged.

In general not a fan of how the tests read now, but it's not your fault. We need a few more pieces in place.

PR #24821, opened 2025-09-04
Updated daily charts of Analytics to show hours

peterzimon:
ref https://linear.app/ghost/issue/PROD-1858/add-hours-to-tooltip-when-in-today-view

- When "Today" is selected as the range for charts, we still show a daily breakdown in tooltips and in the chart labels. For daily range it should show hours which is much more useful for users.


9larsons:
I'm not 100% on this one. The display on the x-axis feels excessive, like we could still omit the timestamps. And the last data point being 11:00pm just feels weird to use. What's your take on it?

CleanShot 2025-09-04 at 09 16 51

9larsons (approved):
LGTM. I don't love the `showHours` prop at the top level that we have to drill down, but I don't see a better option right now. It's a shame that we have some components that have a start of day/eod data point where we can use hours when range === 1.

PR #24823, opened 2025-09-04
Added percentage to Total subscribers

peterzimon:
ref https://linear.app/ghost/issue/PROD-1816/add-growth-percentage-to-newsletter-subscriber-count

- While the chart line is useful, showing the percentage for total subscribers helps users understand the growth of their newsletter subscribers more concretely.


9larsons (approved):
I think it's good after a couple small tweaks. Please take a pass on your own again 👍 .

PR #24827, opened 2025-09-04
Added UTM query param parsing to frontend scripts

9larsons:
ref https://linear.app/ghost/issue/PROD-2560/
- added utm_ params parsing to the url-attribution helper library; this lib returns the referrer/attribution data we use for memberships in ghost
- refactored url-attribution script file to be easier to read/make sense of; updated tests accordingly (tried to decouple a bit w/ implementation)
- updated member-attribution script to pass along utm data; this isn't wired up on the backend yet

The `url-attribution` helpers are used by the `member-attribution` and `ghost-stats` scripts, which are our frontend scripts for handling attribution (member signups + web traffic). We need to update these to pass along the various `utm_{param}` query params in order to support expanding our attribution and analytics data sets.

This change allows us to pass along the data, and we'll need follow-up PRs to handle ingesting that data and properly storing it.

I've included a bit of a refactor here to try and make better sense of the url-attribution script. This has some confusion because historically it has been used for member attribution which has a confusing relationship with referrer data. We haven't been clear about what we call this.

__Testing__
In terms of testing, we should only need to check that the frontend scripts store the additional data. W/r to member-attribution, you can check `sessionStroage` and ensure the `utm_{param}` fields exist in the JSON data. For ghost-stats, you can look at the request payload sent to the traffic analytics proxy.

### DIFF
--- a/ghost/core/core/frontend/src/member-attribution/member-attribution.js
+++ b/ghost/core/core/frontend/src/member-attribution/member-attribution.js
@@ -88,9 +88,17 @@ const LIMIT = 15;
             referrerData = {source: null, medium: null, url: null};
         }
 
-        // Get referrer components from the parsed data
-        const referrerSource = referrerData.source;
-        const referrerMedium = referrerData.medium;
+        // Store all attribution data together
+        // We'll spread this object when creating history entries
+        const attributionData = {
+            referrerSource: referrerData.source,
+            referrerMedium: referrerData.medium,
+            utmSource: referrerData.utmSource,
+            utmMedium: referrerData.utmMedium,
+            utmCampaign: referrerData.utmCampaign,
+            utmTerm: referrerData.utmTerm,
+            utmContent: referrerData.utmContent
+        };
         
         // Use the getReferrer helper to handle same-domain referrer filtering
         // This will return null if the referrer is from the same domain
@@ -117,8 +125,7 @@ const LIMIT = 15;
                     time: currentTime,
                     id: params.get('attribution_id'),
                     type: params.get('attribution_type'),
-                    referrerSource,
-                    referrerMedium,
+                    ...attributionData,
                     referrerUrl
                 });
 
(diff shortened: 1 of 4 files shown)


ghost/core/core/frontend/src/member-attribution/member-attribution.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-env browser */
 /* eslint-disable no-console */
 const urlAttribution = require('../utils/url-attribution');
-const parseReferrer = urlAttribution.parseReferrer;
+const parseReferrerData = urlAttribution.parseReferrerData;
 const getReferrer = urlAttribution.getReferrer;
```
9larsons, line 5: These changes were done to prevent the wildly verbose file from getting more out of hand. There's very little of substantive changes here.

ghost/core/core/frontend/src/utils/url-attribution.js
```diff
@@ -3,63 +3,76 @@
...
+ * Prioritizes: source → medium → url
+ * Filters out same-domain referrers
+ * @private
+ * @param {AttributionData} referrerData - Parsed referrer data
+ * @returns {string|null} Primary referrer value or null
  */
-export function getFinalReferrer(referrerData) {
+function selectPrimaryReferrer(referrerData) {
     const { source, medium, url } = referrerData;
     const finalReferrer = source || medium || url || null;
```
9larsons, line 77: This fn is internal to this file, and I believe it was only exported for easier testing. Rather than trying to tightly test the fn, it made more sense to me to test the `getReferrer` tag more heavily to match our real use cases.

9larsons: I also tried to update the names here to be a bit more descriptive. I'd like to be more descriptive, but the use of referrer is tied to the consuming code, which felt like a massive breach of scope to also update/rethink.

ghost/core/core/frontend/src/utils/url-attribution.js
```diff
@@ -3,63 +3,76 @@
...
- * @returns {Object} Parsed attribution data
+ * @typedef {Object} AttributionData
+ * @property {string|null} source - Primary attribution source (ref || source || utm_source)
+ * @property {string|null} medium - UTM medium parameter
+ * @property {string|null} url - Browser's document.referrer
+ * @property {string|null} utmSource - UTM source parameter
+ * @property {string|null} utmMedium - UTM medium parameter
+ * @property {string|null} utmTerm - UTM term/keyword parameter
+ * @property {string|null} utmCampaign - UTM campaign parameter
+ * @property {string|null} utmContent - UTM content/variant parameter
```
cmraible, line 14: Nice, really helpful to have the types documented here 🙌

cmraible (commented):
Code all looks good to me, nice work! Just want to test a few things on staging, will approve shortly.

cmraible (approved):
Nice one, LGTM!

PR #24830, opened 2025-09-04
Added source tabs component to Shade

peterzimon:
ref https://linear.app/ghost/issue/PROD-2554/create-new-dropdown-components-for-utm-filters

- Source tabs in Analytics is a special feature component that is combining tabs and a dropdown menu. To be reusable in multiple places it's being added to Shade.


9larsons (approved):
I think there's a valid comment from CR but this is working well as-is, as the proposed changes mess up selection. All together, looks great.

I did remove the unused variant.

PR #24841, opened 2025-09-05
Added UTM dropdown to Stats > Web sources

9larsons:
ref https://linear.app/ghost/issue/PROD-2556/
- added new UTM filters to top sources component
- new filters are behind utmTracking flag
- expanded scope of TinybirdService to include new pipes for JWT

There's a few rough edges here, to be expected. The rendering of the Sources subcomponent flashes the full empty state when loading each of the UTM filters. We can fix that down the line, as well as sorting out how we want to do icon display.

__Testing__
- [ ] Make sure the flag shows/hides the UTM filters
- [ ] Make sure the return values display appropriately (can check the Network tab for responses; these are fixtures for the moment)

### DIFF
--- a/apps/stats/src/providers/GlobalDataProvider.tsx
+++ b/apps/stats/src/providers/GlobalDataProvider.tsx
@@ -5,17 +5,8 @@ import {Setting, useBrowseSettings} from '@tryghost/admin-x-framework/api/settin
 import {StatsConfig, useTinybirdToken} from '@tryghost/admin-x-framework';
 import {useBrowseSite} from '@tryghost/admin-x-framework/api/site';
 
-// The actual shape of the config data from the API
-type ConfigData = Config & {
-    config?: {
-        stats?: StatsConfig;
-        labs?: Record<string, boolean>;
-        [key: string]: unknown;
-    };
-};
-
 interface GlobalData {
-    data: ConfigData | undefined;
+    data: Config | undefined;
     site: {
         url?: string;
         icon?: string;
@@ -46,15 +37,10 @@ export const useGlobalData = () => {
 const GlobalDataProvider = ({children}: { children: ReactNode }) => {
     const settings = useBrowseSettings();
     const site = useBrowseSite();
-    // The actual response structure includes nested config with labs and stats
-    const configQuery = useBrowseConfig();
-    const config = {
-        data: configQuery.data?.config as ConfigData | undefined,
-        isLoading: configQuery.isLoading,
-        error: configQuery.error as Error | null
-    };
-    // Only fetch Tinybird token if stats config is present
-    const hasStatsConfig = Boolean(config.data?.config?.stats);
+    const config = useBrowseConfig();
+    // config.data is ConfigResponseType which has shape { config: Config }
+    const configData = config.data?.config;
+    const hasStatsConfig = Boolean(configData?.stats);
     const tinybirdTokenQuery = useTinybirdToken({enabled: hasStatsConfig});
     const [range, setRange] = useState(STATS_RANGE_OPTIONS[STATS_DEFAULT_RANGE_KEY].value);
     // Initialize with all audiences selected (binary 111 = 7)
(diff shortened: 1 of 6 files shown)


apps/stats/src/providers/GlobalDataProvider.tsx
```diff
@@ -5,8 +5,17 @@ import {Setting, useBrowseSettings} from '@tryghost/admin-x-framework/api/settin
 import {StatsConfig, useTinybirdToken} from '@tryghost/admin-x-framework';
 import {useBrowseSite} from '@tryghost/admin-x-framework/api/site';
 
+// The actual shape of the config data from the API
```
9larsons: This is just cleaning up some ugly type handling. We didn't pull labs out of this previously (or we stripped it out when we removed the labs flags), so we need some additional typing. This probably needs a revisit down the road. We shouldn't ultimately need to vary our type here but pull it from framework.

apps/stats/src/views/Stats/Web/Web.tsx
```diff
@@ -82,13 +87,64 @@ const Web: React.FC = () => {
...
+        'UTM sources': 'api_top_utm_sources',
+        'UTM mediums': 'api_top_utm_mediums',
+        'UTM campaigns': 'api_top_utm_campaigns',
+        'UTM contents': 'api_top_utm_contents',
+        'UTM terms': 'api_top_utm_terms'
+    };
+
+    // Get UTM campaign data (only fetch when a campaign is selected)
+    const campaignEndpoint = selectedCampaign ? campaignEndpointMap[selectedCampaign] : '';
+    const {data: utmData, loading: isUtmLoading} = useTinybirdQuery({
```
9larsons, line 109: This is a kind-of annoying block of code to handle the various states. I think it works ok, but we could break this down into a subcomponent if preferred. We're still in the realm of reasonable component size here.

apps/stats/src/views/Stats/Web/components/SourcesCard.tsx
```diff
@@ -136,20 +147,40 @@ export const SourcesCard: React.FC<SourcesCardProps> = ({
...
+                            title={`No visitors ${getPeriodText(range)}`}
+                        >
+                            <LucideIcon.Globe strokeWidth={1.5} />
+                        </EmptyIndicator>
+                    ) : (
+                        // Show skeleton while loading if no data
+                        <div className='mt-4'>
+                            <SkeletonTable lines={5} />
+                        </div>
+                    )
```
9larsons: I was trying to prevent the flashing state when switching. This part may not ultimately be necessary, though I think we should revisit in a follow-up PR the whole loading state behavior.

ghost/core/core/server/services/tinybird/TinybirdService.js
```diff
@@ -53,7 +53,12 @@ const TINYBIRD_PIPES = [
     'api_top_locations',
     'api_top_os',
     'api_top_pages',
-    'api_top_sources'
+    'api_top_sources',
+    'api_top_utm_sources',
+    'api_top_utm_mediums',
+    'api_top_utm_campaigns',
+    'api_top_utm_contents',
+    'api_top_utm_terms'
```
9larsons, line 61: This is needed for the JWT creation to be allowed to access the new endpoints.

apps/stats/src/providers/GlobalDataProvider.tsx
9larsons: Fixed.

apps/stats/src/views/Stats/Web/components/SourcesCard.tsx
9larsons: Reverted.

apps/stats/src/views/Stats/Web/components/SourcesCard.tsx
```diff
@@ -136,7 +147,19 @@ export const SourcesCard: React.FC<SourcesCardProps> = ({
...
+                    <>
+                        <div className='mb-2'>
+                            <SourceTabs
+                                selectedCampaign={selectedCampaign}
+                                selectedTab={selectedTab}
+                                onCampaignChange={onCampaignChange}
+                                onTabChange={onTabChange}
+                            />
+                        </div>
+                        <Separator />
```
peterzimon: I think this separator should be outside the flag condition

peterzimon (dismissed):

PR #24858, opened 2025-09-09
Add Accordion to Shade

jonatansberg:
Add Accordion component from shadcn to Shade. This is needed for the tag details screen.


peterzimon (approved):
@jonatansberg I added accordion animations to tailwind & changed classname ordering to follow standards.

PR #24859, opened 2025-09-09
Moved ghost-stats script above injected code in ghost_head

9larsons:
ref https://linear.app/ghost/issue/PROD-2612/

This is a simple re-ordering change to improve the resiliency of the ghost-stats script as it relates to injected code. Because of how ghost_head is built, if a user doesn't create the correct enclosures in their injected header code, it's possible for the ghost-stats script to be appended to that code. This can result in the script not loading at all.

As a general best practice, we should do all of our script loading before the injected code.


cmraible (approved):
Good idea! LGTM

PR #24861, opened 2025-09-09
Added lint & build guardrails to e2e package

9larsons:
no ref
- linter ignores locally built files, if you end up building them
- changed tsc command to noEmit to prevent building files accidentally

Playwright is able to run Typescript on the fly so we don't need to build actual files - all we need or want is the transpilation checking. So far this package isn't used elsewhere and I don't believe will be, either.


cmraible (approved):
Nice, I came across the same linting thing yesterday — thanks for fixing! 🙌

PR #24862, opened 2025-09-09
Added isolated environment setup to e2e test suite

cmraible:
Currently our e2e test suite runs against an external instance of Ghost, which you have to start manually. Additionally, there's no DB reset or reboot between tests, so any tests that change Ghost's state will pollute all the other tests.

This PR adds a global setup to start shared services like MySQL and Tinybird, and a Playwright fixture that creates a new isolated instance of Ghost for each test. This allows us to change Ghost's state in tests without worry about cleaning up, since each test gets its own fresh database and Ghost container.

The meat of this change is in the following files:
- `e2e/helpers/environment/EnvironmentManager.ts` - contains all the global setup/teardown and the beforeEach/afterEach test logic, in a dedicated class that is fully decoupled from Playwright
- `e2e/helpers/playwright/fixture.ts` - the core Playwright fixture that is used in all the tests to create/teardown an isolated Ghost instance via the `EnvironmentManager`
- `e2e/compose.e2e.yml` - the Docker compose file used by the e2e tests. Completely isolated from the main `compose.yml` in the project's root — no more manually resetting database to run e2e tests

## Testing
Build Ghost and tb-cli containers: `yarn docker:build`
Run the e2e tests: `yarn test:e2e`


.github/actions/setup-docker-registry-mirrors/action.yml
```diff
@@ -0,0 +1,11 @@
...
+description: 'Configure Docker daemon to use mirror.gcr.io registry mirror'
+runs:
+  using: 'composite'
+  steps:
+    - name: Configure Docker daemon for mirror.gcr.io
+      shell: bash
+      run: |
+        echo '{"registry-mirrors": ["https://mirror.gcr.io"]}' | sudo tee /etc/docker/daemon.json
+        sudo systemctl restart docker
+        docker logout || echo "No Docker Hub login found"
```
cmraible, line 11: This was originally hardcoded [here](https://github.com/TryGhost/Ghost/pull/24862/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL819-L823) in the `ci.yml` workflow, and I just used this opportunity to pull it out into a reusable action. 

This configured Docker to use Google Cloud's registry mirror to avoid Docker Hub rate limits in CI.

.github/workflows/ci-docker.yml
```diff
@@ -11,7 +11,110 @@ on:
...
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Install dependencies
+        run: yarn install --prefer-offline --frozen-lockfile
+
+    outputs:
+      dependency_cache_key: ${{ env.cachekey }}
+      node_version: ${{ env.NODE_VERSION }}
+
```
cmraible: For now we still need this, since the e2e tests run "locally" themselves, even though all the infrastructure and Ghost instances are running in Docker. We should be able to get the tests to _also_ run in Docker, but it's a little complicated to get it all working with Docker-in-Docker.

e2e/helpers/utils/setup-user.ts
```diff
@@ -14,8 +16,9 @@ export class GhostUserSetup {
     }
 
     async setup(userOverrides: Partial<User> = {}): Promise<void> {
+        debug('setup-user called');
         if (await this.isSetupAlreadyCompleted()) {
-            logging.info('Ghost user setup is already completed.');
```
cmraible, line 18: With the new setup, this runs for every test, and the logs were nonsensical with more than one worker running

e2e/tests/auth.setup.ts
```diff
@@ -1,43 +0,0 @@
...
-    await loginPage.waitForLoginPageAfterUserCreated();
-
-    await loginPage.signIn(appConfig.auth.email, appConfig.auth.password);
-
-    // waiting for the page to be loaded before storing the credentials
-    const analyticsPage = new AnalyticsOverviewPage(page);
-    await expect(analyticsPage.header).toBeVisible();
-
-    await page.context().storageState({path: authFile});
-});
```
cmraible, line 43: Auth setup is no longer done once globally, but once for each test, since we have a fresh instance per test. We may be able to find a way to do this once (before the DB dump) and save the state, but for now I've just removed this.

e2e/helpers/environment/EnvironmentManager.ts
```diff
@@ -0,0 +1,559 @@
...
+    }
+
+    /**
+     * Execute a command in a container and wait for completion
+     * @param container - The Docker container to execute the command in
+     * @param command - The shell command to execute
+     * @returns The command output
+     * @throws Error if the command fails
+     */
+    private async execInContainer(container: Container, command: string): Promise<string> {
```
9larsons: This feels like we're reinventing the wheel. Dockerode seems to have methods to handle streams. Can you explain why we need this overhead/complexity?

e2e/helpers/environment/EnvironmentManager.ts
```diff
@@ -0,0 +1,559 @@
...
+                TB_HOST: 'http://tinybird-local:7181',
+                TB_LOCAL_HOST: 'tinybird-local',
+                tinybird__stats__endpoint: 'http://tinybird-local:7181',
+                tinybird__stats__endpointBrowser: 'http://localhost:7181',
+                tinybird__tracker__endpoint: 'http://localhost/.ghost/analytics/api/v1/page_hit',
+                tinybird__workspaceId: tinybirdState.workspaceId,
+                tinybird__adminToken: tinybirdState.adminToken
+            };
+
+            const containerConfig = {
```
9larsons:
```suggestion
            const containerConfig: Docker.ContainerCreateOptions = {
```

.github/workflows/ci-docker.yml
```diff
@@ -11,7 +11,110 @@ on:
...
+  setup:
+    name: Setup
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    env:
+      IS_MAIN: ${{ github.ref == 'refs/heads/main' }}
+      IS_DEVELOPMENT: ${{ github.ref == 'refs/heads/main' || github.ref == 'refs/heads/5.x' || github.ref == 'refs/heads/6.x' }}
+      IS_SIX: ${{ github.ref == 'refs/heads/6.x' }}
+      IS_SIX_PR: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.ref == '6.x' }}
```
9larsons: Do we need these IS_SIX and IS_SIX_PR flags?

9larsons (approved):
Overall, looks great and works great, nice work! No major comments/questions or anything that I think should hold this up, though there are things I wouldn't mind following up on. Excited to start using this.

General comments/concerns:
- `EnvironmentManager` is doing _a lot_. As we add more service containers to orchestrate (like AP), I'm concerned with this ballooning. Breaking this up adds complexity, but I think it could help make more clear what the dependency chain and order of operations is with Ghost.
- Debug logging, while helpful, is noisy when trying to read through the code. Not sure what to do about this, as it's obnoxious to add this in temporarily for any debugging.
- I'm curious what a PageHitFactory addition will look like with this setup, and I think this'll test how this works in practice.

.github/workflows/ci-docker.yml
cmraible: Nope, good catch — removed them.

e2e/helpers/environment/EnvironmentManager.ts
cmraible: Updated!

e2e/helpers/environment/EnvironmentManager.ts
cmraible: Yeah agree that this ended up fairly complex, but I haven't been able to find a way to simplify it. We are using dockerode's method for handling the streams (the call to `modem.demuxStream`) but there's still a fair bit of boiler plate necessary to do anything useful with them, AFAICT.

.github/workflows/ci-docker.yml
```diff
@@ -184,3 +284,58 @@ jobs:
...
+          image-tags: ${{ needs.build.outputs.image-tags }}
+
+      - name: Setup Docker Registry Mirrors
+        uses: ./.github/actions/setup-docker-registry-mirrors
+
+      - name: Pull images
+        env:
+          GHOST_IMAGE_TAG: ${{ steps.load.outputs.image-tag }}
+        run: docker compose -f e2e/compose.e2e.yml pull
+
```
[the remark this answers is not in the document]
cmraible: I've tested this on a fork and it works fine.

e2e/helpers/environment/MySQLManager.ts
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Execute a command in a container and wait for completion
+     * @param container - The Docker container to execute the command in
+     * @param command - The shell command to execute
+     * @returns The command output
+     * @throws Error if the command fails
+     */
+    private async execInContainer(container: Container, command: string): Promise<string> {
```
9larsons, line 123: Could you add a touch of context around why this is needed in the header?

e2e/helpers/environment/EnvironmentManager.ts
```diff
@@ -0,0 +1,140 @@
...
+        });
+        this.mysql = new MySQLManager(this.dockerCompose);
+        this.tinybird = new TinybirdManager(this.dockerCompose);
+        this.ghost = new GhostManager(this.docker, this.dockerCompose, this.tinybird);
+    }
+
+    /**
+     * Setup shared global environment for tests (i.e. mysql, tinybird)
+     */
+    public async globalSetup(): Promise<void> {
```
9larsons, line 65: I think this file could be the one to benefit from some additional context in comments as it is the 'conductor'.

e.g. for `globalSetup`, it might help to spell out the purpose of each step:
- ghost-migrations run before snapshot so that we can load the snapshot for each test
- tb-cli sets up the .env for tinybird-local

It's also worth pointing out that that it's a little unintuitive (as in non-declarative) that it appears mysql is booted as a dependency of ghost-migrations. I'd expect to see us starting the mysql container either along with or before the ghost-migrations container.

9larsons: Mentioning that state files are created is also helpful, as the teardown is clearing out state files. For the most part, I'd expect these two to mirror each other.

e2e/helpers/environment/TinybirdManager.ts
```diff
@@ -0,0 +1,92 @@
...
+export class TinybirdManager {
+    private readonly tinybirdStateFile = path.join(STATE_DIR, 'tinybird.json');
+    private dockerCompose: DockerCompose;
+
+    constructor(dockerCompose: DockerCompose) {
+        this.dockerCompose = dockerCompose;
+    }
+
+    /** Ensure the state directory exists. */
+    private ensureDataDir(): void {
```
9larsons: These fs operations feel like utils.

9larsons (approved):
This reads and feels loads better, thanks for giving it another pass!

My only quibble now is seeing the use of `this.` everywhere, ha, but that's just JS. I think the class treatment is right.

Some minor comments about adding some context and possible moving a few methods to a utils file. Don't feel strongly about any of it.

e2e/helpers/environment/EnvironmentManager.ts
cmraible: Yeah good call, I'll add some more context about everything that's happening here!

PR #24869, opened 2025-09-10
Updated ghost-stats UTM payload

9larsons:
ref https://linear.app/ghost/issue/PROD-2619/
- updated payload to only keep referrer fields in parsedReferrer, utm fields at top level
- updated tests
- refactored tests to be a bit more readable/usable

After deciding to change the shape of the payload in the TryGhost/TrafficAnalytics repo, we need this change to the ghost-stats payload.

### DIFF
--- a/ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
+++ b/ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
@@ -182,11 +182,11 @@ export class GhostStats {
                 },
                 pathname: location?.pathname,
                 href: location?.href,
-                utmSource: referrerData.utmSource,
-                utmMedium: referrerData.utmMedium,
-                utmCampaign: referrerData.utmCampaign,
-                utmTerm: referrerData.utmTerm,
-                utmContent: referrerData.utmContent
+                utm_source: referrerData.utmSource,
+                utm_medium: referrerData.utmMedium,
+                utm_campaign: referrerData.utmCampaign,
+                utm_term: referrerData.utmTerm,
+                utm_content: referrerData.utmContent
             });
         }, 300);
     }
--- a/ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
+++ b/ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
@@ -175,9 +175,18 @@ export class GhostStats {
                 'user-agent': navigator?.userAgent,
                 locale,
                 location: country,
-                parsedReferrer: referrerData,
+                parsedReferrer: {
+                    url: referrerData.url,
+                    source: referrerData.source,
+                    medium: referrerData.medium,
+                },
                 pathname: location?.pathname,
                 href: location?.href,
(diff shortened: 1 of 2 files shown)


ghost/core/core/frontend/src/ghost-stats/ghost-stats.js
```diff
@@ -175,9 +175,18 @@ export class GhostStats {
...
                 location: country,
-                parsedReferrer: referrerData,
+                parsedReferrer: {
+                    url: referrerData.url,
+                    source: referrerData.source,
+                    medium: referrerData.medium,
+                },
                 pathname: location?.pathname,
                 href: location?.href,
+                utmSource: referrerData.utmSource,
```
cmraible: suggestion/nit: We haven't really talked about this, but I think it would be good to standardize on either snake_case or camelCase. We _mostly_ use snake_code in JSON payloads, so IMO we should stick with that.

ghost/core/test/unit/frontend/public/ghost-stats.test.js
```diff
@@ -457,17 +524,6 @@ describe('ghost-stats.js', function () {
...
-            // Make it a test environment
-            Object.defineProperty(mockWindow, 'Cypress', {
-                value: {},
-                configurable: true
-            });
-
-            expect(ghostStats.init()).to.be.false;
-            mockWindow.Cypress = undefined;
-        });
-
```
cmraible, line 470: Curious what the rationale for removing this test is — it doesn't look like the code itself has changed

ghost/core/test/unit/frontend/public/ghost-stats.test.js
```diff
@@ -96,30 +161,6 @@ describe('ghost-stats.js', function () {
...
-            expect(browserService.isTestEnvironment()).to.be.true;
-            delete mockWindow.__nightmare;
-
-            Object.defineProperty(mockWindow.navigator, 'webdriver', {
-                value: true,
-                configurable: true
-            });
-            expect(browserService.isTestEnvironment()).to.be.true;
-        });
-
```
cmraible, line 122: I'm curious why this test is removed?

cmraible (commented):
Overall looks good! Left a few questions, and a comment on camelCase vs snake_case. 

Really like the simplification of the tests with the helper functions, it makes them much more readable.

ghost/core/test/unit/frontend/public/ghost-stats.test.js
9larsons: We don't use Cypress so it felt a bit superfluous. I was just cleaning what felt unnecessary as I was trying to make the test file a bit more readable.

ghost/core/test/unit/frontend/public/ghost-stats.test.js
9larsons: Same as above - we don't use these and they felt unnecessary.

cmraible (approved):

PR #24872, opened 2025-09-10
Updated portal messaging to show email

troyciesco:
ref https://linear.app/ghost/issue/PROD-2620

- After a user signs in, if OTC is turned on, the next screen ("check your email") will now display the email address instead of just "your inbox"
- This will make it easier to see if there was a typo in the email address, in the event an email never arrives
- different messages for sign in without OTC available, and sign in with OTC available

Screenshot 2025-09-23 at 12 36 45 PM

sign in without OTC is _not_ impacted:
Screenshot 2025-09-23 at 12 39 53 PM

nor is sign up:
Screenshot 2025-09-23 at 12 40 19 PM


ghost/i18n/locales/af/portal.json
```diff
@@ -12,13 +12,14 @@
...
     "A login link has been sent to your inbox. If it doesn't arrive in 3 minutes, be sure to check your spam folder.": "'n Aanmeldskakel is na u epos gestuur. As dit nie binne 3 minute aankom nie, moet u asseblief u spam-vouer nagaan.",
+    "A sign in link has been sent to {submittedEmailOrInbox}. If it doesn't arrive in 3 minutes, be sure to check your spam folder.": "",
     "Account": "Rekening",
     "Account details updated successfully": "",
     "Account settings": "Rekening instellings",
     "Add a personal note": "",
     "After a free trial ends, you will be charged the regular price for the tier you've chosen. You can always cancel before then.": "Na afloop van die gratis proeftydperk sal u die vasgestelde pry vir die vlak wat u gekies het, betaal. U kan altyd voor die tyd kanselleer.",
     "Already a member?": "Is u reeds 'n lid?",
-    "An email has been sent to your inbox. Use the link inside or enter the code below.": "",
+    "An email has been sent to {submittedEmailOrInbox}. Click the link inside or enter your code below.": "",
```
troyciesco, line 21: this one is meant to be changing - the previous one was the first iteration for OTC and didn't have any translations yet.

kevinansfield (changes requested):
As discussed in our sync call, it would be good to put the updated strings behind a labs flag so we have new strings ready for early translation rather than production strings being cleared out immediately.

ghost/core/test/unit/frontend/helpers/__snapshots__/ghost_head.test.js.snap
```diff
@@ -1097,57 +1097,6 @@ Object {
 }
 `;
 
-exports[`{{ghost_head}} helper includes tinybird tracker script when config is set does not include tracker script when neither trafficAnalytics nor trafficAnalyticsTracking is set 1 1`] = `
```
troyciesco, line 1100: hmm more changes than expected here 👀

troyciesco: ok reviewed a bit, these few analytics snapshots can be cleaned up safely bc the tests no longer exist. In particular, trafficAnalyticsTracking is an old flag that doesn't exist anymore.

troyciesco:
> As discussed in our sync call, it would be good to put the updated strings behind a labs flag so we have new strings ready for early translation rather than production strings being cleared out immediately.

@kevinansfield I'd originally used the otc flag to gate this change, thinking that we'd want to switch to email text at the same time otc is GA, but now realize it doesn't necessarily make sense to do that, especially since a few otc translations have already been done.

I've updated this PR with a new flag specifically for the email text, and cleaned up the creation of the `popupDescription` variable a bit so there wasn't so much hard-to-parse if/else logic.

kevinansfield (approved):

PR #24874, opened 2025-09-10
Added ghost assets build to Dockerfile

cmraible:
Ghost's assets (i.e. `comment-counts.js`, `ghost-stats.js`) have a build step which was missing from the Dockerfile. This is normally done automatically when running `yarn dev` so it went unnoticed, but as we move away from the `dev.js` script, we should include this step in the image build directly.


9larsons (approved):
Good add!

PR #24875, opened 2025-09-11
🎨 Added uuid column to tokens table

troyciesco:
ref https://linear.app/ghost/issue/PROD-2492

- Adds the `uuid` column to the tokens table
- We want to keep `token.id` non-public so we are able to use it any future Admin operations, and so there's no public data used when deriving a token code
- But we need to provide a public lookup reference from the `/send-magic-link/` endpoint so clients can pass the reference+code (future work)
- Adding a `uuid` column follows patterns we use elsewhere

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/6.1/2025-09-11-00-38-13-add-uuid-column-to-tokens.js
+++ b/ghost/core/core/server/data/migrations/versions/6.1/2025-09-11-00-38-13-add-uuid-column-to-tokens.js
@@ -1,0 +1,8 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('tokens', 'uuid', {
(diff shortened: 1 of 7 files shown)


ghost/core/core/server/data/migrations/versions/6.1/2025-09-11-00-38-13-add-uuid-column-to-tokens.js
```diff
@@ -0,0 +1,8 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('tokens', 'uuid', {
```
janedotx, line 3: Does this column need to be indexed? If it does, UUID7 may be better. Also, might be worth calling out what kind of UUID is being used.

kevinansfield (approved):

PR #24879, opened 2025-09-11
🐛 Fixed limit changes not updating without page refresh

aileen:
closes [BAE-492](https://linear.app/ghost/issue/BAE-492/limit-changes-requires-a-page-refresh-to-be-reflected-in-admin)

When users upgraded or changed their subscription plan through the billing iframe, the configuration limits in Admin Settings remained stale until a manual page refresh. This prevented users from immediately accessing newly unlocked features like Stripe Connect, Web Analytics, or additional newsletter creation after upgrading their plan.

The issue occurred because React Query was caching the config data in Admin-X Settings without invalidation when external subscription updates occurred, the `useLimiter` hook's memoization wasn't properly tracking config changes due to using a nested property (`config.hostSettings?.limits`) instead of the full config object as a dependency, and the limit service in Ghost Admin wasn't reloading when subscription changes were received.

This PR implements a comprehensive fix that ensures all admin interfaces update in real-time when plan changes occur. Added React Query cache invalidation in `gh-billing-iframe.js` when subscription update messages are received from the billing iframe, with fallback support for both React Query v4 and v5. Implemented config refresh and limit service reload to ensure Ghost Admin pages can immediately enforce updated limits. Fixed memoization in `useLimiter` hook to properly track config changes by using the full `config` object as a dependency. Added defensive no-op limiter API to prevent runtime errors when the limiter service isn't ready, improving overall stability.

Users now have immediate access to newly available features after plan upgrades without requiring page refreshes. All limit-dependent UI elements update in real-time across both Ghost Admin and Admin-X Settings. Improved error handling prevents potential runtime exceptions when accessing React Query client or during initialization.

Due to the nature of billing integrations, this needs to be verified on staging environments with actual subscription changes.

### DIFF
--- a/apps/admin-x-settings/src/hooks/useLimiter.tsx
+++ b/apps/admin-x-settings/src/hooks/useLimiter.tsx
@@ -74,8 +74,18 @@ export const useLimiter = () => {
     }, [config.hostSettings?.billing]);
 
     return useMemo(() => {
+        // Return a stable no-op API when the limiter isn't ready
+        // This prevents runtime errors while maintaining backward compatibility
+        const noOpLimiter = {
+            isLimited: (): boolean => false,
+            isDisabled: (): boolean => false,
+            checkWouldGoOverLimit: (): Promise<boolean> => Promise.resolve(false),
+            errorIfWouldGoOverLimit: (): Promise<void> => Promise.resolve(),
+            errorIfIsOverLimit: (): Promise<void> => Promise.resolve()
+        };
+
         if (!LimitService || !config.hostSettings?.limits || isLoading) {
-            return;
+            return noOpLimiter;
         }
 
         const limits = {...config.hostSettings.limits} as LimiterLimits;
--- a/apps/admin-x-settings/src/hooks/useLimiter.tsx
+++ b/apps/admin-x-settings/src/hooks/useLimiter.tsx
@@ -122,5 +122,5 @@ export const useLimiter = () => {
             errorIfWouldGoOverLimit: (limitName: string, metadata: Record<string, unknown> = {}): Promise<void> => limiter.errorIfWouldGoOverLimit(limitName, metadata),
             errorIfIsOverLimit: (limitName: string): Promise<void> => limiter.errorIfIsOverLimit(limitName)
         };
-    }, [LimitService, config.hostSettings?.limits, contributorUsers, fetchMembers, fetchNewsletters, helpLink, invites, isLoading, users]);
+    }, [LimitService, config, contributorUsers, fetchMembers, fetchNewsletters, helpLink, invites, isLoading, users]);
 };
--- a/ghost/admin/app/components/gh-billing-iframe.js
+++ b/ghost/admin/app/components/gh-billing-iframe.js
@@ -119,7 +119,7 @@ export default class GhBillingIframe extends Component {
         await this.limit.reload();
 
         // Invalidate React Query cache for config data in admin-x-settings
-        if (window.adminXQueryClient) {
+        if (window?.adminXQueryClient?.invalidateQueries && typeof window.adminXQueryClient.invalidateQueries === 'function') {
             window.adminXQueryClient.invalidateQueries(['ConfigResponseType']);
         }
 
--- a/ghost/admin/app/components/gh-billing-iframe.js
+++ b/ghost/admin/app/components/gh-billing-iframe.js
@@ -48,10 +48,6 @@ export default class GhBillingIframe extends Component {
             if (event.data?.subscription) {
                 this._handleSubscriptionUpdate(event.data);
             }
-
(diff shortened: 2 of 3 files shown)


ghost/admin/app/components/gh-billing-iframe.js
```diff
@@ -46,6 +48,10 @@ export default class GhBillingIframe extends Component {
             if (event.data?.subscription) {
                 this._handleSubscriptionUpdate(event.data);
             }
+
+            if (event.data?.request === 'refreshRequired') {
+                this._handleRefreshRequired();
+            }
```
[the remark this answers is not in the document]
aileen: Good catch

ghost/admin/app/components/gh-billing-iframe.js
```diff
@@ -105,7 +107,18 @@ export default class GhBillingIframe extends Component {
...
+        await this.configManager.fetch();
+
+        // Reload the limit service to ensure all admin pages can enforce limits
+        this.limit.reload();
+
+        // Invalidate React Query cache for config data in admin-x-settings
+        if (window?.adminXQueryClient?.invalidateQueries && typeof window.adminXQueryClient.invalidateQueries === 'function') {
+            window.adminXQueryClient.invalidateQueries(['ConfigResponseType']);
+        }
+
```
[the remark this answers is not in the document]
aileen: We're currently at version 4.36.1 of React Query. Might be a good idea to use this fallback, just so things don't break accidentally when Admin-X updates the dependendy.

ghost/admin/app/components/gh-billing-iframe.js
```diff
@@ -105,7 +107,29 @@ export default class GhBillingIframe extends Component {
...
+        if (window?.adminXQueryClient?.invalidateQueries && typeof window.adminXQueryClient.invalidateQueries === 'function') {
+            try {
+                // React Query v5+
+                window.adminXQueryClient.invalidateQueries({queryKey: ['ConfigResponseType']});
+            } catch (_) {
+                // TODO: remove this fallback when Admin-X-Settings updates to React Query v5+ (@tanstack/react-query)
+                // React Query v4 fallback
+                window.adminXQueryClient.invalidateQueries(['ConfigResponseType']);
+            }
+        }
```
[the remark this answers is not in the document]
aileen: Verify this again, because the dataType assigned is [here](https://github.com/TryGhost/Ghost/blob/5496f113a45f2a20081fd4a8ec73f9502efc1800/apps/admin-x-framework/src/api/config.ts#L93-L94)

ghost/admin/app/services/limit.js
```diff
@@ -60,8 +67,8 @@ export default class LimitsService extends Service {
         });
     }
 
-    async checkWouldGoOverLimit(limitName, metadata = {}) {
-        return this.limiter.checkWouldGoOverLimit(limitName, metadata);
+    reload() {
+        this.loadLimits();
```
sam-lord, line 71: If the AI comment above is right about the general pattern being to have an async reload, then it could be a hint that we should put the call to `await configManager.fetch()` here instead of in the subscription handler

apps/admin-x-settings/src/hooks/useLimiter.tsx
```diff
@@ -74,8 +74,18 @@ export const useLimiter = () => {
     }, [config.hostSettings?.billing]);
 
     return useMemo(() => {
+        // Return a stable no-op API when the limiter isn't ready
+        // This prevents runtime errors while maintaining backward compatibility
+        const noOpLimiter = {
```
sam-lord, line 79: Hmm, could we have a task to replace this later?

I think it would be better to have this interface _always_ use promises so we can await the limits service being available. This feels like it could cause issues, but I'd be happy for the test to go ahead to confirm that the rest of the change works before we address that.

sam-lord (approved):
Approved, with some comments. I think it's worth us getting this in, but I'm most nervous about the noOpLimiter. I don't know enough of the rest of AdminX to know if there would immediately be calls to this which result in bad data before the limits service has initialized.

ghost/admin/app/services/limit.js
aileen: afaik `loadLimits` will trigger the same method on the actual `LimitService` instance. I verified and checked that this is actually a [synchronous method](https://github.com/TryGhost/SDK/blob/main/packages/limit-service/lib/LimitService.js#L28).

apps/admin-x-settings/src/hooks/useLimiter.tsx
aileen: Sounds good to me!

aileen: We're doing the checks in every consuming component of this service anyway and verify that the limit service is ready. There might be a more elegant way of doing this

PR #24885, opened 2025-09-11
Fixed reply count double increment bug in ActivityPub

minimaluminium:
ref https://linear.app/ghost/issue/PROD-2621/reply-count-increases-by-two-on-both-posts-and-notes

- Removed redundant reply handlers that were triggering from both FeedItemStats and APReplyBox
- Replaced manual state management with direct cache-based rendering
- Removed unnecessary useEffect that was syncing reply count from two sources
- Fixed cache updates with `exact: false` to update all reply chain queries (e.g. ['reply_chain', 'post-123'])

### DIFF
--- a/apps/admin-x-activitypub/src/hooks/use-activity-pub-queries.ts
+++ b/apps/admin-x-activitypub/src/hooks/use-activity-pub-queries.ts
@@ -533,7 +533,7 @@ function updateReplyCache(queryClient: QueryClient, id: string, delta: number) {
 function updateReplyCacheOnce(queryClient: QueryClient, id: string, delta: number) {
     // Update reply chain cache (used by Note.tsx and Reader.tsx)
     const replyChainQueryKey = QUERY_KEYS.replyChain(null);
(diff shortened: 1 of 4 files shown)


apps/admin-x-activitypub/src/hooks/use-activity-pub-queries.ts
```diff
@@ -533,7 +533,7 @@ function updateReplyCache(queryClient: QueryClient, id: string, delta: number) {
 function updateReplyCacheOnce(queryClient: QueryClient, id: string, delta: number) {
     // Update reply chain cache (used by Note.tsx and Reader.tsx)
     const replyChainQueryKey = QUERY_KEYS.replyChain(null);
-    queryClient.setQueriesData(replyChainQueryKey, (current?: ReplyChainResponse) => {
+    queryClient.setQueriesData({queryKey: replyChainQueryKey, exact: false}, (current?: ReplyChainResponse) => {
```
allouis, line 536: What does `exact: false` do?

allouis (approved):
This looks good, but I just have one question!

PR #24891, opened 2025-09-11
Added Portal loading E2E tests

9larsons:
no ref

- Fixed Ghost container to use deterministic port assignment based on worker index
- Added URL environment variable so Portal can correctly communicate with Ghost API
- Created Portal page objects (SignUpPage, SignInPage, SignUpSuccessPage)
- Added portal-helpers utility for opening Portal in tests
- Implemented basic Portal loading tests to verify iframe loads and functions

This establishes the foundation for member attribution E2E tests by ensuring Portal (the membership widget) can successfully load and interact with Ghost. From here, we can work towards a complete signin/signup flow.

Re: Port assignment - we pass the `config.url` in various places in Ghost, so we need to know what port we're running on vs. letting Docker auto-assign. There's some janky ways to do this with letting Docker auto-assign (and Ghost probably shouldn't need it fed to it, but rather be able to detect it), so instead we'll deterministically assign based on the worker which should give us the fidelity we need w/o overcomplicating it.


e2e/helpers/pages/portal/SignInPage.ts
```diff
@@ -0,0 +1,46 @@
...
+        // Use role-based selectors for better reliability
+        this.emailInput = this.portalFrame.getByRole('textbox', {name: 'Email'});
+        // Initial button says "Continue" in signin flow
+        this.continueButton = this.portalFrame.getByRole('button', {name: 'Continue'});
+        // After entering email, button changes to "Sign in"
+        this.signinButton = this.portalFrame.getByRole('button', {name: 'Sign in'});
+        this.signupLink = this.portalFrame.getByRole('button', {name: 'Sign up'});
+    }
+
+    async waitForFrame(): Promise<void> {
```
9larsons: I'm not 100% sure on what to do with this waitForFrame. If we need it in every Portal page object, I think creating a PortalPage base class is the right move. Consider this a WIP as I flesh out the full test flow.

e2e/helpers/utils/portal-helpers.ts
```diff
@@ -0,0 +1,34 @@
+import {Page} from '@playwright/test';
+
+/**
+ * Wait for the Portal script to be loaded on the page
+ */
+export async function waitForPortalScript(page: Page): Promise<void> {
```
9larsons, line 6: I think I might want to chat w/ Igor about this. I don't think this is completely necessary, also consider a WIP. The main thing here (imo) is we have some passing tests showing that Portal DOES/CAN load. Once we get full flows, we can cut these because the others will test these pieces implicitly.

e2e/helpers/environment/GhostManager.ts
```diff
@@ -48,6 +48,10 @@ export class GhostManager {
         try {
             const network = await this.dockerCompose.getNetwork();
             const tinybirdState = this.tinybird.loadState();
+            
+            // Use deterministic port based on worker index (or 0 if not in parallel)
+            const hostPort = 30000 + parseInt(process.env.TEST_PARALLEL_INDEX || '0', 10);
```
cmraible, line 53: note (non-blocking): wanted to call out that this is a smell to me — I'd prefer to continue letting Docker assign a host port to guarantee we don't have port conflicts. 

But I see the dilemma — if we let Docker assign the port, then we don't know the port until after Ghost boots, and it's too late to pass it to Portal via Ghost's config.

Before we go too far down this road, I think we need to get the tests themselves running in Docker. That way, we don't need the host port at all: playwright/chrome itself will be running in the docker network, so we can address each Ghost instance by its hostname (which we know in advance) rather than a randomly assigned host port. This way the `url` would be `http://ghost_262b77db-f1c4-4a45-9ab8-b03bdf830c43:2368` rather than `http://localhost:{hostPort}`.

Running playwright/browsers in Docker also speeds up CI and makes local setup easier, since we can bake it into the Docker image, rather than manually installing it locally in development and CI. 

No objection to this approach for now, just thinking out loud 👌

e2e/helpers/pages/portal/SignInPage.ts
cmraible: +1 to `PortalPage` base class, I think that sounds reasonable to me.

Not a fully formed thought, but Playwright has a [Frame class](https://playwright.dev/docs/api/class-frame) with some utilities that might be useful here:
- `frame.waitForLoadState` ([link](https://playwright.dev/docs/api/class-frame#frame-wait-for-load-state))
- `frame.frameLocator` ([link](https://playwright.dev/docs/api/class-frame#frame-frame-locator))

e2e/tests/public/portal-loading.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        await expect(signUpPage.signupButton).toBeVisible();
+        
+        // Switch to signin mode
+        await signUpPage.signinLink.click();
+        
+        // Verify signin form is visible
+        const signInPage = new SignInPage(page);
+        await expect(signInPage.emailInput).toBeVisible();
+        await expect(signInPage.continueButton).toBeVisible();
+    });
```
cmraible, line 63: nit: I'd remove the comments, for the most part. I think these tests read cleanly enough that they are basically redundant with the code

e2e/helpers/utils/portal-helpers.ts
cmraible: Sounds good! 

I'm not totally sure about this, just thinking out loud: since portal loads on every frontend page (I think?), maybe these helpers would make sense as methods on a base `FrontendPage`?

e2e/helpers/pages/portal/SignUpSuccessPage.ts
```diff
@@ -0,0 +1,68 @@
...
+        });
+    }
+
+    async isPortalClosed(): Promise<boolean> {
+        const frame = await this.page.$('[data-testid="portal-popup-frame"]');
+        if (!frame) {
+            return true;
+        }
+        return !(await frame.isVisible());
+    }
```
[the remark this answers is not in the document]
cmraible: Not sure if we need the conditional and early return, but agree that we should use `page.locator` here instead of `page.$()`, which is [deprecated](https://playwright.dev/docs/api/class-page#page-query-selector).

e2e/helpers/pages/portal/SignUpPage.ts
```diff
@@ -0,0 +1,52 @@
...
+        await this.emailInput.fill(email);
+    }
+
+    async fillName(name: string): Promise<void> {
+        await this.nameInput.fill(name);
+    }
+
+    async submit(): Promise<void> {
+        await this.signupButton.click();
+    }
```
cmraible: note (non-blocking): we haven't discussed this, but I think we should align on when we create methods for atomic actions like this vs just exposing the locator and using `page.locator.click()` in tests directly.

My personal take is that methods should be reserved for composing multiple actions into a single unit, and adding methods for these one-liners doesn't add much value. No strong feelings though, just think it's worth mentioning so we can apply it consistently.

cmraible (approved):
Left some comments with thoughts, but nothing blocking! This is all looking really good and starting to feel a lot easier now that we have these running in CI 🎉

One last note, which also isn't blocking but food for thought: if I'm understanding this properly, this is using portal from the CDN, which means we can't use the E2E tests to test changes to Portal. We don't need to cover that in this PR, but I think we should switch to serving the local version of portal from Ghost in our E2E tests.

e2e/helpers/pages/portal/SignInPage.ts
9larsons: Oh good shout, I'll take a look.

e2e/helpers/utils/portal-helpers.ts
9larsons: Also a good call 👍.

PR #24893, opened 2025-09-11
Added mailhog for e2e tests

9larsons:
ref https://linear.app/ghost/issue/PROD-2515/
- added Mailhog to the e2e package to support email related functionality; this is necessary piece of member signups
- added a member sign up flow test

This set of changes is to enable the ability to do full e2e member signup and attribution tests. In order to do that, we need email support. Here we add Mailhog, which is used in our development Docker setup, in order to accomplish this.

We are fairly tightly coupled to Mailhog here as we explore this option and how it works for testing before moving to an adapter pattern like we use elsewhere.


e2e/compose.e2e.yml
```diff
@@ -55,6 +55,16 @@ services:
       tinybird-local:
         condition: service_healthy
 
+  mailhog:
+    image: mailhog/mailhog:latest
+    platform: linux/amd64
+    ports:
+      - "1025:1025" # SMTP server
+      - "8025:8025" # Web interface
```
cmraible:
```suggestion
      - "1025" # SMTP server
      - "8025" # Web interface
```

To avoid port conflicts with our development compose project, we should let docker assign these ports automatically. We should be able to inspect the hostPort on the container for now. 

Alternatively, we could just increment the hostPort here so it doesn't conflict with the root `compose.yml`. Bit of a shortcut, but the problem goes away once playwright/browsers are running in Docker so it feels fine to me.

e2e/tests/public/member-signup-with-email.test.ts
```diff
@@ -0,0 +1,102 @@
...
+import {MailhogClient} from '../../helpers/email/MailhogClient';
+import {faker} from '@faker-js/faker';
+
+test.describe('Member Signup with Email Verification', () => {
+    let mailhog: MailhogClient;
+
+    test.beforeEach(async () => {
+        mailhog = new MailhogClient();
+        // Clear any existing emails before each test
+        await mailhog.deleteAllMessages();
```
cmraible: Since mailhog is shared across all tests, I think we need to find a better way to handle this — one test could be deleting all messages just as another test is checking for a message.

I think we either need to get more specific about which messages we're deleting (i.e. by worker index maybe?) or structure the tests in such a way that we don't need to delete messages at all.

e2e/helpers/email/MailhogClient.ts
```diff
@@ -0,0 +1,192 @@
+import baseDebug from '@tryghost/debug';
```
cmraible, line 1: note: I don't think we should abstract this just yet until we see how this works and feels, but I think we should have an eye toward an adapter pattern here. I could see us eventually wanting to swap mailhog for something else, or e.g. use Mailgun's APIs for this at some point.

e2e/tests/public/member-signup.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        // Wait for success and close
+        const successPage = new SignUpSuccessPage(page);
+        await successPage.waitForSuccess();
+        await successPage.close();
+        
+        // Verify Portal is closed
+        const isClosed = await successPage.isPortalClosed();
+        expect(isClosed).toBe(true);
+    });
+});
```
cmraible, line 37: Were you intending to keep these, in addition to the full flow above?

cmraible (commented):
Nice, surprisingly clean and succinct for a first draft! I don't think we're far off here, the biggest open question to me is how we do the mailhog stuff in such a way that it plays nicely with parallelism.

e2e/tests/public/member-signup-with-email.test.ts
9larsons: Good reminder. Since we do lookup via email, I was planning to not remove any messages and we can use generators/helpers to make sure we're using unique email addresses.

e2e/compose.e2e.yml
9larsons: Let's just increment for now.

e2e/tests/public/member-signup.test.ts
9larsons: Nah I'll toss them.

e2e/helpers/utils/generators.ts
```diff
@@ -0,0 +1,92 @@
+import {faker} from '@faker-js/faker';
```
9larsons: I did this to decouple from faker. That said, we may want to point our data factory generators here.. I was a bit on the fence about that, as those are specific to the factories.

e2e/tests/public/member-signup.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    test('completes full signup flow with magic link', async ({page}) => {
+        const {email} = await signupViaPortal(page);
+        await mailhog.completeSignupVerification(page, email);
+
+        const homePage = new HomePage(page);
+        await homePage.waitForSignedIn();
+    });
+
+    test('receives welcome email with correct content', async ({page}) => {
```
9larsons, line 25: This test has a bit more expects than I'd like in the e2e tests, and arguably shouldn't be an e2e test. I've left it to proof out mailhog and we can scrutinize this further later on.

e2e/helpers/utils/generators.ts
ibalosh: I am wondering if we need generators, at least wondering if we need them this early. We will probably use later more methods from faker, and we would need to add extra wrappers here. Faker is pretty much a standard for generating data, so decoupling might not be needed.

Seems like a too early optimization, that we might not need, at least not yet?

e2e/helpers/utils/generators.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Generate a random username
+ */
+export function randomUsername(): string {
+    return faker.internet.userName();
+}
+
+/**
+ * Generate a random URL
```
ibalosh: I think we should remove the extra comments, they do not add value for obvious functions.

e2e/helpers/flows/signup.ts
```diff
@@ -0,0 +1,25 @@
+import {Page} from '@playwright/test';
+import {HomePage} from '../pages/public';
+import {SignUpPage} from '../pages/portal/SignUpPage';
+import {SignUpSuccessPage} from '../pages/portal/SignUpSuccessPage';
+import {randomEmail, randomName} from '../utils/generators';
+
+/**
+ * Complete the portal signup flow and return the test email used
```
ibalosh: I think this comment is not needed, you wrote the function is nice, easy to read way, that describes it perfectly.

e2e/helpers/flows/signup.ts
```diff
@@ -0,0 +1,25 @@
+import {Page} from '@playwright/test';
```
ibalosh, line 1: `nit` good idea for adding `flows`. Is it a plan to keep it pages related? If so we can call it page-flows or something similar to make it more descriptive. If not, `flows` is cool too.

e2e/helpers/pages/portal/SignUpSuccessPage.ts
```diff
@@ -15,4 +15,11 @@ export class SignUpSuccessPage extends PortalPage {
         this.successMessage = this.portalFrame.getByText('To complete signup, click the confirmation link in your inbox');
         this.closeButton = this.portalFrame.getByRole('button', {name: 'Close'});
     }
+
+    /**
+     * Wait for the signup success message to appear
```
ibalosh: this comment should not be needed, if function is not descriptive enough, we can call it:

`waitForSignUpSuccessMessage`

e2e/tests/public/member-signup.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    test.beforeEach(async () => {
+        mailhog = new MailhogClient();
+    });
+
+    test('completes full signup flow with magic link', async ({page}) => {
+        const {email} = await signupViaPortal(page);
+        await mailhog.completeSignupVerification(page, email);
+
+        const homePage = new HomePage(page);
+        await homePage.waitForSignedIn();
```
ibalosh, line 21: This test has no expectation. I think we should have an `expect(..)` that verifies on which page we landed after signup. In that case, waitForSignedIn should not be needed I think, since expectation will wait for element on the page.

we could wait for on homepage: 

```typescript
export class HomePage extends PublicPage {
     this.accountButton = page.getByRole('link', { name: 'Account' })
}

await expect(homePage.accountButton).toBeVisible();
`

File: e2e/helpers/email/MailhogClient.ts (hunk 3 above)
Comment (ibalosh) [reply e2e/helpers/email/MailhogClient.ts:1]: I agree, I would do though some small changes towards making this easier to maintain later:

Add interface for the `EmailClient` so that in tests, we don't need to mention mailhog client that much, but rather emailClient.

```typescript
let emailClient: EmailClient;

test.beforeEach(async () => {
       emailClient = new MailhogClient();
});
```

Also, we should separate out responsibilities that are not of the email client from it. Client should retrieve, search, delete email.

Email client should be in charge of:

* getMessages
* searchForMessages(filterBy: recipient | emailSubject, options: numberOfEmails, searchTimeLimit -> waitFor
*  deleteMessages(options)

Email utils 

* getMessageTextContent(message: EmailMessage)
* getMessageHtmlContent(message: EmailMessage)
* extractLink(message, magicLinkRegexp)

Parts which should not be part of email client:

* waitForMagicLink - since it's a mix of wait and extract, and too specific for email client
* completeSignupVerification - mix of page and wait

It's ok, to show some of the workflow in test:

```
const {emailAddress} = await signupViaPortal(page);
const message = await this.emailClient.searchForMessage(emailAddress, timeoutMs);
const magicLink = this.extracLinkFromMessage(message, magicLinkRegexp);

const signupVerificationPage = new SignupVerficitationPage(page);
await page.goto(magicLink);
expect(signupVerificationPage.loaded()).toBeTrue();
```

e2e/tests/public/member-signup.test.ts
```diff
@@ -0,0 +1,35 @@
...
+        const {email} = await signupViaPortal(page);
+
+        const message = await mailhog.waitForEmail(email);
+        expect(message.Content.Headers.Subject[0].toLowerCase()).toContain('complete');
+
+        const emailBody = mailhog.getPlainTextContent(message);
+        expect(emailBody).toContain('complete the signup process');
+
+        const magicLink = mailhog.extractMagicLink(message);
+        expect(magicLink).toBeTruthy();
```
ibalosh, line 35: we can skip all three checks, by just visiting the extracted magic link? If it works, we verified it. Current verification goes into implementation details with checking the content of the magic link.

ibalosh (changes requested):
Hey @9larsons 

I took a look at this, and I like the approach, I would pretty much take the exact same route as you did.

One thing that I would improve though is the mailhog client, it does bit too much now, it's mixing things.
 
If you have too much stuff on your plate though, let me know, and I can help out on cleanup there, and we can roll this out earlier.

e2e/tests/public/member-signup.test.ts
9larsons: Agreed!

e2e/tests/public/member-signup.test.ts
9larsons: That's fair. I was on the fence w/ this one. I suppose it makes sense to simply visit and make sure it works, though that's covered by a Portal test. All I really twanted to confirm here was that the email was received (looking appropriate helps). Where does that leave us? Do we feel this is unnecessary and should simply have a signup flow test instead?

e2e/helpers/flows/signup.ts
9larsons: I don't have a plan yet :) I think we can keep as-is and as we add more, we can figure that out.

e2e/tests/public/member-signup.test.ts
ibalosh: yeah, magic link checkup is not needed. As long as you already extract and use magic link elsewhere, and you do in sign up, you are ok, no need to test it here.

You can leave though in the part where you verify the content of the email in terms of subject and body , which you do above the link extraction, and maybe just that link extracted is truthy.

ibalosh (approved):
let's add it in, and we can cleanup later client. I would only move:

```
const magicLink = await this.waitForMagicLink(email, timeoutMs);

await page.goto(magicLink);
await page.waitForLoadState('networkidle');
```

out of mailhog client to the test, not to clutter client with playwright

9larsons:
I included most of the above. We'll refactor the MailhogClient in a later PR as our use takes shape.

PR #24897, opened 2025-09-12
Added full end to end test for traffic analytics page hit tracking

cmraible:
ref https://linear.app/ghost/issue/PROD-2687/

This PR adds the first full end to end test for event tracking:
- Visit site's homepage
- Visit Admin > Analytics
- Check that a unique visitor is recorded

It adds two services to the compose file:
- analytics service, to proxy events to Tinybird
- caddy, to route requests from /.ghost/analytics/* to the analytics service

---

> [!NOTE]
> Adds an end-to-end test that records a homepage visit and verifies a unique visitor via a Caddy-proxied analytics endpoint, plus Docker/Tinybird/CI and helper refactors to support it.
> 
> - **E2E/Testing**:
>   - Add test to record a homepage visit and assert 1 unique visitor in `admin/analytics`.
>   - New helpers: `withIsolatedPage`, `PublicPage.enableAnalyticsRequests()` and `waitForPageHitRequest()`; enhanced `AnalyticsOverviewPage` with `UniqueVisitorsGraph` and `refreshData()`.
>   - Add `data-testid` hooks in `apps/shade` KPI card (`kpi-card-header-value`, `kpi-card-header-diff`).
> - **Docker/Compose**:
>   - Add `e2e/compose.yml` with `caddy` and `analytics` services; route `/.ghost/analytics/*` to analytics via new `.docker/caddy/Caddyfile.e2e`.
>   - Switch analytics healthchecks to Node fetch; expose `tinybird-local` port.
> - **E2E Infra**:
>   - Refactor `DockerCompose`: project scoping, `up()`, `waitForAll()`, status parsing, improved logs; update `down()`/`run` commands to include `-p`.
>   - `EnvironmentManager`: use `up()` + `waitForAll()`; clarify teardown.
>   - `GhostManager`: set `tinybird__tracker__endpoint` to `http://localhost:8080/.ghost/analytics/api/v1/page_hit` and `tinybird__tracker__datasource`.
> - **Tinybird CLI**:
>   - Harden token retrieval: single request, error checks, select admin token by ADMIN scope, reuse response.
> - **CI**:
>   - Use `ubuntu-latest` runners; update compose path to `e2e/compose.yml` for image pulls.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1cf1f2412ce0fc21fba625b6939ea7305c3b4849. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


compose.yml
```diff
@@ -182,7 +182,8 @@ services:
     expose:
       - "3000"
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:3000" ]
+      # Simpler: use Node's global fetch (Node 18+)
```
cmraible, line 185: Analytics healthchecks were failing because `curl` isn't installed in the production image that we're shipping anymore — switched to the same check, but using node fetch instead of curl

e2e/tests/admin/analytics/page-hit-tracking.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        // Create a new browser context for visiting the public site
+        const publicContext = await browser.newContext({
+            baseURL: baseURL
+        });
+        const publicPage = await publicContext.newPage();
+
+        // Set the synthetic monitoring flag to bypass test environment detection
+        await publicPage.addInitScript(() => {
+            window.__GHOST_SYNTHETIC_MONITORING__ = true;
+        });
```
cmraible: This is a little awkward. Workaround to avoid `ghost-stats.js` disabling event tracking since it's running in playwright. Without this, the `page_hit` request is not sent.

e2e/tests/admin/analytics/page-hit-tracking.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        // Navigate to the analytics overview page
+        const analyticsOverviewPage = new AnalyticsOverviewPage(page);
+        await analyticsOverviewPage.goto();
+
+        // Assert
+        // Check that the unique visitors graph shows one visitor
+        const uniqueVisitorsGraph = page.getByTestId('Unique visitors');
+        await expect(uniqueVisitorsGraph).toBeVisible();
+
+        await page.reload();
```
cmraible: Reload to force refresh analytics data. Also a bit awkward, but 🤷

e2e/helpers/environment/DockerCompose.ts
```diff
@@ -17,30 +17,96 @@ export class DockerCompose {
...
+            if (!output) {
+                await new Promise((resolve) => {
+                    setTimeout(resolve, intervalMs);
+                });
+                continue;
+            }
+
+            // Parse docker compose ps output (newline-delimited JSON)
+            const containers = output.split('\n')
+                .filter(line => line.trim())
```
janedotx: Nothing's actually getting filtered out here, is it? Might be clearer as another map, or combining this with the map call below, e.g. `.map(line => JSON.parse(line.trim()))`

.docker/caddy/Caddyfile.e2e
```diff
@@ -0,0 +1,18 @@
+# E2E Test Caddyfile - Routes analytics requests to the analytics service
+:80 {
+    log {
+        output stdout
+        format json
+    }
+
+    # Proxy analytics requests with any prefix (e.g. /.ghost/analytics/ or /blog/.ghost/analytics/)
+    @analytics_paths path_regexp analytics_match ^(.*)/\.ghost/analytics(.*)$
```
ibalosh, line 9: `nit:` Just curious, what is the reason for `Caddyfile.e2e` to be in .docker folder at top, while `compose.e2e.yml` is in e2e? Do we have a convention that we use?

My guess is that we use compose close to piece of codebase that we use it in (like compose at root), while docker files, configs, etc, in `.docker`? 

If that's the case, we can probably just call compose -> compose.yml here in e2e?

e2e/tests/admin/analytics/overview.test.ts
```diff
@@ -4,8 +4,40 @@ import {
...
 } from '../../../helpers/pages/admin';
+import {HomePage} from '../../../helpers/pages/public';
 
 test.describe('Ghost Admin - Analytics Overview', () => {
+    test('records visitor when homepage is visited', async ({page, browser, baseURL}) => {
+        // Create a new browser context for visiting the public site
+        // Ghost does not send tracking requests if it detects that it is running in a test environment, i.e. playwright
+        // We can override this by setting window.__GHOST_SYNTHETIC_MONITORING__ to true
+        // This allows us to test page hit tracking in our e2e tests
+        const publicContext = await browser.newContext({
```
ibalosh: the new context part, I would clean this up a bit, and hide implementation details from the test itself. It could be something like this (names probably need some work):

```typescript
// Create a new browser context and page
async function initPageWithNewContext(browser: Browser, baseURL: string) {
    const context = await browser.newContext({baseURL: baseURL});
    const page = await context.newPage();

    return {
        page,
        context
    };
}

// Ghost does not send tracking requests if it detects that it is running in a test environment, i.e. playwright
// We can override this by setting window.__GHOST_SYNTHETIC_MONITORING__ to true
// This allows us to test page hit tracking in our e2e tests
async function bypassTestEnvironmentSettingsForPage(page: Page) {
    await page.addInitScript(() => {
        window.__GHOST_SYNTHETIC_MONITORING__ = true;
    });
}
```

and then just use in test, something like this:

```typescript
const {page: newPage, context: newContext} = await initPageWithNewContext(browser, baseURL);
await bypassTestEnvironmentSettingsForPage(newPage);

const homePage = new HomePage(newPage);
```

This would shorten the test a bit, and make it more readable. For now I just moved functions above in test, but since this is something that we will probably use in other places too, we could also move it to helpers, something like the helpers/playwright folder or similar.

e2e/tests/admin/analytics/overview.test.ts
```diff
@@ -4,8 +4,40 @@ import {
...
+        });
+        const publicPage = await publicContext.newPage();
+        await publicPage.addInitScript(() => {
+            window.__GHOST_SYNTHETIC_MONITORING__ = true;
+        });
+
+        // Visit the homepage to generate a page hit, then close the context
+        const homePage = new HomePage(publicPage);
+        await homePage.goto();
+        await expect(homePage.title).toBeVisible();
```
ibalosh: to avoid needing this expect, we could have within homePage page object something like, waitForTitle, that would do:

waitFor({state: 'visible'}); , or if the purpose here is pageLoad, something like example below, that I added on different pom:

```typescript
async waitForPageLoad() {
    await this.webPerformanceSection.card.waitFor({state: 'visible'});
}
```

e2e/tests/admin/analytics/overview.test.ts
```diff
@@ -4,8 +4,40 @@ import {
...
+        await publicPage.waitForLoadState('networkidle');
```
ibalosh: I keep getting timeout error on this line for some reason, could you share more details why we need to wait for all resources to load here, maybe we could actually make this a helper in pageobject, that would also describe the intention?

Maybe something like: `homePage.waitForPageToFullyLoad` , with description. page from playwright is the one on which we want to apply this, so I think we could do it that way clearer, with responsibilities all being in same place.

e2e/tests/admin/analytics/overview.test.ts
ibalosh: **or we could do this, that seems even cleaner**:

create helper function that will allow running actions in separate context, with separate page:

```typescript
    async function withIsolatedPage<T>(
        browser: Browser,
        opts: Parameters<Browser['newContext']>[0] = {},
        run: (page: Page, context: BrowserContext) => Promise<T>
    ): Promise<T> {
        const context = await browser.newContext(opts);
        const page = await context.newPage();
        try {
            return await run(page, context);
        } finally {
            await context.close(); // guarantees cleanup even on failure
        }
    }
```

and then just use it like this in the test itself:

```typescript
await withIsolatedPage(browser, {baseURL}, async (page) => {
    bypassTestEnvironmentSettingsForPage(page)
    
    const homePage = new HomePage(page);
    await homePage.goto();
    await expect(homePage.title).toBeVisible();
    await page.waitForLoadState('networkidle');
});
```

this makes it clean, isolated, and steps looks still nice

e2e/tests/admin/analytics/overview.test.ts
```diff
@@ -4,8 +4,40 @@ import {
...
+        await publicContext.close();
+
+        // Navigate to the analytics overview page in Ghost Admin
+        // Reload to ensure we fetch the latest data from Tinybird
+        const analyticsOverviewPage = new AnalyticsOverviewPage(page);
+        await analyticsOverviewPage.goto();
+        await page.reload();
```
ibalosh: I think we should keep this in POM too, and call it something like wait for data to load, or something similar more descriptive. That would describe what needs to be done for action on the page.

e2e/tests/admin/analytics/overview.test.ts
```diff
@@ -4,8 +4,40 @@ import {
...
+        await expect(analyticsOverviewPage.header).toBeVisible();
```
ibalosh: for these two we could probably do a wait and hide all these implementation details that way and just have the expect at bottom

e2e/helpers/environment/DockerCompose.ts
```diff
@@ -17,30 +17,96 @@ export class DockerCompose {
...
+     * @param timeoutMs Maximum time to wait for all services to be ready (default: 60000ms)
+     * @param intervalMs Interval between status checks (default: 500ms)
+     *
+    */
+    async waitForAll(timeoutMs = 60000, intervalMs = 500): Promise<void> {
+        const deadline = Date.now() + timeoutMs;
+
+        while (Date.now() < deadline) {
+            // Get the status of all containers in the project in JSON format
+            const cmd = `docker compose -f ${this.composeFilePath} -p ${this.projectName} ps -a --format json`;
```
ibalosh: This could be separated out so it's easier to understand that this returns back details for docker containers. I would also call `cmd` -> `command`, since we have plenty of space per line, it will look nicer. 

I would also extract the whole parsing of statuses for all containers, and return either list of them or continue. Something like, `fetchContainerDetails` ? This would either continue with delay on empty output, or on zero containers.

We would not need to duplicate delay call

e2e/helpers/environment/DockerCompose.ts
```diff
@@ -17,30 +17,96 @@ export class DockerCompose {
...
+            const output = execSync(cmd, {encoding: 'utf-8'}).trim();
+
+            if (!output) {
+                await new Promise((resolve) => {
```
ibalosh: To make it obvious, we could extract this too as a delay function.  All containers ready can be moved out too.

After all extractions, waitForAll could look something like:

```typescript
const containerDetailsJson = retrieveContainerDetailsInJson();
const containers = fethContainers(containerDetailsJson);
            
if (containers.length === 0) {
    delayVerification(intervalMiliSeconds);
    continue;
}

const allContainersReady = areAllContainersRead(containers);
if (allContainersReady) 
  return;
```

Probably bit more than that, but would be much more readable (this all if you are satisfied with compose idea, if this is not premature optimization)

e2e/helpers/environment/DockerCompose.ts
```diff
@@ -17,30 +17,96 @@ export class DockerCompose {
...
+    /** Bring all services up (detached). */
+    up(): void {
         try {
             logging.info('Starting docker compose services...');
-            execSync(`docker compose -f ${this.composeFilePath} up -d`, {stdio: 'inherit'});
-            if (servicesToWait.length > 0) {
-                // NOTE: `docker compose up -d --wait` will fail if one-shot services are included
-                execSync(`docker compose -f ${this.composeFilePath} wait ${servicesToWait.join(' ')}`);
-            }
+            execSync(`docker compose -f ${this.composeFilePath} -p ${this.projectName} up -d`, {stdio: 'inherit'});
```
ibalosh, line 34: I see we use very similar command multiple times, we might want to reuse it, as separate function (like `runCompose(arguments)` , or something, though as long as it works, for now, not that worried about the environment helpers - nice work with these!)

ibalosh (changes requested):
Hey @cmraible , 

In the review I mostly focused on the readability of the tests, and some basic cleanups we could do in the environment helper files. I did not go too deep there, since it's still new and we are trying it out - thought I would clean them up a bit in terms of extracting some of the methods and use better names for variable names.

Overall I like the idea (sad test containers did not work - will love to hear more). 

My main requests for changes is for decoupling more implementation details from tests, to make them less aware of internals (in terms of logic, for refreshing etc). We should try to hide as much as we can from the test itself.

For some reason I could not get the test to pass, I kept getting times. Let's catchup when you are back.

e2e/tests/admin/analytics/page-hit-tracking.test.ts
ibalosh: yeah saw that, it's fine as long as we describe the intention, like in `browser-service.js` , where we say `isTestEnvironment`. Shared details how you can make it clearer below in comments.

e2e/helpers/environment/DockerCompose.ts
9larsons: Imo we could do this in a separate PR. Not opposed to here, though this is focused on simply adding Tinybird/analytics. I do support all your comments/proposed changes here - good ideas!

e2e/helpers/environment/DockerCompose.ts
cmraible: This filter removes any lines from the CLI's output that only includes whitespace - i.e. `docker compose ps -a` outputs an extra newline at the end, and this filters those lines out, before attempting to parse values out of the line in the map.

.docker/caddy/Caddyfile.e2e
cmraible: > My guess is that we use compose close to piece of codebase that we use it in (like compose at root), while docker files, configs, etc, in .docker?

Yeah, we've never formally aligned on this but that's directionally what I've had in mind. Definitely open to suggestions on how to organize the docker files better.

cmraible: > If that's the case, we can probably just call compose -> compose.yml here in e2e?

Yep, happy to rename it!

e2e/tests/admin/analytics/overview.test.ts
cmraible: Awesome suggestions! I've made a few changes here:

- created and used `withIsolatedPage()` helper, agree that this is much cleaner
- rather than creating a separate helper for `bypassTestEnvironmentSettingsForPage()`, I've just added this to the constructor of the `PublicPage` POM, which should be an ancestor to all pages that need analytics events to be enabled.

e2e/tests/admin/analytics/overview.test.ts
cmraible: Once I figured out the `waitForPageHitRequest()` (see below), I found that we didn't need this expect anymore so I've just removed it!

e2e/tests/admin/analytics/overview.test.ts
cmraible: Good shout - what we really need to wait for is the POST request to `/.ghost/analytics/api/v1/page_hit`. I've reworked this a bit by adding a `waitForPageHitRequest()` method to the `PublicPage` POM, and used it here, thus eliminating the title and load state checks here.

e2e/tests/admin/analytics/overview.test.ts
cmraible: Agreed! Added a `refreshData()` method to the POM and used it here.

e2e/tests/admin/analytics/overview.test.ts
cmraible: Yep, turns out the second one wasn't necessary, so it's just `await analyticsOverviewPage.refreshData()` now

e2e/helpers/environment/DockerCompose.ts
cmraible: Agreed! Definitely some unnecessary duplication in this area. I've made some updates, LMK what you think!

e2e/helpers/environment/DockerCompose.ts
cmraible: I've made a few improvements to this area to reduce the duplication and make it (hopefully) read a lot clearer 👌

e2e/helpers/environment/DockerCompose.ts
cmraible: Good call! I'm not 100% sure this is the best long term approach overall, so I'm going to leave this as-is for now — but definitely agree if we don't come up with a cleaner approach.

e2e/tests/admin/analytics/overview.test.ts
ibalosh: gotcha, tx!

e2e/helpers/environment/DockerCompose.ts
ibalosh: looks good, tx!

ibalosh (approved):
that looks great! 👍

PR #24908, opened 2025-09-15
Added OKLAB/OKLCH support for comments dark-mode detection

sam-lord:
closes [ONC-1155](https://linear.app/ghost/issue/ONC-1155/oss-issue-comments-module-fails-to-detect-dark-mode-if-parent-color-is)

Our old detection works great for all colorspaces that are based on RGB (e.g. hex colours, RGBA, RGB, etc), but not on alternative colorspaces that are retained in the computed style. OKLAB & OKLCH use a perceived luminance value that doesn't need the contrast function to work out how contrasty they are against white.

---

> [!NOTE]
> Improves dark-mode detection in `ContentBox` to handle non-RGB computed colors.
> 
> - Adds `colorStringToLuminance` and `rgbToLuminance` helpers and updates `contrast` to work with luminance values
> - Extends detection logic to parse `oklab`/`oklch` colors and use a simple L>0.6 threshold; retains RGB path using sRGB luminance + contrast
> - Expands tests to cover `oklab`/`oklch` cases (including `none` lightness), and verifies class toggling accordingly
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c78e52f9e5ea7043c88d38bfe5bc8cda67aecc04. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/comments-ui/src/components/content-box.tsx
+++ b/apps/comments-ui/src/components/content-box.tsx
@@ -4,25 +4,51 @@ import React, {useCallback, useEffect, useState} from 'react';
 import {ROOT_DIV_ID} from '../utils/constants';
 import {useAppContext} from '../app-context';
 
+const rgbToLuminance = (r: number, g: number, b: number) => {
+    const a = [r, g, b].map(function (v) {
+        v = v / 255; // normalize to 0..1
+        return v <= 0.03928 ? v / 12.92 : Math.pow((v + 0.055) / 1.055, 2.4);
+    });
+
+    return a[0] * 0.2126 + a[1] * 0.7152 + a[2] * 0.0722;
+};
+
+const colorStringToLuminance = (colorString: string) => {
+    if (colorString.startsWith('oklab') || colorString.startsWith('oklch')) {
+        const regexMatches = colorString.match(/ok(?:lab|lch)\(([^ ]+)/);
+        if (!regexMatches || regexMatches.length < 2) {
+            return 0;
+        }
+        const lum = regexMatches[1];
+        if (lum === 'none') {
+            return 0;
+        } else {
+            return parseFloat(lum);
+        }
+    } else if (colorString.startsWith('rgb')) {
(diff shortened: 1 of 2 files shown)


apps/comments-ui/src/components/content-box.tsx
```diff
@@ -8,17 +8,36 @@ type Props = {
...
+                return parseFloat(lum);
+            }
+        } else if (colorString.startsWith('rgb')) {
+            const colorsOnly = colorString.substring(colorString.indexOf('(') + 1, colorString.lastIndexOf(')')).split(/,\s*/);
+            const r = parseInt(colorsOnly[0]);
+            const g = parseInt(colorsOnly[1]);
+            const b = parseInt(colorsOnly[2]);
+
+            const a = [r, g, b].map(function (v) {
+                v /= 255;
```
lsinger: This is rarely used imho so it makes one trip a bit. I'd probably just do this:

```suggestion
                v = v/255;
```

lsinger: `255` here and many, many numbers further down are magic numbers that would be nice to name if reasonably possible.

apps/comments-ui/src/components/content-box.tsx
```diff
@@ -8,17 +8,36 @@ type Props = {
...
+                return v <= 0.03928 ? v / 12.92 : Math.pow((v + 0.055) / 1.055, 2.4);
```
lsinger: This whole formula could be its own named tiny helper function.

apps/comments-ui/src/components/content-box.tsx
```diff
@@ -8,17 +8,36 @@ type Props = {
...
+            });
+            return a[0] * 0.2126 + a[1] * 0.7152 + a[2] * 0.0722;
```
lsinger: Same as above -- could be a named helper, would probably make it easier to read.

lsinger (commented):
Left a few thoughts ahead of further reviews -- all optional.

PR #24912, opened 2025-09-15
🎨 Added auto-open of labs settings when searched

kevinansfield:
ref https://linear.app/ghost/issue/DES-201

Searching for labs settings required an extra click to open the panel which is annoying when it's the only panel shown. Added auto-open/close for the panel when it's the only one shown.

- Extend search service with autoOpenTarget and target registry (register/unregister)
- Compute single visible target from registered groups + current filter
- Register each TopLevelGroup (navid + keywords) for DRY search participation
- Simplify Sidebar to only compute noResult (remove brittle per-section detection)
- Update Labs to react to autoOpenTarget and preserve manual Open/Close with override reset on filter change


kevinansfield:
Closing in favor of https://github.com/TryGhost/Ghost/pull/24910 that has already been merged

PR #24920, opened 2025-09-17
Upgraded Storybook to v9 in Shade

peterzimon:
ref https://linear.app/ghost/issue/PROD-2644/upgrade-storybook-to-v9

A Storybook addon were automatically [updated](https://github.com/TryGhost/Ghost/commit/67298abcbca68b13782963037106e591c4eda5e4) by Renovate, which made Storybook break when ran locally. To overcome this and future compatible issues in Shade we upgraded Storybook to its latest version.

- The upgrade was executed by running the upgrade script `npx storybook@latest upgrade` ([docs](https://storybook.js.org/docs/releases/upgrading)) which updated all storybook components and stories to use the Storybook `react-vite` module instead of `react`.
- Some Storybook addons (essentials, interactions, links, styling, blocks) were incompatible with v9 so they were removed from `package.json`. We'll add new addons in separate PRs if needed.
- `glob` v10+ is ESM-only and doesn't provide a default export so the import in Vite config was updated accordingly.


apps/shade/.storybook/main.tsx
```diff
@@ -1,30 +1,33 @@
...
 			'@': path.resolve(__dirname, '../src'),
 			crypto: require.resolve('rollup-plugin-node-builtins')
 		}
 		return config;
-	},
+	}
 };
 export default config;
+
+function getAbsolutePath(value: string): any {
```
ErisDS: Do you know why this is needed?

apps/shade/.storybook/main.tsx
```diff
@@ -1,30 +1,33 @@
+import { createRequire } from "node:module";
 import type { StorybookConfig } from "@storybook/react-vite";
-import path from 'path';
+import path, { dirname, join } from 'path';
+
+const require = createRequire(import.meta.url);
```
ErisDS: I understand that these changes - the createRequire and dirname stuff are all to do with ESM vs commonJS modules, but it's weird to me this worked before and doesn't now. Not entirely sure why this belongs with this commit or if there's an easier way to resolve these things?

ErisDS (commented):
I've left a couple of questions about what's in the commit, but I don't see anything majorly wrong to block merge

apps/shade/.storybook/main.tsx
peterzimon: Nope, it was added by the upgrade script. I'll take a look if this can be eliminated

peterzimon: I removed `getAbsolutePath`. It was added by the script but was unnecessary in our setup.

apps/shade/.storybook/main.tsx
peterzimon: I assume these appear now because the Storybook dependencies are upgraded (e.g. `glob` is now on v10 in Shade).

ErisDS: What's weirding me out is the `require.resolve` call on line 24 was there (and presumably working) before, and there's no new usage of require - so the additions on line 1 and line 5 don't make sense to me as belonging to this PR. 

There are reasons to make this pure-ESM but it seems like a separate PR?

peterzimon: Yes, you're right. I've removed the require related lines. Everything works as expected. We can get back to making it pure ESM in a separate PR

ErisDS (approved):

PR #24931, opened 2025-09-17
Added Settings to Admin search

9larsons:


ghost/core/core/shared/settings-search-index.js
peterzimon, line 1: The rest of the files are camelCase in the shared folder, I assume we should be consistent with that.

PR #24939, opened 2025-09-18
🐛 Fixed Esc key not closing post preview 

9larsons:
ref https://linear.app/ghost/issue/DES-1208/
  - Fixed ESC key not closing post preview modal when iframe has focus
  - Added event listener with capture phase to intercept ESC before iframe traps it
  - Added @closeModal prop to preview browser component
  - Added E2E tests to verify ESC key works with and without iframe focus

### DIFF
--- a/e2e/helpers/pages/admin/PostEditorPage.ts
+++ b/e2e/helpers/pages/admin/PostEditorPage.ts
@@ -73,7 +73,7 @@ export class PostPreviewModal {
         }
 
         // Wait for script injection to complete (we have a 500ms setTimeout in browser.js)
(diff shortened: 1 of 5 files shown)


9larsons:
Confirmed this works on staging. However, the e2e test is not passing and it may be due to the fact that the preview modal is not accessing the content. I believe the port mapping is incorrect and we need the changes from https://github.com/TryGhost/Ghost/pull/24891 to be merged first. 

I'll work on merging that PR and we'll see if that takes care of it. We could instead click on the header, but it would make the most sense to click into the content as that more closely represents what users would do.

PR #24949, opened 2025-09-22
🐛 Fixed {{date}} helper rendering English for zh, zh-Hant, and pa locales

### LINKED ISSUE #24783: date helper locales don't match i18n locales
### Issue Summary

The docs say "Dates, with month names, are automatically translated. You don’t need to include them in the translation files."   I was pretty sure I'd seen this before, so I was stumped why the Chinese (zh-Hant) site I'm working on wasn't getting translated dates.  

The problem is that the date helper is looking for a locale from https://github.com/moment/moment/tree/develop/src/locale , and we aren't entirely matching that list with our locales: https://github.com/moment/moment/tree/develop/src/locale .

I think zh-Hant and zh are the only problem, currently.

We could: 
1- declare zh-TW and zh-CN to be the Ghost-supported locales, and add the currently used locales as fallbacks in the i18n module.  [We do this for nn/no/nb already.]  Sites that want dates right would need to switch locales
...or..
2- write a workaround in the date helper to translate the zh locales Ghost uses to the locales moment uses.
...or...
3- do 1 but with a migration that automatically changes locales.

I think 3 is not a good choice, because many themes use the locale field to get the html lang displayed also.

Workaround for anyone who needs one: You can pass locale into the date helper directly.

### Steps to Reproduce

set the locale to zh-Hant or zh, and put a {{date}} somewhere in the theme.  Observe the (untranslated) result.

### Ghost Version

6.0.6

### Node.js Version

22

local, but same problem on Ghost Pro.

### Database type

MySQL 8

### Browser & OS version

_No response_

```shell

```

### Code of Conduct

SamarthBagga:
Fixes #24783 
I have implemented the fix suggested in #24783.

---

> [!NOTE]
> Map Chinese i18n locales ('zh', 'zh-Hant') to Moment.js equivalents ('zh-cn', 'zh-tw') in `core/frontend/helpers/date.js` when setting date locale.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e17c2b60ca77394863c34de54651fa7142e29ab2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/core/core/frontend/helpers/date.js
+++ b/ghost/core/core/frontend/helpers/date.js
@@ -46,7 +46,13 @@ module.exports = function (...attrs) {
     // Documentation: http://momentjs.com/docs/#/i18n/
     // Locales: https://github.com/moment/moment/tree/develop/locale
     if (locale && locale.match('^[^/\\\\]*$') !== null) {
-        dateMoment.locale(locale);
+        let momentLocale = locale;
+        if (locale === 'zh-Hant') {
+            momentLocale = 'zh-tw';
+        } else if (locale === 'zh') {
+            momentLocale = 'zh-cn';
+        }
+        dateMoment.locale(momentLocale);
     }
 
     if (timeago) {
--- a/ghost/core/core/frontend/helpers/date.js
+++ b/ghost/core/core/frontend/helpers/date.js
@@ -46,13 +46,20 @@ module.exports = function (...attrs) {
     // Documentation: http://momentjs.com/docs/#/i18n/
     // Locales: https://github.com/moment/moment/tree/develop/locale
     if (locale && locale.match('^[^/\\\\]*$') !== null) {
-        let momentLocale = locale;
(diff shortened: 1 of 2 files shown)


SamarthBagga:
Hi @mike182uk @sam-lord , Could you please review this?

9larsons:
Thanks for the PR! The bug was real and your fix was directionally right. I've merged this with a rework on top of your commit in order to capture the intent + Mike's concerns above.

Instead of mapping zh/zh-Hant explicitly (per-locale logic in the helper was the concern in #24783, the helper now resolves any locale generically via `Intl.Locale(locale).maximize()` and hands moment a candidate list: zh → zh-cn, zh-Hant → zh-tw, and it turned out to fix pa too.

Same outcome you were after and this should fix future gaps as well.

PR #24952, opened 2025-09-22
✨Added Settings to Admin search

9larsons:
ref [BER-2658](https://linear.app/ghost/issue/BER-2658/)

* updated admin search to index settings
* added search endpoint for settings
* passed through query param to Settings to allow opening the app w/ a specific search term

This PR allows users to search specific settings via keywords in Ghost Admin in order to open up the Settings app with the specified keyword. By passing through the keyword, we generally end up opening the Settings app filtered down to one component which is ideal. We don't launch/open the component in this change, though past changes allow for adding that to each component as desired.

NOTE: This PR does create an area of dual maintenance. We maintain the list of keywords in Ghost Admin for this search instead of also using the keywords in the Settings app. This is NOT ideal, but trying to provide these to both was excessive given the current design with the Ember shell. As we're looking to get rid of Ember, this feels like an ok decision in order to simplify this set of changes. Once we have a singular app, we can *much* more easily maintain this in one spot.

### DIFF
--- a/apps/admin-x-settings/src/App.tsx
+++ b/apps/admin-x-settings/src/App.tsx
@@ -13,12 +13,13 @@ interface AppProps {
     officialThemes: OfficialTheme[];
     zapierTemplates: ZapierTemplate[];
     upgradeStatus?: UpgradeStatusType;
+    initialSearchQuery?: string;
 }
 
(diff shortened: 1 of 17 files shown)


9larsons:
Closing as stale — this predates the settings file renames and the React admin shell's search work, so it would need a ground-up redo against the current architecture.

PR #24953, opened 2025-09-22
🐛 Fixed Analytics > Growth tab MRR chart lookback

9larsons:
ref https://linear.app/ghost/issue/ONC-1157/
- wired up lookback parameter to `/stats/mrr` endpoint
- refactored mrr endpoint tests to use helpers, be more readable
- added mrr tests for >90d lookback (default)

When wiring this up, it appears we missed adding the lookback support to this endpoint. It was adapted from the previous Dashboard which always (and could only) look back 90d, so that's all the more we were returning. The endpoint now correctly looks back over the selected timeframe.


9larsons:
When testing this, the data generator works great. Just run `yarn reset:data` and enable Stripe to see the MRR charts.

cmraible (approved):
Nice work! Works for me locally, looking good 👌

PR #24958, opened 2025-09-23
🐛 Fixed RSA key generation to use 2048-bit keys

### LINKED ISSUE #24831: JWKS exposed under /members/.well-known/jwks.json provides invalid key length for Member JWTs
### Issue Summary

When verifying JWT member tokens, the JSON webtoken keyset (JWKS) exposed at `/members/.well-known/jwks.json` does not provide a public key of sufficient length to validate said tokens.

The issued member token header, retrieved from `members/api/session/` looks like this:
```json
{
  "alg": "RS512",
  "typ": "JWT",
  "kid": "redacted"
}
```
According to the official spec, RS512 requires the RSA key modulus length to be 2048 bits or larger. The current key exposed by Ghost’s JWKS endpoint appears to be smaller than this, resulting in the following error when verifying tokens with a compliant library like `jose`:
```
Error: RS512 requires key modulusLength to be 2048 bits or larger
```

### Steps to Reproduce

Grab a member token at `/member/api/session` and JSON web key at `/members/.well-known/jwks.json`. Make sure to use just the JSON Web Key, excluding the array. 

```json
{"kty":"RSA","kid":"2B9asCdRsVD1PorgY7K8KnJ3dF_0FB3v_3qM6pu00Y8","n":"o66I-4QwmERPboZwnhS6exwhZFRohZVOw_AK74C4BbwavIaCl-P0ZObJBnFcl7AWmNhFSRwDdg1L35CwNLMxhnzvmaLgu0qw4XrIFjFnPIV66zCt9HBSZk2fNQ0IN1bomUA7OQxe_Eou0_YNa-l9ZnfQnVnBIChaXALv89kSmaM","e":"AQAB"}
```

Enter into a validation site like `https://www.jwt.io/` and get the following error:

```
RS512 requires key modulusLength to be 2048 bits or larger
```

### Ghost Version

5.*, 6.*

### Node.js Version

22

both locally as well as on servers installed via docker and ghost pro.

### Database type

MySQL 5.7

### Browser & OS version

_No response_

```shell

```

### Code of Conduct

9larsons:
fixes https://github.com/TryGhost/Ghost/issues/24831
- updated ghost and member settings keys to use 2048 bits for RS512 compatibility
- updated on-the-fly member key gen to use 2048 bits
- added tests for covering these areas (previously untested)
- added migration to regenerate existing keys

Ghost's existing keys only used 1024 bits, which do not meet RS512 compatibility. For those using popular libraries like `jose` to verify tokens, they might see errors. 

__NOTE__: Because we're generating new keys with the migration, we'd expect possible disruptions - existing magic links would not work, currently logged-in members need to re-auth, API tokens would need regenerating, etc. We could create an old key with fallback behavior to try to mitigate this, if necessary.


ghost/core/test/e2e-api/members/jwks-well-known.test.js
```diff
@@ -0,0 +1,115 @@
...
+            should.exist(jwks.keys);
+            jwks.keys.should.be.an.Array();
+            jwks.keys.length.should.be.above(0);
+
+            // Validate each key in the JWKS
+            for (const jwk of jwks.keys) {
+                // Check required JWK properties for RSA keys
+                should.exist(jwk.kty);
+                should.exist(jwk.n);
+                should.exist(jwk.e);
```
ErisDS, line 39: Is there a reason this test isn't using snapshots? Is it pure AI smash?

ghost/core/test/e2e-api/members/jwks-well-known.test.js
```diff
@@ -0,0 +1,115 @@
+const crypto = require('crypto');
+const should = require('should');
```
ErisDS, line 2: Should is deprecated, the intention is to use node:assert for all new tests.

ghost/core/core/server/data/migrations/versions/6.0/2025-09-22-00-00-00-regenerate-rsa-keys-2048.js
```diff
@@ -0,0 +1,11 @@
+const {combineTransactionalMigrations, removeSetting} = require('../../utils');
+const logging = require('@tryghost/logging');
+
+module.exports = combineTransactionalMigrations(
+    removeSetting('ghost_private_key'),
+    removeSetting('ghost_public_key'),
+    removeSetting('members_private_key'),
+    removeSetting('members_public_key')
```
ErisDS, line 8: As I understand all of our settings should be _created_ using migrations. That was an agreement that was made several years ago when we reworked the settings table.

Therefore it seems wrong to remove a setting we want and not re-add it.

ErisDS (commented):
There's an ongoing discussion in slack about the high-level approach. Looking at this I noticed some unexpected implementation details, so calling these out separately.

cathysarisky:
I got hit by this issue, so I'm excited to see a fix in the works!

BUT... did I understand right that **API keys would have to be regenerated**?  Does that include the custom integration keys and staff tokens?  If keys are going to get invalidated, that's the sort of thing that I'd hope for a month or more of lead time and a LOT of effort to let people using Ghost know that it's coming.  Admin API keys get used for a TON of different stuff, including subscriber logins, mission-critical integrations, automated newsletter generations, headless setups, static site setups, apps, search integrations, Discord integrations, etc etc etc etc.  And every single Zapier integration out there, many set up by marginally-technical folks who will NOT understand that you're going to break their stuff until you actually break it.

If it's feasible, I'd much prefer a situation where existing API keys and staff tokens continue to work, either until 7.x, or at least with a couple months of lead time.  Please don't break everything I've built this year all at once, please? 🥹 

If it's "just" the jwks but not the custom integrations changing, that'll only break half a dozen things I've built this year.... 

Apologies if I've misunderstood.  You've got me fearing for my sanity here.

PR #24962, opened 2025-09-23
Updated Tinybird utm endpoints to return real data

9larsons:
ref https://linear.app/ghost/issue/PROD-2668/
- updated pipes to parse and use utm data from columns
- updated fixtures to include utm data
- updated tests

The Tinybird `api_top_utm_*` endpoints were created in a past commit to return fixture data in order to be able to wire up the frontend. We've been collecting data in the background in the `analytics_events` table and are ready to parse and pass this through to the endpoints instead of using fixtures.

This is a relatively simple change, with the exception of changes to `mv_session_data` that were a bit unexpected. The aggregate functions here seem to be incorrect or at least unnecessary, and adding the `utm_*` columns broke local builds, exposing this trouble. While I'm not entirely clear on the behavior, it seems previously we got away with using `finalizeAggregation()` in the few places we did because of fallback behavior in Clickhouse.

finalizeAggregation(countState()) creates state, then finalizes it. count() is direct aggregation. Count should be both simpler and more efficient. All existing tests passed, suggesting that we shouldn't have an issue, though we'll need to test with staging to be sure of the deployments workflow.


ghost/core/core/server/data/tinybird/pipes/mv_session_data.pipe
```diff
@@ -4,10 +4,15 @@ SQL >
...
-        argMin(source, timestamp) as source
+        count() as pageviews,
+        min(timestamp) as first_pageview,
+        max(timestamp) as last_pageview,
+        argMin(source, timestamp) as source,
+        argMin(utm_source, timestamp) as utm_source,
+        argMin(utm_medium, timestamp) as utm_medium,
+        argMin(utm_campaign, timestamp) as utm_campaign,
+        argMin(utm_term, timestamp) as utm_term,
+        argMin(utm_content, timestamp) as utm_content
```
cmraible, line 15: No action needed here, just thinking out loud:

I think this makes sense - we must have added the `finalizeAggregation(countState())` when this was actually a pipe feeding a materialized view. We'd need the `finalizeAggregation` function to convert the `countState` into the final value if using a materialized view with `AggregatingMergeTree`. But since this is just a pipe, Clickhouse just runs the full aggregation on each query, and we don't have to worry about state/finalizing the aggregations. If anything, this should probably lead to a slight performance improvement.

It's a little confusing that this is called `mv_session_data` even though it's just a regular pipe, but I don't think it's worth renaming it right now.

ghost/core/core/server/data/tinybird/datasources/_mv_hits.datasource
```diff
@@ -16,8 +14,13 @@ SCHEMA >
...
+    `utm_source` String,
+    `utm_medium` String,
+    `utm_campaign` String,
+    `utm_term` String,
+    `utm_content` String
 
 ENGINE "MergeTree"
 ENGINE_PARTITION_KEY "toYYYYMM(timestamp)"
-ENGINE_SORTING_KEY "site_uuid, timestamp, session_id"
+ENGINE_SORTING_KEY "site_uuid, timestamp, session_id"
```
cmraible, line 26: Thinking out loud: why don't we need the FORWARD QUERY here? Does Tinybird just re-materialize the view from scratch instead?

cmraible: Ah, from the deployment check step:

```
 Data that will be copied with this deployment:
------------------------------------------------
| Datasource | Backfill type                   |
------------------------------------------------
| _mv_hits   | Using Materialized Pipe mv_hits |
------------------------------------------------
```

So it looks like the answer is yes: since it's a datasource for an MV, it just populates the data from the pipe rather than copying it from the existing MV.

cmraible: This is actually really good to know, since it means we can continue evolving the `_mv_hits.datasource` freely without running into the same issue we've found with the main datasource.

cmraible (approved):
Nice work, LGTM!

ghost/core/core/server/data/tinybird/datasources/_mv_hits.datasource
9larsons: Yep! I should've called that out. It does seem FORWARD QUERY only is for landing data sources.

ghost/core/core/server/data/tinybird/pipes/mv_session_data.pipe
9larsons: Ditto!

PR #24964, opened 2025-09-24
Allowed proxied domains for app preview servers

kevinansfield:
no issue

- added `allowedHosts: true` so any domain can be proxied to the preview server
  - previously if you tried to proxy a domain to one of the preview servers it would error and ask you to add the specific domain in the `allowedHosts` config
  - allows for dev setups using HTTPS and real domains to satisfy Safari's stricter security requirements
- switched from `.js -> .mjs` and `.ts -> .mts` to resolve CJS deprecation warnings on dev startup

---

Example setup:
- Ghost proxy: https://dev.mydomain.com -> http://myip:2368
  - config: `portal.url = https://portal-dev.mydomain.com/portal.min.js`
- Portal proxy: https://portal-dev.mydomain.com -> http://myip:4175

Using `yarn dev --portal` doesn't work in general for Safari if your main Ghost site is HTTPS as it forces the portal URL to http://localhost:4175 so Safari won't load it with a "Mixed content" error (it doesn't have a built-in localhost bypass like Chrome).

An alternative is a local Caddy proxy of `https://localhost:4176` to `http://localhost:4175` and using `yarn dev --portal --https` but then it's harder to test split front-end/admin domains locally or use your dev server across devices for mobile testing.

### DIFF
--- a/apps/announcement-bar/vite.config.mjs
+++ b/apps/announcement-bar/vite.config.mjs
@@ -1,14 +1,12 @@
+/* eslint-env node */
 import {resolve} from 'path';
 import fs from 'fs/promises';
 
 import {defineConfig} from 'vitest/config';
-import cssInjectedByJsPlugin from 'vite-plugin-css-injected-by-js';
 import reactPlugin from '@vitejs/plugin-react';
 import svgrPlugin from 'vite-plugin-svgr';
 
 import pkg from './package.json';
-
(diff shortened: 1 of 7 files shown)


kevinansfield:
@9larsons added you for review as you're on-call and it's not clear who owns this area. Not much to look at here though, I guess main thing is trying to think of any reason why we _shouldn't_ do this?

PR #24972, opened 2025-09-24
E2e cleanup

ibalosh:
Cleaned up e2e suite little bit with some simple changes, to better align with naming conventions for better consistency.

* align test stories, to [not use should](https://github.com/spotify/should-up) and match other tests
* change expectations to use web first assertions, as per [Playwright best practices guide](https://playwright.dev/docs/best-practices)
* cleanup mailhog client


cmraible (approved):
Thanks, these changes all look good to me!

Only comments I have are around maintaining these conventions — I wonder how much of this we could enforce with linting rules?

- Filenames in kebab case: we use `[eslint-plugin-filenames](https://github.com/selaux/eslint-plugin-filenames)` in a few places (see `ghost/core/eslintrc.js`) to enforce certain patterns.

> align test stories, to [not use should](https://github.com/spotify/should-up) and match other tests

TBH I've never thought about this, but agree that it reads cleaner without all the `shoulds`. Should we document this convention in the README.md, and/or add a linting rule (if possible) for it?

9larsons (approved):
I'm on board w/ all the changes. Maybe with the exception of classes like AdminPage being admin-page - this is a convention I'm ok with, though I'll have to get used to it. We've used (inconsistently) the pattern of classes being PascalCase while utils and other files are kebab-case.

ibalosh:
> I'm on board w/ all the changes. Maybe with the exception of classes like AdminPage being admin-page - this is a convention I'm ok with, though I'll have to get used to it. We've used (inconsistently) the pattern of classes being PascalCase while utils and other files are kebab-case.

@9larsons yeah, the kebab-cases on classes, I am not sure with yet either. We caught up with Hannah on that, but I will bring it up on eng call to see consensus on that one. (I might keep that one out of cleanup until I get more info on it, whether that rule is extended)

ibalosh:
> Thanks, these changes all look good to me!
> 
> Only comments I have are around maintaining these conventions — I wonder how much of this we could enforce with linting rules?
> 
> * Filenames in kebab case: we use `[eslint-plugin-filenames](https://github.com/selaux/eslint-plugin-filenames)` in a few places (see `ghost/core/eslintrc.js`) to enforce certain patterns.
> 
> > align test stories, to [not use should](https://github.com/spotify/should-up) and match other tests
> 
> TBH I've never thought about this, but agree that it reads cleaner without all the `shoulds`. Should we document this convention in the README.md, and/or add a linting rule (if possible) for it?

hey @cmraible I will see to update readme, and check out linting rule (check above comment I shared with Steve about that one, I am still not 100% sure about the kebab-case on which files)

PR #24974, opened 2025-09-24
Fixed ghost_head rss feed link with multiple collections

### LINKED ISSUE #24963: ghost_head rss feed generation is wrong - missing href if there are multiple collections.
### Issue Summary

My self-hosted site and a client site on Ghost Pro both have the same issue.  We're getting the following output from ghost_head:

`<link rel="alternate" type="application/rss+xml" title="Actual Site Name" href="">`

The href attribute is blank. 

It appears to be caused by having multiple collections, or not having a collection at /?  Here are examples that work/don't work.  When the feed href is missing, it seems to be sitewide.  Actual rss feeds exist, just the link is broken.

BROKEN example:
```
routes:
collections:
  /news/:
    permalink: /{slug}/
    template: index
    filter: tag:[news]
  /fooz/:
    permalink: /{slug}/
    template: index
taxonomies:
  tag: /tag/{slug}/
  author: /author/{slug}/
```
BROKEN example:
```
routes:
  /: 
    template: index
    controller: channel
    filter: tag:-[news]
collections:
  /news/:
    permalink: /{slug}/
    template: index
    filter: tag:[news]
  /fooz/:
    permalink: /{slug}/
    template: index
taxonomies:
  tag: /tag/{slug}/
  author: /author/{slug}/
```
WORKING example:
```
routes:
  /: 
    template: index
    controller: channel
    filter: tag:-[news]
collections:
  /news/:
    permalink: /{slug}/
    template: index
    filter: tag:[news]
taxonomies:
  tag: /tag/{slug}/
  author: /author/{slug}/
```
WORKING example:
```
routes:
collections:
  /news/:
    permalink: /{slug}/
    template: index
    filter: tag:[news]
  /:
    permalink: /{slug}/
    template: index
    filter: tag:-[news]
taxonomies:
  tag: /tag/{slug}/
  author: /author/{slug}/
```

### Steps to Reproduce

Upload the routes.yaml to a Ghost site, then check the page source on any page (post, index, doesn't matter)

### Ghost Version

6.x

### Node.js Version

ghost pro

ghost pro

### Database type

MySQL 8

### Browser & OS version

_No response_

```shell
none - routes file uploads without errors
```

### Code of Conduct

9larsons:
ref https://linear.app/ghost/issue/ONC-1165/
fixes https://github.com/TryGhost/Ghost/issues/24963
- ghost_head falls back to first listed collection when there's multiple, non-index collections in routes.yaml

Previously, ghost_head would generate an empty href="" attribute in the RSS meta tag when routes.yaml contained multiple collections but none at the root path (/), causing broken feed discovery.

Logic for the rss link tag in ghost_head is now:
- Index collection (/) if it exists and has RSS enabled
- Single collection's RSS if only one collection exists
- First collection with RSS enabled when multiple collections exist
- null if no collections have RSS enabled

---

> [!NOTE]
> Adjusts primary RSS URL resolution and improves router lifecycle handling.
> 
> - Update `getRssUrl` to handle multiple `CollectionRouter`s without an `index`, returning the first with RSS enabled; existing behavior for `index` and single collection remains
> - Harden `resetAllRouters` by guarding `value.reset()` calls and clearing `routers`; add `clearAllRouters` helper
> - Add unit tests for multi-collection cases and test setup/teardown using the new clear/reset utilities
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 679809641f85f2da914d30ca7c2004cd522ced7f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ghost/core/test/unit/frontend/services/routing/registry.test.js
```diff
@@ -47,5 +54,53 @@ describe('UNIT: services/routing/registry', function () {
...
+            registry.setRouter('CollectionRouter-podcast', {
+                name: 'CollectionRouter',
+                routerName: 'podcast',
+                getRssUrl: sinon.stub().returns('/podcast/rss/')
+            });
+
+            registry.getRssUrl().should.eql('/podcast/rss/');
+        });
+
+        it('multiple collections without index, all have RSS disabled', function () {
```
cathysarisky, line 90: What if there is a channel?  (I apologize, I'm not sure if channels are collections under the hood...)

cathysarisky (commented):
This is the first time I've looked at this code, so please bear with me if I'm missing something, but I don't think this currently handles a case where the index route (at /)  is actually a channel, not a collection.  In that case, it'd be ideal to return the rss for the channel.

ghost/core/test/unit/frontend/services/routing/registry.test.js
9larsons: They are handled differently, though it's worth calling out, thanks! I'm also new to this section of the code base.

9larsons:
To my understanding, `/` is not intentioned to be a channel. Channels are supposed to be a division of content, while index is generally everything.

Looking through this:
- Channel RSS feeds are always available (by default) at `/channel-name/rss/`
- Channel RSS feeds have never been discoverable in `ghost_head`
- If you have index set up as a channel, `ghost_head` would continue to have an empty or wrong url

I'm not particularly concerned with the last, because we've drawn a line that we don't make channel RSS feeds discoverable. Whether that's the correct decision is more of a product decision than a 'bug' like the issue you raised leading to this change - where we always intended to provide a discoverable RSS feed URL in `ghost_head` but had unhandled cases.

I had adjusted this to handle channels, though I think it's best to not handle them and leave as-is.

cathysarisky (approved):
Thanks for patching my bug! :)  This looks good to me.

I'll maybe drop a note in the docs for channels & routes about what the current behavior is.  I wouldn't have known that having a channel as the index was a problem from the docs.

ngeorger:
I've added a new comment on the original issue, maybe it could be useful! Thanks

Link: https://github.com/TryGhost/Ghost/issues/24963#issuecomment-3483237259

cathysarisky:
@ngeorger I don't think this patch quite fixes the case you added to the issue, since you've got a channel at /.  I left you a theme-level fix.

@9larsons , see! I'm not the only one who wants to put a channel at / ! 😀

ngeorger:
How we could know if this PR is going to be accepted/merged? Thanks a lot!

9larsons:
Sorry for the delays here. There has been a lot of higher priorities than this fix.

ngeorger:
> Sorry for the delays here. There has been a lot of higher priorities than this fix.

Thanks a lot!

PR #24975, opened 2025-09-24
Improved loading state for sources card

peterzimon:
ref https://linear.app/ghost/issue/PROD-2667/polish-table-loading-state-to-only-focus-on-data-changing

- Moved the loading state indicators in the contents area of the card so there are less happening visually when the card is loading.


9larsons (approved):
Definitely improved, thanks! 👍

PR #24976, opened 2025-09-24
Added Tinybird local development integration

9larsons:
no ref
- Add `yarn dev:tinybird` command for automatic Tinybird setup
- Fix analytics service port mapping (expose port 3000)
- Add docker:dev:analytics commands for container management
- Auto-extract Tinybird config from Docker volumes to environment

So far, our 'no config' setup for doing development with analytics (Tinybird + analytics service) required running everything in Docker, often via `docker compose --profile analytics --profile split up -d`. This change allows a simple command - `yarn dev:tinybird` - that spins up the analytics service, does the tinybird setup, and pushes the relevant configuration into env variables used to run `yarn dev` with a standard local setup. This prevents coordinating all of Docker, which can be burdensome at times, and is a simpler setup when not needing to coordinate changes outside of Ghost.


cmraible (approved):
I haven't manually tested this, but no particular objection to the idea here if it's working well for you!

PR #24981, opened 2025-09-24
Added OTC support for custom sign-in forms

kevinansfield:
ref https://linear.app/ghost/issue/PROD-2573

- adds One-Time Code (OTC) support for custom sign-in forms embedded in Ghost themes. When a custom form is configured with `data-members-otc="true"` the Portal popup will open automatically to collect the OTC after the magic link email is sent

### DIFF
--- a/apps/portal/src/App.js
+++ b/apps/portal/src/App.js
@@ -108,7 +108,7 @@ export default class App extends React.Component {
                 site: contextState.site,
                 member: contextState.member,
                 labs: contextState.labs,
-                onAction: contextState.onAction
+                doAction: contextState.doAction
             });
         }
     }
--- a/apps/portal/src/App.js
+++ b/apps/portal/src/App.js
@@ -106,7 +106,9 @@ export default class App extends React.Component {
             handleDataAttributes({
                 siteUrl,
                 site: contextState.site,
-                member: contextState.member
+                member: contextState.member,
+                labs: contextState.labs,
+                onAction: contextState.onAction
             });
         }
     }
--- a/apps/portal/src/App.js
+++ b/apps/portal/src/App.js
@@ -108,7 +108,8 @@ export default class App extends React.Component {
                 site: contextState.site,
                 member: contextState.member,
                 labs: contextState.labs,
-                doAction: contextState.doAction
+                doAction: contextState.doAction,
+                captureException: Sentry.captureException
             });
         }
     }
--- a/apps/portal/src/data-attributes.js
+++ b/apps/portal/src/data-attributes.js
@@ -16,7 +16,7 @@ function handleError(error, form, errorEl, t) {
 }
 
 export async function formSubmitHandler(
-    {event, form, errorEl, siteUrl, submitHandler},
+    {event, form, errorEl, siteUrl, submitHandler, labs = {}, onAction},
     t = str => str
 ) {
     form.removeEventListener('submit', submitHandler);
@@ -48,6 +48,8 @@ export async function formSubmitHandler(
         emailType = form.dataset.membersForm;
     }
 
+    const wantsOTC = emailType === 'signin' && form?.dataset?.membersOtc === 'true' && labs?.membersSigninOTC;
+
     form.classList.add('loading');
     const urlHistory = getUrlHistory();
     const reqBody = {
@@ -57,6 +59,9 @@ export async function formSubmitHandler(
         name: name,
         autoRedirect: (autoRedirect === 'true')
     };
+    if (wantsOTC) {
+        reqBody.includeOTC = true;
+    }
     if (urlHistory) {
         reqBody.urlHistory = urlHistory;
     }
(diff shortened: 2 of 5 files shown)


apps/portal/src/data-attributes.js
```diff
@@ -85,7 +90,28 @@ export async function formSubmitHandler(
         form.addEventListener('submit', submitHandler);
         form.classList.remove('loading');
         if (magicLinkRes.ok) {
+            let responseBody;
+            if (wantsOTC) {
+                try {
+                    responseBody = await magicLinkRes.clone().json();
```
troyciesco, line 98: question: is clone necessary here? I don't think I've ever used it, but from [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Response/clone#:~:text=The%20clone()%20method%20of,stored%20in%20a%20different%20variable.) it doesn't seem like this is the use case for it

apps/portal/src/data-attributes.js
```diff
@@ -85,7 +90,28 @@ export async function formSubmitHandler(
...
+                } catch (e) {
+                    responseBody = undefined;
+                }
+            }
+
             form.classList.add('success');
```
troyciesco: question/suggestion: does it make sense for this to always be "success" and for it to come in between these two blocks of logic? 

If we always want this process to be "successful" if `magicLinkRes.ok`, regardless of the outcome of trying to set up the OTC popup, it might make more sense to put it at the beginning or end to keep related logic together

(note: i don't have a sample custom form set up yet to try out different placement etc, this is just observation from reading the code)

apps/portal/src/data-attributes.js
```diff
@@ -85,7 +90,28 @@ export async function formSubmitHandler(
...
+            const otcRef = responseBody?.otc_ref;
+            if (otcRef && typeof onAction === 'function') {
+                try {
+                    await onAction('startSigninOTCFromCustomForm', {
+                        email: (email || '').trim(),
+                        otcRef
+                    });
+                } catch (actionError) {
+                    // eslint-disable-next-line no-console
+                    console.error(actionError);
```
troyciesco, line 113: console error might indeed make the most sense here since (i'm guessing) it's not fatal if the OTC popup doesn't happen properly, just noting it in case this is something to be cleaned up

apps/portal/src/App.js
```diff
@@ -106,7 +106,9 @@ export default class App extends React.Component {
             handleDataAttributes({
                 siteUrl,
                 site: contextState.site,
-                member: contextState.member
+                member: contextState.member,
+                labs: contextState.labs,
```
troyciesco, line 110: thought: i wonder if it'd be better to pass the specific flag here instead of a labs object - that way cleaning up the flag later might be a little bit easier since you don't have to trace labs all the way back to its origin, you can just find/remove membersSigninOTC

apps/portal/src/data-attributes.js
kevinansfield: Codex says:

> Response bodies are stream-like and can be consumed only once. In formSubmitHandler we always hand the same magicLinkRes back to our existing logic so it can keep working for non-OTC flows where we don’t need the JSON. Calling json() directly would consume the body, meaning anything else that tried to read it later would fail. Using magicLinkRes.clone().json() keeps the original response untouched (the clone copies the stream) so we can safely read the JSON when OTC is enabled without breaking the legacy path.

TIL too 😄 We potentially call `magicLinkRes.json()` later on in `await HumanReadableError.fromApiResponse(magicLinkRes)` so cloning means that will continue to work.

apps/portal/src/data-attributes.js
kevinansfield: yeah, we'd at least want to log to Sentry too so we have some observability into errors 👀

apps/portal/src/App.js
kevinansfield: I quite like having the labs object passed through as it matches how we use labs elsewhere. It seems cleanup would be simpler that way too as it's just removing `labs?.flag` from any conditionals rather than needing to also remove it from any pass-through args, or do you see that differently? I suppose one issue would be dealing with the `labs` argument not being used somewhere on the outer leaf functions and the linter complaining 🤔

No strong opinion either way though

apps/portal/src/data-attributes.js
troyciesco: Ahhhhh there it is, I didn't notice the second usage down there lol makes sense then

apps/portal/src/App.js
troyciesco: We can certainly leave it for now - I guess another benefit of passing the whole object is that when you see it used farther down the code it's clear that it's a flag and not some other variable. Though imho i wouldn't want to leave `labs` as part of the signature for `handleDataAttributes` and `formSubmitHandler` after this feature is GA (if i'm understanding you correctly).

Maybe something to chat about in an engineering call at some point? I'll look around Monday too to see if an agreed-upon approach is documented somewhere.

kevinansfield: > Though imho i wouldn't want to leave labs as part of the signature for handleDataAttributes and formSubmitHandler after this feature is GA (if i'm understanding you correctly)

You were understanding me correctly. Throughout the codebase we pass `labs` through as a DI arg on places where we can't (or don't want to) require the `labs` service directly. This is one of those places and I think it makes sense to keep it in place once the flag is cleaned up. The alternative of removing means we end up in a situation we've seen multiple times where flag/labs passthrough gets re-implemented each time (usually differently) or duplicated because it wasn't clear how it was already done (usually where we've extracted some value off of labs and used that as the passthrough instead of being explicit that we're using our labs feature).

Labs flags are a central part of our development process, it seems wasteful/confusing to re-implement how we access them every time we want to use one in a particular area.

apps/portal/src/data-attributes.js
kevinansfield: The success state here is for the original form on the page, where "success" is the magic link email being sent which will occur and hit this code whether or not OTC was requested. With that context I think it makes sense to keep it, although I agree moving to the top of the conditional block is probably better. Even if the OTC popup was supposed to open but failed (which should only occur if there was a code bug) the email was still sent with a working magic link in it and therefore "successful".

Agree that moving it to the top of the conditional block is better though 👍

troyciesco (approved):

PR #24984, opened 2025-09-25
Added to containerized environment local version of portal

ibalosh:
The problem arises from the e2e tests using the “published” version of portal rather than the local version, which results in a race condition because the version specified in defaults.json isn’t published yet. With this update, local version of portal is used by tests.

Related to same fix introduced in [legacy test suite](https://github.com/TryGhost/Ghost/pull/23527/files).


ibalosh:
this will still use an image, even if it passes, need to change that

ibalosh:
Experimented with this, locally it's working, but had no luck in CI, @cmraible wondering if this is the approach to this, or we should approach it differently. (used first docker image, for portal, but that would not use latest version). I don't know that much about how we use these in tests yet.

9larsons:
I'll take a crack at this. Essentially, this set of changes is running the Portal process locally, so we can't expect that to work on CI. What we'll need to do is create a container for Portal... and likely do something similar for all of the frontend apps (for the moment), but Portal is the most important by far. As we move towards removing Ember, it's possible we can spin up a (single!) dev server for all of the frontend apps to run against.

We could also serve the built Portal file from Ghost, but Ghost isn't really intended to work that way (as we load from CDN). I'm leaning towards serving Portal in - yet another - container, as that more closely mimics how it works in reality.

ibalosh:
> I'll take a crack at this. Essentially, this set of changes is running the Portal process locally, so we can't expect that to work on CI. What we'll need to do is create a container for Portal... and likely do something similar for all of the frontend apps (for the moment), but Portal is the most important by far. As we move towards removing Ember, it's possible we can spin up a (single!) dev server for all of the frontend apps to run against.
> 
> We could also serve the built Portal file from Ghost, but Ghost isn't really intended to work that way (as we load from CDN). I'm leaning towards serving Portal in - yet another - container, as that more closely mimics how it works in reality.

thanks @9larsons ! 

ah, gotcha, yeah container is the right choice, let me know how it goes, curious of this one for other services. 

(I tried first solution with container, but just with full ghost published repo, and running on it portal as entry point.. but thats too big image for portal, and I am guessing it's using old ghost codebase - already published one)

9larsons:
Give this a try - it seemed to work locally with some changes. Note that the portal container has to be rebuilt before running tests. I bet we can find a better way to handle this.

e2e/helpers/environment/EnvironmentManager.ts
```diff
@@ -55,7 +55,29 @@ export class EnvironmentManager {
...
+            const portMapping = containerInfo.NetworkSettings.Ports['4175/tcp'];
+            if (!portMapping || portMapping.length === 0) {
+                throw new Error('Portal container does not have port 4175 exposed');
+            }
+            const hostPort = portMapping[0].HostPort;
+            const portalUrl = `http://localhost:${hostPort}/portal.min.js`;
+            debug(`Portal is available at: ${portalUrl}`);
+            return portalUrl;
+        } catch (error) {
+            logging.error('Failed to get Portal URL:', error);
```
cmraible, line 73: I think we should just throw an error here, rather than falling back. Port 4175 will almost never work, so better to fail fast here IMO

e2e/helpers/environment/GhostManager.ts
```diff
@@ -69,9 +70,10 @@ export class GhostManager {
                 tinybird__stats__endpointBrowser: 'http://localhost:7181',
                 tinybird__tracker__endpoint: 'http://localhost/.ghost/analytics/api/v1/page_hit',
                 tinybird__workspaceId: tinybirdState.workspaceId,
-                tinybird__adminToken: tinybirdState.adminToken
+                tinybird__adminToken: tinybirdState.adminToken,
+                portal__url: config.portalUrl || 'http://localhost:4175/portal.min.js'
```
cmraible, line 74: Same here, I don't think it makes sense to fallback here

cmraible:
Weird that this is working locally but not in CI 🤔 Looks like we lost our artifact uploads too, which makes it harder to debug in CI...

9larsons:
Closing in lieu of https://github.com/TryGhost/Ghost/pull/25042.

PR #24998, opened 2025-09-29
Added initial conventions file for Claude to make generating new test…

ibalosh:
This is the initial version of claude conventions, we can further improve as we go, to make new tests and helpers that are generated more consistent.

Added it so we can use it to generate tests faster. We will improve it as we go. 


e2e/CLAUDE.md
```diff
@@ -0,0 +1 @@
+AGENTS.md
```
cmraible, line 1: I'm not sure if this will actually impact performance at all, but you can directly "import" other files into CLAUDE.md files using `@`
```suggestion
@./AGENTS.md
```

cmraible (approved):
Nice, LGTM! Have you had much luck with this setup?

e2e/CLAUDE.md
ibalosh: claude says that seems like a better solution, tx!

ibalosh:
> Nice, LGTM! Have you had much luck with this setup?

it does create tests and updates page objects to look like current ones. It might need bit more work for playwright mcp and detecting components in React to update with locator, but we can do that as we experiment more.

Would love to hear @9larsons thoughts on that, he got that working with a bugfix on team week.

PR #25000, opened 2025-09-29
Added `autoComplete` prop/field passthrough in Portal

kevinansfield:
no issue

- `autoComplete='off'` was used in our honeypot fields but it was never applied because the `<InputForm>` and `<InputField>` components didn't pass/accept an `autoComplete` prop

### DIFF
--- a/apps/portal/src/components/common/InputField.js
+++ b/apps/portal/src/components/common/InputField.js
@@ -97,6 +97,7 @@ function InputField({
     tabIndex,
     maxLength,
     autoFocus,
+    autoComplete = '',
     errorMessage
 }) {
     const fieldNode = useRef(null);
@@ -113,7 +114,6 @@ function InputField({
         disabled = true;
     }
 
-    let autoComplete = '';
     let autoCorrect = '';
     let autoCapitalize = '';
     let inputMode;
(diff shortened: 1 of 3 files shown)


apps/portal/src/components/common/InputField.js
```diff
@@ -113,25 +114,24 @@ function InputField({
...
+    let autoCorrect = '';
+    let autoCapitalize = '';
     let inputMode;
     let pattern;
     switch (id) {
     case 'input-email':
-        autocomplete = 'off';
-        autocorrect = 'off';
-        autocapitalize = 'off';
+        autoComplete = 'off';
```
troyciesco, line 123: i think the [cursor bot has a point here](https://github.com/TryGhost/Ghost/pull/25000#pullrequestreview-3280576794) in that adding the `autoComplete` prop but always overwriting it for a few input types feels like it's introducing a point of friction worth addressing in this PR. maybe the autoCompletes in this switch should still use the prop if it's passed?

kevinansfield:
@codex rebase this on upstream/main and examine whether the change is still necessary. Propose any improvements that would address [this comment](https://github.com/TryGhost/Ghost/pull/25000#pullrequestreview-3281754380)

kevinansfield:
@codex rebase this on upstream/main and examine whether the change is still necessary. Propose any improvements that would address https://github.com/TryGhost/Ghost/pull/25000#pullrequestreview-3281754380

PR #25004, opened 2025-09-29
Patched input-otp to focus properly in an iframe

troyciesco:
ref https://linear.app/ghost/issue/PROD-2537 and https://linear.app/ghost/issue/PROD-2676

-  `input-otp` isn’t iframe-aware. It binds to the top-level document, but the OTC input lives in the iframe’s document. As a result, caret/selection tracking and focus checks don’t fire/update, so after the last char the library can’t keep an “active” slot and arrow-key navigation appears broken.
- This PR introduces `patch-package` and patches `input-otp` so that event listeners are properly scoped to the iframe and the OTC input has arrow key functionality and proper focus


troyciesco:
closing in favor of https://github.com/TryGhost/Ghost/pull/25040

PR #25010, opened 2025-09-30
Released @tryghost/portal v2.54.0

kevinansfield:
Changelog for v2.53.3 -> 2.54.0:
  - [Added OTC support for custom forms](https://github.com/TryGhost/Ghost/commit/1ba4f77548)

### DIFF
--- a/apps/portal/package.json
+++ b/apps/portal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/portal",
-  "version": "2.53.3",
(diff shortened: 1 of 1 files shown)


kevinansfield:
Note: The docker e2e tests are currently suffering from a race condition due to how our app shipping process works - if the version is bumped in `defaults.js` they expect that version to be published, but our process means we don't publish until it's merged to main. Therefore the `defaults.js` bump will happen in a separate PR later.

PR #25014, opened 2025-09-30
🎨 Hide What's new banner during onboarding

rob-ghost:
## Why this change?

When setting up a brand new Ghost site, the "What's new" featured banner appears in the sidebar during onboarding. This creates a confusing user experience because:
- Everything is new on a brand new site
- The banner distracts from the onboarding checklist
- Users haven't experienced the baseline yet, so "what's new" has no context

Ref https://linear.app/ghost/issue/DES-1155/hide-whats-new-for-brand-new-sites

## What does this change do?

Hides the "What's new" featured banner while the onboarding checklist is active. Once a user completes or dismisses onboarding, the banner will appear normally if there are new featured updates.

**Technical implementation:**

The `whatsNew` service now encapsulates the concern of deciding if we should show the notification or not, instead of the component itself checking the various conditions on when the notification should be shown.

`whatsNew` then uses the `onboarding` service to check if the Site is currently going through the onboarding checklist or not.

## Why is this needed?

Creates a cleaner, more focused onboarding experience without competing notifications.

## Implementation approach

This PR uses a 3-stage commit strategy:
1. **Baseline tests** - Validates existing behaviour before changes
2. **Refactor** - Moves logic to service layer (no behaviour change)
3. **Feature** - Adds onboarding check with new tests

---

> [!NOTE]
> Hides the "What's new" sidebar banner while onboarding is active by moving visibility logic to `whats-new` service, adding test selectors, and introducing unit/integration tests.
> 
> - **Admin UI**
>   - **Banner visibility**: `gh-nav-menu/footer-banner.js` now uses `whatsNew.shouldShowFeaturedBanner` instead of `hasNewFeatured` to show the banner.
>   - **Test selectors**: Add `data-test-*` attributes to `gh-nav-menu/footer-banner.hbs` for referral and what's new banners/elements.
> - **Service: `whats-new`**
>   - Add `onboarding` service dependency and new computed `shouldShowFeaturedBanner` to hide featured banner when `onboarding.isChecklistShown` is true.
> - **Tests**
>   - New integration tests for `gh-nav-menu/footer-banner` covering referral priority and banner visibility rules.
>   - New unit tests for `whats-new` service covering `hasNew`, `hasNewFeatured`, and `shouldShowFeaturedBanner`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3703894be2a5df2fe34570f729488b364aa98cef. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/admin/app/components/gh-nav-menu/footer-banner.js
+++ b/ghost/admin/app/components/gh-nav-menu/footer-banner.js
@@ -48,7 +48,7 @@ export default class FooterBanner extends Component {
     }
 
     get showWhatsNew() {
-        return !this.showReferralInvite && this.whatsNew.hasNewFeatured;
+        return !this.showReferralInvite && this.whatsNew.shouldShowFeaturedBanner;
     }
 
     @task
--- a/ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
+++ b/ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
@@ -194,4 +194,53 @@ describe('Integration: Component: gh-nav-menu/footer-banner', function () {
 
         expect(find('.gh-whatsnew-toast'), 'what\'s new banner is hidden').to.not.exist;
     });
+
+    it('uses shouldShowFeaturedBanner from whatsNew service', async function () {
+        this.owner.register('service:whatsNew', Service.extend({
+            hasNew: true,
+            hasNewFeatured: true,
+            shouldShowFeaturedBanner: false, // Override to false
+            entries: [{
+                title: 'Test Feature',
+                url: 'https://ghost.org/changelog/test',
+                custom_excerpt: 'Test description',
+                published_at: '2024-12-01T00:00:00.000Z',
+                featured: true
+            }],
+            seen() {}
+        }));
+
+        this.owner.register('service:session', Service.extend({
+            user: {
+                isAdmin: false
+            }
+        }));
+
+        this.owner.register('service:dashboardStats', Service.extend({
+            currentMRR: 0,
+            loadMrrStats() {
+                return Promise.resolve();
+            }
+        }));
+
+        this.owner.register('service:feature', Service.extend({
+            accessibility: {
+                referralInviteDismissed: false
+            }
+        }));
+
+        this.owner.register('service:membersUtils', Service.extend({
+            isStripeEnabled: false
+        }));
+
+        this.owner.register('service:modals', Service.extend({}));
+        this.owner.register('service:settings', Service.extend({
+            stripeConnectLivemode: false
+        }));
+
+        await render(hbs`<GhNavMenu::FooterBanner />`);
+
+        // Should NOT show banner because shouldShowFeaturedBanner=false
+        expect(find('.gh-whatsnew-toast'), 'what\'s new banner respects shouldShowFeaturedBanner').to.not.exist;
+    });
 });
--- a/ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
+++ b/ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
@@ -8,239 +8,150 @@ import {setupRenderingTest} from 'ember-mocha';
 describe('Integration: Component: gh-nav-menu/footer-banner', function () {
     setupRenderingTest();
 
-    it('shows what\'s new banner when hasNewFeatured is true and referral invite is not showing', async function () {
+    // Composable fixture fragments
+    const hasNewFeature = {
+        hasNewFeature: true,
+        shouldShowFeaturedBanner: true,
+        featuredEntry: {
+            title: 'Test Feature',
+            url: 'https://ghost.org/changelog/test',
+            custom_excerpt: 'Test description',
+            published_at: '2024-12-01T00:00:00.000Z',
+            featured: true
+        }
+    };
+
+    const noNewFeature = {
+        hasNewFeature: false
+    };
+
+    const shouldNotShowFeaturedBanner = {
+        shouldShowFeaturedBanner: false
+    };
+
+    const eligibleForReferral = {
+        isAdmin: true,
+        currentMRR: 10100,
+        isStripeEnabled: true,
+        stripeConnectLivemode: true,
+        referralInviteDismissed: false
+    };
+
+    const notEligibleForReferral = {
+        currentMRR: 0,
+        isStripeEnabled: false,
+        stripeConnectLivemode: false
+    };
+
+    const defaultFixture = {
+        ...noNewFeature,
+        ...shouldNotShowFeaturedBanner,
+        ...notEligibleForReferral
+    };
+
+    const setup = function (fixtures = {}) {
+        const config = {
+            ...defaultFixture,
+            ...fixtures
+        };
+
+        // Build whatsNew.entries based on configuration
+        let entries = [];
+        if (config.hasNewFeature && config.featuredEntry) {
+            entries = [config.featuredEntry];
+        }
+
         this.owner.register('service:whatsNew', Service.extend({
-            hasNew: true,
-            hasNewFeatured: true,
-            entries: [{
-                title: 'Test Feature',
-                url: 'https://ghost.org/changelog/test',
-                custom_excerpt: 'Test description',
-                published_at: '2024-12-01T00:00:00.000Z',
-                featured: true
-            }],
-            seen() {}
+            seen() {},
+            init() {
+                this._super(...arguments);
+                this.entries = entries;
+            },
+            hasNew: config.hasNewFeature,
+            hasNewFeatured: config.hasNewFeature,
+            shouldShowFeaturedBanner: config.shouldShowFeaturedBanner
         }));
 
         this.owner.register('service:session', Service.extend({
-            user: {
-                isAdmin: false
+            init() {
+                this._super(...arguments);
+                this.user = {
+                    isAdmin: config.isAdmin
+                };
             }
         }));
 
         this.owner.register('service:dashboardStats', Service.extend({
-            currentMRR: 0,
             loadMrrStats() {
                 return Promise.resolve();
-            }
+            },
+            currentMRR: config.currentMRR
         }));
 
         this.owner.register('service:feature', Service.extend({
-            accessibility: {
-                referralInviteDismissed: false
+            init() {
+                this._super(...arguments);
+                this.accessibility = {
+                    referralInviteDismissed: config.referralInviteDismissed
+                };
             }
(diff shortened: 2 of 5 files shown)


ghost/admin/app/components/gh-nav-menu/footer-banner.js
```diff
@@ -48,12 +48,12 @@ export default class FooterBanner extends Component {
...
 
     get showWhatsNew() {
-        return !this.showReferralInvite && this.whatsNew.hasNewFeatured;
+        return !this.showReferralInvite && this.whatsNew.shouldShowFeaturedBanner;
     }
 
     @task
     *loadCurrentMRR() {
-        if (this.isAdminOrOwnern) {
+        if (this.isAdminOrOwner) {
```
rob-ghost: Fixed a typo, but this might be out-of-scope because I presume before anything relating to `dashboardStats` wasn't working, and by fixing this typo it might start working again, which in turn might cause other bugs / behaviour changes.

Let me know if we're comfortable with this change otherwise I can revert it back and fix it in a separate branch.

jonatansberg: Seems like we have the same typo in `ghost/admin/app/components/gh-referral-invite.js`. The same service method then also gets called by `ghost/admin/app/components/dashboard/charts/anchor-attribution.js` if the site `hasPaidTiers`. 

Maybe this means we're always calling that service regardless of those checks in the footer and the referral invite, and/or this is a no-op, meaning this change doesn't really matter? @kevinansfield do you have any insight into this?

I generally like the idea of tidying things as we go. In this particular case I'm not sure about what's the right course of action. Fixing it separately might be easier to reason about. Especially if you end up squashing this PR.

ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
```diff
@@ -0,0 +1,162 @@
...
+    it('shows what\'s new banner when hasNewFeatured is true and referral invite is not showing', async function () {
+        setup.call(this, {...hasNewFeature, ...notEligibleForReferral});
+
+        await render(hbs`<GhNavMenu::FooterBanner />`);
+
+        expect(find('.gh-whatsnew-toast'), 'what\'s new banner is visible').to.exist;
+        expect(find('.gh-sidebar-banner-subhead'), 'subhead element exists').to.exist;
+        expect(find('.gh-sidebar-banner-subhead').textContent).to.match(/What.s new\?/);
+        expect(find('.gh-sidebar-banner-msg')).to.contain.text(hasNewFeature.featuredEntry.title);
+        expect(find('.gh-sidebar-banner-details')).to.contain.text(hasNewFeature.featuredEntry.custom_excerpt);
```
jonatansberg: Nit: I'm not loving the class name selectors here and in the other assertions, but maybe that's how the other Ember tests are structured?

In the more modern E2E tests we favor role based selectors and fall back to test ids otherwise.

ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
```diff
@@ -0,0 +1,162 @@
...
+    });
+
+    it('hides what\'s new banner when hasNewFeatured is false', async function () {
+        setup.call(this, {...noNewFeature, ...notEligibleForReferral});
```
jonatansberg, line 126: Nit: I don't feel like having the negative fragments here adds that much value since they seem to match the default? I can see the intent with being explicit, but I'm not sure it makes things easier to reason about.

jonatansberg (commented):
Overall, I think this looks good. I have some small nitpick comments, but nothing that should prevent merging from my end. I'll defer the final judgment to Kevin though.

ghost/admin/app/components/gh-nav-menu/footer-banner.js
rob-ghost: I'm happy to take this as a follow-up with its own scope, I think that makes more sense then shoehorning it into this PR scope which isn't related to loading MRR.

ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
rob-ghost: I think this does align with other modules in `/ghost/admin/tests/integration/components`, but I'd be open to changing it if we have a more modern recommendation.

What would you suggest for this component?

ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
rob-ghost: I think there's a few different concerns:

1. What is the shape of the fixture that this suite accepts? These are the fields defined in `defaultFixture`
2. What are sensible defaults for this suite? These are the values in `defaultFixture`
3. What is the declarative setup for this specific test, this is the specific `noNewFeature` fixture etc which describes what the parameters of the test are

Its possible for the test parameters to match the default, but they don't have to.

Would it make more sense if we created the default from fixtures perhaps?

```javascript
const default = {
    ...noNewFeature,
    ...notEligibleForReferral,
    ...shouldNotShowFeaturedBanner,
}
```

This way there would be no re-declaration but the default state is declarative using the same fixtures as the test?

jonatansberg: Those are all valid concerns. I think composing the default fixture using the narrower ones makes it way nicer!

ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
jonatansberg: Selecting the relevant elements based on their ARIA roles makes the tests more robust with regards to styling or markup changes. It would also make the assertions easier to port to the Playwright test suite when we migrate this to React in the not too distant future.<br><br>If role-based selectors aren't available in the Ember tooling, then I would reach for test IDs.<br><br>With that said, I haven't worked with the integration tests in Ember, so take what I'm saying with a pinch of salt. There's definitely a point to just being consistent with what's in the other tests as well.

ghost/admin/app/components/gh-nav-menu/footer-banner.js
kevinansfield: I'd certainly fix this separately as it potentially has an impact on unrelated behaviour.

As for whether or not this would appear broken with the typo... since Ghost 6.0 it will have been broken for anyone with analytics enabled because the old dashboard won't be shown. For anyone without analytics, if they load the dashboard then `charts/anchor-attribution.js` would have loaded the data and allowed it to work.

So fixing the typo will definitely get us back to a working state for the majority of customers.

Interestingly, I don't think `ghost/admin/app/components/gh-referral-invite.js` which has the same typo is used anywhere any more. Looks like it wasn't cleaned up [when it's usage was removed](https://github.com/TryGhost/Ghost/pull/20112/files#diff-dca7732e2a72b5eb4a456d17d64c4f85ae3c7ac72d8cdef360cefaee8078ab46L9).

ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
kevinansfield: The current approach is `data-testid` attributes, there should be lots of examples for that in the Admin codebase 🙂

kevinansfield (commented):
@rob-ghost are we missing something that calls `whatsNew.seen()` when we don't want to display the what's new banner? If we don't call that, as soon as the onboarding checklist is completed/dismissed we'll show the what's new banner despite being considered a "new site" that shouldn't be shown previous release news.

Or has then already been considered and deemed acceptable?

If we don't want to show "what's new" to anyone on a site until something new has been released, perhaps there's a server-side approach we could take instead so new User records are instantiated with the right `accessibility` (legacy naming, is really "user settings" 😬) data to avoid showing them old news?

rob-ghost:
> @rob-ghost are we missing something that calls `whatsNew.seen()` when we don't want to display the what's new banner? If we don't call that, as soon as the onboarding checklist is completed/dismissed we'll show the what's new banner despite being considered a "new site" that shouldn't be shown previous release news.
> 
> Or has then already been considered and deemed acceptable?

No I think this is a valid point. Right now when you dismiss the onboarding checklist you'll be shown the `What's new` toast still, because `lastSeenDate` won't be set so it will show the latest update regardless.

I think it would be better if we only show `What's new` for things that come _after_ onboarding. I'll look into the best place to call `whatsNew.seen()` now and update the PR.

Thanks! 🙌🏻

rob-ghost:
> If we don't want to show "what's new" to anyone on a site until something new has been released, perhaps there's a server-side approach we could take instead so new User records are instantiated with the right `accessibility` (legacy naming, is really "user settings" 😬) data to avoid showing them old news?

Meaning, if we set `lastSeenDate` on `accessibility` in the backend to be the time the user was created, then it renders this whole PR moot because we'd not show any `What's new` toasts until the next update?

I think this is a great suggestion, let me check it out and I'll report back because this would solve the problem more reliably I think and avoid the frontend having to manage `lastSeenDate` state (calling `.seen()` etc).

The only difference between the BE approach and this FE approach is that someone can signup, go away for a while, and come back and still be onboarding but now they see a new toast with updates since they signed up, but we can either:

1. Accept that because it makes sense, there's been updates since you first signed up even if you've not completed onboarding
2. We can still hide it if the onboarding checklist is present, and it will be shown when its dismissed and its contents may still be relevant (due to 1).

ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
rob-ghost: Changed the default fixture to be built from the existing fixtures ✅

ghost/admin/app/components/gh-nav-menu/footer-banner.js
rob-ghost: I reverted this change ✅ I'll create a follow-up ticket in Linear and PR

ghost/admin/tests/integration/components/gh-nav-menu/footer-banner-test.js
rob-ghost: I added `data-test` attributes to the component and updated the testes ✅

jonatansberg:
I think either of those two approaches makes sense. I would lean towards the first approach as it's a pretty harmless edge case, arguably still valuable, and simpler is better.

rob-ghost:
Closing in favour of the approach in https://github.com/TryGhost/Ghost/pull/25075

PR #25018, opened 2025-09-30
Added utm_ columns to events tables

9larsons:
ref https://linear.app/ghost/issue/PROD-2689/
ref https://github.com/TryGhost/Ghost/commit/d8e58b53d0984c198c2a6e12bd73b44356aab350
ref https://github.com/TryGhost/Ghost/commit/2bfd8f8b7ec1f487c92b8497955f573bbd14f422
- added utm_ related fields to members_created_events, members_subscription_created_events, and donation_payment_events
- updated schema
- updated migration `createAddColumnMigration` to not force `algorithm=copy` (added 'auto' option)

This PR contains the migrations necessary to store the utm_ parameters we are adding tracking for. As they are another element of member attribution, they've been added to the tables that already have member attribution tracking: members_created_events, members_subscription_created_events, and donation_payment_events.

The columns added are for all of the utm_ fields: utm_campaign, utm_term, utm_content, utm_medium, utm_source.

We already have referrer_medium and referrer_source, but these are 'processed' fields that do not directly correspond to utm_ fields, so we've added the additional column. At some point, it may be worth consolidating.

__NOTE__: W/r to migrations (see refs above), we had found in investigating changing this migration to use our helpers instead of manually creating the trx and column changes that the performance was _awful_. We traced this back to the use of `algorithm=copy` which, to the best of our knowledge, was useful for the purposes of migrations on the `posts` table due to the very large `long` fields (lexical, mobiledoc, etc). This algorithm forces a copy of the table prior to acting on it, which on a table like `members_created_events` with a sample 2M rows, was incredibly slow (60-250s) vs. letting the sql engine do it's thing (2-3s). 

It seems likely that many of our migrations could actually be improved by letting the sql optimizer create its own plan (although clearly this was not true for posts). While we have some ideas on how to programmatically handle this, it feels safest to add a new param to allow for specifying the algorithm; this is currently ONLY used in adding and dropping a column, so those were the two places updated.

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/6.2/2025-09-30-14-28-09-add-utm-fields.js
+++ b/ghost/core/core/server/data/migrations/versions/6.2/2025-09-30-14-28-09-add-utm-fields.js
@@ -2,23 +2,23 @@ const {combineNonTransactionalMigrations, createAddColumnMigration} = require('.
 
 module.exports = combineNonTransactionalMigrations(
     // members_created_events
-    createAddColumnMigration('members_created_events', 'utm_source', {type: 'string', maxlength: 191, nullable: true}),
-    createAddColumnMigration('members_created_events', 'utm_medium', {type: 'string', maxlength: 191, nullable: true}),
-    createAddColumnMigration('members_created_events', 'utm_campaign', {type: 'string', maxlength: 191, nullable: true}),
-    createAddColumnMigration('members_created_events', 'utm_term', {type: 'string', maxlength: 191, nullable: true}),
-    createAddColumnMigration('members_created_events', 'utm_content', {type: 'string', maxlength: 191, nullable: true}),
+    createAddColumnMigration('members_created_events', 'utm_source', {type: 'string', maxlength: 191, nullable: true, algorithm: 'inplace'}),
+    createAddColumnMigration('members_created_events', 'utm_medium', {type: 'string', maxlength: 191, nullable: true, algorithm: 'inplace'}),
+    createAddColumnMigration('members_created_events', 'utm_campaign', {type: 'string', maxlength: 191, nullable: true, algorithm: 'inplace'}),
+    createAddColumnMigration('members_created_events', 'utm_term', {type: 'string', maxlength: 191, nullable: true, algorithm: 'inplace'}),
+    createAddColumnMigration('members_created_events', 'utm_content', {type: 'string', maxlength: 191, nullable: true, algorithm: 'inplace'}),
 
     // members_subscription_created_events
-    createAddColumnMigration('members_subscription_created_events', 'utm_source', {type: 'string', maxlength: 191, nullable: true}),
-    createAddColumnMigration('members_subscription_created_events', 'utm_medium', {type: 'string', maxlength: 191, nullable: true}),
-    createAddColumnMigration('members_subscription_created_events', 'utm_campaign', {type: 'string', maxlength: 191, nullable: true}),
-    createAddColumnMigration('members_subscription_created_events', 'utm_term', {type: 'string', maxlength: 191, nullable: true}),
-    createAddColumnMigration('members_subscription_created_events', 'utm_content', {type: 'string', maxlength: 191, nullable: true}),
+    createAddColumnMigration('members_subscription_created_events', 'utm_source', {type: 'string', maxlength: 191, nullable: true, algorithm: 'inplace'}),
+    createAddColumnMigration('members_subscription_created_events', 'utm_medium', {type: 'string', maxlength: 191, nullable: true, algorithm: 'inplace'}),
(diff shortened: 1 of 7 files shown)


cmraible:
Any idea if any of these tables are particularly large? I wouldn't be shocked if some of these are upwards of 1 million rows for some sites.

We should probably test these on staging with a representative database size.

9larsons:
I can make a stats service query to check. That said, we've never really been concerned with adding a nulled column - that should be a very quick operation for any size, even with adding a few to each table.

9larsons:
Timing is ~3s for rolling forward and backwards with 2M rows in `members_created_events` on staging. I didn't see any adverse effects when clicking around.

ghost/core/core/server/data/migrations/versions/6.2/2025-09-30-14-28-09-add-utm-fields.js
```diff
@@ -0,0 +1,60 @@
...
+const {createTransactionalMigration} = require('../../utils');
+
+const tables = [
+    'members_created_events',
+    'members_subscription_created_events',
+    'donation_payment_events'
+];
+
+const utmColumns = [
+    {name: 'utm_source', definition: {type: 'string', maxlength: 191, nullable: true}},
```
cmraible: Just curious: where does 191 come from?

cmraible (changes requested):
Nice one, thanks for testing the performance. Apologies for the request this late in the game, I should have done a more thorough review when you first asked for it.

ghost/core/core/server/data/migrations/versions/6.2/2025-09-30-14-28-09-add-utm-fields.js
9larsons: It's common used across our schema and seemed a reasonable limit. It's a varchar limit for InnoDB indices, though arguably we probably don't need to index on these and could extend it. Are you concerned about overflow?

https://dev.mysql.com/doc/refman/8.4/en/innodb-limits.html

cmraible: Nope no concerns, just seemed to be an odd number so I was curious

ghost/core/core/server/data/migrations/versions/6.2/2025-09-30-14-28-09-add-utm-fields.js
```diff
@@ -0,0 +1,60 @@
...
+                if (!hasColumn) {
+                    logging.info(`Adding column ${column.name} to ${table}`);
+                    await knex.schema.table(table, function (t) {
+                        t.string(column.name, column.definition.maxlength).nullable();
+                    });
+                } else {
+                    logging.warn(`Column ${column.name} already exists in ${table} - skipping`);
+                }
+            }
+        }
```
cmraible: suggestion: I think we should use the standard utils we have to ensure consistency in creating the columns here. 

i.e.:
```
 module.exports = combineNonTransactionalMigrations(
      // members_created_events
      createAddColumnMigration('members_created_events', 'utm_source', {type: 'string', maxlength: 191, nullable: true}),
      createAddColumnMigration('members_created_events', 'utm_medium', {type: 'string', maxlength: 191, nullable: true}),
      createAddColumnMigration('members_created_events', 'utm_campaign', {type: 'string', maxlength: 191, nullable: true}),
      createAddColumnMigration('members_created_events', 'utm_term', {type: 'string', maxlength: 191, nullable: true}),
      createAddColumnMigration('members_created_events', 'utm_content', {type: 'string', maxlength: 191, nullable: true}),

      // members_subscription_created_events
      createAddColumnMigration('members_subscription_created_events', 'utm_source', {type: 'string', maxlength: 191, nullable: true}),
      // ... and so on
  );

```

While it is arguably more verbose and somewhat violates DRY, I think it would reduce the cognitive load of understanding what this migration does. There are also some "special" things that the helpers do that the current implementation skips.

In particular (from Claude):

> > The createAddColumnMigration → addColumn function adds , algorithm=copy to MySQL ALTER TABLE statements:
> 
>   if (DatabaseInfo.isMySQL(transaction)) {
>       sql = sql.replace(/;\s*$/, '') + ', algorithm=copy';
>   }
> 
>   What this does: Forces MySQL to use the COPY algorithm instead of INPLACE. This:
>   - Creates a full copy of the table with the new column
>   - More reliable for schema changes
>   - Avoids potential locking issues with INPLACE algorithm
>   - Generally safer for production deployments
> 
>   The current PR implementation uses plain Knex t.string(...).nullable() which defaults to MySQL's INPLACE algorithm when possible.

cmraible (approved):
👌 Thanks for humoring me with these changes — 'twas a bit painful but I'm pretty happy with the solution we came up with, awesome work!

PR #25019, opened 2025-09-30
Move on using real tag factories, instead of mocks

ibalosh:
ref https://github.com/TryGhost/Ghost/pull/24816

* Added this to support @jonatansberg with tags factory to avoid mocking on [his test suite](https://github.com/TryGhost/Ghost/pull/24816).
* We needed a factory for it, basic page object, and test in order to be able to add tag related tests. This will allow team that works on migrating Ember tags section to React, to add tests easier.


9larsons (approved):
Looks great! I'm really enjoying how easily these come together.

e2e/data-factory/factories/tags/tag-factory.ts
```diff
@@ -0,0 +1,73 @@
...
+    twitter_description: string | null;
+    meta_title: string | null;
+    meta_description: string | null;
+    codeinjection_head: string | null;
+    codeinjection_foot: string | null;
+    canonical_url: string | null;
+    accent_color: string | null;
+    created_at: Date;
+    updated_at: Date | null;
+}
```
jonatansberg, line 32: Do we really want to be redefining the interfaces in the tests? 

I would really love for us to find a better way to reuse these type definitions. Obviously, out of scope for this PR, but it seems like something we might want to tackle soon.

e2e/helpers/pages/admin/tags/TagsPage.ts
```diff
@@ -41,4 +43,8 @@ export class TagsPage extends AdminPage {
     getRowByTitle(title: string) {
         return this.tagListRow.filter({has: this.page.getByRole('link', {name: title, exact: true})});
     }
+
+    getTagByName(name: string): Locator {
+        return this.tagNames.filter({hasText: name});
```
jonatansberg: This needs to be an `exact` match, otherwise you might get multiple matches if more than one tag contains the substring you're matching against.

e2e/helpers/pages/admin/tags/TagsPage.ts
```diff
@@ -22,6 +23,7 @@ export class TagsPage extends AdminPage {
         this.pageContent = page.getByTestId('tags-page');
         this.tagList = page.getByTestId('tags-list');
         this.tagListRow = this.tagList.getByTestId('tag-list-row');
+        this.tagNames = page.locator('[data-test-tag-name]');
```
jonatansberg, line 25: The `data-test-tag-name` attribute doesn't exist in the new React frontend.

e2e/data-factory/factories/tags/tag-factory.ts
```diff
@@ -0,0 +1,73 @@
...
+            canonical_url: options.canonical_url ?? null,
+            accent_color: options.accent_color ?? null,
+            created_at: options.created_at ?? now,
+            updated_at: options.updated_at ?? now
+        };
+
+        return {
+            ...defaults,
+            ...options
+        } as Tag;
```
jonatansberg: Do you really need to spread the options on top of the tag since you're already enumerating all of the properties sbove? That would allow you to create tags that aren't actually valid. 

Also, the `as` typecast can probably be replaced with a `satisfies` as to not accidentally introduce typing errors. Or you can remove it entirely since you've already defined the return type of the function.

e2e/tests/admin/tags/tags.test.ts
```diff
@@ -0,0 +1,20 @@
...
+    });
+
+    test('shows newly created tag in public tags list', async ({page}) => {
+        await tagFactory.create({name: 'test new tag'});
+
+        const tagsPage = new TagsPage(page);
+        await tagsPage.goto();
+
+        await expect(tagsPage.getTagByName('test new tag')).toBeVisible();
+    });
```
jonatansberg: Whats the reasoning behind adding a separate test file rather than just integrating the factory with the existing test?

jonatansberg (commented):
The factory's approach looks promising! There's a few things I think we should look at before merging this, but it's a good start 👍

e2e/data-factory/factories/tags/tag-factory.ts
ibalosh: Agreed, I would love a better way that we can use here, but.. we don't really have interfaces to begin with, at least I could not find them in old codebase within core.

Shapes for current factories are picked up by reading db schema and responses from API (which also have in some places on-the-fly data added for some of the factories we plan to add).

I am open for better approaches, but having specified API (with typed requests and responses) is a completely different beast, that is probably as challenging work as current test suite (and probably more)

cc-ing @cmraible and @9larsons for visibility (this topic did come up in our discussions too)

open for ideas on tackling this differently in future

e2e/helpers/pages/admin/tags/TagsPage.ts
ibalosh: updated

e2e/data-factory/factories/tags/tag-factory.ts
ibalosh: missed that from AI.. you are right, removed the spread completely

e2e/tests/admin/tags/tags.test.ts
ibalosh: I did not have your changes in here, this was just the part of merge. I left it outside until I can get your tests running locally, with factories.

(currently I can't run them locally at all, not sure why - it seems to be related to dockerized build of ghost in the test environment)

will leave this comment open, as a note to check once we get settings API merged in here

ibalosh:
@jonatansberg added all the changes I could at this point, will see to get these fully running once settings API can be merged in and I can see react app settings locally

PR #25022, opened 2025-09-30
Updated new paid member newsletter handling

9larsons:
ref https://linear.app/ghost/issue/ONC-939/
ref https://github.com/TryGhost/Ghost/pull/23302 (initial spike)

Previously, we were using Stripe's metadata field to pass the newsletter subscription selection. Stripe has a limit of 500 characters for each key's value, so it was plausible to hit this limit with a dozen or so newsletters. This failed silently by not respecting members' selections when signing up.

With free member creation flows, we set the newsletter data as part of the magic token we generate and send to require email validation. As we're already generating a magic link for the success url (redirect) following the Stripe checkout completion, we can simply parse this for the token and access the newsletter data without having to pass it to Stripe.

In other words, we effectively use the tokens table as a 'pending member'. And by changing our handling here, we align on a single approach for how we handle sign up data.

As a note, the token field allows up to 2000 characters, so even after the other data we pack in there and base64url encoding, we have ample headroom over the previous Stripe metadata handling, as our other fields are reasonably small.

----

Coauthored by @cmraible

### DIFF
--- a/ghost/core/test/e2e-api/members/create-stripe-checkout-session.test.js
+++ b/ghost/core/test/e2e-api/members/create-stripe-checkout-session.test.js
@@ -524,4 +524,174 @@ describe('Create Stripe Checkout Session', function () {
             should(scope.isDone()).eql(true);
         });
     });
+
+    /**
+     * Newsletter preference tests
+     */
+    describe('Newsletter preferences', function () {
+        it('Should not pass newsletter data to Stripe metadata but include in success URL token', async function () {
+            const {body: {tiers}} = await adminAgent.get('/tiers/?include=monthly_price&yearly_price');
+            const paidTier = tiers.find(tier => tier.type === 'paid');
+
+            let stripeSessionMetadata;
+
+            nock('https://api.stripe.com')
+                .persist()
+                .get(/v1\/.*/)
+                .reply((uri) => {
+                    const [match, resource, id] = uri.match(/\/v1\/(\w+)\/(.+)\/?/) || [null];
+                    if (match) {
+                        if (resource === 'products') {
+                            return [200, {
+                                id: id,
+                                active: true
+                            }];
+                        }
+                        if (resource === 'prices') {
+                            return [200, {
+                                id: id,
+                                active: true,
+                                currency: 'usd',
+                                unit_amount: 500,
+                                recurring: {
+                                    interval: 'month'
+                                }
+                            }];
+                        }
+                    }
+                    return [500];
+                });
+
+            nock('https://api.stripe.com')
+                .persist()
+                .post(/v1\/.*/)
+                .reply((uri, body) => {
+                    if (uri === '/v1/checkout/sessions') {
+                        // Capture the metadata sent to Stripe
+                        const params = new URLSearchParams(body);
+                        stripeSessionMetadata = {};
+                        for (const [key, value] of params.entries()) {
+                            if (key.startsWith('metadata[')) {
+                                const metaKey = key.match(/metadata\[([^\]]+)\]/)[1];
+                                stripeSessionMetadata[metaKey] = value;
+                            }
+                        }
+
+                        return [200, {
+                            id: 'cs_123',
+                            url: 'https://checkout.stripe.com/session/cs_123'
+                        }];
+                    }
+                    if (uri === '/v1/prices') {
+                        return [200, {
+                            id: 'price_newsletter',
+                            active: true,
+                            currency: 'usd',
+                            unit_amount: 500,
+                            recurring: {
+                                interval: 'month'
+                            }
+                        }];
+                    }
+                    return [500];
+                });
+
+            await membersAgent.post('/api/create-stripe-checkout-session/')
+                .body({
+                    customerEmail: 'newsletter-test@test.com',
+                    tierId: paidTier.id,
+                    cadence: 'month',
+                    metadata: {
+                        newsletters: JSON.stringify([])
+                    }
+                })
+                .expectStatus(200);
+
+            // Verify that newsletters were NOT sent to Stripe metadata
+            should(stripeSessionMetadata).not.have.property('newsletters');
+        });
+
+        it('Should handle empty newsletter array without sending to Stripe', async function () {
+            const {body: {tiers}} = await adminAgent.get('/tiers/?include=monthly_price&yearly_price');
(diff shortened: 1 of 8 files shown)


ghost/core/test/e2e-api/members/create-stripe-checkout-session.test.js
```diff
@@ -524,4 +524,174 @@ describe('Create Stripe Checkout Session', function () {
...
+            const {body: {tiers}} = await adminAgent.get('/tiers/?include=monthly_price&yearly_price');
+            const paidTier = tiers.find(tier => tier.type === 'paid');
+
+            let stripeSessionMetadata;
+
+            nock('https://api.stripe.com')
+                .persist()
+                .get(/v1\/.*/)
+                .reply((uri) => {
+                    const [match, resource, id] = uri.match(/\/v1\/(\w+)\/(.+)\/?/) || [null];
```
cmraible: question: What is this regex doing? I wonder if we could pull it into a helper to make it easier to understand what's happening here.

ghost/core/test/e2e-api/members/create-stripe-checkout-session.test.js
cmraible: Do you think we should update the attribution while we're at it? Seems a bit weird to handle newsletters one way, and attribution data a different way.

ghost/core/test/e2e-api/members/create-stripe-checkout-session.test.js
```diff
@@ -524,4 +524,174 @@ describe('Create Stripe Checkout Session', function () {
             should(scope.isDone()).eql(true);
         });
     });
+
+    /**
+     * Newsletter preference tests
+     */
+    describe('Newsletter preferences', function () {
+        it('Should not pass newsletter data to Stripe metadata but include in success URL token', async function () {
```
cmraible:
```suggestion
        it('Should not pass newsletter data to Stripe metadata', async function () {
```
In its current form, this only tests that we _don't_ pass newsletters in the Stripe metadata. I think maybe we're missing a test for the success URL token part? Or maybe we test that elsewhere...

ghost/core/test/e2e-api/members/create-stripe-checkout-session.test.js
```diff
@@ -524,4 +524,174 @@ describe('Create Stripe Checkout Session', function () {
...
+                    return [500];
+                });
+
+            await membersAgent.post('/api/create-stripe-checkout-session/')
+                .body({
+                    customerEmail: 'newsletter-test@test.com',
+                    tierId: paidTier.id,
+                    cadence: 'month',
+                    metadata: {
+                        newsletters: JSON.stringify([])
```
cmraible: Since we cover the empty newsletter array case below, I think we should add some newsletters here to be sure it works when there _is_ data.

ghost/core/test/unit/server/services/stripe/services/webhooks/CheckoutSessionEventService.test.js
```diff
@@ -625,10 +631,26 @@ describe('CheckoutSessionEventService', function () {
             assert.deepEqual(memberData.newsletters, [{id: 1, name: 'Newsletter'}]);
         });
 
-        it('should create new member with newsletters if metadata newsletters is not valid JSON', async function () {
+        it('should create new member without newsletters when token has no newsletter data', async function () {
```
cmraible: This is good, but I think we're missing a test for the "happy path", where the token does have newsletter data.

ghost/core/test/e2e-api/members/webhooks.test.js
```diff
@@ -943,55 +943,6 @@ describe('Members API', function () {
             assert.equal(member.newsletters.length, 1, 'The member should have a single newsletter');
         });
 
-        it('Will create a member with signup newsletter preference', async function () {
```
cmraible, line 946: Is this covered elsewhere? The implementation of this test will need to change, but I think we still want this test case.

ghost/core/test/e2e-api/members/create-stripe-checkout-session.test.js
cmraible, line 201: potential issue: what happens in this scenario? I can't think of why we would allow a checkout session without an email address — but if there is a valid case for this, then presumably we don't create a magic link, and this whole approach might not be viable 🤔

cmraible (changes requested):
Nice work! This whole area is such a pain to reason about, sorry it took so long to review. 

A few concerns:
- It looks like we sometimes allow checkout sessions without email addresses — I'm not sure why we do that, but if that's a valid case then this whole approach might not work. Trying to sort out why we'd accept this, but haven't gotten anywhere yet.
- Some of the new tests seem duplicative and/or don't match the test's title.
- I think maybe we're missing an e2e-api test for the case where newsletter _are_ passed?

ghost/core/test/e2e-api/members/create-stripe-checkout-session.test.js
9larsons: I tried putting these mocks into a helper and it became an absolute mess to reason about.. although the current state is effectively the same. I'll play around and see if I can get it in better shape.

It's just trying to match any of the stripe v1 apis. e.g. /v1/products/prod_123 or /v1/prices/price_123

ghost/core/test/e2e-api/members/create-stripe-checkout-session.test.js
9larsons: Updated to do this. I didn't want to bloat this but the changes are certainly overlaid.

cmraible (changes requested):
Upon further inspection, I don't think we need this change at all:
- The 500 character limit on Stripe metadata is _per key_, not an overall limit
- Therefore, even if we add 5 new UTM related keys to the metadata object, there's no additional risk of passing the limit even if we still use metadata for newsletters
- Subscription upgrades (free member -> paid) and donations don't use magic link tokens, so we'd probably still be using Stripe metadata for UTMs regardless.

Ultimately I think change just isn't necessary — but since we have it all queued up and context loaded, I wouldn't be against merging it for _just the newsletters for new subscriptions_ — but either way I don't think should block our UTM implementation.

cmraible:
Closing this Pull request for now to clear it from my review queue. Will post follow up context dump on what we explored here and why we ultimately decided against it soon.

PR #25025, opened 2025-10-01
Added member and member detail base POM

ibalosh:
Initial page object and test was added to support the team with adding additional tests in member attribution area.

- [x] 've written an automated test to prove my change works


e2e/helpers/pages/admin/MemberDetailsPage.ts
```diff
@@ -0,0 +1,38 @@
...
+        await this.emailInput.fill(email);
+    }
+
+    async save(): Promise<void> {
+        await this.saveButton.click();
+        await this.page.waitForLoadState('networkidle');
+    }
+
+    async deleteMember(): Promise<void> {
+        await this.deleteButton.click();
```
9larsons: Do we need a network idle here or confirmation?

9larsons (approved):
Some of the CR suggestions for naming seems reasonable. Overall looks good!

e2e/helpers/pages/admin/MemberDetailsPage.ts
ibalosh: good call, AI.. probably not, I did not use that yet, I will remove the extra method wrappers here, we probably do not need them

PR #25026, opened 2025-10-01
Created page hit promise before visiting home page

cmraible:
Just testing this out for now. 

It seems the `waitForPageHitRequest` is flaky — works in CI and on my machine, but not on others. Suspect there is a race condition where the request finishes before we are able to create the promise.


cmraible:
Closing in favor of https://github.com/TryGhost/Ghost/commit/2364dd7419539795c03b537abbfa9a28510c0fb5

PR #25027, opened 2025-10-01
Upgraded Shade to TailwindCSS version 4

peterzimon:
TBD


peterzimon:
Obsolete, Shade is already on TWCSS 4

PR #23725, opened 2025-06-06
Added `site_uuid` setting migration

cmraible:
ref https://linear.app/ghost/issue/PROD-197/ghost-site-uuid-implementation

The `site_uuid` is intended to be a unique identifier for an instance of Ghost, to aid in integrating with multi-tenant services like Tinybird. This commit adds a migration to add the `site_uuid` setting, and a minimal change set to ensure it doesn't accidentally get overwritten by a JSON content import.

It sets the `site_uuid` from `config.get('site_uuid')` if it is set and valid, otherwise it generates a new uuid. The `site_uuid` value should be written once when the site is created (or at the next boot for existing sites) and it should be immutable.

Additional changes, including the validation of the `site_uuid` on boot and the inclusion of this value in the `/site` endpoint will be included in follow up PRs, to keep this change with the migration as small as possible.

PR #25029, opened 2025-10-01
Added `members_otc_secret` setting

kevinansfield:
closes https://linear.app/ghost/issue/PROD-2519/

`SingleUseTokenProvider` was previously instantiated using the `members_auth_email_secret` setting but that was a hangover from some older auth code in members so it makes sense to switch to a new+specific `members_otc_secret` setting.

- added new setting to `default-settings.json` for implicit creation when missing
- added new setting to the `dynamicDefault` map so it gets populated with a random string of sufficient length during creation
- added migration for explicit creation of the setting prior to Ghost boot
- added new setting to our group+type maps for the settings API input serializer
- updated `SingleUseTokenProvider` instantiation to use the new setting in place of `members_auth_email_secret`


troyciesco:
@kevinansfield i see in your description (and from testing myself locally) that the setting gets automatically added and populated, but should we still have an explicit migration file?

I checked similar PRs from this summer and [one PR did have a migration](https://github.com/TryGhost/Ghost/pull/23725) ~but [another PR didn't](https://github.com/TryGhost/Ghost/pull/24344)~ (EDIT: see below).

If I had to guess from reading the migration guidelines i'd expect that we should have one, but maybe there's guidance somewhere that defines when it can be skipped that I couldn't find, like what's  ~in #24344 and~ currently in this PR

EDIT: actually, #24344 wasn't a migration - https://github.com/TryGhost/Ghost/issues/24340 was the migration for that, and #24344 was just finishing up the feature i think

troyciesco (approved):

PR #25034, opened 2025-10-01
Added Settings factory

9larsons:
no ref

It's a common pattern to need to be able to use labs flags, so this PR adds the settings factory following the existing patterns for interacting with Ghost APIs.

This one is notably a bit different because the settings API is always a bulk operation - fetching and sending all the settings at once.

In general, the expectation is to set a few key settings or labs flags for each test run.


ibalosh:
hey @9larsons since we don't use some of the factory logic anyways, it seems as good alternative to use this in simpler way which I like little bit more. Check it out here: https://github.com/TryGhost/Ghost/pull/25043 

If you like though this one more though, let me know. No strong feelings towards either direction.

9larsons:
Closing in lieu of a simpler implementation here: https://github.com/TryGhost/Ghost/pull/25043.

PR #25036, opened 2025-10-01
Improved email client in e2e package

9larsons:
no ref

Previous implementation of the Mailhog client was limited to raw functionality. This PR attempts to clean up the patterns used by introducing a few layers and separating the Mailhog package from being so tightly tied to the tests, or vice versa.


9larsons:
I'm not 100% on the directory structure here - I was debating putting it in `/helpers/environment` or creating a `services` directory much like we have `/data-factory/`, as I can imagine we may need other services like email down the road. Lmk what you're leaning towards!

ibalosh:
> I'm not 100% on the directory structure here - I was debating putting it in `/helpers/environment` or creating a `services` directory much like we have `/data-factory/`, as I can imagine we may need other services like email down the road. Lmk what you're leaning towards!

hey @9larsons , I was thinking about the approach introduced here. I do like the idea of having `services` directory, and we can go with something like `/services` at top e2e level, or `/helper/services` - I don't have a strong opinion about it.

Regarding introduction of adapter here, I think it's an optimization we don't need. I think we should keep it simple, and stay with the simple email client that you added first. With email client interface for it, we can change it to something else later if we would need. And with utils for parsing email, we are good too, we can further improve that if needed as we go. 

These are going to be separate services, which blend nicely with factories for creating data for tests, we do not need to mix them.

Adapter made sense for factories, since we were thinking of different ways to persist data (db, api). For email, it's too much, as we are not going to build a lot around email, other than search, delete, parsing.

**TLDR:** I would go with what we have, group it as services and use it as api wrapper like we do now.

9larsons:
Great. We've opted to scrap this in lieu of a `/services/` dir that we can put things like this that are light and simple implementations that are made to be more 'disposable' should need be.

PR #25040, opened 2025-10-02
Improved OTP input styles in Portal

peterzimon:
ref https://linear.app/ghost/issue/PROD-2706/improve-the-design-of-the-otc-input-original-component-not-input-otp

- The current styling of the OTP field in Portal was not following common visual and UX standards.

### DIFF
--- a/apps/portal/src/components/pages/MagicLinkPage.js
+++ b/apps/portal/src/components/pages/MagicLinkPage.js
@@ -192,7 +192,7 @@ export default class MagicLinkPage extends React.Component {
                     <div className={`gh-portal-otp-field-container ${errors.otc && 'error'}`}>
                         <input
                             id={`input-${OTC_FIELD_NAME}`}
-                            className='gh-portal-input'
+                            className={`gh-portal-input ${errors.otc && 'error'}`}
                             name={OTC_FIELD_NAME}
                             type="text"
                             value={this.state.otc}
--- a/apps/portal/src/components/pages/MagicLinkPage.js
+++ b/apps/portal/src/components/pages/MagicLinkPage.js
@@ -189,7 +189,6 @@ export default class MagicLinkPage extends React.Component {
         return (
             <form onSubmit={e => this.handleSubmit(e)}>
                 <section className='gh-portal-section gh-portal-otp'>
-                    {/* @TODO: create different input component with updated design */}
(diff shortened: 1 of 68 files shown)


apps/portal/src/components/pages/MagicLinkPage.js
```diff
@@ -180,20 +188,28 @@ export default class MagicLinkPage extends React.Component {
 
         return (
             <form onSubmit={e => this.handleSubmit(e)}>
-                <section className='gh-portal-section'>
+                <section className='gh-portal-section gh-portal-otp'>
                     {/* @TODO: create different input component with updated design */}
-                    <InputField
```
kevinansfield, line 185: If we're switching away from `<InputField>` it would be good to clean up the [OTC-related stuff here](https://github.com/TryGhost/Ghost/blob/main/apps/portal/src/components/common/InputField.js#L131-L139)

apps/portal/src/components/pages/MagicLinkPage.js
```diff
@@ -180,20 +188,28 @@ export default class MagicLinkPage extends React.Component {
 
         return (
             <form onSubmit={e => this.handleSubmit(e)}>
-                <section className='gh-portal-section'>
+                <section className='gh-portal-section gh-portal-otp'>
                     {/* @TODO: create different input component with updated design */}
```
kevinansfield: Can this TODO be removed with the new styling in place or do we want to keep it as a note to switch to a more full-featured component later on?

apps/portal/src/components/pages/MagicLinkPage.js
```diff
@@ -180,20 +188,28 @@ export default class MagicLinkPage extends React.Component {
...
+                        <input
+                            id={`input-${OTC_FIELD_NAME}`}
+                            className='gh-portal-input'
+                            name={OTC_FIELD_NAME}
+                            type="text"
+                            value={this.state.otc}
+                            placeholder="••••••"
+                            inputMode="numeric"
+                            pattern="[0-9]*"
+                            label={t('Code')}
```
kevinansfield: Should this be `aria-label`? Afaik `input` fields don't have a `label` attribute

kevinansfield: Changing that should fix the tests too 🙂

PR #25041, opened 2025-10-02
🎨 Changed "What's new" toast to be hidden for new users.

rob-ghost:
## Why this change?

When setting up a brand new Ghost site, the "What's new" featured banner appears in the sidebar during onboarding. This creates a confusing user experience because:
- Everything is new on a brand new site
- The banner distracts from the onboarding checklist
- Users haven't experienced the baseline yet, so "what's new" has no context

Ref https://linear.app/ghost/issue/DES-1155/hide-whats-new-for-brand-new-sites

## What does this change do?

This change-set sets the newly created User's "last seen" news date with the date that they were created, such that they'll only see "What's new" toasts when newer information is available.

Once a new update is available, the "What's new" toast re-appears, regardless of their onboarding status.

## Technical Implementation

This works by setting the `user.accessibility` column (which is where the `whatsNew.lastSeenDate` is stored) with a default value of today's date.

Formerly this column was `null` by default, and is updated by the frontend as the User changes their settings or dismisses the "What's new" toast. Now it will always have a value by default, which is why the snapshots had to be updated in all cases when there's a new User setup. This is a notable change in behaviour that needs to be considered in this review.

We intentionally made the smallest change that improves the onboarding experience. The current TEXT blob stored in user.accessibility is not ideal for typed/structured data, and we have documented a follow-up to revisit this pattern in more depth. See [ENG-2552](https://linear.app/ghost/issue/ENG-2552/refactor-usersaccessibility-json-blob-pattern)
 for options to refactor the JSON-blob approach and consider a more structured model for accessibility preferences and “What’s new” state.

## Comparison with the [Frontend approach](https://github.com/TryGhost/Ghost/pull/25014)

The benefits of setting this state in the backend for new Users is that we get the same effect of hiding the "What's new" toast without any need to change the frontend or manage any frontend state (such as triggering `whatsNew.seen()`). This has the additional benefit of simplifying the Ember -> React migration as its one less bit of logic in the Ember components which will have to be migrated to React.

The drawback of this approach is that now User's have a non-null `accessibility` value by default which requires many test snapshot updates and we have "What's new" logic baked into the User model logic which is coupling what was formerly a "user preferences" frontend concern with data model logic.

### DIFF
--- a/ghost/core/core/server/models/user.js
+++ b/ghost/core/core/server/models/user.js
@@ -1,5 +1,6 @@
 const validator = require('@tryghost/validator');
 const ObjectId = require('bson-objectid').default;
+const moment = require('moment-timezone');
 const ghostBookshelf = require('./base');
 const baseUtils = require('./base/utils');
 const limitService = require('../services/limits');
@@ -201,6 +202,19 @@ User = ghostBookshelf.Model.extend({
                     columnKey: 'name'
                 })
             });
+        }
+
+        // Initialize accessibility.whatsNew for new users
+        if (options.method === 'insert' && !options.importing) {
+            const existingValue = this.get('accessibility');
+            const accessibility = existingValue ? JSON.parse(existingValue) : {};
+
+            const today = moment.utc().startOf('day').toISOString();
+
+            accessibility.whatsNew ??= {};
+            accessibility.whatsNew.lastSeenDate ??= today;
+
+            this.set('accessibility', JSON.stringify(accessibility));
         }
 
         // If the user's email is set & has changed & we are not importing
--- a/ghost/core/test/e2e-api/admin/__snapshots__/members.test.js.snap
+++ b/ghost/core/test/e2e-api/admin/__snapshots__/members.test.js.snap
@@ -386,7 +386,7 @@ exports[`Members API - member attribution Returns sign up attributions of all ty
 Object {
   "access-control-allow-origin": "http://127.0.0.1:2369",
   "cache-control": "no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0",
-  "content-length": "8230",
+  "content-length": "8290",
   "content-type": "application/json; charset=utf-8",
   "content-version": StringMatching /v\\\\d\\+\\\\\\.\\\\d\\+/,
   "etag": StringMatching /\\(\\?:W\\\\/\\)\\?"\\(\\?:\\[ !#-\\\\x7E\\\\x80-\\\\xFF\\]\\*\\|\\\\r\\\\n\\[\\\\t \\]\\|\\\\\\\\\\.\\)\\*"/,
--- a/ghost/core/test/e2e-api/admin/__snapshots__/pages.test.js.snap
+++ b/ghost/core/test/e2e-api/admin/__snapshots__/pages.test.js.snap
@@ -647,7 +647,7 @@ exports[`Pages API Convert can convert a mobiledoc page to lexical 2: [headers] 
 Object {
   "access-control-allow-origin": "http://127.0.0.1:2369",
   "cache-control": "no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0",
-  "content-length": "4325",
+  "content-length": "4445",
   "content-type": "application/json; charset=utf-8",
   "content-version": StringMatching /v\\\\d\\+\\\\\\.\\\\d\\+/,
   "etag": StringMatching /\\(\\?:W\\\\/\\)\\?"\\(\\?:\\[ !#-\\\\x7E\\\\x80-\\\\xFF\\]\\*\\|\\\\r\\\\n\\[\\\\t \\]\\|\\\\\\\\\\.\\)\\*"/,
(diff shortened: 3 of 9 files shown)


rob-ghost:
@kevinansfield thank you for [your suggestion](https://github.com/TryGhost/Ghost/pull/25014#pullrequestreview-3289213555) on the [frontend PR](https://github.com/TryGhost/Ghost/pull/25014) which addressed this same problem from another angle.

This is what I'd consider the smallest appetite change to adding `user.accessibility.whatsNew.lastSeenDate` to the backend in the most appropriate place (when a new User is being added).

There's more potential refactors in this area (highlighted in [ENG-2552](https://linear.app/ghost/issue/ENG-2552/refactor-usersaccessibility-json-blob-pattern) about the `accessibility` column) but we don't have the appetite to address those concerns in the scope of this work.

My questions for you are:

1. Is this the approach you envisioned in your suggestion?
2. Are there alternative approaches you can think of that you'd prefer to see?
3. Are we comfortable with the scope of this change? (The backend being `whatsNew` aware?)
4. Is this preferable to the frontend approach?

ghost/core/core/server/models/user.js
```diff
@@ -203,6 +204,19 @@ User = ghostBookshelf.Model.extend({
             });
         }
 
+        // Initialize accessibility.whatsNew for new users
+        if (options.method === 'insert' && !options.importing) {
+            const existingValue = this.get('accessibility');
+            const accessibility = existingValue ? JSON.parse(existingValue) : {};
```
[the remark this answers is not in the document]
rob-ghost: Valid concern. Right now there's zero validation of the `accessibility` value stored in the database. The frontend handles it with `JSON.parse` currently.

I'd rather not be too defensive here, if `JSON.parse` fails I'd rather we get an error report rather than handling it gracefully in some way. If we're storing invalid JSON in this column it means we've done something wrong and we should be aware of it ASAP.

So I'm in favour of not being defensive here, but I'd love to hear another reviewer's opinion.

ghost/core/core/server/models/user.js
```diff
@@ -203,6 +204,19 @@ User = ghostBookshelf.Model.extend({
             });
         }
 
+        // Initialize accessibility.whatsNew for new users
+        if (options.method === 'insert' && !options.importing) {
```
kevinansfield, line 208: Was there a reason you opted for this approach compared to using the [defaults() method](https://github.com/TryGhost/Ghost/pull/25041/files#diff-4a35fea52df05aebbbdc1b5bf2fb8937e890f227c030003898de5e91d319358bR63-R77) that sets default data when inserting User models? I suppose we don't have the option of skipping the default value for imports there 🤔

kevinansfield:
> 1. Is this the approach you envisioned in your suggestion?

Yes, this is what I envisioned 👌 

> 2. Are there alternative approaches you can think of that you'd prefer to see?

Possibly only use of `defaults()` in place of `onSaving()` just so it's clearer, although it might come with trade-offs? [Discussion started here](https://github.com/TryGhost/Ghost/pull/25041#discussion_r2399331482)

> 3. Are we comfortable with the scope of this change? (The backend being whatsNew aware?)

Personally, yes. I think it's reasonable for the backend to set/be aware of default user settings, in my mind it's not much different to the various default site settings we have. "whatsNew" _is_ a bit of a stretch as a user setting, but we don't have any other place to store this data at the moment.

> 4. Is this preferable to the frontend approach?

I think it's cleaner to use a known start date rather than trying to infer the new-user status client-side. The client-side approach also has some edge cases from using the onboarding checklist to infer "new user" state because any user that's not an owner will never see the onboarding flow so would always see the last "what's new" changelog. This server-side approach doesn't suffer from the same issue.

kevinansfield:
> 2. Are there alternative approaches you can think of that you'd prefer to see?

Thinking some more, we could have the client set `whatsNew` to today's date if it's not already set. That would mean first load of the Admin interface for any user would skip the banner and not show it again until there is something new. Annoyed I didn't think of that as an alternative sooner but maybe it also has disadvantages I haven't thought of yet 🤔

rob-ghost:
> > 2. Are there alternative approaches you can think of that you'd prefer to see?
> 
> Thinking some more, we could have the client set `whatsNew` to today's date if it's not already set. That would mean first load of the Admin interface for any user would skip the banner and not show it again until there is something new. Annoyed I didn't think of that as an alternative sooner but maybe it also has disadvantages I haven't thought of yet 🤔

Great alternative!

I guess the problem is it assumes every user who hasn't interacted with `What's new` yet (and thus no `whatsNew.lastSeenDate`) is a "new" User (or not interested in `What's new`) and update the `whatsNew.lastSeenDate` meaning the User won't see the toast until the next update.

In practical terms, I think we reach eventual consistency where all new users have no `whatsNew.lastSeenDate` and all existing Users (who have at least visited Ghost admin) will have a `whatsNew.lastSeenDate`, but it's a little implicit and relies on clients to do work for data consistency.

On the other hand, it keeps all things `whatsNew` co-located in the FE service, which has its advantages.

I'll think on this alternative a bit more and see if anything else comes to mind!

ghost/core/core/server/models/user.js
rob-ghost: The short answer is, we set down the `onSaving` path and we didn't pause to consider there were other alternatives.

Longer answer is, because we weren't 100% certain of all the call sites of `User...add/save()` we couldn't be sure if `accessibility` was being passed with a value at any point. The code in `onSaving` handles a few different use cases of merging our default value of `lastSeenDate` but only if we have to, otherwise it accepts what's being passed.

After looking deeper into the use cases, I think `defaults()` does work for this use case as it stands today. There's a couple of things we need to keep in mind if we use `defaults()`:

1. **We'll never need to apply the default on updates** - defaults is ignored on updates unless we pass `defaults: true` on `save()`. Only when inserting a `User` for the first time will the defaults be applied.
2. **We'll never need the dynamic `today` date at save time vs model instantiation time** - `defaults()` runs a little sooner than the actual insert, but since we're (currently) dealing with day resolution of this date on insert it's not a concern for now.
3. **We'll never pass partial `accessibility` data that needs merging** - It's _possible_ to pass a value for `accessibility` when constructing the model which doesn't have `whatsNew.lastDateSeen`. Since `defaults` doesn't do merging, it will just detect that the attribute is present and the default will be ignored. So we'll silently regress this behaviour unknowingly if someone ever passes `accessibility` when inserting a User model.

So `onSaving` is over-engineered for today's use-case given how we're currently saving and editing the `User` model, but it handles more edge cases in the future and would prevent accidental regressions.

We can still choose `defaults()` if we're not worried about these future edge cases and we feel its a more pragmatic or idiomatic approach.

What do you think about these tradeoffs?

rob-ghost:
Closing in favour of the approach in https://github.com/TryGhost/Ghost/pull/25075

PR #25042, opened 2025-10-02
Added Portal build to E2E setup

9larsons:
ref https://github.com/TryGhost/Ghost/pull/24984
- added Portal build step to Docker setup
- added Portal container to E2E environment setup in order to test against local/branch build
- added Playwright trace logging

We need to be able to run tests against changes to Portal. Previously we fetched Portal from CDN and so we could never do this; this adds a build step and container to serve the Portal asset external to Ghost, much like it functions in production.


e2e/package.json
```diff
@@ -8,6 +8,7 @@
     "dev": "tsc --watch --preserveWatchOutput --noEmit",
     "build": "yarn test:types",
     "build:ts": "tsc --noEmit",
+    "docker:build:ghost": "cd .. && docker build -t ghost-monorepo:latest -f .docker/Dockerfile .",
```
9larsons, line 11: @cmraible I'd like to get your eyes on this. It seems we need this because the `yarn docker:build` at the top level doesn't build the same. My brain is a little mushy atm with this so would appreciate fresh eyes/someone with better knowledge of this area. 

This also ties into our other discussion re: the local e2e container setup being the same as local dev and not necessarily the same as CI.. which I suspect is *ok in most cases* and better devx. So maybe this command becomes `docker:build:ci` so we have that option when needed.

.github/workflows/ci-docker.yml
```diff
@@ -334,8 +334,22 @@ jobs:
       - name: Setup Playwright
         uses: ./.github/actions/setup-playwright
 
-
       - name: Run e2e tests
         env:
           GHOST_IMAGE_TAG: ${{ steps.load.outputs.image-tag }}
         run: yarn test:e2e
```
9larsons, line 340: This was added for troubleshooting, however it feels reasonable to add in general. Idk if we pulled it for some reason.

apps/portal/src/index.js
```diff
@@ -8,6 +8,7 @@ const ROOT_DIV_ID = 'ghost-portal-root';
 function addRootDiv() {
     const elem = document.createElement('div');
     elem.id = ROOT_DIV_ID;
+    elem.setAttribute('data-portal-docker-test', 'true');
```
9larsons: This is simply a flag to verify the build catches on CI. Arguably we could remove it now or later, though it's safe to ship as is.

.docker/Dockerfile
```diff
@@ -114,6 +114,15 @@ COPY --from=admin-x-framework-builder /home/ghost/apps/admin-x-framework/dist ap
...
 RUN cd apps/posts && yarn build
 
+# --------------------
+# Portal Builder
+# --------------------
+FROM development-base AS portal-builder
+WORKDIR /home/ghost
+COPY ghost/i18n ghost/i18n
+COPY apps/portal apps/portal
+RUN cd apps/portal && yarn build
```
9larsons, line 124: This is why the container was failing on CI. Without the build stage, we weren't creating the umd file to serve. Locally it very likely picked up your umd file from running `yarn dev` and copied it in to the volume.

ibalosh (approved):
Looks fine by me, maybe a bit nicer locator for portal, but for this I think it's fine.
Chris has better knowledge of our containers, so leaving final call to him

9larsons:
> maybe a bit nicer locator for portal, but for this I think it's fine.

Yeah, the intent is to delete this test file when we have a representative case. This is just to prove that CI is running it successfully.

.docker/Dockerfile
cmraible: Oh nice find!

.github/workflows/ci-docker.yml
cmraible: Definitely good to keep around 👍

apps/portal/src/index.js
cmraible: suggestion: I think we should keep it, but change it to `data-testid="portal-root"` for consistency

e2e/helpers/environment/EnvironmentManager.ts
```diff
@@ -55,7 +55,23 @@ export class EnvironmentManager {
...
+     */
+    private async getPortalUrl(): Promise<string> {
+        const portalContainer = await this.dockerCompose.getContainerForService('portal');
+        const containerInfo = await portalContainer.inspect();
+        const portMapping = containerInfo.NetworkSettings.Ports['4175/tcp'];
+        if (!portMapping || portMapping.length === 0) {
+            throw new Error('Portal container does not have port 4175 exposed');
+        }
+        const hostPort = portMapping[0].HostPort;
+        const portalUrl = `http://localhost:${hostPort}/portal.min.js`;
```
cmraible: suggestion (non-blocking): might be nice to extract `getHostPortForService('portal');` into the `DockerCompose` class

e2e/helpers/environment/GhostManager.ts
```diff
@@ -28,8 +29,8 @@ export class GhostManager {
     }
 
     /** High-level: create, wait, and return a GhostInstance description. */
-    async startInstance(instanceId: string, siteUuid: string): Promise<GhostInstance> {
-        const container = await this.createAndStart({instanceId, siteUuid});
+    async startInstance(instanceId: string, siteUuid: string, portalUrl?: string): Promise<GhostInstance> {
+        const container = await this.createAndStart({instanceId, siteUuid, portalUrl});
```
cmraible, line 33:
```suggestion
    async startInstance(instanceId: string, siteUuid: string, config: GhostConfig): Promise<GhostInstance> {
        const container = await this.createAndStart({instanceId, siteUuid, configl});
```

non-blocking: I think we should generalize `startInstance` to accept an arbitrary configuration object here instead of listing them out individually. This config object would then get passed to Ghost on boot via environment variable. This way we can pass in labs flags or arbitrary config (via the test fixture) to the Ghost instance we are starting.

As an added bonus, we can slowly start building out at least a typescript type to document Ghost's configuration.

e2e/tests/public/portal-docker-build.test.ts
```diff
@@ -0,0 +1,12 @@
+import {test, expect} from '../../helpers/playwright';
+import {HomePage} from '../../helpers/pages/public';
+
+test.describe('Ghost Public - Portal Docker Build Verification', () => {
+    test('portal loads with docker build changes - data attribute present', async ({page}) => {
```
cmraible: I think we should delete this test before merging (or update the descriptions if we want to keep them in)

e2e/tests/public/portal-docker-build.test.ts
```diff
@@ -0,0 +1,12 @@
...
+        const homePage = new HomePage(page);
+        await homePage.goto();
+
+        const portalRoot = page.locator('#ghost-portal-root');
```
cmraible: Moot point if we delete this test, but this might make sense as a locator on the base `PublicPage` POM, so this would be `homePage.portalRoot`

e2e/package.json
cmraible: I don't _think_ we should need this. The build of the docker container should be exactly the same. The only difference I _think_ is the bind mount:
- In the dev setup, your filesystem is mounted into the container, including any build artifacts (e.g. portal)
- In the e2e setup, there's no bind mount, so it uses the exact artifact from the docker build

I do think we need to fix this somehow, because it is at best annoying and is causing lots of confusion I think

cmraible (approved):
Nice work, this is awesome! Nice find on the portal build. Left some comments but nothing blocking 👍

e2e/tests/public/portal-docker-build.test.ts
9larsons: Removed it, yeah.

PR #25043, opened 2025-10-02
Extract settings service for better reusability

ibalosh:
Added basic Ghost client wrapper for updating settings

ref https://github.com/TryGhost/Ghost/pull/25034

The update is based on initial version by @9larsons , that used factories. Since for settings, all we need is just fetching and updating data, I decided to give it a try with approach based on update you did for Mailhog. Mailhog is even better example, since we don't use it in factory/persistence way, so it makes sense to be a service.

Since we use factories for building, I thought we could use alternative, and just reuse the HttpClient interface.
Same as for adapter, which seemed to have too much change to get it align.

It's very simple, so we can always change it to factory if we will need it in factory way, and to adapter.
Let me know your thoughts, I have no strong feelings for either, so if you like more the factory approach, we can use that too.


e2e/tests/admin/tags/list.test.ts
```diff
@@ -1,12 +1,10 @@
...
 import {TagsPage} from '../../../helpers/pages/admin';
-import {overrideFeatureFlags} from '../../../helpers/utils';
 import {mockTagsResponse} from './helpers/mock-tags-response';
 
 test.describe('Ghost Admin - Tags', () => {
-    test.beforeEach(async ({page}) => {
-        await overrideFeatureFlags(page, {tagsX: true});
-    });
+    // Set labs flags for all tests in this describe block
+    test.use({labs: {tagsX: true}});
```
ibalosh, line 7: nice touch @9larsons !

9larsons (commented):
Merged w/ my updates. All in all this is looking and feeling nice to work with.

PR #25056, opened 2025-10-06
WIP: Zod schema generation spike

jonatansberg:
# Add API Contracts Package with Auto-Generated Zod Schemas

## Overview

Introduces a new `@tryghost/api-contracts` package that provides type-safe API contracts with runtime validation using [Zod](https://zod.dev/). Schemas are **auto-generated** from the official [Ghost SDK JSON Schemas](https://github.com/TryGhost/SDK/tree/main/packages/admin-api-schema), ensuring they stay in sync with the canonical API validation rules.

## Motivation

- **Runtime Validation**: Catch API response mismatches immediately rather than at compile time
- **Single Source of Truth**: Generate schemas directly from Ghost SDK to avoid drift
- **Type Safety**: Automatic TypeScript type inference from schemas
- **Backend Compatible**: Zero React dependencies, can be used in Node.js
- **Framework Agnostic**: Works with any JavaScript/TypeScript project (React Query, SWR, etc.)

### Architecture

The package uses a two-file pattern for each resource:

1. **`*.generated.ts`** - Auto-generated, contains pure JSON Schema → Zod conversion
2. **`*.ts`** - Hand-crafted, composes generated schemas with response fields and custom logic

```
Ghost SDK JSON Schema
        ↓ (fetch + resolve $ref + convert)
    tags.generated.ts (TagsInputSchema)
        ↓ (import + compose)
    tags.ts (TagSchema, TagsResponseSchema, etc.)
```

### Generation Pipeline

1. Fetches JSON Schemas from Ghost SDK repository
2. Resolves `$ref` references using `json-refs`
3. Converts to Zod using `json-schema-to-zod`
4. Generates minimal `.generated.ts` files with input schemas only
5. Hand-crafted `.ts` files compose schemas with response-only fields

### Example Usage

```typescript
import { TagsResponseSchema, type TagsResponse } from '@tryghost/api-contracts/tags';

// Validate API response
const response = await fetch('/ghost/api/admin/tags/');
const data = await response.json();
const validated = TagsResponseSchema.parse(data); // Throws if invalid

// With React Query
useQuery<TagsResponse>({
    queryKey: ['tags'],
    queryFn: async () => {
        const res = await fetch('/ghost/api/admin/tags/');
        return TagsResponseSchema.parse(await res.json());
    },
});
```

## Key Features

- ✅ All validation rules preserved (maxLength, patterns, enums, etc.)
- ✅ Uses `snake_case` field names matching Ghost API
- ✅ Tree-shakeable exports (import only what you need)
- ✅ Both ESM and CJS builds
- ✅ Full TypeScript support with auto-generated types
- ✅ Scalable: Easy to add new resources (posts, members, etc.)

## Adding New Resources

1. Add resource name to `RESOURCES` array in `scripts/generate-schemas.ts`
2. Run `yarn generate`
3. Create `src/{resource}.ts` to compose schemas (see `tags.ts` as template)
4. Export from `src/index.ts`

The generator handles all the heavy lifting - fetching, resolving, and converting schemas.

## Package Structure

```
apps/api-contracts/
├── scripts/
│   └── generate-schemas.ts      # Auto-generation script
├── src/
│   ├── common.ts                # Shared schemas (Pagination, Meta, etc.)
│   ├── tags.generated.ts        # AUTO-GENERATED (37 lines)
│   ├── tags.ts                  # HAND-CRAFTED (88 lines)
│   └── index.ts                 # Public exports
├── package.json
├── tsconfig.json
└── README.md                    # Full documentation
```

## Testing

The package includes comprehensive tests. Run with `yarn test` in the package directory.

## Next Steps

This PR includes tags implementation as the first resource. Follow-up PRs can add:
- Posts schemas
- Members schemas
- Users schemas
- Pages schemas
- etc.

Each resource follows the same pattern and can be added independently.

## Migration Path

The new package doesn't replace existing types immediately. It can be adopted incrementally:

1. Start using in new code (e.g., React Query hooks)
2. Add runtime validation to critical API calls
3. Gradually migrate existing API clients
4. Eventually use as single source of truth across frontend/backend

---

**Note**: All field names use `snake_case` to match the Ghost API exactly (e.g., `meta_title`, not `metaTitle`). This ensures schemas validate real API responses without transformation.


rob-ghost:
**Question:** How do we feel about writing tests for schemas?

I think for basic schemas using Zod's built-in validators probably don't need any tests. We can trust that Zod will behave as expected.

There's some cases where we might have to add some custom validation for a particular schema, something like "X is number except when Y < 0", in which case we'll want a test for the new behaviour that we've added.

But as a general rule, if we're using the out-of-the-box validators, I don't think we'll need any tests. If we do have generic custom validators, we can test those validators so we don't need to add the same tests to every schema which uses this custom validator.

I think the ideal tests for these schemas is actually validating if there's any breaking changes to an existing schema definition as we update them. I don't think Zod has these but there are tools for OpenAPI and JSON Schema to ensure there's no breaking changes introduced to existing schemas. This might fit better as a CI check than manual tests.

What do you think?

---

To expand as a bit of a testing meta point (inspired by, but un-related to this proposal), I don't think tests for entrypoints / APIs for example which fuzz the input are all that useful. Ideally if the APIs use the validation trait which validates the input against a defined schema (and react to failures accordingly), then that trait should already be generically tested and we can assume that if an API has that trait, it will behave as expected.

At best we could pass our API to an existing parameterised suite which ensures it returns the appropriate responses on success / failure so we don't have to keep writing the same `it('returns a validation error on invalid input')` for every API. That kind of policy should cut down a lot on the amount of tests we have to write.

If we want to globally ensure our endpoints are tolerant of invalid input then we can have a separate fuzzing suite which sends invalid data to the entire API and we simply expect it to always respond with the appropriate error response.

jonatansberg:
Great comment @rob-ghost, I agree in principle. 

I think there might a point to testing the schemas against snapshots of real responses until they are enforced by the backend. That way we can verify that we're actually describing "reality" in some sense.

Maybe we could integrate that with the existing endpoint snapshots rather than maintaining a separate test suite 🤔

rob-ghost:
Yeah I love that idea, having a corpus of real world request bodies that we can validate against the schemas in CI.

jonatansberg:
I will close this as stale and then we can take another look at this once we're ready.

PR #25068, opened 2025-10-06
Added stats utm endpoints with fixture data

9larsons:
ref https://linear.app/ghost/issue/PROD-2717/
- add new utm endpoint, /stats/utm-growth, accepting a param for the utm return field
- added fixture data for new endpoint until fully wired up

This set of changes adds a new endpoint, /stats/utm-growth, that returns fixture data until we're able to fully wire up the members service to the backend. This effectively unlocks the frontend design from being dependent on the backend support.

The stats endpoint is intended to be used like `/stats/utm-growth?post_id={id}&utm_type=utm_source`. Post is optional; the only required field is utm_type.

### DIFF
--- a/ghost/core/core/server/api/endpoints/stats.js
+++ b/ghost/core/core/server/api/endpoints/stats.js
@@ -514,7 +514,7 @@ const controller = {
         },
         options: [
             'order',
-            'limit', 
+            'limit',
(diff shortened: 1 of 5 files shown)


cmraible (approved):
Nice, LGTM!

PR #25071, opened 2025-10-06
Fixed flaky mysql healthcheck in e2e test suite

cmraible:
The e2e tests were randomly failing with the following error, indicating an issue in the global setup:

```
ERROR Failed to setup test database: Command failed with exit code 1: mysql -uroot -proot -e "UPDATE \`ghost_b3dc0bf5-8ac4-4696-983b-5e092ad6be49\`.settings SET value='b3dc0bf5-8ac4-4696-983b-5e092ad6be49' WHERE \`key\`='site_uuid';"
```
Having reproduced this once locally, we found that the `ghost-migrations` container was logging an `ECONNREFUSED` error when trying to connect to mysql on port 3306 over the docker network.

This makes two small changes to address the failing migrations, and to recover in the event that they do fail:
- changed mysql healthcheck to use a TCP connection rather than a local unix socket, to confirm that mysql is actually ready to accept TCP connections from other containers
- added a restart-on-failure policy to the `ghost-migrations` service, so it will retry up to 5 times if it encounters any errors while running migrations

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -29,3 +29,5 @@ compose.yml
 !.docker/**/*entrypoint.sh
 
 ghost/core/core/built/admin
+
+ghost/core/config.local.json*
--- a/.dockerignore
+++ b/.dockerignore
@@ -29,5 +29,3 @@ compose.yml
 !.docker/**/*entrypoint.sh
 
 ghost/core/core/built/admin
-
-ghost/core/config.local.json*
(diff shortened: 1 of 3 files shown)


cmraible:
@ibalosh @9larsons I'm going to merge this one too, just a tiny change to the compose file for e2e tests.

PR #25072, opened 2025-10-06
Fixed waiting logic in e2e test global setup

cmraible:
In the e2e global setup, we have some custom logic to wait for all docker compose services to:
- be in a healthy state, if they have a health check configured
- exit with code 0 otherwise

There was an error in this check causing it to only check containers that are running, thus ignoring any containers which may have exited with code 1. This fixes that, so if a container exits with code 1, the global setup will fail and output a much clearer error message describing what went wrong.


cmraible:
@ibalosh @9larsons I'm going to go ahead and merge this one without review since it's such a tiny change.

PR #25075, opened 2025-10-06
🎨 Improved What's New to hide updates for new users

rob-ghost:
## Why this change?

When new users first access Ghost admin, they see notifications for changelog entries published before they created their account. This creates confusion because they're being notified about updates to a baseline they never experienced - everything is new to them.

Ref https://linear.app/ghost/issue/DES-1155

## What does this change do?

Automatically sets a default `lastSeenDate` for new users on their first visit to the admin app. This ensures they only receive notifications for updates published after they start using Ghost, rather than seeing notifications for every historical changelog entry.

**Technical implementation:**

When the `whatsNew` service loads for the first time and detects that a user doesn't have a `lastSeenDate` in their accessibility settings, it now automatically sets it to the current date (in UTC ISO format) and persists this to the user model. This happens during the initial `fetchLatest` call using a read-merge-write pattern to preserve other accessibility settings.

## Why is this needed?

New users shouldn't be overwhelmed with notifications about changes they never experienced. This provides eventual consistency for the What's New feature across all users while ensuring notifications remain relevant and contextual.

## Implementation approach

This PR follows a test-driven development approach with three commits:

1. **Test coverage** - Added unit and acceptance tests for the What's New feature, establishing a test baseline for the service
2. **Refactor** - Simplified the service to track only the `whatsNew` slice of user accessibility settings rather than the entire object
3. **Feature** - Implemented automatic default `lastSeenDate` initialization for new users with tests

### DIFF
--- a/ghost/admin/app/components/modals/whats-new.hbs
+++ b/ghost/admin/app/components/modals/whats-new.hbs
@@ -1,17 +1,17 @@
-<div class="modal-content">
-    <h1 class="gh-whasnew-modal-title">What’s new?</h1>
-    <section class="gh-whatsnew-modal-entries" {{did-insert (perform this.whatsNew.updateLastSeen)}}>
+<div class="modal-content" data-test-modal="whats-new">
+    <h1 class="gh-whasnew-modal-title" data-test-modal-title>What’s new?</h1>
+    <section class="gh-whatsnew-modal-entries" data-test-entries {{did-insert (perform this.whatsNew.updateLastSeen)}}>
         {{#each this.whatsNew.entries as |entry|}}
-            <a class="gh-whatsnew-modal-entry" href={{entry.url}} target="_blank" rel="noopener noreferrer">
+            <a class="gh-whatsnew-modal-entry" data-test-entry href={{entry.url}} target="_blank" rel="noopener noreferrer">
                 {{#if entry.feature_image}}
-                    <img class="gh-whatsnew-modal-entry-featureimage" src={{entry.feature_image}} alt={{entry.title}}>
+                    <img class="gh-whatsnew-modal-entry-featureimage" data-test-entry-image src={{entry.feature_image}} alt={{entry.title}}>
                 {{/if}}
                 <div class="gh-whatsnew-modal-entrycontent">
-                    <h2>{{entry.title}}</h2>
+                    <h2 data-test-entry-title>{{entry.title}}</h2>
                     {{#if entry.custom_excerpt}}
-                        <p>{{entry.custom_excerpt}}</p>
+                        <p data-test-entry-excerpt>{{entry.custom_excerpt}}</p>
                     {{/if}}
-                    <span>{{moment-format entry.published_at "DD MMMM YYYY"}}</span>
+                    <span data-test-entry-date>{{moment-format entry.published_at "DD MMMM YYYY"}}</span>
                 </div>
             </a>
         {{/each}}
--- a/ghost/admin/app/services/whats-new.js
+++ b/ghost/admin/app/services/whats-new.js
@@ -7,34 +7,32 @@ import {task} from 'ember-concurrency';
 
 export default Service.extend({
     session: service(),
-    store: service(),
-    response: null,
 
     entries: null,
     changelogUrl: 'https://ghost.org/blog/',
     isShowingModal: false,
 
-    _user: null,
+    // We only want to request the changelog once, so we store the initial
+    // response so we don't request it again.
+    _changelog_response: null,
+
+    // Track only the whatsNew slice of user.accessibility settings
+    _whatsNewSettings: null,
 
     init() {
         this._super(...arguments);
         this.entries = [];
+        this._whatsNewSettings = {};
     },
 
-    whatsNewSettings: computed('_user.accessibility', function () {
-        let settingsJson = this.get('_user.accessibility') || '{}';
-        let settings = JSON.parse(settingsJson);
-        return settings.whatsNew;
-    }),
-
-    hasNew: computed('whatsNewSettings.lastSeenDate', 'entries.[]', function () {
+    hasNew: computed('_whatsNewSettings.lastSeenDate', 'entries.[]', function () {
         if (isEmpty(this.entries)) {
             return false;
         }
 
         let [latestEntry] = this.entries;
 
-        let lastSeenDate = this.get('whatsNewSettings.lastSeenDate') || '2019-01-01 00:00:00';
+        let lastSeenDate = this._whatsNewSettings?.lastSeenDate || '2019-01-01 00:00:00';
         let lastSeenMoment = moment(lastSeenDate);
         let latestDate = latestEntry.published_at;
         let latestMoment = moment(latestDate || lastSeenDate);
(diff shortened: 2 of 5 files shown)


rob-ghost:
👋🏻 @kevinansfield @jonatansberg

Here's the 3rd attempt at this feature (as proposed [here](https://github.com/TryGhost/Ghost/pull/25041#issuecomment-3362111605)).

I think out of the three options ([first the FE depending on onboarding state](https://github.com/TryGhost/Ghost/pull/25014), [second defaulting in the backend](https://github.com/TryGhost/Ghost/pull/25041)) I think this approach is my preference.

It's got low coupling between `whats-new` and other components / services / layers, and high cohesion within the service. It keeps all the `whats-new` concerns co-located in one service.

The drawback compared to the backend approach is the upcoming Ember -> React migration, and adding some more code to be ported, however it will be much easier to port given everything is located in one place.

---

The first two commits add both unit tests for the `whats-new` service, and acceptance tests for the various ways of viewing or interacting with the "What's new" notification and modals (including adding `data-test-*` attributes to the templates), as well as a refactor of the service to simplify its state management.

The third commit would be easiest to review, since its the commit which adds the defaulting of the `lastSeenDate` to the `whatsNew` accessibility state.

ghost/admin/app/services/whats-new.js
```diff
@@ -64,45 +62,59 @@ export default Service.extend({
...
+            ...this._whatsNewSettings,
+            lastSeenDate: latestEntry.published_at
+        });
+
+        // Persist using read-merge-write pattern
+        let user = yield this.session.user;
+        let accessibility = JSON.parse(user.accessibility || '{}');
+        accessibility.whatsNew = this._whatsNewSettings;
+        user.set('accessibility', JSON.stringify(accessibility));
+        yield user.save();
```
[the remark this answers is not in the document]
rob-ghost: The spread is expressly so `hasNew` will re-compute since the value of the attribute has changed and set with `this.set`, this is a false-flag.

Regarding `session.user` being stale, sure, its possible that someone else has updated the `user.accessibility` and for `session.user` not to be aware of it, but I think relying on `session.user` as a source of truth is a good pattern for services and the concern of it going stale should be handled by the `session` service.

ghost/admin/app/services/whats-new.js
```diff
@@ -64,45 +62,59 @@ export default Service.extend({
...
+                yield user.save();
             }
+
+            this.set('_whatsNewSettings', whatsNewSettings);
+
+            this._changelog_response = yield fetch('https://ghost.org/changelog.json');
+            if (!this._changelog_response.ok) {
+                // eslint-disable-next-line
+                return console.error('Failed to fetch changelog', this._changelog_response);
+            }
```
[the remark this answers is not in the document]
rob-ghost: We don't have any retry mechanism, this either works the first time, or it doesn't. Its true that this pattern _prevents_ a retry if we were to add one, but we can always adjust if that time comes.

ghost/admin/app/components/modals/whats-new.hbs
```diff
@@ -1,17 +1,17 @@
-<div class="modal-content">
-    <h1 class="gh-whasnew-modal-title">What’s new?</h1>
```
kevinansfield, line 2: Nitpick: Why were the curly apostrophes swapped here?

rob-ghost: Great spot! You have eagle eyes 👀

Claude ran away with itself when it couldn't get the tests to reliably pass with the apostrophes and it replaced them and I missed it.

I ran into this problem on the other PR too. The solution is to update the test to handle it and not update the template. I'll fix this.

ghost/admin/app/components/modals/whats-new.hbs
```diff
@@ -1,17 +1,17 @@
...
-    <section class="gh-whatsnew-modal-entries" {{did-insert (perform this.whatsNew.updateLastSeen)}}>
+<div class="modal-content" data-test-whats-new-modal>
```
kevinansfield: Nitpick: For common types of component we tend to use test id naming like `data-test-modal="whats-new"`.

Not really a big deal, I've just found it can give us minor benefits such as being able to use `findAll('[data-test-modal]').length` to assert that we're not showing multiple modals at once. Plus it helps keep test id ordering more consistent (e.g. having `data-test-modal-whats-new` ordering in some places vs `data-test-whats-new-modal` ordering in others).

ghost/admin/tests/acceptance/whats-new-test.js
```diff
@@ -0,0 +1,358 @@
...
+
+        beforeEach(function () {
+            toast = new WhatsNewToast();
+        });
+
+        it('shows toast with latest entry when there are new featured entries', async function () {
+            mockChangelog(this.server, [ENTRIES.newFeatured]);
+
+            await visit('/site');
+            await settled();
```
kevinansfield: We shouldn't need `settled()` in almost all cases, the ember test helpers call this internally before resolving so we can clean these extraneous uses up 🙂 If we find they are needed to make a test pass it suggests we have a deeper issue somewhere.

ghost/admin/app/components/modals/whats-new.hbs
rob-ghost: Much prefer this myself 👍🏻 changed the `data-test` attributes and references in the page objects.

kevinansfield (approved):

ghost/admin/tests/acceptance/whats-new-test.js
rob-ghost: Nice, they're gone!

jonatansberg (commented):
I don't have much to add in addition to Kevin's review of the fix. The approach looks solid! 👍 

I have some meta commentary on adding 800 lines of new tests for whats supposedly a minor bugfix. That makes me wonder if we might have gone a little overboard with the testing?

What I mean by that is, does all of the tests add value and make this code easier to maintain in the future, to port to React eventually? We're testing behaviors we care about? Do some of these tests add a future maintenance burden without telling us anything that either the unit tests or the acceptance tests would catch anyhow? 

Take the test on error handling for example. To me, the behavior I would like to verify is that we fail gracefully if something blows up. Exactly if and how we log to the console is rather inconsequential. 

Similarly, tests for the `_whatsNewSettings` state maybe aren't all that desirable? The underscore naming suggests we're testing what was at some point meant to be an implementation detail. 

Make of these observations what you will. The meta meta discussion probably ought to be around whether or not it's worthwhile to invest more time into this particular thing at all :)

rob-ghost:
Thanks @jonatansberg, really good points.

Regarding adding so many tests, the `whats-new` and associated components were not tested at all. So when I came to add the new functionality, I could either add a single suite for _just_ my new changes, or I could ensure the service and interactivity was well covered. I felt we'd be better off if the functionality was tested, and that would give me confidence to make the refactors I'd want to make.

The other reason to want to have this coverage was, when implementing the improvement I also wanted to do some tidy-up refactors to the service to simplify it. In order to do that safely, I needed better test coverage. Specifically acceptance tests, because refactors were breaking the Handlebars templates and it was impossible to tell purely from the service unit tests.

Regarding the contents of the tests (testing outcomes over implementations, `console.error` and `_whatsNewSettings` for example), I agree completely and that's generally how I'd write tests. I'll go back and clean those up now.

> What I mean by that is, does all of the tests add value and make this code easier to maintain in the future, to port to React eventually? We're testing behaviours we care about? Do some of these tests add a future maintenance burden without telling us anything that either the unit tests or the acceptance tests would catch anyhow?

It will be a bit of both.

Yes it helps with overall maintenance and the future port to React. This feature was untested before, so at least we can have a bit of confidence it still works after porting the acceptance tests to the new E2E approach when available. The page objects can help inform the shape of the React components for example.

There are probably some tests which are not required or might be brittle, I'll do another pass on them now to verify. Overall though, I think a brittle test can always be deleted, but I'd rather have a suite with some brittle tests than no suite at all!

> The meta meta discussion probably ought to be around whether or not it's worthwhile to invest more time into this particular thing at all :)

This is a kind of unique situation where its not just about fixing a small bug but getting familiar with the codebase and the review / release / rollout process. We're ok with spending a bit more time than usual on this first PR to explore these things in the future we wouldn't have invested so much time into such an issue 😉

It's also been a good design exercise and discussion to explore this `accessibility` thing and a service which depends on it. I'll probably bring this up in a wider discussion later about coupling and cohesion and the benefits / tradeoffs of each approach we explored.

Thanks for the feedback 🙏🏻 I'll get to doing another pass on the tests right away!

jonatansberg:
It's all good. I'm sorry if my previous comment came off as a bit grumpy 😅 

It's definetly valuable to add tests since there hasn't been any. It also makes sense to invest time in getting familiar with things, and I think we should continue to make room for that as we go 😊

rob-ghost:
Not grumpy at all! They're strong points so I just wanted to give context as to why the approach might have been considered overboard in some cases (which is definitely is!)

rob-ghost:
@jonatansberg I did a pass on the tests and change the acceptance tests to be pure behaviour of the toast and modal, so it doesn't know anything about the fact that its driven by `lastSeenDate` or anything else, just that the toast appears (or doesn't), and disappears (even after refreshes).

I also removed the suite for the internal `_whatsNewSettings` and adjusted the graceful error handling tests to be purely about error handling.

I toyed with removing the badge acceptance tests but I think they're interesting to keep.

I'm happy that the tests now are just testing outcomes and they're not tightly coupled with the implementation of the service.

ghost/admin/app/services/whats-new.js
```diff
@@ -64,45 +63,61 @@ export default Service.extend({
...
+            let accessibility = JSON.parse(user.accessibility || '{}');
+            let whatsNewSettings = accessibility.whatsNew;
+
+            if (!whatsNewSettings?.lastSeenDate) {
+                // New user - persist the defaults from init()
+                whatsNewSettings = this._whatsNewSettings;
+                accessibility.whatsNew = whatsNewSettings;
+                user.set('accessibility', JSON.stringify(accessibility));
+                yield user.save();
             }
```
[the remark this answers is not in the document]
rob-ghost: We're aware of this trade off, but having `this._whatsNewSettings` in the `init` isn't a concern.

PR #25078, opened 2025-10-07
Added e2e tests for page hits with UTM params

9larsons:
ref https://linear.app/ghost/issue/PROD-2686/
- added page hit tests with UTMs in query params
- added helpers

This PR adds a few e2e tests for page hits with UTM params in the url.

### DIFF
--- a/e2e/helpers/pages/admin/analytics/AnalyticsWebTrafficPage.ts
+++ b/e2e/helpers/pages/admin/analytics/AnalyticsWebTrafficPage.ts
@@ -12,6 +12,8 @@ export class AnalyticsWebTrafficPage extends AdminPage {
     readonly pagesButton: Locator;
 
     public readonly topSourcesCard: Locator;
+    public readonly sourcesTab: Locator;
+    public readonly campaignsDropdown: Locator;
 
     constructor(page: Page) {
         super(page);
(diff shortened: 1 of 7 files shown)


cmraible (changes requested):
I've pushed a fix for the current page hit test, which has been rather flaky in CI: https://github.com/TryGhost/Ghost/pull/25108

We should rebase this and use the same pattern there, which should 🤞 fix the flakiness here.

e2e/tests/admin/analytics/utm-tracking.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        });
+    });
+
+    test.describe('utmTracking flag enabled', () => {
+        test.use({labs: {utmTracking: true}});
+
+        test('displays utm_source data correctly', async ({page, browser, baseURL}) => {
+            await withIsolatedPage(browser, {baseURL}, async ({page: publicPage}) => {
+                const homePage = new HomePage(publicPage);
+                await homePage.gotoWithUTMParams({
```
ibalosh: why not call it `gotoWithQueryParams` , that way we can use it for other calls later too

e2e/helpers/pages/admin/analytics/AnalyticsWebTrafficPage.ts
```diff
@@ -28,6 +30,28 @@ export class AnalyticsWebTrafficPage extends AdminPage {
...
+        await this.page.getByRole('menuitem', {name: campaignType}).click();
+        await this.page.waitForLoadState('networkidle');
+    }
+
+    /**
+     * Wait for Tinybird to process analytics data
+     * Tinybird's materialized views need time to process incoming events
+     */
+    async waitForTinybirdProcessing(): Promise<void> {
+        await this.page.waitForTimeout(5000);
```
ibalosh: wondering if there is a different way to do this wait, I saw @cmraible doing something like this in [this PR](https://github.com/TryGhost/Ghost/pull/25108/files) , to wait for request.

Wondering if we can catch API response here too.

We can leave this wait in only if it's absolutely necessary, these tend to give false positives fast

ibalosh (changes requested):
Hey @9larsons 

very nice update! I did not approve it only due that waiForTimeout, hoping we can have better solution.

Other than that, ready to go

cmraible:
@ibalosh would you mind giving this another pass? I was able to fix the flakiness and eliminate the use of `waitForTimeout` and `networkidle`

cmraible (approved):

ibalosh (approved):

PR #25079, opened 2025-10-07
Added UTM fields to Ghost member attribution service

9larsons:
ref https://linear.app/ghost/issue/PROD-2693/update-member-attribution-service-to-pass-through-utm-data
- updated `AttributionBuilder`, `ReferrerTranslator` classes to pass through UTM fields

Since adding the UTM fields to the frontend scripts (member-attribution and ghost-stats), we need to update the backend services to actual utilize the fields. This service is used by the `MemberBREADService`, which lies between Stripe and the Events.

### DIFF
--- a/ghost/core/core/server/services/member-attribution/ReferrerTranslator.js
+++ b/ghost/core/core/server/services/member-attribution/ReferrerTranslator.js
@@ -3,6 +3,11 @@
  * @prop {string|null} [referrerSource]
  * @prop {string|null} [referrerMedium]
  * @prop {string|null} [referrerUrl]
+ * @prop {string|null} [utmSource]
+ * @prop {string|null} [utmMedium]
+ * @prop {string|null} [utmCampaign]
+ * @prop {string|null} [utmTerm]
+ * @prop {string|null} [utmContent]
  */
 
 const {ReferrerParser} = require('@tryghost/referrer-parser');
@@ -36,8 +41,35 @@ class ReferrerTranslator {
             return {
                 referrerSource: null,
                 referrerMedium: null,
-                referrerUrl: null
+                referrerUrl: null,
+                utmSource: null,
+                utmMedium: null,
+                utmCampaign: null,
+                utmTerm: null,
+                utmContent: null
             };
+        }
+
+        // Look for UTM parameters first (most recent entry with UTM data)
+        let utmData = {
+            utmSource: null,
+            utmMedium: null,
+            utmCampaign: null,
+            utmTerm: null,
+            utmContent: null
+        };
+
+        for (const item of history) {
+            if (item.utmSource || item.utmMedium || item.utmCampaign || item.utmTerm || item.utmContent) {
+                utmData = {
+                    utmSource: item.utmSource || null,
+                    utmMedium: item.utmMedium || null,
+                    utmCampaign: item.utmCampaign || null,
+                    utmTerm: item.utmTerm || null,
+                    utmContent: item.utmContent || null
+                };
+                break;
+            }
         }
 
         for (const item of history) {
@@ -53,7 +85,8 @@ class ReferrerTranslator {
                 return {
                     referrerSource,
                     referrerMedium,
-                    referrerUrl
+                    referrerUrl,
+                    ...utmData
                 };
             }
         }
@@ -61,7 +94,8 @@ class ReferrerTranslator {
         return {
             referrerSource: 'Direct',
             referrerMedium: null,
-            referrerUrl: null
+            referrerUrl: null,
+            ...utmData
         };
     }
 
--- a/ghost/core/core/server/services/member-attribution/ReferrerTranslator.js
+++ b/ghost/core/core/server/services/member-attribution/ReferrerTranslator.js
@@ -50,7 +50,9 @@ class ReferrerTranslator {
             };
         }
 
-        // Look for UTM parameters first (most recent entry with UTM data)
+        // Look for UTM parameters (earliest entry with UTM data)
+        // Note: history is ordered newest-to-oldest, so we want the LAST match
+        // This captures the original campaign source rather than subsequent navigations
         let utmData = {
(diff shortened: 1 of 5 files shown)


ghost/core/core/server/services/member-attribution/ReferrerTranslator.js
```diff
@@ -36,10 +41,37 @@ class ReferrerTranslator {
...
+                utmData = {
+                    utmSource: item.utmSource || null,
+                    utmMedium: item.utmMedium || null,
+                    utmCampaign: item.utmCampaign || null,
+                    utmTerm: item.utmTerm || null,
+                    utmContent: item.utmContent || null
+                };
+                break;
+            }
+        }
```
cmraible, line 75: Curious about this second loop - what's the intended behavior here? 

- First loop finds the most recent entry with at least one utm parameter
- Second loop finds (and returns) the most recent entry with at least one of referrerSource, referrerMedium, referrerUrl

In a scenario where:
- A user lands on the homepage, with all the utm parameters set
- User navigates around the site (i.e. visits a few posts)
- User signs up for a free membership

This would take the UTM parameters from the initial entry (landing on the homepage), but the referrerSource, medium, and URL of the most recent post they visited.

I'm not sure that's wrong — in fact it kind of makes sense — I'm just not 100% sure it's right either 😅

ghost/core/test/unit/server/services/member-attribution/referrer-translator.test.js
```diff
@@ -273,7 +328,183 @@ describe('ReferrerTranslator', function () {
...
+                    referrerUrl: 't.co',
+                    utmSource: 'instagram',
+                    utmMedium: null,
+                    utmCampaign: 'summer_promo',
+                    utmTerm: null,
+                    utmContent: null
+                });
+            });
+
+            it('uses first entry with UTM data when multiple entries have UTM', async function () {
```
cmraible:
```suggestion
            it('uses most recent entry with UTM data when multiple entries have UTM', async function () {
```

ghost/core/test/unit/server/services/member-attribution/referrer-translator.test.js
```diff
@@ -273,7 +328,183 @@ describe('ReferrerTranslator', function () {
...
+                utmSource: null,
+                utmMedium: null,
+                utmCampaign: null,
+                utmTerm: null,
+                utmContent: null
+            });
+        });
+
+        describe('UTM parameter extraction', function () {
+            it('extracts all UTM parameters from first history entry with UTM data', async function () {
```
cmraible:
```suggestion
            it('extracts all UTM parameters from most recent history entry with UTM data', async function () {
```

"first" feels ambiguous and potentially misleading — its first entry in the list, but not first entry recorded by the user.

ghost/core/core/server/services/member-attribution/ReferrerTranslator.js
```diff
@@ -36,10 +41,37 @@ class ReferrerTranslator {
...
+                referrerUrl: null,
+                utmSource: null,
+                utmMedium: null,
+                utmCampaign: null,
+                utmTerm: null,
+                utmContent: null
             };
         }
 
+        // Look for UTM parameters first (most recent entry with UTM data)
```
cmraible: Do we want the most recent entry with UTM data, or the first (earliest) entry with UTM data?

ghost/core/test/unit/server/services/member-attribution/referrer-translator.test.js
```diff
@@ -273,7 +328,183 @@ describe('ReferrerTranslator', function () {
...
+                    referrerUrl: null,
+                    utmSource: null,
+                    utmMedium: null,
+                    utmCampaign: null,
+                    utmTerm: null,
+                    utmContent: null
+                });
+            });
+
+            it('extracts UTM from later entry when earlier entries have no UTM', async function () {
```
cmraible: Isn't the first entry here the most recent? Should this be "extracts UTM from earlier entry when more recent entries have no UTM"?

cmraible (commented):
I'm a little unsure of the intended behavior, particularly WRT which history entry we want to take the UTM parameters from, if more than one with UTM parameters exist.

I think we should confirm the intended behavior, then maybe reword some comments/test descriptions to make this more clear.

ghost/core/core/server/services/member-attribution/ReferrerTranslator.js
9larsons: You're right! First/earliest entry for UTMs makes sense. I do think it's correct to separate - the 'content that got them to sign up' should be the most recent, while 'the content (campaign) that got them to Ghost' should be the earliest.

9larsons: Almost sounded like Claude there... 😬.

ghost/core/test/unit/server/services/member-attribution/referrer-translator.test.js
```diff
@@ -273,7 +328,183 @@ describe('ReferrerTranslator', function () {
...
+                utmSource: null,
+                utmMedium: null,
+                utmCampaign: null,
+                utmTerm: null,
+                utmContent: null
+            });
+        });
+
+        describe('UTM parameter extraction', function () {
+            it('extracts all UTM parameters from most recent history entry with UTM data', async function () {
```
cmraible:
```suggestion
            it('extracts all UTM parameters from earliest history entry with UTM data', async function () {
```
Just for consistency since we landed on using the earliest instead of the most recent

cmraible: Arguably we could just make this test use a single entry, and call it "extracts all UTM parameters", since the have the "uses earliest entry with UTM data when multiple entries have UTM" case covered below

cmraible (approved):

PR #25080, opened 2025-10-07
Added UTM fields to Stripe checkout metadata

9larsons:
ref https://linear.app/ghost/issue/PROD-2879/
- added UTM fields to Stripe metadata
- updated Stripe tests to ensure metadata supports UTMs

As the UTM fields are handled just like our attribution fields (it's a form of attribution, or closely related) we need to update Stripe to handle passing these data around. Currently, we handle this in the `session.metadata`, so this PR adds those entries using the same patterns.


cmraible (approved):
Nice, LGTM!

PR #25088, opened 2025-10-08
Added UTM parameter storage to member events

9larsons:
ref https://linear.app/ghost/issue/PROD-2880/
ref https://github.com/TryGhost/Ghost/commit/2ea046c3e88aafea6d8544aafde9e5f473cc5012
- updated event storage to pass along UTM fields to the Ghost database

The Ghost db contains utm_* columns (see migration above) in various events tables. This PR wires up the data flow so that the real data coming from the frontend scripts actually reaches the database.

### DIFF
--- a/ghost/core/core/server/services/donations/DonationBookshelfRepository.ts
+++ b/ghost/core/core/server/services/donations/DonationBookshelfRepository.ts
@@ -23,6 +23,11 @@ type DonationEventModelInstance = BookshelfModelInstance & {
     referrer_source: string | null;
     referrer_medium: string | null;
     referrer_url: string | null;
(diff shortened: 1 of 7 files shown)


cmraible (changes requested):
Nice one! Nothing to note on the code itself 🎉

One thing on behavior though: I'm not seeing the UTM parameters for donation events make it into the database. I'm not sure where we're dropping them yet, but I don't see them in the donation_payment_events table.

9larsons:
D'oh, my bad. I thought I pushed all my commits.. was sitting on one still.

cmraible (approved):

PR #25092, opened 2025-10-08
Updated filter style on React tags list

peterzimon:
ref https://linear.app/ghost/issue/DES-1216/use-toggle-group-in-tag-list-filter

- Changed the component to a toggle group to filter public/internal tags on the tags list for better UX and design parity with the Ember version.

### DIFF
--- a/apps/posts/src/views/Tags/components/TagsHeader.tsx
+++ b/apps/posts/src/views/Tags/components/TagsHeader.tsx
@@ -18,7 +18,7 @@ const TagsHeader: React.FC<TagsHeaderProps> = ({currentTab}) => {
                 <ToggleGroup data-testid="tags-header-tabs" size='button' type="single" value={currentTab} onValueChange={(newTab) => {
                     navigate(newTab === 'internal' ? '/tags?type=internal' : '/tags');
                 }}>
-                    <ToggleGroupItem aria-label="Internal tags" value="public">
+                    <ToggleGroupItem aria-label="Public tags" value="public">
                         Public tags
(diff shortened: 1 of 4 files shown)


apps/posts/src/views/Tags/components/TagsHeader.tsx
```diff
@@ -1,26 +1,30 @@
...
-                    </PageMenuItem>
-                    <PageMenuItem value="internal" asChild>
-                        <Link to="/tags?type=internal">Internal tags</Link>
-                    </PageMenuItem>
-                </PageMenu>
+                <ToggleGroup data-testid="tags-header-tabs" size='button' type="single" value={currentTab} onValueChange={(newTab) => {
+                    navigate(newTab === 'internal' ? '/tags?type=internal' : '/tags');
+                }}>
+                    <ToggleGroupItem aria-label="Public tags" value="public">
+                        Public tags
```
jonatansberg: Can't we use `asChild` here to render proper links? That would let us get rid of that `navigate` call and wont break e.g. command-click to open in a new tab and whatnot. 

It's arguably also a little bit more accessible for screen readers.

jonatansberg (changes requested):
I'm not a huge fan of that controlled navigation. I'd rather have this render links where possible.

apps/posts/src/views/Tags/components/TagsHeader.tsx
peterzimon: Agreed, fixed it

jonatansberg (approved):

PR #25093, opened 2025-10-08
Updated page header in Shade

peterzimon:
ref https://linear.app/ghost/issue/DES-1217/refine-header-spacing-and-alignment

- The vertical spacing and action button alignment was not consistent with the rest of the Admin on the tags list screen (React, using Shade).

### DIFF
--- a/apps/shade/src/components/layout/header.tsx
+++ b/apps/shade/src/components/layout/header.tsx
@@ -22,7 +22,7 @@ function HeaderTitle({className, children}: HeaderTitleProps) {
     return (
         <H1
             className={cn(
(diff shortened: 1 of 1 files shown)


jonatansberg (commented):
I think this looks good. 

One nitpick: When you have a title, no subtitle, and a page menu on a narrow screen, then the spacing is a bit off. 

Image

peterzimon:
CleanShot 2025-10-08 at 16 48 34@2x

good catch, @jonatansberg . Fixed it

jonatansberg (approved):

PR #25097, opened 2025-10-08
Fetch post count when editing tags

jonatansberg:
Refs https://linear.app/ghost/issue/PROD-2711/port-existing-ember-tests-to-playwright

The post count wont show in the tag deletion confirmation modal if you load the tag editor directly instead of going via the tag list. Ensuring we include the counts for the model retrieval of individual tags resolves this.

Without this in place, the [new E2E tests for the tag editor](https://github.com/TryGhost/Ghost/pull/25098) fail.

### DIFF
--- a/ghost/admin/app/routes/tag.js
+++ b/ghost/admin/app/routes/tag.js
@@ -25,7 +25,7 @@ export default class TagRoute extends AuthenticatedRoute {
         this._requiresBackgroundRefresh = false;
 
         if (params.tag_slug) {
-            return this.store.queryRecord('tag', {slug: params.tag_slug});
+            return this.store.queryRecord('tag', {slug: params.tag_slug, include: 'count.posts'});
         } else {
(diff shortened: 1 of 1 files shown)


kevinansfield (approved):
Coderabbit raised [a potentially useful comment here](https://github.com/TryGhost/Ghost/pull/25097/files#r2413622810) with the suggestion to replace the `tag.reload()` call so we're not losing the count again with a partial response from the API. _However_, I don't think that code path will be reachable with the  React tags list as it won't be passing model instances around, the only time it can be reached is from new->edit redirect on save but in that case there won't be any associated posts anyway.

I was wondering if it's worth adding the include at the adapter level but that would increase the performance cost of all the other tag browse requests within Admin that don't need the counts. So, this seems like the best option to me ✅

jonatansberg:
Yeah, I just saw that comment too. I'll take a closer look and see if there's something that needs addressing.

PR #25098, opened 2025-10-08
Port Ember tag editor acceptance tests

jonatansberg:
Refs https://linear.app/ghost/issue/PROD-2711/port-existing-ember-tests-to-playwright

Port the tests that cover user interaction for the tag editor and behavior that spans the tag list and editor. Tests for access control are still left in the Ember acceptance tests until we have the appropriate helpers around users and roles in the new E2E suite.

**Note**: These tests are needed in order to safely remove some of the old acceptance tests in Ember. And we need to to that before we can fully release the new tag list. Since this is effectively blocking that release, I'm hoping we can address any suggestions and feedback through quick successive follow-up PRs rather than having that hold up this from getting merged.


e2e/tests/admin/tags/editor.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        await expect(page.getByText('Page not found')).toBeVisible();
+    });
+
+    test('maintains active state in nav menu when creating a new tag', async ({page}) => {
+        const tagEditor = new TagEditorPage(page);
+
+        await tagEditor.gotoNew();
+
+        await expect(page).toHaveURL('/ghost/#/tags/new');
+        await expect(tagEditor.navMenuItem).toHaveClass(/active/);
```
kevinansfield, line 120: Not a blocker but it feels a bit strange to me that `TagEditorPage` knows about the main nav menu and that we're checking for a class. My gut reaction is that I'd prefer to see something like `expect(adminNavigation.activeItem).toHaveText('Tags')`

jonatansberg: That's completely fair. That was ported more or less one-to-one from the old Ember tests.

kevinansfield (approved):
Cool to see some examples of how our Ember tests can be ported, certainly helps solidify the Ember->React migration process 🙌

I don't see anything here that I'd consider a blocker to merging (once that final trailing-slash assertion failure is fixed 😅)

e2e/helpers/pages/admin/tags/TagEditorPage.ts
```diff
@@ -0,0 +1,86 @@
...
+    readonly deleteModalConfirmButton: Locator;
+
+    readonly navMenuItem: Locator;
+
+    constructor(page: Page) {
+        super(page);
+
+        this.pageUrl = '/ghost/#/tags';
+        
+        this.nameInput = page.locator('[data-test-input="tag-name"]');
```
ibalosh, line 23: any reason these are not test ids, so `page.getByTestId` can be used?

e2e/helpers/pages/admin/tags/TagEditorPage.ts
```diff
@@ -0,0 +1,86 @@
...
+
+        this.deleteModal = page.locator('[data-test-modal="confirm-delete-tag"]');
+        this.deleteModalPostsCount = this.deleteModal.locator('[data-test-text="posts-count"]');
+        this.deleteModalConfirmButton = this.deleteModal.locator('[data-test-button="confirm"]');
+
+        this.navMenuItem = page.locator('[data-test-nav="tags"]');
+    }
+
+    async gotoNew() {
+        await this.page.goto('/ghost/#/tags/new');
```
ibalosh, line 38: this visit seems to not belong in this page object, we could end up with bunch of gotoXY methods. That url will be available in NewTagPage page object, this one should not know about it.

That visit might as well be done in test with something like `newTagsPage.goto()`

unless this is the newTagsPage (but then, it should be just goto, you can see how default pom visit is implemented in base page, it seems like a mix of edit and new

e2e/helpers/pages/admin/tags/TagEditorPage.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    async createTag(name: string, slug: string) {
+        await this.page.goto('/ghost/#/tags/new');
```
ibalosh, line 42: I created tag details page, which is pretty much the same from what I see as edit page, you can borrow logic from there, or probably just extend edit/new from same page that shares the same data - would need littile bit of change to current pom I added before, but could work

e2e/helpers/pages/admin/tags/TagEditorPage.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    async deleteTag() {
+        await this.deleteButton.click();
+    }
+
+    async confirmDelete() {
+        await this.deleteModalConfirmButton.click();
+    }
+
+    async goBackToTagsList() {
```
ibalosh, line 76: we haven't used still this pattern yet -> that returns another POM, any reason for this one, does it read better than doing a click and then create object in test?

e2e/tests/admin/tags/editor.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        await tagEditor.fillTagName('Edited Tag Name');
+        await tagEditor.save();
+        await tagEditor.goBackToTagsList();
+
+        await expect(tagsPage.tagListRow).toHaveCount(1);
+        await expect(tagsPage.getRowByTitle('Edited Tag Name')).toBeVisible();
+    });
+
+    test('can delete tag without posts', async ({page}) => {
+        const tagEditor = new TagEditorPage(page);
```
ibalosh, line 62: is it and editor or just an edit page? I do realize it's not a page in SPA, but `tagsPage`, `newTagsPage`, `editTagsPage` seem more consistent to me, and feels like what I see on a page. Editor feels like something else

e2e/tests/admin/tags/editor.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        const tagsPage = await tagEditor.goBackToTagsList();
+
+        await tagsPage.getRowByTitle('To be deleted').click();
+        await tagEditor.deleteTag();
+
+        await expect(tagEditor.deleteModal).toBeVisible();
+        await tagEditor.confirmDelete();
+
+        await expect(tagEditor.deleteModal).not.toBeVisible();
+        await expect(page).toHaveURL('/ghost/#/tags');
```
ibalosh, line 76: implementation details, we should have zero of these, in test itself

ibalosh (changes requested):
Hey @jonatansberg 

I took a quick look, bit late here sorry I might have missed some of the logic during the review. I will see to do one more look tomorrow

e2e/helpers/pages/admin/tags/TagEditorPage.ts
kevinansfield: This is an Admin rendered screen, Admin/Ember uses `data-test-*` rather than React's `data-testid`. I expect this will change when the edit screens are migrated to React.

ibalosh: ah gotcha, makes sense, thanks @kevinansfield !

e2e/helpers/pages/admin/tags/TagEditorPage.ts
jonatansberg: I can see you merged something after I created this PR. I can go ahead and combine the two. 

Wouldn't TagEditorPage be better aligned with what the other POMs are called tho?

e2e/helpers/pages/admin/tags/TagEditorPage.ts
jonatansberg: There isn't a NewTagPage page object though? That's the same as the TagEditorPage and you get there by visiting that specific URL.

I prefer a method like this over introducing another page object that inherits from the TagEditorPage only to override the URL. Less indirection that way.

e2e/helpers/pages/admin/tags/TagEditorPage.ts
jonatansberg: I think I read that this was a recommended pattern in one of the references for the page object ADR, but I can't remember where now 🤔 

I'd be happy to remove it as I can't say I'm too fond of it either.

e2e/tests/admin/tags/editor.test.ts
jonatansberg: Which part are you not agreeing with? The assertion, or the fact that we're not referencing the URL on the tag list page, rather than having it in plain text here? 

I think this assertion makes sense in that we want to make sure we've actually navigated away from what now would be the empty tag editor.

e2e/tests/admin/tags/editor.test.ts
jonatansberg: From the point of view of the test, the tag editor could be a modal or a split pane view or something else. It wouldn't really matter. 

I like `tagEditor` since it's shorter than `tagEditorPage` and still gets the point across. If you feel strongly about this, I don't mind changing it tho.

jonatansberg:
> I don't see anything here that I'd consider a blocker to merging (once that final trailing-slash assertion failure is fixed 😅)

@kevinansfield That test failure is strange. I don't get the trailing slash when running the tests locally nor in the test case above that uses the same method for navigation. Something must be different with how we're running the tests in CI...

Anyhow, that assertion is kinda redundant, so I'll remove it.

e2e/helpers/pages/admin/tags/TagEditorPage.ts
ibalosh: I don't have a strong opinion about this one, I see Steve used editor on Post, I kind a like more the logic `tags`, `editTags`, `newTags` , but it's a nitpick

e2e/helpers/pages/admin/tags/TagEditorPage.ts
ibalosh: I see, yeah, little bit less indirection, but it comes with a price.. the page object is not really describing just the editor here in this case.

Edit tags page has delete tag, while new tags page doesn't have it, edit page has one url, while new page has another. Newpage should not know about deleting. In a way, it's kinda like breaking little bit the single responsibility rule, but in terms of screen/page.

Divergence of this type tends to grow. I agree, inherited page adds bit more code, but at least the page intentions are clear.

Just a thought, but up to you

e2e/tests/admin/tags/editor.test.ts
ibalosh: no worries, I try to always look at it from customers side, from there it's just a page to me, but def no strong feelings regarding this

e2e/tests/admin/tags/editor.test.ts
ibalosh: We should never test for urls in tests, it's an implementation detail, developer oriented and just like xpath, it can change. 

If you want to check whether you landed on correct page, you can always verify that some of the element on page is there, or title of the page is correct etc. It also makes the test clear, to outsider reader of the test, this url doesn't mean a lot, but having assertion that we landed on tags page or some other provides value.

ibalosh (approved):

e2e/helpers/pages/admin/tags/TagEditorPage.ts
jonatansberg: Lets align and address this (if needed) separately then

e2e/helpers/pages/admin/tags/TagEditorPage.ts
jonatansberg: The moment you save the "new" page becomes the editor tho, so I don't think it's that clear cut. 

It might also be that I come from more of a functional background and I'm really not a huge fan of multi level inheritance and the indirection that it causes. 

I'll leave things as is for now, but let's align on this along with the naming discussion separately.

jonatansberg:
Thanks for the reviews, @kevinansfield and @ibalosh. I'll get this merged so we have the foundations in place for releasing the tag list, and then I'll follow up with some improvements to address the quirks that was ported over from Ember once that's out in the wild.

e2e/tests/admin/tags/editor.test.ts
jonatansberg: I don't have a strong aversion against URLs, I think in many ways they're part of the user interface. But I also don't have any issues with asserting the navigation some other way. 

I'll add a follow-up task to address this.

kevinansfield: > We should never test for urls in tests, it's an implementation detail

I'd have to disagree with that too. URLs are the separator for web applications compared to other types of apps, they enable users to bookmark and share application state and are therefore core to the _users_ experience. SPA's can get into a situation where their internal state is not correctly represented in the URL, breaking that web-experience contract with the user. If we're not allowed to assert that URLs are what we expect them to be, aren't we  missing the ability to test that contract?

ibalosh: Hm, I am not sure I follow. 

What differentiates checking for URL format from whether user is on the right page in the app, with proper content? Users use the app through web interface, they just care that they see a proper content on the page. With url, we are not doing that check. With page content we are, plus we are covering not just that we landed on proper page, but also that the content is there,  two things at once.

Why would it matter if this url is now `/ghost/#/tags` and later maybe `/ghost/#/tagsList` , if the functionality -> what the output (web page content) is correct - which assures url worked. Only the internals changed (url), nothing that user would see. Unless I am missing something here, and it's important to let users know that there is an url change?

If the url has changed internally the test will brake anyways, since it verifies whether it's on a proper page with content, but then we are not going to be updating the test, but code where the change (proper new url) was not added. Checking for url in my opinion makes tests more fragile. The more tests, the more hard coded value checks which don't seem to be needed.

jonatansberg: I think you both have valid points. Asserting that the correct page actually shows is a stronger assertion than just looking at the URL.<br><br>But I agree with Kevin that the URL isn't an implementation detail. It's the contract between the app and the rest of the world as to where things live. It's how you reference individual screens from e.g. a help articles, allow users to save bookmarks and come back in the future, and so on.

kevinansfield: > if the functionality -> what the output (web page content) is correct - which assures url worked

This assumption is only true in a fully server-rendered world.

SPA's remove that fundamental concept of URL==page content by replacing the browsers URL handling and there's no guarantee the app's internal state has been correctly written back to the URL. It's why so often (though thankfully not so much any more) you have web apps where refreshing or sharing links doesn't work how you'd expect. Asserting that the app has correctly written its state back to the URL such that a new load will enable its internal state to be re-derived as expected is really valuable.

I think asserting URLs will be especially valuable during the Ember/React migration because we'll have competing router implementations.

PR #25100, opened 2025-10-08
Refactored What's New initialization to centralize defaults

rob-ghost:
ref https://linear.app/ghost/issue/DES-1155

Moved default lastSeenDate initialization from fetchLatest to `init()`, making the service state predictable from creation and simplifying the initialization logic.

This improves maintainability by ensuring the service is always in a valid state, eliminates defensive fallback code, and makes the separation of concerns clearer: `init()` handles setup, `fetchLatest()` handles data loading/syncing, and `updateLastSeen()` handles user interactions.

### DIFF
--- a/ghost/admin/app/services/whats-new.js
+++ b/ghost/admin/app/services/whats-new.js
@@ -22,7 +22,10 @@ export default Service.extend({
     init() {
         this._super(...arguments);
         this.entries = [];
-        this._whatsNewSettings = {};
+        // Set sensible default for new users - they've "seen" everything up to now
+        this._whatsNewSettings = {
+            lastSeenDate: moment.utc().toISOString()
+        };
     },
 
     hasNew: computed('_whatsNewSettings.lastSeenDate', 'entries.[]', function () {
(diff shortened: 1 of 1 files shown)


rob-ghost:
Follow-up to your feedback from earlier @allouis.

This PR initialises the `_whatsNewSettings` to a default value and removes all the conditionals and defaults from `hasNew` which makes the whole thing a lot clearer and removes any ambiguity over what the default value is.

It does have the issue of `lastSeenDate` being initialised when the service is, rather than when the User was fetched, but I think that's ok since we generally fetch the user immediately after the session is created which is close to when the service is initialised.

I did attempt to merge with `updateLastSeenDate` but it wasn't as clean as I'd like and coupled the need to be able to update the `lastSeenDate` to the latest entry on demand, and when to do it when we're defaulting. It's possible, but it wasn't clean since it required some knowledge of this initial "initialising" state and when we're doing it as a response to a user action. I opted to leave it as it is as a design compromise. There's duplication, but with duplication comes decoupling.

CC @kevinansfield @jonatansberg as the reviewers of the base PR.

PR #25102, opened 2025-10-08
Added a way to run tests on multiple shards in CI

ibalosh:
* Updated Github actions E2E workflow, to run tests o shards
* This reduces CI Docker run run from around 12m to 8m


.github/workflows/ci-docker.yml
```diff
@@ -286,9 +286,14 @@ jobs:
...
   test_e2e:
-    name: E2E Tests
+    name: E2E Tests (Shard ${{ matrix.shardIndex }}/${{ matrix.shardTotal }})
     runs-on: ubuntu-latest
     needs: [build, setup]
+    strategy:
+      fail-fast: false
+      matrix:
+        shardIndex: [1, 2, 3, 4]
+        shardTotal: [4]
```
cmraible: suggestion (non-blocking): have you tried with just 2 or 3 shards? Given that most of the time is spent on setup, we might see a similar benefit with a lower number of shards, which could potentially be more cost efficient.

- No sharding: 1 runner x 12 mins = 12 mins real time, 12 mins usage
- 2 shards: ??
- 3 shards: ??
- 4 shards: 4 runners x 8 mins = 8 mins real time, 32 mins usage

I don't think cost is a huge concern, just throwing it out there 😃

.github/workflows/ci-docker.yml
```diff
@@ -352,4 +357,4 @@ jobs:
       - name: Output command to view report
         if: failure()
         run: |
-          echo "::notice::To view the Playwright report locally, run: gh run download ${{ github.run_id }} -n playwright-report -D /tmp/playwright-\$\$ && npx playwright show-report /tmp/playwright-\$\$/playwright-report"
+          echo "::notice::To view the Playwright report locally, run: gh run download ${{ github.run_id }} -n playwright-report-${{ matrix.shardIndex }} -D /tmp/playwright-\$\$ && npx playwright show-report /tmp/playwright-\$\$/playwright-report"
```
cmraible, line 360: suggestion (non-blocking): might be nice to merge the test reports so we can download a single report with the failures across all shards, but that's a nice-to-have from my POV

cmraible (approved):

.github/workflows/ci-docker.yml
ibalosh: I ran these, and saw that even though 4 shards e2e tests are faster, it's not that big of a difference on compared to 2.

this is just the e2e run step, in CI docker:

* 4 -> 6:30 min 
* 3 -> 7 min 
* 2 -> 8 min 

I switched it to just use 2 shards for now

PR #25106, opened 2025-10-08
Added wrapper script to cleanup environment if tests are interrupted

cmraible:
no ref

Currently if you interrupt the e2e test suite with ctrl+c while the tests are running, the global teardown is skipped, leaving all the docker containers still running. These containers then need to be manually cleaned up, or they can interfere with future test runs — this is really annoying!

Playwright itself doesn't seem to have any built in way to enforce that the `global.teardown.ts` file, which normally does the cleanup, is run when the process is interrupted. Instead, this introduces a wrapper script which registers the signal handlers, then runs `playwright test`, passing any CLI args to playwright.


cmraible:
I don't love the wrapper script idea, but it was the best I could come up with 😞 

Unfortunately it doesn't seem that playwright has a native way to hook into the test runner's process, so I couldn't find any other reliable way to attach event listeners for the signal handling. 

This is one thing that testcontainers handles automatically, so if we could figure out how to get test containers working, we'd effectively get this cleanup for free.

e2e/helpers/environment/GhostManager.ts
```diff
@@ -165,4 +165,61 @@ export class GhostManager {
...
     }
+
+    /** Remove all Ghost test containers (used during cleanup to remove orphaned containers). */
+    async removeAll(): Promise<void> {
+        try {
+            debug('Checking for Ghost test containers...');
+
+            // Retry a few times to catch any containers that might be created during cleanup
+            for (let attempt = 0; attempt < 3; attempt++) {
+                const containers = await this.docker.listContainers({
```
9larsons, line 176: From a readability standpoint I wouldn't mind having some of these dumped into helper fns like `getE2EContainers` here, and the loop below to force rm containers.

9larsons (approved):
I'm not bullish on the helper fns. This looks good and works well! I wish it could have a cleaner exit to a new line entry but that's a terminal issue not the script.

I don't see a better solution for this. I think the alternatives would be a package to manage the process, but that's ultimately the same thing here and we both gain a dependency and lose control/knowledge.

ibalosh:
hey @cmraible I am probably missing something, but was wondering why we can't just on global setup, do docker compose down or something, before starting up.

we can check if cleanup is needed by even something simple and ugly like creating a file as flag that test ran but did not finish

cmraible:
> we can check if cleanup is needed by even something simple and ugly like creating a file as flag that test ran but did not finish

This is a much better idea, I'm going to do this instead.

PR #25107, opened 2025-10-09
Added AGENTS.md and CLAUDE.md

cmraible:
no refs

Adds initial draft of `AGENTS.md` file, and adds `CLAUDE.md` as a symlink to keep them in sync. Using symlinks this way should keep the instructions in sync across different agents (claude, codex, gemini, etc).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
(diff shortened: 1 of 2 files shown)


CLAUDE.md
```diff
@@ -0,0 +1 @@
+AGENTS.md
```
cmraible, line 1: This looks weird in Github's diff, but this indicates that this file is a symlink to AGENTS.md.

ibalosh: yep, appeared the same way in E2E

PR #25108, opened 2025-10-09
Fixed flaky page hit E2E test

cmraible:
The "Analytics > Overview > records a visitor when homepage is visited" E2E test has been flaky in CI, timing out waiting for the page hit response to be received when visiting the homepage.

The flakiness was caused by a race condition: it's possible that the page hit request could be completed before the promise to listen for the request is created. This leads to the test timing out waiting for a request that has already happened.

This fixes it by creating the promise _before_ navigating to the homepage, _then_ waiting for the promise to resolve. It does so in a helper function on the `PublicPage` POM so we can reuse this logic for any tests the depend on the page hit request, and avoid making this mistake again in the future.

### DIFF
--- a/.github/workflows/ci-docker.yml
+++ b/.github/workflows/ci-docker.yml
@@ -334,10 +334,22 @@ jobs:
       - name: Setup Playwright
         uses: ./.github/actions/setup-playwright
 
+      - name: Check for repeat label
+        id: check-repeat
+        if: github.event_name == 'pull_request'
+        run: |
+          LABELS='${{ toJson(github.event.pull_request.labels.*.name) }}'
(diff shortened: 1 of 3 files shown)


cmraible:
I went ahead and merged this one without review, because it's been failing quite frequently in CI which is getting annoying. Happy to make any follow up changes if you guys have feedback or suggestions :slightly_smiling_face:

ibalosh:
great update @cmraible only thing I would change in future, is make it more generic, so that it's usable in other cases too, like `waitForRequest` with url and method of API, instead of just one for page hit, but we can do that as we get more cases of these.

Nice work!

PR #25110, opened 2025-10-09
Updated Tinybird deployment jobs to only run on pushes to main

cmraible:
There is a bug in our Tinybird deploy workflow in CI that allows it to run on commits that have already merged to `main`, potentially leading to stale code being deployed to Tinybird.

This can happen because merged PRs can re-run their workflows in response to e.g. label added events, and since the code is already on the `main` branch, our "is this running on the main branch?" check evaluates to true.

This prevents this from happening by explicitly only running on _pushes_ to the main branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1172,7 +1172,7 @@ jobs:
       job_setup,
       job_tinybird-tests
     ]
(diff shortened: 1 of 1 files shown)


9larsons (approved):

9larsons:
GH kept 500ing with a comment on that review. Thanks for remembering this one!

PR #25111, opened 2025-10-09
Updated `/stats/utm-growth/` endpoint to return real data

cmraible:
ref https://linear.app/ghost/issue/PROD-2881/update-statsutm-growth-to-use-real-data

The `stats/utm-growth/` endpoint currently responds with hard coded fixture data. Now that we are recording the UTM parameters in the events tables, this updates the `stats/utm-growth` endpoint to return actual data.


ghost/core/core/server/services/stats/ReferrersStatsService.js
```diff
@@ -458,18 +458,164 @@ class ReferrersStatsService {
...
+                    // This ensures we don't count conversions that happened outside our date range
+                    .andOn('msce.created_at', '>=', knex.raw('?', [startDateTime]))
+                    .andOn('msce.created_at', '<=', knex.raw('?', [endDateTime]));
+            })
+            .whereNull('msce.id')
+            .whereNotNull(`mce.${utmField}`)
+            .groupBy(`mce.${utmField}`);
+
+        // Filter mce.created_at: only include members created within the date range
+        applyDateFilter(freeSignupsQuery, startDateTime, endDateTime, 'mce.created_at');
```
9larsons, line 496: Filtering only on the created events will be problematic. We need to check for both paid subs and free signups separately, then deduplicate. We've had to do this elsewhere, so I'll look at adapting those here.

ghost/core/core/server/services/stats/ReferrersStatsService.js
```diff
@@ -458,18 +458,164 @@ class ReferrersStatsService {
...
+
+        // Apply post filtering if post_id is provided
+        if (postId) {
+            freeSignupsQuery
+                .where('mce.attribution_id', postId)
+                .where('mce.attribution_type', 'post');
```
9larsons, line 502: While this looks reasonable, note to self to check on indexes - this might be unnecessary if we have an index on postId and not the type.

ghost/core/core/server/services/stats/ReferrersStatsService.js
```diff
@@ -458,18 +458,164 @@ class ReferrersStatsService {
...
+                message: `Invalid UTM field: ${utmField}. Must be one of: ${validUtmFields.join(', ')}`
+            });
+        }
+
+        const knex = this.knex;
+        const {dateFrom: startDateTime, dateTo: endDateTime} = getDateBoundaries(options);
+        const {post_id: postId} = options;
+
+        // Query: Free members who haven't converted to paid within the same time window
+        const freeSignupsQuery = knex('members_created_events as mce')
```
9larsons, line 481: We aren't deduplicating counts. If a free member also becomes a paid member in the reporting period, we only count them as a paid.

ghost/core/core/server/services/stats/ReferrersStatsService.js
```diff
@@ -458,18 +458,164 @@ class ReferrersStatsService {
...
+            .select(knex.raw(`mce.${utmField} as utm_value`))
+            .select(knex.raw('COUNT(DISTINCT mce.member_id) as signups'))
+            .leftJoin('members_subscription_created_events as msce', function () {
+                this.on('mce.member_id', '=', 'msce.member_id')
+                    // Filter msce.created_at: only count conversions within the same time window
```
9larsons, line 486: Oh yeah, signups and paid conversions could have different UTM params, so we definitely need to separate queries and deduplicate/combine.

ghost/core/core/server/services/stats/ReferrersStatsService.js
9larsons: I guess in theory we could have an id shared across multiple types, however unlikely. The index is indeed just on the id so I expect this will be good.

9larsons (approved):
Forget the above - I think we're ok. I think this behavior is appropriate, though I'd like to call it out: 

- if a member signs up as a free member and converts at any point later, we ONLY report the paid signup

I think there's opportunity to centralize some of this data fetching for members_created_events + other events so that we don't have multiple patterns for fetching the data and deduplicating - that's what threw me at first, because this approach is (afaict) actually the same but styled differently from `fetchMemberCountsBySource`. I don't think either is 'right' and I'd like to see us unify behavior where we can.

PR #25118, opened 2025-10-09
Moved invalid OTC error inline

kevinansfield:
ref https://linear.app/ghost/issue/PROD-2706

When an invalid OTC was entered, we were showing a notification popup that was inconsistent with the inline validation error design and had spatial dissociation with the form the user is interacting with. Moving error display inline required a method of surfacing errors from within dispatched in a way that allowed the calling-components to access them.

- added `actionErrorMessage` to App context
  - actions should return this state property rather than `popupNotification` when inline error display is desired
  - when set, should be a plain string
- updated `verifyOTC` action to set `actionError` instead of `popupNotification`
- updated `MagicLinkPage` to use `actionError` to display an inline error state on verification failure to match error design when no code was entered
- refactored `HumanReadableError.fromApiResponse` so it always returns an instance or `undefined` to match the declared return type

### DIFF
--- a/apps/portal/src/App.js
+++ b/apps/portal/src/App.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import * as Sentry from '@sentry/react';
-import i18n from './utils/i18n';
+import i18n, {t} from './utils/i18n';
+import {chooseBestErrorMessage} from './utils/errors';
 import TriggerButton from './components/TriggerButton';
 import Notification from './components/Notification';
 import PopupModal from './components/PopupModal';
@@ -53,6 +54,7 @@ export default class App extends React.Component {
             page: 'loading',
             showPopup: false,
             action: 'init:running',
+            actionErrorMessage: null,
             initStatus: 'running',
             lastPage: null,
             customSiteUrl: props.customSiteUrl,
@@ -683,7 +685,8 @@ export default class App extends React.Component {
     async dispatchAction(action, data) {
         clearTimeout(this.timeoutId);
         this.setState({
-            action: `${action}:running`
+            action: `${action}:running`,
+            actionErrorMessage: null
         });
         try {
             const updatedState = await ActionHandler({action, data, state: this.state, api: this.GhostApi});
@@ -714,6 +717,7 @@ export default class App extends React.Component {
             });
             this.setState({
                 action: `${action}:failed`,
+                actionErrorMessage: chooseBestErrorMessage(error, t('An unexpected error occured. Please try again or <a>contact support</a> if the error persists.')),
                 popupNotification
             });
         }
@@ -960,13 +964,14 @@ export default class App extends React.Component {
 
     /**Get final App level context from App state*/
     getContextFromState() {
-        const {site, member, action, page, lastPage, showPopup, pageQuery, pageData, popupNotification, customSiteUrl, dir, scrollbarWidth, labs, otcRef} = this.state;
+        const {site, member, action, actionErrorMessage, page, lastPage, showPopup, pageQuery, pageData, popupNotification, customSiteUrl, dir, scrollbarWidth, labs, otcRef} = this.state;
         const contextPage = this.getContextPage({site, page, member});
         const contextMember = this.getContextMember({page: contextPage, member, customSiteUrl});
         return {
             api: this.GhostApi,
             site,
             action,
+            actionErrorMessage,
             brandColor: this.getAccentColor(),
             page: contextPage,
             pageQuery,
--- a/apps/portal/src/actions.js
+++ b/apps/portal/src/actions.js
@@ -146,6 +146,10 @@ function startSigninOTCFromCustomForm({data, state}) {
 }
 
 async function verifyOTC({data, api, state}) {
+    const {labs} = state;
+
+    const genericErrorMessage = t('Failed to verify code, please try again');
+
     try {
         const integrityToken = await api.member.getIntegrityToken();
         const response = await api.member.verifyOTC({...data, integrityToken});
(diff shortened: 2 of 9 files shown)


apps/portal/src/App.js
```diff
@@ -714,6 +716,7 @@ export default class App extends React.Component {
             });
             this.setState({
                 action: `${action}:failed`,
+                actionErrorMessage: chooseBestErrorMessage(error, t('An unexpected error occured. Please try again or <a>contact support</a> if the error persists.')),
```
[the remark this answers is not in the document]
kevinansfield: This is copied from https://github.com/TryGhost/Ghost/blob/main/apps/portal/src/components/common/PopupNotification.js#L48, changing this would invalidate all translations and cause a lot of noise that's not desirable in this PR. It can be dealt with separately.

apps/portal/src/data-attributes.js
```diff
@@ -107,10 +107,10 @@ export async function formSubmitHandler(
                         email: (email || '').trim(),
                         otcRef
                     });
-                } catch (actionError) {
+                } catch (e) {
                     // eslint-disable-next-line no-console
-                    console.error(actionError);
-                    captureException?.(actionError);
+                    console.error(e);
+                    captureException?.(e);
```
kevinansfield, line 113: Renamed to avoid any confusion with the `actionErrorMessage` or a future `actionError` state property

apps/portal/src/actions.js
```diff
@@ -146,29 +146,47 @@ function startSigninOTCFromCustomForm({data, state}) {
...
+                return {
+                    action: 'verifyOTC:failed',
+                    actionErrorMessage: chooseBestErrorMessage(response.errors?.[0], genericErrorMessage)
+                };
+            } else {
+                return {
+                    action: 'verifyOTC:failed',
+                    popupNotification: createPopupNotification({
+                        type: 'verifyOTC:failed', autoHide: false, closeable: true, state, status: 'error',
+                        message: response.message || t('Invalid verification code')
```
kevinansfield, line 170: Use of `response.message` here was an existing bug, the API doesn't return a top-level `message` property

apps/portal/src/actions.js
```diff
@@ -146,29 +146,47 @@ function startSigninOTCFromCustomForm({data, state}) {
...
+                action: 'verifyOTC:failed',
+                actionErrorMessage: chooseBestErrorMessage(e, genericErrorMessage)
+            };
         } else {
             return {
                 action: 'verifyOTC:failed',
                 popupNotification: createPopupNotification({
                     type: 'verifyOTC:failed', autoHide: false, closeable: true, state, status: 'error',
-                    message: response.message || t('Invalid verification code')
+                    message: chooseBestErrorMessage(e, genericErrorMessage)
```
kevinansfield, line 186: Diff view is a bit confusing here. The non-alpha flag code hasn't actually changed, it's the same as the code [line 169 code](https://github.com/TryGhost/Ghost/pull/25118/files#diff-82712d3a51811fec93726581502d01d8e81e508d49eb256fb53129f20eef0454L169) in the "deleted" block below

apps/portal/src/actions.js
```diff
@@ -146,29 +146,47 @@ function startSigninOTCFromCustomForm({data, state}) {
...
+                };
+            } else {
+                return {
+                    action: 'verifyOTC:failed',
+                    popupNotification: createPopupNotification({
+                        type: 'verifyOTC:failed', autoHide: false, closeable: true, state, status: 'error',
+                        message: response.message || t('Invalid verification code')
+                    })
+                };
+            }
```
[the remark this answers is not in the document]
kevinansfield: PR change introduces alpha-flag changes, it should not impact any non-alpha flag code in the way you are suggesting.

apps/portal/src/tests/actions.test.js
```diff
@@ -41,3 +42,265 @@ describe('startSigninOTCFromCustomForm action', () => {
...
+
+            const result = await ActionHandler({
+                action: 'verifyOTC',
+                data: {otc: '000000', otcRef: 'ref-123'},
+                state: alphaState,
+                api: mockApi
+            });
+
+            expect(result.action).toBe('verifyOTC:failed');
+            expect(typeof result.actionErrorMessage).toBe('string');
```
jonatansberg: Nit: no need to assert the type separately from the string value

apps/portal/src/tests/actions.test.js
```diff
@@ -41,3 +42,265 @@ describe('startSigninOTCFromCustomForm action', () => {
...
+
+            const result = await ActionHandler({
+                action: 'verifyOTC',
+                data: {otc: '123456', otcRef: 'ref-123'},
+                state: alphaState,
+                api: mockApi
+            });
+
+            expect(result.action).toBe('verifyOTC:failed');
+            expect(typeof result.actionErrorMessage).toBe('string');
```
jonatansberg: Nit: no need to assert the type

apps/portal/src/tests/actions.test.js
```diff
@@ -41,3 +42,265 @@ describe('startSigninOTCFromCustomForm action', () => {
...
+
+            expect(mockApi.member.verifyOTC).toHaveBeenCalledWith({
+                otc: '123456',
+                otcRef: 'ref-123',
+                redirect: 'https://custom-redirect.com',
+                integrityToken: 'token-123'
+            });
+        });
+
+        test('includes integrity token in all API calls', async () => {
```
jonatansberg: Nit: This test case is more or less identical to the one above. You could possibly merge them.

apps/portal/src/tests/actions.test.js
```diff
@@ -41,3 +42,265 @@ describe('startSigninOTCFromCustomForm action', () => {
...
+            };
+
+            await ActionHandler({
+                action: 'verifyOTC',
+                data: {otc: '123456', otcRef: 'ref-123'},
+                state: {}, // no labs
+                api: mockApi
+            });
+
+            expect(mockLocationAssign).toHaveBeenCalled();
```
jonatansberg: What do you expect to happen here?

jonatansberg (approved):
I've tested this locally and everything seems to work, both with and without the alpha feature flag. 

I read a few nitpick comments on the tests. Nothing blocking.

There are some interesting patterns here, especially with the homegrown FLUX-like state management. I don't think it would make sense to try and change any of that now, but it would probably make sense to have a conversation about how we want to deal with shared state, error handling, API calls, etc in React in the future outside of this PR. 

I do find the fact that the action handler returns a pop-up notification to be a bit unexpected. I would have liked to see the API integration decoupled from the presentation of error messages. This PR is a step towards that, which is nice :)

kevinansfield:
> There are some interesting patterns here, especially with the homegrown FLUX-like state management. I don't think it would make sense to try and change any of that now, but it would probably make sense to have a conversation about how we want to deal with shared state, error handling, API calls, etc in React in the future outside of this PR.

Yeah, Portal was our very first React app and does things in quite a different manner to our newer apps. I think we're planning to switch Portal to our Shade components at some point, perhaps that would be a good point to re-examine the app's architecture.

PR #25120, opened 2025-10-09
Revert "Added a way to run tests on multiple shards in CI"

ibalosh:
Reverted TryGhost/Ghost#25102. Introduced sharding to speed up tests, but it seems it generates too much load on CI process.

(**sidenote** : it seems it was GH issue - degradation in performance)


ibalosh:
added a draft just in case, but it seems it was a GH issue after all

PR #25129, opened 2025-10-09
Reduced flakiness of tags e2e tests

9larsons:
ref https://github.com/TryGhost/Ghost/commit/ecb0f3d8f8102e04ad369277bcc9892e20e6552e
- updated tags tests to improve resiliency

A few tests were failing locally. Specifically, the save action was occurring too quickly for the remaining actions in some tests.


cmraible:
Can confirm this has addressed the flakiness locally. I've got a few small tweaks to make, I'll just push them here in a few if you don't mind?

PR #25142, opened 2025-10-13
Updated tags list to use empty indicator component

peterzimon:
ref https://linear.app/ghost/issue/PROD-2565/use-emptyindicator-for-tag-list-blank-state

- Updated the React tags screen to use the Empty indicator component from Shade instead of a custom implementation when there are no tags.


apps/shade/src/components/ui/empty-indicator.tsx
jonatansberg: This would still render an empty paragraph if there isn't a description

jonatansberg (approved):
The visual hierarchy is quite different in the new version compared to the previous one, and the Ember app before that. If this is by choice, then thats fine. Thought I'd call it out. 

I added a comment on a small nitpick. Nothing that would prevent this from being merged though.

kevinansfield (approved):

PR #25146, opened 2025-10-13
Added custom /commit slash command for Ghost-style commit messages

ibalosh:
- Created .claude/commands/commit.md with Ghost commit conventions
- Command guides through creating properly formatted commit messages
- Enforces Ghost's commit format: past tense, emoji for user-facing changes, bullet points
- Automates staging and committing with proper format validation
- Improves developer workflow consistency across the team

what are your thoughts on this


.claude/commands/commit.md
```diff
@@ -0,0 +1,23 @@
+---
+description: Generate a Ghost-style commit message and optionally commit changes
+---
+
+Create a commit message following Ghost's commit conventions:
+
+1. Run `git status` and `git diff` to see all staged and unstaged changes
```
cmraible: Rather than instructing the model to run these commands, we can directly execute them using "!`{command}`"

This should be a bit faster, because the model doesn't need to read the prompt, then run these commands — the output is directly interpolated into the prompt. 

Ref: https://docs.claude.com/en/docs/claude-code/slash-commands#bash-command-execution

.claude/commands/commit.md
```diff
@@ -0,0 +1,23 @@
...
+   - Following lines: Bullet points with context (what changed and why)
+   - Use emoji ONLY if the change is user-facing (✨ Feature, 🎨 Improvement, 🐛 Bug fix, 🌐 i18n)
+3. Show the commit message to the user
+4. Ask: "Would you like me to commit these changes? (yes/no)"
+5. If user says yes:
+   - Stage all modified files with `git add`
+   - Commit using the generated message
+   - Run `git status` to confirm
+6. If user says no, just stop
+
```
cmraible, line 23: It might be worth adding an example commit message or two, but it might not need that 🤷

.claude/commands/commit.md
```diff
@@ -0,0 +1,23 @@
+---
+description: Generate a Ghost-style commit message and optionally commit changes
+---
+
+Create a commit message following Ghost's commit conventions:
+
+1. Run `git status` and `git diff` to see all staged and unstaged changes
+2. Analyze the changes and draft a commit message with:
+   - First line: Past tense summary (max 80 chars)
+   - Second line: [blank]
```
cmraible: This is missing the `ref {issue_url}` line from the commit message guidelines. We could potentially accept an issue URL as an argument to this command, and instruct it to add the `ref` line if it's passed?

cmraible (approved):
Left a few comments with suggestions to potentially improve this, but nothing that should block merging. I'm more than happy to try this out and see how it feels, and we can make tweaks iteratively as needed.

Thanks for doing this, I had the same thought last week 😄

.claude/commands/commit.md
ErisDS: Can we please update these instructions to match up with our rules in CONTRIBUTING.md/internal docs?

ErisDS (changes requested):

.claude/commands/commit.md
ibalosh: nice touch! added in that change

.claude/commands/commit.md
ibalosh: instructed it to contributing file, which should give claude more context and keep it consistent

.claude/commands/commit.md
ibalosh: I just linked it to CONTRIBUTING.md, this way it will allways have a single source of truth, worked good

ErisDS (approved):

PR #25150, opened 2025-10-13
Separated OTC and magic link use count tracking

kevinansfield:
closes https://linear.app/ghost/issue/PROD-2502

We want to count OTC uses separately to magic links because OTCs don't need to allow "uses" by email security link clickers and we don't want those same link clickers to invalidate OTC usage.

- updated `SingleUseTokenProvider.validate` method for OTC validations:
  - increments `otc_used_count` rather than `used_count`
  - updates `first_used_at` if not already set
  - errors with `'OTC_EXPIRED'` error code when the OTC has been used previously (specific error allows for better diagnosis of member sign-in problems)
  - bypasses the magic-link specific `maxUsageCount` expiration
  - bypasses `validityPeriodAfterUsage` expiration
- updated validate method for magic-link validations:
  - errors with `TOKEN_EXPIRED` if an OTC has previously been used

### DIFF
--- a/ghost/core/core/server/services/members/SingleUseTokenProvider.js
+++ b/ghost/core/core/server/services/members/SingleUseTokenProvider.js
@@ -9,6 +9,7 @@ const messages = {
     INVALID_OTC_VERIFICATION_HASH: 'Invalid OTC verification hash',
     INVALID_TOKEN: 'Invalid token provided',
     TOKEN_EXPIRED: 'Token expired',
+    OTC_EXPIRED: 'One-time code expired',
     DERIVE_OTC_MISSING_INPUT: 'tokenId and tokenValue are required'
 };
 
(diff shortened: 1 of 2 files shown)


jonatansberg (changes requested):
I think it would make sense to split the OTC usage validation from the magic link validation into separate methods. That way we won't have to pass around the `isOTC` boolean flag which is cleaner.

kevinansfield:
> I think it would make sense to split the OTC usage validation from the magic link validation into separate methods. That way we won't have to pass around the isOTC boolean flag which is cleaner.

@jonatansberg I had previously tried that but to me it felt much cleaner having the `isOTC` boolean flag because it put the differences front-and-center within each single-purpose method, whereas with a bunch of extra separated functions it was harder to follow and pick apart how OTC validation varies from the standard token validation.

I've pushed a commit that splits into different functions for review in https://github.com/TryGhost/Ghost/pull/25150/commits/625715c16c3745b44c284865e4835cba8f4bbd3f, although I'd prefer to keep the simpler `isOTC` flag approach.

jonatansberg:
> I've pushed a commit that splits into different functions for review in [625715c](https://github.com/TryGhost/Ghost/commit/625715c16c3745b44c284865e4835cba8f4bbd3f), although I'd prefer to keep the simpler `isOTC` flag approach.

It's a little more verbose this way, but I do think I prefer it. At least with one additional change. In that version you have right now you are still doing OTC verifications from within the magic link method/code branch. I think that is a separate constraint that ought to be modeled by composing the the different methods. The OTC methods only ought to be concerned with validating the OTC usage and same for the magic link ones. 

You achive that by pulling out line 74:83 into a separate method and check both the OTC usage and magic links in the link case. That way the individual checks only do one thing. I'd probably consider renaming the magic link checks as well to make the flow clearer.

jonatansberg (approved):
Nice! 👍

PR #25151, opened 2025-10-13
Converted `SingleUseTokenProvider` to TypeScript

kevinansfield:
no issue

- following our scout rule that any files touched should be converted to TypeScript
- updated our job service to pass `['--experimental-strip-types', '--import=tsx']` through to the bree worker so TS can be used in background worker jobs


ghost/core/core/server/services/members/SingleUseTokenProvider.ts
```diff
@@ -13,16 +14,54 @@ const messages = {
...
+    get(key: 'first_used_at'): Date | null;
+    get(key: string): unknown;
+    save(data: Record<string, unknown>, options: {autoRefresh: boolean; patch: boolean; transacting: Knex.Transaction}): Promise<void>;
+}
+
+interface TokenModelStatic {
+    add(data: {data: string}): Promise<TokenModel>;
+    findOne(query: {token?: string; uuid?: string}, options?: {transacting?: Knex.Transaction; forUpdate?: boolean}): Promise<TokenModel | null>;
+    transaction<T>(callback: (transacting: Knex.Transaction) => Promise<T>): Promise<T>;
+}
```
kevinansfield, line 35: Should we be storing model interfaces somewhere more centrally?

Looking in the codebase I found two other examples where model interfaces are used, both have different approaches for inline definitions:
- [BookshelfRepository](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/services/recommendations/service/BookshelfRepository.ts#L17-L33) - "generic" bookshelf model type
- [DonationBookshelfRepository](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/services/donations/DonationBookshelfRepository.ts#L7-L32) - slightly less generic, doesn't account for bookshelf's `get` method?

ghost/core/core/server/services/jobs/job-service.js
```diff
@@ -45,5 +45,11 @@ const initTestMode = () => {
 
 const jobManager = new JobManager({errorHandler, workerMessageHandler, JobModel: models.Job, domainEvents, config, events});
 
+// Add Node.js flags for TypeScript support in worker threads
+// This enables --experimental-strip-types for Node < 22.18.0
+jobManager.bree.config.worker = {
+    execArgv: ['--experimental-strip-types']
+};
+
```
kevinansfield, line 53: Found because CI uses `22.13`, I wasn't seeing problems locally running `22.18`

ghost/core/core/server/services/members/SingleUseTokenProvider.js
```diff
@@ -1,416 +1,2 @@
...
-            updateData.first_used_at = new Date();
-        }
-
-        await model.save(updateData, {autoRefresh: false, patch: true, transacting});
-    }
-}
-
-module.exports = SingleUseTokenProvider;
+/* eslint-disable ghost/filenames/match-regex */
+module.exports = require('./SingleUseTokenProvider.ts');
```
kevinansfield: This was needed to satisfy requires within worker processes

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
```diff
@@ -0,0 +1,422 @@
+/* eslint-disable ghost/filenames/match-regex */
+/* eslint-disable @typescript-eslint/no-var-requires */
+import type {Knex} from 'knex';
+
+const {ValidationError} = require('@tryghost/errors');
+const tpl = require('@tryghost/tpl');
+const crypto = require('node:crypto');
+const {hotp} = require('otplib');
```
kevinansfield: CJS requires here were needed to satisfy Node when a `.ts` file is loaded with `require`

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
```diff
@@ -13,16 +14,54 @@ const messages = {
...
+    findOne(query: {token?: string; uuid?: string}, options?: {transacting?: Knex.Transaction; forUpdate?: boolean}): Promise<TokenModel | null>;
+    transaction<T>(callback: (transacting: Knex.Transaction) => Promise<T>): Promise<T>;
+}
+
+interface SingleUseTokenProviderDependencies {
+    SingleUseTokenModel: TokenModelStatic;
+    validityPeriod: number;
+    validityPeriodAfterUsage: number;
+    maxUsageCount: number;
+    secret?: string;
```
allouis: **question (non-blocking):** Why is secret optional? I know you're just matching existing JSDoc - but is that correct?

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
```diff
@@ -221,15 +239,13 @@ class SingleUseTokenProvider {
...
      *
-     * @param {string} token
-     * @param {Object} transacting
-     * @returns {Promise<Object>}
+     * @param token - The token to find
+     * @param transacting - Database transaction object
+     * @returns The token model
      */
-    async _findAndLockTokenModel(token, transacting) {
+    private async _findAndLockTokenModel(token: string, transacting: Knex.Transaction): Promise<TokenModel> {
```
allouis: **nitpick:** I'm not sure if the `_` prefix to denote privacy is still useful one we're using the `private` keyword

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
```diff
@@ -72,7 +107,7 @@ class SingleUseTokenProvider {
         const isOTC = !!options.otcVerification;
 
         if (isOTC) {
-            await this._validateOTCVerificationHash(options.otcVerification, token);
+            await this._validateOTCVerificationHash(options.otcVerification!, token);
```
allouis: **thought:** We should try not to assert non-null if we can, in this case it's safe - but if the definition of `isOTC` were to change, then this could break type safety.

The simplest way round this would be:

```ts
if (!!options.otcVerification) {
  await this._validateOTCVerificationHash(options.otcVerification, token);
}
```

But then we end up with a partially used `isOTC` definition.

Another approach could be something like this:

```ts

// This type contains both pieces of information, the string, and whether we are using it
// It also encodes that requirement that we always use it if we have it, and never if we don't
type OTCFlowInfo = { active: true;  verification: string } | { active: false; verification: null };

async validate(token: string, options: ValidateOptions = {}): Promise<Record<string, unknown>> {
  if (!options.transacting) {
    // ...
  }
  
  const otc: OTCFlowInfo = !!options.otcVerification ? {
    active: true,
    verification: options.otcVerification
  } : { active: false, verification: null};

  if (otc.active) {
    await this._validateOTCVerificationHash(options.otc.verification, token);
  }

  const model = await this._findAndLockTokenModel(token, options.transacting);
  this._validateUsageCount(model, otc.active);
  this._validateTokenLifetime(model, otc.active);

  await this._incrementUsageCount(model, otc.active, options.transacting);

  try {
    return JSON.parse(model.get('data'));
  } catch (err) {
    return {};
  }
}
```

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
allouis: Yeah so I think probably/maybe? But I'm not sure it's as straightforward as just adding some shared types

I think we have two options:
1. Add a shared interface which our code can depend on, and the models just "happen" to implement
2. Explicitly type the bookshelf models, we can extend the bookshelf type definitions

Going with a shared interface feels like we're cheating a bit. Our code essentially depends on a bookshelf model, but we're pretending it's through an interface, and it's not even an interface we've chosen or designed

But typing the models means our code is explicitly depending on the Bookshelf library (because we'd have to import the model definition), which is probably even more of a PITA when we eventually want (have?) to migrate.

So neither of the options are ideal...

We've already somewhat bought into the repository pattern in places. If we stich with that then we can explicitly type the Bookshelf models (which I think is easier) and as we refactor converge on isolating their usage to repositories.

Now our business logic defines the interfaces of repositories, interfaces we actually design for our use cases, not ones dictated by Bookshelf. We get proper dependency inversion: business logic doesn't depend on infrastructure, infrastructure depends on business logic.

All that being said, I think **TL;DR: Yes, we need to do something about sharing/exposing model types, but I think it needs a larger discussion. And for this PR - I think the inline types are fine, they retain ownership of the interface**

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
```diff
@@ -413,4 +417,4 @@ class SingleUseTokenProvider {
     }
 }
 
-module.exports = SingleUseTokenProvider;
+export = SingleUseTokenProvider;
```
allouis: **question (non-blocking):** Why this style of export? I would have expected something like `export class SingleUseTokenProvider ...` or even `export default ...`

allouis (commented):
This looks like it was a bit of a PITA - thanks for taking the time to convert it!

**question (blocking):** Is there any way we can use TypeScript with Bree? It would be really nice if we can use the full TS feature set! I had a little look at the bree docs (https://github.com/breejs/bree/blob/master/examples/typescript-esm/README.md), which mention using the `NODE_OPTIONS` environment variable (https://nodejs.org/api/cli.html#cli_node_options_options)

I think this suggests we should be able to do something like this in the Ghost core package.json (not sure if we need it twice?)
```json
{
  "scripts": {
    "dev": "NODE_OPTIONS= \"--import=tsx\" node --watch --import=tsx index.js"
  }
}
```

If we can't, we can't - but I think we should try before continuing with merging - what do you think?

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
kevinansfield: > We've already somewhat bought into the repository pattern in places. If we stich with that then we can explicitly type the Bookshelf models (which I think is easier) and as we refactor converge on isolating their usage to repositories.

💯 that we should either go all-in on the repository pattern or choose some other approach that gives us consistency across the codebase. It's a confusing mess right now, especially at edges where we end up with either POJOs or model instances depending on which code path has been taken (`PostsService` and serializers/content gating springs to mind) 😅

> And for this PR - I think the inline types are fine, they retain ownership of the interface

Cool, I think that makes the most sense for now too. Definitely agree we need a larger discussion and a resulting push for standardisation.

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
kevinansfield: Good catch, secret is no longer optional with OTC enabled which will soon be the default. Updated 👍

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
kevinansfield: Ah, good point. I've switched to using `#` private fields instead

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
kevinansfield: Slightly outdated as this was switched to `module.exports =` but still, this was due to Node's incomplete support for using `require` across CJS/ESM(TS) boundaries. It might be fixed if we can get the `import=tsx` support working in bree, I'm looking into that

kevinansfield: Bree is now working with `import=tsx` so I have a bit more clarity here.

If we use `export default class SingleUseTokenProvider` then everywhere we require this module from CJS we need to change to `SingleUseTokenProvider = require('...').default`. Changing to `export = SingleUseTokenProvider` removes that need.

Are we OK with imposing the need for `.default` for any requires?

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
kevinansfield: [I opted for re-ordering](https://github.com/TryGhost/Ghost/pull/25151/commits/1721567e7a6dea733d68922b271179078021ef38) so we use `if (options.otcVerification) {...hash check...}` before we define `isOTC`. That way by the time we define `isOTC` we also know that we definitely have an OTC request or not because anything that wasn't a valid OTC hash will have thrown earlier

ghost/core/core/server/services/members/SingleUseTokenProvider.ts
allouis: I think it's fine to impose the `.default` for requires, and we use `.default` when importing the `bson-objectid` dep in Ghost, so the pattern isn't unknown.

That said, we can also do `export class SingleUseTokenProvider` and then update requires like `const {SingleUseTokenProvider} = require('./path/to');` which is maybe more idiomatic in Ghost?

allouis (approved):

PR #25158, opened 2025-10-14
Fix race condition in tag editor tests

jonatansberg:
Ref https://linear.app/ghost/issue/PROD-2894/troubleshoot-flaky-e2e-tests-running-locally

The tag editor in Ember will redirect to the URL for the newly created tag when saving. If we click the back link in the breadcrumbs too soon we introduce a  race condition between the back navigation and the post-save redirect. 

The navigation to the tag list means that the URL assertion will succeed, but  we will have been redirected back to the editor before the list actually  renders. That makes the following assertion fail.

In order to target the saved state of the save button, I needed to use a use a different locator. While changing that I used the opportunity to align the  other locators with what we consider best practice.


ibalosh (approved):
looks great Jonatan 👌

PR #25161, opened 2025-10-14
Added new ShadCN/UI components to Shade

peterzimon:
ref https://linear.app/ghost/issue/DES-1220/install-missing-shadcnui-components

- [Input group](https://ui.shadcn.com/docs/components/input-group) and [Field](https://ui.shadcn.com/docs/components/field) were missing from Shade. They both are needed for the Tag details screen implementation.


apps/shade/package.json
```diff
@@ -76,13 +76,13 @@
...
-        "@radix-ui/react-label": "2.1.7",
+        "@radix-ui/react-label": "^2.1.7",
         "@radix-ui/react-popover": "1.1.14",
         "@radix-ui/react-radio-group": "1.3.7",
         "@radix-ui/react-select": "2.2.5",
-        "@radix-ui/react-separator": "1.1.7",
+        "@radix-ui/react-separator": "^1.1.7",
         "@radix-ui/react-slider": "1.3.6",
-        "@radix-ui/react-slot": "1.2.3",
+        "@radix-ui/react-slot": "^1.2.3",
```
kevinansfield, line 85: nit: Not a blocker but interested to know where these came from - Renovate will come along and change them back once merged because we have a rule that all dependencies use pinned versions

kevinansfield (approved):

PR #25163, opened 2025-10-14
Cleaned up data factories, readme, env variables

ibalosh:
- Flattened factory folder structure: factories/posts/ → factories/ to keep things simpler, since we don't need subfolders yet
- Updated documentation 
- Removed name parameter, which was introduced as identifier of a particular factory, so that those factories can be filtered and deleted later. Since we don't use that model anymore, I have removed this
- Added custom number of workers per env variable, to custom parallelization locally


cmraible (commented):
I'm not sure what the original intent of the name property was. Do we use it anywhere?

Also just want to call out that using the constructor name would make it e.g. `PostFactory` instead of `post` as it was before, which I suspect it doesn't match the original intent of the `name` field.

Also just want to throw this out there: do we need a name property/getter?

ibalosh:
> I'm not sure what the original intent of the name property was. Do we use it anywhere? ...

hey @cmraible , it's a good question :) That identifier was just used to identify specific factory, so that when you want to cleanup data, you can find for example all post factories, and delete these posts. We have ditched introducing in persistence/factories this. We use way simpler way to clear data, as we should. Therefore I have removed that name completely now - it just introduced confusion.

ibalosh:
I will merge this one in, since it's very small, and made a change to reflect the name identifier (removed it)

PR #25174, opened 2025-10-15
Added Auto-Submitted header to prevent automatic replies

sam-lord:
ref [GVA-567](https://linear.app/ghost/issue/GVA-567/adding-the-sender-header-causes-automatic-replies-to-be-sent-back-for)

This should prevent auto-responders (e.g. out-of-office replies) from being sent, which was a regression introduced when we added the Sender header.


lsinger (approved):
LGTM! 🚢

PR #25178, opened 2025-10-15
Add admin deployment job to CI workflow

jonatansberg:
Ref https://linear.app/ghost/issue/BER-2900/set-up-github-actions

This is the Github action/workflow part needed to build Ghost in a way that will let us use set up continuous delivery of the new React admin. 

- Build admin on merge to main
- Upload assets to artifact storage
- Trigger Ghost-Moya deploy job (like the canary builds)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1203,3 +1203,47 @@ jobs:
           token: ${{ secrets.TINYBIRD_TOKEN }}
           workspace: 'Production'
           slack-webhook: ${{ secrets.ANALYTICS_SLACK_WEBHOOK_URL }}
+
+  deploy_admin:
+    needs: [
+      job_setup,
+      job_required_tests
+    ]
+    name: Deploy Admin
+    runs-on: ubuntu-latest
+    if: |
+      always()
+      && needs.job_setup.result == 'success'
+      && needs.job_setup.outputs.is_main == 'true'
+      && needs.job_required_tests.result == 'success'
+    steps:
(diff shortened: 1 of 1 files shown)


apps/admin/vite-ember-assets.ts
ErisDS, line 10: I'm feeling confused about this line - if the URL starts with `/` then how come it is deemed absolute, rather than relative?

ErisDS, line 13: Also finding it hard to reason about what `base` is, trying to think of a better name for this. As I understand it's either the CDN url, or the local path?

ErisDS (approved):
There's a few places where I think this can be tightened up so it's clearer what's happening, _IF_ these things are going to be long-lived. If they're temporary, there's little point :)

apps/admin/vite-ember-assets.ts
jonatansberg: Urls that start with `/` are absolute. The browser will resolve that to `proto://host:port/<rest of url>`. Urls without a leading slash (or `./`) are relative to the current path.

jonatansberg: In this context `base` refers to the `base` config option in Vite, that in turn would be the either of those two in our usage.

jonatansberg:
@ErisDS The Ember Asset plugin shouldn't be very long-lived. The CI workflow hopefully will stick around, although we might still tweak that. 

 I've replied to your comments. Let me know if there's anything you'd like me to change.

PR #25180, opened 2025-10-15
Added clear commit summary rule to contributing.md

ErisDS:
ref https://github.com/TryGhost/Ghost/pull/25146#pullrequestreview-3341218787 ref https://github.com/TryGhost/Ghost/blob/main/.github/hooks/commit-msg#L94

- Realised this wasn't consistent with our internal guide or our commit-msg hook
- This update makes it clearer exactly what the first line of a commit should be like
- This update makes the text consistent with other references to these rules


ibalosh (approved):
👌

PR #25181, opened 2025-10-15
Added docker build to pretest command when running e2e tests

cmraible:
The E2E tests run against the Ghost docker image. Currently if you make changes to Ghost, you need to manually rebuild the docker image before running the tests again, which is not obvious and annoying to have to run constantly.

This adds the Docker image build as a pretest command, so by default the image will always be rebuilt before running the e2e tests.


ibalosh (approved):
Only thing I would add is to add env variable to .env.example.

Having them all there makes it easy to see what .`env` options we have

PR #25182, opened 2025-10-15
Updated e2e tests global setup to reuse shared containers between test runs

cmraible:
Currently the global setup in e2e tests always starts the shared services in `e2e/compose.yml` for every test run, and stops all services in the global teardown. When iterating on tests locally, this is really slow and annoying, driven mostly by Tinybird setup taking ~30s.

This change leaves the shared services (mysql, tinybird, etc) running, and just cleans up their state so the containers can be reused in subsequent test runs without incurring the 30s delay.

It will also cleanup any leftover ghost containers and databases in the global setup, so if any are left running from previous test runs, we won't need to clean them up manually before running the tests again.

### DIFF
--- a/e2e/helpers/environment/DockerCompose.ts
+++ b/e2e/helpers/environment/DockerCompose.ts
@@ -167,10 +167,6 @@ export class DockerCompose {
         return output;
     }
 
-    /**
-     * Execute a command in a service container using `docker compose run`.
-     * This runs the command with the service's default entrypoint and environment.
-     */
     execInService(service: string, command: string[]): string {
         const cmdArgs = command.map(arg => `"${arg}"`).join(' ');
         const cmd = `docker compose -f ${this.composeFilePath} -p ${this.projectName} run --rm -T ${service} ${cmdArgs}`;
--- a/e2e/helpers/environment/DockerCompose.ts
+++ b/e2e/helpers/environment/DockerCompose.ts
@@ -168,6 +168,18 @@ export class DockerCompose {
     }
 
     /**
+     * Execute a command in a service container using `docker compose run`.
+     * This runs the command with the service's default entrypoint and environment.
+     */
+    execInService(service: string, command: string[]): string {
+        const cmdArgs = command.map(arg => `"${arg}"`).join(' ');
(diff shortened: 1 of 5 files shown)


e2e/helpers/environment/DockerCompose.ts
```diff
@@ -167,6 +167,18 @@ export class DockerCompose {
         return output;
     }
 
+    /**
+     * Execute a command in a service container using `docker compose run`.
+     * This runs the command with the service's default entrypoint and environment.
```
ibalosh: `nit` comment not needed in my opinion, function is super clear

e2e/helpers/environment/EnvironmentManager.ts
```diff
@@ -69,20 +69,61 @@ export class EnvironmentManager {
...
+        try {
+            logging.info('Cleaning up leftover resources from previous test runs...');
+
+            // Clean up leftover Ghost containers
+            await this.ghost.removeAll();
+
+            // Clean up leftover test databases if MySQL is running
+            await this.mysql.dropAllTestDatabases();
+
+            // Delete the MySQL snapshot to ensure fresh migrations
```
ibalosh: `nit` since function is explained at top, and function names for actions you do are clear, these comments can be cleared, and would make the flow of actions easier to read

e2e/helpers/environment/EnvironmentManager.ts
```diff
@@ -69,20 +69,61 @@ export class EnvironmentManager {
...
+     * 4. Recreate the ghost_testing database (if MySQL is running)
+     * 5. Truncate Tinybird analytics_events datasource (if Tinybird is running)
+     *
+     * Note: Docker compose services are left running for reuse across test runs
+     */
+    private async cleanupLeftoverResources(): Promise<void> {
+        try {
+            logging.info('Cleaning up leftover resources from previous test runs...');
+
+            // Clean up leftover Ghost containers
```
ibalosh: `nit` function and intentions are clear, comments seem redundant within function

ibalosh (approved):
love it @cmraible , can't wait to see this in main :) 

If I would have anything to add is that there are too many comments I think. You made the function names pretty clear, so I think that many of comments are not needed and duplicate what code already describes 

Ship it! :)

PR #25184, opened 2025-10-16
Pinned traffic analytics version in compose.yml

cmraible:
Currently docker compose won't ever pull the latest version if there's already a v1.x.x pulled locally. This can be annoying when working on features that span the analytics service and Ghost, as you may be working against a stale version of the analytics service without realizing it.

By pinning the exact version, this will force docker to pull the latest available as soon as it's updated in the compose file (which renovate should handle for us as new versions are released).

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -191,7 +191,7 @@ services:
 
   analytics:
     profiles: [ analytics, all ]
-    image: ghost/traffic-analytics:1
+    image: ghost/traffic-analytics:1.0.18
     platform: linux/amd64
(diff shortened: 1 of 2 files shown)


ibalosh (approved):
My only question regarding this is, is there a way to know when new version is out?

We want to make sure tests use correct version

cmraible:
Yes! As soon as a new image hits docker hub, our renovate config will raise a PR to update to the latest. As long as tests pass, it will then merge it automatically 3 days later, or we can manually merge it if needed.

PR #25191, opened 2025-10-16
Decouple runtime data fixtures for auth vs. public flows

ibalosh:
- Split fixture.ts into base-fixture.ts and authenticated-fixture.ts
- Tests choose authentication level via import path

The reason for introducing this was to make tests import only fixtures they really need, and to keep Playwright fixtures bit easier on the eye, and just contain what is actually used by tests.


e2e/helpers/playwright/fixtures/authenticated-fixture.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Extends base fixture with automatic login
+ *
+ * Usage:
+ * - Import from this file instead of base-fixture.ts
+ * - Use `page` fixture as normal - it will be authenticated
+ * - Optionally set labs flags via test.use({labs: {featureName: true}})
+ */
+export const test = base.extend<AuthenticatedFixture>({
+    // Define labs as an option that can be set per test or describe block
+    labs: [undefined, {option: true}],
```
cmraible, line 58: note: labs may be relevant for some tests that don't need an authenticated admin page, i.e. if the flag impacts theme rendering and nothing else. 

This is fine for now, but I'd eventually like to decouple how we set labs flags from the authenticated page in admin, by e.g. passing the labs flags via config to the Ghost container, rather than enabling flags in the UI.

cmraible (commented):
I'm a bit conflicted on this. In the interest of ensuring that e2e tests are easy to write and to get started writing, I'm afraid this requires the developer to have more intimate knowledge of how the playwright fixtures work, in order to know which fixture to import. With this setup, the very first line I have to add to a new test file requires me to read through our fixture code, develop a mental model of how playwright's fixtures work, and then make a decision of which fixture I want to use.

I like that we're decomposing the monolithic fixture file into multiple files to make the fixtures themselves easier to maintain, extend, and reason about — but I wonder if we could still export them from a single file, or otherwise make it easier for newcomers to figure out which fixtures they need. Documenting this pattern in the README might go a long way toward this too, though.

e2e/helpers/playwright/fixtures/authenticated-fixture.ts
ibalosh: regarding the note: I agree, it's coupled in at the moment, on long run we don't want it to be here [](edited-with-Axolo)

ibalosh:
@cmraible I will close this PR and come back to it later, I need some more thought into this

PR #25194, opened 2025-10-16
🎨 Split helper

cathysarisky:
This is a new "split" handlebars helper that will split strings with a user-specified separator (or defaulting to ',').  Example uses:
```
{{#split "hello,world" as |elements|}}
  {{#foreach elements}}
    |{{this}}|
  {{/foreach}}
{{/split}}
```
Result: `|hello||world|`

```
{{#foreach (split "hello, world" separator=",")}}
   {{this}} {#unless @last}}<br>{{/unless}}
{{/foreach}}
```
Result:   `hello<br> world`


cathysarisky:
Question for review: Should I be returning an array of strings, or SafeStrings ?  Feeling a little muddled here.

ErisDS:
> Should I be returning an array of strings, or SafeStrings ? Feeling a little muddled here.

An array of SafeStrings, I would suggest.

ghost/core/core/frontend/helpers/split.js
```diff
@@ -0,0 +1,49 @@
+module.exports = function split(...args) {
+    const options = args.pop();
+    const data = options.data || {};
+    const separator = options.hash.separator !== undefined ? options.hash.separator : ',';
```
ErisDS, line 13: Can be shortened to 
`const separator = options.hash.separator || ',';` 

Same as the line above and in the concat helper :)

cathysarisky:
> An array of SafeStrings, I would suggest.

Bah. I had that, and then some pesky LLM convinced me I was wrong! 🙄

ghost/core/core/frontend/helpers/split.js
```diff
@@ -0,0 +1,49 @@
...
+        const emptyResult = [];
+        if (options.fn) {
+            return options.fn(emptyResult, {
+                data: data,
+                blockParams: [emptyResult]
+            });
+        } else {
+            return emptyResult;
+        }
+    }
```
ErisDS, line 29: I think there's a way to make this code simpler and more readable.

I understand everything above this comment is handling non-happy path cases - checking if the string is null or undefined first (in that case do the empty handling)
Then coerce to a string
If _that_ is empty, do empty handling again

And then everything below is what we actually want to do.

There are several different ways to do this, one that comes to mind is to make a render function:

```
function renderResult(result) {
  if (options && typeof options.fn === 'function') {
    return options.fn(result, { data, blockParams: [result] });
  }
  return result;
}

```

Then for example, line 26-28 would become:

```
if (string === '') {
   renderResult([]);
}
```

This is pure code cleanliness though - the code is _correct_ and the tests look solid 🙏

ghost/core/core/frontend/helpers/split.js
cathysarisky: Actually, this messes up the case where the separator is "".

ErisDS (approved):

PR #25197, opened 2025-10-17
Improved spacing of the publish page on small screens

lorumic:
fixes https://github.com/TryGhost/Ghost/issues/24898

This PR improves the right spacing of the publish page header, by making it proportional to the left one. It also ensures that the "Back to settings" secondary CTA does not overflow horizontally on very narrow viewports.

- Before

image

- After

image

### DIFF
--- a/ghost/admin/app/styles/components/publishmenu.css
+++ b/ghost/admin/app/styles/components/publishmenu.css
@@ -231,8 +231,7 @@
         align-items: center;
         height: 64px;
         margin: 0;
-        padding: 0;
-        padding-left: 15px;
+        padding: 0 15px;
         background-color: var(--white);
         border-radius: 0;
     }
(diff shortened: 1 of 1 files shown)


joeldesante:
@lorumic 
Wouldn't it be better to drop that secondary button in the CTA below the primary button so that they are stacked vertically after hitting the mobile break point. 

I cant see any reason the user would want the button to trail off like it has in the example above.

lorumic:
> Wouldn't it be better to drop that secondary button in the CTA below the primary button so that they are stacked vertically after hitting the mobile break point.

Yes, I agree, that would be better and was also one of the other things that I had tried while I was implementing the change. I think the final decision is up to the Ghost team, as it's a UX one - if it's confirmed that what we want here is wrapping, I'll be happy to address this change. :)

9larsons:
Hey, added some tweaks! Thanks for this. I can't say I'm ever on mobile, so we appreciate those who use mobile heavily to suggest these. I've added some wrapping and fixed the "Publish" text falling behind the modal header at the same time. It's all related enough.

edit: We're in the process of moving all this to React+Tailwind which should make the mobile behavior a lot friendlier.

PR #25206, opened 2025-10-20
Removed unused `mediaQueries` and `ui` service code

kevinansfield:
no issue

- removed unused properties in `ui` service
  - these properties were hangovers from old approaches to fullscreen routes and mobile design adjustments
- removed unused `mediaQueries` service
  - the only place this was used was in the unused `ui` service properties and a skipped test that is out of date with the current component implementation


jonatansberg (approved):
Nice! 👍

PR #25209, opened 2025-10-21
Moved static url prefixes out of config

allouis:
ref https://linear.app/ghost/issue/PRO-1524

We are already reading the static url prefix for images from url-utils
and we're going to be updating url-utils to be performing logic based on
these constants. The idea here is to consolidate our usage and transfer
ownership of the constants to url-utils. For now we are setting this in
Ghost with `module.exports =` but that is temporary until the package
has been updated to export these values itself.

### DIFF
--- a/ghost/core/core/shared/url-utils.js
+++ b/ghost/core/core/shared/url-utils.js
@@ -2,7 +2,6 @@ const UrlUtils = require('@tryghost/url-utils');
 const config = require('./config');
 
 const BASE_API_PATH = '/ghost/api';
-
 const urlUtils = new UrlUtils({
     getSubdir: config.getSubdir,
     getSiteUrl: config.getSiteUrl,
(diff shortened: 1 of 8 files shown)


ghost/core/core/shared/url-utils.js
```diff
@@ -14,3 +13,6 @@ const urlUtils = new UrlUtils({
 
 module.exports = urlUtils;
 module.exports.BASE_API_PATH = BASE_API_PATH;
+module.exports.STATIC_IMAGE_URL_PREFIX = 'content/images';
```
allouis, line 16: This is already exposed on the `UrlUtils` instance, but I've added it for consistency with the other two.

All three of these `module.exports` assignments for the constants will be removed once `UrlUtils` has been updated to expose the values (I'm currently converting it to TS)

mike182uk (approved):

PR #25211, opened 2025-10-21
Fixed running Ember tests locally on :4200/tests/

kevinansfield:
no issue

- at some point in the not too distant past the chunked JS files output by `ember-auto-import` started to be included in the `tests/index.html` file with no leading `/` resulting in 404s and an empty tests screen
- for CI tests the build runs in the `test` env so our conditional in `ember-cli-build.js` to set `publicAssetURL` to `undefined` and let auto-import do it's thing worked
- for local browser tests when running `yarn dev`, `ember-cli-build.js` is only run in the "dev" env so `publicAssetURL` was always set to the relative `assets/` path which is necessary for dev environments to work in subdirectory setups
- the fix was to add `<base href="{{rootURL}}">` to the tests `index.html` template so any relative `src` attributes are interpreted by the browser as being relative to `/` rather than `/tests/` when not set

### DIFF
--- a/ghost/admin/app/services/onboarding.js
+++ b/ghost/admin/app/services/onboarding.js
@@ -70,7 +70,9 @@ export default class OnboardingService extends Service {
     }
 
     @action
-    async completeChecklist() {
+    async completeChecklist(event) {
+        event?.preventDefault();
+
         const settings = this.settings;
 
         settings.checklistState = 'completed';
@@ -79,7 +81,9 @@ export default class OnboardingService extends Service {
     }
 
     @action
-    async dismissChecklist() {
+    async dismissChecklist(event) {
+        event?.preventDefault();
+
         const settings = this.settings;
 
         settings.checklistState = 'dismissed';
--- a/ghost/admin/tests/index.html
+++ b/ghost/admin/tests/index.html
@@ -2,6 +2,7 @@
 <html>
   <head>
     <meta charset="utf-8">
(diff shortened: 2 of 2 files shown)


jonatansberg (approved):
With these changes, the tests seem to run as expected and there are no surprise redirects in the end. 

I have a handful of failing tests and some that are stuck not completing. I think you mentioned some other time that this is a known issue related to the difference between how the tests run in the browser and in CI. 

Based on that, I think this looks good to go!

kevinansfield:
> I have a handful of failing tests and some that are stuck not completing.

The search modal opening via keyboard tests consistently fail for me locally, both before and after this change. I time-boxed 30min earlier to look at a fix but didn't find why it was failing and alternative implementations didn't improve things.

Another frequent failure point is the navigation tests if you don't have the Chrome tests tab focused. Also failures in general with a non-focused tab that gets put into a sleep/low-power mode.

Do you see any other persistent failures?

jonatansberg:
I've tested a few times now running the tests in the browser while maintaining focus within the tab. That leaves me with two persistent failures:

* returns false when entry published_at equals lastSeenDate exactly
* uses local search if all tags have been loaded by scrolling

rob-ghost:
> returns false when entry published_at equals lastSeenDate exactly

This is a `Whats new` test that I added recently. I'll give it a look because its probably flaky and should be fixed.

PR #25215, opened 2025-10-21
Migrated translations docs from Notion Codex site

cmraible:

- Why are you making it?
- What does it do?

- [ ] I've read and followed the [Contributor Guide](https://github.com/TryGhost/Ghost/blob/main/.github/CONTRIBUTING.md)
- [ ] I've explained my change


docs/translations.md
```diff
@@ -0,0 +1,461 @@
...
+   ```bash
+   yarn test
+   ```
+
+5. **Commit changes:**
+   Include both code changes and updated translation files in your commit.
+
+### Updating Existing Translations
+
+When modifying existing translation keys:
```
cathysarisky, line 344: it would probably be good to offer a note here that changing an existing key with translations and running extraction will cause deletion of any strings already attached to the old key.

(We could use better tooling here, for those times the changes to the string are modest and keeping the existing translation would be better than going back to untranslated.)

cathysarisky (changes requested):
Leaving a note about changing key values, since I'm not sure it's as obvious as it should be that translations will be lost.

PR #25216, opened 2025-10-22
Resolve CSS specificity conflicts between Ember and React apps

jonatansberg:
**Ref:** https://linear.app/ghost/issue/BER-2914/resolve-css-specificity-conflict

This PR fixes CSS specificity conflicts that occur when the Ember admin app and React apps (using the shade design system) are hosted together in the admin shell. 

Specifically:
1. Tailwind utility classes for visibility did not work in React apps due to `!important` rules from Ember
2. Visual regressions in the Ember app when mounted in the shell due to competing CSS resets

#### 1. Scoped show and hidden classes to the Ember app

- The usage of `!important` in the base Ember stylesheet was breaking Tailwind utility classes for visibility in React apps
- By scoping these classes to `.ember-application`, the Tailwind utilities work as expected outside of Ember
- This doesn't affect existing Ember behavior since the increased specificity doesn't matter given the `!important` flag

#### 2. Reduced specificity of the shade CSS reset

- When the shade CSS reset is applied to the admin shell, it affects the mounted Ember app
- Since the shade reset differs from Ember's reset, this caused visual regressions when comparing the Ember app in the shell vs. standalone
- Reducing the shade reset's specificity allows the Ember base stylesheet to take precedence
- The specificity is still high enough to prevent the opposite issue when React apps are hosted within Ember

### DIFF
--- a/ghost/admin/app/styles/patterns/global.css
+++ b/ghost/admin/app/styles/patterns/global.css
@@ -629,11 +629,18 @@ input[type="image"] {
     position: relative;
 }
 
+
 .show {
+    display: block;
+}
+.ember-application .show {
     display: block !important;
 }
 
 .hidden {
+    display: none;
+}
+.ember-application .hidden {
     visibility: hidden !important;
     display: none !important;
 }
--- a/apps/shade/preflight.css
+++ b/apps/shade/preflight.css
@@ -1,4 +1,4 @@
-.shade {
+:where(.shade) {
     /*
     1. Prevent padding and border from affecting element width. (https://github.com/mozdevs/cssremedy/issues/4)
     2. Allow adding a border to an element by just adding a border-width. (https://github.com/tailwindcss/tailwindcss/pull/116)


ghost/admin/app/styles/patterns/global.css
```diff
@@ -629,11 +629,18 @@ input[type="image"] {
     position: relative;
 }
 
+
 .show {
+    display: block;
+}
```
kevinansfield, line 635: What is the purpose for keeping these non-scoped utility classes in Ember's stylesheet?

jonatansberg: I think that's just me being overly cautious. They could likely be safely deleted.

kevinansfield (approved):

kevinansfield:
I couldn't find any issues and can't think of a reason why the additional specificity for the ember styles would be an issue 👍 I do think we could clean up the extra non-scoped classes but not a blocker at all.

jonatansberg:
I can't put my finger on it, but something makes me uneasy about removing those unscoped class names. Since you've tested and verified that it works with them in place, I'd rather leave it as is.

PR #25218, opened 2025-10-22
Added E2E tests for What's New feature with accessibility improvements

rob-ghost:

We are preparing to migrate the What's New feature from Ember to React as part of the admin modernization effort. Before implementing the React version, we need comprehensive E2E test coverage to ensure behavioral parity.

## What does it do?

This PR adds:

1. **Playwright E2E test suite** (`e2e/tests/admin/whats-new.test.ts`)
   - 13 test scenarios covering toast notifications, modal interactions, and badge indicators
   - Uses semantic role-based locators where appropriate
   - Tests against the current Ember implementation

2. **Page Objects** (`e2e/helpers/pages/admin/whats-new/`)
   - `WhatsNewBanner` - Banner notification interactions
   - `WhatsNewModal` - Modal content and interactions
   - `WhatsNewBadges` - Badge indicator locators
   - `WhatsNewMenu` - User menu interactions
   - `WhatsNewPage` - Main orchestrator

3. **UserService helper** (`e2e/helpers/services/user/`)
   - Allows tests to update user accessibility settings via Ghost Admin API
   - Follows existing patterns (similar to SettingsService)

4. **Accessibility improvements to Ember components**
   - Added `role="dialog"` and `aria-modal="true"` to modal
   - Added `aria-label="Dismiss notification"` to dismiss button
   - Added `role="status"` and `aria-live="polite"` to banner notification
   - Added `aria-hidden="true"` to decorative badge indicators (with `data-test-*` attributes)

## Why is this needed?

- **Test Coverage:** Establishes comprehensive test coverage before React migration to prevent regressions
- **Accessibility:** Fixes missing ARIA attributes that should have existed for screen reader users
- **Documentation:** Page objects serve as documentation of the feature's behavior
- **Confidence:** Ensures the React implementation can be validated for behavioral parity

### DIFF
--- a/e2e/helpers/pages/admin/whats-new/WhatsNewBanner.ts
+++ b/e2e/helpers/pages/admin/whats-new/WhatsNewBanner.ts
@@ -1,0 +1,41 @@
+import {AdminPage} from '../AdminPage';
+import {Page, Locator} from '@playwright/test';
+
+export class WhatsNewBanner extends AdminPage {
+    readonly container: Locator;
+    readonly closeButton: Locator;
+    readonly link: Locator;
+    readonly title: Locator;
+    readonly excerpt: Locator;
+
+    constructor(page: Page) {
+        super(page);
+
+        this.container = page.getByRole('status', {name: /what’s new notification/i});
+        this.closeButton = this.container.getByRole('button', {name: /dismiss/i});
+        this.link = this.container.getByRole('link');
+        this.title = this.container.locator('[data-test-toast-title]');
+        this.excerpt = this.container.locator('[data-test-toast-excerpt]');
+    }
+
+    async dismiss(): Promise<void> {
+        await this.closeButton.click();
+        await this.container.waitFor({state: 'hidden'});
+    }
+
+    async clickLink(): Promise<void> {
+        await this.link.click();
+    }
+
+    async clickLinkAndClosePopup(): Promise<void> {
+        const [popup] = await Promise.all([
+            this.page.waitForEvent('popup'),
+            this.clickLink()
+        ]);
+        await popup.close();
+    }
+
+    async waitForBanner(): Promise<void> {
+        await this.container.waitFor({state: 'visible'});
+    }
+}
--- a/e2e/helpers/pages/admin/whats-new/WhatsNewMenu.ts
+++ b/e2e/helpers/pages/admin/whats-new/WhatsNewMenu.ts
@@ -1,0 +1,34 @@
+import {AdminPage} from '../AdminPage';
+import {Page, Locator} from '@playwright/test';
+import {WhatsNewModal} from './WhatsNewModal';
+
+export class WhatsNewMenu extends AdminPage {
+    readonly userMenuTrigger: Locator;
+    readonly whatsNewMenuItem: Locator;
+    readonly avatarBadge: Locator;
+    readonly menuBadge: Locator;
+
+    constructor(page: Page) {
+        super(page);
+
+        this.userMenuTrigger = page.locator('[data-test-nav="arrow-down"]');
+        this.whatsNewMenuItem = page.getByRole('menuitem', {name: 'What’s new?'});
+        this.avatarBadge = page.locator('[data-test-whats-new-avatar-badge]');
+        this.menuBadge = page.locator('[data-test-whats-new-menu-badge]');
+    }
+
+    async openUserMenu(): Promise<void> {
+        await this.userMenuTrigger.click();
+        await this.whatsNewMenuItem.waitFor({state: 'visible'});
+    }
+
+    async openWhatsNewModal(): Promise<WhatsNewModal> {
+        await this.openUserMenu();
+        await this.whatsNewMenuItem.click();
+
+        const modal = new WhatsNewModal(this.page);
+        await modal.waitForModal();
+
+        return modal;
+    }
+}
--- a/e2e/helpers/pages/admin/whats-new/WhatsNewModal.ts
+++ b/e2e/helpers/pages/admin/whats-new/WhatsNewModal.ts
@@ -1,0 +1,49 @@
+import {Page, Locator} from '@playwright/test';
+
+interface ModalEntry {
+    title: string;
+    excerpt: string;
+    hasImage: boolean;
+}
+
+export class WhatsNewModal {
+    private readonly page: Page;
+    readonly modal: Locator;
+    readonly title: Locator;
+    readonly entries: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+
+        this.modal = page.getByRole('dialog', {name: /what’s new/i});
+        this.title = this.modal.getByRole('heading', {name: /what’s new/i});
+        this.entries = this.modal.locator('[data-test-entry]');
+    }
+
+    async waitForModal(): Promise<void> {
+        await this.modal.waitFor({state: 'visible'});
+    }
+
+    async close(): Promise<void> {
+        await this.page.keyboard.press('Escape');
+        await this.modal.waitFor({state: 'hidden'});
+    }
+
+    async getEntries(): Promise<ModalEntry[]> {
+        const entryElements = await this.entries.all();
+
+        return Promise.all(
+            entryElements.map(async (entry) => {
+                const titleText = await entry.locator('[data-test-entry-title]').textContent();
+                const excerptText = await entry.locator('[data-test-entry-excerpt]').textContent();
+                const hasImage = await entry.locator('[data-test-entry-image]').count() > 0;
+
+                return {
+                    title: titleText?.trim() || '',
+                    excerpt: excerptText?.trim() || '',
+                    hasImage
+                };
+            })
+        );
+    }
+}
--- a/e2e/tests/admin/whats-new.test.ts
+++ b/e2e/tests/admin/whats-new.test.ts
@@ -1,0 +1,308 @@
+import {test, expect} from '../../helpers/playwright/fixture';
+import {WhatsNewBanner, WhatsNewMenu} from '../../helpers/pages/admin/whats-new';
+import type {Page} from '@playwright/test';
+
+interface ChangelogEntry {
+    title: string;
+    custom_excerpt: string;
+    published_at: string;
+    url: string;
+    featured: boolean;
+    feature_image?: string;
+}
+
+function daysAgo(days: number): Date {
+    const date = new Date();
+    date.setDate(date.getDate() - days);
+    return date;
+}
+
+function daysFromNow(days: number): Date {
+    const date = new Date();
+    date.setDate(date.getDate() + days);
+    return date;
+}
+
+function createEntry(publishedAt: Date, options: {
+    featured?: boolean;
+    title?: string;
+    excerpt?: string;
+    feature_image?: string;
+} = {}): ChangelogEntry {
+    const title = options.title ?? 'Test Update';
+    return {
+        title,
+        custom_excerpt: options.excerpt ?? 'Test feature',
+        published_at: publishedAt.toISOString(),
+        url: `https://ghost.org/changelog/${title.toLowerCase().replace(/\s+/g, '-')}`,
+        featured: options.featured ?? false,
+        ...(options.feature_image && {feature_image: options.feature_image})
+    };
+}
+
+async function mockChangelog(page: Page, entries: ChangelogEntry[]): Promise<void> {
+    await page.route('https://ghost.org/changelog.json', async (route) => {
+        await route.fulfill({
+            status: 200,
+            contentType: 'application/json',
+            body: JSON.stringify({
+                posts: entries,
+                changelogUrl: 'https://ghost.org/changelog/'
+            })
+        });
+    });
+}
+
+test.describe('Ghost Admin - What\'s New', () => {
+    test.describe('banner notification', () => {
+        test('shows banner for new featured entries the user has not seen', async ({page}) => {
+            await mockChangelog(page, [
+                createEntry(daysFromNow(1), {
+                    featured: true,
+                    title: 'New Featured Update',
+                    excerpt: 'This is an exciting new feature'
+                })
+            ]);
+
+            const banner = new WhatsNewBanner(page);
+            await banner.goto();
+            await banner.waitForBanner();
+
+            await expect(banner.container).toBeVisible();
+            await expect(banner.title).toHaveText('New Featured Update');
+            await expect(banner.excerpt).toHaveText('This is an exciting new feature');
+        });
+
+        test('does not show banner for entries from before user joined', async ({page}) => {
+            await mockChangelog(page, [createEntry(daysAgo(30), {featured: true})]);
+
+            const banner = new WhatsNewBanner(page);
+            await banner.goto();
+
+            await expect(banner.container).not.toBeVisible();
+        });
+
+        test('does not show banner when there are no entries', async ({page}) => {
+            await mockChangelog(page, []);
+
+            const banner = new WhatsNewBanner(page);
+            await banner.goto();
+
+            await expect(banner.container).not.toBeVisible();
+        });
+
+        test('does not show banner when latest entry is not featured', async ({page}) => {
+            await mockChangelog(page, [createEntry(daysFromNow(1))]);
+
+            const banner = new WhatsNewBanner(page);
+            await banner.goto();
+
+            await expect(banner.container).not.toBeVisible();
+        });
+
+        test.describe('dismissal behavior', () => {
+            test('hides banner immediately when close button is clicked', async ({page}) => {
+                await mockChangelog(page, [createEntry(daysFromNow(1), {featured: true})]);
+
+                const banner = new WhatsNewBanner(page);
+                await banner.goto();
+                await banner.waitForBanner();
+
+                await expect(banner.container).toBeVisible();
+
+                await banner.dismiss();
+
+                await expect(banner.container).not.toBeVisible();
+            });
+
+            test('hides banner immediately when link is clicked', async ({page}) => {
+                await mockChangelog(page, [createEntry(daysFromNow(1), {featured: true})]);
+
+                const banner = new WhatsNewBanner(page);
+                await banner.goto();
+                await banner.waitForBanner();
+
+                await expect(banner.container).toBeVisible();
+
+                await banner.clickLinkAndClosePopup();
+
+                await expect(banner.container).not.toBeVisible();
+            });
+
+            test('hides banner immediately when modal is opened', async ({page}) => {
+                await mockChangelog(page, [
+                    createEntry(daysFromNow(1), {featured: true, feature_image: 'https://ghost.org/image1.jpg'}),
+                    createEntry(daysAgo(5))
+                ]);
+
+                const banner = new WhatsNewBanner(page);
+                const menu = new WhatsNewMenu(page);
+
+                await banner.goto();
+                await banner.waitForBanner();
+
+                await expect(banner.container).toBeVisible();
+
+                const modal = await menu.openWhatsNewModal();
+                await modal.close();
+
+                await expect(banner.container).not.toBeVisible();
+            });
+
+            test('banner remains hidden after reload when dismissed', async ({page}) => {
+                await mockChangelog(page, [createEntry(daysFromNow(1), {featured: true})]);
+
+                const banner = new WhatsNewBanner(page);
+                await banner.goto();
+                await banner.waitForBanner();
+
+                await banner.dismiss();
+
+                await banner.goto();
+                await expect(banner.container).not.toBeVisible();
+            });
+
+            test('banner reappears when a new entry is published after dismissal', async ({page}) => {
+                await mockChangelog(page, [createEntry(daysFromNow(1), {featured: true})]);
+
+                const banner = new WhatsNewBanner(page);
+
+                await banner.goto();
+                await banner.waitForBanner();
+                await banner.dismiss();
+
+                await banner.goto();
+                await expect(banner.container).not.toBeVisible();
+
+                await mockChangelog(page, [
+                    createEntry(daysFromNow(2), {
+                        featured: true,
+                        title: 'Second Featured Update'
+                    })
+                ]);
+
+                await banner.goto();
+                await banner.waitForBanner();
+
+                await expect(banner.container).toBeVisible();
+                await expect(banner.title).toHaveText('Second Featured Update');
+            });
+        });
+    });
+
+    test.describe('modal', () => {
+        test('shows modal with all entries when opened from user menu', async ({page}) => {
+            await mockChangelog(page, [
+                createEntry(daysFromNow(1), {
+                    featured: true,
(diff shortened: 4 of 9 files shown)


rob-ghost:
@coderabbitai review

e2e/helpers/pages/admin/whats-new/WhatsNewMenu.ts
```diff
@@ -0,0 +1,41 @@
...
+        return modal;
+    }
+
+    get avatarBadge(): Locator {
+        return this.badges.avatarBadge;
+    }
+
+    get menuBadge(): Locator {
+        return this.badges.menuBadge;
+    }
```
jonatansberg: I think we can add these as properties for the menu. The extra page object and getters add indirection.

ghost/admin/app/components/gh-nav-menu/footer-banner.hbs
```diff
@@ -11,8 +11,8 @@
 
 {{#if (and this.showWhatsNew this.whatsNew.hasNew)}}
     {{#let (get this.whatsNew.entries "0") as |entry|}}
-    <div class="gh-sidebar-banner gh-whatsnew-toast" data-test-toast="whats-new">
-        <button class="gh-sidebar-banner-close" data-test-toast-close type="button" {{on "click" this.dismissWhatsNewToast}}>&#xd7;</button>
+    <div class="gh-sidebar-banner gh-whatsnew-toast" data-test-toast="whats-new" role="status" aria-label="What's new notification" aria-live="polite">
+        <button class="gh-sidebar-banner-close" data-test-toast-close type="button" aria-label="Dismiss notification" {{on "click" this.dismissWhatsNewToast}}>&#xd7;</button>
```
jonatansberg: Nice! 👍

e2e/helpers/pages/admin/whats-new/WhatsNewBanner.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export class WhatsNewBanner extends BasePage {
+    readonly container: Locator;
+    readonly closeButton: Locator;
+    readonly link: Locator;
+    readonly title: Locator;
+    readonly excerpt: Locator;
+
+    constructor(page: Page) {
+        super(page, '/ghost');
```
jonatansberg: I think you can extend the AdminPage instead and that will handle the base URL for you.

e2e/helpers/pages/admin/whats-new/WhatsNewModal.ts
```diff
@@ -0,0 +1,51 @@
...
+        this.page = page;
+
+        this.modal = page.getByRole('dialog', {name: /what's new/i})
+            .or(page.locator('[data-test-modal="whats-new"]'));
+        this.title = this.modal.getByRole('heading', {name: /what's new/i});
+        this.entries = this.modal.locator('[data-test-entry]');
+    }
+
+    async waitForModal(): Promise<void> {
+        await this.modal.waitFor({state: 'visible', timeout: WHATS_NEW_WAIT_TIMEOUT});
```
jonatansberg: Isn't the default timeout enough?

e2e/tests/admin/whats-new.test.ts
```diff
@@ -0,0 +1,263 @@
...
+    test.beforeEach(async ({page}) => {
+        const userService = createUserService(page.request);
+        await userService.updateCurrentUserAccessibility({
+            whatsNew: {
+                lastSeenDate: OLD_LAST_SEEN_DATE
+            }
+        });
+    });
+
+    test.describe('toast notification', () => {
```
jonatansberg: Maybe we shouldn't keep calling it a toast if it's a banner.

e2e/tests/admin/whats-new.test.ts
```diff
@@ -0,0 +1,263 @@
...
+        test('shows toast with latest entry when there are new featured entries', async ({page}) => {
+            await mockChangelog(page, [ENTRIES.newFeatured]);
+            const whatsNewPage = new WhatsNewPage(page);
+
+            await whatsNewPage.goto();
```
jonatansberg: I don't think you need the goto here? Or maybe you do for the mocking to work? 

Either way, the test cases that do not rely on mocks can use the page object straight away since we're loading the dashboard by default.

e2e/helpers/pages/admin/whats-new/WhatsNewPage.ts
```diff
@@ -0,0 +1,19 @@
...
+    readonly menu: WhatsNewMenu;
+
+    constructor(page: Page) {
+        super(page);
+
+        this.banner = new WhatsNewBanner(page);
+        this.modal = new WhatsNewModal(page);
+        this.menu = new WhatsNewMenu(page);
+    }
+}
```
jonatansberg: I think we could use the banner, modal, and menu page objects directly, since you're most of the time only interacting with one of them at a time anyways.

jonatansberg (commented):
I need to run and pick up my kids, so I haven't had time to review everything in full. On a high level, I think this looks good. Added some comments, questions, and nitpicks. 

I'd also be interested in seeing @ibalosh's review.

ghost/admin/app/components/gh-nav-menu/footer-banner.hbs
ibalosh: agreed!

e2e/helpers/pages/admin/whats-new/WhatsNewModal.ts
ibalosh: yeah, I feel the same, wondering if we need it. If we really do, we could keep timeout constant close to the page objects, rather than here near tests. We will use it only on locators anyways

e2e/tests/admin/whats-new.test.ts
```diff
@@ -0,0 +1,263 @@
...
+        });
+    });
+}
+
+test.describe('Ghost Admin - What\'s New', () => {
+    test.beforeEach(async ({page}) => {
+        const userService = createUserService(page.request);
+        await userService.updateCurrentUserAccessibility({
+            whatsNew: {
+                lastSeenDate: OLD_LAST_SEEN_DATE
```
ibalosh: since OLD_LAST_SEEN_DATE is used only by this test, constant seems to be too much, we could just keep it here? It would make test easier to read, and in tests it's fine to hard code it here for readability

e2e/helpers/pages/admin/whats-new/WhatsNewBadges.ts
```diff
@@ -0,0 +1,14 @@
+import {BasePage} from '../../BasePage';
+import {Page, Locator} from '@playwright/test';
+
+export class WhatsNewBadges extends BasePage {
+    readonly avatarBadge: Locator;
+    readonly menuBadge: Locator;
+
+    constructor(page: Page) {
+        super(page, '/ghost');
```
ibalosh: you can use admin as parent for these

e2e/helpers/pages/admin/whats-new/WhatsNewModal.ts
```diff
@@ -0,0 +1,51 @@
...
+    private readonly page: Page;
+    readonly modal: Locator;
+    readonly title: Locator;
+    readonly entries: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+
+        this.modal = page.getByRole('dialog', {name: /what's new/i})
+            .or(page.locator('[data-test-modal="whats-new"]'));
```
ibalosh: hm, curious, would this OR create issues, or is it always going to catch correct element on the page?

e2e/tests/admin/whats-new.constants.ts
```diff
@@ -0,0 +1,2 @@
+export const WHATS_NEW_WAIT_TIMEOUT = 5000;
```
ibalosh: have a feeling we can probably ditch this file by moving timeouts to page objects what's new aread, and date as value to test

e2e/tests/admin/whats-new.test.ts
```diff
@@ -0,0 +1,263 @@
...
+        });
+
+        test('does not show toast again after clicking link', async ({page}) => {
+            await mockChangelog(page, [ENTRIES.newFeatured]);
+            const whatsNewPage = new WhatsNewPage(page);
+
+            await whatsNewPage.goto();
+            await whatsNewPage.banner.waitForBanner();
+
+            const [popup] = await Promise.all([
```
ibalosh: wondering if we can hide this promise by encapsulating this into page object, with something like:

`whatsNewPage.waitForPopup`

ibalosh (changes requested):
Looks nice too me, added couple of things here and there to smoothen it out a bit, but overall 👌

e2e/helpers/pages/admin/whats-new/WhatsNewModal.ts
rob-ghost: Yeah I think this was set to 5s instead of the default 30s to make the local testing loop a little faster. I prefer the code when using the default (and it removes the constant) so I've reverted back to default.

e2e/helpers/pages/admin/whats-new/WhatsNewBanner.ts
rob-ghost: Done!

e2e/helpers/pages/admin/whats-new/WhatsNewMenu.ts
rob-ghost: Agreed, flattened this to remove the indirection.

e2e/helpers/pages/admin/whats-new/WhatsNewPage.ts
rob-ghost: Agreed, the orchestrator was overdoing it. I've removed it in favour of the tests interacting with the individual pages.

e2e/tests/admin/whats-new.test.ts
rob-ghost: Since this became the last constant it's nicer to have it inline so I've updated that 👍🏻

e2e/helpers/pages/admin/whats-new/WhatsNewModal.ts
rob-ghost: Yes, I think it will, this was overly defensive for a test. If we can't get the modal by role then it should fail rather than silently fallback. `.or` was an anti-pattern here. Removed 👍🏻

e2e/tests/admin/whats-new.constants.ts
rob-ghost: Done!

e2e/tests/admin/whats-new.test.ts
rob-ghost: If I'm not mistaken we need to trigger a fresh page reload after calling `mockChangelog` to get the mocked/seeded changelog entries to be pulled down. The previous page will have out-of-date entries so any test which calls `mockChangelog` needs to call `goto` (AFAIK).

e2e/tests/admin/whats-new.test.ts
rob-ghost: Agreed, I've renamed the tests to refer to `banner` instead.

e2e/tests/admin/whats-new.test.ts
rob-ghost: Nice, I moved it to the page object as `clickLinkAndClosePopup`

e2e/tests/admin/whats-new.test.ts
```diff
@@ -0,0 +1,259 @@
...
+
+            await expect(menu.avatarBadge).toBeVisible();
+
+            const modal = await menu.openWhatsNewModal();
+            await modal.close();
+
+            await expect(menu.avatarBadge).not.toBeVisible();
+
+            await menu.goto();
+            await expect(menu.avatarBadge).not.toBeVisible();
```
jonatansberg: Nit: I see this pattern in a few different places. What I assume you're testing is that behavior consistent even after a page reload? It would be nice with a comment or something that calls that out so that this isn't mistaken as an error or cruft in the future.

e2e/helpers/pages/admin/whats-new/WhatsNewModal.ts
jonatansberg: We could definitely have a think about whether we should tighten up the default. 30 seconds is a long time. Something like 10 seconds has worked well for me in the past as a default. It's long enough that things usually have time to load, but still relatively short so that tests don't stall forever before failing.

jonatansberg (commented):
Looks good! 👍  I added a small nit pick that you can address if you want.

e2e/tests/admin/whats-new.test.ts
rob-ghost: Yeah a lot of the dismissal tests for the banner and badge will test that the element is no longer visible on dismiss, and stays that way after a page reload (to ensure persistence).

Would it be preferable to actually have a separate test which illustrates this, instead of this logic in each test + comment?

```js
test('removes badges after viewing What\'s new', () => {
    // ...
    await expect(menu.avatarBadge).not.toBeVisible();

    // Expect that the badge remains removed even after a page reload
    await menu.openUserMenu();
    await expect(menu.menuBadge).not.toBeVisible();
})

// === vs ===

test('removes badges after viewing What\'s new', () => {
    // Same test but without the reload and expect.not.toBeVisible();
})

test('badge does not return on reload after dismissal', () => {
    // Effectively the same test as before but with the extra reload and
    // expect.not.toBeVisible();
})
```

ibalosh: yeah, it's always better if it's a separate test, but with one caveat - as long as there is no big performance penalty for it.

In this case, I guess the performance question is only around mock changelog. Will this be a real API call at once?

In e2e we are going for real call if the service is not external (not under our control)

e2e/helpers/pages/admin/whats-new/WhatsNewBanner.ts
```diff
@@ -0,0 +1,41 @@
...
+    readonly link: Locator;
+    readonly title: Locator;
+    readonly excerpt: Locator;
+
+    constructor(page: Page) {
+        super(page);
+
+        this.container = page.getByRole('status', {name: /what's new notification/i});
+        this.closeButton = this.container.getByRole('button', {name: /dismiss/i});
+        this.link = this.container.getByRole('link');
```
ibalosh, line 16: wondering, will this locator be too general, and multiple links can be identified by it, that could cause issues with tests

e2e/tests/admin/whats-new.test.ts
rob-ghost: From what I understand, we're mocking and self-fulfilling the changelog request (so its not actually making a request).

ibalosh (approved):
looks good to me @rob-ghost , added couple of comments here and there, nothing blocking the approval

my only question would be the mocked data. We strive for not having mocks in e2e, if it's not external (stripe etc).

In this case, I guess it is, since it's coming from ghost.org API, so I think it's fine, just want to make sure.

e2e/helpers/pages/admin/whats-new/WhatsNewBanner.ts
rob-ghost: From a semantic POV I think if there's multiple links then there's a design / accessibility problem with the Banner? Hopefully that kind of issue would cause a failing test.

e2e/tests/admin/whats-new.test.ts
rob-ghost: Do you have any advice @ibalosh on how we might control the response from `/changelog.json` without mocking? I'm just unfamiliar with data generation / setup in e2e tests.

In this case we're testing a few different variations which might require a setup per-test if we're doing it without a mock and using a `ChangelogService` or some other kind of data hydration.

ibalosh: yeah, I wrote that incorrectly :) since we mock, there is not going to be performance penalty. 

I was thinking if we are going to be changing this for real call later. I am guessing this is an external service, so in test we are going to keep it mocked.

ghost/admin/app/components/gh-nav-menu/footer.hbs
```diff
@@ -53,9 +53,9 @@
                         {{else}}
                             <li>
                                 <button class="dropdown-item" tabindex="-1" data-test-nav="whatsnew" type="button" {{on "click" this.openWhatsNew}}>
-                                    What’s new?
+                                    What's new?
```
kevinansfield: Claude being helpful again here? :smiling

e2e/helpers/pages/admin/whats-new/WhatsNewMenu.ts
```diff
@@ -0,0 +1,34 @@
...
+    readonly avatarBadge: Locator;
+    readonly menuBadge: Locator;
+
+    constructor(page: Page) {
+        super(page);
+
+        this.userMenuTrigger = page.locator('[data-test-nav="arrow-down"]');
+        this.whatsNewMenuItem = page.locator('[data-test-nav="whatsnew"]');
+        this.avatarBadge = page.locator('[data-test-whats-new-avatar-badge]');
+        this.menuBadge = page.locator('[data-test-whats-new-menu-badge]');
```
kevinansfield, line 17: We have some inconsistent aria roles for this menu - the menu itself has `role="menu"` but only _some_ items within it have `role="menuitem"`. Is it worth cleaning those up so we can use role selectors or switching to `data-testid=` where appropriate now so that we don't have to change the tests when testing against the React-rendered sidebar?

ghost/admin/app/components/gh-nav-menu/footer.hbs
rob-ghost: Yes 😠 so annoying because it initially kept it and then edited it later 🙄 I'll fix this

rob-ghost: Fixed

e2e/helpers/pages/admin/whats-new/WhatsNewMenu.ts
rob-ghost: Great catch! I did a review of the roles and this looks like the only one that was missing. I've added the role and updated the paged object.

ghost/admin/app/components/modals/whats-new.hbs
```diff
@@ -1,5 +1,5 @@
-<div class="modal-content" data-test-modal="whats-new">
-    <h1 class="gh-whasnew-modal-title" data-test-modal-title>What’s new?</h1>
+<div class="modal-content" data-test-modal="whats-new" role="dialog" aria-modal="true" aria-labelledby="whats-new-modal-title">
+    <h1 id="whats-new-modal-title" class="gh-whasnew-modal-title" data-test-modal-title>What's new?</h1>
```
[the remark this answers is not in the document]
rob-ghost: Existing typo, the style sheet also has the same typo so its not an issue.

e2e/tests/admin/whats-new.test.ts
rob-ghost: I've split these concerns into two distinct test cases/suites, one for immediate dismissal, and one for persistent dismissal which should separate the concern and remove the duplication of the unclear expectation after reload.

ghost/admin/app/components/modals/whats-new.hbs
jonatansberg: Another ` -> '

e2e/helpers/services/user/UserService.ts
```diff
@@ -0,0 +1,28 @@
...
+export class UserService {
+    private readonly request: APIRequest;
+    private readonly adminEndpoint: string;
+
+    constructor(request: APIRequest) {
+        this.request = request;
+        this.adminEndpoint = '/ghost/api/admin';
+    }
+
+    async updateCurrentUserAccessibility(accessibility: UserAccessibility): Promise<void> {
```
jonatansberg: Nit: Maybe we should give this a more meaningful name and make it update only the last seen? I'm not a fan of how we're leaking that questionable implementation into the tests.

jonatansberg (commented):
Two little nitpicks

ghost/admin/app/components/modals/whats-new.hbs
rob-ghost: 🫠 I reviewed the diff to make sure this didn't happen again and I missed it 🙃

It's like whack-a-mole, I fix it manually, and then when an agent is touching this hunk it just goes and re-writes it again. It seems that at least Claude Code doesn't understand `’` and so it will just overwrite it every time and if you're not watching the diffs close enough it can sneak back in.

I wonder if there's anything we can do about that?

In the meantime, I'll fix this 🔨

e2e/helpers/services/user/UserService.ts
rob-ghost: Yes its a good point, I've thought about this and my opinion is that `UserService` should reflect the underlying API its interfacing with and not abstract any of the details because then more business logic gets coupled with the test services.

Since the API refers to `accessibility`, then so should the service.

Happy to hear alternative ideas though! I think if we were to abstract the fact that `whatsNew` belongs to `user.accessibility` then we wouldn't define this `UserService` but instead be `WhatsNewService` which a function like `updateLastSeenDate`.

I think this might make the situation with `accessibility` worse because now we'll have coupled the test code to that concept and would have to update accordingly. If we stick with coupling to the API we only have to change tests when the User API changes which is better for cohesion.

---

It does raise the question if making API requests through Services is the right thing to be doing in these E2E tests? Should the E2E tests instead start from an assumed blank slate and run through the User scenarios?

For eg:
1. By default there's no entries and so no banner
2. If there's entries from the past, no banner
3. If there's newer entries, banner
4. Interacting with these entries, banner dismissed
5. Reloading, the banners are still dismissed
6. Any newer entry will show the banner again

For these tests we don't use the `UserService` at all, instead we're mocking `changelog.json` to show entries with certain timestamps. This would de-couple the tests from `accessibility` entirely because its literally just using the interface and we're only coupled to `changelog.json` for seeding.

jonatansberg: Those are valid concerns. Stepping away from manipulating the backend state via the API would definitely make the tests more portable should the backend implementation change in the future.

If we can make the same assertions without relying on this service at all, then I think that's better.

rob-ghost: Ok so I re-phrased the test suite as user behaviour tests instead of the more unit-test style and I think this works better as e2e tests.

This made it so I could remove the entire `UserService` and just power the behaviour through what's in the changelog.

I prefer this, have a look and let me know what you think.

jonatansberg (approved):
That's definitely better. 

Maybe you could add a small comment on the somewhat unorthodox approach of verifying that things that "happen in the future" show up?

rob-ghost:
> Maybe you could add a small comment on the somewhat unorthodox approach of verifying that things that "happen in the future" show up?

Specifically the `shows banner for new featured entries the user has not seen` test which has `daysFromNow(1)`?

cmraible (approved):
LGTM, nice work! Tests are super easy to read through and understand 🙌

PR #25221, opened 2025-10-22
Fall back to Ember for auth

jonatansberg:
Ref https://linear.app/ghost/issue/BER-2917/fall-back-to-ember-for-authentication

If we can't fetch the current user, fall back to displaying the Ember app so that it can handle authentication for us. Once the user navigates away from any known auth route, refetch the user so that we're up to date on the React side.

This is a temporary solution to handle auth until we bridge the Ember state properly or re-implement auth on the React side.

### DIFF
--- a/apps/admin/src/ember-bridge/EmberAuthSync.tsx
+++ b/apps/admin/src/ember-bridge/EmberAuthSync.tsx
@@ -16,11 +16,8 @@ const isAuthRoute = (path: string) => {
     return AUTH_ROUTES.has(cleanPath);
 };
 
-/**
- * Component that monitors route changes and invalidates React Query caches
- * when navigating away from authentication routes in Ember
- */
-export function EmberAuthSync() {
+
+export function useEmberAuthSync() {
     const location = useLocation();
     const queryClient = useQueryClient();
     const previousPathRef = useRef<string | null>(null);
@@ -37,6 +34,4 @@ export function EmberAuthSync() {
         // Update the previous path reference
         previousPathRef.current = currentPath;
     }, [location, queryClient]);
-
-    return null;
 }
--- a/apps/admin/src/ember-bridge/EmberAuthSync.tsx
+++ b/apps/admin/src/ember-bridge/EmberAuthSync.tsx
@@ -1,16 +1,19 @@
-import { useEffect, useRef } from 'react';
-import { useLocation } from '@tryghost/admin-x-framework';
-import { useQueryClient } from '@tanstack/react-query';
+import { useEffect, useRef } from "react";
+import { useLocation } from "@tryghost/admin-x-framework";
+import { useQueryClient } from "@tanstack/react-query";
 
-/**
- * Helper function to check if a path is an authentication-related route
- */
-const isAuthRoute = (path: string): boolean => {
-    return path.startsWith('/setup') ||
-           path.startsWith('/signin') ||
-           path.startsWith('/signout') ||
-           path.startsWith('/signup') ||
-           path.startsWith('/reset');
+const AUTH_ROUTES = new Set(["setup", "signin", "signout", "signup", "reset"]);
+
+const isAuthRoute = (path: string) => {
+    if (!path) {
+        return false;
+    }
+    
+    // Remove leading slash and extract the first path segment
+    // Also strip any query params or hash fragments
+    const cleanPath = path.replace(/^\//, '').split(/[/?#]/)[0];
+    
+    return AUTH_ROUTES.has(cleanPath);
 };
 
 /**
@@ -37,4 +40,3 @@ export function EmberAuthSync() {
 
     return null;
 }
-
--- a/apps/admin/src/ember-bridge/EmberAuthSync.tsx
+++ b/apps/admin/src/ember-bridge/EmberAuthSync.tsx
@@ -5,10 +5,6 @@ import { useQueryClient } from "@tanstack/react-query";
 const AUTH_ROUTES = new Set(["setup", "signin", "signout", "signup", "reset"]);
 
 const isAuthRoute = (path: string) => {
-    if (!path) {
-        return false;
-    }
-    
     // Remove leading slash and extract the first path segment
     // Also strip any query params or hash fragments
     const cleanPath = path.replace(/^\//, '').split(/[/?#]/)[0];
--- a/apps/admin/src/ember-bridge/EmberAuthSync.tsx
+++ b/apps/admin/src/ember-bridge/EmberAuthSync.tsx
@@ -1,0 +1,40 @@
+import { useEffect, useRef } from 'react';
+import { useLocation } from '@tryghost/admin-x-framework';
+import { useQueryClient } from '@tanstack/react-query';
+
+/**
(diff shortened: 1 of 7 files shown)


apps/admin/src/main.tsx
```diff
@@ -2,9 +2,45 @@ import { StrictMode } from "react";
...
 
 createRoot(document.getElementById("root")!).render(
     <StrictMode>
-        <App />
+        <AppProvider>
+            <FrameworkProvider {...framework}>
+                <RouterProvider prefix={"/"} routes={routes}>
+                    <ShadeApp
+                        className="shade-admin"
+                        darkMode={true}
```
rob-ghost: **Issue:** In the [previous diff of App.tsx](https://github.com/TryGhost/Ghost/pull/25221/files#diff-49c534015f1415829a732211113c75efb25783c5953230067857697eb77893c8L36) `darkMode={false}` and its set to `true` here, was this intentional / does it matter?

apps/admin/src/ember-bridge/EmberAuthSync.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const queryClient = useQueryClient();
+    const previousPathRef = useRef<string | null>(null);
+
+    useEffect(() => {
+        const currentPath = location.pathname;
+        const previousPath = previousPathRef.current;
+
+        // If we're navigating away from an auth route, invalidate all caches
+        if (previousPath && isAuthRoute(previousPath) && !isAuthRoute(currentPath)) {
+            void queryClient.invalidateQueries();
```
rob-ghost, line 27: **Question:** Why are we choosing to fire-and-forget invalidating the queries instead of `await`ing it? Is there any race-condition potential if we're navigating quickly from an auth route, to an un-auth route, back to an auth route and what would the impact be (if any)?

I assume its safe and intentional (due to `void`), but a comment to explain might help.

apps/admin/src/ember-bridge/EmberAuthSync.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Helper function to check if a path is an authentication-related route
+ */
+const isAuthRoute = (path: string): boolean => {
+    return path.startsWith('/setup') ||
+           path.startsWith('/signin') ||
+           path.startsWith('/signout') ||
+           path.startsWith('/signup') ||
+           path.startsWith('/reset');
```
rob-ghost: **Question:** Should we be more explicit than `startsWith`? We probably don't currently have any routes called `/setup-integration` or something but that would be considered an auth route in this case.

I assume we're attempting to match the base component of the path, in which case should we split and compare the first element instead?

```js
const AUTH_ROUTES = new Set([
    'setup',
    'signin',
    'signout',
    'signup',
    'reset'
]);

const isAuthRoute = (path) => {
    const base = path.split('\')[0];

    return AUTH_ROUTES.has(base);
}
```

apps/admin/src/ember-bridge/EmberAuthSync.tsx
```diff
@@ -0,0 +1,40 @@
...
+};
+
+/**
+ * Component that monitors route changes and invalidates React Query caches
+ * when navigating away from authentication routes in Ember
+ */
+export function EmberAuthSync() {
```
rob-ghost: **Question:** Why is this a Component and not a Hook?

Is there a reason this couldn't be `useEmberAuthSync` and have it called from `App`?

rob-ghost (commented):
Overall approach is good 👍🏻

I have a couple of questions and a potential issue.

apps/admin/src/main.tsx
jonatansberg: That must have slipped through as I've been rewriting the history quite heavily as part of getting the spike into a mergeable state. I don't think it matters much, but I could clean that up regardless.

apps/admin/src/ember-bridge/EmberAuthSync.tsx
jonatansberg: We could do that. I don't think it matters much as this is something we ought to throw away as soon as we have a better way of syncing the state.

apps/admin/src/ember-bridge/EmberAuthSync.tsx
jonatansberg: No, there isn't a good reason (other than possibly old habits 😅)

It could definitely be a hook.

apps/admin/src/ember-bridge/EmberAuthSync.tsx
jonatansberg: Effects aren't allowed to be async themselves. Awaiting this would require us to wrap the effect in yet another function. 

I think explicit awaits are enabled in the linting rules for the admin project, so inserting void is necessary to keep the linter from yelling at us.

rob-ghost: > Effects aren't allowed to be async themselves

💯 I didn't trace this to the `useEffect`, this makes sense.

Any thoughts on the potential for a race when moving between auth -> no auth -> auth and how it might affect the caches? I presume its safe but do you have any ideas?

For example, moving auth -> no auth starts the invalidation, but then when we move no auth -> auth quickly, could the component fetch from cache, only to have the cache invalidated immediately after?

apps/admin/src/ember-bridge/EmberAuthSync.tsx
rob-ghost: Just to clarify, since this was a **Question** comment, no need to make changes if you don't think they're needed! Happy to hear "this is temporary" or "not worth changing" as a response, this wouldn't have blocked an approval.

**Suggestion** comments are the ones which would usually be blockers. Appreciate you taking the code suggestion on board though and handling the edge cases my example didn't!

apps/admin/src/ember-bridge/EmberAuthSync.tsx
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useRef } from "react";
+import { useLocation } from "@tryghost/admin-x-framework";
+import { useQueryClient } from "@tanstack/react-query";
+
+const AUTH_ROUTES = new Set(["setup", "signin", "signout", "signup", "reset"]);
+
+const isAuthRoute = (path: string) => {
+    if (!path) {
+        return false;
+    }
```
rob-ghost: **Nitpick(non-blocking):** Since `path: string`, this is guarding against empty strings?

I don't think we need to be defensive here since the path cleaning will return `''` for an empty string and this won't match anything in `AUTH_ROUTES` so its safe.

rob-ghost (approved):
Thanks for addressing the review comments! Nothing blocking remaining 👍🏻

apps/admin/src/ember-bridge/EmberAuthSync.tsx
jonatansberg: I haven't tested this, but my understanding is that React Query should handle that just fine. It's kind of the entire premise of the library :)

Worst case, we overfetch a bit.

PR #25230, opened 2025-10-23
Merge v6.5.1 into main

allouis:


allouis:
I tested rebasing a merge commit here https://github.com/TryGhost/Ghost/pull/25231 and GitHub removed the merge commit.

The resulting branch contains all the code changes we want, but it is missing the merge commit which every other release has - I'm closing this now on the assumption that we would like to have some consistency with our releases

PR #25232, opened 2025-10-23
✨ Added Banner component to Shade design system

rob-ghost:

## What does this PR do?

Adds a new generic `Banner` component to the Shade design system that serves as the foundation for migrating the What's New feature from Ember to React.

We're migrating the What's New notification feature from Ember to React as part of the broader admin modernization effort. Rather than creating a What's New-specific component, this PR introduces a reusable Banner component that can serve multiple use cases:

- What's New notifications (primary use case)
- Referral invite banners
- Error notifications
- Any future inline sidebar notifications

This separation of concerns makes the component more maintainable, testable, and aligns with the design system principles.

### New Components
- `apps/shade/src/components/ui/banner.tsx` - Generic Banner component
- `apps/shade/src/components/ui/banner.stories.tsx` - Comprehensive Storybook documentation
- `apps/shade/test/unit/components/ui/banner.test.tsx` - Unit tests (21 tests, 100% coverage)

### Features
- **6 semantic variants**: default, gradient, info, success, warning, destructive
- **3 size options**: sm, md, lg
- **Dismissible functionality**: Optional close button with proper event handling
- **Full ARIA support**: Configurable `role`, `aria-live`, and `aria-label` attributes
- **Gradient variant**: Replicates the existing cyan/purple gradient shadow with hover animation
- **TypeScript support**: Full type safety with props interface

### Visual Parity
The `gradient` variant matches the existing Ember What's New banner:
- Same box-shadow gradient effect (cyan/purple)
- Same hover animation (translateY + scale)
- Same border radius and padding
- Same transition timing

## Screenshots / Testing Instructions

To review the component in Storybook:

```bash
yarn workspace @tryghost/shade storybook
```

Navigate to **Components / Banner** to see all variants and interactive examples.

### DIFF
--- a/apps/shade/src/components/ui/banner.tsx
+++ b/apps/shade/src/components/ui/banner.tsx
@@ -11,7 +11,7 @@ const bannerVariants = cva(
             variant: {
                 default: 'border border-border bg-background shadow-sm hover:shadow-md',
                 gradient: [
-                    'cursor-pointer border border-grey-100 bg-white',
+                    'cursor-pointer border border-gray-100 bg-white',
                     'shadow-[rgb(75_225_226_/_28%)_-7px_-6px_42px_8px,rgb(202_103_255_/_32%)_7px_6px_42px_8px]',
                     'hover:shadow-[rgb(75_225_226_/_38%)_-7px_-4px_42px_10px,rgb(202_103_255_/_42%)_7px_8px_42px_10px]',
                     'hover:translate-y-[-2px] hover:scale-[1.01]'
@@ -23,7 +23,7 @@ const bannerVariants = cva(
             },
             size: {
                 sm: 'p-2 text-sm',
-                md: 'p-3 pb-4',
+                md: 'p-3',
                 lg: 'p-4'
             }
         },
@@ -85,7 +85,7 @@ const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
                 {dismissible && (
                     <Button
                         aria-label="Dismiss notification"
-                        className="absolute right-2 top-1 size-8"
+                        className="absolute right-1 top-1 size-8 text-gray-600"
                         size="icon"
                         variant="ghost"
                         onClick={handleDismiss}
--- a/apps/shade/src/components/ui/banner.tsx
+++ b/apps/shade/src/components/ui/banner.tsx
@@ -1,0 +1,104 @@
+import * as React from 'react';
+import {cva, type VariantProps} from 'class-variance-authority';
+import {X} from 'lucide-react';
+import {cn} from '@/lib/utils';
+import {Button} from './button';
+
+const bannerVariants = cva(
+    'relative block rounded-lg transition-all duration-300',
+    {
+        variants: {
+            variant: {
+                default: 'border border-border bg-background shadow-sm hover:shadow-md',
+                gradient: [
+                    'cursor-pointer border border-grey-100 bg-white',
+                    'shadow-[rgb(75_225_226_/_28%)_-7px_-6px_42px_8px,rgb(202_103_255_/_32%)_7px_6px_42px_8px]',
+                    'hover:shadow-[rgb(75_225_226_/_38%)_-7px_-4px_42px_10px,rgb(202_103_255_/_42%)_7px_8px_42px_10px]',
+                    'hover:translate-y-[-2px] hover:scale-[1.01]'
+                ],
+                info: 'bg-blue-50 border-blue-200 dark:bg-blue-950/30 dark:border-blue-800 border',
+                success: 'bg-green-50 border-green-200 dark:bg-green-950/30 dark:border-green-800 border',
+                warning: 'bg-yellow-50 border-yellow-200 dark:bg-yellow-950/30 dark:border-yellow-800 border',
+                destructive: 'bg-red-50 border-red-200 dark:bg-red-950/30 dark:border-red-800 border'
+            },
+            size: {
+                sm: 'p-2 text-sm',
+                md: 'p-3 pb-4',
+                lg: 'p-4'
+            }
+        },
+        defaultVariants: {
+            variant: 'default',
+            size: 'md'
+        }
+    }
+);
+
+interface BannerBaseProps
+    extends Omit<React.HTMLAttributes<HTMLDivElement>, 'role'>,
+    VariantProps<typeof bannerVariants> {
+    /**
+     * Semantic role of the banner
+     * @default 'status'
+     *
+     * Note: role="status" has implicit aria-live="polite"
+     *       role="alert" has implicit aria-live="assertive"
+     *       role="region" requires aria-label
+     */
+    role?: 'status' | 'alert' | 'region';
+}
+
+export type BannerProps = BannerBaseProps & (
+    | { dismissible: true; onDismiss: () => void; }
+    | { dismissible?: false; onDismiss?: never; }
+);
+
+const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
+    (props, ref) => {
+        // onDismiss must be destructured to prevent React warning about invalid DOM props
+        const {
+            variant,
+            size,
+            dismissible = false,
+            onDismiss,
+            role = 'status',
+            className,
+            children,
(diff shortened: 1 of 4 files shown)


apps/shade/test/unit/components/ui/banner.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+describe('Banner Component', () => {
+    it('renders children correctly', () => {
+        render(<Banner>Test content</Banner>);
+        const content = screen.getByText('Test content');
+
+        assert.ok(content, 'Banner should render children');
+    });
+
+    it('applies default variant and size classes', () => {
```
rob-ghost, line 20: I tired to parameterise all these tests with `it.each` but the lint rule `ghost/mocha/no-setup-in-describe` didn't allow it, so I had to have individual tests.

apps/shade/src/components/ui/banner.tsx
```diff
@@ -0,0 +1,138 @@
...
+            children,
+            ...props
+        },
+        ref
+    ) => {
+        const [dismissed, setDismissed] = React.useState(false);
+
+        if (dismissed) {
+            return null;
+        }
```
jonatansberg: I'm not sure how I feel about this being stateful. Typically, if you dismiss something, there's already some state outside of this component that determined it would render in the first place. Most likely that state needs to be updated to reflect the change as well. I'd rather have that be the source of truth than maintain an internal state here as well. 

We could use a discriminated union for the types so that you have to provide an onDismiss handler if you've enabled the dismissible feature.

apps/shade/src/components/ui/banner.tsx
```diff
@@ -0,0 +1,138 @@
...
+);
+
+export interface BannerProps
+    extends React.HTMLAttributes<HTMLDivElement>,
+    VariantProps<typeof bannerVariants> {
+    /**
+     * Visual style variant
+     * @default 'default'
+     */
+    variant?: 'default' | 'gradient' | 'info' | 'success' | 'warning' | 'destructive';
```
jonatansberg: Can't this be a keyof rather than redefining the variants?

apps/shade/src/components/ui/banner.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Size of the banner
+     * @default 'md'
+     */
+    size?: 'sm' | 'md' | 'lg';
```
jonatansberg: Same here

jonatansberg (commented):
I'm not sure how I feel about the statefulness of this component. I'd rather have it stateless. That would also remove the need for the tests. 

As an aside, we haven't been maintaining them for new components we're adding since they have a low signal-to-noise ratio. Instead, we've been talking about enabling some automated testing around the Storybook stories, but that hasn't been implemented yet. 

I'm certain Zimo will have some input on the styling, but we can deal with that once he's back.

apps/shade/src/components/ui/banner.tsx
rob-ghost: Yep I agree 💯, this is better off stateless and have the container react to `onDismiss`.

apps/shade/test/unit/components/ui/banner.test.tsx
kevinansfield: I've run into that rule a few times in the past too and ended up doing a next-line disable because it didn't make sense to lint against building something to loop over.

Looking in Ghost core, we even have it disabled globally so we've likely been perpetuating "invalid" uses throughout the codebase 🙈
https://github.com/TryGhost/Ghost/blob/main/ghost/core/test/.eslintrc.js#L42-L43

Worth asking in #engineering-forum to get consensus on the rule's usefulness?

kevinansfield: https://github.com/lo1tuma/eslint-plugin-mocha/blob/main/docs/rules/no-setup-in-describe.md

> If you're using [dynamically generated tests](https://mochajs.org/#dynamically-generating-tests), you should disable this rule.

I think that adds some weight to at least disabling the rule when you know you're not breaking the intention of the rule.

rob-ghost: Nice one thanks for pointing this out! I've re-parameterised the tests and disabled that rule just for this file and I think we can talk about updating this configuration in the forum.

apps/shade/src/components/ui/banner.tsx
rob-ghost: Great suggestion with the "dismissible" discriminated union! I've made the component stateless now.

apps/shade/src/components/ui/banner.tsx
rob-ghost: Turns out it wasn't needed anyway! It was already included in `VariantProps<typeof bannerVariants>`

apps/shade/src/components/ui/banner.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+type NonDismissibleBannerProps = BannerBaseProps & {
+    /** @default false */
+    dismissible?: false;
+    onDismiss?: never;
+};
+
+/**
+ * Discriminated union: when dismissible is true, onDismiss is required
+ */
```
jonatansberg: This comment seems redundant. 

You could perhaps also collapse the four different prop type definitions into something like this if you remove the duplicated aria props: 
```ts
type BannerBaseProps =React.HTMLAttributes<HTMLDivElement> &
  VariantProps<typeof bannerVariants> & (
    | {
        dismissible: true;
        onDismiss: () => void;
      }
    | {
        dismissible?: false;
      }
  );
```

apps/shade/src/components/ui/banner.tsx
```diff
@@ -0,0 +1,117 @@
...
+interface BannerBaseProps
+    extends Omit<React.HTMLAttributes<HTMLDivElement>, 'role'>,
+    VariantProps<typeof bannerVariants> {
+    /** @default 'status' */
+    bannerRole?: 'status' | 'alert' | 'region';
+
+    /** @default 'polite' */
+    ariaLive?: 'polite' | 'assertive' | 'off';
+
+    ariaLabel?: string;
```
jonatansberg: Does it make sense to explicitly enumerate these separately when they will be included in the valid HTML attributes already?

apps/shade/src/components/ui/banner.tsx
```diff
@@ -0,0 +1,117 @@
...
+            size: 'md'
+        }
+    }
+);
+
+interface BannerBaseProps
+    extends Omit<React.HTMLAttributes<HTMLDivElement>, 'role'>,
+    VariantProps<typeof bannerVariants> {
+    /** @default 'status' */
+    bannerRole?: 'status' | 'alert' | 'region';
```
jonatansberg: Doesn't the existing role attribute suffice? It would already be clear that the role targets the banner: `<Banner role="..."` vs `<Banner bannerRole="..."`

jonatansberg (commented):
Much better! I added a few nitpicks that might be worth looking at before merging this. 

Also, since Zimo should be back today, maybe we can allow some time for him to take a look as well.

apps/shade/src/components/ui/banner.tsx
rob-ghost: There was a bit of confusion around the base types and the allowable values, but yes you're right, this is shadowing attributes which already exist with identical types and we're not doing anything meaningful with their defaults or values so I've removed them.

apps/shade/src/components/ui/banner.tsx
rob-ghost: Yes it does, I've changed this to `role` but kept the constrained values and the default.

apps/shade/src/components/ui/banner.tsx
rob-ghost: Yep, thanks to removing the `aria*` props I can collapse the definition to this.

peterzimon (approved):
Hey @rob-ghost — I refined a few minor visual details. Otherwise all looks great to me.

jonatansberg (approved):

PR #25236, opened 2025-10-23
Added indexes to members created events and subscription created events

cmraible:
ref https://linear.app/ghost/issue/NY-712/improve-performance-of-verificationtrigger-queries

The `VerificationTrigger` service runs a few queries on the `members_created_events` and `members_subscription_created_events` tables for each new member added via the Admin API, which are particularly slow (upwards of 40 seconds in some cases). The queries are currently using temporary tables and two full table scans.

This commit adds appropriate indexes to these tables to improve the performance of these queries, and remove the temporary tables + full table scans.


cmraible:
Tested migrations on scale-testing.ghost.is, which has 2 million `members_created_events`, and the migration took 21 seconds to complete. Notably this site does not have _any_ `members_subscription_created_events`, so the total time on a site with a healthy number of each type of event may take longer than the 21 seconds.

cmraible:
Closing this PR, at least for now. We've found a way to avoid these queries entirely on large sites, so adding the indexes would just add overhead for little to no benefit.

PR #25238, opened 2025-10-24
Fixed Tinybird port conflict in e2e tests

cmraible:
The Tinybird service in e2e/compose.yml uses the same hostport as the Tinybird service in compose.yml, which leads to a port conflict when running both.

This changes the host port for Tinybird in the e2e tests to 7182 to avoid the conflict.


cmraible:
Closing as we should handle this with the E2E test setup with dev:forward

PR #25241, opened 2025-10-24
Fixed Analytics > Newsletters percent change calculation

cmraible:
ref https://linear.app/ghost/issue/NY-703/analytics-newsletters-total-subscribers-percent-change-is-incorrect

There are currently two bugs in the percent change calculation on the Analytics > Newsletters view:
- The frontend calculates the percent change using the most recent data point and the second most recent data point, where it should use the most recent data point and the oldest data point in the range
- The backend doesn't return all dates in the range, which leads to the starting value being calculated incorrectly

This fixes both of those bugs to ensure an accurate percent change calculation, and adds an e2e test to cover the percent change calculation.


cmraible:
This fix works from my manual testing, just thinking about how we should test this.

We could conceivably add an E2E test (or add this as part of an E2E test), but currently we don't have a way to add _historical_ members.

e2e/helpers/services/members/MembersImportService.ts
```diff
@@ -0,0 +1,277 @@
...
+        this.adminEndpoint = '/ghost/api/admin';
+    }
+
+    /**
+     * Import members from an array of member objects
+     * @param members Array of member data to import
+     * @param options Import configuration options
+     * @returns Import response with stats and import label
+     */
+    async importMembers(
```
cmraible: We can make this more generic

e2e/helpers/pages/admin/analytics/analytics-newsletters-page.ts
```diff
@@ -1,13 +1,25 @@
 import {AdminPage} from '../AdminPage';
 import {Locator, Page} from '@playwright/test';
 
+class TotalSubscribersTab {
```
9larsons, line 4: This is an interesting subdivision pattern - @ibalosh What's your thoughts on this? I think I'd need to see more of it used and am in favor at first glance.

ghost/core/core/server/services/stats/PostsStatsService.js
```diff
@@ -1039,6 +1053,61 @@ class PostsStatsService {
...
+     * @private
+     * @param {Array<{date: string, value: number}>} values - Sparse array of values with dates
+     * @param {string|null} startDate - Start date in YYYY-MM-DD format (ISO)
+     * @param {string|null} endDate - End date in YYYY-MM-DD format (ISO)
+     * @param {string} timezone - Timezone for date interpretation
+     * @param {number} startingValue - The value to use before the first event (default: 0)
+     * @returns {Array<{date: string, value: number}>} Dense array with all dates filled
+     */
+    _fillMissingDates(values, startDate, endDate, timezone = 'UTC', startingValue = 0) {
+        const moment = require('moment-timezone');
```
9larsons, line 1067: thought: at some point we should investigate alternatives to moment, given we've had concerns about the size/weight of it.

ghost/core/core/server/services/stats/PostsStatsService.js
```diff
@@ -1039,6 +1053,61 @@ class PostsStatsService {
...
+
+        // If no date range provided, return as-is
+        if (!startDate || !endDate) {
+            return values || [];
+        }
+
+        // Determine the date range
+        const rangeStart = moment.tz(startDate, timezone).startOf('day');
```
9larsons, line 1075: nit: there's opportunities to reuse `getDateBoundaries` from `date-utils` (and perhaps to add this there?) but as I've mentioned a few times now, I'd like to redo the whole charting endpoints so I don't feel strongly about putting a lot of energy into cleaning this up now. Arguably this could be the start of it... I think I'd like to spend more time thinking it through.

9larsons (approved):
Some minor comments. Overall, LGTM! This patches a flaky hole that's a bit difficult to see, since we're used to testing with very active sites or with test data that tends to fill completely.

Mostly, this restarts my mental discussion about re-tooling the charting endpoints, so I'm going to take that as a follow up to write up.

W/r to the import, we discussed w/o Igor and I think this is both a reasonably tidy and appropriate solution. I don't think we need a full on seeding structure right now - this gets the job done for the moment, and as mentioned, we don't have a way to *import* much besides members right now. We'd have to take a different approach that'll warrant more discussion.

e2e/helpers/pages/admin/analytics/analytics-newsletters-page.ts
ibalosh: I like it too, I experimented with something similar on members, I called it sections or something. We did not yet polish up the way we want to use these, but as you said Steve, we need to see more of them to decide some rulings around them.

With page objects, common pattern is to divide chunks/section in page to smaller separate subdivisions like here.

These are super valuable to me, they reduce the cognitive load when looking through locators list and actions with them 👍

e2e/helpers/services/members-import/members-import-service.ts
```diff
@@ -0,0 +1,260 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import {HttpClient as APIRequest} from '../../../data-factory/persistence/adapters/http-client';
+import {assert} from 'console';
+
+export interface MemberImportData {
```
ibalosh, line 4: `nit` I am wondering if we could keep types like Members in single place in codebase. I added recently a factory which I guess is using the same API shape, but for different thing (single or multiple posts).

Just a thought, but maybe we could keep these in factories? Maybe factories are not the best final place to put them, but would love if we could keep API shapes in single place, and we use these the most in factories.

e2e/helpers/services/members-import/MembersImportService.ts
```diff
@@ -0,0 +1,260 @@
...
+        members: MemberImportData[],
+        options: ImportMembersOptions = {}
+    ): Promise<MemberImportResponse> {
+        const {
+            pollingTimeout = 30000,
+            pollingInterval = 500,
+            labels = []
+        } = options;
+        const initialCount = await this.getMemberCount();
+        const csvPath = await this.generateCSVFile(members);
```
ibalosh: Any reason for storing CSV as a file first? From what I can see, we don't have a benefit in terms of memory or speed by doing so, but it complicates things little bit more.

We could skip the file entirely and convert the CSV string directly to a Buffer? It would simplify process a bit, reduce coupling between methods, and speed things up a bit.

```typescript
  async import(
      members: MemberImportData[],
      options: ImportMembersOptions = {}
  ): Promise<MemberImportResponse> {
      const initialCount = await this.getMemberCount();
      const csvContent = this.generateCSV(members);  // Just get the string
      const csvBuffer = Buffer.from(csvContent, 'utf-8');  // Convert to buffer

      const {response: importResponse, statusCode} = await this.uploadCSV(
          csvBuffer,
          'members-import.csv',  // Just pass filename
          labels
      );
      // No cleanup needed!
  }

  private async uploadCSV(
      csvBuffer: Buffer,
      filename: string,
      labels: string[] = []
  ): Promise<{response: MemberImportResponse; statusCode: number}> {
      const response = await this.request.post(`${this.adminEndpoint}/members/upload/`, {
          multipart: {
              membersfile: {
                  name: filename,
                  mimeType: 'text/csv',
                  buffer: csvBuffer  // Already have the buffer
              },
              // ...
          }
      });
  }
```

e2e/tests/admin/analytics/newsletters.test.ts
```diff
@@ -1,6 +1,13 @@
 import {AnalyticsNewslettersPage} from '../../../helpers/pages/admin';
+import {MembersImportService} from '../../../helpers/services/members-import';
 import {expect, test} from '../../../helpers/playwright';
 
+function subtractDaysFromCurrentDate(days: number): Date {
+    const date = new Date();
+    date.setDate(date.getDate() - days);
+    return date;
```
ibalosh: `nit` maybe `getDateDaysAgo(days: number): Date` ,  just thinking out loud :) current name is good too

e2e/tests/admin/analytics/newsletters.test.ts
```diff
@@ -24,4 +31,32 @@ test.describe('Ghost Admin - Newsletters', () => {
...
+            {
+                email: 'yesterday@example.com',
+                name: 'Yesterday',
+                created_at: subtractDaysFromCurrentDate(1).toISOString()
+            }
+        ];
+        await membersService.import(members);
+
+        await page.reload();
+        await expect(newslettersPage.newslettersCard).toBeVisible();
```
ibalosh, line 58: love how clean this is 🙌

e2e/tests/admin/analytics/newsletters.test.ts
```diff
@@ -24,4 +31,32 @@ test.describe('Ghost Admin - Newsletters', () => {
...
+            },
+            {
+                email: 'ten-days-ago@example.com',
+                name: 'Ten Days Ago',
+                created_at: subtractDaysFromCurrentDate(10).toISOString()
+            },
+            {
+                email: 'yesterday@example.com',
+                name: 'Yesterday',
+                created_at: subtractDaysFromCurrentDate(1).toISOString()
```
ibalosh: since you switch .toISOString() , we could do that in subtractDaysFromCurrentDate , and if you reuse the function, maybe pass as option formatting to Iso string as choice?

ibalosh (approved):
Hey @cmraible 

overall I think this looks good, none of my comments are show stoppers. Maybe the biggest one is related to using CSV file in members import service, if we load data on the fly, I think it would simplify that service and make that code bit cleaner.

e2e/helpers/pages/admin/analytics/analytics-newsletters-page.ts
cmraible: Just sharing a few other examples we have currently:
- AnalyticsGrowthPage: https://github.com/TryGhost/Ghost/blob/a02eb7dce8f1bb36dec7f6836bdf3b25eb9ceca2/e2e/helpers/pages/admin/analytics/AnalyticsGrowthPage.ts
- AnalyticsOverviewPage: https://github.com/TryGhost/Ghost/blob/a02eb7dce8f1bb36dec7f6836bdf3b25eb9ceca2/e2e/helpers/pages/admin/analytics/AnalyticsOverviewPage.ts

> These are super valuable to me, they reduce the cognitive load when looking through locators list and actions with them 👍

Exactly, without the sub-pages you usually end up with a bunch of locators that share the same (long) prefix

e2e/helpers/services/members-import/members-import-service.ts
cmraible: Good shout! This one is a bit unique since it is specifically the type for a row of the CSV import, which doesn't map 1:1 with the types used in the rest of the API. For example, the `subscribed_to_emails` field only exists in the CSV, but not in the members database table or other API types.

I think in my ideal world, this type would live in Ghost itself (or in a shared types package), and we'd just import it here.

For now since this is very closely coupled to the CSV import, I think it makes sense to keep it close to where it's used.

e2e/helpers/services/members-import/MembersImportService.ts
cmraible: Great idea, implemented!

e2e/tests/admin/analytics/newsletters.test.ts
cmraible: Updated :)

e2e/tests/admin/analytics/newsletters.test.ts
cmraible: Good call, updated :)

ghost/core/core/server/services/stats/PostsStatsService.js
cmraible: Yeah definitely agree, it's a good instinct. I looked at reusing getDateBoundaries for this, but it returns an ISO string instead of the moment object, so it started to get a bit messy. Good opportunity to refactor, but going to leave as-is for this PR.

PR #25268, opened 2025-10-27
Cleaned up unused dashboard code

kevinansfield:
no issue

In Ghost 6.0 the old Ember-powered dashboard was replaced with a React version but the dead code wasn't cleaned up.

- updated `dashboard` route to redirect to the `stats-x` route without creating an unnecessary history entry
- removed dashboard template and all of the unused components
- removed dashboard-mocks service as that was only possible to enable via the old dashboard
  - removed usage of dashboard-mocks service from the dashboard-stats service


cmraible (approved):
Thanks @kevinansfield, LGTM! We should have done this a while ago.

PR #25269, opened 2025-10-27
Cleaned up `audienceFeedback` labs flag

kevinansfield:
no issue

- the `audienceFeedback` labs flag was made GA/always-true 3 years ago but the flag conditionals and related old code were never cleaned up
- removed redundant conditional code
- removed unused "old-analytics" components in Admin


kevinansfield:
Closing. Updated PR opening shortly

PR #25270, opened 2025-10-27
🐛 Fixed outdated dashboard links in design and profile modals

troyciesco:
closes https://linear.app/ghost/issue/NY-690

- We retired the dashboard route in favor of analytics, but it was missed in a few places
- This PR updates it in the design modal that shows up during site setup, as well as for users with permissions below admin
- Now when those modals are closed they'll be redirected to /analytics instead of /dashboard (in the case of low-permission users, it actually falls back to something other than /analytics. For contributors, for example, it falls back to /posts. But that's elsewhere in the code)
- Not user visible since /dashboard was redirecting to /analytics anyway, but more correct
- in DesignModal, also switches from `window.location.hash` to `updateRoute`
- Considered creating a constant for routes, but that seems like a larger lift than this ticket was meant for

---

Location of fix in DesignModal:
redir

Location of fix in UserDetailModal (click "your profile" then close):
Screenshot 2025-10-27 at 8 47 52 AM

### DIFF
--- a/apps/admin-x-settings/src/components/settings/general/UserDetailModal.tsx
+++ b/apps/admin-x-settings/src/components/settings/general/UserDetailModal.tsx
@@ -210,7 +210,7 @@ const UserDetailModalContent: React.FC<{user: User}> = ({user}) => {
         if (canAccessSettings(currentUser)) {
             updateRoute('staff');
         } else {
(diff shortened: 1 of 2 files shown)


9larsons (approved):
That's all we need! LGTM

PR #25273, opened 2025-10-27
Remove activitypub router basepath

jonatansberg:

Moving the base path to a wrapping route and updating all links and urls within the app to use either relative paths or a base-path aware navigation hook makes links and URLs work the same regardless of if the app is mounted by itself within Ember or as a sub-routes in the new React admin shell.

This also makes it easier for us to eventually update the base path from /activitypub to /network.

### DIFF
--- a/apps/activitypub/src/routes.tsx
+++ b/apps/activitypub/src/routes.tsx
@@ -12,9 +12,9 @@ import OnboardingStep2 from '@components/layout/Onboarding/Step2';
 import OnboardingStep3 from '@components/layout/Onboarding/Step3';
 import Preferences from '@views/Preferences';
 import Profile from '@views/Profile';
-import {Navigate, RouteObject} from '@tryghost/admin-x-framework';
+import {Navigate, Outlet, RouteObject} from '@tryghost/admin-x-framework';
 
-export const APP_ROUTE_PREFIX = '/activitypub';
+const basePath = import.meta.env.VITE_TEST ? '' : 'activitypub';
 
 export type CustomRouteObject = RouteObject & {
     pageTitle?: string;
@@ -24,9 +24,11 @@ export type CustomRouteObject = RouteObject & {
 
 export const routes: CustomRouteObject[] = [
     {
-        // Root route configuration
-        path: '',
+        // Root route that defines the app's base path
+        path: basePath,
+        element: <Outlet />,
         errorElement: <Error />, // This will catch all errors in child routes
+        handle: 'activitypub-basepath',
         children: [
             {
                 index: true,
--- a/apps/activitypub/src/App.tsx
+++ b/apps/activitypub/src/App.tsx
@@ -1,7 +1,7 @@
-import {APP_ROUTE_PREFIX, routes} from '@src/routes';
 import {FeatureFlagsProvider} from './lib/feature-flags';
 import {FrameworkProvider, Outlet, RouterProvider, TopLevelFrameworkProps} from '@tryghost/admin-x-framework';
 import {ShadeApp} from '@tryghost/shade';
+import {routes} from '@src/routes';
 
 interface AppProps {
     framework: TopLevelFrameworkProps;
@@ -13,14 +13,9 @@ const App: React.FC<AppProps> = ({framework, activityPubEnabled}) => {
         return null;
     }
 
-    // In test mode, we use hash routing to avoid basename issues
-    const routerProps = import.meta.env.VITE_TEST
-        ? {prefix: '', hash: true, routes}
-        : {prefix: APP_ROUTE_PREFIX, routes};
-
     return (
         <FrameworkProvider {...framework}>
-            <RouterProvider {...routerProps}>
+            <RouterProvider prefix={'/'} routes={routes}>
                 <FeatureFlagsProvider>
                     <ShadeApp className="shade-activitypub" darkMode={false} fetchKoenigLexical={null}>
                         <Outlet />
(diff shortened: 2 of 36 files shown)


apps/activitypub/src/routes.tsx
```diff
@@ -12,20 +12,26 @@ import OnboardingStep2 from '@components/layout/Onboarding/Step2';
...
     pageTitle?: string;
     children?: CustomRouteObject[];
     showBackButton?: boolean;
 };
 
+const ActivityPubToNetworkRedirect = () => {
+    const location = useLocation();
+    const newPath = location.pathname.replace(/^\/activitypub/, '/network');
+    return <Navigate to={`${newPath}${location.search}${location.hash}`} replace />;
+};
```
jonatansberg: I'm not a fan of having this here, but I don't know where else to put it within the directory structure of the app.

sagzy (commented):
@jonatansberg one thing to keep in mind: the `activitypub` package is released independently from Ghost Admin:

- When a change to `apps/activitypub:*` (with a minor/patch version bump in `apps/activitypub/package.json`) is merged to main, a new version of the package is published to NPM and is immediately served to all users in production via jsDelivr ([code](https://github.com/TryGhost/Ghost/blob/8b15610b5e04bda93691a0ad502e4bb2ec359c3d/.github/workflows/ci.yml#L1057)) - both pro and self-hosters
- Ghost Admin is (generally) released with a new Ghost version, and this is rolled out every week for Pro users whereas self-hosters choose when to upgrade

From our conversation this morning, I was imagining the routing change to be outside of `apps/activitypub` (in the React shell app and Ember), so that we don't have to coordinate the release between `apps/activitypub` and Ghost Admin. Is that not an option?

jonatansberg:
No, this change needs to happen within the package. I think it should work the same though? From the point of view of Ember, nothing is really changing in terms of whether the base path is added by the router or if we are adding it to all of the links ourselves. 

We might have to do the change from /activitypub to /network in separate steps tho, so that we can ensure the Ember app will render the ActivityPub app on both routes first 🤔

sagzy:
@jonatansberg yeah keeping the renaming to `/network` for a follow-up PR sounds sensible!  
  
With the renaming, we're making changes to both:

* `apps/activitypub`
* and `ghost/admin/app/router.js`

As the release / rollout lifecycles of the `activitypub` package is independent from Ghost Admin (Ember), we need to make sure the following combination doesn't break things:

> new version of `activitypub` but current version of Ghost Admin

If we take out the renaming piece from that PR, the change becomes easier to ship - so let's do that?

jonatansberg:
@sagzy I think so. I started making the necessary changes, but then I've been in meetings, huddles, mobs and pairing sessions all day. Hopefully I will get back to this tomorrow!

jonatansberg:
@sagzy I ended up taking a slightly different approach where we're inferring the base path from a wrapping route. This lets  us move the app within the component tree rather freely. 

If we want to mount the app under /network in the react admin, all that is required would be something like this:
```diff
diff --git a/apps/admin/src/routes.tsx b/apps/admin/src/routes.tsx
index 4feb871947..20a1724c5e 100644
--- a/apps/admin/src/routes.tsx
+++ b/apps/admin/src/routes.tsx
@@ -6,6 +6,9 @@ import { routes as activityPubRoutes } from "@tryghost/activitypub/src/routes";
 import { routes as statsRoutes } from "@tryghost/stats/src/routes";
 import { EmberFallback } from "./ember-bridge";
 
+// Override the activitypub route to redirect to /network
+activityPubRoutes[0].path = '/network';
+
 export const routes: RouteObject[] = [
     {
         path: "/",
@@ -17,8 +20,11 @@ export const routes: RouteObject[] = [
         children: statsRoutes
     },
     {
-        path: `network`,
-        loader: () => redirect('/activitypub')
+        path: `activitypub/*`,
+        loader: ({ request }) => {
+            const path = request.url.split('/activitypub').slice(1).join('/activitypub');
+            return redirect(`/network${path}`);
+        }
     },
     {
         path: '',
```

Take a look and let me know what you think!

apps/activitypub/src/routes.tsx
```diff
@@ -12,9 +12,9 @@ import OnboardingStep2 from '@components/layout/Onboarding/Step2';
 import OnboardingStep3 from '@components/layout/Onboarding/Step3';
 import Preferences from '@views/Preferences';
 import Profile from '@views/Profile';
-import {Navigate, RouteObject} from '@tryghost/admin-x-framework';
+import {Navigate, Outlet, RouteObject} from '@tryghost/admin-x-framework';
 
-export const APP_ROUTE_PREFIX = '/activitypub';
+const basePath = import.meta.env.VITE_TEST ? '' : 'activitypub';
```
sagzy, line 17: question/double-checking: is the removal of the leading `/` from `activitypub` in `basePath` intentional (compared to the old `APP_ROUTE_PREFIX`)

sagzy (commented):
@jonatansberg new approach looks good!

- Left one question more as a double-check
- I'm seeing one more instance of the old `useNavigate()` (without basePath) in Error.tsx:

https://github.com/TryGhost/Ghost/blob/b1b649981fc22416efb997b5398be7de7dcb77b6/apps/activitypub/src/components/layout/Error/Error.tsx#L9

jonatansberg:
That last `useNavigate` was intentionally left as is. The Error page navigate call goes "outside of the app" so using `useNavigateWithBasePath` would resolve the wrong URL.

apps/activitypub/src/routes.tsx
jonatansberg: Yes, but it shouldn't matter much. Convention with React Router is to use nesting to describe the path rather than have full absolute paths. Having activitypub without the leading the slash at the root level still makes it a root route. 

It's all normalized by React Router in the end regardless of if there is a leading slash or not.

sagzy (approved):

jonatansberg:
I cleaned up some minor formatting changes that was introduced in with one of the previous attempts. Will merge once CI passes.

PR #25274, opened 2025-10-27
Remove router base path for stats app

jonatansberg:

Removing the base path and updating all links and urls within the app to include the base path directly makes links and URLs work the same regardless of if the app is mounted in Ember or as a sub-routes in the new React admin shell.


cmraible (approved):
Nice one, LGTM!

PR #25275, opened 2025-10-27
Implement externalNavigate

jonatansberg:

This is needed so that links across what used to be different micro frontends work within the single router instance in the React admin shell.

Longer term, we probably want to get rid of this altogether, but for now this enables cross-step navigation.

### DIFF
--- a/apps/admin/src/utils/navigation.test.ts
+++ b/apps/admin/src/utils/navigation.test.ts
@@ -1,0 +1,79 @@
+import { describe, expect, it, beforeEach, afterEach, vi } from "vitest";
+import { navigateTo } from "./navigation";
+
+describe("navigateTo", () => {
+    let locationMock: { href: string; hash: string };
+
+    beforeEach(() => {
+        locationMock = { href: "", hash: "" };
+        vi.stubGlobal("location", locationMock as unknown as Location);
+    });
+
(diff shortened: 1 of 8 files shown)


apps/admin/src/utils/navigation.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    beforeEach(() => {
+        // Mock window.location by creating a writable property
+        Object.defineProperty(window, 'location', {
+            value: {
+                href: '',
+                hash: ''
+            },
+            writable: true,
+            configurable: true
+        });
```
kevinansfield: Should this be restored when tests finish?

jonatansberg: I think vitest gives you a clean DOM for each run? Now, I'm not sure though.

kevinansfield: I think you get a new DOM instance per file but tests within a file all share the same instance. In this case its probably fine but not having mock cleanup did jump out as a non-standard practice

jonatansberg: Yeah, that might be how it is. Anyways, I cleaned it up.

kevinansfield (approved):

PR #25277, opened 2025-10-27
Removed unnecessary events queries for sites that are already verified

cmraible:
ref https://linear.app/ghost/issue/NY-712/improve-performance-of-verificationtrigger-queries

For sites that are already verified, the `VerificationTrigger` is still running the `getSignupEvents` queries for every `MemberCreatedEvent`. These queries are unnecessary and expensive for larger sites, so this change removes the event handler entirely for sites that are already verified.


cmraible (commented):
There's some additional refactoring we could do here especially for the tests, but I've tried to keep this changeset minimal since it's partly in response to an incident.

sam-lord (approved):

PR #25279, opened 2025-10-28
Fixed Shade button regressions

peterzimon:
ref https://linear.app/ghost/issue/BER-2933/fix-button-regressions

- During to React port work we introduced a few regressions to buttons: `icon` buttons and icons sizing became off and borders got lost on outline buttons.

### DIFF
--- a/apps/admin/src/layout/app-sidebar/AppSidebar.tsx
+++ b/apps/admin/src/layout/app-sidebar/AppSidebar.tsx
@@ -40,7 +40,7 @@ function AppSidebar({ ...props }: React.ComponentProps<typeof Sidebar>) {
                     <Button
                         variant='ghost'
                         size='icon'
-                        className="size-9 text-gray-800 rounded-full hover:bg-gray-200 -mr-1"
+                        className="size-9 text-gray-800 rounded-full hover:bg-gray-200 -mr-1 [&_svg]:size-auto"
                         title="Search site (Ctrl/⌘ + K)"
                     >
                         <LucideIcon.Search size={20} />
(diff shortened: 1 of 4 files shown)


apps/admin/src/layout/app-sidebar/AppSidebar.tsx
```diff
@@ -62,7 +62,7 @@ function AppSidebar({ ...props }: React.ComponentProps<typeof Sidebar>) {
                         <Button
                             variant="ghost"
                             size="icon"
-                            className="size-9 hover:bg-gray-200 rounded-full p-0 text-gray-800"
+                            className="size-9 hover:bg-gray-200 rounded-full p-0 text-gray-800 [&_svg]:size-auto"
```
jonatansberg, line 65: Can't you set this directly on the icon rather than on the button?

apps/shade/preflight.css
```diff
@@ -211,6 +211,10 @@
         margin: 0; /* 2 */
         padding: 0; /* 3 */
         outline: none;
+
+        /* Reset `border: none;` from global.css */
+        border-width: 0;
+        border-style: solid;
```
jonatansberg, line 217: The one place where this might break would be somehow changing the appearance of borders within Ember when running inside the new admin shell. 

I did a quick run-through and couldn't see if they're happening anywhere tho. And if we would run into that, regressions in the React shell version are nearly as big of a problem as regressions in the Ember version.

jonatansberg (approved):
I think this looks good. I tested it locally (except activitypub, I don't have that running atm) and it seems to resolve the regressions.

apps/admin/src/layout/app-sidebar/AppSidebar.tsx
peterzimon: Not really because of specificity rules, unless I use !important on the icon but I'd like to avoid that if possible.

PR #25280, opened 2025-10-28
🐛 Scope PRESERVE_ENV to global teardown to prevent leaks

ibalosh:
- Removed `PRESERVE_ENV` check from teardownGhostInstance()
- `PRESERVE_ENV` now only affects globalTeardown() to leave containers after all tests complete
- Prevents port conflicts when running tests multiple times with PRESERVE_ENV=true
- Containers are now always cleaned up between individual tests

Previously, `PRESERVE_ENV=true` would skip removing containers after each individual test, causing port allocation errors when running tests again. Now it only preserves the final state after all tests complete, which is the intended debugging behavior.

### DIFF
--- a/e2e/helpers/environment/EnvironmentManager.ts
+++ b/e2e/helpers/environment/EnvironmentManager.ts
@@ -177,9 +177,6 @@ export class EnvironmentManager {
      * Teardown a Ghost instance
      */
     public async teardownGhostInstance(ghostInstance: GhostInstance): Promise<void> {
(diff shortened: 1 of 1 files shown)


ibalosh:
@cmraible I fixed the `PRESERVE_ENV` issue, I will merge this one into main, so I can use it in other branch, and avoid others getting into the port issue.

Let me know if you think it needs more work, and the fix is not adequate

ibalosh:
it will need more work, it preserves for now only last Ghost environment

PR #25283, opened 2025-10-28
Updated the split helper to not produce empty strings

cathysarisky:
Splitting an element with a trailing separator results in a "" element being included in the returned array.  This is typical JavaScript behavior, which I'm preserving here (although I don't love it).  This PR adds some additional tests that show this behavior.

Will also add to docs.


ErisDS (approved):
I think it would be reasonable to change this default behaviour for this helper. It is standard JS behaviour, the helper isn't doing the _wrong_ thing, but it's not doing what you need or expect.

Working around it with a match helper is great, but every extra helper you have to use adds overhead in terms of processing but more importantly maintenance of theme code.

I'd happily approve a PR that changed this behaviour, especially this early in the helper's life.

The way I'd think about this would be changing the behaviour "trim" empty strings from the end of the array by default. If we end up crossing the bridge that the default behaviour is desirable for some users they could add `trim=false` as a property to the helper to disable the trim by default behaviour, but I wouldn't add it unless it was asked for by a theme dev.

cathysarisky:
Hi @ErisDS , thanks for the review! 

I went ahead and implemented the altered behavior for split to eliminate empty strings anywhere they occur.  Given that we're generally using split elements that'll be looped over to produce text, empty items that result in nothing being rendered but that change the length of the array are a potential source of confusion.

I added more tests, including the case where the string is nothing but separators.  

Marking this for re-review since it's in no way what you approved previously, but leaving it on this branch/PR so that we don't lose the context.  Thank you!

ErisDS (approved):

ghost/core/core/frontend/helpers/split.js
```diff
@@ -27,8 +27,15 @@ module.exports = function split(...args) {
...
+    // Filter out all empty strings
+    const result = string.split(separator)
+        .filter(item => item !== '')
+        .map(item => new SafeString(item));
+
+    if (result.length === 0) {
+        return renderResult([], options, data);
+    }
+
     return renderResult(result, options, data);
```
vikaspotluri123, line 40:
```suggestion
return renderResult(result, options, data);
```

I'm not seeing a difference in behavior between the empty and filled cases. Am I missing something?

cathysarisky: hmm. I think you're right, since result would be [] after .map if everything is filtered out.

PR #25288, opened 2025-10-28
Added write to outbox on member creation, processing of outbox with log message

troyciesco:
closes https://linear.app/ghost/issue/NY-709

- Adds the outbox model
- When a member is created, a record is also added to the outbox table with an `event_type` of `MemberCreatedEvent`
- A job runs every five minutes to see if there are any `MemberCreatedEvent` records in the table that are under the retry limit and aren't failed
- if there are, it logs a message (in lieu of actually sending an email) and if it succeeds, deletes the row from the outbox
- Processes up to 1,000 rows every time the job runs. Fetches 100, process each row, then fetches another 100, and so on

---

> [!NOTE]
> On member creation, enqueue a `MemberCreatedEvent` to `outbox` and process it every 5 minutes to log a mock welcome email and remove the entry.
> 
> - **Members / Data**:
>   - **Outbox model**: Add `models/outbox` with statuses (`pending`, `processing`, `failed`) and defaults; wire `Outbox` into members API.
>   - **Member creation**: When `labs:welcomeEmails` is enabled and source is `member`, write a transactional `MemberCreatedEvent` outbox entry with `memberId`, `email`, `name`, `source`, `timestamp`.
> - **Background Jobs**:
>   - **Service**: New `server/services/member-welcome-emails` initialized in `core/boot.js`.
>   - **Recurring job**: Schedule every 5 minutes (non-test) to run `jobs/process-outbox.js`.
>   - **Processing**: Fetch pending `MemberCreatedEvent` entries, mark `processing`, send mock welcome email (log), delete on success; update failed entries with retry/count (`MAX_RETRIES=0`). Batches of 100 up to 1000 per run with performance logs.
> - **Tests**:
>   - Integration tests for job processing and service behavior; unit tests for `outbox` model and `MemberRepository` outbox integration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9a858d9703f9a987e8fd3a62a1576862d4bfc81a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -261,6 +263,15 @@ module.exports = class MemberRepository {
             options = {};
         }
 
+        if (!options.transacting) {
```
troyciesco: this feels odd, need to double-check this pattern

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -402,6 +413,27 @@ module.exports = class MemberRepository {
                 }
             }
         }
+
+        // Insert into outbox if welcome emails feature is enabled
+        if (this._labsService.isSet('welcomeEmails')) {
+            await this._Outbox.add({
+                id: ObjectId().toHexString(),
+                event_type: 'MemberCreatedEvent',
+                payload: JSON.stringify({
```
troyciesco: can likely parse this payload down even further

ghost/core/core/server/services/welcome-emails/jobs/process-outbox.js
```diff
@@ -0,0 +1,89 @@
...
+
+(async () => {
+    const startTime = Date.now();
+    const db = require('../../../data/db');
+
+    if (process.env.NODE_ENV !== 'production') {
+        logging.warn(`[WELCOME-EMAIL] Welcome email processor running in simulation mode (${SIMULATE_FAILURE_RATE * 100}% failure rate)`);
+    }
+
+    const entries = await db.knex('outbox')
```
troyciesco: one concern (from ai chats) with removing the concept of `processed_at` - is there ever a scenario where we might have multiple workers, and thus end up accidentally duplicating emails? Would having `processed_at` even help with that?

I think our general approach with welcome emails, from an event-driven perspective, would be "at-least-once", but whatever we can do to prevent duplicates is useful.

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -402,6 +413,27 @@ module.exports = class MemberRepository {
                 }
             }
         }
+
+        // Insert into outbox if welcome emails feature is enabled
+        if (this._labsService.isSet('welcomeEmails')) {
+            await this._Outbox.add({
+                id: ObjectId().toHexString(),
+                event_type: 'MemberCreatedEvent',
```
troyciesco: need to double-check this event, can it come from that tryghost lib instead of a magic string?

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
troyciesco: i see we use it elsewhere in this file 👍

ghost/core/core/server/services/welcome-emails/jobs/process-outbox.js
troyciesco: i added a `status` column to address this, plus it feels slightly better to have a specific `failed` status instead of relying on the retry_count, which could technically get changed.

ghost/core/core/server/services/member-welcome-emails/jobs/index.js
```diff
@@ -0,0 +1,22 @@
+const path = require('path');
+const jobsService = require('../../jobs');
+
+let hasScheduled = {
+    processOutbox: false
+};
+
+module.exports = {
+    async scheduleWelcomeEmailJob() {
+        if (!hasScheduled.processOutbox && !process.env.NODE_ENV.startsWith('test')) {
```
troyciesco, line 15: need to see if i can make it so this only runs if the labs flag is set 🤔

ghost/core/core/server/services/welcome-emails/jobs/index.js
```diff
@@ -0,0 +1,22 @@
...
+    processOutbox: false
+};
+
+module.exports = {
+    async scheduleWelcomeEmailJob() {
+        if (!hasScheduled.processOutbox && !process.env.NODE_ENV.startsWith('test')) {
+            jobsService.addJob({
+                at: '0 */5 * * * *',
+                job: path.resolve(__dirname, 'process-outbox.js'),
+                name: 'welcome-emails-process-outbox'
```
troyciesco: i wanna come up with a better name for this, it's weirdly long and verbose in the logs

cmraible:
@troyciesco did you just manually create the outbox table for testing this?

troyciesco:
> @troyciesco did you just manually create the outbox table for testing this?

@cmraible nope, this PR is based off [the one to create the migration](https://github.com/TryGhost/Ghost/pull/25286) (so the diff for this piece is clearer), so I ran the migration first.

cmraible:
> @cmraible nope, this PR is based off [the one to create the migration](https://github.com/TryGhost/Ghost/pull/25286) (so the diff for this piece is clearer), so I ran the migration first.

Oh I see, my bad!

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -24,6 +24,14 @@ const messages = {
 
 const SUBSCRIPTION_STATUS_TRIALING = 'trialing';
 
+// Sources allowed to trigger welcome emails via outbox
+// 'member' = self-signups through public forms (Portal, Signup Form, etc.)
+// 'api' in development = allows testing with scripts like gm.js
+const WELCOME_EMAIL_SOURCES = ['member', 'api'];
+// if (process.env.NODE_ENV !== 'production') {
```
troyciesco: will clean up before fully ready for review (i.e. before merge)

ghost/core/core/server/services/welcome-emails/jobs/process-outbox.js
```diff
@@ -0,0 +1,260 @@
...
+ * @returns {Promise<Array>} Array of outbox entries marked as SUBMITTING
+ */
+async function fetchPendingEntries(db, batchSize) {
+    let entries = [];
+    await db.knex.transaction(async (trx) => {
+        const query = trx('outbox')
+            .where('event_type', MemberCreatedEvent.name)
+            .where('status', OUTBOX_STATUSES.PENDING);
+        
+        if (EXPONENTIAL_BACKOFF.enabled) {
```
troyciesco: this was something i added just to see if it's useful or not but haven't done too much with it - may be worth saving for later, if at all.

ghost/core/core/server/services/member-welcome-emails/jobs/process-outbox.js
```diff
@@ -0,0 +1,260 @@
...
+                    .orWhereRaw('last_retry_at < DATE_SUB(NOW(), INTERVAL ? SECOND)', [
+                        Math.floor(EXPONENTIAL_BACKOFF.baseDelayMs / 1000)
+                    ])
+                    .orWhereRaw('last_retry_at < DATE_SUB(NOW(), INTERVAL (? * POW(2, retry_count)) SECOND)', [
+                        Math.floor(EXPONENTIAL_BACKOFF.baseDelayMs / 1000)
+                    ]);
+            });
+        }
+        
+        entries = await query
```
troyciesco, line 26: kind of related to the above, it's worth noting that this doesn't differentiate between attempts that didn't succeed before (like `retry_count` > 0, but status is "pending" bc it hasn't hit max retries yet) and ones that have never been tried (`retry_count` = 0)

since these are 10 batches of 100 each time the job runs, i don't think it'll cause a backup or anything, though the three tries would come during the same job run (like batch 1, 2, 3) when it might make more sense to retry on future runs, which come every five minutes

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
cmraible: note (non-blocking): we should think about how we can unify this pattern. It feels wrong to create the event in the outbox, then emit the `MemberCreatedEvent`, which then subsequently inserts the `member_created_event`. I don't think we need to solve this right now, but something we should keep in mind. 

Just thinking out loud: the outbox should be the definitive source of truth, since we're doing all this work to make sure we can't ever drop one. So in a future state, I think we ultimately will want to listen to the outbox itself, rather than the domain event, for e.g. inserting the `member_created_event`.

I'm also kind of curious why we don't already insert the `member_created_event` within the same transaction...theoretically we could be creating members but missing the `member_created_event`, with the way this is currently working.

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
cmraible: Agree that it feels a bit odd, but it does seem to be the prevailing pattern for ensuring we're in a transaction 🤷

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -402,6 +422,24 @@ module.exports = class MemberRepository {
...
+                id: ObjectId().toHexString(),
+                event_type: MemberCreatedEvent.name,
+                payload: JSON.stringify({
+                    memberId: member.id,
+                    email: member.get('email'),
+                    name: member.get('name'),
+                    source,
+                    timestamp: eventData.created_at
+                }),
+                created_at: new Date()
```
cmraible: suggestion: I think the `created_at` is redundant here, Bookshelf should handle the `created_at` timestamp itself if I'm not mistaken

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -402,6 +422,24 @@ module.exports = class MemberRepository {
...
+        if (this._labsService.isSet('welcomeEmails') && WELCOME_EMAIL_SOURCES.includes(source)) {
+            await this._Outbox.add({
+                id: ObjectId().toHexString(),
+                event_type: MemberCreatedEvent.name,
+                payload: JSON.stringify({
+                    memberId: member.id,
+                    email: member.get('email'),
+                    name: member.get('name'),
+                    source,
+                    timestamp: eventData.created_at
```
cmraible: note (non-blocking): since the payload field is free-form, we'll need to be careful about any changes we make to its shape over time. In a future where we're emitting different kinds of events, we may want a way to clearly define the schema of the payload for each event.

ghost/core/core/server/services/welcome-emails/jobs/process-outbox.js
cmraible: I'd lean toward keeping it simple for now while we're just proving out that this works conceptually, but we'll definitely want something like this eventually.

ghost/core/core/server/services/welcome-emails/jobs/process-outbox.js
```diff
@@ -0,0 +1,260 @@
...
+function shouldSimulateFailure() {
+    return process.env.NODE_ENV !== 'production' && Math.random() < SIMULATE_FAILURE_RATE;
+}
+
+/**
+ * Sends a welcome email to a new member (currently logs only)
+ * @param {Object} payload - Member data containing name and email
+ */
+async function sendWelcomeEmail(payload) {
+    logging.info(`[WELCOME-EMAIL] Welcome email sent to ${payload.name} at ${payload.email}`);
```
cmraible: nit: the member's name isn't a required field, and some sites don't even display the input, in which case `${payload.name}` would be empty here

ghost/core/core/server/services/welcome-emails/jobs/process-outbox.js
```diff
@@ -0,0 +1,260 @@
...
+        payload = JSON.parse(entry.payload);
+
+        if (shouldSimulateFailure()) {
+            throw new errors.InternalServerError({
+                message: MESSAGES.SIMULATED_FAILURE
+            });
+        }
+
+        await sendWelcomeEmail(payload);
+        await deleteProcessedEntry(db, entry.id);
```
cmraible: note: probably not relevant just yet, but any failures here would probably happen _within_ the `sendWelcomeEmail` function, not before it. I think we'll need to put some more thought into how we handle errors and ensure we only delete the entry if we can guarantee it's been handled successfully.

We also need to be careful with action idempotency, i.e. what if we send the email, but then fail to delete the entry for some reason, causing us to reprocess the same event twice? How do we avoid sending a duplicate welcome email?

ghost/core/core/server/models/outbox.js
```diff
@@ -0,0 +1,23 @@
+const ghostBookshelf = require('./base');
+
+const OUTBOX_STATUSES = {
+    PENDING: 'pending',
+    SUBMITTING: 'submitting',
```
troyciesco: I was trying to stay consistent with wording based on other places in the codebase that use `submitting` but having iterated on this more I think `processing` might make more sense.

Will update before merge

troyciesco: for context it was the `emails` and `email_batches` tables. It happens to be our first outbox use case involves email but that's not always true

ghost/core/core/server/services/welcome-emails/jobs/process-outbox.js
```diff
@@ -0,0 +1,260 @@
...
+ * @param {Object} db - Database connection
+ * @param {Object} entry - Outbox entry to process
+ * @returns {Promise<Object>} Result object with success boolean
+ */
+async function processEntry(db, entry) {
+    let payload;
+    try {
+        payload = JSON.parse(entry.payload);
+
+        if (shouldSimulateFailure()) {
```
troyciesco: this will come out shortly - though i wonder if it should be replaced with a check that the entry has a status of `processing`

ghost/core/core/server/services/member-welcome-emails/jobs/process-outbox.js
troyciesco: L22-L23 above addresses this now, but mainly for the sake of discussion. i'm still on the fence. because now failures will take 3 job runs to fully move to failed, so it means they'll block less entries up front...but i wonder if that makes it harder to recover if there's a outage of some sort (i.e. having to process a ton with 1-2 failures before getting back to normal - but maybe that's the best recovery strategy anyway)

ghost/core/core/server/services/member-welcome-emails/jobs/lib/process-entries.js
```diff
@@ -0,0 +1,87 @@
...
+const {MAX_RETRIES, MEMBER_WELCOME_EMAIL_LOG_KEY} = require('./constants');
+const {OUTBOX_STATUSES} = require('../../../../models/outbox');
+const sendMemberWelcomeEmail = require('./send-member-welcome-email');
+
+/**
+ * Deletes a successfully processed outbox entry
+ * @param {Object} db - Database connection
+ * @param {string} entryId - ID of the entry to delete
+ */
+async function deleteProcessedEntry(db, entryId) {
```
9larsons, line 11: I think it's fine for now, but we'd want to conceptually tie these fns to the outbox, not the welcome emails. Each service shouldn't need to implement these, or refer to other services.

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -295,9 +309,11 @@ module.exports = class MemberRepository {
             throw new errors.BadRequestError({message: tpl(messages.moreThanOneProduct)});
         }
 
+        const transactionalOptions = options.transacting ? {transacting: options.transacting} : undefined;
```
troyciesco: I need to look more closely from L312-418, as well as L888. running `yarn test:e2e` in `ghost/core` (and in CI) was having an issue with too many connections that was stemming from adding the transaction above. I think these changes might fix it but need to understand better because it touches a lot of other code to try and get everything in one txn

9larsons (approved):

cmraible (approved):
Reviewed in call, LGTM!

ghost/core/core/server/services/member-welcome-emails/jobs/process-outbox.js
troyciesco: we set MAX_RETRIES for 0 to now to get a sense of how many items even fail, and this won't come up until we increase that, so it can stay as it is for now

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
troyciesco: this was scrapped in favor of just wrapping `_Member.add` and `_Outbox.add` in a transaction, instead of trying to couple everything into the same transaction, which was causing issues with creating the correct stripe data.

ghost/core/core/server/services/member-welcome-emails/jobs/process-outbox.js
```diff
@@ -0,0 +1,131 @@
...
+ * @param {number} batchSize - Maximum number of entries to fetch
+ * @param {string} jobStartISO - ISO-formatted timestamp when the job started (YYYY-MM-DD HH:mm:ss)
+ * @returns {Promise<Array>} Array of outbox entries marked as PROCESSING
+ */
+async function fetchPendingEntries(db, batchSize, jobStartISO) {
+    let entries = [];
+    await db.knex.transaction(async (trx) => {
+        const query = trx('outbox')
+            .where('event_type', MemberCreatedEvent.name)
+            .where('status', OUTBOX_STATUSES.PENDING)
```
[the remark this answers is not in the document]
troyciesco: we know this - it'll be addressed (as will what to do with failed jobs) in a follow up when we add real email sending instead of just logs

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -402,6 +440,7 @@ module.exports = class MemberRepository {
                 }
             }
         }
+
```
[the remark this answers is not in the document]
troyciesco: I'm pretty sure we'll never get orphaned events since the member/outbox creation will have to happen first. 

Technically an outbox row could get processed before those events are created, true, but 1) those two things aren't currently related (though maybe a refactor will make them so later  2) the event emission in this file already wasn't atomic with the creation of a member, so this change isn't introducing that.

ghost/core/core/server/services/member-welcome-emails/jobs/lib/process-entries.js
```diff
@@ -0,0 +1,87 @@
...
+        return {success: true};
+    } catch (err) {
+        await updateFailedEntry({db, entryId: entry.id, retryCount: entry.retry_count, errorMessage: err.message});
+
+        const memberInfo = payload ? `${payload.name} (${payload.email})` : 'unknown member';
+        logging.error(`${MEMBER_WELCOME_EMAIL_LOG_KEY} Failed to send to ${memberInfo}: ${err.message}`);
+
+        return {success: false};
+    }
+}
```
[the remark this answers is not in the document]
troyciesco: if it fails after the email's sent, it won't be reprocessed, because it'll get stuck in the "processing" state. Fixing that stuck processing state is a known piece (mentioned in another comment) we'll address shortly in a different PR

ghost/core/core/server/services/member-welcome-emails/jobs/lib/process-entries.js
```diff
@@ -0,0 +1,91 @@
...
+    } catch (err) {
+        const errorMessage = err?.message ?? 'Unknown error';
+        await updateFailedEntry({db, entryId: entry.id, retryCount: entry.retry_count, errorMessage});
+
+        const email = payload?.email || 'unknown member';
+        const memberInfo = payload?.name ? `${payload.name} (${email})` : email;
+        logging.error(`${MEMBER_WELCOME_EMAIL_LOG_KEY} Failed to send to ${memberInfo}: ${errorMessage}`);
+
+        return {success: false};
+    }
```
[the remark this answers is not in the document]
troyciesco: i think this is a valid concern - will address in https://github.com/TryGhost/Ghost/pull/25342 when sending real emails

PR #25289, opened 2025-10-29
Improved notification grouping with short time window in ActivityPub

minimaluminium:
ref https://linear.app/ghost/issue/BER-2937/improve-notification-grouping-by-adding-date-criteria

-   Grouped same-type notifications only when they occur within a short timeframe
-   Prevented unrelated notifications from being merged hours apart
-   Made grouped notifications feel timely and relevant

### DIFF
--- a/apps/activitypub/src/views/Notifications/Notifications.tsx
+++ b/apps/activitypub/src/views/Notifications/Notifications.tsx
@@ -74,7 +74,7 @@ function groupNotifications(notifications: Notification[], useTimeGrouping: bool
             break;
         case 'follow':
             // Group follows by time window
-            groupKey = `follow_${notification.type}${timeBucket}`;
+            groupKey = `follow_${timeBucket}`;
             break;
         case 'mention':
             // Don't group mentions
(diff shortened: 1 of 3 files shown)


apps/activitypub/src/views/Notifications/Notifications.tsx
```diff
@@ -54,13 +68,13 @@ function groupNotifications(notifications: Notification[]): NotificationGroup[]
...
-                groupKey = `repost_${notification.post.id}`;
+                // Group reposts by the target object and time window
+                groupKey = `repost_${notification.post.id}${timeBucket}`;
             }
             break;
         case 'follow':
-            // Group follows that are next to each other in the array
-            groupKey = `follow_${notification.type}`;
+            // Group follows by time window
+            groupKey = `follow_${notification.type}${timeBucket}`;
```
sagzy: nitpick: `notification.type` here is `"follow"`, which gives us a redundant `follow` in the key:
- before: `follow_follow`
- now: `follow_follow_<bucket>`

Doesn't harm, but looks like a typo!

sagzy (approved):
LGTM! Left one nitpick and there is a linting error to fix

apps/activitypub/src/views/Notifications/Notifications.tsx
minimaluminium: Fixed in https://github.com/TryGhost/Ghost/pull/25289/commits/5b4db79c473bb03c554c013de8cff8bc513bf201

PR #25293, opened 2025-10-29
Updated logging for email analytics job(s)

9larsons:
ref https://linear.app/ghost/issue/NY-721/
- updated `emailAnalytics` config and defaults to have more tunability around alerts/logs
- added new metric for tracking email analytics throughput (behind config)
- added new console warning for delays in the open job (behind config)
- updated logging to be more concise and specific

This is laying the groundwork for better observability in our logs for monitoring the email analytics job. The new metric allows us to monitor the events/s throughput for sites in question - namely, for very large sites where they need to ingest 10ks of events, possibly within a few minutes. Previously we couldn't do this without a lot of manual math and searching in Elastic.

NOTE: The lagging warning for email opens is a flawed measure. We currently only update the job timestamp _if we process events_ so if there was a legitimate period where there were no Mailgun events, this would be a false positive. As such, this is going to issue a warning for sites that have 1) low activity or 2) experience any kind of outage in the site/Mailgun. As such, it may be worth removing and skipping this, though we can try it out.

### DIFF
--- a/ghost/core/core/server/services/email-analytics/EmailAnalyticsService.js
+++ b/ghost/core/core/server/services/email-analytics/EmailAnalyticsService.js
@@ -313,8 +313,6 @@ module.exports = class EmailAnalyticsService {
      */
     async #fetchEvents(fetchData, {begin, end, maxEvents = Infinity, eventTypes = null}) {
         // Start where we left of, or the last stored event in the database, or start 30 minutes ago if we have nothing available
-        logging.info('[EmailAnalytics] Fetching from ' + begin.toISOString() + ' until ' + end.toISOString() + ' (maxEvents: ' + maxEvents + ')');
-
         // Store that we started fetching
         fetchData.running = true;
         fetchData.lastStarted = new Date();
@@ -378,8 +376,6 @@ module.exports = class EmailAnalyticsService {
                 await provider.fetchLatest(processBatch, {begin, end, maxEvents, events: eventTypes});
                 apiPollingTime += (Date.now() - apiStart);
             }
-
-            logging.info('[EmailAnalytics] Fetching finished');
         } catch (err) {
             if (err.message !== 'Fetching canceled') {
                 logging.error('[EmailAnalytics] Error while fetching');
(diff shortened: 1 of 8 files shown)


ghost/core/core/server/services/email-analytics/jobs/index.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     async scheduleRecurringJobs(skipEmailCheck = false) {
         if (
             !hasScheduled &&
-            config.get('emailAnalytics') &&
+            config.get('emailAnalytics:enabled') &&
```
9larsons, line 13: This is significant, however we do not use this in Zuul currently, so we can follow up with a PR to make changes there.

ghost/core/core/server/services/email-analytics/EmailAnalyticsService.js
```diff
@@ -24,9 +24,32 @@ const errors = require('@tryghost/errors');
  * @typedef {'delivered' | 'opened' | 'failed' | 'unsubscribed' | 'complained'} EmailAnalyticsEvent
  */
 
+/**
+ * @typedef {object} EmailAnalyticsFetchResult
+ * @property {number} eventCount - The number of events fetched
+ * @property {number} apiPollingTime - Time spent polling the API in milliseconds
+ * @property {number} processingTime - Time spent processing events in milliseconds
+ * @property {number} aggregationTime - Time spent aggregating stats in milliseconds
```
troyciesco: suggestion: naming these properties with `Ms` at the end might be useful i think

ghost/core/core/server/services/email-analytics/EmailAnalyticsService.js
```diff
@@ -285,18 +309,21 @@ module.exports = class EmailAnalyticsService {
...
         // Store that we started fetching
         fetchData.running = true;
         fetchData.lastStarted = new Date();
         fetchData.lastBegin = begin;
         this.queries.setJobTimestamp(fetchData.jobName, 'started', begin);
 
+        // Timing metrics
+        let apiPollingTime = 0;
+        let processingTime = 0;
+        let aggregationTime = 0;
```
troyciesco: same here re: naming with the time unit

ghost/core/core/server/services/email-analytics/EmailAnalyticsServiceWrapper.js
```diff
@@ -62,51 +63,100 @@ class EmailAnalyticsServiceWrapper {
         });
     }
 
+    /**
+     * Log comprehensive job completion with timing metrics
+     * @param {string} jobType - Type of job (e.g., 'latest-opened', 'latest', 'missing', 'scheduled')
+     * @param {object} fetchResult - The fetch result from EmailAnalyticsService
+     * @param {number} totalDuration - Total duration in milliseconds
```
troyciesco: same here re: `Ms` at the end, especially since the eventual log messages are displayed in seconds instead

troyciesco (approved):
some suggestions on naming that i think will benefit us in the long run - looks good otherwise

PR #25297, opened 2025-10-30
Added topics to Discover mode in Reader

minimaluminium:
ref https://linear.app/ghost/issue/BER-2943/topical-feed

-    Updated Discover to show "Top" plus selectable topic feeds
-    Kept global feed as "Top" and added topic tabs for exploration
-    Updated endpoints from `/feed/global` to `/feed/discover/top` and `/feed/discover/[topic]`

### DIFF
--- a/apps/activitypub/src/api/activitypub.ts
+++ b/apps/activitypub/src/api/activitypub.ts
@@ -450,7 +450,7 @@ export class ActivityPubAPI {
         return this.getPaginatedPosts('.ghost/activitypub/v1/feed/reader', next);
     }
 
-    async getGlobalFeed(next?: string, topic?: string): Promise<PaginatedPostsResponse> {
+    async getDiscoveryFeed(next?: string, topic?: string): Promise<PaginatedPostsResponse> {
         const topicPath = topic && topic !== 'all' ? topic : 'top';
         const endpoint = `.ghost/activitypub/v1/feed/discover/${topicPath}`;
         return this.getPaginatedPosts(endpoint, next);
--- a/apps/activitypub/src/api/activitypub.ts
+++ b/apps/activitypub/src/api/activitypub.ts
@@ -450,9 +450,8 @@ export class ActivityPubAPI {
         return this.getPaginatedPosts('.ghost/activitypub/v1/feed/reader', next);
     }
 
-    async getDiscoveryFeed(next?: string, topic?: string): Promise<PaginatedPostsResponse> {
-        const topicPath = topic && topic !== 'all' ? topic : 'top';
-        const endpoint = `.ghost/activitypub/v1/feed/discover/${topicPath}`;
+    async getDiscoveryFeed(next?: string, topic: string = 'top'): Promise<PaginatedPostsResponse> {
+        const endpoint = `.ghost/activitypub/v1/feed/discover/${topic}`;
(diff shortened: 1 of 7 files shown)


apps/activitypub/src/api/activitypub.ts
```diff
@@ -450,8 +450,10 @@ export class ActivityPubAPI {
         return this.getPaginatedPosts('.ghost/activitypub/v1/feed/reader', next);
     }
 
-    async getGlobalFeed(next?: string): Promise<PaginatedPostsResponse> {
-        return this.getPaginatedPosts('.ghost/activitypub/v1/feed/global', next);
+    async getGlobalFeed(next?: string, topic?: string): Promise<PaginatedPostsResponse> {
```
sagzy: nitpick: should we rename to `getDiscoveryFeed` (instead of "global")?

apps/activitypub/src/api/activitypub.ts
```diff
@@ -450,8 +450,10 @@ export class ActivityPubAPI {
...
+        const topicPath = topic && topic !== 'all' ? topic : 'top';
```
sagzy: question: what's the purpose of that line of code? Wouldn't the `TOPICS` array in TopicFilter.tsx already pass the correct value?

apps/activitypub/src/components/TopicFilter.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import {Button} from '@tryghost/shade';
+
+export type Topic = 'all' | 'news' | 'technology' | 'business' | 'culture' | 'politics' | 'finance' | 'design';
+
+const TOPICS: {value: Topic; label: string}[] = [
+    {value: 'all', label: 'Top'},
```
sagzy: Think we should settle on one value for the "Top" topic - I think "top" is more representative than "all" in terms what the topic contains, so let's drop "all" and use "top"?

apps/activitypub/src/hooks/use-activity-pub-queries.ts
```diff
@@ -1724,8 +1724,8 @@ export function useInboxForUser(options: {enabled: boolean}) {
     return {inboxQuery, updateInboxActivity};
 }
 
-export function useGlobalFeedForUser(options: {enabled: boolean}) {
-    const queryKey = QUERY_KEYS.globalFeed;
+export function useGlobalFeedForUser(options: {enabled: boolean; topic?: string}) {
+    const queryKey = [...QUERY_KEYS.globalFeed, options.topic || 'all'];
```
sagzy: question: is the fallback to 'all' (or after renaming: 'top') necessary here?

sagzy (approved):

apps/activitypub/src/hooks/use-activity-pub-queries.ts
minimaluminium: You're right! As the topic is always passed, we can just make it non-optional and lose the fallback.

apps/activitypub/src/api/activitypub.ts
minimaluminium: Yeah, this doesn't make sense. As we made it consistent by using `top` across everything, we wouldn't need this at all.

PR #25299, opened 2025-10-30
Improved React sidebar design

peterzimon:
ref https://linear.app/ghost/issue/BER-2913/port-sidebar

- Some of the functions in the current (Ember) sidebar had some discoverability problems (e.g. Help being too hidden in the user menu), while others were way too in your face (e.g. dark mode switch).
- The user menu was lacking design details such as menu item icons and proper sizing/spacing.
- Header and footer components were built inside the main component which made code readability harder. Now they have their own components.
- `kbd` component for keyboard shortcuts were missing from Shade

### DIFF
--- a/apps/admin/src/layout/app-sidebar/NavHeader.tsx
+++ b/apps/admin/src/layout/app-sidebar/NavHeader.tsx
@@ -30,7 +30,7 @@ function NavHeader({ ...props }: React.ComponentProps<typeof SidebarHeader>) {
                         </div>
                     </div>
                 </div>
-                <Button variant='outline' className="flex items-center justify-between text-gray-500 hover:text-gray-700 hover:bg-background text-base [&_svg]:stroke-2 pr-2 shadow-xs hover:shadow-sm hover:border-gray-200 h-[38px]">
+                <Button variant="outline" className="flex items-center justify-between text-gray-500 hover:text-gray-700 hover:bg-background text-base [&_svg]:stroke-2 pr-2 shadow-xs hover:shadow-sm hover:border-gray-200 h-[38px]">
                     <div className="flex items-center gap-2">
                         <LucideIcon.Search className="text-muted-foreground" />
                         Search site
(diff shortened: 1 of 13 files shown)


apps/admin/src/layout/app-sidebar/NavHeader.tsx
```diff
@@ -0,0 +1,45 @@
...
+                                alt="Site icon"
+                                className="w-full h-full rounded-md object-cover"
+                                />
+                        </div>
+                        <div className="font-semibold text-[15px] text-black overflow-hidden text-ellipsis whitespace-nowrap">
+                            {title}
+                        </div>
+                    </div>
+                </div>
+                <Button variant='outline' className="flex items-center justify-between text-gray-500 hover:text-gray-700 hover:bg-background hover:border-gray-300 text-base [&_svg]:stroke-2">
```
jonatansberg: Nit: Mixing ' and ". We should just have an auto formatter fore these things...

apps/admin/src/layout/app-sidebar/NavHeader.tsx
```diff
@@ -0,0 +1,45 @@
...
+                    <div className="flex items-center gap-2">
+                        <LucideIcon.Search />
+                        Search
+                    </div>
+                    <Kbd className="text-muted-foreground">⌘K</Kbd>
```
jonatansberg: We would need to be a bit smarter than hardcoding ⌘K here, right? I'm guessing it would be something else on Windows and Linux.

jonatansberg (approved):
Noted a couple of nitpicks. Nothing that prevents us merging this. You can address those in follow-up PRs. 

You probably need to fix the linting error in Shade though that CI is upset about.

apps/admin/src/layout/app-sidebar/NavHeader.tsx
peterzimon: Yes, it's on my list

apps/admin/src/layout/app-sidebar/NavHeader.tsx
peterzimon: Good point, I fixed this one but yeah we should have an auto formatter

PR #25300, opened 2025-10-30
Fixed Ember modals not being visible on React routes

kevinansfield:
no issue

- added `#ember-modal-wormhole` element outside of the `#ember-app` element so it doesn't get hidden when on React routes
- updated Ember application template to render modals into the new `#ember-modal-wormhole` when it exists


jonatansberg (approved):
Nice!

peterzimon (approved):

PR #25301, opened 2025-10-30
Enabled custom domain support for local admin development

jonatansberg:
Fixes https://linear.app/ghost/issue/BER-2944/support-local-development-with-a-custom-domain

This adds a Vite plugin that dynamically resolves the Ghost site URL from the backend API, enabling local development to work when Ghost is configured with a custom domain. Previously, proxy was hard coded to localhost:2368.

### DIFF
--- a/apps/admin/tsconfig.node.json
+++ b/apps/admin/tsconfig.node.json
@@ -21,5 +21,5 @@
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true
   },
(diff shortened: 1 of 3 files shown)


kevinansfield (approved):
Working well when `url` is set to a domain 🙌

It's still failing when `admin.url` is different to the main `url` value. In that case the http://localhost:2368/ghost/api/admin/site/ request gets a 301 redirect to the configured admin domain. That's an edge case though and I don't think we should block on it.

Looks like we'll need something to fix the CI tests? Otherwise ✅

PR #25303, opened 2025-10-30
WIP: Spike on domain warming

sam-lord:
ref [GVA-584](https://linear.app/ghost/issue/GVA-584/spike-implement-domain-warming-using-email-count)

What I've got so far:
* Ghost knows how to send from the fallback domain
* When creating a batch, we can tell it to use the fallback domain or not
* Using the fallback domain changes the behaviour of the `SenderAddressService`, so it uses the fallback address as the from address, and the preferred `reply-to` (or preferred `from` address as a fallback) as the reply-to address
* We also needed to change which domain we send from, which is constructed in the message data sent to the Mailgun, overriding the main domain specified in config

What we don't have:
* Ghost doesn't yet know when to use the fallback domain

---

> [!NOTE]
> Adds domain warmup: limits custom-domain sends per email, splits batches, routes overflow via fallback domain, updates schema, and wires through provider/domain logic.
> 
> - **Email sending (backend)**:
>   - Add `DomainWarmingService` to compute per-send warmup limits from prior `csd_email_count`.
>   - Enhance `BatchSendingService` to split batches by warmup capacity and mark batches with `fallback_domain`; pass `useFallbackDomain` to `SendingService`.
>   - Update `EmailService` to set `csd_email_count` on creation (behind `labs:domainWarmup`) and inject `DomainWarmingService`.
>   - Extend `EmailRenderer`/`EmailAddressService` to support fallback from-address and expose `fallbackDomain`.
>   - Update `SendingService` to select fallback from-address/domain and forward options to the provider.
>   - Update `MailgunClient` to override send domain and fetch events from both primary and fallback domains when `domainWarmup` is enabled.
> - **Data schema/migrations**:
>   - Add `emails.csd_email_count` and `email_batches.fallback_domain` (+ migrations) and update schema hash.
> - **Labs/UI**:
>   - Add `domainWarmup` feature flag and toggle in `PrivateFeatures`.
> - **Tests**:
>   - Expand unit tests to cover warmup limits, batch splitting, fallback routing, and new provider/domain behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8b20867604cce457b916fe6cddcde4571fcdd663. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ghost/core/core/server/services/email-service/DomainWarmingService.ts
sam-lord, line 1: Ignore this file, I pasted in the EmailAddressService as boilerplate to start hacking on, and to have something to require from the BatchSendingService

ghost/core/core/server/services/email-service/DomainWarmingService.js
```diff
@@ -0,0 +1,45 @@
...
+
+        if (lastCount <= 100) {
+            return 200;
+        } else if (lastCount <= 100_000) {
+            return lastCount * 2;
+        } else if (lastCount <= 400_000) {
+            return lastCount * 1.5;
+        } else {
+            return lastCount * 1.25;
+        }
```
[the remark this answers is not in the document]
sam-lord: We do respect the flag, but at the point of creating the email model.

If we did it here, then we'd have to return null or Infinity, both have issues.

ghost/core/core/server/services/email-service/BatchSendingService.js
```diff
@@ -16,6 +16,7 @@ const MAX_SENDING_CONCURRENCY = 2;
  * @typedef {import('./EmailSegmenter')} EmailSegmenter
  * @typedef {import('./EmailRenderer')} EmailRenderer
  * @typedef {import('./EmailRenderer').MemberLike} MemberLike
+ * @typedef {import('./DomainWarmingService')} DomainWarmingService
```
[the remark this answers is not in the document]
sam-lord: Fixed the first issue.

The second one, I actually prefer the spread operator without the extra cruft?? I did have to double check it worked, but knowing it does, I prefer the lack of extra syntax over `{transacting, ...(options || {})}`

ghost/core/core/server/data/migrations/versions/6.6/2025-10-27-16-12-46-add-warming-window.js
```diff
@@ -0,0 +1,6 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('emails', 'csd_email_count', {
```
sam-lord, line 3: If someone has been sending from a CSD for a while then we don't want to put them into a warmup state when we enable the feature.

If the status for the CSD is 'VERIFIED' + the last_updated value was _before_ the last email sent, and the csd_email_count for the last email is null, then we don't need to do domain warmup. Their domain is implicitly warmed since it was activated before the feature was enabled.

We should probably fill in the csd_email_count with the full recipient count for subsequent emails after the feature is enabled --> this means that if someone's email list grows 3x (as an example) overnight, then the warmup process _will_ start for that list.

ghost/core/core/server/data/migrations/versions/6.6/2025-10-30-11-14-39-add-email-batch-fallback-address.js
```diff
@@ -0,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('email_batches', 'fallback_domain', {
```
sam-lord, line 3: Consider `fallback_sending_domain` as a name for clarity?

ghost/core/core/shared/labs.js
```diff
@@ -50,7 +50,7 @@ const PRIVATE_FEATURES = [
     'tagsX',
     'utmTracking',
     'emailUniqueid',
-    'welcomeEmails'
+    'domainWarmup'
```
sam-lord, line 53: Maybe `domainWarming`? First thing to ship here!?

ghost/core/core/server/services/email-service/DomainWarmingService.js
```diff
@@ -0,0 +1,42 @@
...
+
+    async getWarmupLimit() {
+        const lastCount = await this.getLastCount();
+
+        if (lastCount <= 100) {
+            return 200;
+        } else if (lastCount <= 100_000) {
+            return lastCount * 2;
+        } else if (lastCount <= 400_000) {
+            return lastCount * 1.5;
```
sam-lord, line 35: Make sure we return an integer here

ghost/core/core/server/services/email-service/DomainWarmingService.js
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * @param {object} dependencies
+     * @param {{Email: object}} dependencies.models
+     */
+    constructor(dependencies) {
+        this.#emailModel = dependencies.models.Email;
+    }
+
+    async getLastCount() {
```
sam-lord, line 12: What happens if the last email was sent to a small segment (i.e. just paid members) -- we don't want to undo progress with warming up, so we should probably find the highest CSD count from before today.

ghost/core/core/server/services/email-service/DomainWarmingService.js
```diff
@@ -0,0 +1,42 @@
...
+        // Find the most recent email (except today) and return the csd_count
+        const email = await this.#emailModel.findOne({
+            filter: `created_at:<${new Date().toISOString().split('T')[0]}`,
+            order: 'created_at DESC'
```
sam-lord, line 16: Order by csd_email_count DESC to avoid above bug?

sam-lord:
Aileen mentioned that we could enable CSD sending for the premium segment, then for the free segment as a possible next step, to reduce flip-flopping between sends.

ghost/core/core/server/services/email-service/SendingService.js
```diff
@@ -127,16 +129,18 @@ class SendingService {
...
-            from: this.#emailRenderer.getFromAddress(post, newsletter),
+            from: this.#emailRenderer.getFromAddress(post, newsletter, !!options.useFallbackDomain),
             replyTo: this.#emailRenderer.getReplyToAddress(post, newsletter) ?? undefined,
             html: emailBody.html,
             plaintext: emailBody.plaintext,
             recipients,
             emailId: emailId,
-            replacementDefinitions: emailBody.replacements
+            replacementDefinitions: emailBody.replacements,
+            domain: options.useFallbackDomain ? this.#emailRenderer.getFallbackDomain() : undefined
```
sam-lord, line 139: Should the mailgun provider receive the boolean (useFallbackDomain) or the overriden domain.

As a secondary point, should this be named domainOverride to make it clear that the undefined value means letting the Mailgun provider use the domain from config?

ghost/core/core/server/services/email-service/EmailRenderer.js
```diff
@@ -247,11 +247,11 @@ class EmailRenderer {
         return locale;
     }
 
-    getFromAddress(post, newsletter) {
+    getFromAddress(post, newsletter, useFallback = false) {
         // Clean from address to ensure DMARC alignment
         const addresses = this.#emailAddressService.getAddress({
             from: this.#getRawFromAddress(post, newsletter)
-        });
+        }, {useFallbackAddress: useFallback});
```
sam-lord, line 254: Why naming change from `useFallback` to `useFallbackAddress`? Pick one and stick with it

sam-lord: Maybe this should be `useFallbackDomain`, since the implication is that the address should be generated to pass DMARC alignment when combined with _that domain_, instead of with the default bulk config domain

ghost/core/core/server/services/email-service/EmailRenderer.js
```diff
@@ -283,6 +283,10 @@ class EmailRenderer {
         return null;
     }
 
+    getFallbackDomain() {
```
sam-lord, line 286: Should this exist here, or somewhere else? When implementing this it felt like a natural place to put it, but is it correct semantically?

PR #25308, opened 2025-10-30
Added toggle switch to Dark mode button

peterzimon:
ref https://linear.app/ghost/issue/BER-2913/port-sidebar

- We didn't have an indication of the appearance mode in the current implementation. This PR adds a switch component to Shade and uses it in the new user menu for the dark mode button.

### DIFF
--- a/apps/admin/src/layout/app-sidebar/UserMenu.tsx
+++ b/apps/admin/src/layout/app-sidebar/UserMenu.tsx
@@ -89,7 +89,13 @@ function UserMenu({ ...props }: React.ComponentProps<typeof DropdownMenu>) {
                 >
                     <LucideIcon.Moon />
                     <span className="flex-1">Dark mode</span>
-                    <Switch size='sm' checked={dummyDarkMode}/>
+                    <Switch
+                        size='sm'
+                        checked={dummyDarkMode}
+                        onCheckedChange={setDummyDarkMode}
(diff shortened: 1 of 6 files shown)


apps/admin/src/layout/app-sidebar/UserMenu.tsx
```diff
@@ -78,9 +80,16 @@ function UserMenu({ ...props }: React.ComponentProps<typeof DropdownMenu>) {
...
+                    className="cursor-pointer text-base"
+                    onSelect={(e) => {
+                        e.preventDefault();
+                        setDummyDarkMode(!dummyDarkMode);
+                    }}
+                >
                     <LucideIcon.Moon />
-                    <span>Dark mode</span>
+                    <span className="flex-1">Dark mode</span>
+                    <Switch size='sm' checked={dummyDarkMode}/>
```
kevinansfield: Seeing as the menu item already has keyboard support and is clickable too I think we may want some extra props here to keep everything working if the switch itself gets the click event and to make sure it doesn't also receive keyboard selection?

```
onCheckedChange={setDummyDarkMode}
onClick={(e) => e.stopPropagation()}
tabIndex={-1}
```

peterzimon: Ah, good point! I'll add these

kevinansfield (approved):

PR #25310, opened 2025-10-30
Added EmberBridge global to sync state from React shell to Ember

kevinansfield:
ref https://linear.app/ghost/issue/BER-2929/

- added Ember instance initializer that makes our `state-bridge` service instance available on `window.EmberBridge.state`
- updated the react shell's framework setup to use `window.EmberBridge.state.{onUpdate,onInvalidate,onDelete}` in the same way as apps mounted inside Ember via `AdminXComponent`


jonatansberg (approved):
Looks good!

PR #25315, opened 2025-10-30
Added Ember to React data syncing

kevinansfield:
ref https://linear.app/ghost/issue/BER-2929/

- updated Ember's `state-bridge` service
  - extended from `Evented` to add `.{on,off,trigger}` methods to manage event subscription and triggering
  - added `triggerEmberDataChange` method that fires a `'emberDataChange'` event, containing the operation type, model, id, and response data that will allow React to invalidate or update its query cache
- updated `EmbeddedRelationAdapter` to call `stateBridge.triggerEmberDataChange` on any create/update/delete calls made on the Ember Data store
  - our `ApplicationAdapter` extends from `EmbeddedRelationAdapter` so this catches all store activity
- added `useEmberDataSync` hook on the React side
  - polls for `window.EmberBridge` to exist because it will only be available once Ember boots and that's now hosted inside of the React shell
  - once `window.EmberBridge` is available, subscribes to `'emberDataChange'` events
  - when an event occurs, looks up our React Query data type key based on the Ember Data model name and invalidates all queries using that key
  - only does query invalidation for now rather than optimistic query cache updates to keep things simpler, we can add optimistic updates if/when we find a use-case for it

### DIFF
--- a/apps/admin/src/App.tsx
+++ b/apps/admin/src/App.tsx
@@ -2,10 +2,12 @@ import { Outlet } from "@tryghost/admin-x-framework";
 import { useCurrentUser } from "@tryghost/admin-x-framework/api/currentUser";
 import { EmberProvider, EmberFallback, EmberRoot, useEmberAuthSync } from "./ember-bridge";
 import { AdminLayout } from "./layout/AdminLayout";
+import { useEmberDataSync } from "./ember-bridge";
 
 function App() {
     const { data: currentUser } = useCurrentUser();
     useEmberAuthSync();
+    useEmberDataSync();
 
     return (
         <EmberProvider>
--- a/apps/admin/src/ember-bridge/EmberBridge.tsx
+++ b/apps/admin/src/ember-bridge/EmberBridge.tsx
@@ -1,3 +1,6 @@
+import { useEffect } from "react";
+import { useQueryClient } from "@tanstack/react-query";
+
 export interface EmberBridge {
     state: StateBridge;
 }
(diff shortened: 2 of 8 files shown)


apps/admin/src/ember-bridge/EmberBridge.tsx
```diff
@@ -6,10 +9,91 @@ export interface StateBridge {
...
+    'setting': 'SettingsResponseType',
+    'theme': 'ThemesResponseType',
+    'tier': 'TiersResponseType',
+    'user': 'UsersResponseType',
+    'post': 'PostsResponseType',
+    'member': 'MembersResponseType',
+    'tag': 'TagsResponseType',
+    'label': 'LabelsResponseType',
+    'webhook': 'WebhooksResponseType'
+};
```
jonatansberg, line 48: It feels a bit silly that we have this mapping twice. I'd be inclined to either drop the `ResponseType` suffix and lowecase the keys on the react side as well or come up with a way that lets use one mapping bi-directionally by e.g. using the `emberDataTypeMapping` on the Ember side to trigger the events with the needed `responseType`.

apps/admin/src/App.tsx
```diff
@@ -2,10 +2,12 @@ import { Outlet } from "@tryghost/admin-x-framework";
 import { useCurrentUser } from "@tryghost/admin-x-framework/api/currentUser";
 import { EmberProvider, EmberFallback, EmberRoot, useEmberAuthSync } from "./ember-bridge";
 import { AdminLayout } from "./layout/AdminLayout";
+import { useEmberDataSync } from "./ember-bridge";
 
 function App() {
     const { data: currentUser } = useCurrentUser();
     useEmberAuthSync();
+    useEmberDataSync();
```
jonatansberg, line 10: It would be interesting to try and replace the very naive auth sync with something based on this next :)

jonatansberg (approved):
Looks good! The comment re duplicated mapping can be tackled as a followup thing.

apps/admin/src/App.tsx
kevinansfield: I think we'd need to fire some specific auth events from Ember as it won't be firing any of the existing adapter-based events. I can take a look at that whilst we're refining the JSON-field handling in React Query as we'll need something similar for the `customViews` setting that I was otherwise going to look at next

apps/admin/src/ember-bridge/EmberBridge.tsx
kevinansfield: > I'd be inclined to either drop the ResponseType suffix and lowecase the keys on the react side as well

Doing something like that would be my preference compared to trying to find a way to use the map across packages, especially when the Ember side encodes additional Ember-specific metadata.

I do notice one response type here that doesn't fit the pattern is `RecommendationResponseType` because it's not plural for some reason 😕

PR #25321, opened 2025-11-03
Add feature flag to enable admin forward opt-in

jonatansberg:
Allows users on ghost.io to opt into the new React-based admin interface via a private feature flag. The flag manages a cookie that controls which admin shell (Ember vs React) is served and reloading of the page when the flag is updated.

Also updates the state bridge to re-fetch features after settings reload. This ensures adminForward reconciliation runs when settings are updated from React. Previously, the feature service wasn't re-evaluating when settings changed via the state bridge, so the cookie/reload logic wouldn't trigger on toggle.

---

> [!NOTE]
> Adds `adminForward` private flag to opt into the React admin on ghost.io, managing a cookie and re-fetching features after settings reload to keep state in sync.
> 
> - **Private Labs/UI**
>   - Add "New Admin Experience" toggle in `apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx` (`adminForward`).
> - **Ember Admin**
>   - `ghost/admin/app/services/feature.js`:
>     - Add `adminForward` flag.
>     - Add cookie utilities and `inAdminForward` detection.
>     - Implement `_reconcileAdminForwardState` to set/remove `ghost-admin-forward` cookie and reload when flag/shell mismatch; invoke during `fetch()`.
>   - `ghost/admin/app/services/state-bridge.js`:
>     - After settings reload, call `feature.fetch()` to refresh flags.
> - **Core**
>   - Add `adminForward` to `PRIVATE_FEATURES` in `ghost/core/core/shared/labs.js`.
>   - Update config API snapshot to include `labs.adminForward`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ee05ba1fb7c7af7315a8e1fb6a62c49002f1c1dd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/admin/app/services/feature.js
+++ b/ghost/admin/app/services/feature.js
@@ -38,6 +38,27 @@ export function feature(name, options = {}) {
     }));
 }
 
+// Cookie utilities for admin forward functionality
+function getCookie(name) {
+    const value = `; ${document.cookie}`;
+    const parts = value.split(`; ${name}=`);
+    if (parts.length === 2) {
+        return parts.pop().split(';').shift();
+    }
+    return null;
+}
+
+function setCookie(name, value, days) {
+    const date = new Date();
+    date.setTime(date.getTime() + (days * 24 * 60 * 60 * 1000));
+    const expires = `expires=${date.toUTCString()}`;
+    document.cookie = `${name}=${value};${expires};path=/ghost/`;
+}
+
+function deleteCookie(name) {
+    document.cookie = `${name}=;expires=Thu, 01 Jan 1970 00:00:00 UTC;path=/ghost/`;
+}
+
 @classic
 export default class FeatureService extends Service {
     @service lazyLoader;
@@ -59,6 +80,7 @@ export default class FeatureService extends Service {
     @feature('referralInviteDismissed', {user: true}) referralInviteDismissed;
 
     // labs flags
+    @feature('adminForward') adminForward;
     @feature('audienceFeedback') audienceFeedback;
     @feature('webmentions') webmentions;
     @feature('stripeAutomaticTax') stripeAutomaticTax;
@@ -97,9 +119,41 @@ export default class FeatureService extends Service {
         }
     }
 
+    get inAdminForward() {
+        // Detect if Ember is running inside the React admin shell
+        // In React shell: Ember renders to #ember-app
+        // Standalone: Ember renders to body (no #ember-app element)
+        return document.querySelector('#ember-app') !== null;
+    }
+
+    _reconcileAdminForwardState() {
+        // Only proceed if we're on a *.ghost.io or *.ghost.is domain
+        const hostname = window.location.hostname;
+        if (!hostname.endsWith('.ghost.io') && !hostname.endsWith('.ghost.is')) {
+            return;
+        }
+
+        const cookieName = 'ghost-admin-forward';
+        const hasAdminForwardCookie = !!getCookie(cookieName);
+
+        // Update cookie based on feature flag
+        if (hasAdminForwardCookie && !this.adminForward) {
+            deleteCookie(cookieName);
+        } else if (!hasAdminForwardCookie && this.adminForward) {
+            setCookie(cookieName, '1', 365);
+        }
+
+        // Reload if flag state doesn't match current wrapper
+        // (flag enabled but in Ember standalone, or flag disabled but in React shell)
+        if (this.adminForward !== this.inAdminForward) {
+            window.location.reload();
+        }
+    }
+
     fetch() {
         return this.settings.fetch().then(() => {
             this.set('_user', this.session.user);
+            this._reconcileAdminForwardState();
             return this._setAdminTheme().then(() => true);
         });
     }
--- a/ghost/admin/app/services/state-bridge.js
+++ b/ghost/admin/app/services/state-bridge.js
@@ -67,7 +67,11 @@ export default class StateBridgeService extends Service {
             // Blog title is based on settings, but the one stored in config is used instead in various places
             this.config.blogTitle = response.settings.find(setting => setting.key === 'title').value;
 
-            this.settings.reload();
+            // TODO: Reloading settings does not trigger a re-fetch of the
+            // feature flags. We should maybe find a better way to do this.
+            this.settings.reload().then(() => {
+                this.feature.fetch();
+            });
         }
 
         if (dataType === 'TiersResponseType') {
--- a/apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx
+++ b/apps/admin-x-settings/src/components/settings/advanced/labs/PrivateFeatures.tsx
@@ -43,6 +43,10 @@ const features: Feature[] = [{
     title: 'Welcome Emails',
     description: 'Enables features related to sending welcome emails to new members',
     flag: 'welcomeEmails'
+}, {
+    title: 'New Admin Experience',
+    description: 'Try the new React-based admin interface. Only available on ghost.io',
+    flag: 'adminForward'
 }];
 
 const AlphaFeatures: React.FC = () => {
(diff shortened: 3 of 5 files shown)


ghost/admin/app/services/feature.js
```diff
@@ -38,6 +38,27 @@ export function feature(name, options = {}) {
...
+function setCookie(name, value, days) {
+    const date = new Date();
+    date.setTime(date.getTime() + (days * 24 * 60 * 60 * 1000));
+    const expires = `expires=${date.toUTCString()}`;
+    document.cookie = `${name}=${value};${expires};path=/ghost/`;
+}
+
+function deleteCookie(name) {
+    document.cookie = `${name}=;expires=Thu, 01 Jan 1970 00:00:00 UTC;path=/ghost/`;
+}
```
jonatansberg, line 60: Should we move these somewhere else? Use a third party library?

ghost/admin/app/services/feature.js
```diff
@@ -97,9 +119,40 @@ export default class FeatureService extends Service {
         }
     }
 
+    @computed()
+    get inAdminForward() {
```
jonatansberg, line 122: We can use this to toggle visibility of the Ember sidebar, loading of React components, etc. We could move it somewhere else as well if you prefer.

ghost/admin/app/services/feature.js
```diff
@@ -97,9 +119,40 @@ export default class FeatureService extends Service {
...
+        // Detect if Ember is running inside the React admin shell
+        // In React shell: Ember renders to #ember-app
+        // Standalone: Ember renders to body (no #ember-app element)
+        return document.querySelector('#ember-app') !== null;
+    }
+
+    _reconcileAdminForwardState() {
+        // Only proceed if we're on a *.ghost.io domain
+        if (!window.location.hostname.endsWith('.ghost.io')) {
```
kevinansfield: Does this need to account for `ghost.is` staging sites too?

jonatansberg: That's a good idea

ghost/admin/app/services/feature.js
```diff
@@ -97,9 +119,41 @@ export default class FeatureService extends Service {
         }
     }
 
+    @computed()
```
kevinansfield: Nit: No need for `@computed` here

ghost/admin/app/services/feature.js
```diff
@@ -97,9 +119,40 @@ export default class FeatureService extends Service {
...
+        // Update cookie based on feature flag
+        if (hasAdminForwardCookie && !this.adminForward) {
+            deleteCookie(cookieName);
+        } else if (!hasAdminForwardCookie && this.adminForward) {
+            setCookie(cookieName, '1', 365);
+        }
+
+        // Reload if we're in the wrong wrapper
+        if ((hasAdminForwardCookie && !this.inAdminForward) || (!hasAdminForwardCookie && this.inAdminForward)) {
+            window.location.reload();
```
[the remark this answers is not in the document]
jonatansberg: Oops. Fixed that now.

ghost/admin/app/services/feature.js
jonatansberg: Haha, yeah, no. Thats silly. I'll fix that.

ghost/admin/app/services/feature.js
kevinansfield: It feels a bit weird to me that we're handling the cookie management client-side. Especially when trying to sync the cookie state based on fetched labs flags because that introduces timing considerations that may prove odd when working on migrating auth screens - settings/labs flags aren't fetched until after authentication occurs.

Admin (and I think probably our AdminX apps too) start with a blocking request to `/ghost/api/admin/site/` which is an unauthenticated endpoint built on top of config/labs. I wonder if it makes more sense to put the cookie management in some server-side middleware and attach it that endpoint? That way there's no client-side syncing needed and we can simplify the feature toggle to doing a refetch of that endpoint

jonatansberg: We only really need the cookie sync a) when changing the flag and b) when visiting a site that has been enrolled and we lack the cookie flag. 

I'd like to keep the code related to the flagging as co-located as possible, so I'm not a huge fan of starting to mess about with e.g. the `/site` endpoint. We'd also still need some sort of client side logic to determine if we need to reload the page or not since all static assets are served via the CDN and we don't have a way to determine which of the admin versions a request originates from, nor a way to affect that from the server.

Also, I'm not sure I follow with regards to your timing concerns? The `inAdminForward` flag would work immediately in Ember, and if we want to override any of the routes on the React end that also doesn't really matter from the point of view of this feature flagging.

ghost/admin/app/services/feature.js
kevinansfield: 👍 makes sense to keep it co-located

ghost/admin/app/services/state-bridge.js
```diff
@@ -68,7 +67,11 @@ export default class StateBridgeService extends Service.extend(Evented) {
             // Blog title is based on settings, but the one stored in config is used instead in various places
             this.config.blogTitle = response.settings.find(setting => setting.key === 'title').value;
 
-            this.settings.reload();
+            // TODO: Reloading settings does not trigger a re-fetch of the
+            // feature flags. We should maybe find a better way to do this.
```
kevinansfield, line 71: I think this is fine for now as it's quite explicit about what's happening when changes come in from an external source.

If we wanted to change it within Ember it would probably be adding the `feature.fetch()` call to the settings service `reload` method and then removing the explicit `feature.fetch()` call in our [postAuthPreparation() method](https://github.com/TryGhost/Ghost/blob/main/ghost/admin/app/services/session.js#L46).

ghost/admin/app/services/feature.js
kevinansfield: Yeah, I suppose having the cookie set on `/site/` only really helps to cut a few steps in the incognito React shell auth testing scenario because it would be able to automatically refresh before you see the auth screens.

With this `feature` service controlled flow you'd have to go through the sign-in/sign-out step to get to the React auth screens, e.g.

1. open in incognito/new browser (always Ember shell as no cookies)
2. sign-in using Ember shell, reconciliation happens and refresh occurs
3. sign-out in React shell
4. now you're on the React shell sign-in screen

With the `/site/` response setting a cookie then the reconciliation code would exist in the [config-manager service](https://github.com/TryGhost/Ghost/blob/main/ghost/admin/app/services/config-manager.js).

1. open in incognito (still always Ember shell)
    a. `/site` is requested, JS reads cookie and refreshes if cookie exists but we're not in app shell
2. first screen you see is the sign-in (or setup) screen in React shell

kevinansfield: I don't think having the cookie management in the feature service is a blocker, just something we should be aware of when we get to migrating auth screens as it won't necessarily be obvious which shell you've loaded.

I'm also happy keeping the direct cookie twiddling via DOM rather than introducing a new library just for this one use-case 👍

kevinansfield (approved):
I have a minor reservation about the client-side cookie management but it's not a blocker.

Some test snapshots always need updating after adding a new labs flag (source of toil?), otherwise this looks good 🙌

ghost/admin/app/services/feature.js
jonatansberg: > 1. open in incognito/new browser (always Ember shell as no cookies)
> 2. sign-in using Ember shell, reconciliation happens and refresh occurs
> 3. sign-out in React shell
> 4. now you're on the React shell sign-in screen

I can see how that specific testing scenario could be annoying when you try to run it incognito in production or on staging. Locally, you would be running on different origins the way things are now, so that way it's quite easy to tell which is which. 

I think that's a reasonable trade-off if that lets us keep everything in one place.

ghost/admin/app/services/state-bridge.js
jonatansberg: If you're okay with this I'll leave this as is. Feel free to take a stab at cleaning this up on the Ember side, otherwise we can aim for deleting the code rather than fixing it :)

jonatansberg:
Toil indeed.... Pushed a fixed version now.

PR #25323, opened 2025-11-03
Fix blank screen when viewing tag details

jonatansberg:
Ref https://linear.app/ghost/issue/BER-2963/fixed-blank-screen-when-viewing-tag-details

Override blank tag detail route to display Ember screen instead of empty fallback, allowing us to correctly fall back to Ember in the React admin shell without also displaying the error screen when running React within Ember.


kevinansfield:
Nice 👌

kevinansfield (approved):

PR #25330, opened 2025-11-03
Migrations for domain warmup feature

sam-lord:
closes [GVA-583](https://linear.app/ghost/issue/GVA-583/create-migration-in-ghost-to-add-csd-member-count-to-emails)
closes [GVA-587](https://linear.app/ghost/issue/GVA-587/create-a-migration-for-the-email-batches-table-to-say-whether-its-sent)

These migrations are both required for the feature. I ran an experimental spike to see what data we'd need to get domain warmup working, and these two came out:

* A count of the messages that were sent by the CSD for each email => this lets us ramp up the number of emails that we send from the CSD on each subsequent send
* A boolean flag that says whether an email batch was sent by the fallback domain (a back-up domain configured for if we're going to exceed the CSD's budget)

---

> [!NOTE]
> Adds DB migrations and schema/model/test updates for `emails.csd_email_count` and `email_batches.fallback_sending_domain`.
> 
> - **Database/Migrations**
>   - Add `emails.csd_email_count` (nullable integer, unsigned) via `versions/6.7/2025-11-03-15-17-05-add-csd-email-count.js`.
>   - Add `email_batches.fallback_sending_domain` (boolean, default `false`) via `versions/6.7/2025-11-03-15-18-04-add-email-batch-fallback-domain.js`.
>   - Update `core/server/data/schema/schema.js` to include both columns.
> - **Models**
>   - Update `core/server/models/email-batch.js` defaults to include `fallback_sending_domain: false`.
> - **Tests/Fixtures**
>   - Update fixtures in `test/utils/fixtures/data-generator.js` to populate `fallback_sending_domain` and reflect new schema.
>   - Refresh schema hash in `test/unit/server/data/schema/integrity.test.js`.
>   - Adjust E2E/API snapshots to include `csd_email_count` and changed `content-length`s.
>   - Fix link HTML in `test/e2e-server/click-tracking.test.js` (adds missing quotes).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ecb9bec83b0e6452a12f777a2481462700a62939. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -838,6 +838,7 @@ module.exports = {
         error: {type: 'string', maxlength: 2000, nullable: true},
         error_data: {type: 'text', maxlength: 1000000000, fieldtype: 'long', nullable: true},
         email_count: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0},
+        csd_email_count: {type: 'integer', nullable: true, unsigned: true},
         delivered_count: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0},
         opened_count: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0},
         failed_count: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0},
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -867,6 +867,7 @@ module.exports = {
         id: {type: 'string', maxlength: 24, nullable: false, primary: true},
         email_id: {type: 'string', maxlength: 24, nullable: false, references: 'emails.id'},
         provider_id: {type: 'string', maxlength: 255, nullable: true},
+        fallback_sending_domain: {type: 'boolean', nullable: false, defaultTo: false},
         status: {
             type: 'string',
             maxlength: 50,
--- a/ghost/core/test/e2e-server/click-tracking.test.js
+++ b/ghost/core/test/e2e-server/click-tracking.test.js
@@ -39,7 +39,7 @@ describe('Click Tracking', function () {
             body: {
                 posts: [{
                     title: 'My Newsletter',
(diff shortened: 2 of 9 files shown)


ghost/core/core/server/data/schema/schema.js
```diff
@@ -838,6 +838,7 @@ module.exports = {
         error: {type: 'string', maxlength: 2000, nullable: true},
         error_data: {type: 'text', maxlength: 1000000000, fieldtype: 'long', nullable: true},
         email_count: {type: 'integer', nullable: false, unsigned: true, defaultTo: 0},
+        csd_email_count: {type: 'integer', nullable: true, unsigned: true},
```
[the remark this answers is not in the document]
sam-lord: The reason we're not defaulting this to `0` is that the `csd_email_count` will not always be set, if domain warming isn't active, or if there isn't a custom sending domain.

We can use the fact this is null to know when we started warming a domain, too.

aileen (changes requested):
I think the acceptance tests need some care. Other than that, I think this migration is good. The naming of the new columns makes sense to me as well. I also agree with omitting the `defaultTo: 0`, so we have another indicator if the feature is used at all. I would expect that - when we turn the feature on and the user has a custom sending domain, but not send anything yet, we'd set it to `0`?

ghost/core/test/e2e-server/click-tracking.test.js
```diff
@@ -39,7 +39,7 @@ describe('Click Tracking', function () {
             body: {
                 posts: [{
                     title: 'My Newsletter',
-                    html: `<p>External link <a href="https://example.com/a">https://example.com/a</a>; Internal link <a href=${siteUrl.href}/about">${siteUrl.href}/about</a>;Ghost homepage <a href="https://ghost.org">https://ghost.org</a></p>`
+                    html: `<p>External link <a href="https://example.com/a">https://example.com/a</a>; Internal link <a href="${siteUrl.href}/about">${siteUrl.href}/about</a>;Ghost homepage <a href="https://ghost.org">https://ghost.org</a></p>`
```
sam-lord, line 42: I don't understand why I'm having to fix this 🤔

aileen: That is strange indeed. Maybe something to flag up in our development channel?

aileen (approved):
LGTM

PR #25337, opened 2025-11-04
Fixed version mismatch error in React admin shell

jonatansberg:
Closes https://linear.app/ghost/issue/BER-2969/version-mismatch-error-in-production

Omitting the `X-Ghost-Version` header when running in the React admin shell avoids version mismatch errors when the continuously delivered version of the admin is ahead of the backend releases. When the header is not set the backend skips the version check.

This is needed since the continuously delivered admin can be on a newer version than production backends during e.g. the window between a minor release being cut and rolled out. 

Note that removing this check on the backend does not prevent us from implementing our own check in the frontend to handle clients that are lagging behind due e.g. to not reloading their browser for a very long time. 

---

> [!NOTE]
> Conditionally skip sending the X-Ghost-Version header when running the forward admin, updating both React fetch and Ember ajax paths to avoid version mismatch errors.
> 
> - **Header handling**:
>   - In `apps/admin-x-framework/src/utils/api/fetchApi.ts`, only add `x-ghost-version` when `ghostVersion` is provided; keep `app-pragma` default.
> - **Ember Admin**:
>   - `ghost/admin/app/components/admin-x/admin-x-component.js`: pass `framework.ghostVersion` as `''` when `feature.inAdminForward` is true.
>   - `ghost/admin/app/services/ajax.js`:
>     - Inject `feature` service.
>     - Return headers with `App-Pragma` always; include `X-Ghost-Version` only when not in forward admin.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d7cbba08444be49ca4fc405fef4c78d9ab6a009a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/admin-x-framework/src/utils/api/fetchApi.ts
+++ b/apps/admin-x-framework/src/utils/api/fetchApi.ts
@@ -22,9 +22,15 @@ export const useFetchApi = () => {
     return async <ResponseData = any>(endpoint: string | URL, {headers = {}, retry, ...options}: RequestOptions = {}): Promise<ResponseData> => {
         // By default, we set the Content-Type header to application/json
         const defaultHeaders: Record<string, string> = {
-            'app-pragma': 'no-cache',
-            'x-ghost-version': ghostVersion
+            'app-pragma': 'no-cache'
         };
+        
+        // Only include version header if ghostVersion is provided
+        // This allows forward admin deployments to skip version checks
+        if (ghostVersion) {
+            defaultHeaders['x-ghost-version'] = ghostVersion;
+        }
+        
         if (typeof options.body === 'string') {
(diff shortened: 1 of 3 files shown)


kevinansfield (approved):
Looks like a good approach for now 👌

As mentioned earlier it would be great to re-examine these version checks in general, although it's definitely not in scope for this project. It would be helpful to put some sort of version header back in the future as we've made use of it in server logs at times.

jonatansberg:
That's definitely a valid point. I think we should tweak the behavior on the server so that we can reintroduce this header down the line. Thats not a change I'd like to do now, but we can add an issue to the cleanup milestone so that we get back to it.

PR #25339, opened 2025-11-04
✨ Moved members flow to new E2E framework

ibalosh:
ref https://linear.app/ghost/issue/NY-733/port-members-creation-tests

**What**

* Migrated member CRUD flow from legacy browser tests to the new suite
* Migrated member impersonation
* Added member factory - in tests using only `build`, `create` with persistency works too (checked)
* Added appropriate Member POMs and locators
* removed old tests
* cleaned up some of the `playwright lint warnings` along the way

**Why**

* Old tests are tightly coupled, dependent on each other, hard to maintain and read
* If you remove one member test, you get other tests to fail


e2e/tests/admin/members/members.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        await memberDetailsPage.nameInput.fill(memberData.name);
+        await memberDetailsPage.emailInput.fill(memberData.email);
+        await memberDetailsPage.noteInput.fill(memberData.note);
+        await memberDetailsPage.addLabel(memberData.label);
+        await memberDetailsPage.save();
+
+        await membersPage.goto();
+
+        await expect(membersPage.memberListItems).toHaveCount(1);
+        await expect(membersPage.getMemberEmail(memberData.name)).toHaveText(memberData.email);
```
cmraible: This is so much better than:
```
const memberEmail = sharedPage.locator('tbody > tr > a > div > div > p').nth(0);

```
:heart_eyes: [](edited-with-Axolo)

cmraible (commented):
Looking good! Started to review this but it looks like you're still working on it, LMK when you're ready for a full review 😄

e2e/tests/admin/members/members.test.ts
ibalosh: yeah the members test is a good one :slightly_smiling_face: it's so much different that I almost wanted to take a screen comparison screenshot :D

the members test is a good example of what we have been trying to solve, the old one:

- tightly coupled (create test has to run before edit test in order for it to work)
- not readable
- tricky to run locally (your stripe_xyz_key is missing...)
- hard to maintain
all the work on factories, poms, environment management is starting to pay of

I am proud of what we built together @9larsons @cmraible :heart:

(and I think as we make it more sophisticated with fixtures for old data, stripe.. it will be just even better)

Next one after: api... (Chris already seeded that idea in one of our calls :slightly_smiling_face: ) [](edited-with-Axolo)

e2e/tests/admin/members/members.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+        // Edit the member
+        await membersPage.goto();
+        await membersPage.getMemberByName(memberToCreate.name).click();
+
+        const editedMember = {
+            name: 'Test Member Edited',
+            email: 'edited@ghost.org',
+            note: 'This is an edited test member'
+        };
```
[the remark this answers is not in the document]
ibalosh: fixed

e2e/tests/admin/members/members.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+        await memberDetailsPage.emailInput.fill('invalid-email-address');
+        await memberDetailsPage.saveButton.click();
+
+        await expect(memberDetailsPage.retryButton).toBeVisible();
+        await expect(page.getByText('Invalid Email')).toBeVisible();
+
+        await membersPage.goto();
+        await memberDetailsPage.confirmLeaveButton.click();
+        await expect(membersPage.getMemberByName('membertocreate@ghost.org')).toBeVisible();
```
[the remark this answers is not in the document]
ibalosh: fixed

ibalosh:
ok, @9larsons @cmraible , this is ready for review.

It took bit longer than I expected, until I figured out the correlation between tests in old suite.
I decided to port member exporting separately, since PR will get too big.
 [](edited-with-Axolo)

e2e/helpers/pages/admin/members/MemberDetailsPage.ts
```diff
@@ -4,26 +4,64 @@ import {Locator, Page} from '@playwright/test';
...
+        await this.page.keyboard.type(label);
+        await this.page.keyboard.press('Tab');
     }
 
-    async updateName(name: string): Promise<void> {
-        await this.nameInput.fill(name);
+    async removeLabel(): Promise<void> {
+        await this.labelsInput.click();
+        await this.page.keyboard.press('Backspace');
+        await this.page.keyboard.press('Tab');
```
9larsons: Tab in my browser (Safari) takes you up to the URL bar which I think could be disruptive. This also doesn't allow us to specify a label which might be nice if we need/want to test more than one. I'm ok with that change coming later, just clarifying that this is very limited in the current implementation.

e2e/helpers/pages/portal/PortalPage.ts
```diff
@@ -1,24 +1,27 @@
...
+    readonly closeButton: Locator;
+    readonly contactSupportButton: Locator;
+    readonly body: Locator;
 
     constructor(page: Page) {
         super(page);
         this.portalFrame = page.frameLocator(this.frameSelector);
+
+        this.closeButton = this.portalFrame.getByRole('button', {name: 'Close'});
+        this.contactSupportButton = this.portalFrame.getByRole('link', {name: 'Contact support'});
```
9larsons: I don't think contact support is available on every page.

e2e/helpers/pages/portal/PortalPage.ts
```diff
@@ -1,24 +1,27 @@
...
+        this.body = this.portalFrame.locator('body');
```
9larsons: Would you rather assert on body than what's on a given page?

e2e/tests/admin/members/member-impersonation.test.ts
```diff
@@ -0,0 +1,42 @@
+import {HomePage, MemberDetailsPage, MembersPage, PortalPage} from '../../../helpers/pages';
+import {MemberFactory, createMemberFactory} from '../../../data-factory';
+import {expect, test} from '../../../helpers/playwright';
+
+test.describe('Ghost Admin - Member Impersonation', () => {
+    let memberFactory: MemberFactory;
+
+    test.beforeEach(async ({page}) => {
+        memberFactory = createMemberFactory(page.request);
```
9larsons, line 9: I see we use this pattern elsewhere - I think it'd make more sense to only pass page rather than expecting the user to know to pass page.request. Wdyt?

e2e/tests/admin/members/member-impersonation.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    test('impersonates a member and verifies magic link generation', async ({page}) => {
+        const memberToImpersonate = memberFactory.build({email: 'impersonate@ghost.org'});
```
9larsons: This could be even cleaner with destructuring. edit: same for other instances below.

e2e/tests/admin/members/members.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    });
+
+    test('creates a new member with valid details', async ({page}) => {
+        const memberToCreate = memberFactory.build({labels: ['test'], email: 'membertocreate@ghost.org'});
+
+        const membersPage = new MembersPage(page);
+        await membersPage.goto();
+        await membersPage.newMemberButton.click();
+
+        const memberDetailsPage = new MemberDetailsPage(page);
```
9larsons, line 19: We do this same pattern of fill and save - how about a helper?

e2e/tests/admin/members/members.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        await expect(memberDetailsPage.retryButton).toBeVisible();
+        await expect(memberDetailsPage.body).toContainText('Invalid Email');
+
+        await membersPage.goto();
+        await memberDetailsPage.confirmLeaveButton.click();
+        await expect(membersPage.getMemberEmail('membertocreate@ghost.org')).toBeVisible();
+    });
+
+    test('deletes an existing member', async ({page}) => {
+        const toBeDeletedMember = memberFactory.build();
```
9larsons: How come we didn't implement the API here? I get wanting to keep this clean/minimal, but I would rather see this test use `memberFactory.create()` rather than go through the workflow of creating and then deleting, as earlier tests test the actual creation via the UI.

e2e/tests/admin/whats-new.test.ts
```diff
@@ -282,7 +282,7 @@ test.describe('Ghost Admin - What\'s New', () => {
                 await modal.close();
 
                 await menu.goto();
-                await expect(menu.avatarBadge).not.toBeVisible();
+                await expect(menu.avatarBadge).toBeHidden();
```
9larsons, line 285: Maybe we can make a linter rule around this.

9larsons (changes requested):
Hey - several comments. Overall, looks and feels so much nicer! It's pretty damned great to be able to move over some of these crusty tests.

I realized as I finished I should've been using conventional comments so I'll try to remember next time around.

e2e/helpers/pages/admin/members/MemberDetailsPage.ts
cmraible: > Tab in my browser (Safari) takes you up to the URL bar

That seems wrong, I'm not seeing the same in Safari. Tab is pretty universally used for keyboard navigation within a web page, I'd be really surprised if any major browsers handle it differently. I wonder if the behavior you're describing is from an extension or something?

cmraible: > This also doesn't allow us to specify a label which might be nice if we need/want to test more than one

Agree here though, it would be nice to be able to remove a specific label.

e2e/tests/admin/members/member-impersonation.test.ts
ibalosh: I was thinking about this, and the reason it looks weird is the way Playwright names the client they are using to do requests and handle responses. 

The factories we use require some client to make api requests (fetch, axios..). Only reason we need Playwright's client is since we would like a client which also has authentication headers already set in it, so we don't need to worry about authentication.

If we would pass the page, we would pass too much information to the factory, and couple it with page, while in reality factories only need to know about client interface. We want to keep connections with factories minimal.

I agree though that it's harder to see here why do we actually pass that specific client with `page.request`, and not use `fetch`, `axios` or something else for example.

I am wondering, would it be better if we would create a helper in playwright folder/utils like:

`getPlaywrightHttpClient(page)` or something like that. It would make this line longer, but at least we could comment this function in that helper, and describe why we would use playwrights helper at all.

What do you think?

e2e/tests/admin/members/member-impersonation.test.ts
ibalosh: good point, updated 👌

ibalosh: In other tests, I left it non-destructured version, the reason for that is to describe the test scenarios better, like `memberToCreate.name` , `editedMember.name`

e2e/tests/admin/members/members.test.ts
ibalosh: good point 👍 I missed that since I added tests first and then factories. Updated tests to use factories now. Makes them shorter, faster and cleaner. tx!

e2e/tests/admin/whats-new.test.ts
ibalosh: yeah, this actually got caught with latest linting rules for playwright that @cmraible added 🙌

e2e/helpers/pages/portal/PortalPage.ts
ibalosh: I removed it since it's not used yet, once we would need it we can decide whether to have it here, or in specific portal for page where we test it

e2e/helpers/pages/portal/PortalPage.ts
ibalosh: hm, good point, I wanted to assert what's in frame. Indirectly by clicking the account button on homepage, I wanted to assert minimally homepage too.

To describe that better, I updated the name of this locator to `portalFrameBody`

e2e/tests/admin/members/members.test.ts
ibalosh: hm, good point, I added a function action to member details to resolve this, that allows passing more parameters, we can improve it further once we pass even more parameters

e2e/helpers/pages/admin/members/MemberDetailsPage.ts
ibalosh: Hey guys, I followed a bit too much original tests. I updated this to be able to remove specific label, and added helper to be able to remove them all.

(regarding `tab` , it should work for adding labels though, if not that seems like error in UI or something with browser - works for me in FF, Chrome, Safari)

e2e/tests/admin/members/member-impersonation.test.ts
ibalosh: updated this so it's clearer what `page.request` is and how we use it, with description

ibalosh:
hey @9larsons @cmraible , I went through all the code review comments and updates the tests

e2e/tests/admin/members/members.test.ts
```diff
@@ -0,0 +1,118 @@
...
+            email: 'original@example.com',
+            note: 'original note',
+            labels: ['original']
+        });
+
+        const membersPage = new MembersPage(page);
+        await membersPage.goto();
+
+        // Edit the member
+        await membersPage.goto();
```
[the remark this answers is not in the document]
ibalosh: fixed

e2e/tests/admin/members/member-impersonation.test.ts
ibalosh: :+1: reverted that change [](edited-with-Axolo)

e2e/tests/admin/members/member-impersonation.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+test.describe('Ghost Admin - Member Impersonation', () => {
+    let memberFactory: MemberFactory;
+
+    test.beforeEach(async ({page}) => {
+        memberFactory = createMemberFactory(page.request);
+    });
+
+    test('impersonates a member and verifies magic link generation', async ({page}) => {
+        const {name, email, note} = memberFactory.build({email: 'impersonate@ghost.org'});
```
9larsons: nit: This is one where I'd argue for using the factory to create the member instead of going through the creation flow manually, which is tested elsewhere. (I know we don't have the API wired up)

e2e/helpers/pages/admin/members/MembersPage.ts
```diff
@@ -2,16 +2,32 @@ import {AdminPage} from '../AdminPage';
...
+
+    getMemberByName(name: string): Locator {
+        return this.memberListItems.filter({hasText: name});
+    }
+
+    getMemberEmail(memberName: string): Locator {
+        return this.memberListItems.filter({hasText: memberName}).getByRole('paragraph');
+    }
+
+    async getMemberCount(): Promise<number> {
```
9larsons: nit: This could be `getMemberListCount` since it's not a true member count as it's paginated and loaded on scroll.

e2e/helpers/pages/admin/members/MembersPage.ts
```diff
@@ -2,16 +2,32 @@ import {AdminPage} from '../AdminPage';
...
     async clickMemberByEmail(email: string): Promise<void> {
-        await this.membersTable.getByRole('row').filter({hasText: email}).click();
+        await this.memberListItems.filter({hasText: email}).click();
+    }
+
+    getMemberByName(name: string): Locator {
+        return this.memberListItems.filter({hasText: name});
+    }
+
+    getMemberEmail(memberName: string): Locator {
```
9larsons, line 26: nit: This could call `getMemberByName`.

e2e/tests/admin/members/impersonation.test.ts
9larsons, line 1: nit: since this is in `/members/`, could we just call this `impersonation.test.ts`?

ibalosh: yep, updated 👍

e2e/helpers/pages/admin/members/MembersPage.ts
ibalosh: updated

e2e/helpers/pages/admin/members/MembersPage.ts
ibalosh: updated

e2e/tests/admin/members/members.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        const memberDetailsPage = new MemberDetailsPage(page);
+        await memberDetailsPage.fillMemberDetails(memberToCreate.name, memberToCreate.email, memberToCreate.note);
+        await memberDetailsPage.saveButton.click();
+
+        await expect(memberDetailsPage.retryButton).toBeVisible();
+        await expect(memberDetailsPage.body).toContainText('Invalid Email');
+    });
+
+    test('updates an existing member', async ({page}) => {
+        const memberToCreate = await memberFactory.create({
```
9larsons: What does this do? I wouldn't expect this to work as I don't see a .create implementation, so it must be the inherited, or I'm just missing it.

e2e/tests/admin/members/members.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+        await expect(membersPage.memberListItems).toHaveCount(1);
+        await expect(membersPage.getMemberByName(editedMember.name)).toBeVisible();
+        await expect(membersPage.getMemberEmail(editedMember.name)).toHaveText('edited@ghost.org');
+    });
+
+    test('cannot update an existing member with invalid email', async ({page}) => {
+        const {name, email, note} = memberFactory.build({email: 'membertocreate@ghost.org', name: 'Test Member'});
+
+        const membersPage = new MembersPage(page);
```
9larsons, line 84: Same as the above w/ creating a member via API. Not a big deal.

e2e/tests/admin/members/member-impersonation.test.ts
ibalosh: gotcha, updated this one too now, should be up in repo in a minute

e2e/tests/admin/members/members.test.ts
ibalosh: hey, `.create` is inherited from it's [parent factory](https://github.com/TryGhost/Ghost/blob/4f3fb13572885b9c9165e2765e8d01bfaf2f3a8d/e2e/data-factory/factory.ts) 

due this, it did not need an implementation, it works out of the box, and uses API to create member, we only needed to specify adapter in `createMemberFactory` in setup

9larsons (approved):
Changes look great! Thanks for starting this process.

e2e/tests/admin/members/members.test.ts
ibalosh: hah, yeah, I had the same feeling :) updated the name though to describe that better

PR #25342, opened 2025-11-04
Added email sending to welcome email job

troyciesco:
closes https://linear.app/ghost/issue/NY-740

- adds `memberWelcomeEmailTestInbox` to config
- when a new member joins, it sends an email to that inbox (set elsewhere, out of version control) instead of to the member
- this will allow us to test the performance of welcome email sending based on spikes in member sign ups without needing to worry about content of the email etc. yet


troyciesco:
Noting here that there's a few ai review comments from a previous PR that make more sense to address here:
- https://github.com/TryGhost/Ghost/pull/25288#discussion_r2512336623
- https://github.com/TryGhost/Ghost/pull/25288#discussion_r2512340281

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -24,7 +24,7 @@ const messages = {
 
 const SUBSCRIPTION_STATUS_TRIALING = 'trialing';
 
-const WELCOME_EMAIL_SOURCES = ['member'];
+const WELCOME_EMAIL_SOURCES = ['member', 'api'];
```
troyciesco: Note to self to clean this up before merge - this was for easier testing on staging

ghost/core/core/server/services/member-welcome-emails/jobs/lib/mail-context.js
```diff
@@ -0,0 +1,73 @@
...
+        return acc;
+    }, {});
+}
+
+/**
+ * @param {Object} options
+ * @param {import('../../../../data/db')} options.db
+ * @returns {Promise<MailConfig>}
+ */
+async function ensureInitialized({db}) {
```
troyciesco: the point with this file was to only fetch info for the email once per job run, which it does, but also to avoid having to prop drill the mailConfig from `processOutbox` -> `processEntires` -> `processEntry` -> `sendWelcomeEmail`. But the way this is set up might also be harder to reason about 🤔

troyciesco: switched it to get the mailConfig at the beginning of processOutbox and passed it through - simpler to reason about for now.

cmraible (approved):
Working well for me locally 👌 

No notes, LGTM!

PR #25345, opened 2025-11-05
Zimo/comment moderation poc

cmraible:

- Why are you making it?
- What does it do?

- [ ] I've read and followed the [Contributor Guide](https://github.com/TryGhost/Ghost/blob/main/.github/CONTRIBUTING.md)
- [ ] I've explained my change

### DIFF
--- a/apps/posts/src/views/Comments/Comments.tsx
+++ b/apps/posts/src/views/Comments/Comments.tsx
@@ -1,0 +1,61 @@
+import CommentsContent from './components/CommentsContent';
+import CommentsHeader from './components/CommentsHeader';
+import CommentsLayout from './components/CommentsLayout';
+import CommentsList from './components/CommentsList';
+import React from 'react';
+import {Button, EmptyIndicator, LoadingIndicator, LucideIcon} from '@tryghost/shade';
+import {useBrowseComments} from '@tryghost/admin-x-framework/api/comments';
+
+const Comments: React.FC = () => {
+    const {
+        data,
+        isError,
+        isLoading,
+        isFetchingNextPage,
+        fetchNextPage,
+        hasNextPage
+    } = useBrowseComments();
+
+    return (
+        <CommentsLayout>
+            <CommentsHeader />
+            <CommentsContent>
+                {isLoading ? (
+                    <div className="flex h-full items-center justify-center">
+                        <LoadingIndicator size="lg" />
+                    </div>
+                ) : isError ? (
+                    <div className="mb-16 flex h-full flex-col items-center justify-center">
+                        <h2 className="mb-2 text-xl font-medium">
+                            Error loading comments
+                        </h2>
+                        <p className="mb-4 text-muted-foreground">
+                            Please reload the page to try again
+                        </p>
+                        <Button onClick={() => window.location.reload()}>
+                            Reload page
+                        </Button>
+                    </div>
+                ) : !data?.comments.length ? (
+                    <div className="flex h-full items-center justify-center">
+                        <EmptyIndicator
+                            title="No comments yet"
+                        >
+                            <LucideIcon.MessageSquare />
+                        </EmptyIndicator>
+                    </div>
+                ) : (
+                    <CommentsList
+                        fetchNextPage={fetchNextPage}
+                        hasNextPage={hasNextPage}
+                        isFetchingNextPage={isFetchingNextPage}
+                        items={data?.comments ?? []}
+                        totalItems={data?.meta?.pagination?.total ?? 0}
+                    />
+                )}
+            </CommentsContent>
+        </CommentsLayout>
+    );
+};
+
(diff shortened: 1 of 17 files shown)


ghost/core/core/server/api/endpoints/comments.js
```diff
@@ -88,12 +88,23 @@ const controller = {
         validation: {
             options: {
                 post_id: {
-                    required: true
+                    required: false
                 }
             }
         },
         permissions: true,
```
cmraible, line 95: If only admins are able to view the comments in the UI, we should add the same permissions to the API

ghost/core/core/server/api/endpoints/comments.js
```diff
@@ -88,12 +88,23 @@ const controller = {
...
+            // If post_id is not provided, browse all comments
+            if (!frame.options.post_id) {
+                // Use direct model query for browsing all comments
+                const options = {
+                    ...frame.options,
+                    withRelated: ['member', 'post'],
+                    isAdmin: true
+                };
+                const result = await models.Comment.findPage(options);
+                return result;
```
cmraible, line 106: We should move this into the commentsService I think, so the existing `adminBrowse` method can handle this conditional logic/filtering

apps/posts/src/views/Comments/Comments.tsx
```diff
@@ -0,0 +1,61 @@
...
+                ) : (
+                    <CommentsList
+                        fetchNextPage={fetchNextPage}
+                        hasNextPage={hasNextPage}
+                        isFetchingNextPage={isFetchingNextPage}
+                        items={data?.comments ?? []}
+                        totalItems={data?.meta?.pagination?.total ?? 0}
+                    />
+                )}
+            </CommentsContent>
```
cmraible, line 56: I'd like to refactor this, it's really hard to follow what's happening with the multiple nesting ternary operators

apps/posts/src/views/Comments/components/CommentsList.tsx
```diff
@@ -0,0 +1,267 @@
...
+                            return <PlaceholderRow key={key} {...props} />;
+                        }
+
+                        const commentText = stripHtml(item.html);
+                        const truncatedComment = commentText.length > 100
+                            ? `${commentText.substring(0, 100)}...`
+                            : commentText;
+
+                        return (
+                            <TableRow
```
cmraible, line 156: I think the TableRow should be its own component to clean this up a bit, i.e. `CommentListItem`

cmraible (commented):
Splitting my feedback into two sections:
- Technical requirements to ship this safely
- Thoughts on the UX

## Technical Requirements to ship this safely, with (rough) estimates:
- [ ] Refactor the comments endpoint and add/update e2e-api tests - 0.5 days
- [ ] Add basic e2e tests for browsing/hiding/deleting comments - 1 day
- [ ] Add pagination (backend and frontend) for the comments list view - 1 day
- [ ] Some minor refactoring of the frontend components - 0.5 days

All in, I'd say ~1-2 weeks for 1 engineer to get the current POC shipped safely, but I think there are also some UX gaps that we'd want to fill before shipping, especially if we're skipping straight to GA, which would extend that timeline depending on what we decide on.

## Thoughts on the UX

I'm actually more concerned about the UX and whether this is the right thing to build than the more technical things above. I'm curious about the specific use-case for this, and if there might be a simpler solution than creating a whole new CRUD view in Admin.

Personally as a user I'd much rather have the tools to moderate comments in context within the comments UI itself, which we already somewhat have with the ability to hide a comment from the comments UI. Having a single really long list of all the comments out of context in Admin doesn't make a lot of sense to me as a user, but I might be overlooking valid use-cases for larger/more active/more serious sites.

- Where would a publisher see the comment in the first place, and be prompted to take action? How would they find the relevant comment in this UI?

- How would a publisher browse the comments, given they will be out of order and mixed in with comments from different posts and threads? We'll probably need some kind of search / filtering, similar to the members list

- What would a publisher's moderation workflow look like? Are they reviewing every comment and giving a 👍/👎, or is it more reactive in response to e.g. emails from members about specific comments that are hateful?

- What does this UI provide that the existing "Hide comment" functionality in the comments UI itself doesn't (or couldn't) provide? 

- Is this something that all Ghost publishers would use? 50%? 20%? 5%?

- If I see a comment in this list that is suspect, but I want to see the full context to decide whether to hide/delete or not, how can I do that? Going to the post and searching for it feels clumsy, especially for active sites with lots of comments to sort through. 

---

With all the above, this feels too big to me to just merge in a single go — not only for technical reasons, but also for the sake of validating that this is the solution that users want. Personally I'd feel a lot better about putting it behind a flag, running a beta, and getting user feedback on it and iterating before shipping it to everyone, which would also let us break up the code changes into more digestible chunks.

PR #25347, opened 2025-11-05
Finalized topic design with Following as default in ActivityPub

minimaluminium:
ref https://linear.app/ghost/issue/BER-2962/finalise-design-for-topic-based-feeds
ref https://linear.app/ghost/issue/BER-2959/clicking-on-a-topic-that-doesnt-exist-leads-to-generic-error-loading

- Changed default topic view from "Top" to "Following" 
- Added empty state UI for topics with no content
- Unified topic navigation by placing "Following" at the same level as other topics instead of as a separate top-level section

---

> [!NOTE]
> Replace feed mode with a topic filter (default "following"), expand topic list and UI, add empty states, tweak header/layout, and update discovery feed API signature.
> 
> - **Reader/Inbox**:
>   - Replace `feedMode` with topic-based navigation via `TopicFilter`; default to `"following"`.
>   - Enable discovery feed when topic ≠ `"following"`; use inbox when `"following"`.
>   - Handle unknown topics (404) by showing an empty state; use `EmptyIndicator` for zero-content views.
> - **Topics UI**:
>   - Expand topics list and implement compact, scrollable chips with gradient fade.
> - **Header/Layout/Sidebar**:
>   - Add optional header border (`showBorder`), hidden on `reader` when `global-feed` enabled; minor layout min-width and sidebar spacing tweaks.
> - **API/Hooks**:
>   - Change `ActivityPubAPI.getDiscoveryFeed` signature to `(topic, next?)` and update callers in `use-activity-pub-queries`.
>   - Remove `use-feed-mode` hook and `FeedModeDropdown` component.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2c2901e9c099600b4b38b472d37d571c9c85bf6c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/activitypub/src/api/activitypub.ts
```diff
@@ -450,7 +450,7 @@ export class ActivityPubAPI {
         return this.getPaginatedPosts('.ghost/activitypub/v1/feed/reader', next);
     }
 
-    async getDiscoveryFeed(next?: string, topic: string = 'top'): Promise<PaginatedPostsResponse> {
+    async getDiscoveryFeed(next?: string, topic: string = 'following'): Promise<PaginatedPostsResponse> {
```
sagzy: There is no `/feed/discover/following` API endpoint. My suggestion would be to remove the default value here for `topic` and let the app error if this method is called without a topic. 

Similar to e.g. `getPostsByAccount` below and the mandatory `handle` param:
```
    async getPostsByAccount(handle: string, next?: string): Promise<PaginatedPostsResponse> {
        return this.getPaginatedPosts(`.ghost/activitypub/v1/posts/${handle}`, next);
    }
```

sagzy (approved):

PR #25350, opened 2025-11-05
Add prettier config and root level dependency

jonatansberg:
Adding a prettier config and a pinned dependency will allow use of editor formatting features that rely on prettier without offending ESLint as badly.

I've tried to adapt the Prettier config to the formatting imposed by ESLint. There might still be some subtle differences. Running Prettier first and ESLint second helps with resolving those.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "nx": "20.8.0",
     "patch-package": "8.0.1",
     "postinstall-postinstall": "2.1.0",
+    "prettier": "3.6.2",
     "rimraf": "5.0.10",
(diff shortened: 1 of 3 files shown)


cmraible:
note: I've never had issues with ESLint format on save in VSCode, but with this change I'm now getting additional spaces added before/after any `{}`, which is weird. Installing the official Prettier extension in VSCode fixed this for me.

suggestion: maybe we could add a `.vscode/extensions.json` file to the repo with the Prettier and ESLint extensions included as recommendations?

PR #25354, opened 2025-11-05
Added typecheck script and TypeScript configuration improvements

rob-ghost:

We wanted to add an independent `typecheck` script to validate TypeScript types in the `apps/admin` package and all new React code being added as part of the Ember → React migration.

### The Problem
Initially, we attempted to use `tsc --noEmit` for type checking, but this doesn't work with TypeScript project references. The `apps/admin` package uses project references (composite TypeScript configuration) to depend on other packages:
- `apps/admin-x-framework`
- `apps/posts`
- `apps/stats`
- `apps/activitypub`

With project references, TypeScript needs to emit `.d.ts` declaration files from referenced projects so the consuming project can type-check against them. This means we must use `tsc -b` (build mode) instead of `tsc --noEmit`.

### The Side Effect
When running `tsc -b`, TypeScript builds all the referenced dependencies and generates their declaration files into `types/` directories (as configured in each package's `tsconfig.declaration.json`). These generated files were appearing as untracked in git.

We found an established pattern in the codebase where packages like `admin-x-framework`, `admin-x-design-system`, and `shade` already have `types/` in their `.gitignore`. This PR extends that pattern to the remaining dependencies.

## What does this PR do?

1. **Adds `.gitignore` entries** for generated `types/` directories in:
   - `apps/activitypub`
   - `apps/posts`
   - `apps/stats`

2. **Adds `typecheck` npm script** to `apps/admin/package.json`:
   ```json
   "typecheck": "tsc -b"
   ```

3. **Configures TypeScript path aliases** in `apps/admin/tsconfig.app.json`:
   - `@/*` → `./src/*`
   - `@test-utils/*` → `./test-utils/*`
   
4. **Updates TypeScript includes** to include the `test-utils` directory

---

> [!NOTE]
> Adds a `typecheck` script, configures TS path aliases and includes in admin, and ignores generated `types/` in related apps.
> 
> - **TypeScript & Build**:
>   - Add `typecheck` script (`tsc -b`) in `apps/admin/package.json`.
>   - Configure TS path aliases in `apps/admin/tsconfig.app.json` (`@/*`, `@test-utils/*`), include `test-utils`, and add reference to `../admin-x-framework/tsconfig.declaration.json`.
> - **Repo hygiene**:
>   - Ignore generated `types/` in `.gitignore` for `apps/activitypub`, `apps/posts`, and `apps/stats`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2d498b15cb20da259e0161712748f720e5570f5f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/admin/package.json
+++ b/apps/admin/package.json
@@ -8,7 +8,8 @@
     "build": "tsc -b && vite build",
     "lint": "eslint .",
     "preview": "vite preview",
-    "test:unit": "vitest run"
+    "test:unit": "vitest run",
+    "typecheck": "tsc -b"
   },
   "dependencies": {
     "@tryghost/activitypub": "*",
--- a/apps/admin/tsconfig.app.json
+++ b/apps/admin/tsconfig.app.json
@@ -15,6 +15,13 @@
     "noEmit": true,
     "jsx": "react-jsx",
 
+    /* Path aliases */
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["./src/*"],
+      "@test-utils/*": ["./test-utils/*"],
+    },
+
     /* Linting */
     "strict": true,
     "noUnusedLocals": true,
@@ -23,7 +30,7 @@
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true
   },
-  "include": ["src"],
+  "include": ["src", "test-utils"],
   "references": [
     { "path": "../posts/tsconfig.declaration.json" },
     { "path": "../stats/tsconfig.declaration.json" },
--- a/apps/activitypub/.gitignore
+++ b/apps/activitypub/.gitignore
@@ -1,3 +1,4 @@
 dist
+types
 playwright-report
 test-results
--- a/apps/posts/.gitignore
+++ b/apps/posts/.gitignore
@@ -1,3 +1,4 @@
 dist
+types
 playwright-report
 test-results
--- a/apps/stats/.gitignore
+++ b/apps/stats/.gitignore
@@ -1,3 +1,4 @@
 dist
+types
 playwright-report
 test-results


apps/admin/package.json
```diff
@@ -8,7 +8,8 @@
     "build": "tsc -b && vite build",
     "lint": "eslint .",
     "preview": "vite preview",
-    "test:unit": "vitest run"
+    "test:unit": "vitest run",
+    "typecheck": "tsc -b"
```
jonatansberg, line 12: How do you feel about calling this `check:types` and then adding a separate `check` command that runs both the linting and the type check (and potentially future formatting checks)?

apps/admin/tsconfig.app.json
```diff
@@ -23,8 +30,9 @@
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true
   },
-  "include": ["src"],
+  "include": ["src", "test-utils"],
```
jonatansberg, line 33: You don't think we can put this in a separate TS config so we don't end up accidentally importing test things within the main app?

apps/admin/tsconfig.app.json
```diff
@@ -23,8 +30,9 @@
...
   "references": [
+    { "path": "../admin-x-framework/tsconfig.declaration.json" },
```
jonatansberg: Do we need to reference this since this package is being built elsewhere? Typically, we should be hitting the existing build output. Adding a reference here would make TypeScript go and build things as well.

jonatansberg (commented):
Overall approach looks sound. I have a few comments and questions I'd love feedback on before merging.

apps/admin/package.json
rob-ghost: Yes, or `check: parallel(format, lint, typecheck)` perhaps?

apps/admin/tsconfig.app.json
rob-ghost: I do think we can, I tried it but unfortunately things got messy, specifically around detecting _when_ to use `tsconfig.test.json` and how to compose that with `tsconfig.app.json` and `tsconfig.node.json` but only under test. For example, `tsc` wouldn't pickup the tests when type-checking without resolving the config.

In the end I walked it back because it was getting too messy. I can try again if we think its worth the investment?

apps/admin/tsconfig.app.json
rob-ghost: I think I got here through editing `useCurrentUser` to introduce a canary type-error which I wanted to be picked up when type-checking the `admin` package.

Given I was importing `useCurrentUser` in `admin`, and I was type-checking `admin`, I wanted it to raise an error without having to build `admin-x-framework` separately.

Does this violate anything regarding how we're building and publishing packages? If `admin` was pinning a non-local version of `admin-x-framework` we'd expect it to type-check against that version and not the local working copy, I imagine?

jonatansberg (approved):
Reviewed and approved in mob session

apps/admin/tsconfig.app.json
rob-ghost: We discussed this in our mob session and decided it was worth figuring out, but now isn't the time to do it so we'll loop back around once the `whats-new` work is complete to implement this.

apps/admin/tsconfig.app.json
rob-ghost: We discussed this in our mob session and decided that we wanted to move away from these `references` and not towards, so in the spirit of that we've removed this `path`.

In order to type-check against `admin-x-framework` we will need to build that package first (using `dev` or `build`) for types to be references from this project.

PR #25356, opened 2025-11-05
Added user preferences hook with test infrastructure

rob-ghost:

We're porting the What's New feature from Ember to React as part of the broader migration effort. This feature needs to store user preferences (like "last seen date") in the user's `accessibility` field.

This PR introduces the foundational test infrastructure for `apps/admin` and implements the user preferences management hooks. By pairing the test infrastructure with its first real usage, we ensure the tooling evolves naturally with actual requirements rather than anticipating future needs.

### Test Infrastructure (Generic, Reusable)
1. **Vitest PyTest-style fixtures**:
   - `serverFixture`: Provides isolated MSW server for each test
   - `queryClientFixtures`: Provides QueryClient + React wrapper with test-optimized defaults
   
2. **Test utilities**:
   - `test-utils/setup.ts`: Global test setup with jest-extended matchers and shade mocks
   - `test-utils/test-helpers.ts`: Helper for waiting on async query state
   - `test-utils/factories/user.ts`: Factory for generating test user data
   - `test-utils/factories/index.ts`: Barrel export for factories

3. **Configuration**:
   - `vite.config.ts`: Configured Vitest with setup file
   - `tsconfig.app.json`: Includes test-utils in TypeScript compilation
   - `package.json`: Added `jest-extended` dependency

### Admin-X Framework Improvements
- **Refactored `useCurrentUser`**: Now uses `select()` to transform data within the query, providing cleaner API (`data: User | undefined` instead of `data: User | null`)
- **Added `UpdateUserRequestBody` type**: Type-safe request body for user mutations
- **Updated `usePermissions` tests**: Updated to match new API shape

### User Preferences Feature
1. **`useUserPreferencesQuery` hook**:
   - Reads preferences from user's `accessibility` JSON field
   - Zod schema validation for type safety
   - Returns empty object for null/undefined/empty accessibility data
   - Errors on invalid JSON or schema violations

2. **`useUpdateUserPreferences` hook**:
   - Merges new preferences with existing ones
   - Updates user's `accessibility` field via admin-x-framework's `useEditUser`
   - Automatically invalidates queries to reflect changes

3. **Type-safe preferences schema**:
   ```typescript
   type Preferences = {
     whatsNew?: {
       lastSeenDate: string; // ISO 8601 datetime
     };
     // ... extensible for future preferences
   };
   ```

4. **Comprehensive test coverage**:
   - Initialization with various edge cases (null, undefined, empty, invalid JSON/schema)
   - Preference merging behavior
   - Error handling

### Dependencies
- **`zod`**: Runtime schema validation for preferences
- **`jest-extended`**: Additional Vitest matchers

## Why is this needed?

- **Type-safe preferences**: Zod validation prevents runtime errors from malformed data
- **Test infrastructure**: Establishes patterns for testing React hooks with API integration
- **Reusable fixtures**: PyTest-style fixtures make tests cleaner and more maintainable
- **Foundation for What's New**: These hooks will be used by the What's New feature in subsequent PRs

## Testing

```bash
yarn workspace @tryghost/admin test:unit
```

---

> [!NOTE]
> Adds user-preferences hooks with Zod validation and full test fixtures, refactors framework `useCurrentUser` to return a typed user via `select`, adds request body typing, and updates tests/config/deps.
> 
> - **Admin (apps/admin)**:
>   - **Hooks**:
>     - `useUserPreferences`: reads/validates preferences from `user.accessibility` (Zod), memoized via query key incl. `user.id` + `accessibility`.
>     - `useEditUserPreferences`: merges and persists preferences via `useEditUser`.
>   - **Tests/Fixtures**:
>     - MSW + React Query fixtures (`test-utils/fixtures/*`), factories (`mockUser`), helpers, and comprehensive hook tests.
>   - **Config/Deps**:
>     - Vitest setup file, TS path aliases include `test-utils`, add `zod` and `jest-extended`.
> - **Admin-X Framework**:
>   - **`useCurrentUser`**: centralize URL/query key, type `fetchApi<UsersResponseType>`, use `select` to return `users[0]` directly, keep error handling.
>   - **Users API**: add `UpdateUserRequestBody` type.
>   - **Tests**: update `usePermissions` tests to expect `data: undefined` for not-loaded user.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a0ec3ca3857168c4cf5deb532357a44643594fde. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/admin/src/hooks/user-preferences.ts
```diff
@@ -0,0 +1,80 @@
...
+            const parsed = JSON.parse(raw) as unknown;
+
+            return PreferencesSchema.parse(parsed);
+        },
+        enabled: !!user,
+        staleTime: Infinity,
+    });
+};
+
+export const useUpdateUserPreferences = (): UseMutationResult<
```
kevinansfield: Looking at the rest of the codebase we use `useEdit*` everywhere, would it make sense to rename this to match?

apps/admin-x-framework/src/api/currentUser.ts
```diff
@@ -1,32 +1,31 @@
...
             handleError(result.error);
         }
     }, [handleError, result.error]);
 
-    return {
-        ...result,
-        data
-    };
+    return result;
 };
```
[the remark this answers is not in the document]
rob-ghost: I don't think this is accurate because we spread `...result` into the return type before, which is the same as returning `result` now. I had the same feedback previously on a review and I looked into it and found the signature change to not be an issue.

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -0,0 +1,80 @@
...
+    return useMutation({
+        mutationFn: async (updatedPreferences: Preferences) => {
+            if (!user) {
+                throw new Error("User is not loaded");
+            }
+
+            const currentPreferences =
+                queryClient.getQueryData<Preferences>(
+                    userPreferencesQueryKey(user)
+                ) ?? {};
```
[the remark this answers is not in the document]
rob-ghost: This might be a legitimate point, we should invalidate the old `user?.accessibility` key when it mutates. This probably is a `invalidateQueries` on `['userPreferences', user?.id]` which will invalidate all cache entries for a given user id and then we expect the `useQuery` to re-populate it.

apps/admin/test-utils/factories/index.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Factory functions for creating test data.
+ *
+ * Import from here:
+ *   import { mockUser, createMockUser } from "@test-utils/factories";
+ */
+
+export { createMockUser, mockUser } from "./user";
```
jonatansberg, line 8: Do we really want to introduce our own set of factories for test data, seeing how this seemingly duplicates what we're also doing with the new E2E tests? A minimal mock containing the accessibility field and an ID ought to be enough for the tests in this case?

apps/admin/src/hooks/user-preferences.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+// Type definitions
+type SetupUserPreferencesTest = (
+    userOverrides?: Partial<User>
+) => Promise<{ current: ReturnType<typeof useUserPreferencesQuery> }>;
+
+type SetupUpdatePreferencesTest = (initialUser?: Partial<User>) => Promise<{
+    query: { current: ReturnType<typeof useUserPreferencesQuery> };
+    mutation: { current: ReturnType<typeof useUpdateUserPreferences> };
+}>;
```
jonatansberg: Do we need to declare these types? Can't we infer them?

apps/admin/src/hooks/user-preferences.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+                                ...initialUser,
+                            },
+                        ],
+                    });
+                }),
+                http.put<
+                    { id: string },
+                    UpdateUserRequestBody,
+                    UsersResponseType
+                >(USER_UPDATE_API_URL, async ({ request }) => {
```
jonatansberg: Maybe we can manually reformat this as the narrow print width makes the code hard to read.

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -0,0 +1,80 @@
...
+        whatsNew: WhatsNewPreferencesSchema.optional(),
+    })
+    .passthrough();
+
+export type Preferences = z.infer<typeof PreferencesSchema>;
+
+const userPreferencesQueryKey = (user: User | undefined) =>
+    ["userPreferences", user?.id, user?.accessibility] as const;
+
+export const useUserPreferencesQuery = (): UseQueryResult<Preferences> => {
```
kevinansfield: Similar naming question here, do we need the `Query` suffix? It seems like a departure from our existing hooks. I'm not against it because it does give a clue about the return type but if it's a new naming pattern we want to introduce I think we should be intentional about it.

apps/admin/src/hooks/user-preferences.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+}>({
+    ...serverFixture,
+    ...queryClientFixtures,
+    setup: async (
+        {
+            server,
+            wrapper,
+        }: { server: SetupServer; wrapper: TestWrapperComponent },
+        provide: (setup: SetupUserPreferencesTest) => Promise<void>
+    ) => {
```
jonatansberg: This does more than what I would expect something named `setup` to do. It's both arranging and acting. 

Since the tests are already parameterized, moving the acting to the tests themselves shouldn't result in too much duplication and would be cleaner in my eyes. 

Even if we keep this as is, the code is fairly dense. Some inline comments explaining what we're doing could help parse it.

apps/admin/src/hooks/user-preferences.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+            },
+        ].forEach(({ scenario, accessibility }) => {
+            test(`returns empty object when accessibility is ${scenario}`, async ({
+                setup,
+            }) => {
+                const result = await setup({ accessibility });
+
+                expect(result.current.data).toEqual({});
+            });
+        });
```
jonatansberg, line 186: This looks very much like we're testing the Zod schema. Is that valuable?

jonatansberg (commented):
The actual implementation looks good. I have some input on the testing. I've also commented on some of the type stuff in the separate PR.

apps/admin/src/hooks/user-preferences.ts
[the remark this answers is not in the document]
rob-ghost: I attempted to fix this by adding `cacheTime: 0` to immediately invalidate orphaned cache entries.

apps/admin/test-utils/factories/index.ts
rob-ghost: `useEditUser` requires a whole `User` instance to type-check / work.

apps/admin/src/hooks/user-preferences.test.tsx
rob-ghost: Mainly for legibility, pulling the type definitions out to these aliases made the fixture code much easier to follow (assuming you're referring to `SetupUserPreferencesTest/SetupUpdatePreferencesTest`.

Or do you mean removing the types from `setup` entirely and having everything be inferred? I don't think that works with our current config, we require the definitions to have explicit types.

Can you illustrate what you have in mind?

apps/admin/src/hooks/user-preferences.test.tsx
rob-ghost: Agreed, I'm guessing prettier was working with default 80 I'll reformat everything with 100 and hopefully that improves the legibility.

apps/admin/src/hooks/user-preferences.test.tsx
rob-ghost: For me I found `renderHook` and `waitFor` in every test distracting and it was harder to distinguishing between test setup and expectation.

Given the arrange + act was common to pretty much every test (outside of some exceptions) I preferred to have this fixture take care of everything.

I originally had it as a helper function but that meant passing the `server, wrapper` fixtures in every test, having it a fixture allows us to partially apply those.

I agree its very dense so I'll try to make it more legible with formatting, type aliases where possible and comments.

apps/admin/src/hooks/user-preferences.test.tsx
rob-ghost: This is actually not even getting to `zod`, each of these cases is handled by the `user.accessibility || '{}'` before we even get to `JSON.parse`.

This suite is mainly to ensure that, no matter what gets fed into the hook, we're getting an object back that we can destruct on the client side.

apps/admin/src/hooks/user-preferences.test.tsx
rob-ghost: I pulled the setup functions out and also created two distinct fixtures (one for query, one for test) and added some comments to explain the "why" of these fixtures and what they're doing. I hope this is more legible.

It had a side benefit of removing a lot of the type declarations because we could use inference more often.

apps/admin/src/hooks/user-preferences.test.tsx
rob-ghost: Reformatted the whole PR with some of the config you had in your [prettier PR](https://github.com/TryGhost/Ghost/pull/25350/files), I hope this makes things more legible.

apps/admin/src/hooks/user-preferences.test.tsx
rob-ghost: These disappeared with the cleanup of the `setup` fixture functions

apps/admin/test-utils/fixtures/query-client.tsx
```diff
@@ -0,0 +1,95 @@
...
+                retry: false,
+            },
+        },
+        logger: {
+            log: () => {},
+            warn: () => {},
+            error: () => {},
+        },
+    });
+}
```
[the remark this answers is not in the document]
rob-ghost: Duplication is intentional

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -0,0 +1,72 @@
+import { useQuery, useMutation, type UseQueryResult, type UseMutationResult } from "@tanstack/react-query";
+import { z } from "zod";
+
+import { useQueryClient } from "@tryghost/admin-x-framework";
+import { useCurrentUser } from "@tryghost/admin-x-framework/api/currentUser";
+import { useEditUser, type User } from "@tryghost/admin-x-framework/api/users";
+
+const WhatsNewPreferencesSchema = z.object({
+    lastSeenDate: z.iso.datetime(),
+});
```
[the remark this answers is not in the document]
rob-ghost: I added `iso` based on your suggestion and tests pass so 🤷🏻

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const PreferencesSchema = z
+    .object({
+        whatsNew: WhatsNewPreferencesSchema.optional(),
+    })
+    .passthrough();
+
+export type Preferences = z.infer<typeof PreferencesSchema>;
+
+const userPreferencesQueryKey = (user: User | undefined) => ["userPreferences", user?.id, user?.accessibility] as const;
```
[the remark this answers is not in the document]
rob-ghost: I don't know how many rounds we have to go through this issue 🙈

We need to react to any changes to `user.accessibility`, not just from `useEditUserPreferences`. Having `accessibility` in the key is the best way to do this.

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const WhatsNewPreferencesSchema = z.object({
+    lastSeenDate: z.iso.datetime(),
+});
+
+const PreferencesSchema = z
+    .object({
+        whatsNew: WhatsNewPreferencesSchema.optional(),
+    })
+    .passthrough();
```
jonatansberg: Nit: `passthrough()` has been deprecated in Zod 4

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -0,0 +1,72 @@
...
+            return PreferencesSchema.parse(parsed);
+        },
+        enabled: !!user,
+        staleTime: Infinity,
+        // Query key includes user?.accessibility to automatically react to changes from ANY source
+        // (our mutation, other code calling editUser, external updates, etc.). When accessibility
+        // changes, the query key changes, making the old cache entry inactive. cacheTime: 0 ensures
+        // orphaned entries are immediately garbage collected, preventing memory leaks while keeping
+        // the current active entry cached indefinitely.
+        cacheTime: 0,
```
jonatansberg, line 42: We might want to add our own abstraction around this that tucks away the oddities of staleTime and cacheTime for these local-only queries if we see ourselves repeating this pattern.

apps/admin-x-framework/test/unit/hooks/usePermissions.test.ts
```diff
@@ -74,7 +77,8 @@ describe('usePermissions', () => {
             }
         });
 
-        const {result} = renderHook(() => usePermission(['admin', 'editor', 'owner']));
+        const {result} = renderHook(() => usePermission(['admin', 'editor', 'owner'])
+        );
```
jonatansberg: These formatting changes look odd?

jonatansberg (approved):
This looks good. Commented on some minor nitpicks you might want to look at before merging.

apps/admin/src/hooks/user-preferences.ts
rob-ghost: Something like:
```typescript
useLiveQuery(options: Omit<UseQueryOptions, 'staleTime' | 'cacheTime'>) =>
    useQuery({...options, staleTime: Infinity, cacheTime: 0})
```

Definitely something to consider if we see this pattern repeat.

apps/admin-x-framework/test/unit/hooks/usePermissions.test.ts
rob-ghost: Yep since this file changed to align with the new `useCurrentUser` return type (`data: User | undefined`) it got caught up with the reformat. I'll undo the unrelated formatting.

rob-ghost: (I don't know why it decided to put the `);` on a newline, that is odd)

PR #25357, opened 2025-11-05
Added changelog fetching hook

rob-ghost:
ref https://linear.app/ghost/issue/BER-2610

## Why are we making this change?

The What's New feature needs to fetch and display Ghost changelog entries from `https://ghost.org/changelog.json`. This PR implements the changelog fetching logic as part of porting the What's New feature from Ember to React.

This PR builds on the test infrastructure established in #25356.

### Changelog Hook
1. **`useChangelog` hook**:
   - Fetches changelog data from `https://ghost.org/changelog.json`
   - Returns typed `ChangelogEntry[]` and `changelogUrl` string
   - Uses Zod for runtime schema validation
   - Gracefully handles missing/empty data with sensible defaults
   
2. **Type-safe changelog schema**:
   ```typescript
   type ChangelogEntry = {
     id: string;
     title: string;
     excerpt: string;
     url: string;
     publishedAt: Date;
     featured: boolean;
   };
   
   type ChangelogResponse = {
     entries: ChangelogEntry[];
     changelogUrl: string;
   };
   ```

3. **Robust error handling**:
   - HTTP errors (404, 500, etc.)
   - Network failures
   - Invalid JSON responses
   - Schema validation errors
   - All errors include helpful messages

### Test Infrastructure Addition
- **Changelog factory** (`test-utils/factories/changelog.ts`):
  - `createMockChangelogEntry()`: Creates typed changelog entries
  - `createRawChangelogEntry()`: Creates raw API response format
  - `createChangelogResponse()`: Creates full API response
  - `changelogFixtures`: Pre-configured test data for common scenarios

### Comprehensive Test Coverage
Tests cover:
- ✅ Successful data fetching and deserialization
- ✅ Missing/empty fields (posts, changelogUrl)
- ✅ HTTP errors (404, 500)
- ✅ Network failures
- ✅ Invalid JSON responses
- ✅ Schema validation errors (invalid URLs, dates, etc.)

## Why is this needed?

- **Decoupled from Ember**: React hook can be used independently
- **Type safety**: Zod validation prevents runtime errors from malformed API responses
- **Reusable**: Other features can use this hook to display changelog data
- **Foundation for What's New**: The What's New hook will combine this with user preferences

## Testing

```bash
yarn workspace @tryghost/admin test:unit src/whats-new/hooks/use-changelog.test.tsx
```

The tests use MSW to mock the Ghost changelog API and validate all edge cases.

## Dependencies

Depends on:
- #25356 (test infrastructure + user preferences)

PR #25358, opened 2025-11-05
Added What's New orchestration hook

rob-ghost:

This PR completes the React implementation of the What's New feature by orchestrating the changelog data and user preferences to provide notification state. The core challenge is determining whether the current user has seen the latest changelog entry, and providing a simple API for UI components to display notifications and mark entries as seen.

This is the final piece in the What's New feature port from Ember to React.

## What does this PR do?

This PR introduces two hooks: `useWhatsNew` and `useDismissWhatsNew`.

The query hook combines the changelog data with the user's stored `lastSeenDate` preference to compute whether there are new entries. Rather than checking every changelog entry, we only need to compare the latest entry's publication date with when the user last viewed the changelog. This keeps the logic simple and efficient.

The hook also distinguishes between regular new entries and featured entries, since featured entries deserve special UI treatment (like more prominent badges or notifications). This distinction is computed by checking both the "new" status and the entry's featured flag.

For first-time users who have never opened What's New before, the hook automatically initializes their preferences with the current date. This prevents existing changelog entries from appearing as "new" when someone opens the feature for the first time.

The mutation hook provides a simple way to dismiss What's New notifications. When called, it updates the user's `lastSeenDate` to match the latest changelog entry's publication date, which automatically causes the query hook to return `hasNew: false` on subsequent renders.

This PR also refactors the user preferences system to use Zod codecs with explicit encode/decode functions. This allows the internal representation to use native `Date` objects for type safety, while automatically converting to/from ISO datetime strings for storage. This makes working with dates in the preferences system more ergonomic and less error-prone.

## Why is this needed?

The What's New feature needs a centralized source of truth for notification state. UI components shouldn't need to understand the relationship between changelog data and user preferences, or how to compute "new" status. By encapsulating this logic in hooks, we make it easy for multiple components to consistently show or hide notification badges, and we ensure the feature behaves correctly for both new and returning users.

The separate `hasNewFeatured` flag lets us create better UX by highlighting particularly important updates differently from routine changelog entries.

## Testing

```bash
yarn workspace @tryghost/admin test:unit src/whats-new/hooks/use-whats-new.test.tsx
```

Tests cover new user initialization, computing new entry status, the dismiss mutation, and handling edge cases like empty changelogs.

## Dependencies

Depends on:
- #25356 (test infrastructure + user preferences)
- #25357 (changelog hook)

### DIFF
--- a/apps/admin/src/hooks/user-preferences.ts
+++ b/apps/admin/src/hooks/user-preferences.ts
@@ -5,16 +5,22 @@ import { useQueryClient } from "@tryghost/admin-x-framework";
 import { useCurrentUser } from "@tryghost/admin-x-framework/api/currentUser";
 import { useEditUser, type User } from "@tryghost/admin-x-framework/api/users";
 
-const WhatsNewPreferencesSchema = z.object({
-    lastSeenDate: z.iso.datetime(),
+const isoDatetimeToDate = z.codec(z.iso.datetime(), z.date(), {
+    decode: (isoString) => new Date(isoString),
+    encode: (date) => date.toISOString(),
+});
+
+const WhatsNewPreferencesSchema = z.looseObject({
+    lastSeenDate: isoDatetimeToDate.optional().catch(undefined),
 });
 
 const PreferencesSchema = z.looseObject({
-    whatsNew: WhatsNewPreferencesSchema.optional(),
+    whatsNew: WhatsNewPreferencesSchema.optional().catch(undefined),
     nightShift: z.boolean().optional(),
 });
 
 export type Preferences = z.infer<typeof PreferencesSchema>;
+export type WhatsNewPreferences = z.infer<typeof WhatsNewPreferencesSchema>;
 
 const userPreferencesQueryKey = (user: User | undefined) => ["userPreferences", user?.id, user?.accessibility] as const;
 
@@ -57,14 +63,16 @@ export const useEditUserPreferences = (): UseMutationResult<void, Error, Prefere
 
             const currentPreferences = queryClient.getQueryData<Preferences>(userPreferencesQueryKey(user)) ?? {};
 
-            const newPreferences = {
+            const newPreferences: Preferences = {
                 ...currentPreferences,
                 ...updatedPreferences,
             };
 
+            const encodedForStorage = PreferencesSchema.encode(newPreferences);
+
             await editUser({
                 ...user,
-                accessibility: JSON.stringify(newPreferences),
+                accessibility: JSON.stringify(encodedForStorage),
             });
         },
     });
--- a/apps/admin/src/whats-new/hooks/use-whats-new.test.tsx
+++ b/apps/admin/src/whats-new/hooks/use-whats-new.test.tsx
@@ -1,0 +1,426 @@
+import { test as baseTest, describe, expect } from "vitest";
+import { renderHook, waitFor, act } from "@testing-library/react";
+import type { QueryClient } from "@tanstack/react-query";
+import { useWhatsNew, useDismissWhatsNew } from "./use-whats-new";
+import { HttpResponse, http } from "msw";
+import { mockUser, createRawChangelogEntry } from "@test-utils/factories";
+import { waitForQuerySettled } from "@test-utils/test-helpers";
+import type { UpdateUserRequestBody, UsersResponseType } from "@tryghost/admin-x-framework/api/users";
+import type { RawChangelogResponse, RawChangelogEntry } from "./use-changelog";
+import { serverFixture } from "@test-utils/fixtures/msw";
+import { queryClientFixtures, type TestWrapperComponent } from "@test-utils/fixtures/query-client";
+import type { SetupServer } from "msw/node";
+
+// Constants
+const USERS_API_URL = "/ghost/api/admin/users/me/";
+const USER_UPDATE_API_URL = "/ghost/api/admin/users/:id/";
+const CHANGELOG_API_URL = "https://ghost.org/changelog.json";
+
+// Types
+interface SetupQueryOptions {
+    accessibility?: string | null;
+    changelog?: Partial<RawChangelogResponse>;
+}
+
+interface SetupMutationOptions {
+    lastSeenDate?: string;
+    posts?: RawChangelogEntry[];
+}
+
+type SetupQueryTest = (options?: SetupQueryOptions) => ReturnType<typeof setupQuery>;
+type SetupMutationTest = (options?: SetupMutationOptions) => ReturnType<typeof setupMutation>;
+
+// Test fixtures
+const dates = {
+    past: "2025-01-01T00:00:00Z",
+    recent: "2025-01-10T00:00:00Z",
+    current: "2025-01-15T10:00:00Z",
+    future: "2025-01-20T10:00:00Z",
+};
+
+const fixtures = {
+    entries: {
+        newEntry: () =>
+            createRawChangelogEntry({
+                published_at: dates.current,
+            }),
+        oldEntry: () =>
+            createRawChangelogEntry({
+                published_at: dates.past,
+            }),
+        featuredEntry: () =>
+            createRawChangelogEntry({
+                published_at: dates.current,
+                featured: "true",
+            }),
+    },
+    preferences: {
+        empty: {},
+        withOtherSetting: { otherSetting: "value" },
+        withLastSeen: (date: string) => ({
+            whatsNew: { lastSeenDate: date },
+        }),
+    },
+};
+
+// Setup functions
+/**
+ * Setup function for testing `useWhatsNew`.
+ *
+ * This fixture handles the boilerplate of:
+ * 1. Mocking the user API endpoint with customizable accessibility preferences
+ * 2. Mocking the user mutation endpoint (for initializing preferences)
+ * 3. Mocking the changelog API endpoint with customizable changelog data
+ * 4. Rendering the hook with the necessary React Query wrapper
+ * 5. Waiting for the query to settle (success or error state)
+ *
+ * Note: The query is disabled until preferences exist (hasWhatsNewPreferences = true).
+ * This means the query naturally stays in loading state during initialization, then
+ * enables and settles once the mutation completes. Tests only need to wait once.
+ *
+ * This allows tests to focus on asserting behavior rather than setup logic,
+ * making test code more ergonomic and readable.
+ *
+ * @param server - MSW server instance for mocking API endpoints
+ * @param wrapper - React Query wrapper component for hook testing
+ * @param options - Test configuration options (accessibility, changelog data)
+ * @returns The renderHook result with the query in a settled state
+ */
+async function setupQuery(server: SetupServer, wrapper: TestWrapperComponent, options: SetupQueryOptions = {}) {
+    const { accessibility = null, changelog = {} } = options;
+
+    server.use(
+        // Mock the user preferences endpoint
+        http.get(USERS_API_URL, () => {
+            return HttpResponse.json({
+                users: [
+                    {
+                        ...mockUser,
+                        accessibility: accessibility ?? mockUser.accessibility,
+                    },
+                ],
+            });
+        }),
+        // Mock the user mutation endpoint (for initializing preferences)
+        http.put<{ id: string }, UpdateUserRequestBody, UsersResponseType>(USER_UPDATE_API_URL, async ({ request }) => {
+            const body = await request.json();
+            const newAccessibility = body.users[0]?.accessibility;
+            return HttpResponse.json({
+                users: [
+                    {
+                        ...mockUser,
+                        accessibility: newAccessibility,
+                    },
+                ],
+            });
+        }),
+        // Mock the changelog endpoint
+        http.get(CHANGELOG_API_URL, () => {
+            return HttpResponse.json({
+                posts: changelog.posts ?? [],
+                changelogUrl: changelog.changelogUrl ?? "https://ghost.org/changelog",
+            });
+        })
+    );
+
+    const { result } = renderHook(() => useWhatsNew(), {
+        wrapper,
+    });
+
+    // Wait for query to settle. The query is disabled until preferences are initialized,
+    // so it stays in loading state during the mutation, then enables and settles naturally.
+    await waitForQuerySettled(result);
+
+    return result;
+}
+
+/**
+ * Setup function for testing `useDismissWhatsNew`.
+ *
+ * This fixture handles the boilerplate of:
+ * 1. Mocking the user API endpoint with customizable lastSeenDate
+ * 2. Mocking the user mutation endpoint (for updating preferences)
+ * 3. Mocking the changelog API endpoint with customizable posts
+ * 4. Rendering both the query hook (to read state) and mutation hook (to update state)
+ * 5. Waiting for the initial query to settle before tests run
+ *
+ * This allows mutation tests to immediately call `mutation.mutateAsync()` without worrying
+ * about setup, making test code more ergonomic and focused on the mutation behavior.
+ *
+ * @param server - MSW server instance for mocking API endpoints
+ * @param wrapper - React Query wrapper component for hook testing
+ * @param options - Test configuration options (lastSeenDate, posts)
+ * @returns Both query and mutation hook results, ready for testing mutations
+ */
+async function setupMutation(server: SetupServer, wrapper: TestWrapperComponent, options: SetupMutationOptions = {}) {
+    const { lastSeenDate = dates.past, posts = [fixtures.entries.newEntry()] } = options;
+
+    server.use(
+        // Mock the user preferences endpoint
+        http.get(USERS_API_URL, () => {
+            return HttpResponse.json({
+                users: [
+                    {
+                        ...mockUser,
+                        accessibility: JSON.stringify({
+                            whatsNew: { lastSeenDate },
+                        }),
+                    },
+                ],
+            });
+        }),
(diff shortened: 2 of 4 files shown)


apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const parsedSettings = useMemo(() => parseWhatsNewSettings(preferences), [preferences]);
+
+    const settings = useMemo(() => parsedSettings ?? getDefaultWhatsNewSettings(), [parsedSettings]);
+
+    // Initialize default settings if missing
+    useEffect(() => {
+        if (!parsedSettings && isPreferencesLoaded) {
+            const defaultSettings = getDefaultWhatsNewSettings();
+            void updatePreferences({
```
rob-ghost, line 36: Look into multiple side-effects in this case.

apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+const whatsNewQueryKey = (preferences: Preferences | undefined) => ["whatsNew", preferences?.whatsNew] as const;
+
+export const useWhatsNew = (): UseQueryResult<WhatsNewData> => {
+    const { data: preferences, isSuccess: isPreferencesLoaded } = useUserPreferences();
+    const { data: changelog, isSuccess: isChangelogLoaded } = useChangelog();
+    const { mutateAsync: updatePreferences } = useEditUserPreferences();
+
+    const parsedSettings = useMemo(() => parseWhatsNewSettings(preferences), [preferences]);
```
rob-ghost: Factor out `parseWhatsNewSettings` into a `hasSettings` boolean now that we have Zod on preferences.

apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,101 @@
...
+            const hasNew = latestEntry.publishedAt > settings.lastSeenDate;
+            const hasNewFeatured = hasNew && latestEntry.featured === true;
+
+            return { hasNew, hasNewFeatured };
+        },
+        enabled: isPreferencesLoaded && isChangelogLoaded,
+    });
+};
+
+export const useMarkWhatsNewAsSeen = (): UseMutationResult<void, Error, void, unknown> => {
```
rob-ghost: Think of a better name for this, `useDismissWhatsNew` perhaps.

apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    return {
+        lastSeenDate: new Date(preferences.whatsNew.lastSeenDate),
+    };
+}
+
+function serializeWhatsNewSettings(whatsNewSettings: WhatsNewSettings) {
+    return {
+        lastSeenDate: whatsNewSettings.lastSeenDate.toISOString(),
```
rob-ghost: Use [Zod codecs](https://zod.dev/codecs) for encode / decode of the User Preferences

apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,101 @@
...
+
+function parseWhatsNewSettings(preferences: Preferences | undefined): WhatsNewSettings | null {
+    if (!preferences?.whatsNew?.lastSeenDate) {
+        return null;
+    }
+
+    return {
+        lastSeenDate: new Date(preferences.whatsNew.lastSeenDate),
+    };
+}
```
rob-ghost: Move to preferences Zod schema

apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,101 @@
...
+const whatsNewQueryKey = (preferences: Preferences | undefined) => ["whatsNew", preferences?.whatsNew] as const;
+
+export const useWhatsNew = (): UseQueryResult<WhatsNewData> => {
+    const { data: preferences, isSuccess: isPreferencesLoaded } = useUserPreferences();
+    const { data: changelog, isSuccess: isChangelogLoaded } = useChangelog();
+    const { mutateAsync: updatePreferences } = useEditUserPreferences();
+
+    const parsedSettings = useMemo(() => parseWhatsNewSettings(preferences), [preferences]);
+
+    const settings = useMemo(() => parsedSettings ?? getDefaultWhatsNewSettings(), [parsedSettings]);
```
rob-ghost: Check if we need to `getDefaultWhatsNewSettings` if we're doing also in the side-effect.

apps/admin/src/whats-new/hooks/use-whats-new.test.tsx
```diff
@@ -0,0 +1,405 @@
...
+        http.get(USERS_API_URL, () => {
+            return HttpResponse.json({
+                users: [
+                    {
+                        ...mockUser,
+                        accessibility: accessibility ?? mockUser.accessibility,
+                    },
+                ],
+            });
+        }),
```
rob-ghost, line 103: Can we compose this from the existing user preferences tests?

apps/admin/src/whats-new/hooks/use-whats-new.test.tsx
```diff
@@ -0,0 +1,405 @@
...
+        // Mock the changelog endpoint
+        http.get(CHANGELOG_API_URL, () => {
+            return HttpResponse.json({
+                posts: changelog.posts ?? [],
+                changelogUrl: changelog.changelogUrl ?? "https://ghost.org/changelog",
+            });
+        })
```
rob-ghost, line 123: Can we compose this from the changelog tests?

apps/admin/src/whats-new/hooks/use-whats-new.ts
rob-ghost: I spent a long time on this one only to come to the conclusion that the solutions were too complex to warrant doing. We're better off living with the potential race condition.

For reference, the best solution was to have a `WhatsNewProvider` parent component with the side effect of initialising before rendering children which use `useWhatsNew`.

apps/admin/src/whats-new/hooks/use-whats-new.test.tsx
rob-ghost: After looking into it I decided that leaving the fixtures separate was best to avoid coupling the discrete test cases given their subtle differences.

apps/admin/src/whats-new/hooks/use-whats-new.test.tsx
rob-ghost: After looking into it I decided that leaving the fixtures separate was best to avoid coupling the discrete test cases given their subtle differences.

apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            const latestEntry = entries[0];
+            const hasNew = latestEntry.publishedAt > whatsNewPreferences.lastSeenDate;
+            const hasNewFeatured = hasNew && latestEntry.featured === true;
+
+            return { hasNew, hasNewFeatured };
+        },
+        enabled: isPreferencesLoaded && isChangelogLoaded && hasWhatsNewPreferences,
+    });
```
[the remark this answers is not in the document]
rob-ghost: This wasn't a concern because `changelog` is never invalidated, it has a `staleTime: infinity` and is intended to be fetched only once per app load.

That said, the dependant query should follow better conventions and include some aspect of the changelog in the query key to hint that it _should_ re-compute, even if it _doesn't_ in practical terms.

I've since updated `whatsNewQueryKey` to take an optional `ChangelogEntry` and we'll depend on its `publishedAt` date. If the latest entries `publishedAt` date changes, then we'll invalidate the key and re-compute the query result (though this won't happen in reality).

apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,88 @@
...
+            }
+
+            const hasNew = latestEntry.publishedAt > whatsNewPreferences.lastSeenDate;
+            const hasNewFeatured = hasNew && latestEntry.featured === true;
+
+            return { hasNew, hasNewFeatured };
+        },
+        enabled: isPreferencesLoaded && isChangelogLoaded && hasWhatsNewPreferences,
+        staleTime: Infinity,
+        cacheTime: 0,
```
rob-ghost, line 62: This pattern rears its head again. I opted to put the options here for now, but we can consider encapsulating this in a `useLiveQuery` or `useDependantQuery` which curries these into `useQuery`. We'd have to decide what to call it and where to keep it, but the idea of a query derived from another query and the whole point is to not keep cached values and never go stale makes sense as a re-useable hook.

apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,88 @@
...
+    };
+}
+
+interface WhatsNewData {
+    hasNew: boolean;
+    hasNewFeatured: boolean;
+}
+
+const whatsNewQueryKey = (preferences: Preferences | undefined, latestEntry: ChangelogEntry | undefined) =>
+    ["whatsNew", preferences?.whatsNew?.lastSeenDate.toISOString(), latestEntry?.publishedAt.toISOString()] as const;
```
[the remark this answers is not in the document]
rob-ghost: This is factually incorrect. The optional chain operator `?` will **early out** with `undefined` if any part of the chain is `undefined` / `null`, so we will never get to `. toISOString()`.

This is a pretty fundamental error in your javascript knowledge.

apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export const useWhatsNew = (): UseQueryResult<WhatsNewData> => {
```
rob-ghost, line 26: **Meta question:** Does "whats new" make sense as a name anymore?

`whats-new` refers to the title of the notification banner which is shown to the user and the service responsible for showing this banner was named `whats-new`.

In this new hooks landscape we've decomposed the service into many hooks, and now this `useWhatsNew` hook has inherited the name. However, if we look at the return values, `{ hasNew, hasNewFeatured }` the name `whats-new` no longer makes much sense.

Perhaps we should rename this to:

1. `useUnseenChangelog -> { hasUnseen, hasUnseenFeatured }`
2. `useMarkChangelogSeen`

I think these might be more accurate names, referring to unseen changelog entries and then "seeing" them.

What do we think?

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -5,15 +5,32 @@ import { useQueryClient } from "@tryghost/admin-x-framework";
...
+        decode: (stored) => ({
+            ...stored,
+            whatsNew: stored.whatsNew ? WhatsNewPreferencesCodec.decode(stored.whatsNew) : undefined,
+        }),
+        encode: (working) => ({
+            ...working,
+            whatsNew: working.whatsNew ? WhatsNewPreferencesCodec.encode(working.whatsNew) : undefined,
+        }),
+    }
+);
```
jonatansberg: I think it would make more sense to put these codecs on the field themselves rather than the full schema. That way you don't have to reach in to objects or deal with the other existing fields when encoding or decoding.

```ts
const isoDatetimeToDate = z.codec(z.iso.datetime(), z.date(), {
    decode: (isoString) => new Date(isoString),
    encode: (date) => date.toISOString(),
});

const WhatsNewSchema = z.looseObject({ 
    lastSeenDate: isoDatetimeToDate
})

const PreferencesSchema = z.looseObject({ 
    whatsNew: WhatsNewSchema.optional()
});
```

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -30,7 +47,7 @@ export const useUserPreferences = (): UseQueryResult<Preferences> => {
             const raw = user.accessibility || "{}";
             const parsed = JSON.parse(raw) as unknown;
 
-            return PreferencesSchema.parse(parsed);
+            return PreferencesCodec.parse(parsed);
```
jonatansberg: I don't think we should rename the schema. The codec thing is an implementation level concern of making the transforms bidirectional. It's still a schema.

apps/admin/src/whats-new/hooks/use-whats-new.ts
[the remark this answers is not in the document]
jonatansberg: Couldn't `lastSeenDate` potentially still being null (depending on how you define the schema) which would throw?

apps/admin/src/whats-new/hooks/use-whats-new.ts
jonatansberg: The "What's New" name makes sense for us internally and it's aligned with what is stored in the user preferences. I think it's good enough.

apps/admin/src/whats-new/hooks/use-whats-new.ts
jonatansberg: You could also pull out the side effect into an initialize-what's-new-thing and only called that hook once 🤔 

Otherwise, I had something like this in mind:
https://tkdodo.eu/blog/concurrent-optimistic-updates-in-react-query

Did you look into anything like that?

apps/admin/src/whats-new/hooks/use-whats-new.ts
jonatansberg: Let's tackle this as a follow-up.

jonatansberg (changes requested):
I think we might want to make some changes to the schema. Otherwise, this looks good.

apps/admin/src/whats-new/hooks/use-whats-new.ts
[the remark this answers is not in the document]
rob-ghost: The schema claims it not to be null so either `whatsNew` is defined with a `lastSeenDate` or its not, there's no in-between.

IIRC, that's true of the existing data for users also based on how the current `whats-new` service is implemented.

There is a risk that there's some Users out there with invalid preferences and/or whats new preferences, and do we want to be accepting of those (and possibly sanitise them) or do we want to raise exceptions. How permissive do we want to be with this data?

Since we moved `whatsNew` validation into `user-preferences` its a little harder to have error handling / fallback behaviour for individual preference keys, either all the preferences are valid, or none of them are. I'll look into Zod supports defaulting a whole sub-schema in the case where the sub-schema fails to validate so we can handle sanitising `whatsNew`.

apps/admin/src/hooks/user-preferences.ts
rob-ghost: Much cleaner 👍🏻

apps/admin/src/whats-new/hooks/use-whats-new.ts
rob-ghost: > You could also pull out the side effect into an initialize-what's-new-thing and only called that hook once 🤔

Can you give an example of what this looks like?

> Did you look into anything like that?

Yep one of the attempts used `cancelQueries` and `isMutating` but ultimately it was pretty complex to setup and test compared to accepting the race condition so much like the Provider solution, I decided against it. I could put together an example diff if you'd like to see the options before we accept the solution?

apps/admin/src/whats-new/hooks/use-whats-new.ts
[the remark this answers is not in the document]
rob-ghost: We will get loose with our user preferences schema, and add some logging to validation errors.

apps/admin/src/whats-new/hooks/use-whats-new.ts
rob-ghost: TODO after merge; check both the module-level variable lock and also lifting the initialisation into a separate hook which is run from `<App>` (or some other kind of top level context/component).

apps/admin/src/whats-new/hooks/use-whats-new.ts
```diff
@@ -0,0 +1,86 @@
...
+
+            // Safe to assert non-null because query is only enabled when hasWhatsNewPreferences is true
+            const whatsNewPreferences = preferences!.whatsNew!;
+
+            const hasNew = latestEntry.publishedAt > whatsNewPreferences.lastSeenDate;
+            const hasNewFeatured = hasNew && latestEntry.featured === true;
+
+            return { hasNew, hasNewFeatured };
+        },
+        enabled: isPreferencesLoaded && isChangelogLoaded && hasWhatsNewPreferences,
```
rob-ghost: `hasWhatsNewPreferences` makes `isPreferencesLoaded` redundant.

apps/admin/src/whats-new/hooks/use-whats-new.ts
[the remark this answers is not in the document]
rob-ghost: I made the schema members optional and also permissive of validation errors through `catch(undefined)`.

I tried to add logging but I didn't like mixing in `console.error` into each `catch` call, and even with a custom `catchAndLog` helper it would still be mixing concerns I think.

I tried `safeParse` which works but it meant we'd have to fallback to an empty object for all preferences which felt too nuclear just to get a log whenever we have potentially invalid preferences.

I felt catching individually met most of the requirements, it is permissive with defaults defined within the schema and doesn't mix parsing behaviour concerns.

I updated the rest of the code to handle the optional cases for all whats new preference fields.

jonatansberg (approved):
Looks good! There's a merge conflict that need resolving.

PR #25360, opened 2025-11-05
🐛 Fixed magic links only being valid for 10 minutes

kevinansfield:
ref https://linear.app/ghost/issue/BER-2977/

- during a refactor of `SingleUseTokenProvider` the time-since-first-usage check was extracted to a method but in doing so the pre-condition check for the token having been used was lost meaning all magic links had an effective lifetime of `validityAfterFirstUsage` (10 minutes)
- restored pre-condition check and added missing test that would have caught this during the refactor


9larsons (approved):
Looks great, thanks for the details.

PR #25375, opened 2025-11-06
Integrate settings app with react admin shell

jonatansberg:
**Why are we making this change?**
The settings app needs to be fully integrated with the new React-based admin shell to continue the migration away from the Ember admin.

**What does it do?**
- Adds routes and entry points for the settings app
- Updates Vite config to handle settings app quirks
- Exclude the settings app from Ember when the `inAdminForward` flag is set
- Moves SVG icons to admin-x-design-system to avoid transform conflicts
- Implements global query invalidation when editing settings to ensure ripple effects take effect

**Testing**
I've gone through the different views in the settings app locally to confirm that all of the icons (at least the ones I could find) render the same. 

Closes https://linear.app/ghost/issue/BER-2915/integrate-settings-app-with-admin-shell


apps/admin-x-framework/src/utils/api/hooks.ts
```diff
@@ -150,7 +150,7 @@ interface MutationOptions<ResponseData, Payload> extends Omit<QueryOptions<Respo
     headers?: Record<string, string>;
     body?: (payload: Payload) => FormData | object;
     searchParams?: (payload: Payload) => { [key: string]: string; };
-    invalidateQueries?: { dataType: string; };
+    invalidateQueries?: { dataType: string; all?: never } | { all: true, dataType?: never };
```
rob-ghost: **Question:** I know this was an object before, but could this type be simpler?

```typescript
invalidateQueries?: string | 'all'
```
used like:
```typescript
invalidateQueries: dataType
// or
invalidateQueries: 'all'
```

Do we get benefit from it being discriminated union of an object with only one property?

apps/admin-x-framework/src/api/settings.ts
```diff
@@ -46,13 +46,14 @@ export const useEditSettings = createMutation<SettingsResponseType, Setting[]>({
...
-    }
+    },
+    invalidateQueries: {all: true}
 });
 
 export const useDeleteStripeSettings = createMutation<unknown, null>({
     method: 'DELETE',
     path: () => '/settings/stripe/connect/',
-    invalidateQueries: {dataType}
+    invalidateQueries: {all: true}
```
rob-ghost: **Question:** Does this forcefully invalidate _all_ query caches for the whole `<App>`, or is this just for the `<Settings>` query client? I presume this is a nuclear way of re-setting the state? Is there something special about deleting stripe settings which should trigger a complete re-load?

(I can see why invalidating all on editing `settings` above might make sense).

apps/admin/src/settings/Settings.tsx
```diff
@@ -0,0 +1,73 @@
...
+                zIndex: 20,
+            }}
+        >
+            <App
+                designSystem={{
+                    darkMode: false,
+                    fetchKoenigLexical: async () => {},
+                }}
+                officialThemes={officialThemes}
+                zapierTemplates={zapierTemplates}
```
rob-ghost, line 68: **Question:** I presume these hard-coded values are temporary for testing, do we have an idea of an appropriate place to load these from in the future? Having them as global `<App>` state doesn't seem appropriate.

If we're loading them from a query I assume we'll need to handle a loading state somewhere in `<Settings>`?

rob-ghost (commented):
Icon changes 👍🏻 

Showing the Settings app on feature flag 👍🏻 

Query invalidation ❓

CI 🔴 

I don't understand the need for the `nql` workarounds, could you explain why they're needed and if we have an issue to track updating `nql`?

apps/admin-x-framework/src/utils/api/hooks.ts
jonatansberg: That would be cleaner, but it would also break all existing usage. Adding it this way makes the change non-breaking.

apps/admin-x-framework/src/api/settings.ts
jonatansberg: Yeah, you're probably right. I know we have some other things that depend on you having subscribers and payments set up, but this is probably over the top.

apps/admin/src/settings/Settings.tsx
jonatansberg: Fun fact: they live in a similar list that's hardcoded in the Ember app. I think I copied this list from another test configuration. 

We should definetly migrate the canonical list in Ember to the settings app itself. I haven't had the time to look into if that is safe to just move over (and delete in Ember) as I'd rather not duplicate it.

jonatansberg:
> I don't understand the need for the nql workarounds, could you explain why they're needed and if we have an issue to track updating nql?

I don't know where such an issue exists. I copy-pasted what was in the vite config for the settings app. 

I tried without it, but then things blow up. The problem is seemingly there isn't a browser version of the nql package. We end up pulling in what is effectively a Node package, but since we don't use anything in the package that actually depends on Node APIs stubbing them out is enough. 

We should probably fix the nql package though.

apps/admin-x-framework/src/api/settings.ts
kevinansfield: Agreed it doesn't make as much sense when disconnecting Stripe.

I can understand being careful in case other APIs change their output when Stripe is disconnected but I don't believe we have any that do.

However, disconnecting Stripe is a very uncommon action so I don't think there's much harm leaving it here to be on the safer side.

jonatansberg:
@rob-ghost @kevinansfield I had to redo the query invalidation slightly as invalidating all of the queries, including the setting ones, broke the settings E2E test.

As discussed previously, these tests are relying heavily on (indirectly) asserting what's written to the query cache on a successful mutation. Since invalidating the queries will refetch the initial mock data that will overwrite the updated cache value, meaning some assertions fail and other become flaky as there is a race condition between the assertion and the re-render on invalidation.

apps/admin-x-framework/src/api/settings.ts
```diff
@@ -46,6 +46,17 @@ export const useEditSettings = createMutation<SettingsResponseType, Setting[]>({
...
+    // Whenever we update the settings, we want to make sure we invalidate all
+    // other queries to ensure any ripple effects are reflected in the UI. The
+    // updated settings themselves will have been returned in the settings
+    // response, so we don't need to refetch them.
+    invalidateQueries: {
+        filters: {
+            predicate(query) {
+                return query.queryKey[0] !== dataType;
+            }
+        }
```
kevinansfield, line 59: It will be interesting to see what impact this has with extra queries being made on settings changes. I'm guessing it will mostly be a re-fetch of the user and config on each change based on what's displayed alongside settings?

None of our Admin API responses change based on settings updates so this is definitely a belt-and-braces protection. In the Ember app we have the same situation - changing settings only updates the settings record in the store, all other data stays the same until user action triggers a refetch of some specific data.

kevinansfield: It triggers a lot more queries than I thought 😅 Prior to this change it was just the single PUT request when making a change to settings, after this change we're re-fetching the world

Image

kevinansfield (approved):
I'm skeptical about the need to refetch everything on settings updates but aside from that everything is working well 👍

apps/admin-x-framework/src/api/settings.ts
jonatansberg: We could make this less nuclear and refetch the site and config endpoints.. I'll merge this now so we can get on with things but I'll add a follow-up to look into this.

PR #25376, opened 2025-11-06
Fixed member count display in React admin sidebar

jonatansberg:
Replaced hardcoded count with live data. The count syncs automatically with Ember changes through the existing bridge.

Note: When testing this locally, I noticed that it's not possible to delete members within the new admin shell. I'll take a look at that and try to resolve it. 

Closes https://linear.app/ghost/issue/BER-2983/display-correct-member-count-in-sidebar

### DIFF
--- a/apps/admin/src/layout/app-sidebar/hooks/useMemberCount.ts
+++ b/apps/admin/src/layout/app-sidebar/hooks/useMemberCount.ts
@@ -1,0 +1,15 @@
+import { useBrowseMembers } from "@tryghost/admin-x-framework/api/members";
+
+/**
+ * Hook to fetch the total member count efficiently.
+ * Only fetches pagination metadata (limit=1) to minimize API overhead.
+ * Automatically invalidates when members are created/updated/deleted in Ember.
+ */
+export function useMemberCount() {
+    const { data: membersData } = useBrowseMembers({
+        searchParams: { limit: '1' }
+    });
+    
+    return membersData?.meta?.pagination.total;
+}
+
--- a/apps/admin/src/layout/app-sidebar/NavContent.tsx
+++ b/apps/admin/src/layout/app-sidebar/NavContent.tsx
@@ -10,9 +10,11 @@ import {
 } from "@tryghost/shade"
 import { NavMenuItem } from "./NavMenuItem";
 import NavSubMenu from "./NavSubMenu";
+import { useMemberCount } from "./hooks/useMemberCount";
 
 function NavContent({ ...props }: React.ComponentProps<typeof SidebarGroup>) {
     const [postsExpanded, setPostsExpanded] = useState(false);
+    const memberCount = useMemberCount();
 
     return (
         <SidebarGroup {...props}>
@@ -90,7 +92,9 @@ function NavContent({ ...props }: React.ComponentProps<typeof SidebarGroup>) {
                             <LucideIcon.Users />
                             <NavMenuItem.Label>Members</NavMenuItem.Label>
                         </NavMenuItem.Link>
-                        <SidebarMenuBadge>24</SidebarMenuBadge>
+                        {memberCount != null && (
+                            <SidebarMenuBadge>{memberCount}</SidebarMenuBadge>
+                        )}
                     </NavMenuItem>
                 </SidebarMenu>
             </SidebarGroupContent>


apps/admin/src/layout/app-sidebar/hooks/useMemberCount.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Hook to fetch the total member count efficiently.
+ * Only fetches pagination metadata (limit=1) to minimize API overhead.
+ * Automatically invalidates when members are created/updated/deleted in Ember.
+ */
+export function useMemberCount() {
+    const { data: membersData } = useBrowseMembers({
+        searchParams: { limit: '1' }
+    });
```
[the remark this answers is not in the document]
kevinansfield: @jonatansberg we do have [this role gating in Ember](https://github.com/TryGhost/Ghost/blob/main/ghost/admin/app/helpers/members-count-fetcher.js#L37-L44). Are we happy skipping permissions checks in the React shell until later or should it be added alongside this?

kevinansfield (approved):
Aside from the missing permission check - which may be intentional? - everything else looks good 🙌

We do use the member count in a bunch of places, although usually with additional filters in the query to get specific counts. Is it worth moving this hook higher in the directory structure now as we know it's likely to be used generically later on, or is the preference to keep it local until we have the other use-cases migrated over?

apps/admin/src/layout/app-sidebar/hooks/useMemberCount.ts
rob-ghost, line 11: **Question:** `useBrowseMembers` is async and when it finishes loading `useMemberCount()` will flip from `undefined` to a `number` (assuming loading success, [see Codex's comment](https://github.com/TryGhost/Ghost/pull/25376/files#r2499338939)).

This means on mount of the sidebar, `SidebarMenuBadge` won't exist, then after a brief load, the number will appear. Is this the desired behaviour? If its not desired, does it match the way ember worked?

When it comes to async data like this (I presume there might be other badges, total post count perhaps?), do we have any plan to deal with the loading states or will the data simply appear on load success (so the sidebar doesn't hang)?

apps/admin/src/layout/app-sidebar/hooks/useMemberCount.ts
```diff
@@ -0,0 +1,15 @@
+import { useBrowseMembers } from "@tryghost/admin-x-framework/api/members";
+
+/**
+ * Hook to fetch the total member count efficiently.
+ * Only fetches pagination metadata (limit=1) to minimize API overhead.
```
rob-ghost, line 5: **Day dreaming:** It would be super sweet if we had a dedicated `/meta` path for paginated endpoints which would just do the meta pagination counting + totals.

apps/admin/src/layout/app-sidebar/hooks/useMemberCount.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Automatically invalidates when members are created/updated/deleted in Ember.
```
rob-ghost, line 6: **Suggestion(non-blocking):** I found this confusing because the logic to invalidate the queries is not co-located with `useBrowseMembers` and there's no `useEditMembers` or anything like that in `api/memebers.ts` which would indicate invalidation.

I see that its through `MembersResponseType` and its presence in `EMBER_TO_REACT_TYPE_MAPPING` which enables this, but its very opaque because this data isn't co-located anywhere.

My suggestion is a bit of a meta-one on the Ember -> React bridge. Should we introduce a decorator or higher order function to register datatypes with the Ember -> React bridge which replaces `EMBER_TO_REACT_TYPE_MAPPING` and co-locates the binding next to the datatype definition?

Examples of different possible patterns:
```typescript
// Decorator
// =========

// Registers `MembersResponseType` with the Ember -> React mapping
const dataType = registerEmberDataType('member')('MembersResponseType');

// Type-safe constants
// ===================
// This way we can use find references in editors to find all usages

// apps/admin-x-framework/src/api/dataTypes.ts
export const DATA_TYPES = {
    Members: 'MembersResponseType',
    Posts: 'PostsResponseType',
    Tags: 'TagsResponseType',
    // ... etc
} as const;

export const EMBER_MODEL_MAPPING = {
    'member': DATA_TYPES.Members,
    'post': DATA_TYPES.Posts,
    'tag': DATA_TYPES.Tags,
} as const;

// Integration with createQuery
// ============================

// In createQuery
    if (options.emberModelName) {
        emberBridge.register({
            dataType: options.dataType,
            emberModelName: options.emberModelName
        });
    }

// Link with the property when defining the query
export const useBrowseMembers = createQuery<MembersResponseType>({
    dataType,
    path: '/members/',
    emberModelName: 'member' // <---
});
```
It might be a non-issue given the nature of work but I wanted to explore this a bit because I found the implicit invalidation hard to find / follow and I happened to be familiar with `EmberBridge.tsx`

rob-ghost (commented):
I think Codex + Kevin have a good blocking question about permissions.

I'm ✅ with a couple of questions ❓ but nothing blocking and would approve once the permissions question is resolved.

apps/admin/src/layout/app-sidebar/hooks/useMemberCount.ts
rob-ghost: Thinking about it I think the `const dataType = 'MembersResponseType'` is actually our value instance, and in that way I think at least `EMBER_TO_REACT_TYPE_MAPPING` should either be built from these instances, or reference these instances (meaning we import `members.dataType` in `EmberBridge` to add to the mapping).

If we consider it like so:
```typescript
type DataType = string;

const dataType: DataType = 'MembersResponseType';
```

Then we can consider our options on how to add these instances to our lookup map.

Statically (co-located with the bridge):
```typescript
// EmberBridge.tsx

import { dataType: membersDataType } from '...';

const EMBER_TO_REACT_TYPE_MAPPING: Record<DataType, string> = {
    // ...
    membersDataType: 'member',
    // ...
};
```

Dynamically (co-located with the source):
```typescript
// members.ts
import { registerForInvalidation } from 'ember-bridge';

const dataType: DataType = 'MembersResponseType';

registerForInvalidation(dataType);
```

apps/admin/src/layout/app-sidebar/hooks/useMemberCount.ts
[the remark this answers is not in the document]
jonatansberg: That's a good point. We should add that. I think we can do it as a follow-up though – there are lots of things lacking that kind of gating right now.

jonatansberg: I think in this case that's the better trade-off. We could use suspense or something else to only render the full sidebar when all data is settled, but I think that would be a worse experience. The badge itself is very subtle, and having it render a split second after the sidebar itself would barely be noticeable among the other loading states.

apps/admin/src/layout/app-sidebar/hooks/useMemberCount.ts
jonatansberg: While this is a technically interesting discussion, I'm not entirely convinced of the value add. As we port the member list to React, the missing queries would be added and this would continue to work. I do agree, though, that the whole query key thing and calling methods on the query client to invalidate stuff, and all that, it's a bit too magical and too imperative at the same time. Not a huge fan.

Either way, I don't think there is enough value in making the bridging better. We should rather focus on getting rid of the bridging by porting things over.

jonatansberg:
> We do use the member count in a bunch of places, although usually with additional filters in the query to get specific counts. Is it worth moving this hook higher in the directory structure now as we know it's likely to be used generically later on, or is the preference to keep it local until we have the other use-cases migrated over?

@kevinansfield I'd prefer to lift this up when needed rathe than defaulting to a global location. But it's not a hill I'm willing to die on. There's definitely a case for this being reused and maybe it should be global from the get-go.

PR #25378, opened 2025-11-06
Added cache clear of site data when settings are edited

kevinansfield:
ref https://linear.app/ghost/issue/BER-2984/

- the `/site/` API includes data from settings such as the site title that is used across our React apps. It's unauthenticated meaning it gets fetched by the React shell when Ember is going through the site setup flow which ends up changing data exposed on the site API
- with the auth-route based cache invalidation removed it meant we could end up with stale site data in the React query cache after site setup
- adding explicit clearing of `SiteReponseType` cache entries any time Ember updates settings solves the immediate stale data issue
  - longer-term we shouldn't be relying on site data in our React apps, it's only useful for unauthenticated states and has the danger of becoming stale
- on the React side we also want to invalidate `/site/` query caches in `useEditSettings` so that the changes to site title, accent color, etc within our settings app get reflected correctly

### DIFF
--- a/apps/admin-x-framework/src/api/settings.ts
+++ b/apps/admin-x-framework/src/api/settings.ts
@@ -46,7 +46,8 @@ export const useEditSettings = createMutation<SettingsResponseType, Setting[]>({
             ...newData,
             settings: newData.settings
         })
-    }
+    },
+    invalidateQueries: {dataType: 'SiteResponseType'}
 });
 
 export const useDeleteStripeSettings = createMutation<unknown, null>({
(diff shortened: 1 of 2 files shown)


jonatansberg (commented):
Do we need this if we have the invalidate all behavior as introduced in the settings app PR?

kevinansfield:
If we're invalidating all queries in `useEditSettings` then we don't need the more specific invalidation here. We do still have the problem of the setup flow allowing the site title to be changed, but I've also just realised that's not going through the Ember Data store so we'd want a different approach for that too 🤔

PR #25382, opened 2025-11-06
Added fallback sending domain for domain warming

sam-lord:
ref [GVA-588](https://linear.app/ghost/issue/GVA-588/add-the-ability-to-send-from-a-fallback-domain)

The changes here:

* If the boolean is set at the batch level (which we currently never do) -- then we send from the fallback domain in the SendingService (see BatchSendingService)
* In the sending service, we use the bool passed in to determine the from address we pass to the mail provider, and if it's set we also pass in the fallback domain.
* The address service pulls the fallback address + domain from config (injected in the wrapper file)
* The MailgunEmailProvider was just modified to pass the domain through to the MailgunClient, which can now take a domain to override the default one from config

That's all the logic changes, 87 lines of code changes + tests.

### DIFF
--- a/ghost/core/core/server/services/email-address/EmailAddressService.ts
+++ b/ghost/core/core/server/services/email-address/EmailAddressService.ts
@@ -19,24 +19,40 @@ type LabsService = {
     isSet: (flag: string) => boolean
 }
 
+type GetAddressOptions = {
+    useFallbackAddress: boolean
+}
+
 export class EmailAddressService {
     #getManagedEmailEnabled: () => boolean;
     #getSendingDomain: () => string | null;
     #getDefaultEmail: () => EmailAddress;
+    #getFallbackDomain: () => string | null;
+    #getFallbackEmail: () => EmailAddress | null;
     #isValidEmailAddress: (email: string) => boolean;
     #labs: LabsService;
 
     constructor(dependencies: {
         getManagedEmailEnabled: () => boolean,
         getSendingDomain: () => string | null,
+        getFallbackDomain: () => string | null,
         getDefaultEmail: () => EmailAddress,
+        getFallbackEmail: () => string | null,
         isValidEmailAddress: (email: string) => boolean,
         labs: LabsService
 
     }) {
         this.#getManagedEmailEnabled = dependencies.getManagedEmailEnabled;
         this.#getSendingDomain = dependencies.getSendingDomain;
+        this.#getFallbackDomain = dependencies.getFallbackDomain;
         this.#getDefaultEmail = dependencies.getDefaultEmail;
+        this.#getFallbackEmail = () => {
+            const fallbackAddress = dependencies.getFallbackEmail();
+            if (!fallbackAddress) {
+                return null;
+            }
+            return EmailAddressParser.parse(fallbackAddress);
+        };
         this.#isValidEmailAddress = dependencies.isValidEmailAddress;
         this.#labs = dependencies.labs;
     }
@@ -45,12 +61,20 @@ export class EmailAddressService {
         return this.#getSendingDomain();
     }
 
+    get fallbackDomain(): string | null {
+        return this.#getFallbackDomain();
+    }
+
     get managedEmailEnabled(): boolean {
         return this.#getManagedEmailEnabled();
     }
 
     get defaultFromEmail(): EmailAddress {
         return this.#getDefaultEmail();
+    }
+
+    get fallbackEmail(): EmailAddress | null {
+        return this.#getFallbackEmail();
     }
 
     getAddressFromString(from: string, replyTo?: string): EmailAddresses {
@@ -71,7 +95,7 @@ export class EmailAddressService {
      * If we send an email from an email address that doesn't pass, we'll just default to the default email address,
      * and instead add a replyTo email address from the requested from address.
      */
-    getAddress(preferred: EmailAddresses): EmailAddresses {
+    getAddress(preferred: EmailAddresses, options: GetAddressOptions = {useFallbackAddress: false}): EmailAddresses {
         if (preferred.replyTo && !this.#isValidEmailAddress(preferred.replyTo.address)) {
             // Remove invalid replyTo addresses
             logging.error(`[EmailAddresses] Invalid replyTo address: ${preferred.replyTo.address}`);
@@ -90,6 +114,19 @@ export class EmailAddressService {
         if (!this.managedEmailEnabled) {
             // Self hoster or legacy Ghost Pro
             return preferred;
+        }
+
+        // Case: use fallback address when warming up custom domain
+        if (this.#labs.isSet('domainWarmup') && options.useFallbackAddress) {
+            const fallbackEmail = this.fallbackEmail;
+            if (fallbackEmail) {
+                return {
+                    from: fallbackEmail,
+                    replyTo: preferred.replyTo || preferred.from
+                };
+            } else {
+                logging.error('[EmailAddresses] Fallback email address is not configured, cannot use fallback address for sending email.');
+            }
         }
 
         // Case: always allow the default from address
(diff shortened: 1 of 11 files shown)


ghost/core/core/server/services/email-address/EmailAddressService.ts
```diff
@@ -19,24 +19,40 @@ type LabsService = {
...
+    #getFallbackEmail: () => EmailAddress | null;
     #isValidEmailAddress: (email: string) => boolean;
     #labs: LabsService;
 
     constructor(dependencies: {
         getManagedEmailEnabled: () => boolean,
         getSendingDomain: () => string | null,
+        getFallbackDomain: () => string | null,
         getDefaultEmail: () => EmailAddress,
+        getFallbackEmail: () => string | null,
```
aileen, line 40: Does this return a `string` or an `EmailAddress`?

ghost/core/core/server/services/email-service/BatchSendingService.js
```diff
@@ -457,6 +457,7 @@ class BatchSendingService {
                 }, {
                     openTrackingEnabled: !!email.get('track_opens'),
                     clickTrackingEnabled: !!email.get('track_clicks'),
+                    useFallbackAddress: batch.get('fallback_sending_domain'),
```
aileen, line 460: Just out of curiosity: this is not defined on the model at this point, right? So will this be undefined per se? When creating the batches and saving the model, the `fallback_sending_domain` is not yet used. I assume, this will be added later?

ghost/core/core/server/services/email-service/SendingService.js
```diff
@@ -127,16 +137,18 @@ class SendingService {
...
-            from: this.#emailRenderer.getFromAddress(post, newsletter),
+            from: this.#emailRenderer.getFromAddress(post, newsletter, !!options.useFallbackAddress),
             replyTo: this.#emailRenderer.getReplyToAddress(post, newsletter) ?? undefined,
             html: emailBody.html,
             plaintext: emailBody.plaintext,
             recipients,
             emailId: emailId,
-            replacementDefinitions: emailBody.replacements
+            replacementDefinitions: emailBody.replacements,
+            domain: options.useFallbackAddress ? this.#emailAddressService.fallbackDomain : undefined
```
aileen: I think we talked about this briefly that passing `undefined` here will mean that Mailgun sorts out the domain. Is it worth a comment here?

ghost/core/core/server/services/lib/MailgunClient.js
```diff
@@ -106,7 +106,11 @@ module.exports = class MailgunClient {
 
             const mailgunConfig = this.#getConfig();
             startTime = Date.now();
-            const response = await mailgunInstance.messages.create(mailgunConfig.domain, messageData);
+
+            // Use overriden domain if specified in message
+            const mailDomain = message.domain ? message.domain : mailgunConfig.domain;
```
aileen: Related to comment above. This makes it clear, but when reading it originally, I'm wondering why we would pass an `undefined` as `domain`.

ghost/core/test/unit/server/services/email-address/EmailAddressService.test.ts
```diff
@@ -0,0 +1,215 @@
...
+            assert.equal(result.from.address, 'fallback@fallback.example.com');
+            assert.equal(result.replyTo?.address, 'custom@custom.example.com');
+            assert.equal(result.replyTo?.name, 'Custom Sender');
+        });
+
+        it('does not use fallback address when useFallbackAddress is false', function () {
+            labsStub.isSet.withArgs('domainWarmup').returns(true);
+
+            const service = new EmailAddressService({
+                getManagedEmailEnabled: () => false,
```
aileen: I _think_ coderabbit has a point here. Is it intentional that `getManagedEmailEnabled` returns `false`? Looking at the code [here](https://github.com/TryGhost/Ghost/pull/25382/files#diff-d2a7521a7f4fefb6e42f727b1694bfb219d3dc35b0ad6e32fa6d75b022a7cc24R114-R117) it does seem to return `preferred` and not test our intended behavior.

ghost/core/test/unit/server/services/email-address/EmailAddressService.test.ts
```diff
@@ -0,0 +1,215 @@
...
+            assert.equal(result.from.address, 'custom@custom.example.com');
+            assert.equal(result.from.name, 'Custom Sender');
+            assert.equal(result.replyTo, undefined);
+        });
+
+        it('does not use fallback address when domainWarmup flag is disabled', function () {
+            labsStub.isSet.withArgs('domainWarmup').returns(false);
+
+            const service = new EmailAddressService({
+                getManagedEmailEnabled: () => false,
```
aileen: Same as above

ghost/core/test/unit/server/services/email-address/EmailAddressService.test.ts
```diff
@@ -0,0 +1,215 @@
...
+            assert.equal(result.from.address, 'custom@custom.example.com');
+            assert.equal(result.from.name, 'Custom Sender');
+            assert.equal(result.replyTo, undefined);
+        });
+
+        it('does not use fallback address when fallback email is not configured', function () {
+            labsStub.isSet.withArgs('domainWarmup').returns(true);
+
+            const service = new EmailAddressService({
+                getManagedEmailEnabled: () => false,
```
aileen: Same here

ghost/core/test/unit/server/services/email-address/EmailAddressService.test.ts
```diff
@@ -0,0 +1,215 @@
...
+                from: {address: 'custom@custom.example.com', name: 'Custom Sender'},
+                replyTo: {address: 'support@custom.example.com', name: 'Support'}
+            }, {useFallbackAddress: true});
+
+            assert.equal(result.from.address, 'fallback@fallback.example.com');
+            assert.equal(result.replyTo?.address, 'support@custom.example.com');
+            assert.equal(result.replyTo?.name, 'Support');
+        });
+
+        it('uses fallback address before defaulting to managed email', function () {
```
aileen: Is this not the same test as the first one?

aileen (changes requested):
A few notes in the tests, but otherwise looking good

ghost/core/core/server/services/email-address/EmailAddressService.ts
sam-lord: A string, since it's just the value being pulled from config. If we wanted to return an email address then the function in EmailAddressServiceWrapper would need to have an email address parser.

I think that while it isn't as clean in the dependency list, having the actual logic to parse the email address exist within this file rather than the wrapper makes more sense. See the additions at line 49 for the parsing to make #getFallbackEmail into an EmailAddress

ghost/core/core/server/services/email-service/BatchSendingService.js
sam-lord: That's right. This PR is concerned with _how to send from the fallback domain_.

The PR concerned with _when_ to send from the fallback domain is here (stacked on top of this): https://github.com/TryGhost/Ghost/pull/25394

ghost/core/core/server/services/email-service/SendingService.js
sam-lord: Oooh, yes. I think the outcome of the conversation was that naming could help this. Instead of passing `domain`, we could pass something like `domainOverride`, which would be self-documenting, since an absent override implies that a default would be used.

ghost/core/test/unit/server/services/email-address/EmailAddressService.test.ts
sam-lord: Yeah I think this was a mistake. I'll sort this :)

aileen (approved):
Nice!! Looking good now. Thanks for answering my questions 😊

PR #25383, opened 2025-11-06
Wired up dark mode toggle

kevinansfield:
ref https://linear.app/ghost/issue/BER-2952/

- used the new user preferences hooks to wire up the `nightShift` preference to the dark mode toggle UI
- on the Ember side we need to explicitly make the theme change call because we're no longer making the data change via the feature service


jonatansberg (approved):
Neat. I think the bot's comment on error handling could be worth a look. We might want to discuss our overall approach to that in the next mob session?

PR #25390, opened 2025-11-10
Fixed modal event handling in React shell

jonatansberg:
## Problem
Modal buttons (delete, save, etc.) weren't working when Ember runs inside the React shell because modals render to `#ember-modal-wormhole` outside the `#ember-app` container where EventDispatcher listens.

## Solution
Converted `GhTaskButton` and `GhDropdownButton` to tagless components using `{{on}}` modifiers, which attach native event listeners directly to DOM elements instead of relying on EventDispatcher delegation.

## Changes
- `gh-task-button` → tagless with explicit `<button {{on "click"}}>` 
- `gh-dropdown-button` → tagless with explicit `<button {{on "click"}}>` 

This fixes all modal button interactions in React shell while maintaining backward compatibility with standalone Ember.

---

fixes https://linear.app/ghost/issue/BER-2986/event-delegation-does-not-work-in-modals

### DIFF
--- a/ghost/admin/app/components/gh-dropdown-button.js
+++ b/ghost/admin/app/components/gh-dropdown-button.js
@@ -1,34 +1,31 @@
 import Component from '@ember/component';
-import DropdownMixin from 'ghost-admin/mixins/dropdown-mixin';
 import classic from 'ember-classic-decorator';
-import {attributeBindings, tagName} from '@ember-decorators/component';
-import {computed} from '@ember/object';
+import {action, computed} from '@ember/object';
 import {inject as service} from '@ember/service';
 
 @classic
-@tagName('button')
-@attributeBindings('href', 'role', 'type')
-export default class GhDropdownButton extends Component.extend(DropdownMixin) {
+export default class GhDropdownButton extends Component {
+    tagName = '';
+
     @service dropdown;
 
     role = 'button';
+    type = 'button';
+    isOpen = false;
 
     // matches with the dropdown this button toggles
     dropdownName = null;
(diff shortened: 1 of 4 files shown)


ghost/admin/app/components/gh-dropdown-button.js
```diff
@@ -1,14 +1,12 @@
...
 import classic from 'ember-classic-decorator';
-import {attributeBindings, tagName} from '@ember-decorators/component';
-import {computed} from '@ember/object';
+import {action, computed} from '@ember/object';
 import {inject as service} from '@ember/service';
 
 @classic
-@tagName('button')
-@attributeBindings('href', 'role', 'type')
-export default class GhDropdownButton extends Component.extend(DropdownMixin) {
```
kevinansfield, line 11: Dropping the `DropdownMixin` means we lose the open/closed class names

jonatansberg: Hmm. Let me look at that again. The reason it got pulled out is because it adds a click handler that won't fire.

jonatansberg: I added back the class names.

jonatansberg:
@kevinansfield I have your comment and the bot ones.

kevinansfield:
I wonder if a safer option would be to extend and replace our application EventDispatcher so it listens for events on `document.body`

jonatansberg:
Closing in favor of https://github.com/TryGhost/Ghost/pull/25391

PR #25393, opened 2025-11-10
Refactored StateBridge polling

kevinansfield:
no issue

- we're looking at implementing new hooks that depend on different events from `StateBridge`, following the existing pattern would mean creating new polling code for each event handler
- extracted polling to `getStateBridge()` function that uses a shared promise so there's only ever one poll running, returning early if a previous poll has already completed
- extracted event subscription to `onEmberStateBridgeEvent` that handles event handler registration/teardown and provides a simple interface for registering event handlers within their respective hooks


jonatansberg (approved):
This looks good. Seems like CI is a bit unhappy though.

PR #25394, opened 2025-11-10
Added domain warming service

sam-lord:
closes [GVA-591](https://linear.app/ghost/issue/GVA-591/create-a-domain-warming-service)
closes [GVA-592](https://linear.app/ghost/issue/GVA-592/integrate-domain-warming-service-into-the-batch-sending-service)

Added the domain warming service, and the touch points where the BatchSendingService & EmailService use this to send emails from both the primary sending domain and the fallback.

My biggest concern is the createBatches function in BatchSendingService -- it's a complex function and we've added some more complexity. When I tried factoring this logic out, it still felt just as complex within the createBatches function, so I've opted to keep it as is. I'd love to receive review comments in that area as I've been looking at it for a while.

Edit: I'm going to try to cut down the batch-sending-service tests, since they're pretty bloated at 600~ lines added.


sam-lord:
@coderabbitai review

ghost/core/core/server/services/email-service/DomainWarmingService.js
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     *
+     * @returns {Promise<number>} The highest number of messages sent from the CSD in a single email (excluding today)
+     */
+    async #getHighestCount() {
+        const email = await this.#emailModel.findOne({
+            filter: `created_at:<${new Date().toISOString().split('T')[0]}`,
+            order: 'csd_email_count DESC'
+        });
```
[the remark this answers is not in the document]
sam-lord: I don't think this is relevant, will confirm though

ghost/core/core/server/services/email-service/BatchSendingService.js
```diff
@@ -234,6 +239,12 @@ class BatchSendingService {
     async createBatches({email, post, newsletter}) {
         logging.info(`Creating batches for email ${email.id}`);
 
+        // Infinity implies all emails should be sent from the primary domain
+        let domainWarmupLimit = Infinity;
+        if (this.#domainWarmingService.isEnabled()) {
+            domainWarmupLimit = Number.isInteger(email.get('csd_email_count')) ? email.get('csd_email_count') : Infinity;
+        }
```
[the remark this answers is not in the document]
sam-lord: There is no way for emailCount to be a non-integer.

sam-lord:
> ## E2E Tests Failed
> To view the Playwright test report locally, run:
> 
> ```shell
> REPORT_DIR=$(mktemp -d) && gh run download 19263444331 -n playwright-report -D "$REPORT_DIR" && npx playwright show-report "$REPORT_DIR"
> ```

I checked this, it seems like a flaky test. I'll run them locally too, just to validate it.

ghost/core/core/server/services/email-service/DomainWarmingService.js
```diff
@@ -0,0 +1,74 @@
+/**
```
aileen: As a new service, I think we have the working agreement about to get merged to write everything new in TS → https://github.com/TryGhost/Working-Agreements/pull/21

ghost/core/core/server/services/email-service/DomainWarmingService.js
```diff
@@ -0,0 +1,74 @@
...
+        return count || 0;
+    }
+
+    /**
+     *
+     * @param {number} lastCount Highest number of messages sent from the CSD in a single email
+     * @returns {number} The limit for sending from the warming sending domain for the next email
+     */
+    #getTargetLimit(lastCount) {
+        if (lastCount <= 100) {
```
aileen: Might be a good idea to have those magic numbers defined as constants and potentially configurable per injection?

ghost/core/core/server/services/email-service/EmailService.js
```diff
@@ -121,6 +126,10 @@ class EmailService {
         const emailCount = await this.#emailSegmenter.getMembersCount(newsletter, emailRecipientFilter);
         await this.checkLimits(emailCount);
 
+        const csdEmailCount = this.#domainWarmingService.isEnabled()
+            ? await this.#domainWarmingService.getWarmupLimit(emailCount)
+            : undefined;
```
aileen: Do I remember correctly that we use `undefined` here to not intentionally set `0` in order to distinguish between already used (even with 0 sends) and not used at all? Maybe worth a comment here, to remember this context.

ghost/core/core/server/services/email-service/BatchSendingService.js
```diff
@@ -262,14 +273,51 @@ class BatchSendingService {
...
-                    );
-                    batches.push(batch);
+                    // Determine how many members to include in this batch
+                    const remainingCustomDomainCapacity = domainWarmupLimit - totalCount;
+                    const membersToProcess = Math.min(members.length, BATCH_SIZE);
+
+                    // Calculate batch splits
+                    const batchesToCreate = [];
+
+                    if (remainingCustomDomainCapacity > 0 && remainingCustomDomainCapacity < membersToProcess) {
```
aileen: I think this would do better as an explanatory variable 🤔  This is hard to follow otherwise. Maybe something like `shouldUseSplitBatch`?

ghost/core/core/server/services/email-service/BatchSendingService.js
```diff
@@ -262,14 +273,51 @@ class BatchSendingService {
...
+                    } else {
+                        // Single batch: all members use same domain
+                        batchesToCreate.push({
+                            members: members.slice(0, membersToProcess),
+                            useFallbackDomain: totalCount >= domainWarmupLimit
+                        });
+                    }
+
+                    // Create all batches
+                    for (const batchConfig of batchesToCreate) {
```
aileen: Maybe this loop could be a candidate for an outsourced function? I would break the large function a bit up...

ghost/core/test/integration/services/email-service/domain-warming.test.js
```diff
@@ -0,0 +1,323 @@
...
+    before(async function () {
+        const agents = await agentProvider.getAgentsWithFrontend();
+        agent = agents.adminAgent;
+
+        await fixtureManager.init('newsletters', 'members:newsletters');
+        await agent.loginAsOwner();
+    });
+
+    // Helper: Create members with newsletter subscription
+    async function createMembers(count, prefix = 'warmup') {
```
aileen, line 22: Are there other places in our integration tests that create members? This feels a bit like it could be a shared util potentially 🤔

ghost/core/test/integration/services/email-service/domain-warming.test.js
```diff
@@ -0,0 +1,323 @@
...
+            const email = await sendEmail('Test Post Day 1');
+            const totalCount = email.get('email_count');
+            const csdCount = email.get('csd_email_count');
+
+            assert.ok(email);
+            assert.equal(email.get('status'), 'submitted');
+            assert.equal(totalCount, 104); // 100 new + 4 fixture members
+            assert.equal(csdCount, totalCount); // All should use custom domain (104 < 200)
+
+            const {customDomainCount, fallbackDomainCount} = await countRecipientsByDomain(email.id);
```
aileen, line 167: This is a really nice assertion helper 🎉

ghost/core/test/integration/services/email-service/domain-warming.test.js
```diff
@@ -0,0 +1,323 @@
...
+
+            await createMembers(50, 'nowarmup');
+
+            const email = await sendEmail('Test Post No Warmup');
+            const totalCount = email.get('email_count');
+            const csdCount = email.get('csd_email_count');
+
+            assert.ok(totalCount > 0, 'Should have sent to some members');
+            assert.ok(csdCount === null || csdCount === 0 || csdCount === totalCount,
+                `CSD count should be null, 0, or equal total when disabled (got ${csdCount})`);
```
aileen: In this test, domain warming is disabled, is the last or statement still valid?

aileen (approved):
A few questions and comments, but generally looking good to me

ghost/core/core/server/services/email-service/DomainWarmingService.js
sam-lord: I think I'll make them constants, but making them configurable would add a lot of extra configuration values -- I'd prefer to only do that if we're getting feature requests for it, or if we're finding we want to tweak these values for specific sites.

ghost/core/core/server/services/email-service/BatchSendingService.js
sam-lord: Looking at doing that atm. It's fairly intertwined with the surrounding logic given we're keeping track of the count of members, but if I can fix that we could break this up

ghost/core/test/integration/services/email-service/domain-warming.test.js
sam-lord: The shared utils all behave quite differently, creating static fixtures. I believe this is the first instance of needing a _large_ count of members.

I could move this into the shared utils -- I just haven't decided where exactly to put it yet.

ghost/core/core/server/services/email-service/DomainWarmingService.js
aileen: Sounds good!

PR #25395, opened 2025-11-10
Fixed sidebar not showing on React routes

kevinansfield:
ref https://github.com/TryGhost/Ghost/commit/11b56ab9f44219e148f3269ae485c56a91e4455a

- changing `EventDispatcher` to register listeners on `document.body` meant Ember started adding the `.ember-application` class to the body element rather than the nested `#ember-app` element
- updated our React shell override styles for parts of the Ember app by adding `.react-admin` class to the body element and using it in place of `:not(.ember-application)`
- fixed the Ember `.show/.hidden !important` styles to be specific to `.gh-app` rather than `.ember-application` so they don't collide with the Shade classes of the same name

### DIFF
--- a/apps/admin/index.html
+++ b/apps/admin/index.html
@@ -14,7 +14,7 @@
     <meta name="apple-mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-status-bar-style" content="black" />
   </head>
-  <body>
+  <body class="react-admin">
     <div id="root"></div>
     <div id="ember-app"></div>
     <div id="ember-basic-dropdown-wormhole"></div>
--- a/apps/admin/src/index.css
+++ b/apps/admin/src/index.css
@@ -1,43 +1,43 @@
 @import "@tryghost/shade/styles.css";
 
 /* Ensure the Ember app renders in the correct position */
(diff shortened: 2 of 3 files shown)


apps/admin/index.html
```diff
@@ -14,7 +14,7 @@
     <meta name="apple-mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-status-bar-style" content="black" />
   </head>
-  <body>
+  <body class="react-admin">
```
jonatansberg, line 17: Nit: Maybe we can give this a better class name? I don't know what it would be tho...

jonatansberg (approved):

peterzimon (approved):

apps/admin/index.html
kevinansfield: I wondered if `.ghost-admin` would work so it matched the project name, but then the naming gets shared across both the React and Ember versions of Admin.

We could go with `.admin-forward` or (`#admin-forward`) to match the naming we use elsewhere? We could then use it in our `inAdminForward` conditionals which are a little confusing when they look for existence of `#ember-app` to indicate they _aren't_ exactly within an Ember app 😅

PR #25401, opened 2025-11-11
Fixed email preview limiter to use a single limit site-wide

cmraible:
no issue

There is a limit on the number of preview emails a Ghost instance can send per hour to prevent abuse, but it is currently enforced per IP address. This changes the limit to be effectively site-wide, regardless of IP address.


cmraible:
Tested on staging, confirmed changing IPs does not get around the preview email limit.

9larsons (approved):

troyciesco (approved):

PR #25405, opened 2025-11-11
fix(activitypub): rewrite relative portal links

TalibYousuf:
What & Why:
Ensure ActivityPub webhook URLs are absolute by constructing them using siteUrl instead of relative paths. Previously, the URLs started with .ghost/..., which could break when content is read outside the website or in fediverse clients.

Impact:
	•	Prevents broken webhook links for Ghost users using ActivityPub integrations.
	•	Ensures consistency across all environments (local, staging, production).

Code change:
	•	ActivityPubService.js: Updated the construction of webhook target_url so all URLs are absolute using new URL(..., this.siteUrl).

Why this is needed:
ActivityPub webhooks are used by Ghost to communicate with external services. Relative URLs could fail when the webhook is triggered outside the Ghost site context (e.g., federated clients). This fix guarantees reliability.

---

> [!NOTE]
> Adds `ActivityPubService` to initialize/check/remove ActivityPub webhooks with absolute target URLs and handle site enable/disable via owner token.
> 
> - **Backend**:
>   - **New service**: `versions/6.7.0/core/server/services/activitypub/ActivityPubService.js`
>     - Builds expected webhooks with absolute `target_url` via `new URL(path, this.siteUrl)`.
>     - Initializes or resets webhooks for `ghost-activitypub` integration after validating state.
>     - Retrieves webhook secret from `/.ghost/activitypub/v1/site` using owner token.
>     - Supports enable/disable flows: removes webhooks and sends DELETE to site endpoint.
>     - Uses `knex` for persistence and `identityTokenService` to get the Owner user token.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4ded96a8d05d9cbfb07c9ad45aee19c635ae7cc7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/versions/6.7.0/core/server/services/activitypub/ActivityPubService.js
+++ b/versions/6.7.0/core/server/services/activitypub/ActivityPubService.js
@@ -1,0 +1,173 @@
+"use strict";
+var __importDefault = (this && this.__importDefault) || function (mod) {
+    return (mod && mod.__esModule) ? mod : { "default": mod };
+};
+Object.defineProperty(exports, "__esModule", { value: true });
+exports.ActivityPubService = void 0;
+const bson_objectid_1 = __importDefault(require("bson-objectid"));
+const node_fetch_1 = __importDefault(require("node-fetch"));
+class ActivityPubService {
+    knex;
+    siteUrl;
+    logging;
+    identityTokenService;
+    constructor(knex, siteUrl, logging, identityTokenService) {
+        this.knex = knex;
+        this.siteUrl = siteUrl;
+        this.logging = logging;
+        this.identityTokenService = identityTokenService;
+    }
+    getExpectedWebhooks(secret) {
+    const events = [
+        'post.published',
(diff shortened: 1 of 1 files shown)


9larsons:
Thanks for the interest in improving ActivityPub! I'm closing this because the change doesn't apply to this repository:  the file it adds lives under` versions/6.7.0/...`, which is the layout of a local Ghost-CLI installation's compiled output — the actual source lives at `ghost/core/core/server/services/activitypub/activity-pub-service.ts`.  

On current main, the webhook URLs are already absolute (built with `new URL(path, siteUrl)`), and the `/ghost/activitypub/...` paths proposed here would bypass the `.ghost/activitypub` proxy prefix the service depends on. If you're still seeing broken links in federated content, please submit an issue we reproduction details so we can take a closer look.

PR #25409, opened 2025-11-11
Added email analytics for the fallback domain

sam-lord:
closes [GVA-593](https://linear.app/ghost/issue/GVA-593/get-email-analytics-from-both-the-default-sending-domain-the-fallback)

Query both the primary domain and the fallback for email analytics events.


sam-lord:
@coderabbit review

PR #25410, opened 2025-11-11
[WIP] Object Storage Adapter

allouis:
This is an initial spike (not prototype, I don't think we need to throw this away) into how we can improve the developer experience for working on Object Storage, as well as using a more generic S3 Adapter.

It's built upen the changes in https://github.com/TryGhost/SDK/tree/PRO-1524/packages/url-utils as referenced in the first commit here.

We've got a new `S3StorageAdapter` in core along with unit tests - but it's missing e2e tests.

We also have a new `compose.object-storage.yml` file, which adds a new `object-storage` profile, which when enabled will spin up MinIO and configure Ghost to use the S3StorageAdapter.

You should be able to pull this branch locally and run:

```
yarn docker:dev:object-storage
```

### DIFF
--- a/.docker/minio/setup.sh
+++ b/.docker/minio/setup.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+set -euo pipefail
+
+BUCKET=${MINIO_BUCKET:-ghost-dev}
+
+echo "Configuring MinIO alias..."
+mc alias set local http://minio:9000 "${MINIO_ROOT_USER}" "${MINIO_ROOT_PASSWORD}"
+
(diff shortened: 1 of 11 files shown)


muratcorlu:
Great effort, really appreciate it! Please also consider updating [Storage Adapter documentation](https://docs.ghost.org/config#storage-adapters) if this reaches to a release. Current version is not clear enough and outdated. https://forum.ghost.org/t/custom-storage-adapter-documentation-is-incomplete/46499

allouis:
@muratcorlu Thanks! This is all still being actively developed, but once it has settled I'll take a look at the docs ☺️

PR #25411, opened 2025-11-11
Moved `favicon` out of the middleware

rmgpinto:
ref no-issue

- The favicon middleware makes a comparison on every request to check if it should serve the favicon. Adding the path prevents this comparison and skips the function entirely on non-favicon requests. Moving it to the `.get` avoids unnecessary comparisons on other http methos (POST, DELETE, PUT, ...) and only serves it on GET requests.
- Applied the same principles as https://github.com/TryGhost/Ghost/blob/middleware-favicon/ghost/core/core/frontend/web/site.js#L126-L126


ErisDS (approved):
This change makes a lot of sense, and follows existing patterns for how to do this kind of inversion, so I'd be in favour of merging.

PR #25414, opened 2025-11-11
Wired up navigation state to user preferences

kevinansfield:
ref https://linear.app/ghost/issue/BER-3002/

Navigation menu state describing expanded sections and whether the menu is shown are stored on the `navigation` property of our `user.accessibility` (user preferences) JSON blob. In the Ember Admin app this is handled by the `navigation` service, for our React version we want to build on the user-preferences hooks.

- added zod schema for navigation preferences
  - this is a potentially user-modified JSON string so we want to be safe but also not spread knowledge/use of defaults across the codebase, so:
  - if not present, will return default navigation preferences when parsing
  - if anything within the parsed data doesn't match our expectations, will fall back to default preferences
- modified `useEditUserPreferences` to accept partial `UserPreferences` records and perform a deep merge
  - enables simpler consumers that only need to know about their own data within the overall preferences object
- modified `useUserPreferences` to accept a `UseQueryOptions` argument
  - enables consumers to adjust the query behaviour, by e.g. selecting particular keys from the overall preferences object
- added `useNavigationPreferences` and `useEditNavigationPreferences` hooks that wrap the user-preferences hooks for the specific `navigation` key
- added use-case specific hooks `useNavigationExpanded` and `useNavigationMenuVisibility`
  - avoids internal details of the navigation preferences data shape leaking out across consumers
  - provides a state-like `[data, setData]` return value for better DX

### DIFF
--- a/apps/admin/src/hooks/user-preferences.ts
+++ b/apps/admin/src/hooks/user-preferences.ts
@@ -61,13 +61,13 @@ export const useUserPreferences = (): UseQueryResult<Preferences> => {
     });
 };
 
-export const useEditUserPreferences = (): UseMutationResult<void, Error, Preferences, unknown> => {
+export const useEditUserPreferences = (): UseMutationResult<void, Error, Partial<Preferences>, unknown> => {
     const queryClient = useQueryClient();
     const { data: user } = useCurrentUser();
     const { mutateAsync: editUser } = useEditUser();
 
     return useMutation({
-        mutationFn: async (updatedPreferences: Preferences) => {
+        mutationFn: async (updatedPreferences: Partial<Preferences>) => {
             if (!user) {
                 throw new Error("User is not loaded");
             }
--- a/apps/admin/src/hooks/user-preferences.ts
+++ b/apps/admin/src/hooks/user-preferences.ts
@@ -18,10 +18,10 @@ const DEFAULT_NAVIGATION_PREFERENCES = {
 const NavigationPreferencesSchema = z.looseObject({
     expanded: z.object({
         posts: z.boolean(),
-    }).catch(DEFAULT_NAVIGATION_PREFERENCES.expanded),
+    }),
     menu: z.object({
         visible: z.boolean(),
-    }).catch(DEFAULT_NAVIGATION_PREFERENCES.menu),
+    }),
 });
 
 const PreferencesSchema = z.looseObject({
--- a/apps/admin/src/hooks/user-preferences.ts
+++ b/apps/admin/src/hooks/user-preferences.ts
@@ -1,4 +1,4 @@
-import { useQuery, useMutation, type UseQueryResult, type UseMutationResult } from "@tanstack/react-query";
+import { useQuery, useMutation, type UseQueryResult, type UseMutationResult, type UseQueryOptions } from "@tanstack/react-query";
 import { z } from "zod";
 import { useQueryClient } from "@tryghost/admin-x-framework";
 import { useCurrentUser } from "@tryghost/admin-x-framework/api/currentUser";
@@ -78,10 +78,13 @@ export type WhatsNewPreferences = z.infer<typeof WhatsNewPreferencesSchema>;
 
 export const userPreferencesQueryKey = (user: User | undefined) => ["userPreferences", user?.id, user?.accessibility] as const;
 
-export const useUserPreferences = (): UseQueryResult<Preferences> => {
+export function useUserPreferences<TData = Preferences>(
+    options?: Omit<UseQueryOptions<Preferences, Error, TData>, 'queryKey' | 'queryFn'>
+): UseQueryResult<TData> {
     const { data: user } = useCurrentUser();
 
-    return useQuery({
+    return useQuery<Preferences, Error, TData>({
+        ...options,
         queryKey: userPreferencesQueryKey(user),
         queryFn: () => {
             if (!user) {
@@ -102,7 +105,7 @@ export const useUserPreferences = (): UseQueryResult<Preferences> => {
         // the current active entry cached indefinitely.
         cacheTime: 0,
     });
-};
+}
 
 export const useEditUserPreferences = (): UseMutationResult<void, Error, DeepPartial<Preferences>, unknown> => {
     const queryClient = useQueryClient();
--- a/apps/admin/src/hooks/user-preferences.ts
+++ b/apps/admin/src/hooks/user-preferences.ts
@@ -1,6 +1,5 @@
 import { useQuery, useMutation, type UseQueryResult, type UseMutationResult } from "@tanstack/react-query";
 import { z } from "zod";
-
 import { useQueryClient } from "@tryghost/admin-x-framework";
 import { useCurrentUser } from "@tryghost/admin-x-framework/api/currentUser";
 import { useEditUser, type User } from "@tryghost/admin-x-framework/api/users";
@@ -29,7 +28,7 @@ export type NavigationPreferences = z.infer<typeof NavigationPreferencesSchema>;
 const PreferencesSchema = z.looseObject({
     whatsNew: WhatsNewPreferencesSchema.optional().catch(undefined),
     nightShift: z.boolean().optional(),
-    navigation: NavigationPreferencesSchema.optional().catch(DEFAULT_NAVIGATION_PREFERENCES),
+    navigation: NavigationPreferencesSchema.default(DEFAULT_NAVIGATION_PREFERENCES).catch(DEFAULT_NAVIGATION_PREFERENCES),
 });
 
 export type Preferences = z.infer<typeof PreferencesSchema>;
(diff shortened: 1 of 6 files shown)


apps/admin/src/utils/deep-merge.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export type DeepPartial<T> = T extends object ? {
+    [P in keyof T]?: DeepPartial<T[P]>;
+} : T;
+
+/**
+ * Deep merge utility for preferences objects.
+ * Recursively merges nested objects, with values from `source` taking precedence.
+ */
+export function deepMerge<T extends Record<string, unknown>>(target: T, source: DeepPartial<T>): T {
```
kevinansfield, line 26: This util is an AI dump. Q: do we want to import from a utils library instead? If so, which one?

I couldn't find any existing usage of `lodash` across our React apps outside of `admin-x-design-framework`. It is listed as a dependency of shade but it's not used afaics.

rob-ghost: Can we use [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone)? I'd favour that or implementing it ourselves. We could pull in a lib, or `lodash/deepClone` by itself but the less third party dependancies the better.

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -14,20 +14,40 @@ const WhatsNewPreferencesSchema = z.looseObject({
...
     whatsNew: WhatsNewPreferencesSchema.optional().catch(undefined),
     nightShift: z.boolean().optional(),
+    navigation: NavigationPreferencesSchema.default(DEFAULT_NAVIGATION_PREFERENCES).catch(DEFAULT_NAVIGATION_PREFERENCES),
 });
 
 export type Preferences = z.infer<typeof PreferencesSchema>;
 export type WhatsNewPreferences = z.infer<typeof WhatsNewPreferencesSchema>;
 
-const userPreferencesQueryKey = (user: User | undefined) => ["userPreferences", user?.id, user?.accessibility] as const;
+export const userPreferencesQueryKey = (user: User | undefined) => ["userPreferences", user?.id, user?.accessibility] as const;
```
rob-ghost: **Nitpick:** This doesn't need to be exported anymore.

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -14,20 +14,40 @@ const WhatsNewPreferencesSchema = z.looseObject({
...
+export const NavigationPreferencesSchema = z.looseObject({
+    expanded: z.object({
+        posts: z.boolean(),
+    }),
+    menu: z.object({
+        visible: z.boolean(),
+    }),
+});
+
+export type NavigationPreferences = z.infer<typeof NavigationPreferencesSchema>;
```
rob-ghost: **Nitpick:** Lets put this next to the other inferred types below.

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -14,20 +14,40 @@ const WhatsNewPreferencesSchema = z.looseObject({
...
 
 export type Preferences = z.infer<typeof PreferencesSchema>;
 export type WhatsNewPreferences = z.infer<typeof WhatsNewPreferencesSchema>;
 
-const userPreferencesQueryKey = (user: User | undefined) => ["userPreferences", user?.id, user?.accessibility] as const;
+export const userPreferencesQueryKey = (user: User | undefined) => ["userPreferences", user?.id, user?.accessibility] as const;
 
-export const useUserPreferences = (): UseQueryResult<Preferences> => {
+export function useUserPreferences<TData = Preferences>(
+    options?: Omit<UseQueryOptions<Preferences, Error, TData>, 'queryKey' | 'queryFn'>
```
rob-ghost: **Suggestion:** We should omit `staleTime` and `cacheTime` too.

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -48,25 +68,23 @@ export const useUserPreferences = (): UseQueryResult<Preferences> => {
...
             }
 
-            const currentPreferences = queryClient.getQueryData<Preferences>(userPreferencesQueryKey(user)) ?? {};
+            const currentPreferences = queryClient.getQueryData<Preferences>(userPreferencesQueryKey(user)) ?? PreferencesSchema.parse({});
 
-            const newPreferences: Preferences = {
-                ...currentPreferences,
-                ...updatedPreferences,
-            };
+            // TODO: use zod to validate?
```
rob-ghost, line 81: I'm not 100% sure but we can test it, but I think `encode` also validates so we don't need to do it manually before encoding below.

apps/admin/src/utils/deep-merge.ts
kevinansfield: We could possibly use `structuredClone` in parts of the `deepMerge` util or to have something immutable but I can't see any way that `structuredClone` that would allow us to do a deep merge directly.

For third party libraries we could pull in `lodash/merge` but it still doesn't give us immutability. There's [@fastify/deepmerge](https://github.com/fastify/deepmerge) which is fast and doesn't mutate the original object but it's still quite a chunky dependency to add for a single use-case 🤔

rob-ghost (approved):
Some nitpicks but nothing blocking that can't be followed up on later if needed.

I'd like to see if we can use the `structuredClone` instead of hand rolling, but I'd prefer hand-rolling over importing a library just to get deep cloning.

apps/admin/src/utils/deep-merge.ts
kevinansfield: Alternatively, if this is a sticking point we could revert `useEditUserPreferences` back to doing a shallow merge and have the consumers of top-level keys manage how they handle their own nested objects.

rob-ghost: 🤦🏻 I went down the "deep clone" and not "deep merge" hole 🕳️

rob-ghost: Unless we're going to use a lib like `lodash` for many use cases, for utilities like this I'd prefer to just roll our own, especially with agents these days who can spin up pretty thorough test cases if needed.

apps/admin/src/hooks/user-preferences.ts
kevinansfield: Seems like .catch and .defaults are only applied in the forward direction (https://zod.dev/codecs#how-encoding-works). I can't find anything in the docs explicitly about validation, although Claude says it doesn't occur with `.encode`

It may not be necessary, it was just an observation after using `.parse()` on the merged navigation object before passing it to `userEditUserPreferences` in an earlier iteration.

apps/admin/src/utils/deep-merge.ts
jonatansberg: I'd be kind of inclined to go with shallow merging of top keys. Deep mergers are less than ideal performance-wise, tricky to get right, and quite often you end up in unexpected states. I find it especially hard to reason about removing data. You can't really describe an array with one element less or an object with one element removed.

rob-ghost: For me, if the `user-preferences` intends to both be permissive of all shapes, while also attempting to validate, encode, decode and describe the shape that we want it to be, then it makes sense if `user-preferences` handles the merging of "known" preferences being provided by the app, and "unknown" preferences which may or may not exist in `user.accessibility`.

For this example, we have `preferences.navigation.expanded[some type]: boolean`. Its possible that the `user.accessibility` has `preferences.navigation.dismissed: unknown`.

Unless we want to totally ignore unknown keys within `preferences.navigation`, we need to deep merge somewhere. So if we agree that we want to maintain existing data that we might not know about and we permit this with our loose `PreferencesSchema`, then we can either:

1. Deep merge in the hook which manages `preferences.navigation`
2. Deep merge in the `user-preferences` module

I personally think we should have the policy of "loose in, loose out" which is maintained at the lowest level possible (i.e `user-preferences`) because it removes so much boilerplate from dependant queries who want to work with a subset of Preferences data without worrying about merge behaviour.

Alternatively, we adopt a `loose in, strict out` which forces us to provide exact, complete top level key values, which _won't_ get automatically merged, and the intention is to actually clean out unrecognised preferences which will establish eventual consistency. Its ergonomically a little harder to work with because you won't be providing a `Partial<NavigationPreferences>` to `editPreferences(Partial<Preferences>)`, you'll need to give a whole `NavigationPreferences`.

The additional challenge of `NavigationPreferences` is that `expanded` is actually of type `[string]: boolean`, because we allow the app to define any arbitrary key to track expansion state on (such as `"posts"`) which are not known at parsing time.

---

Personally, I don't mind `loose in / loose out` with auto deep merge, vs `loose in / strict out` with shallow top-level merge but enforced complete preference type instances and no deep merging, but I don't like the half-way house of "loose in / loose out is up to you" where we expect preference users to merge themselves (unless they absolutely want to).

If we can force the chosen policy with types, that would. be great!

---

What do you think?

jonatansberg: Those are some valid points. My current thinking is that I would like to get this merged and then we can iterate on the nuances later 😊

PR #25423, opened 2025-11-12
🎨 Added Analytics APIs to robots.txt

jloh:
ref https://linear.app/ghost/issue/ONC-1256/
ref https://linear.app/ghost/issue/NY-705/

- Robots can (and currently do) browser + track hits against the Analytics Endpoints
- We filter this data out at different points but preventing it from being tracked at in the first place is easier


9larsons (approved):

cmraible (approved):
Good idea, thanks!

PR #25432, opened 2025-11-12
Added handling for apple-app-site-association files

ErisDS:
fixes https://linear.app/ghost/issue/ONC-1251/universal-linking-no-longer-work-on-ios#comment-c26d5224

- In Ghost v6 we removed support for serving files from the root of a theme if they don't have an extension
- Apple's apple-app-site-association file is needed for universal linking in apps
- This file is a JSON file, it must not have a file extension but it must be served with a JSON content type
- This requirement from apple is very specific and creates a unique exception case
- We'll solve it by having a unique exception in the code just to support this one file


cmraible (approved):
Nice, LGTM!

PR #25434, opened 2025-11-13
Improving TypeScript support for Adapters

allouis:
When using `export default class` in TS the transpiled JS exposes the
class on `module.exports.default` rather than `module.exports`. We need
to be able to read from both in order to write new adapters in TS.

Migrating the whole of `ghost-storage-base` to TypeScript is outside the
scope of this work, but we do need to provide some types so that the S3
adapter implemntation can work correctly.

---

> [!NOTE]
> Adapter loading now resolves default exports for TS/ES modules; tsconfig includes local d.ts and new `ghost-storage-base` typings are added.
> 
> - **Adapter Loading**:
>   - Add `resolveAdapterExport` in `core/server/services/adapter-manager/AdapterManager.js` to handle function exports and `default` exports.
>   - Use resolved export when instantiating adapters, enabling TS `export default` classes.
> - **TypeScript**:
>   - Update `tsconfig.json` to include `types/**/*.d.ts`.
>   - Add `types/ghost-storage-base.d.ts` providing typings for `ghost-storage-base` (`StorageFile` type and `StorageBase` abstract class).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1d22aca7353372999372623eab64aa1157b1621e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/core/core/server/services/adapter-manager/AdapterManager.js
+++ b/ghost/core/core/server/services/adapter-manager/AdapterManager.js
@@ -1,6 +1,22 @@
 const path = require('path');
 const errors = require('@tryghost/errors');
 
+const resolveAdapterExport = (moduleExport) => {
+    if (!moduleExport) {
+        return moduleExport;
+    }
+
+    if (typeof moduleExport === 'function') {
+        return moduleExport;
+    }
+
+    if (typeof moduleExport === 'object' && typeof moduleExport.default === 'function') {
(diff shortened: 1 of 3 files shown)


vershwal (approved):
LGTM!

PR #25440, opened 2025-11-13
Changed deep merge to only recurse into plain objects.

rob-ghost:
Fixes issue where deep merge would recurse into complex objects (like Dates) which changed their type in the output to plain Object. This would then break validation as the instance was not the expected type.


jonatansberg (approved):
Looks good to me for now as a hotfix. I think we should revisit the deep-merge stuff next week.

PR #25447, opened 2025-11-13
Applied dark mode fixes to React app

peterzimon:
ref https://linear.app/ghost/issue/BER-2998/fix-dark-mode-design-issues

- The sidebar and the main app shell was not prepared completely for dark mode.


apps/admin/src/layout/app-sidebar/UserMenu.tsx
```diff
@@ -44,7 +44,7 @@ function UserMenu(props: UserMenuProps) {
                 >
                 <Avatar>
                     {currentUser.data?.profile_image && <AvatarImage src={currentUser.data?.profile_image} />}
-                    <AvatarFallback className="text-gray-800 hover:text-black">
```
rob-ghost, line 47: **Question:** Does this need a `dark:hover:` style? Will `hover:text-black` be too dark in dark mode? `dark:hover:text-white` perhaps?

rob-ghost (approved):
One question about hover style in dark mode but other than that 👍🏻

PR #25449, opened 2025-11-13
Fixed fallback from name & reply-to

sam-lord:
closes [GVA-603](https://linear.app/ghost/issue/GVA-603/fix-from-address-for-fallback-domain)

The fallback email doesn't have a name, and if there's no preferred address we weren't setting any reply-to. We'll always want a reply-to address if we're using the fallback.


aileen (approved):
👍

PR #25450, opened 2025-11-13
🐛 Fixed double slash in event sending path

troyciesco:
ref https://linear.app/ghost/issue/NY-700

TK


cmraible:
Closing this as we've fixed the issue upstream 👍

PR #25451, opened 2025-11-13
🐛 Fixed missing member attribution stats on pages list view

cmraible:
ref https://linear.app/ghost/issue/ONC-1266/analytics-data-missing-from-post-list-but-present-in-top-pages-section

The member attribution stats on the pages list view in Admin are currently always showing 0 new members, even though the "Top Posts" section of Analytics shows the correct number.

The stats API endpoint that returns the member counts for posts was specifically filtering for `attribution_type = 'post'`, which was filtering out the attribution rows for pages, resulting in 0 new members displaying for all pages on the pages list view in Admin.

### DIFF
--- a/ghost/core/test/unit/server/services/stats/posts.test.js
+++ b/ghost/core/test/unit/server/services/stats/posts.test.js
@@ -137,6 +137,73 @@ describe('PostsStatsService', function () {
         await _createPaidConversionEvent(conversionPostId, finalMemberId, finalSubscriptionId, mrr, referrerSource);
     }
 
+    // Helper functions for page attribution (attribution_type: 'page')
+    async function _createFreeSignupEventForPage(pageId, memberId, referrerSource, createdAt = new Date()) {
+        eventIdCounter += 1;
+        const eventId = `free_event_${eventIdCounter}`;
+        await db('members_created_events').insert({
+            id: eventId,
+            member_id: memberId,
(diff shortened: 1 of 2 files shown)


ghost/core/test/unit/server/services/stats/posts.test.js
```diff
@@ -691,6 +700,20 @@ describe('PostsStatsService', function () {
...
+            await _createPaidConversion('page1', 'page2', 500, 'referrer_1');
+
+            const result = await service.getGrowthStatsForPost('page1');
+
+            const expectedResults = [
+                {post_id: 'page1', free_members: 2, paid_members: 1, mrr: 500}
+            ];
+
+            assert.deepEqual(result.data, expectedResults, 'Results should match expected order and counts for free_members desc');
+        });
```
cmraible, line 716: This test is virtually the same as the existing one above, but the expected result isn't making sense to me.

cmraible: 🤦‍♂️ discussed with Steve, it's correct - the `_createPaidConversion` as used here creates a free signup on page1, then converts to paid on a different paid, so 2 free members here is correct.

9larsons (approved):
Solid 👍. I see a few other places using `'post'` specifically that concern me, I'll write up an issue for us to take a look at those later.

PR #25459, opened 2025-11-17
Updated typescript config with import aliases

ibalosh:
**what**

* added import aliases for most used helpers
* added ESlint rule to enforce path aliases over relative imports

**why**

* import aliases make importing simpler, with clearer intentions on what is imported


cmraible (approved):
Very nice, thanks for doing this!

There are a few places where I _think_ we should be able to use a more specific alias (i.e. `@admin-pages` instead of `@helpers/pages/admin/....` but definitely nothing showstopping - ship it!

ibalosh:
> Very nice, thanks for doing this!
> 
> There are a few places where I _think_ we should be able to use a more specific alias (i.e. `@admin-pages` instead of `@helpers/pages/admin/....` but definitely nothing showstopping - ship it!

I agree @cmraible , this is definitely just a first stab at it, to get imports handled nicer. Linter auto fix is little bit clunky when auto fixing it too. I will merge it in and as we feel it out, we can improve it further.

PR #25460, opened 2025-11-17
Hide sidebar in editor

jonatansberg:
fixes https://linear.app/ghost/issue/BER-3009/hide-sidebar-in-editor

Implemented Ember-React state bridge for sidebar visibility to avoid timing issues between Ember and React Router. The Ember editor route controls visibility via state-bridge and React reads directly from Ember as source of truth using `useSyncExternalStore`.

I also removed the transition animations as it opens and closes to match the current Ember behavior. Also, that animation would stutter and not look very nice.

### DIFF
--- a/apps/admin/src/ember-bridge/EmberBridge.tsx
+++ b/apps/admin/src/ember-bridge/EmberBridge.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect, useState, useSyncExternalStore } from "react";
 import { useQueryClient } from "@tanstack/react-query";
 
 export interface EmberBridge {
@@ -9,6 +9,7 @@ export type StateBridgeEventMap = {
     emberDataChange: EmberDataChangeEvent;
     emberAuthChange: EmberAuthChangeEvent;
     subscriptionChange: SubscriptionState;
+    sidebarVisibilityChange: SidebarVisibilityChangeEvent;
 }
 
 export interface StateBridge {
@@ -17,6 +18,7 @@ export interface StateBridge {
     onDelete: (dataType: string, id: string) => void;
     on<K extends keyof StateBridgeEventMap>(event: K, callback: (event: StateBridgeEventMap[K]) => void): void;
     off<K extends keyof StateBridgeEventMap>(event: K, callback: (event: StateBridgeEventMap[K]) => void): void;
+    sidebarVisible: boolean;
 }
 
 declare global {
@@ -41,6 +43,10 @@ export interface SubscriptionState {
         isActiveTrial: boolean;
         trial_end: string | null;
     };
+}
+
+export interface SidebarVisibilityChangeEvent {
+    isVisible: boolean;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


ghost/admin/app/services/state-bridge.js
```diff
@@ -182,4 +183,16 @@ export default class StateBridgeService extends Service.extend(Evented) {
...
+    setSidebarVisible(isVisible) {
+        this.trigger('sidebarVisibilityChange', {
+            isVisible
+        });
+    }
+
+    get sidebarVisible() {
+        // Sidebar is visible when NOT in fullscreen mode
+        return !this.ui.isFullScreen;
+    }
```
jonatansberg, line 197: Maybe the sidebar naming here is confusing and we should stick with the "full screen" since that's how we reference this in Ember. 

On the other hand, full screen is also ambiguous and could mean other things than the sidebar is visible or not...

ghost/admin/app/routes/settings-x.js
```diff
@@ -9,12 +9,18 @@ export default class SettingsXRoute extends AuthenticatedRoute {
 
     activate() {
         super.activate(...arguments);
-        this.ui.set('isFullScreen', true);
+        
+        if (!this.feature.inAdminForward) {
+            this.ui.set('isFullScreen', true);
+        }
```
rob-ghost, line 16: **Question:** Why do we have this logic here?

I think I know, is it because the settings pages are embedded in the React app / sidebar in `inAdminForward` mode, which means if we collapse the sidebar, we also hide the settings?

A comment to explain this would be helpful because I found it hard to follow on first review.

ghost/admin/app/services/state-bridge.js
rob-ghost: Good question. Sidebar visibility seems like it would be the side-effect of some other kind of state. "When X, sidebar is invisible", which would be more declarative than the imperative "hide the sidebar".

I think we could go either way with it and I wouldn't suggest a change right now in this PR unless we were very sure of the right name at this time. Explicitly opening or closing the sidebar works just as well as a route declaring "I want to be fullscreen" and have the sidebar react to that request.

I personally prefer a declarative statement, so I'd go with "Sidebar is hidden when route is fullscreen" but I don't feel strongly about it.

I do agree that "fullscreen" is a bit of a loaded term and maybe something like `immersive`, `focused` or `expanded` would work better.

ghost/admin/app/services/ui.js
```diff
@@ -43,13 +43,24 @@ export default class UiService extends Service {
...
 
+    get isFullScreen() {
+        return this._isFullScreen;
+    }
+
+    set isFullScreen(value) {
+        this._isFullScreen = value;
+        // Trigger sidebar visibility event whenever fullscreen mode changes
+        this.stateBridge.setSidebarVisible(!value);
+    }
```
rob-ghost, line 62: **Daydreaming (not a suggestion for now):** I wonder would a helper like `@bridged` help us here?

```typescript
@tracked
@bridged(
    'sidebarVisibilityChange',
    (isFullScreen) => ({ isVisible: !isFullScreen })
)
isFullScreen = false;
```

Which would let us mark any property as `@bridged` and we could emit the event automatically / generically? It would save us having to write `get/set` for any other tracked properties we wanted to bridge over to React.

apps/admin/src/ember-bridge/EmberBridge.tsx
```diff
@@ -193,3 +199,30 @@ export function useSubscriptionStatus() {
...
+ * Hook to sync sidebar visibility state from Ember.
+ * 
+ * This hook uses useSyncExternalStore to listen to sidebar visibility changes
+ * triggered by Ember routes (e.g., hiding the sidebar when entering the editor).
+ * 
+ * This is a temporary bridge during the Ember -> React migration and should be
+ * removed once the editor is ported to React.
+ */
+export function useSidebarVisibility(): boolean {
+    return useSyncExternalStore(
```
rob-ghost, line 223: **Praise:** Nice use of `useSyncExternalStore`!

**Question/Suggestion:** Could / should other bridge hooks use `useSyncExternalStore` also? Like `useSubscriptionStatus`? Or is there a difference which makes it possible to use `useSyncExternalStore` for this property which wouldn't work for `useSubscriptionStatus` for example?

```typescript
return useSyncExternalStore(
    (callback) => onEmberStateBridgeEvent('subscriptionChange', callback),
    () => window.EmberBridge?.state.subscription ?? null
);
```

rob-ghost (approved):
Nothing blocking for this PR, just some questions and potential suggestions for the future.

ghost/admin/app/routes/settings-x.js
jonatansberg: Yes, this is indeed so we don't hide the sidebar via Ember. It probably wouldn't do any harm, It's just hard to reason about. 

I can add a comment.

apps/admin/src/ember-bridge/EmberBridge.tsx
jonatansberg: Yes, I think the other hooks that read state out of Ember could benefit from this as well since it handles a bunch of edge cases better than the typical hand rolled version using state and effect.

PR #25465, opened 2025-11-17
Changed member welcome email job to run based on config

troyciesco:
ref https://linear.app/ghost/issue/NY-763

- Modifies the member welcome email job so it no longer references the `welcomeEmails` labs flag, and instead checks to ensure `memberWelcomeEmailTestInbox` is set in config
- This will allow us to stress test a higher volume of emails than we'd get by manually turning on the labs flag for a few sites
- By making sure the config exists before scheduling the job there won't be any unnecessary impact to self-hosted instances
- Doesn't send the email to any real members, just to an internal (to Ghost) test inbox
- The `welcomeEmails` lab flag itself isn't deleted yet - we'll likely use it for UI-related toggles in the future

### DIFF
--- a/ghost/core/core/server/services/member-welcome-emails/jobs/index.js
+++ b/ghost/core/core/server/services/member-welcome-emails/jobs/index.js
@@ -1,6 +1,6 @@
 const path = require('path');
 const jobsService = require('../../jobs');
-const labs = require('../../../../shared/labs');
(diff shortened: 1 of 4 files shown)


cmraible (approved):
No notes, nice one 👨‍🍳

So nice and easy to review 🙌

PR #25466, opened 2025-11-18
Established filter architecture in Tinybird pipes

9larsons:
no ref
- removed old/unused query parameters
- made consistent our use of filter pipes
- distinguished between filter types

When reviewing the Tinybird endpoints and filter logic, it became clear there was a number of discrepancies that ultimately led to confusion and inconsistency. This PR is an attempt to alleviate that. Let's do some plumbing...

image

__Overview__
We have several kinds of filters in Tinybird. Let's call them:
- always filters [date filters, site uuid, timezone]
- (page) hit-level filters [pathname, post_uuid, member_status, 
- session filters

These need to be treated distinctly. 

Always filters are so named because they are _always_ applied - they are our primary and secondary keys in `analytics_events`: `ENGINE_SORTING_KEY "site_uuid, timestamp"`. They optimize queries by reducing the data set.

Page hit data/filters are those that may change on any particular `page_hit`. When clicking around, a user is navigating to different `pathname`s which have unique `poat_uuid`s. If they come to a site and then log in, their `member_status` may change (currently).

Session data/filters are those that are wrapped up or attributed to the session itself. For the various UTM parameters, we set those based on the first page hit in the session within `mv_session_data`. Same for source; these are commonly attribution elements. 

__Architecture__
With this is mind, we have to differentiate in where and how we filter by these various categories. We need to make sure when filtering by page hit data, that we check _every hit in a session_, but when checking for session data, that we _only check the session_ (even if we include every hit from that session in the end results).

This can be accomplished fairly simply: for any query, we filter by page hit data, then session data, then take an intersection of the two. This is what `filtered_sessions.pipe` does.

Endpoints themselves only need to specify the return data. And in some cases, if the filters change the behavior, they made need to be specified (e.g. `api_top_pages` needs `pathname` because otherwise you end up seeing all session data for that path, which doesn't particularly make sense in this case).

Note that `site_uuid` is unique and kept around for query optimization across the board within Clickhouse.

__Changes in this PR__
- Removed our unused filters and pipes; these just add to the mess and maintenance.
- Where possible, moved all varieties of filters into a central filter pipe (`session_data`). This could get renamed to be more descriptive, but I struggled to find a good name, and it would introduce more changes to review.
- Updated tests; minimally, because we want to preserve behavior here. This effort was largely removing the tests that depended on the removed deprecated pipes/endpoints.
- Added readme describing structure and purpose.

The intent here is that we'll have a clearer idea of how to use these and hopefully less issues with consistency across data sets.

---

> [!NOTE]
> Centralizes session vs hit filtering via `filtered_sessions`, simplifies endpoints to rely on it, removes unused top device/browser/os/location endpoints, updates tests, and adds README.
> 
> - **Filtering architecture**:
>   - Introduces clear separation of hit-level (`query_filters`) and session-level (`sessions_filtered_by_session_attributes`) in `filtered_sessions.pipe` with session-level UTM/source/date filters moved here.
> - **Endpoints**:
>   - `api_kpis.pipe`: rely on `filtered_sessions` for filtering; remove per-endpoint date/utm/device/browser/os/location filters; add join in `pathname_pageviews`.
>   - `api_top_pages.pipe`: rely on `filtered_sessions`; keep hit-level `pathname`/`post_uuid`/`post_type` to narrow results.
>   - `api_top_sources` and `api_top_utm_*`: simplify to count from `mv_session_data` joined to `filtered_sessions`; ensure non-empty UTM fields.
> - **Removals**:
>   - Delete deprecated endpoints: `api_top_browsers.pipe`, `api_top_devices.pipe`, `api_top_locations.pipe`, `api_top_os.pipe`.
> - **Tests**:
>   - Remove tests for deleted endpoints; update remaining tests to reflect centralized filtering and expected counts; add combined filter cases (e.g., `pathname` + `source`).
> - **Docs**:
>   - Add `endpoints/README.md` detailing filtering model, placement rules, and endpoint patterns.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 19e791b29a69e5607815417ec3aaadad9787fc57. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ghost/core/core/server/data/tinybird/endpoints/api_kpis.pipe
```diff
@@ -131,27 +120,12 @@ SQL >
                 {% else %}
                     a.date = toDate(toTimezone(timestamp, {{String(timezone, 'Etc/UTC', description="Site timezone", required=True)}}))
                 {% end %}
+            inner join filtered_sessions fs
+                on fs.session_id = h.session_id
```
9larsons, line 124: This should have no significant effect (it didn't change tests) as filtered_sessions should be applying the filters that were removed below.

ghost/core/core/server/data/tinybird/endpoints/api_kpis.pipe
```diff
@@ -79,17 +79,6 @@ SQL >
                 on fs.session_id = sd.session_id
         where
             site_uuid = {{ String(site_uuid, 'mock_site_uuid', description="Tenant ID", required=True) }}
-            {% if defined(date_from) and day_diff(date_from, date_to) == 0 %}
```
9larsons, line 82: This was already joining filtered_sessions and doesn't need the duplicative filters.

ghost/core/core/server/data/tinybird/tests/api_top_pages.yaml
```diff
@@ -123,19 +86,22 @@
   parameters: site_uuid=mock_site_uuid&date_from=2100-01-01&date_to=2100-01-07&timezone=Etc/UTC&utm_medium=social
   expected_result: |
```
9larsons, line 87: UTM test data returns did change because we are properly filtering them at the session level now, meaning we include *all of the pages as part of the session with that UTM data*. If this is incorrect, then we can update the pipe. I have pinged Peter.

PR #25472, opened 2025-11-18
Made domain warming optional based on config values being set

sam-lord:
ref GVA-600

Only enable the domain warming feature if the fallbackDomain & fallbackAddress config values are set. If we don't have those config values, then there isn't a way to do domain warming.


aileen (approved):
LTGM

sam-lord:
Going to re-open this as a new PR, not sure why canary build is failing but this is a relatively simple PR

PR #25478, opened 2025-11-18
Added USE_REACT_SHELL support for e2e tests

rob-ghost:

## Why

As part of the React admin migration (BER-2613), we need to validate that the new React admin shell works correctly by running our e2e test suite against it. This requires infrastructure to serve the React admin build instead of the Ember admin when running tests.

## What

This PR adds support for running e2e tests against the React admin shell by introducing a `USE_REACT_SHELL` environment variable:

- Added `getAdminAssetsPath()` config helper that switches between Ember and React admin based on the environment variable
- Updated admin `app.js` and `controller.js` to use the new config helper
- Added Docker build stage for the React admin (`admin-react-builder`)
- Updated ghost-manager to pass the `USE_REACT_SHELL` env var to the container

### DIFF
--- a/apps/admin/README.md
+++ b/apps/admin/README.md
@@ -13,7 +13,27 @@ Uses an **Ember Bridge** system for smooth migration:
 
 ```bash
 # Start development server
-yarn dev 
+yarn dev
 ```
 
 **Prerequisites:** Ghost and the existing Ember admin must be running on `localhost:2368` for API proxying.
+
+## Building for Production
+
+```bash
+# Build production bundle
+yarn nx run admin:build
+```
+
+This outputs to `apps/admin/dist/`.
+
+## Serving the React Admin
+
+To serve the built React admin instead of the Ember admin, set the `USE_REACT_SHELL` environment variable when starting Ghost from the **monorepo root**:
+
+```bash
+# From the monorepo root
+USE_REACT_SHELL=true yarn dev
+```
+
+This configures the Ghost backend to serve `apps/admin/dist/index.html` at `/ghost/` instead of the Ember admin. The environment variable affects `ghost/core`, not `apps/admin` directly.
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -21,7 +21,7 @@ yarn test:e2e
 
 ### Running Specific Tests
 
-Within `e2e` folder, run one of the following commands: 
+Within `e2e` folder, run one of the following commands:
 
 ```bash
 # All tests
@@ -36,6 +36,18 @@ yarn test --grep "homepage"
 # With browser visible (for debugging)
 yarn test --debug
 ```
+
+### Testing with React Admin Shell
+
+To run e2e tests against the new React admin shell instead of the Ember admin:
+
+From the repository root:
+
+```bash
+USE_REACT_SHELL=true yarn test:e2e
+```
+
+This builds the React admin (`apps/admin`) and configures Ghost to serve it at `/ghost/` instead of the Ember admin.
 
 ## Tests Development
 
--- a/ghost/core/core/shared/config/helpers.js
+++ b/ghost/core/core/shared/config/helpers.js
@@ -98,6 +98,18 @@ const getContentPath = function getContentPath(type) {
 };
 
 /**
+ * Get the path to admin assets, switching between Ember and React admin based on USE_REACT_SHELL env var
+ * @returns {string}
+ */
+const getAdminAssetsPath = function getAdminAssetsPath() {
+    const useReactShell = process.env.USE_REACT_SHELL === 'true';
+    if (useReactShell) {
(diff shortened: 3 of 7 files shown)


apps/admin/README.md
```diff
@@ -13,7 +13,26 @@ Uses an **Ember Bridge** system for smooth migration:
...
+yarn dev
 ```
 
 **Prerequisites:** Ghost and the existing Ember admin must be running on `localhost:2368` for API proxying.
+
+## Building for Production
+
+```bash
+# Build production bundle
+yarn workspace @tryghost/admin vite build
```
jonatansberg: This should probably be something like:
```
yarn nx build @tryghost/admin
```

That way we get all the dependencies built as well.

apps/admin/README.md
```diff
@@ -13,7 +13,26 @@ Uses an **Ember Bridge** system for smooth migration:
...
+```
+
+This outputs to `apps/admin/dist/`.
+
+## Serving the React Admin
+
+To serve the built React admin instead of the Ember admin, set the `USE_REACT_SHELL` environment variable:
+
+```bash
+USE_REACT_SHELL=true yarn dev
```
jonatansberg, line 36: You should probably clarify that this is when you're on dev from the root of the repository? Running your endeavor with that environment variable within the admin package wont do anything.

jonatansberg (approved):
This was a creative solution to get the E2E test to run. There are some things in the README that I think we should update before merging. 

Let's remember to pull some of this out again once we've integrated the admin shell properly.

apps/admin/README.md
rob-ghost: Updated. Some background on why the README ended up with this.

There was some issues running `yarn build` when building `apps/admin` in the `Dockerfile`. The reason for this was the `tsc -b` part of that command needed all the dependancies in order for typescript to compile and that wasn't working at the time.

This then changed to directly running `yarn vite build` which worked. Then this was assumed to be the way to build `apps/admin` for "production" (when really its just the e2e tests).

Since then some of the `COPY` commands were updated in the `Dockerfile` and we never updated the build command.

I just tested both `yarn nx build @tryghost/admin` and `yarn build` and `yarn build` now works.

`yarn nx build` doesn't work in the `Dockerfile` because we're not copying _all_ the required dependancies for the `nx` graph to resolve successfully (we're not copying over `ghost-admin` specifically). This would also produce a double-build in the docker image.

So the updated guide is:

1. Use `yarn nx run admin:build` to build admin in general, resolving all dependencies
2. `Dockerfile` uses `yarn build` because its not setup to work with `nx`, the dependencies are presumed to be pre-built.

apps/admin/README.md
rob-ghost: Adding clarification that the environment variable works at the root because its really for the backend and not `apps/admin`.

e2e/README.md
```diff
@@ -37,6 +37,16 @@ yarn test --grep "homepage"
 yarn test --debug
 ```
 
+### Testing with React Admin Shell
+
+To run e2e tests against the new React admin shell instead of the Ember admin:
+
+```bash
+USE_REACT_SHELL=true yarn test:e2e
```
ibalosh, line 47: I would mention that we run this command from root (since e2e/ doesn't contain that command). 
Or switch it as above one to `USE_REACT_SHELL=true yarn test`

ibalosh (approved):
Added one small comment for running e2e, other than that looks good to me

e2e/README.md
rob-ghost: Its a little confusing because the README starts with talking about commands running from the root but then changes to commands run within this package.

Should we standardise one way or another? I'm in favour of documenting the commands in this package rather than referring to the parent directory, but if we prefer to talk about commands from the root then perhaps all commands should be from the root for consistency?

For now I'll add `From the repository root:` similar to the first `Running Tests` section.

ghost/core/core/shared/config/helpers.js
```diff
@@ -97,6 +97,18 @@ const getContentPath = function getContentPath(type) {
...
+ * Get the path to admin assets, switching between Ember and React admin based on USE_REACT_SHELL env var
+ * @returns {string}
+ */
+const getAdminAssetsPath = function getAdminAssetsPath() {
+    const useReactShell = process.env.USE_REACT_SHELL === 'true';
+    if (useReactShell) {
+        return path.join(this.get('paths:appRoot'), '../../apps/admin/dist');
+    }
+    return this.get('paths:adminAssets');
+};
```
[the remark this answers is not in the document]
rob-ghost: We're choosing not to be overly defensive here. If the path doesn't exist then 💥

e2e/README.md
ibalosh: Good point, I am in favour of documenting commands within package as well, I'll see to clean that up

PR #25479, opened 2025-11-18
Added React/Ember matrix for E2E tests in CI

rob-ghost:
## Summary

- Adds `shell: [ember, react]` dimension to E2E test matrix in CI workflow
- React shell tests use `continue-on-error: true` so they don't block CI
- Sets `USE_REACT_SHELL` environment variable based on matrix selection

**Depends on:** #25478

### DIFF
--- a/.github/workflows/ci-docker.yml
+++ b/.github/workflows/ci-docker.yml
@@ -286,7 +286,7 @@ jobs:
           } >> $GITHUB_STEP_SUMMARY
 
   test_e2e:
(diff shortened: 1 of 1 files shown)


jonatansberg (approved):
Nice!

I've been thinking about how we integrate this with admin deploy job (which is in the main ci workflow) so that we only deploy admin if the E2E tests pass. 

I think it might make sense to either fold this workflow into the main one, or we could turn most of this workflow into reusable actions and then call them from the main workflow 🤔

ibalosh (approved):

PR #25480, opened 2025-11-18
Established Tinybird filter architecture

9larsons:
no ref
- Where possible, moved all varieties of filters into a central filter pipe (session_data). This could get renamed to be more descriptive, but I struggled to find a good name, and it would introduce more changes to review.
- Updated tests; minimally, because we want to preserve behavior here. This effort was largely removing the tests that depended on the removed deprecated pipes/endpoints.
- Added readme describing structure and purpose.

When reviewing the Tinybird endpoints and filter logic, it became clear there was a number of discrepancies that ultimately led to confusion and inconsistency. This PR is an attempt to alleviate that. Let's do some plumbing...

**Overview**
We have several kinds of filters in Tinybird. Let's call them:

always filters [date filters, site uuid, timezone]
(page) hit-level filters [pathname, post_uuid, member_status,
session filters
These need to be treated distinctly.

Always filters are so named because they are always applied - they are our primary and secondary keys in analytics_events: ENGINE_SORTING_KEY "site_uuid, timestamp". They optimize queries by reducing the data set.

Page hit data/filters are those that may change on any particular page_hit. When clicking around, a user is navigating to different pathnames which have unique poat_uuids. If they come to a site and then log in, their member_status may change (currently).

Session data/filters are those that are wrapped up or attributed to the session itself. For the various UTM parameters, we set those based on the first page hit in the session within mv_session_data. Same for source; these are commonly attribution elements.

**Architecture**
With this is mind, we have to differentiate in where and how we filter by these various categories. We need to make sure when filtering by page hit data, that we check every hit in a session, but when checking for session data, that we only check the session (even if we include every hit from that session in the end results).

This can be accomplished fairly simply: for any query, we filter by page hit data, then session data, then take an intersection of the two. This is what filtered_sessions.pipe does.

Endpoints themselves only need to specify the return data. And in some cases, if the filters change the behavior, they made need to be specified (e.g. api_top_pages needs pathname because otherwise you end up seeing all session data for that path, which doesn't particularly make sense in this case).

**Note that site_uuid is unique and kept around for query optimization across the board within Clickhouse.**


ghost/core/core/server/data/tinybird/pipes/filtered_sessions.pipe
```diff
@@ -38,4 +39,11 @@ SQL >
       on qf.session_id = sd.session_id
     where
         site_uuid = {{ String(site_uuid, 'mock_site_uuid', description="Tenant ID", required=True) }}
+        {% if defined(date_from) %} and toDate(toTimezone(first_pageview, {{String(timezone, 'Etc/UTC', description="Site timezone", required=True)}})) >= {{ Date(date_from) }} {% else %} and toDate(toTimezone(first_pageview, {{String(timezone, 'Etc/UTC', description="Site timezone", required=True)}})) >= timestampAdd(today(), interval -7 day) {% end %}
+        {% if defined(date_to) %} and toDate(toTimezone(first_pageview, {{String(timezone, 'Etc/UTC', description="Site timezone", required=True)}})) <= {{ Date(date_to) }} {% else %} and toDate(toTimezone(first_pageview, {{String(timezone, 'Etc/UTC', description="Site timezone", required=True)}})) <= today() {% end %}
```
cmraible: nit: personally I'd prefer to split these conditionals into multiple lines for readability

ghost/core/core/server/data/tinybird/pipes/filtered_sessions.pipe
```diff
@@ -2,7 +2,9 @@ TOKEN "axis" READ
 
 NODE query_filters
```
cmraible: suggestion: `query_filters` is a vague name. The description makes a lot of sense though. Maybe we could rename `query_filters` to `sessions_filtered_by_hit_attributes` to match the node below this?

ghost/core/core/server/data/tinybird/tests/api_top_pages.yaml
```diff
@@ -94,19 +94,22 @@
...
     {"post_uuid":"6b8635fb-292f-4422-9fe4-d76cfab2ba31","pathname":"\/blog\/hello-world\/","visits":1}
 
 - name: Filtered by utm_term - discount
   description: Filtered by utm_term - discount
   parameters: site_uuid=mock_site_uuid&date_from=2100-01-01&date_to=2100-01-07&timezone=Etc/UTC&utm_term=discount
   expected_result: |
+    {"post_uuid":"06b1b0c9-fb53-4a15-a060-3db3fde7b1fc","pathname":"\/about\/","visits":1}
     {"post_uuid":"6b8635fb-292f-4422-9fe4-d76cfab2ba31","pathname":"\/blog\/hello-world\/","visits":1}
 
 - name: Filtered by utm_content - post_123
```
cmraible, line 115: Nice, test updates all make sense. Thanks for splitting the other refactor into its own PR, this makes it much easier to understand what behavior is actually changing here 🙌

cmraible (approved):
Nice one! Took me a few minutes to grok this, but definitely agree it is a cleaner way to handle the filtering. The `README.md` is great for posterity too, thanks for including that.

I left a few optional comments just on naming & readability, but no show stoppers 🙌

ghost/core/core/server/data/tinybird/pipes/filtered_sessions.pipe
9larsons: Yeah I didn't love the names. Sounds good to me.

PR #25482, opened 2025-11-18
Added dual-mode email analytics processing to allow batch processing

9larsons:
- Added batched processing option for the Email Analytics job (enabled via config, default: false)
- Updated test coverage

For particularly large newsletter audiences, the sequential processing of the email analytics job doesn't have enough throughput to be a 'live' measure of email events. A small delay is appropriate, but it's possible to fall an hour+ behind if reaching 100ks of recipients. Batched processing improves throughput, although there's some caveats to the current implementation.

The primary concern with the batched processing introduced here is that the email events, e.g. `EmailDeliveredEvent`) are all emitted at the same time for a batch (up to 300). This results in saturating all the available connections to make updates. While we haven't seen this pose a problem to normal traffic, it could pose a difficulty to a shared db for multiple sites, or for small/low provisioned resource databases. Because of this, we've put it behind config while we refine the behavior, or even redesign the approach altogether.

Batched Processing Mode (when enabled):
  - Pre-fetches all recipients in 1-2 queries instead of N individual lookups (EmailEventProcessor.batchGetRecipients())
  - Accumulates database updates in memory and flushes with SQL CASE statements (EmailEventStorage.flushBatchedUpdates())
  - Processes member stats aggregation in batches of 100 with GROUP BY (queries.aggregateMemberStatsBatch())
  - Reduces queries from ~20,000 to ~68 per 10,000 events (99% reduction)
  - Improves throughput from ~50 to 400-800 events/s (8-16x faster)

Sequential Mode (when disabled):
  - Preserves original behavior: immediate individual queries and updates
  - Maintains backward compatibility

### DIFF
--- a/ghost/core/core/server/services/email-analytics/lib/queries.js
+++ b/ghost/core/core/server/services/email-analytics/lib/queries.js
@@ -201,5 +201,89 @@ module.exports = {
         await db.knex('members')
             .update(updateQuery)
             .where('id', memberId);
+    },
+
+    async aggregateMemberStatsBatch(memberIds) {
+        if (!memberIds || memberIds.length === 0) {
+            return;
+        }
+
+        // Batch query to get stats for all members at once
+        const stats = await db.knex('email_recipients')
+            .leftJoin('emails', 'emails.id', 'email_recipients.email_id')
+            .select(
+                'email_recipients.member_id',
+                db.knex.raw('COUNT(email_recipients.id) as email_count'),
+                db.knex.raw('SUM(CASE WHEN email_recipients.opened_at IS NOT NULL THEN 1 ELSE 0 END) as email_opened_count'),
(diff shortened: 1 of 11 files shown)


ghost/core/core/server/services/email-analytics/lib/queries.js
```diff
@@ -201,5 +201,89 @@ module.exports = {
...
+
+        // Execute batched update with CASE statements
+        await db.knex.raw(`
+            UPDATE members
+            SET
+                email_count = CASE id ${emailCountCases.join(' ')} END,
+                email_opened_count = CASE id ${emailOpenedCountCases.join(' ')} END,
+                email_open_rate = CASE id ${emailOpenRateCases.join(' ')} END
+            WHERE id IN (${memberIds.map(() => '?').join(',')})
+        `, bindings);
```
cmraible, line 287: This is really clever, nice one 🎉

cmraible (approved):
LGTM, ship it!

PR #25381, opened 2025-11-06
Moved member exports and member filtering flow

ibalosh:
ref https://linear.app/ghost/issue/NY-733/port-members-creation-tests

**What**

* Migrated member export flow from legacy browser tests to the new suite
* Migrated member filtering flow
* Improved factory with option to create and build many

**Why**

* Old tests are tightly coupled, dependent on each other, hard to maintain and read

related [PR](https://github.com/TryGhost/Ghost/pull/25339)

PR #25486, opened 2025-11-19
Fixed What's New changelog entry format and tests

rob-ghost:

## Summary

- Fixed changelog entry `featured` field to use string `'true'` instead of boolean
- Added avatar badge indicator for new changelog entries
- Updated What's New banner to use proper anchor link
- Aligned React component data-test attributes with Ember patterns
- Added `.first()` selector for badge/menu locators due to dual-shell element overlap during migration

### DIFF
--- a/e2e/helpers/pages/admin/whats-new/whats-new-menu.ts
+++ b/e2e/helpers/pages/admin/whats-new/whats-new-menu.ts
@@ -11,10 +11,14 @@ export class WhatsNewMenu extends AdminPage {
     constructor(page: Page) {
         super(page);
 
-        this.userMenuTrigger = page.locator('[data-test-nav="arrow-down"]');
-        this.whatsNewMenuItem = page.getByRole('menuitem', {name: 'What’s new?'});
-        this.avatarBadge = page.locator('[data-test-whats-new-avatar-badge]');
-        this.menuBadge = page.locator('[data-test-whats-new-menu-badge]');
+        // TODO: Remove .first() after React shell fully replaces Ember admin
+        // During migration, both shells render the arrow-down icon
+        this.userMenuTrigger = page.locator('[data-test-nav="arrow-down"]').first();
+        this.whatsNewMenuItem = page.getByRole('menuitem', {name: /What’s new\?/i});
+        // TODO: Remove .first() after React shell fully replaces Ember admin
+        // During migration, both shells render badges - .first() selects React's
+        this.avatarBadge = page.locator('[data-test-whats-new-avatar-badge]').first();
+        this.menuBadge = page.locator('[data-test-whats-new-menu-badge]').first();
     }
 
     async openUserMenu(): Promise<void> {
--- a/e2e/tests/admin/whats-new.test.ts
+++ b/e2e/tests/admin/whats-new.test.ts
@@ -2,14 +2,17 @@ import {WhatsNewBanner, WhatsNewMenu} from '@/admin-pages';
 import {expect, test} from '@/helpers/playwright/fixture';
 import type {Page} from '@playwright/test';
 
-interface ChangelogEntry {
+// Local type definition matching the API response format
+type RawChangelogEntry = {
+    slug: string;
     title: string;
     custom_excerpt: string;
     published_at: string;
     url: string;
-    featured: boolean;
+    featured: string;
     feature_image?: string;
-}
+    html?: string;
+};
 
 function daysAgo(days: number): Date {
     const date = new Date();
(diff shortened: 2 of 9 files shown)


e2e/tests/admin/whats-new.test.ts
```diff
@@ -2,14 +2,17 @@ import {WhatsNewBanner, WhatsNewMenu} from '@/admin-pages';
 import {expect, test} from '@/helpers/playwright/fixture';
 import type {Page} from '@playwright/test';
 
-interface ChangelogEntry {
+// Local type definition matching the API response format
+type RawChangelogEntry = {
```
rob-ghost, line 6: I did try importing this from `use-changelog` but it couldn't resolve the import paths correctly (the aliased ones inside `use-changelog`) so I opted to duplicate instead of dealing with that at this time.

e2e/helpers/pages/admin/whats-new/whats-new-menu.ts
```diff
@@ -11,10 +11,15 @@ export class WhatsNewMenu extends AdminPage {
...
-        this.menuBadge = page.locator('[data-test-whats-new-menu-badge]');
+        // TODO: Remove .first() after React shell fully replaces Ember admin
+        // During migration, both shells render the arrow-down icon
+        this.userMenuTrigger = page.locator('[data-test-nav="arrow-down"]').first();
+        // Use regex to match both straight (') and curly (') apostrophes
+        this.whatsNewMenuItem = page.getByRole('menuitem', {name: /What.s new\?/i});
+        // TODO: Remove .first() after React shell fully replaces Ember admin
+        // During migration, both shells render badges - .first() selects React's
+        this.avatarBadge = page.locator('[data-test-whats-new-avatar-badge]').first();
+        this.menuBadge = page.locator('[data-test-whats-new-menu-badge]').first();
```
jonatansberg, line 21: nit: maybe we can migrate to the proper testid format here and do an `.or()` for the Ember one?

jonatansberg (approved):
It does indeed look like they solved the tests failing. I don't understand why you needed to update the Ember acceptance test tho?

e2e/helpers/pages/admin/whats-new/whats-new-menu.ts
ibalosh: We should strive for `getBy..` nomenclature and avoid `.locator` by Playwright, so we should add `testid`. When linting you might even get a warning for this one.

rob-ghost:
@jonatansberg 

> I don't understand why you needed to update the Ember acceptance test tho?

TL;DR: the test fixtures were wrong which caused a bug to go undetected, and fixing the bug required more tests to be updated.

When running the e2e tests in React, we're now using Zod to validate the Changelog response. The e2e fixtures were not passing validation because `featured` is a `stringbool` and not a `boolean`.

Fixing the fixtures in the e2e tests then made the Ember tests fail, because of [this latent bug in the ember service](https://github.com/TryGhost/Ghost/pull/25486/files#diff-2d84694ac4b925d6241d7083e6edcb432cf9e08b6160430d37c6d16a1a1ed7d9R49). Essentially we had a bug where all changelog entry items were considered `featured` because `'true' | 'false'` as string values are always truthy.

So updating the fixtures + ember code then made both the React and Ember e2e tests pass, but then caused the acceptance tests to fail since we now fixed the service it was behaving differently given the fixtures. So that's why we had to update the acceptance tests.

e2e/helpers/pages/admin/whats-new/whats-new-menu.ts
rob-ghost: I had the two options, the `first()` or the `or()`.

I chose `first()` because I didn't want any conditions in the test to handle specifically React or Ember, I wanted the same logic to run on both versions identically.

In this case, Ember uses `data-test-whats-new-menu-badge` and therefore React should use `data-test-whats-new-menu-badge`.

If we want to move to proper testid format, then we'd change it here for **both** Ember and React, and update their markup respectively. But I think our objective in this PR is conformance, then we can update them together later.

rob-ghost:
FYI @ibalosh I thought the merges were blocked on codeowner approval so when I could set auto-merge I assumed everything was approved.

I just noticed you're still reviewing / leaving comments so if you have any feedback or if you wanted this reverted pending review from you or someone on your team please let me know.

Sorry for not waiting 🙏🏻

e2e/helpers/pages/admin/whats-new/whats-new-menu.ts
ibalosh: The way I did this is by adding the `testid` with same name in Ember templates, something like [here](https://github.com/TryGhost/Ghost/pull/25381/files#diff-383f486770f120c439c45ce5f445d952114c6558cc767ba98a3547efae3a5423). That way, test is not going to be changed later, will work for both and once you port it to React, you will just add the new `testid` locators. 

Otherwise I have a feeling we will not change these later in tests. Also on React side, you will not have duplicates to remove later. 

I think there is no good reason to port the old locator format from ember templates to React. Does that make sense?

rob-ghost: > The way I did this is by adding the testid with same name in Ember templates

Yes, I think we should do that as a follow up. The objective here is to have React conform to the current state of Ember as much as possible, and then we can look at amending the markup + tests together to a better format.

If we go an add `testid` now to both places, it makes it harder to reason about if they're conforming or if there's been a test behaviour change from updating the `testid` attributes.

By having React conform to Ember and achieve 🟢 without having to change the Ember side we have greater confidence that React is behaving correctly. Then we have a better foundation for changing the attributes because we'll know that everything is conformant and passing.

PR #25491, opened 2025-11-19
Added member_welcome_emails_enabled setting

troyciesco:
closes https://linear.app/ghost/issue/NY-762

- adds `member_welcome_emails_enabled` setting
- defaults to false
- eventually this will toggle whether member welcome emails will send, instead of the config we currently have set up for testing


troyciesco:
Closing because this is no longer needed - the `automated_emails` table will have a status field that will control this instead, see #25494

PR #25492, opened 2025-11-19
Added member welcome emails toggle to settings

troyciesco:
ref https://linear.app/ghost/issue/NY-757

TK


troyciesco:
Closing because this is no longer needed - the `automated_emails` table will have a status field that will control this instead, see #25494

PR #25493, opened 2025-11-19
Suppress auto-responders from Microsoft email servers for bulk email

sam-lord:
ref [ONC-1193](https://linear.app/ghost/issue/ONC-1193/increase-in-automatic-replies)

We can use this non-standard header (X-Auto-Response-Suppress) to prevent Outlook auto-responders, which should help with certain Outlook configurations not handled by the standard `Auto-Submitted` header.


aileen (approved):
Great! Let's hope this is the final fix for this 🤞

PR #25494, opened 2025-11-20
Added `automated_emails` table

cmraible:
ref https://linear.app/ghost/issue/NY-770/create-a-database-migration-to-add-automated-emails-table

## What

This PR adds a new table to Ghost's database, called `automated_emails`. It also adds a new Claude Code "skill," which provides Claude Code with instructions on how to create a database migration using the slimer CLI tool.

## Why

We are working on adding welcome emails for new members to Ghost. Currently Ghost can be configured to send a welcome email to members as soon as they sign up, but the content of the welcome email is hardcoded. 

Our next step is to allow users to update the content in the welcome email, including the subject and body of the email. This table is where we'll store the content of the welcome email, and any other future automated emails we may add to Ghost.


ghost/core/core/server/data/schema/schema.js
```diff
@@ -1131,5 +1131,17 @@ module.exports = {
         '@@INDEXES@@': [
             ['event_type', 'status', 'created_at']
         ]
+    },
+    automated_emails: {
+        id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+        status: {type: 'string', maxlength: 50, nullable: false, defaultTo: 'inactive'},
```
troyciesco: Commenting separately so I hopefully don't confuse the rabbit, because I do want it to catch this stuff - but I think it's fine not to have the `isIn` validation yet until we've worked on this a bit more and we uncover how exactly we want to use `status`. 

Most likely it'll stay this way, but even just for the `outbox` I've modified it as we've learned stuff so it's nice that there's not validation there currently.

I added a ticket to our backlog to follow up on this towards the end of the project though.

troyciesco (approved):

cmraible:
@troyciesco I've made a few tweaks here:
- Added slug field, as we discussed
- Added an index on slug and status
- Added some validations to the schema for `status`, `sender_email`, `sender_reply_to`

troyciesco (approved):

PR #25499, opened 2025-11-20
Serve Ember admin assets rather than proxy

jonatansberg:
refs https://linear.app/ghost/issue/BER-2989/integrate-new-admin-shell-with-yarn-dev

As a part of integrating the new admin into the main dev workflow we will end up proxying the react admin dev server. Having the react dev server also proxy assets creates a circular request pattern that gets hard to reason about.

This PR tweaks our middleware setup so that we serve the assets directly using `sirv` which is the same static assets server Vite uses internally.

### DIFF
--- a/apps/admin/package.json
+++ b/apps/admin/package.json
@@ -35,6 +35,7 @@
     "globals": "16.4.0",
     "jest-extended": "^6.0.0",
     "jsdom": "24.1.0",
+    "sirv": "3.0.2",
     "typescript": "5.8.3",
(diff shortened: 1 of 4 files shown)


kevinansfield (approved):

kevinansfield:
The failing Ember test _looks_ like it's unrelated/flaky but worth checking before merge - locally it passes so its either flaky or a CI problem

jonatansberg:
@kevinansfield the failing test is is indeed one of the flaky ones.

PR #25503, opened 2025-11-20
Added static components of welcome email settings

peterzimon:
ref https://linear.app/ghost/issue/NY-761/static-settings-ui-for-editing-saving-welcome-emails
ref https://linear.app/ghost/issue/NY-760/static-settings-ui-for-activating-welcome-emails

- As the first step towards automations UI, this PR adds a static version of settings for welcome emails, for both activating and editing email contents.


cmraible (approved):
Looks good, thanks Zimo! 

I think we will need to somehow enable the user to edit the email before enabling it, but we can handle that in a follow up — this is a great start so we can start to wire everything up to the API and start iterating.

PR #25504, opened 2025-11-20
Migrate theme assets to settings app

jonatansberg:
This PR migrates the theme and Zapier assets from Ember to the Settings app. Most assets have been moved. I duplicated the Ghost orbs since they were referenced elsewhere as well. 

This is needed so that we display the correct themes and Zapier integrations in the React admin version of the settings app.


kevinansfield (commented):
This appears to bundle the images into the JS source and displays them with `data:image/png;base64,...` meaning the data cost is paid even if the theme/integration parts of settings aren't accessed. Is that expected?

jonatansberg:
Hmm. No. That's not what I intended. Good catch. I wonder why it doesn't treat them as normal static assets? 🤔

jonatansberg:
Looking at the Vite docs, it seems any assets below 4KB will be inlined by default to avoid the extra network request. You can opt out of that though a flag on the import. 

https://vite.dev/config/build-options#build-assetsinlinelimit
https://vite.dev/guide/assets#explicit-inline-handling

kevinansfield:
👌 seems like a reasonable default for now. Let's leave it as-is, we can always review the config later if needed

kevinansfield (approved):

PR #25507, opened 2025-11-20
Fixed race conditions in E2E tests with React shell

rob-ghost:

While trying to make the React e2e tests required for CI to pass, we also want to make sure there's no flakiness in the test suite, either pre-existing or freshly introduced with the React changes.

This PR was produced by running the E2E test suite 5x times and observing which tests failed at least once. We then fixed these tests in various ways (adding explicit waits, grouping actions and waits in page objects).

We have observed the test suite passing after 5x runs with no failures, so we're confident that this should have addressed flaky tests in the e2e suite as they exist today.


rob-ghost:
@coderabbitai review

ibalosh (approved):
thanks for doing this @rob-ghost 🙌

PR #25511, opened 2025-11-23
fix: Use local data:image/png instead of spacergif.org

SteffoSpieler:
This pull request removes the use of spacergif.org and thus eliminates an external dependency that can go offline. This makes it possible to host Ghost fully GDPR-compliant, as having an external dependency without consent from users will send their information (IP, user agent, etc.) to spacergif.org. As this can't be disabled using a cookie banner, this needs to be removed.

---

> [!NOTE]
> Inline 1x1 PNG data URIs replace external spacergif.org placeholders in Koenig embed/video renderers; tests updated accordingly.
> 
> - **Koenig renderers**
>   - `core/server/services/koenig/node-renderers/embed-renderer.js`: For email video previews, replace spacer image `src` with a base64 `data:image/png` and drop unused spacer width calc.
>   - `core/server/services/koenig/node-renderers/video-renderer.js`: Use base64 `data:image/png` for `poster`/spacer in web and email templates; remove external `spacergif.org` usage.
> - **Tests**
>   - Update expectations in `embed-renderer.test.js` and `video-renderer.test.js` to match inline data URI spacers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7759b0764a696335572472dc06b720924522a130. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/core/core/server/services/koenig/node-renderers/embed-renderer.js
+++ b/ghost/core/core/server/services/koenig/node-renderers/embed-renderer.js
@@ -29,15 +29,15 @@ function renderTemplate(node, document, options) {
     if (isEmail && isVideoWithThumbnail) {
         const emailTemplateMaxWidth = 600;
         const thumbnailAspectRatio = metadata.thumbnail_width / metadata.thumbnail_height;
-        const spacerWidth = Math.round(emailTemplateMaxWidth / 4);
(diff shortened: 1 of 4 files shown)


SteffoSpieler:
In my tests this wasn't an issue: 
grafik
(both are using the local poster)

PR #25518, opened 2025-11-24
Fixed react admin sidebar active state

jonatansberg:
Closes https://linear.app/ghost/issue/BER-3010/fix-sidebar-active-state-for-submenu-items-and-add-tests

- Added getRouteUrl/isRouteActive methods to the Ember bridge that exposes the full route/query param state
- Integrated bridged routing state routing via a controlled isActive state and  new useEmberRouting hook                                                      
- Added comprehensive unit tests

This enables the React admin shell to properly reflect Ember routing state during the Ember-to-React migration phase, including support for "sticky" query params/filters.

### DIFF
--- a/apps/admin/src/ember-bridge/EmberBridge.tsx
+++ b/apps/admin/src/ember-bridge/EmberBridge.tsx
@@ -10,6 +10,7 @@ export type StateBridgeEventMap = {
     emberAuthChange: EmberAuthChangeEvent;
     subscriptionChange: SubscriptionState;
     sidebarVisibilityChange: SidebarVisibilityChangeEvent;
+    routeChange: RouteChangeEvent;
 }
 
 export interface StateBridge {
@@ -19,6 +20,8 @@ export interface StateBridge {
     on<K extends keyof StateBridgeEventMap>(event: K, callback: (event: StateBridgeEventMap[K]) => void): void;
     off<K extends keyof StateBridgeEventMap>(event: K, callback: (event: StateBridgeEventMap[K]) => void): void;
     sidebarVisible: boolean;
+    getRouteUrl: (routeName: string, queryParams?: Record<string, string | null> | null) => string;
+    isRouteActive: (routeNames: string | string[], queryParams?: Record<string, string | null> | null) => boolean;
 }
 
 declare global {
@@ -48,6 +51,13 @@ export interface SubscriptionState {
 export interface SidebarVisibilityChangeEvent {
     isVisible: boolean;
 }
+
+export interface RouteChangeEvent {
+    routeName: string;
+    queryParams: Record<string, unknown>;
+}
+
+export type EmberRouting = Pick<StateBridge, 'getRouteUrl' | 'isRouteActive'>;
 
 /**
  * Maps Ember Data model names to React ResponseType strings.
@@ -226,3 +236,57 @@ export function useSidebarVisibility(): boolean {
         getSidebarVisibility // Server snapshot (same as client for now)
     );
 }
+
+// Default no-op routing for when the bridge isn't available yet
+const defaultRouting: EmberRouting = {
+    getRouteUrl: (routeName) => routeName,
+    isRouteActive: () => false
+};
+
+/**
+ * Hook to access Ember routing state.
+ * Returns routing methods that re-render when Ember's route changes.
+ * 
+ * @example
+ * ```tsx
+ * const routing = useEmberRouting();
+ * const postsUrl = routing.getRouteUrl('posts');
+ * const customUrl = routing.getRouteUrl('posts', {type: 'draft'});
+ * const isActive = routing.isRouteActive('posts', {type: 'draft'});
+ * ```
+ */
+export function useEmberRouting(): EmberRouting {
+    const [bridge, setBridge] = useState<StateBridge | null>(() => window.EmberBridge?.state ?? null);
+    const [, forceUpdate] = useState(0);
+
+    useEffect(() => {
+        // Wait for bridge to be available
+        if (!bridge) {
+            void getStateBridge().then((stateBridge) => {
+                if (stateBridge) {
+                    setBridge(stateBridge);
+                }
+            });
+            return;
+        }
+        
+        // Subscribe to route changes to force re-renders
+        const handleRouteChange = () => {
+            forceUpdate(n => n + 1);
+        };
+        
+        bridge.on('routeChange', handleRouteChange);
+        return () => bridge.off('routeChange', handleRouteChange);
+    }, [bridge]);
+
+    // Return default no-op routing until bridge is available
+    if (!bridge) {
+        return defaultRouting;
+    }
+
+    return {
+        getRouteUrl: bridge.getRouteUrl,
+        isRouteActive: bridge.isRouteActive
+    };
+}
+
--- a/apps/admin/src/layout/app-sidebar/NavMenuItem.tsx
+++ b/apps/admin/src/layout/app-sidebar/NavMenuItem.tsx
@@ -19,17 +19,20 @@ type NavMenuLinkProps = React.ComponentProps<typeof SidebarMenuButton> & {
     target?: string
     rel?: string
     activeOnSubpath?: boolean
+    isActive?: boolean
 };
 function NavMenuLink({
     to,
     target,
     rel,
     activeOnSubpath = false,
+    isActive: controlledIsActive,
     children,
     ...props
 }: NavMenuLinkProps) {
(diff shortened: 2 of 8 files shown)


apps/admin/src/ember-bridge/EmberBridge.tsx
```diff
@@ -226,3 +236,51 @@ export function useSidebarVisibility(): boolean {
...
+    // Return only routing methods
+    if (!bridge) {
+        return null;
+    }
+
+    return {
+        getRouteUrl: bridge.getRouteUrl,
+        isRouteActive: bridge.isRouteActive
+    };
+}
```
rob-ghost: **Suggestion:** I think we could use `useEmberRouting` for this and avoid the `forceUpdate` re-render trick:

```typescript
export function useEmberRouting(): EmberRouting | null {
    const bridge = useSyncExternalStore(
        () => {
            const stateBridge =
  window.EmberBridge?.state;

            if (!stateBridge) return () => {};

            stateBridge.on('routeChange', handleRouteChange);

            return () => stateBridge.off('routeChange', handleRouteChange);
        },
        () => window.EmberBridge?.state ?? null,
        () => null
    );

    if (!bridge) return null;

    return {
        getRouteUrl: bridge.getRouteUrl,
        isRouteActive: bridge.isRouteActive
    };
}
```

ghost/admin/app/services/state-bridge.js
```diff
@@ -195,4 +217,140 @@ export default class StateBridgeService extends Service.extend(Evented) {
...
+                    .reduce((acc, [key, value]) => {
+                        acc[key] = value;
+                        return acc;
+                    }, {});
+
+                const url = this.router.urlFor(routeName, {queryParams: cleanParams});
+                return url;
+            } catch (error) {
+                // eslint-disable-next-line no-console
+                console.error(`Failed to generate URL for route ${routeName}:`, error);
```
rob-ghost: **Question:** Is the convention to use `console.error` but disable the line? Or should we be throwing errors / using some other form of logging?

**Question:** Should we be catching errors here at all? I don't immediately see anything that we would expect to throw an error in the `try` block.

ghost/admin/app/services/state-bridge.js
```diff
@@ -195,4 +217,140 @@ export default class StateBridgeService extends Service.extend(Evented) {
...
+        // Generate URL with controller's current query params
+        try {
+            const params = this._getControllerQueryParams(routeName);
+
+            const cleanParams = Object.entries(params)
+                .filter(([, value]) => value !== null && value !== undefined && value !== '')
+                .reduce((acc, [key, value]) => {
+                    acc[key] = value;
+                    return acc;
+                }, {});
```
rob-ghost: **Suggestion(non-blocking):** This is repeated above and could be a `cleanParams` helper function.

ghost/admin/app/services/state-bridge.js
```diff
@@ -195,4 +217,140 @@ export default class StateBridgeService extends Service.extend(Evented) {
...
+                const value = controller[controllerKey];
+                if (value !== null && value !== undefined) {
+                    params[urlKey] = value;
+                }
+            }
+
+            return params;
+        } catch (error) {
+            return {};
+        }
```
rob-ghost: **Suggestion:** I think we should make error handling consistent in this file. Here we silently return a default, other places we log an error.

**Question:** We're pretty defensive in a lot of this code with wide `try/catch` blocks. Broad question of, should we be so broadly defensive? Or should we allow exceptions to happen. If we want to handle certain expected errors, should we choose to wrap certain lines very explicitly?

I'd personally prefer we don't get defensive, but if we do get defensive I'd like to keep the scope of `try/catch` to places that genuinely might throw and we likely can't control.

rob-ghost (commented):
A couple of suggestions and comments but nothing I'd consider blocking if we wanted to move forward with the code as it is.

Nice 👍🏻

apps/admin/src/ember-bridge/EmberBridge.tsx
```diff
@@ -226,3 +236,51 @@ export function useSidebarVisibility(): boolean {
...
+        };
+        
+        bridge.on('routeChange', handleRouteChange);
+        return () => bridge.off('routeChange', handleRouteChange);
+    }, [bridge]);
+
+    // Return only routing methods
+    if (!bridge) {
+        return null;
+    }
```
rob-ghost, line 285: **Suggestion:** Instead of returning `null` and requiring us to optional-chain and provide defaults in all cases, could we instead return some no-op default functions?

```typescript
return {
    getRouteUrl: (route) => route,
    isRouteActive: () => false
}
```

It seems this is the common fallback pattern:
```typescript
routing?.getRouteUrl('members') ?? 'members'

routing?.isRouteActive('members') ?? false
```

jonatansberg: That's a good idea. I'll make the change 👍

apps/admin/src/ember-bridge/EmberBridge.tsx
jonatansberg: I'm not sure I'm following your suggestion. Your example still has the handle route change callback, which would do what in this scenario? The reason we need the force update thing is so that the consuming components re-render when the state changes on the Ember side.

ghost/admin/app/services/state-bridge.js
jonatansberg: I think we can do one more pass at the error handling here. Some of it has been left in place from previous iterations that worked slightly differently.

ghost/admin/app/services/state-bridge.js
jonatansberg: I ended up restructuring the code to remove the duplication.

ghost/admin/app/services/state-bridge.js
jonatansberg: I'll clean this up.

apps/admin/src/layout/app-sidebar/NavMenuItem.tsx
```diff
@@ -19,17 +19,20 @@ type NavMenuLinkProps = React.ComponentProps<typeof SidebarMenuButton> & {
...
     rel,
     activeOnSubpath = false,
+    isActive: controlledIsActive,
     children,
     ...props
 }: NavMenuLinkProps) {
-    const href = `#/${to}`;
-    const isActive = useIsActiveLink({ path: to, activeOnSubpath });
+    const href = `#/${to?.replace(/^\/?#\//, '')}`;
+    const computedIsActive = useIsActiveLink({ path: to, activeOnSubpath });
```
rob-ghost, line 34: **Daydream (not for now):** I feel like there's something here related to `to` and `isActive` attributes.

In general, we have this combination for `NavMenu.Link` usage:

```typescript
<NavMenuItem.Link 
    to={routing.getRouteUrl(path)}
    isActive={routing.isRouteActive(path, queryParams)}
>
```

Then inside `NavMenuItem.Link` we also check `useIsActiveLink({ path: to, activeOnSubpath });` and this is overridden by `controlledIsActive`.

So generally when we're having a `NavItem.Link` it will have some kind of `to` destination which could be a tuple/object of `[path, queryParams]` and then it would do the `routing.getRouteUrl(path)` and `routing.isRouteActive(path, queryParams)}` inside the component.

This way the usage across the board changes to:
```typescript
<NavMenuItem.Link 
    to="posts"
>

<NavMenuItem.Link 
    to=["posts", { type: "scheduled" }]
>
```

And the `href` URL and `isActive` is handled automatically.

This isn't a suggestion for now but I think there's an ergonomic interface buried here somewhere.

apps/admin/src/ember-bridge/EmberBridge.tsx
rob-ghost: My bad, my snippet wasn't complete. The `useSyncExternalStore` gives a `callback` which is effectively `forceUpdate`:

```typescript
(callback) => {
    const stateBridge = window.EmberBridge?.state;

    if (!stateBridge) return () => {};

    stateBridge.on('routeChange', callback);

    return () => stateBridge.off('routeChange', callback);
},
```

jonatansberg: I think we can leave it as is then. The main point with using `useSyncExternalStore` as far as I've understood it is to prevent state tearing, and since we're anyways not actually using the state I think what's there is fine as well. 

I think there is a more elegant abstraction around how we detect the bridge hiding somewhere in plain sight that would let us get away from a lot of these things. But I don't think it's worth the effort to go and try and figure that out.

PR #25521, opened 2025-11-24
Updated member welcome emails job to emit event instead of processing outbox

troyciesco:
ref https://linear.app/ghost/issue/NY-805

- this makes it so the welcome emails job works similar to the email analytics job, in that it just emits a `StartMemberWelcomeEmailJobEvent`
- The `MemberWelcomeEmailsService` subscribes to that event and then processes the outbox
- This had a positive improvement on CPU utilization, most visible/useful for app servers running many Ghost instances; https://github.com/TryGhost/Ghost/pull/25534 will further improve it by randomizing when the job starts for each instance


cmraible:
I think we should do this, since this is an established pattern in Ghost already and I'd rather stick with an existing convention rather than diverge for this one use-case...but I'm not a huge fan of this pattern and I think we should fix the underlying gaps in the JobManager next so we can skip the worker thread entirely.

As I understand it, this pattern is strictly using bree and the worker thread just to schedule the job to run on the main thread. Spawning a worker (from the main thread) just for the worker to effectively call a function _back on the main thread_ is effectively just a really convoluted and expensive `setTimeout()`. I had a very quick and dirty AI session with the JobManager itself and it doesn't seem like too heavy a left to rip out bree in favor of something that supports scheduling recurring jobs on the main thread, i.e. `node-cron`: https://github.com/TryGhost/framework/pull/308.

ghost/core/core/server/services/member-welcome-emails/index.js
```diff
@@ -8,8 +11,15 @@ class MemberWelcomeEmailsServiceWrapper {
 
         jobs.scheduleMemberWelcomeEmailJob();
 
+        // We currently cannot trigger a non-offloaded job from the job manager
+        // So the member welcome email job simply emits an event that we listen for here
+        // This allows the actual processing to run on the main thread instead of in a worker
+        domainEvents.subscribe(StartMemberWelcomeEmailJobEvent, async () => {
+            await processOutbox();
```
cmraible: What happens if `processOutbox()` takes more than 5 minutes to run? What if it takes 25 minutes to run? Do we end up with 5 concurrent instances of the `processOutbox()` job running?

troyciesco: oh hmm good point - before Bree would only allow one job of a certain type to run at a time, but obviously now the job part is basically just telling this to run

i think we'd end up with multiple instances of it running, but the actual processing is written in such a way that rows won't get picked up by more than one process. But multiple instances still isn't necessarily desirable.

i see email analytics also has a guard so that only one runs at a time - i'll add that here too

cmraible (approved):
Looking good to me, nice one 🚀

PR #25522, opened 2025-11-25
Added ingestion monitoring fields to tinybird materialized view

cmraible:
ref https://linear.app/ghost/issue/NY-748/create-tinybird-mv-to-speed-up-monitoring-endpoints

## Problem
The Tinybird monitoring endpoints (and raw queries used in our monitoring dashboard) are slow enough that they frequently timeout. Part of the problem is the date parsing of the `payload.meta.received_timestamp` field at query time. 

## Solution
This change aims to improve the performance of these endpoints by parsing the `inserted_at` and `received_at` fields and pre-calculating the ingestion latency in the `mv_hits` materialized view, thereby moving the parsing & calculating to ingest time, rather than query time.

Now the monitoring endpoints or any ad-hoc monitoring queries can directly query `mv_hits`, rather than recalculating all these values at query time. 

---

- Adds `inserted_at`, `received_at`, `ingestion_latency_ms` fields to `_mv_hits` datasource with appropriate data types
- Adds above fields to `mv_hits` pipe to populate the materialized view
- Modifies `api_monitoring_ingestion` and `api_monitoring_ingestion_aggregated` to pull from `_mv_hits` instead of `analytics_events`

---

## Deployment
- This doesn't change any landing datasources, so the deployment should be simple
- The changes to the materialized view will be backfilled using the associated pipe
- The endpoints are stateless, so no data will need to change for it

```
» Validating deployment...


* Changes to be deployed:
--------------------------------------------------------
status: modified
name: _mv_hits
type: datasource
path: datasources/_mv_hits.datasource
--------------------------------------------------------
status: modified
name: api_monitoring_ingestion
type: endpoint
path: endpoints/api_monitoring_ingestion.pipe
--------------------------------------------------------
status: modified
name: api_monitoring_ingestion_aggregated
type: endpoint
path: endpoints/api_monitoring_ingestion_aggregated.pipe
--------------------------------------------------------
status: modified
name: mv_hits
type: materialization
path: pipes/mv_hits.pipe
--------------------------------------------------------
* No changes in tokens to be deployed

 Data that will be copied with this deployment:
------------------------------------------------
| Datasource | Backfill type                   |
------------------------------------------------
| _mv_hits   | Using Materialized Pipe mv_hits |
------------------------------------------------


✓ Deployment is valid
```

### DIFF
--- a/ghost/core/core/server/data/tinybird/datasources/_mv_hits.datasource
+++ b/ghost/core/core/server/data/tinybird/datasources/_mv_hits.datasource
@@ -3,7 +3,7 @@ SCHEMA >
     `timestamp` DateTime,
     `received_at` DateTime64(3),
     `inserted_at` DateTime64(3),
-    `ingestion_latency_ms` IntervalMillisecond,
+    `ingestion_latency_ms` Int64,
     `action` LowCardinality(String),
     `version` LowCardinality(String),
     `session_id` String,
--- a/ghost/core/core/server/data/tinybird/datasources/_mv_hits.datasource
+++ b/ghost/core/core/server/data/tinybird/datasources/_mv_hits.datasource
@@ -1,6 +1,7 @@
 SCHEMA >
     `site_uuid` LowCardinality(String),
     `timestamp` DateTime,
+    `received_timestamp` DateTime,
     `action` LowCardinality(String),
     `version` LowCardinality(String),
     `session_id` String,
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/data/tinybird/datasources/_mv_hits.datasource
```diff
@@ -1,6 +1,9 @@
 SCHEMA >
     `site_uuid` LowCardinality(String),
     `timestamp` DateTime,
+    `received_at` DateTime64(3),
+    `inserted_at` DateTime64(3),
+    `ingestion_latency_ms` Int64,
```
cmraible, line 6: - `Int64` allows values between [-9223372036854775808 : 9223372036854775807], which should be plenty for latency in milliseconds, which should be a couple thousand max
- Signed integer instead of `UInt64` because invalid values will be set to -1

9larsons (approved):
On board with all of it! LGTM 🙌

JoeeGrigg (approved):
LGTM

PR #25529, opened 2025-11-25
Added E2E tests for sidebar navigation during Ember-to-React migration

rob-ghost:
ref https://linear.app/ghost/issue/BER-3038

## What does this do?

Adds E2E tests for the admin sidebar and updates markup to support testing with semantic selectors:

- Added E2E tests covering main navigation, posts submenu, custom views, and footer interactions
- Updated markup in both Ember and React sidebars with ARIA attributes (`aria-current`, `aria-expanded`, `aria-label`, `role="menuitem"`) to enable semantic locators
- Removed Ember sidebar when running in React shell

These have been segmented into 3 separate commits for easier reviewing if you want to review each change in isolation.

## Why?

We need stable tests during the Ember-to-React migration. Using semantic selectors based on ARIA attributes allows tests to work with both implementations regardless of markup changes.

## ❌ Known Failing Tests

The React tests for the active nav items are currently failing as the logic is implemented in https://github.com/TryGhost/Ghost/pull/25518. The React e2e tests are still currently optional and do not block merge on failure.

## ⏩ Skipped Dark Mode tests

Dark mode test was skipped because of differences in the markup between Ember and React. We need to decide if we:
1. Keep it skipped until we settle on a markup for dark mode (user menu or sidebar toggle)
2. Conditionally skip it depending on the shell
3. Have two tests, one for Ember and React, skipped conditionally depending on the shell
4. Custom code in the Sidebar page object to deal with the differences in markup

### DIFF
--- a/apps/admin/src/layout/app-sidebar/UserMenu.tsx
+++ b/apps/admin/src/layout/app-sidebar/UserMenu.tsx
@@ -118,14 +118,14 @@ function UserMenu(props: UserMenuProps) {
                     )}
                 </DropdownMenuItem>
                 <DropdownMenuItem className="cursor-pointer text-base" asChild>
-                    <Link to={`/settings/staff/${currentUser.data?.slug}`} role="menuitem">
+                    <Link to={`/settings/staff/${currentUser.data?.slug}`}>
                         <LucideIcon.User />
                         <span>Your profile</span>
                     </Link>
                 </DropdownMenuItem>
                 <DropdownMenuSeparator />
                 <DropdownMenuItem className="cursor-pointer text-base" asChild>
-                    <a href="https://ghost.org/resources?utm_source=admin&utm_campaign=resources" target="_blank" rel="noopener noreferrer" role="menuitem">
+                    <a href="https://ghost.org/resources?utm_source=admin&utm_campaign=resources" target="_blank" rel="noopener noreferrer">
                         <LucideIcon.Book />
                         Resources & guides
                     </a>
--- a/apps/admin/src/layout/app-sidebar/UserMenu.tsx
+++ b/apps/admin/src/layout/app-sidebar/UserMenu.tsx
@@ -41,6 +41,7 @@ function UserMenu(props: UserMenuProps) {
                 <SidebarMenuButton
                     size="lg"
                     className="data-[state=open]:bg-sidebar-accent data-[state=open]:text-sidebar-accent-foreground"
+                    aria-label="User menu"
                 >
                 <div className="relative">
                     <Avatar>
@@ -117,14 +118,14 @@ function UserMenu(props: UserMenuProps) {
                     )}
                 </DropdownMenuItem>
                 <DropdownMenuItem className="cursor-pointer text-base" asChild>
-                    <Link to={`/settings/staff/${currentUser.data?.slug}`}>
+                    <Link to={`/settings/staff/${currentUser.data?.slug}`} role="menuitem">
                         <LucideIcon.User />
                         <span>Your profile</span>
                     </Link>
                 </DropdownMenuItem>
                 <DropdownMenuSeparator />
                 <DropdownMenuItem className="cursor-pointer text-base" asChild>
-                    <a href="https://ghost.org/resources?utm_source=admin&utm_campaign=resources" target="_blank" rel="noopener noreferrer">
+                    <a href="https://ghost.org/resources?utm_source=admin&utm_campaign=resources" target="_blank" rel="noopener noreferrer" role="menuitem">
                         <LucideIcon.Book />
                         Resources & guides
                     </a>
--- a/apps/shade/src/components/ui/sidebar.tsx
+++ b/apps/shade/src/components/ui/sidebar.tsx
@@ -184,6 +184,7 @@ const Sidebar = React.forwardRef<
                         'flex h-full w-[--sidebar-width] flex-col bg-sidebar text-sidebar-foreground',
                         className
                     )}
+                    role="navigation"
                     {...props}
                 >
                     {children}
@@ -198,6 +199,7 @@ const Sidebar = React.forwardRef<
                         className="w-[--sidebar-width] bg-sidebar p-0 text-sidebar-foreground [&>button]:hidden"
                         data-mobile="true"
                         data-sidebar="sidebar"
+                        role="navigation"
                         side={side}
                         style={
             {
@@ -219,6 +221,7 @@ const Sidebar = React.forwardRef<
                 data-side={side}
                 data-state={state}
                 data-variant={variant}
+                role="navigation"
             >
                 {/* This is what handles the sidebar gap on desktop */}
                 <div
--- a/e2e/helpers/pages/admin/posts/custom-view-modal.ts
+++ b/e2e/helpers/pages/admin/posts/custom-view-modal.ts
@@ -1,0 +1,48 @@
+import {Locator, Page} from '@playwright/test';
+
+export class CustomViewModal {
+    private readonly page: Page;
+    public readonly modal: Locator;
+    public readonly nameInput: Locator;
+    public readonly nameError: Locator;
+    public readonly saveButton: Locator;
+    public readonly deleteButton: Locator;
+    public readonly cancelButton: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+        this.modal = page.getByRole('dialog');
+        this.nameInput = page.getByLabel('View name');
+        this.nameError = page.locator('[data-test-error="custom-view-name"]');
+        this.saveButton = this.modal.getByRole('button', {name: 'Save'});
+        this.deleteButton = this.modal.getByRole('button', {name: 'Delete'});
+        this.cancelButton = this.modal.getByRole('button', {name: 'Cancel'});
+    }
+
+    async waitForModal(): Promise<void> {
+        await this.modal.waitFor({state: 'visible'});
+    }
+
+    async enterName(name: string): Promise<void> {
+        await this.nameInput.fill(name);
+    }
+
+    async selectColor(color: string): Promise<void> {
+        await this.page.getByLabel(color).click();
+    }
+
+    async save(): Promise<void> {
+        await this.saveButton.click();
+        await this.modal.waitFor({state: 'hidden'});
+    }
+
+    async delete(): Promise<void> {
+        await this.deleteButton.click();
+        await this.modal.waitFor({state: 'hidden'});
+    }
+
+    async cancel(): Promise<void> {
+        await this.cancelButton.click();
+        await this.modal.waitFor({state: 'hidden'});
+    }
+}
--- a/e2e/helpers/pages/admin/posts/posts-page.ts
+++ b/e2e/helpers/pages/admin/posts/posts-page.ts
@@ -2,9 +2,22 @@ import {AdminPage} from '@/admin-pages';
 import {Locator, Page} from '@playwright/test';
 
 export class PostsPage extends AdminPage {
-    readonly postsList: Locator;
-    readonly postsListItem: Locator;
-    readonly newPostButton: Locator;
+    public readonly postsList: Locator;
+    public readonly postsListItem: Locator;
+    public readonly newPostButton: Locator;
+
+    public readonly postsFilters: Locator;
+
+    public readonly typeFilter: Locator;
+    public readonly visibilityFilter: Locator;
+    public readonly authorFilter: Locator;
+    public readonly tagFilter: Locator;
+    public readonly orderFilter: Locator;
+
+    public readonly saveViewButton: Locator;
+    public readonly editViewButton: Locator;
+
+    public readonly pageTitle: Locator;
 
     constructor(page: Page) {
         super(page);
@@ -13,6 +26,18 @@ export class PostsPage extends AdminPage {
         this.postsList = page.getByTestId('posts-list');
         this.postsListItem = this.postsList.getByTestId('posts-list-item');
         this.newPostButton = page.getByRole('link', {name: 'New post'});
+
+        this.postsFilters = page.getByTestId('posts-filters');
+        this.typeFilter = this.postsFilters.getByRole('button', {name: 'Type filter'});
+        this.visibilityFilter = this.postsFilters.getByRole('button', {name: 'Visibility filter'});
+        this.authorFilter = this.postsFilters.getByRole('button', {name: 'Author filter'});
+        this.tagFilter = this.postsFilters.getByRole('button', {name: 'Tag filter'});
+        this.orderFilter = this.postsFilters.getByRole('button', {name: 'Sort filter'});
+
+        this.saveViewButton = page.getByRole('button', {name: /save as view/i});
+        this.editViewButton = page.getByRole('button', {name: /edit current view/i});
+
+        this.pageTitle = page.getByRole('heading', {level: 2});
     }
 
     getPostByTitle(title: string): Locator {
@@ -21,5 +46,44 @@ export class PostsPage extends AdminPage {
 
     async refreshData() {
         await this.page.reload();
+    }
+
+    async selectType(typeName: string): Promise<void> {
+        await this.typeFilter.click();
+        await this.page.getByRole('option', {name: typeName, exact: true}).click();
+    }
+
+    async selectVisibility(visibilityName: string): Promise<void> {
+        await this.visibilityFilter.click();
+        await this.page.getByRole('option', {name: visibilityName, exact: true}).click();
+    }
+
+    async selectAuthor(authorName: string): Promise<void> {
+        await this.authorFilter.click();
+        await this.page.getByRole('option', {name: authorName, exact: true}).click();
+    }
+
+    async selectTag(tagName: string): Promise<void> {
+        await this.tagFilter.click();
+        await this.page.getByRole('option', {name: tagName, exact: true}).click();
+    }
+
+    async selectOrder(orderName: string): Promise<void> {
+        await this.orderFilter.click();
+        await this.page.getByRole('option', {name: orderName, exact: true}).click();
+    }
+
+    async openSaveViewModal(): Promise<void> {
+        await this.saveViewButton.waitFor({state: 'visible'});
+        await this.saveViewButton.click();
+    }
+
+    async openEditViewModal(): Promise<void> {
+        await this.editViewButton.waitFor({state: 'visible'});
+        await this.editViewButton.click();
+    }
+
+    async getActiveViewName(): Promise<string | null> {
+        return await this.pageTitle.textContent();
     }
 }
--- a/e2e/helpers/pages/admin/sidebar/sidebar-page.ts
+++ b/e2e/helpers/pages/admin/sidebar/sidebar-page.ts
@@ -1,0 +1,69 @@
+import {AdminPage} from '@/admin-pages';
+import {Locator, Page} from '@playwright/test';
+
+/**
+ * SidebarPage uses semantic, accessibility-first locators.
+ * This approach tests the UI as users interact with it, not implementation details.
+ *
+ * Accessibility features:
+ * ✓ Active nav links have aria-current="page"
+ * ✓ Posts toggle button has aria-expanded
+ */
+export class SidebarPage extends AdminPage {
+    public readonly sidebar: Locator;
+    public readonly postsToggle: Locator;
+    public readonly userDropdownTrigger: Locator;
+    public readonly nightShiftToggle: Locator;
+    public readonly whatsNewButton: Locator;
+    public readonly userProfileLink: Locator;
+    public readonly signOutLink: Locator;
+
+    constructor(page: Page) {
+        super(page);
+        this.sidebar = page.getByRole('navigation');
+        this.postsToggle = this.sidebar.getByRole('button', {name: /toggle post views/i});
+        this.userDropdownTrigger = page.locator('[data-test-nav="arrow-down"]');
+        this.nightShiftToggle = page.getByRole('button', {name: /dark mode/i}).or(page.getByRole('menuitem', {name: /dark mode/i}).getByRole('switch'));
+        this.whatsNewButton = page.getByRole('menuitem', {name: /what's new/i});
+        this.userProfileLink = page.getByRole('menuitem', {name: /your profile/i});
+        this.signOutLink = page.getByRole('menuitem', {name: /sign out/i});
+    }
+
+    getNavLink(name: string): Locator {
+        return this.sidebar
+            .getByRole('link')
+            .filter({hasText: new RegExp(name, 'i')});
+    }
+
+    async expandPostsSubmenu(): Promise<void> {
+        const isExpanded = await this.postsToggle.getAttribute('aria-expanded');
+        if (isExpanded !== 'true') {
+            await this.postsToggle.click();
+        }
+    }
+
+    async collapsePostsSubmenu(): Promise<void> {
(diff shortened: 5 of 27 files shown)


rob-ghost:
@coderabbitai review

apps/shade/src/components/ui/sidebar.tsx
```diff
@@ -219,6 +219,7 @@ const Sidebar = React.forwardRef<
                 data-side={side}
                 data-state={state}
                 data-variant={variant}
+                role="navigation"
```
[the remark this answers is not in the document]
rob-ghost: Could be a good improvement to make now, I added `role="navigation"` primarily for e2e tests in this branch. Someone else weigh in if we want to make this change now because we agree its a good thing to do.

apps/admin/src/layout/app-sidebar/UserMenu.tsx
```diff
@@ -117,14 +118,14 @@ function UserMenu(props: UserMenuProps) {
...
-                    <Link to={`/settings/staff/${currentUser.data?.slug}`}>
+                    <Link to={`/settings/staff/${currentUser.data?.slug}`} role="menuitem">
                         <LucideIcon.User />
                         <span>Your profile</span>
                     </Link>
                 </DropdownMenuItem>
                 <DropdownMenuSeparator />
                 <DropdownMenuItem className="cursor-pointer text-base" asChild>
-                    <a href="https://ghost.org/resources?utm_source=admin&utm_campaign=resources" target="_blank" rel="noopener noreferrer">
+                    <a href="https://ghost.org/resources?utm_source=admin&utm_campaign=resources" target="_blank" rel="noopener noreferrer" role="menuitem">
```
jonatansberg: Take a look if we still need the `role="menuitem"`, if so, maybe move it to the `DropdownMenuItem`?

apps/shade/src/components/ui/sidebar.tsx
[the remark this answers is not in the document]
jonatansberg: I think we can go ahead and add this everywhere!

ghost/admin/app/components/gh-nav-menu/main.hbs
```diff
@@ -30,12 +30,12 @@
                 <ul class="gh-nav-list gh-nav-main">
                     {{#if (and (gh-user-can-admin this.session.user))}}
                         <li class="relative">
-                            <LinkTo @route="stats-x">{{svg-jar "graph-chart-up-arrow"}}Analytics</LinkTo>
+                            <LinkTo @route="stats-x" @current-when="stats-x" aria-current={{if this.isOnAnalytics "page"}}>{{svg-jar "graph-chart-up-arrow"}}Analytics</LinkTo>
```
jonatansberg: As per @kevinansfield the `@current-when` is redundant when the `@route` and `@current-when` are the same.

e2e/helpers/pages/admin/sidebar/sidebar-page.ts
```diff
@@ -0,0 +1,128 @@
...
+    public readonly userDropdownTrigger: Locator;
+    public readonly nightShiftToggle: Locator;
+
+    constructor(page: Page) {
+        super(page);
+        this.sidebar = page.getByRole('navigation');
+        // Button has aria-label="Toggle post views" and aria-expanded to convey state
+        this.postsToggle = this.sidebar.getByRole('button', {name: /toggle post views/i});
+        // Submenu is identified by its id, which is linked via aria-controls on the toggle button
+        this.postsSubmenu = this.sidebar.locator('#posts-submenu');
```
jonatansberg: Do we need this locator? Ideally we should access the sub menu items via semantic locators.

ghost/admin/app/components/gh-nav-menu/main.hbs
```diff
@@ -61,17 +61,19 @@
...
                     <GhLinkToCustomViewsIndex @route="posts" @query={{reset-query-params "posts" }}
-                        data-test-nav="posts">{{svg-jar "posts" class="gh-nav-posts-icon"}}Posts</GhLinkToCustomViewsIndex>
+                        data-test-nav="posts" aria-current={{unless this.customViews.activeView "page"}}>{{svg-jar "posts" class="gh-nav-posts-icon"}}Posts</GhLinkToCustomViewsIndex>
                     <LinkTo @route="lexical-editor.new" @model="post" class="gh-secondary-action gh-nav-new-post"
                         @alt="New post" title="New post" data-test-nav="new-story"><span>{{svg-jar "plus"}}</span>
                     </LinkTo>
                     {{#if this.session.user.isAuthorOrContributor}}
                     {{#if this.customViews.forPosts}}
-                    <ul class="gh-nav-view-list">
+                    <ul class="gh-nav-view-list" id="posts-submenu">
```
jonatansberg: Remove this if we can have a better locator

ghost/admin/app/components/gh-nav-menu/main.hbs
```diff
@@ -87,17 +89,19 @@
...
-                        aria-label="{{if this.navigation.settings.expanded.posts " Collapse custom post
-                        types" "Expand custom post types" }}">
+                        aria-label="Toggle post views"
+                        aria-expanded="{{if this.navigation.settings.expanded.posts "true" "false"}}">
                         {{svg-jar (if this.navigation.settings.expanded.posts "arrow-down-stroke"
                         "arrow-right-stroke")}}
                     </button>
                     {{#liquid-if this.navigation.settings.expanded.posts}}
-                    <ul class="gh-nav-view-list">
+                    <ul class="gh-nav-view-list" id="posts-submenu">
```
jonatansberg: Remove this if we can have a better locator

ghost/admin/app/components/gh-nav-menu/main.js
```diff
@@ -44,6 +41,21 @@ export default class Main extends Component.extend(ShortcutsMixin) {
...
+
+    @match('router.currentRouteName', /^(pages|page)/)
+        isOnPages;
+
+    @match('router.currentRouteName', /^(tags|tag)/)
+        isOnTags;
+
+    @match('router.currentRouteName', /^(members|member)/)
+        isOnMembers;
+
```
jonatansberg: Let's override the LinkTo component in Ember rather than re-implementing the same behavior for `@current-when` using these predicates. 

```
// app/components/link-to.js
import LinkComponent from '@ember/routing/link-component';
import { computed } from '@ember/object';

export default LinkComponent.extend({
  attributeBindings: ['ariaCurrent:aria-current'],

  ariaCurrent: computed('isActive', function () {
    return this.isActive ? 'page' : null;
  })
});
```

If the link-to component is used in a context outside of the sidebar where the aria-current attribute wouldn't make sense, then extend it and have a nav-link-to component for the sidebar usage.

e2e/helpers/pages/admin/sidebar/sidebar-page.ts
```diff
@@ -0,0 +1,128 @@
...
+        return this.getNavLink(name).locator('[aria-current="page"]');
+    }
+
+    isCustomViewActive(name: string): Locator {
+        return this.getCustomView(name).locator('[aria-current="page"]');
+    }
+
+    getVisibleItemsForRole(role: UserRole): NavItem[] {
+        return SidebarPage.NAV_ITEMS.filter(item => item.roles.includes(role));
+    }
```
jonatansberg: This seems to be unused

e2e/helpers/pages/admin/sidebar/sidebar-page.ts
```diff
@@ -0,0 +1,128 @@
...
+export class SidebarPage extends AdminPage {
+    public static readonly NAV_ITEMS: NavItem[] = [
+        {name: 'Analytics', path: '/ghost/#/analytics', roles: ['Administrator']},
+        {name: 'View site', path: '/ghost/#/site', roles: ['Administrator', 'Editor']},
+        {name: 'Posts', path: '/ghost/#/posts', roles: ['Administrator', 'Editor', 'Author', 'Contributor']},
+        {name: 'Pages', path: '/ghost/#/pages', roles: ['Administrator', 'Editor']},
+        {name: 'Tags', path: '/ghost/#/tags', roles: ['Administrator', 'Editor']},
+        {name: 'Members', path: '/ghost/#/members', roles: ['Administrator', 'Editor']},
+        {name: 'Settings', path: '/ghost/#/settings', roles: ['Administrator', 'Editor']}
+    ];
```
jonatansberg: This seems unused as well

e2e/helpers/pages/admin/sidebar/sidebar-page.ts
```diff
@@ -0,0 +1,128 @@
...
+        // Filter by text content (not aria-label) to avoid matching external link buttons
+        // Not using exact match to handle cases like "Settings (CTRL/⌘ + ,)"
+        return this.sidebar
+            .getByRole('link')
+            .filter({hasText: new RegExp(name, 'i')});
+    }
+
+    getCustomView(name: string): Locator {
+        return this.postsSubmenu.getByRole('link', {name: new RegExp(name, 'i')});
+    }
```
jonatansberg: We should be able to just use getNavLink

e2e/helpers/pages/admin/sidebar/sidebar-page.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    async clickNavItem(name: string): Promise<void> {
+        await this.getNavLink(name).click();
+    }
+
+    async clickCustomView(name: string): Promise<void> {
+        await this.getCustomView(name).click();
+    }
```
jonatansberg: Just do `.click()` in the test instead

e2e/helpers/pages/admin/sidebar/sidebar-page.ts
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    isNavItemActive(name: string): Locator {
+        return this.getNavLink(name).locator('[aria-current="page"]');
+    }
+
+    isCustomViewActive(name: string): Locator {
+        return this.getCustomView(name).locator('[aria-current="page"]');
+    }
```
jonatansberg: Just use isNavItemActive

e2e/tests/admin/posts/custom-views.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        test('clicking Drafts view - filters posts and shows active state', async ({page}) => {
+            const sidebar = new SidebarPage(page);
+
+            await sidebar.goto('/ghost/#/posts');
+            await sidebar.expandPostsSubmenu();
+
+            await sidebar.clickCustomView('Drafts');
+
+            await expect(page).toHaveURL(/type=draft/);
+            await expect(sidebar.getCustomView('Drafts')).toHaveAttribute('aria-current', 'page');
```
jonatansberg: Use page object helper here (and in the other tests)

e2e/tests/admin/posts/custom-views.test.ts
```diff
@@ -0,0 +1,93 @@
+import {SidebarPage} from '@/admin-pages';
+import {expect, test} from '@/helpers/playwright/fixture';
+
+test.describe('Ghost Admin - Post Custom Views', () => {
```
jonatansberg: This suite seems redundant alongside the existing navigation tests. It would potentially be worthwhile to have tests asserting that you can create, update, and delete the custom views and that this is reflected in the sidebar.

e2e/tests/admin/sidebar/navigation.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    test.describe('main navigation', () => {
+        mainNavItems.forEach(({name, expectedPath}) => {
+            test(`clicking ${name} - navigates and shows active state`, async ({page}) => {
+                const sidebar = new SidebarPage(page);
+
+                await sidebar.goto('/ghost');
+
+                await sidebar.clickNavItem(name);
+
+                await expect(page).toHaveURL(new RegExp(expectedPath.replace(/[#]/g, '\\$&')));
```
jonatansberg: It would be nice if we can clean up this regex and not do a bunch of string manipulation. Tweaking the expected paths above would allow us to do so.

e2e/tests/admin/sidebar/navigation.test.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    test.describe('role-based visibility', () => {
+        test('Administrator - sees Settings link', async ({page}) => {
+            const sidebar = new SidebarPage(page);
+
+            await sidebar.goto('/ghost');
+
+            await expect(sidebar.getNavLink('Settings')).toBeVisible();
+        });
+    });
```
jonatansberg: This seems incomplete. We don't have any roles in the current test setup, so maybe we can mark this as a to-do to flesh it out?

e2e/tests/admin/sidebar/navigation.test.ts
```diff
@@ -0,0 +1,161 @@
...
+            const sidebar = new SidebarPage(page);
+
+            await sidebar.goto('/ghost');
+
+            // React has night shift toggle inside dropdown, Ember has it outside
+            // For React, open dropdown to access toggle. For Ember, toggle is always visible.
+            const isReactShell = process.env.USE_REACT_SHELL === 'true';
+            if (isReactShell) {
+                await sidebar.openUserDropdown();
+            }
```
jonatansberg: I think we can remove this conditional since opening the user menu in Ember wouldn't interfere with toggling night shift / dark mode.

jonatansberg (commented):
Here are the comments from or mob review session

e2e/helpers/pages/admin/posts/custom-view-modal.ts
```diff
@@ -0,0 +1,48 @@
...
+    public readonly nameError: Locator;
+    public readonly saveButton: Locator;
+    public readonly deleteButton: Locator;
+    public readonly cancelButton: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+        this.modal = page.getByRole('dialog');
+        this.nameInput = page.getByLabel('View name');
+        this.nameError = page.locator('[data-test-error="custom-view-name"]');
```
ibalosh, line 16: `nit` I assume later this one will end up being `page.getByTestId("error-customer-view")` or something like that

e2e/helpers/pages/admin/posts/posts-page.ts
```diff
@@ -13,13 +24,75 @@ export class PostsPage extends AdminPage {
...
+
+        this.pageTitle = page.getByRole('heading', {level: 2});
     }
 
     getPostByTitle(title: string): Locator {
         return this.postsListItem.filter({has: this.page.getByRole('heading', {name: title, exact: true, level: 3})});
     }
 
+    async waitForPostsToLoad(): Promise<void> {
+        await Promise.race([
```
ibalosh: `nit:` maybe we could describe the intention of this wait function with a comment, so it's clear why we handle the race condition and silence rejections

ibalosh: `nit` and maybe make it even clearer with named locators:

```
await Promise.race([
    firstPost.waitFor({state: 'visible'}).catch(() => {}),
    emptyState.waitFor({state: 'visible'}).catch(() => {})
]);
```

e2e/helpers/pages/admin/posts/posts-page.ts
```diff
@@ -13,13 +24,75 @@ export class PostsPage extends AdminPage {
...
+            this.postsListItem.first().waitFor({state: 'visible'}).catch(() => {}),
+            this.page.getByText('No posts match the current filter').waitFor({state: 'visible'}).catch(() => {})
+        ]);
+    }
+
     async refreshData() {
         await this.page.reload();
```
ibalosh, line 48: `nit` little bit hidden (I am guilty as charged with too long inheritance path :) ) , you can use from AdminPage -> BasePage parent `this.refresh()` method if that makes sense

e2e/helpers/pages/admin/sidebar/sidebar-page.ts
```diff
@@ -0,0 +1,73 @@
...
+    public readonly whatsNewButton: Locator;
+    public readonly userProfileLink: Locator;
+    public readonly signOutLink: Locator;
+
+    constructor(page: Page) {
+        super(page);
+        this.sidebar = page.getByRole('navigation');
+        this.postsToggle = this.sidebar.getByRole('button', {name: /toggle post views/i});
+        this.userDropdownTrigger = page.locator('[data-test-nav="arrow-down"]');
+        this.nightShiftToggle = page.getByRole('button', {name: /dark mode/i}).or(page.getByRole('menuitem', {name: /dark mode/i}).getByRole('switch'));
```
ibalosh, line 26: `nit` since we use `or` for locating here, will this always work, will it find correct locators (I assume so just a double check here)

e2e/helpers/pages/admin/sidebar/sidebar-page.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    getNavLink(name: string): Locator {
+        return this.sidebar
+            .getByRole('link')
+            .filter({hasText: new RegExp(name, 'i')});
+    }
+
+    getCustomViewColorIndicator(viewName: string): Locator {
+        return this.getNavLink(viewName).locator('[data-color]');
```
ibalosh, line 39: can we use more friendly way of locating then `.locator` here, no way to go by getByRole or TestId route?

e2e/tests/admin/sidebar/navigation.test.ts
```diff
@@ -0,0 +1,152 @@
...
+];
+
+test.describe('Ghost Admin - Sidebar Navigation', () => {
+    test.describe('main navigation', () => {
+        NAV_ITEMS.forEach(({name, path}) => {
+            test(`clicking ${name} - navigates and shows active state`, async ({page}) => {
+                const sidebar = new SidebarPage(page);
+
+                await sidebar.goto('/ghost');
+
```
ibalosh, line 28: `mega mega small nit` probably not needed empty line

e2e/tests/admin/sidebar/navigation.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    {name: 'Members', path: /\/ghost\/#\/members\/?$/, roles: ['Administrator', 'Editor']}
+];
+
+test.describe('Ghost Admin - Sidebar Navigation', () => {
+    test.describe('main navigation', () => {
+        NAV_ITEMS.forEach(({name, path}) => {
+            test(`clicking ${name} - navigates and shows active state`, async ({page}) => {
+                const sidebar = new SidebarPage(page);
+
+                await sidebar.goto('/ghost');
```
ibalosh, line 27: I am wondering if these goto visits can go to sidebar to a method there, so that we don't have hard coded url's here, but visit with parameter options in sidebar

ghost/admin/app/components/gh-nav-menu/main.hbs
```diff
@@ -97,9 +98,11 @@
                         {{#each this.customViews.forPosts as |view|}}
                         <li>
                             <LinkTo @route="posts" @query={{reset-query-params "posts" view.filter}}
-                                data-test-nav-custom="{{view.route}}-{{view.name}}" title="{{view.name}}">
+                                data-test-nav-custom="{{view.route}}-{{view.name}}" title="{{view.name}}"
+                                @activeClass="active"
+                                aria-current={{if (and this.customViews.activeView (eq view.name this.customViews.activeView.name)) "page"}}>
                                 <span class="gh-nav-viewname">{{view.name}}</span>
-                                <span class="flex items-center svg-{{view.color}}">
+                                <span class="flex items-center svg-{{view.color}}" data-color="{{view.color}}" aria-hidden="true">
```
ibalosh, line 105: is this test only attribute, if so why not `data-testid`?

ghost/admin/app/components/modals/custom-view-form.hbs
```diff
@@ -1,9 +1,9 @@
-<div class="modal-content">
-    <header class="modal-header" data-test-modal="custom-view-form">
-        <h1>{{if @data.customView.isNew "New view" "Edit view"}}</h1>
+<div class="modal-content" role="dialog" aria-modal="true" aria-labelledby="custom-view-modal-title" data-test-modal="custom-view-form">
```
ibalosh, line 1: here too, wondering if it's test only locator why not `data-test-id`

ibalosh (approved):
Couple of nitpicks and questions around locator choices (data-testids mostly). I don't have strong feelings against it though

ghost/admin/app/components/modals/custom-view-form.hbs
rob-ghost: Pre-existing test attribute for acceptance tests, removing it makes them fail. Not being used in the e2e tests.

e2e/helpers/pages/admin/posts/posts-page.ts
rob-ghost: This line isn't in the changeset but good to know 👍🏻 I'll look to update it while I'm actioning other comments.

e2e/helpers/pages/admin/sidebar/sidebar-page.ts
rob-ghost: This is trying to work around the different markup for React and Ember (hence the `or`). They have different roles / labels.

apps/admin/src/layout/app-sidebar/NavContent.tsx
```diff
@@ -66,7 +65,7 @@ function NavContent({ ...props }: React.ComponentProps<typeof SidebarGroup>) {
                     </NavMenuItem>
 
                     {/* Posts submenu */}
-                    <NavSubMenu isExpanded={postsExpanded} id="posts-submenu">
```
rob-ghost, line 69: Revert this

e2e/helpers/pages/admin/posts/posts-page.ts
rob-ghost: We're going to change this approach to wait for the filters themselves to settle loading before proceeding with interactions, instead of waiting for posts to load or the pages loading state to resolve.

e2e/helpers/pages/admin/posts/custom-view-modal.ts
rob-ghost: The challenge with the error component is how widespread its used. In order to update it to use a semantic selector the blast radius would be quite large.

So we're sticking with the older conventional `data-test-error` for now.

A future improvement would be to add `aria-describedby` (or `aria-errormessage`) to the input pointing at an error element identified with an `id`.

e2e/helpers/pages/admin/sidebar/sidebar-page.ts
rob-ghost: For all the comments relating to `data-color`, `[data-color]` is a semantic data attribute used to identify the colour indicator. There's no appropriate aria role for colours like this and its better than using a `data-testid`. So its not perfect, but its the best option we have for a semantic selector for this colour indicator.

e2e/tests/admin/sidebar/navigation.test.ts
rob-ghost: Valid suggestion, but I'll defer it for now.

e2e/helpers/pages/admin/posts/posts-page.ts
rob-ghost: The change to align this was out-of-scope of this PR, but a good cleanup 🪣 we can do later

PR #25532, opened 2025-11-25
Replaced `esm` patch with compatibility check in Ember build

kevinansfield:
no issue

We'd like to avoid having any `postinstall` scripts as its best security practice to have npm install scripts disabled. The `esm` patch that we install with `patch-package` via `postinstall-postinstall` was only necessary for a specific range of Node versions (22.10.0-22.17.x) and is needed during the Ember Admin build that only impacts core developers who can more easily switch to a compatible Node version.

- added a node version compatibility check to `ember-cli-build.js` that is called on all Ember builds, outputting a helpful message when we detect an incompatible Node version being used before any build error can be hit
- removed the `esm` patch and associated packages
- bumped Node version to 22.18.0 in CI to support Ember Admin builds


sam-lord (approved):
Looks great!

PR #25538, opened 2025-11-26
Fixed unfollow action failing in ActivityPub

minimaluminium:
ref https://linear.app/ghost/issue/BER-2924/clicking-unfollow-on-a-profile-flashes-and-then-says-youre-still

The unfollow API returns a `202 Accepted` status with an empty response body, but the fetchJSON method was only handling `204 No Content` responses. This caused the client to attempt parsing the empty response as JSON, throwing an error and triggering the mutation error callback, which reverted the button state back to "Following".

### DIFF
--- a/apps/activitypub/package.json
+++ b/apps/activitypub/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/activitypub",
-  "version": "2.0.1",
+  "version": "2.0.2",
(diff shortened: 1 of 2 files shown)


lsinger (approved):
Thanks for taking care of this -- and thank you also for the great explanation in the PR description, that was really helpful context.

PR #25540, opened 2025-11-26
Split React and Ember E2E tests to make React tests optional

rob-ghost:
## Why?

The React E2E tests in `ci-docker.yml` are currently experimental as we migrate from Ember to React. These tests were configured with `continue-on-error` at the matrix level, but this doesn't prevent the overall workflow from being marked as failed when React tests fail.

## What does it do?

This PR splits the combined `test_e2e` job into two separate jobs:

- **`test_e2e_ember`** - Required tests that must pass for CI to succeed
- **`test_e2e_react`** - Optional tests with `continue-on-error: true` that run but don't block the workflow

Additionally, a new **`required_tests`** job acts as the final gate, only checking Ember test results to determine the workflow's overall status.

## Why is this needed?

Developers shouldn't be blocked by experimental React E2E test failures while the migration is in progress. This change allows:
- React tests to run and provide feedback
- React test failures to be visible in the workflow UI
- The overall CI workflow to pass when Ember tests succeed, regardless of React test results

| Ember Tests | React Tests | Workflow Status |
|-------------|-------------|-----------------|
| Pass        | Pass        | ✅ Success      |
| Pass        | Fail        | ✅ Success      |
| Fail        | Pass        | ❌ Failure      |
| Fail        | Fail        | ❌ Failure      |

### DIFF
--- a/.github/workflows/ci-docker.yml
+++ b/.github/workflows/ci-docker.yml
@@ -285,8 +285,8 @@ jobs:
 
           } >> $GITHUB_STEP_SUMMARY
 
-  test_e2e:
-    name: E2E Tests (${{ matrix.shell == 'react' && 'React' || 'Ember' }} ${{ matrix.shardIndex }}/${{ matrix.shardTotal }})
+  test_e2e_ember:
+    name: E2E Tests (Ember ${{ matrix.shardIndex }}/${{ matrix.shardTotal }})
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


rob-ghost:
Closing this PR as the CI status is already correct when only React e2e tests fail and this PR makes no difference to that. PRs having "Some checks were not successful" is a separate, non-blocking issue that is just a bit counter-intuitive.

PR #25552, opened 2025-11-27
Integrate local E2E test with dev environment

jonatansberg:
I've extended the default E2E test fixtures to detect the presence of containers initialized through `yarn dev`. When those containers are present we can reuse existing resources and use the lightweight dev backend image and gateway setup for E2E test isolation.

Since the `yarn dev` command won't start Tinybird by default I've split the default project into a separate analytics project and main admin project so that we can conditionally run those tests. In CI, the tests will still fail if Tinybird isn't available. 

Note: This PR includes the changes from #25551, which should be merged first.

### DIFF
--- a/e2e/helpers/environment/constants.ts
+++ b/e2e/helpers/environment/constants.ts
@@ -40,3 +40,54 @@ export const MAILPIT = {
     PORT: 1025
 };
 
+/**
+ * Configuration for dev environment mode.
+ * Used when yarn dev infrastructure is detected.
+ */
+export const DEV_ENVIRONMENT = {
+    projectNamespace: 'ghost-dev',
+    networkName: 'ghost_dev'
+} as const;
+
+export const TEST_ENVIRONMENT = {
+    projectNamespace: 'ghost-dev-e2e',
+    gateway: {
+        image: 'ghost-dev-ghost-dev-gateway'
+    },
+    ghost: {
+        image: 'ghost-dev-ghost-dev',
+        workdir: '/home/ghost/ghost/core',
+        port: 2368,
+        env: [
+            // Environment configuration
+            'NODE_ENV=development',
+            'server__host=0.0.0.0',
+            `server__port=2368`,
+
+            // Database configuration (database name is set per container)
+            'database__client=mysql2',
+            `database__connection__host=ghost-dev-mysql`,
+            `database__connection__port=3306`,
+            `database__connection__user=root`,
+            `database__connection__password=root`,
+
+            // Redis configuration
+            'adapters__cache__Redis__host=ghost-dev-redis',
+            'adapters__cache__Redis__port=6379',
+
+            // Email configuration
+            'mail__transport=SMTP',
+            'mail__options__host=ghost-dev-mailpit',
+            'mail__options__port=1025',
+
+            // Public assets via gateway (same as compose.dev.yaml)
+            'portal__url=/ghost/assets/portal/portal.min.js',
+            'comments__url=/ghost/assets/comments-ui/comments-ui.min.js',
+            'sodoSearch__url=/ghost/assets/sodo-search/sodo-search.min.js',
+            'sodoSearch__styles=/ghost/assets/sodo-search/main.css',
+            'signupForm__url=/ghost/assets/signup-form/signup-form.min.js',
+            'announcementBar__url=/ghost/assets/announcement-bar/announcement-bar.min.js'
+        ]   
+    }
+} as const;
+
--- a/e2e/helpers/environment/environment-factory.ts
+++ b/e2e/helpers/environment/environment-factory.ts
@@ -1,0 +1,19 @@
+import {DevEnvironmentManager} from './dev-environment-manager';
+import {EnvironmentManager} from './environment-manager';
+import {isDevEnvironmentAvailable} from './service-availability';
+
+// Cached manager instance (one per worker process)
+let cachedManager: EnvironmentManager | DevEnvironmentManager | null = null;
+
+/**
+ * Get the environment manager for this worker.
+ * Creates and caches a manager on first call, returns cached instance thereafter.
+ */
+export async function getEnvironmentManager(): Promise<EnvironmentManager | DevEnvironmentManager> {
+    if (!cachedManager) {
+        const useDevEnv = await isDevEnvironmentAvailable();
+        cachedManager = useDevEnv ? new DevEnvironmentManager() : new EnvironmentManager();
+    }
+    return cachedManager;
+}
+
--- a/e2e/helpers/environment/service-availability.ts
+++ b/e2e/helpers/environment/service-availability.ts
@@ -1,0 +1,85 @@
+import Docker from 'dockerode';
+import baseDebug from '@tryghost/debug';
+import {DEV_ENVIRONMENT, TINYBIRD} from './constants';
+
+const debug = baseDebug('e2e:ServiceAvailability');
+
+/**
+ * Find running Tinybird containers for a specific Docker Compose project.
+ */
+async function isServiceAvailable(docker: Docker, serviceName: string) {
+    const containers = await docker.listContainers({
+        filters: {
+            label: [
+                `com.docker.compose.service=${serviceName}`,
+                `com.docker.compose.project=${DEV_ENVIRONMENT.projectNamespace}`
+            ],
+            status: ['running']
+        }
+    });
+    return containers.length > 0;
+}
+
+export async function isDevNetworkAvailable(docker: Docker): Promise<boolean> {
+    try {
+        const networks = await docker.listNetworks({
+            filters: {name: [DEV_ENVIRONMENT.networkName]}
+        });
+
+        if (networks.length === 0) {
+            debug('Dev environment not available: network not found');
+            return false;
+        }
+        debug('Dev environment is available');
+        return true;
+    } catch (error) {
+        debug('Error checking dev environment:', error);
+        return false;
+    }
+}
+
+/**
+ * Check if the dev environment (yarn dev) is running.
+ * Detects by checking for the ghost_dev network and running MySQL container.
+ */
+export async function isDevEnvironmentAvailable(): Promise<boolean> {
+    const docker = new Docker();
+
+    if (!await isDevNetworkAvailable(docker)) {
+        debug('Dev environment not available: network not found');
+        return false;
+    }
+
+    if (!await isServiceAvailable(docker, 'mysql')) {
+        debug('Dev environment not available: MySQL container not running');
+        return false;
+    }
+
+    if (!await isServiceAvailable(docker, 'redis')) {
+        debug('Dev environment not available: Redis container not running');
+        return false;
+    }
+
+    if (!await isServiceAvailable(docker, 'mailpit')) {
+        debug('Dev environment not available: Mailpit container not running');
+        return false;
+    }
+
+    return true;
+}
+
+// Cache availability checks per process
+const tinybirdAvailable: boolean | null = null;
+
+/**
+ * Check if Tinybird is running.
+ * Checks for tinybird-local service in ghost-dev compose project.
+ */
+export async function isTinybirdAvailable(): Promise<boolean> {
+    if (tinybirdAvailable !== null) {
+        return tinybirdAvailable;
+    }
+
+    const docker = new Docker();
+    return isServiceAvailable(docker, TINYBIRD.LOCAL_HOST);
+}
--- a/e2e/helpers/environment/service-managers/dev-ghost-manager.ts
+++ b/e2e/helpers/environment/service-managers/dev-ghost-manager.ts
@@ -1,0 +1,318 @@
+import Docker from 'dockerode';
+import baseDebug from '@tryghost/debug';
+import path from 'path';
+import {DEV_ENVIRONMENT, TEST_ENVIRONMENT, TINYBIRD} from '@/helpers/environment/constants';
+import {fileURLToPath} from 'url';
+import {isTinybirdAvailable} from '@/helpers/environment/service-availability';
+import type {Container, ContainerCreateOptions} from 'dockerode';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const debug = baseDebug('e2e:DevGhostManager');
+
+export interface DevGhostManagerConfig {
+    workerIndex: number;
+}
+
+/**
+ * Manages Ghost and Gateway containers for dev environment mode.
+ * Creates worker-scoped containers that persist across tests.
+ */
+export class DevGhostManager {
+    private readonly docker: Docker;
+    private readonly config: DevGhostManagerConfig;
+    private ghostContainer: Container | null = null;
+    private gatewayContainer: Container | null = null;
+
+    constructor(config: DevGhostManagerConfig) {
+        this.docker = new Docker();
+        this.config = config;
+    }
+
+    get ghostContainerId(): string | null {
+        return this.ghostContainer?.id ?? null;
+    }
+
+    get gatewayContainerId(): string | null {
+        return this.gatewayContainer?.id ?? null;
+    }
+
+    getGatewayPort(): number {
+        return 30000 + this.config.workerIndex;
+    }
+
+    async setup(): Promise<void> {
+        debug(`Setting up containers for worker ${this.config.workerIndex}...`);
+
+        const ghostName = `ghost-e2e-worker-${this.config.workerIndex}`;
+        const gatewayName = `ghost-e2e-gateway-${this.config.workerIndex}`;
+
+        // Try to reuse existing containers (handles process restarts after test failures)
+        this.ghostContainer = await this.getOrCreateContainer(ghostName, () => this.createGhostContainer(ghostName));
+        this.gatewayContainer = await this.getOrCreateContainer(gatewayName, () => this.createGatewayContainer(gatewayName, ghostName));
+
+        debug(`Worker ${this.config.workerIndex} containers ready`);
+    }
+
+    /**
+     * Get existing container if running, otherwise create new one.
+     * This handles Playwright respawning processes after test failures.
+     */
+    private async getOrCreateContainer(name: string, create: () => Promise<Container>): Promise<Container> {
+        try {
+            const existing = this.docker.getContainer(name);
+            const info = await existing.inspect();
+            
+            if (info.State.Running) {
+                debug(`Reusing running container: ${name}`);
+                return existing;
+            }
+            
+            // Exists but stopped - start it
+            debug(`Starting stopped container: ${name}`);
+            await existing.start();
+            return existing;
+        } catch {
+            // Doesn't exist - create new
+            debug(`Creating new container: ${name}`);
+            const container = await create();
+            await container.start();
+            return container;
+        }
+    }
+
+    async teardown(): Promise<void> {
+        debug(`Tearing down worker ${this.config.workerIndex} containers...`);
+
+        if (this.gatewayContainer) {
+            await this.removeContainer(this.gatewayContainer);
+            this.gatewayContainer = null;
+        }
+        if (this.ghostContainer) {
+            await this.removeContainer(this.ghostContainer);
+            this.ghostContainer = null;
+        }
+
+        debug(`Worker ${this.config.workerIndex} containers removed`);
+    }
+
+    async restartWithDatabase(databaseName: string, extraConfig?: Record<string, string>): Promise<void> {
+        if (!this.ghostContainer) {
+            throw new Error('Ghost container not initialized');
+        }
+
+        debug('Restarting Ghost with database:', databaseName);
+
+        const info = await this.ghostContainer.inspect();
+        const containerName = info.Name.replace(/^\//, '');
+
+        // Remove old and create new with updated database
+        await this.removeContainer(this.ghostContainer);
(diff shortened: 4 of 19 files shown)


e2e/helpers/environment/service-managers/dev-ghost-manager.ts
```diff
@@ -0,0 +1,359 @@
...
+                }
+            },
+            Labels: {
+                'com.docker.compose.project': this.config.projectNamespace,
+                'tryghost/e2e': 'ghost-dev'
+            }
+        };
+
+        return this.docker.createContainer(config);
+    }
```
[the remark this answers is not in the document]
jonatansberg: This is intentional since we want to use the shared volume from the main ghost-dev project.

e2e/helpers/environment/service-managers/dev-ghost-manager.ts
```diff
@@ -0,0 +1,359 @@
...
+            await existing.start();
+            return existing;
+        } catch {
+            // Doesn't exist - create new
+            debug(`Creating new container: ${name}`);
+            const container = await create();
+            await container.start();
+            return container;
+        }
+    }
```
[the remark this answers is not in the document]
jonatansberg: This is a non-issue in practice.

e2e/helpers/environment/service-availability.ts
```diff
@@ -0,0 +1,80 @@
...
+            return false;
+        }
+
+        // Check if healthy
+        const container = containers[0];
+        if (!container.Status.includes('healthy')) {
+            debug('Tinybird not available: container not healthy');
+            tinybirdAvailable = false;
+            return false;
+        }
```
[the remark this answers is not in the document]
jonatansberg: Removing the health check as we actually only care about existence at this point.

e2e/helpers/environment/service-managers/dev-ghost-manager.ts
```diff
@@ -0,0 +1,359 @@
...
+                const logs = await container.logs({stdout: true, stderr: true});
+                debug('Migration logs:', logs.toString());
+            } catch {
+                // Container may already be removed
+            }
+            throw new Error(`Migrations failed with exit code ${result.StatusCode}`);
+        }
+
+        debug('Migrations completed successfully');
+    }
```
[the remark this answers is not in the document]
jonatansberg: Fixed

e2e/helpers/environment/service-availability.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const docker = new Docker();
+
+    try {
+        const containers = await docker.listContainers({
+            filters: {
+                name: [TINYBIRD.LOCAL_HOST],
+                status: ['running']
+            }
+        });
```
rob-ghost: **Issue:** `TINYBIRD.LOCAL_HOST` is defined as `tinybird-local` but `yarn dev:analytics` spins up `ghost-dev-tinybird` so running `yarn dev:analytics` and then running the tests still skips over the tinybird tests.

rob-ghost (approved):
I've run the following scenarios locally and all but one passed:

1. ✅ Running with `yarn dev:forward` - the running containers are used, analytics tests are skipped
2. ✅ Running standalone - e2e spins up the required containers (including tinybird)

I did find that running `yarn dev:analytics` which spins up tinybird and then running the tests they were still skipped even though the containers are available. I think this is intentional but I also think it would be possible to use them with a slight tweak to the service availability module. See [this comment](https://github.com/TryGhost/Ghost/pull/25552/files#r2577098442).

Assuming that might be out-of-scope, PR approved 👍🏻 (pending linting is fixed and the base branch is merged)

e2e/helpers/environment/service-availability.ts
jonatansberg: We should align those so we can reuse the Tinybird instance.

jonatansberg:
Oh man, that's a lot of linting things. I think my local ESLint setup is broken for the e2e project for some reason...

Edit: I've fixed them now.

e2e/helpers/environment/constants.ts
```diff
@@ -40,3 +40,30 @@ export const MAILPIT = {
     PORT: 1025
 };
 
+/**
+ * Configuration for dev environment mode.
+ * Used when yarn dev:forward infrastructure is detected.
+ */
+export const DEV_ENVIRONMENT = {
```
ibalosh, line 47: To avoid confusion, these seem to make more sense to be a separate file, so we have something like:

* constants.test-env.ts
* constants.dev-end.ts

we will have some duplicates, but cleaner separation between environment configurations and cleaner switch between them in future. At the moment, this looks confusing, with having configurations overlap in a single file and no clear shape, which will differ depending on file from which you are loading them

e2e/helpers/environment/service-availability.ts
jonatansberg: I tried fixing this. Now it works great for the `dev:analytics` case, but it's for some unknown reason a bit racy for the standalone E2E tests. I'll come back to this tomorrow.

e2e/helpers/environment/service-availability.ts
```diff
@@ -0,0 +1,83 @@
...
+export function isCI(): boolean {
+    return process.env.CI === 'true';
+}
+
+/**
+ * Determine if analytics tests should be skipped.
+ * Returns true if Tinybird is not available AND we're not in CI.
+ * In CI, we want tests to fail if Tinybird isn't available.
+ */
+export async function shouldSkipAnalyticsTests(): Promise<boolean> {
```
ibalosh: It feels like we are over-complicating this with detection whether tinybird is available. Tomorrow we will add something else, and the process will be even more complex. Ti feels wrong for tests to know, what to skip and prone to future errors.

I think that by default all tests should run, and you can control whether you want that or not by something simpler like [tag annotations](https://playwright.dev/docs/test-annotations), which you can use in your dev environment to run specific tests that you would like.

With tags, you could choose which tests to include or exclude, by single or multiple groups. I would rather go for:

* start what you want 2. run tests that you chose against it
* rather than start what you want and tests detect what they should ignore (this feels like it could go wrong in future, and too much of an optimization)

e2e/helpers/pages/public/public-page.ts
```diff
@@ -86,13 +87,17 @@ export class PublicPage extends BasePage {
 
     async goto(url?: string, options?: pageGotoOptions): Promise<void> {
         await this.enableAnalyticsRequests();
-        const pageHitPromise = this.pageHitRequestPromise();
+        const pageHitPromise = this.waitForPageHitRequest();
```
ibalosh: `thinkingOutLoud` yeah, this wait should probably be just a choice for analytics tests, goto should ignore this for everything else (that way it would work even if we use tags and certain services are off like analytics

e2e/helpers/environment/environment-factory.ts
```diff
@@ -0,0 +1,60 @@
...
+const debug = baseDebug('e2e:EnvironmentFactory');
+
+// Cached manager instance (one per worker process)
+let cachedManager: EnvironmentManager | DevEnvironmentManager | null = null;
+
+/**
+ * Check if the dev environment (yarn dev:forward) is running.
+ * Detects by checking for the ghost_dev network and running MySQL container.
+ */
+export async function isDevEnvironmentAvailable(): Promise<boolean> {
```
ibalosh: Do you find this way sufficient to check whether dev environment is running locally?

ibalosh (changes requested):
I’ll take another pass tomorrow with a fresher head to fully understand the newly introduced managers, so please take this initial review with a bit of caution. I wanted to get some early feedback to you quickly.

Some of the concerns I had by looking at the implementation:

**Too much behaviour is hidden behind auto detection** 

* get environment manager silently decides dev and test (standalone) mode. If something is missconfigured, you silently fallback instead of failing loudly
* something like const useDev = process.env.E2E_DEV === 'true' maybe?
* also for analytics exclusion I would simply use [tag annotations](https://playwright.dev/docs/test-annotations#tag-tests)

there's a lot of coupling to docker internals, like if changing compose.dev.yaml naming, hard to find the error (I admit though our initial implementation has similar issue, but it's less complex in my opinion)

**Two environment managers instead of one**

Original already encapsulates the general idea of start, cleanup, stop etc, why create one more? Ideally, they should be two configurations, rather than two classes plus factory and auto detection?

We might want to add simpler setup of env to choose, simplify a bit env manager, and per config and then just:

```
export function getEnvironmentManager(): EnvironmentManager {
  const config = loadFromEnv();
  return new EnvironmentManager(config);
}
```

Thats some initial reactions I had by look today. I do realize that this can't be simple, but we should keep it still friendly if possible to new-commers to the e2e repo and keep it very well documented to be able to read it. At first glance, it seems more complex 

I also got errors on custom-views.tests when yarn dev:forward is running in one shell and I do yarn test in other

e2e/helpers/environment/service-availability.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (tinybirdAvailable !== null) {
+        return tinybirdAvailable;
+    }
+
+    const docker = new Docker();
+
+    try {
+        const containers = (await Promise.all([
+            findTinybirdContainers(docker, DOCKER_COMPOSE_CONFIG.PROJECT),
+            findTinybirdContainers(docker, 'ghost-dev')
```
9larsons: Shouldn't this be in the env/constants?

e2e/helpers/environment/service-availability.ts
9larsons: I tend to agree on principle. While this is slick and a proactive setup, I'd rather force the engineer to be more prescriptive.

On the other hand, I think if you run `yarn test:e2e` it should run what we expect to run on CI. If that means we need to boot extra containers, then we should do that. And we shouldn't require the dev to know that they need to disable tinybird. What Jonathan is trying to do here is conserve resources/increase speed of running the tests. 

I don't think I like the default being a slimmer version of our tests, and this seems like a reasonable compromise. I'm ultimately on the fence.

e2e/helpers/environment/service-managers/dev-ghost-manager.ts
```diff
@@ -0,0 +1,359 @@
...
+    networkName: string;
+    projectNamespace: string;
+    workerIndex: number;
+    mysql: {host: string; port: number; user: string; password: string};
+    redis: {host: string; port: number};
+    images: {ghost: string; gateway: string};
+}
+
+// Base environment matching compose.dev.yaml - only set what Ghost needs
+const BASE_GHOST_ENV = [
```
9larsons: If we have `/environment/dev-environment-manager` this almost feels appropriate to house there, and awkward to have here.

9larsons:
I'm not seeing this actually faster. Could you clarify how it's faster? I'm actually seeing much slower rates locally, and suspect something is different on my end. 

Would you mind sending over a clip/vid of you running this with and without local dev?

9larsons (dismissed):
In a note above, plus a couple questions.

ibalosh:
> I'm not seeing this actually faster. Could you clarify how it's faster? I'm actually seeing much slower rates locally, and suspect something is different on my end.
> 
> Would you mind sending over a clip/vid of you running this with and without local dev?

Same for me as for @9larsons , I have not seen speed increase, or at least not sure which one. I am wondering if I am missing something when I run it

e2e/helpers/environment/service-availability.ts
jonatansberg: I'm happy to consider other ways of doing this but I don't buy the slippery slope argument in this regard. Would you be open to iterating on that post merge, or do you consider this a blocker? 

With that said, I do actually think having the tests be conditional makes sense. Tinybird is an optional component that's not a requirement for running Ghost. On top of that Ti is exceptionally expensive to run and uses more resources than everything else combined. The way the tests have been set up mean they won't even run on most people's machines. I mean, my M4 Pro MacBook with 24 GB of RAM is struggling most of the time, especially if you're also running the dev environment alongside the dedicated E2E test containers. 

If we don't make these tests conditional on the availability of Tinybird then either: 
A. You can't run the E2E tests against the existing dev containers without forcing everyone to either always run Tinybird (and whatever pinned version of the analytics backend we happen to have in our Docker configuration).
B. We need a special E2E testing command rather than the "default one".
C. We always skip the tests unless you opt in.

Having the test be conditional means that if you explicitly start the analytics stuff, they will run, and if you don't, they won't. They always run in CI. To me, this feels like a very pragmatic trade-off that avoids the very clear drawbacks of all of the above alternatives. 

If/when the analytics backend is properly integrated into the monorepo, going with the "always run everything option" seems more sensible. By then we should hopefully be rid of Ember as well, which would free up a lot of resources.

e2e/helpers/environment/environment-factory.ts
jonatansberg: It's been working well enough in my local testing. Any specific reason you're concerned about this?

jonatansberg:
@9larsons @ibalosh The biggest speed increases come from:
1. You no longer need to rebuild any images to test frontend changes. This saves **minutes** on every test run. 
2. Less resource usage overall. The existing E2E setup forces my computer to start SWAPing if I leave my dev environment running. This makes everything slow, the CPU is pinned at 100% due to I/O wais, tests become flaky, and eventually my computer is sent into orbit. 

If I instead kill my dev environment to run the tests, then yes, the tests run faster, but the overall iteration cycle becomes worse as I now need to restart my dev environment if I want to e.g. verify a change in my browser, before killing it to run the tests again. 

If you have machines with more working memory, I imagine you'd get different results. Also, serving the app from the dev servers will be slower than serving the built version. So if your testing patterns mostly involve iterating on backend code, leaving the frontend unchanged, that's also a completely different story.

jonatansberg:
@ibalosh I missed your main comment, sorry. 

Your concerns around auto-detection and having two managers instead of one are definitely reasonable. I went for building a second manager and auto-detection rather than trying to iterate on the existing one, as doing so while maintaining support for both modes of operation became a mess. 

The auto-detection makes the default DX nicer, but we can definitely have an `e2e:forward` command or something to that effect and toggle based on environment variables if you feel more comfortable that way. 

Ultimately my goal would be for what's currently called the dev environment manager in this PR to supersede the current one. In that world, we would still have two different modes of running the tests, either against the local development environment or against pre-built assets, but we would be rid of most of the duplication. 

This would also allow running the tests locally in the same way as today, which would be slightly faster if you're only doing backend changes.

ibalosh:
hey @jonatansberg let's catchup on a call to talk more to move this one forward. Even though number of files is small-ish, the PR is a bit bigger, with multiple concepts added, on some I am sold, some not, so I would like to get clarity

cmraible (changes requested):
Sorry it's taken a while for me to take a good look at this; I've waited because I generally agree with @ibalosh and @9larsons feedback. That said, I do think we should work to enable this for the quicker inner loop when developing locally, so I'm motivated to find a solution here that we can all agree on. I'm also eager to fully adopt the `dev:forward` experience across the board and ultimately remove the need for the more complex, static Docker build.

One thing I'll add, which maybe (hopefully?) clears a path to a simpler solution: I don't see any reason to maintain two different "modes" of running the e2e tests, which is where I think a lot of this additional complexity is coming from. I'd much prefer to _replace_ the current method of running tests (both locally and in CI for parity) than to add another way to run them.

jonatansberg:
@cmraible No worries, I've been off for a bit with sick kids and then some sick leave of my own. I had a good chat with @ibalosh about this PR and the general direction I see this going in. Long story short, what you're suggesting is indeed the direction we're heading in.

I've done some more work on reshaping this which will remove the environment sniffing. That's almost ready. 

We'll still have two parallel setups for a while until the version I've been working on is ready for general use, including CI. But at that point, we can converge on a single setup again.

jonatansberg:
I've done another pass at this incorporating the changes that @ibalosh and I discussed before the holidays. 

Would you be okay with merging this now? If so I'll move on to how we run the tests in CI, including running the full E2E suite against the latest released version of Ghost in addition to main (as we discussed in the engineering forum.)

ibalosh (approved):

ibalosh:
all good 👍

e2e/helpers/environment/environment-factory.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Check if the dev environment (yarn dev) is running.
+ * Detects by checking for the ghost_dev network and running MySQL container.
+ */
+export async function isDevEnvironmentAvailable(): Promise<boolean> {
+    const docker = new Docker();
+
+    try {
+        const networks = await docker.listNetworks({
+            filters: {name: [DEV_ENVIRONMENT.networkName]}
+        });
```
cmraible: I would still prefer to avoid the environment sniffing, in favor of an explicit environment variable. I know this will go away soon anyway once this becomes the default and only option, but in the meantime I'd prefer to be explicit.

If you feel really strongly that the environment sniffing is important, I think the implementation of it needs a tweak: even if you stop `yarn dev`, the docker network still exists until you run `docker compose -f compose.dev.yaml down`, so it will try to use the dev env even if it's not running. It might be better to check for an actively running container or something along those lines.

e2e/playwright.config.mjs
```diff
@@ -43,13 +43,22 @@ const config = {
...
         },
+        {
+            name: 'analytics',
+            testDir: './tests',
+            testMatch: ['analytics/**/*.test.ts'],
+            use: {
+                viewport: {width: 1920, height: 1080}
+            },
+            dependencies: ['global-setup']
+        },
```
cmraible, line 62: 🚨 This makes a lot of sense, but it also excludes the analytics tests from CI since it only runs the main project. Please make sure the analytics project is included in CI before merging.

e2e/tests/public/homepage.test.ts
```diff
@@ -9,4 +9,15 @@ test.describe('Ghost Public - Homepage', () => {
...
+
+    test('sends page hit request', async ({page}) => {
+        const homePage = new HomePage(page);
+
+        // Set up request expectation before navigation
+        const response = homePage.waitForPageHitRequest();
+        await homePage.goto();
+
+        // Wait for the request to be fulfilled
+        await expect(response).resolves.toBeDefined();
```
cmraible: Could we cleanup this syntax at all? 

One of our goals with the e2e test suite is to make the tests as readable as possible, such that someone non-technical can read and understand them. Keeping the assertion isolated in the Page Object abstracts away this more technical detail and makes the tests themselves read a lot nicer, and this feels like a regression in that sense.

It's also easy to forget that this is necessary when writing tests, and if we forget it our tests end up being significantly flakier

e2e/helpers/pages/public/public-page.ts
```diff
@@ -86,13 +86,11 @@ export class PublicPage extends BasePage {
 
     async goto(url?: string, options?: pageGotoOptions): Promise<void> {
         await this.enableAnalyticsRequests();
-        const pageHitPromise = this.pageHitRequestPromise();
```
cmraible, line 89: Agree it makes sense to skip the pageHitPromise if it's not an analytics test, but I'd really prefer to keep this implementation detail in the POM rather than the tests, because it's easy to forget and results in flaky tests if it's missing, plus it makes the tests/assertions a bit harder to read.

I've raised a PR to this PR here with a suggestion that will automatically wait for the `pageHitPromise` only if running in the analytics project: https://github.com/TryGhost/Ghost/pull/26053

e2e/helpers/environment/environment-factory.ts
jonatansberg: The reason I kept it in there is to ensure that this works out of the box with the Playwright integration in VS Code or Cursor and for people who want to run the tests while developing things. I'm planning to move on to getting the standalone version of this to run next, which means we would be able to pull out the old setup. 

I think the network ought to be cleaned up when `yarn dev` exits since [we do call `docker compose down`](https://github.com/TryGhost/Ghost/blob/main/package.json#L136). I did have a check for if the MySQL container previously but I removed that in order to clean up the config.. I could add that back in tho 🤔

e2e/helpers/pages/public/public-page.ts
jonatansberg: I was thinking it'd make more sense to make this an explicit assertion rather than having the POM do this implicitly, meaning that when it fails, you get a proper test failure rather than just a timeout. I haven't seen any flakiness, but I also haven't tested this extensively. 

If you prefer it integrated into the POM, we can definitely do it that way instead.

jonatansberg: I think I understand what you mean by flakiness now! It's the other analytics tests that become flaky 🤦

e2e/helpers/environment/environment-factory.ts
jonatansberg: I updated this to check for all the required services rather than assuming they're there.

e2e/playwright.config.mjs
jonatansberg: Done!

e2e/helpers/pages/public/public-page.ts
jonatansberg: I integrated your fix and that seems to have resolved the flakiness.

cmraible (approved):
Thanks for incorporating my feedback, LGTM 🚢

jonatansberg:
@9larsons let me know if you have any more feedback. I'd be happy to incorporate that as a follow-up. 

I'll go ahead and merge this so I can move on to getting the CI version of this done and ensuring we test against the latest public release in addition to main.

PR #25563, opened 2025-12-01
Added multipart upload in S3 storage adapter

vershwal:
ref https://linear.app/ghost/issue/PRO-1538

- GCS doesn't support AWS streaming signature format, causing uploads to fail with SignatureDoesNotMatch. This can be fixed by using buffer instead of stream for uploads. But this will cause memory issues for large files
- This commit implements a multipart upload for files > 10MB
- Small files (<10MB): Use buffer upload (fs.readFile)
- Large files (≥10MB): Use S3 multipart upload API for memory efficiency. Multipart uploads read file in 10MB chunks, keeping RAM usage constant
- Added automatic cleanup with AbortMultipartUpload on errors
- Multipart upload is S3-standard and works with both AWS S3 and GCS


ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +153,101 @@ export default class S3Storage extends StorageBase {
...
+                while (uploadedBytes < fileSize) {
+                    // Read one part into buffer
+                    const remainingBytes = fileSize - uploadedBytes;
+                    const currentPartSize = Math.min(this.partSize, remainingBytes);
+                    const buffer = Buffer.alloc(currentPartSize);
+                    
+                    const {bytesRead} = await fileHandle.read(buffer, 0, currentPartSize, uploadedBytes);
+                    
+                    if (bytesRead === 0) {
+                        break;
```
allouis: I think this should be an error

vershwal:
@coderabbitai review

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -48,6 +56,8 @@ export interface S3StorageOptions {
     sessionToken?: string;
     tenantPrefix?: string;
     s3Client?: S3Client;
+    multipartThreshold?: number; // File size threshold for multipart upload (default: 10MB)
+    partSize?: number; // Size of each part for multipart upload (default: 10MB)
```
allouis: I think we can remove comments here, we can make the names more explicit if we're worried about losing context - maybe `multipartUploadThresholdBytes` and `multipartChunkSizeBytes` would capture everything?

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -93,6 +107,11 @@ export default class S3Storage extends StorageBase {
             });
         }
 
+        // 10MB threshold - files larger than this use multipart upload
+        this.multipartThreshold = options.multipartThreshold || 10 * 1024 * 1024;
```
allouis: I would prefer we have these as required options - it removes branching and ensures that Pro always sets them because we would then error if they're missing

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -115,6 +134,17 @@ export default class S3Storage extends StorageBase {
         const relativePath = await this.getUniqueFileName(file, dir);
 
         const key = this.buildKey(relativePath);
+
+        // Get file size to determine upload strategy
```
allouis: Can remove this comment

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -115,6 +134,17 @@ export default class S3Storage extends StorageBase {
...
+        const stats = await fs.promises.stat(file.path);
+
+        if (stats.size >= this.multipartThreshold) {
+            logging.info(`Large file (${Math.round(stats.size / 1024 / 1024)}MB), using multipart upload...`);
+            return await this.uploadMultipart(file, key, stats.size);
+        }
+ 
+        // Use simple upload for small files (faster, less overhead)
```
allouis: Can remove this comment

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -115,6 +134,17 @@ export default class S3Storage extends StorageBase {
         const relativePath = await this.getUniqueFileName(file, dir);
 
         const key = this.buildKey(relativePath);
+
+        // Get file size to determine upload strategy
+        const stats = await fs.promises.stat(file.path);
+
+        if (stats.size >= this.multipartThreshold) {
+            logging.info(`Large file (${Math.round(stats.size / 1024 / 1024)}MB), using multipart upload...`);
```
allouis: I don't think we should have this logic of converting bytes to human readable embedded in a template string, I would either pull that logic out OR potentially just use the bytes? It might also be usefult to include the threshold number too (maybe filename/path?) if we're ever debugging this

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +157,106 @@ export default class S3Storage extends StorageBase {
         return `${this.cdnUrl}/${key}`;
     }
 
+    private async uploadMultipart(file: UploadFile, key: string, fileSize: number): Promise<string> {
+        let uploadId: string | undefined;
+
+        try {
+            // Step 1: Initiate multipart upload
```
allouis: Can remove this comment

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +157,106 @@ export default class S3Storage extends StorageBase {
...
+                Key: key,
+                ContentType: file.type
+            }));
+
+            uploadId = createResponse.UploadId;
+            if (!uploadId) {
+                throw new errors.InternalServerError({
+                    message: tpl(messages.multipartUploadInitFailed)
+                });
+            }
```
allouis: I think we can move this out of the `try/catch` as it is just caught and rethrown

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +157,106 @@ export default class S3Storage extends StorageBase {
...
+
+            // Step 2: Upload parts
```
allouis: Can remove this comment

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +157,106 @@ export default class S3Storage extends StorageBase {
...
+            const parts: {ETag: string; PartNumber: number}[] = [];
+            const fileHandle = await fs.promises.open(file.path, 'r');
```
allouis: I think we can also move this outside of the try/catch - and even attempt to open the file before initiating the multipart upload, so we can exit early if there's an error

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +157,106 @@ export default class S3Storage extends StorageBase {
...
+                        });
+                    }
+
+                    // Upload this part
+                    const uploadPartResponse = await this.client.send(new UploadPartCommand({
+                        Bucket: this.bucket,
+                        Key: key,
+                        UploadId: uploadId,
+                        PartNumber: partNumber,
+                        Body: buffer.slice(0, bytesRead)
```
allouis: I guess we have to slice the buffer because we might have read less bytes than we requested? I'm wondering if it might be easier to understand this if we use a stream to read the bytes and buffer it into chunks, rather than manually pulling them 🤔 

I'm thinking something like this:

```ts
const stream = fs.createReadStream(file.path); // maybe pass highwaterMark to set the chunk size to the part size?
let buffer = Buffer.alloc(0);
let partNumber = 1;

for await (const chunk of stream) { 
   buffer = Buffer.concat([buffer, chunk]); // loop and add chunks to buffer

   while (buffer.length >= this.partSize) { // while it contains at least one part upload it
       const part = buffer.slice(0, this.partSize);
       buffer = buffer.slice(this.partSize);

       await uploadPart(part, partNumber);
       partNumber += 1;
   }
}

if (buffer.length > 0) { // check if there's a remainder part at the end
   await uploadPart(buffer, partNumber);
}
```

Not sure if this feels simpler to you?

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +157,106 @@ export default class S3Storage extends StorageBase {
...
+                    
+                    const {bytesRead} = await fileHandle.read(buffer, 0, currentPartSize, uploadedBytes);
+                    
+                    if (bytesRead === 0) {
+                        throw new errors.InternalServerError({
+                            message: tpl(messages.multipartUploadReadFailed)
+                        });
+                    }
+
+                    // Upload this part
```
allouis: Can remove this comment

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +157,106 @@ export default class S3Storage extends StorageBase {
...
+            // Step 2: Upload parts
+            const parts: {ETag: string; PartNumber: number}[] = [];
+            const fileHandle = await fs.promises.open(file.path, 'r');
+            
+            try {
+                let partNumber = 1;
+                let uploadedBytes = 0;
+
+                while (uploadedBytes < fileSize) {
+                    // Read one part into buffer
```
allouis: Can remove this comment

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +157,106 @@ export default class S3Storage extends StorageBase {
...
+                    partNumber += 1;
+
+                    const progress = Math.round((uploadedBytes / fileSize) * 100);
+                    logging.info(`Uploaded part ${partNumber - 1}/${Math.ceil(fileSize / this.partSize)} (${progress}%)`);
+                }
+            } finally {
+                await fileHandle.close();
+            }
+
+            // Step 3: Complete multipart upload
```
allouis: Can remove this comment

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +157,106 @@ export default class S3Storage extends StorageBase {
...
+                    parts.push({
+                        ETag: uploadPartResponse.ETag,
+                        PartNumber: partNumber
+                    });
+
+                    uploadedBytes += bytesRead;
+                    partNumber += 1;
+
+                    const progress = Math.round((uploadedBytes / fileSize) * 100);
+                    logging.info(`Uploaded part ${partNumber - 1}/${Math.ceil(fileSize / this.partSize)} (${progress}%)`);
```
allouis: I think moving this before the increment makes it easier to understand, I couldn't immediately see why we do `+ 1`

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +157,106 @@ export default class S3Storage extends StorageBase {
...
+                UploadId: uploadId,
+                MultipartUpload: {
+                    Parts: parts
+                }
+            }));
+
+            logging.info(`Multipart upload completed: ${parts.length} parts`);
+            return `${this.cdnUrl}/${key}`;
+        } catch (error) {
+            // Abort multipart upload on error to clean up
```
allouis: Can remove this comment I think 🤔

ghost/core/core/server/adapters/storage/S3Storage.ts
allouis: Or even we can pull that looping out into a separate method:
```ts
async function* readFileInParts(filePath: string, partSize: number): AsyncGenerator<Buffer> {
    const stream = fs.createReadStream(filePath);
    let buffer = Buffer.alloc(0);

    for await (const chunk of stream) {
         buffer = Buffer.concat([buffer, chunk]);

         while (buffer.length >= partSize) {
              yield buffer.slice(0, partSize);
              buffer = buffer.slice(partSize);
         }
    }

    if (buffer.length > 0) {
         yield buffer;
    }
}
```

And then we can just do something like

```ts
const parts = this.readFileInParts(filePath, partSize);

for await (const part of parts) {
    // part is a buffer of the correct size and we can upload it
}
```

ghost/core/core/server/adapters/storage/S3Storage.ts
vershwal: done!

ghost/core/core/server/adapters/storage/S3Storage.ts
vershwal: ya, make sense

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -93,6 +107,20 @@ export default class S3Storage extends StorageBase {
...
+            });
+        }
+        this.multipartUploadThresholdBytes = options.multipartUploadThresholdBytes;
+
+        if (!options.multipartChunkSizeBytes) {
+            throw new errors.IncorrectUsageError({
+                message: tpl(messages.missingMultipartChunkSize)
+            });
+        }
+        this.multipartChunkSizeBytes = options.multipartChunkSizeBytes;
```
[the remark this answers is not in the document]
vershwal: Thanks for this. This is important because if someone adds a config where minimum chunk size is less than 5MB, GCS will throw error and the upload would not be completed. I faced this issue locally. I've added the validation.

ghost/core/core/server/adapters/storage/S3Storage.ts
vershwal: Good call! I've refactored this to use a stream-based approach with an async generator. The upload loop is now much cleaner. This is definitely easier to understand than the manual fileHandle.read() with offset tracking

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -18,6 +18,9 @@ import {
     S3ClientConfig
 } from '@aws-sdk/client-s3';
 
+// Minimum chunk size for multipart uploads (5 MiB) - required by S3/GCS
```
allouis, line 21: Do we know if GCS has this minimum too? Can we check? Maybe worth adding a link to some docs here

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +172,101 @@ export default class S3Storage extends StorageBase {
...
+                UploadId: uploadId,
+                MultipartUpload: {
+                    Parts: parts
+                }
+            }));
+
+            logging.info(`Multipart upload completed: file=${key} parts=${parts.length}`);
+            return `${this.cdnUrl}/${key}`;
+        } catch (error) {
+            if (uploadId) {
```
allouis: We can get rid of this as it will always be defined due to the check on line 201

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -127,6 +172,101 @@ export default class S3Storage extends StorageBase {
...
+
+                parts.push({
+                    ETag: uploadPartResponse.ETag,
+                    PartNumber: partNumber
+                });
+
+                uploadedBytes += chunk.length;
+
+                const progress = Math.round((uploadedBytes / fileSize) * 100);
+                logging.info(`Uploaded part ${partNumber}/${totalParts} (${progress}%)`);
```
allouis: What are your thoughts on the logging here? It's pretty verbose, which is great, but we've got a lot of variables defined that are soley used for logging

I'm wondering a few things:
- How useful are logs for the progress of the upload?
- If they're useful - can we just use parts OR bytes - rather than combination?

allouis (commented):
Few small comments but this is looking good!

ghost/core/core/server/adapters/storage/S3Storage.ts
vershwal: Yes, GCS has the limit too: https://docs.cloud.google.com/storage/quotas#requests

ghost/core/core/server/adapters/storage/S3Storage.ts
vershwal: Nice! done.

ghost/core/core/server/adapters/storage/S3Storage.ts
vershwal: hmm..these logs were helpful when I was testing locally - to see the upload is happening. But in production i don't think these will be of much use. We already have logs in the start and end of upload - I think those are enough for debugging in production.

allouis (approved):

PR #25564, opened 2025-12-01
Fixed search bar opening when using Cmd+K inside editor

kevinansfield:
closes https://linear.app/ghost/issue/BER-3078/

- we moved search and settings keyboard shortcuts from the Ember sidebar component to the application route so we still have working shortcuts inside React shell where we don't render the Ember sidebar at all
- this caused a slight change in behaviour because the shortcuts are now always active, even when Ember wouldn't have rendered the sidebar such as inside the editor, resulting in duplicate keyboard shortcuts / unexpected search bar opening
- added an early return to the shortcut handlers that uses `ui.isFullScreen` as a proxy for the sidebar not being shown and therefore the shortcuts not being active

### DIFF
--- a/ghost/admin/tests/acceptance/search-test.js
+++ b/ghost/admin/tests/acceptance/search-test.js
@@ -221,6 +221,17 @@ describe('Acceptance: Search', function () {
             await closeSearchWithBackdrop();
         });
 
+        it('does not open search modal if the sidebar is hidden', async function () {
+            const post = this.server.create('post', {title: 'Test post', slug: 'test-post', status: 'draft'});
+            await visit(`/editor/post/${post.id}`);
+            assertSearchModalClosed();
+            await triggerKeyEvent(document, 'keydown', 'K', {
+                metaKey: ctrlOrCmd === 'command',
+                ctrlKey: ctrlOrCmd === 'ctrl'
+            });
+            assertSearchModalClosed();
+        });
+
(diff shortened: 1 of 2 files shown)


ghost/admin/tests/acceptance/search-test.js
```diff
@@ -221,6 +221,17 @@ describe('Acceptance: Search', function () {
...
+        it('does not open search modal if the sidebar is hidden', async function () {
+            const post = this.server.create('post', {title: 'Test post', slug: 'test-post', status: 'draft'});
+            await visit(`/editor/post/${post.id}`);
+            assertSearchModalClosed();
+            await triggerKeyEvent(document, 'keydown', 'K', {
+                metaKey: ctrlOrCmd === 'command',
+                ctrlKey: ctrlOrCmd === 'ctrl'
+            });
+            assertSearchModalClosed();
+        });
```
rob-ghost, line 233: **Question:** Should this be an e2e instead? I'll approve regardless.

rob-ghost (approved):
I think the approach is sensible 👍🏻

ghost/admin/tests/acceptance/search-test.js
kevinansfield: Eventually, yes. I've opened an issue to track moving these to the e2e suite, for this PR I wanted to get changes in quickly to unblock the next Ghost release.

PR #25568, opened 2025-12-01
🌵SPIKE: i18next-cli integration to replace deprecated i18next-parser

cathysarisky:
SPIKE:  Testing replacing i18next-parser with i18next-cli.  It has KEY REPLACEMENT. I'm in love.  

A basic sanity check has it working and extracting keys, so it'll be basically a drop-in replacement, with added new features available as we want them.  (And I want them.)

It can be run (from within the i18n folder) with `NAMESPACE=portal i18next-cli extract -c i18next.config.ts `

TODO:  Update scripts in the package to use new tooling

TODO:  Make it parse .hbs files.


ghost/i18n/package.json
```diff
@@ -8,6 +8,7 @@
   "types": "./build/i18n.d.ts",
   "scripts": {
     "dev": "echo \"Implement me!\"",
+    "postinstall": "patch-package",
```
cathysarisky:
```suggestion
```

cathysarisky:
TODO: Check if email template strings are getting picked up correctly

PR #25578, opened 2025-12-02
Added isolated Docker networks per E2E test worker

rob-ghost:
ref https://linear.app/ghost/issue/BER-3096

## Summary

- Creates an isolated Docker network per parallel test worker (`ghost_e2e_worker_0`, `ghost_e2e_worker_1`, etc.)
- Ghost containers now connect to both the shared compose network (for MySQL, Tinybird, etc.) and their worker-specific network
- Worker networks are automatically cleaned up during global teardown

## Context

Analysis of flaky React E2E tests revealed that `ERR_NETWORK_CHANGED` errors were causing blank page failures. React tests had a 5.5x higher JS error rate than Ember tests, with failures clustered in the first 0-133ms of page load when the main `ghost-*.js` bundle was being fetched.

The root cause: when parallel test workers create/destroy Ghost containers, Docker network operations on the shared `ghost_e2e` network can briefly affect all containers on that network, interrupting in-flight browser requests. React is more vulnerable because it makes more JS requests during initial page load.

This change isolates each worker's container operations to its own network, preventing cross-worker interference.

### DIFF
--- a/e2e/helpers/environment/service-managers/ghost-manager.ts
+++ b/e2e/helpers/environment/service-managers/ghost-manager.ts
@@ -15,6 +15,7 @@ export interface GhostInstance {
     port: number;
     baseUrl: string;
     siteUuid: string;
+    workerNetworkId?: string; // isolated network for this worker
 }
 
 export interface GhostStartConfig {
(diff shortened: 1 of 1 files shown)


rob-ghost:
Closing as this 1. wasn't working and 2. is probably the wrong path to go down.

`ERR_NETWORK_CHANGED` won't be related to docker networks.

PR #25579, opened 2025-12-02
Reduced worker count in CI

jonatansberg:
refs https://linear.app/ghost/issue/BER-3096/investigate-e2e-flakiness-with-react

The default runners on GitHub have 4 vCPU, so running 4 workers in parallel is too high of a concurrency setting.

### DIFF
--- a/.github/workflows/ci-docker.yml
+++ b/.github/workflows/ci-docker.yml
@@ -341,11 +341,79 @@ jobs:
       - name: Setup Playwright
         uses: ./.github/actions/setup-playwright
 
+      - name: Install psutil
+        run: pip install psutil
+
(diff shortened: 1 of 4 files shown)


ibalosh (changes requested):
reducing workers doesnt seem to reduce the flakyness, from what I could see it's way more frequent on React project, and on this branch, saw it only on React tests

jonatansberg:
@ibalosh Dropping all the way to one worker per shard seems to have eliminated the flakiness without negatively impacting total runtime if we add a few more shards.

ibalosh:
@jonatansberg you are right, seems it resolved it 👍 🙌

PR #25584, opened 2025-12-02
Wired up Koenig editor to welcome emails modal

9larsons:
ref https://linear.app/ghost/issue/NY-787/
- added Koenig editor to welcome emails modal

This editor allows the user to manipulate the content of the email. However, this isn't wired up to anything while we're waiting for the persistence changes to be merged.

### DIFF
--- a/apps/admin-x-design-system/src/global/form/HtmlEditor.tsx
+++ b/apps/admin-x-design-system/src/global/form/HtmlEditor.tsx
@@ -14,6 +14,7 @@ export interface HtmlEditorProps {
     nodes?: 'DEFAULT_NODES' | 'BASIC_NODES' | 'MINIMAL_NODES'
     emojiPicker?: boolean;
     darkMode?: boolean;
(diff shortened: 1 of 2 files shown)


apps/admin-x-design-system/src/global/form/HtmlEditor.tsx
```diff
@@ -14,6 +14,7 @@ export interface HtmlEditorProps {
     nodes?: 'DEFAULT_NODES' | 'BASIC_NODES' | 'MINIMAL_NODES'
     emojiPicker?: boolean;
     darkMode?: boolean;
+    singleParagraph?: boolean;
```
9larsons, line 17: Arguably this limited editor was only ever intended for a single line use case in settings. I've kept `true` for backwards compatibility.

If we want the welcome email to allow for all of our cards/be a full featured editor, we should revert this part.

cmraible (approved):
That was quick! Nice one, LGTM 🚀

PR #25589, opened 2025-12-03
fix(mail): stringify defaultFromEmail before passing to getAddressFromString

hodorogandrei:
## Summary

This PR fixes a bug in `GhostMailer.js` where transactional emails fail to send when no explicit `from` address is provided.

### The Problem

In `getFromAddress()` (line 34-54), when `requestedFromAddress` is not provided, the code sets:

```javascript
requestedFromAddress = emailAddress.service.defaultFromEmail;
```

However, `defaultFromEmail` returns an **object** `{address: string, name: string}`, not a string.

This object is then passed to `getAddressFromString(requestedFromAddress, ...)` which internally calls `EmailAddressParser.parse(from)`. The `parse()` function expects a **string** input and returns `null` when given an object:

```javascript
// Testing EmailAddressParser.parse() behavior:
EmailAddressParser.parse('test@example.com')  // => {address: 'test@example.com', name: undefined}
EmailAddressParser.parse({address: 'test@example.com'})  // => null
```

This causes downstream failures when trying to access `addresses.from.name` on a null value.

### Impact

This bug affects **all transactional emails** that don't explicitly specify a `from` address:
- **Staff device verification (2FA) emails** - Users cannot log in when `staffDeviceVerification` is enabled
- **Password reset emails** - Users cannot reset their passwords
- **Member magic link emails** - Members cannot sign in
- Any other system emails using the default sender

The error message shown to users is unhelpful: "Failed to send email. Please check your site configuration and try again." - even when the SMTP configuration is correct.

### Root Cause Analysis

The bug was introduced when the email address handling was refactored to use the new `EmailAddressService`. The `defaultFromEmail` getter was changed to return an object instead of a string, but the consumer in `GhostMailer.js` was not updated to handle this change.

### The Fix

Convert the email object to a string using `EmailAddressParser.stringify()` before passing it to `getAddressFromString()`:

```javascript
// Before (buggy):
requestedFromAddress = emailAddress.service.defaultFromEmail;

// After (fixed):
const defaultEmail = emailAddress.service.defaultFromEmail;
requestedFromAddress = EmailAddressParser.stringify(defaultEmail);
```

## Test Plan

1. Enable `staffDeviceVerification` in config
2. Configure SMTP settings (any provider)
3. Attempt to log in to Ghost Admin
4. Verify that the verification code email is sent successfully
5. Also test password reset and member magic link emails

## Reproduction Steps (for the bug)

1. Set up Ghost with SMTP configured (not Mailgun)
2. Enable `staffDeviceVerification: true` in config
3. Try to log in to Ghost Admin
4. Observe the "Failed to send email" error even though SMTP credentials are correct
5. Check logs - you'll see an EmailError but no underlying SMTP error (because it fails before reaching SMTP)

---

> [!NOTE]
> Fixes default sender handling by stringifying `defaultFromEmail`; adds/fills Romanian translations across comments, newsletter, portal, and search.
> 
> - **Backend/Mail**
>   - Fix `getFromAddress` in `core/server/services/mail/GhostMailer.js` to stringify `emailAddress.service.defaultFromEmail` via `EmailAddressParser.stringify` before `getAddressFromString` when no `from` is provided.
> - **i18n (Romanian)**
>   - `locales/ro/comments.json`: add translation for `"Sort by"`.
>   - `locales/ro/newsletter.json`: fill numerous missing strings (e.g., status labels, actions, notices).
>   - `locales/ro/portal.json`: add multiple missing strings (plan/change, verification, rate-limit and checkout messages, unsubscribe notices, etc.).
>   - `locales/ro/search.json`: add translations for search UI strings (`Authors`, `Posts`, `Tags`, etc.).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2315758b92ea0bb83dc65006bbffe38b69ff9af3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


cathysarisky:
Hi!  Thanks for the contribution!  I'm a volunteer Ghost contributor and I handle the i18n PRs, which is why I'm looking at this one.

I think something has gone slightly wrong with this PR, as it seems to have both Romanian language translations and a bug fix for email.  Both of these are good things, but they don't belong together.  I'm going to close this PR, but please re-open as two separate PRs so that they can be handled separately.

For the email fix, you should probably also add some tests that cover this case, and make sure that no existing tests are breaking. :)

hodorogandrei:
> Hi! Thanks for the contribution! I'm a volunteer Ghost contributor and I handle the i18n PRs, which is why I'm looking at this one.
> 
> I think something has gone slightly wrong with this PR, as it seems to have both Romanian language translations and a bug fix for email. Both of these are good things, but they don't belong together. I'm going to close this PR, but please re-open as two separate PRs so that they can be handled separately.
> 
> For the email fix, you should probably also add some tests that cover this case, and make sure that no existing tests are breaking. :)

Thank you for all the constructive feedback! I have now opened a new Pull Request:

https://github.com/TryGhost/Ghost/pull/25594

PR #25594, opened 2025-12-03
fix(mail): stringify defaultFromEmail before passing to getAddressFromString

hodorogandrei:
## Summary

This PR fixes a bug in `GhostMailer.js` where transactional emails fail to send when no explicit `from` address is provided.

### The Problem

In `getFromAddress()` (line 32-51), when `requestedFromAddress` is not provided, the code sets:

```javascript
requestedFromAddress = emailAddress.service.defaultFromEmail;
```

However, `defaultFromEmail` returns an **object** `{address: string, name: string}`, not a string.

This object is then passed to `getAddressFromString(requestedFromAddress, ...)` which internally calls `EmailAddressParser.parse(from)`. The `parse()` function expects a **string** input and returns `null` when given an object:

```javascript
// Testing EmailAddressParser.parse() behavior:
EmailAddressParser.parse('test@example.com')  // => {address: 'test@example.com', name: undefined}
EmailAddressParser.parse({address: 'test@example.com'})  // => null
```

This causes downstream failures when trying to access `addresses.from.name` on a null value.

### Impact

This bug affects **all transactional emails** that don't explicitly specify a `from` address:
- **Staff device verification (2FA) emails** - Users cannot log in when `staffDeviceVerification` is enabled
- **Password reset emails** - Users cannot reset their passwords  
- **Member magic link emails** - Members cannot sign in
- Any other system emails using the default sender

The error message shown to users is unhelpful: "Failed to send email. Please check your site configuration and try again." - even when the SMTP configuration is correct.

### The Fix

Convert the email object to a string using `EmailAddressParser.stringify()` before passing it to `getAddressFromString()`:

```javascript
// Before (buggy):
requestedFromAddress = emailAddress.service.defaultFromEmail;

// After (fixed):
const defaultEmail = emailAddress.service.defaultFromEmail;
requestedFromAddress = EmailAddressParser.stringify(defaultEmail);
```

### Test Added

Added a test case in `GhostMailer.test.js` that verifies emails are sent correctly when no `from` address is provided, which exercises the `defaultFromEmail` code path.

## Test Plan

1. Enable `staffDeviceVerification` in config
2. Configure SMTP settings (any provider)
3. Attempt to log in to Ghost Admin
4. Verify that the verification code email is sent successfully
5. Also test password reset and member magic link emails

## Reproduction Steps (for the bug)

1. Set up Ghost with SMTP configured (not Mailgun)
2. Enable `staffDeviceVerification: true` in config
3. Try to log in to Ghost Admin
4. Observe the "Failed to send email" error even though SMTP credentials are correct

---

> [!NOTE]
> **Fix transactional email sender handling**
> 
> - In `ghost-mailer.js`, stringifies `defaultFromEmail` before passing to `getAddressFromString`, ensuring correct fallback to `"<site title>" <noreply@<domain>>`
> - Adds unit test in `ghost-mailer.test.js` verifying default `from` behavior when none is provided
> - Updates E2E `member-signup.test.ts` to expect `From.Name` as site title and `From.Address` with `noreply@`
> - Adds Romanian (`ro`) newsletter translations in `i18n/locales/ro/newsletter.json`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3b81225f26f446a2c0545ea02a32f9876752e2bb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


9larsons:
Thanks for the contribution! After digging into this I'm going to close. I'm not seeing a path where the scenario this guards against materializes in practice. When `defaultFromEmail` is something other than a plain string, the existing fallback path in the mail service already handles the failure mode — `getAddressFromString`'s result is only used when it parses successfully, so the stringify here adds a defensive layer in front of behavior that's already safe.

Without a reproducible case where the current code misbehaves, I'd rather not add the extra transformation. If you've actually hit a crash or a malformed From header in the wild, please open an issue with the config shape that caused it please reopen and I'll gladly revisit.

PR #25604, opened 2025-12-03
Fixed broken member signup test, expecting wrong From name

ibalosh:
**what**

* Fixed member signup test that was failing on welcome email since the following [PR got merged](https://github.com/TryGhost/Ghost/commit/0e08ace799ef8fd9da7b6a52f9e059f3157d4e27)
* The solution was to use blog title as a from name, instead of generic `Ghost` name that was used before

### DIFF
--- a/e2e/tests/public/member-signup.test.ts
+++ b/e2e/tests/public/member-signup.test.ts
@@ -63,7 +63,7 @@ test.describe('Ghost Public - Member Signup', () => {
 
         latestMessage = await retrieveLatestEmailMessage(emailInbox);
 
-        expect(latestMessage.From.Name).toContain('Ghost');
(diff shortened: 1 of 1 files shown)


rob-ghost (approved):

ibalosh:
**For some context**

1. We create [account owner / user in Playwright fixture](https://github.com/TryGhost/Ghost/blob/d620b3e281fca1af1a63fc47fdf2f1ccd9ada537/e2e/helpers/playwright/fixture.ts#L105)
2. The created user will be created with defaults like [blog title here](https://github.com/TryGhost/Ghost/blob/b589cb6dc267eaa0b7d6d546fc90fc3e59abeadc/e2e/data-factory/factories/user-factory.ts#L24)
3. That is from where the From name comes from

PR #25609, opened 2025-12-03
Updated welcome emails to run on member creation, periodic job for cleanup

troyciesco:
ref https://linear.app/ghost/issue/NY-821

- makes it so if welcome emails are configured and a member signs up via portal, a welcome email will send right away instead of waiting for the periodic job
- checking `this.processing` prevents multiple instantiations; and because of how the batching in the outbox processing works, new members created while the outbox is processing will get picked up
- we still have the periodic (5min) job that runs to process the outbox - this will be useful for retries and making sure entries don't get missed

---

> [!NOTE]
> Immediately triggers the member welcome email job after creating a member and increases job retries from 0 to 1.
> 
> - **Backend**
>   - **Immediate welcome email dispatch**: In `MemberRepository.create`, after adding the outbox `MemberCreatedEvent`, dispatches `StartMemberWelcomeEmailJobEvent` to start processing right away (when `memberWelcomeEmailTestInbox` is enabled and source is `member`).
> - **Jobs**
>   - **Retries**: Bumps `MAX_RETRIES` in `member-welcome-emails` job `constants.js` from `0` to `1`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ecf2b45eab40bb05ff1ac091aa4f27b04d4e9ae7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ghost/core/test/unit/server/services/member-welcome-emails/index.test.js
```diff
@@ -0,0 +1,94 @@
+const sinon = require('sinon');
```
troyciesco: This test file will move (and maybe look a bit different, not sure) when i decouple the outbox from the member welcome email service a bit better, but this provides a starting point for that at least

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -360,6 +361,8 @@ module.exports = class MemberRepository {
                     })
                 }, {transacting});
 
+                this.dispatchEvent(StartMemberWelcomeEmailJobEvent.create());
```
troyciesco: Originally i had `MemberWelcomeEmailServiceWrapper` subscribe to the`MemberCreatedEvent`, but in looking into decoupling outbox from member welcome emails in [this PR](https://github.com/TryGhost/Ghost/pull/25610) i think dispatching this here makes more sense (in the next PR, it'll be `StartOutboxProcessingEvent`, and i think it makes sense that we want to explicitly start that right after writing to the outbox here)

ghost/core/test/unit/server/services/member-welcome-emails/index.test.js
```diff
@@ -0,0 +1,94 @@
...
+
+describe('Member Welcome Emails Service', function () {
+    let service;
+    let processOutboxStub;
+    let configStub;
+    let loggingStub;
+    let jobsStub;
+
+    beforeEach(function () {
+        service = rewire('../../../../../core/server/services/member-welcome-emails/index.js');
```
cmraible: note (non-blocking): we have a few different patterns for testing services, and I _think_ this (using rewire to pass stubbed dependencies) is one of the "older" ones. I should have called this out when we originally created the member-welcome-emails service, but it escaped my brain at the time! I definitely wouldn't block this PR on this since it would require a refactor of the service itself.

The pattern that we've aligned on for this is to pass all of the dependencies of the service in the constructor (manual dependency injection essentially). This way we can create stubbed dependencies and just pass them into the constructor when creating the service in unit tests, removing the need for something like rewire. An example would be the MembersBREADService: https://github.com/TryGhost/Ghost/blob/NY-821_send-member-welcome-email-right-away/ghost/core/core/server/services/members/members-api/services/MemberBREADService.js.

This is probably a good topic for the engineering forum, since I'm not sure if/where this is documented, and it would be really nice to formally align on our current best practices for instantiating services.

cmraible (approved):
Very nice, worked like a charm - welcome email in my inbox as soon as I signed up 🙌

cmraible (approved):

PR #25610, opened 2025-12-03
Decoupled outbox processing from welcome emails

troyciesco:
ref https://linear.app/ghost/issue/NY-824

- this decouples the processing of the outbox from the member welcome emails
- outbox has a handler pattern so it knows what to do based on certain events (like `MemberCreated`)
- this will allow us to reuse the member welcome email sending elsewhere too, like a "test email" in the UI

---

> [!NOTE]
> Introduces a generic outbox service and handler pipeline, enqueues `MemberCreatedEvent` on member creation, and routes welcome emails via the outbox instead of a dedicated service.
> 
> - **Outbox Service**
>   - Add generic outbox service in `core/server/services/outbox` with event `StartOutboxProcessingEvent`, scheduler, and job `outbox-job.js`.
>   - Implement batch processing (`jobs/lib/process-outbox.js`, `process-entries.js`) with handler dispatch (e.g., `handlers/member-created.js`).
>   - Centralize logging under `OUTBOX_LOG_KEY` and rename scheduling to `process-outbox`.
> - **Member Creation Flow**
>   - On member create, write `MemberCreatedEvent` to `outbox` and dispatch `StartOutboxProcessingEvent` (`MemberRepository.js`).
> - **Welcome Emails**
>   - Extract reusable mail config (`member-welcome-emails/get-mail-config.js`) and sender (`send-member-welcome-email.js`), exported via `member-welcome-emails/index.js`.
>   - Route welcome email sending through outbox handler (`outbox/handlers/member-created.js`); fix template import paths.
> - **Boot**
>   - Initialize `outbox` service during background services (`core/boot.js`); remove `member-welcome-emails` service init.
> - **Tests**
>   - Add/adjust tests for outbox processing and service behavior; update unit test to new `send-member-welcome-email` path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e51cda2f011a0527fed06250758b40f783c9d66f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -359,6 +361,8 @@ module.exports = class MemberRepository {
                     })
                 }, {transacting});
 
+                this.dispatchEvent(StartOutboxProcessingEvent.create());
```
troyciesco: i'd say this is probably the main "change" in this cleanup PR - before we had the job subscribing to the `MemberCreatedEvent`, but since the job is now less tied to welcome emails as a whole and is more about processing the outbox, i think this makes more sense. 

Benefit is that when processing the outbox we don't have to double-check info about whether the member creation should even trigger an outbox process, like [here](https://github.com/TryGhost/Ghost/pull/25609/files#diff-8a88bd4716b7919bf85ad9e994c3174849cf68e18e8d6ed5955cbae00f8019c8R25)

Might be something we change further later

troyciesco: actually, i think i'm even just gonna do this right away in that original PR lol

ghost/core/core/server/services/outbox/jobs/lib/process-entries.js
```diff
@@ -0,0 +1,95 @@
...
+
+    await db.knex('outbox')
+        .where('id', entryId)
+        .update({
+            status: newStatus,
+            retry_count: newRetryCount,
+            last_retry_at: db.knex.raw('CURRENT_TIMESTAMP'),
+            message: truncatedMessage,
+            updated_at: db.knex.raw('CURRENT_TIMESTAMP')
+        });
```
cmraible, line 30: is there a reason we're using knex directly here vs using the outbox model? I don't think it's a problem for any reason, just curious 😄

ghost/core/core/server/services/outbox/jobs/lib/process-entries.js
```diff
@@ -0,0 +1,95 @@
+const logging = require('@tryghost/logging');
+const {MAX_RETRIES, OUTBOX_LOG_KEY} = require('./constants');
+const {OUTBOX_STATUSES} = require('../../../../models/outbox');
+const MemberCreatedEvent = require('../../../../../shared/events/MemberCreatedEvent');
+const memberCreatedHandler = require('../../handlers/member-created');
+
+const EVENT_HANDLERS = {
+    [MemberCreatedEvent.name]: memberCreatedHandler
+};
```
cmraible, line 9: note: really nice! Eventually I think we can extend this to be more dynamic, i.e. maybe the `member-welcome-emails` service can dynamically register these handlers depending on whether there are any `automated_emails` in an `active` state, and maybe the status of the labs flag.

This is great for now though, and gives us a perfect entrypoint for a more dynamic handler registry in the future 👌

ghost/core/core/server/services/outbox/jobs/lib/process-entries.js
```diff
@@ -0,0 +1,95 @@
...
+        .where('id', entryId)
+        .update({
+            status: OUTBOX_STATUSES.COMPLETED,
+            message: 'Processed, but failed to delete outbox entry',
+            updated_at: db.knex.raw('CURRENT_TIMESTAMP')
+        });
+}
+
+async function processEntry({db, entry, mailConfig}) {
+    const handler = EVENT_HANDLERS[entry.event_type];
```
cmraible, line 44: suggestion (non-blocking): wondering if we should support having multiple handlers for the same event 🤔

We don't need this right now (hence non-blocking) but in the future with more customizable automations we could definitely have more than one "action" associated with a single "trigger"

cmraible, line 44: suggestion (non-blocking): wondering if we should support having multiple handlers for the same event 🤔

We don't need this right now (hence non-blocking) but in the future with more customizable automations we could definitely have more than one "action" associated with a single "trigger"

cmraible (approved):
Nice work! This feels like a huge step in the right direction for future extensibility. I left a few comments, but mostly just food for thought.

cmraible:
btw: E2E tests should be passing now after a rebase, they're failing for an unrelated reason

ghost/core/core/server/services/outbox/jobs/lib/process-entries.js
troyciesco: ahhh it's because before when the worker thread was running this I couldn't use the model (or maybe it was that it was too cumbersome to import, something like that).

I made a backlog ticket to look at this once the critical path for the feature set is done, I wonder if there'd be any sort of noticeable difference in performance/on a flamegraph etc

ghost/core/core/server/services/outbox/jobs/lib/process-entries.js
troyciesco: yeah i agree, i wondered this too!

PR #25611, opened 2025-12-03
🐛 Fixed stale tier data setting members to comped status

mike182uk:
ref https://linear.app/ghost/issue/ONC-1305

  When a member's subscription was cancelled while an admin had the member
  page open, saving the member (e.g. changing a note) would send stale tier
  data to the API. The backend interpreted this as adding a complimentary
  tier, incorrectly changing the member's status from 'free' to 'comped'

  The fix removes tiers from the member serializer payload since tier changes
  are handled through dedicated API calls (add/remove complimentary), not
  through normal member saves

---

> [!NOTE]
> Prevents stale tier data from comping members by removing `tiers` from the member save payload and only processing tier changes in mirage when explicitly provided, with added regression and serializer unit tests.
> 
> - **Serializer (`ghost/admin/app/serializers/member.js`)**:
>   - Remove `tiers` from serialized payload; continue excluding `stripe`, `geolocation`, `status`, `last_seen_at`, `comped`.
> - **Mirage (`ghost/admin/mirage/config/members.js`)**:
>   - Guard `attrs.tiers` access and only process removals when tiers are explicitly provided.
>   - Set `member.status = 'comped'` when adding complimentary tier; manage tiers/subscriptions via relationship methods.
>   - Exclude `tiers`/`subscriptions` from `update()` payload.
> - **Tests**:
>   - Acceptance: add regression test ensuring saving with stale tier data does not set status to `comped`; minor helper import.
>   - Unit: add serializer tests asserting excluded fields (`stripe`, `geolocation`, `status`, `last_seen_at`, `comped`, `tiers`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 87423178088dd5847e5234a9b2c24449ee0aa939. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


allouis (approved):
I think this works to fix the bug - but the bug is still present for if e.g. the email or name is updated whilst the page is open in the admin right?

Maybe a more thorough fix would be to compare the updated_at or something and reject if it's older? I realise these are edgecases though, and right now the subscription one is more likely due to Ghost not being able to cancel immediately from the admin

mike182uk:
Nice one @fabien

I did a quick spike on the more comprehensive solution:

Ghost already has collision detection via `@tryghost/bookshelf-collision`, but it's only enabled for posts (it compares `updated_at` timestamps and throws `UpdateCollisionError` if stale)

To enable for members, I think it would be a case of:

- Frontend - Send `updated_at`, handle conflict errors in UI
- Backend - Extend collision plugin or add logic to `MemberRepository`
- UX - Design what happens on conflict (i.e error message or force reload)

BUT this is obviously a much bigger change that will require a little more thought

Given this fixes the billing related issue which is more common + impactful, I'm gonna merge as is, and create a follow up for a more general concurrency handling solution 👍

PR #25616, opened 2025-12-04
Removed E2E test for locations tab

cmraible:
no issue

This tab has been removed from the analytics UI, so we no longer need this test. It was able to be merged with a failing test because the E2E tests are not currently taken into account by this repo's branch protection, which I'll fix in a separate PR. For now, this gets the E2E tests back into a passing state.


cmraible:
@ibalosh @9larsons going to merge this one without review since this test is failing on main currently. The locations tab was removed from the UI, so we just don't need this test anymore.

PR #25623, opened 2025-12-04
Made admin forward feature flag work for self-hosting

jonatansberg:
This PR ensures we're copying the admin forward build assets to the location expected by Ghost core and conditionally serves the correct entrypoint based on the existing feature flag. Since the admin forward assets is a superset of the Ember assets, this should be safe. 

I've also updated the development Docker image as well as the archive script to correctly copy the assets. Finally, I removed some leftovers from a previous integration of the admin forward assets based on the `USE_REACT_SHELL` env var.

### DIFF
--- a/apps/admin/package.json
+++ b/apps/admin/package.json
@@ -64,6 +64,10 @@
         ]
       },
       "build": {
+        "outputs": [
+          "{projectRoot}/dist",
+          "{workspaceRoot}/ghost/core/core/built/admin"
+        ],
         "dependsOn": [
           "build",
           {
--- a/Dockerfile
+++ b/Dockerfile
@@ -210,9 +210,6 @@ COPY --from=posts-builder /home/ghost/apps/posts/dist apps/posts/dist
 COPY --from=portal-builder /home/ghost/apps/portal/umd apps/portal/umd
 COPY --from=admin-x-settings-builder /home/ghost/apps/admin-x-settings/dist apps/admin-x-settings/dist
 COPY --from=activitypub-builder /home/ghost/apps/activitypub/dist apps/activitypub/dist
-COPY --from=admin-ember-builder /home/ghost/ghost/admin/dist ghost/admin/dist
-COPY --from=admin-ember-builder /home/ghost/ghost/core/core/built/admin ghost/core/core/built/admin
-RUN rm -rf apps/admin/dist
-COPY --from=admin-react-builder /home/ghost/apps/admin/dist apps/admin/dist
+COPY --from=admin-react-builder /home/ghost/ghost/core/core/built/admin ghost/core/core/built/admin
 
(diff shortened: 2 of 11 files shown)


apps/admin/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "tsc -b && vite build",
+    "build": "tsc -b && vite build && node scripts/copy-admin-assets.mjs",
```
rob-ghost: **Question:** Is there a reason the copy logic needs to be a separate script? The `vite-ember-assets.ts` plugin already has a `closeBundle()` hook that copies Ember assets into the React build output. This could be extended to also copy the merged result back to ghost/core:
```typescript
closeBundle() {
    if (config.command === 'build') {
        const emberAssetsPath = path.resolve(GHOST_ADMIN_PATH, 'assets');
        const buildAssetsPath = path.resolve(config.build.outDir, 'assets');

        // Existing: copy Ember assets into React build
        fs.cpSync(emberAssetsPath, buildAssetsPath, { recursive: true });

        // New: copy merged build back to ghost/core
        fs.cpSync(buildAssetsPath, emberAssetsPath, { recursive: true, force: true });
        fs.copyFileSync(
            path.resolve(config.build.outDir, 'index.html'),
            path.resolve(GHOST_ADMIN_PATH, 'index-forward.html')
        );
    }
}
```

This keeps all asset copying logic in one place and removes a build step. I tested this locally and it produces the same result, but I'm not sure if I'm missing some different behaviour between building Ember and building React which would require the separate script.

rob-ghost (commented):
Aside from the question about if we need another script of if we can sue `vite-ember-assets.ts` (so we can just run `vite build`), I have a suggestion for the future of how we build and bundle these assets.

Longer term, would it be cleaner if packages only wrote to their own `dist/` directories, and didn't copy assets to other packages? Right now `apps/admin` copies the `index{-forward}.html` to `ghost/core/built/admin` during its build stage.

Would a better uni-directional flow be:

1. `ghost/admin` -> `ghost/admin/dist`
2. `apps/admin` -> `apps/admin/dist`
3. `ghost/core` -> pulls from `ghost/admin/dist` + `apps/admin/dist` -> `ghost/core/built/admin`

This probably breaks loading `/ghost` from the running vite dev servers, but shouldn't we be requesting that from the running `ghost/core` such that the labs flags work like they do in production?

I don't think we need to decide this now / make this change, we can proceed with this iteration, but I'd like to discuss this for the future.

apps/admin/package.json
jonatansberg: Good point. Fixed that!

rob-ghost (approved):
Nice work 👏🏻

PR #25624, opened 2025-12-04
Changed `servePublicFile` from middleware to router

rmgpinto:
ref https://linear.app/ghost/issue/BER-3095

- Extracted the public files code from the middleware to the router. This avoids route comparison on every requests Ghost receives.

---

> [!NOTE]
> Refactors public asset serving from middleware to explicit routes, adds theme/privately-aware robots.txt handling, and updates site wiring and tests.
> 
> - **Routers (public files)**:
>   - Introduce `web/routers/serve-public-file.js` with `servePublicFiles(siteApp)` registering explicit routes for `sitemap.xsl`, `public/*` assets (CSS/JS), `.well-known/recommendations.json`, and card assets via `cardAssets.serveMiddleware()`.
>   - Add conditional `/robots.txt` serving that defers to theme file if present and skips when site is private; otherwise serves default `robots.txt`.
>   - Export `servePublicFile` and `createPublicFileMiddleware` (retained behavior) alongside new router.
> - **Theme Engine**:
>   - In `services/theme-engine/active.js`, detect theme `robots.txt` on init and expose `hasRobotsTxt()`.
> - **Site wiring**:
>   - Replace middleware-based public file serving with router-based `servePublicFiles(siteApp)` in `web/site.js`; remove related middleware wiring.
> - **Middleware index**:
>   - Remove `servePublicFile` export.
> - **Tests**:
>   - Update test imports to use router path for `servePublicFile`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 26de7266e0089498065ddf8c073b6cd2ad355a86. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/core/core/frontend/web/routers/serve-public-file.js
+++ b/ghost/core/core/frontend/web/routers/serve-public-file.js
@@ -5,6 +5,9 @@ const errors = require('@tryghost/errors');
 const config = require('../../../shared/config');
 const urlUtils = require('../../../shared/url-utils');
 const tpl = require('@tryghost/tpl');
+const {cardAssets} = require('../../services/assets-minification');
+const themeEngine = require('../../services/theme-engine');
+const settingsCache = require('../../../shared/settings-cache');
 
 const messages = {
     imageNotFound: 'Image not found',
@@ -99,7 +102,6 @@ function createPublicFileMiddleware(location, file, mime, maxAge, options = {}) 
     };
 }
 
-// ### servePublicFile Middleware
 // Handles requests to robots.txt and favicon.ico (and caches them)
 function servePublicFile(location, file, type, maxAge, options = {}) {
     const publicFileMiddleware = createPublicFileMiddleware(location, file, type, maxAge, options);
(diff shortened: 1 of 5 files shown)


ghost/core/core/frontend/web/routers/serve-public-file.js
```diff
@@ -113,6 +115,72 @@ function servePublicFile(location, file, type, maxAge, options = {}) {
...
+        const themeSitemapXslPath = path.join(activeTheme.path, 'sitemap.xsl');
+        fs.access(themeSitemapXslPath, fs.constants.R_OK, (err) => {
+            if (err) {
+                // Theme doesn't have sitemap.xsl, serve default
+                return defaultSitemapXslMiddleware(req, res, next);
+            }
+            // Theme has sitemap.xsl, let staticTheme middleware handle it
+            return next();
+        });
+    });
```
cmraible: I think this is a (unintended?) behavior change. On `main`, Ghost doesn't serve the `sitemap.xsl` from the theme even if it exists in the theme - it always serves the default/hardcoded `sitemap.xsl`.

ghost/core/core/frontend/web/routers/serve-public-file.js
```diff
@@ -113,6 +115,72 @@ function servePublicFile(location, file, type, maxAge, options = {}) {
...
+
+        const themeRobotsTxtPath = path.join(activeTheme.path, 'robots.txt');
+        fs.access(themeRobotsTxtPath, fs.constants.R_OK, (err) => {
+            if (err) {
+                // Theme doesn't have robots.txt, serve default
+                return defaultRobotsTxtMiddleware(req, res, next);
+            }
+            // Theme has robots.txt, let staticTheme middleware handle it
+            return next();
+        });
```
cmraible: suggestion: this logic feels a bit out of place here, and it also adds a file system call on each request to `/robots.txt`. I think it would make more sense to move the "does the active theme have a robots.txt file?" logic into [active.js](https://github.com/TryGhost/Ghost/blob/serve-public-file-router/ghost/core/core/frontend/services/theme-engine/active.js) as a `hasRobotsTxt` getter or similar.

Then this bit could become something like:

```
const activeTheme = themeEngine.getActive();
if (activeTheme?.hasRobotsTxt) {
     return next();
}
return defaultRobotsTxtMiddleware(req, res, next);
```

This would have the added benefit of checking for the robots.txt in the theme once when activated, rather than on each request.

cmraible (changes requested):
A few minor suggestions I think are worth addressing before merging: 

1. I'm not sure how these were included, but the diffs to the source theme submodule shouldn't be included here. I'm not sure how they got here to begin with, they're usually automatically removed from commits with git hooks 🤔

2. This introduces a behavior change to the `sitemap.xsl` handling that I _think_ was unintentional. On `main` we don't allow the theme to override `sitemap.xsl`, but this change carves out an exception to allow overriding it.

3. We should be able to avoid the additional file system call in each request to robots.txt but moving the "does the active theme have a robots.txt file?" logic into the theme engine itself, which also improves the readability / separation of concerns in serve-public-file.js

Other than those notes, it all looks good to me 👌

ghost/core/core/frontend/web/routers/serve-public-file.js
rmgpinto: Fixed in latest push.

ghost/core/core/frontend/web/routers/serve-public-file.js
rmgpinto: Fixed in latest push.

rmgpinto:
I've removed the submodule changes.
All issues addresses now.

cmraible (approved):
LGTM, thanks for addressing the feedback!

PR #25625, opened 2025-12-04
Renamed files to kebab-case for admin design system

ibalosh:
ref https://linear.app/ghost/issue/NY-722/migrate-ghost-codebase-to-kebab-case-file-naming-convention

**what**

* Renamed all files to match to kebab-case format
* Added ESLint kebab-case rule

**why**

* To match our working agreement on filenames in Ghost repo


ibalosh:
hey @peterzimon 

this is just a simple rename in design system. It passes all checks, and I took a look manually of storybook stories, they look fine.

If you will have a minute or two, would love if you could take a quick glance through it, just in case you can think of something that could break. Thanks

peterzimon (approved):

PR #25626, opened 2025-12-04
Updated member welcome email with automated_email content

troyciesco:
ref https://linear.app/ghost/issue/NY-773

- Updates member welcome emails so that they fetch content from `automated_emails` instead of hard-coded
- Includes logic to check if the automated email is active
- Cleans up some of the boundaries between member welcome emails and the outbox; outbox doesn't need to worry about the mail configuration anymore


ghost/core/core/server/services/member-welcome-emails/service.js
```diff
@@ -0,0 +1,105 @@
...
+
+        const template = this.#templates[templateType];
+
+        if (!template) {
+            throw new errors.IncorrectUsageError({
+                message: MESSAGES.NO_EMAIL_TEMPLATE
+            });
+        }
+
+        if (template.status !== 'active') {
```
troyciesco: I added this part with the thinking that I want to avoid a scenario where an outbox entry is added while a template is active, and somehow, like maybe if the immediate send fails, and if the template is then sent to inactive, I don't then want this outbox row to get processed on the next run. 

And if the template is reset to active like a week later, I don't want to then reprocess the outbox row. 

So right now I'm just throwing an error, but I'm wondering if I should like actually delete the row or something. I feel like this is something we'll probably fine-tune with usage, and it's very much an edge case.

ghost/core/core/server/services/member-welcome-emails/service.js
```diff
@@ -0,0 +1,105 @@
...
+    #renderer;
+    #templates = {free: null, paid: null};
+
+    constructor() {
+        emailAddressService.init();
+        this.#mailer = new mail.GhostMailer();
+        this.#renderer = new MemberWelcomeEmailRenderer();
+    }
+
+    async loadTemplates() {
```
troyciesco: ironic that templates makes sense here with our back and forth around naming 😂 i think i might change it to either `loadAutomatedEmails` or `loadMemberWelcomeEmails` or something though

ghost/core/core/server/services/member-welcome-emails/service.js
```diff
@@ -0,0 +1,105 @@
...
+            });
+        }
+
+        await this.#mailer.send({
+            to: toEmail,
+            subject,
+            html,
+            text,
+            forceTextContent: true
+        });
```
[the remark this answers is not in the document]
troyciesco: This will be addressed in the follow-up PR https://github.com/TryGhost/Ghost/pull/25637

ghost/core/core/server/services/member-welcome-emails/MemberWelcomeEmailRenderer.js
```diff
@@ -0,0 +1,80 @@
...
+        try {
+            content = await lexicalLib.render(lexical, {target: 'email'});
+        } catch (err) {
+            throw new errors.IncorrectUsageError({
+                message: MESSAGES.INVALID_LEXICAL_STRUCTURE,
+                context: err.message
+            });
+        }
+
+        const memberName = member.name || 'there';
```
cmraible, line 41: lol idk why but this made me laugh

ghost/core/core/server/services/member-welcome-emails/MemberWelcomeEmailRenderer.js
```diff
@@ -0,0 +1,80 @@
...
+                name: memberName,
+                email: member.email || '',
+                firstname: firstName
+            },
+            siteTitle: siteSettings.title,
+            siteUrl: siteSettings.url,
+            accentColor: siteSettings.accentColor
+        };
+
+        const contentWithReplacements = this.Handlebars.compile(content)(templateData);
```
cmraible, line 59: interesting 🤔 so if the lexical includes i.e. {{member.name}}, this should substitute it for us?

ghost/core/core/server/services/outbox/handlers/member-created.js
```diff
@@ -1,10 +1,12 @@
...
+const memberWelcomeEmailService = require('../../member-welcome-emails/service');
 
 const LOG_KEY = `${OUTBOX_LOG_KEY}[MEMBER-WELCOME-EMAIL]`;
 
-async function handle({payload, mailConfig}) {
-    const sendMemberWelcomeEmail = require('../../member-welcome-emails').sendMemberWelcomeEmail;
-    await sendMemberWelcomeEmail({payload, mailConfig});
+async function handle({payload}) {
+    // TODO: derive memberStatus from payload when paid welcome emails are added
+    const memberStatus = 'free';
```
cmraible, line 8: Does this mean that we're currently just sending the free welcome email to all members, regardless of status?

ghost/core/test/unit/server/services/member-welcome-emails/service.test.js
```diff
@@ -0,0 +1,328 @@
...
+
+        serviceModule = rewire('../../../../../core/server/services/member-welcome-emails/service');
+        serviceModule.__set__('logging', loggingStub);
+        serviceModule.__set__('config', configStub);
+        serviceModule.__set__('settingsCache', settingsCacheStub);
+        serviceModule.__set__('urlUtils', urlUtilsStub);
+        serviceModule.__set__('emailAddressService', emailAddressServiceStub);
+        serviceModule.__set__('mail', mailStub);
+        serviceModule.__set__('MemberWelcomeEmailRenderer', MockRenderer);
+        serviceModule.__set__('AutomatedEmail', automatedEmailStub);
```
cmraible: note (non-blocking): I don't have a better suggestion, but just wanted to call out that this is _a lot_ of stubs, which then couples these tests quite closely with the implementation. Wondering if we need all of these tests, or if some of these might be better placed as integration tests 🤔

ghost/core/test/unit/server/services/outbox/jobs/lib/process-entries.test.js
```diff
@@ -0,0 +1,301 @@
...
+
+        dbStub = {knex: knexFn};
+
+        processEntries = rewire('../../../../../../../core/server/services/outbox/jobs/lib/process-entries');
+        processEntries.__set__('logging', loggingStub);
+        processEntries.__set__('OUTBOX_STATUSES', OUTBOX_STATUSES);
+        processEntries.__set__('EVENT_HANDLERS', {
+            MemberCreatedEvent: mockHandler
+        });
+    });
```
cmraible: note (non-blocking): similar to above comment, this is a lot of stubs. I wonder if most of these cases are already covered adequately by integration tests (and if not, maybe they could be?)

ghost/core/test/unit/server/services/outbox/jobs/lib/process-outbox.test.js
```diff
@@ -0,0 +1,247 @@
...
+        processOutbox = rewire('../../../../../../../core/server/services/outbox/jobs/lib/process-outbox');
+        processOutbox.__set__('logging', loggingStub);
+        processOutbox.__set__('db', dbStub);
+        processOutbox.__set__('processEntries', processEntriesStub);
+        processOutbox.__set__('memberWelcomeEmailService', memberWelcomeEmailServiceStub);
+        processOutbox.__set__('OUTBOX_STATUSES', OUTBOX_STATUSES);
+        processOutbox.__set__('MESSAGES', MESSAGES);
+        processOutbox.__set__('BATCH_SIZE', 100);
+        processOutbox.__set__('MAX_ENTRIES_PER_JOB', 1000);
+    });
```
cmraible: note (non-blocking): same thing here - given that we have the integration tests for this job, I'm wondering how much value these unit tests add, given that we have to mock so much to test it effectively. Seems like these unit tests are really closely coupled with the implementation.

cmraible (commented):
Overall looking pretty good, nice work! I think the implementation is pretty solid. Most of my comments are on the unit tests, which generally feel quite closely coupled to the implementation. I'm wondering if we can simplify them or eliminate some of them or change them to more integration-y tests so we can focus more on the behavior that the internal implementation details.

ghost/core/core/server/services/member-welcome-emails/MemberWelcomeEmailRenderer.js
troyciesco: yeah! results in something like this:

Screenshot 2025-12-04 at 3 20 58 PM

ghost/core/core/server/services/member-welcome-emails/MemberWelcomeEmailRenderer.js
troyciesco: lol same - i'm actually not sure if we want it, because if someone uses {{member.name}} for anything other than the greeting (like "Hi Jon Doe," vs "Hi there,") it'll look weird.

I'm gonna add a ticket to check on this next week

ghost/core/core/server/services/outbox/handlers/member-created.js
troyciesco: yeah in my head i was thinking that if we don't have the differentiation on the frontend yet, it makes sense that you'd want to send a welcome email to everyone...though we do refer to things as "free" everywhere else 🤔

ghost/core/core/server/services/outbox/jobs/lib/process-outbox.js
```diff
@@ -36,19 +38,16 @@ async function fetchPendingEntries({db, batchSize, jobStartISO}) {
...
 
 async function processOutbox() {
-    const db = require('../../../../data/db');
-    const getMailConfig = require('../../../member-welcome-emails/get-mail-config');
-
     const jobStartMs = Date.now();
     const jobStartISO = new Date(jobStartMs).toISOString().slice(0, 19).replace('T', ' ');
 
-    let mailConfig;
+    memberWelcomeEmailService.init();
```
troyciesco, line 44: it feels like there's probably a better place for this so the outbox doesn't have to care at all, but it also keeps us from processing jobs that are going to fail anyway if the email service doesn't work for some reason. 

but i think we can probably clean this up as we add more handlers etc, outbox is way less coupled to welcome emails in general now

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -93,6 +96,7 @@ module.exports = class MemberRepository {
         this.tokenService = tokenService;
         this._newslettersService = newslettersService;
         this._labsService = labsService;
+        this._AutomatedEmail = AutomatedEmail;
```
cmraible, line 99: note: this is fine for now, but I don't think the MemberRepository should "know" about automated emails. Haven't thought deeply about this, but off the cuff here's what I'm thinking:

- Rather than passing the Outbox model and AutomatedEmail model to the MemberRepository, we should pass the Outbox _service_
- The Outbox service would have an `emit(event_type, payload, transaction)` method, which the MemberRepository would _always_ call
- The Outbox service would then decide if it should _actually_ publish to the outbox (depending on whether any handlers are registered, i.e. if any automated_emails exist)
- ... profit?

ghost/core/core/server/services/outbox/handlers/member-created.js
cmraible: Cool, I think that's fine for now

ghost/core/core/server/services/outbox/jobs/lib/process-outbox.js
cmraible: Yeah agreed 🙌 this is where I think having some kind of registry for the outbox will come in handy, so it can decide if anything is listening to a particular event type, without having to know anything about the subscriber's implementation

cmraible (approved):
Looks good to me! Tests are much easier to reason about now, thanks for that!

Left one note for future refactor opportunity, but nothing blocking 🙌

PR #25629, opened 2025-12-04
🐛 Fixed contributor profile editor navigation to posts page

rob-ghost:
## Summary

Contributors couldn't close their profile editor modal because a race condition would redirect them back. The modal's external navigation to `/posts` was being overridden by a useEffect that detected the route change and redirected contributors back to their profile.

the issue was AdminXSettings is always mounted and it has a side effect to re-direct to the user profile page if the User isn't permitted to see all the settings. This effect was running all the time (because the settings are always mounted) and it was affecting the navigation away from the profile.

Rather than amend the effect, or conditionally mount the settings (too many changes), I opted to lift the permissions for the settings to the router level (which should be responsible for routes and their permissions) which fixes the issue.

---

> [!NOTE]
> Restricts Authors/Contributors to their own profile via Ember routing, removes React-side redirect, updates modal close navigation, and adapts tests.
> 
> - **Routing/Access Control**:
>   - Add Ember `beforeModel` guard in `ghost/admin/app/routes/settings-x.js` to redirect Authors/Contributors to `staff/<slug>` and block other settings paths.
> - **React UI**:
>   - In `main-content.tsx`, stop React-side redirects; hide main settings entirely for non-`canAccessSettings` users so only the profile modal renders via routing.
>   - In `user-detail-modal.tsx`, change close behavior for non-settings users to external `''` route to let the shell handle navigation.
> - **Tests**:
>   - Update acceptance tests to navigate directly to `/#/settings/staff/owner` and verify profile-only view for Author/Contributor; remove URL assertions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0623ad83a83c864cb614868653e826119162fede. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/admin-x-settings/src/components/settings/general/user-detail-modal.tsx
+++ b/apps/admin-x-settings/src/components/settings/general/user-detail-modal.tsx
@@ -210,7 +210,8 @@ const UserDetailModalContent: React.FC<{user: User}> = ({user}) => {
         if (canAccessSettings(currentUser)) {
             updateRoute('staff');
         } else {
-            updateRoute({isExternal: true, route: 'analytics'});
+            // Contributors can't access settings, exit to let the shell handle navigation
+            updateRoute({isExternal: true, route: ''});
         }
     }, [currentUser, updateRoute]);
 
--- a/apps/admin-x-settings/test/acceptance/permissions.test.ts
+++ b/apps/admin-x-settings/test/acceptance/permissions.test.ts
@@ -16,20 +16,22 @@ test.describe('User permissions', async () => {
         await expect(page.getByTestId('title-and-description')).toBeHidden();
     });
 
+    // Note: Author/Contributor redirect to profile is handled by the Ember router (settings-x.js),
+    // not by the React app. These tests verify the UI renders correctly when on the profile route.
     test('Authors can only see their own profile', async ({page}) => {
         await mockApi({page, requests: {
             ...globalDataRequests,
             browseMe: {...globalDataRequests.browseMe, response: meWithRole('Author')}
         }});
 
-        await page.goto('/');
+        // Navigate directly to profile route using hash-based routing
+        // (Ember router handles redirect in production)
+        await page.goto('/#/settings/staff/owner');
 
         await expect(page.getByTestId('user-detail-modal')).toBeVisible();
         await expect(page.getByTestId('sidebar')).toBeHidden();
         await expect(page.getByTestId('users')).toBeHidden();
         await expect(page.getByTestId('title-and-description')).toBeHidden();
-
-        expect(page.url()).toMatch(/\/owner$/);
     });
 
     test('Contributors can only see their own profile', async ({page}) => {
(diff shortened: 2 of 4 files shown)


apps/admin-x-settings/src/components/settings/general/user-detail-modal.tsx
```diff
@@ -210,7 +210,8 @@ const UserDetailModalContent: React.FC<{user: User}> = ({user}) => {
         if (canAccessSettings(currentUser)) {
             updateRoute('staff');
         } else {
-            updateRoute({isExternal: true, route: 'analytics'});
+            // Contributors can't access settings, redirect to posts page
+            updateRoute({isExternal: true, route: 'posts'});
```
kevinansfield: Nit: It might be preferable to use `route: ''` so that the router can then redirect to whichever route we currently consider the landing page for contributors

kevinansfield (approved):

apps/admin-x-settings/src/components/settings/general/user-detail-modal.tsx
rob-ghost: I did try that and it didn't work but maybe it was related to the other issue we were seeing, I can try again

rob-ghost: Yep the default route worked this time, 👍🏻

apps/admin-x-settings/test/acceptance/permissions.test.ts
```diff
@@ -38,13 +40,13 @@ test.describe('User permissions', async () => {
...
+        // (Ember router handles redirect in production)
+        await page.goto('/#/settings/staff/owner');
 
         await expect(page.getByTestId('user-detail-modal')).toBeVisible();
         await expect(page.getByTestId('sidebar')).toBeHidden();
         await expect(page.getByTestId('users')).toBeHidden();
         await expect(page.getByTestId('title-and-description')).toBeHidden();
-
-        expect(page.url()).toMatch(/\/owner$/);
     });
```
kevinansfield, line 51: Could these tests be improved? The assertions don't really match the naming which makes them difficult to reason about and potentially misleading.

If the expectation is that opening another user's profile directly in the settings app only renders the logged-in user's profile should we have some assertions that we're showing the right data rather than just that a profile modal is shown? Alternatively if these tests no longer make sense perhaps we should remove them?

rob-ghost: They probably definitely can be improved, however I'd prefer to convert these to E2E tests as soon as we have the ability to assume different roles and delete this suite.

The changes in this PR are just to make them pass after we moved the redirect to the router.

In order to not block the React shell rollout, would you be happy to merge this fix as it stands and we can follow up with E2E tests in [this ticket](https://linear.app/ghost/issue/BER-3112/add-e2e-route-navigation-tests-for-multiple-user-roles)?

kevinansfield: 👍 if we're getting rid of these tests when we add e2e tests for the profile screen then its not a blocker

kevinansfield (approved):

PR #25612, opened 2025-12-03
Moved E2E test suite from ci-docker.yml to main ci.yml workflow

cmraible:
no issue

Currently the E2E tests are not included in the branch protection rules, which allows a PR to merge with failing E2E tests. This is the case because the E2E tests are in a separate workflow file and not included in the "All required tests passed or skipped" check in ci.yml.

This PR moves the E2E tests (and its dependencies) into the main `ci.yml` workflow, and adds them to the "All required tests passed or skipped" check, which will enforce passing E2E tests before merging PRs and deploying Admin.

Note: when re-running this multiple times, the only failures I saw were the global setup timing out at 60 seconds, so this also increases the timeout for the global setup to mitigate this.

---

> [!NOTE]
> Consolidates Docker build/inspect and sharded E2E tests into `ci.yml`, adds E2E to required checks, and increases the E2E global setup timeout to 2 minutes.
> 
> - **CI**:
>   - Remove `/.github/workflows/ci-docker.yml`.
>   - Add to `/.github/workflows/ci.yml`:
>     - `job_docker_build`, `job_inspect_image` for Docker image build/publish and analysis.
>     - `job_e2e_tests` (sharded Ember/React) and `job_merge_e2e_reports` for Playwright report aggregation.
>   - Include `job_e2e_tests` in `job_required_tests` to enforce passing E2E before deploys.
> - **E2E**:
>   - Increase global setup timeout to 2 minutes in `e2e/tests/global.setup.ts`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c0cf7b791da2f2e764811c8fe96f35f2758362db. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #25632, opened 2025-12-04
Merge ci workflows with merge report

ibalosh:
Related to [following PR ](https://github.com/TryGhost/Ghost/pull/25612) - this PR updates report merge for e2e on top of related PR.

**what**

from Chris notes:

This PR moves the E2E tests (and its dependencies) into the main ci.yml workflow, and adds them to the "All required tests passed or skipped" check, which will enforce passing E2E tests before merging PRs and deploying Admin.

Note: when re-running this multiple times, the only failures I saw were the global setup timing out at 60 seconds, so this also increases the timeout for the global setup to mitigate this.

In addition to Chrises PR in this branch,  e2e reports for React, Ember are working correctly and are separated: 

* merged properly
* you can see show report handy commands in comments (like here in PR below) if any of e2e tests fail

**why**

Currently the E2E tests are not included in the branch protection rules, which allows a PR to merge with failing E2E tests. This is the case because the E2E tests are in a separate workflow file and not included in the "All required tests passed or skipped" check in ci.yml.

Also, you would want to be able to easily review traces/reports to see why tests failed.


cmraible:
Looks like the merge reports jobs are failing if the tests pass now 😂

ibalosh:
> Looks like the merge reports jobs are failing if the tests pass now 😂

yo @cmraible , fixed, it seems it was a dependency issue in CI (I forgot to change the `needs` section to depend on current job). I tested for failure, and for pass in CI.

cmraible (approved):
Thanks @ibalosh, looks good to me 👌

PR #25634, opened 2025-12-04
Added device endpoint and filter to analytics endpoints

9larsons:
ref https://linear.app/ghost/issue/NY-827/
- added api_top_devices
- added device as a session-level field

User agent/device is a session-level metric, so it's been included in the session_data mv.

---

> [!NOTE]
> Adds a top devices endpoint and promotes device to a session-level field with filtering, updating pipes, fixtures, and tests.
> 
> - **Tinybird endpoints**:
>   - `endpoints/api_top_devices.pipe`: new endpoint returning `device` with visit counts; integrates with `filtered_sessions` and supports pagination.
> - **Pipes/data model**:
>   - `pipes/mv_hits.pipe`: extract `device` from payload and default to `unknown`; remove UA-derived device; propagate `device` downstream.
>   - `pipes/mv_session_data.pipe`: aggregate and expose session-level `device` via `argMin`.
>   - `pipes/filtered_sessions.pipe`: add session-level filter `device` and update description.
> - **Fixtures & tests**:
>   - `fixtures/analytics_events.ndjson`: add `device` field values (e.g., `desktop`, `bot`).
>   - Add `tests/api_top_devices.yaml`; extend `tests/api_kpis.yaml`, `tests/api_top_locations.yaml`, `tests/api_top_pages.yaml` with device-filter cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 992fc004a274f11ec8e3f6919e6d0a7c9e2333e2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


cmraible (changes requested):
We'll also want to update mv_hits I think, which is still doing the naive user agent parsing in the query itself. The analytics service now does this parsing, so we should be able to directly use `payload.device` rather than parsing the UA in the mv_hits.pipe

9larsons:
> We'll also want to update mv_hits I think, which is still doing the naive user agent parsing in the query itself. The analytics service now does this parsing, so we should be able to directly use `payload.device` rather than parsing the UA in the mv_hits.pipe

Good shout! I want to find a better way to keep that schema in sync with the fixtures we have.

I didn't update them in order to limit the scope of the changes, but ideally we'd vary a few of the sessions in the fixture data rather than having only one bot session. Something to figure out later.

ghost/core/core/server/data/tinybird/pipes/mv_hits.pipe
```diff
@@ -24,7 +24,8 @@ SQL >
         JSONExtractString(payload, 'member_status') as member_status,
         JSONExtractString(payload, 'post_uuid') as post_uuid,
         JSONExtractString(payload, 'post_type') as post_type,
-        lower(JSONExtractString(payload, 'user-agent')) as user_agent,
```
cmraible, line 27: was removing the `lower()` intentional here?

ghost/core/core/server/data/tinybird/pipes/mv_hits.pipe
```diff
@@ -106,15 +107,7 @@ SQL >
...
-        case
-            when match(user_agent, 'wget|ahrefsbot|curl|urllib|bitdiscovery|\+https://|googlebot')
-            then 'bot'
-            when match(user_agent, 'android')
-            then 'mobile-android'
-            when match(user_agent, 'ipad|iphone|ipod')
-            then 'mobile-ios'
-            else 'desktop'
-        END as device,
+        device,
```
cmraible:
```suggestion
        case
            when device = '' then 'unknown'
            else device
        end
```

I'm _pretty sure_ every event should have a device, but just to be defensive we should probably provide a default in case it doesn't exist, or JSONExtractString fails for whatever reason.

cmraible (changes requested):
One small suggestion: just in case some events don't have the device type in the payload, we should provide a default value for device.

ghost/core/core/server/data/tinybird/pipes/mv_hits.pipe
9larsons: Good catch. I meant to just remove it from the device when it got re-added.

cmraible (approved):
LGTM!

PR #25637, opened 2025-12-04
Updated member welcome emails to send to members instead of test inbox

troyciesco:
ref https://linear.app/ghost/issue/NY-822

- The `memberWelcomeEmailsTestInbox` is no longer needed
- Instead member welcome emails will be behind a labs flag
- Also updates the member welcome emails service so that it sends to the member email instead of that test inbox


troyciesco:
closing bc we did this in a different PR

PR #25638, opened 2025-12-04
Refactored `HtmlEditor` component used in settings

cmraible:
ref https://linear.app/ghost/issue/NY-836/add-minimaleditor-component-to-admin-x-design-system

In the member emails section of settings, we need to allow the user to edit the content of the welcome email. Initially we'll just include basic text formatting, but soon we'll want to add additional cards, like buttons for call-to-actions.

Currently this modal uses the HtmlEditor, which already existed and was used in e.g. newsletter settings to edit the footer content. This worked initially, but the HtmlEditor outputs HTML, not Lexical, which breaks when we try to wire it up to the API (which expects lexical).

This PR:
- Creates a new `KoenigEditorBase` component with the common boilerplate needed to launch an instance of the editor
- Updates `HtmlEditor` to be a thin wrapper around `KoenigEditorBase` to add the HTML output functionality
- Exports the `KoenigEditorBase` from the design system so we can create custom editor instances for other use-cases, i.e. for member welcome emails


apps/admin-x-design-system/src/global/form/KoenigEditorBase.tsx
```diff
@@ -0,0 +1,193 @@
...
+            throw response;
+        default:
+            return response;
+        }
+    };
+
+    return {read};
+};
+
+export type EditorResource = ReturnType<typeof loadKoenig>;
```
9larsons: This doesn't seem to be imported anywhere.

apps/admin-x-design-system/src/global/form/KoenigEditorBase.tsx
```diff
@@ -0,0 +1,193 @@
...
+    interface Window {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        '@tryghost/koenig-lexical': any;
+    }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type KoenigInstance = { [key: string]: any };
+
+export const loadKoenig = function (fetchKoenigLexical: FetchKoenigLexical) {
```
9larsons: This doesn't seem to be imported anywhere. Seems like the export got added in the refactor.

9larsons (approved):
LGTM! Couple minor comments, nbd.

cmraible:
Good catch, removed both the unused imports 👍

PR #10101, opened 2018-11-05
Fixed frame context being empty

kirrg001:
no issue

See code comment.
Any functional call e.g. `api.posts.browse()` which does not pass a context, ends in `frame.options.context = {}`. This is expected. But the serializers need to respect and interpret this as a content api access. This is the default behaviour for routing.

PR #25655, opened 2025-12-08
Improve pure black accent color bug in member emails

peterzimon:
ref https://linear.app/ghost/issue/DES-1169/email-cta-for-signups-lacks-contrast-when-received-in-dark-mode

- Transactional member emails had a bug on Apple Mail which caused low contrast buttons when the accent color was set to pure black (`#000000`). This change updates the accent color to its closest gray variant (`#010101`) when set to `#000000` so that the automatic color application is not causing a bug in Apple Mail.


peterzimon:
Closing this PR as the solution results in unpredicted outcomes.

PR #25657, opened 2025-12-08
🐛 Fixed incorrect text color in HTML cards with dark email background

kevinansfield:
closes https://linear.app/ghost/issue/BER-2265/

- simplified `p` selector for setting text color so it captures all paragraphs instead of a long list of complicated specific selectors
  - fixes paragraphs nested inside divs not getting the right background-appropriate text color
- added specific `inherit` text color overrides for cards with fixed background colors
  - fixes a few pieces of text inside cards that have per-card adjustments for the selected card background
- fixed list items in product cards appearing too dark/bright compared to the other card content


vershwal (approved):
Nice one! I tested this on my local :+1:
The HTML cards look good after the fix. Everything else is also looking good.

PR #25660, opened 2025-12-08
Updated welcome email modal to utilize useForm

troyciesco:
ref https://linear.app/ghost/issue/NY-856

- Removes placeholder logic and updates form handling in `welcome-email-modal.tsx` to utilize our `useForm` hook and fixes a few related issues in the process
- Save button now shows "Saved" on successful save instead of closing the modal (closes https://linear.app/ghost/issue/NY-854)
- `cmd + S` now saves changes, without closing modal (closes https://linear.app/ghost/issue/NY-853)
- Hides reply-to if it's the same as from, tweaks spacing (ref https://linear.app/ghost/issue/NY-851)
- Makes it so From and Reply-to aren't editable and are the same as the default newsletter values; also updates styles (closes https://linear.app/ghost/issue/NY-846)

### DIFF
--- a/apps/admin-x-settings/src/components/settings/membership/member-emails/member-email-editor.tsx
+++ b/apps/admin-x-settings/src/components/settings/membership/member-emails/member-email-editor.tsx
@@ -22,9 +22,12 @@ const MemberEmailsEditor: React.FC<MemberEmailsEditorProps> = ({
     // but the API expects a JSON string
     const handleChange = useCallback((data: unknown) => {
         if (onChange && data && typeof data === 'object') {
-            onChange(JSON.stringify(data));
+            const stringified = JSON.stringify(data);
+            if (stringified !== value) {
+                onChange(stringified);
+            }
         }
-    }, [onChange]);
+    }, [onChange, value]);
 
     return (
         <KoenigEditorBase
--- a/apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
+++ b/apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
@@ -1,96 +1,82 @@
 import NiceModal from '@ebay/nice-modal-react';
-import validator from 'validator';
 import {useEffect, useRef, useState} from 'react';
 
 import MemberEmailEditor from './member-email-editor';
-import {Button, Modal, TextField, showToast} from '@tryghost/admin-x-design-system';
+import {Button, Hint, Modal, TextField} from '@tryghost/admin-x-design-system';
+import {useForm, useHandleError} from '@tryghost/admin-x-framework/hooks';
+
+import {getSettingValues} from '@tryghost/admin-x-framework/api/settings';
 import {useCurrentUser} from '@tryghost/admin-x-framework/api/current-user';
 import {useEditAutomatedEmail} from '@tryghost/admin-x-framework/api/automated-emails';
+import {useGlobalData} from '../../../../components/providers/global-data-provider';
 import {useRouting} from '@tryghost/admin-x-framework/routing';
 import type {AutomatedEmail} from '@tryghost/admin-x-framework/api/automated-emails';
 
 interface WelcomeEmailModalProps {
-    emailType?: 'free' | 'paid';
-    automatedEmail?: AutomatedEmail;
+    emailType: 'free' | 'paid';
+    automatedEmail: AutomatedEmail;
 }
 
+const isEmptyLexical = (lexical: string | null | undefined): boolean => {
+    if (!lexical) {
+        return true;
+    }
+    
+    try {
+        const parsed = JSON.parse(lexical);
+        const children = parsed?.root?.children;
+        
+        // Empty if no children or only an empty paragraph
+        if (!children || children.length === 0) {
+            return true;
+        }
+        if (children.length === 1 && 
+            children[0].type === 'paragraph' && 
+            (!children[0].children || children[0].children.length === 0)) {
+            return true;
+        }
+        
+        return false;
+    } catch {
+        return true;
+    }
+};
+
 const WelcomeEmailModal = NiceModal.create<WelcomeEmailModalProps>(({emailType = 'free', automatedEmail}) => {
-    const modal = NiceModal.useModal();
     const {updateRoute} = useRouting();
     const {data: currentUser} = useCurrentUser();
     const {mutateAsync: editAutomatedEmail} = useEditAutomatedEmail();
     const [showTestDropdown, setShowTestDropdown] = useState(false);
     const [testEmail, setTestEmail] = useState(currentUser?.email || '');
     const dropdownRef = useRef<HTMLDivElement>(null);
+    const handleError = useHandleError();
+    const {settings} = useGlobalData();
+    const [siteTitle, defaultEmailAddress] = getSettingValues<string>(settings, ['title', 'default_email_address']);
 
-    // Form state for editable fields
-    const [formData, setFormData] = useState({
-        subject: automatedEmail?.subject || 'Welcome',
-        lexical: automatedEmail?.lexical || '',
-        sender_name: automatedEmail?.sender_name || '',
-        sender_email: automatedEmail?.sender_email || '',
-        sender_reply_to: automatedEmail?.sender_reply_to || ''
+    const {formState, saveState, updateForm, handleSave, okProps, errors} = useForm({
+        initialState: {
+            subject: automatedEmail?.subject || 'Welcome',
+            lexical: automatedEmail?.lexical || ''
+        },
+        savingDelay: 500,
+        onSave: async (state) => {
+            await editAutomatedEmail({...automatedEmail, ...state});
+        },
+        onSaveError: handleError,
+        onValidate: (state) => {
+            const newErrors: Record<string, string> = {};
+
+            if (!state.subject) {
+                newErrors.subject = 'A subject is required';
+            }
+
+            if (isEmptyLexical(state.lexical)) {
+                newErrors.lexical = 'Email content is required';
+            }
+
+            return newErrors;
+        }
     });
-    const [isSaving, setIsSaving] = useState(false);
-    const [errors, setErrors] = useState<{sender_email?: string; sender_reply_to?: string}>({});
-
-    // Track if form has changes
-    const hasChanges = automatedEmail && (
-        formData.subject !== (automatedEmail.subject || '') ||
-        formData.lexical !== (automatedEmail.lexical || '') ||
-        formData.sender_name !== (automatedEmail.sender_name || '') ||
-        formData.sender_email !== (automatedEmail.sender_email || '') ||
-        formData.sender_reply_to !== (automatedEmail.sender_reply_to || '')
-    );
-
-    const updateFormData = (key: keyof typeof formData, value: string) => {
-        setFormData(prev => ({...prev, [key]: value}));
-        // Clear error when user starts typing
-        if (key === 'sender_email' || key === 'sender_reply_to') {
-            setErrors(prev => ({...prev, [key]: undefined}));
-        }
-    };
-
-    const validateForm = (): boolean => {
-        const newErrors: typeof errors = {};
-
-        if (formData.sender_email && !validator.isEmail(formData.sender_email)) {
-            newErrors.sender_email = 'Enter a valid email address';
-        }
-
-        if (formData.sender_reply_to && !validator.isEmail(formData.sender_reply_to)) {
-            newErrors.sender_reply_to = 'Enter a valid email address';
-        }
-
-        setErrors(newErrors);
-        return Object.keys(newErrors).length === 0;
-    };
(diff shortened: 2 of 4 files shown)


apps/admin-x-settings/src/components/settings/membership/member-emails/member-email-editor.tsx
```diff
@@ -22,9 +22,12 @@ const MemberEmailsEditor: React.FC<MemberEmailsEditorProps> = ({
     // but the API expects a JSON string
     const handleChange = useCallback((data: unknown) => {
         if (onChange && data && typeof data === 'object') {
-            onChange(JSON.stringify(data));
+            const stringified = JSON.stringify(data);
+            if (stringified !== value) {
```
troyciesco, line 26: This is so the form knows if the lexical content changes - Koenig seems to call onChange immediately which makes useForm thinks that the form is dirty

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -14,83 +18,68 @@ interface WelcomeEmailModalProps {
     automatedEmail?: AutomatedEmail;
 }
 
+const isEmptyLexical = (lexical: string | null | undefined): boolean => {
```
troyciesco, line 20: I don't know for sure if we actually want this - i'm assuming we wouldn't want an empty welcome email to be possible. 

worth noting that idk where the best spot to delay the error message tho, will follow up.

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -14,83 +18,68 @@ interface WelcomeEmailModalProps {
...
     const {data: currentUser} = useCurrentUser();
     const {mutateAsync: editAutomatedEmail} = useEditAutomatedEmail();
     const [showTestDropdown, setShowTestDropdown] = useState(false);
     const [testEmail, setTestEmail] = useState(currentUser?.email || '');
     const dropdownRef = useRef<HTMLDivElement>(null);
+    const handleError = useHandleError();
+    const {settings} = useGlobalData();
+    const [defaultEmailAddress] = getSettingValues<string>(settings, ['default_email_address']);
+    const {localSettings} = useSettingGroup();
+    const [siteTitle] = getSettingValues(localSettings, ['title']) as string[];
```
troyciesco: not totally sure if these are necessary or if i can do it in the same spot as defaultEmailAddress (was mimicking newsletters)

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -14,83 +18,68 @@ interface WelcomeEmailModalProps {
...
-        setErrors(newErrors);
-        return Object.keys(newErrors).length === 0;
-    };
-
-    const handleSave = async () => {
-        if (!automatedEmail || !validateForm()) {
-            return;
-        }
+            if (isEmptyLexical(formState.lexical)) {
+                newErrors.lexical = 'Content is required for your welcome email';
```
troyciesco: As noted above, not sure if we want this.

(Github made the diff look ridiculous too)

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -168,61 +172,40 @@ const WelcomeEmailModal = NiceModal.create<WelcomeEmailModalProps>(({emailType =
...
                             </div>
                             <Button
-                                color="black"
-                                disabled={!hasChanges || isSaving}
-                                label={isSaving ? 'Saving...' : 'Save'}
-                                onClick={handleSave}
+                                color={okProps.color}
+                                disabled={saveState === 'saving'}
+                                label={okProps.label || 'Save'}
+                                onClick={async () => await handleSave({fakeWhenUnchanged: true})}
```
troyciesco, line 178: this mimics newsletters. could also disable the button until the form has changes too

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -168,61 +172,40 @@ const WelcomeEmailModal = NiceModal.create<WelcomeEmailModalProps>(({emailType =
...
-                                error={Boolean(errors.sender_email)}
-                                hint={errors.sender_email}
-                                maxLength={191}
-                                placeholder='noreply@example.com'
-                                value={formData.sender_email}
-                                onChange={e => updateFormData('sender_email', e.target.value)}
-                            />
+                        <div className='flex grow items-center gap-1'>
+                            <span>{automatedEmail?.sender_name || siteTitle}</span>
+                            <span className='text-grey-700'>{`<${senderEmail}>`}</span>
```
troyciesco, line 186: funny side note that i spent like 5 minutes wondering why my style changes weren't working until i realized this codebase uses "grey" instead of "gray" lol

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -168,61 +172,40 @@ const WelcomeEmailModal = NiceModal.create<WelcomeEmailModalProps>(({emailType =
...
-                            <TextField
-                                className='!h-[34px] w-full'
-                                error={Boolean(errors.sender_reply_to)}
-                                hint={errors.sender_reply_to}
-                                maxLength={191}
-                                placeholder='reply@example.com'
-                                value={formData.sender_reply_to}
-                                onChange={e => updateFormData('sender_reply_to', e.target.value)}
-                            />
+                    {replyToEmail !== senderEmail && (
```
troyciesco, line 189: in the original [static UI PR](https://github.com/TryGhost/Ghost/pull/25503/files#diff-6231a1c8c1e9b4d5a9e06f069407c5801a244eb268434afaec7be88140a4fcb5R111) `replyTo` was hidden if it was the same as sender so i put that back. 

I know we have a ticket to look at how all this works with custom domains etc. so may make sense to look at this again there

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -168,61 +172,40 @@ const WelcomeEmailModal = NiceModal.create<WelcomeEmailModalProps>(({emailType =
...
                         </div>
-                    </div>
-
-                    <div className='-mt-1 flex items-center'>
+                    )}
+                    <div className='flex items-center'>
                         <div className='w-20 font-semibold'>Subject:</div>
                         <div className='grow'>
                             <TextField
-                                className='!h-[34px] w-full'
```
troyciesco, line 219: this style didn't seem necessary

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -116,18 +105,33 @@ const WelcomeEmailModal = NiceModal.create<WelcomeEmailModalProps>(({emailType =
         };
     }, [showTestDropdown]);
 
+    useEffect(() => {
```
troyciesco, line 106: this copies how it's done in other parts of the codebase, except that it includes handleSave (which is in a `useCallback`) in the dependency array so it doesn't run on every render, which i'm pretty sure is the correct thing to do here

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -168,58 +172,38 @@ const WelcomeEmailModal = NiceModal.create<WelcomeEmailModalProps>(({emailType =
...
+                    {replyToEmail !== senderEmail && (
+                        <div className='flex items-center py-0.5'>
+                            <div className='w-20 font-semibold'>Reply-to:</div>
+                            <div className='grow text-grey-700'>
+                                {replyToEmail}
+                            </div>
                         </div>
-                    </div>
-
-                    <div className='-mt-1 flex items-center'>
```
troyciesco, line 215: I tweaked the spacing as requested in https://linear.app/ghost/issue/NY-851 - i think it looks good, but i marked this one as "ref" instead of "closes" in the PR description so I remember to review against the design

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
cmraible: Good catch, I think we definitely do want this

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
cmraible: Had a little play with this, and I think we don't want handleSave in the dependency array.

With the handleSave in the dependency array as it is here, it actually _does_ remove/add the listener on every render; if I remove handleSave from the dependency array, it's added once when the modal is opened, and removed if the modal closes as expected.

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
cmraible: We should be able to get it from the getSettingValues hook where we already get defaultEmailAddress, since we only need to read the siteTitle. (FWIW I think it's also not needed in the newsletters modal).

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -116,18 +105,33 @@ const WelcomeEmailModal = NiceModal.create<WelcomeEmailModalProps>(({emailType =
...
+                    e.preventDefault();
+                    handleSave({fakeWhenUnchanged: true});
+                }
+            };
+            window.addEventListener('keydown', handleCMDS);
+            return () => {
+                window.removeEventListener('keydown', handleCMDS);
+            };
+        }
+    }, [handleSave]);
```
cmraible, line 108:
```suggestion
    }, []);
```

See reply above 😄

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
```diff
@@ -168,58 +172,38 @@ const WelcomeEmailModal = NiceModal.create<WelcomeEmailModalProps>(({emailType =
                                 )}
                             </div>
                             <Button
-                                color="black"
-                                disabled={!hasChanges || isSaving}
-                                label={isSaving ? 'Saving...' : 'Save'}
-                                onClick={handleSave}
+                                color={okProps.color}
+                                disabled={okProps.disabled ?? saveState === 'saving'}
```
cmraible:
```suggestion
                                disabled={okProps.disabled'}
```

The additional `saveState === 'saving'` is redundant; see [useForm](https://github.com/TryGhost/Ghost/blob/770f657ae91e95a9874be0909a08db315c205962/apps/admin-x-framework/src/hooks/useForm.ts#L143)

cmraible (approved):
Nice! Definitely good call on `useForm`, that's my bad for not checking our existing patterns for form handling. Everything's working nicely for me locally, and I've confirmed that it fixes the issues mentioned in the PR description.

Left a couple suggestions that are worth looking at before merge, but they're all pretty minor so I'll go ahead and approve so you're not blocked getting this merged in the AM 🙌

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
troyciesco: ahh right because `handleSave` is changing every time `formState` does because that's a dep in its `useCallback`.

But an empty dependency array (other than showing the exhaustive-deps warning) means that it'll contain an old reference to `handleSave`, and that still has the _original_ `formState`.

Note this is for the `cmd+S` behavior, not clicking the button - i tested with an empty dependency array after reading your comment and that tripped me up for a minute lol

Below: empty dependency array, saving with `cmd+S`. doesn't save the new comment
dep-array

So to recap:
- no dep array: renders every time someone types
- empty dep array: renders once, but doesn't save new content because old `handleSave` ref
- handleSave in dep array: renders every time someone types

I think this can be solved with `useRef`, i'll dust off the ol' React knowledge

troyciesco: yep that fixes it
ref-handleSave

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-modal.tsx
troyciesco: oh good call!

PR #25661, opened 2025-12-08
Added stripe to `yarn dev:forward`

cmraible:
no issue

Currently to run Ghost locally with Stripe with the new `yarn dev:forward` setup, you need to install and run the Stripe CLI locally, and pass the webhook secret to Ghost via environment variable. This change makes it easier to run Ghost with Stripe with the `yarn dev:forward` command.

---
To use:
- Add `COMPOSE_PROFILES=stripe` to `.env` file at root of repo
- Add `STRIPE_SECRET_KEY` to `.env` file at root of repo (or export it in your shell)
- Run `yarn dev:forward`

---
How it works
- The stripe CLI runs in docker. The entrypoint file runs to fetch the WEBHOOK_SECRET required by Ghost to verify Stripe webhooks, and saves it to a shared volume
- When `ghost-dev` starts, the entrypoint reads WEBHOOK_SECRET from the shared volume into its environment before booting Ghost

---

> [!NOTE]
> Enables local Stripe integration without a host-installed CLI by wiring a Dockerized Stripe CLI to Ghost via a shared config volume.
> 
> - Adds `stripe` service (profile `stripe`) running `stripe/stripe-cli` with an entrypoint that fetches `STRIPE_WEBHOOK_SECRET`, writes it to `shared-config/.env.stripe`, and forwards webhooks to `${GHOST_URL}/members/webhooks/stripe/`
> - Updates `ghost-dev` entrypoint to read `STRIPE_WEBHOOK_SECRET` from shared config and export it as `WEBHOOK_SECRET`
> - Mounts `shared-config` into `ghost-dev` and adds optional `depends_on` for `stripe`; healthcheck waits for `.env.stripe`
> - In `compose.yml`, sets `GHOST_URL` for the `stripe` service; `.env.example` hints enabling the `stripe` profile
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 06adeb123b40a31393a7c9329a78db5d257c619b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,7 +4,7 @@
 # Docker Compose profiles to enable
 ## Run `docker compose config --profiles` to see all available profiles
 ## See https://docs.docker.com/compose/how-tos/profiles/ for more information
-COMPOSE_PROFILES=ghost
+# COMPOSE_PROFILES=stripe
 
 # Debug level to pass to Ghost
 # DEBUG=
--- a/compose.dev.yaml
+++ b/compose.dev.yaml
@@ -98,6 +98,9 @@ services:
         condition: service_healthy
       mailpit:
         condition: service_healthy
+      stripe:
+        condition: service_healthy
+        required: false
     healthcheck:
       test: ["CMD", "node", "-e", "fetch('http://localhost:2368',{redirect:'manual'}).then(r=>process.exit(r.status<500?0:1)).catch(()=>process.exit(1))"]
       timeout: 5s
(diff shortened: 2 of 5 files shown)


.env.example
```diff
@@ -4,7 +4,7 @@
 # Docker Compose profiles to enable
 ## Run `docker compose config --profiles` to see all available profiles
 ## See https://docs.docker.com/compose/how-tos/profiles/ for more information
-COMPOSE_PROFILES=ghost
+# COMPOSE_PROFILES=stripe
```
9larsons, line 7: Why remove the default profile? Just because it's not necessary?

9larsons (approved):
My only real concern here is mixing our configuration with .env and config.{x} and adding more confusion... however I think we'll never really avoid that since we need .env for Docker, unless we do some patching to pass through the config items.

Overall this looks good 🙌 . Definitely an improvement and step forward towards parity with our non-Docker dev setup.

.env.example
cmraible: Yep! The default profile is part of the "legacy" docker environment, where by default only mysql + redis run.

PR #25664, opened 2025-12-09
Fixed free welcome emails being sent to new paid members

cmraible:
closes https://linear.app/ghost/issue/NY-857/free-member-welcome-email-is-sent-to-new-paid-members

Currently if free welcome emails are enabled, and a new member signs up for a paid subscription, they will receive the free welcome email in error. This is because new paid members are initially created in the `free` status, until the Stripe subscription is fully processed, at which point they transition to `paid`. 

This fixes the bug by opting to skip sending the welcome email if the member has a `stripeCustomer` attached, which indicates that the member at least _intends_ to become a paid member.

### DIFF
--- a/ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
+++ b/ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
@@ -369,7 +369,8 @@ module.exports = class MemberRepository {
                             email: newMember.get('email'),
                             name: newMember.get('name'),
                             source,
-                            timestamp
+                            timestamp,
+                            memberStatus: 'free'
                         })
                     }, {transacting});
                 }
--- a/ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
+++ b/ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
@@ -355,7 +355,10 @@ module.exports = class MemberRepository {
                     labels
                 }, {...memberAddOptions, transacting});
                 
-                if (isFreeWelcomeEmailActive) {
+                // Only send the free welcome email if:
+                // 1. The free welcome email is active
+                // 2. The member is not signing up for a paid subscription (no stripeCustomer)
+                if (isFreeWelcomeEmailActive && !stripeCustomer) {
                     const timestamp = eventData.created_at || newMember.get('created_at');
 
                     await this._Outbox.add({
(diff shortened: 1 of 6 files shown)


ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -366,7 +369,8 @@ module.exports = class MemberRepository {
                             email: newMember.get('email'),
                             name: newMember.get('name'),
                             source,
-                            timestamp
+                            timestamp,
+                            memberStatus: 'free'
```
cmraible:
```suggestion
                            status: 'free'
```

Thinking we should keep this as just `status`, dropping the `member` prefix, since we don't include it for anything else (except the Id...)

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -380,7 +384,7 @@ module.exports = class MemberRepository {
                 member = await this._Member.transaction(runMemberCreation);
             }
 
-            if (isFreeWelcomeEmailActive) {
+            if (isFreeWelcomeEmailActive && !stripeCustomer) {
```
cmraible:
```suggestion
            const isFreeSignup = !stripeCustomer;
            if (isFreeWelcomeEmailActive && isFreeSignup) {
```

Just for the sake of readability

ghost/core/core/server/services/outbox/handlers/member-created.js
```diff
@@ -4,9 +4,7 @@ const memberWelcomeEmailService = require('../../member-welcome-emails/service')
 const LOG_KEY = `${OUTBOX_LOG_KEY}[MEMBER-WELCOME-EMAIL]`;
 
 async function handle({payload}) {
-    // TODO: derive memberStatus from payload when paid welcome emails are added
-    const memberStatus = 'free';
-    await memberWelcomeEmailService.api.send({member: payload, memberStatus});
+    await memberWelcomeEmailService.api.send({member: payload, memberStatus: payload.memberStatus});
```
cmraible:
```suggestion
    await memberWelcomeEmailService.api.send({member: payload});
```

I think we can skip the payload.memberStatus, since we're passing it in the payload now

cmraible (commented):
Couple minor naming notes to self

ghost/core/core/server/services/outbox/handlers/member-created.js
```diff
@@ -4,9 +4,7 @@ const memberWelcomeEmailService = require('../../member-welcome-emails/service')
...
 }
```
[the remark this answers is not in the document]
cmraible: Not worried about this since this feature is behind a private labs flag.

ghost/core/core/server/services/outbox/handlers/member-created.js
cmraible: Eh, actually being more explicit feels okay here I think - nevermind.

PR #25667, opened 2025-12-09
Handled redirects to CDN in S3Storage

allouis:
ref https://linear.app/ghost/issue/PRO-1550

After moving to CDN based storage with the S3Storage adapter, there may still be references to the old URLs in the wild. Rather than 404 we want to redirect to the CDN on the assumption that files have been migrated.

### DIFF
--- a/ghost/core/core/server/adapters/storage/S3Storage.ts
+++ b/ghost/core/core/server/adapters/storage/S3Storage.ts
@@ -1,5 +1,6 @@
 import fs from 'node:fs';
 import path from 'node:path';
+import type {Request, Response, NextFunction, RequestHandler} from 'express';
 import StorageBase from 'ghost-storage-base';
 import tpl from '@tryghost/tpl';
 import errors from '@tryghost/errors';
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -335,8 +335,19 @@ export default class S3Storage extends StorageBase {
     }
 
     serve() {
-        return function (_req: unknown, _res: unknown, next: (err?: unknown) => void) {
-            next();
+        return (
+            req: {path: string},
```
allouis: let's import express types here and use them

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -334,9 +335,16 @@ export default class S3Storage extends StorageBase {
...
+    serve(): RequestHandler {
+        return (req: Request, res: Response, next: NextFunction) => {
+            const relativePath = req.path.replace(/^\/+/, '');
+
+            if (!relativePath) {
+                return next();
+            }
+
+            const key = this.buildKey(relativePath);
+            res.redirect(301, `${this.cdnUrl}/${key}`);
```
vershwal, line 347: Missing return statement on redirect? Not a bug here but adding return `res.redirect(...)` is a common pattern. Also makes it consistent with the early return `return next();`

PR #25671, opened 2025-12-09
Fixed brittle `scrollToTop` functions in Stats and PostAnalytics

kevinansfield:
ref https://linear.app/ghost/issue/BER-2999/
ref https://github.com/TryGhost/Ghost/pull/25583
ref https://github.com/TryGhost/Ghost/pull/25501

## Problem
- the [recently introduced](https://github.com/TryGhost/Ghost/pull/25583) `scrollToTop` functions use a brittle approach of assuming the first element found with an `overflow-y-scroll` class is the main scrollable content area
- in the React admin shell we were re-working the overall app layout to accommodate alert bars and reduce the number of scrollable elements, this resulted in the removal of the `overflow-y-scroll` class on the main content area breaking these functions

## Fix
- pulled the `getScrollParent` utility out of the posts `virtual-table` components into a re-usable Shade utility (`@tryghost/shade/utils`), and pointed its existing consumers at it
- added container refs to both Stats and PostAnalytics layout components so an element is available from which to find the scrollable parent
- updated the `scrollToTop` functions to use `getScrollParent` with the container ref so no matter what classes get used we're always scrolling the right parent scroll container


cmraible (approved):
Thanks @kevinansfield, looks good to me. Agree this approach is significantly less brittle 👌

PR #25675, opened 2025-12-09
Fixed Koenig editor failing to load in React admin shell

cmraible:
ref https://linear.app/ghost/issue/BER-3111/

The Koenig editor was not loading when using `yarn dev:forward` because `Settings.tsx` was passing an empty no-op function for `fetchKoenigLexical`.

After implementing a proper `fetchKoenigLexical` function that loaded the UMD bundle from `/ghost/assets/koenig-lexical/`, a second issue emerged: React hooks errors ("Invalid hook call", "Cannot read properties of null reading 'useMemo'"). This happened because the UMD bundle includes its own bundled React, which conflicted with Vite's React instance in dev mode.

The fix imports `@tryghost/koenig-lexical` as an ESM module directly, allowing Vite to handle React deduplication. This required adding the package as a dependency to the admin app.

Changes:
- Added `@tryghost/koenig-lexical` dependency to apps/admin
- Created `fetch-koenig-lexical.ts` utility that imports the ESM module
- Updated `settings.tsx` to use the new utility function


cmraible:
Tested this out and it's working locally, but I'm still getting the red "An error occurred loading editor. Please refresh and try again." error when I tried deploying this to staging. Not sure if it's an issue with the code in this PR, or if we're maybe not versioning the assets properly, so deploying a PR to a staging site isn't using the correct files.

jonatansberg:
@cmraible I suspect that the PR deploy-to-staging workflow isn't working. The load balancer is likely intercepting all asset requests and serving the latest build from main. 

We should try to fix that or come up with some other way of doing preview deployments. I see no reason why this fix wouldn't work though and I think it should be safe to merge once the rename conflict has been resolved.

apps/admin/package.json
```diff
@@ -15,6 +15,7 @@
     "@tryghost/activitypub": "*",
     "@tryghost/admin-x-framework": "*",
     "@tryghost/admin-x-settings": "*",
+    "@tryghost/koenig-lexical": "*",
```
cmraible: Should we pin a version here, since koenig-lexical doesn't live in the monorepo?

apps/admin/src/utils/fetch-koenig-lexical.ts
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * In development (Vite), we import the ESM module directly from node_modules.
+ * This ensures React is properly deduped by Vite's bundler, avoiding the
+ * "Invalid hook call" errors that occur when the UMD bundle (which includes
+ * its own bundled React) is loaded alongside Vite's React.
+ */
+export async function fetchKoenigLexical(): Promise<unknown> {
+    // Import the ESM module directly - Vite will handle React deduplication
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const koenig = await import('@tryghost/koenig-lexical');
```
cmraible, line 12: This change would also mean we're no longer importing @tryghost/koenig-lexical from the CDN assets, as we do in the Ember app. I think that's okay, but I'm not 100% sure this doesn't break any of our processes for e.g. releasing changes to the editor.

apps/admin/package.json
cmraible: Should we pin a version here? Thinking probably yes, since koenig-lexical isn't part of the monorepo.

cmraible:
@jonatansberg thanks, sounds good!

@kevinansfield I'd also like to get your eyes on this before merging. Do you see any issue with just installing koenig-lexical in admin, rather than importing it from the CDN?

apps/admin/package.json
jonatansberg: That's probably a good idea. Likely the Renovate bot would do that anyways the moment this PR gets merged.

apps/admin/src/utils/fetch-koenig-lexical.ts
jonatansberg: I don't think that's a problem. It does create a bit of overhead in the migration stage, but lazy loading it like this is how we planned on doing it as we bring it into the monorepo and port the post/page editor.

jonatansberg (approved):

jonatansberg:
@cmraible both Kevin and I (and Rob and Igor) looked at this yesterday in our mob session, and that's what led to my comment. You can go ahead and merge this.

apps/admin/src/utils/fetch-koenig-lexical.ts
kevinansfield: > also mean we're no longer importing @tryghost/koenig-lexical from the CDN assets

If you mean we're no longer loading `@tryghost/koenig-lexical` from cdnjs then we haven't done that for a really long time.

This will create a duplicate of `@tryghost/koenig-lexical` in our built assets that we copy to our own CDN, with the Ember hosted editors loading one version and the React uses loading another. That's not a huge deal though, it's still lazy loaded, the only thing we're losing is the preload to reduce loading state/flicker _but_ that was mainly for the core editor which this won't impact as Ember will still pre-load its version for now.

apps/admin/package.json
kevinansfield: Yeah, pinned version would be good.

kevinansfield:
> Do you see any issue with just installing koenig-lexical in admin, rather than importing it from the CDN?

Explained in an inline comment above too but koenig-lexical has been [part of the Admin build since 2023](https://github.com/TryGhost/Ghost/commit/639be25f1dfcd0a1014bd534742d198d6447d600) so there's no substantial change from this 🙂

kevinansfield (approved):

PR #25676, opened 2025-12-09
Added paid member welcome email sending

troyciesco:
closes https://linear.app/ghost/issue/NY-804

- The settings UI allows publishers to enable and configure a paid welcome email (with a flag enabled), but Ghost doesn't actually send the welcome emails for new paid members.

- This adds the functionality to send the paid welcome emails for newly created paid members, and members upgrading from a free account to a paid plan. 

---

> [!NOTE]
> Implements paid welcome email sending for new or upgraded paid members, leveraging Outbox and existing automated email templates.
> 
> - In `MemberRepository.linkSubscription`, when status changes to `paid`, enqueue `MemberCreatedEvent` to `Outbox` and dispatch processing if `memberWelcomeEmailTestInbox` is set, source is `member`, and the paid template is active
> - Determines source from request context; disallows `import`, `admin`, and `api` sources for sending
> - Validates paid template existence and `active` status before enqueuing
> - Keeps free welcome email behavior intact and ensures no free email is sent for paid signups
> - Integration tests seed paid template and verify Outbox processing no-ops when paid template is inactive/missing
> - Unit tests cover Outbox enqueue on paid transition, and negative cases (config unset, disallowed sources, inactive template)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0aae0f3929a641404315ca9e0c3762c524d46e46. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -972,10 +972,35 @@ module.exports = class MemberRepository {
...
         }
 
+        const context = options?.context || {};
+        const source = this._resolveContextSource(context);
+        const shouldSendPaidWelcomeEmail = config.get('memberWelcomeEmailTestInbox') && WELCOME_EMAIL_SOURCES.includes(source);
+        let isPaidWelcomeEmailActive = false;
+        if (shouldSendPaidWelcomeEmail && this._AutomatedEmail) {
+            const paidWelcomeEmail = await this._AutomatedEmail.findOne({slug: MEMBER_WELCOME_EMAIL_SLUGS.paid}, options);
+            isPaidWelcomeEmailActive = paidWelcomeEmail && paidWelcomeEmail.get('lexical') && paidWelcomeEmail.get('status') === 'active';
+        }
```
cmraible: I'm realizing that this logic feels out of place here. Thinking maybe the `member-welcome-emails` service should export a `shouldSendWelcomeEmail({status: 'paid', source: 'api'})` method or something like that 🤔

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -1278,6 +1306,9 @@ module.exports = class MemberRepository {
                     batchId: options.batch_id
                 });
                 this.dispatchEvent(subscriptionActivatedEvent, options);
+                if (isPaidWelcomeEmailActive) {
+                    await queuePaidWelcomeEmail(newStripeCustomerSubscriptionModel);
+                }
```
cmraible: I think we should do this at the very end of `linkSubscription` where we handle the member status changes. 

If member status has changed, and it is now `paid`, we should send the welcome email. This way we can do it only in one place, and not worry about the intricacies of stripe subscriptions.

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
cmraible: I still think we should do this refactor, and I've gotten started on it here: https://github.com/TryGhost/Ghost/pull/25693

Since this is also how we do it for free welcome emails, the refactor got a bit more complex than I'd hoped. It didn't feel right to include all that with an actual behavior change, so will instead opt to do the refactor after this.

cmraible (approved):

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -1420,6 +1420,30 @@ module.exports = class MemberRepository {
                 to_status: updatedMember.get('status'),
                 ...eventData
             }, options);
+
+            const context = options?.context || {};
+            const source = this._resolveContextSource(context);
+            const shouldSendPaidWelcomeEmail = config.get('memberWelcomeEmailTestInbox') && WELCOME_EMAIL_SOURCES.includes(source);
```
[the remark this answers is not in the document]
troyciesco: i don't think this is true, based on the testing i've done against the admin api, but i made a ticket to double check before GA just in case https://linear.app/ghost/issue/NY-891

PR #25681, opened 2025-12-10
Renamed activitypub app files and directories to kebab-case

mike182uk:
ref https://linear.app/ghost/issue/BER-3044

As part of standardizing naming conventions across the codebase, renamed all `PascalCase` and `camelCase` files / directories in the activitypub app to `kebab-cas`e. This aligns with the RFC for consistent file naming and matches the convention used in `shade` and `admin-x-settings`


ibalosh (approved):
Hey @mike182uk 

it looks good, only thing I would add is I don't see linting rule to enforce it

mike182uk:
Ah good catch @ibalosh I've added that in https://github.com/TryGhost/Ghost/pull/25681/commits/314cb5a164764d1b919b4c803a1b276340c44a0b 👍

ibalosh:
thanks! looks great!

PR #25686, opened 2025-12-10
Added route lazy loading

jonatansberg:
refs https://linear.app/ghost/issue/BER-3126/generated-bundle-is-huge

- Added a simple helper to enable code splitting and lazy loading at the route level. 
- Disabled the redundant preloading of settings in Ember

This reduces the amount of JS we are serving to just render the shell from ~12mb to ~5.4mb pre compression. The minified pre-zip size is in many ways a more relevant metric than the gzip size since the former correlates better with actual JS parse and execution time.

Before:
Screenshot 2025-12-10 at 15 45 05

After: 
Screenshot 2025-12-10 at 15 39 46

There is plenty of room to iterate on this further by pulling out some of our larger dependencies in favor of smaller ones.

### DIFF
--- a/apps/activitypub/src/routes.tsx
+++ b/apps/activitypub/src/routes.tsx
@@ -1,18 +1,6 @@
 import AppError from '@components/layout/error';
-import BlueskySharing from '@views/preferences/components/bluesky-sharing';
-import Explore from '@views/explore';
-import Feed from './views/feed/feed';
-import Inbox from '@views/inbox';
-import Moderation from '@views/preferences/components/moderation';
-import Note from './views/feed/note';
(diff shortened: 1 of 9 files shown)


kevinansfield:
> Disabled the redundant preloading of settings in Ember

@jonatansberg is this part missing from the PR?

jonatansberg:
Yeah, I realized. I pushed that change just now.

kevinansfield (approved):
Tested with non-dev build locally with throttling turned on. The additional loading states weren't particularly distracting and the reduced initial loading from smaller bundle sizes was noticeable 🚀

PR #25687, opened 2025-12-10
🐛 Fixed force upgrade mode not blocking React shell routes

rob-ghost:
ref https://linear.app/ghost/issue/BER-3127

### What is Force Upgrade mode?

Force Upgrade is a Ghost(Pro) billing feature that locks users out of most of the admin when a subscription action is required (e.g., payment failed, plan expired). In this mode, users are redirected to the `/pro` billing page, with only settings and sign-out remaining accessible.

### The Bug

The force upgrade protection was implemented in Ember's routing layer, which intercepts route transitions and redirects to `/pro`. However, as we've migrated routes to React (Analytics, Network/ActivityPub, Tags), these routes bypassed the Ember protection entirely - React was unaware of the force upgrade state, so these pages would render normally when they should have been blocked.

### The Fix

We chose a declarative approach using React Router's `handle` metadata:

1. **React reads forceUpgrade from the API config** - The `hostSettings.forceUpgrade` flag is set by the server and available via the config endpoint
2. **React derives the effective state** - If config says `forceUpgrade: true` but subscription status is `active`, billing was just completed (server hasn't restarted yet), so we clear the lock locally. This mirrors Ember's logic.
3. **A ForceUpgradeGuard wraps all React routes** - This guard checks the derived forceUpgrade state and route metadata
4. **Routes opt-in to being allowed** - Routes set `handle: { allowInForceUpgrade: true }` to bypass the guard (used for settings and the Ember fallback which handles `/pro`)

This approach mirrors how Ember handles it (checking route transitions) but uses React Router's native patterns.

### DIFF
--- a/apps/admin/src/ember-bridge/ember-bridge.tsx
+++ b/apps/admin/src/ember-bridge/ember-bridge.tsx
@@ -1,5 +1,6 @@
 import { useEffect, useState, useSyncExternalStore } from "react";
 import { useQueryClient } from "@tanstack/react-query";
+import { useBrowseConfig } from "@tryghost/admin-x-framework/api/config";
 
 export interface EmberBridge {
     state: StateBridge;
@@ -42,9 +43,10 @@ export interface EmberAuthChangeEvent {
 }
 
 export interface SubscriptionState {
-    subscription: {
+    subscription?: {
         isActiveTrial: boolean;
         trial_end: string | null;
+        status: string;
     };
 }
 
@@ -278,5 +280,34 @@ export function useEmberRouting(): EmberRouting {
         getRouteUrl: bridge.getRouteUrl,
         isRouteActive: bridge.isRouteActive
     };
+}
+
+/**
+ * Hook to get the forceUpgrade state.
+ *
+ * Returns true when the site is in force upgrade mode (requires billing action).
+ *
+ * Force upgrade state is determined by:
+ * 1. Config hostSettings.forceUpgrade (set by server, requires restart to change)
+ * 2. Subscription status (if subscription becomes 'active', forceUpgrade is cleared)
+ */
+export function useForceUpgrade(): boolean {
+    const { data: config } = useBrowseConfig();
+    const subscriptionStatus = useSubscriptionStatus();
+
+    const configForceUpgrade = config?.config?.hostSettings?.forceUpgrade;
+
+    // If config doesn't have forceUpgrade, we're not in force upgrade mode
+    if (!configForceUpgrade) {
+        return false;
+    }
+
+    // If subscription has become active, billing was completed successfully
+    // The server config hasn't restarted yet, but we can clear forceUpgrade locally
+    if (subscriptionStatus?.subscription?.status === 'active') {
+        return false;
+    }
+
+    return true;
 }
 
--- a/apps/admin/src/ember-bridge/force-upgrade-guard.tsx
+++ b/apps/admin/src/ember-bridge/force-upgrade-guard.tsx
@@ -1,0 +1,44 @@
+import { Navigate, Outlet, useMatches } from "@tryghost/admin-x-framework";
+import { useForceUpgrade } from "./ember-bridge";
+
+export interface RouteHandle {
+    allowInForceUpgrade?: boolean;
+}
+
+/**
+ * Guard component that redirects to /pro when site is in force upgrade mode.
+ *
+ * Routes can opt-out of being blocked by setting `handle: { allowInForceUpgrade: true }`
+ * in their route definition. This is used for routes that should remain accessible
+ * during force upgrade (e.g., /pro itself, /settings, /signout).
+ *
+ * When forceUpgrade is active:
+ * - Routes with allowInForceUpgrade: true render normally
+ * - All other routes redirect to /pro (billing page)
+ *
+ * @example
+ * ```tsx
+ * // In routes.tsx
+ * {
+ *     path: "settings/*",
+ *     element: <Settings />,
+ *     handle: { allowInForceUpgrade: true }
+ * }
+ * ```
+ */
+export function ForceUpgradeGuard() {
+    const forceUpgrade = useForceUpgrade();
+    const matches = useMatches();
+
+    // Check if any matched route allows access in force upgrade mode
+    const isAllowed = matches.some((match) => {
+        const handle = match.handle as RouteHandle | undefined;
+        return handle?.allowInForceUpgrade === true;
+    });
+
+    if (forceUpgrade && !isAllowed) {
+        return <Navigate to="/pro" replace />;
+    }
+
+    return <Outlet />;
+}
--- a/apps/admin/src/routes.tsx
+++ b/apps/admin/src/routes.tsx
@@ -13,53 +13,66 @@ import GlobalDataProvider from "@tryghost/stats/src/providers/global-data-provid
 import { routes as statsRoutes } from "@tryghost/stats/src/routes";
 
 // Ember
-import { EmberFallback } from "./ember-bridge";
+import { EmberFallback, ForceUpgradeGuard } from "./ember-bridge";
+import type { RouteHandle } from "./ember-bridge";
 
 export const routes: RouteObject[] = [
     {
-        // Override the tag detail route from the posts app to ensure we
-        // correctly delegate to Ember since we can't remove the blank screen in
-        // the posts app. The blank screen needs to be there to prevent the
-        // router error fallback from triggering when navigating from the tag
-        // list to a tag detail page.
-        path: "/tags/:tagSlug",
-        Component: EmberFallback,
-    },
-    {
-        element: (
-            <PostsAppContextProvider value={{ fromAnalytics: true }}>
-                <Outlet />
-            </PostsAppContextProvider>
-        ),
-        children: postRoutes[0].children!.filter((route) => route.path !== "*"),
-    },
-    {
-        element: (
-            <GlobalDataProvider>
-                <Outlet />
-            </GlobalDataProvider>
-        ),
-        children: statsRoutes,
-    },
-    {
-        path: `network`,
-        loader: () => redirect("/activitypub"),
-    },
-    {
-        path: "",
-        element: (
-            <FeatureFlagsProvider>
-                <Outlet />
-            </FeatureFlagsProvider>
-        ),
-        children: activityPubRoutes,
-    },
-    {
-        path: `settings/*`,
-        lazy: lazyComponent(() => import("./settings/settings")),
-    },
-    {
-        path: "*",
-        Component: EmberFallback,
+        // ForceUpgradeGuard wraps all routes to redirect to /pro when in force upgrade mode.
+        // Routes with handle.allowInForceUpgrade: true bypass this protection.
+        element: <ForceUpgradeGuard />,
+        children: [
+            {
+                // Override the tag detail route from the posts app to ensure we
+                // correctly delegate to Ember since we can't remove the blank screen in
+                // the posts app. The blank screen needs to be there to prevent the
+                // router error fallback from triggering when navigating from the tag
+                // list to a tag detail page.
+                path: "/tags/:tagSlug",
+                Component: EmberFallback,
+                handle: { allowInForceUpgrade: true } satisfies RouteHandle,
+            },
+            {
+                element: (
+                    <PostsAppContextProvider value={{ fromAnalytics: true }}>
+                        <Outlet />
+                    </PostsAppContextProvider>
+                ),
+                children: postRoutes[0].children!.filter((route) => route.path !== "*"),
+            },
+            {
+                element: (
+                    <GlobalDataProvider>
+                        <Outlet />
+                    </GlobalDataProvider>
+                ),
+                children: statsRoutes,
+            },
+            {
+                path: `network`,
+                loader: () => redirect("/activitypub"),
+            },
+            {
+                path: "",
+                element: (
+                    <FeatureFlagsProvider>
+                        <Outlet />
+                    </FeatureFlagsProvider>
+                ),
+                children: activityPubRoutes,
+            },
+            {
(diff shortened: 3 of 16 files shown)


apps/admin/src/ember-bridge/ember-bridge.tsx
```diff
@@ -43,8 +44,9 @@ export interface EmberAuthChangeEvent {
 
 export interface SubscriptionState {
     subscription: {
-        isActiveTrial: boolean;
-        trial_end: string | null;
+        isActiveTrial?: boolean;
+        trial_end?: string | null;
+        status?: string;
```
jonatansberg: Nit: Are these all individually optional, or is it actually the full object that should be optional?

apps/admin/src/ember-bridge/ember-bridge.tsx
```diff
@@ -280,3 +282,36 @@ export function useEmberRouting(): EmberRouting {
...
+ * checking if config says forceUpgrade but subscription is now active.
+ */
+export function useForceUpgrade(): boolean {
+    const { data: config } = useBrowseConfig();
+    const subscriptionStatus = useSubscriptionStatus();
+
+    const configForceUpgrade = config?.config?.hostSettings?.forceUpgrade;
+
+    // If config doesn't have forceUpgrade, we're not in force upgrade mode
+    if (!configForceUpgrade) {
```
jonatansberg, line 301: Thought: Maybe redundant, but would it make sense to also check if billing is enabled?

apps/admin/src/routes.tsx
```diff
@@ -16,53 +16,66 @@ import { routes as statsRoutes } from "@tryghost/stats/src/routes";
...
+                    <FeatureFlagsProvider>
+                        <Outlet />
+                    </FeatureFlagsProvider>
+                ),
+                children: activityPubRoutes,
+            },
+            {
+                path: `settings/*`,
+                element: <Settings />,
+                handle: { allowInForceUpgrade: true } as RouteHandle,
```
jonatansberg: Can't we fix the `RouteObject` rather than assert here?

e2e/tests/admin/sidebar/force-upgrade.test.ts
```diff
@@ -0,0 +1,152 @@
+import {NAV_ITEMS, SidebarPage} from '@/helpers/pages';
```
jonatansberg: Nit: shouldn't this be a billing test rather than a sidebar test?

e2e/tests/admin/sidebar/force-upgrade.test.ts
```diff
@@ -0,0 +1,152 @@
...
+     * We check for the iframe element itself since the content may not load in test environment.
+     */
+    async function waitForBillingIframe(page: import('@playwright/test').Page) {
+        // Target the billing iframe specifically (there may be other iframes like Explore)
+        const billingIframe = page.locator('iframe[title="Billing"]');
+        await billingIframe.waitFor({state: 'visible', timeout: 10000});
+        return billingIframe;
+    }
+
+    test.describe('sidebar navigation shows billing iframe', () => {
```
jonatansberg: Is this meant to test the redirect? It's not the sidebar that shows the iframe, no?

ghost/admin/app/routes/application.js
```diff
@@ -233,6 +234,12 @@ export default Route.extend(ShortcutsRoute, {
             this.billing.openBillingWindow(this.router.currentURL, '/pro');
         }
 
+        // Notify React of the initial subscription state
+        // React uses this to derive forceUpgrade state (config.forceUpgrade && subscription.status !== 'active')
+        this.stateBridge.triggerSubscriptionChange({
+            subscription: this.billing.subscription || {}
+        });
```
jonatansberg, line 241: Non blocking question: Is this needed? I assumed that iframe would load way after the bridge has been initialized, but maybe that's a naive assumption.

apps/admin/src/ember-bridge/force-upgrade-guard.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * ```
+ */
+export function ForceUpgradeGuard() {
+    const forceUpgrade = useForceUpgrade();
+    const matches = useMatches();
+
+    // Check if any matched route allows access in force upgrade mode
+    const isAllowed = matches.some((match) => {
+        const handle = match.handle as RouteHandle | undefined;
+        return handle?.allowInForceUpgrade === true;
```
jonatansberg, line 36: Nice use of handles! 👍

jonatansberg (commented):
Overall this looks good. I added a few comments on some minor things that we might want to look at before merging.

e2e/tests/admin/sidebar/force-upgrade.test.ts
```diff
@@ -0,0 +1,152 @@
...
+            });
+        });
+    });
+
+    /**
+     * Helper to wait for the billing iframe to be visible.
+     * In forceUpgrade mode, the billing iframe should appear in the main content area.
+     * We check for the iframe element itself since the content may not load in test environment.
+     */
+    async function waitForBillingIframe(page: import('@playwright/test').Page) {
```
ibalosh: any reason for not moving the helper into the page object? 
Also the import for type works, but looks awkward.

e2e/tests/admin/sidebar/force-upgrade.test.ts
```diff
@@ -0,0 +1,152 @@
...
+            test(`${name} link - billing iframe blocks access`, async ({page}) => {
+                const sidebarPage = new SidebarPage(page);
+                await sidebarPage.goto();
+
+                // Wait for billing iframe to appear (force upgrade kicks in)
+                const billingIframe = await waitForBillingIframe(page);
+                await expect(billingIframe).toBeVisible();
+
+                // Try to navigate via sidebar link
+                const navLink = sidebarPage.getNavLink(name);
```
ibalosh: I think we could remove this comment. It could be removed with self explanatory function name in sidebarPage, like `sidebarPage.navigateViaLink(linkName)` or something similar

e2e/tests/admin/sidebar/force-upgrade.test.ts
```diff
@@ -0,0 +1,152 @@
...
+            await sidebarPage.getNavLink('Settings').click();
+
+            // Settings routes should be accessible - URL should change to settings
+            await expect(page).toHaveURL(/#\/settings/);
+        });
+
+        test('direct navigation to /settings is allowed', async ({page}) => {
+            const sidebarPage = new SidebarPage(page);
+            await sidebarPage.goto('/ghost/#/settings');
+            // Settings should be allowed - should NOT show billing iframe
```
ibalosh: Any reason we need this comment here? If billing iframe should be hiddent, should not that be an assertion?

this seems like instead of comment could be something like:

``` typescript
 expect(settingsPage.someElement).toBeVisible(); 
 expect(billingIframe).toBeHidden();
```

e2e/tests/admin/sidebar/force-upgrade.test.ts
```diff
@@ -0,0 +1,152 @@
...
+            await expect(page).toHaveURL(/#\/settings/);
+        });
+
+        test('Sign out is accessible', async ({page}) => {
+            const sidebarPage = new SidebarPage(page);
+            await sidebarPage.goto();
+
+            // Wait for billing iframe
```
ibalosh: not needed comment, it just duplicates the functionName which is self explanatory

e2e/tests/admin/sidebar/force-upgrade.test.ts
```diff
@@ -0,0 +1,152 @@
...
+            await waitForBillingIframe(page);
+
+            // Open user dropdown and click sign out
+            await sidebarPage.userDropdownTrigger.click();
+            await sidebarPage.signOutLink.click();
+
+            // Should navigate to signin page after logout
```
ibalosh: not sure we need this comment either

ibalosh (changes requested):

apps/admin/src/ember-bridge/ember-bridge.tsx
rob-ghost: Good question, this comes from the fact that Ember is sending `subscription || {}` through, so we might get an empty object.

I think we can either:

1. Leave it as-is with the individual fields marked as optional
2. Make the `subscription` field a `Partial` to allow for optional fields
3. Make the `subscription` field itself optional, and change the Ember code to return `|| null` and handle that downstream

Do you have any thoughts and what would be your preference?

apps/admin/src/ember-bridge/ember-bridge.tsx
rob-ghost: `forceUpgrade` is only set by Ghost(Pro) when billing is enabled so it wouldn't be required, but we can add it if you'd prefer to be explicit though!

apps/admin/src/routes.tsx
rob-ghost: I tried to avoid this but `RouteObject` is re-exported directly from `react-router`, which types handle as `unknown` so we'd need to create our own custom extended type for this property and update the types here but I opted to leave them as-is and conform.

Would you prefer to have the custom extended type?

e2e/tests/admin/sidebar/force-upgrade.test.ts
rob-ghost: Yep you're right, the work stared in the sidebar before moving it to the router guard. I'll get this moved over to `/billing`

e2e/tests/admin/sidebar/force-upgrade.test.ts
rob-ghost: Yep the redirect came after these tests were added initially. I'll get this re-named with the move.

ghost/admin/app/routes/application.js
rob-ghost: I do think its needed, but it might be left-over from before we added the router redirect.

I think before we'd get `triggerSubscriptionChange` after the billing iframe renders, so there would be a time where the `useForceUpgrade` hook wouldn't return accurate data.

With triggering this in the application, this seeds the state with the right data immediately and makes sure the router and `useForceUpgrade` resolves with the right data first.

apps/admin/src/ember-bridge/force-upgrade-guard.tsx
rob-ghost: Thanks! Glad you liked the approach

e2e/tests/admin/sidebar/force-upgrade.test.ts
rob-ghost: Agreed on the import.

I guess I just wasn't imaging it as a Page, but you're right, I'll make a `BillingPage` or similar to encapsulate this and get rid of the awkward import.

e2e/tests/admin/sidebar/force-upgrade.test.ts
rob-ghost: I'll address all the un-necessary comments

e2e/tests/admin/sidebar/force-upgrade.test.ts
rob-ghost: Made this an assertion

apps/admin/src/ember-bridge/ember-bridge.tsx
rob-ghost: I ended up going with `subscription?`

apps/admin/src/routes.tsx
```diff
@@ -13,53 +13,66 @@ import GlobalDataProvider from "@tryghost/stats/src/providers/global-data-provid
...
+                Component: EmberFallback,
+                handle: { allowInForceUpgrade: true } as RouteHandle,
+            },
+            {
+                element: (
+                    <PostsAppContextProvider value={{ fromAnalytics: true }}>
+                        <Outlet />
+                    </PostsAppContextProvider>
+                ),
+                children: postRoutes[0].children!.filter((route) => route.path !== "*"),
```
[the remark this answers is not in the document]
rob-ghost: Out of scope

apps/admin/src/routes.tsx
jonatansberg: I was thinking you would create a type for this at the top of the file where you override the handle, e.g. something like `type RouteWithHandle = Omit<RouteObject, 'handle'> & { handle?: RouteHandle }`... 

But maybe that gets gnarly with the self-referential nature of route objects and children and whatnot 🤔

apps/admin/src/ember-bridge/ember-bridge.tsx
```diff
@@ -280,3 +282,36 @@ export function useEmberRouting(): EmberRouting {
...
+ *
+ * Returns true when the site is in force upgrade mode (requires billing action).
+ *
+ * Force upgrade state is determined by:
+ * 1. Config hostSettings.forceUpgrade (set by server, requires restart to change)
+ * 2. Subscription status (if subscription becomes 'active', forceUpgrade is cleared)
+ *
+ * When a user completes billing, the subscription status becomes 'active' but the
+ * server config won't update until restart. React derives the effective state by
+ * checking if config says forceUpgrade but subscription is now active.
```
jonatansberg: Nit: this comment seems a bit redundant with the breakdown above and the inline comments below

apps/admin/src/routes.tsx
jonatansberg: If the above is a dead end, then we should at least be able to swap the `as` for `satisfies`, no?

rob-ghost: Can't do the `type RouteWithHandle` because it won't typecheck with `RouterProvider` because `RouteWithHandle[] != RouteObject[]`.

`satisfies` seems to work so I'll change it to that.

ibalosh (approved):

jonatansberg (approved):
Looks good! CI seems unhappy for some reason, though.

PR #25688, opened 2025-12-10
:art: Added net change to tooltip on change in paid subscriptions chart

prschulz:
ref https://linear.app/ghost/issue/NY-863

- Allows publishers to easily see net change in subscriptions for a given time period


peterzimon (approved):
Tested it, looks good to me

cmraible (approved):
Looks good to me! Tested locally and the net change is showing up as expected 👌

PR #25699, opened 2025-12-11
Basic comment moderation

jonatansberg:
Reworked Zimo's earlier POC so that we can integrate it into the new React admin.


apps/posts/src/views/comments/comments.tsx
```diff
@@ -0,0 +1,138 @@
...
+                        </Button>
+                    </div>
+                ) : !data?.comments.length ? (
+                    <div className="flex h-full items-center justify-center">
+                        <EmptyIndicator
+                            title="No comments yet"
+                        >
+                            <LucideIcon.MessageSquare />
+                        </EmptyIndicator>
+                    </div>
```
kevinansfield, line 122: Minor future enhancement: It would be nice to change the wording to "No comments match selected filter" or similar when a filter is in use

kevinansfield (approved):

PR #25702, opened 2025-12-11
Updated test welcome emails to send in-progress content

troyciesco:
no ref

Original PR first saves the automated email before sending the test email, but that creates a scenario where you'll _also_ have the changes live, which you might not want.

This PR, pointed at that one for comparison, switches the logic around so whatever is currently in the form for subject and lexical is what gets sent. This also happens to make validation and error/loading/success states a lot less jumpy.


troyciesco:
@coderabbitai review

PR #25703, opened 2025-12-11
Fixed <GhFullscreenModal> display in Admin-forward

kevinansfield:
closes https://linear.app/ghost/issue/BER-3114/

Some modals in Ember (post history and member-related modals like import) that use an older style of modal implementation were rendering inside of the main content area rather than fullscreen.

- added wormhole element for the liquid-wormhole based modals to the top-level React template
- added override of `liquid-wormhole` service with a replaced `addDefaultDestination()` method that appends to our new wormhole element when in admin-forward
  - the original problem arose because the default `liquid-wormhole` behaviour is to render to the configured `rootElement` rather than the document, we had to set `rootElement` to a nested element for Ember to render alongside the React sidebar which then doesn't work for modals that expect to cover the entire screen

### DIFF
--- a/apps/admin/index.html
+++ b/apps/admin/index.html
@@ -29,6 +29,7 @@
     </div>
     <div id="ember-basic-dropdown-wormhole"></div>
     <div id="ember-modal-wormhole"></div>
+    <div id="ember-liquid-wormhole"></div>
     <script type="module" src="/src/main.tsx"></script>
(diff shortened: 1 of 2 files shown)


jonatansberg (approved):
Nice! I verified that this works locally both in the React shell and in the old Ember admin.

PR #25709, opened 2025-12-12
Added materialized view for mv_session_data pipe

cmraible:

The `mv_session_data` pipe aggregates session data that is then used in almost all our downstream endpoints. Originally we created this as a materialized view (hence the `mv` prefix), but we encountered a bug that caused us to double count pageviews within a session. We've now reached a point where we are encountering performance issues with the current implementation, with requests to Tinybird endpoints timing out after 10 seconds.

This reimplements the materialized view for `mv_session_data`, using the `AggregatingMergeTree` table engine and the appropriate -State and -Merge [aggregate function combinators.](https://clickhouse.com/docs/sql-reference/aggregate-functions/combinators). These combinators were the missing piece in our original implementation which caused the double counting.

### DIFF
--- a/e2e/tests/admin/analytics/overview.test.ts
+++ b/e2e/tests/admin/analytics/overview.test.ts
@@ -20,6 +20,35 @@ test.describe('Ghost Admin - Analytics Overview', () => {
         expect(await analyticsOverviewPage.uniqueVisitors.count()).toBe(1);
     });
 
+    test('records multiple pageviews in single session correctly', async ({page, browser, baseURL}) => {
+        const analyticsWebTrafficPage = new AnalyticsWebTrafficPage(page);
+
+        const context = await browser.newContext({baseURL});
+        const publicBrowserPage = await context.newPage();
+        const homePage = new HomePage(publicBrowserPage);
+
+        await homePage.goto();
+        await page.waitForTimeout(1000);
+        await analyticsWebTrafficPage.goto();
+        await expect(analyticsWebTrafficPage.totalViewsTab).toContainText('1');
+        await expect(analyticsWebTrafficPage.totalUniqueVisitorsTab).toContainText('1');
+
+        await homePage.goto();
+        await page.waitForTimeout(1000);
+        await analyticsWebTrafficPage.refreshData();
+        await expect(analyticsWebTrafficPage.totalViewsTab).toContainText('2');
+        await expect(analyticsWebTrafficPage.totalUniqueVisitorsTab).toContainText('1');
+
+        await homePage.goto();
+        await page.waitForTimeout(1000);
+        await analyticsWebTrafficPage.refreshData();
+        await expect(analyticsWebTrafficPage.totalViewsTab).toContainText('3');
+        await expect(analyticsWebTrafficPage.totalUniqueVisitorsTab).toContainText('1');
+
+        await publicBrowserPage.close();
+        await context.close();
+    });
+
     test('latest post', async ({page}) => {
         const analyticsOverviewPage = new AnalyticsOverviewPage(page);
         await analyticsOverviewPage.goto();
--- a/e2e/tests/admin/analytics/overview.test.ts
+++ b/e2e/tests/admin/analytics/overview.test.ts
@@ -23,30 +23,35 @@ test.describe('Ghost Admin - Analytics Overview', () => {
     test('records multiple pageviews in single session correctly', async ({page, browser, baseURL}) => {
         const analyticsWebTrafficPage = new AnalyticsWebTrafficPage(page);
 
+        const waitForViewCount = async (expectedCount: string) => {
+            await expect.poll(async () => {
+                await analyticsWebTrafficPage.refreshData();
+                return await analyticsWebTrafficPage.totalViewsTab.textContent();
+            }, {timeout: 10000}).toContain(expectedCount);
+        };
+
         const context = await browser.newContext({baseURL});
         const publicBrowserPage = await context.newPage();
-        const homePage = new HomePage(publicBrowserPage);
 
-        await homePage.goto();
-        await page.waitForTimeout(1000);
-        await analyticsWebTrafficPage.goto();
-        await expect(analyticsWebTrafficPage.totalViewsTab).toContainText('1');
-        await expect(analyticsWebTrafficPage.totalUniqueVisitorsTab).toContainText('1');
+        try {
+            const homePage = new HomePage(publicBrowserPage);
 
-        await homePage.goto();
-        await page.waitForTimeout(1000);
-        await analyticsWebTrafficPage.refreshData();
-        await expect(analyticsWebTrafficPage.totalViewsTab).toContainText('2');
-        await expect(analyticsWebTrafficPage.totalUniqueVisitorsTab).toContainText('1');
+            await homePage.goto();
+            await analyticsWebTrafficPage.goto();
+            await waitForViewCount('1');
+            await expect(analyticsWebTrafficPage.totalUniqueVisitorsTab).toContainText('1');
 
-        await homePage.goto();
-        await page.waitForTimeout(1000);
-        await analyticsWebTrafficPage.refreshData();
-        await expect(analyticsWebTrafficPage.totalViewsTab).toContainText('3');
-        await expect(analyticsWebTrafficPage.totalUniqueVisitorsTab).toContainText('1');
+            await homePage.goto();
+            await waitForViewCount('2');
+            await expect(analyticsWebTrafficPage.totalUniqueVisitorsTab).toContainText('1');
 
-        await publicBrowserPage.close();
-        await context.close();
+            await homePage.goto();
+            await waitForViewCount('3');
+            await expect(analyticsWebTrafficPage.totalUniqueVisitorsTab).toContainText('1');
+        } finally {
+            await publicBrowserPage.close();
+            await context.close();
+        }
     });
 
     test('latest post', async ({page}) => {
--- a/ghost/core/core/server/data/tinybird/datasources/_mv_session_data.datasource
+++ b/ghost/core/core/server/data/tinybird/datasources/_mv_session_data.datasource
@@ -1,0 +1,16 @@
+SCHEMA >
+    `site_uuid` LowCardinality(String),
+    `session_id` String,
+    `pageviews` AggregateFunction(count, UInt64),
+    `first_pageview` AggregateFunction(min, DateTime),
+    `last_pageview` AggregateFunction(max, DateTime),
+    `source` AggregateFunction(argMin, String, DateTime),
+    `device` AggregateFunction(argMin, String, DateTime),
+    `utm_source` AggregateFunction(argMin, String, DateTime),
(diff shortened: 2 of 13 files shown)


e2e/tests/admin/analytics/overview.test.ts
```diff
@@ -20,6 +20,35 @@ test.describe('Ghost Admin - Analytics Overview', () => {
...
 
+    test('records multiple pageviews in single session correctly', async ({page, browser, baseURL}) => {
+        const analyticsWebTrafficPage = new AnalyticsWebTrafficPage(page);
+
+        const context = await browser.newContext({baseURL});
+        const publicBrowserPage = await context.newPage();
+        const homePage = new HomePage(publicBrowserPage);
+
+        await homePage.goto();
+        await page.waitForTimeout(1000);
```
cmraible: These constant timeouts are definitely less than ideal. I'd much prefer a more robust way to wait until the page hit has been fully processed by tinybird, but I'm not aware of a better way to do this at the moment. 🤔

ghost/core/core/server/data/tinybird/datasources/_mv_session_data.datasource
```diff
@@ -0,0 +1,16 @@
...
+    `source` AggregateFunction(argMin, String, DateTime),
+    `device` AggregateFunction(argMin, String, DateTime),
+    `utm_source` AggregateFunction(argMin, String, DateTime),
+    `utm_medium` AggregateFunction(argMin, String, DateTime),
+    `utm_campaign` AggregateFunction(argMin, String, DateTime),
+    `utm_term` AggregateFunction(argMin, String, DateTime),
+    `utm_content` AggregateFunction(argMin, String, DateTime)
+
+ENGINE "AggregatingMergeTree"
+ENGINE_SORTING_KEY "site_uuid, session_id"
```
cmraible, line 16: This is the data source that the mv_session_data pipe will materialize into.

ghost/core/core/server/data/tinybird/pipes/mv_session_data.pipe
```diff
@@ -1,49 +1,22 @@
...
+        argMinState(utm_term, timestamp) as utm_term,
+        argMinState(utm_content, timestamp) as utm_content
     FROM _mv_hits
-    WHERE site_uuid = {{ String(site_uuid, 'mock_site_uuid', description="Tenant ID", required=True) }}
-        {% if defined(date_from) %}
-            AND timestamp >= toDateTime({{ Date(date_from) }}, {{ String(timezone, 'Etc/UTC', description="Site timezone", required=False) }})
-        {% end %}
-        {% if defined(date_to) %}
-            AND timestamp < toDateTime({{ Date(date_to) }}, {{ String(timezone, 'Etc/UTC', description="Site timezone", required=False) }}) + interval 1 day
-        {% end %}
```
cmraible, line 26: Since this will be running at ingest time to populate the materialized view, we don't have the site_uuid or date filters available to filter on. This is correct, because we want the materialized view itself to include all sessions, regardless of site_uuid or time filters.

ghost/core/core/server/data/tinybird/datasources/_mv_session_data.datasource
```diff
@@ -0,0 +1,16 @@
...
+    `last_pageview` AggregateFunction(max, DateTime),
+    `source` AggregateFunction(argMin, String, DateTime),
+    `device` AggregateFunction(argMin, String, DateTime),
+    `utm_source` AggregateFunction(argMin, String, DateTime),
+    `utm_medium` AggregateFunction(argMin, String, DateTime),
+    `utm_campaign` AggregateFunction(argMin, String, DateTime),
+    `utm_term` AggregateFunction(argMin, String, DateTime),
+    `utm_content` AggregateFunction(argMin, String, DateTime)
+
+ENGINE "AggregatingMergeTree"
```
cmraible, line 15: "AggregatingMergeTree" is a Clickhouse "Table Engine" that is specifically built for, well, aggregations. But it behaves a little differently than a default "MergeTree", so we have to use special -State functions in the `mv_session_data.pipe` pipe that feeds this data source, and then use special -Merge functions anywhere we query this datasource.

ghost/core/core/server/data/tinybird/pipes/filtered_sessions.pipe
```diff
@@ -37,16 +37,31 @@ SQL >
...
+            argMinMerge(utm_content) as utm_content
+        FROM _mv_session_data
+        WHERE site_uuid = {{ String(site_uuid, 'mock_site_uuid', description="Tenant ID", required=True) }}
+        GROUP BY site_uuid, session_id
+    ) sd
     inner join sessions_filtered_by_hit_attributes sfha
       on sfha.session_id = sd.session_id
     where
-        site_uuid = {{ String(site_uuid, 'mock_site_uuid', description="Tenant ID", required=True) }}
+        1 = 1
```
troyciesco, line 64: Is this because all of the other where conditions might not happen so we need to have at least something here?

ghost/core/core/server/data/tinybird/pipes/mv_session_data.pipe
```diff
@@ -1,49 +1,22 @@
...
+        countState() as pageviews,
+        minState(timestamp) as first_pageview,
+        maxState(timestamp) as last_pageview,
+        argMinState(source, timestamp) as source,
+        argMinState(device, timestamp) as device,
+        argMinState(utm_source, timestamp) as utm_source,
+        argMinState(utm_medium, timestamp) as utm_medium,
+        argMinState(utm_campaign, timestamp) as utm_campaign,
+        argMinState(utm_term, timestamp) as utm_term,
+        argMinState(utm_content, timestamp) as utm_content
```
cmraible, line 17: Since this pipe is now feeding materialized view, and more specifically an AggregatingMergeTree, we need to use these -State versions of all our aggregation functions. This changes how the data is stored in Clickhouse, and then requires that we use the corresponding -Merge functions to "collapse" the state into the aggregations current value whenever we query the materialized view.

ghost/core/core/server/data/tinybird/endpoints/api_kpis.pipe
```diff
@@ -56,6 +56,22 @@ SQL >
...
+    Read session data from AggregatingMergeTree MV using -Merge combinators
+
+SQL >
+    %
+    SELECT
+        site_uuid,
+        session_id,
+        countMerge(pageviews) as pageviews,
+        minMerge(first_pageview) as first_pageview,
+        maxMerge(last_pageview) as last_pageview
```
cmraible, line 70: These -Merge functions effectively calculate the actual value (count, min, max as the case may be) from the state that is stored in the `AggregatingMergeTree`. This is the piece that we were missing in our original implementation, which led to double counting pageviews because we weren't using the -State or -Merge functions to keep track of the state appropriately and then collapse it into the correct value at query time.

ghost/core/core/server/data/tinybird/pipes/filtered_sessions.pipe
cmraible: good question, I'm not sure if we need this 1 = 1 or not 🤔

cmraible: Oh, it's actually because all the subsequent `where` clauses start with "and", so we need to have some kind of condition before them, otherwise this could end up as `where and ...` which would throw a syntax error

ghost/core/core/server/data/tinybird/endpoints/api_top_devices.pipe
```diff
@@ -1,17 +1,26 @@
...
         count() as visits
-    from mv_session_data sd
+    from session_data sd
       inner join filtered_sessions fs
          on fs.session_id = sd.session_id
-    where
-        site_uuid = {{ String(site_uuid, 'mock_site_uuid', description="Tenant ID", required=True) }}
     group by device
     order by visits desc
     limit {{ Int32(skip, 0) }},{{ Int32(limit, 50) }}
```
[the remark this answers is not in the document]
cmraible: This is a valid point, but:
- adding the additional field to the join could have unintended performance side effects
- the filtered_sessions and session_data nodes are already filtered by the same site_uuid, so in practice there is practically no risk of a collision
- the current implementation before this change does include this defense-in-depth

so while this is not a bad idea, it feels out of place in this change.

e2e/tests/admin/analytics/overview.test.ts
```diff
@@ -20,6 +20,40 @@ test.describe('Ghost Admin - Analytics Overview', () => {
...
 
+    test('records multiple pageviews in single session correctly', async ({page, browser, baseURL}) => {
+        const analyticsWebTrafficPage = new AnalyticsWebTrafficPage(page);
+
+        const waitForViewCount = async (expectedCount: string) => {
+            await expect.poll(async () => {
+                await analyticsWebTrafficPage.refreshData();
+                return await analyticsWebTrafficPage.totalViewsTab.textContent();
+            }, {timeout: 10000}).toContain(expectedCount);
+        };
```
cmraible, line 31: This is actually a neat little pattern that Claude suggested. It should both avoid a constant `waitForTimeout()` while (hopefully) eliminating the flakiness from any tests that depend on page_hit ingestion. Will plan to update our other page_hit related tests to use this soon.

cmraible:
- I've tested this pretty extensively locally, and it's all working as expected. 
- The existing Tinybird fixture tests didn't have to change at all for this, which gives me a lot of confidence that this is correct
- Considered deploying this now (Thursday night) but ultimately decided to hold off, at least until morning when I have my wits about me

ghost/core/core/server/data/tinybird/datasources/_mv_session_data.datasource
troyciesco: Do we also need `ENGINE_PARTITION_KEY`?

From reading about it in the docs ([here](https://www.tinybird.co/docs/sql-reference/engines/mergetree#engine-settings)) i'm guessing no, and it is optional anyway, I just noticed it on tb's AggregatingMergeTree [example](https://www.tinybird.co/docs/sql-reference/engines/aggregatingmergetree). Looking at the schema I don't know if there's anything we could base it on anyway 🤔

e2e/tests/admin/analytics/overview.test.ts
ibalosh: oh, nice one!

e2e/tests/admin/analytics/overview.test.ts
```diff
@@ -20,6 +20,40 @@ test.describe('Ghost Admin - Analytics Overview', () => {
...
+
+        const context = await browser.newContext({baseURL});
+        const publicBrowserPage = await context.newPage();
+
+        try {
+            const homePage = new HomePage(publicBrowserPage);
```
ibalosh, line 37: this looks like `withIsolatedPage` that was added before, so we don't need try catch etc ?

ibalosh (approved):
one small note on test side other than that looks good to me

9larsons (approved):

PR #25714, opened 2025-12-15
✨ Added comment permalinks to share direct links to specific comments

rob-ghost:
Adds the ability to permalink to a specific comment anchored on the comment's timestamp.

When someone visits a comment permalink, the page bypasses lazy loading to ensure the comments section loads immediately, then scrolls the referenced comment into view and briefly highlights it with a yellow flash so it's easy to spot. If the comment isn't on the first page, then we check if the comment exists at all, and if it does we keep fetching pages until we find it.

The URL format follows the pattern `{post-url}#ghost-comments-{comment-id}`.

ref https://linear.app/ghost/issue/FEA-486

### DIFF
--- a/apps/comments-ui/src/app.tsx
+++ b/apps/comments-ui/src/app.tsx
@@ -7,18 +7,27 @@ import React, {useCallback, useEffect, useMemo, useState} from 'react';
 import i18nLib from '@tryghost/i18n';
 import setupGhostApi from './utils/api';
 import {ActionHandler, SyncActionHandler, isSyncAction} from './actions';
-import {AppContext, DispatchActionType, EditableAppContext} from './app-context';
+import {AppContext, Comment, DispatchActionType, EditableAppContext} from './app-context';
 import {CommentsFrame} from './components/frame';
 import {setupAdminAPI} from './utils/admin-api';
 import {useOptions} from './utils/options';
 
 type AppProps = {
     scriptTag: HTMLElement;
+    initialCommentId: string | null;
+    pageUrl: string;
 };
 
 const ALLOWED_MODERATORS = ['Owner', 'Administrator', 'Super Editor'];
 
-const App: React.FC<AppProps> = ({scriptTag}) => {
+/**
+ * Check if a comment ID exists in the comments array (either as a top-level comment or reply)
+ */
+function isCommentLoaded(comments: Comment[], targetId: string): boolean {
+    return comments.some(c => c.id === targetId || c.replies?.some(r => r.id === targetId));
+}
+
+const App: React.FC<AppProps> = ({scriptTag, initialCommentId, pageUrl}) => {
     const options = useOptions(scriptTag);
     const [state, setFullState] = useState<EditableAppContext>({
         initStatus: 'running',
@@ -33,7 +42,9 @@ const App: React.FC<AppProps> = ({scriptTag}) => {
         order: 'count__likes desc, created_at desc',
         adminApi: null,
         commentsIsLoading: false,
-        commentIdToHighlight: null
+        commentIdToHighlight: null,
+        commentIdToScrollTo: initialCommentId,
+        pageUrl
     });
 
     const iframeRef = React.createRef<HTMLIFrameElement>();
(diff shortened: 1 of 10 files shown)


rob-ghost:
@coderabbitai review

apps/comments-ui/src/app.tsx
```diff
@@ -165,8 +169,45 @@ const App: React.FC<AppProps> = ({scriptTag}) => {
         try {
             // Fetch data from API, links, preview, dev sources
             const {member, labs} = await api.init();
-            const {comments, pagination, count} = await fetchComments();
-            const state = {
+
+            // Load comments, fetching additional pages if needed to find permalink target
```
rob-ghost: We need to handle replies which are behind "load more"

apps/comments-ui/src/components/content/comment.tsx
```diff
@@ -320,7 +329,14 @@ const CommentHeader: React.FC<CommentHeaderProps> = ({comment, className = ''})
...
                         <MemberExpertise comment={comment}/>
-                        <span title={formatExplicitTime(comment.created_at)}><span className="mx-[0.3em]">·</span>{createdAtRelative}</span>
+                        <a
+                            className="hover:underline"
+                            href={buildCommentPermalink(pageUrl, comment.id)}
+                            target="_parent"
+                            title={formatExplicitTime(comment.created_at)}
+                        >
+                            <span className="mx-[0.3em]">·</span>{createdAtRelative}
+                        </a>
```
rob-ghost: Clicking "Reply to <parent>" should also behave like a permalink click on the parent comment

apps/comments-ui/src/app.tsx
```diff
@@ -33,7 +35,9 @@ const App: React.FC<AppProps> = ({scriptTag}) => {
         order: 'count__likes desc, created_at desc',
         adminApi: null,
         commentsIsLoading: false,
-        commentIdToHighlight: null
+        commentIdToHighlight: null,
+        commentIdToScrollTo: initialCommentId,
```
rob-ghost, line 46: Do we want this behind a flag? Can we actually use flags on comment-ui?

apps/comments-ui/src/components/content/comment.tsx
```diff
@@ -52,13 +56,18 @@ export const CommentComponent: React.FC<CommentProps> = ({comment, parent}) => {
...
     } else if (showCommentContent && !showHiddenMessage) {
-        return <PublishedComment comment={comment} openEditMode={openEditMode} parent={parent} />;
+        content = <PublishedComment comment={comment} openEditMode={openEditMode} parent={parent} />;
     }
 
-    return null;
+    return (
+        <div id={id}>
+            {content}
+        </div>
```
rob-ghost: Is this still necessary? Ideally we can keep the `id` on the `Transition`

apps/comments-ui/src/components/content/comment.tsx
```diff
@@ -320,7 +329,14 @@ const CommentHeader: React.FC<CommentHeaderProps> = ({comment, className = ''})
                 <div className="flex items-baseline pr-4 font-sans text-base leading-snug text-neutral-900/50 sm:text-sm dark:text-white/60">
                     <span>
                         <MemberExpertise comment={comment}/>
-                        <span title={formatExplicitTime(comment.created_at)}><span className="mx-[0.3em]">·</span>{createdAtRelative}</span>
+                        <a
+                            className="hover:underline"
+                            href={buildCommentPermalink(pageUrl, comment.id)}
```
rob-ghost: Question: Can you use this `href` to semantically fetch the appropriate comment element so we don't have to use `id`

rob-ghost: If the `id` of the comment element is the same as this anchor hash value, will the browser automatically scroll to it on click?

apps/comments-ui/src/utils/helpers.ts
```diff
@@ -194,6 +208,43 @@ export const scrollToElement = (element: HTMLElement) => {
...
+            break;
+        }
+        currentWindow = currentWindow.parent;
+    }
+
+    // currentWindow is now the top-level window (works for both iframe and non-iframe)
+    currentWindow.scrollTo({
+        top: absoluteTop - (currentWindow.innerHeight / 2) + (element.offsetHeight / 2),
+        behavior: 'smooth'
+    });
```
rob-ghost: Can we just use `scrollIntoView()` on the element? Seems to work on console when selecting the comment element.

apps/comments-ui/src/styles/iframe.css
```diff
@@ -2,6 +2,16 @@
...
+/* Highlight animation for comment permalinks */
+@keyframes highlight {
+    0% {
+        background-color: rgb(253 224 71 / 0.5); /* yellow-300/50 */
+    }
+    100% {
+        background-color: transparent;
+    }
+}
+
```
rob-ghost: Highlighting for linking to `Reply to:` works, we shouldn't need this if we're re-using that functionality

apps/comments-ui/src/app.tsx
```diff
@@ -165,8 +169,43 @@ const App: React.FC<AppProps> = ({scriptTag}) => {
...
+                        const nextPage = await api.comments.browse({
+                            page: pagination.page + 1,
+                            postId: options.postId,
+                            order: state.order
+                        });
+                        comments = [...comments, ...nextPage.comments];
+                        pagination = nextPage.meta.pagination;
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
rob-ghost: We don't have console logs in this code by convention.

apps/comments-ui/src/app.tsx
```diff
@@ -160,13 +171,113 @@ const App: React.FC<AppProps> = ({scriptTag}) => {
...
+        if (!hasMoreReplies) {
+            return comments;
+        }
+
+        const repliesResponse = await api.comments.replies({commentId: parentId, limit: 100});
+        return comments.map(c => (c.id === parentId
+            ? {...c, replies: repliesResponse.comments}
+            : c
+        ));
+    };
```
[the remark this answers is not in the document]
rob-ghost: 1. The type is correct, we don't need to be overly defensive
2. The limit is the existing limit for loading more replies, but its true that we should handle the case where there might be more than 1 page of replies. I'll action this.

kevinansfield (approved):

kevinansfield:
Works well and I haven't managed to find any edge cases in testing 🙌 Would be nice to optimise once we have a better API for loading comments mid-list

PR #25715, opened 2025-12-15
Removed comment tombstone serialization

rob-ghost:
## Why?

PR #25700 introduced "tombstones" - a pattern where deleted comments with replies would be returned with minimal data (`id`, `parent_id`, `status` only). This was intended to preserve thread structure while hiding deleted content.

However, tombstones caused issues because they were missing expected attributes like `member` and `replies`, which broke assumptions in consuming code and made the API response inconsistent.

## What does it do?

**Both Admin and Public APIs** now preserve thread structure by including deleted parents when they have published replies. Deleted comments return full attributes with `html: null`, so the UI can show appropriate placeholder text ("This comment has been removed").

**New Browse All endpoint** (`GET /comments/`): Uses `browseAll: true` flag to exclude deleted comments entirely. This provides a flat list without thread preservation, suitable for admin moderation tables.

## Changes

- Removed tombstone serialization from comment mapper
- Deleted comments now return `html: null` (content never exposed)
- Both Admin and Public APIs preserve thread structure for deleted parents with replies
- Added `browseAll` option for the new `/comments` endpoint to exclude deleted comments

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/comments.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/comments.js
@@ -48,25 +48,6 @@ const commentMapper = (model, frame) => {
 
     const isPublicRequest = utils.isMembersAPI(frame);
 
-    // Deleted comments are tombstones - just enough info to show "Reply to deleted comment"
-    if (jsonModel.status === 'deleted') {
-        const tombstone = {
-            id: jsonModel.id,
-            parent_id: jsonModel.parent_id || null,
-            status: 'deleted'
-        };
-        // Include post relation if loaded
-        if (jsonModel.post) {
-            url.forPost(jsonModel.post.id, jsonModel.post, frame);
-            tombstone.post = _.pick(jsonModel.post, postFields);
-        }
-        // Include replies if present (for tree structure), recursively mapped
-        if (jsonModel.replies) {
-            tombstone.replies = jsonModel.replies.map(reply => commentMapper(reply, frame));
-        }
-        return tombstone;
-    }
-
     if (jsonModel.inReplyTo && (jsonModel.inReplyTo.status === 'published' || (!isPublicRequest && jsonModel.inReplyTo.status === 'hidden'))) {
         jsonModel.in_reply_to_snippet = htmlToPlaintext.commentSnippet(jsonModel.inReplyTo.html);
     } else if (jsonModel.inReplyTo && jsonModel.inReplyTo.status !== 'published') {
@@ -113,6 +94,11 @@ const commentMapper = (model, frame) => {
         if (jsonModel.status !== 'published') {
             response.html = null;
         }
+    }
+
+    // Deleted comments should never expose their content
+    if (jsonModel.status === 'deleted') {
+        response.html = null;
     }
 
     return response;
--- a/ghost/core/core/server/models/comment.js
+++ b/ghost/core/core/server/models/comment.js
@@ -63,18 +63,24 @@ const Comment = ghostBookshelf.Model.extend({
 
     // Called by our filtered-collection bookshelf plugin
     applyCustomQuery(options) {
-        const excludedCommentStatuses = options.isAdmin ? ['deleted'] : ['hidden', 'deleted'];
+        const excludedStatuses = options.isAdmin ? ['deleted'] : ['hidden', 'deleted'];
 
         this.query((qb) => {
-            qb.where(function () {
-                this.whereNotIn('comments.status', excludedCommentStatuses)
-                    .orWhereExists(function () {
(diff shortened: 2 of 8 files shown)


ghost/core/core/server/api/endpoints/utils/serializers/output/mappers/comments.js
jonatansberg, line 97: It would be sensible to still clear the HTML content for deleted comments, like in this commit:
https://github.com/TryGhost/Ghost/pull/25699/changes/f43b3aa2e2b2dae5a17952fc1553e00d4084df36

jonatansberg (changes requested):
We shouldn't include the content for deleted comments in the responses. I've verified that this already works in my branch.

ghost/core/core/server/models/comment.js
```diff
@@ -63,19 +63,28 @@ const Comment = ghostBookshelf.Model.extend({
...
-                        this.select(1)
-                            .from('comments as replies')
-                            .whereRaw('replies.parent_id = comments.id')
-                            .whereNotIn('replies.status', excludedCommentStatuses);
-                    });
+        if (options.isAdmin) {
+            // Admin: simply exclude deleted comments, no exceptions
+            this.query((qb) => {
+                qb.whereNot('comments.status', 'deleted');
+            });
```
kevinansfield: Was this tested on the front-end when logged in as an Admin? In that scenario, Comments-UI uses the Admin API for all comments display and relies on deleted comments still being included when they have replies

apps/comments-ui/test/e2e/content.test.ts
```diff
@@ -135,5 +136,138 @@ test.describe('Deleted and Hidden Content', async () => {
         // parent comment is hidden but shows text
         await expect (frame.getByText('This comment has been hidden')).toBeVisible();
     });
+
+    test.describe('Deleted Comment Behavior', async () => {
+        // These tests verify the correct handling of deleted comments:
+        // - Leaf deleted comments (no replies) should NOT be shown
+        // - Deleted comments WITH replies should show a placeholder
+        // - Replies can reference deleted comments via in_reply_to_id
```
kevinansfield: There's potentially some mismatch here, these tests use `MockedApi` so if that doesn't match our real API behaviour (afaics `MockedAPI` always includes deleted comments/replies for Admin API requests which this PR changes in the real API) then these tests are misleading.

ghost/core/core/server/models/comment.js
rob-ghost: Testing is a bit of a mess for this atm, but yes Comments UI uses the same admin endpoints and this is a regression I'm working on it.

jonatansberg: From the admin frontend perspective, it would be fine to include the deleted comments as long as their shape matches the other comments.

rob-ghost: The change came from `/comments` returning deleted comments but only if they had replies, which doesn't make sense in the context of the table view since there's no threading context so its really confusing to have some deleted comments but not others.

So the goal was to have the admin API ignore all deleted comments, but since the code is tightly coupled it regressed behaviour of `admin/comments/{id}` which Comments UI uses if you're logged in as admin.

apps/comments-ui/test/e2e/content.test.ts
rob-ghost: So much for `e2e` 🙄

kevinansfield (approved):

ghost/core/core/server/models/comment.js
```diff
@@ -63,18 +63,24 @@ const Comment = ghostBookshelf.Model.extend({
...
-                    .orWhereExists(function () {
-                        this.select(1)
-                            .from('comments as replies')
-                            .whereRaw('replies.parent_id = comments.id')
-                            .whereNotIn('replies.status', excludedCommentStatuses);
-                    });
-            });
+            if (options.browseAll) {
+                // Browse All: simply exclude statuses, no thread structure preservation
+                qb.whereNotIn('comments.status', excludedStatuses);
```
kevinansfield, line 71: Non blocker: It would be great to find a more long-term solution to this behaviour switch that will scale/be less likely to bite us as we build out more features.

jonatansberg (approved):

ghost/core/core/server/models/comment.js
rob-ghost: Yep I have a few ideas but I didn't want to bloat this PR with trying to solve it now.

It will boil down to removing the flags and having something more composable which each individual use case can build upon, as theres:

1. What status to include
2. If we should be including threaded replies or not
3. Should we be including roots or leaves

Depends on which API you're using and we're getting by with flags but they're a major 🚩

PR #25718, opened 2025-12-15
Added rendering output tests for URL transformations

vershwal:
ref https://linear.app/ghost/issue/PRO-1516/
ref https://linear.app/ghost/issue/PRO-1520/

- Added tests to verify that `__GHOST_URL__` placeholders are transformed to absolute site URLs or CDN url while rendering posts(depending on the config)
- Added `e2e-frontend/rendered_content.test.js` for HTML and RSS rendering
- Added email preview tests in `e2e-api/admin/email-previews.test.js`
- Email preview tests are placed in `e2e-api/admin/email-previews.test.js` because email previews are fetched via the Admin API endpoint, not rendered as frontend routes.
- Email preview tests only check image URLs (feature, inline, gallery, thumbnails) because email rendering doesn't include file/video/audio source URLs directly - these cards link to the post page instead of the actual media files.


allouis (approved):
Perfect

PR #25720, opened 2025-12-15
🐛 Fixed missing member discount data after migrations

sagzy:
ref https://linear.app/ghost/issue/ONC-1284

- before: during migrations, paid members with a Stripe coupon would be created without the corresponding offer in Ghost. As a result, while they were charged the right discounted amount by Stripe, they would appear as if they were paying the full subscription price in Ghost Admin or Portal
- now: we create an offer based on Stripe coupons during migrations. That way, the subscription price in Ghost matches the subscription price in Stripe, taking into account the discount
- the offer is marked as archived, so that it cannot be used for signups. The name is computed based on the coupon type, duration and code for uniqueness, e.g. `"15% off forever (dai8HKn)"`


ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -1036,22 +1038,26 @@ module.exports = class MemberRepository {
...
+                    name: ghostProduct.get('name')
+                },
+                transacting: options.transacting
+            });
+
+            if (ensuredOfferId) {
+                offerId = ensuredOfferId;
             }
-        } else if (offerId) {
-            offer = await this._offerRepository.getById(offerId, {transacting: options.transacting});
```
sagzy, line 1054: Note: we don't actually use `offer` in the rest of the `linkSubscription` method, so this code wasn't actually being used

ghost/core/core/server/services/members/members-api/repositories/MemberRepository.js
```diff
@@ -1036,22 +1038,26 @@ module.exports = class MemberRepository {
...
 
-        if (stripeCouponId) {
-            // Get the offer from our database
-            offer = await this._offerRepository.getByStripeCouponId(stripeCouponId, {transacting: options.transacting});
-            if (offer) {
-                offerId = offer.id;
-            } else {
-                logging.error(`Received an unknown stripe coupon id (${stripeCouponId}) for subscription - ${stripeSubscriptionData.id}.`);
+        if (stripeCouponId && !offerId && ghostProduct) {
+            const cadence = _.get(subscriptionPriceData, 'recurring.interval', 'month');
```
allouis: Why do we default to `'month'` here? Prices used for subscriptions must have a recurring property

ghost/core/core/server/services/offers/OffersImportService.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Generate a human-readable name/title for an offer based on coupon.
+     * @private
+     */
+    _generateNameFromCoupon(coupon) {
+        const durationText = coupon.duration === 'repeating' && coupon.duration_in_months
+            ? `for ${coupon.duration_in_months} months`
+            : coupon.duration;
```
allouis: I don't think this case should be possible - `duration_in_months` is `null` if `duration` is `'once'` or `'forever'` - otherwise it's a positive integer

ghost/core/core/server/services/offers/OffersImportService.js
```diff
@@ -0,0 +1,138 @@
...
+     * Ensure an offer exists for the given Stripe coupon.
+     * Returns offerId or null.
+     *
+     * @param {object} params
+     * @param {object} params.coupon
+     * @param {'month'|'year'} params.cadence
+     * @param {{id: string, name?: string}} params.tier
+     * @param {import('knex').Knex.Transaction} [params.transacting]
+     */
+    async ensureOfferForCoupon({coupon, cadence, tier, transacting}) {
```
allouis: I don't think this method should return `null`

We can make `coupon` a required input, as well as the `id` property of it.

As for the missing `offerRepository` and `offersAPI` we can validate that in the constructor and throw there, failing fast means we'll know immediately if Ghost is broken

I also don't think `_mapCouponToOfferData` should return `null` either - that just leaves us with a failure to create the coupon

ghost/core/core/server/services/offers/OffersImportService.js
```diff
@@ -0,0 +1,138 @@
...
+            type = 'percent';
+            amount = coupon.percent_off;
+        } else if (coupon.amount_off) {
+            type = 'fixed';
+            amount = coupon.amount_off;
+            currency = coupon.currency;
+            if (!currency) {
+                return null;
+            }
+        } else {
```
allouis: I don't think this branch should be possible - coupons have either `amount_off` or `percent_off`

ghost/core/core/server/services/offers/OffersImportService.js
```diff
@@ -0,0 +1,138 @@
...
+            return null;
+        }
+
+        const duration = coupon.duration;
+        const durationInMonths = duration === 'repeating'
+            ? coupon.duration_in_months
+            : null;
```
allouis: Same here - this branch should not be possible

ghost/core/core/server/services/offers/OffersImportService.js
```diff
@@ -0,0 +1,138 @@
...
+
+        const offerData = this._mapCouponToOfferData(coupon, {cadence, tier});
+        if (!offerData) {
+            this.logger.warn(`Skipped creating offer from Stripe coupon ${coupon.id} due to unsupported coupon configuration.`);
+            return null;
+        }
+
+        try {
+            const created = await this.offersAPI.createOffer(offerData, {transacting});
+            return created?.id || null;
```
allouis: `offersAPI.createOffer` always returns an Offer - so we don't need this clause

ghost/core/core/server/services/offers/OffersImportService.js
```diff
@@ -0,0 +1,138 @@
...
+        let currency = null;
+
+        if (coupon.percent_off) {
+            type = 'percent';
+            amount = coupon.percent_off;
+        } else if (coupon.amount_off) {
+            type = 'fixed';
+            amount = coupon.amount_off;
+            currency = coupon.currency;
+            if (!currency) {
```
allouis: Is this possible?

ghost/core/core/server/services/offers/OffersImportService.js
```diff
@@ -0,0 +1,138 @@
...
+        if (!coupon?.id) {
+            return null;
+        }
+
+        if (!this.offerRepository || !this.offersAPI) {
+            this.logger.warn('Offer import unavailable: missing repository or offersAPI');
+            return null;
+        }
+
+        const existing = await this.offerRepository.getByStripeCouponId(coupon.id, {transacting});
```
allouis: This gets called as part of `linkSubscription` which gets called on incoming webhooks - is there chance of a race condition here? Let's say we get two incoming webhooks, both of them pass this step with `existing === null`, and then they both go on to attempt to create the offer?

allouis (commented):
Overall I think the solution of creating offers from coupons make sense, I have one concern about a race condition, and I also think we can simplify this implementation a bit!

**`null` checks and flow**

There is a lot of defensive programming here which makes the flow more complicated, we do a bunch of checks against `null` where `null` should not be possible under the Stripe API contract

IMO we should use types to make it clear the data structures we're working with (and expect) - and we can do a single top level validation and throw (fail fast) if things don't match. That way we're immediately alerted to the fact that something is broken between us and Stripe.

_If we really don't want to throw we can log an error, but at least the downstream logic doesn't need null checks everywhere_

**new OffersImportService**

I'm not 100% sure we need this 🤔 The existing pattern for creating Offers is a static `create` factory method on the `Offer` domain object - I think we could add a `createFromStripeCoupon` factory too, which would encapsulate the logic of generating the name and offer data. 

And then we have the `ensureOfferForCoupon` - which could just live on the `OffersAPI`? This second part is less of a deal to me tbh but curious what you think

ghost/core/core/server/services/offers/OffersImportService.js
sagzy: Yes good shout! Think in this case, the second offer creation would fail due to the unique constraint on `code` or `stripe_coupon_id` in the database

ghost/core/core/server/services/offers/domain/models/StripeCoupon.js
```diff
@@ -0,0 +1,106 @@
...
+            throw new InvalidStripeCoupon({
+                message: 'Stripe coupon `duration` is required and must be a string.'
+            });
+        }
+
+        if (input.duration_in_months && typeof input.duration_in_months !== 'number') {
+            throw new InvalidStripeCoupon({
+                message: 'Stripe coupon `duration_in_months` must be a number.'
+            });
+        }
```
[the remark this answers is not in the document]
sagzy: thanks fixed!

ghost/core/core/server/services/offers/OffersImportService.js
sagzy: Great catch! 🎣

Added protection against race condition in 76db34b550cd5d96e7d3a1ee10e74ad000048e5d, we now:
1. Check that the offer exists
2. Try to create a new one
3. If it fails, check again if the offer exists
4. Else throw

https://github.com/TryGhost/Ghost/blob/76db34b550cd5d96e7d3a1ee10e74ad000048e5d/ghost/core/core/server/services/offers/application/OffersAPI.js#L144C1-L164C14

sagzy:
@allouis great review, thank you, really appreciate you taking the time!

I have refactored the code in 76db34b550cd5d96e7d3a1ee10e74ad000048e5d based on your feedback, curious to hear what you think! The code now fails loudly, uses the existing offersAPI instead of a new service and handles the race condition

ghost/core/core/server/services/offers/domain/models/StripeCoupon.js
```diff
@@ -0,0 +1,112 @@
...
+ * @property {string} id
+ * @property {number} [percent_off]
+ * @property {number} [amount_off]
+ * @property {string} [currency]
+ * @property {string} duration
+ * @property {number} [duration_in_months]
+ */
+
+/** @extends ValueObject<StripeCouponInput> **/
+class StripeCoupon extends ValueObject {
```
sagzy, line 16: @allouis I was on the fence about these validations: 
- on one hand, we shouldn't need to validate the format of Stripe Coupons
- on the other hand, if the API contract changes after a Stripe upgrade and our code needs adapting, it'd be great to see a readable error in the logs and an unit test failure

ghost/core/core/server/services/offers/application/OffersAPI.js
```diff
@@ -122,6 +121,57 @@ class OffersAPI {
...
+                // Handle race condition: another request may have created the offer
+                // between the check and save. If so, return the existing offer.
+                if (err.code === 'ER_DUP_ENTRY' || err.code === 'SQLITE_CONSTRAINT') {
+                    const createdOffer = await this.repository.getByStripeCouponId(coupon.id, txOptions);
+                    if (createdOffer) {
+                        return OfferMapper.toDTO(createdOffer);
+                    }
+                }
+                throw err;
+            }
```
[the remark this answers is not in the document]
allouis: I do agree here @sagzy - this is one of those "should never happen" cases, where it might be useful to have an error, similar to here https://github.com/TryGhost/ActivityPub/blob/5663f91015bf9b2541a48a57a0cf364167ca9d7d/src/account/account.service.ts#L158-L164

It should never happen - but if it does, super useful to have a specific string to search for wtf is going on.

Not a big deal though - either way we do error, and I'm sure we can track it down

ghost/core/core/server/services/offers/domain/models/StripeCoupon.js
allouis: I think these are great - we've moved all of those null checks into explicit assertions, and they're all in one place which is easy to find. Maybe it doesn't belong in the "domain" and could be validation at the API layer - but I think that's nitpicking 🐜

ghost/core/test/e2e-api/members/webhooks.test.js
```diff
@@ -1680,17 +1684,22 @@ describe('Members API', function () {
             });
         });
 
-        it('Silently ignores an invalid offer id in metadata', async function () {
+        it('Supports creating an offer from a Stripe coupon', async function () {
```
allouis, line 1687: We had an explicit test for the opposite of this behaviour! At least it was half expected I guess 😁

allouis (approved):

ghost/core/core/server/services/offers/application/OffersAPI.js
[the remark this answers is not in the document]
sagzy: added in c8183be3b3e9f87b56029667320ceec85552d1d2

ghost/core/test/e2e-api/members/webhooks.test.js
sagzy: Haha yeah! Outside of migration flows, I think it's fair to assume that the offer will pre-exist in 99% of the cases and we don't need to "generate" it? Maybe a limitation we were aware of, but didn't want to complicate things for an edge case (outside of migrations) - don't know :)

PR #25724, opened 2025-12-16
Improved date filter for comments moderation

jonatansberg:
closes https://linear.app/ghost/issue/FEA-490/fix-date-filter-behavior-and-operators-for-comments

Simplified date filter operators to is, before, after and fixed "is" operator to match entire calendar day instead of exact timestamp.

### DIFF
--- a/apps/posts/src/views/comments/comments.tsx
+++ b/apps/posts/src/views/comments/comments.tsx
@@ -11,54 +11,59 @@ function buildCommentsFilter(filters: Filter[]): string | undefined {
     const parts: string[] = [];
     
     for (const filter of filters) {
+        if (!filter.values[0]) {
+            continue;
+        }
+
         switch (filter.field) {
-        case 'status':
-            if (filter.values[0]) {
-                parts.push(`status:${filter.values[0]}`);
-            }
+        case 'status': 
+            parts.push(`status:${filter.values[0]}`);
             break;
-        case 'created_at':
+
+        case 'created_at': 
             if (filter.operator === 'before' && filter.values[0]) {
                 parts.push(`created_at:<'${filter.values[0]}'`);
             } else if (filter.operator === 'after' && filter.values[0]) {
                 parts.push(`created_at:>'${filter.values[0]}'`);
             } else if (filter.operator === 'is' && filter.values[0]) {
-                parts.push(`created_at:'${filter.values[0]}'`);
-            } else if (filter.operator === 'between' && filter.values[0] && filter.values[1]) {
-                parts.push(`created_at:>='${filter.values[0]}'+created_at:<='${filter.values[1]}'`);
+                // Match all items from the selected day in the user's timezone
+                const dateValue = String(filter.values[0]); // Format: YYYY-MM-DD
+                    
+                // Create Date objects in user's local timezone, then convert to UTC
+                const startOfDay = new Date(dateValue + 'T00:00:00').toISOString();
+                const endOfDay = new Date(dateValue + 'T23:59:59.999').toISOString();
+                    
+                parts.push(`created_at:>='${startOfDay}'+created_at:<='${endOfDay}'`);
             }
             break;
-        case 'body':
-            if (filter.values[0]) {
-                const value = filter.values[0] as string;
-                // Escape single quotes in the value
-                const escapedValue = value.replace(/'/g, '\\\'');
-                
-                if (filter.operator === 'contains') {
-                    parts.push(`html:~'${escapedValue}'`);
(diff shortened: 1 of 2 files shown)


apps/posts/src/views/comments/comments.tsx
rob-ghost: **Nitpick(out of scope):** If we move this to be a guard then we can remove this common check from all the subsequent cases:

```typescript
for (const filter of filters) {
    const [ value ] = filter.values;

    if (!value) {
        continue;
    }

    switch (filter.field) {
        case 'status':
            // No need to have a condition here
            parts.push(`status:${values}`);
            break;
        // ...
    }
}
```

apps/posts/src/views/comments/comments.tsx
```diff
@@ -23,9 +23,14 @@ function buildCommentsFilter(filters: Filter[]): string | undefined {
             } else if (filter.operator === 'after' && filter.values[0]) {
                 parts.push(`created_at:>'${filter.values[0]}'`);
             } else if (filter.operator === 'is' && filter.values[0]) {
-                parts.push(`created_at:'${filter.values[0]}'`);
-            } else if (filter.operator === 'between' && filter.values[0] && filter.values[1]) {
-                parts.push(`created_at:>='${filter.values[0]}'+created_at:<='${filter.values[1]}'`);
+                // Match all items from the selected day in the user's timezone
+                const dateValue = filter.values[0] as string; // Format: YYYY-MM-DD
```
rob-ghost: **Question:** I see `Filter` has `values` as `T[]`. Is there any occasion where we're not using string values for the filters?

Could we define this as `Filter[string]` so we don't have to cast or do we use other types like `number`, and would it be simpler to just use `string` for everything and do the parsing to other types here?

rob-ghost (approved):
A couple of nitpick/question comments but 👍🏻 this can be merged, nice one!

apps/posts/src/views/comments/comments.tsx
jonatansberg: I think we imported the filter components from somewhere. They sit at 2k+ lines of code without any tests. If you start pulling a thread somewhere then it all kinda falls apart. 

Long term, we should do something about that. For now, I'd generally prefer to touch them as little as possible.

apps/posts/src/views/comments/comments.tsx
jonatansberg: Fixed

apps/posts/src/views/comments/comments.tsx
jonatansberg: I cleared up the ugly typecast as least.

apps/posts/src/views/comments/comments.tsx
```diff
@@ -11,54 +11,59 @@ function buildCommentsFilter(filters: Filter[]): string | undefined {
...
-            } else if (filter.operator === 'between' && filter.values[0] && filter.values[1]) {
-                parts.push(`created_at:>='${filter.values[0]}'+created_at:<='${filter.values[1]}'`);
+                // Match all items from the selected day in the user's timezone
+                const dateValue = String(filter.values[0]); // Format: YYYY-MM-DD
+                    
+                // Create Date objects in user's local timezone, then convert to UTC
+                const startOfDay = new Date(dateValue + 'T00:00:00').toISOString();
+                const endOfDay = new Date(dateValue + 'T23:59:59.999').toISOString();
+                    
+                parts.push(`created_at:>='${startOfDay}'+created_at:<='${endOfDay}'`);
```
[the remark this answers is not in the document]
jonatansberg: .toISOString() does convert the generated datetime to UTC. That's correct. But it's offset according to the user's local time zone. This is the intended behavior since the server time is also in UTC. 

This way, the time zones do line up and you only see the expected results corresponding to the correct calendar date window in your local time zone.

apps/posts/src/views/comments/comments.tsx
```diff
@@ -11,54 +11,59 @@ function buildCommentsFilter(filters: Filter[]): string | undefined {
...
+            const value = filter.values[0] as string;
+            // Escape single quotes in the value
+            const escapedValue = value.replace(/'/g, '\\\'');
+                    
+            if (filter.operator === 'contains') {
+                parts.push(`html:~'${escapedValue}'`);
+            } else if (filter.operator === 'not_contains') {
+                parts.push(`html:-~'${escapedValue}'`);
             }
             break;
```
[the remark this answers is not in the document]
jonatansberg: I added some brackets, but our current code formatting makes that look like shit. Since there aren't any conflicts for these variables, I don't think this is an issue.

PR #25726, opened 2025-12-16
Improved comments moderation status filters and hidden state visibility

jonatansberg:
refs https://linear.app/ghost/issue/FEA-496/adjust-status-filters-and-distinguish-published-vs-hidden

- Removed "Deleted" filter option since deleted comments aren't returned
- Added EyeOff icon and muted styling to distinguish hidden comments

### DIFF
--- a/apps/posts/src/views/comments/components/comments-list.tsx
+++ b/apps/posts/src/views/comments/components/comments-list.tsx
@@ -137,13 +137,15 @@ function CommentsList({
                                 data-testid="comment-list-row"
                             >
                                 <TableCell className="static col-start-1 col-end-1 row-start-1 row-end-1 flex min-w-0 flex-col p-0 md:relative lg:table-cell lg:p-4">
-                                    {item.html ? (
-                                        <div dangerouslySetInnerHTML={{__html: item.html}} className="prose block text-base" />
-                                    ) : (
-                                        <span className="text-muted-foreground">
-                                            Deleted comment
-                                        </span>
-                                    )}
+                                    {item.status === 'hidden' 
+                                        ? (
+                                            <div title="This comment is hidden">
+                                                <LucideIcon.EyeOff className="float-left size-5 pr-2 text-muted-foreground" />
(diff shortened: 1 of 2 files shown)


apps/posts/src/views/comments/components/comments-list.tsx
```diff
@@ -137,13 +137,21 @@ function CommentsList({
...
-                                        </span>
-                                    )}
+                                    {item.status === 'hidden' 
+                                        ? (
+                                            <div title="This comment is hidden">
+                                                <LucideIcon.EyeOff className="float-left size-5 pr-2 text-muted-foreground" />
+                                                <div dangerouslySetInnerHTML={{__html: item.html || ''}} className="prose block text-base text-muted-foreground" />
+                                            </div>
+                                        ) 
+                                        : item.html
```
rob-ghost: **Suggestion:** We handle optional `html` with `__html: item.html || ''` in the above case, we should do the same here since we shouldn't have any deleted comments in this view anyway. This simplifies the ternaries into hidden or not hidden with optional HTML.

jonatansberg: I thought about including this from the get go. I wasn't sure if we would actually completely eliminate the display of deleted comments in the admin. That's why I left it in. 

Fixed now.

apps/posts/src/views/comments/components/comments-list.tsx
```diff
@@ -137,13 +137,21 @@ function CommentsList({
...
+                                            </div>
+                                        ) 
+                                        : item.html 
+                                            ? (
+                                                <div dangerouslySetInnerHTML={{__html: item.html}} className="prose block text-base" />
+                                            ) : (
+                                                <span className="text-muted-foreground">
+                                                    Deleted comment
+                                                </span>
+                                            )}
```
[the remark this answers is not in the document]
jonatansberg: The HTML is sanitized in the backend. I did consider adding extra sanitization here too tho as user comments would be untrusted.

rob-ghost (approved):

PR #25728, opened 2025-12-16
Changed link redirects from middleware to router

rmgpinto:
ref https://linear.app/ghost/issue/BER-3095

Extracted the link redirects code from the middleware to the router. This avoids route comparison on every requests Ghost receives.

---

> [!NOTE]
> Moves link redirect handling out of global middleware and into a dedicated router for clearer routing and fewer per-request checks.
> 
> - New `routers/link-redirects.js` registers `GET` handler at `linkRedirects.service.relativeRedirectPrefix() + '*'`; `site.js` mounts this router
> - `LinkRedirectsService`: adds `relativeRedirectPrefix()` and removes in-method prefix checks; `handleRequest` now assumes routing filtered paths
> - Tests: add coverage for `relativeRedirectPrefix`, update redirect behavior tests, and drop non-matching-prefix cases
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 93fa62fbb0e9b7b7110a2d648ecf418988e3aa55. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/core/core/frontend/web/routers/link-redirects.js
+++ b/ghost/core/core/frontend/web/routers/link-redirects.js
@@ -1,0 +1,5 @@
+const linkRedirects = require('../../../server/services/link-redirection');
+
+module.exports = function handleRedirects(siteApp) {
+    siteApp.get(linkRedirects.service.relativeRedirectPrefix() + '*', linkRedirects.service.handleRequest);
+};
--- a/ghost/core/core/frontend/web/site.js
+++ b/ghost/core/core/frontend/web/site.js
@@ -16,7 +16,7 @@ const themeMiddleware = themeEngine.middleware;
 const membersService = require('../../server/services/members');
 const offersService = require('../../server/services/offers');
 const customRedirects = require('../../server/services/custom-redirects');
-const linkRedirects = require('../../server/services/link-redirection');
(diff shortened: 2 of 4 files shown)


rmgpinto:
@coderabbitai review

9larsons (changes requested):
I'll need to look deeper into this to understand the original behavior/intent. Could you resolve the conflict so we can see if the cursor bot comment is valid? More tests may need updating.

Is it actually cleaner to use this router? I'm finding that a confusing abstraction vs. using siteApp.get with the redirects service directly in `site.js`.

9larsons:
Removed tests that are no longer relevant; `handleRequest` should no longer receive redirects as they are handled in the router. We have redirect related e2e tests in `ghost/core/test/e2e-server/click-tracking.test.js` and testing the router would effectively be testing Express.

We should test this on staging before merging, however I'm in favor of this change overall. We should see better performance with the `/r/` endpoint which is one of the highest-traffic sources for the container to handle.

9larsons (approved):
Approved but would like to see tested on staging, especially with a subdirectory setup.

Ghost-Slimer:
Deployed to **staging** with ID: `3960`

PR #25731, opened 2025-12-16
Hid the upgrade required banner in admin forward

jonatansberg:
no ref 

The "Ghost has been updated! To get access to the latest features, refresh or click here" banner doesn't really make sense in it's current format while in the continuously delivered version of the admin and it's confusing when it shows up on staging.


kevinansfield (approved):
I agree it's reasonable to disable this entirely until we get to some form of re-implementation.

We so rarely have backwards-incompatible API changes outside of feature flags that the upgrade banner isn't protecting anyone from errors, it's mainly there to encourage refreshes so new features show up. It would be interesting to have some stats on how often this banner is shown/actioned to even know if it's worth re-implementing.

PR #25736, opened 2025-12-17
Optimized filtered_sessions pipe to skip unnecessary join

cmraible:

The filtered_sessions pipe always joins with `mv_session_data`, even when no session-level filters (source, device, utm_*) were provided. This added overhead for the majority of analytics queries that only use hit-level filters. The pipe now conditionally skips the join when no session-level filters are present, passing session IDs directly from the first filtering stage.

The vast majority of requests don't include filters on these fields. By skipping the join, we avoid running the `mv_session_data` pipe one additional time. From local testing, this reduces the memory usage in our endpoints by ~15%, cpu time by ~14-25%, and ultimately reduces the amount of rows and bytes read by ~30% across all our Tinybird endpoints.


9larsons (approved):
I like this, good plan. I don't love how declarative we have to be but that's not something we can resolve without Tinybird adding more support.

PR #25737, opened 2025-12-17
Switched Tinybird endpoints from uniqExact to uniq for session counting

cmraible:

The uniqExact function computes exact unique counts which is more expensive than necessary for analytics dashboards. Using uniq (HyperLogLog) provides approximate counts with ~2-3% error margin, which is acceptable for visitor/session metrics and offers better query performance.


cmraible:
Opting not to merge this, as we prefer to keep using uniqExact() for more precise metrics

PR #25738, opened 2025-12-17
Optimized top_locations and top_pages endpoints to avoid double-scan

cmraible:

These endpoints were scanning `_mv_hits` twice: once through `filtered_sessions` to get qualifying session IDs, then again to aggregate by location/pathname. When no session-level filters (source, device, `utm_*`) are provided, the endpoints now query `_mv_hits` directly in a single pass. This should significantly reduce query time and timeouts for these high-traffic endpoints.


cmraible:
Closing because other improvements have gotten us into a much better spot now, and next steps will likely involve a more thorough rethink of how our Tinybird schema and endpoints are structured.

PR #25740, opened 2025-12-17
Updated Bluesky sharing copy in activitypub app

mike182uk:
no ref

Updated the Bluesky sharing copy in activitypub app so that it reads better


sagzy (approved):
No need for a review on copy changes right? :D

PR #25742, opened 2025-12-17
Changed `require()` on sentry to boot faster

rmgpinto:
ref https://linear.app/ghost/issue/BER-3132

- The `require('@tryghost/errors')` adds 50-100ms on boot time, this is due to the package's own initialization and `require()`. Since it's not needed at boot, it can be moved inside the `beforeSend` function.

### DIFF
--- a/ghost/core/core/shared/sentry.js
+++ b/ghost/core/core/shared/sentry.js
@@ -1,7 +1,7 @@
 const config = require('./config');
 const SentryKnexTracingIntegration = require('./SentryKnexTracingIntegration');
 const sentryConfig = config.get('sentry');
-const errors = require('@tryghost/errors');
+let _errors;
 
 /**
  * @param {import('@sentry/node').Event} event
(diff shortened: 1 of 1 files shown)


rmgpinto:
@coderabbitai review

kevinansfield (approved):

kevinansfield:
Approved 👍 Out of interest, is there anything we can do in @tryghost/errors to improve that packages performance across the board?

rmgpinto:
IIRC, Claude mentioned: https://github.com/TryGhost/framework/blob/main/packages/errors/src/utils.ts#L1 as the culprit.

kevinansfield:
In that case, would it be preferable to do the same delayed require in that package so we only pay the cost on the first stack trace generation?

rmgpinto:
I'd say so. I wanted to address the boot time on Ghost, but probably makes sense to get the package changed with what you suggested.

PR #25743, opened 2025-12-17
Cleaned up member import service to use factory interfaces in e2e

ibalosh:
Cleaned up member import service to use factory interfaces


maxbritto:
I see this PR introduces the notion of `import_tier` which is a great idea when importing existing users. Though the UI doesn't seem to accept this column if we add it in the CSV. 
Is it something that is planned for the future, or is there a way right now to work around this limited UI? 
CleanShot 2026-01-14 at 11 24 40

PR #25748, opened 2025-12-17
Reduced volume of network requests on Web traffic views

troyciesco:
ref https://linear.app/ghost/issue/NY-865

- reduced volume of api requests to Tinybird endpoints form posts/stats web views

The frontend apps are querying for all filter results at all times, which is an excessive amount of additional network traffic. We've pared back the handling to only request when filters are being selected by introducing a lazy loading pattern. While verbose, this is ok for now, while we get better definition on what we want the handling to be.

Acknowledged and not done: When we apply a filter, we store the list of values and the counts to that filter dropdown. If someone applies two filters, then goes back to click the first to change it, we have a stale set of results, as the results did _not_ have the second filter applied to them. Patching this up requires additional frequent queries, and we're going to validate whether or not we feel strongly about having the counts at all. We can follow up with a separate change for this.


troyciesco:
@coderabbitai review

9larsons (approved):

PR #25751, opened 2025-12-17
Removed Tinybird v2 endpoints

cmraible:

- The v2 endpoints were created as an experiment to see if the materialized view for `mv_session_data` would help improve response times and eliminate timeouts at Tinybird endpoints
- The v2 endpoints backed by the materialized view have a subtle, unintended behavioral change to how visits and pageviews are aggregated and bucketed by time, so we ultimately decided not to promote them to the "default" endpoints
- This removes the v2 endpoints, the mv_session_data_v2 materialized view and datasource, since they are currently live and consuming server resources, despite not being in active use by any sites.


cmraible:
Need to figure how to actually deploy this change. The Tinybird tests are failing because we can't delete materialized views or datasources without passing a special flag to allow destructive operations. Thinking maybe we can add the flag in CI for this commit, then as soon as the deployment is complete we can go back and remove the flag to avoid accidentally deleting stateful resources in the future.

PR #25753, opened 2025-12-18
Improve comment moderation filter UX

jonatansberg:
- Fixed issue with stale filter options 
- Reworked filter state implementation 
- Improved integration with browser history 
- Added scroll restoration


apps/posts/src/views/comments/hooks/use-filter-state.ts
```diff
@@ -0,0 +1,191 @@
+import {useMemo, useCallback} from 'react';
+import {useSearchParams} from '@tryghost/admin-x-framework';
+import type {Filter} from '@tryghost/shade';
+
+/**
+ * Comment filter field keys - single source of truth for filter definitions
+ */
+export const COMMENT_FILTER_FIELDS = ['status', 'created_at', 'body', 'post', 'author'] as const;
```
rob-ghost: **Question:** `reported` isn't listed here, should it be?

Its a case in `buildNqlFilter`.

apps/posts/src/views/comments/hooks/use-filter-options.ts
```diff
@@ -0,0 +1,63 @@
...
+        
+        // Add/update with search results (these take priority)
+        for (const item of searchResults ?? []) {
+            optionsMap[item.id] = transformToOption(item);
+        }
+
+        // Add ID fallback for active filter on deleted items
+        const activeFilter = filters.find(f => f.field === filterFieldName);
+        if (activeFilter && activeFilter.values[0]) {
+            const filterValue = String(activeFilter.values[0]) as string;
```
rob-ghost: **Nitpick:** No need to cast here, `String` returns `string`

apps/posts/src/views/comments/hooks/use-search-members.ts
```diff
@@ -7,7 +7,7 @@ export function useSearchMembers(searchTerm: string) {
     return useBrowseMembers({
         searchParams: {
             ...(debouncedSearch && {search: debouncedSearch}),
-            limit: '20',
+            limit: '100',
```
rob-ghost, line 10: **Suggestion:** Should we make a constant for these limits? `useBrowseMembers` and `useBrowsePosts` seem to share this limit and are being updated at the same time.

Or make them default in the `searchParams`.

apps/posts/src/views/comments/hooks/use-search-members.ts
rob-ghost: **Suggestion:** I know its out of scope of the changeset but this and `use-search-posts.ts` debounce the search term but with slightly different numbers, `300` and `200` respectively.

Should these be the same number? Should they then share a constant? Or should it be a default value in `useDebounce` and remove the specialisation if its not needed?

apps/posts/src/views/comments/components/comments-list.tsx
```diff
@@ -125,16 +126,22 @@ function CommentsList({
...
     hasNextPage?: boolean;
     isFetchingNextPage?: boolean;
     fetchNextPage: () => void;
     onAddFilter: (field: string, value: string, operator?: string) => void;
+    isLoading?: boolean;
 }) {
     const parentRef = useRef<HTMLDivElement>(null);
+    
+    // Restore scroll position when navigating back from filtered views
+    useScrollRestoration({parentRef, isLoading, debug: true});
```
rob-ghost: **Issue:** Un-used / unsupported `debug: true` parameter.

Pointed out by coderabbit elsewhere but wanted to leave a comment at the call site.

apps/posts/src/views/comments/hooks/use-filter-state.ts
```diff
@@ -0,0 +1,191 @@
+import {useCallback, useMemo} from 'react';
+import {useSearchParams} from '@tryghost/admin-x-framework';
+import type {Filter} from '@tryghost/shade';
+
+/**
+ * Comment filter field keys - single source of truth for filter definitions
+ */
+export const COMMENT_FILTER_FIELDS = ['status', 'created_at', 'body', 'post', 'author'] as const;
```
rob-ghost: **Suggestion(if-minor):** This is a bigger one so we can ignore it for now.

If we want to update our list of filters, we have to update 3 places:

1. Here in `COMMENT_FILTER_FIELDS`
2. Down below in `buildNqlFilter`
3. `filterFields` in `comments-filters.tsx`

This means we have some low-cohesion when it comes to the filter definitions.

We could define a filter type which supports all 3 concerns, and would give us high cohesion and one single source of truth for all things filters:

```typescript
interface CommentFilterFieldDefinition extends FilterFieldConfig {
    toNql: (filter: Filter) => string | undefined;
}

export const COMMENT_FILTER_DEFINITIONS: CommentFilterFieldDefinition[] = [
    {
        key: 'status',
        label: 'Status',
        type: 'select',
        icon: <LucideIcon.Circle className="size-4" />,
        options: [
            { value: 'published', label: 'Published' },
            { value: 'hidden', label: 'Hidden' },
        ],
        toNql: (filter) => filter.values[0] ? `status:${filter.values[0]}` : undefined,
    },
    // ...
]

const COMMENT_FILTER_DEFINITION_MAP = Object.fromEntries(
    COMMENT_FILTER_DEFINITIONS.map(def => [def.key, def])
);

export function buildNqlFilter(filters: Filter[]): string | undefined {
    return filters
        .map(filter => {
            const def = COMMENT_FILTER_DEFINITION_MAP[filter.field];
            return def!.toNql(filter);
        })
        .join('+');
  }
```

jonatansberg: Yeah, I've been thinking about this as well. It's not as easy as you'd wish. For e.g. the status filters it works fine, but with the dynamic filters it gets messy. 

I do agree tho that this is ripe for refactoring.

apps/posts/src/views/comments/components/comments-list.tsx
jonatansberg: I had already fixed this, but for some reason it wasn't committed 😅

apps/posts/src/views/comments/hooks/use-filter-state.ts
jonatansberg: Yes. Good catch! I think I missed this when I rebased on main.

apps/posts/src/views/comments/hooks/use-search-members.ts
jonatansberg: I made the numbers the same now. I think we can leave the rest as a follow-up.

apps/posts/src/views/comments/hooks/use-filter-state.ts
rob-ghost: Yeah I thought the dynamic ones would be harder. We could probably do a merge in the hook to support the dynamic stuff:

```typescript
const dynamicOptions = {
    author: { options: members.options, isLoading: members.isLoading, ... },
    post: { options: posts.options, isLoading: posts.isLoading, ... },
};

const filterFields = Object.entries(COMMENT_FILTER_CONFIG_MAP).map(
    ([key, config]) => ({
        ...config,
        ...dynamicOptions[key],
    })
);
```

jonatansberg: I tried to get Claude to straighten this out for me, but it made an even bigger mess. Seeing how I have about one hour left of my day before going off for the holidays, I think we'll have to leave this as a follow-up :)

apps/posts/src/views/comments/hooks/use-search-members.ts
jonatansberg: Not sure what the best course of action is here. I think for now the repetition is okay.

rob-ghost (approved):
Nice work 👍🏻 happy to revisit the refactoring / cleanups later

PR #25759, opened 2025-12-18
Added support for activitypub with improved dev workflow

mike182uk:
ref https://github.com/TryGhost/ActivityPub/pull/1487

Added ActivityPub routing and HMR support for external domains
- Added `dev-gateway` routing for ActivityPub API (`/.ghost/activitypub/*`) and federation endpoints (`/.well-known/webfinger`, `/.well-known/nodeinfo`)
- Fixed HMR WebSocket connections through reverse proxies by updating `vite-deep-links.ts` to skip WebSocket upgrades and root path requests
- Changed `allowedHosts` to `true` to support external domains (Tailscale, etc.)
- Added `/ghost/.well-known/*` route to Ghost backend for JWKS endpoint
- Simplified admin reverse proxy config (Caddy handles WebSocket automatically)

### DIFF
--- a/apps/admin/vite-deep-links.ts
+++ b/apps/admin/vite-deep-links.ts
@@ -16,19 +16,6 @@ export function deepLinksPlugin(): Plugin {
 
         return () => {
             server.middlewares.use((req, res, next) => {
-                // Skip WebSocket upgrade requests (used for HMR)
-                if (req.headers.upgrade?.toLowerCase() === 'websocket') {
-                    next();
-                    return;
-                }
-
-                // Skip root path with query string only (e.g. /ghost/?token=xxx for Vite HMR)
-                const urlPath = req.originalUrl?.split('?')[0];
-                if (urlPath === `${base}/` || urlPath === base) {
-                    next();
-                    return;
-                }
-
                 const match = req.originalUrl?.match(pathRegex);
 
                 if (match) {
--- a/apps/admin/vite-deep-links.ts
+++ b/apps/admin/vite-deep-links.ts
@@ -16,6 +16,19 @@ export function deepLinksPlugin(): Plugin {
 
         return () => {
             server.middlewares.use((req, res, next) => {
+                // Skip WebSocket upgrade requests (used for HMR)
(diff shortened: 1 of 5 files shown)


apps/admin/vite.config.ts
```diff
@@ -18,12 +18,7 @@ export default defineConfig({
     },
     server: {
         host: true,
-        allowedHosts: [
-            "localhost",
-            "127.0.0.1",
-            "::1",
-            "host.docker.internal",
-        ]
+        allowedHosts: true
```
jonatansberg, line 21: Any reason for changing this? I went for the more restrictive list based on the recommendations in the Vite docs. In practice, this should probably be fine as well.

apps/admin/vite-deep-links.ts
```diff
@@ -16,6 +16,19 @@ export function deepLinksPlugin(): Plugin {
...
+                    next();
+                    return;
+                }
+
+                // Skip root path with query string only (e.g. /ghost/?token=xxx for Vite HMR)
+                const urlPath = req.originalUrl?.split('?')[0];
+                if (urlPath === `${base}/` || urlPath === base) {
+                    next();
+                    return;
+                }
```
jonatansberg: We might need to handle the token case in the gateway rather than this middleware? I think the same goes for the WebSocket thing. If that has made it all the way here, then something's off, no? 🤔

jonatansberg (commented):
Overall, this looks really good. I have a couple of questions that would be relevant to go over before merging.

apps/admin/vite.config.ts
mike182uk: This was to allow support for external domains (i.e when funneling via tailscale)

apps/admin/vite-deep-links.ts
mike182uk: Heh this ain't needed at all 😅

I think I added it because I couldn't get the Caddy changes working at one point, but works fine without this so all good

sagzy (approved):

PR #25765, opened 2025-12-18
Refactored filtered_sessions to eliminate redundant scans

cmraible:
The filtered_sessions pipe previously returned only session IDs, which forced endpoints to JOIN back to _mv_hits to get the data they needed. This caused two full table scans per query, leading to timeouts on high-traffic sites.

Now filtered_sessions returns full hit rows with all filtering logic centralized in one place. Endpoints query it directly (for hit-level aggregations) or use a session_id IN subquery (for session-level aggregations). This eliminates the redundant scan.

### DIFF
--- a/ghost/core/core/server/data/tinybird/endpoints/api_kpis.pipe
+++ b/ghost/core/core/server/data/tinybird/endpoints/api_kpis.pipe
@@ -75,10 +75,31 @@ SQL >
             is_bounce,
             duration as session_sec
         from mv_session_data sd
-            inner join filtered_sessions fs
-                on fs.session_id = sd.session_id
         where
             site_uuid = {{ String(site_uuid, 'mock_site_uuid', description="Tenant ID", required=True) }}
+            -- Date filters on session start
+            {% if defined(date_from) %}
+                and first_pageview >= toDateTime({{ Date(date_from, description="Starting day for filtering a date range", required=False) }}, {{String(timezone, 'Etc/UTC', description="Site timezone", required=True)}})
+            {% else %}
+                and first_pageview >= toDateTime(dateAdd(day, -7, today()), {{String(timezone, 'Etc/UTC', description="Site timezone", required=True)}})
+            {% end %}
+            {% if defined(date_to) %}
+                and first_pageview < toDateTime({{ Date(date_to, description="Finishing day for filtering a date range", required=False) }}, {{String(timezone, 'Etc/UTC', description="Site timezone", required=True)}}) + interval 1 day
+            {% else %}
+                and first_pageview < toDateTime(dateAdd(day, 1, today()), {{String(timezone, 'Etc/UTC', description="Site timezone", required=True)}})
+            {% end %}
+            -- Session-level filters
+            {% if defined(source) %} and source = {{ String(source, description="Source to filter on", required=False) }} {% end %}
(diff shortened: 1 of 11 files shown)


ghost/core/core/server/data/tinybird/pipes/filtered_sessions.pipe
```diff
@@ -1,27 +1,51 @@
 TOKEN "axis" READ
```
cmraible, line 1: Before: apply hit-level and session-level filters, and return a unique list of session_ids. Consuming pipes then JOIN this with _mv_hits to get the hit data for the returned session_ids.

After: apply hit-level and session-level filters, and return all hit data matching the filters. Consuming pipes can then skip the JOIN to _mv_hits, because this returns all the required data already.

ghost/core/core/server/data/tinybird/endpoints/api_kpis.pipe
```diff
@@ -114,20 +135,12 @@ SQL >
                 {% end %}
                 count() pageviews
             from timeseries a
-            inner join _mv_hits h on
+            inner join filtered_sessions h on
```
troyciesco, line 138: nit: `h` doesn't make sense now. also do we need to change anything in this block to explicitly use the `filtered_sessions` table?

ghost/core/core/server/data/tinybird/endpoints/api_kpis.pipe
```diff
@@ -75,10 +75,31 @@ SQL >
...
+            {% if defined(device) %} and device = {{ String(device, description="Device type to filter on", required=False) }} {% end %}
+            {% if defined(utm_source) %} and utm_source = {{ String(utm_source, description="UTM source to filter on", required=False) }} {% end %}
+            {% if defined(utm_medium) %} and utm_medium = {{ String(utm_medium, description="UTM medium to filter on", required=False) }} {% end %}
+            {% if defined(utm_campaign) %} and utm_campaign = {{ String(utm_campaign, description="UTM campaign to filter on", required=False) }} {% end %}
+            {% if defined(utm_term) %} and utm_term = {{ String(utm_term, description="UTM term to filter on", required=False) }} {% end %}
+            {% if defined(utm_content) %} and utm_content = {{ String(utm_content, description="UTM content to filter on", required=False) }} {% end %}
+            -- Hit-level filters: use filtered_sessions for session lookup
+            {% if defined(member_status) or defined(location) or defined(pathname) or defined(post_uuid) %}
+            and session_id IN (select distinct session_id from filtered_sessions)
+            {% end %}
```
cmraible, line 102: This is all largely duplicated with filtered_sessions, which isn't super maintainable. Not 100% sure why this is all here vs just using the filtered_sessions pipe

cmraible:
This actually doesn't seem to have yielded the improvement I expected.

It seems to remove the additional scan from `top_locations`, but not `api_kpis` or `top_sources` 🤔

ghost/core/core/server/data/tinybird/endpoints/api_kpis.pipe
cmraible: good catch - although maybe it's actually `filtered_sessions` that is named incorrectly now, since it's returning hits instead of session_ids

PR #25773, opened 2025-12-27
✨ Added IndexNow integration for faster search engine indexing

betschki:
[IndexNow](https://www.indexnow.org) is a protocol that allows websites to notify search engines (Bing, Yandex, Seznam.cz, Naver) about content changes for faster indexing. This integration automatically pings IndexNow when posts are published or updated, reducing the time for content to appear in search results from days to minutes.

Ghost already integrates Ping-O-Matic, though the actual benefit of that service is rather limited. While IndexNow is not supported by Google, it supports a wider user base than Ping-O-Matic

The feature is behind the 'indexnow' labs flag in developer experiments. API keys are auto-generated on first use and the verification file is served dynamically at /{key}.txt.

---

- Why are you making it?
- What does it do?

---

> [!NOTE]
> Introduces IndexNow for faster indexing, gated by `labs.indexnow`.
> 
> - New `core/server/services/indexnow` listens to `post.published` and `post.published.edited`, pings `https://api.indexnow.org/indexnow` when SEO-relevant fields change; initialized in `boot.js`
> - Frontend middleware `serve-indexnow-key` serves the verification file at `/{key}.txt`; mounted in `site.js`
> - New setting `indexnow_api_key` (in default settings), allowed via settings input serializer and excluded from exports
> - Adds `indexnow` to labs flags and Admin Labs UI (with keyed list items)
> - Updates tests: unit tests for service behavior and middleware exposure; snapshot/config hash updated
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8650feb0e7b289b1564a5c9ee32706043f304861. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ghost/core/core/server/services/indexnow.js
```diff
@@ -0,0 +1,234 @@
...
+
+        // Save the generated key to settings
+        await models.Settings.edit([{
+            key: 'indexnow_api_key',
+            value: key
+        }], {context: {internal: true}});
+    }
+
+    return key;
+}
```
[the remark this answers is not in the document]
betschki: This sounds like too much of an edge case. The API key either exists from the beginning or when the lab feature is activated. Neither of which would likely coincide with the publishing of a post

ghost/core/core/frontend/web/middleware/serve-indexnow-key.js
```diff
@@ -0,0 +1,64 @@
...
+ * - Only responds to 32-character hex patterns (valid IndexNow key format)
+ * - Returns 404 (via next()) if IndexNow is disabled or no key configured
+ *
+ * Route collision note:
+ * The pattern /[a-f0-9]{32}.txt is unlikely to collide with user content
+ * since pages/posts don't have .txt extensions. The middleware runs after
+ * theme static assets, so legitimate .txt files in themes take precedence.
+ *
+ * @see ghost/core/core/server/services/indexnow.js - Main IndexNow service
+ */
```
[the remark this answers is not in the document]
betschki: Updated comment to say middleware runs BEFORE theme static assets

ghost/core/core/server/services/indexnow.js
```diff
@@ -0,0 +1,234 @@
...
+    const key = await getOrCreateApiKey();
+
+    // Get the site URL for the keyLocation parameter
+    const siteUrl = urlUtils.urlFor('home', true);
+
+    // Build the IndexNow request URL
+    const indexNowUrl = new URL(INDEXNOW_ENDPOINT);
+    indexNowUrl.searchParams.set('url', url);
+    indexNowUrl.searchParams.set('key', key);
+    indexNowUrl.searchParams.set('keyLocation', `${siteUrl}${key}.txt`);
```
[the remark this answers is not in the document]
betschki: Fixed. Now using urlUtils.urlJoin() instead of string concatenation

JohnONolan (approved):

ghost/core/core/server/data/migrations/versions/6.10/2025-12-27-11-17-40-add-indexnow-api-key-setting.js
```diff
@@ -0,0 +1,8 @@
+const {addSetting} = require('../../utils');
+
+module.exports = addSetting({
+    key: 'indexnow_api_key',
+    value: null,
+    type: 'string',
+    group: 'indexnow'
+});
```
sagzy: As database migrations have caused pain in the past, we generally prefer to create a separate PR/commit for database migrations, that contains minimal changes (example: https://github.com/TryGhost/Ghost/pull/15705/changes)

Would you mind moving the database changes to a separate PR, so that we can merge the database changes separately from the code changes?

ghost/core/core/server/services/indexnow.js
```diff
@@ -0,0 +1,251 @@
...
+};
+
+const defaultPostSlugs = [
+    'welcome',
+    'the-editor',
+    'using-tags',
+    'managing-users',
+    'private-sites',
+    'advanced-markdown',
+    'themes'
```
sagzy: `coming-soon` could be added here :)

ghost/core/core/server/services/indexnow.js
```diff
@@ -0,0 +1,251 @@
...
+                key: 'indexnow_api_key',
+                value: key,
+                type: 'string',
+                group: 'indexnow'
+            }).save(null, {context: {internal: true}});
+        }
+    }
+
+    return key;
+}
```
sagzy: I think we can get rid of `getOrCreateApiKey` logic here, and let the settings service populate the key on boot. Here's how it would work:
1. `settings-service.init()` is called on boot
2. It calls models.Settings.populateDefaults() 
3. For each default setting, it calls `defaultSetting.getDefaultValue()`

So if we add `indexnow_api_key: () => crypto.randomBytes(16).toString('hex')` here https://github.com/TryGhost/Ghost/blob/602976a0d3680a11e8b1a92561384ecd722cca09/ghost/core/core/server/models/settings.js#L63

We shouldn't be needing any logic to generate / edit the setting in this service

ghost/core/core/server/data/migrations/versions/6.10/2025-12-27-11-17-40-add-indexnow-api-key-setting.js
betschki: Makes perfect sense! Created #25808 for that

sagzy (approved):

ghost/core/core/server/services/indexnow.js
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        // Get the site URL for the keyLocation parameter
+        const siteUrl = urlUtils.urlFor('home', true);
+
+        // Build the IndexNow request URL
+        const indexNowUrl = new URL(INDEXNOW_ENDPOINT);
+        indexNowUrl.searchParams.set('url', url);
+        indexNowUrl.searchParams.set('key', key);
+        indexNowUrl.searchParams.set('keyLocation', urlUtils.urlJoin(siteUrl, `${key}.txt`));
```
vikaspotluri123, line 103: Is there a requirement for the key to be in the URL? e.g. would it make sense to use `.well-known/index_now_key.txt`, or make it part of the API? I think this would also simplify the router

e: Since we have to support sub-directories, we're using option 2 ([docs](https://www.indexnow.org/documentation)), which requires the API key to be hosted on the (sub-dir) root. I think we can use a static path for the key location, but it can't be nested.

muratcorlu: In case of serving Ghost from sub-directory, ```urlUtils.urlJoin(siteUrl, `${key}.txt`)``` will result a txt file url under that sub-directory as well, right?

vikaspotluri123: Yes, but we can use a static route rather than the key (e.g indexnow.txt)

muratcorlu: Ah ok. You mean simplifying the route definition. Good point.

PR #25783, opened 2026-01-05
✨ Added ability to ban members from commenting (computed approach)

rob-ghost:
## Summary

Adds the ability for administrators to ban members from commenting, with support for:
- **Permanent bans** - member cannot comment indefinitely
- **Temporary bans** - member cannot comment until a specific date
- **Ban history** - all bans are preserved for audit purposes

### Why this approach?

We evaluated two approaches for tracking ban status:

| Approach | Pros | Cons |
|----------|------|------|
| **Simple flag** (`can_comment` column on members) | Fast reads, simple | Temporary bans need scheduled jobs; no history |
| **Computed field** (separate ban table + query-time computation) | Auto-expiry via SQL; full history; no sync jobs | Slightly more complex queries |

We chose the **computed approach** because:

1. **No scheduled jobs** - Temporary bans auto-expire naturally. The SQL query checks `expires_at > NOW()` at query time, so when a ban's expiry date passes, the member can immediately comment again without any background process.

2. **Data consistency** - There's no flag to get out of sync. The `can_comment` value is always derived from the current state of the bans table.

3. **Full audit trail** - Every ban is preserved. When unbanning, we set `expires_at = NOW()` rather than deleting the record, so administrators can see the complete history of actions taken.

4. **Follows Ghost patterns** - Uses the same `countRelations()` pattern that Ghost uses for `count.likes`, `count.replies`, and other computed aggregates.

### API Endpoints

```
GET    /ghost/api/admin/members/:member_id/comment-bans     # List ban history
POST   /ghost/api/admin/members/:member_id/comment-bans     # Create ban
DELETE /ghost/api/admin/members/:member_id/comment-bans/:id # Remove specific ban
```

### Schema

New table `members_comment_bans`:
- `id`, `member_id` - identifiers
- `expires_at` - null for permanent, timestamp for temporary
- `reason` - optional text explanation
- `created_at`, `created_by` - audit fields
- `updated_at`, `updated_by` - for tracking unban actions


rob-ghost:
Closing in favour of https://github.com/TryGhost/Ghost/pull/25791

PR #25800, opened 2026-01-07
🐛 Fixed attribution for email-only posts

9larsons:
ref https://linear.app/ghost/issue/NY-878/

When a free member clicks a link to view an email-only post, it takes them to a url like `/email/{uuid}`. We append the appropriate attribution to the url via query params to create the `ghost-history` entry for attributing to the email, however, this won't work unless the post is also published.

We have a gap where the attribution lookup via uuid is only finding published posts.

This PR allows us to find published and sent posts, so that we correctly attribute the url when directly navigating to an email-only post (which in general is 'unlisted').

---

> [!NOTE]
> Ensures email-only posts (status `sent`) are correctly resolved for attribution.
> 
> - Updates `getResourceById('post')` in `url-translator.js` to query with `filter: 'status:[published,sent]'`
> - Refactors page branch to return `page || null` for consistency
> - Extends unit tests: mocks post status based on id and adds a test validating `sent` posts are returned
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ae3107a14c5cae7339a82eb51b25af4462a97dc9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ghost/core/test/unit/server/services/member-attribution/url-translator.test.js
```diff
@@ -4,10 +4,14 @@ const UrlTranslator = require('../../../../../core/server/services/member-attrib
 
 const models = {
     Post: {
-        findOne({id}) {
+        findOne({id}, options = {}) {
             if (id === 'invalid') {
                 return null;
             }
+            // Simulate real Post model behavior: without proper filter, only published posts are found
+            if (id === 'sent-post' && options.filter !== 'status:[published,sent]') {
```
troyciesco: i'm slightly confused by the comment - i know this isn't the use case in the code, but is a `status:[sent]` filter valid? or does `published` always need to be included for some reason?

could be something you end up cleaning up in your follow up tests anyway, just calling out that if you keep it we might want to reword to say the UrlTranslator uses this particular filter or something to that effect

ghost/core/core/server/services/member-attribution/UrlTranslator.js
```diff
@@ -132,7 +132,11 @@ class UrlTranslator {
         switch (type) {
         case 'post':
         case 'page': {
-            const post = await this.models.Post.findOne({id}, {require: false});
+            // Include sent status for email-only posts, and specify type to override default type:post filter
+            const post = await this.models.Post.findOne({id}, {
+                require: false,
+                filter: `status:[published,sent]+type:${type}`
```
cmraible: suggestion(non-blocking): rather than adding the `+type:${type}`, what would you think about splitting up the `case`s in the switch statement:

`case 'post'`: use this filtering logic, without the added `+type:${type}` filter
`case 'page'`: use the current/previous implementation without specifying the filters at all

I don't feel strongly about it, just throwing an idea out there.

9larsons: Yeah I like that. Done.

PR #25812, opened 2026-01-08
added url transform to default email content

troyciesco:
added url transform to default email content


ghost/core/test/integration/services/member-welcome-emails.test.js
```diff
@@ -330,6 +330,28 @@ describe('Member Welcome Emails Integration', function () {
             const entriesAfterJob = await models.Outbox.findAll();
             assert.equal(entriesAfterJob.length, 0);
         });
+
+        it('transforms __GHOST_URL__ placeholders to actual site URL in default templates', async function () {
```
troyciesco, line 334: i was about to delete this test because it's only for the default template but i see we have some other default template related ones, so i'll keep it even tho it's temporary

PR #25816, opened 2026-01-08
Skipped creating offers for invalid Stripe coupons during migrations

sagzy:
ref https://linear.app/ghost/issue/BER-3162

During migrations, some Stripe coupons from other platforms may not compatible with Ghost offers. For example, a `yearly` coupon set to `repeating` (Ghost only accepts `once` or `forever` for `yearly` coupons).

When we attempt to create a Ghost offer based on an incompatible Stripe coupon, we can either:
    1. Surface the offer validation errors during migrations, so that they can be resolved case-by-case, until all invalid Stripe coupons have been removed from paid members
    2. Skip creating the offer but still create the paid member

The migration team has opted for solution 2, in particular in the context of self-migrations where resolving those errors may be difficult.

As we discover more incompatibilities with Stripe coupons from other platforms, we can expand the list of error codes here: https://github.com/TryGhost/Ghost/blob/43c984792b5a03830a131fc8ba27947ffa2cdd67/ghost/core/core/server/services/members/members-api/repositories/member-repository.js#L1057

### DIFF
--- a/ghost/core/core/server/services/members/members-api/repositories/member-repository.js
+++ b/ghost/core/core/server/services/members/members-api/repositories/member-repository.js
@@ -1054,14 +1054,11 @@ module.exports = class MemberRepository {
                 );
                 offerId = offer.id;
             } catch (e) {
-                if (e instanceof errors.ValidationError) {
-                    // During migrations, some Stripe coupons from other platforms may not compatible with Ghost offers.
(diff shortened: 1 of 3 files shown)


PaulAdamDavis (approved):
LGTM! :shipit:

PR #25822, opened 2026-01-08
🐛 Fixed sorting in Growth tab Sources table

troyciesco:
closes https://linear.app/ghost/issue/NY-696

- Sorting on the Growth tab's Sources table was not working due to a parameter name mismatch. The API was sending `order`, but the backend service was expecting `orderBy`. This meant the sort parameter was never correctly applied, and data always appeared in the default order
- This PR fixes the parameter destructuring in `ghost/core/core/server/services/stats/referrers-stats-service.js` to correctly map the incoming `order` parameter to the internal `orderBy` variable.

-----

source-sort


cmraible (approved):
Nice one, works for me 👌

PR #25836, opened 2026-01-12
Updated member welcome email token replacement

troyciesco:
closes https://linear.app/ghost/issue/NY-837

- Replaced Handlebars template syntax ({{member.name}}) with simpler brace syntax ({name}, {first_name}, {email}, {site_title}, {site_url})
- Added support for custom fallback values (e.g., {first_name, "friend"} renders "friend" when name is missing)
- Removed hardcoded "there" fallback

### DIFF
--- a/e2e/data-factory/factories/automated-email-factory.ts
+++ b/e2e/data-factory/factories/automated-email-factory.ts
@@ -26,7 +26,7 @@ export class AutomatedEmailFactory extends Factory<Partial<AutomatedEmail>, Auto
             status: 'active',
             name: 'Welcome Email (Free)',
             slug: 'member-welcome-email-free',
-            subject: 'Welcome to {{site.title}}!',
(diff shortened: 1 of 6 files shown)


cmraible (changes requested):
Nice one, I see what you were getting at now. Mostly looks good to go from my POV, just a few small things:

1. Coderabbit's comments, especially the one re: escaping seem worth looking into
2. There are a few places where we still use the old `{{}}` handlebars syntax in other files, which we should update with this:
- In the default subject line - this would break: ghost/core/core/server/services/member-welcome-emails/constants.js
- In some of our integration tests (less important but might as well update for consistency):
   - ghost/core/test/integration/services/member-welcome-emails.test.js
   - ghost/core/test/integration/jobs/process-outbox.test.js

cmraible (approved):
LGTM!

PR #25839, opened 2026-01-12
SPIKE: Added welcome emails to member events

troyciesco:
ref https://linear.app/ghost/issue/NY-889

spike of how we might show welcome email sends in the member activity feed

### Notes

**`automated_email_recipients` table pattern**: The `member_id` column intentionally does not have a foreign key constraint, following the same pattern as `email_recipients`. This is because:
1. Members can be deleted, but email send history should be preserved for audit/analytics
2. The denormalized fields (`member_uuid`, `member_email`, `member_name`) capture the member's state at send time

See commit 76c1b60a4d for the original rationale.

TK


troyciesco:
closing bc we did this in separate PRs

PR #25842, opened 2026-01-13
Add admin deploy lock/toggle

jonatansberg:
ref https://linear.app/ghost/issue/BER-3160

This should let us disable the automated admin releases if needed. I've gone ahead and created the variable as well.


rob-ghost:
Closing as the workflows this PR targets are now gone and we need to approach this from another angle.

PR #25853, opened 2026-01-14
Used comment permalinks in notification emails

kevinansfield:
ref https://linear.app/ghost/issue/BER-3183/

- removes hardcoded `#ghost-comments-root` from email templates in favor of including that in the `postUrl` data that's passed through
- added conditional `postUrl` generation that uses `#ghost-comments-root` or the specific comment permalink depending on the `commentPermalinks` labs flag
- leaves "View comments" wording in place for now rather than switching to "View comment" to avoid i18n locale churn


rob-ghost (approved):
Looks good 👍🏻

PR #25854, opened 2026-01-14
Deferred Featurebase load+initialization until interaction

kevinansfield:
ref https://linear.app/ghost/issue/FEA-509/

We shouldn't need to pay the network and CPU cost of loading the Featurebase script until we actually need it to display the Feedback widget.

- Delayed Featurebase SDK loading and JWT token fetching until the user interacts with the Feedback button
- Added preloading on hover/focus to hide latency while avoiding unnecessary page load overhead
- Clicks will wait for all preloading to finish (starting if not already started) before opening the feedback modal

### DIFF
--- a/apps/admin/src/layout/app-sidebar/hooks/use-featurebase.ts
+++ b/apps/admin/src/layout/app-sidebar/hooks/use-featurebase.ts
@@ -91,11 +91,13 @@ export function useFeaturebase(): Featurebase {
                 defaultBoard: DEFAULT_BOARD,
                 featurebaseJwt: token
             }, (err) => {
-                // Only gate actions on first init - re-inits (e.g. theme/token changes) are
-                // best-effort. Resolving/rejecting an already-settled promise is a no-op.
                 if (err) {
                     console.error('[Featurebase] Failed to initialize widget:', err);
                     deferredInitRef.current.reject(err);
+
+                    // reset so we can retry on next interaction
+                    deferredInitRef.current = deferred();
+                    setShouldLoad(false);
                 } else {
                     deferredInitRef.current.resolve();
                 }
--- a/apps/admin/src/layout/app-sidebar/hooks/use-featurebase.ts
+++ b/apps/admin/src/layout/app-sidebar/hooks/use-featurebase.ts
@@ -1,4 +1,4 @@
-import {useCallback, useEffect, useRef} from 'react';
+import {useCallback, useEffect, useRef, useState} from 'react';
 import {getFeaturebaseToken} from '@tryghost/admin-x-framework';
 import {useBrowseConfig} from '@tryghost/admin-x-framework/api/config';
 import {useCurrentUser} from '@tryghost/admin-x-framework/api/current-user';
@@ -46,6 +46,7 @@ function loadFeaturebaseSDK(): Promise<void> {
 
 interface Featurebase {
     openFeedbackWidget: (options?: {board?: string}) => void;
+    preloadFeedbackWidget: () => void;
 }
 
 export function useFeaturebase(): Featurebase {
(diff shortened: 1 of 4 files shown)


apps/admin/src/layout/app-sidebar/hooks/use-featurebase.ts
```diff
@@ -1,23 +1,28 @@
-import {useCallback, useEffect, useRef} from 'react';
+import {useCallback, useEffect, useRef, useState} from 'react';
 import {getFeaturebaseToken} from '@tryghost/admin-x-framework';
 import {useBrowseConfig} from '@tryghost/admin-x-framework/api/config';
-import {useCurrentUser} from '@tryghost/admin-x-framework/api/current-user';
 import {useFeatureFlag} from '@/hooks/use-feature-flag';
 import {useUserPreferences} from '@/hooks/user-preferences';
+import {deferred} from '@/utils/deferred';
+import type {Deferred} from '@/utils/deferred';
```
jonatansberg: Nit: this could be a single import

apps/admin/src/layout/app-sidebar/hooks/use-featurebase.ts
```diff
@@ -1,23 +1,28 @@
...
 }
 
 const SDK_URL = 'https://do.featurebase.app/js/sdk.js';
 
+let featurebaseSDKPromise: Promise<void> | null = null;
 function loadFeaturebaseSDK(): Promise<void> {
-    return new Promise((resolve, reject) => {
+    if (featurebaseSDKPromise) {
+        return featurebaseSDKPromise;
+    }
```
jonatansberg: Nit: maybe this would be slightly cleaner if we, instead of doing the early return, did the assignment in the if statement and then always return the promise at the end 🤔

apps/admin/src/layout/app-sidebar/hooks/use-featurebase.ts
```diff
@@ -34,83 +39,96 @@ function loadFeaturebaseSDK(): Promise<void> {
...
     const {data: tokenData} = getFeaturebaseToken({
-        enabled: featurebaseEnabled
+        enabled: featurebaseEnabled && shouldLoad
     });
     const token = tokenData?.featurebase?.token;
 
     useEffect(() => {
-        if (!featurebaseEnabled || !featurebaseOrg || !currentUser || !token) {
-            return;
+        if (shouldLoad ) {
```
jonatansberg: Nit: ` )`

apps/admin/src/layout/app-sidebar/hooks/use-featurebase.ts
```diff
@@ -34,83 +39,96 @@ function loadFeaturebaseSDK(): Promise<void> {
...
-        window.postMessage({
-            target: 'FeaturebaseWidget',
-            data: {
-                action: 'openFeedbackWidget',
-                ...(options?.board && {setBoard: options.board})
-            }
-        }, '*');
+        setShouldLoad(true);
+
+        void initPromise.current.promise.then(() => {
```
jonatansberg: Nit: maybe we can give the ref a better name? `initPromise.current.promise.then` looks a bit silly.

apps/admin/src/utils/deferred.ts
```diff
@@ -0,0 +1,10 @@
+export const deferred = <T>() => {
```
jonatansberg: Maybe we can generate some unit tests for this for good measure?

jonatansberg (approved):
It feels a bit silly to review code that I wrote a lot of myself, even more so when almost all my comments are nitpicks on things I did 😂 

Maybe we can clean up some of them? Otherwise, I do think this is more or less ready to go.

apps/admin/src/layout/app-sidebar/hooks/use-featurebase.ts
```diff
@@ -1,116 +1,141 @@
...
-                    initializedWithRef.current = {theme, token};
+                    deferredInitRef.current.resolve();
                 }
             });
-        }).catch(() => {
-            isInitializingRef.current = false;
-            initializedWithRef.current = null;
         });
-    }, [featurebaseEnabled, featurebaseOrg, currentUser, token, theme]);
+    }, [organization, theme, token, shouldLoad]);
```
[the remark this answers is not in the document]
kevinansfield: The reset is immediately undone by the next user interaction. Both openFeedbackWidget and preloadFeedbackWidget call setShouldLoad(true), so any hover or click on the button re-triggers init automatically. There's no scenario where the user is actually blocked.

PR #25855, opened 2026-01-14
Re-enabled the nx daemon in `yarn dev` script

cmraible:
no ref

## Problem

Following the official install from source instructions from Ghost's docs, using `yarn dev`, currently fails with an NX error message:

`NX   Daemon process terminated and closed the connection`

The nx daemon is required when running `yarn dev`, because some of the sub-commands use `nx watch -- ..` to rebuild apps on file changes, and `nx watch` requires the daemon to be running.

## Investigation

Suspect commit: ac366a3c290e8447510108c326a24e8ab7d2edd7

Before this commit, which upgraded to NX 22, the `yarn dev` script explicitly started the nx daemon. This was removed, so now the NX daemon doesn't run by default. It can be started manually with `yarn nx daemon --start`, which fixes the NX Daemon errors.

## Solution

Add back the explicit `yarn nx reset` and `yarn nx daemon --start` commands to the `yarn dev` script, so these run automatically when running `yarn dev`. This should guarantee the NX daemon is running whenever `yarn dev` is run, thus avoiding the `NX Daemon process terminated and closed the connection` errors.

---

> [!NOTE]
> Ensures `yarn dev` reliably runs with the Nx daemon.
> 
> - In `.github/scripts/dev.js`, adds `yarn nx reset --onlyDaemon` and `yarn nx daemon --start` before launching concurrent dev processes
> - Aims to prevent `NX Daemon process terminated and closed the connection` errors when using `nx watch`-based tasks
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5532df56732bdcb7803e81a7c2e6a609a1a66124. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


cmraible:
@jonatansberg added you as a reviewer since you originally removed these bits from `dev.js` — mostly as an FYI in case you had other reasons for removing them that I'm overlooking 😃

.github/workflows/ci.yml
```diff
@@ -221,6 +221,60 @@ jobs:
       node_version: ${{ env.NODE_VERSION }}
       node_test_matrix: ${{ steps.node_matrix.outputs.matrix }}
 
+  job_development-environment:
+    name: Development Environment
+    runs-on: ubuntu-latest
+    needs: [job_setup]
+    if: needs.job_setup.outputs.changed_any_code == 'true'
```
troyciesco: oh sick, so this won't run every commit?

.github/workflows/ci.yml
```diff
@@ -221,6 +221,60 @@ jobs:
...
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        env:
+          FORCE_COLOR: 0
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      # Step 3: Run yarn setup (as documented)
+      # Uses -y flag to auto-accept prompts in CI
+      - name: Run yarn setup
```
troyciesco: this creates 1000 members by default, there's a way to set it lower isn't there? maybe that wouldn't save that much though.

.github/workflows/ci.yml
cmraible: yep, if you just update i.e. a markdown file, this will be skipped

.github/workflows/ci.yml
cmraible: yeah I think it's pretty quick — would prefer to keep it as close to identical to the docs as possible, in case the member generation or something fails for whatever reason

troyciesco (approved):

jonatansberg:
NX should start the daemon automatically. The fact that it's not running might be related to some other issue in our setup. Forcefully resetting and restarting the daemon like this does help to some extent, but it also messes up any other NX commands that you might have running at the same time. I removed the restart as it supposedly shouldn't be needed with the version we're on. 

Zooming out a birt furrther, I'd really want to move away from using the daemon altogether. We're seeing odd issues with the daemon hanging if you switch branches or similar that are really unintuitive to get out of. 

I think we should be able to swap the NX watch usage for the built-in watch support in Vite and that would let us get off the daemon altogether.

cmraible:
> Zooming out a birt furrther, I'd really want to move away from using the daemon altogether.

I am 100% with you this, the daemon creates way more issues than it solves in my experience. This feels like the best next step, rather than continuing to try to debug why we seem to need this force restart when we theoretically shouldn't. Then we can just disable the daemon outright, like you wanted to do [here.](https://github.com/TryGhost/Ghost/commit/555192031123d486d53d20e0ff4d530e86dccf0d)

PR #25858, opened 2026-01-15
🐛 Fixed reply form showing parent comment author's details

rob-ghost:
## Summary

When opening a reply form on a comment, the form header was incorrectly displaying the parent comment author's name and expertise instead of the current user's data. This happened because the code used fallback patterns that pulled values from the comment being replied to when the current member had no name or expertise set.

The fix removes these fallbacks so the reply form only shows the logged-in user's own details (or the appropriate placeholder if they have none).

ref https://linear.app/ghost/issue/BER-3212

### DIFF
--- a/apps/comments-ui/src/components/content/forms/form.tsx
+++ b/apps/comments-ui/src/components/content/forms/form.tsx
@@ -241,7 +241,7 @@ const Form: React.FC<FormProps> = ({
     const [progress, setProgress] = useState<Progress>('default');
     const formEl = useRef(null);
 
-    const memberName = member?.name ?? comment?.member?.name;
+    const memberName = member?.name;
 
     if (progress === 'sending' || (memberName && isAskingDetails)) {
         // Force open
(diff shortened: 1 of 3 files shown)


apps/comments-ui/src/components/content/forms/form.tsx
```diff
@@ -308,7 +308,7 @@ const FormWrapper: React.FC<FormWrapperProps> = ({
     const {member, dispatchAction} = useAppContext();
 
     const memberName = member?.name ?? comment?.member?.name;
```
kevinansfield: Member names are also optional at a data level so this has the same fallback problem for the reply form.

Comments-UI itself does enforce a name before commenting which is handled via a popup when clicking on the primary comment form, however for reply forms it just looks like it's broken because the input box is disabled with no indication why. That's a separate issue though.

kevinansfield (approved):
Left a comment about a similar fallback issue with member name, although that's a little more of an edge-case. Otherwise looks good 👌

apps/comments-ui/src/components/content/forms/form.tsx
rob-ghost: Nice! I've amended the PR to be more generally about showing the parents details when the member details are missing, please give it another look.

kevinansfield (approved):

PR #25859, opened 2026-01-15
Changed comment moderation 'View post' action to use comment permalinks

rob-ghost:
## Summary

- The "View post" action in the comment moderation dropdown now navigates directly to the comment on the post using the comment permalink
- Renamed the action to "View on post" to better describe this behaviour 

ref https://linear.app/ghost/issue/BER-3200

### DIFF
--- a/apps/posts/src/views/comments/comments.tsx
+++ b/apps/posts/src/views/comments/comments.tsx
@@ -6,11 +6,14 @@ import CommentsList from './components/comments-list';
 import React, {useCallback} from 'react';
 import {Button, EmptyIndicator, LoadingIndicator, LucideIcon, createFilter} from '@tryghost/shade';
 import {useBrowseComments} from '@tryghost/admin-x-framework/api/comments';
+import {useBrowseConfig} from '@tryghost/admin-x-framework/api/config';
 import {useFilterState} from './hooks/use-filter-state';
 import {useKnownFilterValues} from './hooks/use-known-filter-values';
 
 const Comments: React.FC = () => {
     const {filters, nql, setFilters} = useFilterState();
+    const {data: configData} = useBrowseConfig();
+    const commentPermalinksEnabled = configData?.config?.labs?.commentPermalinks === true;
 
     const handleAddFilter = useCallback((field: string, value: string, operator: string = 'is') => {
         setFilters((prevFilters) => {
@@ -72,6 +75,7 @@ const Comments: React.FC = () => {
                     </div>
                 ) : (
                     <CommentsList
+                        commentPermalinksEnabled={commentPermalinksEnabled}
                         fetchNextPage={fetchNextPage}
                         hasNextPage={hasNextPage}
                         isFetchingNextPage={isFetchingNextPage}
--- a/apps/posts/src/views/comments/components/comments-list.tsx
+++ b/apps/posts/src/views/comments/components/comments-list.tsx
@@ -133,7 +133,8 @@ function CommentsList({
     isFetchingNextPage,
     fetchNextPage,
     onAddFilter,
-    isLoading
+    isLoading,
+    commentPermalinksEnabled
 }: {
     items: Comment[];
     totalItems: number;
@@ -142,6 +143,7 @@ function CommentsList({
     fetchNextPage: () => void;
     onAddFilter: (field: string, value: string, operator?: string) => void;
     isLoading?: boolean;
+    commentPermalinksEnabled?: boolean;
 }) {
     const parentRef = useRef<HTMLDivElement>(null);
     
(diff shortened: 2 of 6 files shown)


apps/posts/src/views/comments/components/comments-list.tsx
```diff
@@ -340,14 +340,12 @@ function CommentsList({
...
-                                                            View post
-                                                            </a>
-                                                        </DropdownMenuItem>
-                                                    )}
+                                                    <DropdownMenuItem asChild>
+                                                        <a href={`${item.post!.url}#ghost-comments-${item.id}`} rel="noopener noreferrer" target="_blank">
+                                                            <LucideIcon.ExternalLink className="mr-2 size-4" />
+                                                        View on post
+                                                        </a>
+                                                    </DropdownMenuItem>
```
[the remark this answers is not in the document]
rob-ghost: The API includes the `post` relation, and there's a cascading delete on comments when a post is deleted, so in practice, in this code path, its not possible for `item.post` to be `null`.

`item.post` is optional because there's code paths which don't include the `post` relation, and therefore it can be `null`.

This could be solved by having a sub-type for the comment which forces `post` but I don't see the value over ensuring it with `!`

apps/posts/src/views/comments/comments.tsx
```diff
@@ -6,11 +6,14 @@ import CommentsList from './components/comments-list';
...
 import {Button, EmptyIndicator, LoadingIndicator, LucideIcon, createFilter} from '@tryghost/shade';
 import {useBrowseComments} from '@tryghost/admin-x-framework/api/comments';
+import {useBrowseConfig} from '@tryghost/admin-x-framework/api/config';
 import {useFilterState} from './hooks/use-filter-state';
 import {useKnownFilterValues} from './hooks/use-known-filter-values';
 
 const Comments: React.FC = () => {
     const {filters, nql, setFilters} = useFilterState();
+    const {data: configData} = useBrowseConfig();
+    const commentPermalinksEnabled = configData?.config?.labs?.commentPermalinks === true;
```
jonatansberg, line 16: Don't we have a feature flag hook somewhere?

e2e/tests/admin/comments/comment-moderation.test.ts
```diff
@@ -0,0 +1,80 @@
...
+            member_id: member.id,
+            html: '<p>Test comment with permalink</p>'
+        });
+
+        await settingsService.updateLabsSettings({commentPermalinks: true});
+
+        const commentsPage = new CommentsPage(page);
+        await commentsPage.goto();
+        // Reload to ensure config changes are picked up by react-query
+        await page.reload();
```
jonatansberg: I don't think you need the reload here; the goto should already do a reload.

e2e/tests/admin/comments/comment-moderation.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        const commentRow = commentsPage.getCommentRowByText('Test comment without permalink');
+        await commentsPage.openMoreMenu(commentRow);
+
+        const viewPostMenuItem = commentsPage.getViewPostMenuItem();
+        await expect(viewPostMenuItem).toBeVisible();
+
+        const href = await viewPostMenuItem.getAttribute('href');
+        expect(href).toContain(post.slug);
+        expect(href).not.toContain('#ghost-comments-');
```
jonatansberg: Do we need a negative test for this? I think the comment permalinks should just be the way this works. Do we even need a feature flag?

jonatansberg (commented):
The code looks fine. The main thing I'm wondering about is the feature flagging and the negative tests. Do we need that?

apps/posts/src/views/comments/comments.tsx
rob-ghost: I thought the same, and apparently not. There's no `useLabs` or equivalent I could find.

Would you like to add something like this in this PR?

e2e/tests/admin/comments/comment-moderation.test.ts
rob-ghost: Permalinks are behind a labs flag atm. So we can't link directly to a comment if the flag isn't turned on.

So the tests are for:

1. `View post` when permalink flag is off and links to the post itself
2. `View on post` when permalink flag is on which links directly to the comment

Given the flag exists, I think we should err on the side of honouring it. This test is then to check that the conditional which handles which menu item is rendered is present and working.

e2e/tests/admin/comments/comment-moderation.test.ts
rob-ghost: We do need the `reload` because `goto` does a SPA navigation so it doesn't refresh the config.

However, I found we have a `labs` test fixture that we can use for the tests so we don't have to manually set them and reload.

e2e/tests/admin/comments/comment-moderation.test.ts
```diff
@@ -0,0 +1,79 @@
...
+            await commentsPage.waitForComments();
+
+            const commentRow = commentsPage.getCommentRowByText('Test comment without permalink');
+            await commentsPage.openMoreMenu(commentRow);
+
+            const viewPostMenuItem = commentsPage.getViewPostMenuItem();
+            await expect(viewPostMenuItem).toBeVisible();
+
+            const href = await viewPostMenuItem.getAttribute('href');
+            expect(href).toContain(post.slug);
```
ibalosh: I am wondering if it would be better to actually visit the link in both tests rather then check the shape based on factory. With maybe something along the lines `commentsPage.getCommentRowByText('blah').visitPostThatOwnsTheComment()`  and then assert page visited details

It makes it more e2e-ish?

ibalosh (changes requested):

e2e/tests/admin/comments/comment-moderation.test.ts
rob-ghost: Yeah I can see the value in this approach.

I've changed the test to click the action and assert that the new page that pops up has the location we expect.

ibalosh (approved):

PR #25862, opened 2026-01-15
Added billing portal configuration persistence

sam-lord:
ref [FEA-502](https://linear.app/ghost/issue/FEA-502/portal-billing-management)

### DIFF
--- a/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
+++ b/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
@@ -240,8 +240,11 @@ module.exports = class RouterController {
             customer = await this._stripeAPIService.getCustomer(subscription.get('customer_id'));
         }
 
+        const configurationId = this._settingsCache.get('stripe_billing_portal_configuration_id');
+
         const session = await this._stripeAPIService.createBillingPortalSession(customer, {
-            returnUrl: req.body.returnUrl
+            returnUrl: req.body.returnUrl,
+            ...(configurationId && {configurationId})
         });
         const sessionInfo = {
             url: session.url
--- a/ghost/core/core/server/services/stripe/billing-portal-manager.js
+++ b/ghost/core/core/server/services/stripe/billing-portal-manager.js
@@ -109,7 +109,8 @@ class BillingPortalManager {
 
         if (updateOnly) {
             return {
-                business_profile: defaultOptions.business_profile
+                features: defaultOptions.features,
+                default_return_url: this.siteUrl
             };
         } else {
             return defaultOptions;
--- a/ghost/core/core/server/services/stripe/billing-portal-manager.js
+++ b/ghost/core/core/server/services/stripe/billing-portal-manager.js
@@ -1,0 +1,122 @@
+const CONFIGURATION_ID_SETTING = 'stripe_billing_portal_configuration_id';
+
+class BillingPortalManager {
+    /** @type {object} */
+    SettingsModel;
+    /** @type {object} */
+    settingsCache;
+    /** @type {object} */
+    api;
+    /** @type {string|null} */
+    siteUrl = null;
+    /** @type {boolean} */
+    configured = false;
+
+    /**
+     * @param {object} deps
+     * @param {object} deps.api
+     * @param {object} deps.models
+     * @param {object} deps.models.Settings
+     * @param {object} deps.settingsCache
+     */
+    constructor({api, models, settingsCache}) {
+        this.SettingsModel = models.Settings;
+        this.settingsCache = settingsCache;
+        this.api = api;
+        this.configured = false;
+    }
+
+    /**
+     * Configures the billing portal manager, passing in additional dependencies
+     * in a different stage of the Stripe service lifecycle.
+     * @param {object} config
+     * @param {string} config.siteUrl
+     */
+    configure(config) {
+        this.siteUrl = config.siteUrl;
+        this.configured = true;
+    }
+
+    /**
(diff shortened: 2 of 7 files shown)


ghost/core/core/server/services/stripe/stripe-api.js
```diff
@@ -933,4 +933,20 @@ module.exports = class StripeAPI {
...
+     * @param {string} [options.business_profile.headline]
+     * @param {object} options.features
+     * @param {string} options.default_return_url
+     *
+     * @returns {Promise<import('stripe').Stripe.BillingPortal.Configuration>}
+     */
+    async createBillingPortalConfiguration(options) {
+        await this._rateLimitBucket.throttle();
+        return await this._stripe.billingPortal.configurations.create(options);
+    }
```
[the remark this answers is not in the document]
sam-lord: I'm not sure this is true -- we're able to creating billing portal configurations. I've confirmed this experimentally at least.

We probably should be concerned with upgrading at some point.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -233,8 +233,11 @@ module.exports = class RouterController {
             customer = await this._stripeAPIService.getCustomer(subscription.get('customer_id'));
         }
 
+        const configurationId = this._settingsCache.get('stripe_billing_portal_configuration_id');
+
         const session = await this._stripeAPIService.createBillingPortalSession(customer, {
-            returnUrl: req.body.returnUrl
+            returnUrl: req.body.returnUrl,
+            configurationId: configurationId
```
sam-lord: Need to think about what happens when this is falsey 🤔

ghost/core/core/server/services/stripe/billing-portal-manager.js
sam-lord, line 1: Should be typescript & needs respective test file

ghost/core/core/server/services/stripe/billing-portal-manager.js
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Configures the billing portal manager, passing in additional dependencies
+     * in a different stage of the Stripe service lifecycle.
+     *
+     * @param {BillingPortalConfig} config
+     */
+    configure(config) {
+        this.siteUrl = config.siteUrl;
+        this.configured = true;
```
sam-lord, line 49: Should probably be explicitly initialized in the constructor.

ghost/core/core/server/services/stripe/billing-portal-manager.js
sam-lord: I switched it to typescript, but it broke because this file is included in a bree job. Will need to be very deliberate about use of TypeScript in Ghost -- was fortunate that this was covered by tests this time. Would be an easy way to introduce a runtime error that only occurs if the file is run within a job. Unit test coverage is not sufficient to confirm this.

ghost/core/core/server/data/migrations/versions/6.13/2026-01-15-10-41-19-add-billing-portal-setting.js
```diff
@@ -0,0 +1,8 @@
+const {addSetting} = require('../../utils');
+
+module.exports = addSetting({
```
aileen, line 3: Migrations = separate PR?

ghost/core/core/server/services/stripe/billing-portal-manager.js
```diff
@@ -0,0 +1,122 @@
...
+    configured = false;
+
+    /**
+     * @param {object} deps
+     * @param {object} deps.api
+     * @param {object} deps.models
+     * @param {object} deps.models.Settings
+     * @param {object} deps.settingsCache
+     */
+    constructor({api, models, settingsCache}) {
```
aileen, line 34: What's our normal pattern here? Do we do guard clauses for deps?

ghost/core/core/server/services/stripe/billing-portal-manager.js
```diff
@@ -0,0 +1,122 @@
...
+     * Get the configuration options for the Stripe Billing Portal.
+     * @param {boolean} [updateOnly=false]
+     * @returns {object}
+     */
+    getConfigurationOptions(updateOnly = false) {
+        const defaultOptions = {
+            business_profile: {
+                headline: `Manage your ${this.settingsCache.get('title')} subscription`
+            },
+            features: {
```
aileen: Could the features be a const? Easy to see the default settings

ghost/core/core/server/services/stripe/billing-portal-manager.js
```diff
@@ -0,0 +1,122 @@
...
+                subscription_cancel: {
+                    enabled: true
+                }
+            },
+            default_return_url: this.siteUrl
+        };
+
+        if (updateOnly) {
+            return {
+                business_profile: defaultOptions.business_profile
```
aileen: I wonder if we need that? It might be important to always ensure the default features are enabled. If a publisher disables a feature (by accident), but we rely (and the customer is lead to believe so by the UI) that those features are available, we might be in a pickle

aileen (changes requested):
Nitpick comment, but I think we have to enforce the default features, even on update

ghost/core/core/server/services/stripe/billing-portal-manager.js
sam-lord: Not typically, especially in cases where it's just dependency injection.

aileen (approved):
Great!! This LGMT 👍

PR #25864, opened 2026-01-15
✨ Released automatic domain warm up

sam-lord:
ref [GVA-640](https://linear.app/ghost/issue/GVA-640/remove-labs-flag-for-domain-warming)

Removing the labs flag makes this generally available, configurable with the `fallbackAddress` and `fallbackDomain` options (which are required to be able to warm-up the primary domain).


aileen (approved):
👍

PR #25866, opened 2026-01-15
Added deep-link by `id` support to comments list

kevinansfield:
ref https://linear.app/ghost/issue/BER-3199/

We want to allow linking to a specific comment for moderation from the front-end and in report notifications.

- added URL filtering support for `/comments/?id=is:{comment_id}` to `useFilterState`
  - no UI is added for manually selecting a specific id
  - adds an `isSingleIdFilter` property to the return object, being `true` when a deep-link id filter is present so we can adapt the filtering UI accordingly
  - adds an options param to the returned `clearFilters` function to support URL history behaviour
- modified comments list for deep-link by `id` state
  - hides filtering UI because it doesn't make sense when there's no display of or option to change an `id` filter
  - adds "Show all comments" button underneath the list showing a single comment, uses newly added `{replace: false}` option so it's possible to get back to the single comment using the Back button after showing all

### DIFF
--- a/apps/posts/src/views/comments/hooks/use-filter-state.ts
+++ b/apps/posts/src/views/comments/hooks/use-filter-state.ts
@@ -5,24 +5,28 @@ import type {Filter} from '@tryghost/shade';
 /**
  * Comment filter field keys - single source of truth for filter definitions
  */
-export const COMMENT_FILTER_FIELDS = ['status', 'created_at', 'body', 'post', 'author', 'reported'] as const;
+export const COMMENT_FILTER_FIELDS = ['id', 'status', 'created_at', 'body', 'post', 'author', 'reported'] as const;
 
 export type CommentFilterField = typeof COMMENT_FILTER_FIELDS[number];
 
 export function buildNqlFilter(filters: Filter[]): string | undefined {
     const parts: string[] = [];
-    
+
     for (const filter of filters) {
         if (!filter.values[0]) {
             continue;
         }
 
         switch (filter.field) {
-        case 'status': 
+        case 'id':
+            parts.push(`id:'${filter.values[0]}'`);
+            break;
+
+        case 'status':
             parts.push(`status:${filter.values[0]}`);
             break;
 
-        case 'created_at': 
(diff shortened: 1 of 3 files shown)


apps/posts/src/views/comments/hooks/use-filter-state.ts
```diff
@@ -5,53 +5,57 @@ import type {Filter} from '@tryghost/shade';
...
+
     for (const filter of filters) {
         if (!filter.values[0]) {
             continue;
         }
 
         switch (filter.field) {
-        case 'status': 
+        case 'id':
+            parts.push(`id:'${filter.values[0]}'`);
```
kevinansfield, line 22: I added `''` instinctively because we were burned a while back with fully-numeric object id's being treated as a number rather than a string which took a really long time to debug. That was in a server-side filter though so may not be necessary here.

apps/posts/src/views/comments/hooks/use-filter-state.ts
```diff
@@ -181,11 +192,16 @@ export function useFilterState(): UseFilterStateReturn {
     }, [filters, setSearchParams]);
 
     // Clear all filter params from URL
-    const clearFilters = useCallback(() => {
-        setSearchParams(new URLSearchParams(), {replace: true});
+    const clearFilters = useCallback(({replace = true}: {replace?: boolean} = {}) => {
```
rob-ghost, line 195: **Nitpick:** Not sure if we need a whole options object here, why not `(replace: boolean = true)` for the arguments? Do we expect more options later?

rob-ghost (approved):
Just a nitpick question, otherwise this looks good 👍🏻

apps/posts/src/views/comments/hooks/use-filter-state.ts
kevinansfield: I followed the same pattern we use for `setFilters` option because I felt like a single boolean when calling like `clearFilters(true)` / `clearFilters(false)` was ambiguous as to what the argument is doing

PR #25880, opened 2026-01-19
POC: Zod schema-driven validation for member commenting

rob-ghost:
## What does this PR do?

This is a proof-of-concept demonstrating how Zod can improve our approach to schema handling across the codebase. It refactors the member commenting feature to use Zod schemas for validation and transformation at both the data and API layers.

## Why are we exploring this?

Our current approach to parsing JSON from the database and serializing domain objects for API responses involves manual code that's spread across multiple layers. This POC explores whether Zod can provide a cleaner, more maintainable pattern.

## What improvements does Zod bring?

The main benefit is that schemas become the single source of truth. Instead of writing separate parsing logic, validation code, and TypeScript interfaces, a Zod schema handles all three. When we define `MemberCommentingFromDB`, it simultaneously describes what valid data looks like, transforms it into a domain object, and provides the TypeScript type.

This means changes to the data shape happen in one place. The schema validates input, transforms it, and if something's wrong, provides clear error messages. The domain class becomes simpler because it no longer needs to know about JSON strings or database formats - it just receives properly typed data.

The read-mutate-save pattern also becomes clearer. The service reads a member, the schema transforms the raw DB value into a domain object, the domain object creates a new state via immutable methods, and another schema handles serialization back to the database format.


rob-ghost:
Closed as it is too far out of date and will be replaced by a modern proposal

PR #25887, opened 2026-01-19
Changed from Stripe checkout to Stripe billing portal

sam-lord:
ref [FEA-502](https://linear.app/ghost/issue/FEA-502/portal-billing-management)

Added members API for creating the Stripe billing portal session & switched the front-end to use that API.


ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -195,6 +195,64 @@ module.exports = class RouterController {
...
+        } catch (err) {
+            logging.error(err);
+            res.writeHead(401);
+            return res.end('Unauthorized');
+        }
+
+        const member = email ? await this._memberRepository.get({email}) : null;
+
+        if (!member) {
+            res.writeHead(403);
```
aileen, line 219: I assume we intentionally return a `403` instead of a `404` here?

ghost/i18n/locales/af/portal.json
```diff
@@ -23,7 +23,7 @@
     "An unexpected error occured. Please try again or <a>contact support</a> if the error persists.": "'n Onverwagte fout het voorgekom. Probeer asseblief weer of <a>kontak kliëntediens</a> as die fout voortduur.",
     "Back": "Terug",
     "Back to Log in": "Terug na aanmelding",
-    "Billing info": "Rekening inligting",
+    "Billing info & receipts": "",
```
aileen, line 26: Having this string empty will now fall back to English, if I understand correctly. Does this mean that members on a translated site will see this as mixed content? (question applies to all updated translation strings) that now have no content in their respective language.

aileen (approved):
I left a comment regarding the translation string, which will appear in Portal. I'm just a bit concerned that we remove already translated strings and will replace them with fallback English. But I also don't see another way of solving this 🤔

ghost/i18n/locales/af/portal.json
sam-lord: Yeah that's my understanding.

sam-lord: I also think that's the intention with these string changes. When we make changes to user facing strings, it needs to be translated again.

sam-lord:
The error seems like something unrelated in the canary build :)

PR #25890, opened 2026-01-19
🐛 Allowed complimentary members to upgrade to a paid plan

lsinger:
ref https://linear.app/ghost/issue/BER-2985

Allowing all complimentary members to upgrade to a paid plan. In turn removed `allowCompMemberUpgrade` (because that would now always return true for all comped members) and simplified the places where it was used. 

---

> [!NOTE]
> Makes complimentary members eligible to upgrade like free members and removes now-redundant gating logic.
> 
> - Removes `allowCompMemberUpgrade` from `utils/helpers` and all call sites
> - Simplifies upgrade flow: treats complimentary members as upgrade-eligible in `account-plan-page` and `paid-account-actions` (always shows `Change` button unless only free plans)
> - Adjusts checkout/update logic to route complimentary members through `checkoutPlan` instead of subscription update paths
> - Cleans `app.js` page resolution (no longer forces `accountPlan` to `accountHome` for complimentary members)
> - Adds/updates tests to cover complimentary upgrade paths (single/multi-tier, monthly/yearly, cancelled subscriptions, offers) and adds `complimentaryWithCancelledSubscription` fixture
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8090bc937685134423ebe10dd26225532c34aa67. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
+++ b/apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
@@ -1,5 +1,5 @@
 import AppContext from '../../../../app-context';
-import {allowCompMemberUpgrade, getCompExpiry, getMemberSubscription, getMemberTierName, getUpdatedOfferPrice, hasMultipleProductsFeature, hasOnlyFreePlan, isComplimentaryMember, isPaidMember, isInThePast, subscriptionHasFreeTrial} from '../../../../utils/helpers';
+import {getCompExpiry, getMemberSubscription, getMemberTierName, getUpdatedOfferPrice, hasMultipleProductsFeature, hasOnlyFreePlan, isComplimentaryMember, isPaidMember, isInThePast, subscriptionHasFreeTrial} from '../../../../utils/helpers';
 import {getDateString} from '../../../../utils/date-time';
 import {ReactComponent as LoaderIcon} from '../../../../images/icons/loader.svg';
 import {ReactComponent as OfferTagIcon} from '../../../../images/icons/offer-tag.svg';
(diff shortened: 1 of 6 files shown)


apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
```diff
@@ -163,7 +161,7 @@ const PaidAccountActions = () => {
                         <h3>{planLabel}</h3>
                         <PlanLabel price={price} isComplimentary={isComplimentary} subscription={subscription} />
                     </div>
-                    <PlanUpdateButton isComplimentary={isComplimentary} isPaid={isPaid} isCancelled={isCancelled} />
```
sagzy, line 166: Looks like `isCancelled` prop was never used 😬

sagzy (commented):
If I create a comped member locally, the subscriptions array is an empty array in the `GET /members/api/member/` call and therefore `allowCompMemberUpgrade` is true. 

So this does seem to work already without the changes in the normal use case. Reading through the conversation on the Linear issue, it seems this fixes an edge case where member has a paid sub -> cancelled it -> got a comped sub - which is indeed pretty much an edge case.

I don't think we need to cover this specific edge case via automated test, but it would be great to have a test for the more general case: "a comped member can upgrade to a paid plan" - what do you think?

We should be able reuse the existing upgrade tests for free members and apply them to comped members: [apps/portal/test/upgrade-flow.test.js](https://github.com/TryGhost/Ghost/blob/b8581b315b7f093262a9d63db43c1c5087b17358/apps/portal/test/upgrade-flow.test.js#L186)

apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
lsinger: Yeah 😬

lsinger:
> I don't think we need to cover this specific edge case via automated test, but it would be great to have a test for the more general case: "a comped member can upgrade to a paid plan" - what do you think?

Thank you @sagzy -- ~I~ Claude added tests for the regular and the edge case in [73a03fd](https://github.com/TryGhost/Ghost/pull/25890/commits/73a03fd102cc2cf3ad4577cee1a154890171f8fb).

sagzy (approved):
looks good!

PR #25891, opened 2026-01-19
🐛 Fixed email-only posts stuck as "sent" after limit/verification failure

kevinansfield:
ref https://linear.app/ghost/issue/ONC-1382/

When sending a scheduled email-only post, no rollback occurred if the email creation part failed meaning the post got stuck in a "sent" state.

- added `EmailService.checkCanSendEmail` method
  - performs checks that were previously performed as part of `createEmail` method
  - allows calls from outside of `EmailService` to gate calling `createEmail`
- added pre-email-send checks to `PostsService.editPost`
  - determines if an email would be sent based on query params, existing post state, and new post status
  - uses `EmailService.checkCanSendEmail` to throw early before we get to any post saving or email creation

### DIFF
--- a/ghost/core/core/server/services/email-service/email-service.js
+++ b/ghost/core/core/server/services/email-service/email-service.js
@@ -102,12 +102,15 @@ class EmailService {
     }
 
     /**
+     * Pre-check if email sending would be allowed before making any post changes.
+     * This validates limits and verification requirements early to avoid leaving
+     * posts in a stuck "sent" state if email creation would fail.
      *
-     * @param {Post} post
-     * @returns {Promise<Email>}
+     * @param {object} newsletter - The newsletter model to send to
+     * @param {string} emailRecipientFilter - The recipient filter for the email
+     * @returns {Promise<{emailCount: number}>} The email count if checks pass, throws if email cannot be sent
      */
-    async createEmail(post) {
-        let newsletter = await post.getLazyRelation('newsletter');
+    async checkCanSendEmail(newsletter, emailRecipientFilter) {
         if (!newsletter) {
             throw new errors.EmailError({
                 message: tpl(messages.missingNewsletterError)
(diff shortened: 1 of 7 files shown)


ghost/core/core/server/services/posts/posts-service.js
```diff
@@ -60,20 +60,57 @@ class PostsService {
...
+        if (['published', 'sent'].includes(newStatus)) {
+            existingPost = await this.models.Post.findOne(
+                {id: frame.options.id, status: 'all'},
+                {columns: ['id', 'status', 'newsletter_id', 'email_recipient_filter']}
+            );
+            const previousStatus = existingPost?.get('status');
+
+            // Check if we'll be sending an email - either via new newsletter param or existing newsletter on post
+            if (frame.options.newsletter || existingPost?.get('newsletter_id')) {
+                sendingEmail = this.shouldSendEmail(newStatus, previousStatus);
```
sagzy: On lines 120-122, we have 
```
const sendEmail = model.wasChanged() && this.shouldSendEmail(model.get('status'), model.previous('status'));

if (sendEmail) { ... }
```

Is this now redundant?

sagzy (approved):

ghost/core/core/server/services/posts/posts-service.js
kevinansfield: One would hope so _but_ we have business logic in the model around status changes so we can't be 100% sure that the model status was actually set to what we had in the request. In the refactor this detail does get extracted into the `createOrRetryEmail` method so there's more distance/less coupling between pre-edit validation state and final model state

ghost/core/core/server/services/email-service/email-service.js
```diff
@@ -117,15 +120,28 @@ class EmailService {
         if (newsletter.get('status') !== 'active') {
             // A post might have been scheduled to an archived newsletter.
             // Don't send it (people can't unsubscribe any longer).
-            throw new errors.EmailError({
+            throw new errors.BadRequestError({
```
sagzy, line 123: Is this change from EmailError → BadRequestError intentional?

sagzy (approved):

ghost/core/core/server/services/email-service/email-service.js
kevinansfield: That was intentional because it bubbles up to the front-end if an edit post API request is made specifying an email to an archived newsletter - previously that would serve a 500 rather than the more appropriate 400.

Email records only get created via the post edit API, even if it's the scheduler that is making that API request. I couldn't see a reason to keep it as a 500 but happy to revert if you're aware of problems it would cause.

PR #25893, opened 2026-01-19
Improved keyboard events in member welcome email modal

troyciesco:
ref https://linear.app/ghost/issue/NY-852

- Fixes issue where `Cmd+K` would open both site search and "Enter url" input when focused in the editor and with text selected. Now just opens the url input (note it's an existing Koenig shortcut)
- Makes it so hitting `Escape` with the send test email dropdown open will first close that dropdown if it's open, instead of closing the modal entirely. Hitting `Escape` again will close the modal.

### DIFF
--- a/apps/admin-x-settings/src/components/settings/membership/member-emails/member-email-editor.tsx
+++ b/apps/admin-x-settings/src/components/settings/membership/member-emails/member-email-editor.tsx
@@ -29,25 +29,34 @@ const MemberEmailsEditor: React.FC<MemberEmailsEditorProps> = ({
         }
     }, [onChange, value]);
 
+    // Stop Cmd+K from bubbling to global search handler
+    const handleKeyDown = useCallback((e: React.KeyboardEvent) => {
+        if ((e.metaKey || e.ctrlKey) && e.key === 'k') {
+            e.stopPropagation();
+        }
+    }, []);
+
     return (
-        <KoenigEditorBase
-            className={className}
-            emojiPicker={false}
-            inheritFontStyles={false}
-            initialEditorState={value}
-            nodes={nodes}
-            placeholder={placeholder}
-            singleParagraph={singleParagraph}
-            onChange={handleChange}
-        >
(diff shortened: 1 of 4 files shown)


troyciesco:
@coderabbitai review

cmraible (commented):
I don't think this quite fixes the bug described in https://linear.app/ghost/issue/NY-852. 

If I hit CMD+K to insert a link, it works now so that part is good — but if I hit escape to dismiss the link's URL input, it just exits the whole modal.

---

I'm also thinking about tests — it feels uncomfy to merge a bug fix without any tests. I wonder if we could add a few tests in the admin-x-settings acceptance test suite? A full E2E test feels a bit too heavy for this, but maybe the admin-x-settings tests would be appropriate - WDYT?

cmraible:
Thanks for adding the tests, they look good! Not sure if still in progress or not, but I'm still seeing the same behavior with the create link popup, as described in the original issue.

I'm fine to merge this as it's definitely an improvement in its current form, but I don't think it fully addresses the bug described in the issue.

troyciesco:
> Thanks for adding the tests, they look good! Not sure if still in progress or not, but I'm still seeing the same behavior with the create link popup, as described in the original issue.
> 
> I'm fine to merge this as it's definitely an improvement in its current form, but I don't think it fully addresses the bug described in the issue.

@cmraible ah sorry, was still in progress, coming back to give an update on this one kept getting bumped yesterday lol

So at first my position (I posted on the ticket) was that fixing the `"Enter url" input doesn't close with esc` might not be worth fixing, because it closes with enter (presumably after you've typed what you need to type). 

Then I went to the minimal Koenig demo to see if the behavior was the same there, and saw that even in the minimal editor you still get access to post links that you can click. Then I realized: maybe the behavior should be the same here, or at least it's worth figuring out why it's acting differently in a modal. I think it might be something in Koenig, so I'm going to check it out there.

Good idea on merging these fixes though, i'll do that, and i've updated the description of the PR so it doesn't close the associated ticket.

PR #25899, opened 2026-01-19
Updated top devices endpoint to default to desktop for unknown values

9larsons:
ref https://linear.app/ghost/issue/NY-935/

Currently, 'unknown' is ~0.35% of our traffic. Because this is parsed from the user-agent, it would be awfully uncommon to not encounter, and is generally not helpful to the user. Given that desktop is the biggest value at ~100x the unknown (~40%), we're going to default to Desktop. It should be a very minor increment and in exchange we'll have a cleaner ui.

I've made this change in the pipes so that we don't need to migrate all the data in Tinybird, although that's also a viable option.

### DIFF
--- a/ghost/core/core/server/data/tinybird/endpoints/api_top_devices.pipe
+++ b/ghost/core/core/server/data/tinybird/endpoints/api_top_devices.pipe
@@ -5,7 +5,7 @@ NODE top_devices
 SQL >
     %
     select
-        device,
+        case when device = 'unknown' then 'desktop' else device end as device,
         count() as visits
     from mv_session_data sd
       inner join filtered_sessions fs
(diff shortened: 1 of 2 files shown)


cmraible:
I'm a little surprised this didn't change any of our test's expectations. Do you think we should have one of these "unknown" cases in our test fixture?

It's a tiny change so I don't feel _that_ strongly about it, but it feels a bit uncomfortable to ship a behavior change like this with no test additions/updates.

EvanHahn (approved):
As long as this isn't changing the underlying data from "unknown" to "desktop" (which it doesn't appear to be), I'm ok reporting it this way.

9larsons:
Ooh yeah, I was tunnel visioned yesterday. This is actually problematic. We'd be obscuring the ~0.34% results... and as for why it didn't break tests, I saw this before w/ Tinybird not using the branch data. Let me investigate further... I think I may close this PR and re-open. While we could still do this and deal with the 0.35% 'fudge factor' to clean up the UI, I think it's fine until we find a real solution.

PR #25902, opened 2026-01-20
Added `automated_email_recipients` database table

cmraible:
ref https://linear.app/ghost/issue/NY-936

## Summary

- Adds `automated_email_recipients` table to track automated emails sent to members
- First step in implementing welcome emails in the member activity feed
- Follows the `email_recipients` pattern: no FK on `member_id` to preserve audit history when members are deleted

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/6.14/2026-01-19-23-55-51-add-automated-email-recipients-table.js
+++ b/ghost/core/core/server/data/migrations/versions/6.14/2026-01-19-23-55-51-add-automated-email-recipients-table.js
@@ -3,9 +3,9 @@ const {addTable} = require('../../utils');
 module.exports = addTable('automated_email_recipients', {
     id: {type: 'string', maxlength: 24, nullable: false, primary: true},
     automated_email_id: {type: 'string', maxlength: 24, nullable: false, references: 'automated_emails.id'},
-    member_id: {type: 'string', maxlength: 24, nullable: false},
+    member_id: {type: 'string', maxlength: 24, nullable: false, index: true},
     processed_at: {type: 'dateTime', nullable: false},
-    member_uuid: {type: 'string', maxlength: 36, nullable: true},
-    member_email: {type: 'string', maxlength: 191, nullable: true},
+    member_uuid: {type: 'string', maxlength: 36, nullable: false},
+    member_email: {type: 'string', maxlength: 191, nullable: false},
     member_name: {type: 'string', maxlength: 191, nullable: true}
 });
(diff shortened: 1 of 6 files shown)


ghost/core/core/server/data/migrations/versions/6.14/2026-01-19-23-55-51-add-automated-email-recipients-table.js
```diff
@@ -0,0 +1,11 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('automated_email_recipients', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    automated_email_id: {type: 'string', maxlength: 24, nullable: false, references: 'automated_emails.id'},
+    member_id: {type: 'string', maxlength: 24, nullable: false},
```
troyciesco: blocking: hmm wasn't there an index here in the original plan?

ghost/core/core/server/data/migrations/versions/6.14/2026-01-19-23-55-51-add-automated-email-recipients-table.js
```diff
@@ -0,0 +1,11 @@
...
+    processed_at: {type: 'dateTime', nullable: false},
+    member_uuid: {type: 'string', maxlength: 36, nullable: true},
+    member_email: {type: 'string', maxlength: 191, nullable: true},
```
troyciesco: blocking (i think?): these weren't nullable in the original plan - would they ever actually be nullable if we have a legit record? member_name def could be so that makes sense

troyciesco (changes requested):
marking as request changes bc i think we want to do the two things i noted but lmk if i'm wrong!

ghost/core/core/server/data/migrations/versions/6.14/2026-01-19-23-55-51-add-automated-email-recipients-table.js
cmraible: Yes, good catch! There should be one for member_id. Wonder where we lost that along the way 🤔 Will address in a minute

ghost/core/core/server/data/migrations/versions/6.14/2026-01-19-23-55-51-add-automated-email-recipients-table.js
cmraible: Yep good call, this is fixed now

troyciesco (approved):

PR #25904, opened 2026-01-20
Added UI to disable member commenting from comment list

rob-ghost:

## What does this PR do?

This adds the ability for staff to disable a member's commenting privileges directly from the comment moderation screen in Ghost Admin. When reviewing comments, staff can now click the menu on any comment and choose "Disable commenting" to prevent that member from commenting in the future. A confirmation dialog explains the action and the member's comments will show a visual indicator once they've been banned.

The feature includes:
- "Disable commenting" / "Enable commenting" menu items in the comment row dropdown
- A confirmation dialog when disabling
- A visual indicator (crossed-out message icon) next to banned members' names
- API mutation hooks for the disable/enable endpoints

## Why is this needed?

Comment moderation currently allows hiding or deleting individual comments, but there's no way to prevent a problematic member from continuing to post. This gives staff a quick way to ban repeat offenders without leaving the comments screen.

## How is this gated?

The UI is behind the `disableMemberCommenting` labs flag. The flag doesn't exist yet - it will be added to `GA_FEATURES` when the backend API is deployed, which will automatically enable this UI for all users.

## Testing

E2E tests are included and will run once the labs flag is enabled. Manual testing requires the backend API to be available.

### DIFF
--- a/apps/admin-x-framework/src/api/members.ts
+++ b/apps/admin-x-framework/src/api/members.ts
@@ -1,9 +1,11 @@
-import {Meta, createQuery} from '../utils/api/hooks';
+import {Meta, createMutation, createQuery} from '../utils/api/hooks';
 
 export type Member = {
     id: string;
     name?: string;
     email?: string;
+    avatar_image?: string;
+    can_comment?: boolean;
 };
 
 export interface MembersResponseType {
@@ -16,4 +18,30 @@ const dataType = 'MembersResponseType';
 export const useBrowseMembers = createQuery<MembersResponseType>({
     dataType,
     path: '/members/'
+});
+
+export const useDisableMemberCommenting = createMutation<
+    MembersResponseType,
+    {id: string; reason: string}
+>({
+    method: 'POST',
+    path: ({id}) => `/members/${id}/commenting/disable`,
+    body: ({reason}) => ({
+        reason
+    }),
+    invalidateQueries: {
+        dataType: 'CommentsResponseType'
+    }
+});
+
+export const useEnableMemberCommenting = createMutation<
+    MembersResponseType,
+    {id: string}
+>({
+    method: 'POST',
+    path: ({id}) => `/members/${id}/commenting/enable`,
+    body: () => ({}),
+    invalidateQueries: {
+        dataType: 'CommentsResponseType'
+    }
 });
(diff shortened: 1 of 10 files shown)


apps/admin-x-framework/src/api/members.ts
```diff
@@ -17,3 +19,29 @@ export const useBrowseMembers = createQuery<MembersResponseType>({
...
+    MembersResponseType,
+    {id: string}
+>({
+    method: 'POST',
+    path: ({id}) => `/members/${id}/commenting/enable`,
+    body: () => ({}),
+    invalidateQueries: {
+        dataType: 'CommentsResponseType'
+    }
+});
```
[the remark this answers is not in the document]
rob-ghost: This is future-proofing for a scenario that doesn't exist yet. We don't show the status on the Member screen ATM, we can cross that bridge when we come to it.

apps/posts/src/views/comments/comments.tsx
```diff
@@ -14,6 +14,7 @@ const Comments: React.FC = () => {
     const {filters, nql, setFilters, clearFilters, isSingleIdFilter} = useFilterState();
     const {data: configData} = useBrowseConfig();
     const commentPermalinksEnabled = configData?.config?.labs?.commentPermalinks === true;
+    const disableMemberCommentingEnabled = configData?.config?.labs?.disableMemberCommenting === true;
```
kevinansfield, line 17: Nit / Non-blocker: We have some inconsistency here, most places use the `useFeatureFlag` hook rather than browsing config.

It seems like the `useFeatureFlag` hook in `posts` app uses settings which differs to all the others that use config. We could simplify it to match seeing as our [`FeatureToggle` component also updates the config cache](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-settings/src/components/settings/advanced/labs/feature-toggle.tsx#L25-L33)?

kevinansfield (approved):

weylandswart (approved):
Design tested and looks good to me!

apps/posts/src/views/comments/comments.tsx
rob-ghost: I refactored `useFeatureFlag` here: https://github.com/TryGhost/Ghost/pull/26038

PR #25911, opened 2026-01-20
Added data-attribute for Stripe customer billing portal

sam-lord:
ref [FEA-545](https://linear.app/ghost/issue/FEA-545/add-new-data-attribute-to-use-billing-portal-flow)

### DIFF
--- a/apps/portal/test/data-attributes.test.js
+++ b/apps/portal/test/data-attributes.test.js
@@ -3,7 +3,7 @@ import {site as FixturesSite, member as FixtureMember} from './utils/test-fixtur
 import {fireEvent, appRender, within} from './utils/test-utils';
 import setupGhostApi from '../src/utils/api';
 import * as helpers from '../src/utils/helpers';
-import {formSubmitHandler, planClickHandler} from '../src/data-attributes';
+import {formSubmitHandler, planClickHandler, handleDataAttributes} from '../src/data-attributes';
 import {vi} from 'vitest';
 
 // Mock data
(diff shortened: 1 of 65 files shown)


apps/portal/test/data-attributes.test.js
```diff
@@ -348,6 +348,227 @@ describe('Member Data attributes:', () => {
...
+            handleDataAttributes({siteUrl});
+
+            const button = document.querySelector('[data-members-manage-billing]');
+            button.click();
+
+            await new Promise((resolve) => {
+                setTimeout(resolve, 0);
+            });
+
+            // Should still call billing portal endpoint with null identity
```
aileen, line 586: Why is that? Is that expected to still call the endpoint, even when the identity is `null`?

aileen (approved):
Left a comment, as I'm not sure about the behavior that was asserted in the test. Maybe it's intentional, maybe an oversight. Just wanted to flag it up. Everything else looks good to me.

apps/portal/test/data-attributes.test.js
sam-lord: It's the same functionality as the original `edit-billing` data attribute -- they both call the members API even if the identity is null. The API returns a 400 with an error message, and I think the front-end has its own error message it displays for that case.

It's a bit weird, and the comment doesn't do anything to explain that weirdness, but it's an established pattern that we're replicating. We _could_ prevent the request from being sent but it doesn't make a huge difference either way.

PR #25912, opened 2026-01-20
Simplified audience filter handling in posts and stats apps

9larsons:
ref https://linear.app/ghost/issue/NY-825/

When working on the stats-filter components (behind the previous utmTracking feature flag), we moved the Locations view's component into the Web view. We also changed the filter from an Audience dropdown to the filter component.

In doing this, we no longer needed to coordinate the audience state across views, and we also can handle the audience logic via the filter component. This allows us to significantly clean up the handling in the frontend apps.


apps/posts/src/views/PostAnalytics/components/audience-select.tsx
```diff
@@ -1,15 +1,14 @@
...
-import {useGlobalData} from '@src/providers/post-analytics-context';
 
 export const AUDIENCE_TYPES = [
     {name: 'Public visitors', value: 'undefined', bit: AUDIENCE_BITS.PUBLIC},
     {name: 'Free members', value: 'free', bit: AUDIENCE_BITS.FREE},
     {name: 'Paid members', value: 'paid', bit: AUDIENCE_BITS.PAID}
 ];
 
+// Re-export ALL_AUDIENCES for convenience
+export {ALL_AUDIENCES};
```
troyciesco: this seems odd 🤔 is it even used anywhere that couldn't import it from `@src/utils/constants`?

apps/stats/src/hooks/use-top-sources-growth.ts
```diff
@@ -1,16 +1,14 @@
+import {ALL_AUDIENCES, getAudienceQueryParam} from '@views/Stats/components/audience-select';
```
troyciesco: This too - shouldn't `ALL_AUDIENCES` come from `@src/utils/constants`? 
(i see this one's from a different app than the other spot i commented but it looks like they both have it defined)

troyciesco (changes requested):

EvanHahn (approved):
LGTM (though I admit I don't completely understand the context yet).

apps/posts/src/views/PostAnalytics/Web/web.tsx
```diff
@@ -29,23 +27,19 @@ interface postAnalyticsProps {}
 const Web: React.FC<postAnalyticsProps> = () => {
     const navigate = useNavigate();
     const {postId} = useParams();
-    const {statsConfig, isLoading: isConfigLoading, range, audience: globalAudience, data: globalData, post, isPostLoading} = useGlobalData();
+    const {statsConfig, isLoading: isConfigLoading, range, data: globalData, post, isPostLoading} = useGlobalData();
 
     // Use URL-synced filter state for bookmarking and sharing
     const {filters: utmFilters, setFilters: setUtmFilters} = useFilterParams();
```
9larsons, line 33: Side note: This utmFilters convention needs to be cleaned up to simply be queryParams or webFilters. That kind of naming.

apps/posts/src/views/PostAnalytics/Web/web.tsx
```diff
@@ -217,28 +211,19 @@ const Web: React.FC<postAnalyticsProps> = () => {
     return (
         <>
             <PostAnalyticsHeader currentTab='Web'>
-                {!utmTrackingEnabled ?
```
9larsons, line 220: The previous default behavior was the flag was disabled and so we'd display the old filter. That's no longer the case, and with the change to the new filters we can remove all refs to the old AudienceSelect component.

apps/posts/src/views/PostAnalytics/components/audience-select.tsx
```diff
@@ -1,8 +1,4 @@
-import React from 'react';
 import {ALL_AUDIENCES, AUDIENCE_BITS} from '@src/utils/constants';
-import {Button, DropdownMenu, DropdownMenuCheckboxItem, DropdownMenuContent, DropdownMenuTrigger, LucideIcon} from '@tryghost/shade';
-import {useAppContext} from '@src/providers/posts-app-context';
-import {useGlobalData} from '@src/providers/post-analytics-context';
 
 export const AUDIENCE_TYPES = [
     {name: 'Public visitors', value: 'undefined', bit: AUDIENCE_BITS.PUBLIC},
```
9larsons, line 4: This file is now essentially a utils file. I will follow up with a refactor PR - I mostly want to strip out the dead code here.

apps/posts/src/views/PostAnalytics/components/audience-select.tsx
9larsons: Agreed, rm'd.

troyciesco (approved):

PR #25914, opened 2026-01-20
Fixed adminForward cookie handling for subdirectory sites

kevinansfield:
ref https://linear.app/ghost/issue/BER-3225/

- on subdirectory sites the fixed `/ghost/` cookie path is incorrect resulting in a refresh loop where the old Admin assets keep getting loaded whilst the `adminForward` flag is enabled because it's not able to read the cookie
- used the existing `ghostPaths.adminRoot` property to set a dynamic cookie path based on the loaded admin URL

### DIFF
--- a/ghost/admin/app/services/feature.js
+++ b/ghost/admin/app/services/feature.js
@@ -48,19 +48,20 @@ function getCookie(name) {
     return null;
 }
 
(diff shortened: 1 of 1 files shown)


jonatansberg (approved):
Looks good!

PR #25918, opened 2026-01-20
Fixed subdir handling in admin-forward dev/build

kevinansfield:
no issue

Ghost can be run on a subdir such as `http://localhost:2368/blog/` which is useful for finding edge cases with path handling in local development. However the new admin-forward dev server was not handling that correctly as it was hardcoded to expect `/ghost/` on the root.

- updated backend proxy to hit and proxy the API on the right path when configured with a subdirectory URL
- updated vite config to generate relative paths for assets in the `index-forward.html` file used for serving `/ghost/` when not building for CDN deployment

### DIFF
--- a/apps/admin/vite-backend-proxy.ts
+++ b/apps/admin/vite-backend-proxy.ts
@@ -4,6 +4,16 @@ import type { IncomingMessage } from "http";
 const GHOST_URL = process.env.GHOST_URL ?? "http://localhost:2368/";
 
 /**
+ * Extracts the subdirectory path from GHOST_URL.
+ * e.g., "http://localhost:2368/blog/" -> "/blog"
+ *       "http://localhost:2368/" -> ""
+ */
+function getSubdir(): string {
+    const url = new URL(GHOST_URL);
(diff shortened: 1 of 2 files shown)


apps/admin/vite-backend-proxy.ts
```diff
@@ -11,7 +21,7 @@ async function resolveGhostSiteUrl() {
     const MAX_ATTEMPTS = 20;
     for (let attempt = 1; attempt <= MAX_ATTEMPTS; attempt++) {
         try {
-            const siteEndpoint = new URL("/ghost/api/admin/site/", GHOST_URL);
+            const siteEndpoint = new URL('ghost/api/admin/site/', GHOST_URL);
```
kevinansfield, line 24: Forward slash removed from the path here so `URL` construction correctly builds the path on top of `GHOST_URL`

kevinansfield:
This is only part of the problem, the generated `index-forward.html` file doesn't work with subdirectories currently

kevinansfield:
Updated with the `index-forward.html` fix

PR #25920, opened 2026-01-20
Normalized email addresses more reliably

EvanHahn:
closes https://linear.app/ghost/issue/NY-942/create-and-use-parse-email-address-package-in-core
ref https://github.com/TryGhost/Ghost/pull/25919
ref https://github.com/TryGhost/framework/pull/342

This creates the `parse-email-address` package (previously reviewed [here][0]--all I did was move it over) and uses it in email normalization, used when sending magic links.

This should not change behavior for "normal", lowercase ASCII email addresses. However, it does make several subtler tweaks to the way emails are normalized:

- Some totally invalid emails weren't normalized.

  Before: if `req.body.email` wasn't a string or lacked an `@`, no normalization would occur. The system would then attempt to send an email that'd almost certainly fail.

  After: these throw a "bad request" error.

- Invalid emails with multiple `@` signs were considered valid.

  Before: `foo@bar@example.com` was accepted.

  After: it is not accepted and throws a "bad request" error.

- Domains are now lowercased.

  Before: `FOO@Example.COM` was normalized to `FOO@Example.COM`.

  After: `FOO@Example.COM` is normalized to `FOO@example.com`.

[0]: https://github.com/TryGhost/framework/pull/340

---

> [!NOTE]
> **Introduce and integrate email parsing utility**
> 
> - New TS package `@tryghost/parse-email-address` (wrapper around `parse-email-address`) with tests and build config
> - Core `normalize-email` now uses `parseEmailAddress` to punycode domains and lowercase them; returns `null` for invalid emails (e.g. multiple `@`)
> - `members` router controller updated to use strict normalization; rejects invalid emails earlier and logs normalization changes
> 
> **Tests and tooling**
> 
> - Updated unit tests reflecting stricter normalization (invalid inputs now `null`, domain lowercased)
> - CI: build TS packages before acceptance/legacy tests; Dockerfile adds builder stage for `ghost/parse-email-address` and includes its build output
> - `ghost/core/package.json` adds dependency on `@tryghost/parse-email-address`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eaad95ab03b0711527e990fe341cd5d029c9a3d8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
+++ b/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
@@ -682,20 +682,14 @@ module.exports = class RouterController {
         }
 
         // Normalize email to prevent homograph attacks
-        let normalizedEmail;
-
-        try {
-            normalizedEmail = normalizeEmail(email);
-
-            if (normalizedEmail !== email) {
-                logging.info(`Email normalized from ${email} to ${normalizedEmail} for magic link`);
-            }
-        } catch (err) {
-            logging.error(`Failed to normalize [${email}]: ${err.message}`);
-
+        const normalizedEmail = normalizeEmail(email);
+        if (!normalizedEmail) {
             throw new errors.BadRequestError({
                 message: tpl(messages.invalidEmail)
             });
+        }
+        if (normalizedEmail !== email) {
+            logging.info(`Email normalized from ${email} to ${normalizedEmail} for magic link`);
         }
 
         if (honeypot) {
--- a/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
+++ b/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
@@ -688,9 +688,6 @@ module.exports = class RouterController {
                 message: tpl(messages.invalidEmail)
             });
         }
-        if (normalizedEmail !== email) {
-            logging.info(`Email normalized from ${email} to ${normalizedEmail} for magic link`);
-        }
 
         if (honeypot) {
             logging.warn('Honeypot field filled, this is likely a bot');
--- a/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
+++ b/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
@@ -681,7 +681,7 @@ module.exports = class RouterController {
             });
         }
 
-        // Normalize email to prevent homograph attacks
+        // Normalize email to avoid invalid addresses and mitigate homograph attacks
         const normalizedEmail = normalizeEmail(email);
         if (!normalizedEmail) {
             throw new errors.BadRequestError({
--- a/ghost/parse-email-address/src/index.ts
+++ b/ghost/parse-email-address/src/index.ts
@@ -1,0 +1,20 @@
+import {parseEmailAddress as upstreamParseEmailAddress} from 'parse-email-address';
+import {domainToASCII} from 'node:url';
+
+export const parseEmailAddress = (
+    emailAddress: string
+): null | { local: string; domain: string } => {
+    const upstreamParsed = upstreamParseEmailAddress(emailAddress);
+    if (!upstreamParsed) {
+        return null;
+    }
+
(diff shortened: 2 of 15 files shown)


ghost/parse-email-address/src/index.ts
EvanHahn, line 1: Everything in these files is identical to what it was in the Framework repo. I don't think it's worth reviewing anything in `ghost/parse-email-address/`.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -682,21 +682,15 @@ module.exports = class RouterController {
...
-
+        const normalizedEmail = normalizeEmail(email);
+        if (!normalizedEmail) {
             throw new errors.BadRequestError({
                 message: tpl(messages.invalidEmail)
             });
         }
+        if (normalizedEmail !== email) {
+            logging.info(`Email normalized from ${email} to ${normalizedEmail} for magic link`);
+        }
```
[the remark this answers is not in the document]
9larsons: I agree with this. I think we can make this a debug statement, if anything. How often would we expect this to occur?

ghost/parse-email-address/tsconfig.json
```diff
@@ -0,0 +1,110 @@
...
+    "noUnusedParameters": true,                          /* Raise an error when a function parameter isn't read. */
+    "exactOptionalPropertyTypes": true,                  /* Interpret optional property types as written, rather than adding 'undefined'. */
+    "noImplicitReturns": true,                           /* Enable error reporting for codepaths that do not explicitly return in a function. */
+    "noFallthroughCasesInSwitch": true,                  /* Enable error reporting for fallthrough cases in switch statements. */
+    "noUncheckedIndexedAccess": true,                    /* Add 'undefined' to a type when accessed using an index. */
+    "noImplicitOverride": true,                          /* Ensure overriding members in derived classes are marked with an override modifier. */
+    "noPropertyAccessFromIndexSignature": true,          /* Enforces using indexed accessors for keys declared using an indexed type. */
+    "allowUnusedLabels": true,                           /* Disable error reporting for unused labels. */
+    "allowUnreachableCode": true,                        /* Disable error reporting for unreachable code. */
```
9larsons: Feels odd to allow unused and unreachable. I'm assuming this is part of the defaults?

9larsons (changes requested):
Just the minor comment about removing email address logging. I missed adding this to the new review functionality. 

I was also going to say we could remove punycode, but it seems we never explicitly listed it as a dependency...

ghost/parse-email-address/tsconfig.json
EvanHahn: Whoops. Fixed in 99916f045fbfdc375aa95d0cc5bdcaa1c16da398.

EvanHahn:
> Just the minor comment about removing email address logging.

I couldn't find the comment, but I think you might be talking about this line being removed:

https://github.com/TryGhost/Ghost/blob/de8f7ee6f1732d8055130633a9585022c7cefd71/ghost/core/core/server/services/members/members-api/controllers/router-controller.js#L694

I deleted this because it's not a bug that we failed to normalize the email. If someone gives us a bogus email address, that's not a Ghost bug. Happy to put it back (or change it to a warning), though, if that's the kind of thing we want to log.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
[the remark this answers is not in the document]
EvanHahn: Fixed in f885a546de0d14cbae06d0e426530a5614092d31.

EvanHahn:
Updated a comment (no user impact) in 189c33bff42546fd967030e4a8ba0180d136e7b0.

9larsons (approved):
LGTM!

PR #25921, opened 2026-01-20
Added tracking and retrieval for `automated_email_recipients` events

cmraible:
ref https://linear.app/ghost/issue/NY-938/track-and-retrieve-welcome-email-events

This change introduces the `AutomatedEmailRecipient` Bookshelf model, modifies the `member-created` outbox handler to record welcome email sends in the `automated_email_recipients` database table, and adds a `getAutomatedEmailSentEvents` method to the `EventRepository` to retrieve these events for the member activity feed. It also ensures the member's UUID is included in the outbox payload for accurate tracking and wires up all necessary dependencies.


ghost/core/core/server/services/members/members-api/repositories/event-repository.js
```diff
@@ -880,6 +886,47 @@ module.exports = class EventRepository {
...
+            useBasicCount: true,
+            mongoTransformer: chainTransformers(
+                replaceCustomFilterTransformer(filter),
+                ...mapKeys({
+                    'data.created_at': 'processed_at',
+                    'data.member_id': 'member_id'
+                })
+            )
+        };
+        options.order = options.order.replace(/created_at/g, 'processed_at');
```
cmraible: This is a little awkward. I wonder if we should change the schema to use the standard `created_at`, `updated_at` timestamp columns. Not a big deal either way, but it would simplify this little chunk of code a bit.

troyciesco: hmm tricky, since `email_recipients` uses `processed_at` too.
Questions I'm asking myself:
- if `email_recipients` didn't exist and we were creating it today, would we still used `processed_at`?
- what does `processed_at` mean? is it really the same as `created_at`? (for either table). it seems like it might actually be _different_ in `email_recipients` because it's nullable, which gives me pause about making them 1:1 in `automated_email_recipients`

i kinda lean towards `created_at` and `updated_at` but curious if anything about those questions above changes anything for you.

cmraible: Great questions! I will answer them in reverse order:

* `processed_at` means that's when Ghost sent the request to Mailgun to send the email. It's nullable because the `email_recipients` rows are created at the same time its `email_batch` is created, and we create the batches in the DB before sending them to mailgun to be sent. This way if the batch fails to submit to mailgun for whatever reason, we can detect that and resubmit the batch.
* I think we would still use `processed_at` for `email_recipients` if we built it again today. BUT only because of the batching, which isn't a concern for `automated_email_recipients`, which are more one-piece-flow.

I think I'm with you that `created_at` and `updated_at` makes sense for `automated_email_recipients`. I don't see any need to ever create an `automated_email_recipient` until we've sent it to mailgun already. If submitting to mailgun fails for whatever reason, it should stay in the outbox and be retried, so there's no need for a transient state in `automated_email_recipients` where it exists but hasn't been processed yet.<br><br>I'll go ahead and make that change to the migration and the downstream changes in these PRs.

troyciesco (approved):

cmraible:
(Just rebased on `main` after merging the migration, no new code since this was last reviewed)

PR #25941, opened 2026-01-21
Fixed closing of link input in member welcome email editor

troyciesco:
closes https://linear.app/ghost/issue/NY-852

When using Cmd+K to add a link in the welcome email editor, pressing ESC would close the entire modal (losing edits) and navigate away from settings, instead of just closing the link input.

### Changes

- `modal.tsx`: Skip closing when Koenig's link input is focused (`data-testid="link-input"`), allowing the ESC event to propagate to Koenig's handler which closes the input
- `main-content.tsx`: Skip navigation to dashboard when a modal is open (`#modal-backdrop` exists), since the modal's ESC handler doesn't stop propagation when it skips closing

### Manual Testing
- Open Settings > Welcome emails > Edit on a welcome email
- Type some text, select it, press Cmd+K to open the link input
- Press ESC
- Link input should close, modal should stay open, settings page should remain visible
- Press ESC again — modal should close normally

Related Koenig PR https://github.com/TryGhost/Koenig/pull/1693

### DIFF
--- a/apps/admin-x-settings/src/main-content.tsx
+++ b/apps/admin-x-settings/src/main-content.tsx
@@ -32,6 +32,12 @@ const MainContent: React.FC = () => {
     useEffect(() => {
         const handleKeyDown = (event: KeyboardEvent) => {
             if (event.key === 'Escape') {
+                // Don't navigate away if a modal is open - let the modal handle ESC
+                const modalBackdrop = document.getElementById('modal-backdrop');
+                if (modalBackdrop) {
+                    return;
+                }
+
                 confirmIfDirty(isDirty, () => {
                     navigateAway('/');
                 });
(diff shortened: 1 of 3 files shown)


troyciesco:
@coderabbitai review

apps/admin-x-settings/src/main-content.tsx
```diff
@@ -32,6 +32,12 @@ const MainContent: React.FC = () => {
     useEffect(() => {
         const handleKeyDown = (event: KeyboardEvent) => {
             if (event.key === 'Escape') {
+                // Don't navigate away if a modal is open - let the modal handle ESC
+                const modalBackdrop = document.getElementById('modal-backdrop');
```
troyciesco, line 36: need to see if there's a more robust way of doing this

troyciesco: So there is a context for the modal package we use and we could check for if a modal is visible, but apparently there's [two separate contexts](https://github.com/TryGhost/Ghost/blob/main/apps/admin-x-settings/src/app.tsx#L17)? 🤔 

Which worries me that there'd be a way to get into an incorrect state or something if we tried to use the context value, as opposed to checking the DOM directly. (I don't have a specific example of how, just gut reaction)

This does still feel brittle though.

cmraible: Yeah agree it's not ideal, but I think this is the right judgement call given the two separate contexts

cmraible (approved):

PR #25946, opened 2026-01-22
Added support for member uuid substitution in emails

9larsons:
ref https://linear.app/ghost/issue/NY-927/
ref https://github.com/TryGhost/Product/issues/2383
- added token substitution for {uuid} for tracked links

Currently, {uuid} replacements work everywhere but in tracked links. This appears to have been restricted because of two reasons:
- we didn't fully wire up support for this with link click tracking, and chose to disable it rather than support it at the time
- we weren't previously clear on uuid being 'public' data, although we expose it in various places

This PR changes that to intentionally support replacements by removing the imposed blocks and adding support for substitution in tracked links via tokenization and replacement via the link click tracking service.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced link tracking to better handle personalized member links containing dynamic identifiers.
  * Improved redirect handling with proper substitution of member-specific identifiers in URLs.
  * Better encoding support for redirect URLs with special characters.


ghost/core/core/server/services/link-redirection/link-redirects-service.js
```diff
@@ -109,8 +113,29 @@ class LinkRedirectsService {
...
+                if (memberUuid) { // assume valid if present as m={uuid}
+                    redirectUrl = decodedUrl.split(MEMBER_UUID_PLACEHOLDER).join(memberUuid);
+                } else { // remove if no member UUID is present
+                    redirectUrl = decodedUrl.split(MEMBER_UUID_PLACEHOLDER).join('');
+                }
+            }
+
             res.setHeader('X-Robots-Tag', 'noindex, nofollow');
-            return res.redirect(link.to.href);
+            return res.redirect(redirectUrl);
```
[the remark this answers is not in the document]
9larsons: Unconcerned - this should be handled in the code.

ghost/core/test/e2e-browser/admin/publishing.spec.js
```diff
@@ -265,19 +265,15 @@ test.describe('Publishing', () => {
             await sharedPage.goto('/ghost');
             await createPage(sharedPage, pageData);
 
-            // Schedule the post to publish asap (by setting it to 00:00, it will get auto corrected to the minimum time possible - 5 seconds in the future)
-            await publishPost(sharedPage, {time: '00:00', type: null});
+            // Schedule the page (uses default 10 minutes in future)
+            await publishPost(sharedPage, {date: '2050-01-01', time: '12:00', type: null});
```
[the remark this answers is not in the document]
9larsons: This was mistakenly included here. Reverted.

ghost/core/test/e2e-browser/admin/publishing.spec.js
```diff
@@ -265,19 +265,15 @@ test.describe('Publishing', () => {
...
-
-            // Now wait for 5 seconds
-            await sharedPage.waitForTimeout(5000);
+            // Verify it shows as scheduled
+            await checkPostStatus(sharedPage, 'Scheduled');
 
-            // Check again, now it should have been added to the page
-            await checkPostPublished(sharedPage, pageData);
+            // Verify the page is NOT published yet
+            await checkPostNotPublished(sharedPage, pageData);
```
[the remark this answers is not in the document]
9larsons: This file is not in the set of changes.

ghost/core/core/server/services/email-service/email-renderer.js
```diff
@@ -443,10 +439,13 @@ class EmailRenderer {
...
-                // because otherwise we would add tracking to links that need to be replaced first
-                if (url.toString().indexOf('%%{uuid}%%') !== -1) {
-                    return url.toString();
+                // Skip tracking for links that contain replacement patterns other than %%{uuid}%%
+                // %%{uuid}%% is substituted at redirect time via the ?m= parameter, but other
+                // patterns (e.g. %%{key}%%) require direct Mailgun substitution
+                const urlStr = url.toString();
+                const withoutUuid = urlStr.replace(/%%\{uuid\}%%/g, '');
+                if (/%%\{[^}]+\}%%/.test(withoutUuid)) {
+                    return urlStr;
```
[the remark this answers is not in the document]
9larsons: In practice, this isn't a real issue. The originalPath check at lines 432-434 already handles the case where the entire href is a Mailgun variable (like %%{unsubscribe_url}%%). The code at 445-448 handles the case where a Mailgun variable appears inside a larger URL. The url object here comes from the link replacer which parses href attributes — and Mailgun variables embedded in otherwise-valid URLs would already be in their raw %%{...}%% form in the HTML, not percent-encoded, because the URL parser preserves them in the query string. The encoding scenario the bot describes (braces becoming %7B/%7D) would only happen if the placeholder were in the URL path, which isn't a realistic use case for Mailgun template variables in email links.

ghost/core/core/server/services/link-redirection/link-redirects-service.js
```diff
@@ -11,6 +11,13 @@ const LinkRedirect = require('./link-redirect');
  * @prop {(linkRedirect: LinkRedirect) => Promise<void>} save
  */
 
+// Placeholder pattern for member UUID in redirect destinations
+// %%{uuid}%% is substituted with the actual UUID at redirect time
+const MEMBER_UUID_PLACEHOLDER = '%%{uuid}%%';
+
+// UUID v4 pattern (8-4-4-4-12 hex format) for validating member UUIDs
+const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
```
EvanHahn, line 20: nit: I don't think this is quite right. v4 UUIDs have a few fixed characters. I _think_ [this regular expression](https://gist.github.com/johnelliott/cf77003f72f889abbc3f32785fa3df8d#file-uuidv4test-js-L7) is more correct, but you might want to check [the RFC](https://www.rfc-editor.org/rfc/rfc9562.html#name-uuid-version-4).

ghost/core/test/unit/server/services/link-redirection/link-redirects-service.test.js
```diff
@@ -136,5 +136,253 @@ describe('LinkRedirectsService', function () {
...
+            });
+            const req = {
+                originalUrl: 'no_r/prefix'
+            };
+            const res = {};
+            const next = sinon.fake();
+
+            await instance.handleRequest(req, res, next);
+
+            assert.equal(next.callCount, 1);
```
EvanHahn, line 157: nit, applies to this whole file: this is fine (and matches the rest of the file). But [Sinon's assertions](https://sinonjs.org/releases/v20/assertions/) offer clearer code and better error messages, if you want to use them.

ghost/core/test/unit/server/services/link-redirection/link-redirects-service.test.js
```diff
@@ -136,5 +136,253 @@ describe('LinkRedirectsService', function () {
...
+                }
+            };
+            const instance = new LinkRedirectsService({
+                linkRedirectRepository,
+                config: {
+                    baseURL: new URL('https://localhost:2368/')
+                }
+            });
+            const req = {
+                originalUrl: '/r/abc?m=some-uuid'
```
EvanHahn: nit: should this be a valid UUID?

ghost/core/test/unit/server/services/link-redirection/link-redirects-service.test.js
```diff
@@ -136,5 +136,253 @@ describe('LinkRedirectsService', function () {
...
+                redirect: sinon.fake(),
+                setHeader: sinon.fake()
+            };
+            await instance.handleRequest(req, res);
+            assert.equal(res.redirect.callCount, 1);
+            assert.equal(res.redirect.getCall(0).args[0], 'https://example.com/normal-link?foo=bar');
+        });
+
+        it('substitutes double-encoded %%{uuid}%% placeholder in query string', async function () {
+            // When stored as %25%25%7Buuid%7D%25%25, decodeURIComponent recovers %%{uuid}%%
```
EvanHahn: nit: I think it's clear without this comment.

```suggestion
```

EvanHahn (commented):
Partial review—I still need to dig into a few other things.

ghost/core/core/server/services/link-redirection/link-redirects-service.js
```diff
@@ -109,8 +116,36 @@ class LinkRedirectsService {
...
+            } catch (e) {
+                // decodeURIComponent fails on malformed sequences (e.g. %%%7Buuid%7D%%)
+                // Normalize URL-encoded braces so we can still detect the placeholder
+                decodedUrl = redirectUrl.replace(/%7B/gi, '{').replace(/%7D/gi, '}');
+            }
+            const hasMemberUuidPlaceholder = decodedUrl.includes(MEMBER_UUID_PLACEHOLDER);
+            if (hasMemberUuidPlaceholder) {
+                const memberUuid = url.searchParams.get('m');
+                if (memberUuid && UUID_REGEX.test(memberUuid)) {
+                    redirectUrl = decodedUrl.split(MEMBER_UUID_PLACEHOLDER).join(memberUuid);
```
EvanHahn: nit: Would [`String.prototype.replaceAll`][0] be clearer here?

```suggestion
                    redirectUrl = decodedUrl.replaceAll(MEMBER_UUID_PLACEHOLDER, memberUuid);
```

Same comment applies to the replacement below.

[0]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll

ghost/core/test/unit/server/services/email-service/email-renderer.test.js
```diff
@@ -1882,6 +1882,55 @@ describe('Email renderer', function () {
...
+                post,
+                newsletter,
+                segment,
+                options
+            );
+
+            // Verify tracking was called for the Transistor link
+            addTrackingToUrlStub.called.should.be.true();
+            const transistorCall = addTrackingToUrlStub.getCalls().find(call => call.args[0].href.includes('transistor.fm')
+            );
```
EvanHahn, line 1912: nit: formatting here is a little weird.

EvanHahn (approved):
A few non-blocking nits. LGTM overall.

ghost/core/core/server/services/link-redirection/link-redirects-service.js
9larsons: I've updated the comment. I don't think we need strict checking here - if a bad uuid is supplied, it won't match up, and it's not that worrisome. The purpose is some simple validation without blocking valid uuidv4.

PR #25947, opened 2026-01-22
Added database migration to store start/end of a discount

sagzy:
ref https://linear.app/ghost/issue/FEA-537

- As we introduce retention offers that can be applied to existing subscriptions, we need a reliable way to calculate whether the offer applies to the next payment
- instead of computing it ourselves, we're going to retrieve the discount start / end from Stripe and store it in the database

### DIFF
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -699,6 +699,8 @@ module.exports = {
         updated_at: {type: 'dateTime', nullable: true},
         mrr: {type: 'integer', unsigned: true, nullable: false, defaultTo: 0},
         offer_id: {type: 'string', maxlength: 24, nullable: true, unique: false, references: 'offers.id'},
+        discount_start: {type: 'dateTime', nullable: true},
+        discount_end: {type: 'dateTime', nullable: true},
         trial_start_at: {type: 'dateTime', nullable: true},
         trial_end_at: {type: 'dateTime', nullable: true},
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/data/schema/schema.js
```diff
@@ -698,6 +698,8 @@ module.exports = {
         updated_at: {type: 'dateTime', nullable: true},
         mrr: {type: 'integer', unsigned: true, nullable: false, defaultTo: 0},
         offer_id: {type: 'string', maxlength: 24, nullable: true, unique: false, references: 'offers.id'},
+        discount_start: {type: 'dateTime', nullable: true},
+        discount_end: {type: 'dateTime', nullable: true},
```
mike182uk, line 703: **suggestion (non-blocking):** `_at` suffix similar to to `trial_(start|end)_at`?

Looking at the schema file the convention is a bit mixed (i.e not all `dateTime` fields have `_at` suffix), although **_most_** fields do end in `_at`

sagzy: Yeh we do have a mix, with e.g. `current_period_end` versus `trial_end_at` on that same table

IMO the `_at` is useful when there is ambiguity without it: for example, `canceled_at` without `_at` would be read as boolean, not a datetime. With `_start`, `_end` a datetime is already expected.

Stripe also uses `discount.start` / `discount.end` without the extra `_at` ([docs](https://docs.stripe.com/api/discounts)) - so I'd vote for that!

PR #25948, opened 2026-01-22
Stored subscription discount start/end during Stripe sync

sagzy:
ref https://linear.app/ghost/issue/FEA-537

- As we introduce retention offers that can be applied to existing subscriptions, we need a reliable way to calculate whether the offer applies to the next payment
- instead of computing it ourselves, we retrieve it from Stripe and store it in our database

### DIFF
--- a/ghost/core/test/unit/server/services/members/members-api/repositories/member-repository.test.js
+++ b/ghost/core/test/unit/server/services/members/members-api/repositories/member-repository.test.js
@@ -685,6 +685,208 @@ describe('MemberRepository', function () {
             // Verify subscription was NOT created because the error was rethrown
             StripeCustomerSubscription.add.called.should.be.false();
         });
+
+        it('persists discount_start and discount_end for a new subscription', async function () {
+            const discountStart = 1768940436;
+            const discountEnd = 1784578836;
+
+            const subscriptionWithDiscount = {
+                ...subscriptionData,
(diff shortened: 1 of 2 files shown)


ghost/core/test/unit/server/services/members/members-api/repositories/member-repository.test.js
```diff
@@ -607,6 +607,142 @@ describe('MemberRepository', function () {
...
+                context: {}
+            });
+
+            StripeCustomerSubscription.add.calledOnce.should.be.true();
+            const addedSubscriptionData = StripeCustomerSubscription.add.firstCall.args[0];
+
+            assert.ok(addedSubscriptionData.discount_start instanceof Date);
+            assert.equal(addedSubscriptionData.discount_start.getTime(), discountStart * 1000);
+            assert.equal(addedSubscriptionData.discount_end, null);
+        });
```
mike182uk, line 889: **suggestion**: Should we also add a test for the `edit` path? I can see that the `subscriptionData` object is used for both paths - A test specifically aimed at the `edit` path would surface issues if the implementation changed (e.g., `linkSubscription` was refactored and forgot to include the fields for edits)

sagzy: Yes good one - added an additional new test

PR #25952, opened 2026-01-22
Added disableMemberCommenting labs flag

rob-ghost:
This PR adds the `disableMemberCommenting` private labs flag, which will gate the upcoming feature allowing staff to disable commenting for individual members.

The flag is being merged separately from the feature implementation to provide a clean integration point for testing the various components of this feature independently.


kevinansfield (approved):
There will be some failing snapshots for this change but as long as those are added, all looks good 👌

PR #25955, opened 2026-01-22
Release new admin shell and sidebar

jonatansberg:
closes https://linear.app/ghost/issue/BER-3230/release-new-admin-shell

Removed the feature flagging and conditionals so that we always serve the new admin shell. This includes the new sidebar UI and a significantly improved local developer experience via `yarn dev:forward` (soon to be renamed `yarn dev`.)


kevinansfield (approved):

cmraible:
Does this mean we can remove the duplicate e2e tests for Ember in CI now? No opinion on whether that should be included here or in a follow up, just curious and excited to simplify that 😄

jonatansberg:
Yeah, thats correct! I have a separate tasks for that since there are some additional cleanup that goes along with it.

ibalosh (approved):

PR #25957, opened 2026-01-22
Renamed dev:forward script to dev

jonatansberg:
closes https://linear.app/ghost/issue/BER-3231/promote-devforward-to-dev

The existing dev script has been renamed to `dev:legacy` and should be removed in the future. Similarly, there is an alias left in place for the `dev:forward` command that we can remove in the near future as well. 

The old `dev` command does not correctly start the new admin shell and should not be used other than as a temporary escape hatch.


kevinansfield (approved):
All looks good to me and worked well in testing 🙌

PR #25960, opened 2026-01-22
DRYed out signin return in Portal

EvanHahn:
towards https://linear.app/ghost/issue/NY-943/show-sniperlink-for-gmailcom-email-addresses-when-signing-up-for-an

We can shorten this code a bit. I think this is a useful change on its own, but also makes a future change a little easier.

---

> [!NOTE]
> Refactors `signin` flow in `apps/portal/src/actions.js` to consolidate the post-magic-link return object.
> 
> - Always returns `page: "magiclink"` and `lastPage: "signin"`, conditionally including `otcRef` only when `response.otc_ref` is present
> - Removes duplicated branching and keeps existing `pageData.email` behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3eea9b4a17f13acc9e72e863e6845712a25f2c4c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/portal/src/actions.js
+++ b/apps/portal/src/actions.js
@@ -88,22 +88,11 @@ async function signin({data, api, state}) {
             includeOTC: true
         };
         const response = await api.member.sendMagicLink(payload);
-
-        if (response?.otc_ref) {
-            return {
-                page: 'magiclink',
-                lastPage: 'signin',
-                otcRef: response.otc_ref,
-                pageData: {
-                    ...(state.pageData || {}),
-                    email: (data?.email || '').trim()
-                }
-            };
-        }
-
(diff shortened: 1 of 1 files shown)


apps/portal/src/actions.js
```diff
@@ -88,22 +88,11 @@ async function signin({data, api, state}) {
             includeOTC: true
         };
         const response = await api.member.sendMagicLink(payload);
-
-        if (response?.otc_ref) {
```
EvanHahn, line 92: I removed the `?` because `sendMagicLink` will always return a value.

PR #25961, opened 2026-01-22
Improved URLs logged after admin dev server loads

kevinansfield:
no issue

Nx shows the admin dev server in the bottom right pane when running `yarn dev` which leaves a prominent list of network URLs for the admin dev server in view once it finishes loading. Those network URLs are misleading because they are internal URLs rather than ones a developer should be using.

- added override of vite's `printUrls` method to output the real Ghost frontend and admin URLs instead of the internal dev server URLs

Before:
Screenshot 2026-01-22 at 18 07 50

After:
Screenshot 2026-01-22 at 18 11 53

### DIFF
--- a/apps/admin/vite.config.ts
+++ b/apps/admin/vite.config.ts
@@ -1,5 +1,5 @@
 import { resolve } from "path";
-import { defineConfig } from "vitest/config";
+import { createLogger, defineConfig } from "vite";
 import tsconfigPaths from "vite-tsconfig-paths";
 import react from "@vitejs/plugin-react-swc";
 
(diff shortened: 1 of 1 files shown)


apps/admin/vite.config.ts
```diff
@@ -8,6 +8,26 @@ import { ghostBackendProxyPlugin } from "./vite-backend-proxy";
...
+ */
+function ghostUrlPlugin() {
+    return {
+        name: 'ghost-url',
+        configureServer(server: { printUrls: () => void }) {
+            return () => {
+                server.printUrls = () => {
+                    const logger = createLogger();
+                    logger.info(`  \x1b[32m➜\x1b[0m  \x1b[1mGhost Frontend\x1b[0m: \x1b[36m${GHOST_URL}\x1b[0m`);
+                    logger.info(`  \x1b[32m➜\x1b[0m  \x1b[1mGhost Admin\x1b[0m:    \x1b[36m${GHOST_URL}ghost/\x1b[0m`);
```
jonatansberg, line 24: Should this use the proper urls from config, if we are using a different path?

jonatansberg (commented):
Overall, this looks good. My main concern is that we've previously run into issues when assuming things are always on the default paths. Shouldn't we read the correct URLs and post them here? We could potentially merge this URL printing thing into one of the existing plugins that already has that context.

apps/admin/vite.config.ts
kevinansfield: Doesn't that mean the `GHOST_URL` variable we use across vite config for subdirectory setups is also incorrect?

kevinansfield:
@9larsons I'm assuming this isn't much use with the work you're doing to remove the TUI so I'll close. Pinging just in case it's still relevant in some way.

9larsons:
@kevinansfield Ah this could be useful, but yeah, I think ultimately we may be looking to get rid of the nx tui.

PR #25966, opened 2026-01-22
Removed unused member permission code from comments API

rob-ghost:
## Summary

- Removed dead member permission code that was never enforced
- Simplified the public API bypass condition
- Cleaned up Comment model's `permissible()` function

## Context

The member permission system for comments contained several latent bugs that went undetected until we attempted to implement per-member comment restrictions:

1. **Public bypass was always true**: The condition `apiType !== 'comments'` checked for an API type that doesn't exist (Content API has `apiType: 'content'`, Members API has `apiType: 'members'`). This meant all public requests bypassed permission checks.

2. **providers.member() was a no-op**: It loaded ALL comment permissions for ALL members, so every member effectively had every permission.

3. **Comment.permissible() returned wrong type**: It returned `true` instead of `Promise.resolve()`, which could cause issues in the async permission chain.

Since the entire system was effectively a no-op, we've opted to remove the unused code rather than fix a system that was never enforcing anything.

Admin permission checks remain unchanged and functional.

### DIFF
--- a/ghost/core/core/server/api/endpoints/comments-members.js
+++ b/ghost/core/core/server/api/endpoints/comments-members.js
@@ -1,6 +1,5 @@
 const commentsService = require('../../services/comments');
 const ALLOWED_INCLUDES = ['member', 'replies', 'replies.member', 'replies.count.likes', 'replies.liked', 'count.replies', 'count.likes', 'liked', 'post', 'parent'];
-const UNSAFE_ATTRS = ['status'];
 
 /** @type {import('@tryghost/api-framework').Controller} */
 const controller = {
@@ -25,7 +24,7 @@ const controller = {
                 include: ALLOWED_INCLUDES
             }
         },
-        permissions: true,
+        permissions: false,
         query(frame) {
             return commentsService.controller.browse(frame);
         }
@@ -50,7 +49,7 @@ const controller = {
                 include: ALLOWED_INCLUDES
             }
         },
-        permissions: 'browse',
+        permissions: false,
         query(frame) {
             return commentsService.controller.replies(frame);
         }
@@ -72,7 +71,7 @@ const controller = {
                 include: ALLOWED_INCLUDES
             }
         },
-        permissions: true,
+        permissions: false,
         query(frame) {
             return commentsService.controller.read(frame);
         }
@@ -96,7 +95,7 @@ const controller = {
                 }
             }
         },
-        permissions: true,
+        permissions: false,
         query(frame) {
             return commentsService.controller.edit(frame);
         }
@@ -120,9 +119,7 @@ const controller = {
                 }
             }
         },
-        permissions: {
-            unsafeAttrs: UNSAFE_ATTRS
-        },
+        permissions: false,
         query(frame) {
             return commentsService.controller.add(frame);
         }
@@ -142,7 +139,7 @@ const controller = {
                 include: ALLOWED_INCLUDES
             }
         },
-        permissions: true,
+        permissions: false,
         query() {
             return commentsService.controller.destroy();
         }
@@ -171,7 +168,7 @@ const controller = {
         ],
         validation: {
         },
-        permissions: true,
+        permissions: false,
         async query(frame) {
             return await commentsService.controller.like(frame);
         }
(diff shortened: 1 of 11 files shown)


ghost/core/core/server/api/endpoints/utils/permissions.js
```diff
@@ -90,9 +90,9 @@ module.exports = {
         // @TODO: https://github.com/TryGhost/Ghost/issues/10099
         frame.options.context = permissions.parseContext(frame.options.context);
 
-        // CASE: Content API access
-        if (frame.options.context.public && frame.apiType !== 'comments') {
```
rob-ghost, line 94: The `apiType: 'comments'` didn't exist in the code, this was always `true`. Removed for clarity,

ghost/core/core/server/api/endpoints/comments-members.js
```diff
@@ -25,7 +24,7 @@ const controller = {
                 include: ALLOWED_INCLUDES
             }
         },
-        permissions: true,
+        permissions: false,
```
rob-ghost, line 27: Since we're removing the members permission checks, which always resolved in success, we don't need permission checks on these endpoints anymore.

We have to keep `permissions` as its a required field on controllers, so `permissions: false`.

ghost/core/core/server/api/endpoints/comments-members.js
```diff
@@ -50,7 +49,7 @@ const controller = {
                 include: ALLOWED_INCLUDES
             }
         },
-        permissions: 'browse',
+        permissions: false,
```
rob-ghost, line 52: Worth noting that `permissions: 'browse'` is actually an invalid way to configure permissions, and the JSDoc type for the controller would catch this if it was typechecked.

However, we're only type-checking TS files, and not JS files, and we're not type-checking JSDoc comments. I'd turn these checks on, but the blast radius is probably huge right now and out of scope of this change.

ghost/core/core/server/models/comment.js
```diff
@@ -170,47 +168,14 @@ const Comment = ghostBookshelf.Model.extend({
...
-            return Promise.reject(new errors.NoPermissionError({
-                message: tpl(messages.notYourCommentToEdit)
-            }));
-        }
-
-        if (action === 'destroy' && commentModelOrId.get('member_id') !== context.member.id) {
-            return Promise.reject(new errors.NoPermissionError({
-                message: tpl(messages.notYourCommentToDestroy)
-            }));
+            return Promise.resolve();
```
rob-ghost: All the user permission checks were duplicated here and in the user provider so these were redundant and removed.

This function was also returning a `bool` when we were expecting a resolve or a reject. Resolving with `false` is still considered a success, so we were "failing open" when attempting to deny a member from using a comment action.

We need to keep this function for the admin/user permission checks.

rob-ghost: Turns out this whole `permissible` was redundant

ghost/core/core/server/api/endpoints/comments-members.js
```diff
@@ -25,7 +24,7 @@ const controller = {
                 include: ALLOWED_INCLUDES
             }
         },
-        permissions: true,
+        permissions: false,
         query(frame) {
```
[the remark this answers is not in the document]
rob-ghost: The service handles this case and filters based on status so there's no risk of field injection.

ghost/core/test/e2e-api/members-comments/__snapshots__/comments.test.js.snap
```diff
@@ -747,6 +747,36 @@ Object {
 }
 `;
 
+exports[`Comments API when commenting enabled for all when authenticated Can not delete a comment which does not belong to you 1: [body] 1`] = `
+Object {
+  "errors": Array [
+    Object {
+      "code": null,
+      "context": "Unable to find member",
```
kevinansfield, line 755: Thought there might be something wrong here, but it's a known issue https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/services/comments/comments-service.js#L406-L409

kevinansfield (approved):
Agree that deleting and removing a source of confusion is preferable to fixing/re-implementing at this stage 👍

PR #25967, opened 2026-01-22
Decouple e2e tests from environment

cmraible:
Note: this is a very exploratory work-in-progress

The ultimate goal here is to get our E2E test suite working natively with the `yarn dev:forward` experience to make local development and testing much easier. The moonshot goal would be to allow running the e2e test suites against our staging environment or really _any_ running Ghost site, regardless of where and how it's being run.

This PR is a first attempt to decouple the E2E tests themselves from the environment more with a cleaner interface. This isn't quite right just yet, but feels like a step in the right direction.

### DIFF
--- a/e2e/helpers/environment/types.ts
+++ b/e2e/helpers/environment/types.ts
@@ -1,0 +1,98 @@
+/**
+ * Provider-agnostic types for e2e test environment management
+ *
+ * These interfaces decouple tests from specific environment implementations
+ * (Docker, external Ghost, dev:forward, etc.)
+ */
+
+/**
+ * Provider-agnostic representation of a test context
+ * No Docker-specific fields - just what tests need
+ */
+export interface TestContext {
+    /** Unique identifier for this test context */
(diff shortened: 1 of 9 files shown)


e2e/helpers/environment/types.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * URLs for services that tests may need to interact with
+ */
+export interface ServiceUrls {
+    ghost: string;
+    /** Mailpit web UI URL */
+    mailpit?: string;
+    /** Mailpit API URL */
+    mailpitApi?: string;
```
cmraible, line 31: Ideally we could make this more generic so we're not coupled to mailpit. I think we'd need to create an interface for "MailProvider" or something along those lines, and mailpit would be a concrete implementation thereof.

cmraible:
Closing this as we've made some progress on this separately in being able to run against the new `yarn dev` environment.

PR #25968, opened 2026-01-22
Added sniperlink when signing up

EvanHahn:
ref https://linear.app/ghost/issue/NY-943

When you sign up with a `@gmail.com` email, the "Now check your email!" has an updated button which takes you to `mail.google.com`.

https://github.com/user-attachments/assets/c961e86a-46cc-4880-9dca-99c13f462be8

*Note: I stubbed real email sending, so this screencast takes me to a bogus place in Gmail.*

In the future, we plan to show this link in other parts of the portal.


apps/portal/src/components/common/action-button.js
```diff
@@ -82,7 +82,7 @@ const Styles = ({brandColor, disabled, style = {}, isPrimary}) => {
 };
 
 function ActionButton({
```
EvanHahn, line 86: I wanted `<ActionButton>` to be an `<a>` tag underneath, so I made some tweaks to allow that.

apps/portal/src/utils/api.js
```diff
@@ -265,6 +265,15 @@ function setupGhostApi({siteUrl = window.location.origin, apiUrl, apiKey}) {
             }
         },
 
+        /**
+         * @returns {{
```
EvanHahn, line 269: A step toward type checking for this function.

ghost/core/core/server/lib/get-sniper-links.ts
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Extract the domain from an email address. Returns `null` if it can't be extracted.
+ *
+ * This is naïve — it doesn't fully respect RFC 5321 — but covers most normal emails.
+ *
+ * @example
+ * parseEmailDomain('foo@example.com');
+ * // => 'example.com'
+ */
+function parseEmailDomain(email: string): null | string {
```
EvanHahn: This function may get reworked significantly if #25920 is merged.

ghost/core/core/server/lib/get-sniper-links.ts
```diff
@@ -0,0 +1,115 @@
...
+
+type GetLinkFn = (options: Readonly<{recipient: string; sender: string}>) => string;
+
+type Provider = {
+    domains: ReadonlyArray<string>;
+    getDesktopLink: GetLinkFn;
+    getAndroidLink: GetLinkFn;
+};
+
+const PROVIDERS: ReadonlyArray<Provider> = [
```
EvanHahn, line 61: Parts of this file, such as this `PROVIDERS` object, might seem a little odd—why have an array when we only have one? This is because we plan to support more providers soon.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -718,19 +722,28 @@ module.exports = class RouterController {
...
-
                 if (signIn.otcRef) {
-                    res.writeHead(201, {'Content-Type': 'application/json'});
-                    return res.end(JSON.stringify({otc_ref: signIn.otcRef}));
+                    resBody.otc_ref = signIn.otcRef;
                 }
             }
 
-            res.writeHead(201);
-            return res.end('Created.');
```
EvanHahn: Slight API change here.

apps/portal/src/components/common/action-button.js
9larsons, line 1: This could be a separate, very small change to help reduce the span of this PR.

ghost/core/core/server/lib/get-sniper-links.ts
9larsons, line 1: It would be nice to see the backend changes separate from Portal for this one. This is an awful lot of files (not including translations) and changes that'll be squashed to one commit.

apps/portal/src/components/pages/magic-link-page.js
```diff
@@ -285,15 +286,44 @@ export default class MagicLinkPage extends React.Component {
...
+        );
+    }
+
     render() {
-        const {otcRef} = this.context;
-        const showOTCForm = !!otcRef;
+        const {otcRef, sniperLinks} = this.context;
+
+        /** @type {ReactNode} */ let footer;
+        if (otcRef) {
```
9larsons: Have we validated this behavior, that we only show sniper links if there's no OTC? OTC goes to inbox anyways, so I'm a little confused here.

apps/portal/src/components/common/action-button.js
9larsons: nit: I'm not a huge fan of the polymorphism here. Portal has a lot of things I'd prefer to do different, and this feels reasonable for the moment with the use case.

ghost/core/core/server/lib/get-sniper-links.ts
9larsons: We are merging that, so should we revisit this?

9larsons (changes requested):
Generally some questions here. I'm fine with this PR on the whole, though with future changes it'd be helpful to break them up a bit more. I know this one got held up by the other PR w/ package location changes (framework v ghost) that didn't help.

ghost/core/core/server/lib/get-sniper-links.ts
EvanHahn: All backend stuff was removed from this PR, so this comment is no longer relevant.

ghost/core/core/server/lib/get-sniper-links.ts
EvanHahn: All backend stuff was removed from this PR, so this comment is no longer relevant.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
EvanHahn: All backend stuff was removed from this PR, so this comment is no longer relevant.

ghost/core/core/server/lib/get-sniper-links.ts
EvanHahn: Done in #25997.

apps/portal/src/components/common/action-button.js
EvanHahn: Reverted the polymorphism in 8c7ae5c8891b4b2e6ae7f962c8db0180bbd95c2e.

apps/portal/src/components/common/action-button.js
EvanHahn: Reverted the polymorphism. There's still a tiny change I made in this file, but I didn't think it was worth the PR overhead.

apps/portal/src/components/common/action-button.js
```diff
@@ -14,6 +14,10 @@ export const ActionButtonStyles = `
         border: none;
     }
 
+    .gh-portal-btn-primary {
```
EvanHahn, line 17: The changes in this file _could_ be a separate PR, but I didn't think it was worth the time.

There is no reason to set the font color with JavaScript as far as I could see, so I just set it with regular CSS. That's a useful improvement on its own but also makes it easier to use this class in `<SniperLinkButton>`.

apps/portal/src/utils/is-android-chrome.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @param {Readonly<{
+ *     userAgentData?: Readonly<{
+ *         platform?: string;
+ *     }>;
+ *     userAgent: string;
+ * }>} navigator
+ * @returns {boolean}
+ */
+export const isAndroidChrome = navigator => isAndroid(navigator) && isChrome(navigator);
```
EvanHahn, line 32: I realized these links only work on Android Chrome, so I changed this from `isAndroid` to `isAndroidChrome`. (The internal `isAndroid` function still remained.)

EvanHahn (commented):
Reworked this to be frontend-only.

9larsons (approved):
question: Why did we choose to only display OTCs if available, rather than both sniper links and OTC entry?

I have minor concerns the css change will cause issues.. somewhere.. however, it does make sense for it to be in css.

EvanHahn:
> question: Why did we choose to only display OTCs if available, rather than both sniper links and OTC entry?

That's the next thing I'll be working on!

PR #25970, opened 2026-01-23
Fixed timezone bug in browser tests

cmraible:
Test only change that shouldn't impact any user facing behavior.

## Summary

Fixed a timezone-related flaky test failure in the publishing browser tests.

## The Problem

The scheduling tests in `publishing.spec.js` were only setting `time: '00:00'` without specifying a date. The assumption was that midnight today would always be in the past, triggering the auto-correction to "5 seconds in the future."

However, this fails when the **UTC date differs from the local date**:

- **For Americas (UTC-8 to UTC-5)**: Tests fail roughly **4 PM to midnight local time**
  - Example: At 5 PM Pacific (UTC-8), UTC is already 1 AM the next day
  - If the scheduler uses the UTC date, "00:00" on that date is still in the future
  - No auto-correction happens → test fails expecting "in a few seconds"

## The Fix

Explicitly set `date: getTodayDateString()` along with `time: '00:00'`. The helper uses `DateTime.now()` to get the current system date, ensuring that "00:00 today" is always in the past regardless of timezone differences between the test runner and the server.

## Changes

- Added `getTodayDateString()` helper function
- Updated 4 test cases to explicitly pass today's date when scheduling posts

### DIFF
--- a/ghost/core/test/e2e-browser/admin/publishing.spec.js
+++ b/ghost/core/test/e2e-browser/admin/publishing.spec.js
@@ -9,7 +9,7 @@ const {createTier, createMember, createPostDraft, impersonateMember} = require('
  * Used for scheduling tests to ensure the date is always "today" regardless of time.
  */
 const getTodayDateString = () => {
-    return DateTime.now().toFormat('yyyy-MM-dd');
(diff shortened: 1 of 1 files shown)


cmraible:
This fix doesn't actually fix the timezone bug.

PR #25971, opened 2026-01-24
Modified assets hash to be SHA256 based

jloh:
Closes https://linear.app/ghost/issue/HKG-1574

Currently Ghost generates asset hashes based off a MD5 of the current time. Every time Ghost gets restarted or the theme is modified the assets hash gets cleared. This means if JS/CSS or other asset based files aren't modified they have to get re-cached by the end users browser or CDN sitting in front.

This changes the hash to be based on a SHA256 hash of the underlying file where possible. This means that hashes are more static to improve caching on all layers.

A hash length of 16 was chosen to make it more easier to spot vs the current hash implementation and to allow a little lower chance of a collision.

### DIFF
--- a/ghost/core/core/frontend/meta/asset-url.js
+++ b/ghost/core/core/frontend/meta/asset-url.js
@@ -1,8 +1,11 @@
 const crypto = require('crypto');
+const path = require('path');
 const config = require('../../shared/config');
 const {blogIcon} = require('../../server/lib/image');
 const urlUtils = require('../../shared/url-utils');
 const {SafeString} = require('../services/handlebars');
+const assetHash = require('../services/asset-hash');
+const themeEngine = require('../services/theme-engine');
 
 /**
  * Serve either uploaded favicon or default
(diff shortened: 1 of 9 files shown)


ghost/core/core/frontend/meta/asset-url.js
```diff
@@ -12,10 +15,72 @@ function getFaviconUrl() {
...
+    if (builtFilePath) {
+        const builtPath = pathModule.join(builtFilePath, filename);
+        const hash = assetHash.getHashForFile(builtPath);
+        if (hash) {
+            return hash;
+        }
+    }
+
+    return null;
+}
```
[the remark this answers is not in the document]
jloh: Not applicable. Theme file paths are validated at upload time through multiple layers:
   - `extract-zip` library uses `realpath()` + relative path verification to prevent extraction outside the destination directory
   - `@tryghost/zip` blocks symlinks and excessively long filenames  
   - Theme names are sanitized (special chars replaced with hyphens)
   - GScan performs structural validation before activation
   
   The paths reaching `getAssetUrl` come from theme Handlebars templates, not user input, and those themes have already passed extraction-time security checks.

ghost/core/core/frontend/services/asset-hash/index.js
```diff
@@ -0,0 +1,74 @@
...
+        const mtimeMs = stat.mtimeMs;
+
+        // Check cache
+        const cached = hashCache.get(filePath);
+        if (cached && cached.mtimeMs === mtimeMs) {
+            return cached.hash;
+        }
+
+        // Read file and compute hash
+        const content = fs.readFileSync(filePath);
```
vershwal, line 37: The sync `fs.statSync()` + `fs.readFileSync()` calls will make the first request after boot slower for themes with many assets.

Not a blocker for this PR, but we could consider optimizing this. One approach could be pre-computing asset hashes eagerly during theme activation (in `mount()` or `activateTheme()`). This would populate the cache before the first user request arrives, moving the I/O cost to the boot/activation phase rather than the request path.

ghost/core/core/frontend/services/asset-hash/index.js
```diff
@@ -0,0 +1,74 @@
...
+ * Provides SHA256-based hashing for theme asset files.
+ * Hashes are cached in memory and invalidated when file mtime changes.
+ */
+
+// Hash length: 16 hex characters (64 bits) for low collision probability
+// This is longer than the legacy 10-char global hash, making it easy to distinguish
+const HASH_LENGTH = 16;
+
+// Cache structure: { filePath: { hash: string, mtimeMs: number } }
+const hashCache = new Map();
```
vershwal, line 18: Should we consider adding a size limit to this in-memory cache? Could act as a safeguard for unusually large themes (e.g., themes with large icon packs).

ghost/core/core/frontend/services/asset-hash/index.js
jloh: Hmmm yeah I hadn't considered that it'd slow the first render. Could be quite significant if the initial page being hit has a load of asset helpers as well. I don't know what the best approach would be here, would stopping boot be preferable to a slow initial serving of a page? I guess a slower initial serving would at least allow Admin to be served?

Then again if Ghost can't `fs.stat` a file we don't want it to block the frontend loading either 🤔

ghost/core/core/frontend/services/asset-hash/index.js
jloh: Yeah thats a good call. What is a reasonable size you think?

ghost/core/core/frontend/meta/asset-url.js
```diff
@@ -1,8 +1,11 @@
 const crypto = require('crypto');
+const pathModule = require('path');
```
allouis: I think it's pretty standard for this to be called `path` - if there's conflicts with other variables I would prefer them to be renamed to e.g. `pathToCheck` or `imagePath` or whatever...

ghost/core/core/frontend/services/asset-hash/index.js
```diff
@@ -0,0 +1,74 @@
...
+        const cached = hashCache.get(filePath);
+        if (cached && cached.mtimeMs === mtimeMs) {
+            return cached.hash;
+        }
+
+        // Read file and compute hash
+        const content = fs.readFileSync(filePath);
+        const hash = crypto.createHash('sha256')
+            .update(content)
+            .digest('hex')
```
allouis: Given we're truncating the string `'base64url'` encoding would decrease risk of collision here

ghost/core/core/frontend/services/asset-hash/index.js
allouis: Great point! I wonder if we can ship this behind either a labs flag, or a config flag so we can get it out, and then see how it affects sites in staging/prod?

ghost/core/core/frontend/services/asset-hash/index.js
jloh: Have moved to that!

ghost/core/core/frontend/services/asset-hash/index.js
vershwal: Node can easily handle millions of small Map entries. So this is mainly a safeguard against bugs not large themes and we should almost never hit it in practice.
Something like 100,000 entries feels reasonable. It would be ~25MB at most(Claude did this calculation for me) and still allowing even the largest themes to work without restriction.

allouis (approved):

PR #25983, opened 2026-01-26
🎨 Added disabled state UI for members with commenting disabled

rob-ghost:
## Why?

Staff members can now disable commenting for individual members (via the `disableMemberCommenting` labs flag). When a member has their commenting disabled, the comments UI needs to reflect this state appropriately.

## What does it do?

When a member has their commenting ability disabled:
- Shows a "Commenting disabled" message with a link to contact support instead of the comment form
- Hides the reply button entirely
- Disables the like button functionality  
- Hides the edit/delete menu on their own comments

The support email link uses the site's configured portal support email.


rob-ghost:
**Future thought:**

We have a few states where commenting isn't available:

1. Commenting is turned off for the publication
2. Commenting is members-only, visitor isn't a member
3. Commenting is paid-only, member is on free tier
4. Commenting is enabled but the member's ability to comment is disabled

These kind of boil down to:

- Can comment
- Can't comment (no upsell path)
- Can't comment but can be upsold

Right now the logic for determining this is spread across components. We could enumerate these as explicit states in the context (similar to how we now have `isCommentingDisabled`), then components just check the state rather than re-deriving it.

Something like:
```typescript
type CommentingState = 'CAN_COMMENT' | 'COMMENTING_OFF' | 'NEEDS_MEMBERSHIP' | 'NEEDS_PAID_TIER' | 'MEMBER_DISABLED';
```

apps/comments-ui/src/components/content/buttons/like-count.tsx
```diff
@@ -0,0 +1,26 @@
...
+                className={`mr-[6px] ${
+                    liked ? 'fill-black stroke-black dark:fill-white dark:stroke-white' : 'stroke-black/50 dark:stroke-white/60'
+                }`}
+            />
+            {count}
+        </span>
+    );
+};
+
+export default LikeCount;
```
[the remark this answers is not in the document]
rob-ghost: The comment is technically valid - using React.FC without importing React should cause a TypeScript error. However, this is a pre-existing pattern throughout the codebase - like-button.tsx and reply-button.tsx use the same pattern.

The tsconfig has "jsx": "react-jsx" (new JSX transform) which means React doesn't need to be imported for JSX, but for types like React.FC, you'd still typically need the import.

This pattern works because of how the bundler (Vite) and TypeScript are configured together - it's not strictly correct but is the established pattern in this codebase. Fixing only this file would be inconsistent with the rest of the codebase.

apps/comments-ui/src/components/content/comment.tsx
```diff
@@ -402,28 +403,41 @@ type CommentMenuProps = {
...
     highlightReplyButton: boolean;
     openEditMode: () => void;
-    parent?: Comment;
     className?: string;
 };
 const CommentMenu: React.FC<CommentMenuProps> = ({comment, openReplyForm, highlightReplyButton, openEditMode, className = ''}) => {
-    const {admin, t} = useAppContext();
+    const {admin, member, t, isMember, isCommentingDisabled} = useAppContext();
+
+    const isAdmin = !!admin;
```
kevinansfield: Could `isAdmin` be moved to app context too so it matches the new calculated `isMember` property?

apps/comments-ui/src/components/content/comment.tsx
```diff
@@ -402,28 +403,41 @@ type CommentMenuProps = {
...
-                {<MoreButton comment={comment} toggleEdit={openEditMode} />}
-            </div>
-        );
-    } else {
-        return (
-            <div className={`flex items-center gap-4 ${className}`}>
-                {<LikeButton comment={comment} />}
-                {<ReplyButton isReplying={highlightReplyButton} openReplyForm={openReplyForm} />}
-                {<MoreButton comment={comment} toggleEdit={openEditMode} />}
+                {showMoreButton && <MoreButton comment={comment} toggleEdit={openEditMode} />}
```
kevinansfield: Considering this is the admin view, does it need the `showMoreButton` conditional?

apps/comments-ui/src/components/content/comment.tsx
rob-ghost: Done, and I updated any boolean context references to `admin` with `isAdmin`

apps/comments-ui/src/components/content/comment.tsx
rob-ghost: Done!

apps/comments-ui/src/components/content/buttons/like-button.tsx
```diff
@@ -6,13 +6,11 @@ type Props = {
...
 const LikeButton: React.FC<Props> = ({comment}) => {
-    const {dispatchAction, member, commentsEnabled} = useAppContext();
+    const {dispatchAction, isMember, hasRequiredTier} = useAppContext();
     const [animationClass, setAnimation] = useState('');
     const [disabled, setDisabled] = useState(false);
 
-    const paidOnly = commentsEnabled === 'paid';
-    const isPaidMember = member && !!member.paid;
-    const canLike = member && (isPaidMember || !paidOnly);
+    const canLike = isMember && hasRequiredTier;
```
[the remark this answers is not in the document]
rob-ghost: `LikeButton` is only rendered when `isCommentingDisabled` otherwise we render a non-interactive `LikeLabel`

kevinansfield (approved):

PR #25985, opened 2026-01-26
Added commenting column to members table

rob-ghost:

## Summary

- Adds a nullable `commenting` column to the `members` table
- Stores per-member commenting permissions, allowing admins to disable commenting for specific members with an optional reason

## Why TEXT instead of JSON?

The column stores JSON data but uses `TEXT` type rather than `JSON` because:

1. **SQLite has no native JSON type** - it stores JSON as TEXT anyway, so `type: 'json'` would create inconsistent behavior between MySQL (native JSON with validation) and SQLite (TEXT without validation)

2. **Matches Ghost's established pattern** - all existing JSON-like data uses TEXT columns with manual `JSON.parse`/`JSON.stringify`:
   - `settings.value`
   - `sessions.session_data`  
   - `custom_theme_settings.value`

3. **Cross-database consistency** - TEXT behaves identically on both MySQL and SQLite

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/6.15/2026-01-26-12-00-00-add-commenting-column-to-members.js
+++ b/ghost/core/core/server/data/migrations/versions/6.15/2026-01-26-12-00-00-add-commenting-column-to-members.js
@@ -1,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('members', 'commenting', {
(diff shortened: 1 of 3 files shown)


jonatansberg (approved):
This looks good to me. I haven't really done much with the migrations though, so it might be worthwhile to wait for @kevinansfield's review as well.

kevinansfield (approved):
All looks good to me 👌

PR #25987, opened 2026-01-26
Use calculated `GHOST_URL` for dev commands

kevinansfield:
no issue

`apps/admin`'s vite config and the Stripe container both make use of the `GHOST_URL` env var but that requires you to manually specify it each time which is annoying and potentially conflicts with the URL that Ghost will actually be running on.

- added `ghost/core/bin/get-config` script that uses Ghost's existing nconf based config loader to write out the value of the specified config key
- updated `yarn dev` command to fetch and set `GHOST_URL` using the `get-config` script
- updated `yarn dev:*` commands to call `yarn dev` rather manually calling `nx` so we don't need to duplicate the `GHOST_URL` setup


jonatansberg (commented):
I'm not sure this is the right approach when looking at this again and digging into the usage of the GHOST_URL environment variable. I think its sole purpose is to inform Vite about where the Ghost backend lives so that we can resolve the correct URLs from the site endpoint.

Since that anyway redirects, I think maybe we don't even need to set it? We could likely update the vite-backend-proxy plugin to print the expected URLs since we already resolved them there.

kevinansfield:
For subdirectories there's no redirect if we don't hit the right path. We also use `GHOST_URL` in the Stripe container which was why I kept the variable and added the auto-set as part of the dev command.

Happy to learn about alternative approaches though!

jonatansberg:
I did see that too. that was added very recently… 

If we do want to set it, then maybe we should do it through a small script or something rather than inlining the config call? I don't have a strong idea for what better would look like, but I think parsing the config by invoking Node from the package script feels iffy.

kevinansfield:
Cleaned up the inline node call by adding `get-config` script to `ghost/core`. Initially I tried using the script via a `config:get` package.json script but the need to run it via `yarn workspace` from the root meant yarn was trashing the script output.

kevinansfield:
@jonatansberg another issue I've run into with subdir configs is that Caddy only proxies `/ghost/assets/*` but Admin will load `admin-x-settings` and `posts` from `/${subdir}/ghost/assets/*` causing them to 404.

PR #25992, opened 2026-01-26
Improved welcome email editor styles

peterzimon:
ref. https://linear.app/ghost/issue/NY-954/refine-welcome-email-styles

- Using serif font looked odd in the context of an email, especially that the welcome emails default to sans-serif.
- Email modal was a bit wider than necessary for the email body
- Dark mode needed a few tweaks


apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
```diff
@@ -49,7 +49,7 @@ const EmailPreview: React.FC<{
                 </div>
             </div>
             <Button
-                className='rounded-md border border-grey-200 font-semibold hover:shadow-xs'
+                className='border border-grey-200 font-semibold hover:border-grey-300 hover:!bg-white dark:border-grey-900 dark:hover:border-grey-800 dark:hover:!bg-grey-950'
```
troyciesco: i don't think the `!important` modifiers on these background hovers should be needed - are you seeing differently, or is there some edge case i'm not thinking of?

defining the `hover:bg-white` also doesn't seem to be needed, since the color prop is set to white and it's not a link

peterzimon: Good points, fixed

PR #25996, opened 2026-01-26
Added migration for Stripe billing portal configuration

sam-lord:
ref [FEA-502](https://linear.app/ghost/issue/FEA-502/portal-billing-management-converted-to-project)

Migration as originally implemented in PR: https://github.com/TryGhost/Ghost/pull/25862/

This will hold the id of the billing portal configuration object defined in Stripe. This way we can create our own billing portal configuration with any of the features we need, and override the default headline.


aileen (approved):
LGTM

PR #25997, opened 2026-01-26
Added utility for getting sniper links

EvanHahn:

This adds `getSniperLinks`, a function that basically turns an email address into a link to open the email app/URL. We'll use it soon.

It's missing two big features:

- Support for additional mail providers, like Outlook and Proton.
- Do DNS lookups for MX records, which can get more domains. For example, `ghost.org`'s MX records refer to Gmail.

### DIFF
--- a/ghost/core/core/server/lib/get-sniper-links.ts
+++ b/ghost/core/core/server/lib/get-sniper-links.ts
@@ -1,0 +1,97 @@
+/**
+ * This module is heavily modified from [Buttondown's Sniper Link project][0].
+ * Its license is copied below.
+ *
+ * [0]: https://github.com/buttondown/sniper-link
+ *
+ * MIT License
+ *
+ * Copyright (c) 2024 Buttondown
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/lib/get-sniper-links.ts
```diff
@@ -0,0 +1,97 @@
...
+    {
+        domains: ['gmail.com', 'googlemail.com', 'google.com'],
+        getDesktopLink: ({recipient, sender}) => (
+            `https://mail.google.com/mail/u/${encodeURIComponent(
+                recipient
+            )}/#search/from%3A(${encodeURIComponent(
+                sender
+            )})+in%3Aanywhere+newer_than%3A1h`
+        ),
+        getAndroidLink: () => (
```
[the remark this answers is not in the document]
EvanHahn: Inaccurate.

ghost/core/package.json
EvanHahn, line 1: Previously, we weren't importing `@tryghost/parse-email-address` from TypeScript. Now that we are, we need to build it.

ghost/core/core/server/lib/get-sniper-links.ts
EvanHahn, line 1: This file, and its tests, have not changed since #25968, except it now uses `parseEmailAddress` instead of a redundant email parsing function.

9larsons (approved):

PR #26005, opened 2026-01-26
Fixed performance tests in CI

EvanHahn:
Performance tests broke because they expected TypeScript projects to be built with `yarn build:tsc`, not `yarn build:ts`. This updates the `parse-email-address` package to match that.

Also lets you run perf tests with a new `perf-tests` label.


9larsons (approved):

9larsons:
Assuming CI passes, all good from me.

PR #26006, opened 2026-01-26
Changed welcome emails to only be gated by labs flag

cmraible:
ref https://linear.app/ghost/issue/NY-822/remove-the-memberwelcomeemailtestinbox-config-and-send-to-user

## Summary

Currently all welcome emails features are gated behind the presence of a `memberWelcomeEmailTestInbox` configuration parameter. This PR changes the behavior such that `memberWelcomeEmailTestInbox` can still be provided to override the email address welcome emails are sent to, but the lack of a value for this parameter will not disable welcome emails entirely.

Now, as long as the `welcomeEmails` labs flag is set, welcome emails should be enabled. The `memberWelcomeEmailTestInbox` only determines where the welcome emails are sent, rather than enabling/disabling the features entirely.

- Changed welcome email feature to be gated by `labs.welcomeEmails` flag instead of `memberWelcomeEmailTestInbox` config
- Made `memberWelcomeEmailTestInbox` an optional recipient override (falls back to member's email when not set)
- Added tests for recipient behavior (test inbox vs member email)

---

## Testing

Without `memberWelcomeEmailTestInbox` set, with labs flag set to true
- [x] Welcome emails UI should display in Admin
- [x] Free signup test, welcome email should be addressed to the member
- [x] Paid signup test, welcome email should be addressed to the member

With `memberWelcomeEmailTestInbox` set, with labs flag set to true
- [x] Welcome emails UI should display in Admin
- [x] Free signup test, welcome email should be addressed to the config value
- [x] Paid signup test, welcome email should be addressed to the config value

With `memberWelcomeEmailTestInbox` set, with labs flag set to false:
- [x] Welcome emails UI should not display in Admin
- [x] Free signup test, should not send welcome email
- [x] Paid signup test, should not send welcome email

Without `memberWelcomeEmailTestInbox` set, with labs flag set to false:
- [x] Welcome emails UI should not display in Admin
- [x] Free signup test, should not send welcome email
- [x] Paid signup test, should not send welcome email

### DIFF
--- a/e2e/tests/public/member-signup.test.ts
+++ b/e2e/tests/public/member-signup.test.ts
@@ -8,10 +8,13 @@ import {signupViaPortal} from '@/helpers/playwright/flows/signup';
 test.describe('Ghost Public - Member Signup', () => {
     let emailClient: EmailClient;
 
-    test.use({config: {
-        memberWelcomeEmailSendInstantly: 'true',
-        memberWelcomeEmailTestInbox: `test+welcome-email@ghost.org`
-    }});
+    test.use({
+        config: {
+            memberWelcomeEmailSendInstantly: 'true',
+            memberWelcomeEmailTestInbox: `test+welcome-email@ghost.org`
+        },
+        labs: {welcomeEmails: true}
+    });
 
     test.beforeEach(async () => {
         emailClient = new MailPit();
--- a/e2e/tests/public/member-signup.test.ts
+++ b/e2e/tests/public/member-signup.test.ts
@@ -8,13 +8,7 @@ import {signupViaPortal} from '@/helpers/playwright/flows/signup';
 test.describe('Ghost Public - Member Signup', () => {
     let emailClient: EmailClient;
 
-    test.use({
-        config: {
-            memberWelcomeEmailSendInstantly: 'true',
-            memberWelcomeEmailTestInbox: `test+welcome-email@ghost.org`
-        },
-        labs: {welcomeEmails: true}
-    });
+    test.use({labs: {welcomeEmails: true}});
 
     test.beforeEach(async () => {
         emailClient = new MailPit();
@@ -51,11 +45,10 @@ test.describe('Ghost Public - Member Signup', () => {
         expect(emailTextBody).toContain('complete the signup process');
     });
 
-    test('received welcome email', async ({page, config}) => {
+    test('received welcome email', async ({page}) => {
         const automatedEmailFactory = createAutomatedEmailFactory(page.request);
         await automatedEmailFactory.create();
 
-        const emailInbox = config!.memberWelcomeEmailTestInbox!;
         const homePage = new HomePage(page);
         await homePage.goto();
         const {emailAddress} = await signupViaPortal(page);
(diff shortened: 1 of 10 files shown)


ghost/core/test/integration/jobs/process-outbox.test.js
```diff
@@ -7,7 +7,7 @@ const models = require('../../../core/server/models');
 const {OUTBOX_STATUSES} = require('../../../core/server/models/outbox');
 const db = require('../../../core/server/data/db');
 const mailService = require('../../../core/server/services/mail');
-const config = require('../../../core/shared/config');
+const mockManager = require('../../utils/e2e-framework-mock-manager');
```
cmraible, line 10: The diff in this file looks bigger than it really is - definitely hide whitespace changes.

The jist of the change here is to create dedicated describe blocks for the labs flag on/off states

cmraible:
I tried to include a video for testing each of these scenarios, but it ended up being too large to upload to Github. Rest assured, I did manually test each of the scenarios in the description.

e2e/tests/public/member-signup.test.ts
```diff
@@ -8,10 +8,7 @@ import {signupViaPortal} from '@/helpers/playwright/flows/signup';
 test.describe('Ghost Public - Member Signup', () => {
     let emailClient: EmailClient;
 
-    test.use({config: {
-        memberWelcomeEmailSendInstantly: 'true',
```
cmraible, line 12: I have a follow up task to remove this configuration parameter entirely, since we now send instantly by default.

e2e/tests/public/member-signup.test.ts
```diff
@@ -65,7 +61,11 @@ test.describe('Ghost Public - Member Signup', () => {
         await publicPage.goto(magicLink);
         await homePage.waitUntilLoaded();
 
-        latestMessage = await retrieveLatestEmailMessage(emailInbox);
+        const welcomeMessages = await emailClient.search(
+            {to: emailAddress, subject: 'Welcome'},
+            {timeoutMs: 10000}
+        );
+        latestMessage = await emailClient.getMessageDetailed(welcomeMessages[0]);
```
troyciesco, line 68: I like this change 👍 seems less likely to have some weird timing issue down the road.

ghost/core/core/server/services/member-welcome-emails/service.js
```diff
@@ -77,12 +77,8 @@ class MemberWelcomeEmailService {
...
         });
 
-        const toEmail = config.get('memberWelcomeEmailTestInbox');
-        if (!toEmail) {
-            throw new errors.IncorrectUsageError({
-                message: MESSAGES.MISSING_TEST_INBOX_CONFIG
-            });
-        }
+        const testInbox = config.get('memberWelcomeEmailTestInbox');
+        const toEmail = testInbox || member.email;
```
troyciesco, line 81: Question: Do you think we should still have the defensive check here? 

It does seem like an unlikely scenario, though without it we could possibly pass it along to the mailer without one. Which does _also_ have a check, but the error would be "Incomplete message data" which isn't very clear.

troyciesco (approved):
One question i'll leave up to you (we can chat about if you want), otherwise LGTM!

ghost/core/core/server/services/member-welcome-emails/service.js
cmraible: Oh yeah, good call. Agree it's unlikely, but can't hurt to add it back in.

cmraible: Re-added!

PR #26007, opened 2026-01-26
Added MX record lookup for sniper links

EvanHahn:
closes https://linear.app/ghost/issue/NY-945
ref https://github.com/TryGhost/Ghost/pull/25997

We need the email provider, like Gmail or Proton, to create a sniper link. If we can't get the provider from the email address, we now fall back to looking up the [MX records] for the domain and look there.

For example, though `ghost.org` isn't in our hard-coded list, it includes Gmail domains in the MX records, so we'd show a Gmail sniper link.

[MX records]: https://en.wikipedia.org/wiki/MX_record

### DIFF
--- a/ghost/core/core/server/lib/get-sniper-links.ts
+++ b/ghost/core/core/server/lib/get-sniper-links.ts
@@ -27,7 +27,10 @@
  * SOFTWARE.
  */
 
+import type {MxRecord} from 'node:dns';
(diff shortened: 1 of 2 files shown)


9larsons (approved):
Nicely done. Feels very overbuilt, but not in a bad way 😄.

Also don't think I've seen `.shuffle` used before, that's a neat trick for tests.

PR #26008, opened 2026-01-27
Moved member email editor styles

troyciesco:
no ref

This moves the classes around so the container-related styles are in `welcome-email-modal.tsx`, and the editor related styles (which are, effectively, koenig overwrites) into `member-email-editor.tsx`. 

Also tries to make things just a bit easier to read, because multiple `[&_.koenig-lexical-editor-input-placeholder]` selectors creates a lot of noise.

I did try moving things out into an external stylesheet, but it required importing it and i found it harder to reason about where the styles were and what was getting applied.

Screenshot 2026-01-26 at 7 05 43 PM


troyciesco:
closed in favor of https://github.com/TryGhost/Ghost/pull/25992

PR #26009, opened 2026-01-27
Fix flaky location filter test by configuring analytics tracking

cmraible:
## Why are you making it?

The location filter test was previously skipped due to flakiness in CI. The root cause was that the synthetic traffic generated for testing wasn't being properly tracked by Ghost's analytics system, likely because webdriver detection was preventing tracking.

## What does it do?

This change re-enables the previously skipped test and fixes the flakiness by:

1. **Creating a dedicated browser context** with proper configuration instead of using `withIsolatedPage`
2. **Setting timezone to America/New_York** to ensure consistent location data from Ghost's timezone mapping
3. **Injecting `__GHOST_SYNTHETIC_MONITORING__` flag** via `addInitScript` to bypass webdriver detection in the Ghost stats script, allowing analytics to be properly tracked during test execution
4. **Properly managing context lifecycle** with try/finally to ensure resources are cleaned up

## Why is this something Ghost users or developers need?

This fix ensures that the location filtering feature in post analytics is properly tested and reliable. By re-enabling this test with proper analytics tracking configuration, we can catch regressions in location-based filtering functionality and maintain confidence in the analytics feature's correctness.

## Checklist

- [x] I've written an automated test to prove my change works (re-enabled previously skipped test with proper configuration)

https://claude.ai/code/session_015umCpcHA54U9gqEcCknQU7

### DIFF
--- a/e2e/tests/admin/analytics/post-analytics/web-filters.test.ts
+++ b/e2e/tests/admin/analytics/post-analytics/web-filters.test.ts
@@ -124,13 +124,26 @@ test.describe('Ghost Admin - Post Analytics Web Filters', () => {
             await expect(postAnalyticsPage.getActiveFilter('Source')).toBeHidden();
         });
 
-        // TODO: This is flaky on CI, so we're skipping it for now.
-        test.skip('click on location row adds location filter', async ({page, browser, baseURL}) => {
-            // Generate traffic to the post
-            await withIsolatedPage(browser, {baseURL}, async ({page: publicPage}) => {
+        test('click on location row adds location filter', async ({page, browser, baseURL}) => {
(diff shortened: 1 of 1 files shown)


ibalosh:
hey @cmraible , I see you closed this one, did it not work? curious about this update

cmraible:
@igor this PR was mostly an experiment with Claude code's Slack app ([context](<https://ghost.slack.com/archives/C02G9E68C/p1769472161027849>)). It may indeed work, but Claude ended up working on a different test than I intended it to, so I just closed the PR. Feel free to reopen though if you're interested, and what Claude came up with here looks good to you!

PR #26011, opened 2026-01-27
Added support for backporting Portal changes to 5.x

sagzy:
ref INC-232

- we currently don't have support for publishing packages such as Portal for 5.x users
- with this change, we check if the current version from the 5.x branch exists on NPM. If not, we publish it under the tag `ghost-5x` - so that it doesn't get tagged as `latest`, which is the default


hadret:
This LGTM 👍🏻

kevinansfield (approved):

PR #26015, opened 2026-01-27
Fixed issue where members could be enumerated

sam-lord:
ref [GVA-647](https://linear.app/ghost/issue/GVA-647/member-email-enumeration-during-signin)

So the changes here:
* If you're registering to a site: No change (this already automatically sends a sign-in email if you have an account)
* If you're signing into a site and do have account: No change
* If you're signing in to a site, but don't have an account, and the site is private: You'll get an email explaining that you don't have an account and the site is private
* If you're signing into a site, but don't have an account, and the site is public: You'll get a sign-up email with magic link

This also meant adding the blocked domains feature to the sign-in endpoint, but only when it's being used to create an account.

Screenshot 2026-01-27 at 15 58 19

This is the attempted sign-in email.

---

> [!NOTE]
> **Medium Risk**
> Changes authentication-adjacent magic-link behavior and email sending, which could impact signup/signin flows and rate-limiting/blocked-domain edge cases if misconfigured.
> 
> **Overview**
> Prevents member email enumeration during `send-magic-link` sign-in by **no longer returning a 400 when the email doesn’t exist**. Instead, when self-signup is enabled the API sends a *signup* magic-link email (without OTC) and returns an empty `201`, while invite-only sites silently return an empty `201` with no email.
> 
> Adds blocked-domain handling to this “signin-but-create-account” path (silently drops blocked domains), updates Portal copy to show a non-revealing invite-only message, and standardizes “e-mail” → “email” strings across errors/tests/locales.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b175ec741a116fdcb56b3a0890cddaabd576d953. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Sends signup magic-link emails for unknown addresses when self-signup is enabled; invite-only sites remain silent.
  * Portal shows quick "open email"/provider links and new sign-in informational copy.

* **Bug Fixes**
  * Prevents account enumeration in sign-in/sign-up flows.
  * Enforces blocked email-domain handling and consistent OTC behavior for non-existent members.

* **Documentation & Localization**
  * Standardized "e-mail" → "email" across portal translations; added new sign-in message.


ghost/core/core/server/services/members/emails/signin-restricted.js
```diff
@@ -0,0 +1,131 @@
+module.exports = ({t, siteTitle, email, siteDomain, siteUrl}) => `
+<!doctype html>
+<html>
+  <head>
+    <meta name="viewport" content="width=device-width">
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+    <meta name="color-scheme" content="light only">
+    <meta name="supported-color-schemes" content="light">
+    <title>${t('Sign in attempt for {siteTitle}', {siteTitle, interpolation: {escapeValue: false}})}</title>
```
aileen: [nitpick] Most of the other emails use an emoji in the title, except for the confirm email one. Not an issue, just wanted to flag it up.

ghost/core/core/server/services/members/emails/signin-restricted.js
```diff
@@ -0,0 +1,131 @@
...
+      margin-bottom: 34px;
+      border-bottom-width: 1px;
+      border-bottom-color: #EEF5F8;
+    }
+    a {
+      color: #3A464C;
+    }
+    </style>
+  </head>
+  <body style="background-color: #FFFFFF; font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'; -webkit-font-smoothing: antialiased; font-size: 14px; line-height: 1.5em; margin: 0; padding: 0; -ms-text-size-adjust: 100%; -webkit-text-size-adjust: 100%;">
```
aileen: All the other email templates follow the pattern of having comments for the different sections. Email HTML is very difficult to read and I think adding those comments and better intendation will improve the readability a lot. See [here](https://github.com/TryGhost/Ghost/blob/1d0def6fe24756cdbfd86e38e387c9f87bbf9437/ghost/core/core/server/services/members/emails/update-email.js#L104-L174)

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -849,8 +849,36 @@ module.exports = class RouterController {
...
-                message: this._allowSelfSignup() ? tpl(messages.memberNotFoundSignUp) : tpl(messages.memberNotFound)
+            // Member doesn't exist - to prevent enumeration, we don't reveal this
+            // If self-signup is allowed, send a signup email so they can create an account
+            // If self-signup is disabled, send an informational email explaining no account exists
+            if (this._allowSelfSignup()) {
+                const blockedEmailDomains = this._settingsCache.get('all_blocked_email_domains');
+                const emailDomain = normalizedEmail.split('@')[1]?.toLowerCase();
+                if (emailDomain && blockedEmailDomains.includes(emailDomain)) {
+                    // To prevent enumeration, we don't reveal this
+                    return {};
```
aileen, line 880: [question] What happens in the UI in this case?

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -849,8 +849,36 @@ module.exports = class RouterController {
...
+                return await this._sendEmailWithMagicLink({
+                    email: normalizedEmail,
+                    tokenData,
+                    requestedType: 'signup',
+                    referrer
+                });
+            }
+
+            // Self-signup disabled: send informational email (no magic link)
+            return await this._sendEmailWithMagicLink({
```
aileen: [question] I assume there's no other method for sending the email without the magic link? If you don't read the comment, you would assume that we're sending the magic link because of the `_sendEmailWithMagicLink` method 🤔

ghost/core/test/e2e-api/members/send-magic-link.test.js
```diff
@@ -43,42 +43,64 @@ describe('sendMagicLink', function () {
...
+            to: email,
+            subject: /Sign in attempt for Ghost/
+        });
+
+        // Verify the email explains the situation
+        assert.ok(mail.text.includes('no account exists with this email'));
+        assert.ok(mail.text.includes('invite-only'));
+
+        // Verify there's no magic link in the email (it's just informational)
+        assert.ok(!mail.text.includes('?token='));
```
aileen: [nitpick] Would this be a good place for an assertion message? e.g. `assert.ok(!mail.text.includes('?token='), 'No magic link was sent')`?

ghost/core/test/e2e-api/members/send-magic-link.test.js
```diff
@@ -424,6 +446,26 @@ describe('sendMagicLink', function () {
...
+                        emailType: 'signin'
+                    })
+                    .expectEmptyBody()
+                    .expectStatus(201);
+
+                // Verify no email was sent (prevents enumeration)
+                // sentEmail will throw if no email was sent
+                assert.throws(() => {
+                    mockManager.assert.sentEmail({to: email});
+                }, /Expected at least 1 emails sent/);
```
aileen, line 476: I find this comment confusing wrt to the comment above, but maybe that's just me?

ghost/i18n/locales/af/ghost.json
```diff
@@ -31,6 +32,7 @@
     "Name": "",
     "New comment on {postTitle}": "",
     "New reply to your comment on {siteTitle}": "",
+    "No member exists with this e-mail address.": "",
```
aileen: I agree with the agent review: this should be `email` and not `e-mail`?

aileen (approved):
A few comments, but non are blocking.

ghost/core/core/server/services/members/emails/signin-restricted.js
sam-lord: Yeah I don't think this deserves an emoji. It's a pretty specific email, considering it's only sent if you:

* Click to sign-in on a private site
* Give an email address that doesn't have an account

I don't want to copy the emoji from another email (this isn't a 🙌 situation, it doesn't have a 🔑 to sign-in) -- this is a less interesting email.

I will request a copy review from Peter though!

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
sam-lord: Image

Same as any other action :)

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
sam-lord: It's a bit weird, but `sendMagicLink` is just the holder for the functionality of sending an email to a potential member. The email itself definitely belongs with the others there, and it would be unusual to need to add another dependency here just to separate the magic link emails from the not-magic-link-but-still-relating-to-signup emails.

I think the longer term modification here would be to rename the magic link module (and the `sendEmailWithMagicLink` that comes with it) to something that more accurately reflects that this is a way of sending emails to members relating to signup/signin.

ghost/i18n/locales/af/ghost.json
sam-lord: I copied it from here: https://github.com/TryGhost/Ghost/blob/1d0def6fe24756cdbfd86e38e387c9f87bbf9437/ghost/core/core/server/services/members/members-api/controllers/router-controller.js#L23

I don't disagree though, we should be more consistent in e-mail vs email!

I'll change it 👍

ghost/core/test/e2e-api/members/send-magic-link.test.js
sam-lord: Good shout, removed comment & added assertion message for this one

PR #26022, opened 2026-01-27
Updated magic links endpoint to return sniper links

EvanHahn:
ref https://linear.app/ghost/issue/NY-943
ref https://github.com/TryGhost/Ghost/pull/25968

This updates the `send-magic-links` endpoint to return sniper links.

This change should have no user impact yet, but will soon once the frontend supports it.

### DIFF
--- a/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
+++ b/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
@@ -1,3 +1,4 @@
+const dns = require('node:dns/promises');
 const tpl = require('@tryghost/tpl');
 const logging = require('@tryghost/logging');
 const sanitizeHtml = require('sanitize-html');
@@ -5,6 +6,7 @@ const {BadRequestError, NoPermissionError, UnauthorizedError, DisabledFeatureErr
 const errors = require('@tryghost/errors');
 const {isEmail} = require('@tryghost/validator');
 const normalizeEmail = require('../utils/normalize-email');
+const {getSniperLinks} = require('../../../../lib/get-sniper-links');
 
 const messages = {
     emailRequired: 'Email is required.',
@@ -51,6 +53,8 @@ function extractRefererOrRedirect(req) {
 }
 
 module.exports = class RouterController {
+    #sniperLinksDnsResolver = new dns.Resolver({maxTimeout: 1000});
+
     /**
      * RouterController
      *
@@ -69,6 +73,7 @@ module.exports = class RouterController {
      * @param {any} deps.newslettersService
      * @param {any} deps.sentry
      * @param {any} deps.settingsCache
+     * @param {any} deps.settingsHelpers
      * @param {any} deps.urlUtils
      */
     constructor({
@@ -87,6 +92,7 @@ module.exports = class RouterController {
         newslettersService,
         sentry,
         settingsCache,
+        settingsHelpers,
         urlUtils
     }) {
         this._offersAPI = offersAPI;
@@ -104,6 +110,7 @@ module.exports = class RouterController {
         this._newslettersService = newslettersService;
         this._sentry = sentry || undefined;
         this._settingsCache = settingsCache;
+        this._settingsHelpers = settingsHelpers;
         this._urlUtils = urlUtils;
     }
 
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -712,19 +719,29 @@ module.exports = class RouterController {
...
+                    resBody.otc_ref = signIn.otcRef;
                 }
             }
 
-            res.writeHead(201);
-            return res.end('Created.');
+            const sniperLinks = await getSniperLinks({
+                recipient: normalizedEmail,
+                sender: this._settingsHelpers.getMembersSupportAddress(),
+                dnsResolver: this.#sniperLinksDnsResolver
```
EvanHahn, line 737: The new `dnsResolver` option is the only difference between this PR and what's in #25968.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -698,7 +705,7 @@ module.exports = class RouterController {
             // Honeypot field is filled, this is a bot.
             // Pretend that the email was sent successfully.
             res.writeHead(201);
-            return res.end('Created.');
```
9larsons, line 701: issue: Why did this change? It feels wrong to change the response here, even for a bot/honeypot response.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -712,19 +719,29 @@ module.exports = class RouterController {
...
-            res.writeHead(201);
-            return res.end('Created.');
+            const sniperLinks = await getSniperLinks({
+                recipient: normalizedEmail,
+                sender: this._settingsHelpers.getMembersSupportAddress(),
+                dnsResolver: this.#sniperLinksDnsResolver
+            });
+            if (sniperLinks) {
+                resBody.sniperLinks = sniperLinks;
+            }
```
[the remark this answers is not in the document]
9larsons: Reasonable, but already in a try catch block.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
9larsons: suggestion: While helpful for testing, it feels awkward for the calling code to need to instantiate and pass in the dns resolver. Would it make sense to have a default?

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
9larsons, line 1: note: This is all w/o a flag, which is what we discussed/suggested previously. We'll control the frontend display only in order to simplify behavior.

9larsons (changes requested):
One main question.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
EvanHahn: This is a stylistic choice.

In addition to being useful for testing, it also tells callers that this function might do a DNS lookup, and [it keeps the function reasonably pure][0]. This is also similar to the dependency injection I've spotted elsewhere in the codebase.

If you feel strongly, I'm happy to change it.

[0]: https://github.com/google/wuffs/blob/072595ae7df55544874caf5096cd9ef4ecf1bdbb/doc/note/hermeticity.md

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
EvanHahn: To match non-honeypot responses.

I don't want bots to know they're being detected. Because we changed the non-honeypot response, I think we also need to change the honeypot one.

9larsons (approved):

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -698,7 +705,7 @@ module.exports = class RouterController {
             // Honeypot field is filled, this is a bot.
             // Pretend that the email was sent successfully.
             res.writeHead(201);
-            return res.end('Created.');
+            return res.end('{}');
```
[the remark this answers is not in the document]
EvanHahn: Missed this. Addressed in #26027.

PR #26023, opened 2026-01-27
Added more providers to sniper link detection

EvanHahn:
closes https://linear.app/ghost/issue/NY-944

Before this change, `getSniperLinks()` only supported Gmail. Now, it supports Yahoo, Outlook, Proton, iCloud, Hey, AOL, and Mail.ru.

### DIFF
--- a/ghost/core/core/server/lib/get-sniper-links.ts
+++ b/ghost/core/core/server/lib/get-sniper-links.ts
@@ -31,6 +31,7 @@ import type {MxRecord} from 'node:dns';
 import * as dns from 'node:dns/promises';
 import {parseEmailAddress} from '@tryghost/parse-email-address';
 import logging from '@tryghost/logging';
+import {constant} from 'lodash';
 
 type GetLinkFn = (options: Readonly<{recipient: string; sender: string}>) => string;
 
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/lib/get-sniper-links.ts
```diff
@@ -31,6 +31,7 @@ import type {MxRecord} from 'node:dns';
 import * as dns from 'node:dns/promises';
 import {parseEmailAddress} from '@tryghost/parse-email-address';
 import logging from '@tryghost/logging';
+import {constant} from 'lodash';
```
9larsons: issue (non-blocking): we try to limit the use of lodash because it's heavy for what it is. Could we move to a simple arrow fn or const here instead of lodash?

9larsons (approved):
Overall LGTM, one comment about lodash.

ghost/core/core/server/lib/get-sniper-links.ts
EvanHahn: Done in 042c4657e64d22377728e41fa4739c5120e3ef08.

PR #26025, opened 2026-01-27
Removed `memberWelcomeEmailSendInstantly` configuration parameter

cmraible:
ref https://linear.app/ghost/issue/NY-964/cleanup-remove-welcomeemailsendinstantly-configuration

Initially the process outbox job would run at most every 5 minutes, so we created this `memberWelcomeEmailSendInstantly` configuration parameter to run the job much more frequently for the sake of our e2e tests. We've since updated the implementation so the process-outbox job is triggered immediately when a member signs up anyways, so this configuration parameter is now redundant and unnecessary.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,8 +8,6 @@ node_modules
 build
 dist
 
-apps/*/types
-
 coverage
 
 .eslintcache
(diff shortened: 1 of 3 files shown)


cmraible:
Still planning to do this, but this isn't the right way to do it 😅

troyciesco (approved):

PR #26030, opened 2026-01-27
Started including feature flags in Portal

EvanHahn:

This change should have no user impact because it introduces a dormant feature.

We want to hide sniper links behind a feature flag, but Portal didn't support feature flags as far as I could tell. Now it does!

### DIFF
--- a/apps/portal/src/app.js
+++ b/apps/portal/src/app.js
@@ -199,7 +199,7 @@ export default class App extends React.Component {
     async initSetup() {
         try {
             // Fetch data from API, links, preview, dev sources
-            const {site, member, offers, page, showPopup, popupNotification, lastPage, pageQuery, pageData} = await this.fetchData();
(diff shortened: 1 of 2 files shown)


9larsons (approved):

9larsons (changes requested):
So the old pattern for this was we used the data attributes from `ghost_head`, which probably isn't the best model, and we can probably switch to using the API like you have outlined here. Alternatively, you could pivot and use the existing pattern of these data attributes. (see i18n as an example)

Personally, I'm in favor of moving to the API-based model as it's a bit more intuitive. The main thing is I just don't want to have multiple patterns here. For what it's worth, the i18n data attribute actually should just be cleaned up. It's been GA for quite a while.

EvanHahn:
Turns out this was already available on `site.labs`. Closing!

PR #26038, opened 2026-01-28
Refactored useFeatureFlag hook to use config instead of settings

rob-ghost:
The `useFeatureFlag` hook in the posts app was checking feature flags by parsing a JSON string from the settings API, while everywhere else in the codebase reads flags directly from `config.labs`. This created an inconsistency where toggling a feature flag via the Labs settings page wouldn't immediately reflect in components using this hook, since the `FeatureToggle` component updates the config cache but not the settings cache.

This PR simplifies the hook to read directly from `config.labs`, bringing it in line with how other apps check feature flags and ensuring changes are reflected immediately when flags are toggled.

### DIFF
--- a/apps/posts/test/unit/hooks/use-feature-flag.test.tsx
+++ b/apps/posts/test/unit/hooks/use-feature-flag.test.tsx
@@ -1,183 +1,75 @@
-import {beforeEach, describe, expect, it} from 'vitest';
-import {createTestWrapper, mockServer} from '../../utils/msw-helpers';
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import React from 'react';
+import {beforeEach, describe, expect, it, vi} from 'vitest';
 import {renderHook} from '@testing-library/react';
 import {useFeatureFlag} from '@src/hooks/use-feature-flag';
 
-/*
- * TODO: Fix useFeatureFlag tests - currently failing due to React Context dependency
- *
- * PROBLEM: useFeatureFlag depends on useGlobalData from PostAnalyticsContext, but MSW
- * only mocks HTTP requests, not React Context. The PostAnalyticsProvider requires
- * complex setup with multiple API dependencies.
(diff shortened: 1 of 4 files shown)


apps/posts/test/unit/hooks/use-feature-flag.test.tsx
```diff
@@ -10,40 +10,32 @@ import {useFeatureFlag} from '@src/hooks/use-feature-flag';
...
- *    This would make it work perfectly with MSW since it only needs HTTP calls
- * 3. Create a minimal PostAnalyticsContext provider wrapper for tests
- *
- * CURRENT STATUS: Tests skipped since this code is not currently in use
- * When this feature is needed, choose solution #2 (refactor) for better architecture
+ * CURRENT STATUS: Tests skipped since this code is not currently in use.
+ * The hook now uses config.labs instead of settings, which simplifies mocking.
  */
 
 describe.skip('useFeatureFlag', () => {
```
kevinansfield: This skipping seems a bit contradictory when the hook is actually being used in the codebase. The comment isn't particularly clear about the why unless its just outdated, did you spot any hints whilst making the changes?

rob-ghost: I glazed over the `.skip`, doing another pass on the tests as a whole

rob-ghost: Ended up with a pretty significant re-write of this and associated tests:

* Removed a bunch of duplicated tests with different names testing the same input / output
* Removed setting up global mocks for a bunch of things, only mocks the `useGlobalData` we use as the boundary
* Removed query mocks because it doesn't use react query anymore

I was of two minds about keeping the MSW layer or mocking the hook but the hook was just so much easier to setup.

kevinansfield (approved):

PR #26040, opened 2026-01-28
Centralized member commenting serialization in Member model

rob-ghost:

## What

Moves member `commenting` and `can_comment` serialization into the Member model's Bookshelf lifecycle hooks (`parse`/`format`/`toJSON`).

## Why

The `commenting` field (a JSON blob storing disabled state, reason, and expiry) and computed `can_comment` boolean were being parsed and serialized inconsistently across Ghost's API surfaces:

- **REST API** via `MemberBREADService.read()` correctly parsed and enriched the data
- **Webhooks** received raw Bookshelf models, missing the enrichment entirely
- **Comments API** had its own inline parsing logic

This created maintenance burden and potential for bugs when new API surfaces were added without remembering to apply the same transformations.

## Design Decision

Rather than adding shared utility functions or view models that each API surface would need to remember to use, this change leverages Bookshelf's existing lifecycle hooks:

- **`parse()`** transforms the raw JSON string from the database into a `MemberCommenting` domain object and computes `can_comment`
- **`format()`** reverses this for database writes
- **`toJSON()`** serializes the domain object to the snake_case API format

This ensures **every** Member load—regardless of which API surface or code path—automatically has the correctly formatted data. The serialization happens once, at the model layer, eliminating redundant calls and guaranteeing consistency.


ghost/core/core/server/models/member.js
```diff
@@ -403,6 +439,11 @@ const Member = ghostBookshelf.Model.extend({
             attrs.avatar_image = gravatar.url(attrs.email, {size: 250, default: 'blank'});
         }
 
+        // Serialize commenting domain object to API format
+        if (attrs.commenting) {
+            attrs.commenting = serializeCommenting(attrs.commenting);
+        }
```
kevinansfield, line 444: Could this be moved to `MemberCommentingCodec.toJSON(attrs.commenting)`?

It would remove the backwards import from serializer to model and reduce the output serializer's concern to just plucking/filtering properties based on access level

ghost/core/core/server/models/member.js
```diff
@@ -21,6 +23,40 @@ const Member = ghostBookshelf.Model.extend({
...
+     * Transform data coming from the database.
+     * Parses the `commenting` JSON string into a MemberCommenting domain object
+     * and computes the `can_comment` boolean.
+     */
+    parse(attrs) {
+        attrs = ghostBookshelf.Model.prototype.parse.call(this, attrs);
+
+        if (attrs.commenting !== undefined) {
+            const commenting = MemberCommentingCodec.parse(attrs.commenting);
+            attrs.commenting = commenting;
```
kevinansfield, line 35: Nit: With the parsing happening at the data layer it feels even more odd to me having the mixed `snake_case`/`camelCase` split

ghost/core/core/server/models/member.js
kevinansfield: If we go with the column codec plugin it would also keep all of the model-layer concerns in one place

rob-ghost: Yeah good idea, doing this now. Things might get a bit tricky when we move to Zod for these codecs because technically they're in/out, so we'd need a codec for the `parse/format` from/to the DB and a transformer for the API to take the place of `toJSON`

ghost/core/core/server/models/member.js
rob-ghost: Yeah I understand this.

I'm coming down on the side of, everything beyond the API boundary is `camelCase`, including what's saved in the JSON blob in the DB. Everything above the API boundary is `snake_case`.

Even if they did follow the same naming style, I'd still maintain discrete codecs / functions for `parse/format` and `toJSON` because we want control over how the different types and values are interpreted, so keeping the distinction between the attribute name styles doesn't feel so bad in that lens.

ghost/core/core/server/services/members/utils.js
```diff
@@ -31,7 +29,7 @@ module.exports.formattedMemberResponse = function formattedMemberResponse(member
         created_at: member.created_at,
         enable_comment_notifications: member.enable_comment_notifications,
         can_comment: member.can_comment,
-        commenting: serializeCommenting(member.commenting)
+        commenting: member.commenting
```
kevinansfield, line 32: Aside to this PR: This is the "front-end" `/members/api/member/` serialization that is accessible to the members themselves - are we ok exposing the disabled reason to them or should it be considered an admin-only property?

kevinansfield: Ah, I see we also expose the reason in API error messages so it looks like the assumption is that it's a public reason and not an internal Staff User note for why a member's commenting was disabled 👍

kevinansfield (approved):
Much cleaner than having multiple format/parse steps spread across the codebase ✨

ghost/core/core/server/services/members/utils.js
rob-ghost: Yeah, honestly speaking I thought `reason` would take a greater importance than it has done. The point of returning `reason` was so we could say to the member `You have been disabled from commenting for <reason> unit <disabledUntil>` but it doesn't look like that's currently a concern.

I've been keeping it for posterity mainly.

PR #26042, opened 2026-01-28
Removed accent color override logic from Portal

EvanHahn:
ref #26030
ref #26043

_This change should have no user impact._

When Portal fetches the site data, it overrides the accent color with the `data-accent-color` attribute on the Portal `<script>` tag. But [that property is set to the same value][0], so it's unnecessary.

As a followup, I removed the code that sets this in the HTML. See #26043.

Tested this manually by changing my accent color and making sure it's correct in Portal:

[0]: https://github.com/TryGhost/Ghost/blob/0a94c3df12ac30359541c48a322dfff5639a6bd9/ghost/core/core/frontend/helpers/ghost_head.js#L60

### DIFF
--- a/apps/portal/src/app.js
+++ b/apps/portal/src/app.js
@@ -580,27 +580,12 @@ export default class App extends React.Component {
         return {};
     }
 
-    /* Get the accent color from data attributes */
-    getColorOverride() {
-        const scriptTag = document.querySelector('script[data-ghost]');
-        if (scriptTag && scriptTag.dataset.accentColor) {
-            return scriptTag.dataset.accentColor;
-        }
-        return false;
-    }
-
     /** Fetch site, member session data and member offers with Ghost Apis  */
     async fetchApiData() {
         const {siteUrl, customSiteUrl, apiUrl, apiKey} = this.props;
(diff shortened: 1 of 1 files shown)


apps/portal/src/app.js
```diff
@@ -580,27 +580,12 @@ export default class App extends React.Component {
         return {};
     }
 
-    /* Get the accent color from data attributes */
-    getColorOverride() {
-        const scriptTag = document.querySelector('script[data-ghost]');
-        if (scriptTag && scriptTag.dataset.accentColor) {
```
EvanHahn, line 586: I assume this attribute is never set by anything other than `ghost_head`. If that's wrong (maybe users can customize portal's accent color separately?), we should keep this code.

kevinansfield: This is used for site previews via the Design Settings area in Admin. We have the override so that the temporary accent colour can be provided across the front-end without saving the settings and affecting the live site via the settings API.

EvanHahn: Ah, didn't realize this was necessary. Closing.

PR #26043, opened 2026-01-28
Removed accent-color data attribute from Portal <script> tag

EvanHahn:
ref https://github.com/TryGhost/Ghost/pull/26042

**I recommend reviewing #26042 first, as it provides context for this change.**

This should be safe to deploy any time. If it's deployed before #26042 is merged, the frontend will just use the accent color that's already returned from the server (which is the same). If it's deployed after, the frontend is ignoring the value anyway, so it doesn't matter.

### DIFF
--- a/ghost/core/core/frontend/helpers/ghost_head.js
+++ b/ghost/core/core/frontend/helpers/ghost_head.js
@@ -48,7 +48,7 @@ function finaliseStructuredData(meta) {
     return head;
 }
 
-function getMembersHelper(data, frontendKey, excludeList) {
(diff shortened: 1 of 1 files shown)


kevinansfield (changes requested):
Duplicate of comment here https://github.com/TryGhost/Ghost/pull/26042#discussion_r2737667253

The data attributes override is there to support design previews of the front-end. For those renders, Portal will still hit the settings API which won't contain the unsaved accent color, this data attribute is our way of overriding that with the appropriate preview colour.

EvanHahn:
Ah, didn't realize this was necessary. Closing.

PR #26044, opened 2026-01-28
Refactored bulk operations to use composable where strategies

rob-ghost:

## Summary

Ghost's bulk operations (`bulkEdit`, `bulkDestroy`) coupled query construction with execution — callers fetched ID arrays, then passed them into methods that iterated those arrays with chunked `whereIn` clauses. This worked, but every bulk operation followed the same two-step pattern regardless of whether the ID fetch was necessary.

This PR replaces `bulkEdit` and `bulkDestroy` with `bulkUpdate` and `bulkDelete` that accept composable *where strategies* — generator functions (`byNQL`, `byIds`, `byColumnValues`) that yield query modifier callbacks. The strategies handle their own chunking internally, and the bulk methods simply iterate whatever the strategy yields.

The practical benefit is that direct-table operations (comment status updates, link redirect updates) can now pass an NQL filter straight through via `byNQL`, skipping the fetch-all-IDs round trip entirely. Junction-table operations (members_newsletters, members_labels, posts_products) still need to fetch parent IDs first because the API filter targets the parent table, not the junction — so those use `byIds` or `byColumnValues` with the same two-step flow, just through the cleaner strategy interface.

Action logging (`addActions` for the audit trail) is preserved via an explicit `actionOptions` parameter separated from query parameters, making the contract clearer than the previous approach where logging options were mixed into the same object as query data.

The first commit adds missing E2E test coverage for the `addLabel` bulk member operation (which previously had none, unlike `removeLabel`). These tests pass against the original implementation and establish baseline coverage before the refactor.

### DIFF
--- a/ghost/core/core/server/lib/post-revisions.ts
+++ b/ghost/core/core/server/lib/post-revisions.ts
@@ -1,3 +1,5 @@
+import {byIds} from '../models/base/plugins/bulk-filters';
+
 type PostLike = {
     id: string;
     lexical: string;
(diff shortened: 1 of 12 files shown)


rob-ghost:
Closing as we're back-porting some of these ideas to the base branch.

PR #26045, opened 2026-01-28
🎨 Lazy-loaded URL service in app.js

rmgpinto:
ref https://linear.app/ghost/issue/BER-3132

- Moved urlService require from top-level to inside isMaintenanceModeEnabled function
- Allows server to start listening faster before urlService module is loaded

### DIFF
--- a/ghost/core/core/app.js
+++ b/ghost/core/core/app.js
@@ -2,17 +2,16 @@ const sentry = require('./shared/sentry');
 const express = require('./shared/express');
 const config = require('./shared/config');
 const logging = require('@tryghost/logging');
-const urlService = require('./server/services/url');
 
 const fs = require('fs');
 const path = require('path');
 
 const isMaintenanceModeEnabled = (req) => {
-    if (req.app.get('maintenance') || config.get('maintenance').enabled || !urlService.hasFinished()) {
+    if (req.app.get('maintenance') || config.get('maintenance').enabled) {
         return true;
     }
-
-    return false;
+    const urlService = require('./server/services/url');
+    return !urlService.hasFinished();
 };
 
 // We never want middleware functions to be anonymous


ghost/core/core/app.js
```diff
@@ -2,17 +2,16 @@ const sentry = require('./shared/sentry');
...
 
 const isMaintenanceModeEnabled = (req) => {
-    if (req.app.get('maintenance') || config.get('maintenance').enabled || !urlService.hasFinished()) {
+    if (req.app.get('maintenance') || config.get('maintenance').enabled) {
         return true;
     }
-
-    return false;
+    const urlService = require('./server/services/url');
+    return !urlService.hasFinished();
```
sagzy, line 14: As this middleware is called on every request, wouldn't we be creating more runtime cost here?

Wonder if we should cache the require at a module level, with e.g.

```
  let _urlService;
  const getUrlService = () => {
      if (!_urlService) {
          _urlService = require('./server/services/url');
      }
      return _urlService;
  };

```

Also I think there is an opportunity to extract the url service logic out of "isMaintenanceModeEnabled" as it doesn't semantically belong here, so maybe:

```
  const isMaintenanceModeEnabled = (req) => {
      return req.app.get('maintenance') || config.get('maintenance').enabled;
  };

  const isUrlServiceReady = () => {
      return getUrlService().hasFinished();
  };

  const maintenanceMiddleware = function maintenanceMiddleware(req, res, next) {
      if (!isMaintenanceModeEnabled(req) && isUrlServiceReady()) {
          return next();
      }

      // serve maintenance page
  };
```

sagzy (commented):
@rmgpinto left one suggestion on the code

Little note on commit message: I think we shouldn't be adding 🎨 emoji here, as this won't be a user-facing change

PR #26047, opened 2026-01-28
Removed i18n labs flag; it's been on since 2024

EvanHahn:
ref bb9a69edfede00cb30cfcbc9213489ca48101cf6

i18n has been enabled since November 2024. We can remove it from our list of always-on feature flags, remove `labs.isSet('i18n')` calls, and a bit of associated code.

### DIFF
--- a/ghost/core/core/frontend/helpers/ghost_head.js
+++ b/ghost/core/core/frontend/helpers/ghost_head.js
@@ -2,7 +2,7 @@
 // Usage: `{{ghost_head}}`
 //
 // Outputs scripts and other assets at the top of a Ghost theme
-const {labs, metaData, settingsCache, config, blogIcon, urlUtils, getFrontendKey, settingsHelpers} = require('../services/proxy');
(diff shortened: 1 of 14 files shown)


ghost/core/core/frontend/helpers/ghost_head.js
```diff
@@ -59,7 +59,7 @@ function getMembersHelper(data, frontendKey, excludeList) {
 
         const colorString = (_.has(data, 'site._preview') && data.site.accent_color) ? data.site.accent_color : '';
         const attributes = {
-            i18n: labs.isSet('i18n'),
+            i18n: true,
```
9larsons, line 62: nit: This does feel a little odd to continue passing this through/setting it, vs. removing it. That could be done separately.

9larsons (approved):
LGTM, seems to check out fine. @ErisDS will be happy to see this gone 😁.

ghost/core/core/frontend/helpers/ghost_head.js
EvanHahn: Yeah, I couldn't remove it yet, because Portal still expects it. But I can do that in a followup.

PR #26053, opened 2026-01-28
Added conditional await for pageHitPromise to base public-page POM

cmraible:
This automatically waits for a page hit request on any page on Ghost's frontend if it's running in the `analytics` project, but skips this step if running in any other project.


jonatansberg:
In closing this since the changes have been incorporated into the parent branch.

PR #26054, opened 2026-01-28
Enabled parallel browser tests across CI jobs

9larsons:
ref https://linear.app/ghost/issue/NY-958/

Browser tests previously had a global concurrency lock allowing only one job to run at a time across all PRs. This was because Stripe Connect test accounts used predictable emails (test0@example.com, etc.) that would conflict between concurrent jobs.
- Use GITHUB_RUN_ID in Stripe account emails for unique isolation
- Remove global concurrency lock from job_browser-tests
- Add scheduled cleanup workflow to delete test accounts older than 24h

This allows multiple PRs to run browser tests simultaneously without Stripe account conflicts, while preventing unbounded account growth.

Stripe doesn't appear to have a limit on Connect accounts, and we're allowed to create 5/s before hitting rate limits. We create four per test suite run and that should not restrict our needs with commits/PRs.


cmraible (approved):
This is amazing, nice one!

Assuming we don't see any issues with this next week, we should be able to run browser tests on _all_ PRs again!!!

PR #26055, opened 2026-01-28
Removed Should.js from core service unit tests

EvanHahn:
no ref

*This test-only change should have no user impact.*

Should.js is out, `node:assert/strict` and Sinon assertions are in.


EvanHahn:
Closing in favor of smaller changes.

PR #26063, opened 2026-01-29
Changed spam complaint suppressions to automatically delete from Mailgun

sam-lord:
ref [GVA-646](https://linear.app/ghost/issue/GVA-646/sync-spam-complaint-suppressions-from-mailgun-to-disable-members)

At the moment, spam complaints persist in Mailgun. This means that for sites that share a domain, a member creating a spam complaint on one site can impact another.

Also, if the spam complaint is made in error, the site admin has no way to fix the issue on behalf of the member.


sagzy (approved):
LGTM

PR #26064, opened 2026-01-29
Skipped retention offers for members in trial period

mike182uk:
ref https://linear.app/ghost/issue/FEA-535

Members on a tier with a trial period were being shown retention offers, which led to a confusing "Free trial - Ends {date}" message after redemption. This change skips retention offers for subscriptions with an active trial period (`trial_end_at` in the future), matching the existing behavior for offer-based trials


sagzy (approved):
Looks good! Do we want to apply the same logic to `applyOfferToSubscription`, to be extra safe?

sagzy (approved):

PR #26070, opened 2026-01-29
Added "View in admin" link to comment actions

kevinansfield:
ref https://linear.app/ghost/issue/BER-3199/

- adds link to Comments-UI comment admin menu that deep links to the comment inside the new admin-area comment moderation screen

### DIFF
--- a/apps/comments-ui/test/e2e/admin-moderation.test.ts
+++ b/apps/comments-ui/test/e2e/admin-moderation.test.ts
@@ -314,4 +314,53 @@ test.describe('Admin moderation', async () => {
         const {frame} = await initializeTest(page);
         await expect(frame.getByTestId('count')).toContainText('3 comments');
     });
+
+    test.describe('View in admin link', () => {
+        test('shows View in admin link when commentModeration flag is enabled', async ({page}) => {
+            mockedApi.addComment({id: 'test-comment-id', html: '<p>This is a comment</p>'});
+
+            await mockAdminAuthFrame({page, admin});
+
+            const {frame} = await initialize({
+                mockedApi,
+                page,
+                publication: 'Publisher Weekly',
+                title: 'Member discussion',
+                count: true,
+                admin,
+                labs: {
+                    commentModeration: true
+                }
+            });
+
+            const moreButtons = frame.getByTestId('more-button');
+            await moreButtons.nth(0).click();
+
+            const viewInAdminLink = frame.getByTestId('view-in-admin-button');
+            await expect(viewInAdminLink).toBeVisible();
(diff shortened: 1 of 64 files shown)


apps/comments-ui/test/e2e/admin-moderation.test.ts
```diff
@@ -314,4 +314,53 @@ test.describe('Admin moderation', async () => {
...
+                }
+            });
+
+            const moreButtons = frame.getByTestId('more-button');
+            await moreButtons.nth(0).click();
+
+            const viewInAdminLink = frame.getByTestId('view-in-admin-button');
+            await expect(viewInAdminLink).toBeVisible();
+            await expect(viewInAdminLink).toHaveAttribute('href', `${admin}#/comments/?id=is:test-comment-id`);
+            await expect(viewInAdminLink).toHaveAttribute('target', '_blank');
```
rob-ghost, line 342: **Suggestion (non-blocking):** For the "view in context" button (admin -> comments-ui) our test opened the page and expected that the location was correct, rather than inspecting that the DOM had the right attributes.

This means we test the user behaviour of clicking the link and a tab opening, and the implementation can change underneath.

rob-ghost (approved):
Code-wise it looks good 👍🏻

apps/comments-ui/test/e2e/admin-moderation.test.ts
kevinansfield: These are independent comments-ui tests rather than our main e2e suite so there's no admin app being served anywhere. We could try and capture the test opening a new window and verifying the URL that it tried to open is correct?

kevinansfield: I gave it a go but as I suspected Playwright isn't happy, the new page URL is always `"chrome-error://chromewebdata/"`

kevinansfield: We can mock the admin page with a basic HTML response which works. I wonder if we're getting a bit too deep for something that's essentially an integration test where we're asserting we render what we expect?

```
// Mock the admin page so navigation succeeds
await page.context().route(`${admin}**`, async (route) => {
    await route.fulfill({
        status: 200,
        body: '<html><body>Admin</body></html>'
    });
});

const newPagePromise = page.context().waitForEvent('page');
await viewInAdminLink.click();
const newPage = await newPagePromise;

await expect(newPage).toHaveURL(`${admin}#/comments/?id=is:test-comment-id`);
await newPage.close();
```

rob-ghost: If its not straight forward forget about it! Not worth the effort

PR #26085, opened 2026-01-29
Fixed over-eager newsletter redirect, possibly fixing flaky test

EvanHahn:
ref https://github.com/TryGhost/Ghost/actions/runs/21494158108/job/61924761703

One of the newsletter tests is [flaky in CI][0]. I think this is because app settings are loading and we redirect away before they do.

The old logic was, roughly: redirect if newsletters are disabled *or* if app settings are still loading.

The new logic is: redirect if newsletters are disabled. Don't redirect if app settings are still loading.

I'm not certain this will fix CI flakiness--I can't reproduce locally--but I think this is an improvement anyway.

[0]: https://github.com/TryGhost/Ghost/actions/runs/21494158108/job/61924761703


cmraible (approved):
Nice find! 

I tested this out and noticed that the redirect leading to `/` wasn't working properly and displaying a blank screen, so I pushed a little update just to update the redirect to `/analytics` as it should be.

PR #26086, opened 2026-01-29
Fixed paid welcome emails incorrectly sent to admin-created members

cmraible:
ref https://linear.app/ghost/issue/NY-891/before-ga-double-check-that-only-members-createdupdated-through-portal

## Summary

- Added unit tests for `member-bread-service.add()` to verify context is passed to `linkStripeCustomer`
- Fixed bug where `context` was not included in `sharedOptions` when calling `linkStripeCustomer`

## Problem

When an admin creates a member via the Admin API with a `stripe_customer_id`, the paid welcome email was incorrectly being sent. This happened because:

1. `member-bread-service.add()` calls `linkStripeCustomer()` with `sharedOptions`
2. `sharedOptions` only included `transacting`, not `context`
3. `linkStripeCustomer()` → `linkSubscription()` uses context to determine the source
4. Without context, `_resolveContextSource({})` defaulted to `'member'`
5. Since `'member'` is in `WELCOME_EMAIL_SOURCES`, the paid welcome email was queued

## Fix

Updated `sharedOptions` in `member-bread-service.js` to include `context`:

```javascript
const sharedOptions = {
    ...(options.transacting && {transacting: options.transacting}),
    ...(options.context && {context: options.context})
};
```

Now the source correctly resolves to `'admin'` for admin-created members, and no welcome email is sent.

---

## Side effect in members API tests

This change also corrects a bug in the staff notifications behavior, which was being hidden by the bug described above. Previously in the members API tests we were asserting that a "Paid subscription started" staff notification was sent. This was the case because, due to the bug described and fixed in this PR, the member in these tests was being incorrectly created with source = 'member'.

Now these members are correctly created with source = 'admin', and we correctly do not send the "Paid subscription started" notification anymore.

--- 

### Signup as a paid member via portal
- Signup via portal for a paid subscription
- [x] Member should receive welcome email
- [x] Staff user should receive "Paid subscription started" email

### Add a paid member via Admin API
- Create a new Stripe customer in your stripe dashboard
- Add payment method using the 4242 4242 4242 4242 test card
- Add an active subscription to the customer
- Grab the stripe customer ID
- Grab your staff access token, or an integration's Admin API token
- Create a new member by POSTing to `/ghost/api/admin/members`, using your access token to authenticate and passing the stripe_customer_id
- [x] Member should be created, with `status = paid`
- [x] Member should not receive welcome email
- [x] Staff user should receive "New free member" email. Note: this is a pre-existing bug. This shouldn't really happen, but it's not new behavior introduced by this PR
- [x] Staff user should receive "Paid subscription started" email

### Add a paid member via members CSV import
- Create a new Stripe customer in your stripe dashboard
- Add payment method
- Add an active subscription to the customer
- Grab the stripe customer id
- Create a members CSV import with a unique (i.e. not already in use by existing member) email address, and include the stripe_customer_id
- Import the CSV file in the members screen
- [x] Member should be created, with `status = paid`
- [x] Member should not receive welcome email
- [x] Staff user should not receive "Paid subscription started" email


cmraible:
I did some digging and found this issue from the original implementation: https://github.com/TryGhost/Product/issues/1864

> 
> The expected behavior for triggering email alerts for members created via different sources -
> 
>     Importer = our import tooling ❌
>     Admin = manually in the UI ❌
>     API = external / zapier ✅
>     Member = member triggered ✅


So I think this actually is correct, and the members webhook tests were only passing because of this bug

ghost/core/test/e2e-api/members/webhooks.test.js
```diff
@@ -379,11 +380,6 @@ describe('Members API', function () {
             // Check that the staff notifications has been sent
             await DomainEvents.allSettled();
 
-            mockManager.assert.sentEmail({
-                subject: /Paid subscription started: Cancel me at the end of the billing cycle/,
-                to: 'jbloggs@example.com'
-            });
-
```
cmraible, line 386: This assertion should not have been here in the first place, but it was required in the test due to the underlying bug we found.

The member created above in this test now correctly has source = 'admin', and Ghost correctly should not send a "Paid subscription started" notification email in this case. There isn't really a behavioral change here, because there isn't any UI in Admin that allows creating paid members with `stripe_customer_id`.

ghost/core/test/e2e-api/members/webhooks.test.js
```diff
@@ -545,11 +542,6 @@ describe('Members API', function () {
             // Check that the staff notifications has been sent
             await DomainEvents.allSettled();
 
-            mockManager.assert.sentEmail({
-                subject: /Paid subscription started: Cancel me now/,
-                to: 'jbloggs@example.com'
-            });
-
```
cmraible, line 552: Ditto here

ghost/core/test/unit/server/services/members/members-api/services/members-bread-service.test.js
```diff
@@ -5,6 +5,152 @@ const NextPaymentCalculator = require('../../../../../../../core/server/services
...
+
+            return {service, memberRepository, linkStripeCustomerStub, createStub};
+        }
+
+        it('passes context to linkStripeCustomer when stripe_customer_id is provided', async function () {
+            // This test verifies that when a member is created via Admin API with a stripe_customer_id,
+            // the context (containing {user: true}) is passed through to linkStripeCustomer.
+            // This is important because linkStripeCustomer -> linkSubscription uses context
+            // to determine the source for welcome emails. Without context, the source defaults
+            // to 'member', causing welcome emails to be incorrectly sent for admin-created members.
```
troyciesco, line 83: random observation, not for this PR: i wonder if we should even be defaulting to "member" as the source at all 🤔

troyciesco (approved):
lgtm -  i did look a bit to see if there was more tests we need to add, but i think we're good, we already test that welcome emails are sent/not sent based on what's passed

ghost/core/test/unit/server/services/members/members-api/services/members-bread-service.test.js
cmraible: Yeah that's a great point. It might be best to default to 'unknown' (or similar) as long as that doesn't break anything...then we can at least see how often we get 'unknown' and start to eliminate any gaps in our source logic until we can be 100% confident that we've covered all the edges 🤔

PR #26093, opened 2026-01-30
Added loader to public assets

9larsons:
ref https://ghost.slack.com/archives/C018EKC56JF/p1769803075901969

The loader animation was not showing when loading admin.


jonatansberg:
I'm not sure this actually changes anything?

We're merging the Ember and React asset so the loading video should already be at that path. AFAIK the loader has been showing consistently until now so I wonder if something else has changed 🤔

9larsons:
Closing in lieu of https://github.com/TryGhost/Ghost/pull/26100.

PR #26098, opened 2026-02-01
Threaded comment replies

weylandswart:
Draft PR to add threaded replies to comments. Triggered from a parent (Comment) or child (Reply).

### DIFF
--- a/apps/admin-x-framework/src/api/comments.ts
+++ b/apps/admin-x-framework/src/api/comments.ts
@@ -92,11 +92,7 @@ export const useHideComment = createMutation<CommentsResponseType, {id: string}>
         }]
     }),
     invalidateQueries: {
-        filters: {
-            predicate: (query) => {
-                return query.queryKey[0] === 'CommentsResponseType' || query.queryKey[0] === 'CommentThreadResponseType';
-            }
-        }
+        dataType
     }
 });
 
@@ -110,11 +106,7 @@ export const useShowComment = createMutation<CommentsResponseType, {id: string}>
         }]
     }),
     invalidateQueries: {
-        filters: {
-            predicate: (query) => {
-                return query.queryKey[0] === 'CommentsResponseType' || query.queryKey[0] === 'CommentThreadResponseType';
-            }
-        }
+        dataType
     }
 });
 
@@ -128,22 +120,12 @@ export const useDeleteComment = createMutation<CommentsResponseType, {id: string
         }]
     }),
     invalidateQueries: {
-        dataType: 'CommentsResponseType',
-        filters: {
-            predicate: (query) => {
-                return query.queryKey[0] === 'CommentsResponseType' || query.queryKey[0] === 'CommentThreadResponseType';
-            }
-        }
+        dataType
     }
 });
 
-export interface CommentThreadResponseType {
-    meta?: Meta;
-    comments: Comment[];
-}
-
-const useCommentRepliesQuery = createQueryWithId<CommentThreadResponseType>({
-    dataType: 'CommentThreadResponseType',
+export const useCommentReplies = createQueryWithId<CommentsResponseType>({
+    dataType,
     path: (id: string) => `/comments/${id}/replies/`,
     defaultSearchParams: {
         include: 'member,post,count.replies,count.likes,count.reports,parent',
@@ -151,52 +133,10 @@ const useCommentRepliesQuery = createQueryWithId<CommentThreadResponseType>({
     }
 });
 
-export const useCommentReplies = (commentId: string | null, options?: {enabled?: boolean}) => {
-    // Always call the hook to maintain hook order, but disable when commentId is null
-    // Use a placeholder ID to avoid invalid URLs in queryKey when commentId is null
-    const result = useCommentRepliesQuery(commentId || '__placeholder__', {
-        enabled: options?.enabled !== false && !!commentId
-    });
-    
-    // Return empty state when commentId is null to match previous behavior
-    if (!commentId) {
-        return {
-            data: undefined,
-            isLoading: false,
-            isError: false,
-            error: null,
-            refetch: () => Promise.resolve({data: undefined, isLoading: false, isError: false, error: null} as unknown as ReturnType<typeof result.refetch>)
-        };
-    }
-    
-    return result;
-};
-
-const useReadCommentQuery = createQueryWithId<CommentsResponseType>({
-    dataType: 'CommentsResponseType',
+export const useReadComment = createQueryWithId<CommentsResponseType>({
+    dataType,
     path: (id: string) => `/comments/${id}/`,
     defaultSearchParams: {
         include: 'member,post,count.replies,count.likes,count.reports,parent'
     }
 });
-
-export const useReadComment = (commentId: string | null, options?: {enabled?: boolean}) => {
-    // Always call the hook to maintain hook order, but disable when commentId is null
-    // Use a placeholder ID to avoid invalid URLs in queryKey when commentId is null
-    const result = useReadCommentQuery(commentId || '__placeholder__', {
-        enabled: options?.enabled !== false && !!commentId
-    });
-    
-    // Return empty state when commentId is null to match previous behavior
-    if (!commentId) {
-        return {
-            data: undefined,
-            isLoading: false,
-            isError: false,
-            error: null,
-            refetch: () => Promise.resolve({data: undefined, isLoading: false, isError: false, error: null} as unknown as ReturnType<typeof result.refetch>)
-        };
-    }
-    
-    return result;
-};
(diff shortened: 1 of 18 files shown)


apps/admin-x-framework/src/api/comments.ts
```diff
@@ -114,6 +128,75 @@ export const useDeleteComment = createMutation<CommentsResponseType, {id: string
...
     }
 });
+
+export interface CommentThreadResponseType {
+    meta?: Meta;
+    comments: Comment[];
+}
+
+const useCommentRepliesQuery = createQueryWithId<CommentThreadResponseType>({
+    dataType: 'CommentThreadResponseType',
```
jonatansberg: I don't think we need to create a separate data type for the comment thread responses. We can treat them the same as the comments, especially since we in any case always clear both of them. That would clean up the implementation significantly.

apps/admin-x-framework/src/api/comments.ts
```diff
@@ -86,7 +92,11 @@ export const useHideComment = createMutation<CommentsResponseType, {id: string}>
         }]
     }),
     invalidateQueries: {
-        dataType: 'CommentsResponseType'
+        filters: {
+            predicate: (query) => {
+                return query.queryKey[0] === 'CommentsResponseType' || query.queryKey[0] === 'CommentThreadResponseType';
+            }
+        }
```
jonatansberg: Revert all of these changes and use the `dataType` instead.

apps/admin-x-framework/src/api/comments.ts
```diff
@@ -114,6 +128,75 @@ export const useDeleteComment = createMutation<CommentsResponseType, {id: string
...
+            data: undefined,
+            isLoading: false,
+            isError: false,
+            error: null,
+            refetch: () => Promise.resolve({data: undefined, isLoading: false, isError: false, error: null} as unknown as ReturnType<typeof result.refetch>)
+        };
+    }
+    
+    return result;
+};
```
jonatansberg: I think we can get rid of this altogether, rename `useCommentRepliesQuery` to `useCommentReplies` and use that directly.

apps/admin-x-framework/src/api/comments.ts
```diff
@@ -114,6 +128,75 @@ export const useDeleteComment = createMutation<CommentsResponseType, {id: string
...
+
+const useReadCommentQuery = createQueryWithId<CommentsResponseType>({
+    dataType: 'CommentsResponseType',
+    path: (id: string) => `/comments/${id}/`,
+    defaultSearchParams: {
+        include: 'member,post,count.replies,count.likes,count.reports,parent'
+    }
+});
+
+export const useReadComment = (commentId: string | null, options?: {enabled?: boolean}) => {
```
jonatansberg: Same here

apps/admin-x-framework/src/api/members.ts
```diff
@@ -42,6 +46,10 @@ export const useEnableMemberCommenting = createMutation<
     path: ({id}) => `/members/${id}/commenting/enable`,
     body: () => ({}),
     invalidateQueries: {
-        dataType: 'CommentsResponseType'
+        filters: {
+            predicate: (query) => {
+                return query.queryKey[0] === 'CommentsResponseType' || query.queryKey[0] === 'CommentThreadResponseType';
+            }
+        }
```
jonatansberg: Revert these changes in this file as well

apps/posts/src/hooks/use-post-success-modal.ts
```diff
@@ -149,7 +149,7 @@ export const usePostSuccessModal = () => {
             author: getAuthorsText(post.authors),
             onClose: handleClose
         };
-    }, [post, isModalOpen, postCount, site?.title]);
+    }, [post, isModalOpen, postCount, site?.title, site?.icon]);
```
jonatansberg: This looks like an unrelated change?

apps/posts/src/views/comments/components/comment-thread-list.tsx
```diff
@@ -0,0 +1,414 @@
...
+    const date = new Date(dateString);
+    const formatted = new Intl.DateTimeFormat('en-US', {
+        month: 'short',
+        day: 'numeric',
+        year: 'numeric',
+        hour: 'numeric',
+        minute: 'numeric'
+    }).format(date);
+    return formatted.replace(/(\d+),(\s+\d{4})/, '$1$2');
+}
```
jonatansberg: I think we have some existing utilities for formatting dates already. If possible, we should be using them rather than introducing new ones.

apps/posts/src/views/comments/components/comment-thread-list.tsx
```diff
@@ -0,0 +1,414 @@
...
+
+    const avatar = (
+        <div className={`relative mb-3 flex size-6 min-w-6 shrink-0 items-center justify-center overflow-hidden rounded-full bg-accent md:mb-4 md:size-8 md:min-w-8 ${comment.status === 'hidden' && 'opacity-50'}`}>
+            {comment.member?.id && comment.member.avatar_image && (
+                <div className='absolute inset-0'><img alt="Member avatar" className="size-full rounded-full object-cover" src={comment.member.avatar_image} /></div>
+            )}
+            <div>
+                <LucideIcon.User className='!size-3 text-muted-foreground md:!size-4' size={12} />
+            </div>
+        </div>
```
jonatansberg: Should this be a separate component? I feel like we already have this in multiple variations elsewhere.

ghost/admin/app/services/state-bridge.js
```diff
@@ -7,6 +7,8 @@ import {run} from '@ember/runloop';
 
 const emberDataTypeMapping = {
     AutomatedEmailsResponseType: null, // automated emails only exist in React admin
+    CommentsResponseType: null, // comments only exist in React admin
+    CommentThreadResponseType: null, // comment threads only exist in React admin
```
jonatansberg: Remove the CommentThreadResponseType

apps/posts/src/views/comments/components/comment-thread-list.tsx
```diff
@@ -0,0 +1,414 @@
...
+        
+        // Recursively build tree for each direct reply
+        const nestedReplies = directReplies.map(reply => buildCommentTree(reply, allReplies, new Set(visitedIds)));
+        
+        // Return parent with nested replies
+        return {
+            ...parent,
+            replies: nestedReplies
+        };
+    };
```
jonatansberg: Since we already have the parent and replies, do we even need to build a tree at this point? They would all have the same parent ID.

apps/posts/src/views/comments/components/comment-thread-list.tsx
```diff
@@ -0,0 +1,414 @@
...
+        }
+    };
+
+    const handleEnableCommenting = (member: Comment['member']) => {
+        if (member?.id) {
+            enableCommenting({id: member.id});
+        }
+    };
+
+    // Check if this comment has replies (from the nested structure we'll build)
```
jonatansberg: This is a weird comment. You can remove it. We should consider if we want to simplify this component as well since we only really have a parent and children, not a full tree.

jonatansberg (commented):
Good job with putting this together! Most of my comments are related to simplifying the admin API integration. It's mostly about pulling out things we don't need to reduce the complexity.

The other thing I think we should take a look at before merging this is the remaining duplication between the different comment components. Maybe you can work with Zimo to refactor that or we can pair up some time tomorrow to take that on?

apps/posts/src/views/comments/components/comment-thread-list.tsx
weylandswart: Updated in https://github.com/TryGhost/Ghost/pull/26098/commits/9124fbc4f04a3772375a6902662050435eede699

apps/posts/src/hooks/use-post-success-modal.ts
weylandswart: Whoops, sorry. Removed in https://github.com/TryGhost/Ghost/pull/26098/commits/9124fbc4f04a3772375a6902662050435eede699

apps/posts/src/views/comments/components/comment-thread-list.tsx
weylandswart: It could be indeed. I had previously just used what existed on `CommentsList.tsx` I do now have to do a check for `!comment.member.avatar_image.includes('d=blank')` which _feels_ wrong? I don't know enough about Gravatar to know whether we could have this return nothing if it's blank.

Updated in https://github.com/TryGhost/Ghost/pull/26098/commits/9124fbc4f04a3772375a6902662050435eede699 in any case.

PR #26103, opened 2026-02-02
Removed unused Ember whats-new code

kevinansfield:
ref https://linear.app/ghost/issue/BER-3075/

- all whats-new implementation has moved to the React Admin shell so we can remove the unused code from Ember Admin


rob-ghost (approved):
💣

PR #26107, opened 2026-02-02
Added test session isolation for concurrent runs (database + port)

rob-ghost:
All ghost/core mocha test suites (integration, e2e-api, e2e-server, etc.) share the same hardcoded database name and server port. Running any two suites concurrently — whether across worktrees or terminal tabs — causes failures from database corruption and EADDRINUSE port conflicts.

## Database isolation

Generates a unique database name per mocha session using nconf-style environment variables (`database__connection__database` for MySQL, `database__connection__filename` for SQLite). Using env vars ensures child processes — such as bree job workers spawned during tests — inherit the correct database configuration. The config singleton is also updated via `config.set()` for code that has already loaded config.

An `afterAll` mocha hook cleans up by dropping the MySQL database or removing the SQLite files when the session ends, but only for databases we generated (not externally-provided ones).

## Port isolation

Assigns a random HTTP server port (10000–60000) per mocha session via the `server__port` env var, replacing the hardcoded port 2369. The `url` config is only updated to match the random port when `server: true` is used (7 test files that start a real HTTP listener). Unit tests — which never start a server — continue using the canonical `http://127.0.0.1:2369` URL, so their hardcoded assertions are unaffected.

The snapshot manager is monkey-patched to normalize response URLs back to port 2369 before comparison, so committed `.snap` files remain stable regardless of which port was used at runtime.

## Overrides

Both `database__connection__*` and `server__port` can be set explicitly to override the generated values, following the same pattern used in CI workflows and Docker compose files.

The top-level `e2e/` Playwright suite already has its own per-test isolation and is unaffected by this change.


rob-ghost:
## Verification script

Save as `ghost/core/scripts/test-concurrent-isolation.sh` and run from `ghost/core/`:

```bash
chmod +x scripts/test-concurrent-isolation.sh
./scripts/test-concurrent-isolation.sh
```

<details>
<summary>test-concurrent-isolation.sh</summary>

```bash
#!/usr/bin/env bash
#
# Prove that concurrent test runs don't conflict due to shared database
# or port resources. Runs two mocha processes in parallel and verifies
# both exit cleanly.
#
# Usage:  ./scripts/test-concurrent-isolation.sh
#
# What it tests:
#   1. Database isolation: each process gets a unique SQLite DB file
#   2. Port isolation: each process binds its HTTP server to a random port
#   3. Snapshot stability: snapshot comparisons normalize ports to 2369
#
set -euo pipefail

SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
GHOST_CORE="$(cd "$SCRIPT_DIR/.." && pwd)"

echo "=== Concurrent Test Isolation Verification ==="
echo ""
echo "Running two test processes in parallel, each with:"
echo "  - A unique SQLite database (random filename in /tmp)"
echo "  - A random HTTP server port (avoids EADDRINUSE)"
echo ""

# Pick two test files that start a real HTTP server (server: true).
# These are the hardest case — they would conflict on port 2369 without
# the isolation changes.
TEST_A="test/e2e-server/1-options-requests.test.js"
TEST_B="test/e2e-server/click-tracking.test.js"

TMPDIR_A=$(mktemp -d)
TMPDIR_B=$(mktemp -d)

cd "$GHOST_CORE"

echo "Starting run A: $TEST_A"
yarn test:single "$TEST_A" > "$TMPDIR_A/output.log" 2>&1 &
PID_A=$!

echo "Starting run B: $TEST_B"
yarn test:single "$TEST_B" > "$TMPDIR_B/output.log" 2>&1 &
PID_B=$!

echo ""
echo "Waiting for both runs to complete (PIDs: $PID_A, $PID_B)..."
echo ""

EXIT_A=0
EXIT_B=0

wait $PID_A || EXIT_A=$?
wait $PID_B || EXIT_B=$?

# Extract results
PASS_A=$(grep -oE '[0-9]+ passing' "$TMPDIR_A/output.log" || echo "? passing")
PASS_B=$(grep -oE '[0-9]+ passing' "$TMPDIR_B/output.log" || echo "? passing")
FAIL_A=$(grep -oE '[0-9]+ failing' "$TMPDIR_A/output.log" || echo "0 failing")
FAIL_B=$(grep -oE '[0-9]+ failing' "$TMPDIR_B/output.log" || echo "0 failing")

echo "--- Results ---"
echo ""
echo "  Run A ($TEST_A): exit=$EXIT_A  $PASS_A  $FAIL_A"
echo "  Run B ($TEST_B): exit=$EXIT_B  $PASS_B  $FAIL_B"
echo ""

# Check for EADDRINUSE (the specific error this change prevents)
if grep -q "EADDRINUSE" "$TMPDIR_A/output.log" "$TMPDIR_B/output.log" 2>/dev/null; then
    echo "FAIL: EADDRINUSE detected — port isolation is not working"
    echo ""
    grep "EADDRINUSE" "$TMPDIR_A/output.log" "$TMPDIR_B/output.log" 2>/dev/null
    rm -rf "$TMPDIR_A" "$TMPDIR_B"
    exit 1
fi

# Check for database contention errors
if grep -qi "table.*doesn't exist\|database is locked\|database.*not found" "$TMPDIR_A/output.log" "$TMPDIR_B/output.log" 2>/dev/null; then
    echo "FAIL: Database contention detected — DB isolation is not working"
    rm -rf "$TMPDIR_A" "$TMPDIR_B"
    exit 1
fi

rm -rf "$TMPDIR_A" "$TMPDIR_B"

if [ "$EXIT_A" -ne 0 ] || [ "$EXIT_B" -ne 0 ]; then
    echo "FAIL: One or both test runs failed"
    exit 1
fi

echo "PASS: Both test runs completed successfully in parallel"
echo "  No EADDRINUSE errors (port isolation working)"
echo "  No database contention (DB isolation working)"
exit 0
```

</details>

**Local test output:**

```
=== Concurrent Test Isolation Verification ===

Running two test processes in parallel, each with:
  - A unique SQLite database (random filename in /tmp)
  - A random HTTP server port (avoids EADDRINUSE)

Starting run A: test/e2e-server/1-options-requests.test.js
Starting run B: test/e2e-server/click-tracking.test.js

Waiting for both runs to complete...

--- Results ---

  Run A (test/e2e-server/1-options-requests.test.js): exit=0  12 passing  0 failing
  Run B (test/e2e-server/click-tracking.test.js): exit=0  1 passing  0 failing

PASS: Both test runs completed successfully in parallel
  No EADDRINUSE errors (port isolation working)
  No database contention (DB isolation working)
```

Also verified two full integration suites (167 tests each) pass concurrently with zero failures.

PR #26109, opened 2026-02-02
🐛 Fixed comp subscription not being cancelled on upgrade to paid

rob-ghost:
When a comped member upgrades to a paid subscription, the complimentary access was not being removed. This meant that when the paid subscription was later cancelled, the member would revert to `comped` status instead of `free` — effectively giving them perpetual access after a single payment cycle. The bug manifests when the comp tier is different from the paid subscription tier, which is why it was tricky to reproduce locally.

This PR consolidates the two flavours of complimentary access removal into a single `removeComplimentarySubscription` method in MemberRepository. Ghost has two types of comp access: Stripe-backed subscriptions (with `plan_nickname: 'Complimentary'` and `unit_amount: 0`) and Ghost-only comps (products added directly via the Admin UI with no backing Stripe subscription). The old `cancelComplimentarySubscription` only handled Stripe-backed comps, and didn't filter specifically for complimentary subscriptions — it would cancel all non-canceled subscriptions. The new method handles both types correctly: it cancels Stripe-backed comp subscriptions via the Stripe API (then syncs via `linkSubscription`), and removes Ghost-only comp products by checking which products are backed by active Stripe subscriptions.

Rather than embedding this logic inside `linkSubscription` (which was the v1 approach and raised recursion concerns), the comp removal is called from the webhook handlers — `SubscriptionEventService` and `CheckoutSessionEventService` — after `linkSubscription` completes. If the member's status is now `paid`, any remaining comp access is cleaned up. This keeps `linkSubscription` focused on its single responsibility of syncing one Stripe subscription to the Ghost database.

Supersedes https://github.com/TryGhost/Ghost/pull/26074.

### DIFF
--- a/ghost/core/core/server/services/members/members-api/repositories/member-repository.js
+++ b/ghost/core/core/server/services/members/members-api/repositories/member-repository.js
@@ -152,7 +152,11 @@ module.exports = class MemberRepository {
     }
 
     isComplimentarySubscription(subscription) {
-        return subscription.plan && subscription.plan.nickname && subscription.plan.nickname.toLowerCase() === 'complimentary';
+        return this.isComplimentaryPlanNickname(subscription.plan?.nickname);
+    }
+
+    isComplimentaryPlanNickname(nickname) {
+        return nickname && nickname.toLowerCase() === 'complimentary';
     }
 
     /**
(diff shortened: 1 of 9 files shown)


ghost/core/core/server/services/members/members-api/repositories/member-repository.js
```diff
@@ -1991,34 +1991,66 @@ module.exports = class MemberRepository {
...
-        });
-
-        const subscriptions = await member.related('stripeSubscriptions').fetch();
+        const member = await this._Member.findOne({id});
+        const subscriptions = await member.related('stripeSubscriptions').fetch(options);
 
+        // 1. Cancel Stripe-backed complimentary subscriptions
         for (const subscription of subscriptions.models) {
-            if (subscription.get('status') !== 'canceled') {
+            const isComp = this.isComplimentarySubscription({plan: {nickname: subscription.get('plan_nickname')}});
```
sagzy: **nitpick (non-blocking)**: instead of having to construct a "Stripe-like" subscription object here, maybe we should extract out the `nickname.toLowerCase() === 'complimentary'` logic - so that it can be called from a Stripe subscription or Ghost subscription

ghost/core/core/server/services/stripe/services/webhook/checkout-session-event-service.js
```diff
@@ -251,6 +251,12 @@ module.exports = class CheckoutSessionEventService {
                     });
                 }
             }
+
+            // If member is now paid, remove any complimentary access
+            const updatedMember = await memberRepository.get({id: member.id});
+            if (updatedMember && updatedMember.get('status') === 'paid') {
+                await memberRepository.removeComplimentarySubscription({id: member.id});
+            }
```
sagzy, line 259: AFAIK, there isn't an instance where we'd receive a checkout webhook without a `customer.subscription.*`

Therefore, during a member upgrade, we'd be calling `memberRepository.removeComplimentarySubscription` twice - which shouldn't harm, but also isn't strictly necessary. 

**Suggestion**: remove logic from here, and keep it in the subscription event handler only

sagzy (approved):

ghost/core/core/server/services/stripe/services/webhook/checkout-session-event-service.js
rob-ghost: I tried this, but Stripe doesn't guarantee delivery order of webhooks. If `customer.subscription.created` arrives before `checkout.session.completed`, the member won't exist yet (member creation happens in the checkout handler), so the subscription handler silently skips. Keeping it in both covers the race condition.

sagzy: Ah! That's a good catch 👍

ghost/core/core/server/services/members/members-api/repositories/member-repository.js
```diff
@@ -1991,34 +1995,66 @@ module.exports = class MemberRepository {
...
+                .map(sub => sub.related('stripePrice')?.related('stripeProduct')?.get('product_id'))
+                .filter(Boolean)
+        );
+
+        const currentProducts = member.related('products').toJSON();
+        const filteredProducts = currentProducts.filter(product => activeSubscriptionProductIds.has(product.id));
+
+        if (filteredProducts.length !== currentProducts.length) {
+            await this._Member.edit({products: filteredProducts}, {id});
+        }
```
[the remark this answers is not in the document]
rob-ghost: This method is only called after we've confirmed `updatedMember.get('status') === 'paid'` — the paid subscription has already been linked via `linkSubscription`, which handles the status update. So by the time we're removing comp products, the status is already correct. No stale `comped` state is possible in this flow.

PR #26110, opened 2026-02-02
Enable Stripe billing portal

sam-lord:
As well as re-enabling the billing portal, this PR also addresses the feedback from Peter:

* GVA-656 -- Disabling cancellation in the billing portal
* GVA-653 -- Changing the "Manage" button back to "Update"
* GVA-654 -- Re-open Portal when returning from the billing portal
* GVA-652 -- Card wasn't updated correctly after setting a default in the billing portal

The big one here is GVA-652, which required a little change to how we handle Stripe webhooks. We weren't handling the case where a customer has a default payment method (only the case where a subscription has a default payment method). I believe that's because the card update checkout flow just touched the subscription, not the customer.

I'd recommend reviewing the individual PRs to address the concerns individually. I'm making this as one big PR to avoid the individual changes being released in Portal, will need to stick a Portal release commit on the end of this PR when ready.


ghost/i18n/locales/context.json
```diff
@@ -26,7 +26,7 @@
     "Become a paid member of {publication} to start commenting.": "Call to action in the comments app",
     "Become a paid member of {site} to get access to all premium content.": "Call to action for paid content",
     "Best": "Text appears in the sorting selector for comments",
-    "Billing info": "A label for the user billing information section",
+    "Billing info & receipts": "",
```
aileen: This is the context file, so it shouldn't be empty

ghost/core/core/server/services/members/members-api/repositories/member-repository.js
```diff
@@ -1000,12 +1001,19 @@ module.exports = class MemberRepository {
...
         const stripeSubscriptionData = await this._stripeAPIService.getSubscription(data.subscription.id);
         let paymentMethodId;
-        if (!stripeSubscriptionData.default_payment_method) {
-            paymentMethodId = null;
-        } else if (typeof stripeSubscriptionData.default_payment_method === 'string') {
+        if (stripeSubscriptionData.default_payment_method) {
             paymentMethodId = stripeSubscriptionData.default_payment_method;
         } else {
-            paymentMethodId = stripeSubscriptionData.default_payment_method.id;
+            // If the subscription doesn't have a payment method, we fall back to the customer's default.
```
aileen, line 1010: Why would the subscription not have a payment method? What changed here 🤔

apps/portal/src/utils/api.js
```diff
@@ -583,52 +582,6 @@ function setupGhostApi({siteUrl = window.location.origin, apiUrl, apiKey}) {
             });
         },
```
aileen, line 585: Unless we introduce a breaking change for the data attributes, should this remain?

apps/portal/src/actions.js
```diff
@@ -337,20 +337,6 @@ async function applyOffer({data, state, api}) {
     }
 }
 
-async function editBilling({data, state, api}) {
```
aileen, line 338: Same as mentioned above, possible breaking change?

ghost/core/core/server/services/members/members-api/repositories/member-repository.js
aileen: If the default payment method primarily lives on the customer, this should probably be updated for global usage, but we still need to verify this.

aileen (changes requested):
I would recommend to look and investigate why the payment method moved from being accessible from the subscription to the customer and if we could generally switch to _just_ fetch it from the customer for consistency and to avoid another API call.

ghost/core/core/server/services/members/members-api/repositories/member-repository.js
sam-lord: It turns out this is a very big function (600+ lines), called in a lot of places.

Even if we could change to just looking up the customer, I don't think it's worth changing the calling pattern to include the customerId, we're better off getting the customer from the subscription.

ghost/core/core/server/services/members/members-api/repositories/member-repository.js
```diff
@@ -1001,12 +1001,19 @@ module.exports = class MemberRepository {
...
+            // This is set when the customer uses the Stripe billing portal to update their payment method.
+            const customerId = stripeSubscriptionData.customer;
+            const customer = await this._stripeAPIService.getCustomer(customerId);
+            const customerPaymentMethod = customer.invoice_settings?.default_payment_method;
+            if (customerPaymentMethod) {
+                paymentMethodId = customerPaymentMethod;
+            } else {
+                paymentMethodId = null;
+            }
```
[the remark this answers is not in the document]
sam-lord: I don't think we need to do this -- we're not handling multiple versions of the Stripe API. I think if we upgrade we'll need to do it wholesale given we don't do this sort of handling in the rest of the file.

aileen (approved):
Looks good to me now

PR #26120, opened 2026-02-02
Added transistor podcast landing page to Portal account details

9larsons:
ref https://linear.app/ghost/issue/NY-925/

Added Podcasts landing page link in Portal > Member Details when transistor.fm integration is enabled and the member has access to at least some shows.

## Summary by CodeRabbit

* **New Features**
  * Transistor Podcasts added to account pages: shows a Podcasts section with a Subscribe button linking to Transistor.fm for eligible members.

* **Localization**
  * Added "Podcasts" translation key across portal locales with a context entry.

* **Style**
  * Global frame styles extended to include a Podcasts action fade-in animation.

* **Tests**
  * New unit tests covering UI rendering and integration behavior for the Podcasts feature.


apps/portal/src/components/pages/AccountHomePage/components/account-actions.js
```diff
@@ -34,8 +37,6 @@ const AccountActions = () => {
         });
     };
 
-    // Extract helper functions for complex conditions
```
EvanHahn, line 37: ❤️

apps/portal/src/components/pages/AccountHomePage/components/transistor-podcasts-action.js
```diff
@@ -0,0 +1,42 @@
...
+
+const TransistorPodcastsAction = ({hasPodcasts, memberUuid}) => {
+    if (!hasPodcasts || !memberUuid) {
+        return null;
+    }
+
+    const transistorUrl = `https://partner.transistor.fm/ghost/${memberUuid}`;
+
+    return (
+        <section className="gh-portal-action-transistor">
```
EvanHahn: nit: there's some inconsistency about whether these use single or double quotes. I think we should use single quotes everywhere.

In the long term, an autoformatter should fix this.

apps/portal/src/components/pages/AccountHomePage/components/transistor-podcasts-action.js
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+`;
+
+const TransistorPodcastsAction = ({hasPodcasts, memberUuid}) => {
+    if (!hasPodcasts || !memberUuid) {
+        return null;
+    }
+
+    const transistorUrl = `https://partner.transistor.fm/ghost/${memberUuid}`;
```
EvanHahn, line 25: nit: is there any risk that `memberUuid` is some unexpected value? I don't want someone to, for example, put a query string in there.

apps/portal/src/components/pages/AccountHomePage/components/use-integrations.js
```diff
@@ -0,0 +1,45 @@
...
+
+        const checkTransistor = async () => {
+            try {
+                const response = await fetch(`https://partner.transistor.fm/ghost/member/${memberUuid}`);
+                if (response.ok) {
+                    const data = await response.json();
+                    setTransistorPodcasts(data.member === true);
+                }
+            } catch (e) {
+                // Silently fail - don't show the button if we can't reach Transistor
```
EvanHahn: Two nits:

1. I like this but I think it could be good to `console.warn()` or equivalent, to help us debug in the future.
2. This could also happen if we can't parse the response as JSON, or if `data` isn't an object.

```suggestion
                // Don't show the button if Transistor fails
                console.warn('Error in Transistor request', e);
```

apps/portal/src/components/pages/AccountHomePage/components/use-integrations.js
```diff
@@ -0,0 +1,45 @@
...
+    const [transistorPodcasts, setTransistorPodcasts] = useState(false);
+
+    useEffect(() => {
+        if (!isTransistorEnabled || !memberUuid) {
+            return;
+        }
+
+        const checkTransistor = async () => {
+            try {
+                const response = await fetch(`https://partner.transistor.fm/ghost/member/${memberUuid}`);
```
EvanHahn: nit: We should cancel this if the component unmounts. The `useEffect` callback should return something that cancels an `AbortController`, probably. And make sure to cancel both the `fetch` and the response handling.

apps/portal/test/unit/components/pages/AccountHomePage/transistor-podcasts-action.test.js
```diff
@@ -0,0 +1,164 @@
...
+            expect(queryByText('Podcasts')).not.toBeInTheDocument();
+            expect(queryByText('Subscribe')).not.toBeInTheDocument();
+        });
+
+        test('renders nothing when memberUuid is missing', () => {
+            const {queryByText} = setup({hasPodcasts: true, memberUuid: null});
+            expect(queryByText('Podcasts')).not.toBeInTheDocument();
+        });
+
+        test('renders nothing when both props are missing', () => {
```
EvanHahn: nit: IMO, both props should be required. If you agree, I think we should remove this test.

apps/portal/test/unit/components/pages/AccountHomePage/transistor-podcasts-action.test.js
```diff
@@ -0,0 +1,164 @@
...
+
+        test('Subscribe link opens in new tab', () => {
+            const {getByText} = setup({hasPodcasts: true, memberUuid: 'test-uuid-123'});
+            const link = getByText('Subscribe');
+            expect(link.getAttribute('target')).toBe('_blank');
+            expect(link.getAttribute('rel')).toBe('noopener noreferrer');
+        });
+    });
+
+    describe('Integrated with AccountActions', () => {
```
EvanHahn: nit: Should these be part of `AccountActions` tests instead?

apps/portal/test/unit/components/pages/AccountHomePage/transistor-podcasts-action.test.js
```diff
@@ -0,0 +1,164 @@
...
+                overrideContext: {site, member}
+            });
+
+            await waitFor(() => {
+                expect(window.fetch).toHaveBeenCalled();
+            });
+
+            expect(queryByText('Podcasts')).not.toBeInTheDocument();
+        });
+
```
EvanHahn, line 27: nit: Could you also test what happens if `fetch` returns a 200 response, but invalid JSON?

EvanHahn (changes requested):
LGTM other than several nitpicks.

apps/portal/src/components/pages/AccountHomePage/components/transistor-podcasts-action.js
9larsons: Only if someone manually manipulated their database. Given this is outgoing, it's on the transistor side to protect against odd query params (though we should do what we can to prevent questionable behavior...).

apps/portal/test/unit/components/pages/AccountHomePage/transistor-podcasts-action.test.js
9larsons: I had considered that - was on the fence about leaving all integration-based tests in a single file. Moved, though.

apps/portal/test/unit/components/pages/AccountHomePage/transistor-podcasts-action.test.js
9larsons: Good idea.

apps/portal/src/components/pages/AccountHomePage/components/transistor-podcasts-action.js
EvanHahn: Sounds good. If you want to do basic validation on this, great—but I think it's fine as is.

apps/portal/src/components/pages/AccountHomePage/components/use-integrations.js
```diff
@@ -0,0 +1,58 @@
...
+                // Don't show the button if Transistor fails
+                // eslint-disable-next-line no-console
+                console.warn('Error in Transistor request', e);
+            }
+        };
+
+        checkTransistor();
+
+        return () => controller.abort();
+    }, [isTransistorEnabled, memberUuid]);
```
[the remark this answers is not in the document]
EvanHahn: I don't know if this comment is worth addressing. This is theoretically a problem if the props ever change, because we could show stale data. If we ever expect `memberUuid` or `isTransistorEnabled` to change—which maybe we don't—then this probably isn't an issue.

apps/portal/test/unit/components/pages/AccountHomePage/account-actions.test.js
```diff
@@ -0,0 +1,136 @@
...
+
+            const {queryByText} = render(<AccountActions />, {
+                overrideContext: {site, member}
+            });
+
+            await waitFor(() => {
+                expect(window.fetch).toHaveBeenCalled();
+            });
+
+            expect(queryByText('Podcasts')).not.toBeInTheDocument();
```
EvanHahn, line 53: nit: Does this necessarily check the correct behavior? I imagine the following sequence being possible:

1. `window.fetch` is called, unblocking the `waitFor` above
2. `Podcasts` is not in the document because the React component hasn't updated, so this assertion passes
3. `Podcasts` gets added to the document

Not sure if this is an issue given the way `waitFor` works and the way these matchers work. But wanted to flag.

(This comment applies elsewhere in this file too.)

EvanHahn (approved):
Two non-blocking comments.

apps/portal/test/unit/components/pages/AccountHomePage/account-actions.test.js
9larsons: Good take. I suspect it's fine, though it's often good to be prudent.

PR #26127, opened 2026-02-03
Improved `yarn dev` usage with Stripe

kevinansfield:
no issue

Previously with the Dockerised `yarn dev` command, in order to enable Stripe webhooks it was necessary to modify your local `.env` file to add the `COMPOSE_PROFILES=stripe` line, and either specify Stripe keys in the `.env` or export ENV variables in your shell file.

- added `docker/stripe/with-stripe.sh` command that adds `stripe` to `COMPOSE_PROFILES` and performs requirement checks to enable early exit with a useful log rather than needing to dig into container logs to find the problem
- added `yarn dev:stripe` command to make Stripe availability easier to find than stumbling across the `.env.example` file
- updated `yarn dev:all` to use `with-stripe.sh` so Stripe webhook support is included alongside other services


docker/stripe/with-stripe.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo "================================================================================"
+    echo "ERROR: STRIPE_SECRET_KEY is not set"
+    echo ""
+    echo "To use the Stripe service, set STRIPE_SECRET_KEY in your .env file or ENV vars:"
+    echo "  STRIPE_SECRET_KEY=sk_test_..."
+    echo ""
+    echo "You can find your secret key at: https://dashboard.stripe.com/test/apikeys"
+    echo "================================================================================"
+    echo ""
+    exit 1
```
kevinansfield, line 36: This mirrors the error message output in the container itself via the [entrypoint.sh script](https://github.com/TryGhost/Ghost/blob/main/docker/stripe/entrypoint.sh#L30-L40).

I think it's useful to keep both because the stripe container can still be started without using this script if manually enabled via your `.env`/`COMPOSE_PROFILES`

sagzy (approved):
Love this @kevinansfield 

Makes using Stripe much more intuitive: run `yarn dev:stripe` → see error message on how to set the required private key → Stripe works. No need to remember another process!

jonatansberg (approved):

PR #26215, opened 2026-02-04
Removed non nullable constraint from `offers.product_id`

mike182uk:
ref https://linear.app/ghost/issue/BER-3257

Removed non nullable constraint from `offers.product_id` so that offers can be created that are not assigned to a specific tier. This is required for retention offers where an offer is applicate to the cadence (`monthly`, `yearly`) of any tier, and not the specific tier

---

> [!NOTE]
> **Medium Risk**
> Alters a production DB constraint on a foreign-keyed column; low code surface area, but schema changes can impact existing queries/assumptions and rollback behavior.
> 
> **Overview**
> Allows `offers` to be created without a linked tier by making `offers.product_id` nullable.
> 
> Adds a new `6.17` migration to set the column nullable (with foreign key checks disabled for safe rollback on some MySQL configs), updates the schema definition accordingly, and refreshes the schema integrity test hash.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 92c2c1547b5e25020470294f10655d2ea4b07565. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/6.17/2026-02-04-10-42-20-offers-nullable-product-id.js
+++ b/ghost/core/core/server/data/migrations/versions/6.17/2026-02-04-10-42-20-offers-nullable-product-id.js
@@ -1,3 +1,4 @@
 const {createSetNullableMigration} = require('../../utils');
 
+// We need to disable foreign key checks because if MySQL is missing the STRICT_TRANS_TABLES mode, we cannot revert the migration
 module.exports = createSetNullableMigration('offers', 'product_id', {disableForeignKeyChecks: true});
--- a/ghost/core/core/server/data/migrations/versions/6.17/2026-02-04-10-42-20-offers-nullable-product-id.js
+++ b/ghost/core/core/server/data/migrations/versions/6.17/2026-02-04-10-42-20-offers-nullable-product-id.js
@@ -1,0 +1,3 @@
+const {createSetNullableMigration} = require('../../utils');
+
+module.exports = createSetNullableMigration('offers', 'product_id', {disableForeignKeyChecks: true});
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/data/migrations/versions/6.17/2026-02-04-10-42-20-offers-nullable-product-id.js
```diff
@@ -0,0 +1,3 @@
+const {createSetNullableMigration} = require('../../utils');
+
+module.exports = createSetNullableMigration('offers', 'product_id', {disableForeignKeyChecks: true});
```
sagzy, line 4: Question: why do we need `{disableForeignKeyChecks: true}` here? As it's not very intuitive, shall we call out the reason with an inline comment?

Or better, if this applies to every migrations that sets a column to be nullable, we could move that directly to the utility function?

sagzy (approved):

ghost/core/core/server/data/migrations/versions/6.17/2026-02-04-10-42-20-offers-nullable-product-id.js
mike182uk: Good shout, we call this out in other migrations like this so should do the same here - Have added comment

> Or better, if this applies to every migrations that sets a column to be nullable, we could move that directly to the utility function?

It only applies to columns that have a foreign key constraint, not every nullable migration (the utility would need to introspect the schema to detect FK columns)

ghost/core/core/server/data/schema/schema.js
```diff
@@ -483,7 +483,7 @@ module.exports = {
         active: {type: 'boolean', nullable: false, defaultTo: true},
         name: {type: 'string', maxlength: 191, nullable: false, unique: true},
         code: {type: 'string', maxlength: 191, nullable: false, unique: true},
-        product_id: {type: 'string', maxlength: 24, nullable: false, references: 'products.id'},
+        product_id: {type: 'string', maxlength: 24, nullable: true, references: 'products.id'},
```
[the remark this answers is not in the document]
mike182uk: Implementation is updated in https://github.com/TryGhost/Ghost/pull/26221

sagzy (approved):

PR #26220, opened 2026-02-04
🐛 Fixed admin auth re-fetch overwriting paginated comments

rob-ghost:
When a user navigates to a comment permalink that requires loading multiple pages of comments, `initSetup` correctly paginates through all pages to find the target. However, `initAdminAuth` then independently re-fetches only page 1 via the admin API (which has a limit of 20), and overwrites the entire comment list with this truncated result. This causes the permalink target to disappear from the DOM, breaking both scroll-to and highlight behavior.

The root cause is that `initAdminAuth` used a direct `setState({...state, comments, pagination})` call with a stale closure over `state`, rather than the callback form. This meant it always replaced whatever comments were currently loaded with its own page-1 fetch. The fix switches to a callback-form `setState` that checks whether `initSetup` has already loaded beyond page 1 — if so, it preserves the existing comments and only updates admin-related state (`adminApi`, `admin`, `isAdmin`).

An E2E test reproduces the exact scenario: 25 comments with a permalink to the last one, with admin auth enabled. It verifies the target comment remains visible after admin auth completes its re-fetch.

---

> [!NOTE]
> **Low Risk**
> Small, localized state-update change plus an E2E regression test; low risk aside from potential edge cases around pagination detection.
> 
> **Overview**
> Fixes a permalink regression where `initAdminAuth` re-fetching admin comments (page 1 only) could overwrite comments that `initSetup` had already paginated in to reach a permalink target.
> 
> `initAdminAuth` now uses a functional `setState` and preserves the existing `comments`/`pagination` when the current pagination indicates multiple pages were loaded, updating only admin-related state in that case. Adds an E2E test to reproduce the 25-comment permalink scenario with admin auth enabled and assert the target comment remains visible after admin auth completes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b960d284206c9634efc1703f7b1f0b2c6744311a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented admin authentication from overwriting previously loaded paginated comments, so permalinks and multi-page comment views retain visible comments after admin auth.

* **Tests**
  * Added an end-to-end test validating admin auth does not replace member-paginated comments when following a permalink.

### DIFF
--- a/apps/comments-ui/src/app.tsx
+++ b/apps/comments-ui/src/app.tsx
@@ -140,11 +140,23 @@ const App: React.FC<AppProps> = ({scriptTag, initialCommentId, pageUrl}) => {
                 if (admin) {
                     // this is a bit of a hack, but we need to fetch the comments fully populated if the user is an admin
                     const adminComments = await adminApi.browse({page: 1, postId: options.postId, order: state.order, memberUuid: state.member?.uuid});
-                    setState({
-                        ...state,
-                        adminApi: adminApi,
-                        comments: adminComments.comments,
-                        pagination: adminComments.meta.pagination
+                    setState((currentState) => {
+                        // Don't overwrite paginated comments when initSetup loaded
+                        // multiple pages (e.g., for permalink scrolling)
+                        if (currentState.pagination && currentState.pagination.page > 1) {
+                            return {
+                                adminApi,
+                                admin,
+                                isAdmin: true
+                            };
+                        }
+                        return {
+                            adminApi,
+                            admin,
+                            isAdmin: true,
+                            comments: adminComments.comments,
+                            pagination: adminComments.meta.pagination
+                        };
                     });
                 }
             } catch (e) {
--- a/apps/comments-ui/test/e2e/permalink.test.ts
+++ b/apps/comments-ui/test/e2e/permalink.test.ts
@@ -1,6 +1,6 @@
 import {E2E_PORT} from '../../playwright.config';
 import {FrameLocator, Page} from '@playwright/test';
-import {MOCKED_SITE_URL, MockedApi, initialize} from '../utils/e2e';
+import {MOCKED_SITE_URL, MockedApi, initialize, mockAdminAuthFrame} from '../utils/e2e';
 import {expect, test} from '@playwright/test';
 
 /**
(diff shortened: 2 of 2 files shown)


apps/comments-ui/test/e2e/permalink.test.ts
```diff
@@ -363,4 +363,84 @@ test.describe('Comment Permalinks', async () => {
...
+        // which re-fetches comments via admin API (page 1 only, limit 20)
+        const adminFrame = page.locator('iframe[data-frame="admin-auth"]');
+        await expect(adminFrame).toBeAttached();
+
+        // Give admin auth time to complete and potentially overwrite state
+        await page.waitForTimeout(1000);
+
+        // The target comment (beyond admin page 1) must STILL be visible
+        await expect(commentsFrame.getByText('Target comment beyond admin page 1')).toBeVisible();
+    });
```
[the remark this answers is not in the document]
rob-ghost: Fixed, replaced with a semantic locator for admin actions which indicate admin auth is complete.

apps/comments-ui/src/app.tsx
```diff
@@ -140,11 +140,23 @@ const App: React.FC<AppProps> = ({scriptTag, initialCommentId, pageUrl}) => {
...
-                        pagination: adminComments.meta.pagination
+                    setState((currentState) => {
+                        // Don't overwrite paginated comments when initSetup loaded
+                        // multiple pages (e.g., for permalink scrolling)
+                        if (currentState.pagination && currentState.pagination.page > 1) {
+                            return {
+                                adminApi,
+                                admin,
+                                isAdmin: true
+                            };
```
jonatansberg, line 151: Wouldn't you still need to spread the current state here? Otherwise you'd anyways be dropping those fields, no?

jonatansberg (commented):
I'm not sure if this is correct or maybe I'm misunderstanding the code. Note that I haven't tested this locally (yet).

apps/comments-ui/src/app.tsx
rob-ghost: The `setState` wrapper does the shallow merge (lines 66-76) but its non-obvious from here. That merge would make spreading here redundant.

The `setState` wrapper is probably causing a bit of confusion here, what do you think would make it more obvious that this isn't dropping keys from the state, but overriding the returned keys instead?

rob-ghost: The "proper" fix would be to not use this `setState` anti-pattern and instead use `useReducer` for the complex state object, but its a refactor I'd rather do separately.

jonatansberg: Oh, I missed that part. That is indeed confusing and not the behavior I'd expect from a setState callback. I think we should definitely clean that up at some point, I'm not sure if that point is right now though.

jonatansberg (approved):
Thanks for clearing up my confusion. I think this is good to go. Are you confident in the change or would you like me to test it out locally as well?

rob-ghost:
@jonatansberg I've tested it locally and added a test for this scenario which reliably tests the bug so I'm pretty confident in this particular change.

Whether or not it fixes everything, we'll see.

PR #26229, opened 2026-02-04
🎨 Showed site timezone, instead of device timezone, in post analytics

EvanHahn:
towards https://linear.app/ghost/issue/NY-951

| Before (device in America/Chicago) | After (site in Pacific/Honolulu) |
|---|---|
| Before | After |

Previously, post analytics showed the time in the device's timezone. Now, they show the time in the site's timezone.

## Release Notes

* **New Features**
  * Post analytics now display dates and times with timezone-aware formatting, ensuring accurate time representation across different regions.

* **Tests**
  * Added comprehensive unit tests for timezone handling and date/time formatting utilities.


apps/posts/src/views/PostAnalytics/components/post-analytics-header.tsx
```diff
@@ -1,7 +1,38 @@
...
+    DropdownMenuSeparator,
+    DropdownMenuTrigger,
+    H1,
+    LucideIcon,
+    Navbar,
+    PageMenu,
+    PageMenuItem,
+    PostShareModal,
+    formatDisplayDate,
+    formatDisplayTime,
```
EvanHahn, line 31: These were the two imports I added.

apps/posts/test/.eslintrc.cjs
```diff
@@ -2,5 +2,9 @@ module.exports = {
     plugins: ['ghost'],
     extends: [
         'plugin:ghost/ts-test'
-    ]
+    ],
+    rules: {
+        // We aren't using Mocha so we can disable some Ghost test rules.
```
EvanHahn, line 7: I could've done this in a separate PR, but I think it's clearer to couple them with a patch that needs this.

apps/shade/src/lib/utils.ts
```diff
@@ -220,6 +220,17 @@ export const formatDisplayDate = (dateString: string, timezone?: string): string
...
 
+/**
+ * Format a plain time in a given time zone.
+ *
+ * @example
+ * formatDisplayTime('2020-04-20T18:09:12.345Z', 'Africa/Lagos')
+ * // 7:09pm
+ */
+export const formatDisplayTime = (dateString: string, timezone: string): string => (
+    moment(dateString).tz(timezone).format('h:mma')
```
EvanHahn, line 231: Note that this uses 12-hour time, not 24-hour time, like before. Happy to revert.

troyciesco: From what I can tell it's a complete mix across the codebase.

Screenshot 2026-02-05 at 9 22 50 AM
Screenshot 2026-02-05 at 9 21 19 AM

this seems consistent with what else is in shade, at least, so makes sense to me.

troyciesco (approved):
lgtm - just noting i saw the year disappeared too in the new version, but that looks correct with how that function works (i.e. doesn't show currentYear)

PR #26231, opened 2026-02-04
Renamed "sniper links" to "inbox links"

EvanHahn:
closes https://linear.app/ghost/issue/NY-1002

We need to bump Portal after this change. I did *not* handle the case where the server was still sending the `sniperLinks` key because, luckily, this feature has not been released yet.

`git grep -i sniper` returns no relevant results after this change.

## Summary by CodeRabbit

* **Chores**
  * Renamed "Sniper Links" feature to "Inbox Links" throughout the application to improve clarity and consistency. Changes include updated feature configuration in admin dashboard settings, refreshed UI component names and styling, modified authentication flows for signin, signup, and one-time code authentication processes, plus adjusted feature flag references.


9larsons (approved):
Should be good. Only concerns:
- API mismatch; I don't think it's a problem based on how it's consumed in portal
- flag in settings; I think we bulk update this endpoint so it'd clear out stale entries, and we don't have this enabled for any beta participants/sites afaik (also it's now GA)

PR #26238, opened 2026-02-04
Fixed flaky Admin X Settings acceptance tests

EvanHahn:
towards https://linear.app/ghost/issue/NY-1000

*This test-only change should have no user impact.*

These tests, which were a little flaky on my machine and on CI, are now more reliable. Basically, I'm just ensuring that things are visible before moving on, or relying on a little polling.

## Summary by CodeRabbit

* **Tests**
  * Improved reliability and consistency across migration tools, membership portal, and recommendations suites.
  * Consolidated navigations with a shared helper to reduce duplication.
  * Replaced ad-hoc waits with explicit visibility assertions for modals and sections.
  * Switched several assertions to polling-based checks to better handle asynchronous updates.

### DIFF
--- a/apps/admin-x-settings/test/acceptance/membership/portal.test.ts
+++ b/apps/admin-x-settings/test/acceptance/membership/portal.test.ts
@@ -111,8 +111,8 @@ test.describe('Portal Settings', async () => {
 
         await modal.getByRole('tab', {name: 'Look & feel'}).click();
 
-        await modal.getByRole('switch').click();
         await modal.getByRole('textbox').fill('become a member of something epic');
+        await modal.getByRole('switch').click();
         await modal.getByRole('button', {name: 'Save'}).click();
 
         await expect.poll(() => lastApiRequests.editSettings?.body).toEqual({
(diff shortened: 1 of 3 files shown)


apps/admin-x-settings/test/acceptance/membership/portal.test.ts
```diff
@@ -104,20 +103,19 @@ test.describe('Portal Settings', async () => {
...
-        await page.waitForSelector('[data-testid="portal-modal"]');
-
-        const modal = await page.getByTestId('portal-modal');
+        const modal = page.getByTestId('portal-modal');
+        await expect(modal).toBeVisible();
 
         await modal.getByRole('tab', {name: 'Look & feel'}).click();
 
         await modal.getByRole('switch').click();
         await modal.getByRole('textbox').fill('become a member of something epic');
```
9larsons:
```suggestion
await modal.getByRole('textbox').fill('become a member of something epic');
await modal.getByRole('switch').click();
```

My read is that these should be reversed - the fixtures set this to checked by default, and by unchecking it we hide the textbox, leading to a race condition as the frontend is trying to drop the content.

9larsons (changes requested):
Aside from this, I say the rest looks good and we'll see what CI has to say about it.

apps/admin-x-settings/test/acceptance/membership/portal.test.ts
EvanHahn: Done in eba7068eb8637302a14ac063135a3cd7055995fe.

9larsons (approved):

PR #26240, opened 2026-02-04
Increased MySQL setup timeout in CI

EvanHahn:
towards https://linear.app/ghost/issue/NY-1000

This allows more time for MySQL to start. [A recent CI failure][0] shows that this timed out after 3 minutes, and at least two images failed to pull in that time.

[0]: https://github.com/TryGhost/Ghost/actions/runs/21683986870/job/62525954605

## Summary by CodeRabbit

* **Chores**
  * Increased timeout for database initialization in CI/CD workflows to improve test execution reliability when using MySQL 8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -541,7 +541,7 @@ jobs:
 
       - uses: tryghost/mysql-action@main
         if: matrix.env.DB == 'mysql8'
-        timeout-minutes: 3
+        timeout-minutes: 5
         with:
           authentication plugin: 'caching_sha2_password'
           mysql version: '8.0'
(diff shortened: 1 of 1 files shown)


9larsons (approved):

9larsons:
question: Do we know that it's a slowness to start such that 3 minutes isn't enough? It looks like it was active in the logs - wish we had timestamps.

EvanHahn:
I admit I don't know that. But (1) I think we capture stderr, so I don't think we're missing error messages (2) it does look like a few images failed to pull in the time allotted.

I'm not sure this will fix any of our woes, but it might?

PR #26247, opened 2026-02-04
🐛 Fixed race condition in reply-to-reply comment form

rob-ghost:
The `reply to reply comment` E2E test was flaky because `openCommentForm` in comments-ui awaited `loadMoreReplies` before adding the reply form to `openCommentForms` state. When the replies API was slow — or when a concurrent `initAdminAuth` request tied up the network — the form never appeared within the test timeout, causing `replyEditor.focus()` to fail.

The fix splits `openCommentForm` into a sync and async part. A new sync action (`addOpenCommentForm`) immediately updates `openCommentForms` state so the reply form renders instantly, while `openCommentForm` continues to load additional replies in the background. This matches the existing pattern where `SyncActions` are dispatched via `setState` functional updates to avoid race conditions with other concurrent state changes.

A new E2E test (`reply-form-race.test.ts`) reproduces the race deterministically by intercepting the replies endpoint after initialization so it never responds, then clicking reply on a reply and asserting the form appears. This test fails 100% without the fix and passes reliably with it.

### DIFF
--- a/apps/comments-ui/package.json
+++ b/apps/comments-ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/comments-ui",
-  "version": "1.3.8",
+  "version": "1.3.9",
   "license": "MIT",
   "repository": "https://github.com/TryGhost/Ghost",
   "author": "Ghost Foundation",
--- a/apps/comments-ui/src/actions.ts
+++ b/apps/comments-ui/src/actions.ts
@@ -76,7 +76,7 @@ async function loadMoreReplies({state, api, data: {comment, limit}, isReply}: {s
         while (hasMore) {
             const data = await fetchReplies(afterReplyId, 100);
             allComments.push(...data.comments);
-            hasMore = !!data.meta.pagination.next;
+            hasMore = !!data.meta?.pagination?.next;
 
             if (data.comments && data.comments.length > 0) {
                 afterReplyId = data.comments[data.comments.length - 1]?.id;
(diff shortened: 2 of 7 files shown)


kevinansfield (approved):
All looks good to me 👌 I have vague recollections of a desire to show all replies before replying to a comment to help avoid posting without context into a conversation that has diverged but in hindsight I expect we were optimising for a bad assumption.

rob-ghost:
Yeah I think you're right as to the intention of what it was doing before, it would load replies, then show the box. That's what was causing the flaky test, the replies wouldn't load or not load in time and the box never appeared.

In reality, in production, it probably didn't happen much or at all.

So, do we change the behaviour for the sake of tests? Or do we go back to loading replies before showing the box? The question being, why did they load so slowly?

I think there's a good reason to not "jump" when you click to add a reply, but the jump will happen on submit now which refreshes the list or auto-expands the list.

@weylandswart any design opinion on this? When a user clicks reply to a comment, we open the reply box immediately, and on submission we refresh the replies to see your new reply added at the bottom.

Vs the previous behaviour of clicking reply, the replies auto-expand and the reply box opens at the bottom for you to submit.

rob-ghost:
@kevinansfield I've amend the change. Now it will load more replies if it needs to, and I kept two improvements from the last iteration:

1. Automatically scroll to your new comment / reply using permalinks
2. Refresh all replies when posting a new comment

So now it loads more if it has to, and refreshes on submit to show your new comment in the list along with any other new comments posted since.

The flakiness has hopefully been solved by improving error handling over `loadMoreReplies`, it would catch errors and silently fail before, now it should fall through and allow the comment box to open regardless.

PR #26262, opened 2026-02-05
🐛 Fixed admin auth overwriting permalink-expanded replies

rob-ghost:
When a logged-in admin visits a comment permalink, the comments UI expands replies and paginates to find the target comment during `initSetup`. However, `initAdminAuth` runs concurrently and re-fetches page 1 of comments, overwriting the expanded state. This meant the permalink target would vanish after admin auth completed — the user would briefly see the comment, then it would disappear.

The existing guard in `initAdminAuth` only checked for multi-page pagination (`currentState.pagination.page > 1`), which missed the case where replies had been expanded but comments were still on page 1. The fix adds an `initialCommentId` check so that when a permalink is active, admin auth never overwrites the comment data that `initSetup` prepared.

An E2E test reproduces the exact scenario: a parent comment with 5 replies where only 3 are returned by browse, the permalink targets reply 5, `initSetup` expands all replies, then admin auth completes — the test asserts the target reply remains visible throughout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where expanded replies in permalinks could be overwritten when admin authentication initializes. Comments now properly remain visible as previously loaded, maintaining state even after admin authentication context is applied.

* **Tests**
  * Added end-to-end test ensuring admin authentication does not overwrite expanded replies loaded through permalink anchors.

### DIFF
--- a/apps/comments-ui/src/app.tsx
+++ b/apps/comments-ui/src/app.tsx
@@ -141,9 +141,9 @@ const App: React.FC<AppProps> = ({scriptTag, initialCommentId, pageUrl}) => {
                     // this is a bit of a hack, but we need to fetch the comments fully populated if the user is an admin
                     const adminComments = await adminApi.browse({page: 1, postId: options.postId, order: state.order, memberUuid: state.member?.uuid});
                     setState((currentState) => {
-                        // Don't overwrite paginated comments when initSetup loaded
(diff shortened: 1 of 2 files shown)


jonatansberg (approved):
Looks like a reasonable fix. The whole thing is ripe for refactoring tho.

PR #26263, opened 2026-02-05
Refactored comments-ui to use single CommentApiProvider for auth

rob-ghost:
The previous comments-ui architecture had admin auth initialization happening inside the main App component as a side-effect that raced with comment loading. The raw GhostApi and AdminApi leaked into state, actions, and multiple components, making it difficult to reason about which API path was being used and when admin data was available.

This refactoring introduces a `CommentApiProvider` that resolves both the member session and admin auth before exposing a single unified `commentApi` interface to the rest of the app. The provider checks whether the admin auth frame is available (via a fetch to the auth-frame endpoint — a 204 means the user isn't logged into Ghost admin, so the iframe is skipped entirely), then sets up the admin API and verifies the user's role. The `App` component splits into `App` (creates the raw API, wraps in provider) and `AppInner` (consumes only `commentApi` from context). No component or action ever touches the raw `GhostApi` directly.

The `CommentApi` type is the single interface everything goes through — browse, replies, read, count, add, edit, like, unlike, report, updateMember, hideComment, and showComment. The factory function uses an early-return pattern for admin vs non-admin branches, which gives TypeScript clean narrowing without non-null assertions. Admin auth state is modeled as a tri-state (`AdminApi | null | undefined`) where undefined means pending, null means resolved-no-admin, and an AdminApi instance means resolved-admin.

Other cleanups bundled in: permalink scroll-target helpers extracted as pure functions (no closure over component state), dead parameters removed from `SyncActionHandler`, the `isReply` bypass in `loadMoreReplies` removed (it always goes through `commentApi.replies()` now), and immutable provider data (`isAdmin`, `labs`, `supportEmail`) moved out of mutable `EditableAppContext` into `AppContextType` where it flows directly from the provider.

This stacks on top of the bugfix PR and should be reviewed after that one merges.

### DIFF
--- a/apps/comments-ui/src/actions.ts
+++ b/apps/comments-ui/src/actions.ts
@@ -125,73 +125,6 @@ async function addReply({state, commentApi, data: {reply, parent}}: {state: Edit
     };
 }
 
-async function hideComment({state, commentApi, data: comment}: {state: EditableAppContext, commentApi: CommentApi, data: {id: string}}) {
-    if (commentApi.hideComment) {
-        await commentApi.hideComment(comment.id);
-    }
-    return {
-        comments: state.comments.map((c) => {
-            const replies = c.replies.map((r) => {
-                if (r.id === comment.id) {
-                    return {
-                        ...r,
-                        status: 'hidden'
-                    };
-                }
-
-                return r;
-            });
-
-            if (c.id === comment.id) {
-                return {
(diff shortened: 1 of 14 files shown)


rob-ghost:
Closing in favour of https://github.com/TryGhost/Ghost/pull/26297

PR #26270, opened 2026-02-05
Disabled Playwright `maxFailures: 1` when running in UI mode

kevinansfield:
no issue

- when running the e2e tests in UI mode you usually want all tests to run and then you can dig into individual failures with the UI still open, having the tests abort on the first failure is annoying
- changed `maxFailures` to use a conditional that checks for `--ui` argument, disabling early-exit when present

### DIFF
--- a/e2e/playwright.config.mjs
+++ b/e2e/playwright.config.mjs
@@ -22,7 +22,7 @@ const config = {
         timeout: process.env.CI ? 30 * 1000 : 10 * 1000
     },
     retries: 0, // Retries open the door to flaky tests. If the test needs retries, it's not a good test or the app is broken.
-    maxFailures: 1,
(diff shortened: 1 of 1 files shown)


e2e/playwright.config.mjs
```diff
@@ -22,7 +22,7 @@ const config = {
         timeout: process.env.CI ? 30 * 1000 : 10 * 1000
     },
     retries: 0, // Retries open the door to flaky tests. If the test needs retries, it's not a good test or the app is broken.
-    maxFailures: 1,
+    maxFailures: process.argv.includes('--ui') ? 0 : 1,
```
cmraible, line 25: suggestion (non-blocking): it might be more convenient to remove the maxFailures from the config entirely, and supply using the `--max-failures` CLI arg in CI

cmraible (approved):

PR #26272, opened 2026-02-05
Updated DOM structure and accessibility

troyciesco:
Made some tweaks so we don't have issues with nested buttons and strange click behavior, and added some accessibility updates.

---

> [!NOTE]
> **Low Risk**
> UI/DOM and test-selector changes only; main risk is regressions in click/focus behavior or E2E flakiness around the new loading/interaction guards.
> 
> **Overview**
> Refactors the welcome email preview row to avoid invalid nested interactive elements by splitting the clickable “Edit” area from the enable/disable `Toggle`, adding proper focus-visible ring handling and improving click behavior.
> 
> Tightens loading/mutation handling by preventing toggles/edits while initial data is loading or mutations are in-flight, and shows a small placeholder instead of a disabled switch during initial load. Updates Playwright page objects/tests to target the new `data-testid`/role-based structure and removes an assertion that the edit control disappears when disabled.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 391b0ef9a697f6fa6e3842c1fe6ccf3207751d5b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
+++ b/apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
@@ -20,14 +20,14 @@ const EmailPreview: React.FC<{
     automatedEmail: AutomatedEmail,
     emailType: 'free' | 'paid',
     enabled: boolean,
-    isLoading: boolean,
+    isInitialLoading: boolean,
     onEdit: () => void,
     onToggle: () => void
 }> = ({
     automatedEmail,
     emailType,
     enabled,
-    isLoading,
+    isInitialLoading,
     onEdit,
     onToggle
 }) => {
@@ -38,13 +38,15 @@ const EmailPreview: React.FC<{
     const senderName = automatedEmail.sender_name || siteTitle || 'Your Site';
 
     return (
-        <button
-            className='flex w-full cursor-pointer items-center justify-between gap-6 rounded-lg border border-grey-100 bg-grey-50 p-5 text-left transition-all hover:border-grey-200 hover:shadow-sm dark:border-grey-925 dark:bg-grey-975 dark:hover:border-grey-800'
+        <div
+            className='relative flex w-full items-center justify-between gap-6 rounded-lg border border-grey-100 bg-grey-50 p-5 text-left transition-all hover:border-grey-200 hover:shadow-sm dark:border-grey-925 dark:bg-grey-975 dark:hover:border-grey-800'
             data-testid={`${emailType}-welcome-email-preview`}
-            type='button'
-            onClick={onEdit}
         >
-            <div className='flex w-full items-center justify-between'>
+            <button
+                className='flex w-full cursor-pointer items-center justify-between before:absolute before:inset-0 before:rounded-lg before:content-[""] focus-visible:outline-none focus-visible:before:ring-2 focus-visible:before:ring-green'
+                type='button'
+                onClick={onEdit}
+            >
                 <div className='flex items-start gap-3'>
                     {icon ?
                         <div className='size-10 min-h-10 min-w-10 rounded-sm bg-cover bg-center' style={{
@@ -54,27 +56,29 @@ const EmailPreview: React.FC<{
                         <div className='flex aspect-square size-10 items-center justify-center overflow-hidden rounded-full p-1 text-white' style={{
                             backgroundColor: color
                         }}>
-                            <img alt="Logo" className='h-auto w-8' src={FakeLogo} />
+                            <img alt="" className='h-auto w-8' src={FakeLogo} />
                         </div>
                     }
-                    <div>
+                    <div className='text-left'>
                         <div className='font-semibold'>{senderName}</div>
                         <div className='text-sm'>{automatedEmail.subject}</div>
                     </div>
                 </div>
                 <div className='text-sm font-semibold opacity-100 transition-all hover:opacity-80'>
                     Edit
-                    {/* <Icon name='pen' size='sm' /> */}
                 </div>
+            </button>
+            <div className='relative z-10 rounded-full has-[:focus-visible]:ring-2 has-[:focus-visible]:ring-green'>
+                {isInitialLoading ? (
+                    <div className="h-4 w-7 rounded-full bg-grey-300 dark:bg-grey-800" />
+                ) : (
+                    <Toggle
+                        checked={enabled}
+                        onChange={onToggle}
+                    />
+                )}
             </div>
-            <div onClick={e => e.stopPropagation()}>
-                <Toggle
-                    checked={enabled}
-                    disabled={isLoading}
-                    onChange={onToggle}
-                />
-            </div>
-        </button>
+        </div>
     );
 };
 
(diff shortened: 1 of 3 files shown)


apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
```diff
@@ -38,13 +38,15 @@ const EmailPreview: React.FC<{
     const senderName = automatedEmail.sender_name || siteTitle || 'Your Site';
 
     return (
-        <button
-            className='flex w-full cursor-pointer items-center justify-between gap-6 rounded-lg border border-grey-100 bg-grey-50 p-5 text-left transition-all hover:border-grey-200 hover:shadow-sm dark:border-grey-925 dark:bg-grey-975 dark:hover:border-grey-800'
+        <div
```
troyciesco, line 41: since this is a container, it should be a div - but we can still make the whole thing clickable, i'll explain in another comment.

apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
```diff
@@ -38,13 +38,15 @@ const EmailPreview: React.FC<{
...
+            className='relative flex w-full items-center justify-between gap-6 rounded-lg border border-grey-100 bg-grey-50 p-5 text-left transition-all hover:border-grey-200 hover:shadow-sm dark:border-grey-925 dark:bg-grey-975 dark:hover:border-grey-800'
             data-testid={`${emailType}-welcome-email-preview`}
-            type='button'
-            onClick={onEdit}
         >
-            <div className='flex w-full items-center justify-between'>
+            <button
+                className='flex w-full cursor-pointer items-center justify-between before:absolute before:inset-0 before:rounded-lg before:content-[""] focus-visible:outline-none focus-visible:before:ring-2 focus-visible:before:ring-green'
```
troyciesco, line 46: couple changes to this button:
- Using this :before pseudo-class makes it so that we can make that entire container a clickable button without having the issue of the toggle being a child of a button. Now with this, the button can sit right next to the toggle and they both work and are accessible. 
- Adding the focus visible makes it obvious if you're tabbing through the interactive pieces of settings where you actually are. 

(below img shows me hovering the button, that's why there's a shadow lol)

Image

apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
```diff
@@ -54,27 +56,29 @@ const EmailPreview: React.FC<{
                         <div className='flex aspect-square size-10 items-center justify-center overflow-hidden rounded-full p-1 text-white' style={{
                             backgroundColor: color
                         }}>
-                            <img alt="Logo" className='h-auto w-8' src={FakeLogo} />
+                            <img alt="" className='h-auto w-8' src={FakeLogo} />
```
troyciesco, line 59: Since this is decorative i don't think we need actual content in the alt

apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
```diff
@@ -54,27 +56,29 @@ const EmailPreview: React.FC<{
...
                     </div>
                 </div>
                 <div className='text-sm font-semibold opacity-100 transition-all hover:opacity-80'>
                     Edit
-                    {/* <Icon name='pen' size='sm' /> */}
                 </div>
+            </button>
+            <div className='relative z-10 rounded-full has-[:focus-visible]:ring-2 has-[:focus-visible]:ring-green'>
+                {isInitialLoading ? (
+                    <div className="h-4 w-7 rounded-full bg-grey-300 dark:bg-grey-800" />
```
troyciesco, line 73: This is the part that I added that I feel the least good about, but what was happening was that the toggle was not appropriately getting the initial state of the row. So whenever you reloaded the page, it would show as inactive whether or not the email was actually active. 

I didn't want to dig into the weeds of the toggle component itself because I know it's used in a lot of spots, but by checking for this initial load and having this fake loader that doesn't actually do anything means that we can render the toggle with the appropriate state as soon as it's ready and it will work properly. And this also helps avoid layout shift. 

fwiw this is basically never seen

apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
```diff
@@ -54,27 +56,29 @@ const EmailPreview: React.FC<{
...
+                ) : (
+                    <Toggle
+                        checked={enabled}
+                        onChange={onToggle}
+                    />
+                )}
             </div>
-            <div onClick={e => e.stopPropagation()}>
```
troyciesco, line 70: Since the button and the toggle are siblings now, as opposed to parent-child, we don't need the stop propagation anymore.

apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
```diff
@@ -142,6 +148,10 @@ const MemberEmails: React.FC<{ keywords: string[] }> = ({keywords}) => {
         const existing = automatedEmails.find(email => email.slug === slug);
         const label = emailType === 'free' ? 'Free members' : 'Paid members';
 
+        if (isBusy) {
```
troyciesco, line 151: These are just to avoid potential issues if you're clicking really fast or whatever.

apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
```diff
@@ -54,27 +56,29 @@ const EmailPreview: React.FC<{
...
-                    {/* <Icon name='pen' size='sm' /> */}
                 </div>
+            </button>
+            <div className='relative z-10 rounded-full has-[:focus-visible]:ring-2 has-[:focus-visible]:ring-green'>
+                {isInitialLoading ? (
+                    <div className="h-4 w-7 rounded-full bg-grey-300 dark:bg-grey-800" />
+                ) : (
+                    <Toggle
+                        checked={enabled}
+                        onChange={onToggle}
```
troyciesco, line 77: Removed the disabled state here because it was causing a weird little flash of the not allowed pointer, and that didn't feel ideal. We are still preventing unnecessary updates using the isBusy variable below.

apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
```diff
@@ -197,11 +211,10 @@ const MemberEmails: React.FC<{ keywords: string[] }> = ({keywords}) => {
                     title='Free members welcome email'
                 />
                 <EmailPreview
-                    key={`free-${freeWelcomeEmailEnabled}`}
```
troyciesco, line 200: should only need keys in a list, to keep track of order/position of an item. I think having it here might cause an unneeded rerender every time one of the toggles is clicked

apps/admin-x-settings/src/components/settings/membership/member-emails.tsx
```diff
@@ -212,11 +225,10 @@ const MemberEmails: React.FC<{ keywords: string[] }> = ({keywords}) => {
                             title='Paid members welcome email'
                         />
                         <EmailPreview
-                            key={`paid-${paidWelcomeEmailEnabled}`}
                             automatedEmail={paidEmailForDisplay}
                             emailType='paid'
                             enabled={paidWelcomeEmailEnabled}
-                            isLoading={isLoading}
```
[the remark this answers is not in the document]
troyciesco: yeah but we don't show the toggle now until the initial load is done. 

with the key and extra remounts it breaks the toggle animation, and i'm also pretty sure the toggle state ends up incorrect when the page is first loaded (i.e. it'll always show off when you first load, even when the email is active)

PR #26277, opened 2026-02-05
Added pagination to thread sidebar view

kevinansfield:
ref https://linear.app/ghost/issue/BER-3283

- Added "Load more replies" button to thread sidebar when there are more than 100 replies
- Wired up existing `fetchNextPage`, `hasNextPage`, `isFetchingNextPage` from `useThreadComments` infinite query
- Button shows loading state while fetching and hides when all replies are loaded

### DIFF
--- a/apps/posts/src/views/comments/components/comment-thread-list.tsx
+++ b/apps/posts/src/views/comments/components/comment-thread-list.tsx
@@ -1,6 +1,6 @@
 import CommentContent from './comment-content';
 import React from 'react';
-import {Button, LucideIcon} from '@tryghost/shade';
+import {Button, LoadingIndicator, LucideIcon} from '@tryghost/shade';
 import {Comment, useHideComment, useShowComment} from '@tryghost/admin-x-framework/api/comments';
 import {CommentAvatar} from './comment-avatar';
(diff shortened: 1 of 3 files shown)


apps/posts/src/views/comments/components/comment-thread-list.tsx
```diff
@@ -130,13 +130,19 @@ interface CommentThreadListProps {
     replies: Comment[];
     selectedCommentId: string;
     commentPermalinksEnabled?: boolean;
+    fetchNextPage?: () => void;
+    hasNextPage?: boolean;
+    isFetchingNextPage?: boolean;
```
rob-ghost: **Nitpick:** AFAIK, `fetchNextPage` and `isFetchingNextPage` are not optional in the infinite query response, so we could make these required and not worry about the `null|undefined` case (oddly `hasNextPage` _is_ optional):

```typescript
fetchNextPage: () => void;
hasNextPage?: boolean;
isFetchingNextPage: boolean
```

Just so we don't have to do something like `fetchNextPage?.()`.

rob-ghost (approved):
Non-blocking nitpick comment, nice 👍🏻

PR #26279, opened 2026-02-05
Updated transistor renderer

9larsons:
ref https://linear.app/ghost/issue/NY-1005/
- added dynamic replacement of {uuid} using the same mechanism as content-gating at the API level
- updated renderer
- added tests

This PR fully wires up the substitution of `{uuid}` in the outputted content (API response). The token is set within the rendered html content so that we don't need to re-render the content at request time for each member. This instead allows us to do it at the API, just like we do with content visibility, which is also leveraged by this functionality. 

This handles the actual display on the frontend. This is difficult to test because you need the integration set up with an available domain - you can see the build on site 352618 on staging.

This does not change the email rendering pipeline.

The Editor > Post Preview > Web view is still broken - we have to follow up with changes to inject a valid uuid in order to be able to preview it there. That is tracked separately.


ghost/core/test/unit/server/services/koenig/node-renderers/transistor-renderer.test.js
```diff
@@ -0,0 +1,124 @@
+const assert = require('assert/strict');
```
9larsons, line 1: I should've added these tests with the original renderer commit.

ghost/core/core/server/services/koenig/node-renderers/transistor-renderer.js
```diff
@@ -34,7 +29,7 @@ function frontendTemplate(node, document, options) {
 
     const iframe = document.createElement('iframe');
     iframe.setAttribute('width', '100%');
-    iframe.setAttribute('height', '400');
+    iframe.setAttribute('height', '325');
```
9larsons, line 32: This matches the content height based on the embed. You can see this here with https://partner.transistor.fm/ghost/embed/7d75f320-d0ed-4e6a-9ef5-840163033dbd for the moment.

ghost/core/core/server/services/koenig/node-renderers/transistor-renderer.js
```diff
@@ -16,14 +16,9 @@ function renderTransistorNode(node, options = {}) {
...
-    const memberUuid = options.memberUuid;
 
-    if (!memberUuid) {
-        // Transistor does not support public/non-member embeds for now, so we return null
-        return null;
-    }
-
-    const embedUrl = new URL(`https://partner.transistor.fm/ghost/embed/${memberUuid}`);
+    // Use {uuid} placeholder - content.js will substitute with member UUID at request time
+    const embedUrl = new URL(`https://partner.transistor.fm/ghost/embed/{uuid}`);
```
9larsons, line 21: We no longer get to null check here because it must get substituted at the API level.

ghost/core/core/server/api/endpoints/utils/serializers/output/utils/post-gating.js
```diff
@@ -110,6 +110,12 @@ const forPost = (attrs, frame) => {
         _updateTextAttrs(attrs);
     }
 
+    // Replace member UUID placeholder for Transistor embeds (URL-encoded {uuid})
+    const member = frame.original.context.member;
+    if (member && member.uuid && attrs.html) {
+        attrs.html = attrs.html.replace(/%7Buuid%7D/gi, member.uuid);
+    }
```
9larsons, line 117: It seemed best to handle this as encoded, though I'm ambivalent here.

ghost/core/test/unit/api/canary/utils/serializers/output/utils/post-gating.test.js
```diff
@@ -91,6 +91,72 @@ describe('Unit: endpoints/utils/serializers/output/utils/post-gating', function
...
+                const attrs = {
+                    visibility: 'public',
+                    html: '<iframe src="https://partner.transistor.fm/ghost/embed/{uuid}"></iframe>'
+                };
+
+                frame.original.context.member = {uuid: 'test-member-uuid-456'};
+
+                gating.forPost(attrs, frame);
+
+                // Non-encoded {uuid} should NOT be replaced
```
EvanHahn: nit: I think this comment is unnecessary, based on the title of the test.

```suggestion
```

ghost/core/test/unit/server/services/koenig/node-renderers/transistor-renderer.test.js
```diff
@@ -0,0 +1,124 @@
...
+    function renderForEmail(data, options) {
+        return callRenderer('transistor', data, {...options, target: 'email'});
+    }
+
+    describe('web', function () {
+        it('renders iframe with URL-encoded %7Buuid%7D placeholder', function () {
+            const result = renderForWeb(getTestData());
+
+            assert.ok(result.html);
+            // URL constructor encodes {uuid} to %7Buuid%7D
```
EvanHahn: nit: I think this comment is unnecessary.

```suggestion
```

ghost/core/test/unit/server/services/koenig/node-renderers/transistor-renderer.test.js
```diff
@@ -0,0 +1,124 @@
...
+
+    function renderForEmail(data, options) {
+        return callRenderer('transistor', data, {...options, target: 'email'});
+    }
+
+    describe('web', function () {
+        it('renders iframe with URL-encoded %7Buuid%7D placeholder', function () {
+            const result = renderForWeb(getTestData());
+
+            assert.ok(result.html);
```
EvanHahn: nit: I think all of these `assert.ok(result.html)`s are unnecessary because you immediately check something about them later. I'd probably remove unless you're getting type errors.

ghost/core/test/unit/server/services/koenig/node-renderers/transistor-renderer.test.js
```diff
@@ -0,0 +1,124 @@
...
+            assert.ok(result.html.includes('color: #ff5500') || result.html.includes('color:#ff5500'));
+        });
+
+        it('uses default accent color when not provided', function () {
+            const result = renderForEmail(getTestData({accentColor: null}));
+
+            assert.ok(result.html);
+            // Default is #15171A
```
EvanHahn: nit: I think it's clear without this comment.

```suggestion
```

ghost/core/core/server/services/koenig/node-renderers/transistor-renderer.js
EvanHahn: A little concerned that we have a fixed height and `scrolling=no` below. But that's probably an issue for another ticket.

EvanHahn (approved):
Some non-blocking comments.

I think the `ref` in the PR description is wrong. (Should it be [NY-998]?)

[NY-998]: https://linear.app/ghost/issue/NY-998

PR #26282, opened 2026-02-05
🐛 Fixed Unsplash integration not loading when selecting Publication Cover image

cmraible:
ref https://linear.app/ghost/issue/ONC-1453/unsplash-integration-not-loading

## Problem

In Admin Settings > Design and Branding, the unsplash selector for the Publication Cover isn't working properly. Rather than showing Unsplash images, it's just showing a blank white screen on the Unsplash modal. the Unsplash selector on the main Post editor is working fine.

## Diagnosis

When loading the Unsplash selector for Publication Cover, the request to Unsplash returns a 401: Unauthorized error. Upon further inspection, the new React admin wrapper in apps/admin is supplying blank values for the unsplash configuration.

## Fix

Supply the same unsplash configuration in apps/admin that we do in Ember admin.

### DIFF
--- a/apps/admin/src/main.tsx
+++ b/apps/admin/src/main.tsx
@@ -15,10 +15,10 @@ const framework = {
         navigateTo(link.route);
     },
     unsplashConfig: {
-        Authorization: "",
-        "Accept-Version": "",
-        "Content-Type": "",
-        "App-Pragma": "",
+        Authorization: "Client-ID 8672af113b0a8573edae3aa3713886265d9bb741d707f6c01a486cde8c278980",
+        "Accept-Version": "v1",
+        "Content-Type": "application/json",
+        "App-Pragma": "no-cache",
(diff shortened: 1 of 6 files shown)


apps/admin/src/main.tsx
```diff
@@ -15,10 +15,10 @@ const framework = {
         navigateTo(link.route);
     },
     unsplashConfig: {
-        Authorization: "",
-        "Accept-Version": "",
-        "Content-Type": "",
-        "App-Pragma": "",
+        Authorization: "Client-ID 8672af113b0a8573edae3aa3713886265d9bb741d707f6c01a486cde8c278980",
```
cmraible, line 18: Note: this felt wrong, but this isn't meant to be a secret value. This is also hardcoded in Ember admin and has been forever. It's just an identifier for the Ghost application, which is exposed on the frontend anyways, not a secret key or anything like that.

PR #26285, opened 2026-02-05
🐛 Fixed 404 when navigating between post stats and editor

troyciesco:
Closes [ONC-1421: 404 when navigating between post analytics & editor](https://linear.app/ghost/issue/ONC-1421/404-when-navigating-between-post-analytics-and-editor)

Problem
- Opening editor from analytics set a bad back path, so `< Analytics` could point to an invalid post URL and 404.

Solution
- In `lexical-editor.js`, derive the previous analytics path from transition route info instead of guessing.
- Handle both analytics families (`posts-x`, `stats-x`), including subroutes and query params.
- Store the resolved path in `controller.fromAnalytics` for existing posts only.
- Reset `fromAnalytics` on route reset to prevent stale breadcrumbs.
- Add acceptance coverage in `editor-test.js` for:
  - post analytics root/subpage
  - stats root/subpage/query params
  - new post (no analytics breadcrumb)
  - default posts breadcrumb fallback

stat-edit-post-stat

### DIFF
--- a/ghost/admin/app/routes/lexical-editor.js
+++ b/ghost/admin/app/routes/lexical-editor.js
@@ -3,6 +3,97 @@ import AuthenticatedRoute from 'ghost-admin/routes/authenticated';
 import {run} from '@ember/runloop';
 import {inject as service} from '@ember/service';
 
+const POSTS_ANALYTICS_ROUTE = 'posts-x';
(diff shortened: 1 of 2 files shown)


cmraible (approved):
LGTM!

PR #26294, opened 2026-02-09
🐛 Fixed automatic Medium.com bookmark creation

kevinansfield:
ref https://linear.app/ghost/issue/ONC-1429/
ref https://linear.app/ghost/issue/ONC-1451/

- Medium.com and Itch.com have added Cloudflare bot protection (or CF bot protection has changed) on their oEmbed endpoint that was preventing our oEmbed service from using it to create embeds when pasting links
- Changed the user-agent our oEmbed service uses to match latest Chrome which appears to bypass the current Cloudflare restrictions
  - the effectiveness of this technique for bypassing itch.com restrictions hasn't been consistent in testing but has shown improvement in some cases

### DIFF
--- a/ghost/core/core/server/services/oembed/oembed-service.js
+++ b/ghost/core/core/server/services/oembed/oembed-service.js
@@ -10,7 +10,7 @@ const iconv = require('iconv-lite');
 const path = require('path');
 
 // Some sites block non-standard user agents so we need to mimic a typical browser
-const USER_AGENT = 'Mozilla/5.0 (compatible; Ghost/5.0; +https://ghost.org/)';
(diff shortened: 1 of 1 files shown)


cmraible:
@kevinansfield thanks! I've tried to make these same requests with the updated user agent from a staging server, and I've still gotten 403s — I think the requests are being blocked partly because they're coming from datacenter IPs. I'm going to deploy this to a staging site to make sure it's not something I'm doing wrong, but I don't think this will be enough to get around the blocks in staging/production.

cmraible (approved):
I've taken this for a spin on staging:
- ✅ Medium links work again, so this should fix [ONC-1429](https://linear.app/ghost/issue/ONC-1429/bookmark-embeds-not-working-for-medium-links)
- ❌ itch.io links still do not, so this does not resolve [ONC-1451](https://linear.app/ghost/issue/ONC-1451/bookmark-previews-no-longer-working)

That said, Medium is probably much more commonly used, so this seems like a worthwhile fix to me. 🚢

PR #26297, opened 2026-02-09
Refactored comments-ui to use CommentApiProvider for API abstraction

rob-ghost:
The comments-ui codebase had a recurring code smell: every action that needed to fetch data contained inline `if (state.admin && state.adminApi) { ... } else { ... }` conditionals to choose between the admin API and the member API. This duplicated branching logic was scattered across `loadMoreComments`, `setOrder`, `loadMoreReplies`, `showComment`, and other functions.

This PR introduces a `CommentApi` polymorphic facade that abstracts away the admin/member distinction. Actions now call methods like `commentApi.browse()` without knowing which underlying API is being used. The facade uses a discriminated union (`MemberCommentApi | AdminCommentApi`) so TypeScript enforces correct usage—admin-only methods like `hideComment` are only available when `commentApi.isAdmin` is true.

The new `CommentApiProvider` centralizes several things that were previously scattered in `app.tsx`:
- Admin authentication via AuthFrame
- Role checking against allowed moderators
- Member initialization via `api.init()`
- Creation of the appropriate CommentApi facade

The provider uses a preflight HEAD request to check if an admin session cookie exists before loading the AuthFrame iframe. If the server returns 204 (no session), we skip the iframe entirely and resolve as non-admin immediately. This avoids the overhead of loading the iframe for non-admin users and eliminates the need for arbitrary timeouts.

`EditableAppContext` is cleaned up by removing `admin` and `adminApi` fields—these are now internal to the provider. The provider exposes `member`, `labs`, and `supportEmail` from the init response, plus the resolved `commentApi`.

## Review notes

A couple of admin-only action instances (`hideComment`, `showComment`) still have `commentApi.isAdmin` guards in `actions.ts`. These are being extracted into a separate `AdminActionsProvider` in the follow-up PR `refactor/comments-admin-actions`, which leverages the discriminated union to remove those guards entirely.

### DIFF
--- a/apps/comments-ui/src/app.tsx
+++ b/apps/comments-ui/src/app.tsx
@@ -1,16 +1,14 @@
 /* eslint-disable no-shadow */
 
-import AuthFrame from './auth-frame';
 import ContentBox from './components/content-box';
 import PopupBox from './components/popup-box';
-import React, {useCallback, useEffect, useMemo, useState} from 'react';
+import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import i18nLib from '@tryghost/i18n';
 import setupGhostApi from './utils/api';
 import {ActionHandler, SyncActionHandler, isSyncAction} from './actions';
 import {AppContext, Comment, DispatchActionType, EditableAppContext} from './app-context';
+import {CommentApiProvider, useCommentApi} from './components/comment-api-provider';
 import {CommentsFrame} from './components/frame';
-import {createCommentApi} from './components/comment-api-provider';
-import {setupAdminAPI} from './utils/admin-api';
 import {useOptions} from './utils/options';
 
 type AppProps = {
@@ -19,8 +17,6 @@ type AppProps = {
     pageUrl: string;
 };
 
-const ALLOWED_MODERATORS = ['Owner', 'Administrator', 'Super Editor'];
-
 /**
  * Check if a comment ID exists in the comments array (either as a top-level comment or reply)
  */
@@ -28,12 +24,15 @@ function isCommentLoaded(comments: Comment[], targetId: string): boolean {
     return comments.some(c => c.id === targetId || c.replies?.some(r => r.id === targetId));
 }
 
-const App: React.FC<AppProps> = ({scriptTag, initialCommentId, pageUrl}) => {
-    const options = useOptions(scriptTag);
+const AppContent: React.FC<{
+    options: ReturnType<typeof useOptions>;
+    initialCommentId: string | null;
+    pageUrl: string;
+}> = ({options, initialCommentId, pageUrl}) => {
+    const {commentApi, member, labs, supportEmail} = useCommentApi();
     const [state, setFullState] = useState<EditableAppContext>({
         initStatus: 'running',
         member: null,
-        admin: null,
         comments: [],
         pagination: null,
         commentCount: 0,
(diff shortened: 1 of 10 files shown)


apps/comments-ui/src/components/comment-api-provider.tsx
```diff
@@ -0,0 +1,177 @@
...
+    return (
+        <>
+            {hasSession && <AuthFrame adminUrl={adminUrl} onLoad={onAdminAuthLoad} />}
+            {contextValue && (
+                <CommentApiContext.Provider value={contextValue}>
+                    {children}
+                </CommentApiContext.Provider>
+            )}
+        </>
+    );
```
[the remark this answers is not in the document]
rob-ghost: Fixed with the `onLoad` and `onError` callback combo, I think that's good enough.

apps/comments-ui/src/app.tsx
```diff
@@ -388,10 +321,34 @@ const App: React.FC<AppProps> = ({scriptTag, initialCommentId, pageUrl}) => {
...
+const App: React.FC<AppProps> = ({scriptTag, initialCommentId, pageUrl}) => {
+    const options = useOptions(scriptTag);
+
+    const api = useMemo(() => {
+        return setupGhostApi({
+            siteUrl: options.siteUrl,
+            apiUrl: options.apiUrl!,
+            apiKey: options.apiKey!
+        });
+    }, [options]);
```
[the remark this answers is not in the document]
rob-ghost: Pre-existing, just showing up in the diff due to code moves. Out-of-scope for now.

apps/comments-ui/src/components/comment-api-provider.tsx
```diff
@@ -0,0 +1,183 @@
...
+export type CommentApi = MemberCommentApi | AdminCommentApi;
+
+const ALLOWED_MODERATORS = ['Owner', 'Administrator', 'Super Editor'];
+
+function createCommentApi(api: GhostApi, adminApi: AdminApi | null, memberUuid?: string): CommentApi {
+    if (adminApi) {
+        return {
+            isAdmin: true,
+            browse: p => adminApi.browse({...p, memberUuid}),
+            replies: p => adminApi.replies({...p, memberUuid}),
```
kevinansfield, line 51: Nit: I think the type for `adminApi.replies` needs updating to accept `limit: 'all'`

apps/comments-ui/src/components/comment-api-provider.tsx
```diff
@@ -0,0 +1,183 @@
...
+    children: React.ReactNode;
+    api: GhostApi;
+    adminUrl: string | undefined;
+};
+
+export function CommentApiProvider({children, api, adminUrl}: CommentApiProviderProps) {
+    // undefined = pending, value = resolved
+    const [memberData, setMemberData] = useState<MemberData | undefined>(undefined);
+    const [adminApi, setAdminApi] = useState<AdminApi | null | undefined>(adminUrl ? undefined : null);
+    const [hasSession, setHasSession] = useState<boolean | undefined>(adminUrl ? undefined : false);
```
kevinansfield: Nit: Renaming to `hasAdminSession` would more accurately describe its usage

kevinansfield (approved):
This is much cleaner! 🚀

apps/comments-ui/src/components/comment-api-provider.tsx
rob-ghost: Actually `replies` doesn't support `limit: 'all'` since that's on the app side, not the API side AFAIK. I think the underlying type declaration was wrong, I've updated it.

ErisDS:
- Workflow: `CI`
- Failed Step: `Run yarn nx run @tryghost/comments-ui:test`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/23005387148)
**Action:**

PR #26301, opened 2026-02-09
Added React Query for comment/reply submission

rob-ghost:
This branch reimplements the comment/reply submission using React Query instead of manual fetch calls. The base branch already refactored submission to refetch from the server after posting, which ensures concurrent replies from other users become visible. This branch takes the same approach but uses React Query's mutation primitives.

Components now call `useAddReply()` and `useAddComment()` hooks directly rather than dispatching actions. The mutations handle the API call and refetch, then invoke sync actions to update app state. This adds `@tanstack/react-query` as a dependency (~15KB gzipped) and introduces a new `utils/query.ts` file with the mutation hooks.

Both branches pass the same tests and achieve identical behavior. This PR exists to compare the two approaches and decide whether React Query's declarative pattern justifies the added dependency and complexity for this use case.

### DIFF
--- a/apps/comments-ui/package.json
+++ b/apps/comments-ui/package.json
@@ -44,6 +44,7 @@
   },
   "dependencies": {
     "@headlessui/react": "1.7.19",
+    "@tanstack/react-query": "^4",
     "@tiptap/core": "2.26.3",
     "@tiptap/extension-blockquote": "2.26.3",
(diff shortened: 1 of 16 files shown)


rob-ghost:
Closing as we don't have the appetite for such a large-scale refactor to introduce React Query right now.

PR #26304, opened 2026-02-09
Fixed template token formatting in welcome emails

troyciesco:
Closes [NY-995: Fix Mono Font Styling on Welcome Email Template Variables](https://linear.app/ghost/issue/NY-995/fix-mono-font-styling-on-welcome-email-template-variables)

- Replacement strings like `{name}` or `{email}` were showing up with code formatting when the email was sent
- This PR removes the `<code>` html that gets wrapped around those before rendering, while making sure that we still properly format real code blocks

Before:
Screenshot 2026-02-03 at 3 21 45 PM

After:
Screenshot 2026-02-03 at 3 20 01 PM

With all the variables plus some code that _should_ still be styled:
Screenshot 2026-02-10 at 4 52 53 PM

### DIFF
--- a/ghost/core/core/server/services/member-welcome-emails/member-welcome-email-renderer.js
+++ b/ghost/core/core/server/services/member-welcome-emails/member-welcome-email-renderer.js
@@ -53,14 +53,12 @@ class MemberWelcomeEmailRenderer {
      * Applies replacement tokens to a string
      * Supports fallback values: {first_name, "friend"} renders "friend" if name is empty
      * @param {Object} options
+     * @param {{id: string, getValue: () => string|undefined}[]} options.definitions - Replacement token definitions
      * @param {string} options.text - The text to process (content body or subject line)
-     * @param {Object} options.member - Member data
-     * @param {Object} options.siteSettings - Site settings
      * @param {boolean} [options.escapeHtml=false] - Whether to HTML-escape replaced values
      * @returns {string}
      */
-    #applyReplacements({text, member, siteSettings, escapeHtml = false}) {
-        const definitions = this.#buildReplacementDefinitions({member, siteSettings});
+    #applyReplacements({definitions, text, escapeHtml = false}) {
         let processed = wrapReplacementStrings(text);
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/services/member-welcome-emails/member-welcome-email-renderer.js
```diff
@@ -96,8 +94,17 @@ class MemberWelcomeEmailRenderer {
             });
         }
 
-        const contentWithReplacements = this.#applyReplacements({text: content, member, siteSettings, escapeHtml: true});
-        const subjectWithReplacements = this.#applyReplacements({text: subject, member, siteSettings, escapeHtml: false});
+        const definitions = this.#buildReplacementDefinitions({member, siteSettings});
```
troyciesco, line 97: by building the definitions here like this, we can define it in one place (`#buildReplacementDefinitions`) and then apply the regex to each. so if we add more we don't have to define them in two spots

we also then pass the definitions into `#applyReplacements` instead of building them twice

ghost/core/test/unit/server/services/member-welcome-emails/member-welcome-email-renderer.test.js
```diff
@@ -306,5 +306,53 @@ describe('MemberWelcomeEmailRenderer', function () {
             assert(!result.html.includes('}%%'));
             assert.equal(result.subject, 'Welcome !');
         });
+
+        it('removes code wrappers around replacement strings', async function () {
```
troyciesco, line 310: i think these two tests are covering the only ways that this would come into play:
- token replacement, wrapped in `<code>`, and we want to remove `<code>`
- actual code, for which we don't want to remove styling, even if includes a replacement like `{name}`

cmraible (approved):
Pushed [this](https://github.com/TryGhost/Ghost/pull/26304/changes/af73bd95582acb9896b4a4ea9b956a67e3727d07), otherwise LGTM pending CI

PR #26309, opened 2026-02-09
Fixed Ember theme response mutations leaking into React

kevinansfield:
ref https://linear.app/ghost/issue/BER-3275/

The `InstalledTheme` type used `gscan_errors` as the property name for theme validation errors, but the API returns `errors`.

The `gscan_errors` name came from Ember's serializer which remaps the property internally.
Additionally, Ember's `pushPayload` in the state bridge was mutating the response object in place (renaming `errors` via the serializer attrs), which meant React code holding a reference to the same object could not read the original properties. This is fixed by cloning the response before passing it to Ember's store.

---

The original code was working but relied on a side-effect of the Ember state bridge. With more of Ember being ripped out we would have hit problems with this sooner or later.

### DIFF
--- a/apps/admin-x-settings/src/components/settings/site/theme/theme-installed-modal.tsx
+++ b/apps/admin-x-settings/src/components/settings/site/theme/theme-installed-modal.tsx
@@ -24,7 +24,7 @@ export const ThemeProblemView = ({problem}:{problem: ThemeProblem}) => {
                             <div dangerouslySetInnerHTML={{__html: problem.details}} className='mb-4' />
                             <Heading level={6}>Affected files:</Heading>
                             <ul className='mt-1'>
-                                {problem.failures.map(failure => <li><code>{failure.ref}</code>{failure.message ? `: ${failure.message}` : ''}</li>)}
+                                {problem.failures.map(failure => <li key={failure.ref}><code>{failure.ref}</code>{failure.message ? `: ${failure.message}` : ''}</li>)}
                             </ul>
                         </div> :
                         null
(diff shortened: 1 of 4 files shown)


apps/admin-x-settings/src/components/settings/site/theme/theme-installed-modal.tsx
```diff
@@ -47,10 +47,10 @@ const ThemeInstalledModal: React.FC<{
     const handleError = useHandleError();
 
     let errorPrompt = null;
-    if (installedTheme && installedTheme.gscan_errors) {
+    if (installedTheme && installedTheme.errors) {
         errorPrompt = <div className="mt-6">
             <List hint={<>Highly recommended to fix, functionality <strong>could</strong> be restricted</>} title="Errors">
-                {installedTheme.gscan_errors?.map(error => <ThemeProblemView problem={error} />)}
+                {installedTheme.errors?.map((error, index) => <ThemeProblemView key={error.code || index} problem={error} />)}
```
kevinansfield: Fixed missing `key` prop here seeing as we're touching this file anyway.

rob-ghost: **Question:** Would using `error.code` be a potential problem if there's multiple errors with the same code?

ghost/admin/app/services/state-bridge.js
```diff
@@ -76,17 +76,23 @@ export default class StateBridgeService extends Service.extend(Evented) {
 
         const {type, singleton} = emberDataTypeMapping[dataType];
 
+        // Clone the response before pushing to the Ember store because
+        // pushPayload mutates the object in place (e.g. renaming keys via
+        // serializer attrs). Without cloning, React code holding a reference
+        // to the same object would see the mutated property names.
+        const clonedResponse = structuredClone(response);
```
rob-ghost, line 83: **Question:** Should `pushPayload` do this clone instead so it doesn't mutate its arguments?

This way we wouldn't have to remember to clone it ourselves if we're pushing in other places and not expecting it to mutate our argument.

rob-ghost (commented):
**Question:** How could we have discovered this issue sooner / automatically? (The incorrect API response type)

apps/admin-x-settings/src/components/settings/site/theme/theme-installed-modal.tsx
kevinansfield: We _shouldn't_ have duplicates because we group all affected files under a specific code but even so, using just `index` would be safer if that ever changes

ghost/admin/app/services/state-bridge.js
kevinansfield: `pushPayload` is an Ember Data core method, we could try overriding but its not clear what else within Ember may be relying on that mutation behaviour.

As discussed in call, we'll opt to keep the clone in this bridge method as it will go away eventually anyway.

rob-ghost (approved):

PR #26310, opened 2026-02-09
Improved (fixed?) flaky migration tools test timing out on CI

kevinansfield:
no issue

The "Built-in migrators" test was doing 4 full page.goto('/') reloads within a single 30s test timeout. Each reload requires API fetches and a full React render, which on CI runners easily exceeds the budget. Replaced with a single page load and page.goBack() between assertions.

### DIFF
--- a/apps/admin-x-settings/test/acceptance/advanced/migration-tools.test.ts
+++ b/apps/admin-x-settings/test/acceptance/advanced/migration-tools.test.ts
@@ -8,14 +8,16 @@ test.describe('Migration tools', async () => {
         }});
 
         const openMigrator = async (name: string, route: string) => {
-            await page.goto('/');
-
             const migrationSection = page.getByTestId('migrationtools');
             await expect(migrationSection).toBeVisible();
 
             await migrationSection.getByRole('button', {name}).click();
             await expectExternalNavigate(page, {route});
+
+            await page.goBack();
         };
+
+        await page.goto('/');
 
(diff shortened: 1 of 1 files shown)


kevinansfield:
@EvanHahn I noticed you recently added this multiple `page.goto()` behaviour, was that mirroring existing code or something new in an attempt to reduce flakiness?

EvanHahn:
I'm out this week, will look next.

rob-ghost (approved):
It's worth merging and seeing if this improves the flakiness of these migrator tests.

rob-ghost:
An alternative fix would be to parameterise the tests for each route and that way they each get their own 30s test timeout and there would only be one `goto` per-route.

kevinansfield:
> An alternative fix would be to parameterise the tests for each route and that way they each get their own 30s test timeout and there would only be one goto per-route.

Yeah, that's definitely an option if we still run into issues. I'd like to merge as-is, seems like we'd learn something more quickly if this is still flaky - pushing to individual 30s timeouts could more easily hide an underlying issue?

rob-ghost:
@kevinansfield 💯 merge as-is 👍🏻

kevinansfield:
Instant fail on the next run on `main` 😅 The playwright trace from CI was helpful, ended up with a better fix in a new PR https://github.com/TryGhost/Ghost/pull/26320

PR #26312, opened 2026-02-09
🐛 Fixed from address for gmail inbox links

9larsons:
ref https://linear.app/ghost/issue/NY-1012/

Inbox links used the support address (reply) instead of the specified from address. Gmail allows us to pass params to automatically filter to a message, and in this case we're filtering to the wrong one, making it look like the email never got sent.

### DIFF
--- a/ghost/core/core/server/services/email-address/email-address-service.ts
+++ b/ghost/core/core/server/services/email-address/email-address-service.ts
@@ -24,6 +24,7 @@ export class EmailAddressService {
     #getDefaultEmail: () => EmailAddress;
     #getFallbackDomain: () => string | null;
     #getFallbackEmail: () => EmailAddress | null;
+    #getMembersSupportAddress: () => string;
     #isValidEmailAddress: (email: string) => boolean;
 
     constructor(dependencies: {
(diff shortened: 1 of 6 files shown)


ghost/core/core/server/services/email-address/email-address-service.ts
```diff
@@ -68,6 +71,16 @@ export class EmailAddressService {
         return this.#getFallbackEmail();
     }
 
+    /**
```
9larsons, line 74: This is the meat of the PR. Almost all of the other changes are just passing the email service into the members service.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -733,7 +736,7 @@ module.exports = class RouterController {
 
             const inboxLinks = await getInboxLinks({
                 recipient: normalizedEmail,
-                sender: this._settingsHelpers.getMembersSupportAddress(),
+                sender: this._emailAddressService.getMembersSupportAddress(),
```
9larsons, line 739: And then instead of using the setting, we use the service to get this.

ghost/core/core/server/services/email-address/email-address-service.ts
```diff
@@ -68,6 +71,16 @@ export class EmailAddressService {
...
+     * Get the actual sender address for member emails after DMARC transformation.
+     * On managed platforms, this may differ from the configured support address to ensure DMARC compliance.
+     */
+    getMembersSupportAddress(): string {
+        const configuredAddress = this.#getMembersSupportAddress();
+        const transformed = this.getAddressFromString(configuredAddress);
+        return transformed.from.address;
+    }
```
cmraible, line 82: This function shouldn't be necessary - since we actually want the from address, not the members support address, we should be able to use the existing getter a few lines above this

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
cmraible, line 739:
```suggestion
                sender: this._emailAddressService.defaultFromEmail.address,
```

cmraible (changes requested):

ghost/core/core/server/services/email-address/email-address-service.ts
cmraible: Disregard, misunderstanding on my part.

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
cmraible: Disregard, misunderstanding on my part.

cmraible (approved):

PR #26320, opened 2026-02-10
Fixed flaky migration tools test by avoiding real page navigations

kevinansfield:
ref https://github.com/TryGhost/Ghost/pull/26310

The test's `externalNavigate` stub was setting `window.location.href` which
triggered real browser navigations to non-existent URLs (~3s each) followed
by `page.goBack()` to reload the React app (~3.5s each), repeated 4 times.
This easily exceeded the 30s timeout on CI.

The stub now stores the navigation link as a data attribute on `document.body`
instead of navigating, and the `expectExternalNavigate` helper asserts against
that attribute. This lets the test verify all 4 migrator buttons with a
single page load and zero navigations.

### DIFF
--- a/apps/admin-x-framework/src/test/acceptance.ts
+++ b/apps/admin-x-framework/src/test/acceptance.ts
@@ -358,5 +358,7 @@ export async function testUrlValidation(input: Locator, textToEnter: string, exp
 };
 
 export async function expectExternalNavigate(page: Page, link: Partial<ExternalLink>) {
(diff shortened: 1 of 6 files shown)


kevinansfield:
An alternative here would just be deleting this test, its usefulness is questionable in any case. This test is also the only consumer of `expectExternalNavigate` so we could potentially just rip all of that handling out.

rob-ghost (approved):

PR #26334, opened 2026-02-10
Port members list from ember to react

weylandswart:
No ref

Draft PR to port our Members list from Ember to React. Accounts for filtering updates made on the Comments page and new components used for the newer React pages.


apps/posts/src/views/members/components/members-actions.tsx
```diff
@@ -0,0 +1,229 @@
...
+    if (filter) {
+        params.set('filter', filter);
+    }
+    params.set('limit', 'all');
+
+    // Use the Ghost Admin API URL
+    const response = await fetch(`/ghost/api/admin/members/upload/?${params}`, {
+        method: 'GET',
+        credentials: 'include'
+    });
```
rob-ghost: **Issue:** We shouldn't be using a raw `fetch` and favour `useFetchApi` for consistency. If there's a reason to use raw `fetch` lets call it out with a comment.

apps/posts/src/views/members/components/members-actions.tsx
```diff
@@ -0,0 +1,229 @@
...
+    const [showDeleteModal, setShowDeleteModal] = useState(false);
+    const [isExporting, setIsExporting] = useState(false);
+
+    const handleExport = useCallback(async () => {
+        setIsExporting(true);
+        try {
+            await exportMembers(nql);
+        } catch (error) {
+            // eslint-disable-next-line no-console
+            console.error('Export failed:', error);
```
rob-ghost: **Issue:** This swallows the error without any feedback to the user. The ember version shows a toast on failure.

If we intend to handle failure differently in React, or if we're waiting on Toast porting, we should add a TODO here explaining why we're handling the error.

apps/posts/src/views/members/members.tsx
```diff
@@ -0,0 +1,162 @@
...
+    const canBulkDelete = useMemo(() => {
+        return !filters.some(f => BULK_DELETE_RESTRICTED_FILTERS.includes(f.field));
+    }, [filters]);
+
+    // Build search params for the API query
+    const searchParams: Record<string, string> = {
+        include: 'labels,tiers',
+        limit: '50',
+        order: 'created_at desc'
+    };
```
rob-ghost: **Suggestion:** This object is created on every render, not a big deal but I don't know if passing it to `useBrowseMembersInfinite` causes it to re-fetch un-necessarily.

We could use `useMemo`:
```typescript
const searchParams = useMemo(() => {
    const searchParams: Record<string, string> = {
        include: 'labels,tiers',
        limit: '50',
        order: 'created_at desc'
    };

    if (nql) {
        searchParams.filter = nql;
    }

    return searchParams;
}, [nql])
```

apps/posts/src/views/members/components/members-filters.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const tiers = tiersData?.tiers || [];
+    const newsletters = newslettersData?.newsletters || [];
+    const hasMultipleTiers = tiers.filter(t => t.type === 'paid' && t.active).length > 1;
+
+    // Get filter configuration
+    const filterFields = useMembersFilterConfig({
+        labels,
+        tiers: tiers.filter(t => t.type === 'paid' && t.active),
+        newsletters: newsletters.filter(n => n.status === 'active'),
+        offers: [], // TODO: Fetch offers if needed
```
rob-ghost: **Question:** Is the lack of offers being tracked anywhere? Linear issue?

rob-ghost (approved):
Some comments that could probably be addressed now with the help of an agent, otherwise we can merge and iterate.

apps/posts/src/views/members/components/members-list.tsx
```diff
@@ -0,0 +1,247 @@
...
+            </div>
+        </div>
+    );
+});
+
+function formatDate(dateString: string): string {
+    const date = new Date(dateString);
+    const options: Intl.DateTimeFormatOptions = {day: 'numeric', month: 'short', year: 'numeric'};
+    return date.toLocaleDateString('en-US', options);
+}
```
rob-ghost: **Suggestion:** Shade provides `formatDisplayDate()` and `formatTimestamp()` we should use for consistency.

apps/posts/src/views/members/hooks/use-members-filter-state.ts
```diff
@@ -0,0 +1,378 @@
...
+        is_any_of: '',
+        is_not_any_of: '-',
+        greater_than: '>',
+        less_than: '<',
+        equals: '',
+        not_equals: '-'
+    };
+
+    return relationMap[relation] ?? '';
+}
```
rob-ghost, line 76: **Suggestion(not-now):** These are duplicated from Ember, along with some other utilities (`escapeNqlString`, `getFilterRelationOperator`, `buildMemberNqlFilter`, the filter option constants)

We probably want to centralise these into `admin-x-framework` or something at some point.

jonatansberg (commented):
We don't need the changes on the Ember side. 

It's enough to mount the React version on a different path in the new admin shell. With the Ember changes as they are, we will still try to render some version of this in React when the labs flag is set rather than the old Ember view.

apps/posts/src/views/members/hooks/use-resource-search.ts
```diff
@@ -0,0 +1,90 @@
...
+ * - `post` type: fetches published posts and pages
+ * - `email` type: fetches published/sent posts that have a newsletter (i.e. emails)
+ */
+export function useResourceSearch(resourceType: ResourceType): UseResourceSearchReturn {
+    const [searchValue, setSearchValue] = useState('');
+
+    const isPostType = resourceType === 'post';
+
+    const postFilter = buildFilter(
+        isPostType ? 'status:published' : '(status:published,status:sent)+newsletter_id:-null',
```
weylandswart, line 34: Extract email filter to variable for better readability

PR #26341, opened 2026-02-10
Added migration for Transistor integration Portal customization

9larsons:
ref https://linear.app/ghost/issue/NY-996

This migration creates a new JSON object setting to store Transistor integration Portal > Account Details configuration (enabled, heading, description, button text, URL template). These settings will be available in Admin > Settings > Portal customization > Account details when the integration is configured & enabled.

The frontend changes will follow, both within Admin > Settings and Portal.


ghost/core/core/server/data/migrations/versions/6.19/2026-02-10-12-00-00-add-transistor-portal-settings.js
```diff
@@ -0,0 +1,15 @@
...
+    key: 'transistor_portal_settings',
+    value: JSON.stringify({
+        enabled: true,
+        heading: 'Podcasts',
+        description: 'Access your RSS feeds',
+        button_text: 'Manage',
+        url_template: 'https://partner.transistor.fm/ghost/{memberUuid}'
+    }),
+    type: 'object',
+    group: 'transistor',
```
9larsons: transistor.transistor is the main on/off. This should arguably have been enabled and we could do another migration to clean this up. 

For this PR, I think grouping of `transistor` makes sense because they are transistor integration-specific, not portal-specific.

ghost/core/test/e2e-api/content/__snapshots__/settings.test.js.snap
```diff
@@ -84,6 +84,7 @@ Object {
     "support_email_address": "noreply@127.0.0.1",
     "timezone": "Etc/UTC",
     "title": "Ghost",
+    "transistor_portal_settings": null,
```
9larsons: I am trying to figure out why this is showing as null in the snap.

9larsons:
ref https://linear.app/ghost/issue/NY-1025/. These settings tests cause nothing but pain.

ghost/core/test/e2e-api/admin/__snapshots__/settings.test.js.snap
```diff
@@ -328,6 +328,14 @@ Object {
       "key": "explore_ping_growth",
       "value": false,
     },
+    Object {
+      "key": "transistor",
+      "value": false,
+    },
```
9larsons, line 334: We add the transistor key here because the fixtures were updated. I didn't notice this was missing in the last update. (I also didn't know our snapshot tests test fixtures instead of defaults, which is nonsense.)

ghost/core/test/e2e-api/admin/settings.test.js
```diff
@@ -10,7 +10,7 @@ const membersService = require('../../../core/server/services/members');
 const {anyErrorId} = matchers;
 
 // Updated to reflect current total based on test output
-const CURRENT_SETTINGS_COUNT = 93;
+const CURRENT_SETTINGS_COUNT = 95;
```
9larsons: Because transistor and transistor_portal_settings got added to the fixtures, that's why we're incrementing by two.

ghost/core/core/server/data/migrations/versions/6.19/2026-02-10-12-00-00-add-transistor-portal-settings.js
```diff
@@ -0,0 +1,15 @@
+const {addSetting} = require('../../utils');
```
troyciesco: heads up the version for this file should be higher now

troyciesco (changes requested):
still reviewing but wanted to drop the comment about the version first so we don't forget! since 6.18 went out yesterday

ghost/core/core/server/data/migrations/versions/6.19/2026-02-10-12-00-00-add-transistor-portal-settings.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = addSetting({
+    key: 'transistor_portal_settings',
+    value: JSON.stringify({
+        enabled: true,
```
troyciesco: question: should this default to false? that's what it seems like at first glance, but totally possible i'm misunderstanding what this is controlling and its relationship to the main on/off you referenced in the other comment

9larsons: This just controls whether we display the transistor fields in Portal. It's a fair question - my impression is we want to enable all functionality by default and provide a toggle to disable. Zimo has some mockups for providing the toggle in the integration pane itself so it's very obvious what is enabled. On by default makes sense to me.

ghost/core/core/server/data/migrations/versions/6.19/2026-02-10-12-00-00-add-transistor-portal-settings.js
9larsons: Very good shout, thanks. Updated.

troyciesco (approved):

cmraible (changes requested):
I think we should avoid using a JSON field for the settings. 

This buried real deeply in [an old Notion doc spec](https://www.notion.so/ghost/Settings-Table-8fed3136c32b4549bd2cbcf0bb9266e8?source=copy_link#7288deaea2774384a7da8189edd4cd78), which I happened to come across a while ago and remembered it only when looking at the second PR in this stack.

I think it's a valid point though - by keeping the settings in a free form JSON object, it makes parsing and validating and typing that JSON object much more involved than if each value were in its own row in the settings table.

Let's chat about this tomorrow? I don't mean to block if you feel strongly about the JSON object, but requesting changes for now until we have a chance to chat.

9larsons:
Great find. I'll work on those changes. I would almost want a subgroup here but that is what it is. The fact that we have 'object' as a `type` in the schema is annoying, and we only use it for labs. Feels like we could re-work that and strip it out at some point to make it clear we shouldn't use it.

I had json object in mind already as we were talking through what this would look like for plugins and trying to not require migrations for everything (though I think official integrations should have them). I'll have to continue thinking on that, or maybe that's another area we carve out. Or we don't end up using `settings` for it at all.

ghost/core/core/server/data/migrations/versions/6.19/2026-02-10-12-00-00-add-transistor-portal-settings.js
```diff
@@ -0,0 +1,108 @@
...
+            await knex('settings').insert({
+                id: (new ObjectID()).toHexString(),
+                key: 'transistor_portal_enabled',
+                value: 'true',
+                type: 'boolean',
+                group: 'transistor',
+                flags: 'PUBLIC',
+                created_at: knex.raw('current_timestamp')
+            });
+        }
```
cmraible: Was there a reason not to use the `addSetting` helper for these?

ghost/core/core/server/api/endpoints/utils/serializers/input/settings.js
```diff
@@ -80,7 +80,12 @@ const EDITABLE_SETTINGS = [
     'explore_ping',
     'explore_ping_growth',
     'indexnow_api_key',
-    'transistor'
+    'transistor',
+    'transistor_portal_enabled',
```
cmraible, line 84: question: what's the difference between these two? Would there be a scenario where transistor is enabled on the site, but not in portal?

9larsons: Troy had the same question - transistor is the integration, while portal is the portal settings, which is a subset of the integration functionality.

ghost/core/core/server/data/migrations/versions/6.19/2026-02-10-12-00-00-add-transistor-portal-settings.js
9larsons: Nope - resolved.

cmraible (approved):

PR #26343, opened 2026-02-10
Added Settings > Portal customization options for Transistor integration

9larsons:
ref https://linear.app/ghost/issue/NY-996
ref https://github.com/TryGhost/Ghost/pull/26341
- wired up transistor integration portal settings in portal customizations
- updated Account details tab behavior in Settings > Portal customizations to navigate the display to Account details as well on the left hand side

This wires up the Transistor.fm integration Portal settings to Admin > Settings > Portal customizations > Account details. We should be able to load in and adjust the settings.


apps/admin-x-settings/src/components/settings/membership/portal/account-page.tsx
```diff
@@ -1,21 +1,51 @@
...
 
+interface TransistorPortalSettings {
+    enabled: boolean;
+    heading: string;
+    description: string;
+    button_text: string;
+    url_template: string;
+}
+
+const defaultTransistorSettings: TransistorPortalSettings = {
```
9larsons: I don't love defaults being here as well as in the migration and the settings-defaults.json. One of the problems is that we don't have a default value in the key-value pair returned by the settings API. We'd need to update the serializer to support that and do a lot more to establish that pattern. This seems like the best option for now, and it should be largely untouched.

Alternatively we could set this in the component, which might be more functional.

apps/admin-x-settings/src/components/settings/membership/portal/transistor-settings.tsx
```diff
@@ -0,0 +1,78 @@
...
+    url_template: string;
+}
+
+const defaultTransistorSettings: TransistorPortalSettings = {
+    enabled: true,
+    heading: 'Podcasts',
+    description: 'Access your RSS feeds',
+    button_text: 'Manage',
+    url_template: 'https://partner.transistor.fm/ghost/{memberUuid}'
+};
```
9larsons: This is annoying to have here but we don't support default values in the settings API serializer, nor do we have any sort of default pattern. I don't feel strongly about implementing one for this just to clean it up a touch.

apps/admin-x-settings/src/components/settings/membership/portal/portal-modal.tsx
```diff
@@ -21,9 +21,9 @@ const Sidebar: React.FC<{
     updateTier: (tier: Tier) => void
     errors: Record<string, string | undefined>
     setError: (key: string, error: string | undefined) => void
-}> = ({localSettings, updateSetting, localTiers, updateTier, errors, setError}) => {
-    const [selectedTab, setSelectedTab] = useState('signupOptions');
-
+    selectedTab: string
```
9larsons, line 24: This is a simple ux improvement slipped into this: when clicking signup or account on either the left or right set of tabs, we auto-switch the other to match to prevent dragging the mouse across the screen + extra clicks.

apps/admin-x-settings/src/components/settings/membership/portal/portal-modal.tsx
```diff
@@ -189,15 +186,33 @@ const PortalModal: React.FC = () => {
 
     const onSelectURL = (id: string) => {
         setSelectedPreviewTab(id);
+        // Sync sidebar tab with preview tab
```
9larsons, line 189: This is the logic for the tab switching.

apps/admin-x-settings/src/components/settings/membership/portal/account-page.tsx
```diff
@@ -1,16 +1,20 @@
 import React, {type FocusEventHandler, useEffect, useState} from 'react';
```
9larsons, line 1: Adds the transistor block to Portal > Account details.

apps/admin-x-settings/src/utils/get-portal-preview-url.ts
```diff
@@ -60,5 +60,10 @@ export const getPortalPreviewUrl = ({settings, config, tiers, siteData, selected
 
     settingsParam.append('disableBackground', 'false');
 
+    const transistorPortalSettings = getSettingValue<string>(settings, 'transistor_portal_settings');
+    if (transistorPortalSettings) {
+        settingsParam.append('transistorPortalSettings', transistorPortalSettings);
+    }
+
```
9larsons, line 73: This allows us to pass in settings with the Portal preview pane in Settings to display changes 'live'.

apps/portal/src/components/pages/AccountHomePage/components/transistor-podcasts-action.js
```diff
@@ -13,29 +13,48 @@ export const TransistorPodcastsActionStyles = `
...
 
-    const transistorUrl = `https://partner.transistor.fm/ghost/${memberUuid}`;
+    const hasCustomHeading = settings.heading && settings.heading !== DEFAULT_HEADING;
+    const hasCustomDescription = settings.description && settings.description !== DEFAULT_DESCRIPTION;
+    const hasCustomButtonText = settings.button_text && settings.button_text !== DEFAULT_BUTTON_TEXT;
+
+    const heading = hasCustomHeading ? settings.heading : t(DEFAULT_HEADING);
+    const description = hasCustomDescription ? settings.description : t(DEFAULT_DESCRIPTION);
+    const buttonText = hasCustomButtonText ? settings.button_text : t(DEFAULT_BUTTON_TEXT);
+    const urlTemplate = settings.url_template || DEFAULT_URL_TEMPLATE;
```
9larsons: We have to do this convoluted value checking in order to preserve i18n where possible.

apps/portal/test/unit/components/pages/AccountHomePage/transistor-podcasts-action.test.js
```diff
@@ -17,29 +17,29 @@ describe('TransistorPodcastsAction', () => {
     test('renders nothing when hasPodcasts is false', () => {
         const {queryByText} = setup({hasPodcasts: false, memberUuid: 'test-uuid'});
         expect(queryByText('Podcasts')).not.toBeInTheDocument();
-        expect(queryByText('Subscribe')).not.toBeInTheDocument();
+        expect(queryByText('Manage')).not.toBeInTheDocument();
```
9larsons: This changed in the migration from the previous placeholder value.

apps/portal/src/components/pages/AccountHomePage/components/transistor-podcasts-action.js
```diff
@@ -13,29 +13,53 @@ export const TransistorPodcastsActionStyles = `
...
+
+    .gh-portal-action-transistor .gh-portal-list-detail p {
+        word-break: break-word;
+    }
 `;
 
-const TransistorPodcastsAction = ({hasPodcasts, memberUuid}) => {
+// Single source of truth for Transistor defaults - must match default-settings.json
+// Translatable strings use t() when displaying defaults, custom values display as-is
+export const TRANSISTOR_DEFAULTS = {
```
9larsons, line 22: Having even more defaults is awful. I'm trying to think of solutions outside of defaults from the settings API... which may be required.

apps/admin-x-settings/src/components/settings/membership/portal/transistor-settings.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const TransistorSettings: React.FC<{
+    localSettings: Setting[]
+    updateSetting: (key: string, setting: SettingValue) => void
+}> = ({localSettings, updateSetting}) => {
+    const [transistorPortalSettingsJson] = getSettingValues<string>(localSettings, ['transistor_portal_settings']);
+    const transistorSettings: TransistorPortalSettings = transistorPortalSettingsJson
+        ? {...DEFAULT_SETTINGS, ...JSON.parse(transistorPortalSettingsJson) as Partial<TransistorPortalSettings>}
```
cmraible: Will transistorPortalSettingsJson always be valid JSON? The unguarded JSON.parse could break rendering if not

apps/admin-x-settings/src/components/settings/membership/portal/portal-modal.tsx
cmraible: Nice touch!

apps/portal/src/components/pages/AccountHomePage/components/transistor-podcasts-action.js
cmraible: For welcome emails, we do something sort of similar: the defaults live in the settings frontend app, but as soon as you enable a welcome email the first time, we POST it to the backend and it's persisted from that point forward.

Maybe the settings app could POST (or PUT technically I think) the default values and persist in the DB when Transistor is first enabled?

9larsons: It doesn't work as cleanly here. We enable the integration from the integration page, not the Portal settings, so we'd be setting the defaults elsewhere vs. where the user is interacting with them as we don't have a 'setup flow' here like Welcome emails.

cmraible: Gotcha, makes sense!

9larsons:
In order to test this, you can spoof the `hasPodcasts: true` in `use-integrations.ts`.
```
    return {
        transistor: {
            enabled: isTransistorEnabled,
            hasPodcasts: transistorPodcasts, // set to true
            memberUuid,
            settings: transistorSettings
        }
    };
```

apps/admin-x-settings/src/components/settings/membership/portal/transistor-settings.tsx
```diff
@@ -0,0 +1,84 @@
...
+    ]);
+
+    // Don't show the section if the main Transistor integration is disabled
+    const integrationEnabled = transistorIntegrationEnabled === true;
+    if (!integrationEnabled) {
+        return null;
+    }
+
+    const enabled = transistorPortalEnabled === true;
+    const heading = transistorPortalHeading as string ?? '';
```
9larsons: I'm not sure we need the ?? '' here.

ghost/core/core/shared/settings-cache/cache-manager.js
```diff
@@ -234,6 +234,14 @@ class CacheManager {
...
         }
 
+        // Compute transistor_portal_enabled: only true if main integration AND portal setting are both enabled
+        //  - this is done server-side because we shouldn't expose the transistor integration setting
+        if ('transistor_portal_enabled' in settings) {
+            const transistorEnabled = this._doGet('transistor');
+            const portalEnabled = settings.transistor_portal_enabled;
+            settings.transistor_portal_enabled = Boolean(transistorEnabled) && Boolean(portalEnabled);
+        }
+
```
9larsons, line 244: I added this because the transistor (integration) setting is not public, nor does it feel it should be. Unfortunately this adds a layer of logic here that would be nice to avoid. I don't have a better idea atm. Maybe with fresh eyes next week.

I suppose one way to do this would be to set the Portal setting to false if the integration is ever disabled, and true if ever enabled, that way we don't have to do double-duty.

The case we're trying to cover here is ensuring the frontend Portal > Account details doesn't try to display the Podcast segment when `transistor_portal_enabled` = true and `transistor` = false.

9larsons:
Ah right - Portal CI task will fail until it's shipped as part of this. I'll include that after we're ok with the changes.

apps/portal/src/components/pages/AccountHomePage/components/transistor-podcasts-action.js
```diff
@@ -13,29 +13,50 @@ export const TransistorPodcastsActionStyles = `
...
+        word-break: break-word;
+    }
 `;
 
-const TransistorPodcastsAction = ({hasPodcasts, memberUuid}) => {
+export const TRANSISTOR_DEFAULTS = {
+    heading: 'Podcasts',
+    description: 'Access your RSS feeds',
+    button_text: 'Manage',
+    url_template: 'https://partner.transistor.fm/ghost/{memberUuid}'
```
cmraible, line 26: Should these match the defaults created in the migration? I think otherwise the `maybeTranslate` function won't internationalize the default strings since they won't match.

cmraible (approved):
I think we should update the defaults in the frontend to match the defaults that are set in the migration; other than that this looks good to me!

PR #26347, opened 2026-02-11
Removed labeled and unlabeled triggers from CI workflow

cmraible:
refs https://ghost.slack.com/archives/C02G9E68C/p1770769239070299

## Problem

Currently, adding or removing any label to a PR triggers the CI workflow to run again. For _most_ labels, this isn't desired and results in CI _failing every time_ just because you added an innocent label.

For example, if you have a PR with all checks passing, if you add the "ok to merge for me" label, CI will re-run and the "Merge Reports" job will fail every time. In order to actually merge the PR, you'd need to re-run all the checks again, which is a waste of time.

## Context

Labels that include "test" or "deploy" can trigger additional jobs to run in CI. This is the historical reason that we added the `labeled`/`unlabeled` trigger on PRs for the CI workflow in the first place. These labels include:
- `browser-tests` - to run the browser tests
- `test` - simply used to trigger a full CI re-run
- `perf-tests` - to run the performance tests (Ghost boot time) to run
- `deploy-to-staging` - to trigger a staging build based on the PR branch

Applying any of these labels would trigger all of CI to run again, plus whatever additional jobs they are intended to trigger.

## Fix

This fix removes the `labeled` and `unlabeled` events as triggers for CI entirely. This fixes the "adding a label causes CI to fail on my PR" problem. There are only a few valid use-cases for these triggers, and the costs now outweigh the benefits.

With this change, applying or removing a label from a PR shouldn't trigger CI to run at all. The `label-actions.yml` workflow will still run — this is fine since it's a completely separate workflow and doesn't impact CI checks.

## Breaking changes to CI

With this change, the following labels will no longer trigger any action in the CI workflow simply by adding/removing the label:
- `browser-tests` - this label is no longer necessary
- `deploy-to-staging` - 0/300 open PRs have this label currently
- `test` - 0/300 open PRs have this label
- `perf-tests` - 0/300 open PRs have this label

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 on:
   pull_request:
-    types: [opened, synchronize, reopened, labeled, unlabeled]
+    types: [opened, synchronize, reopened]
   push:
     # Ref: GHA Filter pattern syntax: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet
     # Run on pushes to main, release branches, and previous/future major version branches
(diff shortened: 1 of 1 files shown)


cmraible:
Discussion in Slack for context: https://ghost.slack.com/archives/C09K9BSGYQM/p1770783947904959

PR #26348, opened 2026-02-11
Added srcset support for CDN images

vershwal:
ref [HKG-1581](https://linear.app/ghost/issue/HKG-1581)

- Ghost generates responsive image `srcset` for local `/content/images/` URLs, but CDN/Object Storage URLs were not recognized as content images. As sites move image delivery to CDN, those images should also have `srcset` for responsive loading.
- This PR supports CDN-hosted Ghost images while keeping existing local-image and Unsplash behavior unchanged.
- The updated renderer paths are safe for CDN URLs because they only do URL/HTML string transformations (e.g. inserting `/size/w{width}/`), not file reads.
- If `urls:image` is not configured, behavior falls back to local-content-image detection.
- Existing local-image and Unsplash flows continue to work.

### DIFF
--- a/ghost/core/core/server/services/koenig/render-utils/is-content-image.js
+++ b/ghost/core/core/server/services/koenig/render-utils/is-content-image.js
@@ -1,0 +1,22 @@
+const isLocalContentImage = function (url, siteUrl = '') {
+    const normalizedSiteUrl = siteUrl.replace(/\/$/, '');
+    const imagePath = url.replace(normalizedSiteUrl, '');
+    return /^(\/.*|__GHOST_URL__)\/?content\/images\//.test(imagePath);
+};
+
+const isContentImage = function (url, siteUrl = '', imageBaseUrl = '') {
+    if (isLocalContentImage(url, siteUrl)) {
+        return true;
+    }
+    if (imageBaseUrl) {
+        const normalizedBaseUrl = imageBaseUrl.replace(/\/$/, '');
+        const imagePath = url.replace(normalizedBaseUrl, '');
+        return /^\/?content\/images\//.test(imagePath);
+    }
+    return false;
+};
+
+module.exports = {
+    isLocalContentImage,
+    isContentImage
+};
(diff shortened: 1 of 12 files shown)


ghost/core/test/unit/server/services/koenig/render-utils/is-content-image.test.js
```diff
@@ -0,0 +1,113 @@
+const assert = require('assert/strict');
+const {isLocalContentImage, isContentImage} = require('../../../../../../core/server/services/koenig/render-utils/is-content-image');
+
+describe('services/koenig/render-utils/is-local-content-image', function () {
```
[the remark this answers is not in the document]
vershwal: Ya that's right, will change before merging.

ghost/core/core/server/services/koenig/render-utils/is-content-image.js
```diff
@@ -0,0 +1,22 @@
...
+};
+
+const isContentImage = function (url, siteUrl = '', imageBaseUrl = '') {
+    if (isLocalContentImage(url, siteUrl)) {
+        return true;
+    }
+    if (imageBaseUrl) {
+        const normalizedBaseUrl = imageBaseUrl.replace(/\/$/, '');
+        const imagePath = url.replace(normalizedBaseUrl, '');
+        return /^\/?content\/images\//.test(imagePath);
```
allouis: This is _almost_ the same as the regex above, if remove the `__GHOST_URL__` from the above regex, then we can treat all checks the same and compose them

```ts
const matchesContentImagePath = function (url, baseUrl) {
    const normalized = baseUrl.replace(/\/$/, '');
    const path = url.replace(normalized, '');
    return /^\/?content\/images\//.test(path);
};

const isLocalContentImage = function (url, siteUrl = '') {
    return matchesContentImagePath(url, siteUrl) || matchesContentImagePath(url, '__GHOST_URL__');
};

const isContentImage = function (url, siteUrl = '', imageBaseUrl = '') {
    return isLocalContentImage(url, siteUrl) || (imageBaseUrl && matchesContentImagePath(url, imageBaseUrl));
};
```

Stylistic, NABD

allouis (approved):

ghost/core/core/server/services/koenig/render-utils/is-content-image.js
vershwal: Thank you. I like this. Updated 👍

PR #26349, opened 2026-02-11
Ensured retention offers are not shown post cancellation

mike182uk:
ref https://linear.app/ghost/issue/BER-3297

With the custom cancellation button, it was possible that after a cancellation had occurred that the button would still trigger the retention offers to be shown. Seeming though the user has already cancelled, there is no need to show the offers. This prevents an issue where the user can take up an offer, but the cancellation period does not get reset.

### DIFF
--- a/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
+++ b/ghost/core/core/server/services/members/members-api/controllers/router-controller.js
@@ -980,6 +980,10 @@ module.exports = class RouterController {
             return res.end(JSON.stringify({offers}));
         }
 
+        function sendNoOffersAvailable() {
+            return sendOffersResponse([]);
+        }
+
         if (!identity) {
             res.writeHead(401);
             return res.end('Unauthorized');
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/services/members/members-api/controllers/router-controller.js
```diff
@@ -1039,6 +1039,11 @@ module.exports = class RouterController {
 
         const activeSubscription = activeSubscriptions[0];
 
+        // If subscription is already set to cancel, don't show retention offers
+        if (activeSubscription.get('cancel_at_period_end')) {
+            return sendOffersResponse();
```
sagzy: nitpick: this is not a new addition in this PR, but noticing that I had to scroll up to read the content of `sendOffersResponse` to understand what we were returning, i.e. no offers available.

So I'm thinking, we could either make `[]` explicit when calling 
`sendOffersResponse([])` 

Or, take it a step further and add an additional helper to increase readability:  
```
function sendNoOffersAvailable() {
    return sendOffersResponse([]);
}
```

sagzy (commented):
Looks good, added one nitpick about increasing readability (not directly related to the change in this commit)

Also: should we apply the same change in `applyOfferToSubscription`?

ghost/core/core/server/services/members/members-api/controllers/router-controller.js
mike182uk: Nice, this is a great suggestion - Have updated in https://github.com/TryGhost/Ghost/pull/26349/commits/7c1eb36a5811a76ac8fe858f97511666bcc585a2

mike182uk:
> Also: should we apply the same change in applyOfferToSubscription?

Yep, good catch. Have added in https://github.com/TryGhost/Ghost/pull/26349/commits/693b7fbfbb3b6a17e6ee18c807aa759568ff2eb0

sagzy (approved):

PR #26352, opened 2026-02-11
Improved the domain warming tests to run faster

sam-lord:
no issue

This should prevent the flakiness we've seen where the domain tests take too long to run. Instead of using the ORM layer, we're bulk inserting / deleting using the knex layer. I wouldn't want to do this in the source, but in tests this feels like a suitable optimisation to check bulk cases quickly.


aileen (approved):
Very nice!

PR #26358, opened 2026-02-11
🐛 Fixed welcome emails using noreply address instead of default newsletter reply-to address

cmraible:
ref https://linear.app/ghost/issue/NY-1028/welcome-email-uses-noreply-instead-of-configured-sender

## Problem

Welcome emails aren't currently respecting the sender info of the default newsletter as intended. Instead, they are being sent with a fallback `sender_name` and `sender_reply_to` address:
- For a site with custom sending domain, this is `noreply@example.com`
- For a site without a custom sending domain, this is `subdomain@ghost.io`

## Fix

Welcome emails should use the sender info from the default newsletter's settings, so that new members can reply directly to the welcome email and the publisher will receive that email. This fix gets the settings from the default newsletter, and passes this through to mailgun when sending welcome emails (real ones and test emails).

### DIFF
--- a/ghost/core/core/server/services/member-welcome-emails/service.js
+++ b/ghost/core/core/server/services/member-welcome-emails/service.js
@@ -3,9 +3,11 @@ const errors = require('@tryghost/errors');
 const urlUtils = require('../../../shared/url-utils');
 const settingsCache = require('../../../shared/settings-cache');
 const emailAddressService = require('../email-address');
+const settingsHelpers = require('../settings-helpers');
+const EmailAddressParser = require('../email-address/email-address-parser');
 const mail = require('../mail');
 // @ts-expect-error type checker has trouble with the dynamic exporting in models
-const {AutomatedEmail} = require('../../models');
+const {AutomatedEmail, Newsletter} = require('../../models');
 const MemberWelcomeEmailRenderer = require('./member-welcome-email-renderer');
 const {MEMBER_WELCOME_EMAIL_LOG_KEY, MEMBER_WELCOME_EMAIL_SLUGS, MESSAGES} = require('./constants');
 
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/services/member-welcome-emails/service.js
```diff
@@ -28,6 +30,59 @@ class MemberWelcomeEmailService {
...
+                    address: fromAddress,
+                    name: senderName || undefined
+                },
+                replyTo: replyToSetting === 'newsletter' ? undefined : {address: replyToSetting}
+            });
+
+            if (addresses.replyTo) {
+                replyTo = EmailAddressParser.stringify(addresses.replyTo);
+            }
+        }
```
cmraible, line 77: This logic is largely duplicated from the email [renderer](https://github.com/TryGhost/Ghost/blob/ad0cbc88fd9be2eaf70bb5b806dbafcfa915a455/ghost/core/core/server/services/email-service/email-renderer.js#L261-L287). We should refactor this out into a helper after we fix this, potentially in the emailAddress service?

troyciesco (approved):

ghost/core/core/server/services/member-welcome-emails/service.js
```diff
@@ -83,7 +84,18 @@ class MemberWelcomeEmailService {
...
+        if (this.#defaultNewsletterSenderOptions) {
+            return this.#defaultNewsletterSenderOptions;
+        }
+
+        this.#defaultNewsletterSenderOptions = await this.#getDefaultNewsletterSenderOptions();
+        return this.#defaultNewsletterSenderOptions;
+    }
+
     async loadMemberWelcomeEmails() {
+        this.#defaultNewsletterSenderOptions = await this.#getDefaultNewsletterSenderOptions();
```
cmraible, line 97: always fetch the newsletter sending options when running loadMemberWelcomeEmails. Cache the value in #defaultNewsletterSenderOptions

ghost/core/core/server/services/member-welcome-emails/service.js
```diff
@@ -137,7 +149,7 @@ class MemberWelcomeEmailService {
             siteSettings: this.#getSiteSettings()
         });
 
-        const senderOptions = await this.#getDefaultNewsletterSenderOptions();
+        const senderOptions = await this.#getSenderOptions();
```
cmraible, line 152: This should always read from #defaultNewsletterSenderOptions, but will fallback to the DB query if somehow it isn't set

cmraible:
Tested in both staging and production and working as expected, both with/without custom sending domains.

PR #26361, opened 2026-02-11
Added api top pages routing pipe

9larsons:
ref https://linear.app/ghost/issue/NY-956

This pipe detects if filters are used and if not, uses the optimized view for api_top_pages return data (used by top_content endpoint). This has dramatically sped up the unfiltered view and provides an easy way to switch live a new endpoint without having to change versions in Ghost.


ghost/core/core/server/data/tinybird/endpoints/api_top_pages_router.pipe
```diff
@@ -0,0 +1,28 @@
...
+
+NODE router
+DESCRIPTION >
+    Routes to the appropriate implementation based on filter parameters.
+    Session-level filters (device, location, source, utm_*) require the original pipe.
+    Simple queries use the faster v3 materialized view.
+
+SQL >
+    %
+    {% if defined(device) or defined(location) or defined(source) or defined(utm_source) or defined(utm_medium) or defined(utm_campaign) or defined(utm_term) or defined(utm_content) %}
```
9larsons, line 22: This is very easy to iterate on if we want to have views with the most common filters. Additionally, it makes it very easy to deploy a change to this router pipe instead of Ghost in order to deploy changes, assuming we aren't changing the return shape.

ghost/core/core/server/data/tinybird/tests/api_top_pages_router.yaml
```diff
@@ -0,0 +1,271 @@
+
+- name: basic_route_to_v3
```
9larsons: Arguably this is an excessive amount of tests. I think all we need are some smoke tests that data returns with and without a filter provided. 🤷

ghost/core/core/server/data/tinybird/tests/api_top_pages_router.yaml
```diff
@@ -0,0 +1,271 @@
...
+    Please provide a value or set a default value in the pipe code.'
+  expected_http_status: 400
+
+- name: default_date_range
+  description: Test query without explicit date range (routes to api_top_pages_v3)
+  parameters: site_uuid=mock_site_uuid&timezone=Etc/UTC
+  expected_result: 'Template Syntax Error: Required parameter is not defined. Check
+    the parameters date_from, date_to, member_status, pathname, post_uuid and post_type.
+    Please provide a value or set a default value in the pipe code.'
+  expected_http_status: 400
```
cmraible: suggestion (blocking): we probably want to handle these more gracefully - either by making these parameters not required in `api_top_pages_v3` (if possible, they may truly be required 🤷 ) or by including them in the routers conditional so we don't accidentally return a 400

ghost/core/core/server/data/tinybird/tests/api_top_pages_router.yaml
```diff
@@ -0,0 +1,271 @@
...
+- name: all_non_session_filters
+  description: Test query with all non-session filters (routes to api_top_pages_v3)
+  parameters: site_uuid=mock_site_uuid&date_from=2100-01-01&date_to=2100-01-07&timezone=Etc/UTC&member_status=paid&pathname=%2Fabout%2F&post_uuid=06b1b0c9-fb53-4a15-a060-3db3fde7b1fc&post_type=page&skip=0&limit=10
+  expected_result: |
+    {"post_uuid":"06b1b0c9-fb53-4a15-a060-3db3fde7b1fc","pathname":"\/about\/","visits":3}
+
+- name: all_session_filters
+  description: Test query with all possible session filters (routes to api_top_pages)
+  parameters: site_uuid=mock_site_uuid&date_from=2100-01-01&date_to=2100-01-07&timezone=Etc/UTC&device=desktop&location=GB&source=bing.com&utm_source=google&utm_medium=social&utm_campaign=brand_awareness&utm_term=discount&utm_content=post_123
+  expected_result: ''
```
cmraible: Is this expected?

ghost/core/core/server/data/tinybird/tests/api_top_pages_router.yaml
```diff
@@ -0,0 +1,271 @@
...
+
+- name: high_skip_value_v3
+  description: Test high skip value without session filters (routes to api_top_pages_v3)
+  parameters: site_uuid=mock_site_uuid&date_from=2100-01-01&date_to=2100-01-07&timezone=Etc/UTC&skip=100&limit=10
+  expected_result: ''
+
+- name: high_skip_value_with_source
+  description: Test high skip value with source filter (routes to api_top_pages)
+  parameters: site_uuid=mock_site_uuid&date_from=2100-01-01&date_to=2100-01-07&timezone=Etc/UTC&source=google.com&skip=100&limit=10
+  expected_result: ''
```
cmraible: Are these all expected to be empty?

ghost/core/core/server/data/tinybird/tests/api_top_pages_router.yaml
```diff
@@ -0,0 +1,271 @@
...
+- name: combined_source_and_post_type
+  description: Test query combining source (session) and post_type (non-session) filters
+  parameters: site_uuid=mock_site_uuid&date_from=2100-01-01&date_to=2100-01-07&timezone=Etc/UTC&source=bing.com&post_type=post
+  expected_result: |
+    {"post_uuid":"6b8635fb-292f-4422-9fe4-d76cfab2ba31","pathname":"\/blog\/hello-world\/","visits":1}
+
+- name: combined_utm_and_non_session
+  description: Test query combining UTM (session) and non-session filters
+  parameters: site_uuid=mock_site_uuid&date_from=2100-01-01&date_to=2100-01-07&timezone=Etc/UTC&utm_source=google&member_status=undefined&pathname=%2Fabout%2F
+  expected_result: ''
```
cmraible, line 58: Is this expected? Not sure how much value the tests with empty expected results have

ghost/core/core/server/data/tinybird/tests/api_top_pages_router.yaml
```diff
@@ -0,0 +1,271 @@
...
+  description: Test query with multiple session-level filters
+  parameters: site_uuid=mock_site_uuid&date_from=2100-01-01&date_to=2100-01-07&timezone=Etc/UTC&device=desktop&location=GB&source=bing.com
+  expected_result: |
+    {"post_uuid":"06b1b0c9-fb53-4a15-a060-3db3fde7b1fc","pathname":"\/about\/","visits":1}
+    {"post_uuid":"","pathname":"\/","visits":1}
+
+- name: device_with_utm_filters
+  description: Test query combining device and UTM filters
+  parameters: site_uuid=mock_site_uuid&date_from=2100-01-01&date_to=2100-01-07&timezone=Etc/UTC&device=desktop&utm_source=google&utm_medium=social
+  expected_result: ''
```
cmraible: Is this expected to be empty?

ghost/core/core/server/data/tinybird/endpoints/api_top_pages_router.pipe
cmraible: This is really neat! I think this makes a lot of sense conceptually, as long as we're careful about breaking changes as you mentioned.

ghost/core/core/server/data/tinybird/tests/api_top_pages_router.yaml
cmraible: Yeah agree - I also can't think of a good way to _truly_ test that the results are coming from the correct pipe. For any request that should be routed to api_top_pages_v3, I'd expect api_top_pages to also return the same exact data, so we can't _really_ know for sure. 

But the ones with filters at least should be sufficient, since as long as the results are correct, we know for sure they aren't coming from api_top_pages_v3.

I think I'd be for reducing the number of tests - especially since they're a huge pain to manually validate anyhow. If we can get the same confidence from 10 tests that we get out of these 38, I think that'd be a win.

cmraible (changes requested):
Cool idea! [This one](https://github.com/TryGhost/Ghost/pull/26361/changes#r2796063003) is the only truly blocking comment, but the other tests, particularly the ones with empty expected results could either be adjusted or removed.

ghost/core/core/server/data/tinybird/tests/api_top_pages_router.yaml
cmraible: I think this is an actual problem that we should solve in the implementation, not just an issue that deleting the test solves.

If a request doesn't have location/device/source/etc filters, it will be routed to api_top_pages_v3. If it also doesn't supply _both_ `date_from` and `date_to`, then the request will return a 400, since both date parameters are required for `api_top_pages_v3`.

For a concrete example: `getPostStats` in ghost/core/core/server/services/stats/posts-stats-service.js:1153 calls `api_top_pages` currently with only a `date_from`. If/when we switch that function to use the router instead, all those requests will 404.

I think we should do one of the following:
1. Preferred: Remove the `date_to` requirement from `api_top_pages_v3`. I'm not sure if this was made required for a good reason, or if it was just an oversight. This seems like the simplest path, assuming it's not truly required for `api_top_pages_v3` to work properly
2. Capture this in the routing logic: if `date_to` is not supplied, then always route to `api_top_pages` instead of `api_top_pages_v3`

cmraible (changes requested):
See comment here: https://github.com/TryGhost/Ghost/pull/26361#discussion_r2800209830

cmraible (approved):

PR #26371, opened 2026-02-12
Unified E2E test infrastructure for monorepo and production images

rob-ghost:
closes https://linear.app/ghost/issue/BER-3293/

## Problem

E2E tests currently run against the monorepo dev image — a fundamentally different artefact from what gets deployed. Configuration differences, missing or differently-served assets, and divergent startup paths all mean passing E2E tests don't guarantee the production image works. We're building production Docker images now; we should be testing them.

At the same time, the local dev experience (yarn dev + fast iteration with HMR) must stay untouched. Developers shouldn't pay a tax for production confidence.

## Proposal

There are only two modes: **dev** (yarn dev running, reuses host dev servers) and **production** (Docker container from production image). There is no "monorepo container mode" — nobody runs containerised E2E tests against the dev image outside of dev mode.

The E2E CI job now depends on `job_docker_build_production` and pulls the full production image (`ghost`) from GHCR. A thin `Dockerfile.e2e` layers pre-built UMD bundles onto the production image so public apps (portal, comments, search, signup, announcements) are served from `/ghost/assets/` paths — the same way they'd be served in production with CDN overrides.

This also fixes an inconsistency where portal had a dedicated Vite preview container while the other 4 public apps silently fell back to CDN. All 6 are now tested the same way.

The migration command is unified (`node node_modules/.bin/knex-migrator init` works on both images) and build commands are reusable npm scripts (`build:apps`, `build:docker`).

## Not Doing

- Changing the dev mode workflow — `yarn dev` + `cd e2e && yarn test` is completely untouched
- Changing how public apps are built or bundled
- Adding production E2E as a separate job alongside monorepo E2E — we replaced, not duplicated

## Trade-offs

- **No more auto-build for container mode.** Previously `pretest` auto-built a monorepo Docker image. Container mode now requires an explicit production image (`GHOST_E2E_IMAGE`). This is intentional — building a production image locally is a heavier operation that shouldn't happen silently. Dev mode (where auto-detection matters) is unchanged.
- **Portal loses its dedicated Vite container in container mode.** It's now served as a static asset like the other 5 apps. Dev mode (where hot-reload matters) is unchanged. Container mode never benefited from Vite — containers are ephemeral per-test.
- **UMD bundles built on host, not in the Docker layer.** Building inside the Dockerfile would require copying all app source + node_modules (slow, large image). Host builds via Nx are fast and cached. The tradeoff is the host must have Node/Yarn — which it always does in CI and local dev.
- **E2E tests now block on production image build.** The production build is heavier than the monorepo build (includes `yarn build:production` + `pack:standalone`). This adds wall-clock time to the E2E pipeline, but the whole point is testing the real artifact.

## Rollback

**Easy.** Revert two commits. No data migrations, no infrastructure changes, no customer impact.

## Blast Radius

- **E2E test infrastructure only.** No changes to Ghost core, admin, or any published package.
- **CI workflow** — E2E job now depends on production image build instead of dev image build.
- **Dev mode** is completely unaffected.
- **Self-hosters and Ghost(Pro)** are not affected.


cmraible:
I'll take a closer look at this later today but directionally I'm very excited about this 🙌

cmraible (approved):
Really nice work, very excited for this 🙌

PR #26380, opened 2026-02-12
Added `offer_redemptions` to subscription API response

mike182uk:

Adds `subscription.offer_redemptions` (sourced from offer_redemptions table) to the member API response, and updates admin to display all historical offers on the member page instead of only the current active offer. This is to support retention offers where the system now has to account for multiple offer redemptions potentially occurring on the same sub

### DIFF
--- a/ghost/admin/app/components/member/subscription-detail-box.hbs
+++ b/ghost/admin/app/components/member/subscription-detail-box.hbs
@@ -20,7 +20,7 @@
             </p>
             {{/if}}
         </div>
-        {{#if (or @sub.cancellationReason @sub.offer)}}
+        {{#if (or @sub.cancellationReason @sub.offer_redemptions.length)}}
         <div class="gh-membertier-offer-cancellation">
             {{#if @sub.cancellationReason}}
             <div>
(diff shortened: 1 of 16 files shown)


ghost/admin/app/components/member/subscription-detail-box.hbs
```diff
@@ -20,40 +20,40 @@
...
         <div class="gh-membertier-offer-cancellation">
             {{#if @sub.cancellationReason}}
             <div>
                 <h4>Cancellation reason</h4>
                 <div class="gh-membertier-cancelreason">{{@sub.cancellationReason}}</div>
             </div>
             {{/if}}
-            {{#if @sub.offer}}
-                {{#if (eq @sub.offer.type "trial")}}
+            {{#each @sub.offer_redemptions as |offer|}}
```
sagzy: For Pro users: given that the Admin change will go out instantly, but the core change will only go out with the next full rollout, there will a period of time with a regression in Admin: the member subscription box won't render the details of an offer (as `@sub.offer_redemptions` is undefined)

ghost/admin/app/components/members/filters/offers.js
```diff
@@ -10,7 +10,9 @@ export const OFFERS_FILTER = {
     getColumnValue: (member) => {
         return {
             class: 'gh-members-list-labels',
-            text: (member.subscriptions ?? []).map(label => label.offer?.name).join(', ')
+            text: (member.subscriptions ?? [])
+                .flatMap(sub => (sub.offer_redemptions ?? []).map(o => o.name))
+                .join(', ')
```
sagzy, line 18: Same potential issue here for Pro users due to the timing difference between Admin and core rollouts

ghost/core/core/server/services/offers/offer-bookshelf-repository.js
```diff
@@ -208,6 +208,27 @@ class OfferBookshelfRepository {
         return redemptions.map(r => r.get('offer_id'));
     }
 
+    /**
+     * @param {object} options
+     * @param {string[]} options.subscriptionIds
+     * @param {import('knex').Knex.Transaction} [options.transacting]
+     * @returns {Promise<Array<{subscription_id: string, offer_id: string}>>}
+     */
+    async getRedeemedOfferIdsForSubscriptions({subscriptionIds, transacting}) {
```
sagzy: nitpick: shall this method signature be `getRedeemedOfferIdsForSubscriptions(subscriptionIds, options = {})` (instead of `{subscriptionIds, transacting}`), to stay consistent with the rest of the methods in that repository?

sagzy (commented):
Code looks good, but we need to be careful about the timing difference between Admin / backend rollout on Pro

ghost/core/core/server/services/offers/offer-bookshelf-repository.js
mike182uk: Thanks, I've updated both `getRedeemedOfferIdsForSubscriptions` and `getRedeemedOfferIdsForSubscription` to be consistent with the other method signatures → https://github.com/TryGhost/Ghost/pull/26380/commits/9ccecca060eb1c80207753dd047792bebd198ba2

ghost/admin/app/components/member/subscription-detail-box.hbs
mike182uk: Good catch, wdyt to https://github.com/TryGhost/Ghost/pull/26380/commits/455e1123b5528ccb1f58a684463c3c793e2b028c ?

Then add a linear task to come back and clean this up later?

mike182uk: https://linear.app/ghost/issue/BER-3336

sagzy: Looks good @mike182uk

ghost/core/core/server/services/offers/offer-bookshelf-repository.js
sagzy: nice one

sagzy (approved):

PR #26383, opened 2026-02-12
Updated `linkSubscription` to clear stale `offer_id` on webhook sync

mike182uk:

The `offer_id` column previously persisted forever, even after the Stripe discount expired. This changes the sync guard to only preserve `offer_id` for active trials, and broadens the `OfferRedemptionEvent` trigger to fire on any offer_id change (not just `null` → `new`), enabling retention offer redemption tracking

### DIFF
--- a/ghost/core/core/server/services/members/members-api/repositories/member-repository.js
+++ b/ghost/core/core/server/services/members/members-api/repositories/member-repository.js
@@ -1159,20 +1159,28 @@ module.exports = class MemberRepository {
                 await stripeCustomerSubscriptionModel.destroy(options);
             }
         } else if (stripeCustomerSubscriptionModel) {
-            // CASE: Offer is already mapped against sub, don't overwrite it with NULL
-            // Needed for trial offers, which don't have a stripe coupon/discount attached to sub
+            const previousOfferId = stripeCustomerSubscriptionModel.get('offer_id');
+
+            // CASE: Only preserve offer_id for active trials (trial offers don't have Stripe discounts)
+            // Otherwise, allow offer_id to be cleared when the Stripe discount expires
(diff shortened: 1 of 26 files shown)


ghost/core/core/server/services/members/members-api/repositories/member-repository.js
```diff
@@ -1159,20 +1159,28 @@ module.exports = class MemberRepository {
...
         } else if (stripeCustomerSubscriptionModel) {
-            // CASE: Offer is already mapped against sub, don't overwrite it with NULL
-            // Needed for trial offers, which don't have a stripe coupon/discount attached to sub
+            const previousOfferId = stripeCustomerSubscriptionModel.get('offer_id');
+
+            // CASE: Only preserve offer_id for active trials (trial offers don't have Stripe discounts)
+            // Otherwise, allow offer_id to be cleared when the Stripe discount expires
             if (!subscriptionData.offer_id) {
-                delete subscriptionData.offer_id;
+                const trialEndAt = stripeCustomerSubscriptionModel.get('trial_end_at');
```
sagzy: Instead of reading `trial_end_at` from the cached value in the database, should we use the live value coming from Stripe? i.e. `subscriptionData.trial_end_at`

They should _normally_ be the same, but feels safer to use Stripe as source of truth rather than our cached value

mike182uk: Yeh, they shouldn't drift but your right that makes sense to read directly from the Stripe data seeming though we have it already. Updated in https://github.com/TryGhost/Ghost/pull/26383/commits/73ad4d60501b91af38b04ea67d5ff5f87c29cfe1

sagzy (approved):

PR #26384, opened 2026-02-12
Updated default values for portal transistor settings

9larsons:
ref https://linear.app/ghost/issue/NY-1034/


9larsons:
Changing some default string values before the migration goes out - does not need full review.

PR #26387, opened 2026-02-12
Trimmed leading empty entries in all time data for traffic api data

9larsons:
ref https://linear.app/ghost/issue/ONC-1457/

Added util to strip leading zeroes in the api_kpis response for 'all time'. Previously, this would show 1000 days of data, regardless of when the data started for a given site.

Ultimately we'll want to fix this in the api - this is a short term and easier fix for now.

### DIFF
--- a/apps/stats/src/utils/chart-helpers.ts
+++ b/apps/stats/src/utils/chart-helpers.ts
@@ -18,6 +18,20 @@ export const getPeriodText = (range: number): string => {
     return '';
 };
 
+/**
+ * Truncates leading empty data points from the beginning of a dataset,
+ * keeping one zero entry before the first real data for a smooth chart transition.
+ * Ultimately, we should fix the API to return only what we want to see.
+ */
+export function truncateLeadingEmptyData<T extends {value?: number}>(data: T[]): T[] {
+    const firstNonEmptyIndex = data.findIndex(item => Number(item.value) > 0);
+    if (firstNonEmptyIndex > 1) {
+        // Keep one zero entry before the first real data
+        return data.slice(firstNonEmptyIndex - 1);
+    }
+    return data;
+}
+
 type AggregationType = 'sum' | 'avg' | 'exact';
 
 type AggregationStrategy = 'none' | 'weekly' | 'monthly' | 'monthly-exact';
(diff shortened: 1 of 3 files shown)


apps/stats/src/utils/chart-helpers.ts
```diff
@@ -18,6 +18,20 @@ export const getPeriodText = (range: number): string => {
     return '';
 };
 
+/**
+ * Truncates leading empty data points from the beginning of a dataset,
+ * keeping one zero entry before the first real data for a smooth chart transition.
+ * Ultimately, we should fix the API to return only what we want to see.
```
EvanHahn, line 24: When do we plan to do this? Can we update this comment with a Linear ticket?

apps/stats/test/unit/utils/chart-helpers.test.ts
```diff
@@ -51,6 +52,87 @@ describe('chart-helpers', () => {
...
+        it('handles all zero values', () => {
+            const data = [
+                {date: '2024-01-01', value: 0},
+                {date: '2024-02-01', value: 0},
+                {date: '2024-03-01', value: 0}
+            ];
+
+            const result = truncateLeadingEmptyData(data);
+
+            // No non-empty data found, returns original
```
EvanHahn: nit: I think it's clear without the comment.

```suggestion
```

apps/stats/src/views/Stats/Web/components/web-kpis.tsx
```diff
@@ -23,15 +24,23 @@ const WebKPIs: React.FC<WebKPIsProps> = ({data, range, isLoading}) => {
             return [];
         }
 
-        return sanitizeChartData<KpiDataItem>(data, range, currentMetric.dataKey as keyof KpiDataItem, 'sum')?.map((item: KpiDataItem) => {
+        const sanitizedData = sanitizeChartData<KpiDataItem>(data, range, currentMetric.dataKey as keyof KpiDataItem, 'sum')?.map((item: KpiDataItem) => {
```
EvanHahn, line 27: nit: `sanitizeChartData` seems like it should always return an array so we can remove this `?` and the `|| []` below. (Arguably out of scope for this PR—feel free to ignore and we can address in a followup.)

EvanHahn (approved):
A few non-blocking comments.

apps/stats/src/views/Stats/Web/components/web-kpis.tsx
9larsons: Looks like it's visitors data that is sometimes null - this is another drop in the bucket of why we need to revise the stats APIs. We should always have a data return. Good shout and fixed via other changes.

PR #26392, opened 2026-02-12
Match small portal popover's X to large version

cathysarisky:
Currently, the colors used in portal for the gh-portal-close (X) button are inconsistent.  the small popover uses --grey10 (which has poor contrast), while the larger popover (when showing tiers) uses --grey6.  This PR changes gh-portal-close to consistently use --grey6, matching the current visual styling of the large popover's close button.


9larsons (approved):
Thanks! 👍

PR #26394, opened 2026-02-13
minor accessibility updates for Portal 

cathysarisky:
This PR makes minor adjustments to portal to increase accessibility for screen readers.  No visual changes, just better accessibility for users who need to tab navigate or use a screen reader.

The close button is now a button, not a div, which buys it automatic focus-ability.  I added css to remove the default button styling, and the click handler on the button, so that screen users and tab-navigators can access it.

Also updated the benefits "checks" to not get read by the screen reader.


9larsons:
Pushed a follow-up commit addressing review feedback:
- added `aria-label="Close popup"`, `type="button"`, and `aria-hidden="true"` on the inner SVG so the X is actually announced and won't submit surrounding forms
- removed a duplicate onClick that was firing closePopup twice via bubbling
- applied the same button conversion to popup-notification's close, which had the same `<div>`+onClick pattern
- reverted the comments-ui SVG change to keep this PR Portal-only
- scoped two magic-link tests by form, since the X now has a "Close" accessible name and was colliding with the existing action button

PR #26406, opened 2026-02-16
Removed unnecessary PII from member geolocation data

betschki:
closes https://github.com/TryGhost/Ghost/issues/26315

The geolocation service stored the full geojs.io response (IP, coordinates, ASN, org, etc.) but only country, country_code, and region are ever used. This strips the response to just those three fields to reduce stored PII.

Got some code for us? Awesome 🎊!

Please take a minute to explain the change you're making:
- Why are you making it?
- What does it do?
- Why is this something Ghost users or developers need?

Please check your PR against these items:

- [x] I've read and followed the [Contributor Guide](https://github.com/TryGhost/Ghost/blob/main/.github/CONTRIBUTING.md)
- [x] I've explained my change
- [x] I've written an automated test to prove my change works

We appreciate your contribution! 🙏

---

> [!NOTE]
> **Low Risk**
> Small behavioral change that narrows stored data; main risk is any downstream code implicitly relying on removed geojs fields.
> 
> **Overview**
> Member geolocation storage is reduced to only `country`, `country_code`, and `region`, stripping other fields returned by geojs.io (e.g. IP, coordinates, ASN/org data) before persisting.
> 
> Test fixtures and assertions are updated to expect the trimmed geolocation object, and the members seeder now generates only those three geolocation fields.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e3c87f1c3d933e65e841ce5858e8e2e1cfe90049. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #26414, opened 2026-02-16
Added data migration to strip PII from existing member geolocation

betschki:
ref https://github.com/TryGhost/Ghost/pull/26406

The geolocation column stored the full geojs.io response including IP addresses, coordinates, and ASN data. This migration rewrites existing rows to keep only country, country_code, and region.

- Why are you making it?
- What does it do?

---

> [!NOTE]
> **Medium Risk**
> Updates existing member data in the `members` table and discards fields that cannot be restored, so any bugs could cause permanent data loss in that column.
> 
> **Overview**
> Adds a new transactional data migration that rewrites existing `members.geolocation` JSON to retain only `country`, `country_code`, and `region`, removing other previously stored fields.
> 
> The migration skips null/invalid JSON rows and no-ops when records are already clean, updates rows in 1000-record batches, and marks the change as *irreversible* on rollback (logs a warning instead of restoring stripped data).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 91b8c8e36a4c72119dd9087112ac1fdf5476f89f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


kevinansfield (commented):
## Migration efficiency review

The current approach selects all rows into Node.js, parses JSON in JS, then fires individual `UPDATE` statements per row via `Promise.all` in batches of 1000. For a site with 750k+ members this means ~750k individual UPDATE queries, significant memory usage loading all rows into Node, and row-level locks held for the entire transaction.

### Suggestion: Batched raw SQL with MySQL JSON functions

MySQL 5.7+ (which Ghost requires) has `JSON_EXTRACT`, `JSON_UNQUOTE`, and `JSON_OBJECT` — the entire transformation can happen in the database with no data transferred to Node.js.

Key changes:
1. **Use `createNonTransactionalMigration`** so each batch commits independently — only ~5000 rows locked at a time instead of all 750k
2. **Use raw SQL with JSON functions** — each batch is a single UPDATE statement instead of 5000 individual ones
3. **Gate behind `DatabaseInfo.isMySQL`** — SQLite JSON function support is inconsistent and PII stripping isn't needed for dev databases
4. **Self-healing batching** — already-cleaned rows won't match the WHERE clause, so interrupted migrations resume safely

```js
const logging = require('@tryghost/logging');
const DatabaseInfo = require('@tryghost/database-info');
const {createNonTransactionalMigration} = require('../../utils');

const BATCH_SIZE = 5000;

module.exports = createNonTransactionalMigration(
    async function up(knex) {
        if (!DatabaseInfo.isMySQL(knex)) {
            logging.info('Skipping PII strip on non-MySQL database');
            return;
        }

        let totalUpdated = 0;

        // eslint-disable-next-line no-constant-condition
        while (true) {
            const batch = await knex('members')
                .select('id')
                .whereNotNull('geolocation')
                .whereRaw("JSON_EXTRACT(geolocation, '$.ip') IS NOT NULL")
                .limit(BATCH_SIZE);

            if (!batch.length) {
                break;
            }

            const ids = batch.map(r => r.id);

            await knex.raw(`
                UPDATE members
                SET geolocation = JSON_OBJECT(
                    'country', JSON_UNQUOTE(JSON_EXTRACT(geolocation, '$.country')),
                    'country_code', JSON_UNQUOTE(JSON_EXTRACT(geolocation, '$.country_code')),
                    'region', JSON_UNQUOTE(JSON_EXTRACT(geolocation, '$.region'))
                )
                WHERE id IN (?)
            `, [ids]);

            totalUpdated += ids.length;
            logging.info(`Stripped PII from ${totalUpdated} members so far...`);
        }

        if (totalUpdated === 0) {
            logging.info('No members with PII in geolocation data found — skipping');
        } else {
            logging.info(`Stripped PII from member geolocation data: ${totalUpdated} rows updated`);
        }
    },
    async function down() {
        logging.warn('Rollback not possible: stripped geolocation fields cannot be restored');
    }
);
```

### Performance comparison at 750k members

| | Current | Suggested |
|---|---|---|
| DB round-trips | ~750k | ~300 (150 SELECT + 150 UPDATE) |
| Memory | All geolocation JSON loaded into Node | Only IDs (~5000 at a time) |
| Lock duration | Entire transaction (all rows) | ~5000 rows at a time |
| Resumable if interrupted | No (transactional) | Yes (already-cleaned rows are skipped) |

Note: `JSON_VALID()` was omitted — `JSON_EXTRACT` on invalid JSON returns `NULL`, so the `$.ip IS NOT NULL` check naturally skips corrupt rows without needing an explicit validity scan on every row.

betschki:
@kevinansfield that's a great suggestion and makes perfect sense. Pivoted to that

betschki:
@kevinansfield do you have any more thoughts on this?

9larsons:
Hey @betschki, sorry for the long silence here, and thanks for sticking with it — the reworked batched raw-SQL approach looks solid. However, I think we need to support SQLite. I chatted w/ Kevin and the main concern was the new approach with MySQL batching wouldn't work with SQLite, though I really don't want to see divergences in data.

I think it's fine to fall back to your original approach for SQLite support (fetch rows, strip in JS, update in batches) — it's slower, but sites with members tables big enough for that to matter shouldn't be running SQLite anyway, so a simple `isMySQL ? <current raw-SQL path> : <row-based path>` split inside the same migration is perfectly acceptable. Keep the MySQL fast path exactly as-is.

Would you mind tackling that? And bump the migration version while you're at it.

The other piece holding this up was testing this with large dbs, which I'll put on my list once the final version is ready.

And after that's clear, we'll just have to make sure the migration version is accurate.

betschki:
@9larsons good points, thank you! I've updated the migration. Let me know if we need anything else here

9larsons:
@betschki Thanks! Sorry for the delays, going to try and get this tested and merged next week. This week got away from me and I want some fresher eyes.

9larsons:
@betschki I've tested this and we've got a couple dbs that have a _lot_ of rows that we'll need to handle before doing this as a downtime migration. Stay tuned, haven't forgotten about this!

Issue #26381, opened 2026-02-12
Welcome email sender still appears as noreply@ in preview


molly:
### Issue Summary

I see that #26358 fixed the issue where welcome emails are being sent from the noreply@ address. However I see that the welcome email preview in settings still shows the noreply@ address for the sender. A test email shows the welcome emails are properly being sent with the expected newsletter@ address, so this is probably just a display issue.

<img width="694" height="711" alt="Image" src="https://github.com/user-attachments/assets/2eae531b-4102-40e3-9e97-b3378f97cdfa" />

<img width="708" height="424" alt="Image" src="https://github.com/user-attachments/assets/40643dd6-34c7-490f-893f-e3d123b78944" />

### Steps to Reproduce

1. Open the Welcome emails section in Settings and click "Edit" on one of the email templates

### Ghost Version

v6.18.2

### Node.js Version

v22.18.0

### How did you install Ghost?

Ubuntu, v24.04.2 LTS

### Database type

MySQL 8

### Browser & OS version

OSX Sonoma 14.5, Chrome 144.0.7559.98

### Relevant log / error output

```shell

```

### Code of Conduct

- [x] I agree to be friendly and polite to people in this repository

vikaspotluri123:
@troyciesco I think GH only linked the PR (not 100% sure, but it might only look for <closes #> vs <closes # and # and #>). This should be good to close, right?

cmraible:
Yep, we can close this - the fix above will be included in the next Ghost release. Thanks @molly for reporting!

PR #26420, opened 2026-02-16
Fixed display of member welcome email sender in modal

troyciesco:
closes [NY-1041](https://linear.app/ghost/issue/NY-1041/oss-issue-welcome-email-sender-still-appears-as-noreply-in-preview) and https://github.com/TryGhost/Ghost/issues/26381

- In https://github.com/TryGhost/Ghost/pull/26358 we made it so that member welcome emails, when sent, would use sender info and reply_to from the default newsletter
- this PR updates the modal to display that information properly as well
- Since this info is used in both the button that opens the modal and the modal itself, added a hook to de-duplicate some of the logic for determining exactly what to display
- No logic yet for updating sender name, email, and reply to specifically for welcome emails, but this code is written in such a way that once that's supported it can easily take precedence over default newsletter values 

### State 1 - defaults
The default values for the newsletter (first screenshot) are what welcome emails use.

<table> <tr> <th>Default Newsletter Settings</th> <th>Welcome Email Buttons</th> <th>Modal</th> </tr> <tr> <td align="center">Screenshot 2026-02-16 at 3 01 30 PM</td> <td align="center">Screenshot 2026-02-16 at 3 02 04 PM</td> <td align="center">Screenshot 2026-02-16 at 3 01 43 PM</td> </tr> </table>

### State 2 - custom name and sender email
Updated sender name and sender email for the default newsletter. Sender name is updated in the button that opens the modal, and name + sender email are updated in the modal. The test email matches the same values.

<table> <tr> <th>Newsletter Settings</th> <th>Welcome Email Buttons</th> </tr> <tr> <td align="center">Screenshot 2026-02-16 at 3 02 38 PM</td> <td align="center">Screenshot 2026-02-16 at 3 02 48 PM</td> </tr> <tr> <th>Modal</th> <th>Resulting Test Email</th> </tr> <tr> <td align="center">Screenshot 2026-02-16 at 3 02 54 PM</td> <td align="center">Screenshot 2026-02-16 at 3 03 08 PM</td> </tr> </table>

### State 3 - custom reply to
Same as above, but with a custom reply-to address. It appears in the modal and in the actual sent email.

<table> <tr> <th>Newsletter Reply-To Setting</th> <th>Modal (w/ Reply-To)</th> <th>Resulting Test Email</th> </tr> <tr> <td align="center">Screenshot 2026-02-16 at 3 03 24 PM</td> <td align="center">Screenshot 2026-02-16 at 3 03 33 PM</td> <td align="center">Screenshot 2026-02-16 at 3 03 42 PM</td> </tr> </table>

### DIFF
--- a/apps/admin-x-settings/src/hooks/use-welcome-email-sender-details.ts
+++ b/apps/admin-x-settings/src/hooks/use-welcome-email-sender-details.ts
@@ -1,0 +1,54 @@
+import {getSettingValues} from '@tryghost/admin-x-framework/api/settings';
+import {renderReplyToEmail, renderSenderEmail} from '../utils/newsletter-emails';
+import {useBrowseNewsletters} from '@tryghost/admin-x-framework/api/newsletters';
+import {useGlobalData} from '../components/providers/global-data-provider';
+import {useMemo} from 'react';
+import type {AutomatedEmail} from '@tryghost/admin-x-framework/api/automated-emails';
+
+type AutomatedEmailSenderFields = Pick<AutomatedEmail, 'sender_name' | 'sender_email' | 'sender_reply_to'> | null | undefined;
+
+const trimValue = (value: string | null | undefined) => value?.trim() || '';
+
+export const useWelcomeEmailSenderDetails = (automatedEmail: AutomatedEmailSenderFields) => {
+    const {settings, config} = useGlobalData();
+    const [siteTitle, defaultEmailAddress, supportEmailAddress] = getSettingValues<string>(settings, ['title', 'default_email_address', 'support_email_address']);
+    const {data: newslettersData} = useBrowseNewsletters({
+        searchParams: {
+            filter: 'status:active',
+            limit: '1'
+        }
+    });
+    const defaultNewsletter = newslettersData?.newsletters?.[0];
+
+    return useMemo(() => {
+        const automatedSenderName = trimValue(automatedEmail?.sender_name);
+        const automatedSenderEmail = trimValue(automatedEmail?.sender_email);
+        const automatedSenderReplyTo = trimValue(automatedEmail?.sender_reply_to);
+
+        const defaultNewsletterSenderName = trimValue(defaultNewsletter?.sender_name);
+        const defaultNewsletterSenderEmail = defaultNewsletter ? trimValue(renderSenderEmail(defaultNewsletter, config, defaultEmailAddress)) : '';
+        const defaultNewsletterReplyTo = defaultNewsletter ? trimValue(renderReplyToEmail(defaultNewsletter, config, supportEmailAddress, defaultEmailAddress)) : '';
+
+        const resolvedSenderName = automatedSenderName || defaultNewsletterSenderName || trimValue(siteTitle) || 'Your Site';
+        const resolvedSenderEmail = automatedSenderEmail || defaultNewsletterSenderEmail || trimValue(defaultEmailAddress) || '';
+        const resolvedReplyToEmail = automatedSenderReplyTo || defaultNewsletterReplyTo || '';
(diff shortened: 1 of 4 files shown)


apps/admin-x-settings/src/hooks/use-welcome-email-sender-details.ts
```diff
@@ -0,0 +1,54 @@
+import {getSettingValues} from '@tryghost/admin-x-framework/api/settings';
+import {renderReplyToEmail, renderSenderEmail} from '../utils/newsletter-emails';
```
troyciesco, line 2: it feels slightly odd to import these from `newsletter-emails`, but we're already kind of blurring the lines here by using the default newsletter values anyway.

apps/admin-x-settings/src/hooks/use-welcome-email-sender-details.ts
```diff
@@ -0,0 +1,54 @@
...
+        const automatedSenderReplyTo = trimValue(automatedEmail?.sender_reply_to);
+
+        const defaultNewsletterSenderName = trimValue(defaultNewsletter?.sender_name);
+        const defaultNewsletterSenderEmail = defaultNewsletter ? trimValue(renderSenderEmail(defaultNewsletter, config, defaultEmailAddress)) : '';
+        const defaultNewsletterReplyTo = defaultNewsletter ? trimValue(renderReplyToEmail(defaultNewsletter, config, supportEmailAddress, defaultEmailAddress)) : '';
+
+        const resolvedSenderName = automatedSenderName || defaultNewsletterSenderName || trimValue(siteTitle) || 'Your Site';
+        const resolvedSenderEmail = automatedSenderEmail || defaultNewsletterSenderEmail || trimValue(defaultEmailAddress) || '';
+        const resolvedReplyToEmail = automatedSenderReplyTo || defaultNewsletterReplyTo || '';
+        const hasDistinctReplyTo = resolvedReplyToEmail !== '' && resolvedReplyToEmail !== resolvedSenderEmail;
```
cmraible, line 35: non-blocking suggestion: it smells a bit to me that this logic is repeated in the frontend, which could lead to drift. I'm wondering if this would be better to add as a sort of "calculated field" in the `GET /automated_emails` endpoint?

Currently the endpoint responds with sender_name, sender_email and sender_reply_to straight from the DB. In a future world where users can use these fields to override the value, I think we'll still want to return these fields straight from the DB, but I wonder if we should also return a `derived_*` or `calculated_*` or `actual_*` version of these fields based on this logic from the backend?

PS I think you suggested this last week when we were working on the OG reply-to fix, I should have listened 😃

cmraible (commented):
See comment below - WDYT? I don't feel too strongly about it, could go either way.

apps/admin-x-settings/src/hooks/use-welcome-email-sender-details.ts
troyciesco: ooo interesting, yeah that might be better...the tradeoff is we'd have to invalidate the automated emails query when newsletters are edited so we don't show stale data. That feels clunky too, but also seems like a path less likely to introduce bugs for users down the line. I'll explore this tomorrow!

cmraible: > we'd have to invalidate the automated emails query when newsletters are edited so we don't show stale data

Good point - I hadn't considered that! I also wonder if that clunkiness is just more signal that we should add the ability to edit these values explicitly for welcome emails - in which case I think we could temporarily ignore the invalidation when newsletters are edited, given it's a bit of an edge case and will go away soon.

Just thinking out loud!

cmraible (approved):

PR #26438, opened 2026-02-17
Replaced `offer_id` guards with `hasActiveOffer` check in retention flow

mike182uk:

Members with expired offer discounts were blocked from seeing or claiming retention offers because the guards only checked for `offer_id` presence. The new `hasActiveOffer` utility checks `discount_start`/`discount_end`, active trials, and falls back to offer duration lookup for legacy data


ghost/core/test/unit/server/services/members/members-api/utils/has-active-offer.test.js
```diff
@@ -0,0 +1,177 @@
...
+        const offersAPI = createOffersAPI(null);
+
+        const result = await hasActiveOffer(model, offersAPI);
+
+        assert.equal(result, false);
+    });
+
+    // Priority: discount_start takes precedence over trial and legacy fallback
+
+    it('discount_start takes precedence over trial_end_at', async function () {
```
sagzy, line 162: This shouldn't really happen in real life (we either have an active discount or an active trial), but the extra unit test doesn't harm 👍

ghost/core/core/server/services/members/members-api/utils/has-active-offer.js
```diff
@@ -0,0 +1,60 @@
...
+
+            return new Date() < end;
+        }
+    } catch (e) {
+        // If we can't look up the offer, err on the side of blocking
+        return true;
+    }
+
+    return false;
+};
```
sagzy, line 60: Nice!

Suggestion (non-blocking): there seems to be an opportunity to extract the common logic here and in next-payment-calculator ([code](https://github.com/TryGhost/Ghost/blob/6e6e604402194260565d4a50c3511cf809edf87d/ghost/core/core/server/services/members/members-api/services/next-payment-calculator.js#L101)). Maybe next-payment-calculator logic calls hasActiveOffer first, then extract the start / end dates as a second step? 🤔 Maybe also something that is better suited for a follow-up tidying PR

sagzy (approved):
LGTM

ghost/core/core/server/services/members/members-api/utils/has-active-offer.js
mike182uk: Yeh there is definitely some overlap, but there was enough differences that it felt like a new fn was warranted. I think we could definitely DRY these up though - It would probably be easier in a focused tidying rather than muddying this PR - https://linear.app/ghost/issue/BER-3338

PR #26441, opened 2026-02-17
Swapped Ghost core use of node --watch to nodemon for Docker hot reload

jonatansberg:
`node --watch` missed changes in lazily loaded modules when run in Docker, so `yarn dev` now runs through `nodemon` for reliable restarts on source edits.

When testing locally this seems to reliably restart the backend regardless of which files are edited. 

Upstream Node bug: https://github.com/nodejs/node/issues/51621


kevinansfield (approved):
I've been running this and all is working well 👌

PR #26442, opened 2026-02-17
Removed dev.js and legacy scripts

9larsons:
no ref

With the update to the React-based Admin app, we've overhauled the dev setup to lean fully into Docker. This commit removes the old/legacy setup to prevent confusion on what's the intended way to develop with Ghost.

The `dev.js` script has handled many tasks in the past. It appears to me as if the browser tests are the only necessary use case right now, so I've stripped away the unnecessary pieces and moved/renamed it to a dedicated script describing its purpose.


e2e/package.json
```diff
@@ -12,7 +12,7 @@
     "docker:build:ghost": "cd .. && docker build -t ghost-monorepo:latest -f Dockerfile .",
     "docker:update": "docker compose pull && docker compose up -d --force-recreate",
     "prepare": "tsc --noEmit",
-    "pretest": "(test -n \"$GHOST_E2E_SKIP_BUILD\" || test -n \"$CI\") && echo 'Skipping Docker build (GHOST_E2E_SKIP_BUILD or CI is set)' || docker compose -f ../compose.yml build ghost tb-cli",
+    "pretest": "(test -n \"$GHOST_E2E_SKIP_BUILD\" || test -n \"$CI\") && echo 'Skipping Docker build (GHOST_E2E_SKIP_BUILD or CI is set)' || docker compose -f ../compose.dev.yaml build ghost-dev",
```
cmraible: I _think_ we can just remove this? `yarn dev` has to already be running for the e2e tests to work now I think, so the extra docker build is redundant

.env.example
```diff
@@ -14,7 +14,7 @@
 
 # GHOST_DEV_APP_FLAGS=
```
cmraible: nit: we can remove these lines too, they're irrelevant now that dev.js and compose.yml are gone

package.json
```diff
@@ -26,46 +26,26 @@
...
-    "docker:sleep:stop": "docker stop ghost-devcontainer",
-    "docker:test:unit": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm --no-deps ghost yarn test:unit",
-    "docker:test:browser": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} docker compose run --rm ghost yarn test:browser",
-    "docker:test:all": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm ghost yarn nx run ghost:test:all",
-    "docker:test:e2e": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm ghost yarn test:e2e",
-    "docker:reset": "docker compose --profile all down -v && docker compose up -d --wait",
-    "docker:restart": "docker compose down && docker compose up -d --wait",
-    "docker:down": "docker compose --profile all down",
+    "docker:build": "yarn docker:clean && yarn build:clean && docker compose -f compose.dev.yaml --profile all build",
+    "docker:clean": "echo \"Deleting node_modules volumes...\" && docker compose -f compose.dev.yaml --profile all down --remove-orphans && docker volume ls -q -f name=ghost_node_modules | xargs -I{} docker volume rm {}",
```
cmraible: I don't think we need `docker:clean` anymore: compose.dev.yaml doesn't store the node_modules in a volume, so there's nothing to cleanup 😄

package.json
```diff
@@ -26,46 +26,26 @@
...
+    "docker:reset": "docker compose -f compose.dev.yaml --profile all down -v && docker compose -f compose.dev.yaml up -d --wait",
+    "docker:restart": "docker compose -f compose.dev.yaml down && docker compose -f compose.dev.yaml up -d --wait",
```
cmraible: Not sure if either of these make sense to keep around anymore either, since we generally don't run `docker compose up -d` without the `yarn dev` wrapper.

package.json
```diff
@@ -26,46 +26,26 @@
...
-    "docker:sleep": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} docker compose run -d --name ghost-devcontainer --rm -it ghost /bin/bash -c 'sleep infinity'",
-    "docker:sleep:stop": "docker stop ghost-devcontainer",
-    "docker:test:unit": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm --no-deps ghost yarn test:unit",
-    "docker:test:browser": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} docker compose run --rm ghost yarn test:browser",
-    "docker:test:all": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm ghost yarn nx run ghost:test:all",
-    "docker:test:e2e": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm ghost yarn test:e2e",
-    "docker:reset": "docker compose --profile all down -v && docker compose up -d --wait",
-    "docker:restart": "docker compose down && docker compose up -d --wait",
-    "docker:down": "docker compose --profile all down",
+    "docker:build": "yarn docker:clean && yarn build:clean && docker compose -f compose.dev.yaml --profile all build",
```
cmraible:
```suggestion
    "docker:build": "docker compose -f compose.dev.yaml build",
```

We can simplify this now:
- docker:clean - see comment below
- build:clean - we shouldn't need to do this before building the image defined in `compose.dev.yaml`
- --profile all - this profile doesn't exist in compose.dev.yaml

cmraible (changes requested):
Noted a couple things that we can simplify further, but overall this looks good! Happy to take another pass in the morning 👍

e2e/package.json
9larsons: I've updated this to point to the right image. I still think it's useful to have around, since I don't love the requirement that `yarn dev` be running (and afaict, it's not required - we just have handling for both cases now).

package.json
9larsons: I've updated the scripts. I still like `docker:clean`'s other functions. Sometime a nuke is necessary and it's nice to have a script for it.

package.json
9larsons: Ah, right - very good point. I suppose let's clean heavily now and add back whatever later we find helpful.

package.json
```diff
@@ -26,46 +26,26 @@
...
-    "docker:test:browser": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} docker compose run --rm ghost yarn test:browser",
-    "docker:test:all": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm ghost yarn nx run ghost:test:all",
-    "docker:test:e2e": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm ghost yarn test:e2e",
-    "docker:reset": "docker compose --profile all down -v && docker compose up -d --wait",
-    "docker:restart": "docker compose down && docker compose up -d --wait",
-    "docker:down": "docker compose --profile all down",
+    "docker:build": "docker compose -f compose.dev.yaml build",
+    "docker:clean": "docker compose -f compose.dev.yaml down -v --remove-orphans --rmi local",
+    "docker:reset": "docker compose -f compose.dev.yaml --profile all down -v && docker compose -f compose.dev.yaml up -d --wait",
+    "docker:restart": "docker compose -f compose.dev.yaml down && docker compose -f compose.dev.yaml up -d --wait",
```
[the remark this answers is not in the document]
9larsons: We no longer have `docker:reset`. 

`docker:clean` and `down` have been updated to use `${DEV_COMPOSE_FILES}` - `--profile all` because of the additional compose files (analytics & storage).

package.json
```diff
@@ -26,46 +26,23 @@
...
-    "docker:test:e2e": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm ghost yarn test:e2e",
-    "docker:reset": "docker compose --profile all down -v && docker compose up -d --wait",
-    "docker:restart": "docker compose down && docker compose up -d --wait",
-    "docker:down": "docker compose --profile all down",
+    "reset:data": "docker exec ghost-dev bash -c 'cd /home/ghost/ghost/core && node index.js generate-data --clear-database --quantities members:1000,posts:100 --seed 123'",
+    "reset:data:empty": "docker exec ghost-dev bash -c 'cd /home/ghost/ghost/core && node index.js generate-data --clear-database --quantities members:0,posts:0 --seed 123'",
+    "reset:data:xxl": "docker exec ghost-dev bash -c 'cd /home/ghost/ghost/core && node index.js generate-data --clear-database --quantities members:2000000,posts:0,emails:0,members_stripe_customers:0,members_login_events:0,members_status_events:0 --seed 123'",
+    "docker:build": "docker compose -f compose.dev.yaml build",
+    "docker:clean": "docker compose -f compose.dev.yaml down -v --remove-orphans --rmi local",
+    "docker:down": "docker compose -f compose.dev.yaml down",
```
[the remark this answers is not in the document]
9larsons: Updated `build` to use dev compose files.

package.json
```diff
@@ -26,46 +26,23 @@
...
-    "docker:sleep": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} docker compose run -d --name ghost-devcontainer --rm -it ghost /bin/bash -c 'sleep infinity'",
-    "docker:sleep:stop": "docker stop ghost-devcontainer",
-    "docker:test:unit": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm --no-deps ghost yarn test:unit",
-    "docker:test:browser": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} docker compose run --rm ghost yarn test:browser",
-    "docker:test:all": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm ghost yarn nx run ghost:test:all",
-    "docker:test:e2e": "COMPOSE_PROFILES=${COMPOSE_PROFILES:-ghost} NX_DAEMON=false docker compose run --rm ghost yarn test:e2e",
-    "docker:reset": "docker compose --profile all down -v && docker compose up -d --wait",
-    "docker:restart": "docker compose down && docker compose up -d --wait",
-    "docker:down": "docker compose --profile all down",
+    "reset:data": "docker exec ghost-dev bash -c 'cd /home/ghost/ghost/core && node index.js generate-data --clear-database --quantities members:1000,posts:100 --seed 123'",
```
cmraible, line 36: I'd been doing this manually when needed - this is much more convenient, good call!

cmraible (approved):

PR #26452, opened 2026-02-17
Removed welcomeEmails feature flag checks from Ember Admin

cmraible:
ref https://linear.app/ghost/issue/NY-971/post-ga-cleanup-remove-the-welcomeemails-labs-flag-from-admin

There should be no behavioral changes in this commit, since the `welcomeEmails` flag has already been promoted to GA.

Now that welcome emails are in GA, we want to remove the flag completely. This commit removes any conditional behavior in the Admin ember app that depends on the `welcomeEmails` flag.

### DIFF
--- a/ghost/admin/app/utils/member-event-types.js
+++ b/ghost/admin/app/utils/member-event-types.js
@@ -14,9 +14,7 @@ export const ALL_EVENT_TYPES = [
 export function getAvailableEventTypes(settings, feature, hiddenEvents = []) {
     const extended = [...ALL_EVENT_TYPES];
 
-    if (feature.welcomeEmails) {
-        extended.push({event: 'automated_email_sent_event', icon: 'filter-dropdown-email-received', name: 'Welcome email received', group: 'emails'});
-    }
+    extended.push({event: 'automated_email_sent_event', icon: 'filter-dropdown-email-received', name: 'Welcome email received', group: 'emails'});
     if (settings.commentsEnabled !== 'off') {
         extended.push({event: 'comment_event', icon: 'filter-dropdown-comments', name: 'Comments', group: 'others'});
     }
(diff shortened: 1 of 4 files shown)


ghost/admin/app/utils/member-event-types.js
```diff
@@ -14,9 +14,7 @@ export const ALL_EVENT_TYPES = [
 export function getAvailableEventTypes(settings, feature, hiddenEvents = []) {
     const extended = [...ALL_EVENT_TYPES];
 
-    if (feature.welcomeEmails) {
-        extended.push({event: 'automated_email_sent_event', icon: 'filter-dropdown-email-received', name: 'Welcome email received', group: 'emails'});
-    }
+    extended.push({event: 'automated_email_sent_event', icon: 'filter-dropdown-email-received', name: 'Welcome email received', group: 'emails'});
```
troyciesco: how come this (and `feedback_event`) aren't in `ALL_EVENT_TYPES`?

troyciesco (changes requested):
this is a very light use of the request changes button, but i think the rabbit's nit is valid and i think my other comment will actually result in a change, unless i'm misunderstanding a step before `automated_email_sent_event` is bestowed the honor of being included in `ALL_EVENT_TYPES` 😁

cmraible:
Oh good call, I'll take a look 👀

ghost/admin/app/utils/member-event-types.js
cmraible: Fixed [here](https://github.com/TryGhost/Ghost/pull/26452/changes/ad32dd5a0a20dbc282694d2111aaf75108ebab53)!

Note: this changes the order of feedback and comments in the filter dropdown, but I don't think that they were ordered with any intention originally so that should be fine.

troyciesco (approved):

PR #26454, opened 2026-02-17
Removed welcomeEmails feature flag checks in the members repository

cmraible:
ref https://linear.app/ghost/issue/NY-971/post-ga-cleanup-remove-the-welcomeemails-labs-flag-from-admin
closes https://linear.app/ghost/issue/NY-965/post-ga-cleanup-remove-labsservice-from-memberrepository

There should be no behavioral changes in this commit, since the `welcomeEmails` flag has already been promoted to GA.

Now that welcome emails are in GA, we want to remove the flag completely. This commit removes any conditional behavior in Ghost/core's Member Repository that depends on the `welcomeEmails` flag.

Note: this also removes the labsService dependency from the MembersRepository entirely.

I've manually tested the following:
- Free welcome email disabled, free signup does not receive welcome email
- Free welcome email enabled, free signup does receive welcome email
- Paid welcome email disabled, paid signup does not receive welcome email
- Paid welcome email enabled, paid signup does receive welcome email

### DIFF
--- a/ghost/core/core/server/services/members/members-api/repositories/member-repository.js
+++ b/ghost/core/core/server/services/members/members-api/repositories/member-repository.js
@@ -359,8 +359,7 @@ module.exports = class MemberRepository {
         const memberAddOptions = {...(options || {}), withRelated};
         let member;
 
-        const welcomeEmailsEnabled = this._labsService.isSet('welcomeEmails');
-        if (welcomeEmailsEnabled && WELCOME_EMAIL_SOURCES.includes(source)) {
+        if (WELCOME_EMAIL_SOURCES.includes(source)) {
             const freeWelcomeEmail = this._AutomatedEmail ? await this._AutomatedEmail.findOne({slug: MEMBER_WELCOME_EMAIL_SLUGS.free}) : null;
             const isFreeWelcomeEmailActive = freeWelcomeEmail && freeWelcomeEmail.get('lexical') && freeWelcomeEmail.get('status') === 'active';
             const isFreeSignup = !stripeCustomer;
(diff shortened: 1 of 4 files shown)


cmraible:
The rabbit makes a good point here. Planning to address the [comment](https://github.com/TryGhost/Ghost/pull/26454#pullrequestreview-3816912949) before merging this one.

cmraible:
> isFreeWelcomeEmailActive can only ever be true if shouldCheckFreeWelcomeEmail was true, and that already requires isFreeSignup === true (line 360). The second && isFreeSignup at line 368 is therefore dead-code as written today.

Coderabbit's nit pick here is technically true, but I think I prefer to keep the `isFreeSignup` condition here just to be super explicit. Happy to be swayed otherwise though.

ghost/core/core/server/services/members/members-api/repositories/member-repository.js
```diff
@@ -359,39 +356,38 @@ module.exports = class MemberRepository {
...
+
+        if (shouldCheckFreeWelcomeEmail) {
             const freeWelcomeEmail = this._AutomatedEmail ? await this._AutomatedEmail.findOne({slug: MEMBER_WELCOME_EMAIL_SLUGS.free}) : null;
-            const isFreeWelcomeEmailActive = freeWelcomeEmail && freeWelcomeEmail.get('lexical') && freeWelcomeEmail.get('status') === 'active';
-            const isFreeSignup = !stripeCustomer;
+            isFreeWelcomeEmailActive = freeWelcomeEmail && freeWelcomeEmail.get('lexical') && freeWelcomeEmail.get('status') === 'active';
+        }
 
+        if (isFreeWelcomeEmailActive && isFreeSignup) {
             const runMemberCreation = async (transacting) => {
```
cmraible, line 369: This got a bit hairier than I expected for just removing a flag, but I'm pretty sure this is correct. Very open to suggestions/simplification

cmraible:
This should be the last one before removing the flag itself 😄

cmraible:
Sorry for the commits after requesting a review - I noticed there was some mocking still hanging around in the integration tests, then I rebased so I can stack the lab removal itself on top of this.

ghost/core/core/server/services/members/members-api/repositories/member-repository.js
troyciesco: yeah this file is always tricky. I think this is fine though, to me the logic is clear and easy to follow even if we could technically do it in fewer lines.

troyciesco (approved):

PR #26457, opened 2026-02-18
Added yarn dev:local command to run Ghost backend locally

cmraible:
In some circumstances, it can be more convenient to run Ghost's backend locally, rather than in Docker. In particular, when testing Ghost against local versions of dependencies that live in other monorepos with `yarn link`.

This leaves the default `yarn dev` as-is, but adds a new `yarn dev:local` command that runs:
- Docker services like mysql and redis
- Ghost's backend directly on the host
- Dev servers for all the apps

### DIFF
--- a/ghost/core/package.json
+++ b/ghost/core/package.json
@@ -22,6 +22,7 @@
   "scripts": {
     "archive": "npm pack",
     "dev": "node --watch --import=tsx index.js",
+    "dev:wait-mysql": "bash -lc 'until nc -z 127.0.0.1 \"${database__connection__port:-3306}\"; do echo \"Waiting for MySQL on 127.0.0.1:${database__connection__port:-3306}\"; sleep 1; done; yarn dev'",
     "build:assets": "yarn build:assets:css && yarn build:assets:js",
     "build:assets:js": "node bin/minify-assets.js",
     "build:assets:css": "postcss core/frontend/public/ghost.css --no-map --use cssnano -o core/frontend/public/ghost.min.css",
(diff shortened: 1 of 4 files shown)


ghost/core/package.json
```diff
@@ -22,6 +22,7 @@
   "scripts": {
     "archive": "npm pack",
     "dev": "node --watch --import=tsx index.js",
+    "dev:wait-mysql": "bash -lc 'until nc -z 127.0.0.1 \"${database__connection__port:-3306}\"; do echo \"Waiting for MySQL on 127.0.0.1:${database__connection__port:-3306}\"; sleep 1; done; yarn dev'",
```
cmraible, line 25: A bit weird, but it works. Since Ghost is running outside of Docker, we lose the benefit of the `depends_on` in the compose file, so Ghost tries to start before mysql is ready without this.

package.json
```diff
@@ -31,6 +31,7 @@
...
     "dev:storage": "DEV_COMPOSE_FILES='-f compose.dev.storage.yaml' nx run ghost-monorepo:docker:dev",
     "dev:all": "DEV_COMPOSE_FILES='-f compose.dev.analytics.yaml -f compose.dev.storage.yaml' nx run ghost-monorepo:docker:dev",
+    "dev:local": "NODE_TLS_REJECT_UNAUTHORIZED=0 database__client=mysql2 database__connection__host=127.0.0.1 database__connection__user=root database__connection__password=${MYSQL_ROOT_PASSWORD:-root} database__connection__database=${MYSQL_DATABASE:-ghost_dev} server__host=0.0.0.0 server__port=2369 mail__transport=SMTP mail__options__host=127.0.0.1 mail__options__port=1025 adapters__cache__Redis__host=127.0.0.1 adapters__cache__Redis__port=6379 portal__url=/ghost/assets/portal/portal.min.js comments__url=/ghost/assets/comments-ui/comments-ui.min.js sodoSearch__url=/ghost/assets/sodo-search/sodo-search.min.js sodoSearch__styles=/ghost/assets/sodo-search/main.css signupForm__url=/ghost/assets/signup-form/signup-form.min.js announcementBar__url=/ghost/assets/announcement-bar/announcement-bar.min.js DEV_COMPOSE_FILES=\"-f compose.dev.local.yaml ${DEV_LOCAL_COMPOSE_FILES:-}\" nx run ghost-monorepo:docker:dev:local",
```
cmraible, line 34: Pretty ugly...might be better to put some of the environment variables in a special .env file that we source here instead? Some of this is probably not entirely necessary anyways.

cmraible:
Closing this for now at least as I don't have time to polish and merge this. Will keep it around as a reference for future work in this area.

PR #26464, opened 2026-02-18
Added tests to verify CDN image size lookup uses HTTP path

vershwal:
ref https://linear.app/ghost/issue/HKG-1582/

- Ensures `ImageSize.getImageSizeFromUrl` routes local `/content/images` URLs to storage reads while CDN `/content/images` URLs use URL-based dimension lookup, with unit coverage and email-service integration checks.


vershwal:
@coderabbitai review

allouis (approved):

PR #26469, opened 2026-02-18
Simplify admin builds

jonatansberg:
Remove all redundant build steps, except for the admin-x-design-system. It still needs to be built separately in order to support continued usage of the SVGR Vite plugin.


jonatansberg:
Closing since stale.

PR #26478, opened 2026-02-18
Added CDN image parity for sizing and dimensions

vershwal:
ref https://linear.app/ghost/issue/HKG-1618/

Ghost is migrating image storage from local storage to GCS. Image resizing is now handled by the image proxy rather than local storage, but the rest of the codebase still needs to treat CDN-hosted images the same as local ones -- generating `/content/images/size/...` URLs, fetching dimensions correctly, and applying email optimizations. This PR updates the non-resize paths so CDN images behave identically to local images across rendering, serving, and metadata flows.

- Updated `mobiledoc.populateImageSizes` so CDN/non-local image cards use HTTP URL-based dimension fetching instead of storage-path reads, while preserving local image behavior (including legacy local path compatibility in subdir setups).
- Centralised internal image detection into `storageUtils.isInternalImage`, which checks both local storage paths and configured CDN base URLs (`urls:image`), so callers no longer duplicate CDN detection logic.
- Updated email image rendering and frontend `img_url` helper to use `storageUtils.isInternalImage` for generating `/content/images/size/...` variants, keeping local-like optimization behavior for CDN images while preventing accidental rewrites of third-party `/content/images/...` URLs.
- Added unit tests for all CDN paths above, including a double-rewrite prevention test for already-sized CDN URLs, plus additional non-editor CDN coverage for metadata image sizing, blog icon sizing, and oEmbed image storage flows.

### DIFF
--- a/ghost/core/core/frontend/utils/images.js
+++ b/ghost/core/core/frontend/utils/images.js
@@ -1,11 +1,14 @@
 const url = require('url');
 const imageTransform = require('@tryghost/image-transform');
+const config = require('../../shared/config');
 const urlUtils = require('../../shared/url-utils');
 
 module.exports.detectInternalImage = function detectInternalImage(requestedImageUrl) {
     const siteUrl = urlUtils.getSiteUrl();
+    const imageBaseUrl = (config.get('urls:image') || '').replace(/\/+$/, '');
     const isAbsoluteImage = /https?:\/\//.test(requestedImageUrl);
-    const isAbsoluteInternalImage = isAbsoluteImage && requestedImageUrl.startsWith(siteUrl);
+    const matchesImageBase = imageBaseUrl && requestedImageUrl.startsWith(imageBaseUrl + '/');
+    const isAbsoluteInternalImage = isAbsoluteImage && (requestedImageUrl.startsWith(siteUrl) || matchesImageBase);
 
     // CASE: imagePath is a "protocol relative" url e.g. "//www.gravatar.com/ava..."
     //       by resolving the the imagePath relative to the blog url, we can then
--- a/ghost/core/core/frontend/utils/images.js
+++ b/ghost/core/core/frontend/utils/images.js
@@ -1,14 +1,12 @@
 const url = require('url');
 const imageTransform = require('@tryghost/image-transform');
-const config = require('../../shared/config');
 const urlUtils = require('../../shared/url-utils');
+const storageUtils = require('../../server/adapters/storage/utils');
 
 module.exports.detectInternalImage = function detectInternalImage(requestedImageUrl) {
     const siteUrl = urlUtils.getSiteUrl();
-    const imageBaseUrl = (config.get('urls:image') || '').replace(/\/+$/, '');
     const isAbsoluteImage = /https?:\/\//.test(requestedImageUrl);
-    const matchesImageBase = imageBaseUrl && requestedImageUrl.startsWith(imageBaseUrl + '/');
-    const isAbsoluteInternalImage = isAbsoluteImage && (requestedImageUrl.startsWith(siteUrl) || matchesImageBase);
+    const isAbsoluteInternalImage = isAbsoluteImage && (requestedImageUrl.startsWith(siteUrl) || storageUtils.isInternalImage(requestedImageUrl));
 
     // CASE: imagePath is a "protocol relative" url e.g. "//www.gravatar.com/ava..."
     //       by resolving the the imagePath relative to the blog url, we can then
--- a/ghost/core/core/server/lib/mobiledoc.js
+++ b/ghost/core/core/server/lib/mobiledoc.js
@@ -3,6 +3,7 @@ const errors = require('@tryghost/errors');
 const logging = require('@tryghost/logging');
 const config = require('../../shared/config');
 const storage = require('../adapters/storage');
+const storageUtils = require('../adapters/storage/utils');
 
 let cardFactory;
 let cards;
@@ -132,8 +133,16 @@ module.exports = {
             }
 
             const isUnsplash = payload.src.match(/images\.unsplash\.com/);
+            const isRelativeImagePath = /^\/content\/images\//.test(payload.src);
             try {
-                const size = isUnsplash ? await getUnsplashSize(payload.src) : await imageSize.getOriginalImageSizeFromStorageUrl(payload.src);
+                let size;
+                if (isUnsplash) {
+                    size = await getUnsplashSize(payload.src);
+                } else if (isRelativeImagePath || storageUtils.isLocalImage(payload.src)) {
+                    size = await imageSize.getOriginalImageSizeFromStorageUrl(payload.src);
+                } else {
+                    size = await imageSize.getImageSizeFromUrl(payload.src);
+                }
 
                 if (size && size.width && size.height) {
                     payload.width = size.width;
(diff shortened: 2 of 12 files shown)


ghost/core/core/server/services/email-service/email-renderer.js
```diff
@@ -1423,10 +1425,11 @@ class EmailRenderer {
...
                 }
 
-                if (this.#storageUtils.isLocalImage(href)) {
+                if (this.#storageUtils.isLocalImage(href) || CONTENT_IMAGES_PATH_REGEX.test(href)) {
+                    const sizePath = 'size/w' + (visibleWidth * 2) + (visibleHeight ? 'h' + (visibleHeight * 2) : '') + '/';
                     // we can safely request a 1200px image - Ghost will serve the original if it's smaller
                     return {
-                        href: href.replace(/\/content\/images\//, '/content/images/size/w' + (visibleWidth * 2) + (visibleHeight ? 'h' + (visibleHeight * 2) : '') + '/'),
+                        href: href.replace(CONTENT_IMAGES_PATH_WITHOUT_SIZE_REGEX, '/content/images/' + sizePath),
                         width: size.width,
```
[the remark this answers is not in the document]
vershwal: Fixed it! Thanks

ghost/core/core/frontend/utils/images.js
```diff
@@ -1,11 +1,13 @@
...
 const imageTransform = require('@tryghost/image-transform');
+const config = require('../../shared/config');
 const urlUtils = require('../../shared/url-utils');
 
 module.exports.detectInternalImage = function detectInternalImage(requestedImageUrl) {
     const siteUrl = urlUtils.getSiteUrl();
+    const imageBaseUrl = (config.get('urls:image') || '').replace(/\/+$/, '');
     const isAbsoluteImage = /https?:\/\//.test(requestedImageUrl);
-    const isAbsoluteInternalImage = isAbsoluteImage && requestedImageUrl.startsWith(siteUrl);
+    const isAbsoluteInternalImage = isAbsoluteImage && (requestedImageUrl.startsWith(siteUrl) || (imageBaseUrl && requestedImageUrl.startsWith(imageBaseUrl)));
```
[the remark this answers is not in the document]
vershwal: Thanks, done

ghost/core/core/server/services/email-service/email-renderer.js
```diff
@@ -1449,6 +1452,33 @@ class EmailRenderer {
...
+            const imageBaseUrl = this.#urlUtils?.urlFor?.('image', {image: ''}, true);
+            if (imageBaseUrl && href.startsWith(imageBaseUrl)) {
+                return true;
+            }
+        } catch {
+            return false;
+        }
+
+        return false;
+    }
```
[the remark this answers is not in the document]
vershwal: Done!

ghost/core/core/server/lib/mobiledoc.js
```diff
@@ -132,8 +133,16 @@ module.exports = {
             }
 
             const isUnsplash = payload.src.match(/images\.unsplash\.com/);
+            const isRelativeImagePath = /^\/content\/images\//.test(payload.src);
             try {
-                const size = isUnsplash ? await getUnsplashSize(payload.src) : await imageSize.getOriginalImageSizeFromStorageUrl(payload.src);
+                let size;
+                if (isUnsplash) {
+                    size = await getUnsplashSize(payload.src);
+                } else if (isRelativeImagePath || storageUtils.isLocalImage(payload.src)) {
```
allouis, line 141: Is it possible that we have a relative image path when using the S3Adapter? I don't think so, right?

ghost/core/core/server/services/email-service/email-renderer.js
```diff
@@ -1449,6 +1456,32 @@ class EmailRenderer {
             height: null
         };
     }
+
+    #isGhostContentImageUrl(href) {
```
allouis: I'm not convinced this method belongs here 🤔 Could the storageUtils expose a method like `isInternalImage` or something?

Ideally whatever is calculating this doesn't know about `imageBaseUrl` at all I think - and would somehow ask the image storage adapter - "Hey, is this yours?" - maybe moving this to storage utils gets us closer to that?

ghost/core/core/frontend/utils/images.js
```diff
@@ -1,11 +1,14 @@
...
+const config = require('../../shared/config');
 const urlUtils = require('../../shared/url-utils');
 
 module.exports.detectInternalImage = function detectInternalImage(requestedImageUrl) {
     const siteUrl = urlUtils.getSiteUrl();
+    const imageBaseUrl = (config.get('urls:image') || '').replace(/\/+$/, '');
     const isAbsoluteImage = /https?:\/\//.test(requestedImageUrl);
-    const isAbsoluteInternalImage = isAbsoluteImage && requestedImageUrl.startsWith(siteUrl);
+    const matchesImageBase = imageBaseUrl && requestedImageUrl.startsWith(imageBaseUrl + '/');
+    const isAbsoluteInternalImage = isAbsoluteImage && (requestedImageUrl.startsWith(siteUrl) || matchesImageBase);
```
allouis: Similar to last commit - but it feels like we're doing the same/similar checks everywhere and we could consider centralising them?

allouis (commented):
Looks good, just a few comments on placement of checks

ghost/core/test/unit/server/adapters/storage/utils.test.js
```diff
@@ -161,4 +162,64 @@ describe('storage utils', function () {
...
+
+        it('should fall back to isLocalImage when no CDN config', function () {
+            urlForStub = sinon.stub(urlUtils, 'urlFor');
+            urlForStub.withArgs('home').returns('http://myblog.com/');
+            urlGetSubdirStub = sinon.stub(urlUtils, 'getSubdir');
+            urlGetSubdirStub.returns('');
+
+            assert.equal(storageUtils.isInternalImage('/content/images/2026/02/photo.png'), true);
+            assert.equal(storageUtils.isInternalImage('https://external.com/photo.png'), false);
+        });
```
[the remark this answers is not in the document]
vershwal: Addressed in commit https://github.com/TryGhost/Ghost/commit/fd729e678af8687b3728774045bdb5324b52119d

ghost/core/core/server/lib/mobiledoc.js
vershwal: You're right that new images uploaded via the S3 adapter won't have relative paths -- they'll always be absolute CDN URLs. But before this PR, everything that wasn't Unsplash was routed through `getOriginalImageSizeFromStorageUrl`, which only works for local/storage paths. The `isRelativeImagePath` check preserves that existing behavior for old content: if a site was using local storage before switching to S3, mobiledoc cards already saved in the database would still contain relative `/content/images/...` paths. Without this check, those old cards would fall through to `getImageSizeFromUrl` which would fail since relative paths aren't valid HTTP URLs.

ghost/core/core/frontend/utils/images.js
vershwal: Done

ghost/core/core/server/services/email-service/email-renderer.js
vershwal: Thanks for this. I think it’s great to move all the checks into one place. Added `isInternalImage` in https://github.com/TryGhost/Ghost/pull/26478/changes/fd1db2a91c7e4b000b59a9fa6885bd52920c1576

allouis (approved):

PR #26484, opened 2026-02-18
🐛 Fixed invisible comments filter dropdown in Safari

kevinansfield:
fixes https://linear.app/ghost/issue/BER-3153/

- Wrapped shade `PopoverContent` in `PopoverPrimitive.Portal` to fix filter popover positioning in Safari
- Without a Portal, Safari resolved the offset parent differently from Chrome, causing the popover's `translateX` to be off by exactly the sidebar width (300px), pushing it off-screen
- Matches the Portal pattern already used by `SelectContent` and `HoverCardContent` in shade, and the standard shadcn/ui popover implementation


weylandswart (approved):
Sorry for taking so long with the review. Tested and all looks good to me!

PR #26487, opened 2026-02-18
Optimized requests with identical get helpers

9larsons:
ref https://linear.app/ghost/issue/ONC-1431

#get helpers from the frontend are processed as individual API requests, although there may be duplicates. This change creates a per-request cache that prevents duplicative queries hitting the database, while allowing theme creators to continue using a simple syntax.

This is configured to be used with a new config setting: `optimization:getHelper:deduplication`.


ghost/core/core/frontend/helpers/get.js
```diff
@@ -303,6 +329,41 @@ async function makeAPICall(resource, controllerName, action, apiOptions) {
...
+    const rendered = options.fn(response, {
+        data: data,
+        blockParams: blockParams
+    });
+
+    if (response['@@ABORTED_GET_HELPER@@']) {
+        return new SafeString(`<span data-aborted-get-helper>Could not load content</span>` + rendered);
+    }
+    return rendered;
+}
```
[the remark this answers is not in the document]
9larsons: - renderResponse already clones before mutation in ghost/core/core/frontend/helpers/get.js:331.
  - It uses:
      - cloned top-level response object
      - cloned response[resource] array/data
      - then runs prepareContextResource on the clone, not the cached original.
  - This already prevents shared-mutable/double-wrapping issues for deduped cached responses, and is stricter than the requested shallow-clone approach.

ghost/core/core/frontend/helpers/get.js
```diff
@@ -46,6 +46,22 @@ const pathAliases = {
     'post.author': 'post.author.slug'
 };
 
+/**
```
9larsons, line 64: We could make a more robust cache string, however this is simple and fast and doesn't depend on maintaining a list of properties on the object.

ghost/core/core/frontend/helpers/get.js
```diff
@@ -400,3 +466,5 @@ module.exports.async = true;
 module.exports.optimiseFilterCacheability = optimiseFilterCacheability;
 
 module.exports.querySimplePath = querySimplePath;
+
+module.exports.generateCacheKey = generateCacheKey;
```
9larsons: Exported for testing.

ghost/core/core/frontend/helpers/get.js
```diff
@@ -340,37 +396,47 @@ module.exports = async function get(resource, options) {
     // Parse the options we're going to pass to the API
     apiOptions = parseOptions(ghostGlobals, this, apiOptions);
     apiOptions.context = {member: data.member};
-    try {
```
9larsons, line 343: This is now focused on the cache handling and whether or not we should make the request and most of the other handling got refactored into `renderResponse`.

ghost/core/core/frontend/helpers/get.js
EvanHahn: nit, take it or leave it: we should mention, in the code, that this is exported for testing. I like naming it something like `__generateCacheKeyForTests`, but a comment would work too.

```suggestion
module.exports.__generateCacheKeyForTests = generateCacheKey;
```

or

```suggestion
// Exported for tests. Shouldn't be used externally.
module.exports.generateCacheKey = generateCacheKey;
```

Alternatively, you could _not test this function at all_, because it's pretty well covered by higher level tests.

ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js
```diff
@@ -37,11 +38,15 @@ function updateLocalTemplateOptions(req, res, next) {
         status: req.member.status
     } : null;
 
+    // Only create cache object if deduplication feature is enabled
```
EvanHahn: nit: I think it's clear without this comment.

```suggestion
```

ghost/core/test/unit/frontend/helpers/get.test.js
```diff
@@ -614,4 +614,310 @@ describe('{{#get}} helper', function () {
             assert.deepEqual(args.posts, []);
         });
     });
+
+    describe('generateCacheKey', function () {
+        it('should produce deterministic keys for identical queries', function () {
+            const apiOptions1 = {filter: 'featured:true', limit: 5};
+            const apiOptions2 = {filter: 'featured:true', limit: 5};
+
+            assert.equal(generateCacheKey('posts', apiOptions1), generateCacheKey('posts', apiOptions2));
```
EvanHahn: praise: Love that this doesn't test the internals, just the stability of the cache key.

ghost/core/test/unit/frontend/helpers/get.test.js
```diff
@@ -614,4 +614,310 @@ describe('{{#get}} helper', function () {
...
+            const apiOptions1 = {filter: 'featured:true', collection: 'home'};
+            const apiOptions2 = {filter: 'featured:true', collection: 'archive'};
+
+            assert.notEqual(generateCacheKey('posts', apiOptions1), generateCacheKey('posts', apiOptions2));
+        });
+
+        it('should handle missing optional fields', function () {
+            const apiOptions = {filter: 'featured:true'};
+
+            // Should not throw
```
EvanHahn: nit: consider [`assert.doesNotThrow()`](https://nodejs.org/api/assert.html#assertdoesnotthrowfn-error-message).

ghost/core/test/unit/frontend/helpers/get.test.js
```diff
@@ -614,4 +614,310 @@ describe('{{#get}} helper', function () {
...
+        });
+
+        it('should produce different keys for different filters', function () {
+            const apiOptions1 = {filter: 'featured:true'};
+            const apiOptions2 = {filter: 'featured:false'};
+
+            assert.notEqual(generateCacheKey('posts', apiOptions1), generateCacheKey('posts', apiOptions2));
+        });
+
+        it('should include member uuid in key', function () {
```
EvanHahn: nit: `generateCacheKey` doesn't care about the shape of its second argument. Do we want these tests? Personally, I think we should remove them. Specifically:

- "should include member uuid in key"
- "should include additional query options to avoid collisions"
- "should handle missing optional fields"
- "should include all relevant query parameters"

ghost/core/core/frontend/helpers/get.js
```diff
@@ -340,37 +396,47 @@ module.exports = async function get(resource, options) {
...
-        const response = await makeAPICall(resource, controllerName, action, apiOptions);
 
-        // prepare data properties for use with handlebars
-        if (response[resource] && response[resource].length) {
-            response[resource].forEach(prepareContextResource);
+    // Per-request deduplication: check if we have a cached result for this query
+    const enableDeduplication = config.get('optimization:getHelper:deduplication');
```
EvanHahn: nit: It feels wrong to check the config twice—once here and once in `updateLocalTemplateOptions`. Instead, I think we should assume dedupe is on if `options.data._queryCache` is present.

ghost/core/core/frontend/helpers/get.js
```diff
@@ -46,6 +46,22 @@ const pathAliases = {
...
+ * @param {String} resource - The resource type (posts, tags, etc.)
+ * @param {Object} apiOptions - The API options (filter, limit, include, etc.)
+ * @returns {String} - Deterministic cache key
+ */
+function generateCacheKey(resource, apiOptions) {
+    const sortedOptions = Object.fromEntries(
+        Object.entries(apiOptions).sort(([a], [b]) => a.localeCompare(b))
+    );
+
+    return `${resource}|${JSON.stringify(sortedOptions)}`;
```
EvanHahn: question: Is there any risk of being unable to stringify an option here? Some example worries I have:

- `JSON.stringify({foo: undefined}) === '{}'`
- `JSON.stringify({fn: () => {}}) === '{}'`
- `JSON.stringify(123n)` → error
- `JSON.stringify([object with circular references])` → error

ghost/core/core/frontend/helpers/get.js
```diff
@@ -46,6 +46,22 @@ const pathAliases = {
     'post.author': 'post.author.slug'
 };
 
+/**
+ * Generate a deterministic cache key for a {{#get}} query.
+ * Sorts top-level option keys for deterministic serialization.
```
EvanHahn, line 66: question: Do we also want to sort lower level object keys? For example, these two objects would have different cache keys:

```javascript
{a: {b: 1, c: 2}}
{a: {c: 2, b: 1}}
```

But maybe that's fine.

ghost/core/core/frontend/helpers/get.js
```diff
@@ -46,6 +46,22 @@ const pathAliases = {
...
+ *
+ * @param {String} resource - The resource type (posts, tags, etc.)
+ * @param {Object} apiOptions - The API options (filter, limit, include, etc.)
+ * @returns {String} - Deterministic cache key
```
EvanHahn: Several nits here:

1. TypeScript currently treats `{String}` and `{string}` the same, but [that's changing](https://devblogs.microsoft.com/typescript/progress-on-typescript-7-december-2025/#javascript-checking-and-jsdoc-compatibility). This should be `{string}`.
2. It'd be nice to have a more exact type for `apiOptions`. (We can be fairly loose because this function doesn't care much about its internals.)
3. We should remove the dash in `@returns` because it creates a redundant line (and I think we should do the same for the rest for consistency). See screenshot:
   Screenshot

```suggestion
 * @param {string} resource The resource type (posts, tags, etc.)
 * @param {Readonly<Record<string, unknown>>} apiOptions The API options (filter, limit, include, etc.)
 * @returns {string} Deterministic cache key
```

ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js
```diff
@@ -37,11 +38,15 @@ function updateLocalTemplateOptions(req, res, next) {
...
+    const enableDeduplication = config.get('optimization:getHelper:deduplication');
+
     hbs.updateLocalTemplateOptions(res.locals, _.merge({}, localTemplateOptions, {
         data: {
             member: member,
             site: siteData,
-            custom: customData
+            custom: customData,
+            ...(enableDeduplication && {_queryCache: new Map()})
```
EvanHahn, line 48: - suggestion: rather than putting this on the `data` key, could we put it on the top level options? (And maybe remove the underscore from `_queryCache`?) The cache isn't really part of the data, so I think it belongs a level up.
- question: just to make sure I understand: this cache is only for the lifetime of this render, right? I don't want it to grow indefinitely.

ghost/core/core/frontend/helpers/get.js
```diff
@@ -340,37 +396,47 @@ module.exports = async function get(resource, options) {
...
+    let cacheKey;
+
+    if (queryCache) {
+        cacheKey = generateCacheKey(resource, apiOptions);
+
+        if (queryCache.has(cacheKey)) {
+            try {
+                // Await cached promise (handles both resolved and in-flight)
+                const cachedResponse = await queryCache.get(cacheKey);
+                returnedRowsCount = cachedResponse[resource] && cachedResponse[resource].length;
```
EvanHahn: nit: I believe this line does nothing.

```suggestion
```

ghost/core/core/frontend/helpers/get.js
```diff
@@ -340,37 +396,47 @@ module.exports = async function get(resource, options) {
...
+                return renderResponse(cachedResponse, resource, options, data);
```
EvanHahn: - issue: This early return prevents all the logging below.  Cached requests will probably be below the notify threshold, but I don't think we should assume that. 
- issue: If `cachedResponse` resolves but `renderResponse` fails, we will needlessly delete the cache key.

ghost/core/core/frontend/helpers/get.js
```diff
@@ -303,6 +319,46 @@ async function makeAPICall(resource, controllerName, action, apiOptions) {
...
 }
 
+/**
+ * Prepare and render the response from a {{#get}} query
+ *
+ * @param {Object} response - API response
+ * @param {String} resource - Resource type (posts, tags, etc.)
+ * @param {Object} options - Handlebars options
+ * @param {Object} data - Handlebars data frame
+ * @returns {String|SafeString} - Rendered template output
```
EvanHahn: nit: similar to the above, but we don't really know the types of those `Object`s so we can leave them vague for now.

```suggestion
 * @param {Object} response API response
 * @param {string} resource Resource type (posts, tags, etc.)
 * @param {Object} options Handlebars options
 * @param {Object} data Handlebars data frame
 * @returns {string | SafeString} Rendered template output
```

ghost/core/test/unit/frontend/helpers/get.test.js
```diff
@@ -614,4 +614,310 @@ describe('{{#get}} helper', function () {
...
+
+            // Should only make one API call
+            assert.equal(browseStub.callCount, 1);
+        });
+
+        it('should handle concurrent identical requests', async function () {
+            configUtils.set('optimization:getHelper:deduplication', true);
+            locals = {root: {_locals: {}}, _queryCache: new Map()};
+
+            // Make a slow browse stub
```
EvanHahn: thought: I'm a little worried this short timeout is brittle. Depending on the shuffle of the event loop, this could resolve before the second `get` is called. Maybe irrelevant, feel free to ignore.

EvanHahn (changes requested):
LGTM overall, but a number of nits.

ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js
9larsons: Correct, scoped only to the request. We definitely don't want it hanging around.

ghost/core/core/frontend/helpers/get.js
9larsons: I'm perfectly ok with skipping this. I looked into it more, and in general we should not have nested objects. I added some jsdoc to coarsely fill, although we don't know everything in `apiOptions` because themes can pass their own values. We just want to make sure we capture the most common use cases, e.g. "get 3 posts with this tag". 

The only nested object I'm aware of is the member uuid, which should never vary in the context of a single request.

ghost/core/core/frontend/helpers/get.js
9larsons: In general, *no*, but I've added a defensive catch to the stringify call. We'll now check for a cacheKey value (which'll be null if the catch triggers). Good to be safe.

ghost/core/core/frontend/helpers/get.js
9larsons: Agreed! Fixed.

ghost/core/core/frontend/helpers/get.js
9larsons: Currently yes, because of the early exit you mentioned below. Though if we still check the timing, we should include the rows.

ghost/core/core/frontend/helpers/get.js
9larsons: Oh, yep! Good catch. Fixed. I was mostly thinking about skipping the logging in the event of a cached response, but it feels more reasonable to keep these on the same path. Plus, it's entirely possible they take forever for some reason.

ghost/core/core/frontend/helpers/get.js
9larsons: Removed - I agree, I don't think they're necessary.

ghost/core/core/frontend/services/theme-engine/middleware/update-local-template-options.js
9larsons: While I agree it doesn't fit into `data`, that's what is scoped at the request level to pass to the get helpers.

ghost/core/test/unit/frontend/helpers/get.test.js
9larsons: Removed these tests.

ghost/core/test/unit/frontend/helpers/get.test.js
9larsons: Removed tests.

ghost/core/test/unit/frontend/helpers/get.test.js
9larsons: It seems stable, but I agree. CI wrecks any hope, too. I've moved this to a promise that we hold until after the get calls, which should be fairly bulletproof.

ghost/core/core/frontend/helpers/get.js
```diff
@@ -46,6 +46,42 @@ const pathAliases = {
     'post.author': 'post.author.slug'
 };
 
+/**
+ * @typedef {Object.<string, unknown>} GetHelperAPIOptions
```
EvanHahn, line 50: nit: I _think_ this should be a Record, but I might be wrong.

```suggestion
 * @typedef {Record<string, unknown>} GetHelperAPIOptions
```

ghost/core/core/frontend/helpers/get.js
EvanHahn: It still looks like we have an early return preventing the logging. Could you fix that?

EvanHahn (approved):
Two non-blocking comments. LGTM.

PR #26494, opened 2026-02-18
Fixed flaky "no OTC in email" test

EvanHahn:
ref https://github.com/TryGhost/Ghost/actions/runs/22154549172/job/64055168837

We had an assertion error like this (highlights mine):

<pre>Email HTML should not contain OTC. Found: "<b>860572</b>" near: "0.1:4015/members/?token=th0fWUDq5WKKteGjd3CSaLYwIC<b>860572</b>&action=signin"</pre>

If the token has six consecutive numbers, the test will fail.

To fix that, we only check the actual text, not HTML attributes.


cmraible (approved):
Nice find!

PR #26497, opened 2026-02-18
🐛 Fixed CORS errors for embedded signup forms after a custom domain is added

kevinansfield:
closes https://linear.app/ghost/issue/BER-3063/

## Problem

Embedded signup forms are used via copy/paste and include the currently configured URL in their data attributes. Any forms created whilst a site is on a hosted subdomain (e.g. `*.ghost.io`) stop working when the site switches over to custom domain due to a CORS error on the `/ghost/api/members/token-integrity/` endpoint which requires a cookie to function. The request failed CORS because we returned an `Access-Control-Allow-Origin: *` header on the Members API for which browsers refuse to make credentialed requests cross-origin.

## Summary

Replaced the blanket `cors({maxAge})` middleware on the members API with an origin-aware policy. Known origins (site URL, admin URL, localhost) get the specific origin reflected back with `Access-Control-Allow-Credentials: true` so credentialed requests like integrity-token work correctly. All other requests (unknown, missing, or invalid origins) still get `*` so the public API remains as accessible as before from any site embedding Ghost widgets.

### `access-control-allow-origin` behaviour

| Request origin | Before | After |
|---|---|---|
| Site URL (`config.url`) | `*` | Reflected origin + credentials |
| Admin URL (`config.admin.url`) | `*` | Reflected origin + credentials |
| Localhost | `*` | Reflected origin + credentials |
| Third-party site | `*` | `*` |
| No origin header | `*` | `*` |
| Invalid origin | `*` | `*` |

### DIFF
--- a/ghost/core/core/server/web/members/middleware/cors.js
+++ b/ghost/core/core/server/web/members/middleware/cors.js
@@ -1,0 +1,58 @@
+const {URL} = require('url');
+const cors = require('cors');
+const config = require('../../../../shared/config');
+
+const maxAge = config.get('caching:cors:maxAge');
+const ENABLE_CORS = {origin: true, maxAge, credentials: true};
+const WILDCARD_CORS = {origin: '*', maxAge};
+
+function isAllowedOrigin(origin) {
+    if (!origin || origin === 'null') {
+        return false;
+    }
+
+    try {
+        const originUrl = new URL(origin);
+
+        // allow localhost requests no matter the port
+        if (originUrl.hostname === 'localhost' || originUrl.hostname === '127.0.0.1') {
+            return true;
+        }
+
+        // allow the configured site host through on any protocol
+        const siteUrl = new URL(config.get('url'));
+        if (originUrl.host === siteUrl.host) {
+            return true;
+        }
+
+        // allow the configured admin host through on any protocol
+        if (config.get('admin:url')) {
+            const adminUrl = new URL(config.get('admin:url'));
+            if (originUrl.host === adminUrl.host) {
+                return true;
+            }
+        }
+    } catch (e) {
+        // unparsable origin
+        return false;
+    }
+
+    return false;
+}
+
+/**
+ * @param {import('express').Request} req
+ * @param {Function} callback
+ */
+function corsOptionsDelegate(req, callback) {
+    const origin = req.get('origin');
+
+    if (isAllowedOrigin(origin)) {
+        return callback(null, ENABLE_CORS);
+    } else {
+        return callback(null, WILDCARD_CORS);
+    }
+}
+
+module.exports = cors(corsOptionsDelegate);
+
--- a/ghost/core/core/server/web/members/app.js
+++ b/ghost/core/core/server/web/members/app.js
@@ -1,5 +1,4 @@
 const debug = require('@tryghost/debug')('members');
-const cors = require('cors');
 const bodyParser = require('body-parser');
 const express = require('../../../shared/express');
 const sentry = require('../../../shared/sentry');
@@ -15,6 +14,7 @@ const api = require('../../api').endpoints;
 
 const commentRouter = require('../comments');
 const announcementRouter = require('../announcement');
+const corsMiddleware = require('./middleware/cors');
 
 /**
  * @returns {import('express').Application}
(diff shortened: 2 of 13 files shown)


kevinansfield:
## Context: CORS middleware comparison

We now have three `cors.js` files handling different parts of the API. Here's how they relate:

### Commonalities

All three files:
- Use the `cors` npm package with a `corsOptionsDelegate` pattern
- Check the `Origin` header and short-circuit when missing/`"null"`
- Allow `localhost`/`127.0.0.1` origins
- Allow the configured site URL host
- Allow the configured `admin:url` host (if set)
- Read `maxAge` from `config.get('caching:cors:maxAge')`
- All get `Vary: Origin` added by the `cors` library when the origin is dynamically determined

### Differences

| Aspect | **Frontend** (`frontend/web`) | **Admin API** (`server/web/api`) | **Members API** (`server/web/members`) |
|---|---|---|---|
| **Default (no origin)** | `origin: false` (deny) | `origin: false` (deny) | `origin: '*'` (wildcard) |
| **Unrecognized origin** | `origin: false` (deny) | `origin: false` (deny) | `origin: '*'` (wildcard) |
| **Credentials** | `credentials: true` | Not set | `credentials: true` (for known origins) |
| **Localhost check** | `hostname` match via `new URL()` | `hostname` match via `os.networkInterfaces()` + allowlist | `hostname` match via `new URL()` |
| **URL parsing** | `new URL()` (WHATWG) | `url.parse()` (legacy Node) | `new URL()` (WHATWG) |
| **Error handling** | Throws `BadRequestError` on invalid origin | No try/catch (would crash on bad URL) | Catches errors, falls back to wildcard |
| **Extra Vary handling** | Adds explicit `Vary: Origin` on OPTIONS via `handleCaching` | Adds explicit `Vary: Origin` on OPTIONS via `handleCaching` | Relies solely on `cors` library's automatic Vary |
| **Local IP detection** | Only `localhost` + `127.0.0.1` | Enumerates all network interfaces (`os.networkInterfaces()`) | Only `localhost` + `127.0.0.1` |
| **Export shape** | `[handleCaching, cors()]` (array) | `[handleCaching, cors()]` (array) | `cors()` (single middleware) |

### Key observations

1. **The Admin API file is the oldest-style** — it uses the deprecated `url.parse()`, enumerates all local IPs via `os.networkInterfaces()`, and uses a cached allowlist. The other two use the modern WHATWG `URL` constructor.

2. **The Members API is the most permissive** — unrecognized or missing origins get `origin: '*'` (wildcard CORS), meaning any website can make requests to the Members API. This makes sense because the Members API is called from arbitrary customer sites embedding signup forms, portal, comments, etc. However, `credentials: true` is only sent for *known* origins (site/admin/localhost).

3. **The Frontend file is the only one that returns an error** on a malformed origin (`BadRequestError`). The Members file silently falls back to wildcard, and the Admin API file has no protection against malformed URLs.

4. **There's meaningful duplication** — the origin-checking logic (localhost, siteUrl, adminUrl) is essentially the same in the Frontend and Members files, while the Admin API does the same thing via a different mechanism (allowlist array with `indexOf`). This could potentially be extracted into a shared helper.

ghost/core/core/server/web/members/middleware/cors.js
```diff
@@ -0,0 +1,57 @@
+const {URL} = require('url');
+const cors = require('cors');
+const config = require('../../../../shared/config');
+
+const maxAge = config.get('caching:cors:maxAge');
+const ENABLE_CORS = {origin: true, maxAge, credentials: true};
+const WILDCARD_CORS = {origin: '*', maxAge};
+
+function isAllowedOrigin(origin) {
```
allouis, line 9: If `isAllowedOrigin` handled the `try/catch` for URL construction internally and also handled `null`/missing origins (returning `false`), then the `corsOptionsDelegate` below could collapse into a single `if` statement, something like:
 
```js
function corsOptionsDelegate(req, callback) {
    const origin = req.get('origin');
    if (isAllowedOrigin(origin)) {
        return callback(null, ENABLE_CORS);
    }
    return callback(null, WILDCARD_CORS);
}
```

Not a big deal tho!

allouis (approved):

PR #26503, opened 2026-02-19
Fixed rounding for percent-off offer prices

sagzy:
closes https://linear.app/ghost/issue/BER-3344

- Added rounding for percent-off prices, so that the displayed price matches the amount being billed by Stripe. Example: 20% off $5.99 renders $4.79 (not $4.792)
- Cleaned up duplicate code

### DIFF
--- a/apps/portal/src/utils/helpers.js
+++ b/apps/portal/src/utils/helpers.js
@@ -808,17 +808,18 @@ export const getOfferOffAmount = ({offer}) => {
 };
 
 export const getUpdatedOfferPrice = ({offer, price, useFormatted = false}) => {
-    const originalAmount = price.amount;
-    let updatedAmount;
+    const originalAmountInCents = price.amount;
+    let updatedAmountInCents = originalAmountInCents;
 
     if (offer.type === 'fixed' && isSameCurrency(offer.currency, price.currency)) {
-        updatedAmount = ((originalAmount - offer.amount)) / 100;
-        updatedAmount = updatedAmount > 0 ? updatedAmount : 0;
+        updatedAmountInCents = Math.max(0, (originalAmountInCents - offer.amount));
     } else if (offer.type === 'percent') {
-        updatedAmount = (originalAmount - ((originalAmount * offer.amount) / 100)) / 100;
-    } else {
-        updatedAmount = originalAmount / 100;
+        const discountInCents = Math.round(originalAmountInCents * (offer.amount / 100));
+        updatedAmountInCents = Math.max(0, (originalAmountInCents - discountInCents));
     }
+
+    const updatedAmount = updatedAmountInCents / 100;
+
     if (useFormatted) {
         return Intl.NumberFormat('en', {currency: price?.currency, style: 'currency'}).format(updatedAmount);
     }
--- a/apps/portal/package.json
+++ b/apps/portal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/portal",
-  "version": "2.64.6",
+  "version": "2.64.7",
   "license": "MIT",
   "repository": "https://github.com/TryGhost/Ghost",
   "author": "Ghost Foundation",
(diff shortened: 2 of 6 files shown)


apps/portal/src/utils/helpers.js
```diff
@@ -815,7 +815,7 @@ export const getUpdatedOfferPrice = ({offer, price, useFormatted = false}) => {
         updatedAmount = ((originalAmount - offer.amount)) / 100;
         updatedAmount = updatedAmount > 0 ? updatedAmount : 0;
     } else if (offer.type === 'percent') {
-        updatedAmount = (originalAmount - ((originalAmount * offer.amount) / 100)) / 100;
+        updatedAmount = Math.round(originalAmount - ((originalAmount * offer.amount) / 100)) / 100;
```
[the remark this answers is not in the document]
sagzy: valid, updated!

apps/portal/src/utils/helpers.js
```diff
@@ -808,17 +808,18 @@ export const getOfferOffAmount = ({offer}) => {
...
     if (offer.type === 'fixed' && isSameCurrency(offer.currency, price.currency)) {
-        updatedAmount = ((originalAmount - offer.amount)) / 100;
-        updatedAmount = updatedAmount > 0 ? updatedAmount : 0;
+        updatedAmountInCents = Math.max(0, (originalAmountInCents - offer.amount));
     } else if (offer.type === 'percent') {
-        updatedAmount = (originalAmount - ((originalAmount * offer.amount) / 100)) / 100;
-    } else {
-        updatedAmount = originalAmount / 100;
+        const discountInCents = Math.round(originalAmountInCents * (offer.amount / 100));
+        updatedAmountInCents = Math.max(0, (originalAmountInCents - discountInCents));
```
mike182uk, line 818: The codex agent commented on this, but Im not sure if you've updated since? I think it was suggesting:

```
updatedAmountInCents = Math.round(originalAmountInCents * (100 - offer.amount) / 100);
```

to match the backend?

(no need for `Math.max` as `offer.amount` is always a percentage, so result should be non-negative?)

mike182uk: should we add a test that would catch this drift?

sagzy: Hehe actually that was the original change, that I fixed it after the bot commented.

Looking at how Stripe handles this rounding edge case: they apply the rounding to the discount amount, not original amount - discount.

--- 

Here's a repro of the edge case with Stripe:
CleanShot 2026-02-19 at 14 26 38@2x

In this example,
- `originalAmountInCents` is 101 ($1.01)
- `offer.amount` is 50 (50%)


With this logic:
```
updatedAmountInCents = Math.round(originalAmountInCents * (100 - offer.amount) / 100);
```

The final amount would be `Math.round(101 * (100 - 50) / 100)` = `Math.round(101 * 0.5)` = `Math.round(50.5)`= `51` ($0.51)



Versus with this logic:
```
        const discountInCents = Math.round(originalAmountInCents * (offer.amount / 100));
        updatedAmountInCents = originalAmountInCents - discountInCents;
```

The final amount would be `101 - Math.round(101 * 50 / 100)` = `101 - Math.round(50.5)` = `101 - 51` = `50` ($0.50)
 
 Which is what Stripe is charging to the member!

sagzy: The next payment calculator on the backend side has the correct logic already: https://github.com/TryGhost/Ghost/blob/b1670182a0d864e1edfaf2926e5097282dfa21ea/ghost/core/core/server/services/members/members-api/services/next-payment-calculator.js#L160-L171 

So that is good

Where did you see the other logic on the backend side?

sagzy: Agree the `Math.max(0, ...)` is unnecessary for values coming from the backend, as they're validated to be between 1 and 100. The only use case for it is Portal previews, that renders whatever value the user is entering in the input field - having a Math.max(0, ...) avoids printing weird negative values in the preview

sagzy: Going to add a unit test with that edge case, and will remove Math.max from the next payment calculator, as there isn't any value in having it there (keeping it in Portal though, for Portal previews)

mike182uk: Nice one, thanks for the detailed write up. I think you've covered it, it was the [member repository](https://github.com/TryGhost/Ghost/blob/18d19b4124ab54f507b977c82d8424dbeba20fee/ghost/core/core/server/services/members/members-api/repositories/member-repository.js#L212) that had a similar formula, but this is being used for MRR so not related to this. Sorry for the confusion!

PR #26504, opened 2026-02-19
Updated members filtering by retention offer 

mike182uk:
closes https://linear.app/ghost/issue/BER-3349

This change ensures that only 1 retention offer per cadence is visible in the offer select when filtering members by retention. This also updates the retention modal to link the members list pre-filtered when clicking on the redemption count

### DIFF
--- a/apps/admin-x-settings/src/components/settings/growth/offers/offers-retention.tsx
+++ b/apps/admin-x-settings/src/components/settings/growth/offers/offers-retention.tsx
@@ -120,7 +120,7 @@ const OffersRetention: React.FC = () => {
     const {data: {offers: allOffers = []} = {}} = useBrowseOffers();
     const retentionOffers = getRetentionOffers(allOffers);
 
-    const handleRetentionOfferClick = (id: string) => {
+    const handleOfferEdit = (id: string) => {
         updateRoute(`offers/edit/retention/${id}`);
     };
 
@@ -142,13 +142,13 @@ const OffersRetention: React.FC = () => {
                     return (
                         <tr key={offer.id} className='group relative scale-100 border-b border-b-grey-200 dark:border-grey-800' data-testid='retention-offer-item'>
                             <td className='p-0'>
-                                <a className='block cursor-pointer p-5 pl-0' onClick={() => handleRetentionOfferClick(offer.id)}>
+                                <a className='block cursor-pointer p-5 pl-0' onClick={() => handleOfferEdit(offer.id)}>
                                     <span className='font-semibold'>{offer.name}</span><br />
                                     <span className='text-sm text-grey-700'>{offer.description}</span>
                                 </a>
                             </td>
                             <td className='whitespace-nowrap p-0 text-sm'>
-                                <a className='block cursor-pointer p-5' onClick={() => handleRetentionOfferClick(offer.id)}>
+                                <a className='block cursor-pointer p-5' onClick={() => handleOfferEdit(offer.id)}>
                                     {offer.terms ? (
                                         <>
                                             <span className='text-[1.3rem] font-medium uppercase'>{offer.terms}</span><br />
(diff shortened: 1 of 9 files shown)


apps/admin-x-settings/src/components/settings/growth/offers/offers-retention.tsx
```diff
@@ -120,43 +134,54 @@ const OffersRetention: React.FC = () => {
...
+                                        <span className='text-grey-700'>&ndash;</span>
+                                    )}
+                                </a>
+                            </td>
+                            <td className='whitespace-nowrap p-0 text-sm'>
+                                <a
+                                    className={`block cursor-pointer p-5 ${redemptionFilterUrl ? 'hover:underline' : ''}`}
+                                    data-testid={`retention-redemptions-link-${offer.id}`}
+                                    href={redemptionFilterUrl}
+                                    onClick={!redemptionFilterUrl ? () => handleRetentionOfferClick(offer.id) : undefined}
```
sagzy: nitpick: This is slightly hard to parse. What about renaming `handleRetentionOfferClick(offer.id)` to what it does, e.g. `editRetentionOffer(offer.id)`?

ghost/admin/app/components/offers/segment-select.js
```diff
@@ -38,31 +51,108 @@ export default class OffersSegmentSelect extends Component {
...
     }
 
+    getOfferById(id) {
+        return this.offers.find((offer) => {
+            return offer.id === id;
+        });
+    }
+
+    getOfferRedemptionType(offer) {
+        return offer.redemptionType || offer.redemption_type || 'signup';
```
sagzy: Question: is this necessary? I'd expect the Offer model to give us a `offer.redemptionType`, with a default value fallback

https://github.com/TryGhost/Ghost/blob/5d77c88f0b191eeb7528825a28cf902f28271843/ghost/admin/app/models/offer.js#L10

ghost/admin/app/components/offers/segment-select.js
```diff
@@ -38,31 +51,108 @@ export default class OffersSegmentSelect extends Component {
...
+            }
+        });
+
+        let ids = Array.from(offerIds).map((id) => {
+            const offer = this.getOfferById(id);
             return {
-                id: offer.id,
-                name: offer.name
+                id,
+                name: offer?.name || id
```
sagzy: Question: can offer be null/falsy here? Do we need a fallback from missing name?

ghost/admin/app/components/offers/segment-select.js
```diff
@@ -71,21 +161,31 @@ export default class OffersSegmentSelect extends Component {
...
             offers.forEach((offer) => {
+                if (this.getOfferRedemptionType(offer) === 'retention') {
+                    return;
+                }
+
                 offersGroup.options.push({
                     name: offer.name,
                     id: offer.id,
-                    count: offers.count?.members,
+                    count: offer.count?.members,
```
sagzy: nitpick: this looks like dead code (returned undefined before, returns undefined now). There is an inline comment saying `TODO: add include: 'count.members' to query once API supports` above lol. Let's delete it? 🧹

sagzy (approved):
Left a few nitpick / questions, but nothing blocking merge 👍

apps/admin-x-settings/src/components/settings/growth/offers/offers-retention.tsx
mike182uk: I've renamed `handleOfferEdit` as the `handle` prefix is the pattern used everywhere else 👍

ghost/admin/app/components/offers/segment-select.js
mike182uk: Yeh, this was being over defensive, removed

ghost/admin/app/components/offers/segment-select.js
mike182uk: lol no, removed

ghost/admin/app/components/offers/segment-select.js
mike182uk: Yeh cool, i've cleaned this up 👍

PR #26510, opened 2026-02-19
🐛 Fixed editor alerts positioning

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/26125
closes https://linear.app/ghost/issue/BER-3271/

- Alerts rendered into `#ember-alerts-wormhole` (body grid row 1) did not push editor content down and appeared behind the editor header because `.gh-editor-fullscreen-container` used `position: fixed`, covering the entire viewport and ignoring the grid layout
- Changed the editor fullscreen container from `position: fixed` to `position: absolute` and established a flex height chain so `.gh-koenig-editor` remains the sole scroll container for editor content
- Simplified the layout so each page context has one scroll container: SidebarInset for non-editor pages, `.gh-koenig-editor` for editor content
- Fixed the post settings sidebar to use flex column layout with a pinned header and independently scrollable content area to resolve issues on mobile
- Moved Ember CSS overrides (`.gh-app` position, `.gh-main` overflow) from the React shell into Ember source styles directly, removing redundant overrides
- Fixed mobile editor layout to fill the full viewport when the mobile navbar is hidden

### DIFF
--- a/apps/admin/src/index.css
+++ b/apps/admin/src/index.css
@@ -40,18 +40,6 @@ body.react-admin #ember-app {
     height: 100%;
 }
 
-/* Override Ember's fixed positioning to work within our flex layout */
-body.react-admin #ember-app .gh-app {
-    position: static;
-    width: 100%;
-    height: 100%;
-}
-
-/* Remove Ember's overflow since we handle scrolling at the SidebarInset level */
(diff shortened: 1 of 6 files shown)


weylandswart (approved):
Clicked around a bit more after our Tuple and everything looks solid.

PR #26516, opened 2026-02-19
Added Stripe service and webhook e2e test

9larsons:
no ref
- added mock Stripe server
- added webhook builders, client
- added Stripe fixtures
- added first Stripe webhook test to /e2e/

We'd like to build out Stripe mocks in the /e2e/ package in order to move all existing browser tests into our new /e2e/ testing suite. This is the first of several slices in an attempt to break up all of the pieces of Stripe functionality we need.

The first test here is creating a Ghost member on receiving the checkout webhook, which is effectively the back end of the Portal workflow where Portal -> Stripe -> Ghost. 

The Stripe server is stateful, unlike the other options out there for Stripe mocking/testing, meaning we are stuck rolling our own.

### DIFF
--- a/e2e/helpers/environment/service-managers/mysql-manager.ts
+++ b/e2e/helpers/environment/service-managers/mysql-manager.ts
@@ -27,12 +27,20 @@ export class MySQLManager {
         this.containerName = containerName;
     }
 
-    async setupTestDatabase(databaseName: string, siteUuid: string): Promise<void> {
+    async setupTestDatabase(databaseName: string, siteUuid: string, options: {
+        stripe?: {
+            secretKey: string;
+            publishableKey: string;
+        };
+    } = {}): Promise<void> {
         debug('Setting up test database:', databaseName);
         try {
             await this.createDatabase(databaseName);
             await this.restoreDatabaseFromSnapshot(databaseName);
             await this.updateSiteUuid(databaseName, siteUuid);
+            if (options.stripe) {
+                await this.updateStripeSettings(databaseName, options.stripe.secretKey, options.stripe.publishableKey);
+            }
 
             debug('Test database setup completed:', databaseName, 'with site_uuid:', siteUuid);
         } catch (error) {
@@ -165,6 +173,26 @@ export class MySQLManager {
         await this.exec(command);
 
         debug('site_uuid updated in database settings:', siteUuid);
+    }
+
+    async updateStripeSettings(database: string, secretKey: string, publishableKey: string): Promise<void> {
+        debug('Updating Stripe settings in database:', database);
+
+        const escapedSecretKey = secretKey.replace(/'/g, '\\\'');
+        const escapedPublishableKey = publishableKey.replace(/'/g, '\\\'');
+
+        // Use INSERT ... ON DUPLICATE KEY UPDATE so this works whether or not
+        // the settings rows already exist. In dev mode the base DB is empty
+        // (only schema, no seeded rows), so a plain UPDATE would be a no-op.
+        const command = 'mysql -uroot -proot -e "INSERT INTO \\`' + database + '\\`.settings ' +
+            '(id, \\`group\\`, \\`key\\`, value, type, flags, created_at, updated_at) VALUES ' +
+            '(SUBSTRING(REPLACE(UUID(), \'-\', \'\'), 1, 24), \'members\', \'stripe_secret_key\', \'' + escapedSecretKey + '\', \'string\', NULL, NOW(), NOW()), ' +
+            '(SUBSTRING(REPLACE(UUID(), \'-\', \'\'), 1, 24), \'members\', \'stripe_publishable_key\', \'' + escapedPublishableKey + '\', \'string\', NULL, NOW(), NOW()) ' +
+            'ON DUPLICATE KEY UPDATE value = VALUES(value), updated_at = NOW();"';
+
+        await this.exec(command);
+
+        debug('Stripe settings updated in database');
     }
 
     private async exec(command: string) {
--- a/e2e/helpers/playwright/fixture.ts
+++ b/e2e/helpers/playwright/fixture.ts
@@ -1,5 +1,6 @@
 import baseDebug from '@tryghost/debug';
 import {Browser, BrowserContext, Page, TestInfo, test as base} from '@playwright/test';
+import {FakeStripeServer, StripeTestService, WebhookClient} from '@/helpers/services/stripe';
 import {GhostInstance, getEnvironmentManager} from '@/helpers/environment';
 import {SettingsService} from '@/helpers/services/settings/settings-service';
 import {faker} from '@faker-js/faker';
@@ -7,6 +8,10 @@ import {loginToGetAuthenticatedSession} from '@/helpers/playwright/flows/sign-in
 import {setupUser} from '@/helpers/utils';
 
 const debug = baseDebug('e2e:ghost-fixture');
+const STRIPE_FAKE_SERVER_PORT = 40000 + parseInt(process.env.TEST_PARALLEL_INDEX || '0', 10);
+const STRIPE_SECRET_KEY = 'sk_test_e2eTestKey';
+const STRIPE_PUBLISHABLE_KEY = 'pk_test_e2eTestKey';
+
 export interface User {
     name: string;
     email: string;
@@ -23,7 +28,9 @@ export interface GhostInstanceFixture {
     ghostInstance: GhostInstance;
     labs?: Record<string, boolean>;
     config?: GhostConfig;
-    stripeConnected?: boolean;
+    stripeEnabled?: boolean;
+    stripeServer?: FakeStripeServer;
+    stripe?: StripeTestService;
     ghostAccountOwner: User;
     pageWithAuthenticatedUser: {
         page: Page;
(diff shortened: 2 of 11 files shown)


e2e/helpers/playwright/fixture.ts
```diff
@@ -60,20 +64,26 @@ async function setupNewAuthenticatedPage(browser: Browser, baseURL: string, ghos
...
-    ghostInstance: async ({config}, use, testInfo: TestInfo) => {
+    // Each test gets its own Ghost instance with isolated database.
+    ghostInstance: async ({config, stripeEnabled}, use, testInfo: TestInfo) => {
         debug('Setting up Ghost instance for test:', testInfo.title);
+        const stripeConfig = stripeEnabled ? {
+            STRIPE_API_HOST: 'host.docker.internal',
+            STRIPE_API_PORT: String(STRIPE_MOCK_PORT),
+            STRIPE_API_PROTOCOL: 'http'
+        } : {};
+        const mergedConfig = {...(config || {}), ...stripeConfig};
```
EvanHahn, line 101: thought: Maybe `stripeEnabled` should be part of the config? Not sure.

e2e/helpers/services/stripe/mock-stripe-server.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    private handleRequest(req: http.IncomingMessage, res: http.ServerResponse): void {
+        const url = req.url ?? '';
+        const method = req.method ?? '';
+
+        debug(`${method} ${url}`);
+
+        // GET /v1/customers/:id — returns customer with embedded subscriptions
+        const customerMatch = url.match(/^\/v1\/customers\/([^/?]+)/);
```
EvanHahn: nit: This custom routing feels like an opportunity for a framework like Express or Fastify (or a lightweight thing like [router]).

[router]: https://github.com/stripe/stripe-mock

e2e/helpers/services/stripe/mock-stripe-server.ts
```diff
@@ -0,0 +1,188 @@
+import baseDebug from '@tryghost/debug';
+import http from 'http';
+import type {StripeCustomer, StripePaymentMethod, StripeSubscription} from './builders';
+
+const debug = baseDebug('e2e:mock-stripe');
+
+export class MockStripeServer {
```
EvanHahn: question: Did you look at [stripe-mock], Stripe's official mock server? As a reference, see [how they set it up in their Node SDK](https://github.com/stripe/stripe-node/blob/47cdb5eb972d00bd682aa9078e0cfdc1049f7226/test/testUtils.ts#L69-L98).

[stripe-mock]: https://github.com/stripe/stripe-mock

e2e/helpers/services/stripe/builders.ts
```diff
@@ -0,0 +1,182 @@
+import crypto from 'crypto';
```
cmraible, line 1: suggestion: this file feels like it reinvents our Factory concept, which I think we should reuse here - e.g. `StripeCustomerFactory`, `StripePriceFactory`, etc.

e2e/helpers/services/stripe/mock-stripe-server.ts
cmraible: +1, I'd bias toward Express since that's what Ghost already uses

e2e/helpers/services/stripe/mock-stripe-server.ts
cmraible: nittiest of nits: I think this would technically be a `FakeStripeServer` https://martinfowler.com/bliki/TestDouble.html

e2e/helpers/services/stripe/webhook-client.ts
```diff
@@ -0,0 +1,34 @@
+import crypto from 'crypto';
```
cmraible, line 1: nit: maybe I'm thinking about this backwards, but in my mind Ghost is the webhook client, and this would be a webhook server

e2e/tests/admin/members/stripe-webhooks.test.ts
```diff
@@ -0,0 +1,17 @@
...
+    // Ghost creates a member from Stripe checkout webhooks; this tests the post-Portal redirect flow
+    test('member created via webhooks - has paid status', async ({stripe, page}) => {
+        const membersService = new MembersService(page.request);
+        const email = `stripe-test-${Date.now()}@example.com`;
+
+        await stripe!.createPaidMemberViaWebhooks({email, name: 'Test User'});
+
+        const member = await membersService.getByEmail(email);
+        expect(member.status).toBe('paid');
+    });
```
cmraible, line 16: I'm curious what this would look like for a truly e2e test, where a member signs up via portal. Do we create a mock Stripe Checkout frontend - probably not? But I think we'll need to mirror the redirect behavior of the Stripe checkout.

This also raises questions around how we handle different scenarios - e.g. what if the user aborts stripe checkout, or what if the payment fails. I realize this is just a first iteration and you probably haven't gotten to this point yet, just thinking out loud 😄

e2e/helpers/playwright/fixture.ts
```diff
@@ -86,6 +96,24 @@ export const test = base.extend<GhostInstanceFixture>({
...
+        await server.start();
+        debug('Mock Stripe server started on port', STRIPE_MOCK_PORT);
+
+        const webhookClient = new WebhookClient(baseURL);
+        const service = new StripeTestService(server, webhookClient);
+        await use(service);
+
+        await server.stop();
+        debug('Mock Stripe server stopped');
+    },
```
cmraible: I'm not sure, but this may need to come after the baseURL fixture below, which is where it gets set properly for each Ghost instance.

e2e/helpers/services/stripe/mock-stripe-server.ts
```diff
@@ -0,0 +1,188 @@
...
+            const timer = setTimeout(() => {
+                reject(new Error('Timed out waiting for Ghost to configure billing portal'));
+            }, timeoutMs);
+
+            this.billingPortalWaiters.push(() => {
+                clearTimeout(timer);
+                resolve();
+            });
+        });
+    }
```
cmraible: I'm rather confused what this is actually doing, but that may be a me problem. I'd have thought we could just get away with seeding the settings table with the (fake) stripe keys, but I'm not completely familiar with the actual Stripe Connection setup.

ghost/core/core/server/services/stripe/stripe-api.js
```diff
@@ -125,9 +125,19 @@ module.exports = class StripeAPI {
...
+        };
+        if (process.env.STRIPE_API_HOST) {
+            stripeConfig.host = process.env.STRIPE_API_HOST;
+        }
+        if (process.env.STRIPE_API_PORT) {
+            stripeConfig.port = parseInt(process.env.STRIPE_API_PORT, 10);
+        }
+        if (process.env.STRIPE_API_PROTOCOL) {
+            stripeConfig.protocol = process.env.STRIPE_API_PROTOCOL;
+        }
```
cmraible: suggestion: using `process.env` directly here feels like an anti-pattern, could we use `config.get()` instead?

cmraible:
One higher level thought/suggestion: this creates a new instance of the stripe server for each instance of Ghost, so we can pass the Ghost URL to the stripe server for webhooks. 

Did you consider creating a single persistent instance that would live for the lifetime of the test run? Maybe not worth the additional complexity, but we could conceivably handle the webhook URL mapping within the StripeTestService via the Stripe API keys or something along those lines.

No strong opinions on which way is better, just curious if you explored this at all

e2e/helpers/services/stripe/mock-stripe-server.ts
9larsons: Damnit, you're right. It started that way and I just don't like the name fake. It felt cheap. 😂

ghost/core/core/server/services/stripe/stripe-api.js
9larsons: Yes, good catch!

e2e/helpers/services/stripe/webhook-client.ts
9larsons: Since it's POSTing to Ghost, it feels reasonable that it's a client.

e2e/helpers/services/stripe/mock-stripe-server.ts
9larsons: 100% agree. It also cleans up a lot of extra lines. Done.

e2e/helpers/services/stripe/mock-stripe-server.ts
9larsons: I did try to do this a few times and previous implementations made this too difficult. I took another swing with this cleaner structure and it looks to work, thankfully! This feels much cleaner.

If the data isn't seeded into the snapshot, Ghost will read in the keys and set them. It's definitely a step we want to avoid if possible since it adds flakiness, I suppose with the exception of a test specifically testing setting up Stripe.

e2e/helpers/services/stripe/builders.ts
9larsons: I would like more discussion on this. I think we could rename them to factory-style names but I don't want them to live in the `/data-factory` dir nor do I want them to be confused with the function of those factories. 

Services are distinct and I think there's limited use case (if any?) where we'd want to create Stripe data in a test that isn't via Ghost. Conceptually I think that's more where I draw the line.

e2e/helpers/playwright/fixture.ts
9larsons: Dependencies are not based on where it appears in the file, if that's what you mean. I think we're ok here, though if we want to order it such that it feels progressive, I'm good w/ that.

e2e/helpers/playwright/fixture.ts
9larsons: Stripe config is determined via the presence of the `sk_` and `pk_` keys. This handling has been updated so I think it's much more clear now as that data is able to be seeded in the db.

e2e/tests/admin/members/stripe-webhooks.test.ts
9larsons: We discussed this one, but to add notes here: 

This iteration is the simplest slice - I think the end-state would be a test where the Stripe server receives the request from Portal w/ the redirect url, kicks back the expected success url, and does all the webhook sending to Ghost.

In order to do the Portal side of things, we need to make paid tiers fully functional, which'll be its own mini project and slice, plus whatever extra is needed in Portal.

e2e/helpers/services/stripe/webhook-client.ts
```diff
@@ -0,0 +1,34 @@
...
+import type {StripeEvent} from './builders';
+
+const DEFAULT_WEBHOOK_SECRET = 'DEFAULT_WEBHOOK_SECRET';
```
troyciesco: should this be `process.env.WEBHOOK_SECRET || 'DEFAULT_WEBHOOK_SECRET'` like it is in [config.js](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/services/stripe/config.js#L56)?

from what I/agents can tell it's at least possible to end up with a mismatch depending on how your environment is set up

e2e/helpers/environment/service-managers/mysql-manager.ts
```diff
@@ -170,6 +178,22 @@ export class MySQLManager {
         debug('site_uuid updated in database settings:', siteUuid);
     }
 
+    async updateStripeSettings(database: string, secretKey: string, publishableKey: string): Promise<void> {
```
cmraible, line 178: Nice, this feels much cleaner 🙌

e2e/helpers/playwright/fixture.ts
cmraible: Oh gotcha, yeah I think I had it in my head that it was order dependent. I do think it would read a bit easier if they were in order (e.g. if baseURL was defined before anything that depends on it), but I don't feel strongly about it, not blocking.

cmraible (changes requested):
We chatted offline, just submitting the review formally: 

- We discussed using the factory base class in place of the builders
- And fixing the tests in CI

cmraible (approved):
LGTM 🚢

PR #26518, opened 2026-02-19
Added inbox link to Mailpit in development

EvanHahn:
no ref

This is a development-only change.

If you use an `example.com` email address, show an inbox link to `localhost:8025`.

Note that this contains a Portal patch release.

### DIFF
--- a/ghost/core/core/server/lib/get-inbox-links.ts
+++ b/ghost/core/core/server/lib/get-inbox-links.ts
@@ -34,7 +34,7 @@ import logging from '@tryghost/logging';
 
 type GetLinkFn = (options: Readonly<{recipient: string; sender: string}>) => string;
 
-type ProviderName = 'gmail' | 'yahoo' | 'outlook' | 'proton' | 'icloud' | 'hey' | 'aol' | 'mailru';
+type ProviderName = 'gmail' | 'yahoo' | 'outlook' | 'proton' | 'icloud' | 'hey' | 'aol' | 'mailru' | 'dev-mailpit';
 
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/lib/get-inbox-links.ts
```diff
@@ -115,7 +115,13 @@ const PROVIDERS: ReadonlyArray<Provider> = [
         domains: ['mail.ru'],
         getDesktopLink: ({sender}) => buildUrl('https://e.mail.ru/search/', 'q_from', sender),
         getAndroidLink: () => getAndroidIntentUrl('ru.mail.mailapp', 'https://e.mail.ru/')
-    }
+    },
+    ...(process.env.NODE_ENV === 'development' ? [{
+        name: 'dev-mailpit' as const,
+        domains: ['example.com'],
+        getDesktopLink: () => 'http://localhost:8025',
+        getAndroidLink: () => 'http://localhost:8025'
```
cmraible, line 123: nit: I'm not sure if this would ever work for android?

cmraible (approved):

ghost/core/core/server/lib/get-inbox-links.ts
EvanHahn: We have to define an Android link, so I just set it to the same thing. I *could* make it optional, but it feels weird to do that for this dev-only behavior.

cmraible: Yeah makes sense to me, no objection 👌

PR #26521, opened 2026-02-19
[Draft] Removed GHOST_CDN_URL from Ghost(Pro) admin CI build

rob-ghost:
> **Draft proposal** — not ready to merge. See [full proposal](https://www.notion.so/ghost/Proposal-Relative-admin-build-with-CDN-asset-proxy-build-once-deploy-everywhere-30c51439c0308187a658dd8f42fff126).

## Context

Ghost(Pro) is moving to relative-path admin builds so a single artifact works across staging and production. The `GHOST_CDN_URL` config option is fully preserved for anyone who uses it — this PR only stops setting it in Ghost(Pro)'s CI build.

## This PR

Removes the `GHOST_CDN_URL` env var from the `deploy_admin` CI job. Without it set, Vite defaults to `base: './'` (relative paths) and Ember defaults to relative asset URLs. The existing code paths that read `GHOST_CDN_URL` are untouched — self-hosters and contributors who set the env var get the same behaviour as before.

Depends on BlackPearl's asset proxy being deployed first so that relative paths in `index.html` resolve correctly via the CDN.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1768,8 +1768,6 @@ jobs:
         env:
           DEPENDENCY_CACHE_KEY: ${{ needs.job_setup.outputs.dependency_cache_key }}
       - name: Build Admin
-        env:
-          GHOST_CDN_URL: https://assets.ghost.io/admin-forward/
         run: yarn nx run admin:build
       - name: Upload build artifact
         id: upload-artifacts


muratcorlu:
@rob-ghost @JohnONolan This is a breaking change, and should not be released before v7, if we care semantic versioning. I believe there are many self-hosted users of this undocumented feature, as many other undocumented (or not well documented) things in Ghost. Please give extra attention when you make breaking changes on an open-source project. This is also not something that will give you any extra hassle. Just don't use env variable on Ghost(Pro).

Of course I have no clue about BlackPearl or the relative admin build proposal, but still, being able to serve static assets from a separate domain is a good thing to have. And I can't see any reason to drop this feature.

rob-ghost:
@muratcorlu thank you for the feedback and sorry for any alarm. This PR is actually a part of a proposal draft that hasn't been reviewed or accepted and we definitely won't be making any breaking changes as a part of that. I've changed this to a draft PR and also reverted the removal of the build option to be absolutely clear that we're not intending to do that anytime soon.

Sorry for the confusion caused, we'll look into how to make it more clear when PRs are a part of a proposal / suggestion and when they're intended for merge.

muratcorlu:
Hello @rob-ghost. Thanks for the reply. I'm glad to hear that you re-consider this change. And yes, we really need to have more clear and open communication about development of Ghost. I understand you have Ghost(Pro) related tasks, but since we all (other hosting providers and self-hosters) use the same code-base, it's important for us to be able to track what is going on. Thanks for giving attention to that topic.

JohnONolan:
@muratcorlu you are welcome to share feedback and we're happy to receive it, but please take it down a notch.

muratcorlu:
Hello @JohnONolan . If I wrote in a disturbing tone, I apologize. Since English is not my native language, I may sometimes have difficulty adjusting the tone. This may result in excessive directness. I will be more careful.

rob-ghost:
Superseded by #26556 — same intent (remove `GHOST_CDN_URL` from CI), but now depends on the deploy-time `sed` rewrite approach instead of the proxy-based approach.

See the full set of PRs:
- [Ghost #26555](https://github.com/TryGhost/Ghost/pull/26555) — meta tag + bridge script
- [Ghost-Moya #135](https://github.com/TryGhost/Ghost-Moya/pull/135) — sed-based deploy
- [Ghost #26556](https://github.com/TryGhost/Ghost/pull/26556) — remove GHOST_CDN_URL from CI

PR #26523, opened 2026-02-20
Added entitlements JWT endpoint

9larsons:
ref https://ghost.slack.com/archives/C018EKC56JF/p1771527734858719

The existing JWT is used for identity and only contains the email address. It has a TTL of 10 minutes which is reasonable for the use case but little else.

We have another use case where integrations often want to validate the source, but are left with no option but to hit the members Admin API endpoint with an Admin API key because the email isn't sufficient to know if the member has access to a tier, is paid, isn't specific to email (which could change), etc., all of which are used for integrations.

This endpoint would still require the members session cookie and has the following props, e.g.:
```
 sub: darylmayer880479@example.com
 scope: members:entitlements:read
 member_uuid: 629c5190-f068-4f49-b533-00d7477a7aed
 paid: false
 active_tier_ids: []
 jti: present
 iat: 1771555398
 exp: 1771555518
 aud/iss: http://localhost:2368/members/api
 Lifetime: 120s (2 minutes)
```

### DIFF
--- a/ghost/core/core/server/services/members/members-api/members-api.js
+++ b/ghost/core/core/server/services/members/members-api/members-api.js
@@ -330,7 +330,6 @@ module.exports = function MembersAPI({
 
         return tokenService.encodeEntitlementToken({
             sub: member.email,
-            memberId: member.id,
             memberUuid: member.uuid,
             paid: member.status !== 'free',
             activeTierIds
--- a/ghost/core/core/server/services/members/members-api/members-api.js
+++ b/ghost/core/core/server/services/members/members-api/members-api.js
@@ -329,8 +329,7 @@ module.exports = function MembersAPI({
             .filter(Boolean))];
 
         return tokenService.encodeEntitlementToken({
-            sub: member.email,
-            memberUuid: member.uuid,
+            sub: member.uuid,
             paid: member.status !== 'free',
             activeTierIds
         });
--- a/ghost/core/core/server/services/members/members-api/members-api.js
+++ b/ghost/core/core/server/services/members/members-api/members-api.js
@@ -329,7 +329,8 @@ module.exports = function MembersAPI({
             .filter(Boolean))];
 
         return tokenService.encodeEntitlementToken({
-            sub: member.uuid,
+            sub: member.email,
+            memberUuid: member.uuid,
             paid: member.status !== 'free',
             activeTierIds
         });
--- a/ghost/core/core/server/services/members/members-api/members-api.js
+++ b/ghost/core/core/server/services/members/members-api/members-api.js
@@ -317,6 +317,26 @@ module.exports = function MembersAPI({
         return tokenService.encodeIdentityToken({sub: member.email});
     }
 
+    async function getMemberEntitlementToken(transientId) {
+        const member = await getMemberIdentityDataFromTransientId(transientId);
+        if (!member) {
+            return null;
+        }
+
+        const activeTierIds = [...new Set((member.subscriptions || [])
+            .filter(sub => users.isActiveSubscriptionStatus(sub.status))
+            .map(sub => sub?.tier?.id)
+            .filter(Boolean))];
+
+        return tokenService.encodeEntitlementToken({
+            sub: member.email,
+            memberId: member.id,
+            memberUuid: member.uuid,
+            paid: member.status !== 'free',
+            activeTierIds
+        });
+    }
+
     async function setMemberGeolocationFromIp(email, ip) {
         if (!email || !ip) {
             throw new errors.IncorrectUsageError({
@@ -417,6 +437,7 @@ module.exports = function MembersAPI({
         middleware,
         getMemberDataFromMagicLinkToken,
         getMemberIdentityToken,
+        getMemberEntitlementToken,
         getMemberIdentityDataFromTransientId,
         getMemberIdentityData,
         cycleTransientId,
--- a/ghost/core/core/server/services/members/members-api/services/token-service.js
+++ b/ghost/core/core/server/services/members/members-api/services/token-service.js
@@ -47,7 +47,7 @@ module.exports = class TokenService {
             keyid: jwk.kid,
             algorithm: 'RS512',
             audience: this._issuer,
-            expiresIn: '2m',
+            expiresIn: '5m',
             issuer: this._issuer
         });
     }
--- a/ghost/core/core/server/services/members/members-api/services/token-service.js
+++ b/ghost/core/core/server/services/members/members-api/services/token-service.js
@@ -31,6 +31,7 @@ module.exports = class TokenService {
 
     async encodeEntitlementToken({
         sub,
+        memberUuid,
         paid,
         activeTierIds = []
     }) {
@@ -40,6 +41,7 @@ module.exports = class TokenService {
             sub,
             kid: jwk.kid,
             scope: 'members:entitlements:read',
+            member_uuid: memberUuid,
             paid,
             active_tier_ids: activeTierIds,
             jti: crypto.randomUUID()
--- a/ghost/core/core/server/services/members/members-api/services/token-service.js
+++ b/ghost/core/core/server/services/members/members-api/services/token-service.js
@@ -1,5 +1,6 @@
 const jose = require('node-jose');
 const jwt = require('jsonwebtoken');
+const crypto = require('crypto');
 
 module.exports = class TokenService {
     constructor({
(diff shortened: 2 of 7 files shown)


cathysarisky (commented):
This would solve a number of cases (only some of which I wrote) where an external integration uses the jwt to validate who the member is (based on email), but then has to use an admin api key for whatever information is needed.  I'd love to reduce the need to store an admin api key when all I need is to read a little information about the member, like a non-changing ID, or what tier's privs they should get.

ghost/core/core/server/services/members/members-api/services/token-service.js
```diff
@@ -28,6 +29,31 @@ module.exports = class TokenService {
...
+            scope: 'members:entitlements:read',
+            member_uuid: memberUuid,
+            paid,
+            active_tier_ids: activeTierIds,
+            jti: crypto.randomUUID()
+        }, this._privateKey, {
+            keyid: jwk.kid,
+            algorithm: 'RS512',
+            audience: this._issuer,
+            expiresIn: '2m',
```
rob-ghost: **Question:** What are the tradeoffs for choosing 2m expiry time for the entitlements token? What's the intended usage pattern for the token?

ghost/core/core/server/services/members/members-api/services/token-service.js
```diff
@@ -28,6 +29,31 @@ module.exports = class TokenService {
...
+    async encodeEntitlementToken({
+        sub,
+        memberUuid,
+        paid,
+        activeTierIds = []
+    }) {
+        const jwk = await this._keyStoreReady;
+
+        return jwt.sign({
+            sub,
```
rob-ghost, line 41: **Question:** `sub` is the member's email, should it be the member's `uuid` instead?

ghost/core/core/server/services/members/members-api/members-api.js
```diff
@@ -317,6 +317,25 @@ module.exports = function MembersAPI({
...
+
+        const activeTierIds = [...new Set((member.subscriptions || [])
+            .filter(sub => users.isActiveSubscriptionStatus(sub.status))
+            .map(sub => sub?.tier?.id)
+            .filter(Boolean))];
+
+        return tokenService.encodeEntitlementToken({
+            sub: member.email,
+            memberUuid: member.uuid,
+            paid: member.status !== 'free',
```
rob-ghost, line 334: **Question:** Should we derive `paid` from `activeTierIds`? Could they disagree (perhaps in a comp'd scenario where they're not free, but have no active tiers, depending on how comps are implemented)?

rob-ghost (commented):
**Question:** What are the alternatives considered and why is the JWT approach the best one?

**Question:** What's the blast radius of this change? Goes this affect any of the following:

1. Ghost(pro)
2. Self-hosters
3. Clients (first-party, third-party)
4. Other teams - do any docs / schemas need updating?

**Question:** Have we considered the security of this? If the JWT was exposed and we're sharing membership subscription status, is that a concern?

cathysarisky:
"Question: Have we considered the security of this? If the JWT was exposed and we're sharing membership subscription status, is that a concern?"

Not my PR, but I'll comment here.  If a user has a membership somewhere embarrassing and that fact leaks, that's bad.   The existing /members/api/session already has the member's email and that a membership exists.   Is it a marginally worse leak if it's a paid membership?  I guess arguably?

But... in the current situation, every integration that needs access to any membership data beyond what's in the current JWT is asking the site owner to give them unscoped admin api access.  There are a lot of admin API keys out there that don't need to be, because there's not currently another way to figure out basic user entitlements.  The blast radius on a hypothetical admin api key leak from an integration is significant, and would include the ability to exfiltrate all membership data, send phishing/malware emails, deface the site, etc.

In the absence of being able to do tightly-scoped keys, I think this is a net security enhancement.

ghost/core/core/server/services/members/members-api/services/token-service.js
9larsons: After talking with Cathy, 5m feels like a more functional duration. The intended use would be to request as needed to validate access for outside integrations with Ghost. I expect it to be requested as-needed and then it simply needs to last the lifetime of whatever the process is, which to my understanding, should in most cases be short-lived - e.g. verifying tier access in order to get increased access in a Discord server.

ghost/core/core/server/services/members/members-api/services/token-service.js
9larsons: Updated - I was on the fence about keeping it consistent with the other, though I think `uuid` makes more sense here.

cathysarisky: If this were the uuid, then would we not have the member email?  I can see integrations wanting both, and if the email isn't here, they're going to need to also get the session jwt for the email.  (Or go back to the admin api...)

ghost/core/core/server/services/members/members-api/members-api.js
9larsons: They should already be in agreement, afaict. If someone is comped they'll have an active tier id, and anyone free should have an empty active tier id.

cathysarisky: I /think that a subscription with cancel_at_period_end:true still has status:active, correct?  Definitely want those members to be "paid" until the subscription actually cancels.

ghost/core/test/unit/server/services/members/members-api/services/token-service.test.js
```diff
@@ -30,6 +30,27 @@ describe('TokenService', function () {
         });
     });
 
+    describe('encodeEntitlementToken', function () {
+        it('can encode an entitlement token and decode it afterwards', async function () {
+            const token = await tokenService.encodeEntitlementToken({
+                sub: 'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee',
```
cathysarisky: Earlier you had the sub being an email. Once you settle on what's in the token, you might wand to update the test to match?

cathysarisky (commented):
I think the ideal (from an integrations standpoint) jwt would contain both the uuid and the member email.  Integrations may need either or both, and it'd be annoying to have to get both this jwt and the session jwt to get them.  I don't have any strong opinion on which one occupies the 'sub' spot.

ghost/core/core/server/services/members/members-api/services/token-service.js
9larsons: Talked w/ Cathy and reverted changes - `sub` is still email, consistent with the identity token.

9larsons:
## Problem

Third-party integrations and theme developers need a way to verify a member's entitlements (paid status, active tiers) without making server-side API calls or exposing sensitive member data. The existing identity token (`/members/api/session`) provides authentication but doesn't carry authorization/entitlement claims, forcing integrations to make additional API calls to determine what a member has access to.

## Proposal

Add a new public endpoint `GET /members/api/entitlements` that returns a short-lived (5 minute) RS512-signed JWT containing the authenticated member's entitlement claims:

- `sub` — member email
- `member_uuid` — member UUID
- `paid` — boolean indicating paid vs free status
- `scope` — `members:entitlements:read`
- `active_tier_ids` — array of tier IDs the member currently has active subscriptions for
- `jti` — unique token ID (replay protection)

The token is issued from the existing session cookie (same auth flow as the identity token) and signed with the same RSA key pair. No new dependencies or infrastructure required.

## Not Doing

- **Write/mutation scopes** — this token is read-only entitlement data
- **Custom claims or extensibility** — no plugin system for adding arbitrary claims
- **Token refresh/rotation mechanism** — clients re-fetch when the 5-minute TTL expires
- **New key management** — reuses existing members API key pair
- **Caching or rate limiting** — handled at the infrastructure layer if needed later

## Trade-offs

- **5-minute TTL is a balance**: short enough to limit stale entitlement data, long enough to avoid excessive re-fetching. A member who upgrades/downgrades won't see the change reflected in already-issued tokens for up to 5 minutes.
- **Email in `sub` claim**: follows the existing identity token convention, but means the JWT contains PII.
- **No revocation mechanism**: tokens are stateless. If a member's subscription is cancelled, existing tokens remain valid until expiry.
- **Adds a new public endpoint**: increases the API surface area that needs to be maintained and secured.

## Rollback

- **Easy**: Remove the route registration in `members/app.js` and the endpoint stops responding. No database changes, no migrations, no data to clean up. Pure code-only change.

## Blast Radius

- **New public endpoint**: `GET /members/api/entitlements` on the public-facing site
- **No existing behavior changed**: identity token, session management, and all other member endpoints are untouched
- **Files touched**: token-service, members-api, members-ssr, middleware, members app router
- **Consumers**: no current consumers — this is a new capability for future integrations/themes to adopt

rob-ghost (approved):
Solid proposal 🎯 nice work! 👍🏻

cathysarisky (approved):
LGTM! Thanks for getting both uuids and emails in there.

PR #26528, opened 2026-02-21
Fixed duplicate tag insertion in assignTagToUserPosts

Ray0907:
## What does this PR do?

  Fixed a bug in `assignTagToUserPosts` where posts that already had the
  user tag were being re-inserted, causing duplicate tag assignments.

  ## Why is this change needed?

  The original filter logic compared raw row objects (`{post_id: 'xxx'}`)
  against string IDs using `Array.includes()`, so it never matched.
  This meant every call to `assignTagToUserPosts` would insert duplicate
  `posts_tags` rows for posts that already had the tag.

  ## How does it work?

  - Extract `post_id` from the raw query results into a `Set`
  - Filter `usersPostIds` (already mapped to strings) against the Set
  - `Set.has()` also improves performance from O(n) to O(1) per lookup

  ## Checklist

  - [x] I've read and followed the [Contributor
  Guide](https://github.com/TryGhost/Ghost/blob/main/.github/CONTRIBUTING.md)

---

> [!NOTE]
> **Low Risk**
> Small, well-scoped change to tag-assignment filtering logic plus a targeted unit test; low risk aside from potential edge cases in the tag lookup/query results.
> 
> **Overview**
> Prevents `assignTagToUserPosts` from re-inserting `posts_tags` rows for posts that already have the user’s internal `#slug` tag by correctly filtering existing tagged posts (building a `Set` of `post_id`s from the query results) and early-returning when nothing remains to tag.
> 
> Adds a unit test that stubs `DomainEvents.dispatch` and verifies only untagged posts are inserted, with `Post.addActions` and the dispatched event receiving the filtered post IDs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6256827fb572da66853fe1be693c8111fd8ff650. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


9larsons (approved):
Thanks for the contribution!

Pushed two follow-ups:

- `036808e` — restored both files to main and reapplied only the substantive bits (filter fix, early-return guard, new test). The earlier commits had run Prettier across both files, which buried the real diff.
- `f7a8ed3` — added a second test for the empty-after-filter path so the new early-return guard is actually exercised, and factored the shared mock setup between the two tests.

Small nit on the description: "every call to `assignTagToUserPosts` would insert duplicate `posts_tags` rows for posts that already had the tag" is technically true, but the function only runs once per user deletion against a tag named after the deleted user. 

Duplication only occurs in narrow scenarios — e.g. slug reuse where the prior user's tag persists and overlaps with the new user's posts, or a manually pre-applied `#slug` tag. The fix is correct; severity is just lower than the wording implies.

PR #26530, opened 2026-02-22
🌐 Updated Bulgarian translations

yovko:
**Changes:**

Updated Bulgarian translation of comments.json and ghost.json to include recently added strings.

### DIFF
--- a/ghost/i18n/locales/bg/ghost.json
+++ b/ghost/i18n/locales/bg/ghost.json
@@ -26,7 +26,7 @@
     "Keep reading": "Продължете четенето",
     "Less like this": "По-малко такива",
     "Manage subscription": "Управление на абонамента",
-    "Manage your preferences": "",
+    "Manage your preferences": "Управление на настройките",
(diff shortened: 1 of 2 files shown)


ghost/i18n/locales/bg/ghost.json
```diff
@@ -26,7 +26,7 @@
     "Keep reading": "Продължете четенето",
     "Less like this": "По-малко такива",
     "Manage subscription": "Управление на абонамента",
-    "Manage your preferences": "",
+    "Manage your preferences": "Управление на настройките",
```
cathysarisky, line 29: 🤖 AI: [Possible consistency issue] In "Manage your preferences" → "Управление на настройките", I suspect it might be more consistent with the rest of the BG UI to mirror the imperative style used elsewhere (e.g. "Управлявайте настройките си" or similar), since this string is a call-to-action rather than a noun label. Would you consider whether an imperative form would fit better in context, please?

cathysarisky (approved):

PR #26539, opened 2026-02-23
Unified environment manager

jonatansberg:
## Summary

This PR reworks E2E environment orchestration into a single unified flow for local and CI, while adding support for build-mode testing against configurable Ghost backend images (`GHOST_E2E_IMAGE`).

The main goal is to remove drift between “dev” and “CI” behaviors and make backend-version coverage easier to maintain.

## Main changes

- Replaced split environment managers with one unified `EnvironmentManager` + worker-scoped `GhostManager`.
- Standardized lifecycle across modes:
  - shared infra (MySQL/Redis/Mailpit/Tinybird) is pre-started
  - global setup creates/snapshots a migrated base DB
  - each test clones DB + restarts Ghost with isolated DB
- Removed older environment-specific managers and compose wrapper utilities.
- Added build mode runtime support:
  - new `docker/dev-gateway/Caddyfile.build`
  - updated `e2e/Dockerfile.e2e` to layer public app bundles into `/content/files`
- Reworked CI E2E orchestration to run Playwright inside the Playwright container with dedicated prep scripts, better failure log collection, and explicit infra up/down.
- Added explicit infra and runtime scripts (`infra:up/down`, preflight/build prep, host/container Playwright runners, Tinybird state sync).
- Updated E2E docs and package scripts to match the new flow.
- Tests no longer implicitly bootstrap all infra themselves; infra is expected to be running (`yarn dev` or `yarn workspace @tryghost/e2e infra:up`).
- CI now executes tests inside a Playwright container (instead of host Playwright setup) (reworked https://github.com/TryGhost/Ghost/pull/26037)


jonatansberg:
@coderabbitai review

e2e/package.json
```diff
@@ -11,14 +11,18 @@
...
-    "test": "playwright test --project=main",
-    "test:analytics": "playwright test --project=analytics",
-    "test:all": "playwright test --project=main --project=analytics",
-    "test:single": "playwright test --project=main -g",
-    "test:debug": "playwright test --project=main --headed --timeout=60000 -g",
+    "pretest": "(test -n \"$GHOST_E2E_SKIP_BUILD\" || test -n \"$CI\") && echo 'Skipping Docker build' || echo 'Tip: run yarn dev or yarn workspace @tryghost/e2e infra:up before running tests'",
+    "infra:up": "bash ./scripts/infra-up.sh",
+    "infra:down": "bash ./scripts/infra-down.sh",
+    "tinybird:sync": "node ./scripts/sync-tinybird-state.mjs",
+    "tinybird:fetch-config": "yarn tinybird:sync",
```
jonatansberg: Why do we need both of these? Can't we get rid of one?

e2e/package.json
```diff
@@ -11,14 +11,18 @@
...
+    "preflight:build": "bash ./scripts/prepare-ci-e2e-build-mode.sh",
+    "test": "bash ./scripts/run-local-playwright.sh playwright test --project=main",
```
jonatansberg: Nit: let's align the naming between the CI and Playwright versions of the run scripts. Right now they're subtly different. 

Maybe we could even unify them so that it's a single run-playwright script that checks the appropriate environment variables to determine if we should run inside of Docker or on host, making it easier to run the test within Docker locally if you'd like.

e2e/scripts/run-playwright-ci.sh
```diff
@@ -0,0 +1,31 @@
...
+    exit 1
+fi
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+
+cd "$REPO_ROOT"
+
+PLAYWRIGHT_VERSION="$(node -p 'require("./e2e/package.json").devDependencies["@playwright/test"]')"
+PLAYWRIGHT_IMAGE="mcr.microsoft.com/playwright:v${PLAYWRIGHT_VERSION}-noble"
```
jonatansberg: This is repeated in multiple places. Would it make sense to have a small env script that we're including?

e2e/scripts/sync-tinybird-state.mjs
```diff
@@ -0,0 +1,140 @@
...
+        adminToken: env.TINYBIRD_ADMIN_TOKEN,
+        trackerToken: env.TINYBIRD_TRACKER_TOKEN
+    }, null, 2));
+}
+
+try {
+    if (!isTinybirdRunning()) {
+        clearConfigIfPresent();
+        log(`Tinybird is not running for compose project ${composeProject}; skipping Tinybird state sync`);
+        process.exit(0);
```
jonatansberg, line 98: Verify that it makes sense to exit with a zero here. Maybe we should be failing more loudly. Same applies for the next check.

e2e/scripts/sync-tinybird-state.mjs
```diff
@@ -0,0 +1,140 @@
...
+    const rawEnv = fetchConfigFromTbCli();
+    const env = parseEnv(rawEnv);
+
+    if (!env.TINYBIRD_WORKSPACE_ID || !env.TINYBIRD_ADMIN_TOKEN) {
+        clearConfigIfPresent();
+        log('Tinybird config is not available yet; continuing without e2e/data/state/tinybird.json');
+        process.exit(0);
+    }
+
+    writeConfig(env);
```
jonatansberg, line 109: If we're using this for the local development setup now, are there anything we can pull out of that compose file and whatnot?

jonatansberg:
@coderabbitai review

e2e/helpers/environment/service-managers/mysql-manager.ts
```diff
@@ -13,20 +14,21 @@ interface ContainerWithModem extends Container {
...
 export class MySQLManager {
-    private readonly dockerCompose: DockerCompose;
+    private readonly docker: Docker;
     private readonly containerName: string;
 
-    constructor(dockerCompose: DockerCompose, containerName: string = 'mysql') {
-        this.dockerCompose = dockerCompose;
+    constructor(containerName: string = 'ghost-dev-mysql') {
+        this.docker = new Docker();
         this.containerName = containerName;
```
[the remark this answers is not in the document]
jonatansberg: This seems to be a false positive. MySQLManager defaults to ghost-dev-mysql, as does compose.dev.yaml.

e2e/scripts/load-playwright-container-env.sh
```diff
@@ -0,0 +1,21 @@
...
+fi
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+
+cd "$REPO_ROOT"
+
+PLAYWRIGHT_VERSION="$(node -p 'require("./e2e/package.json").devDependencies["@playwright/test"]')"
+PLAYWRIGHT_IMAGE="mcr.microsoft.com/playwright:v${PLAYWRIGHT_VERSION}-noble"
+WORKSPACE_PATH="${GITHUB_WORKSPACE:-$REPO_ROOT}"
```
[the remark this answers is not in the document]
jonatansberg: Since we're only sourcing this from other scripts, I don't think this is a concern.

e2e/helpers/environment/constants.ts
```diff
@@ -6,105 +6,106 @@ const __dirname = path.dirname(__filename);
...
+ * Build mode assets are baked into the E2E image via ENV vars in e2e/Dockerfile.e2e.
+ */
+export const LOCAL_ASSET_URLS = [
+    'portal__url=/ghost/assets/portal/portal.min.js',
+    'comments__url=/ghost/assets/comments-ui/comments-ui.min.js',
+    'sodoSearch__url=/ghost/assets/sodo-search/sodo-search.min.js',
+    'sodoSearch__styles=/ghost/assets/sodo-search/main.css',
+    'signupForm__url=/ghost/assets/signup-form/signup-form.min.js',
+    'announcementBar__url=/ghost/assets/announcement-bar/announcement-bar.min.js'
+] as const;
```
jonatansberg: Can we remove this part since we are baking these in to the E2E docker image?

rob-ghost (approved):
CI is 🟢 so I'm happy no matter what 👍🏻 

I think there might be a couple of orphaned files:

`e2e/helpers/environment/service-managers/portal-manager.ts`'s only consumer was the `docker-compose.ts` removed in this PR.

`e2e/compose.yml` doesn't seem to have any references after this PR.

Can be done as a fast-follow though! 🚀

.github/workflows/ci.yml
```diff
@@ -1261,27 +1261,28 @@ jobs:
...
         env:
           GHOST_E2E_BASE_IMAGE: ${{ steps.load.outputs.image-tag }}
-        run: yarn workspace @tryghost/e2e build:docker
+        run: bash ./e2e/scripts/prepare-ci-e2e-job.sh
 
-      - name: Pull images
+      - name: Run e2e tests in Playwright container
         env:
+          TEST_WORKERS_COUNT: 1
+          GHOST_E2E_MODE: build
```
cmraible: What is this?

e2e/helpers/pages/public/public-page.ts
```diff
@@ -100,11 +99,32 @@ export class PublicPage extends BasePage {
         });
     }
 
+    protected async waitForMemberAttributionReady(): Promise<void> {
+        // TODO: Ideally we should find a way to get rid of this. This is currently needed
+        // to prevent flaky attribution-dependent assertions in CI.
+        await this.page.waitForFunction(() => {
```
cmraible, line 105: When running locally in "dev" mode, I get 49 tests failing, most of which fail here, e.g.:

```
1) [main] › tests/admin/members/member-activity-events.test.ts:28:5 › Ghost Admin - Member Activity Events › welcome email event appears in member activity feed

    Test timeout of 30000ms exceeded.

    Error: page.waitForFunction: Target page, context or browser has been closed

       at helpers/pages/public/public-page.ts:105

      103 |         // TODO: Ideally we should find a way to get rid of this. This is currently needed
      104 |         // to prevent flaky attribution-dependent assertions in CI.
    > 105 |         await this.page.waitForFunction(() => {
          |                         ^
      106 |             try {
      107 |                 const raw = window.sessionStorage.getItem('ghost-history');
      108 |
        at PublicPage.waitForMemberAttributionReady (/Users/chris/Developer/TryGhost/Ghost/review_26539_ber-3363-unified-environment-manager/e2e/helpers/pages/public/public-page.ts:105:25)
        at PublicPage.openPortalViaSubscribeButton (/Users/chris/Developer/TryGhost/Ghost/review_26539_ber-3363-unified-environment-manager/e2e/helpers/pages/public/public-page.ts:122:20)
        at signupViaPortal (/Users/chris/Developer/TryGhost/Ghost/review_26539_ber-3363-unified-environment-manager/e2e/helpers/playwright/flows/signup.ts:8:22)
        at /Users/chris/Developer/TryGhost/Ghost/review_26539_ber-3363-unified-environment-manager/e2e/tests/admin/members/member-activity-events.test.ts:34:44

    Error Context: test-results/admin-members-member-activ-4bb0f-ars-in-member-activity-feed-main/error-context.md

```

e2e/helpers/pages/public/public-page.ts
```diff
@@ -81,7 +81,6 @@ export class PublicPage extends BasePage {
         const testInfo = test.info();
         let pageHitPromise = null;
         if (testInfo.project.name === 'analytics') {
-            await this.enableAnalyticsRequests();
```
cmraible, line 84: Was this removed on purpose? I think this is still necessary?

e2e/scripts/sync-tinybird-state.mjs
```diff
@@ -0,0 +1,139 @@
...
+    if (!env.TINYBIRD_WORKSPACE_ID || !env.TINYBIRD_ADMIN_TOKEN) {
+        clearConfigIfPresent();
+        throw new Error('Tinybird is running but required config values are missing in /mnt/shared-config/.env.tinybird');
+    }
+
+    writeConfig(env);
+    log(`Wrote Tinybird config to ${configPath}`);
+
+    try {
+        truncateAnalyticsEvents();
```
cmraible: suggestion(blocking): we shouldn't truncate the Tinybird events here - this deletes all events in the Tinybird instance from `compose.dev.yaml`, which would be an unexpected side effect of running end to end tests.

All the events are partitioned by site_uuid, and each test gets its own site_uuid, so this is unnecessary anyhow.

e2e/scripts/run-playwright-host.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
cmraible, line 1: suggestion (non-blocking): it might be neat if we just always ran playwright in the docker image, whether locally or in CI. Not a big deal, but just one more potential source of drift 🤷

cmraible (dismissed):
Really nice! It looks like CI is very stable here, with all tests passing without even a single retry. Overall this is awesome, but a few things I'd like to see addressed before merging:

1. We shouldn't truncate Tinybird events. See [comment](https://github.com/TryGhost/Ghost/pull/26539/changes#r2849885244)

2. I'm seeing a lot of tests failing when running locally in "dev" mode in the `waitForMemberAttributionReady` method. See [comment](https://github.com/TryGhost/Ghost/pull/26539/changes#r2849823903).

e2e/helpers/pages/public/public-page.ts
jonatansberg: Interesting, I'll double-check why that is. I didn't look too closely at the dev failures because there are some other tests that failed due to mismatches caused by SVGs being minified in prod and not in development that I've been meaning to fix separately.

e2e/helpers/pages/public/public-page.ts
jonatansberg: Seemed redundant since we're already enabling analytics requests a few rows down? This would effectively be a no-op.

e2e/scripts/run-playwright-host.sh
jonatansberg: I know, I saw that comment on the other PR as well. I think that's a meaningful follow-up to explore. I think there might be some other things that become harder if we go that route tho, e.g. integrating with the Playwright runner in VS Code and possibly a performance drop locally compared to running directly on host. 

Because of that I decided to stop here for now since I've been going with this for way too long and want to get it merged.

jonatansberg:
@rob-ghost I removed the unused compose file. I can't find the portal manager though; I think it might have already been removed.

e2e/scripts/sync-tinybird-state.mjs
jonatansberg: Good catch. Codex snuck that in there, and to me it seemed sensible, but I don't know the Tinybird and Analytics set up very well.

e2e/helpers/pages/public/public-page.ts
jonatansberg: It probably makes more sense to keep this one though... I don't think we need to enable an analytics request for every single page. I'll make that change and see if everything still passes in CI.

e2e/helpers/pages/public/public-page.ts
jonatansberg: I've been trying to reproduce this locally, but I haven't been able to. You might want to ensure you have restarted yarn dev locally so that all dependencies and the Docker images are up to date.

e2e/scripts/run-playwright-host.sh
cmraible: Ah yeah, makes sense - no problem

e2e/helpers/pages/public/public-page.ts
cmraible: Okay no worries, might just be a me thing. I'll investigate some more myself 👍

PR #26545, opened 2026-02-23
Added validation for outbox statuses

troyciesco:
closes [NY-809](https://linear.app/ghost/issue/NY-809/cleanup-consider-adding-constraints-to-status-fields-on-automated)

- No user-facing changes
- Adds schema validation to the `outbox.status` field
- While developing outbox functionality we deliberately left this validation off so that we could develop more easily. Now that we have a live use case (welcome emails) it makes sense to formalize the valid values for this field

### DIFF
--- a/ghost/core/test/unit/server/models/outbox.test.js
+++ b/ghost/core/test/unit/server/models/outbox.test.js
@@ -1,4 +1,5 @@
 const assert = require('node:assert/strict');
+const errors = require('@tryghost/errors');
 const models = require('../../../../core/server/models');
 const {OUTBOX_STATUSES} = require('../../../../core/server/models/outbox');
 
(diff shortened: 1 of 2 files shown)


ghost/core/test/unit/server/models/outbox.test.js
```diff
@@ -41,4 +42,23 @@ describe('Unit: models/outbox', function () {
...
+                status: 'not-a-real-status'
+            })
+                .then(function () {
+                    throw new Error('expected ValidationError');
+                })
+                .catch(function (err) {
+                    assert(Array.isArray(err));
+                    assert.equal(err.length, 1);
+                    assert.equal((err[0] instanceof errors.ValidationError), true);
+                    assert.match(err[0].context, /outbox\.status/);
```
troyciesco, line 60: this pattern is how it's done elsewhere, like [member-created-event.test.js](https://github.com/TryGhost/Ghost/blob/main/ghost/core/test/unit/server/models/member-created-event.test.js#L21)

troyciesco (commented):
@cmraible i think an `isIn` change doesn't require a migration but shout if i'm wrong!

cmraible (approved):
LGTM!

PR #26546, opened 2026-02-23
Added trigger_cd job and streamlined CI Docker builds

rob-ghost:
ref https://linear.app/ghost/issue/BER-3294/

## Commit 1: Added trigger_cd job and admin artifact upload to CI

Ghost CI builds production Docker images and admin assets but doesn't tell Ghost-Moya about them. The existing `canary` and `deploy_admin` jobs use `aurelien-baudet/workflow-dispatch` to trigger Ghost-Moya's legacy `deploy.yml` and `deploy-admin.yml` workflows. The new `cd.yml` pipeline in Ghost-Moya expects a `repository_dispatch` event instead, with the core image tag and admin artifact details in the payload.

**Admin artifact upload** is added as a step inside `job_docker_build_production` — after `yarn build:production` completes, the built admin is uploaded as an `admin-build-cd` artifact. No separate job needed since admin is already built as part of the production build.

**`trigger_cd`** dispatches a `ghost-artifacts-ready` event to Ghost-Moya with the core image SHA tag, admin artifact ID, and run metadata. It depends only on `job_setup` and `job_docker_build_production` — not on test completion. Tests run in parallel as a quality signal but don't gate the CD trigger; the environment's own smoke test in cd.yml is the deployment gate.

`dry_run` is forced to `true` on all automated dispatches until cd.yml is validated and ready to replace the existing canary deploy.

Requires Ghost-Moya cd.yml to be on the default branch — `repository_dispatch` only triggers workflows on main. Done: [#127](https://github.com/TryGhost/Ghost-Moya/pull/127), [#132](https://github.com/TryGhost/Ghost-Moya/pull/132), [#133](https://github.com/TryGhost/Ghost-Moya/pull/133), [#134](https://github.com/TryGhost/Ghost-Moya/pull/134).

## Commit 2: Removed development Docker image build and folded image inspect inline

**Removed `job_docker_build`** — the development image (`ghost-development` from root `Dockerfile`) had no meaningful consumers. Its only dependent was `job_inspect_image` which just reports image size and layers to the step summary.

**Removed `job_inspect_image`** as a separate job and moved the inspect step directly into `job_docker_build_production`. The full image build now always sets `load: true` so the image is available locally for `docker inspect`/`docker history`.

**Deleted root `Dockerfile`** — the 228-line multi-stage development Dockerfile is no longer referenced anywhere. Local dev uses `docker/ghost-dev/Dockerfile` (via `compose.dev.yaml`), CI uses `Dockerfile.production`, and E2E uses `e2e/Dockerfile.e2e`.


.github/workflows/ci.yml
```diff
@@ -1736,3 +1635,50 @@ jobs:
...
+            echo "pr_number=${{ github.event.pull_request.number }}" >> $GITHUB_OUTPUT
+          else
+            echo "skip=true" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+          echo "skip=false" >> $GITHUB_OUTPUT
+
+      - name: Dispatch to Ghost-Moya cd.yml
+        if: steps.params.outputs.skip != 'true'
+        uses: peter-evans/repository-dispatch@v3
```
cmraible, line 1670: nit: we may want to use v4: https://github.com/peter-evans/repository-dispatch/releases/tag/v4.0.1

although renovate should pick this up regardless.

cmraible (approved):

.github/workflows/ci.yml
rob-ghost: Yeah Claude has a tendency to pick older versions of pretty much everything its really annoying!

I'll update the versions as a fast-follow after this is merged 👍🏻

Ghost-Slimer:
Deployed to **staging** with ID: `4090` (via cd.yml)

Ghost-Slimer:
Deployed to **staging** with ID: `4090` (via cd.yml)

PR #26555, opened 2026-02-24
Replaced cdnUrl config with script-origin asset base derivation

rob-ghost:
ref https://linear.app/ghost/issue/ENG-1326

## Summary

- Replaced Ember's build-time `cdnUrl` config with a runtime `assetBase()` utility that derives the CDN base from where the Ember scripts were loaded — the same principle ES modules use for relative imports
- Removed the `ghost-cdn-url` meta tag and bridge script from `index.html` (no longer needed)
- Fixed `vite-ember-assets` plugin reading from the combined output directory instead of the Ember dist, which caused `././assets/` path prefixes to accumulate on repeated builds

## Background

Ghost(Pro) currently bakes absolute CDN URLs into the admin build at CI time via `GHOST_CDN_URL`, then string-replaces them per environment at deploy time. The new `cd.yml` flow ([Ghost-Moya #135](https://github.com/AnywhereGhost/Ghost-Moya/pull/135), merged) replaces this with a simpler approach: build once with relative paths, then rewrite only `index.html` at deploy time with `sed`.

Vite's ES modules handle this automatically — `import.meta.url` reflects where the module was loaded from, so dynamic imports resolve to CDN if the entry script was loaded from CDN. Ember uses classic AMD scripts with no module-relative resolution, so it previously needed a `cdnUrl` config value to construct absolute URLs for lazy-loaded assets.

Rather than shimming `cdnUrl` through a meta tag, this PR makes Ember work the same way as Vite: derive the asset base from where the scripts were loaded.

## Details

**`assetBase()` utility** (`ghost/admin/app/utils/asset-base.js`): Finds the Ember `<script>` element in the DOM, reads its `src` (always resolved to an absolute URL by the browser), and extracts the base path up to `/assets/`. If the script loaded from CDN (because `sed` rewrote `index.html`), the result is the CDN base. If local, it's the local admin root. The result is memoized. Falls back to `ghostPaths().adminRoot` if no matching script is found.

**Replaced 6 duplicated conditionals**: `lazy-loader.js`, `parse-history-event.js`, `admin-x-component.js`, `user.js`, and `fetch-koenig-lexical.js` all had `config.cdnUrl ? ... : ghostPaths()...` patterns. Each is now a single `assetBase()` call with unused imports removed.

**Removed `cdnUrl` from `config/environment.js`**: No longer needed at runtime. `GHOST_CDN_URL` still works at build time for `ember-cli-build.js` (`fingerprint.prepend`, `publicAssetURL`) — self-hosters who set it get the same behaviour as before.

**././ fix**: The `vite-ember-assets` plugin was reading Ember asset paths from `ghost/core/core/built/admin/index.html` — the combined output directory that gets overwritten by the plugin's own `closeBundle` hook. On each build, paths that already had `./` got another `./` prepended. Now reads from `ghost/admin/dist/index.html` (the Ember build's own output), producing clean `./assets/` paths.

### DIFF
--- a/apps/admin/vite-ember-assets.ts
+++ b/apps/admin/vite-ember-assets.ts
@@ -4,6 +4,7 @@ import fs from 'fs';
 import sirv from 'sirv';
 
 const GHOST_ADMIN_PATH = path.resolve(__dirname, '../../ghost/core/core/built/admin');
+const GHOST_ADMIN_DIST = path.resolve(__dirname, '../../ghost/admin/dist');
 
 function isAbsoluteUrl(url: string): boolean {
     return url.startsWith('http://') ||
@@ -29,12 +30,14 @@ export function emberAssetsPlugin() {
         transformIndexHtml: {
             order: 'post',
             handler() {
-                // Path to the Ghost admin index.html file
-                const indexPath = path.resolve(GHOST_ADMIN_PATH, 'index.html');
+                // Read from Ember's own build output (not the combined output
+                // in built/admin which gets overwritten by closeBundle and would
+                // accumulate duplicate path prefixes on repeated builds)
+                const indexPath = path.resolve(GHOST_ADMIN_DIST, 'index.html');
                 try {
                     const indexContent = fs.readFileSync(indexPath, 'utf-8');
                     const base = config.base || '/';
-                    
+
                     // Extract stylesheets
                     const styleRegex = /<link[^>]*rel="stylesheet"[^>]*href="([^"]*)"[^>]*>/g;
                     const styles: HtmlTagDescriptor[] = [];
--- a/ghost/admin/app/components/admin-x/admin-x-component.js
+++ b/ghost/admin/app/components/admin-x/admin-x-component.js
@@ -1,6 +1,7 @@
 import * as Sentry from '@sentry/ember';
 import Component from '@glimmer/component';
 import React, {Suspense} from 'react';
+import assetBase from 'ghost-admin/utils/asset-base';
 import config from 'ghost-admin/config/environment';
 import fetch from 'fetch';
 import fetchKoenigLexical from 'ghost-admin/utils/fetch-koenig-lexical';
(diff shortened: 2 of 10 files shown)


apps/admin/index.html
```diff
@@ -13,6 +13,21 @@
     <meta name="apple-mobile-web-app-title" content="Ghost" />
     <meta name="apple-mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-status-bar-style" content="black" />
+
+    <meta name="ghost-cdn-url" content="">
+    <script>
+    (function() {
+      var c = document.querySelector('meta[name="ghost-cdn-url"]');
```
jonatansberg: Where are we setting the ghost-cdn-url meta tag?

rob-ghost: So it works, but its kind of backwards for now.

We build with `GHOST_CDN_URL` which puts the value in ember's meta config (URL encoded). `ghost-cdn-url` is empty by default, so the bridge leaves the ember meta config alone.

If we set `ghost-cdn-url` by doing the string replacement, this overrides the existing build value.

The string replacement takes place during deploy depending on the environment we're deploying to: https://github.com/TryGhost/Ghost-Moya/pull/135

A different approach might be to move the `GHOST_CDN_URL` handling to this new meta tag and overwrite it with the replacement. Then `ghost-cdn-url` becomes the source-of-truth.

jonatansberg (commented):
Overall implementation looks reasonable. My main concerns would be around how this will interact with the existing Ghost Moya flows that try to do path replace and our release flows outside of the new CD setup. 

It should probably be fine, but it would be nice to double check that before we merge this and it goes out into production for everyone.

rob-ghost:
I built and ran a script locally which would test each possible permutation and check that all the requests resolved successfully:

1. Building without CDN config - relative paths
2. Building without CDN config - replace paths with CDN paths (i.e the new workflow)
3. Building with CDN config - absolute paths (i.e the current workflow)

Everything resolved successfully. I have as much confidence as I can have that this is good to go.

@jonatansberg is there anything else you'd like tested before we move?

rob-ghost:
@jonatansberg on further testing on staging, I don't think we need this at all.

For Ghost 6.x, we're always serving the Vite built index.html which supports the es modules. That works with the CDN URL replacement we're doing on Ghost(Pro)'s CI/CD flow without any changes to Ghost.

We're never serving Embers index.html, only for Ghost 5.x, and even then, that wouldn't be going through the new CI/CD pipeline, so they're not affected. We'd have to _backport_ this fix for it to even matter.

So I think this is all dead code as far as Ghost 6.x and Ghost(Pro) is concerned and we don't have to worry about this at all. I've tested `main` builds using the new CI/CD pipeline and it works with staging CDN URL just fine.

I'll consider closing this, but I'd like to get your thoughts on it first.

rob-ghost:
@jonatansberg Actually, I've been wrong, the libraries _are_ still dynamically loaded and we need this resolution for the CDN to work. Going to test this PR on the new staging flow and if it works I'll merge.

rob-ghost:
I can confirm this works on staging and fixes the CDN url issue. Merging.

PR #26561, opened 2026-02-24
Fixed incorrect usage of file upload progress

EvanHahn:
towards https://linear.app/ghost/issue/NY-1075

`progressTracker.current` is a `Map`, but this line treated it like an object. This fixes that.


9larsons (approved):

9larsons:
Would love for us to get further along with Typescript to catch more of these!

PR #26562, opened 2026-02-24
Improved E2E fixtures with reusable role auth state

jonatansberg:
ref https://linear.app/ghost/issue/BER-3366/rework-e2e-fixtures

## What changed
- Reworked global setup to create and reuse a local fixture package (DB snapshot + per-role auth states)
- Added role-aware Playwright auth context helper and role fixture support
- Added staff invite + role onboarding setup flow for owner/administrator/editor/author/contributor
- Added fixture cache invalidation checks (missing files/migration mismatch/forced reset)
- Updated dev/build gateway Caddy config to rewrite Origin consistently for admin/API traffic

## Notes
- This branch is rebased on `ber-3363-unified-environment-manager`
- Local override env var is `E2E_FORCE_FIXTURE_RESET=1`

## Validation
- `yarn workspace @tryghost/e2e test:types`
- `yarn workspace @tryghost/e2e lint`
- Focused and full Playwright runs executed in dev mode to inventory failures


9larsons (commented):
I validated this locally against a build-mode setup rather than `pnpm dev`, including a PR-matched production/E2E image pair (`ghost-production:pr26562` and `ghost-e2e:pr26562`). The core approach looks promising: forced fixture/auth generation passed, cache reuse passed, a 2-worker Admin UI probe passed, and `members-legacy/stripe-webhooks.test.ts` passed through repeated isolated env setup using the cached auth package.

A few changes I’d suggest before we lean on this as the auth/cycle-time path:

1. Bind the auth-state files to the snapshot manifest

Right now cache validation checks that each `e2e/data/state/auth/*.json` file exists, but it doesn’t prove those files belong to the same `/tmp/dump.sql` snapshot/manifest. If the DB snapshot and auth JSON get out of sync locally, we can still get a cache hit and then fail later with confusing auth/session behavior.

Suggestion: include auth-state hashes or a generated fixture package ID in the manifest. Ideally the same package ID is written into both the manifest metadata and the generated auth-state files, or the manifest records the auth-state file hashes after generation.

2. Include auth/session implementation inputs in local cache invalidation

The current fingerprint covers E2E fixture/helper files and migrations, but not the Ghost server code that defines session shape, origin validation, staff invite/signup behavior, or auth persistence. CI always rebuilds, but local cache reuse can hide auth regressions unless the developer remembers `E2E_FORCE_FIXTURE_RESET=1`.

Suggestion: add a small set of auth/session/staff setup implementation paths to the cache fingerprint, or document that auth/server changes require `E2E_FORCE_FIXTURE_RESET=1`. I’d prefer automatic invalidation for auth-related paths because this PR specifically optimizes reusable auth.

3. Add one browser-level Admin auth smoke

`role-auth-state.test.ts` currently proves the cached cookie/session works through `page.request`, which is useful, but it doesn’t fully prove Admin browser navigation with the stored state across worker gateway origins. The validation I ran caught this by running real UI files with `--workers=2`.

Suggestion: add one smoke that uses cached role state, navigates to Admin UI on a normal worker environment, and verifies the current user/role from the rendered Admin app or a browser-context request after navigation. This would protect the subtle bit here: storage state + cloned DB session origin + Admin frontend behavior working together.

Validation I ran locally:

- `pnpm --filter @tryghost/e2e test:types`
- targeted ESLint for changed fixture/auth files
- build-mode image validation with `ghost-e2e:pr26562`
- `E2E_FORCE_FIXTURE_RESET=1 ... role-auth-state.test.ts`: passed, 14 passed in ~1.2m
- no-force rerun: cache hit, auth/user setup skipped, 7 passed / 7 skipped in ~25.6s
- `settings-search.test.ts` + `sidebar/upgrade-banner.test.ts` with `--workers=2`: cache hit, 6 passed / 7 skipped in ~45s
- `members-legacy/stripe-webhooks.test.ts`: cache hit, per-test isolated webhook flow passed, 6 passed / 7 skipped in ~1.3m

So: the architecture looks sound from the targeted probes. The main thing I’d tighten is making the cache package harder to get into a locally stale or half-mismatched state, plus adding a browser-level smoke so CI covers the exact auth path we care about.

9larsons:
^That's from a guided automated review, dumped it here to keep it colocated. I've pushed some changes to E2E this week and I'm still aiming to take a deeper look at this myself.

jonatansberg:
The initial version of this tried to auto-detect auth related changes, like the review suggets, and that just ended up either with a massive list of files to hash or just always blowing the cache. We're making changes that would break the auth state extremely seldom, and it's more likely that would break that file list than the actual signed cookie, so I pulled that out.

If one would like to be defensive about it, we could add a separate step that just validates that the credentials are still valid, and then blows the cache if they're not. That would be a lot more robust than trying to hash things. I'd lean towards that we should let things blow up if the auth state changes tho, as that would also invalidate all active sessions for every single user.

PR #26569, opened 2026-02-24
Removed file upload hook's Ember dependency

EvanHahn:
towards https://linear.app/ghost/issue/NY-1097
ref https://github.com/TryGhost/Ghost/pull/26591
ref https://github.com/TryGhost/Ghost/pull/26593

This change should have limited user impact.

I want to move `useFileUpload` to a separate file. To do this, I need to remove its dependency on Ember's AJAX service. That meant exporting `useFetchApi` from `admin-x-framework` and depending on it there.

IMO, the "root" of the change is in `koenig-lexical-editor.js`. I'd start there if reviewing this patch.

`useFetchApi` is _slightly_ different from Ember's AJAX service, but it's our closest analog in the React world.


EvanHahn:
Abandoned in favor of other work.

PR #26571, opened 2026-02-24
Removed `free_months` concept from domain

mike182uk:
ref https://linear.app/ghost/issue/BER-3364

The `free_months` concept we added to support retention offers is no longer needed as we are changing the system from using `trial` to Stripe coupon equivalent (100% off repeating x months). This is to simplify the implementation so that we do not have weird billing metrics in Stripe and Ghost


ghost/core/test/e2e-api/members/member-offers.test.js
```diff
@@ -958,20 +956,20 @@ describe('Members API - Member Offers', function () {
...
+                discount_amount: 100,
+                duration: 'repeating',
+                duration_in_months: 1,
                 interval: cadence,
                 product_id: null,
                 currency: null,
                 active: true,
                 redemption_type: 'retention',
-                stripe_coupon_id: null // Free months offer don't rely on Stripe coupons
+                stripe_coupon_id: null
```
sagzy: I would get rid of that E2E test entirely: the previous test "successfully applies retention offer to subscription" is sufficient now that the free month offer is just UI-sugar and behaves likes a standard percent retention offer in the domain

sagzy (approved):
LGTM! Left one suggestion on removing a E2E test that is now redundant IMO

ghost/core/test/e2e-api/members/member-offers.test.js
mike182uk: Agree, removed 👍

PR #26574, opened 2026-02-24
Updated retention offers admin to account for `free_months` removal

mike182uk:
ref https://github.com/TryGhost/Ghost/pull/26571
ref https://linear.app/ghost/issue/BER-3364

With the `free_months` concept removed from the domain, we now need to update the admin to create `free months` offers using 100% off repeating x months. This changeset handles that, aswell as removing the `free months` option for the yearly retention offer

PR #26575, opened 2026-02-24
Updated portal to `free months` for 100% off x months retention offers

mike182uk:
ref #26574
ref #26571
ref https://linear.app/ghost/issue/BER-3364

With the `free_months` concept removed from the domain, we now need to update portal to show 100% off repeating x months offers as `Free Months`. This changeset handles that.


apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
```diff
@@ -203,8 +203,9 @@ function FreeTrialLabel({subscription}) {
 
 // TODO: Add i18n once copy is finalized
 function FreeMonthsLabel({nextPayment, subscription}) {
-    const months = nextPayment?.discount?.amount ?? 0;
-    const renewalDate = getDateString(subscription?.current_period_end);
+    const months = subscription?.offer?.duration_in_months ?? 0;
```
sagzy, line 206: Ah this makes me realise that we should probably have `duration_in_months` in the next payment object. That way, we have all the information needed to display next payment information in that object (and don't have to read from the subscription object). This is going to be helpful when we build the theme helper(s) for account

sagzy (approved):
LGTM, left one comment that can be part of a follow-up PR

sagzy (changes requested):
When redeeming a free-month offer, the "3 months free - Renews {date}" displays a different than than the top-level "Your subscription will renew on {date}"

CleanShot 2026-02-25 at 09 49 49@2x

mike182uk:
> When redeeming a free-month offer, the "3 months free - Renews {date}" displays a different than than the top-level "Your subscription will renew on {date}"

@sagzy Ah good catch, this was because we was using the trial logic before. I've updated so the the dates should now be in sync

sagzy (approved):

apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
sagzy: added follow-up issue: [BER-3377](https://linear.app/ghost/issue/BER-3377/add-missing-duration-in-months-to-next-payment-object)

PR #26577, opened 2026-02-24
Changed Sinon to always restore after every test

EvanHahn:
no ref

This test-only change should have no user impact.

This always calls `sinon.restore()` after every test, instead of having to remember to do it. I feel this is less error-prone, and is a little shorter.


EvanHahn:
Too tricky to get this quite right. Abandoning.

PR #26580, opened 2026-02-25
🌐 Capitalize a standout instance of “stripe” in the portal.

southp:
This PR replaces “stripe” by “Stripe” in the string: `Could not create stripe checkout session`. Since Stripe is a product name, it should be capitalized.

In the same vein, it’s also worth considering if we should also capitalize “checkout”, since [Checkout](https://stripe.com/payments/checkout) is actually also a Stripe product. I didn’t go that far in this PR, since I think referring it as a functionality also makes sense here.

cc @cathysarisky since it’s a language PR :)


southp:
Hmm, it looks like the i18n test has failed since this PR updates existing translations, and there is a `failOnUpdate` option causing that. 

Can someone help me understand what I should do instead? I’d be grateful :)

cathysarisky:
You've missed context.json.  
Looks like the CI got changed recently to enforce portal versions, but happily the error message tells you what you need to do: https://github.com/TryGhost/Ghost/actions/runs/22381462378/job/64783663882?pr=26580

southp:
Thanks for the fast response, @cathysarisky . Super helpful 🙇🏻
It turned out the i18n test was failing because I missed updating one instance of the same string that this PR is updating. It is now fixed in https://github.com/TryGhost/Ghost/pull/26580/commits/459e1dd818bbc650c937bf1755f618e8d0c4346c. The version of `portal` is also bumped by https://github.com/TryGhost/Ghost/pull/26580/commits/3700b5a6aac5ee11316b4a3f987fd752f95a05c3. All tests are now passed :)

PR #26584, opened 2026-02-25
Added offers filter to members-forward page

weylandswart:
Ref https://linear.app/ghost/issue/BER-3359/add-offers-as-a-member-filter-option

Includes the recent changes made to `/members` to account for retention offers.

### DIFF
--- a/apps/posts/src/views/members/components/members-filters.tsx
+++ b/apps/posts/src/views/members/components/members-filters.tsx
@@ -1,13 +1,70 @@
-import React from 'react';
-import {Filter, Filters, LucideIcon} from '@tryghost/shade';
+import React, {useCallback, useMemo} from 'react';
+import {Filter, FilterOption, Filters, LucideIcon} from '@tryghost/shade';
+import {Offer} from '@tryghost/admin-x-framework/api/offers';
 import {getSettingValue, useBrowseSettings} from '@tryghost/admin-x-framework/api/settings';
 import {getSiteTimezone} from '@src/utils/get-site-timezone';
 import {useBrowseConfig} from '@tryghost/admin-x-framework/api/config';
 import {useBrowseLabels} from '@tryghost/admin-x-framework/api/labels';
 import {useBrowseNewsletters} from '@tryghost/admin-x-framework/api/newsletters';
+import {useBrowseOffers} from '@tryghost/admin-x-framework/api/offers';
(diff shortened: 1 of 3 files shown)


apps/posts/src/views/members/components/members-filters.tsx
```diff
@@ -33,15 +91,82 @@ const MembersFilters: React.FC<MembersFiltersProps> = ({
...
+        return retentionOffersEnabled ? buildRetentionOfferIdMap(offers) : new Map<string, string[]>();
+    }, [offers, retentionOffersEnabled]);
+
+    const offersOptions = useMemo(() => {
+        return buildOfferOptions(offers, retentionOffersEnabled, retentionMap);
+    }, [offers, retentionOffersEnabled, retentionMap]);
+
+    // When retention grouping is active, translate between synthetic IDs (for display)
+    // and real offer IDs (stored in URL / sent to API)
+    const displayFilters = useMemo(() => {
```
rob-ghost, line 116: **Question:** This is the view layer, is this the right place to host synthetic ID expansion / collapse?

It's probably fine given the current scope but would it make more sense to live in the `use-members-filter-state` hook so its centralised and any other hook users gets the behaviour for free?

apps/posts/src/views/members/components/members-filters.tsx
```diff
@@ -33,15 +91,82 @@ const MembersFilters: React.FC<MembersFiltersProps> = ({
...
     const hasMultipleTiers = activePaidTiers.length > 1;
 
+    // Build offer options with retention grouping
+    const retentionMap = useMemo(() => {
+        return retentionOffersEnabled ? buildRetentionOfferIdMap(offers) : new Map<string, string[]>();
+    }, [offers, retentionOffersEnabled]);
+
+    const offersOptions = useMemo(() => {
+        return buildOfferOptions(offers, retentionOffersEnabled, retentionMap);
+    }, [offers, retentionOffersEnabled, retentionMap]);
```
rob-ghost, line 112: **Question:** Could we use `FilterOption.metadata` to hold the real offer ids instead of a separate map?

This would co-locate the mapping with the option itself and probably simplify the memo and callbacks into a single expansion step in `handleFiltersChange`.

rob-ghost (commented):
Forgot to submit the review before merge but there was nothing blocking. Up to you if you want to follow up on any of these questions!

PR #26590, opened 2026-02-25
Add Mailgun tag for member welcome emails

cmraible:
ref https://linear.app/ghost/issue/NY-1027/add-a-tag-to-welcome-emails-in-mailgun-to-make-filtering-easier

### Motivation
- Make welcome emails easier to filter in Mailgun by attaching a dedicated tag to those sends.
- Allow callers to pass per-message Mailgun tags so specific email types can be identified in logs and analytics.

### Description
- Added `MEMBER_WELCOME_EMAIL_TAG = 'member-welcome-email'` and exported it from `member-welcome-emails/constants.js`.
- Passed that tag from `MemberWelcomeEmailService` into the mail payload via a new `mailgunTags` message field.
- Extended `GhostMailer` to accept `mailgunTags` in `send()` and merge them with default tags via the updated `getTags(additionalTags = [])` implementation, with deduplication.
- Added/updated tests to assert that the `mailgunTags` are propagated from the welcome-email flow and merged into Mailgun `o:tag` values.

### Manual testing

I tested this out locally with Mailgun, and confirmed that I can see the tags coming through in the logs:
Screenshot 2026-02-25 at 15 15 58

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_699e5f8f682c8329b62447df8ddfee52)

### DIFF
--- a/ghost/core/core/server/services/mail/ghost-mailer.js
+++ b/ghost/core/core/server/services/mail/ghost-mailer.js
@@ -198,6 +198,11 @@ module.exports = class GhostMailer {
         return tpl(messages.messageSent);
     }
 
+    /**
+     * Builds the Mailgun tag list from defaults, site tag, and optional extra tags.
+     * @param {string[]} [additionalTags]
+     * @returns {string[]}
+     */
     getTags(additionalTags = []) {
         const tagList = [...DEFAULT_TAGS];
 
--- a/ghost/core/core/server/services/mail/ghost-mailer.js
+++ b/ghost/core/core/server/services/mail/ghost-mailer.js
@@ -63,12 +63,13 @@ function getFromAddress(requestedFromAddress, requestedReplyToAddress) {
  */
 function createMessage(message) {
     const encoding = 'base64';
-    const generateTextFromHTML = !message.forceTextContent;
+    const {forceTextContent, tags, ...cleanMessage} = message;
+    const generateTextFromHTML = !forceTextContent;
 
     const addresses = getFromAddress(message.from, message.replyTo);
 
     return {
-        ...message,
+        ...cleanMessage,
         ...addresses,
         generateTextFromHTML,
         encoding,
--- a/ghost/core/core/server/services/mail/ghost-mailer.js
+++ b/ghost/core/core/server/services/mail/ghost-mailer.js
@@ -206,7 +206,12 @@ module.exports = class GhostMailer {
         }
 
         if (Array.isArray(additionalTags) && additionalTags.length > 0) {
-            tagList.push(...additionalTags);
+            const cleanedTags = additionalTags
+                .filter(tag => typeof tag === 'string')
+                .map(tag => tag.trim())
+                .filter(tag => tag.length > 0);
+
+            tagList.push(...cleanedTags);
         }
 
         const uniqueTags = [...new Set(tagList)];
--- a/ghost/core/core/server/services/mail/ghost-mailer.js
+++ b/ghost/core/core/server/services/mail/ghost-mailer.js
@@ -118,7 +118,7 @@ module.exports = class GhostMailer {
      * @param {string} [message.replyTo]
      * @param {string} [message.from] - sender email address
      * @param {string} [message.text] - text version of this message
-     * @param {string[]} [message.mailgunTags] - optional additional Mailgun tags
+     * @param {string[]} [message.tags] - optional additional Mailgun tags
      * @param {boolean} [message.forceTextContent] - maps to generateTextFromHTML nodemailer option
      * which is: "if set to true uses HTML to generate plain text body part from the HTML if the text is not defined"
      * (ref: https://github.com/nodemailer/nodemailer/tree/da2f1d278f91b4262e940c0b37638e7027184b1d#e-mail-message-fields)
@@ -134,7 +134,7 @@ module.exports = class GhostMailer {
 
         const messageToSend = createMessage(message);
         if (this.state.usingMailgun) {
-            const tags = this.getTags(message.mailgunTags);
+            const tags = this.getTags(message.tags);
             if (tags.length > 0) {
                 messageToSend['o:tag'] = tags;
             }
--- a/ghost/core/core/server/services/mail/ghost-mailer.js
+++ b/ghost/core/core/server/services/mail/ghost-mailer.js
@@ -131,7 +131,7 @@ module.exports = class GhostMailer {
 
         const messageToSend = createMessage(message);
         if (this.state.usingMailgun) {
-            const tags = this.getTags();
+            const tags = this.getTags(message.mailgunTags);
             if (tags.length > 0) {
                 messageToSend['o:tag'] = tags;
             }
@@ -195,7 +195,7 @@ module.exports = class GhostMailer {
         return tpl(messages.messageSent);
     }
 
-    getTags() {
+    getTags(additionalTags = []) {
         const tagList = [...DEFAULT_TAGS];
 
         const siteId = config.get('hostSettings:siteId');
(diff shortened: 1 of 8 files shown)


ghost/core/core/server/services/mail/ghost-mailer.js
```diff
@@ -195,14 +195,18 @@ module.exports = class GhostMailer {
...
             tagList.push(`blog-${siteId}`);
         }
 
-        return tagList;
+        if (Array.isArray(additionalTags) && additionalTags.length > 0) {
+            tagList.push(...additionalTags);
+        }
+
+        return [...new Set(tagList)];
     }
```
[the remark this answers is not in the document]
cmraible: The limit is actually 10 tags. I've addressed this in https://github.com/TryGhost/Ghost/pull/26590/commits/8dedf8d8468db55a321cdcb6548c33adba31e2c6

ghost/core/core/server/services/mail/ghost-mailer.js
```diff
@@ -195,14 +197,29 @@ module.exports = class GhostMailer {
...
+
+        const uniqueTags = [...new Set(tagList)];
+
+        if (uniqueTags.length > MAX_MAILGUN_TAGS) {
+            const keptTags = uniqueTags.slice(0, MAX_MAILGUN_TAGS);
+            const droppedTags = uniqueTags.slice(MAX_MAILGUN_TAGS);
+
+            logging.warn(`[MAIL] Mailgun tag count exceeded ${MAX_MAILGUN_TAGS}; truncating tags. Kept: ${keptTags.join(', ')}. Dropped: ${droppedTags.join(', ')}`);
+
+            return keptTags;
```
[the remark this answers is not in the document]
cmraible: I don't see this being an issue, we wouldn't use any sensitive values in a tag

ghost/core/core/server/services/mail/ghost-mailer.js
```diff
@@ -195,14 +198,34 @@ module.exports = class GhostMailer {
...
 
-        return tagList;
+        if (Array.isArray(additionalTags) && additionalTags.length > 0) {
+            const cleanedTags = additionalTags
+                .filter(tag => typeof tag === 'string')
+                .map(tag => tag.trim())
+                .filter(tag => tag.length > 0);
+
+            tagList.push(...cleanedTags);
+        }
```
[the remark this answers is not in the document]
cmraible: I'm not too concerned about hitting the 128 character limit or including non-ascii characters

ghost/core/test/unit/server/services/mail/ghost-mailer.test.js
```diff
@@ -400,6 +401,66 @@ describe('Mail: Ghostmailer', function () {
...
+                html: 'content',
+                tags: ['member-welcome-email']
+            });
+
+            const sentMessage = sendMailSpy.firstCall.args[0];
+            assert(sentMessage['o:tag'].includes('transactional-email'));
+            assert(sentMessage['o:tag'].includes('member-welcome-email'));
+        });
+
+        it('should truncate tags to Mailgun maximum and log warning', async function () {
```
troyciesco, line 428: Truncation test assertion could be tighter (nit)

  In the truncation test, the assertion checks that tags 8 and 9 are missing:
  assert.equal(sentMessage['o:tag'].includes('tag-8'), false);
  assert.equal(sentMessage['o:tag'].includes('tag-9'), false);

  Consider also asserting that exactly the expected 10 tags are present (the 3 defaults + first 7 additional), to ensure the ordering is correct:
  assert.deepEqual(sentMessage['o:tag'], [
      'ghost-email', 'transactional-email', 'blog-123123',
      'tag-1', 'tag-2', 'tag-3', 'tag-4', 'tag-5', 'tag-6', 'tag-7'
  ]);

ghost/core/core/server/services/mail/ghost-mailer.js
```diff
@@ -131,7 +134,7 @@ module.exports = class GhostMailer {
 
         const messageToSend = createMessage(message);
         if (this.state.usingMailgun) {
-            const tags = this.getTags();
+            const tags = this.getTags(message.tags);
```
troyciesco, line 140: tags leaks through createMessage into the transport payload

The createMessage function (line 64-79):

  function createMessage(message) {
      const encoding = 'base64';
      const generateTextFromHTML = !message.forceTextContent;
      const addresses = getFromAddress(message.from, message.replyTo);

      return {
          ...message,    // <-- spreads EVERYTHING from the input
          ...addresses,
          generateTextFromHTML,
          encoding,
          headers: { Sender: addresses.from }
      };
  }

  The key is ...message — this copies every property from the input object into the returned object.

  The send method (line 127-153):

  async send(message) {
      // message = { to, subject, html, text, forceTextContent, tags: ['member-welcome-email'], ...senderOptions }

      const messageToSend = createMessage(message);
      // messageToSend now has EVERYTHING from message, including:
      //   tags: ['member-welcome-email']     <-- raw array, carried over from the spread

      if (this.state.usingMailgun) {
          const tags = this.getTags(message.tags);
          messageToSend['o:tag'] = tags;
          // messageToSend now ALSO has:
          //   'o:tag': ['ghost-email', 'transactional-email', 'blog-xyz', 'member-welcome-email']
      }

      const response = await this.sendMail(messageToSend);
      // ...
  }

  So what gets sent to transport.sendMail():

  {
      to: 'user@example.com',
      subject: 'Welcome!',
      html: '...',
      text: '...',
      forceTextContent: true,
      tags: ['member-welcome-email'],          // <-- stowaway from ...message spread
      'o:tag': ['ghost-email', 'transactional-email', 'blog-xyz', 'member-welcome-email'],  // <-- the real one Mailgun reads
      from: '...',
      replyTo: '...',
      generateTextFromHTML: false,
      encoding: 'base64',
      headers: { Sender: '...' }
  }

  tags is Ghost's internal field name. o:tag is Mailgun's actual field name. Both end up on the same object passed to the transport.

  Why it's harmless: Mailgun's API (and nodemailer transports in general) ignore properties they don't recognize. tags isn't a Mailgun field — it looks at o:tag. So the stray tags property just gets silently
  ignored.

  Why it's still worth noting: It's a leaky abstraction. An internal concern (tags as input to getTags) bleeds into the wire-level payload. If a future transport or middleware ever interprets a tags property, it
  could cause unexpected behavior. The same issue already exists with forceTextContent — it's also a Ghost-internal flag that leaks through the spread. So this isn't a new pattern, just one more instance of it.

  The fix would be minimal — either destructure out internal-only fields before spreading, or delete them after:

  const { tags, forceTextContent, ...rest } = message;
  return { ...rest, ...addresses, generateTextFromHTML, encoding, headers: { ... } };

troyciesco (changes requested):

ghost/core/core/server/services/mail/ghost-mailer.js
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/26590/commits/5fdf41ca0bce7dc31b205ceb9ce1f5cf61da85a8

ghost/core/test/unit/server/services/mail/ghost-mailer.test.js
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/26590/commits/4728f6e7acca01192a9e88b2df2bb98a2adb00d6

Switched this assertion to a deepEqual

cmraible:
Also sanitized the logs to not include the values of the tags that have been truncated, but instead just a count: https://github.com/TryGhost/Ghost/pull/26590/commits/87440916cc9b79677380d383c0e774883bfc1c1c

troyciesco (approved):
approved for whenever CI passes

cmraible:
There was one more snapshot test in our unit tests that included `forceTextContent` in the payload to mailgun, so I updated that. The other test failure was a flaky test that has since been fixed on main, so I also rebased and force pushed.

PR #26591, opened 2026-02-25
Fixed outdated types when building admin-x-framework

EvanHahn:
towards https://linear.app/ghost/issue/NY-1097

The `nx` cache needs to be more properly purged for some of these projects, otherwise we get [CI failures][0]. This defines what files are part of the output, so it can properly cache them.

[0]: https://github.com/TryGhost/Ghost/actions/runs/22409070748/job/64877688336?pr=26569


EvanHahn:
Addressed @coderabbit's comments in #26596.

EvanHahn:
I think this is no longer necessary. Closing.

PR #26599, opened 2026-02-25
Converted initial outbox logs to structured logging

cmraible:
refs https://linear.app/ghost/issue/NY-1020/convert-outbox-logs-to-structured-logs

## Summary

- converted the outbox missing-slug warning and outbox already-running guard logs to structured logging with a top-level `system` object and message second argument
- added red/green unit coverage that captures real `@tryghost/logging` Bunyan JSON output in-memory and asserts structured fields
- extracted shared test log capture utility (`ghost/core/test/utils/logging-utils.js`) with JSDoc and updated outbox unit tests to use top-level capture hooks

## Reviewer notes

Mostly I'm looking for feedback on the new testing pattern introduced here. The new `captureLoggerOutput()` test helper redirects the actual JSON logs emitted by @tryghost/logging into an in-memory stream. This way we can assert on the _actual_ JSON that is output by the logger, which feels more useful to me than asserting on which arguments we pass to the logger.

This doesn't cover all the logs that I want to change to structured logs for the referenced issue, but I wanted to align on the pattern here while keeping the PR relatively small; once we're aligned on the approach I can extend it to the other outbox/welcome email logs in a follow up PR.

### DIFF
--- a/ghost/core/core/server/services/outbox/handlers/member-created.js
+++ b/ghost/core/core/server/services/outbox/handlers/member-created.js
@@ -12,7 +12,12 @@ async function handle({payload}) {
     try {
         const slug = MEMBER_WELCOME_EMAIL_SLUGS[payload.status];
         if (!slug) {
-            logging.warn(`${LOG_KEY} No automated email slug found for member status: ${payload.status}`);
+            logging.warn({
+                system: {
+                    event: 'outbox.member_created.no_slug_mapping',
+                    member_status: payload.status
+                }
+            }, 'No automated email slug found for member status');
             return;
         }
 
--- a/ghost/core/core/server/services/outbox/handlers/member-created.js
+++ b/ghost/core/core/server/services/outbox/handlers/member-created.js
@@ -17,7 +17,7 @@ async function handle({payload}) {
                     event: 'outbox.member_created.no_slug_mapping',
                     member_status: payload.status
                 }
-            }, 'No automated email slug found for member status');
+            }, `${LOG_KEY} No automated email slug found for member status`);
             return;
         }
 
@@ -40,7 +40,7 @@ async function handle({payload}) {
                 event: 'outbox.member_created.track_send_failed'
             },
             err
-        }, 'Failed to track automated email send');
+        }, `${LOG_KEY} Failed to track automated email send`);
     }
 }
 
--- a/ghost/core/core/server/services/outbox/handlers/member-created.js
+++ b/ghost/core/core/server/services/outbox/handlers/member-created.js
@@ -36,9 +36,11 @@ async function handle({payload}) {
         });
     } catch (err) {
         logging.error({
-            message: `${LOG_KEY} Failed to track automated email send`,
+            system: {
+                event: 'outbox.member_created.track_send_failed'
+            },
             err
-        });
+        }, 'Failed to track automated email send');
     }
 }
 
--- a/ghost/core/test/unit/server/services/outbox/handlers/member-created.test.js
+++ b/ghost/core/test/unit/server/services/outbox/handlers/member-created.test.js
@@ -2,7 +2,7 @@ const assert = require('node:assert/strict');
 const sinon = require('sinon');
 const rewire = require('rewire');
 const logging = require('@tryghost/logging');
-const {captureLoggerOutput} = require('../../../../../utils/logging-utils');
+const {captureLoggerOutput, findByEvent} = require('../../../../../utils/logging-utils');
 
 describe('member-created handler', function () {
     let handler;
@@ -90,7 +90,7 @@ describe('member-created handler', function () {
         });
 
         sinon.assert.calledOnce(memberWelcomeEmailServiceStub.api.send);
-        const warningLog = logCapture.output.find(log => log.level === 40 && log.msg === 'No automated email slug found for member status');
+        const warningLog = findByEvent(logCapture.output, 'outbox.member_created.no_slug_mapping');
         assert.ok(warningLog);
         assert.deepEqual(warningLog.system, {
             event: 'outbox.member_created.no_slug_mapping',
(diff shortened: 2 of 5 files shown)


ghost/core/test/unit/server/services/outbox/handlers/member-created.test.js
```diff
@@ -31,13 +27,15 @@ describe('member-created handler', function () {
             add: sinon.stub().resolves()
         };
 
+        logCapture = captureLoggerOutput();
```
cmraible, line 30: Redirects logs to an in-memory stream where we collect all the logs from this test.

ghost/core/test/unit/server/services/outbox/handlers/member-created.test.js
```diff
@@ -55,7 +53,9 @@ describe('member-created handler', function () {
         });
 
         sinon.assert.calledOnce(memberWelcomeEmailServiceStub.api.send);
-        sinon.assert.calledOnce(loggingStub.error);
+        const errorLog = findByEvent(logCapture.output, 'outbox.member_created.track_send_failed');
+        assert.ok(errorLog);
+        assert.ok(errorLog.msg.startsWith('Failed to track automated email send'));
```
cmraible: Queries the captured logs by event, then asserts on the actual JSON from the log

ghost/core/test/unit/server/services/outbox/handlers/member-created.test.js
```diff
@@ -31,13 +27,15 @@ describe('member-created handler', function () {
...
+        logCapture = captureLoggerOutput();
+
         handler.__set__('memberWelcomeEmailService', memberWelcomeEmailServiceStub);
-        handler.__set__('logging', loggingStub);
         handler.__set__('AutomatedEmail', AutomatedEmailStub);
         handler.__set__('AutomatedEmailRecipient', AutomatedEmailRecipientStub);
     });
 
     afterEach(function () {
+        logCapture.restore();
```
cmraible, line 38: Resets to logger to its original settings

ghost/core/test/utils/logging-utils.js
```diff
@@ -0,0 +1,96 @@
...
+            })
+        }
+    };
+
+    return {
+        output,
+        restore() {
+            buffered = parseBufferedJsonLogs(buffered, output);
+            logging.streams = originalStreams;
+            stream.destroy();
```
troyciesco, line 67: Potential stream race condition in restore()

  In logging-utils.js:62-65, restore() flushes the buffer and then immediately destroys the stream:

  restore() {
      buffered = parseBufferedJsonLogs(buffered, output);
      logging.streams = originalStreams;
      stream.destroy();
  }

  The data event handler on the PassThrough stream is async — there could be buffered data in the stream that hasn't fired its data event yet when restore() is called. Consider draining with stream.read() before
  destroying, or calling stream.end() (which flushes) instead of stream.destroy() (which does not).

  In practice this is likely fine for synchronous test code, but worth noting for robustness.

ghost/core/test/utils/logging-utils.js
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Find the first structured log record matching a system event.
+ *
+ * @param {Array<object>} output - Captured log records.
+ * @param {string} event - Structured event name to match.
+ * @returns {object|undefined} First matching log record.
+ */
+function findByEvent(output, event) {
```
troyciesco, line 79: findAllByEvent is exported but unused

ghost/core/test/unit/server/services/outbox/handlers/member-created.test.js
```diff
@@ -72,10 +72,10 @@ describe('member-created handler', function () {
             }
         });
 
-        sinon.assert.calledOnce(loggingStub.error);
-        const errorCall = loggingStub.error.getCall(0);
-        assert.ok(errorCall.args[0].message.includes('Failed to track automated email send'));
-        assert.equal(errorCall.args[0].err, dbError);
+        const errorLog = findByEvent(logCapture.output, 'outbox.member_created.track_send_failed');
+        assert.ok(errorLog);
+        assert.ok(errorLog.msg.startsWith('Failed to track automated email send'));
```
troyciesco: better to not assert the message in a test (here and elsewhere)

ghost/core/core/server/services/outbox/handlers/member-created.js
```diff
@@ -12,7 +12,12 @@ async function handle({payload}) {
     try {
         const slug = MEMBER_WELCOME_EMAIL_SLUGS[payload.status];
         if (!slug) {
-            logging.warn(`${LOG_KEY} No automated email slug found for member status: ${payload.status}`);
+            logging.warn({
+                system: {
+                    event: 'outbox.member_created.no_slug_mapping',
+                    member_status: payload.status
+                }
+            }, 'No automated email slug found for member status');
```
troyciesco: consider adding the log key back so it's easier to see stuff when developing locally. also consider tweaking the log key to more closely match the event above

ghost/core/test/utils/logging-utils.js
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Temporarily redirects Ghost logger streams to an in-memory Bunyan stream.
+ *
+ * This allows tests to assert on real serialized JSON output from
+ * `@tryghost/logging` without stubbing logger methods.
+ *
+ * @returns {{output: Array<object>, restore: () => void}} Capture handle.
+ */
+function captureLoggerOutput() {
```
troyciesco, line 35: captureLoggerOutput doesn't preserve serializers

The replacement bunyan logger in the capture utility doesn't use the same serializers as the original Ghost logger. This means the captured JSON may differ from production output for objects that have custom serialization (like err objects). For the current test cases this doesn't matter, but as this utility gets used more broadly it could cause subtle differences between test and production log output. Worth a comment or future enhancement.

troyciesco (changes requested):

troyciesco (approved):

ghost/core/test/utils/logging-utils.js
cmraible: Removed!

ghost/core/test/utils/logging-utils.js
cmraible: I did not address this because I don't think it's a problem we'll actually have, but it's good to have the note here in case we do find that this causes problems.

PR #26609, opened 2026-02-26
Added import to members forward page

weylandswart:
Ref https://linear.app/ghost/issue/BER-3381/add-bulk-member-importing-flow

Somewhere in the original port, importing members was lost. This adds it back.


apps/posts/src/views/members/components/bulk-action-modals/import-members/field-mapping.ts
```diff
@@ -0,0 +1,273 @@
...
+    for (let i = 1; i < rows.length; i++) {
+        const obj: Record<string, string> = {};
+        headers.forEach((header, j) => {
+            obj[header] = rows[i][j] || '';
+        });
+        data.push(obj);
+    }
+
+    return data;
+}
```
kevinansfield, line 251: Is Ember using a CSV parsing library or does it also implement CSV parsing itself? Lets use a library if possible

PR #26610, opened 2026-02-26
Add legacy members filter conversion support

jonatansberg:
When we release the new members list, we want to continue to support the old filter syntax in the URL so that bookmarked versions of the member list will continue to work.

The old filter syntax is converted to the new one, and raw NQL is used as a fallback when a straightforward conversion isn't possible.

### DIFF
--- a/apps/posts/package.json
+++ b/apps/posts/package.json
@@ -40,6 +40,7 @@
     },
     "dependencies": {
         "@tryghost/admin-x-framework": "0.0.0",
+        "@tryghost/nql-lang": "0.6.4",
         "@tryghost/shade": "0.0.0",
         "i18n-iso-countries": "7.14.0",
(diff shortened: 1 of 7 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Run unit tests`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22713231709)
**What failed:** Coverage for some modules does not meet global threshold
**Why:** The logs show that the coverage for some modules in the Ghost project fell below the global coverage threshold, causing the unit tests to fail. This is a code issue that the author should address by improving test coverage for the affected modules.
**Action:**
The author should review the coverage report and add more tests to improve coverage for the modules that did not meet the threshold.

jonatansberg:
No longer needed after #26828

PR #26617, opened 2026-02-26
Added saved views for filtered members

weylandswart:
Ref https://linear.app/ghost/issue/BER-3346/ability-to-save-static-filters

Allows users to filter the members page and then save the result of that filtering as a custom view.


apps/admin/src/layout/app-sidebar/nav-custom-views.tsx
```diff
@@ -32,9 +32,39 @@ interface NavCustomViewsProps {
...
+                />
+            </NavMenuItem.Link>
+        </NavMenuItem>
+    );
+}
+
 export function NavCustomViews({ route = 'posts' }: NavCustomViewsProps) {
     const { data: settingsData } = useBrowseSettings();
     const routing = useEmberRouting();
+    const [showAll, setShowAll] = useState(false);
```
kevinansfield: Discussed pulling out the show all collapse. We'll remove for now. Also noted this isn't behind a labs flag and would impact the custom posts views list.

apps/admin/src/layout/app-sidebar/hooks/use-member-views.ts
```diff
@@ -0,0 +1,33 @@
...
+const memberViewSchema = z.object({
+    name: z.string(),
+    route: z.string(),
+    filter: z.record(z.string(), z.string().nullable())
+});
+const sharedViewsSchema = z.array(memberViewSchema);
+
+export type MemberView = z.infer<typeof memberViewSchema>;
+
+export function useMemberViews() {
```
kevinansfield: There's some questions around the duplication with the same data structure being used in `NavCustomViews` and how `NavCustomViews` naming is now a little odd being so generic.

apps/admin/src/layout/app-sidebar/nav-content.tsx
```diff
@@ -126,10 +131,56 @@ function NavContent({ ...props }: React.ComponentProps<typeof SidebarGroup>) {
                         </NavMenuItem>
                     )}
 
-                    {showMembers && (
+                    {showMembers && membersForwardEnabled && (
+                        <>
+                            <NavMenuItem>
```
kevinansfield, line 136: Discussed duplication with the posts custom views list. Decided it was fine, this list will have enough variation by not having colours and potentially having a count for each view.

apps/admin/src/layout/app-sidebar/nav-member-views.tsx
```diff
@@ -0,0 +1,118 @@
...
+    const currentKeys = Array.from(currentParams.keys()).sort();
+    const viewKeys = Array.from(viewParams.keys()).sort();
+
+    if (currentKeys.length !== viewKeys.length) {
+        return false;
+    }
+
+    return currentKeys.every((key, i) =>
+        key === viewKeys[i] && currentParams.get(key) === viewParams.get(key)
+    );
```
kevinansfield: This logic is a result of us not using NQL in the query params for React but storing that as the filter. For Posts this logic isn't needed because Ember still uses NQL in the query params but will probably need to be extracted at some point.

There may also be some overlap with @jonatansberg's work on mapping existing Ember member list bookmarks to the React screen (#26610)

apps/posts/src/views/members/components/manage-view-popover.tsx
```diff
@@ -0,0 +1,166 @@
...
+interface SaveViewPopoverProps {
+    filters: Filter[];
+    existingViews: MemberView[];
+    activeView?: MemberView | null;
+    onDeleted?: () => void;
+}
+
+const SaveViewPopover: React.FC<SaveViewPopoverProps> = ({filters, existingViews, activeView, onDeleted}) => {
+    const [open, setOpen] = useState(false);
+    const [name, setName] = useState('');
```
kevinansfield, line 15: Should we have separate state for the name, or does it make more sense to be editing a view instance?

apps/posts/src/views/members/hooks/use-member-views.ts
```diff
@@ -0,0 +1,174 @@
...
+
+// Schema for a single view (color optional for member views)
+const memberViewSchema = z.object({
+    name: z.string(),
+    route: z.string(),
+    color: z.string().optional(),
+    icon: z.string().optional(),
+    filter: z.record(z.string(), z.string().nullable())
+});
+
```
kevinansfield, line 15: This schema is duplicated with custom views.

`color` and `icon` aren't settable in the UI for member views. `icon` isn't settable/used for either.

apps/admin/src/layout/app-sidebar/hooks/use-member-views.ts
```diff
@@ -0,0 +1,33 @@
+import { useMemo } from 'react';
+import { z } from 'zod';
+import { useBrowseSettings, getSettingValue } from '@tryghost/admin-x-framework/api/settings';
+
+const memberViewSchema = z.object({
+    name: z.string(),
+    route: z.string(),
+    filter: z.record(z.string(), z.string().nullable())
+});
+const sharedViewsSchema = z.array(memberViewSchema);
```
kevinansfield: There's some odd duplication here with the changes in `posts/src/views/members/hooks/use-member-views.ts`

apps/posts/src/views/members/hooks/use-member-views.ts
kevinansfield: Should probably live as a top-level hook so schemas and utils can be shared across sidebar and members screen.

apps/posts/src/views/members/hooks/use-member-views.ts
```diff
@@ -0,0 +1,174 @@
...
+        const parsed: unknown = JSON.parse(json);
+        const result = sharedViewsSchema.safeParse(parsed);
+        if (!result.success) {
+            return [];
+        }
+        return result.data;
+    } catch {
+        return [];
+    }
+}
```
kevinansfield, line 82: There's duplication here with the logic in `NavCustomViews`

ErisDS:
- Workflow: `CI`
- Failed Step: `Build TS packages`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22580117825)
**What failed:** TypeScript type errors
**Action:**

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -55,6 +56,7 @@ export function useUserPreferences<TData = Preferences>(
             return PreferencesSchema.parse(parsed);
         },
         enabled: !!user,
+        keepPreviousData: true,
```
rob-ghost, line 59: FYI `keepPreviousData` is now [deprecated](https://tanstack.com/query/v5/docs/framework/react/guides/migrating-to-v5#removed-keeppreviousdata-in-favor-of-placeholderdata-identity-function) in the next major version of Tanstack Query.

apps/posts/src/views/members/hooks/use-member-views.ts
rob-ghost: Agreed, this is duplicated in `layout/app-sidebar/hooks/use-member-views.ts` with subtle differences. We should lift this hook or schema and share it across.

apps/admin/src/layout/app-sidebar/nav-member-views.tsx
```diff
@@ -0,0 +1,88 @@
...
+            params.set(key, value);
+        }
+    }
+    return params.toString();
+}
+
+/**
+ * Check if the current URL search params match a view's filter
+ */
+function isViewActive(currentSearch: string, viewFilter: Record<string, string | null>): boolean {
```
rob-ghost: **Suggestion:** These two helper functions re-create `filterRecordToSearchParams` and `isFilterEqual` from the posts app hook under different names.

This, along with `use-member-views.ts` duplication is hinting that we need to lift some things up to be shared across containers, perhaps grouped under a `members-views` domain?

apps/posts/src/views/members/hooks/use-member-views.ts
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * Convert Filter[] (UI state) to a record for storage
+ * e.g., [{field: 'status', operator: 'is', values: ['paid']}] → {status: 'is:paid'}
+ */
+export function filtersToRecord(filters: Filter[]): Record<string, string> {
+    const record: Record<string, string> = {};
+    for (const filter of filters) {
+        if (filter.values[0] !== undefined) {
+            record[filter.field] = `${filter.operator}:${String(filter.values[0])}`;
```
rob-ghost: **Question:** Do filters ever have more than one `value`? If they do, this would only take the first one.

So if we had `values: [free, paid]` it would only take `free`.

It might be this never happens, but I'd be lead to believe it could from the dostring and the attribute names.

apps/posts/src/views/members/hooks/use-member-views.ts
```diff
@@ -0,0 +1,174 @@
...
+ * Hook to delete a member view
+ */
+export function useDeleteMemberView() {
+    const {data: settingsData} = useBrowseSettings();
+    const {mutateAsync: editSettings} = useEditSettings();
+
+    const deleteView = useCallback(async (view: MemberView) => {
+        const allViews = parseSharedViews(settingsData);
+        const updatedViews = allViews.filter(v => !(v.route === 'members' && v.name === view.name && isFilterEqual(v.filter, view.filter))
+        );
```
rob-ghost: **Issue:** There's some inconsistency around how we identify views when saving or deleting them.

`useSaveMemberView` matches existing views by filter equality (same filter = update, different filter = create). But `useDeleteMemberView` matches by `name + filter`. And the popover's duplicate check is by `name` only.

This means: if I save a view "Paid members" with `status:paid`, then change the filters to `status:free` and hit "Save view" with the same name, the hook throws "A view with this name already exists." But if I change filters slightly and use a different name, it creates a new view, even though the user might have expected to update the existing one. The identity model is muddled between "a view is identified by its name" vs "a view is identified by its filter."

We should consider picking a consistent way to identify saved views and use that in all cases.

apps/posts/src/views/members/hooks/use-member-views.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    return save;
+}
+
+/**
+ * Hook to delete a member view
+ */
+export function useDeleteMemberView() {
+    const {data: settingsData} = useBrowseSettings();
+    const {mutateAsync: editSettings} = useEditSettings();
```
rob-ghost, line 153: **Issue:** Both `useSaveMemberView` and `useDeleteMemberView` use `useBrowseSettings` and `useEditSettings` which does a full re-write of the json settings value. This can cause consistency problems especially if we're open in two different tabs or in two different sessions trying to edit views at the same time.

One user's request can be easily over-written by another if there's a race or the caches are out of sync, which seems likely.

This is an existing problem with the posts views too I think so it might not be worth fixing right now, just want to point it out.

apps/posts/src/views/members/components/manage-view-popover.tsx
```diff
@@ -0,0 +1,166 @@
...
+            if (!isOpen) {
+                setName('');
+                setError('');
+            }
+        }}>
+            <PopoverTrigger asChild>
+                <Button
+                    variant="outline"
+                >
+                    {isEditing ? 'Edit view' : 'Save view'}
```
rob-ghost, line 103: **Suggestion(if-minor):** I wonder should we have two distinct modals, one for saving a new view and one for editing an existing view such that we have one top level conditional and not sprinkled among the content:

```typescript
isEditing ? <EditViewPopover/> : <SaveViewPopover/>
```

Having both under the name `SaveViewPopover` is not initially clear and it removes the decision making from the body of the component.

apps/posts/src/views/members/components/save-view-popover.tsx
```diff
@@ -0,0 +1,166 @@
...
+                        <div className="flex items-center justify-between">
+                            <Button
+                                className="text-red hover:bg-red/5 hover:text-red"
+                                disabled={deleting}
+                                size="sm"
+                                variant="ghost"
+                                onClick={() => void handleDelete()}
+                            >
+                                {deleting ? 'Deleting...' : 'Delete'}
+                            </Button>
```
rob-ghost: **Suggestion:** Delete is a single click, with something destructive like this do we want any kind of confirmation modal? "Are you sure you want to delete this view?"

rob-ghost (commented):
I think there's some universal truth to lifting the member concerns to their own higher level domain (directory) and sharing them down to avoid the duplication we have between hooks and the potential drift between them.

I think we can do another pass on this one to tighten it up.

kevinansfield:
Superseded by #26681

PR #26196, opened 2026-02-03
Removed `announcementBar` feature flag

EvanHahn:
ref f9f5f72752722ae835efbe871e1c9c7a4359a827

*This change should have no user impact.*

The `announcementBar` feature flag has been enabled since 2023. We can remove the flag and assume it's always on.

---

> [!NOTE]
> **Low Risk**
> Small flag-removal change that mainly deletes gating and updates tests; risk is limited to unintentionally exposing `/api/announcement` where it was previously blocked by misconfiguration.
> 
> **Overview**
> Removes the `announcementBar` labs feature flag from the admin feature service and from shared labs GA flag handling.
> 
> The members `/api/announcement` route is no longer protected by `labs.enabledMiddleware('announcementBar')`, and related test fixtures/snapshots are updated to stop expecting or setting the flag in config/labs payloads.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f61501548cdab44fdceedac806783dd65fd56eef. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #26622, opened 2026-02-26
Updated labs tests to not rely on specific flags

troyciesco:
no ref

- a test for labs ensures that config values take precedence over GA keys. However, the test referenced the `announcementBar` flag was removed in https://github.com/TryGhost/Ghost/pull/26196. That means this test was looking at a state that wasn't really true
- instead, we can check for the first item in the `GA_KEYS` list. If there's nothing in the list, the test can be skipped.
- a few other tests had this same issue, so this PR updates them as well
- there was a test for alpha flags that wasn't actually doing anything, so it's deleted

---

> [!NOTE]
> **Low Risk**
> Test-only changes that reduce brittleness by avoiding references to specific feature flags; no production logic is modified.
> 
> **Overview**
> Updates `labs.test.js` to stop hard-coding specific labs flag names and instead derive test flags from `labs.WRITABLE_KEYS_ALLOWLIST` and `labs.GA_KEYS`, skipping tests when those lists are empty.
> 
> Removes a brittle/ineffective alpha-flag test and adjusts remaining assertions to validate precedence rules (config overrides settings/GA) using dynamically selected flags.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ef1b59286e2a4778136c749f90f7ebc64fa2be5e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/core/test/unit/shared/labs.test.js
+++ b/ghost/core/test/unit/shared/labs.test.js
@@ -28,54 +28,47 @@ describe('Labs Service', function () {
         }));
     });
 
-    it('returns an alpha flag when dev experiments in toggled', function () {
-        configUtils.set('enableDeveloperExperiments', true);
-        sinon.stub(process.env, 'NODE_ENV').value('production');
-        const getSpy = sinon.stub(settingsCache, 'get');
-        getSpy.withArgs('labs').returns({
-            urlCache: true
-        });
-
-        // NOTE: this test should be rewritten to test the alpha flag independently of the internal ALPHA_FEATURES list
-        //       otherwise we end up in the endless maintenance loop and need to update it every time a feature graduates from alpha
-        assert.deepEqual(labs.getAll(), expectedLabsObject({
-            urlCache: true,
-            members: true
-        }));
-
-        assert.equal(labs.isSet('members'), true);
-        assert.equal(labs.isSet('urlCache'), true);
-    });
-
     it('respects the value in config over settings', function () {
+        if (labs.WRITABLE_KEYS_ALLOWLIST.length === 0) {
+            this.skip();
+        }
+
+        const flag = labs.WRITABLE_KEYS_ALLOWLIST[0];
+
         configUtils.set('labs', {
-            collections: false
+            [flag]: false
         });
         const getSpy = sinon.stub(settingsCache, 'get');
         getSpy.withArgs('labs').returns({
-            collections: true,
+            [flag]: true,
             members: true
         });
 
         assert.deepEqual(labs.getAll(), expectedLabsObject({
-            collections: false,
+            [flag]: false,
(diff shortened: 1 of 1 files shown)


ghost/core/test/unit/shared/labs.test.js
```diff
@@ -28,54 +28,47 @@ describe('Labs Service', function () {
...
-        getSpy.withArgs('labs').returns({
-            urlCache: true
-        });
+    it('respects the value in config over settings', function () {
+        if (labs.WRITABLE_KEYS_ALLOWLIST.length === 0) {
+            this.skip();
+        }
 
-        // NOTE: this test should be rewritten to test the alpha flag independently of the internal ALPHA_FEATURES list
-        //       otherwise we end up in the endless maintenance loop and need to update it every time a feature graduates from alpha
```
troyciesco, line 40: i *think* this test wasn't doing anything. `enableDeveloperExperiments` doesn't seem to change anything about what labs are returned, i think it just controls whether the tab for editing `PRIVATE_FEATURES` shows up in the UI.

to further check, i create the inverse of this test ("does not return an alpha flag when dev experiments is false") and it failed.

based on how labs works with `getAll()` returning all beta and private keys, I don't think a test is needed here.

EvanHahn:
I'll plan to review this next week. Let me know if you want me to look sooner.

ghost/core/test/unit/shared/labs.test.js
```diff
@@ -28,54 +28,47 @@ describe('Labs Service', function () {
...
 
-    it('returns an alpha flag when dev experiments in toggled', function () {
-        configUtils.set('enableDeveloperExperiments', true);
-        sinon.stub(process.env, 'NODE_ENV').value('production');
-        const getSpy = sinon.stub(settingsCache, 'get');
-        getSpy.withArgs('labs').returns({
-            urlCache: true
-        });
+    it('respects the value in config over settings', function () {
+        if (labs.WRITABLE_KEYS_ALLOWLIST.length === 0) {
```
EvanHahn, line 32: praise: I like this. Keeps us from having to change this test if we ever have nothing in the list.

ghost/core/test/unit/shared/labs.test.js
```diff
@@ -28,54 +28,47 @@ describe('Labs Service', function () {
...
+            this.skip();
```
EvanHahn, line 33: nit: we should `return` after skipping.

```suggestion
            this.skip();
            return;
```

The same comment applies to the other skips.

EvanHahn (approved):
LGTM other than one non-blocking comment.

PR #26624, opened 2026-02-26
⚡ Optimized comment count subquery to avoid full table scan

9larsons:
ref https://linear.app/ghost/issue/ONC-1509/

The `OR` condition between `parent_id` and `in_reply_to_id` columns in the `direct_replies` count subquery defeats MySQL index usage, causing a full table scan per comment row. Split into two separate indexed COUNT subqueries summed together, allowing MySQL to use the existing foreign key indexes.

This could easily end up reading several million rows for a single query to get the top 20 comments+replies.

### DIFF
--- a/ghost/core/core/server/models/comment.js
+++ b/ghost/core/core/server/models/comment.js
@@ -265,22 +265,21 @@ const Comment = ghostBookshelf.Model.extend({
             },
             direct_replies(modelOrCollection, options) {
                 const excludedCommentStatuses = options.isAdmin ? ['deleted'] : ['hidden', 'deleted'];
+                const statusPlaceholders = excludedCommentStatuses.map(() => '?').join(',');
 
-                modelOrCollection.query('columns', 'comments.*', (qb) => {
-                    qb.count('replies.id')
-                        .from('comments AS replies')
-                        .where(function () {
-                            // Root comments: count direct replies (parent_id = this, in_reply_to_id IS NULL)
-                            this.where(function () {
-                                this.whereRaw('replies.parent_id = comments.id')
-                                    .whereNull('replies.in_reply_to_id');
-                            })
-                                // Child comments: count replies-to-this-child (in_reply_to_id = this)
-                                .orWhereRaw('replies.in_reply_to_id = comments.id');
-                        })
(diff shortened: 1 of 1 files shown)


9larsons:
## EXPLAIN before/after (15K comments, dev dataset)

**Before** — OR condition defeats index usage, full table scan per comment:

| table | type | key | rows | Extra |
|-------|------|-----|------|-------|
| replies | **ALL** | NULL | **15,369** | Range checked for each record |

**After** — two separate indexed subqueries:

| table | type | key | rows | Extra |
|-------|------|-----|------|-------|
| r1 | **ref** | comments_parent_id_foreign | **17** | Using where |
| r2 | **ref** | comments_in_reply_to_id_foreign | **4** | Using where |

**~730x fewer rows scanned per comment** (15,369 → ~21).

For a page load on the busiest post (40 comment objects needing `count__direct_replies`): **614,760 → 840 total rows scanned**. Both `/members/api/comments/post/:id/` and `/ghost/api/admin/comments/post/:id/` go through the same model code path.

On the reported production dataset (77K comments), this would reduce from ~8.9M row scans per request to ~2,400.

ghost/core/core/server/models/comment.js
```diff
@@ -265,22 +265,21 @@ const Comment = ghostBookshelf.Model.extend({
...
-                                // Child comments: count replies-to-this-child (in_reply_to_id = this)
-                                .orWhereRaw('replies.in_reply_to_id = comments.id');
-                        })
-                        .whereNotIn('replies.status', excludedCommentStatuses)
-                        .as('count__direct_replies');
-                });
+                const statusPlaceholders = excludedCommentStatuses.map(() => '?').join(',');
+
+                // Split into two separate indexed subqueries instead of a single OR-based query.
+                // The OR between parent_id and in_reply_to_id defeats MySQL index usage,
```
EvanHahn: **nit:** This implies the existence of an `OR` that doesn't exist.

```suggestion
                // An OR between parent_id and in_reply_to_id defeats MySQL index usage,
```

EvanHahn (approved):
Love having to work around the query planner...

9larsons:
It's a harsh mistress.

PR #26626, opened 2026-02-26
Used shared Koenig file upload hook

EvanHahn:
closes https://linear.app/ghost/issue/NY-1097

This change should have no user impact. It's just a cleanup.

### DIFF
--- a/apps/admin-x-design-system/src/utils/format-url.ts
+++ b/apps/admin-x-design-system/src/utils/format-url.ts
@@ -1,4 +1,4 @@
-import isEmail from 'validator/es/lib/isEmail.js';
+import isEmail from 'validator/es/lib/isEmail';
 
 export const formatUrl = (value: string, baseUrl?: string, nullable?: boolean) => {
     if (nullable && !value) {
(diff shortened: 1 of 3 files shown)


apps/admin-x-design-system/src/utils/format-url.ts
```diff
@@ -1,4 +1,4 @@
-import isEmail from 'validator/es/lib/isEmail';
+import isEmail from 'validator/es/lib/isEmail.js';
```
EvanHahn, line 1: I admit I'm not certain why this is required.

I _think_ this is because `validator/es/lib/isEmail` isn't directly exported from the validator package, and we need to make sure it's resolving a file.

This change _could_ be done separately, but it only started causing problems after this PR was made.

Let me know if you want me to understand this further.

9larsons: Per Claude analysis, it's because Admin uses webpack v5 which uses `fullySpecified: true` so we need the extension.

9larsons (approved):

PR #26631, opened 2026-02-28
🎨 Changed email verification error messages to be configurable via hostSettings

betschki:
Managed hosting providers other than Ghost(Pro) currently see hardcoded "support@ghost.org" and "account is currently in review" messages that don't apply to their platform. This adds hostSettings configuration keys (emailSendingDisabledMessage, emailVerificationNeededMessage) under the existing emailVerification namespace, with the current Ghost(Pro) messages kept as defaults. Also replaced fragile regex-based error detection in publish-limit modal with proper error code checking.

- Why are you making it?
- What does it do?


JohnONolan (approved):
Good idea

betschki:
@JohnONolan any chance we could get a review on this?

ghost/admin/app/utils/publish-options.js
9larsons, line 1: I'm on the fence if this is something we should add to `defaults.json` instead of hard-coding here, but it probably makes sense to leave as-is, especially for the scope of this PR 👍.

ghost/core/core/server/services/email-service/email-service-wrapper.js
```diff
@@ -141,7 +141,8 @@ class EmailServiceWrapper {
             membersRepository,
             verificationTrigger: membersService.verificationTrigger,
             emailAnalyticsJobs,
-            domainWarmingService
+            domainWarmingService,
+            emailSendingDisabledMessage: configService.get('hostSettings:emailVerification:emailSendingDisabledMessage')
```
9larsons: I dislike passing a message through the wrapper vs passing config, as this feels too 'low level' and an anti pattern if we were to have several messages.

ghost/core/core/server/services/verification-trigger.js
```diff
@@ -205,7 +208,7 @@ class VerificationTrigger {
 
                 if (throwOnTrigger) {
```
9larsons, line 206: Currently, aside from tests, we have no callers that pass `throwOnTrigger = true`. This would then be dead code. Could you let me know why you updated this path?

9larsons (changes requested):
A more general comment: I'd love for us to define the available `hostSettings` (and more generally, our config options) _somewhere_. The fact that this is totally ambiguous is troublesome and hard to reason about.

I think I'd like to see a more generic pattern for passing the messages, although the DI pattern is the right one. Perhaps something like `hostSettings.messages.emailVerification.<items>`?

Then DI would/could still pass in `hostSettings.messages.emailVerification`, or if we use other hostSettings/config, the more broad item that's appropriate.

ghost/admin/app/utils/publish-options.js
betschki: Agree. It would make more sense there − though as you pointed out, I see that outside of the scope of this PR.

9larsons: Yeah - I should've been clearer with my comments. This is non-blocking comment and a todo for me.

ghost/core/core/server/services/email-service/email-service-wrapper.js
betschki: Good point! Switched to passing the full config service via DI, same pattern as DomainWarmingService.

ghost/core/core/server/services/verification-trigger.js
betschki: You're right! Reverted it back.

9larsons (approved):
Super clean! I am annoyed we pass `config: configService`, as I think simply passing `configService` is more clear... but we do this almost everywhere in core so it's better to be consistent.

PR #26638, opened 2026-03-02
Fixed linear-triage workflow failing due to disabled noop safe-output

kevinansfield:
## Summary
- The `linear-triage` agentic workflow was failing because the `noop` safe-output was disabled, so the agent had no tool to call after successfully triaging issues — causing all downstream jobs to fail
- Re-enabled `noop` in the workflow source `.md` file
- Upgraded `gh-aw` from v0.49.3 to v0.51.5 which properly includes `noop` in the `<safe-output-tools>` prompt when compiling
- Recompiled the lock file and cleaned up the stale v0.49.3 entry from the actions-lock

fixes https://github.com/TryGhost/Ghost/issues/26570


ErisDS:
- Workflow: `CI`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22576914703)
**What failed:** Playwright tests failed due to test assertion failures
**Why:** The logs show that the Playwright tests failed due to test assertion failures, which indicates a code issue rather than an infrastructure problem. The final error message mentions that 2 tests did not run and 54 passed, suggesting that some tests failed their assertions.
**Action:**
The author should review the failed Playwright tests, debug the issues, and fix the code to ensure the tests pass.

PR #26640, opened 2026-03-02
Fixed sign-out redirecting to wrong URL on subdirectory sites

kevinansfield:
## Summary

Sites using a subdirectory/proxy setup (e.g. `example.com/blog/ghost/`) could not sign out via the admin UI — the sign-out handler in the React admin app hardcoded `/ghost` paths without accounting for the subdirectory, resulting in a redirect to `/ghost` instead of `/blog/ghost/` and a "Cannot GET /ghost" error.

- Replaced hardcoded `/ghost/api/admin/session` fetch URL with `getGhostPaths().apiRoot` which derives the correct subdirectory from the current URL
- Replaced hardcoded `/ghost` redirect with `getGhostPaths().adminRoot` so the post-signout redirect respects the subdirectory

fixes https://linear.app/ghost/issue/ONC-1510

### DIFF
--- a/apps/admin/src/layout/app-sidebar/user-menu.tsx
+++ b/apps/admin/src/layout/app-sidebar/user-menu.tsx
@@ -11,6 +11,7 @@ import {
     Switch
 } from "@tryghost/shade"
 import { useCurrentUser } from "@tryghost/admin-x-framework/api/current-user";
+import { getGhostPaths } from "@tryghost/admin-x-framework/helpers";
 import { useUserPreferences, useEditUserPreferences } from "@/hooks/user-preferences";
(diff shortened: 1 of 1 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Run yarn nx affected -t lint --base=59399d8850befd5095ae1149e31f375cac9f0514`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22582608764)
**What failed:** Lint errors in admin app code
**Why:** The CI step failed due to lint errors in the admin app code, specifically related to unsafe assignment and calls of error-typed values. This is a code issue that the author should fix.
**Action:**
The author should address the lint errors in the admin app code to resolve the CI failure.

PR #26641, opened 2026-03-02
✨ Added Feedbin to list of supported Inbox Links providers

EvanHahn:
no ref

When someone signs up with a [Feedbin] email address, we now have a link to take them straight to the app. (Feedbin is an RSS reader that has custom email inboxes you can use to read everything all in one place.)

Note that Feedbin doesn't have an Android app, so we use the regular desktop link for both.

[Feedbin]: https://feedbin.com/


9larsons (approved):
nit: A context.json entry could be helpful.

ghost/i18n/locales/context.json
```diff
@@ -189,6 +189,7 @@
     "One hour ago": "Time a comment was placed",
     "One min ago": "Time since a comment was made",
     "Open AOL Mail": "Shown on signup and signin if your email is detected as AOL Mail",
+    "Open Feedbin": "",
```
cathysarisky: Any time you add a new string, please remember to populate context.json.


```suggestion
    "Open Feedbin": "Shown on signup and signin if your email is detected as Feedbin",
```

cathysarisky (changes requested):
teeny i18n fix!

ghost/i18n/locales/context.json
EvanHahn: Ah, didn't realize...thanks!!

cathysarisky (approved):
Thanks for resolving the context.json issue!  I cruise through and try to fill in blanks every so often, but if the folks with the most context fill it out as we go, it ends up better. :)

PR #26647, opened 2026-03-02
Fixed member deletion errors from concurrent API requests

kevinansfield:
## Summary
- When multiple `DELETE /members/{id}` requests target the same member concurrently, a race condition causes Bookshelf to throw "No Rows Deleted" errors that bubble up as 500s to Sentry
- The `findOne` check passes for both requests (member exists), but by the time the second request's `destroy()` runs, the member is already gone
- Fixed by passing `require: false` to Bookshelf's `destroy()` so it treats a missing row as a no-op — the member is already deleted, which is the desired outcome

refs https://linear.app/ghost/issue/ONC-1516

### DIFF
--- a/ghost/core/core/server/services/members/members-api/repositories/member-repository.js
+++ b/ghost/core/core/server/services/members/members-api/repositories/member-repository.js
@@ -805,9 +805,10 @@ module.exports = class MemberRepository {
             }
         }
 
+        // require: false so concurrent deletes don't throw "No Rows Deleted"
         return this._Member.destroy({
             id: data.id
-        }, options);
+        }, {...options, require: false});
     }
 
     async bulkDestroy(options) {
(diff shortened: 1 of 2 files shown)


ErisDS:
- Workflow: `CI`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22589714962)
**What failed:** Playwright browser tests failed with exit code 1
**Why:** The final error in the logs indicates that the Playwright browser tests failed with exit code 1, which is a soft failure caused by a code issue, likely a test assertion failure or a problem with the application code under test.
**Action:**
The author should investigate the Playwright test failures and fix any issues in the application code or test suite to resolve the CI failure.

ghost/core/core/server/services/members/members-api/repositories/member-repository.js
```diff
@@ -805,9 +805,18 @@ module.exports = class MemberRepository {
             }
         }
 
-        return this._Member.destroy({
-            id: data.id
-        }, options);
+        try {
+            return await this._Member.destroy({
+                id: data.id
+            }, options);
```
allouis: You can pass `{...options, require: false}` here and avoid the whole try/catch and error inspection

allouis (approved):
This can be simplified a little, but it's not blocking

sagzy (approved):

PR #26651, opened 2026-03-02
Specified more exact types for fetching bookmark data

EvanHahn:
no ref

This types-only change should have no user impact.


cmraible (approved):
Apologies, I could have sworn I had already approved this one! LGTM

PR #26658, opened 2026-03-02
Added card styles to member email renderer

9larsons:
ref https://linear.app/ghost/issue/NY-1114/
- added card styles to member email editor to support image, button, callout, etc cards

### DIFF
--- a/ghost/core/core/server/services/member-welcome-emails/email-templates/partials/card-styles.hbs
+++ b/ghost/core/core/server/services/member-welcome-emails/email-templates/partials/card-styles.hbs
@@ -4,11 +4,37 @@ td.kg-card-spacing {
     padding: 0 0 1.5em 0;
 }
 
+figure {
+    margin: 0 0 1.5em;
+    padding: 0;
+}
+
+figcaption {
+    text-align: center;
+    font-family: -apple-system, BlinkMacSystemFont, Roboto, Helvetica, Arial, sans-serif;
+    font-size: 13px;
+    padding-top: 10px;
+    padding-bottom: 10px;
+    line-height: 1.5em;
+    {{#if backgroundIsDark}}
+        color: #ffffff;
+        color: rgba(255, 255, 255, .6);
+    {{else}}
+        color: #15212a;
+        color: rgba(0, 0, 0, 0.6);
+    {{/if}}
+}
+
+figcaption a {
+    text-decoration: none!important;
+}
+
 .kg-image-card img {
     display: block;
     margin: 0 auto;
     height: auto;
     width: auto;
+    max-width: 100%;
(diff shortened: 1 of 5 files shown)


ghost/core/core/server/services/member-welcome-emails/email-templates/wrapper.hbs
```diff
@@ -311,6 +311,8 @@
         font-size: 19px;
         line-height: 1.3em;
     }
+
+    {{>cardStyles}}
```
9larsons, line 315: This injects the styles in the wrapper. We'll want to refactor this wrapper later w/ more changes to the render engine.

ghost/core/core/server/services/member-welcome-emails/email-templates/partials/card-styles.hbs
```diff
@@ -0,0 +1,161 @@
+/* Welcome-email Koenig card styles (minimal subset) */
```
9larsons, line 1: Largely copied from `styles.hbs`, a monolith.

9larsons: I think we'll want to segmentize that file into smaller ones like this `card-styles`.

ghost/core/core/server/services/member-welcome-emails/member-welcome-email-renderer.js
```diff
@@ -15,17 +17,51 @@ class MemberWelcomeEmailRenderer {
 
     constructor({t}) {
         this.Handlebars = require('handlebars').create();
+        this.Handlebars.registerHelper('if', function (conditional, options) {
+            if (conditional) {
+                return options.fn(this);
+            } else {
+                return options.inverse(this);
+            }
+        });
```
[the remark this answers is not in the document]
9larsons: Seems valid, looking in to handling.

ghost/core/core/server/services/member-welcome-emails/member-welcome-email-renderer.js
```diff
@@ -19,6 +20,11 @@ class MemberWelcomeEmailRenderer {
             let hash = options?.hash;
             return t(key, hash || options || {});
         });
+        const cardStylesSource = fs.readFileSync(
```
EvanHahn, line 23: I don't know how often this gets called but I could see this eventually becoming a performance issue if we're constructing a lot of these objects. Just something to keep in mind.

EvanHahn (approved):
Could we create a ticket to DRY out the duplicated code? (Maybe we're going to tackle this more holistically. so it might not be worth doing.)

Other than that, not much to say about this. LGTM.

PR #26666, opened 2026-03-03
Added label picker with inline editing to members list

rob-ghost:
The members list filter and bulk-action modals used simple Select dropdowns for choosing labels, which limited users to picking one label at a time with no way to manage labels without leaving the context. This replaces those dropdowns with a searchable, multi-select label picker built on Popover + Command (cmdk) from Shade — the same proven pattern used by the existing filter options popovers.

The picker is driven by a `useLabelPicker` hook that encapsulates all label CRUD operations (browse, create, edit, delete) via new mutations added to admin-x-framework. The component itself adapts to three contexts through prop presence rather than an explicit mode prop: the filter value cell renders as an inline popover with applied-label pills above the search, the add-label bulk modal includes a "Create" option for new labels, and the remove-label bulk modal scopes available labels to those actually assigned to the filtered members. In all contexts, labels can be renamed or deleted inline via edit/delete buttons that appear on hover, with confirmation for destructive actions.

The Ember state bridge was updated with `LabelsResponseType` and `MembersResponseType` mappings so that mutation cache updates propagate correctly between the React and Ember query caches. The filter state serialization was also adjusted to handle multiselect fields with empty values, so a label filter row persists in the URL even before any labels are selected.

ref https://linear.app/tryghost/issue/BER-3342/

---

> [!NOTE]
> **Medium Risk**
> Touches multiple user-facing flows (filters and bulk member operations) and introduces new label mutation paths with cache-update logic; issues could cause incorrect selections or label assignments/removals.
> 
> **Overview**
> Replaces the members label dropdown UX with a new searchable, multi-select `LabelPicker` used in both the members list label filter (inline popover renderer) and bulk add/remove label modals, including inline rename/delete and optional create-from-search.
> 
> Adds label CRUD mutations to `admin-x-framework` (`useCreateLabel`, `useEditLabel`, `useDeleteLabel`) with query-cache updates, introduces a shared `useLabelPicker` hook to coordinate selection + mutations (including slug-change handling), and updates bulk member actions to apply multiple labels and to scope removable labels to those present on the filtered members.
> 
> Improves filter URL handling by serializing empty multiselect values so newly-added filter rows persist, and updates the Ember state bridge mappings for `LabelsResponseType`/`MembersResponseType` to support React-side cache update semantics.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c96b7652a69a8a44d53b2ad714ee995db806f7b2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/admin-x-framework/src/api/labels.ts
+++ b/apps/admin-x-framework/src/api/labels.ts
@@ -1,4 +1,4 @@
-import {Meta, createQuery} from '../utils/api/hooks';
+import {Meta, createMutation, createQuery} from '../utils/api/hooks';
 
 export type Label = {
     id: string;
@@ -18,4 +18,48 @@ const dataType = 'LabelsResponseType';
 export const useBrowseLabels = createQuery<LabelsResponseType>({
     dataType,
     path: '/labels/'
+});
+
+export const useCreateLabel = createMutation<LabelsResponseType, Pick<Label, 'name'>>({
+    method: 'POST',
+    path: () => '/labels/',
+    body: label => ({labels: [label]}),
+    updateQueries: {
+        dataType,
+        emberUpdateType: 'createOrUpdate',
+        update: (newData, currentData) => {
+            const current = currentData as LabelsResponseType;
+            return current && {...current, labels: current.labels.concat(newData.labels)};
+        }
+    }
+});
+
+export const useEditLabel = createMutation<LabelsResponseType, Pick<Label, 'id' | 'name'>>({
+    method: 'PUT',
+    path: label => `/labels/${label.id}/`,
+    body: label => ({labels: [label]}),
+    updateQueries: {
+        dataType,
+        emberUpdateType: 'createOrUpdate',
+        update: (newData, currentData) => {
+            const current = currentData as LabelsResponseType;
+            return current && {
+                ...current,
+                labels: current.labels.map(label => newData.labels.find(({id}) => id === label.id) || label)
+            };
+        }
+    }
+});
+
+export const useDeleteLabel = createMutation<void, string>({
+    method: 'DELETE',
+    path: id => `/labels/${id}/`,
+    updateQueries: {
+        dataType,
+        emberUpdateType: 'delete',
+        update: (_, currentData, id) => {
+            const current = currentData as LabelsResponseType;
+            return current && {...current, labels: current.labels.filter(label => label.id !== id)};
+        }
+    }
 });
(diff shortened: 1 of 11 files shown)


weylandswart (changes requested):
Made tweaks to what I could in https://github.com/TryGhost/Ghost/pull/26666/commits/39bddb521f99f352f4478a2d09a23f31101d5ff7. 

A few more changes that I think are worth making. 

- Clicking on save when in the label edit state seems to save the value while remaining in the editing state
- It’s not currently possible to edit a label when initially selecting a filter of type label (Before the filter is added) 
Image

- Once a label filter has been added, the label dropdown is not aligned with the left side of the dropdown trigger (Label name or x labels). 
Image

- The combobox should use Shade's Input or Select styles (Height, focus state, borders, colours, etc.)
- I get an error when trying to add or remove a label ` Error: A mutation invalidating MembersResponseType succeeded in React Admin but there is no mapping to an Ember type. Add one to emberDataTypeMapping
    at E.onInvalidate (state-bridge.js:147:1)
    at onInvalidate (main.tsx:29:35)
    at Object.onSuccess (hooks.ts:198:13)
    at async members-actions.tsx:60:17 `
- Does the empty header part of the dropdown have to be there when the option to create a new label is shown? It looks a bit broken to me.  
Image

ErisDS:
- Workflow: `CI`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22647943937)
**What failed:** Playwright tests failed due to a test assertion failure
**Why:** The final error in the logs indicates that a Playwright test failed due to an assertion error. This is a code issue, as the test is failing due to a problem with the application code under test, not an infrastructure problem.
**Action:**
The developer should review the failed test case, debug the issue, and fix the underlying code to ensure the test passes.

weylandswart (changes requested):
- Save isn’t exiting the editing state for a label (The changes are persisted, though)
- The dropdown when bulk adding a label (And likely removing) isn’t scrollable 
- I think, UX wise, each new label addition should close the dropdown
- The add/remove label bulk actions work, but I stay in the modal and I see a "Failed to remove label. There was a problem removing this label. Please try again.” toast.
- Labels are not editable when initially adding a label filter from the top right filter dropdown (There's no pencil icon on hover)
- If you add a label (From bulk), the existing searched-for text remains in the input

weylandswart (approved):

PR #26669, opened 2026-03-03
Fixed review issues in paginated label dropdowns

kevinansfield:
## Summary

Addresses review feedback from #26655 and fixes additional issues found during follow-up testing.

**Review comment fixes:**
- Added `aria-selected="false"` to loading option in `power-select-options-with-scroll.hbs` to fix a11y lint violation instead of suppressing it
- Fixed `gh-member-single-label-input` to not overwrite a preselected label with the first available option when the selected label isn't in the first loaded page — only calls `onChange` for default (no pre-selection) case
- Changed `options.unshift` to `options.push` in `gh-members-recipient-select` so the Labels group appears after Tiers, keeping infinite scroll predictable
- Added `@tracked` to `_meta` in `labels-manager` so `hasLoaded`/`hasLoadedAll` getters reactively update
- Added `parseInt` to pagination comparison in `loadMoreTask` for consistency with `hasLoadedAll`

**Additional fixes:**
- Used public `labelsManager.labels` getter instead of private `labelsManager._labels` in `gh-member-label-input`
- Switched `addSearchedLabels` deduplication to ID-based `Set` lookup (reference equality can fail with Ember Data models)
- Made `searchLabelsTask` restartable to prevent stacking concurrent search requests
- Removed redundant `|| []` after `.filter(Boolean)` in `selectedLabels`
- Added explicit `waitFor` calls in E2E label selection helpers to prevent flaky tests with paginated dropdown loading

refs https://github.com/TryGhost/Ghost/pull/26655

### DIFF
--- a/e2e/helpers/pages/admin/members/members-page.ts
+++ b/e2e/helpers/pages/admin/members/members-page.ts
@@ -88,13 +88,19 @@ class SettingsSection extends BasePage {
     }
 
     private async selectLabelOption(labelName: string): Promise<void> {
+        await this.selectLabel.waitFor({state: 'visible'});
         await this.selectLabel.click();
 
         const dropdown = this.page.locator('.ember-power-select-dropdown').last();
+        await dropdown.waitFor({state: 'visible'});
+
         const searchInput = dropdown.locator('.ember-power-select-search input');
+        await searchInput.waitFor({state: 'visible'});
         await searchInput.fill(labelName);
 
-        await dropdown.getByRole('option', {name: labelName, exact: true}).click();
+        const option = dropdown.getByRole('option', {name: labelName, exact: true});
+        await option.waitFor({state: 'visible'});
+        await option.click();
     }
 }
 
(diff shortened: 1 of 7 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Run yarn nx affected -t lint --base=08056d2c1816341df4dc254ba93d8afbaca637f8`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22628671133)
**What failed:** Lint error in power-select-options-with-scroll.hbs template
**Why:** The failure is caused by a lint error in the power-select-options-with-scroll.hbs template. This is a code issue that the author should fix, as it is a violation of the 'require-mandatory-role-attributes' rule in the Ember template linter.
**Action:**
The author should run `npx ember-template-lint app/components/power-select-options-with-scroll.hbs --clean-todo` to remove the todo from the todo list and fix the lint error.

kevinansfield:
Self-merged as changes were minimal and I've been testing other fixes on top of these changes locally.

PR #26670, opened 2026-03-03
Added newsletter selection to bulk member unsubscribe

rob-ghost:
The React members list bulk unsubscribe action previously removed members from all newsletters unconditionally. Sites with multiple active newsletters had no way to target a specific newsletter, unlike the Ember admin which already supported this.

The unsubscribe modal now detects how many active newsletters exist. When there are two or more, it renders a newsletter selector with an "All newsletters" default and individual newsletter options. When there's only one (or none), it shows the existing simple confirmation dialog. These two flows are implemented as separate components to keep the logic clean. The backend already accepted an optional `newsletter` parameter on the bulk unsubscribe endpoint, so no server-side changes were needed.

### DIFF
--- a/apps/admin-x-framework/src/api/members.ts
+++ b/apps/admin-x-framework/src/api/members.ts
@@ -182,6 +182,7 @@ export interface BulkEditAction {
     meta?: {
         label?: {id: string};
     };
+    newsletter?: string | null;
 }
 
 export interface BulkOperationResponseType {
@@ -204,7 +205,8 @@ export const useBulkEditMembers = createMutation<
     body: ({action}) => ({
         bulk: {
             action: action.type,
-            meta: action.meta || {}
+            meta: action.meta || {},
+            newsletter: action.newsletter
         }
     }),
     searchParams: ({filter, all}) => {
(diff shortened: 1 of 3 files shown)


weylandswart:
- I don't think the modal logic is properly respecting archived newsletters. In Ember if there's only 1 active newsletter then the modal doesn't show a select. In React the radios are still show and the select still includes archived newsletters.
- The modal state isn't being reset when its submitted. If I unsubscribe and reopen the modal the selected newsletter pills are still present in the input.

rob-ghost:
@weylandswart 

> I don't think the modal logic is properly respecting archived newsletters. In Ember if there's only 1 active newsletter then the modal doesn't show a select. In React the radios are still show and the select still includes archived newsletters.

I think this is a bug in the Ember -> React bridge, I just tested and archived newsletters don't appear in the modal, however if you go to settings and archive them, then navigate back to members and attempt to unsubscribe again, **without reloading the page**, then the archived newsletters are still there.

This probably means the ember side is not invalidating the query cache in React so we're seeing stale newsletters. I think we can fix this separately.

Here's what 1 newsletter looks like now when there's only one newsletter:
Screenshot 2026-03-04 at 19 59 25

weylandswart:
@rob-ghost Yeah, weird, all good now testing again. Not sure what the issue was. All looks and feels good to me!

PR #26676, opened 2026-03-03
Improved URL service and sitemap boot performance

rob-ghost:
The URL service is one of the slowest parts of Ghost's boot sequence. It fetches every published resource from the database, generates URLs, and then notifies the sitemap service one resource at a time via events. This PR tackles four independent bottlenecks in that pipeline.

First, the `Urls` data store was coupled to event emission — every call to `Urls.add()` fired a `url.added` event, which the sitemap service consumed one-by-one during boot. This meant N event dispatches, N moment.js date parses, and N XML node constructions before the server could start. The store is now a pure data structure, and the URL service emits a single bulk `url.init` event after all URLs are generated. The sitemap manager handles this bulk event to initialize all entries at once, while runtime additions/removals still use per-resource events emitted by the `UrlGenerator`.

Second, the resource config used exclude lists that grew with every new schema column and still fetched ~20 columns per post when only ~10 are needed for URL generation, sitemap XML, and change detection. These are now explicit include lists, making the contract between the URL service and the database self-documenting and reducing query payload.

Third, `raw-knex.js` existed specifically to bypass Bookshelf ORM overhead, yet still called `toJSON()`, `fixBools()`, and `fixDatesWhenFetch()` through the Bookshelf prototype on every row. `toJSON`/`serialize` just shallow-copy attributes with no meaningful transformation in this context. `fixDatesWhenFetch` parses dates with moment.js redundantly since knex already returns JavaScript Date objects. The only necessary operation — boolean coercion — is now a pre-computed column loop without Bookshelf prototype lookups.

Fourth, relation queries (tags, authors) used `WHERE IN` with every post ID materialized as a literal. For 10k posts that's 10k string values the query planner must parse. This is replaced with a subquery that mirrors the main query's NQL filter, letting the database use an index scan instead of parsing a literal list.

An integration test suite (`url-service-and-sitemap.test.js`) was added first as a safety net, covering both default routing and custom multi-collection routing with custom fixtures. It verifies URL generation, sitemap content, canonical exclusions, and draft filtering end-to-end.

### DIFF
--- a/ghost/core/core/server/models/base/plugins/raw-knex.js
+++ b/ghost/core/core/server/models/base/plugins/raw-knex.js
@@ -30,12 +30,20 @@ module.exports = function (Bookshelf) {
                     limit
                 } = options;
 
-                const bookshelfPrototype = Bookshelf.registry.models[modelName].prototype;
                 const tableNames = {
                     Post: 'posts',
                     User: 'users',
                     Tag: 'tags'
                 };
+
+                const tableName = tableNames[modelName];
+                const tableDef = schema.tables[tableName];
+                const booleanColumns = [];
+                for (const key in tableDef) {
+                    if (!key.startsWith('@@') && tableDef[key].type === 'boolean') {
+                        booleanColumns.push(key);
+                    }
+                }
 
                 const relations = {
                     tags: {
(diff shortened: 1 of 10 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Legacy tests`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22644498628)
**What failed:** Test assertion failure: expected 301 to equal 200 in frontend behavior tests
**Why:** The root cause is a test assertion failure, where the expected HTTP status code of 301 did not match the actual 200. This is a code issue, as the test is validating the application's behavior and the failure indicates a problem with the code under test.
**Action:**
The author should investigate the frontend behavior tests and fix the issue that is causing the assertion failure. This is likely a bug in the application code that needs to be addressed.

ghost/core/core/server/models/base/plugins/raw-knex.js
```diff
@@ -73,8 +82,10 @@ module.exports = function (Bookshelf) {
                     query.limit(limit);
                 }
 
-                // exclude fields if provided
-                if (exclude) {
+                // select only the fields needed
+                if (include) {
```
vikaspotluri123, line 86: Should there be a debug assertion that errors when both include and exclude are provided?

rob-ghost:
Closing in favour of decomposing into more reviewable chunks, first is here: https://github.com/TryGhost/Ghost/pull/26689

ghost/core/core/server/models/base/plugins/raw-knex.js
rob-ghost: FYI this was a PoC which is being decomposed into more reviewable chunks, the first of which is [here](https://github.com/TryGhost/Ghost/pull/26689) which removes `exclude` as its not used by anything after this change.

PR #26678, opened 2026-03-04
🐛 Fix Admin API Post update method to not save a revision when save_revision=false

markstos:
fixes https://github.com/TryGhost/ghost/issues/26677

There's a sort order bug in Ghost's revision logic. Here's the chain:

    post-revision.js:32-33 — orderDefaultRaw() returns 'created_at_ts DESC' — newest first
    post.js:920-926 — revisions are fetched via findAll → gets them in DESC order (newest first)
    post-revisions.ts:56 — const latestRevision = revisions[revisions.length - 1] — grabs the last element

So latestRevision is actually the oldest revision, not the newest. The time check on line 84 compares Date.now() against the oldest revision's timestamp — which is almost always >10 minutes old. This means the background_save condition passes every time content has changed, regardless of how recently the last revision was created.

This bug is masked in the Ghost Admin UI because it always sends save_revision=true and there was no test coverage in the test suite for the condition.

The flag is not documented in the Admin API docs, helping the issue to go undetected.
Steps to Reproduce

    Make an Admin API call to create a post.
    Check revisions.
    Make an Admin API call to update a post with save_revision=false in the query string.
    Check revision count.
    Make an Admin API call to update a post with save_revision=false in the query string.

Revision count should not be increasing, but it is.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit fc10425340517efcd21e515b920108ed307b9096. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ghost/core/core/server/models/post.js
+++ b/ghost/core/core/server/models/post.js
@@ -932,7 +932,7 @@ Post = ghostBookshelf.Model.extend({
                         columns: ['id', 'lexical', 'created_at', 'author_id', 'title', 'reason', 'post_status', 'created_at_ts', 'feature_image']
                     }, _.pick(options, 'transacting')));
 
-                const revisions = revisionModels.toJSON();
+                const revisions = revisionModels.toJSON().reverse();
 
                 const current = {
                     id: model.id,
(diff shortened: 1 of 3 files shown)


9larsons:
@markstos Looks good! Thanks for this. Pushed a couple minor changes.

PR #26683, opened 2026-03-04
WIP: Changed the verification system to use a webhook

sam-lord:
ref [GVA-690](https://linear.app/ghost/issue/GVA-690/change-logic-in-ghost-to-send-webhook-instead-of-email)

Draft, this was a one-shot PR to see whether it was possible to swap to using a webhook (based on a URL defined in config:hostSettings) instead of directly sending the verification email from Ghost


aileen:
Closing, moved it to #26778

PR #26684, opened 2026-03-04
Added reusable Dropzone component in Shade

jonatansberg:
This is PR1 of a stacked members import rework. The immediate goal is to introduce a reusable, accessible Dropzone primitive in Shade so the members import rewrite can consume a tested, shared drag-and-drop/file-select component instead of ad-hoc logic.

Why this change:
- Establishes a shared upload/dropzone primitive for admin apps
- Reduces repeated manual drag-and-drop handling
- Improves accessibility and reliability with focused unit tests

What changed:
- Added `Dropzone` component to `@tryghost/shade` using `react-dropzone`
- Exported `Dropzone` from Shade public API
- Added Storybook states: idle, drag-active, disabled, rejection, custom content
- Added unit tests for keyboard/accessibility behavior and accept/reject callbacks

Checklist:
- [x] I've read and followed the Contributor Guide
- [x] I've written automated tests to prove the change works

Verification:
- `yarn --cwd apps/shade lint`
- `yarn --cwd apps/shade test:unit`


ErisDS:
- Workflow: `CI`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22668313514)
**What failed:** Test assertion failure: expected 332 to equal 392 in domain warming integration test
**Why:** The root cause of the failure is a test assertion failure in the domain warming integration test. This is a code issue, as the test is checking the expected behavior of the application code and the failure indicates a mismatch between the expected and actual values.
**Action:**
The author should investigate the domain warming functionality and update the test to ensure it accurately reflects the expected behavior, or fix the underlying code that is causing the test to fail.

ErisDS:
- Workflow: `CI`
- Failed Step: `Determine added packages`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22893165306)
**What failed:** Rate limited by external service
**Why:** Infrastructure issue detected: rate limited by external service. This is typically not caused by code changes.
**Action:**
Try re-running the workflow. If it persists, investigate the infrastructure.

ErisDS:
- Workflow: `CI`
- Failed Step: `Run yarn nx affected -t lint --base=d7bb54f26e999932cf2f611a2a76609caeb1b963`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22999046419)
**What failed:** Lint errors in code
**Action:**

apps/shade/src/components/ui/dropzone.tsx
```diff
@@ -0,0 +1,96 @@
...
+    };
+
+    const dropzoneRootProps = getRootProps({
+        ...props,
+        role: 'button',
+        tabIndex: disabled ? -1 : 0,
+        'aria-disabled': disabled,
+        className: cn(
+            'flex cursor-pointer flex-col items-center justify-center rounded-md border-2 border-dashed p-10 transition-colors outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2',
+            disabled && 'cursor-not-allowed opacity-60 pointer-events-none',
```
kevinansfield, line 75: Nit: `cursor-not-allowed` never takes effect here because `pointer-events-none` prevents any pointer interactions. Reproducible in the storybook demo.

apps/shade/src/components/ui/dropzone.tsx
```diff
@@ -0,0 +1,96 @@
...
+            isDragAccept && 'border-green-500 bg-green-50 dark:bg-green-950/20',
+            isDragReject && 'border-red-500 bg-red-50 dark:bg-red-950/20',
```
kevinansfield: ~These only get set if `accept` is provided, if it isn't provided then there's no feedback when dragging a file over the drop zone. Storybook shows this when dragging a file over the "Drag active" demo - no styles change.~

kevinansfield: This might be a Safari issue actually, I'm not getting many interactions there at all 👀

kevinansfield: It seems there's a known limitation where in Safari react-dropzone can't determine if a file is accepted or rejected. Would it be better to always show the green state on `isDragActive` unless we get `isDragReject` or the component is disabled?

kevinansfield (changes requested):
All looks good but I think the Safari-forced styling tweak should be applied so we have some file drag feedback across browsers.

kevinansfield (approved):

PR #26689, opened 2026-03-04
Optimised raw-knex resource fetching for URL service boot

rob-ghost:
The URL service boots by fetching every published post, page, tag, and author from the database via `raw-knex.js` — a fast path that bypasses Bookshelf's ORM layer. On large sites (250k+ posts), this boot path was taking ~9s. This PR brings it down to ~5.4s (a 39% reduction) by addressing three distinct bottlenecks in how raw-knex fetches and processes data.

**Include lists instead of exclude lists** — The resource config previously listed columns to *exclude*, which meant every new column added to the schema was fetched by default even if the URL service didn't need it. This inverts the approach: each resource type now declares exactly which columns it needs (id, slug, type, featured, dates, images, canonical_url). This makes the contract explicit and keeps the query payload minimal.

**Removed Bookshelf ORM overhead** — After fetching raw rows, the old code wrapped each row in a fake Bookshelf model just to call `toJSON()`, `fixBools()`, and `fixDatesWhenFetch()`. That's a lot of ceremony for what amounts to boolean coercion. This replaces all of that with a direct loop over schema-derived boolean columns, eliminating lodash, prototype binding, and serialize overhead entirely. This single change accounts for the largest speedup (~1.3s on 250k posts).

**Parallel split-query relation loading** — Relations (tags, authors) were previously loaded sequentially using a JOIN query whose `WHERE IN` clause contained every post ID as a literal value. This rewrites relation loading to use a split-query pattern: one query for the pivot table (e.g. `posts_tags`) filtered by a subquery, and a second query for the lookup table (e.g. `tags`), joined in JS. Because the relation queries use subqueries instead of materialised ID lists, they don't depend on the main query's results — all queries fire in parallel via `Promise.all`. Posts sharing the same tag or author reference the same object, avoiding hundreds of thousands of per-row allocations.

The PR also includes an integration test suite (commit 0) that boots the URL service with custom fixtures and verifies both URL cache state and sitemap XML output. This covers default and custom routing configurations, ensuring drafts, canonical posts, and orphan tags are correctly excluded. This test suite serves as a safety net for all the optimisation work — any regression in the boot path or event wiring will surface here.

### DIFF
--- a/ghost/core/core/server/models/base/plugins/raw-knex.js
+++ b/ghost/core/core/server/models/base/plugins/raw-knex.js
@@ -2,6 +2,8 @@ const _ = require('lodash');
 const debug = require('@tryghost/debug')('models:base:raw-knex');
 const plugins = require('@tryghost/bookshelf-plugins');
 
+const schema = require('../../../data/schema');
+
 /**
  * @param {import('bookshelf')} Bookshelf
  */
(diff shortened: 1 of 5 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Run yarn nx affected -t lint --base=d066dba172c7de052b0d1a22823ac7e8b50bbd9b`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22671368880)
**What failed:** Lint errors in test file: unused variables 'forKnex' and 'markdownToMobiledoc'
**Why:** The failure is caused by lint errors in the test file, specifically two unused variables. This is a code issue that the author should fix, as it does not involve any infrastructure problems.
**Action:**
The author should address the lint errors by either removing the unused variables or marking them as allowed unused variables in the lint configuration.

PR #26693, opened 2026-03-04
Fixed relative urls in member welcome emails

9larsons:
closes https://linear.app/ghost/issue/NY-1127/

- made urls full instead of relative, which won't work in email


ErisDS:
- Workflow: `CI`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22673506145)
**What failed:** Test assertion failure in domain warming integration test
**Why:** The root cause is a test assertion failure in the domain warming integration test. This is a code issue, as the test is failing due to an incorrect expected value in the test assertion.
**Action:**
The author should investigate the domain warming integration test and fix the issue with the expected value in the assertion. Once the test is fixed, the CI build should pass.

EvanHahn (approved):

PR #26698, opened 2026-03-04
Improved content API filter validation

kevinansfield:
ref https://linear.app/ghost/issue/ONC-1525/

- Improved field matching in `rejectPrivateFieldsTransformer` on public content API endpoints
- Extracted the transformer into a shared utility

### DIFF
--- a/ghost/core/core/server/api/endpoints/authors-public.js
+++ b/ghost/core/core/server/api/endpoints/authors-public.js
@@ -1,23 +1,13 @@
 const tpl = require('@tryghost/tpl');
 const errors = require('@tryghost/errors');
-const {mapQuery} = require('@tryghost/mongo-utils');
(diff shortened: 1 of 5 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Build & push core image`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22676088171)
**What failed:** Docker build failed due to apt-get install error
**Why:** The final error in the logs indicates a failure in the Docker build process, specifically during the apt-get install step. This is an infrastructure issue and not a code problem, as it is related to the build environment and not the application code itself.
**Action:**
Investigate the build environment and dependencies to identify the root cause of the apt-get install error. This could be related to network issues, resource constraints, or compatibility problems with the build environment.

PR #26715, opened 2026-03-05
🐛 Fixed contributor posts list not being scrollable

kevinansfield:
fixes https://linear.app/ghost/issue/ONC-1512/

The contributor layout used `h-full overflow-auto` on the `<main>` element, with the EmberRoot child also using `h-full`. This constrained the Ember content to exactly the scroll container's height, so `.gh-viewport`'s `overflow: hidden` clipped the posts list with no way to scroll.

Changed the contributor layout to use the same flex column pattern as the non-contributor `SidebarInset` layout — a `flex flex-col overflow-y-auto` container with a `flex-1` wrapper for children. This allows the content to grow beyond the container via `min-height: auto`, enabling scroll.

### DIFF
--- a/apps/admin/src/layout/admin-layout.tsx
+++ b/apps/admin/src/layout/admin-layout.tsx
@@ -23,7 +23,9 @@ export function AdminLayout({ children }: AdminLayoutProps) {
     if (isContributor) {
         return (
             <div className="relative h-full bg-background">
-                <main className="h-full overflow-auto">{children}</main>
+                <main className="flex h-full flex-col overflow-y-auto">
+                    <div className="flex-1">{children}</div>
+                </main>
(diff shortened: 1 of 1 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Load Image`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22724851413)
**What failed:** Docker login to ghcr.io failed due to network timeout
**Why:** The logs show a network error when trying to log in to the GitHub Container Registry (ghcr.io). This is an infrastructure issue outside the control of the code changes, and is likely due to a temporary network problem or rate limiting.
**Action:**
This is a transient infrastructure issue, so retrying the build may resolve the problem. If the issue persists, the maintainers should investigate the GitHub Actions runner configuration or contact GitHub support.

PR #26717, opened 2026-03-05
Added IP format normalization to request-external

kevinansfield:
ref https://linear.app/ghost/issue/ONC-1533/

- replaced regex IP matching with a WHATWG URL parser normalization step and octet matching


ErisDS:
**Classification:** ⚪ CI FAIL
- Workflow: `CI`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22726622807)
**What failed:** Workflow `CI` failed.
**Why:** Velo could not extract enough detail to provide a full explanation.
**Action:**
Inspect the failed step/logs and update the PR to address the failure.

PR #26719, opened 2026-03-05
🐛 Fixed double-prefixed asset URLs when using CDN

rob-ghost:
After #26555 replaced `cdnUrl` config with the runtime `assetBase()` utility, customers on CDN-served installs started seeing 404s for admin assets. The browser was requesting URLs like `https://assets.ghost.io/admin-forward/https://assets.ghost.io/admin-forward/assets/ghost-dark-{hash}.css` — the CDN origin doubled up.

The root cause is that `broccoli-asset-rev` rewrites string literals in the compiled JavaScript at build time, prepending the full CDN origin and fingerprint hash. So a source-level `'assets/ghost-dark.css'` becomes `'https://assets.ghost.io/admin-forward/assets/ghost-dark-{hash}.css'` in the production build. The old code handled this with `config.cdnUrl ? '' : this.ghostPaths.adminRoot` — when a CDN was in use, the prefix was an empty string, so the already-absolute URL passed through unchanged. The new `assetBase()` always returns a non-empty absolute URL, causing double-prefixing.

This adds a `prefixAssetUrl()` helper to `asset-base.js` that checks whether the URL is already absolute before prepending `assetBase()`. All call sites that concatenate `assetBase()` with potentially-rewritten asset paths (`lazy-loader.js`, `user.js`, `parse-history-event.js`) now use this helper instead of direct concatenation. The `admin-x-component.js` and `fetch-koenig-lexical.js` call sites are unaffected because they concatenate `assetBase()` with partial directory paths that `broccoli-asset-rev` won't match as fingerprintable filenames.

Tests cover the double-prefix scenario at both the unit level (`prefixAssetUrl` with absolute URLs) and the integration level (lazy-loader receiving a CDN-rewritten URL).

### DIFF
--- a/ghost/admin/app/components/admin-x/admin-x-component.js
+++ b/ghost/admin/app/components/admin-x/admin-x-component.js
@@ -1,7 +1,6 @@
 import * as Sentry from '@sentry/ember';
 import Component from '@glimmer/component';
 import React, {Suspense} from 'react';
-import assetBase from 'ghost-admin/utils/asset-base';
 import config from 'ghost-admin/config/environment';
 import fetch from 'fetch';
 import fetchKoenigLexical from 'ghost-admin/utils/fetch-koenig-lexical';
@@ -9,6 +8,7 @@ import ghostPaths from 'ghost-admin/utils/ghost-paths';
 import {action} from '@ember/object';
 import {camelize} from '@ember/string';
 import {inject} from 'ghost-admin/decorators/inject';
+import {prefixAssetUrl} from 'ghost-admin/utils/asset-base';
 import {inject as service} from '@ember/service';
 import {tracked} from '@glimmer/tracking';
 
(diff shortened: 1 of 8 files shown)


kevinansfield (approved):

ErisDS:
- Workflow: `CI`
- Failed Step: `Load Image`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22729290087)
**What failed:** Docker pull failed due to a network error connecting to the container registry
**Why:** The final error in the logs indicates a network error when trying to pull the Docker image from the container registry. This is an infrastructure issue, not a code problem, as it is related to the network connection to the external service (the container registry).
**Action:**
Investigate the network connectivity issues between the CI runner and the container registry. This could be a temporary outage or a configuration problem with the CI environment.

PR #26733, opened 2026-03-09
Updated cancel_link helper default label to 'Resume subscription'

minimaluminium:
ref https://linear.app/ghost/issue/BER-3417/refresh-custom-cancellation-button-state-without-manual-page-reload

The `{{cancel_link}}` theme helper's default `continueLabel` was "Continue subscription", while Portal was updated to use "Resume subscription" in #26679.                                                         
                  
  - Updated the default `continueLabel` from "Continue subscription" to "Resume subscription"
  - Updated the corresponding unit test

### DIFF
--- a/ghost/core/core/frontend/helpers/cancel_link.js
+++ b/ghost/core/core/frontend/helpers/cancel_link.js
@@ -4,7 +4,7 @@
 // Should be used inside of a subscription context, e.g.: `{{#foreach @member.subscriptions}} {{cancel_link}} {{/foreach}}`
 // Outputs cancel/renew links to manage subscription renewal after the subscription period ends.
 //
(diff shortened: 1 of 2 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Load Image`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22848827095)
**What failed:** Docker login to ghcr.io failed due to a context deadline exceeded
**Why:** The root cause of the failure is a Docker login error to the GitHub Container Registry (ghcr.io) due to a context deadline exceeded. This is an infrastructure issue and not a code problem, as it is related to network connectivity or a temporary issue with the external service.
**Action:**
This is likely a temporary issue with the GitHub Container Registry. Try rerunning the workflow or check the status of the registry. If the issue persists, the infrastructure team should investigate the problem.

PR #26736, opened 2026-03-09
Increased retention offer code entropy from 16-bit to 32-bit

kevinansfield:
no issue

- Retention offer codes now use `Uint32Array` instead of `Uint16Array`, generating 8-character hex codes instead of 4-character ones
- Increases the code space from ~65K to ~4.3 billion possible values


sagzy:
Closing in favor of https://github.com/TryGhost/Ghost/pull/26738, which contains an additional length check for the offer name, as Stripe imposes a 40 char limit to coupon names

PR #26737, opened 2026-03-09
Stripped redundant fields from public member offers endpoint

mike182uk:
no ref

The `/members/api/member/offers/` endpoint was returning the full admin DTO but some of the fields were not are not needed by Portal. Added a `PublicOfferDTO` and `toPublicDTO` mapper that returns only the fields the public endpoint requires


sagzy (approved):
LGTM

PR #26742, opened 2026-03-09
🐛 Fixed bookmark filtering in Welcome Emails editor

EvanHahn:
closes https://linear.app/ghost/issue/NY-1143

`useFilterableApi` filtered the response if data was already loaded, but not if it wasn't. This fixes that bug, which fixes the end-user bug.

https://github.com/user-attachments/assets/ab2f193f-943a-4fe7-82f3-39781cf5cc61

https://github.com/user-attachments/assets/356c4a9f-4eb3-491c-82df-1d01cc8bf86d

### DIFF
--- a/apps/admin-x-framework/src/hooks/use-filterable-api.ts
+++ b/apps/admin-x-framework/src/hooks/use-filterable-api.ts
@@ -2,6 +2,16 @@ import {useRef} from 'react';
 import {apiUrl, useFetchApi} from '../utils/api/fetch-api';
 import {Meta} from '../utils/api/hooks';
 
+const filterData = <
+    Data extends {[k in FilterKey]: string},
+    FilterKey extends keyof Data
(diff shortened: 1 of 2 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Build TS packages`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22871866742)
**What failed:** Failure to build @tryghost/admin-x-framework package
**Why:** The root cause of the failure is a failure to build the @tryghost/admin-x-framework package, which is likely due to a code issue in that package. This is a soft failure because it points to a problem in the application code that the author should fix.
**Action:**
The author should investigate the build failure for the @tryghost/admin-x-framework package and make the necessary code changes to resolve the issue.

9larsons (approved):
Looks great! Fixes a few other parts of Admin, too, as an added bonus.

PR #26744, opened 2026-03-09
Updated CI to support multi-source builds from same-org forks

rob-ghost:
The CI workflow currently hardcodes `TryGhost/Ghost` as the only repository allowed to push Docker images to GHCR and trigger the Ghost-Moya CD pipeline. This means private forks like Ghost-Security — which share the same org and secrets — can't produce builds that flow through the standard Pro CD path. Their images end up as ephemeral workflow artifacts with no downstream consumer.

This changes the GHCR push guard from an exact repository match to an org-level check (`repository_owner == 'TryGhost'`), so any same-org repo can push to GHCR. To avoid image name collisions, the GHCR image name is now derived from the repository name rather than hardcoded — `TryGhost/Ghost` continues to push to `ghcr.io/tryghost/ghost-core` while `TryGhost/Ghost-Security` would push to `ghcr.io/tryghost/ghost-security-core`. The same derivation applies to the full image and build cache references.

Separately, the admin and server build artifacts are now uploaded unconditionally regardless of GHCR push strategy. Previously, external fork PRs got the server `.tar.gz` artifact but not the admin build — an oversight where the Docker push strategy flag was reused to gate an unrelated artifact upload. Now every CI run (including external forks) produces both artifacts, enabling downstream consumers like Ghost-Moya to build from any fork given a successful CI run.

The `trigger_cd` job also passes `source_repo` in the dispatch payload so Ghost-Moya can resolve SHAs, discover artifacts, and report commit statuses against the correct source repository rather than assuming `TryGhost/Ghost`. The companion PR on Ghost-Moya adds the receiving side of this parameter.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -944,23 +944,39 @@ jobs:
       - name: Determine push strategy
         id: strategy
         run: |
-          # Use artifact-based image transfer (instead of GHCR) for fork PRs
-          # and non-main repos (e.g. Ghost-Security)
+          # Same-org repos (e.g. TryGhost/Ghost, TryGhost/Ghost-Security) push to GHCR.
+          # External forks and cross-repo PRs use artifact-based image transfer instead.
           USE_ARTIFACT="false"
-          if [ "${{ github.repository }}" != "TryGhost/Ghost" ]; then
+          if [ "${{ github.repository_owner }}" != "TryGhost" ]; then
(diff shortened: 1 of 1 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Determine changed packages`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22874034455)
**Action:**

PR #26745, opened 2026-03-09
Changed skeleton component to use an integer-based key instead of a UUID

cmraible:
no ref

When running Ghost in development with `yarn dev` in a remote VM, the analytics app crashes when loading over a plaintext http connection because `crypto.randomUUID` is only available in a secure context (https or localhost). 

`crypto.randomUUID` is also overkill for React keys anyways, so this changes to using integers instead, to make it easier to develop Ghost in a remote VM environment.


cmraible:
@coderabbit review

cmraible:
@CodeRabbit review

cmraible:
@troyciesco no rush at all this, just a random little one I came across in my adventures

troyciesco (approved):

PR #26748, opened 2026-03-10
Fixed retention offer not rendering for 1 mo repeating offers

sagzy:
closes https://linear.app/ghost/issue/BER-3413

- when a member signs ups with a repeating 1 month offer, the next payment is not discounted
- therefore, they should be able to redeem a retention offer


ErisDS:
- Workflow: `CI`
- Failed Step: `E2E tests`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22900685726)
**Action:**

mike182uk (approved):

PR #26755, opened 2026-03-10
🐛 Fixed N+1 queries and broken reply limit in comments API

rob-ghost:
Comments load time was slow due to an N+1 query pattern in the Comment model's `findPage()` method. Profiling against seeded data (5000 comments across 100 posts) showed the browse endpoint executing 26 SQL queries per request, the read endpoint hitting 83 queries, and the replies endpoint running 19 queries.

The N+1 issue came from a sequential `for...await model.load()` loop that loaded reply relations one comment at a time. Each comment in the page triggered separate queries for its replies, reply members, reply in_reply_to references, and reply counts. This has been replaced with a single `Collection.load()` call that batches all reply relations across all comments in the page. Additionally, the reply relations were being double-loaded — included in both the base `findPage` query's `withRelated` and then re-loaded in the loop — which has been fixed by separating the relations before the base query runs.

| Endpoint | Before | After | Reduction |
|----------|--------|-------|-----------|
| Browse (15 comments) | 26 queries | 7 queries | 73% |
| Read (30 replies) | 83 queries | 5 queries | 94% |
| Replies (15 replies) | 19 queries | 4 queries | 79% |

### Reply limit tradeoff

The `replies()` relation previously had `.query('limit', 3)` which **did work** with the old per-model `model.load()` approach — each call issued a separate `SELECT ... WHERE parent_id = ? LIMIT 3` query scoped to a single parent, so the limit correctly capped replies to 3 per comment.

However, `LIMIT 3` is incompatible with the batch `Collection.load()` approach, which fetches all replies in a single `WHERE parent_id IN (...) LIMIT 3` query — the limit would apply to the total result set across all parents rather than per-parent.

Rather than implementing a more complex per-parent trimming step after the batch load, we've removed the server-side limit entirely. **This is a deliberate tradeoff**: the API now returns all replies inline (increasing payload size for comments with many replies) in exchange for significantly fewer database queries. The frontend's `Replies` component handles the display limit client-side — showing 3 initially with a "Show more" button that reveals the rest from already-loaded data without an additional API call.

ref https://linear.app/tryghost/issue/ONC-1544/

### DIFF
--- a/ghost/core/core/server/models/comment.js
+++ b/ghost/core/core/server/models/comment.js
@@ -52,9 +52,7 @@ const Comment = ghostBookshelf.Model.extend({
 
     replies() {
         return this.hasMany('Comment', 'parent_id', 'id')
-            .query('orderBy', 'created_at', 'ASC')
-            // Note: this limit is not working
-            .query('limit', 3);
+            .query('orderBy', 'created_at', 'ASC');
     },
 
     // Called by our filtered-collection bookshelf plugin
@@ -210,7 +208,6 @@ const Comment = ghostBookshelf.Model.extend({
                     options.withRelated = [
                         // Relations
                         'member', 'in_reply_to', 'count.replies', 'count.direct_replies', 'count.likes', 'count.liked',
-                        // Replies (limited to 3)
                         'replies', 'replies.member', 'replies.in_reply_to',
                         'replies.count.direct_replies', 'replies.count.likes', 'replies.count.liked'
                     ];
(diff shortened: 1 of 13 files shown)


ghost/core/core/server/models/comment.js
```diff
@@ -52,9 +52,7 @@ const Comment = ghostBookshelf.Model.extend({
 
     replies() {
         return this.hasMany('Comment', 'parent_id', 'id')
-            .query('orderBy', 'created_at', 'ASC')
-            // Note: this limit is not working
```
kevinansfield, line 56: 🫠

ErisDS:
- Workflow: `CI`
- Failed Step: `Load Image`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22906737039)
**What failed:** Operation timed out
**Why:** Infrastructure issue detected: operation timed out. This is typically not caused by code changes.
**Action:**
Try re-running the workflow. If it persists, investigate the infrastructure.

ErisDS:
- Workflow: `CI`
- Failed Step: `Run yarn nx run @tryghost/comments-ui:test`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22904570794)
**Action:**

PR #26761, opened 2026-03-10
Changed the E2E isolation model defaults

jonatansberg:
## Summary
- default E2E isolation to per-file and add a root-level `test.describe.configure({mode: 'parallel'})` opt-in for per-test isolation
- add fixture mode resolution, mode guard rails, `resolvedIsolation`, and `resetEnvironment()` to support mixed isolation safely
- support both local workflows by auto-selecting build vs dev mode and opt stateful member/billing suites back into per-test isolation

## Performance
- Over the last 3 days, Velo reports Ghost CI averaging 29.7 minutes per successful run. The latest successful run on this branch completed in 22.6 minutes, a 7.1 minute improvement (23.8%). 

- Looking specifically at the E2E matrix, the last 10 successful CI runs averaged a 17.8 minute critical path across the 4 E2E shards; this branch completed that path in 13.1 minutes, a 4.7 minute improvement (26.2%). 

- Total E2E shard compute time also dropped from 66.9 minutes on average to 45.6 minutes, reducing runner time by 21.3 minutes (31.9%).

ref https://linear.app/ghost/issue/BER-3428/rework-e2e-isolation-model


ErisDS:
- Workflow: `CI`
- Failed Step: `Set up Docker Buildx`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22911238823)
**Action:**

9larsons:
Planning to look at this tomorrow morning!

9larsons:
Ok, I pushed some changes - we can revert if desired, it just felt easier to push them in for now.

The main gripe was the monkey patching. Even if valid, it felt fragile and, well, icky. Afaict the main reason to do this was to avoid multiple commands for the test-writer to avoid needing to call `test.use({isolation: 'per-test'});` along with `mode: parallel`. I created a helper for this.

The other function of monkey patching was avoiding the deprecated `test.describe.parallel` and `test.describe.serial` so I've created an ESLint rule to capture those instead.

Overall, I'd still like to maintain full isolation, but agree it's:
1) A performance gain we sorely need atm
2) Not terribly difficult to reason about per-test-file state (retries can cause problems here, sometimes)

I think the next point of interest would be not needing to fully restart the Ghost container after db reset, which might buy us back some of the gap here with many restarts per file, or just be additional gains.

9larsons (approved):
Approved as long as you're ok with the changes I pushed - just re-request review if you'd like to discuss more.

PR #26763, opened 2026-03-10
Changed users.bio soft limit from 250 to 2000

PaulAdamDavis:
no ref

Increases the soft limit on users.bio from 250 to 2000 characters

https://github.com/TryGhost/Ghost/commit/c75ab5317ab3e3e5f16200bdc32844d9a952c5b4 was used as an implementation reference

### DIFF
--- a/apps/admin-x-settings/test/acceptance/general/users/profile.test.ts
+++ b/apps/admin-x-settings/test/acceptance/general/users/profile.test.ts
@@ -44,13 +44,13 @@ test.describe('User profile', async () => {
         await modal.getByRole('button', {name: 'Save'}).click();
         await expect(modal).toContainText('Enter a valid email address');
 
-        // Test location validation
-        await modal.getByLabel('Location').fill(new Array(195).join('a'));
+        // Test location validation, limited to 150 characters
+        await modal.getByLabel('Location').fill('a'.repeat(151));
         await modal.getByRole('button', {name: 'Save'}).click();
         await expect(modal).toContainText('Location is too long');
 
-        // Test bio validation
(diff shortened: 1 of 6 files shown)


apps/admin-x-settings/test/acceptance/general/users/profile.test.ts
```diff
@@ -50,7 +50,7 @@ test.describe('User profile', async () => {
         await expect(modal).toContainText('Location is too long');
 
         // Test bio validation
-        await modal.getByLabel('Bio').fill(new Array(255).join('a'));
+        await modal.getByLabel('Bio').fill(new Array(2005).join('a'));
```
EvanHahn: nit: I'd change this to 2001 to be more exact. (Not exactly new as part of your work.)

ErisDS:
- Workflow: `CI`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22922844333)
**What failed:** Test assertion failed
**Action:**

EvanHahn (changes requested):
Looks like tests are failing. Also, I recommend implementing CodeRabbit's comments.

And does this need a database migration?

LGTM otherwise!

apps/admin-x-settings/test/acceptance/general/users/profile.test.ts
PaulAdamDavis: Turns out that `new Array(2005).join('a')` makes a string of 2004 characters, so changing this to `2001` would actually be valid. `'a'.repeat(2001)` _does_ create a string of the length you would expect from the int, so I think it makes sense to change, which I have done.

EvanHahn (approved):

PR #26764, opened 2026-03-10
Update validateSiteUuid to be synchronous

EvanHahn:
no ref

This function never `await`s or returns a promise. Let's make it synchronous.


cmraible (approved):
Good call, thanks for finding and improving this!

PR #26765, opened 2026-03-10
Modernize frontend apps service initializer

EvanHahn:
no ref

This change should have no user impact. It just takes advantage of `async` functions for a little clarity.

It has a *slight* functionality change: it now returns `Promise<void>` instead of `Promise<Array>`. We never used the return value so this is fine.


EvanHahn:
I don't think this is worth anyone's time. Closing.

PR #26766, opened 2026-03-10
Cleaned up tier repository initializer

EvanHahn:
no ref

This change should have no user impact. It:

- Creates `Tier` models in parallel, for performance
- Uses `Set` instead of an object

This should slightly improve boot time and makes the code a bit clearer.


ErisDS:
- Workflow: `CI`
- Failed Step: `Run unit tests`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22925982767)
**Action:**

ghost/core/core/server/services/tiers/tier-repository.js
```diff
@@ -31,16 +31,14 @@ module.exports = class TierRepository {
     }
 
     async init() {
-        this.#store = [];
-        this.#ids = {};
```
cmraible, line 35: nit: if `init()` is called twice (as it is in test fixture utils [here](https://github.com/TryGhost/Ghost/blob/aed859f10ed263f392196e80f17e066c09c38c70/ghost/core/test/utils/fixture-utils.js#L977-L979)), this could leave stale ids in the Set since we're no longer explicitly clearing `this.#ids`. 

Unlikely to cause any real problems I think, but a slight unintended behavioral change.

Adding a `this.#ids.clear()` at the beginning of `init()` would solve this, but I'm wouldn't consider this blocking

cmraible (approved):
Approved with a non-blocking comment. Good find!

ghost/core/core/server/services/tiers/tier-repository.js
EvanHahn: Good catch. Addressed in d8bad23b810df63c134d694cda8e2c4104f841d3.

PR #26769, opened 2026-03-11
🎨 Added separate Server and Admin build versions to About dialog

rob-ghost:
On Ghost(Pro), the About dialog currently shows a single version string that doesn't distinguish between the server container and the admin client — both ship independently and can be at different commits. This makes it harder to debug issues or verify which code is actually running.

This change introduces two new optional build-time inputs: `GHOST_BUILD_VERSION` (a Docker build arg baked as an ENV var, read by the public-config endpoint at runtime) and `VITE_ADMIN_BUILD_VERSION` (a Vite define constant baked into the admin bundle). When both are set — as they will be on Ghost(Pro) CI builds — the About dialog shows separate "Server" and "Admin" lines, each linking to their respective GitHub commit. When neither is set (self-hosted installs, tagged releases), the dialog continues to show a single "Version" line pointing at the release tag, identical to today's behavior.

The `link-to-github-releases` utility was extended to handle the new `semver+epoch.sha` (server) and `semver+sha` (admin) formats, alongside the existing legacy `+moya` and git-describe formats. New test cases cover all format variations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -934,7 +934,14 @@ jobs:
           DEPENDENCY_CACHE_KEY: ${{ needs.job_setup.outputs.dependency_cache_key }}
 
       - name: Build server and admin assets
-        run: yarn build:production
+        run: |
+          PKG_VERSION=$(node -p "require('./ghost/core/package.json').version")
+          SHORT_SHA="${GITHUB_SHA:0:7}"
+          if [ "${{ github.ref_type }}" != "tag" ]; then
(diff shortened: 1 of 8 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `Run yarn nx run @tryghost/admin-x-settings:test:acceptance`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/22930152912)
**Action:**

cmraible (approved):
Nice one, thanks!

PR #26801, opened 2026-03-12
Added `members_single_opt_in` computed setting for single opt-in signup

mike182uk:
no ref

This is the first step toward native single opt-in support. The setting is derived from the `hostSettings:membersSingleOptIn:enabled` config value (managed by host) and exposed to frontends via the Content API public settings response.


mike182uk:
The work for this change is currently postponed

PR #26804, opened 2026-03-12
Added composite index on offer_redemptions(offer_id, created_at)

kevinansfield:
## Summary

- Adds a composite index on `offer_redemptions(offer_id, created_at)` to optimize the query in `mapToOffer` that finds the last redemption date per offer
- `mapToOffer` is called for every offer when listing offers via the Admin API (`GET /offers/`), when checking available offers for members (`listOffersAvailableToSubscription`), and on boot for redirect setup — each call queries `offer_redemptions` filtered by `offer_id` and ordered by `created_at DESC`, currently causing unindexed filesorts
- The composite index lets MySQL satisfy both the filter and sort directly from the index, and the existing single-column FK index on `offer_id` is subsumed by the leftmost prefix of the new index

### DIFF
--- a/ghost/admin/package.json
+++ b/ghost/admin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ghost-admin",
-  "version": "6.21.2",
+  "version": "6.22.0-rc.0",
   "description": "Ember.js admin client for Ghost",
   "author": "Ghost Foundation",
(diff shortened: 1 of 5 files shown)


kevinansfield:
Closing as this is no longer a hot query path. We can re-evaluate in future as needed

PR #26809, opened 2026-03-12
Cleaned up outbox entries when welcome email is toggled inactive

cmraible:
closes https://linear.app/ghost/issue/NY-835/cleanup-what-happens-to-outbox-rows-when-an-email-is-toggled-between

## Summary

- **Proactive cleanup**: When a welcome email template is toggled from active to inactive, pending outbox entries for that template are now deleted from the database immediately
- **Graceful skip**: In the outbox handler, if a template becomes inactive after an entry was created, the handler skips sending and allows the entry to be deleted naturally
- **Tests**: Added unit test to verify the handler skips sending inactive templates

This prevents zombie FAILED rows from accumulating in the outbox table when templates are disabled.


ErisDS:
- Workflow: `CI`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/23016288329)
**What failed:** Test assertion failed
**Action:**

PR #26810, opened 2026-03-12
Added subscription lifecycle e2e tests for Stripe webhooks

9larsons:
## Summary
- Added e2e tests for subscription cancellation, deletion, and invoice payment webhook flows
- Extended the fake Stripe server to support `expand[]` query params on subscription lookups
- Added `cancelSubscription`, `deleteSubscription`, and `sendInvoicePaymentSucceeded` methods to the Stripe test service
- Extended `MembersService` with `getByEmailWithSubscriptions` for asserting subscription state

## Test plan
- [x] All 4 Stripe webhook tests pass locally (`yarn test tests/admin/members/stripe-webhooks.test.ts`)
- [x] TypeScript types check (`yarn test:types`)
- [x] Linting passes (`yarn lint`)

### DIFF
--- a/e2e/helpers/services/members/members-service.ts
+++ b/e2e/helpers/services/members/members-service.ts
@@ -1,7 +1,31 @@
 import {HttpClient as APIRequest, Member} from '@/data-factory';
 
+export interface MemberSubscription {
+    id: string;
+    status: string;
+    cancel_at_period_end: boolean;
+    current_period_end: string;
+    start_date: string;
+    default_payment_card_last4: string | null;
+    plan: {
+        id: string;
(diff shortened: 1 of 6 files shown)


9larsons:
Picked up the test nit and folded in two follow-ups while I was there:

- added a subscription length assertion before dereferencing `member.subscriptions[0]` in the cancel-at-period-end test
- fixed the fake Stripe subscription route to read Stripe's indexed `expand` query encoding (`expand[0]=...`), not just `expand[]`
- tightened the invoice webhook test so it waits for the resulting `payment_event` instead of only re-checking the already-paid member status

PR #26819, opened 2026-03-14
Replaced cheerio with lightweight html-utils

kevinansfield:
## Summary
- Replaced the direct `cheerio` (0.22) dependency with a thin `html-utils` utility built on `htmlparser2`, `css-select`, `domutils`, and `dom-serializer` — the same underlying packages cheerio wraps
- The cheerio 0.22 dependency causes yarn hoisting conflicts with the `cheerio ^1.0` that `juice` requires, leading to version resolution issues across the monorepo
- Companion to https://github.com/TryGhost/SDK/pull/821 which does the same for `@tryghost/url-utils`

## Changes
- **New:** `ghost/core/core/server/lib/html-utils.js` — lightweight HTML manipulation utility providing `load()` and `select()` functions with a jQuery-like API (`attr`, `remove`, `addClass`, `each`, `before`, `text`, `html`, etc.)
- **6 production files** updated to use `html-utils` instead of `cheerio`: email-renderer, mention-sending-service, mention-discovery-service, mention.js, oembed-service, generate-feed
- **14 test files** updated to use `html-utils`
- **package.json:** Removed `cheerio: 0.22.0`, added `css-select`, `dom-serializer`, `domhandler`, `domutils`, `htmlparser2`
- **Snapshot updates:** Entity encoding changed from hex format (cheerio 0.22 artifact) to the original format used in source code — e.g. `&#x2007;` → `&#8199;`, `&#xAD;` → `&shy;`. Same characters, different (now correct) encoding. Additionally, `juice` now correctly resolves to cheerio 1.0 (its declared dep) instead of accidentally picking up 0.22 via hoisting.

## Benchmark results

Tested cheerio 0.22 vs html-utils across all Ghost usage patterns:

```
Scenario                          Speedup
OEmbed link discovery              2.3x
Extract links (mentions)           5.8x
Mention verification               6.8x
Segment extraction                 5.9x
Full email post-processing         2.1x
Segment filtering + serialize      2.4x
RSS feed generation               18.8x
```


ErisDS:
- Workflow: `CI`
- Failed Step: `E2E tests`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/23096522715)
**Action:**

ErisDS:
- Workflow: `CI`
- Failed Step: `E2E tests`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/23106877624)
**What failed:** Type error in code
**Action:**

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+        it('$.html() serializes the full document', function () {
+            const $ = load('<p>Hello</p>');
+            assert.ok($.html().includes('<p>Hello</p>'));
+        });
+
+        it('$.text() returns text content of the document', function () {
+            const $ = load('<p>Hello</p> <p>World</p>');
+            assert.ok($.text().includes('Hello'));
+            assert.ok($.text().includes('World'));
+        });
```
kevinansfield: This should test the full return value, currently `.text()` could be returning HTML and these assertions would still match

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+            const $ = load('<p>Hello</p>');
+            assert.equal(typeof $, 'function');
+            assert.equal(typeof $.html, 'function');
+            assert.equal(typeof $.text, 'function');
+        });
+
+        it('$.html() serializes the full document', function () {
+            const $ = load('<p>Hello</p>');
+            assert.ok($.html().includes('<p>Hello</p>'));
+        });
```
kevinansfield: Lets use a more complex document than a single element.

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+const {load, select} = require('../../../../core/server/lib/html-utils');
+
+describe('html-utils', function () {
+    describe('load', function () {
+        it('returns a query function with html() and text() methods', function () {
+            const $ = load('<p>Hello</p>');
+            assert.equal(typeof $, 'function');
+            assert.equal(typeof $.html, 'function');
+            assert.equal(typeof $.text, 'function');
+        });
```
kevinansfield: This test is a little pointless if we're already testing the functions separately. We don't have a test of the top-level `$` function

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+
+        it('preserves HTML entities with decodeEntities: false', function () {
+            const $ = load('<div>&#x2007;&#xA0;&shy;&nbsp;</div>');
+            const html = $.html();
+            assert.ok(html.includes('&#x2007;'));
+            assert.ok(html.includes('&#xA0;'));
+            assert.ok(html.includes('&shy;'));
+            assert.ok(html.includes('&nbsp;'));
+        });
```
kevinansfield: What's the behaviour without `decodeEntities: false`?

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+            const links = $('a');
+            assert.equal(links.length, 2);
+        });
+
+        it('wraps a raw element when passed a non-string', function () {
+            const $ = load('<p>Hello</p>');
+            const el = $('p').get(0);
+            const wrapped = $(el);
+            assert.equal(wrapped.text(), 'Hello');
+        });
```
kevinansfield: This is difficult to understand, try to better capture the intent of "wrap". Assert the HTML rather than the text so it's clearer what element(s) we're dealing with.

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+                const html = $.html();
+                assert.equal(html.match(/target="_blank"/g).length, 2);
+            });
+        });
+
+        describe('removeAttr', function () {
+            it('removes attribute from elements', function () {
+                const $ = load('<div data-gh-segment="free">content</div>');
+                $('[data-gh-segment]').removeAttr('data-gh-segment');
+                assert.ok(!$.html().includes('data-gh-segment'));
```
kevinansfield: Assert the full html string

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+            });
+        });
+
+        describe('remove', function () {
+            it('removes elements from DOM', function () {
+                const $ = load('<div><p class="keep">keep</p><p class="remove">remove</p></div>');
+                $('p.remove').remove();
+                assert.ok($.html().includes('keep'));
+                assert.ok(!$.html().includes('remove'));
```
kevinansfield: Assert the full html string

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+            });
+        });
+
+        describe('addClass', function () {
+            it('adds class to elements', function () {
+                const $ = load('<figcaption>caption</figcaption>');
+                $('figcaption').addClass('kg-card-figcaption');
+                assert.ok($.html().includes('class="kg-card-figcaption"'));
```
kevinansfield: Assert the full html string

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+            });
+
+            it('does not duplicate existing class', function () {
+                const $ = load('<p class="existing">text</p>');
+                $('p').addClass('existing');
+                assert.equal($('p').get(0).attribs.class, 'existing');
```
kevinansfield: Assert the full html string

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+            });
+
+            it('appends to existing classes', function () {
+                const $ = load('<p class="a">text</p>');
+                $('p').addClass('b');
+                assert.equal($('p').get(0).attribs.class, 'a b');
```
kevinansfield: Assert the full html string

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+            it('returns empty string for empty set', function () {
+                const $ = load('<p>Hello</p>');
+                assert.equal($('div').html(), '');
+            });
+
+            it('sets inner HTML', function () {
+                const $ = load('<div>old</div>');
+                $('div').html('<p>new</p>');
+                assert.ok($.html().includes('<p>new</p>'));
+                assert.ok(!$.html().includes('old'));
```
kevinansfield: Assert the full html string

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+                const html = $.html();
+                assert.ok(html.indexOf('img') < html.indexOf('first'));
+            });
+
+            it('handles multi-node inserts with correct sibling links', function () {
+                const $ = load('<p>target</p>');
+                $('p').before('<span>A</span><span>B</span>');
+                const html = $.html();
+                assert.ok(html.indexOf('A') < html.indexOf('B'));
+                assert.ok(html.indexOf('B') < html.indexOf('target'));
```
kevinansfield: Assert the full html string

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,338 @@
...
+        describe('element mutation', function () {
+            it('supports direct tagName mutation', function () {
+                const $ = load('<figure><figcaption>test</figcaption></figure>');
+                $('figure, figcaption').each((i, elem) => {
+                    elem.tagName = 'div';
+                });
+                const html = $.html();
+                assert.ok(html.includes('<div>'));
+                assert.ok(!html.includes('<figure>'));
+                assert.ok(!html.includes('<figcaption>'));
```
kevinansfield: Assert the full html string

ghost/core/test/e2e-api/admin/__snapshots__/email-previews.test.js.snap
EvanHahn, line 1: Why did this file change?

ghost/core/test/unit/server/services/email-service/email-renderer.test.js
```diff
@@ -3479,7 +3479,7 @@ describe('Email renderer', function () {
             assert(!response.html.includes('members only section'));
             assert(response.html.includes('some text for both'));
             assert(!response.html.includes('finishing part only for members'));
-            assert(response.html.includes('Devenez un(e) abonn&#xE9;(e) payant de Cathy&#39;s Blog pour acc&#xE9;der &#xE0; du contenu exclusif'));
+            assert(response.html.includes('Devenez un(e) abonné(e) payant de Cathy&#39;s Blog pour accéder à du contenu exclusif'));
```
EvanHahn, line 3482: I'm a little concerned about this behavior change. I admit I know little about the email renderer, but I think escaped HTML entities are probably more reliable than full Unicode characters. Maybe this is fine, but wanted to flag.

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,350 @@
...
+            assert.equal($(null).length, 0);
+            assert.equal($(undefined).length, 0);
+        });
+
+        it('$ wraps an array of elements', function () {
+            const $ = load('<p>A</p><p>B</p>');
+            const elements = $('p').get();
+            const wrapped = $(elements);
+            assert.equal(wrapped.length, 2);
+            assert.equal(wrapped.attr('href'), undefined);
```
EvanHahn, line 25: nit: do we need this assertion?

```suggestion
```

ghost/core/core/server/lib/html-utils.js
```diff
@@ -0,0 +1,219 @@
...
+ *   const {load} = require('./lib/html-utils');
+ *   const $ = load(html);
+ *   $('a').each((i, el) => { ... });
+ *   const result = $.html();
+ */
+
+class WrappedSet {
+    constructor(elements, root) {
+        this._elements = Array.isArray(elements) ? elements : [elements];
+        this._root = root;
```
EvanHahn, line 21: nit: can we use proper JavaScript private members (`this.#elements` and `this.#root`) instead?

ghost/core/core/server/lib/html-utils.js
EvanHahn, line 1: I'd love to see JSDoc types for this file. Happy to help with that.

ghost/core/core/server/lib/html-utils.js
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    attr(name, value) {
+        if (value === undefined) {
+            // Getter
+            const el = this._elements[0];
+            if (!el || !el.attribs) {
+                return undefined;
+            }
+            return el.attribs[name];
```
EvanHahn, line 62: nit: Do we want to protect against keys on `Object.prototype`? For example, `$('a').get('hasOwnProperty')` could return an unexpected value.

ghost/core/core/server/lib/html-utils.js
```diff
@@ -0,0 +1,219 @@
...
+
+    map(fn) {
+        return this._elements.map((el, i) => fn(i, el));
+    }
+
+    attr(name, value) {
+        if (value === undefined) {
+            // Getter
+            const el = this._elements[0];
+            if (!el || !el.attribs) {
```
EvanHahn, line 59: question: when is `el.attribs` missing? This question applies everywhere it's checked in this file.

ghost/core/core/server/lib/html-utils.js
```diff
@@ -0,0 +1,219 @@
...
+            const el = this._elements[0];
+            const children = getChildren(el);
+            return render(children, {decodeEntities: false});
+        }
+        // Set inner HTML
+        for (const el of this._elements) {
+            const parsed = parseDocument(content, {decodeEntities: false});
+            const newChildren = getChildren(parsed);
+            el.children = newChildren;
+            for (let j = 0; j < newChildren.length; j++) {
```
EvanHahn, line 125: nit: I've only seen `j` used as an iteration index when `i` has already been used. Maybe we could change this to `i`?

ghost/core/core/server/lib/html-utils.js
```diff
@@ -0,0 +1,219 @@
...
+            const idx = parent.children.indexOf(el);
+            parent.children.splice(idx, 0, ...newNodes);
+
+            // Rebuild prev/next links only for the affected window
+            const end = Math.min(parent.children.length, idx + newNodes.length + 1);
+            for (let i = Math.max(0, idx - 1); i < end; i++) {
+                const node = parent.children[i];
+                node.parent = parent;
+                node.prev = i > 0 ? parent.children[i - 1] : null;
+                node.next = i < parent.children.length - 1 ? parent.children[i + 1] : null;
```
EvanHahn, line 164: nit: I wonder if this could be DRYed out with what's above in `html()`.

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,350 @@
...
+            assert.equal(result.length, 0);
+        });
+    });
+
+    describe('WrappedSet', function () {
+        describe('get', function () {
+            it('returns all elements when called without index', function () {
+                const $ = load('<p>1</p><p>2</p>');
+                const elements = $('p').get();
+                assert.equal(elements.length, 2);
```
EvanHahn, line 84: nit: could we also assert on what's returned here, beyond the length? A check for `tagName` like you've done below is probably enough.

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,350 @@
...
+                assert.equal(links[1].attribs.href, 'b');
+            });
+        });
+
+        describe('toArray', function () {
+            it('returns a copy of the elements array', function () {
+                const $ = load('<p>1</p><p>2</p>');
+                const arr = $('p').toArray();
+                assert.equal(arr.length, 2);
+                assert.notEqual(arr, $('p').get());
```
EvanHahn, line 114: nit: these will never be equal because `notEqual` checks reference equality.

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,350 @@
...
+                assert.deepEqual(hrefs, ['a', 'b']);
+            });
+        });
+
+        describe('attr', function () {
+            it('gets attribute value', function () {
+                const $ = load('<a href="http://example.com">link</a>');
+                assert.equal($('a').attr('href'), 'http://example.com');
+            });
+
```
EvanHahn, line 163: nit: I'd love to see tests for:

- Getting an empty attribute value (e.g., `<a href="">link</a>`)
- What happens when the same attribute is duplicated (e.g., `<a href="x" href="y">link</a>`)
- If capitalization is affected (e.g., `<a hREF="#foo">link</a>`)

ghost/core/core/server/lib/html-utils.js
```diff
@@ -0,0 +1,219 @@
...
+        for (const el of this._elements) {
+            removeElement(el);
+        }
+        return this;
+    }
+
+    addClass(className) {
+        for (const el of this._elements) {
+            if (el.attribs) {
+                const existing = el.attribs.class || '';
```
EvanHahn, line 92: nit: the ternary on the line below makes this `|| ''` unnecessary.

EvanHahn (changes requested):
Partial review. Will read more soon.

ghost/core/test/e2e-api/admin/__snapshots__/email-previews.test.js.snap
kevinansfield: The email output switched over to the same character encoding we used in the source files and matches the expected `juice` behaviour when it hasn't had its `cheerio` dependency hoisted from 1.0.0 back to 0.22.

ghost/core/core/server/lib/html-utils.js
```diff
@@ -0,0 +1,219 @@
+const {parseDocument} = require('htmlparser2');
```
EvanHahn, line 1: Based on my reading of the Cheerio source, it uses [Parse5](https://parse5.js.org/modules/parse5.html) unless you're in XML mode. Why use `htmlparser2` here?

ghost/core/test/unit/server/lib/html-utils.test.js
```diff
@@ -0,0 +1,350 @@
...
+                $('p').first().before('<img src="test.jpg">');
+                assert.equal($.html(), '<img src="test.jpg"><p>first</p><p>second</p>');
+            });
+
+            it('handles multi-node inserts with correct sibling links', function () {
+                const $ = load('<p>target</p>');
+                $('p').before('<span>A</span><span>B</span>');
+                assert.equal($.html(), '<span>A</span><span>B</span><p>target</p>');
+            });
+
```
EvanHahn, line 272: nit: could you add a test like this?

```js
it('inserts before each match', function () {
  const $ = load('<p>a</p><p>b</p>');
  $('p').before('<span>x</span>');
  assert.equal($.html(), '<span>x</span><p>a</p><span>x</span><p>b</p>');
});
```

EvanHahn (changes requested):

kevinansfield:
The incompatible cheerio/juice versioning that was the driver for investigating the viability of dropping cheerio has been resolved in recent versions and made easier with the adoption of pnpm.

Closing this as the API surface layer required to replace our cheerio usage was fairly large without additional refactors and introduced maintenance complexity for minor performance benefits in areas that aren't known to be bottlenecks.

PR #26825, opened 2026-03-16
Converted em line-height values to unitless in email templates

minimaluminium:
ref https://linear.app/ghost/issue/DES-1306/text-size-issues-when-using-accessibility-features-on-iphones

When iPhone users increase text size via iOS Dynamic Type, Ghost newsletter emails have excessively large line spacing, making them nearly unreadable. The root cause is em-based line-height values in the email template — when Dynamic Type scales font sizes, em line-heights compound the scaling.                                                                                                                              
                                                                                                                                                                                                                     
This converts all line-height: Xem values to unitless line-height: X in the email stylesheet. Unitless values are the recommended best practice for email accessibility, as they scale proportionally without compounding.

### DIFF
--- a/ghost/core/core/server/services/email-rendering/partials/base-styles.hbs
+++ b/ghost/core/core/server/services/email-rendering/partials/base-styles.hbs
@@ -103,7 +103,7 @@ table td {
     padding-top: 10px;
     padding-left: 30px;
     padding-right: 30px;
-    line-height: 1.5em;
+    line-height: 1.5;
 }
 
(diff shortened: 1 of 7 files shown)


ErisDS:
- Workflow: `CI`
- Failed Step: `E2E tests`
- Run: [View failed run](https://github.com/TryGhost/Ghost/actions/runs/23133071954)
**Action:**

kevinansfield (approved):

PR #26837, opened 2026-03-16
🐛 Fixed "labels is not defined" error when creating comped members via API

troyciesco:
closes https://linear.app/tryghost/issue/ONC-1553

- The `add()` method in `memberBREADService` passed the full `options` object (including `withRelated: ['labels', 'newsletters']`) to `setComplimentarySubscription()`. 
- When the Stripe customer had existing subscriptions, fetching related stripeSubscriptions with those options caused a "labels is not defined on the model" error on StripeCustomerSubscription models. 
- This fix uses `sharedOptions` (context + transacting only), matching the pattern already used for `linkStripeCustomer()`.
---
### Summary

When creating a comped member via the Admin API with a `stripe_customer_id` pointing to a Stripe customer that has existing subscriptions, `setComplimentarySubscription()` received the full options object including `withRelated: ['labels', 'newsletters']`. Inside that method, `member.related('stripeSubscriptions').fetch(options)` attempted to eager-load `labels` on `StripeCustomerSubscription` models, which don't have that relation — throwing "labels is not defined on the model".

### Impact

The bug's impact depends on the Stripe customer's existing subscriptions:

- **Customer with a paid subscription + `comped: true`**: The comp conversion never runs. The member ends up with `paid` status instead of `comped`, and the API returns a 500. This is a real data issue — the member doesn't get their complimentary access.
- **Customer with an existing comp subscription**: All database records are written correctly by `linkStripeCustomer()` before the error occurs, so the member is properly set up. However, the API still returns a 500 to the caller.
- **Customer with no subscriptions**: The bug doesn't trigger because there are no `StripeCustomerSubscription` models to eager-load on.

### Fix

Uses `sharedOptions` (only `context` + `transacting`) instead of the full `options` when calling `setComplimentarySubscription()` on line 406 of `member-bread-service.js`. This matches the pattern already established for the `linkStripeCustomer()` call on line 381.


cmraible (approved):
LGTM! I just pushed a single e2e-api test to cover this, mostly just to prevent incidental regressions in the future 🚢

PR #26839, opened 2026-03-16
🐛 Fixed translations for Inbox Links button

EvanHahn:
closes https://linear.app/ghost/issue/ONC-1548
ref https://github.com/TryGhost/Ghost/pull/26308

| Before | After |
|---|---|
| Before | After |

We always showed English text on the Inbox Links button. Now we translate it properly.

This happened because we computed the text before changing the language and never updated it. Now we compute it at the last minute, which fixes the problem.

### DIFF
--- a/apps/portal/package.json
+++ b/apps/portal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/portal",
-  "version": "2.66.3",
+  "version": "2.66.4",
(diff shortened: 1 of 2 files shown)


EvanHahn:
@coderabbitai full review

9larsons (approved):
Good catch 👍.

PR #26844, opened 2026-03-17
Simplified browser filter syntax for members and comments

jonatansberg:
## Summary

- replaced live `filter=<nql>` browser state with browser-owned filter params for both members and comments
- kept NQL as an API output format only, with targeted one-way legacy member import for old Ember links
- aligned comments with the mainline browser-param contract instead of carrying a separate NQL compatibility path
- simplified the shared filter contract and removed dead parse-side code and unused exports

## Why this change

The current rewrite still carries a lot of complexity from using one format for three different jobs:

- browser URL state
- API filter syntax
- legacy deep-link compatibility

This spike separates those concerns.

The browser now owns its own filter state, members keep a small legacy import path for old Ember URLs, comments follow the shipped browser-param contract from `main`, and NQL is only generated when calling the backend.

That reduces live parse logic, removes a large amount of compatibility code from the normal runtime path, and makes the filter state hooks easier to reason about.

## Dependency

- depends on #26829
- base PR: #26829

## What changed

- added a shared browser filter param codec under `apps/posts/src/views/filters`
- changed members and comments filter state hooks to use browser params as the source of truth
- moved legacy NQL parsing into import-only helpers for members
- removed comment-side NQL import compatibility and matched the existing mainline comment URL/operator format
- reduced the field contract so parse behavior only exists where legacy import still needs it
- cleaned up some dead code and unused exports in the filter and members surfaces

## Browser filter format

The browser URL now uses named filter params keyed by field, with repeated params preserving repeated filter instances.

Examples:

- `?status=is:paid`
- `?label=is-any:[vip,alpha]&label=is-not-any:[test]`
- `?reported=is:true&post=is_not:post_123`
- `?search=jane`

Members still compile those filters to backend NQL and still import old Ember `filter=` links once before rewriting to the browser-owned format.

## Test plan

- `yarn --cwd apps/posts tsc --noEmit`
- focused `vitest` runs across the changed filters, members, and comments suites
- `knip --workspace apps/posts --production --no-exit-code`


jonatansberg:
Dropping as we want to keep nql filter format

PR #26849, opened 2026-03-17
Moved shared email CSS into a unified email-rendering service

9larsons:
ref https://linear.app/ghost/issue/NY-1115
- created card styles template
- created typography styles template
- updated welcome emails to use styles from newsletter renderer (not copy+paste)
- moved styles to generic `email-rendering` dir for reuse

Welcome emails and newsletter emails maintained ~1500 lines of duplicated Koenig card and typography CSS independently. Extracted shared typography and card styles into a new email-rendering service so both renderers draw from a single source of truth. Welcome emails now inherit the newsletter's more complete typography (proper heading margins, dark mode support, code styling). Also shared the juice CSS inlining + plaintext finalization logic.


ghost/core/core/server/services/email-rendering/partials/typography.hbs
```diff
@@ -0,0 +1,374 @@
+/* Shared email content typography */
```
EvanHahn: Two nits:

- I don't think this comment is necessary.
- "Typography" doesn't feel like a good name for this file—it does much more than that. Is there a better name?

ghost/core/core/server/services/email-rendering/partials/card-styles.hbs
```diff
@@ -1,316 +1,4 @@
...
-        color: rgba(0, 0, 0, 0.6);
-    {{/if}}
-}
-
-figcaption a {
-    text-decoration: none!important;
-}
-
-
+/* Shared Koenig card styles */
```
EvanHahn: nit: I don't think this comment is necessary.

ghost/core/core/server/services/email-rendering/partials/card-styles.hbs
EvanHahn, line 1: This will include a bunch of unnecessary styles in welcome emails. For example, it supports audio cards which aren't present. Is that okay?

EvanHahn (approved):
I didn't review the moves closely. I'm assuming that all the CSS is basically the same, but in different places.

Three non-blocking comments.

ghost/core/core/server/services/email-rendering/partials/card-styles.hbs
9larsons: We chatted - we're ok with this being fully opt-in to make it easier and less fragile.

PR #26868, opened 2026-03-18
Add newsletter setting for footer Share button

peterzimon:
ref https://linear.app/ghost/issue/FEA-480/native-share-buttons-with-link-based-referral-param-for-attribution

Adds a newsletter setting to show/hide the footer Share action and updates the settings preview footer actions UI to match the email layout.


ghost/core/core/server/data/migrations/versions/6.19/2026-03-17-12-00-00-add-newsletter-show-share-button.js
```diff
@@ -0,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
```
troyciesco, line 1: haven't fully reviewed yet but gonna mark "request changes" real quick so this doesn't get merged under an old version

troyciesco (changes requested):

ghost/core/test/unit/server/services/email-service/email-renderer.test.js
```diff
@@ -1451,6 +1455,30 @@ describe('Email renderer', function () {
...
+            const options = {};
+
+            const response = await emailRenderer.renderBody(
+                post,
+                newsletter,
+                segment,
+                options
+            );
+
+            assert(!response.html.includes('#/portal/share'));
```
troyciesco: i think this should be `#/share` now right? (i also wonder if there's a better way to check for the absence of sharing 🤔 )

ghost/core/test/integration/services/email-service/batch-sending.test.js
```diff
@@ -900,6 +900,38 @@ describe.skip('Batch sending tests', function () {
...
+            const defaultNewsletter = await getDefaultNewsletter();
+            await models.Newsletter.edit({show_share_button: true}, {id: defaultNewsletter.id});
+
+            try {
+                const {html} = await sendEmail(agent, {
+                    title: 'This is a test post title',
+                    mobiledoc: mobileDocExample
+                });
+
+                assert.match(html, /#\/portal\/share/);
```
troyciesco: a few more (old?) `#/portal/share` here and on L932

troyciesco:
now that we've looked at each branch individually, going to merge this one into its parent branch ([FEA-480/email-share-button](https://github.com/TryGhost/Ghost/tree/FEA-480/email-share-button)) so it's easier to add any finishing touches

PR #26869, opened 2026-03-18
Added file type validation and content-type control for file uploads

vershwal:
ref https://linear.app/ghost/issue/HKG-1643/

The files upload endpoint accepted any file type without validation, unlike images and media which check extension and MIME type. With the move to a shared CDN domain, we don't want to upload files with executable content types.

This PR adds:
- An allowlist of supported file extensions (based on Ghost Pro data)
- Extension-only validation middleware on the upload route
- Content-type resolution that serves browser-renderable types (images, PDF, JSON, audio, video, fonts) with their natural type, overrides formats (HTML, JS, CSS, XML) to text/plain, and defaults everything else to application/octet-stream (forced download)
- Users can zip unsupported file types as an escape hatch

### These are the lists and rules we have for extensions and content-types for files:-

**1. Allowed Extensions (48 total)**
These are the top 44 extensions from our Ghost Pro data. These are all the extensions that have count > 40 (and 4 more that were already added in Ghost so I didn't remove those from the allow list).
Total: 48 extensions, \~177,444 files

If someone uploads a file with an extension that's NOT in this list, the upload is rejected with a 415 error. They can always zip the file and upload as `.zip`.

### Once the file is uploaded, we decide on the content-type to store in the GCS. This is done in 3 phases:-

1. **Neutralized Extensions (served as text/plain)**
   These extensions can be executed by browsers, which is a security risk on a shared CDN domain. We override their content type to text/plain so browsers display them as plain text instead of executing them.

   ```yaml
   .html → text/plain
   .htm  → text/plain
   .js   → text/plain
   .css  → text/plain
   .xml  → text/plain
   ```
2. **Browser-Renderable (served with original content type)**
   These are safe file types where we keep the real content type so the browser can display/play them inline (e.g. show a PDF, play a video, render an image)

   ```yaml
   Images:  .jpg .jpeg .png .gif .webp → image/*
   PDF:     .pdf → application/pdf
   JSON:    .json → application/json
   Audio:   .mp3 .wav .m4a → audio/*
   Video:   .mp4 .mov → video/*
   Fonts:   .otf .woff .woff2 → font/*
   Text:    .txt → text/plain
   ```
3. **Everything else - Default** (e.g. .docx, .xlsx, .psd, .zip, .epub, .svg, etc.) is served as application/octet-stream, which forces the browser to download the file instead of trying to open it.


ghost/core/core/shared/config/overrides.json
```diff
@@ -46,33 +46,52 @@
             "extensions": [
                 ".pdf",
                 ".json",
-                ".jsonld",
-                ".odp",
                 ".ods",
```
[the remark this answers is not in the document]
vershwal: Thanks for this. Fixed here 4c895b0d923878661d54006ececd841c89b97815

ghost/core/test/e2e-api/admin/files.test.js
```diff
@@ -94,4 +95,72 @@ describe('Files API', function () {
         const fileArg = saveSpy.firstCall.args[0];
         assert.equal(fileArg.type, 'application/json', 'save() should receive the correct content type for JSON files');
     });
+
+    it('Rejects an unsupported file extension with 415', async function () {
+        const tempFile = path.join(os.tmpdir(), 'test.exe');
+        fs.writeFileSync(tempFile, 'not a real executable');
+
```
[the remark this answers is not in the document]
vershwal: Done 4c895b0d923878661d54006ececd841c89b97815

ghost/core/core/shared/config/overrides.json
```diff
@@ -46,34 +46,54 @@
             "extensions": [
                 ".pdf",
                 ".json",
-                ".jsonld",
-                ".odp",
```
vershwal, line 50: Removed these extensions because their count was very low in our data.

ghost/core/core/shared/config/overrides.json
```diff
@@ -46,34 +46,54 @@
             "extensions": [
                 ".pdf",
                 ".json",
-                ".jsonld",
```
allouis, line 49: IMO `.jsonld` is reasonable to continue support for

ghost/core/core/server/api/endpoints/files.js
```diff
@@ -14,7 +14,7 @@ const controller = {
             const filePath = await storage.getStorage('files').save({
                 name: frame.file.originalname,
                 path: frame.file.path,
-                type: mime.lookup(frame.file.originalname) || 'application/octet-stream'
+                type: getStorageContentType(frame.file.originalname)
```
allouis, line 17: I wonder, should this just be enforced at the file storage level, so that future callers can't bypass it? 

Maybe that's a larger change and not worth it it here 🤔

ghost/core/core/server/services/files/file-type-utils.js
```diff
@@ -0,0 +1,59 @@
+const path = require('path');
```
allouis: I think this file can be TS

allouis (commented):
Few comments!

ghost/core/core/shared/config/overrides.json
vershwal: I agree it's reasonable. Removed this because surprisingly we don't have any entry for `. jsonld` in our data. But I am okay to add this back.

ghost/core/core/server/services/files/file-type-utils.js
vershwal: Nice catch. Converted file-type-utils to TS in https://github.com/TryGhost/Ghost/pull/26869/changes/7371ec12c1355834b54f55c39355d99a94f16058

ghost/core/core/server/api/endpoints/files.js
vershwal: Yeah, I think you’re right that enforcing it at the file storage layer would be safer long-term shape. But I would keep it here for now to keep the scope of this PR smaller. At the moment we only have a single file upload path/caller, so this is still covering the current behavior safely.

allouis: Cool, in that case, I agree 🙏

allouis (approved):

ghost/core/core/shared/config/overrides.json
```diff
@@ -47,33 +47,54 @@
...
-                "application/vnd.ms-powerpoint",
-                "application/vnd.openxmlformats-officedocument.presentationml.presentation",
-                "application/rtf",
-                "text/plain",
-                "application/vnd.ms-excel",
-                "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
-                "application/xml",
-                "application/atom+xml"
-            ]
+            "contentTypes": []
```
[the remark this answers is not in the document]
vershwal: The key detail you missed is that the `/files/upload` route uses `fileValidation`, **not** the generic `validation` middleware. Looking at the route:

```322:327:ghost/core/core/server/web/api/endpoints/admin/routes.js
    router.post('/files/upload',
        mw.authAdminApi,
        apiMw.upload.single('file'),
        apiMw.upload.fileValidation({type: 'files'}),
        http(api.files.upload)
    );
```

The `fileValidation` middleware (lines 373-397) only checks **extensions** — it never reads `contentTypes` at all. The empty `contentTypes: []` in `overrides.json` exists solely to prevent `undefined` from leaking into the `ImportManager` (which aggregates content types from all upload handlers via `lodash.flatMap`). It has zero impact on the upload validation path.

The older `validation()` middleware is the one that checks both extensions and content types, but that's used by images/media — not files.

PR #26878, opened 2026-03-18
🐛 Fixed search navigating away during CJK IME composition

davehuang-tw:
When using CJK input methods (Chinese, Japanese, Korean) in the sodo-search modal (Cmd+K), pressing Enter to confirm a character during IME composition incorrectly triggers navigation to the first search result. This makes search unusable for CJK users.

The keyboard handler in the `Results` component used `keyup`, but `compositionend` fires before `keyup`, so `event.isComposing` is already `false` by then. Switched to `keydown` where `isComposing` reliably reflects the composition state.

### Before (main branch)

https://github.com/user-attachments/assets/248f9403-d896-4cdc-ac5b-6105379a9a42

- Type "測試" using zh-Hant IME while characters are still being composed
- Press Enter to confirm composition
- The search immediately redirects to the 1st result instead of finalizing the composed text and letting the user select their intented result

### After (PR branch)

https://github.com/user-attachments/assets/d1f9e041-1e86-4549-a95b-7ece822e207b

- Type "測試" using zh-Hant IME while characters are still being composed
- Press Enter to confirm composition
- Search results remain, letting user manually pick search item

---

> [!NOTE]
> **Low Risk**
> Low risk: small change to keyboard event handling in the search results list; main risk is unintended key handling differences between `keyup` and `keydown` for navigation/Enter behavior.
> 
> **Overview**
> Fixes Cmd+K search result navigation for CJK IME users by moving the results keyboard listener from `keyup` to `keydown` and ignoring events during composition (`event.isComposing` / legacy `keyCode === 229`).
> 
> This prevents Enter presses used to confirm IME text from triggering immediate navigation to the currently selected search result.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 866ececceb6a4ce4b43c8b6a98f4b96c2743bd3a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


9larsons (commented):
@sailor95 Thanks for this! Made a few additions.

PR #26884, opened 2026-03-19
🌐 Added missed string wrapping for Portal's offer page

cathysarisky:
no ref

* Abstracted the translateCadence function to utils since we need it in multiple places.
* Fixed two more places cadence wasn't wrapped for translations.
* Removed "{amountOff} off" string, since we are already using "{amount} off" elsewhere.


apps/portal/src/utils/helpers.js
```diff
@@ -996,3 +996,13 @@ export function isRecentMember({member}) {
 
     return diffHours < 24;
 }
+
+// Translate cadence to human readable string
```
9larsons, line 1000: What an ugly file this is. 1000 lines of 'helpers' 😁.

9larsons (approved):

apps/portal/src/utils/helpers.js
cathysarisky: Problem for another day!!

PR #26894, opened 2026-03-19
🌵 Spike: Adding `gifted` status to member subscription

mike182uk:
no ref

Spike into adding `gifted` membership subscription status

### DIFF
--- a/apps/admin-x-framework/src/api/members.ts
+++ b/apps/admin-x-framework/src/api/members.ts
@@ -65,7 +65,7 @@ export type Member = {
     name?: string;
     email?: string;
     avatar_image?: string;
-    status: 'free' | 'paid' | 'comped';
+    status: 'free' | 'paid' | 'comped' | 'gifted';
     note?: string;
     subscribed: boolean;
     labels?: MemberLabel[];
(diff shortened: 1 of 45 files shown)


cathysarisky (changes requested):
Hey @mike182uk ,

I hope this will move forward - gift subs are something I get asked about a LOT.  

Just a quick note that you probably need to consider ramifications on https://github.com/TryGhost/Ghost/blob/52dc4107c6de393955d8524fadc07cd387ed6af9/ghost/core/core/server/services/email-service/email-templates/template.hbs#L231 of subbing in 'gifted'.  At least American English would definitely say "gift subscription", not "gifted subscription".  

The existing free|paid|complimentary substitution is already a major headache for translators, because they don't consistently want the same word order for all three status values, and a fourth is going to exacerbate that.  Could we consider splitting this out to four separate full sentence status strings, which could then each be properly localized?

mike182uk:
Thanks for the comment @cathysarisky 

Yep this was very much a spike to see what the blast radius would be for adding something like this. As you can see, it's rather large 😅 The spike doesn't capture every single thing we'd need to modify to support this, so appreciate your input here, these are definitely the things that are easy to miss. I've made a note of this feedback so we can discuss further in the team when thinking more about implementation 👍

PR #26896, opened 2026-03-19
Added server-side search and infinite scroll for filter select dropdowns

kevinansfield:
refs https://linear.app/ghost/issue/BER-3334/

## Summary
- Added server-side search and infinite scroll for label and tier filter dropdowns in the members view
- Added `useFilterSearch` hook that wraps any infinite query with debounced search, pagination, and loading states
- When all items fit on a single page, automatically switches to instant local filtering instead of making server round-trips
- Added loading indicators during search (covering debounce wait and fetch) for both the standard select popovers and the label picker's custom renderer
- Fixed `defaultNextPageParams` in labels and tiers infinite queries to return `undefined` when there are no more pages, matching the existing pattern in tags/comments/members
- Used `initialCount` from the unfiltered query for `hasMultipleTiers` so the tier filter doesn't disappear when search narrows results
- Removed unused `useBrowseLabels` query and consolidated the infinite query + mutations under a single `dataType`. Label mutations now use `invalidateQueries` (triggering a refetch) instead of `updateQueries` (optimistic cache patching), matching the pattern used by tiers. The optimistic patching was inherited from the original simple query and would have needed reworking to handle the `InfiniteData` page structure anyway.
- Cached selected labels in `useLabelPicker` so they stay visible when server-side search narrows the result set

### DIFF
--- a/apps/admin-x-framework/src/api/labels.ts
+++ b/apps/admin-x-framework/src/api/labels.ts
@@ -24,15 +24,10 @@ export const useBrowseLabels = createQuery<LabelsResponseType>({
 export const useBrowseInfiniteLabels = createInfiniteQuery<LabelsResponseType & {isEnd: boolean}>({
     dataType: 'InfiniteLabelsResponseType',
     path: '/labels/',
-    defaultNextPageParams: (lastPage, otherParams) => {
-        if (!lastPage.meta?.pagination.next) {
-            return undefined;
-        }
-        return {
-            ...otherParams,
-            page: lastPage.meta.pagination.next.toString()
-        };
-    },
+    defaultNextPageParams: (lastPage, otherParams) => ({
+        ...otherParams,
+        page: (lastPage.meta?.pagination.next || 1).toString()
+    }),
     returnData: (originalData) => {
         const {pages} = originalData as InfiniteData<LabelsResponseType>;
         const labels = pages.flatMap(page => page.labels);
--- a/apps/admin-x-framework/src/api/labels.ts
+++ b/apps/admin-x-framework/src/api/labels.ts
@@ -1,4 +1,5 @@
-import {Meta, createMutation, createQuery} from '../utils/api/hooks';
+import {InfiniteData} from '@tanstack/react-query';
+import {Meta, createInfiniteQuery, createMutation, createQuery} from '../utils/api/hooks';
 
 export type Label = {
     id: string;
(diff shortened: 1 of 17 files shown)


apps/posts/src/hooks/use-label-picker.ts
```diff
@@ -18,15 +19,45 @@ export interface UseLabelPickerResult {
...
+
+    const searchParams = useMemo(() => {
+        const params: Record<string, string> = {limit: '100'};
+        if (serverSearchTerm.trim()) {
+            params.filter = `name:~'${escapeNqlValue(serverSearchTerm.trim())}'`;
+        }
+        return params;
+    }, [serverSearchTerm]);
+
+    const {data: labelsData, isLoading, isFetching, fetchNextPage, hasNextPage, isFetchingNextPage} = useBrowseInfiniteLabels({searchParams});
```
kevinansfield: [P1] `useLabelPicker` now reads from `useBrowseInfiniteLabels`, whose cache lives under `InfiniteLabelsResponseType`, but the label mutations still only update `LabelsResponseType`. After creating, editing, or deleting a label, `picker.labels` stays stale. That breaks the downstream flows that resolve `selectedSlugs` back through `picker.labels` before submit (for example add-label and import-members), so a newly created label can be selected in the UI but never actually sent to the backend.

apps/posts/src/components/label-picker/label-picker.tsx
```diff
@@ -359,7 +377,11 @@ const LabelPicker: React.FC<LabelPickerProps> = ({
...
     isDuplicateName,
     inline = false,
-    align = 'start'
+    align = 'start',
+    onSearchChange,
+    onLoadMore,
+    hasMore,
+    isLoadingMore
 }) => {
     const selectedLabels = selectedSlugs
```
kevinansfield, line 386: [P2] With server-side search enabled, `labels` only contains the current search result/page, but `selectedLabels` is still rebuilt by searching that filtered array. If I already selected label A and then search for label B, A drops out of the trigger text and selected-pill area even though `selectedSlugs` still contains it. The generic async select in `shade` preserves selected options across searches; this custom picker needs the same caching behavior.

apps/admin-x-framework/src/api/labels.ts
```diff
@@ -15,51 +16,44 @@ export interface LabelsResponseType {
...
     body: label => ({labels: [label]}),
-    updateQueries: {
-        dataType,
-        emberUpdateType: 'createOrUpdate',
-        update: (newData, currentData) => {
-            const current = currentData as LabelsResponseType;
-            return current && {...current, labels: current.labels.concat(newData.labels)};
-        }
-    }
+    invalidateQueries: {dataType}
```
kevinansfield, line 45: [P1] Switching the label mutations over to `invalidateQueries` fixes the cache-key mismatch, but it still leaves `picker.labels` stale until the refetch completes. `createLabel()` returns immediately after invalidation, and the picker toggles the new slug before the refreshed query data arrives. Downstream flows like add-label/import-members still resolve `selectedSlugs` back through `picker.labels`, so a user can create a label and click confirm before the refetch finishes, and that new label will still be dropped. This probably needs a synchronous cache/local-state update for the mutation result, not just invalidation.

apps/posts/src/views/members/components/members-filters.tsx
```diff
@@ -42,13 +44,30 @@ const MembersFilters: React.FC<MembersFiltersProps> = ({
...
+
+    const tierSearch = useFilterSearch({
+        useQuery: useBrowseTiers,
+        dataKey: 'tiers',
+        valueKey: 'id',
+        labelKey: 'name',
+        baseFilter: 'type:paid',
+        buildSearchFilter: (term: string) => `name:~'${escapeNqlValue(term)}'`,
+        limit: '100'
+    });
```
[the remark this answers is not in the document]
kevinansfield: This was changed to match the filter select behaviour in the previous Ember implementation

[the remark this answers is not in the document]
kevinansfield: Did you verify that or just accept what I said?

kevinansfield:
Closing in favour of https://github.com/TryGhost/Ghost/pull/26958

PR #26899, opened 2026-03-19
🐛 Fixed attribution on links in email-only posts

cmraible:
refs https://linear.app/ghost/issue/ONC-1569/post-attribution-for-links-in-emails-seems-to-be-broken-on-pro
closes https://linear.app/ghost/issue/ONC-1568/email-only-post-attribution-incorrectly-resolves-to-url-type-instead

## Problem

Attribution for links in email-only posts does not correctly resolve to the post itself. For example, if a free member clicks a link in an email-only post they received, then upgrade to a paid membership within the same browser session, the newly created subscription should be attributed to the email only post. Currently though, the subscription would be attributed as a `url` and, depending on the link, the URL might be the `/email/:uuid` route of the post, or the Homepage.

## Root cause

In `/ghost/core/core/server/services/member-attribution/url-translator.js`, Ghost had previously been filtering out email only posts, because the Post.findOne() query implicitly filters to only `published` posts, while an email-only post will have a status of `sent`.

## Fix

Updated the query in the url-translator to filter for `status:[published,sent]` and `type:[post,page]`. With this change, the attribution correctly resolves to the `post` itself, rather than the generic `url`.

There is also a change to the attribution builder to support this: since email only post's don't have a URL registered in the URL service, the `attribution_url` was being set to `null`. This adds a conditional to check if the post is email only, and if so, it sets the `attribution_url` to `/email/:uuid` for the post.

Finally, it also updated the link on the members page to the post to point to the Post Analytics, rather than the post on the frontend. This change is made for regular `published` posts in addition to the `sent` posts for consistency.

### Files changed
- **`url-translator.js`** — Added fallback `findOne` query with `status:sent` for email-only posts; added `/email/:uuid/` URL for sent posts
- **`attribution-builder.js`** — Use `/email/:uuid/` URL when resolving attribution for email-only posts at read time

## Test plan
- [ ] Send an email-only post containing an offer link
- [ ] Redeem the offer via the email link
- [ ] Verify the subscription attribution shows the post title, not "Homepage"
- [ ] Verify the attribution URL is `/email/:uuid/` (not `/404/`)
- [ ] Repeat with a published+sent post and verify attribution works correctly
- [ ] Verify page attribution still resolves correctly (not broken by the query change)
- [ ] Verify draft post attribution still degrades gracefully to URL type

### DIFF
--- a/ghost/admin/app/components/member/subscription-detail-box.hbs
+++ b/ghost/admin/app/components/member/subscription-detail-box.hbs
@@ -14,9 +14,15 @@
                 Source&nbsp;&mdash;&nbsp;<span>{{@sub.attribution.referrerSource}}</span>
             </p>
             {{/if}}
-            {{#if (and @sub.attribution @sub.attribution.url @sub.attribution.title)}}
+            {{#if (and @sub.attribution @sub.attribution.title)}}
             <p>
+                {{#if (and @sub.attribution.id (eq @sub.attribution.type "post"))}}
+                Page&nbsp;&mdash;&nbsp;<a href="/ghost/#/posts/analytics/{{@sub.attribution.id}}">{{ @sub.attribution.title }}</a>
+                {{else if @sub.attribution.url}}
                 Page&nbsp;&mdash;&nbsp;<a href="{{@sub.attribution.url}}" target="_blank" rel="noopener noreferrer">{{ @sub.attribution.title }}</a>
+                {{else}}
+                Page&nbsp;&mdash;&nbsp;{{ @sub.attribution.title }}
+                {{/if}}
             </p>
             {{/if}}
         </div>
--- a/ghost/core/core/server/services/member-attribution/attribution-builder.js
+++ b/ghost/core/core/server/services/member-attribution/attribution-builder.js
@@ -96,12 +96,7 @@ class Attribution {
             };
         }
 
-        // Email-only posts (status:sent) have no public URL — the URL service
-        // returns /404/ for them. Use the /email/:uuid path instead.
-        const isEmailOnly = this.type === 'post' && model.get('status') === 'sent';
-        const updatedUrl = isEmailOnly
-            ? this.#urlTranslator.relativeToAbsolute(`/email/${model.get('uuid')}/`)
-            : this.#urlTranslator.getUrlByResourceId(this.id, {absolute: true});
+        const updatedUrl = this.#urlTranslator.getResourceUrl(this.id, this.type, model, {absolute: true});
 
         return {
             id: model.id,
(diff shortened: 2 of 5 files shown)


ghost/core/core/server/services/member-attribution/url-translator.js
```diff
@@ -132,7 +132,7 @@ class UrlTranslator {
         switch (type) {
         case 'post':
         case 'page': {
-            const post = await this.models.Post.findOne({id}, {require: false});
+            const post = await this.models.Post.findOne({id}, {require: false, context: {internal: true}});
```
cmraible: This fixes the missing attribution for the link on email-only posts. Without the `context: {internal: true}`, Post.findOne() includes a `status: published` filter by default, which excludes posts sent as email-only. 

When resolving the attribution for a newly created subscription, it therefore does not find a post, and falls back to the Homepage.

ghost/admin/app/components/member/subscription-detail-box.hbs
```diff
@@ -16,7 +16,11 @@
             {{/if}}
             {{#if (and @sub.attribution @sub.attribution.url @sub.attribution.title)}}
             <p>
+                {{#if (eq @sub.attribution.post_status "sent")}}
```
cmraible: The other issue is that email only posts don't have a regular post URL, so attribution URL actually leads to a 404.

This checks if the post status is `sent` instead of `published`, and if so, it links to the post analytics > growth for the post, rather than 404ing.

ghost/core/core/server/services/member-attribution/attribution-builder.js
```diff
@@ -103,6 +103,7 @@ class Attribution {
             type: this.type,
             url: updatedUrl,
             title: model.get('title') ?? model.get('name') ?? this.#urlTranslator.getUrlTitle(this.url),
+            postStatus: this.type === 'post' ? model.get('status') : undefined,
```
cmraible: Plumbing to pass the postStatus through to the frontend, so we can link to Post Analytics instead of to a 404

ghost/core/core/server/api/endpoints/utils/serializers/output/members.js
```diff
@@ -379,6 +379,7 @@ function serializeAttribution(attribution) {
         type: attribution?.type,
         url: attribution?.url,
         title: attribution?.title,
+        post_status: attribution?.postStatus,
```
cmraible: More plumbing to pass the post_status through to the frontend to avoid linking to a 404

9larsons (approved):

PR #26914, opened 2026-03-23
Clean up sidebar collapsible composition

jonatansberg:
## Summary
This refactor pulls the sidebar collapsible behavior into `NavMenuItem` so Posts and Members can share the same composition pattern instead of each carrying their own toggle, chevron, and submenu wiring in `nav-content`.

It introduces `NavMenuItem.Collapsible`, `NavMenuItem.CollapsibleItem`, and `NavMenuItem.CollapsibleMenu`, extracts `useCustomSidebarViews` so the posts sidebar state can be reused outside the renderer, and removes the old standalone submenu wrapper. It also changes the collapsed-child fallback so an active hidden child highlights the parent item rather than forcing the section open.

## Stacked PR
This is a stacked PR on top of [#26893](https://github.com/TryGhost/Ghost/pull/26893). Please review it against `ber-3346-member-saved-views` rather than `main`.

## Test Plan
- `../../node_modules/.bin/vitest run src/layout/app-sidebar/nav-content.helpers.test.ts src/layout/app-sidebar/nav-custom-views.test.tsx src/layout/app-sidebar/nav-menu-item.test.tsx` (from `apps/admin`)
- `yarn workspace @tryghost/admin typecheck`


apps/admin/src/layout/app-sidebar/nav-menu-item.tsx
```diff
@@ -14,6 +16,93 @@ function NavMenuItem({ children, ...props }: React.ComponentProps<typeof Sidebar
...
+    id: string;
+    onExpandedChange: (expanded: boolean) => void | Promise<void>;
+}
+
+function NavMenuCollapsible({children, expanded, id, onExpandedChange}: NavMenuCollapsibleProps) {
+    const value = React.useMemo(() => ({
+        expanded,
+        id,
+        onExpandedChange
+    }), [expanded, id, onExpandedChange]);
```
rob-ghost: **Suggestion(non-blocking):** `onExpandedChange` is currently being passed the `setExpanded` of the `useNavigationExpanded` callback.

This is re-defined each render, which re-computes this `useMemo` every render.

We can either wrap the `onExpandedChange` in a `useCallback` upstream, or drop the `useMemo` since its currently not doing anything. Or we could leave it as it is because its better for `NavMenuCollapsible` to try to memoise anyway, even if we're not passing in a wrapped callback.

apps/admin/src/layout/app-sidebar/nav-menu-item.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+        Button: ({children, ...props}: React.ComponentProps<'button'>) => <button {...props}>{children}</button>,
+        LucideIcon: {
+            ChevronRight: (props: React.ComponentProps<'svg'>) => <svg {...props} />
+        },
+        useSidebar: () => ({
+            isMobile: false,
+            setOpenMobile: vi.fn()
+        })
+    };
+});
```
rob-ghost: **Question:** This is a pretty wide mock surface area, could we not use the actual components out of Shade?

Its possible for the Shade components to change in a breaking way but this test won't fail because of the mocks.

rob-ghost (approved):
Nothing blocking 👍🏻

PR #26915, opened 2026-03-23
Require erasable TypeScript syntax

EvanHahn:
no ref

**What?** This enables [TypeScript's `erasableSyntaxOnly` option][0] across the codebase. This means we can't use enums, namespaces, or a few other things. This change should have no user impact.

**Why?** The big reason: [Node supports running TypeScript files directly][1], but only if all the syntax is erasable. This means we can remove build steps in many cases! It has a few other advantages, too: simplifying source maps, restricting some tricky parts of TypeScript, slightly accelerating TypeScript compilation,

This doesn't enable some of the other options that are sometimes recommended in tandem, like `verbatimModuleSyntax`. Nor does it actually run `.ts` files with Node. Those are for another day.

[0]: https://www.typescriptlang.org/tsconfig/#erasableSyntaxOnly
[1]: https://nodejs.org/api/typescript.html#type-stripping


9larsons (approved):
Great idea!

PR #26923, opened 2026-03-24
Added theme helpers for subscription pricing and offer duration

minimaluminium:
ref https://linear.app/ghost/issue/BER-3384/add-interval-option-to-price-helper
ref https://linear.app/ghost/issue/BER-3461/new-offer-duration-helper

Adds two small helpers to make it easier for theme developers to build custom account pages that display subscription pricing and active discounts.

  **1. `interval` option on `{{price}}`**

  `{{price plan interval="true"}}` now outputs `$10/month` instead of requiring manual concatenation (`{{price plan}}/{{plan.interval}}`). Works with any object that has `.amount`, `.currency`, and `.interval` —
  including `plan` and `next_payment`.

  **2. New `{{offer_duration}}` helper**

  Outputs the duration context of an active discount: "Forever", "Ends 2 May 2026", or empty string if no discount. Used inside a subscription loop alongside `{{price}}` to match Portal's offer label format.

  Together, displaying a discounted subscription in a theme becomes:

  ```handlebars
  {{#foreach @member.subscriptions}}
    {{#if next_payment.discount}}
      <s>{{price plan interval="true"}}</s>
      <p>{{price next_payment interval="true"}} — {{offer_duration}}</p>
    {{else}}
      <p>{{price plan interval="true"}}</p>
    {{/if}}
  {{/foreach}}

### DIFF
--- a/ghost/core/core/frontend/helpers/offer_duration.js
+++ b/ghost/core/core/frontend/helpers/offer_duration.js
@@ -15,7 +15,7 @@ function formatDate(isoDate) {
         return '';
     }
     const event = new Date(isoDate);
-    const options = {year: 'numeric', month: 'short', day: 'numeric'};
+    const options = {year: 'numeric', month: 'short', day: 'numeric', timeZone: 'UTC'};
     return event.toLocaleDateString('en-GB', options);
 }
 
--- a/ghost/core/core/frontend/helpers/offer_duration.js
+++ b/ghost/core/core/frontend/helpers/offer_duration.js
@@ -1,0 +1,61 @@
+// # {{offer_duration}} Helper
+// Usage: `{{offer_duration}}`
+//
+// Should be used inside of a subscription context, e.g.:
+// `{{#foreach @member.subscriptions}} {{price next_payment interval="true"}} — {{offer_duration}} {{/foreach}}`
+//
+// Outputs the duration context of an active discount on a subscription.
+// Returns "Forever", "Ends {date}", or an empty string if no active discount.
+// Consistent with the duration part of Portal's offer label on the member account page.
+const {SafeString} = require('../services/handlebars');
+const {labs} = require('../services/proxy');
+
+function formatDate(isoDate) {
+    if (!isoDate) {
+        return '';
+    }
+    const event = new Date(isoDate);
+    const options = {year: 'numeric', month: 'short', day: 'numeric'};
+    return event.toLocaleDateString('en-GB', options);
+}
+
+function offer_duration() { // eslint-disable-line camelcase
+    const nextPayment = this.next_payment;
+
+    if (!nextPayment) {
+        return '';
+    }
+
+    const discount = nextPayment.discount;
+
+    if (!discount) {
+        return '';
+    }
+
+    let durationLabel = '';
+    if (discount.duration === 'forever') {
+        durationLabel = 'Forever';
+    } else if (discount.duration === 'repeating' && discount.end) {
+        durationLabel = `Ends ${formatDate(discount.end)}`;
+    } else if (discount.duration === 'once' && this.current_period_end) {
+        // By design, "once" offers don't have a discount end in Stripe.
+        // They expire at the end of the current billing period.
+        durationLabel = `Ends ${formatDate(this.current_period_end)}`;
+    }
+
+    return new SafeString(durationLabel);
+}
+
+module.exports = function offerDurationLabsWrapper() {
+    let self = this;
+    let args = arguments;
+
+    return labs.enabledHelper({
+        flagKey: 'members',
+        flagName: 'Members',
+        helperName: 'offer_duration',
+        helpUrl: 'https://ghost.org/docs/themes/members/'
+    }, () => {
+        return offer_duration.apply(self, args); // eslint-disable-line camelcase
+    });
+};
--- a/ghost/core/core/frontend/helpers/price.js
+++ b/ghost/core/core/frontend/helpers/price.js
@@ -72,7 +72,7 @@ module.exports = function price(planOrAmount, options) {
             locale
         });
 
-        if (interval === 'true' && plan.interval) {
+        if (interval && interval !== 'false' && plan.interval) {
             result = `${result}/${plan.interval}`;
         }
 
--- a/ghost/core/core/frontend/helpers/price.js
+++ b/ghost/core/core/frontend/helpers/price.js
@@ -5,11 +5,14 @@
 // Usage: `{{price plan numberFormat="long"}}`
 // Usage: `{{price plan currencyFormat="code"}}`
 // Usage: `{{price plan currencyFormat="name"}}`
+// Usage: `{{price plan interval="true"}}`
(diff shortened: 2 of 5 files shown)


ghost/core/core/frontend/helpers/price.js
```diff
@@ -59,15 +62,21 @@ module.exports = function price(planOrAmount, options) {
...
-        return formatter({
+        let result = formatter({
             amount: plan.amount && (plan.amount / 100),
             currency: currency || plan.currency,
             currencyFormat,
             numberFormat,
             locale
         });
+
+        if (interval && interval !== 'false' && plan.interval) {
```
sagzy, line 75: suggestion: shall we use a positive condition and stricter condition here, so it's easier to reason about?

```
const intervalRequested = interval === true || interval === "true";

if (intervalRequested && plan.interval) {
   ...
}
```

ghost/core/core/frontend/helpers/offer_duration.js
```diff
@@ -0,0 +1,61 @@
...
+const {SafeString} = require('../services/handlebars');
+const {labs} = require('../services/proxy');
+
+function formatDate(isoDate) {
+    if (!isoDate) {
+        return '';
+    }
+    const event = new Date(isoDate);
+    const options = {year: 'numeric', month: 'short', day: 'numeric', timeZone: 'UTC'};
+    return event.toLocaleDateString('en-GB', options);
```
sagzy, line 19: On hardcoding `en-GB` for the date formatting: should we instead use `options.data.site.locale` like the price helper?

ghost/core/core/frontend/helpers/price.js
```diff
@@ -5,11 +5,14 @@
 // Usage: `{{price plan numberFormat="long"}}`
 // Usage: `{{price plan currencyFormat="code"}}`
 // Usage: `{{price plan currencyFormat="name"}}`
+// Usage: `{{price plan interval="true"}}`
```
sagzy, line 8: Shall we also document `interval = true` as option here, as both "true" and true are accepted?

ghost/core/core/frontend/helpers/offer_duration.js
sagzy: Another reference: the existing {{date}} helper

ghost/core/core/frontend/helpers/offer_duration.js
```diff
@@ -0,0 +1,61 @@
...
+    let durationLabel = '';
+    if (discount.duration === 'forever') {
+        durationLabel = 'Forever';
+    } else if (discount.duration === 'repeating' && discount.end) {
+        durationLabel = `Ends ${formatDate(discount.end)}`;
+    } else if (discount.duration === 'once' && this.current_period_end) {
+        // By design, "once" offers don't have a discount end in Stripe.
+        // They expire at the end of the current billing period.
+        durationLabel = `Ends ${formatDate(this.current_period_end)}`;
+    }
```
sagzy, line 44: question: How would translations work for this helper? What if I have a site in Korean?

The "Forever" string can be translated, with using `{{t (offer_duration)}}` in the theme then adding the string in "locales/{locale}.json"

But wouldn't "Ends 3 May 2026" be impossible to translate, as the date is embedded in the string?

— 

This makes me wonder if, instead of building a helper that outputs an English string, we should have a helper that outputs the end date of a discount. This will still abstract away any knowledge of how the different offers work (once expires at current billing end, repeating expires at discount end, forever has no end date), but gives more control over the actual rendering of the string.

Example:

`{discount_end_date}}` returns a date for repeating or once offers, or null for forever offers.

Then theme developers can use it as 

```
{{#if next_payment.discount}}
      <s>{{price plan}}/{{plan.interval}}</s>
      <p>
          {{price next_payment}}/{{next_payment.interval}}
          {{#if (discount_end_date)}}
              — Ends {{date discount_end_date format="D MMM YYYY"}}
          {{else}}
              — Forever
          {{/if}}
      </p>
  {{else}}
      <p>{{price plan}}/{{plan.interval}}</p>
  {{/if}}
```

It's a bit more code than with the current `{{offer_duration}}` helper, but solves the translation issue (and gives more control over formatting the date too)

sagzy: I'm actually now thinking that maybe we should alter the API response, and abstract away the difference between once/repeating/forever in a way that allows theme developers to directly write:

```
{{#if next_payment.discount}}
      <s>{{price plan}}/{{plan.interval}}</s>
      <p>
          {{price next_payment}}/{{next_payment.interval}}
          {{#if (next_payment.discount.end)}}
              — Ends {{date next_payment.discount.end format="D MMM YYYY"}}
          {{else}}
              — Forever
          {{/if}}
      </p>
  {{else}}
      <p>{{price plan}}/{{plan.interval}}</p>
  {{/if}}
  ```
  
This eliminates the need for adding an additional helper, and we end up with a theme-level code that is much more readable than the current `match`-based one:

```
{{#if next_payment.discount}}
    <!-- {{next_payment.discount.duration}} -->
    <s>{{price plan}}/{{plan.interval}}</s>
    <p>
    {{price next_payment}}/{{next_payment.interval}}
    {{#match next_payment.discount.duration "=" "forever"}}
        — Forever
    {{/match}}
    {{#match next_payment.discount.duration "=" "repeating"}}
        — Ends {{date next_payment.discount.end format="D MMM YYYY"}}
    {{/match}}
    {{#match next_payment.discount.duration "=" "once"}}
        — Ends {{date current_period_end format="D MMM YYYY"}}
    {{/match}}
    </p>
{{else}}
    <p>{{price plan}}/{{plan.interval}}</p>
{{/if}}
```

Comment (sagzy) [reply ghost/core/core/frontend/helpers/offer_duration.js:44]: Note: that API change can be made in a forward-compatible way, as we won't be modifying the current field that is used for once offers (`current_period_end`), so that shouldn't break any existing code in Portal nor implementations in themes

Comment (sagzy) [reply ghost/core/core/frontend/helpers/offer_duration.js:44]: Only tradeoff is that the "discount end" we expose under the next payment object will not always match what is in Stripe (once offers have no discount end in Stripe), but I think that's acceptable. The database will still store the original value from Stripe, we would override it within the scope of the next payment object, whose responsibility is to answer "what is the next payment amount and does a discount apply to it?"

File: ghost/core/core/frontend/helpers/price.js [hunk 5]
```diff
@@ -59,15 +62,21 @@ module.exports = function price(planOrAmount, options) {
...
+        let result = formatter({
             amount: plan.amount && (plan.amount / 100),
             currency: currency || plan.currency,
             currencyFormat,
             numberFormat,
             locale
         });
+
+        if (interval && interval !== 'false' && plan.interval) {
+            result = `${result}/${plan.interval}`;
```
sagzy, line 76: @minimaluminium

question: wouldn't we run into the same translation issue here? Result string would be e.g. `$5.00/month`, which would be impossible to translate as the price is variable.

PR #26929, opened 2026-03-24
Fixed S3Storage.exists() crash when called without targetDir

allouis:
## Summary

- Made `targetDir` optional in `S3Storage.exists()` and `S3Storage.delete()`, matching `LocalStorageBase`'s behavior
- When `targetDir` is omitted, uses `fileName` directly as the relative path instead of crashing in `path.posix.join(undefined, fileName)`
- Fixes breakage in `handle-image-sizes` middleware and `image-size` service which both call `exists(path)` with a single argument

## Context

Several callers pass a single path to `exists()` rather than separate `fileName` + `targetDir` args. This works fine with `LocalStorageBase` (which defaults to `this.storagePath` when `targetDir` is undefined) but crashes with `S3Storage`. Both `saveRaw` and `exists` go through `buildKey` with the same input, so the S3 keys match correctly.

### DIFF
--- a/ghost/core/core/server/adapters/storage/S3Storage.ts
+++ b/ghost/core/core/server/adapters/storage/S3Storage.ts
@@ -310,14 +310,14 @@ export default class S3Storage extends StorageBase {
         return relativePath.slice(this.storagePath.length + 1);
     }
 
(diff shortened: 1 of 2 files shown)


vershwal (approved):
Nice one, looks good!

PR #16402, opened 2023-03-13
Added `show_latest_posts` column to newsletter table

SimonBackx:
fixes https://github.com/TryGhost/Team/issues/2709

Adds the `show_latest_posts` column to the newsletter table. Disabled by default.

PR #26939, opened 2026-03-24
Added show_share_button column to newsletters

troyciesco:
ref https://linear.app/ghost/issue/FEA-480

- adds the `show_share_button` column to newsletters to prep for future share button functionality
- includes the bare minimum for setting up the migration and for tests to pass. In this case it includes updating snapshots related to the newsletter
- also bumps to `6.23.0-rc.0`


apps/admin-x-framework/src/api/newsletters.ts
```diff
@@ -32,6 +32,7 @@ export type Newsletter = {
     show_comment_cta: boolean;
     show_subscription_details: boolean;
     show_latest_posts: boolean;
+    show_share_button: boolean;
```
troyciesco: i dont think this type update and the newsletters.json normally get included in a migration PR...but i think they should?

When i first set up this PR i basically just had the bare min migration files, but migration-related changes in tests failed [here](https://github.com/TryGhost/Ghost/actions/runs/23503341707/job/68405618164).

So i added it to the model and it seemed to pass from there, and i had to then update the snapshots accordingly. with that in mind it makes sense to me that the type and the test response data also should be updated.

Unless i hit a weird sqlite edge-case on the first run in that test? i'm not sure, i'd never hit that error before when doing a migration.

ghost/core/core/server/data/migrations/versions/6.23/2026-03-24-15-55-52-add-show-share-button-to-newsletters.js
```diff
@@ -0,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('newsletters', 'show_share_button', {
+    type: 'boolean',
+    nullable: false,
```
EvanHahn, line 5: praise: love when a field is not nullable.

apps/admin-x-framework/src/api/newsletters.ts
cmraible: Genuinely just a question, I'm not sure: given that the admin pieces of this will deploy right away ahead of the migration, will this cause any issues if this field doesn't exist in the API responses?

cmraible: > Unless i hit a weird sqlite edge-case on the first run in that test? i'm not sure, i'd never hit that error before when doing a migration.

I've also never encountered that error, but it does seem a bit fishy. You've defined a default value, so it shouldn't encounter any NOT NULL constraints 🤔

EvanHahn (approved):
LGTM other than @cmraible's question, which is a good one.

cmraible (approved):
LGTM!

apps/admin-x-framework/src/api/newsletters.ts
troyciesco: we've talked outside this thread about this, but to summarize here for the curious (cc @EvanHahn):
- the model change is needed and included in other migrations of this type (like https://github.com/TryGhost/Ghost/pull/16402)

> given that the admin pieces of this will deploy right away ahead of the migration, will this cause any issues if this field doesn't exist in the API responses?
we don't think so. once this is merged, these server changes won't deploy to Pro right away, they'll need a rollout. but because there are no changes to the admin code to try and reference/use `show_share_button` we should be ok, even if it's still hitting server version `6.22.1`, effectively. once the server is next rolled out it will be `6.23-rc.0`, and then we can safely add code that references it.

PR #26944, opened 2026-03-24
Refactored share modal files

troyciesco:
Refactors a bit of the share modal setup:
- moves into 3 separate files for component, styles, hook
- gathers all the data in one function and simplifies the logic a bit


apps/portal/src/components/pages/share/use-share-data.js
```diff
@@ -0,0 +1,89 @@
...
+
+const gatherShareData = () => {
+    const hostWindow = getHostWindow();
+    const doc = hostWindow.document;
+    const loc = hostWindow.location;
+
+    // helper to reduce some repetition of querySelector
+    const q = (selector, attr = 'content') => doc.querySelector(selector)?.[attr] || '';
+
+    const shareUrl = q('link[rel="canonical"]', 'href') || (loc.origin + loc.pathname + loc.search);
```
troyciesco, line 52: to me, this (and the others below) is easier to read and reason about than custom helpers like `getCanonicalUrl` for all of them, but open to feedback on that

apps/portal/src/components/pages/share/use-share-data.js
```diff
@@ -0,0 +1,89 @@
...
+        threads: createShareLink('https://www.threads.net/intent/post', {text: threadsText}),
+        linkedin: createLinkedinShareLink({shareTitle, shareUrl}),
+        bluesky: createShareLink('https://bsky.app/intent/compose', {text: threadsText})
+    };
+
+    return {shareUrl, shareTitle, shareExcerpt, shareImage, shareFavicon, shareSiteName, shareAuthor, socialLinks};
+};
+
+const useShareData = () => {
+    const [data] = useState(gatherShareData);
```
troyciesco, line 84: i think memoizing each function wasn't necessary - though since all this data fetching is just dom selectors state might not be necessary either, will revisit

troyciesco: yeah i don't think it's strictly necessary but it's not causing any problems either

PR #26946, opened 2026-03-24
Removed unused values from email renderer design options

EvanHahn:
towards https://linear.app/ghost/issue/NY-1163

We pass a number of options when we render the email's base HTML: accent color, background color, link style...but most of them are completely unused! Let's remove them.

I think this is a useful change on its own, but it also makes [an upcoming one][0] easier.

In addition to the existing automated tests, I also did a manual newsletter test and everything looks the same.

[0]: https://linear.app/ghost/issue/NY-1163

### DIFF
--- a/ghost/core/core/server/services/email-service/email-renderer.js
+++ b/ghost/core/core/server/services/email-service/email-renderer.js
@@ -331,30 +331,9 @@ class EmailRenderer {
             postUrl
         };
 
-        const accentColor = this.#getAccentColor();
-        const accentContrastColor = this.#getAccentContrastColor();
-
         renderOptions.design = {
-            accentColor,
-            accentContrastColor,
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/services/email-service/email-renderer.js
```diff
@@ -331,30 +331,9 @@ class EmailRenderer {
             postUrl
         };
 
-        const accentColor = this.#getAccentColor();
-        const accentContrastColor = this.#getAccentContrastColor();
-
         renderOptions.design = {
-            accentColor,
```
EvanHahn, line 338: For reference, `buttonCorners` and `buttonStyle` are referenced in the Koenig repo. The rest are not.

Please let me know if there's something I'm missing!

EvanHahn:
Never mind, I think this is wrong.

PR #26953, opened 2026-03-25
Improved member list query performance

jonatansberg:
ref https://linear.app/ghost/issue/BER-3471/fix-default-members-list-query-performance-with-a-composite-index

This PR adds a composite index for the default members list query. The default members browse path orders by `created_at DESC, id DESC`. Without a matching composite index, large member tables fall back to a full scan plus filesort. This change keeps the schema aligned with that query shape by adding `members(created_at, id)` to the members schema definition and shipping a non-transactional migration that creates the index with online DDL on MySQL.

### Performance impact:

With 2m members cold queries went from ~7s to ~0.1s and hot queries from ~0.8s to ~0.03s.

### Verification:

- Ran `yarn knex-migrator migrate`, `yarn knex-migrator rollback --force`, and `yarn knex-migrator migrate` against MySQL
- Confirmed the `members_created_at_id_index` index exists after re-migrating and that the default members query uses it again
- Verified the migration's `up`/`down` and idempotent paths against SQLite with an isolated direct migration check
- Ran `mocha test/unit/server/data/schema/schema.test.js`

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/6.37/2026-05-04-07-51-14-add-members-created-at-id-index.js
+++ b/ghost/core/core/server/data/migrations/versions/6.37/2026-05-04-07-51-14-add-members-created-at-id-index.js
@@ -1,0 +1,63 @@
+const logging = require('@tryghost/logging');
+const DatabaseInfo = require('@tryghost/database-info');
+const {createNonTransactionalMigration} = require('../../utils');
+
+const INDEX_NAME = 'members_created_at_id_index';
+
+async function hasIndex(knex) {
+    if (DatabaseInfo.isSQLite(knex)) {
+        const result = await knex.raw(`select * from sqlite_master where type = 'index' and tbl_name = 'members' and name = '${INDEX_NAME}'`);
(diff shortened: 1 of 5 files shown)


ghost/core/core/server/data/migrations/versions/6.37/2026-05-04-07-51-14-add-members-created-at-id-index.js
```diff
@@ -0,0 +1,63 @@
...
+                ADD INDEX ${INDEX_NAME} (created_at, id),
+                ALGORITHM=INPLACE,
+                LOCK=NONE
+            `);
+            return;
+        }
+
+        await knex.schema.table('members', (table) => {
+            table.index(['created_at', 'id']);
+        });
```
kevinansfield, line 38: Not a blocker but I wonder if we should improve the existing `addIndex` and `dropIndex` commands to support index existence checks and `algorithm` and `lock` options? We have precedent for that in the `addColumn` and `dropColumn` commands

PR #26956, opened 2026-03-25
Tightened validation for colors in email renderer

EvanHahn:
no ref

Before this change, `foo #ff9900` was considered a valid hex color. After this change, it is not.

I believe it's difficult to have bad data in the database, but if that happens, we want to check it properly.


cmraible (approved):
Nice find!

PR #26958, opened 2026-03-25
Consolidated filter search into unified useFilterSearch hook

kevinansfield:

## Summary

- Shared `useFilterSearch` hook handles local/server search mode detection, debouncing, infinite scroll, and async resolution of selected values not on the current page
- Comments and members filter views use `useFilterSearch` instead of separate scattered implementations
- `createQueryWithId`-based getters (`getLabelBySlug`, `getNewsletter`, etc.) in admin-x-framework resolve display names for filter values outside the paginated window
- `useLabelPicker` caches selected labels independently (mirroring shade's `cachedSelectedOptions` pattern) so label renames and search/remount cycles don't lose resolved names
- Fixed `useDeleteLabel`'s cache updater crashing on infinite query data
- Disabled cmdk's `shouldFilter` on filter dropdowns that use `onSearchChange` — cmdk v1 reorders DOM elements via `appendChild` during scored search and never restores the original order when the search is cleared


apps/admin-x-framework/src/api/offers.ts
kevinansfield, line 66: Aside: The `useBrowseOffersById` naming here is an inconsistency, all other APIs use `get{Resource}` with `createQueryWithId`

apps/posts/src/views/members/hooks/use-email-resource-search.ts
```diff
@@ -0,0 +1,22 @@
...
+import {useFilterSearch} from '@src/hooks/use-filter-search';
+
+export function useEmailResourceSearch() {
+    return useFilterSearch({
+        useQuery: useBrowsePostsInfinite,
+        dataKey: 'posts',
+        serverSearchParams: term => ({
+            filter: term
+                ? `(status:published,status:sent)+newsletter_id:-null+title:~${escapeNqlString(term)}`
+                : '(status:published,status:sent)+newsletter_id:-null',
```
kevinansfield: The base filter `(status:published,status:sent)+newsletter_id:-null` could be extracted here so we don't repeat it

apps/posts/src/views/members/components/members-filters.tsx
```diff
@@ -49,30 +62,88 @@ const MembersFilters: React.FC<MembersFiltersProps> = ({
...
-    const {data: tiersData} = useBrowseTiers({searchParams: {limit: '100'}});
-    const {data: offersData} = useBrowseOffers({});
-    const {data: newslettersData} = useBrowseNewsletters({searchParams: {limit: '100'}});
+    const activeLabelValues = getActiveFilterValues(filters, 'label');
+    const activeTierValues = getActiveFilterValues(filters, 'tier_id');
+
+    const labelSearch = useFilterSearch({
+        useQuery: useBrowseLabelsInfinite,
+        dataKey: 'labels',
+        serverSearchParams: (term): Record<string, string> => (term ? {filter: `name:~${escapeNqlString(term)}`} : {}),
```
kevinansfield, line 70: Why does this have an explicit return type but others `serverSearchParams` don't?

apps/posts/src/views/members/components/members-filters.tsx
```diff
@@ -49,30 +62,88 @@ const MembersFilters: React.FC<MembersFiltersProps> = ({
...
+                seen.add(offer.id);
+                merged.push(offer);
+            }
+        }
+        return buildOfferOptions(merged);
+    }, [offerItems, allOfferItems]);
+
+    // Use unfiltered items for field visibility (must not change during search)
+    const allPaidTiers = tierSearch.allItems.filter(tier => tier.type === 'paid' && tier.active);
+    const hasMultipleTiers = allPaidTiers.length > 1;
```
kevinansfield: We've only fetched paid tiers from the API, why do we have the extra client-side filter here?

apps/posts/src/views/members/components/members-filters.tsx
```diff
@@ -91,35 +162,37 @@ const MembersFilters: React.FC<MembersFiltersProps> = ({
         onFiltersChange(mapOfferRedemptionFilters(newFilters, values => fromOfferFilterDisplayValues(values, offersOptions)));
     }, [onFiltersChange, offersOptions]);
 
-    const postSearch = useResourceSearch('post');
-    const emailSearch = useResourceSearch('email');
+    const postSearch = usePostResourceSearch();
+    const emailSearch = useEmailResourceSearch();
```
kevinansfield: Why is `useEmailResourceSearch` extracted when it's just wrapping a plain `useFilterSearch` call as we have inline above?

apps/posts/src/views/members/hooks/use-email-resource-search.ts
kevinansfield: Done — extracted to `EMAIL_BASE_FILTER` constant.

apps/posts/src/views/members/components/members-filters.tsx
kevinansfield: It's needed here because TS can't narrow the empty object `{}` to `Record<string, string>` without it — the other `serverSearchParams` callbacks always return an object with a `filter` key so TS infers the type correctly. Left as-is since the alternative would be an inline cast.

apps/posts/src/views/members/components/members-filters.tsx
```diff
@@ -24,7 +32,13 @@ interface MembersFiltersProps {
     activeView?: MemberView | null;
 }
 
-const EMPTY_OFFERS: typeof buildOfferOptions extends (offers: infer T) => unknown ? T : never = [];
+function toSearchProps(search: {onSearchChange: (s: string) => void; searchValue: string; isLoading: boolean}): FilterSearchProps {
+    return {
+        onSearchChange: search.onSearchChange,
+        searchValue: search.searchValue,
+        isLoading: search.isLoading
+    };
```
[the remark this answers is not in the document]
kevinansfield: This PR isn't handling infinite scroll within dropdowns so we can keep the scope focused on consolidating the data handling. Infinite scroll will come later

kevinansfield:
There's an alternative approach in https://github.com/TryGhost/Ghost/pull/26986 that tries to avoid the monolithic `useFilterSearch` hook

kevinansfield:
Closing in favour of https://github.com/TryGhost/Ghost/pull/26986

PR #26959, opened 2026-03-25
Removed unnecessary measureElement from virtual scroll

rob-ghost:
## Summary
- Removes `virtualizer.measureElement` ref from virtual scroll item props since all consumers use fixed-height rows (`estimateSize: () => 72` or default `() => 100`)
- Dynamic measurement forces synchronous layout/reflow on every row mount, which is unnecessary for fixed-height rows
- The `virtualizer` object is already returned from the hook, so any future consumer needing dynamic measurement can access `virtualizer.measureElement` directly

Built on top of #26952 which fixes the scroll container resolution.

## Test plan
- [x] `yarn workspace @tryghost/posts test:unit` passes (114 tests)
- [x] Verified members table renders correctly with 1,000 members (19 visible rows, 738 total DOM elements)
- [x] Scrolling and infinite loading work correctly

### DIFF
--- a/apps/posts/src/components/virtual-table/use-infinite-virtual-scroll.tsx
+++ b/apps/posts/src/components/virtual-table/use-infinite-virtual-scroll.tsx
@@ -45,7 +45,6 @@ export function useInfiniteVirtualScroll<T>({
         key: virtualItem.key,
         item: items[virtualItem.index],
         props: {
-            ref: virtualizer.measureElement,
             'data-index': virtualItem.index
         }
     }));


rob-ghost:
Closing this PR — the `measureElement` removal is unnecessary. The performance issue (3.4s blocking task) was caused by measuring 900+ items when the scroll container was misconfigured. PR #26952's overflow fix ensures only ~15 items render at a time, making `measureElement` negligible. Removing it actually breaks scroll position tracking for lists where `estimateSize` doesn't match actual row heights (e.g. tags: 77px actual vs 100px default estimate).

PR #26969, opened 2026-03-25
Added snapshot tests for post analytics CSV export endpoint

cmraible:
towards https://linear.app/ghost/issue/NY-645/add-post-analytics-export-loaderdata-stream-in-settings

Added snapshots tests to the post analytics CSV export endpoint with rich fixture data covering all export code paths: email analytics (sends/opens/clicks), multiple newsletters, member feedback, signup/conversion attribution, and all access types. Includes 6 settings variation tests that snapshot different column sets when features are toggled off.

This change is in preparation for a fairly significant refactor of the post analytics export to use data streaming, to ensure the change doesn't introduce meaningful behavioral changes.

### DIFF
--- a/ghost/core/core/server/services/posts/posts-exporter.js
+++ b/ghost/core/core/server/services/posts/posts-exporter.js
@@ -35,7 +35,7 @@ class PostsExporter {
     async export({filter, order, limit}) {
         const posts = await this.#models.Post.findPage({
             filter: filter ?? 'status:published,status:sent',
-            order,
+            order: order ?? 'published_at DESC, id DESC',
             limit,
             withRelated: [
                 'tiers',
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/services/posts/posts-exporter.js
```diff
@@ -35,7 +35,7 @@ class PostsExporter {
     async export({filter, order, limit}) {
         const posts = await this.#models.Post.findPage({
             filter: filter ?? 'status:published,status:sent',
-            order,
+            order: order ?? 'published_at DESC, id DESC',
```
cmraible: This is the only potential behavioral change to this endpoint included in this change, which makes the ordering deterministic for snapshot stability when no order is passed via the API (as is the case in Admin).

PR #26975, opened 2026-03-26
Added migration to create default `email_design_settings` row

cmraible:
closes https://linear.app/ghost/issue/NY-1179/insert-default-email-design-settings-row

## Summary

Seeds a default row into the `email_design_settings` table. This row will be referenced by all existing `automated_emails` rows when the FK is added in a subsequent migration.

## Changes

- Added DML migration to insert a default `email_design_settings` row with slug `default-automated-email`, using schema defaults for all design columns. Idempotent — skips if the row already exists.
- Added `EmailDesignSetting` Bookshelf model (required by the fixture system)
- Added fixture entry in `fixtures.json` so new databases created via `knex-migrator init` also get the default row
- Updated fixtures integrity hash


ghost/core/core/server/models/email-design-setting.js
```diff
@@ -0,0 +1,30 @@
+const ghostBookshelf = require('./base');
```
cmraible, line 1: note: creating the model is required for the fixtures to work

`yarn knex-migrator init` doesn't run all the migrations in the `versions` directory, so we have to add any default data like this in the migration (for existing sites) _and_ the fixtures (for new sites). The fixture manager depends on the model, hence including it in this PR.

cmraible:
@CodeRabbit review

ghost/core/core/server/models/email-design-setting.js
```diff
@@ -0,0 +1,30 @@
...
+            title_font_weight: 'bold',
+            image_corners: 'square',
+            show_badge: true
+        };
+    }
+});
+
+module.exports = {
+    EmailDesignSetting: ghostBookshelf.model('EmailDesignSetting', EmailDesignSetting)
+};
```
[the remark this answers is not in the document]
cmraible: ghost/core/core/server/models/index.js auto-discovers models, they don't need to be explicitly registered

ghost/core/core/server/data/migrations/versions/6.25/2026-03-26-15-47-00-insert-default-email-design-settings-row.js
```diff
@@ -0,0 +1,42 @@
...
+        logging.info('Inserting default email_design_settings row');
+
+        const existing = await knex('email_design_settings').where({slug: DEFAULT_SLUG}).first();
+
+        if (existing) {
+            logging.warn('Default email_design_settings row already exists, skipping');
+            return;
+        }
+
+        await knex('email_design_settings').insert({
```
troyciesco, line 18: i'd meant to bring this up yesterday: would it make sense to seed these styles from the first active newsletter (avoiding saying "default" lol) if it exists?

one spot it could get clunky is if there's something like a header image in the newsletter, it might not make sense in the welcome email.

i think seeding like how you have it is probably right but just wanted to bring it up before we ship this piece. A "copy design from newsletter" button would probably be a better way to smooth out the experience for people that do want stuff the same or want to save most of the effort of copying colors etc.

ghost/core/core/server/data/migrations/versions/6.25/2026-03-26-15-47-00-insert-default-email-design-settings-row.js
```diff
@@ -0,0 +1,42 @@
...
+module.exports = createTransactionalMigration(
+    async function up(knex) {
+        logging.info('Inserting default email_design_settings row');
+
+        const existing = await knex('email_design_settings').where({slug: DEFAULT_SLUG}).first();
+
+        if (existing) {
+            logging.warn('Default email_design_settings row already exists, skipping');
+            return;
+        }
```
EvanHahn, line 16: question: When would this ever happen? If it did, should we throw an error instead?

ghost/core/core/server/data/migrations/versions/6.25/2026-03-26-15-47-00-insert-default-email-design-settings-row.js
EvanHahn: The risk of this, I think, is if someone already has a welcome email set up, they might be surprised to see its design change. But you could make the argument the other way too.

ghost/core/core/server/models/email-design-setting.js
```diff
@@ -0,0 +1,30 @@
+const ghostBookshelf = require('./base');
+
+const EmailDesignSetting = ghostBookshelf.Model.extend({
+    tableName: 'email_design_settings',
+
+    /**
+     * @returns {object} Default values for email design settings columns
+     */
```
EvanHahn: nit: This reduces type safety, I believe. TypeScript can infer the value that's being returned here. But if you specify it as `object`, all it'll know is that it's _some_ object. I think it's worth removing this entire annotation.

```suggestion
```

EvanHahn (approved):
A few non-blocking comments.

troyciesco (changes requested):
dropping the same "request changes" here as the parent pr just to make sure we bump the migration verison

ghost/core/core/server/data/migrations/versions/6.25/2026-03-26-15-47-00-insert-default-email-design-settings-row.js
cmraible: > When would this ever happen?

In the normal course of a Ghost site, it's _highly_ unlikely that it would, so honest answer is probably never. 

>  If it did, should we throw an error instead?

Good question - I think not, but I'm happy to be challenged on that. Rationale for _not_ throwing an error: it doesn't hurt anything to no-op here: if the row already exists, nothing really breaks in Ghost. 

If we instead throw an error, it would block any subsequent migrations from running, and prevent Ghost from booting. I do think there's a valid fail-fast argument to be made that preventing Ghost from booting if the schema/data is in an unexpected state is the right thing to do, which would be an interesting discussion. Practically speaking though, I'm mostly just following the existing conventions in previous migrations:

```
rg -U 'if \(existing\w*\).*\n.*\n.*return;' ghost/core/core/server/data/migrations/versions/

ghost/core/core/server/data/migrations/versions/5.40/2023-03-21-19-02-add-self-serve-integration-api-key.js
39:        if (existingKey) {
40:            logging.warn('Admin API key for "Self-Serve Migration Integration" already exists');
41:            return;

ghost/core/core/server/data/migrations/versions/6.14/2026-01-15-12-01-00-add-transistor-integration-api-key.js
37:        if (existingKey) {
38:            logging.warn('Admin API key already exists for Transistor integration');
39:            return;

ghost/core/core/server/data/migrations/versions/5.40/2023-03-21-18-42-add-self-serve-integration-role.js
14:        if (existingRole) {
15:            logging.warn('"Self-Serve Migration Integration" role already exists, skipping');
16:            return;

ghost/core/core/server/data/migrations/versions/6.14/2026-01-15-12-00-00-add-transistor-integration.js
14:        if (existing) {
15:            logging.warn('Found existing Transistor integration');
16:            return;

ghost/core/core/server/data/migrations/versions/5.113/2025-03-07-12-24-00-add-super-editor.js
14:        if (existingRole) {
15:            logging.warn('"Super Editor" role already exists, skipping');
16:            return;

ghost/core/core/server/data/migrations/versions/5.40/2023-03-21-18-52-add-self-serve-integration.js
15:        if (existingIntegration) {
16:            logging.warn('Integration already exists, skipping');
17:            return;

ghost/core/core/server/data/migrations/versions/6.23/2026-03-26-05-50-35-insert-default-email-design-settings-row.js
13:        if (existing) {
14:            logging.warn('Default email_design_settings row already exists, skipping');
15:            return;

ghost/core/core/server/data/migrations/versions/5.3/2022-07-06-09-26-add-ghost-explore-integration-api-key.js
39:        if (existingKey) {
40:            logging.warn('Admin API key already exists');
41:            return;

ghost/core/core/server/data/migrations/versions/5.100/2024-11-06-04-45-15-add-activitypub-integration.js
17:        if (existing) {
18:            logging.warn('Found existing Ghost ActivityPub integration');
19:            return;

ghost/core/core/server/data/migrations/versions/5.3/2022-07-06-07-58-add-ghost-explore-integration-role.js
14:        if (existingRole) {
15:            logging.warn('"Ghost Explore Integration" role already exists, skipping');
16:            return;

ghost/core/core/server/data/migrations/versions/5.3/2022-07-06-09-17-add-ghost-explore-integration.js
15:        if (existingIntegration) {
16:            logging.warn('Integration already exists, skipping');
17:            return;

```

ghost/core/core/server/models/email-design-setting.js
cmraible: Good shout, TIL!

ghost/core/core/server/data/migrations/versions/6.25/2026-03-26-15-47-00-insert-default-email-design-settings-row.js
```diff
@@ -0,0 +1,42 @@
...
+
+        await knex('email_design_settings').insert({
+            id: (new ObjectID()).toHexString(),
+            slug: DEFAULT_SLUG,
```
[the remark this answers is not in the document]
cmraible: This is fine - there should never be data already in this table anyhow, and if we're rolling this migration back, it _should_ be destructive

ghost/core/core/server/data/migrations/versions/6.25/2026-03-26-15-47-00-insert-default-email-design-settings-row.js
cmraible: Yeah I think I agree that this is probably right as-is, to avoid surprises with the welcome email design changing without any user intervention.

> A "copy design from newsletter" button would probably be a better way to smooth out the experience for people that do want stuff the same or want to save most of the effort of copying colors etc.

This is a good idea though, might be something we can fit in before GA

troyciesco (approved):

PR #26985, opened 2026-03-26
Fixed flaky data seeder test

kevinansfield:
no issue

When seeder database timestamp strings were reparsed in non-UTC CI environments, they could shift into the future and make Faker reject inverted date ranges causing flaky CI failures. Parsing those values as UTC and guarding the date range logic keeps the data generator test stable across timezones.

## What this does
- parse seeder database timestamp strings as UTC instead of local time
- guard date generation against inverted ranges when referenced records are at or after the current time
- add regression coverage for timezone parsing and date range handling

Example flaky run: https://github.com/TryGhost/Ghost/actions/runs/23592298819/job/68700344677?pr=26958

### DIFF
--- a/ghost/core/core/server/data/seeders/importers/comment-reports-importer.js
+++ b/ghost/core/core/server/data/seeders/importers/comment-reports-importer.js
@@ -49,9 +49,7 @@ class CommentReportsImporter extends TableImporter {
             return null;
         }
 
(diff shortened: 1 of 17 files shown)


kevinansfield:
Self-merging because I just hit this again in https://github.com/TryGhost/Ghost/actions/runs/23839885077/job/69492642560

PR #26986, opened 2026-03-26
Changed admin filters to use composable value sources

kevinansfield:

## What changed
- added first-class `ValueSource` support to `shade` filters, including clearer loading states and load-more support for async option sources
- introduced composable filter-source factories in `posts` for remote, local, hybrid, combined, and Ghost browse-backed value sources
- migrated the comments and members filters to field-owned value sources and removed the old bespoke search hooks they depended on
- updated the label picker and label sources to preserve selected labels during search, keep edit/delete flows working, and invalidate label queries safely after mutations
- kept the analytics filters on the simpler static-options path while restoring the loading bridge in `shade` so their UI behavior does not regress
- added focused unit coverage for the `shade` filter path and the hybrid value-source behavior

## Why
The old filter implementations duplicated the same search, hydration, and loading logic across screens while still leaking those concerns into the filter UI layer. This leaves the end state with one model: `shade` owns transient search UI, each field owns its value lookup strategy, small datasets can stay local, large datasets can stay remote or hybrid, and the existing members/comments/label UX keeps working while the remaining admin surfaces continue to converge.


rob-ghost (approved):
👍🏻 solid refactor and I much prefer this design

PR #27003, opened 2026-03-27
Fixed crash in Redis adapter when storeConfig is not provided

muratcorlu:
When storeConfig was omitted from the Redis cache adapter config, the retryStrategy callback would throw a TypeError whenever ioredis tried to reconnect to Redis, crashing Ghost instead of recovering gracefully. Fixed by using optional chaining on config.storeConfig.

Also fixed missing (config.username) JSDoc definition.

- Why are you making it?
- What does it do?


muratcorlu:
Hi @allouis 

Since you are already working on Redis Adapter, can you also have a look at this? Simple fix but I think it's nice to handle that null check to avoid application crashes.

allouis:
Hey @muratcorlu - thanks for your contribution!

PR #27005, opened 2026-03-27
[POC] 🎨 Added post editing presence prototype

9larsons:
## What this does

- adds a soft editing lease for posts and pages in posts_meta
- heartbeats from the editor every 30 seconds to surface an active editor
- expires the lease after 2 minutes without a heartbeat
- refreshes the lease on successful save when the current lease is unowned or expired
- does not steal the lease from another actively editing user
- shows presence in the editor header and the posts list
- keeps the normal save path untouched so presence does not update post timestamps or create revisions

## Notes

- this is intentionally a POC and not a final UX recommendation
- the editor treatment is clearer than the list treatment; the list UI still needs design input
- there is no independent polling on the posts list yet, so list presence updates when row data refreshes

## Testing

- yarn eslint on the touched admin and core files
- yarn test:single test/unit/server/services/posts/posts-service.test.js
- pre-commit lint-staged ran ember-template-lint and eslint on staged files

## Why draft

- the mechanics are working, but the list presentation should be reviewed by design before shipping

### DIFF
--- a/ghost/admin/app/components/gh-editor-post-status.hbs
+++ b/ghost/admin/app/components/gh-editor-post-status.hbs
@@ -66,4 +66,6 @@
         Draft
         {{unless @hasDirtyAttributes "- Saved"}}
     {{/if}}
+
+    <GhPostEditingIndicator @post={{@post}} />
 </div>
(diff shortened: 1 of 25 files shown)


9larsons:
CleanShot 2026-03-27 at 11 20 38@2x
CleanShot 2026-03-27 at 11 20 59@2x
Some ss - obviously the post list needs some significant reworking by design.

PR #27006, opened 2026-03-27
Added email design settings database permissions

cmraible:
closes https://linear.app/ghost/issue/NY-1181/add-email-design-setting-permissions

Adds CRUD permissions for the new `email_design_setting` resource, restricted to `Administrator` and `Admin Integration` roles.

**Changes:**
- Added database migration to create browse, read, edit, add, and destroy permissions for `email_design_setting`
- Updated fixtures to include the new permissions and assign them to Administrator and Admin Integration roles
- Updated fixtures hash in integrity test

PR #27008, opened 2026-03-27
Added nullable `email_design_setting_id` FK to automated_emails

cmraible:
closes https://linear.app/ghost/issue/NY-1180/add-email-design-setting-id-fk-to-automated-emails

## Summary

- Added nullable `email_design_setting_id` FK column to `automated_emails` table
- Plan to backfill all existing `automated_emails` rows with the default `email_design_settings` row, then make this column non-nullable in two follow up migration PRs to come.

Stacked on #27006


cmraible:
Closing this for the time being to take a different approach. This bakes in the default `email_design_settings` row into the DB schema definition for automated_emails, which I don't think is what we really want. I think we need to do the 3 step process here instead:
1. Create FK as a nullable field
2. Backfill existing automated_emails, using the default email_design_settings row
3. Make the FK non-nullable again

cmraible:
@Coderabbit review

troyciesco (approved):

PR #27013, opened 2026-03-30
Updated email verification flow to GA

aileen:
## What changed

This removes the `verificationFlow` Labs flag and the old escalation email fallback from Ghost so the webhook-based verification flow is now the only code path.

## Why this changed

`verificationFlow` is now GA. Keeping the flag and the legacy fallback logic around would leave Ghost with split verification behavior, dead admin Labs controls, and test coverage that still exercised code we no longer want to ship.

## Impact

Ghost now relies on the host verification webhook everywhere member verification is triggered.
The private Labs toggle for `verificationFlow` is gone.
Member creation and member imports no longer fall back to the legacy escalation email path.
Tests and snapshots were updated to match the GA behavior.

## Validation

- `yarn test:single test/unit/server/services/verification-trigger.test.js`
- `yarn test:single test/unit/server/services/verification/verification-webhook-service.test.ts`
- `node --check ghost/core/core/server/services/members/service.js`
- `node --check ghost/core/core/server/services/verification-trigger.js`
- `node --check ghost/core/test/unit/server/services/verification-trigger.test.js`
- `git diff --check`

## Notes

The database-backed admin and legacy suites could not be run in this environment because the local `sqlite3` native binding for the available Node runtime is missing.


ghost/core/test/integration/services/email-service/batch-sending.test.js
```diff
@@ -501,58 +611,8 @@ describe.skip('Batch sending tests', function () {
...
-
-        const newsletterSlug = fixtureManager.get('newsletters', 0).slug;
-        await agent.put(`posts/${id}/?newsletter=${newsletterSlug}`)
-            .body({posts: [updatedPost]})
-            .expectStatus(403);
-        sinon.assert.calledTwice(getSignupEvents);
-        assert.equal(settingsCache.get('email_verification_required'), true);
-
-        await configUtils.restore();
+    it('Cannot send an email when webhook verification is required', async function () {
```
aileen, line 594: This test is skipped through the  parent `describe` block. I updated it anyway and ensured it passes in isolation.

aileen:
Closing in favor of #27045

PR #27017, opened 2026-03-30
Migrated from yarn v1 to pnpm

9larsons:
ref https://linear.app/ghost/issue/PLA-17/

## Summary

- Replaced yarn v1 with pnpm as the monorepo package manager
- Using `shamefully-hoist=true` to match yarn v1's flat `node_modules` layout for compatibility
- All scripts, CI workflows, Docker configs, and documentation updated
- Fixed phantom dependencies exposed by pnpm's stricter resolution
- Simplified CI install script — pnpm's `onlyBuiltDependencies` handles native module builds natively

## Key changes

**Package manager bootstrap:**
- `pnpm-workspace.yaml` defines workspace packages (`ghost/*`, `apps/*`, `e2e`)
- `.npmrc` configures hoisted mode with `shamefully-hoist=true`
- `packageManager` field in root `package.json` pins pnpm version via corepack
- `pnpm.overrides` replaces per-package `resolutions` fields

**CI improvements:**
- Simplified `install-deps.sh` — removed manual sqlite3 rebuild (pnpm handles via allowlist)
- Removed redundant `cache: pnpm` from browser-tests job
- Removed redundant `prepare` scripts from 5 workspace packages

**Benchmarks (local, cold cache):**
| Metric | yarn v1 | pnpm | Change |
|--------|---------|------|--------|
| Cold install | 50.7s | 14.3s | 3.5x faster |
| Warm install | 0.7s | 2.7s | Slightly slower |
| Cold build (16 projects) | 1m29s | 1m14s | 17% faster |
| Docker dev build | 1m44s | 61s | 1.7x faster |
| Unit tests | 5818 pass / 12 fail | identical | No change |


.github/actions/restore-cache/action.yml
9larsons, line 1: I have a follow-up to use the `pnpm` store instead of caching `node_modules`. For now, this works, and limits our changes.

.github/scripts/clean.js
9larsons, line 1: Generally, we shouldn't need to clear the pnpm store. Sometimes, `pnpm store prune`, but we shouldn't need it in the repo nuke script.

.github/scripts/enforce-package-manager.js
9larsons, line 1: New script to help with those still trying to use their well-trained `yarn` commands.

.github/scripts/dependency-inspector.js
9larsons, line 1: Nothing points to this. I've created a follow-up to look into the origin - we may want to clean this up.

.github/scripts/install-deps.sh
9larsons, line 1: I'd like to try and replace this with GHAs' retry logic. Another todo.

.github/workflows/release.yml
9larsons, line 1: This'll be cleaned up as well when we move to the `pnpm` cache.

apps/comments-ui/vite.config.mts
9larsons, line 1: We should be able to clean up this resolution when we remove `shamefully-hoist`.

apps/shade/vite.config.ts
9larsons, line 1: Needed because hoisting resolves to glob 8.x (CJS) which only has a default export.

apps/signup-form/tsconfig.json
9larsons, line 1: Needed because of hoisting.

e2e/helpers/pages/admin/login-page.ts
9larsons, line 1: Linter update - `no-extra-semi`.

e2e/helpers/pages/admin/login-verify-page.ts
9larsons, line 1: Linter update - `no-extra-semi`.

ghost/admin/package.json
9larsons, line 1: `pnpm` doesn't support direct GitHub refs.

9larsons, line 1: `pnpm` doesn't use resolutions.

apps/shade/src/components/ui/animated-number.tsx
9larsons, line 1: Needed to fix type resolution.

.npmrc
9larsons, line 1: engine-strict = false is the default for pnpm, but it's listed to be clear. We may want to use true as we progress.

package.json
9larsons, line 1: The new resolutions/overrides are pulled up from ghost/admin as pnpm only reads from the root package.json. 

onlyBuiltDependencies is an allow list for those with post install scripts (eg those with native bindings like sqlite, or those using nx, esbuild).

ghost/core/monobundle.js
```diff
@@ -142,11 +169,52 @@ function getWorkspaces(from) {
...
+            if (!workspaceNames.has(key)) {
+                filteredOverrides[key] = value;
+            }
+        }
+
+        if (!pkgInfo.pkg.pnpm) {
+            pkgInfo.pkg.pnpm = {};
+        }
+        pkgInfo.pkg.pnpm.overrides = {...filteredOverrides, ...pkgInfo.pkg.pnpm.overrides};
+    }
```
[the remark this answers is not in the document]
9larsons: We want core to take precedence.

package.json
```diff
@@ -21,38 +17,39 @@
...
-    "test": "nx run-many -t test",
-    "test:unit": "nx run-many -t test:unit",
-    "test:e2e": "yarn workspace @tryghost/e2e test",
-    "test:e2e:analytics": "yarn workspace @tryghost/e2e test:analytics",
-    "test:e2e:all": "yarn workspace @tryghost/e2e test:all",
-    "test:e2e:debug": "DEBUG=@tryghost/e2e:* yarn test:e2e",
-    "main": "yarn main:monorepo && yarn main:submodules",
-    "main:monorepo": "git checkout main && git pull ${GHOST_UPSTREAM:-origin} main && yarn",
+    "lint": "pnpm nx run-many -t lint",
+    "test": "pnpm nx run-many -t test",
```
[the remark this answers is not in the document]
9larsons: More than likely our contributor docs are the ones that are out of date. These will get revised later.

PR #27018, opened 2026-03-30
Fixed Nx cache outputs for parse-email-address package

EvanHahn:
no ref

The build target compiled to `build/` but Nx's default cache outputs only tracked `dist`, `es`, `types`, and `umd`, causing the directory to be missing after cache hits.

### DIFF
--- a/ghost/parse-email-address/package.json
+++ b/ghost/parse-email-address/package.json
@@ -34,5 +34,12 @@
   },
   "dependencies": {
     "parse-email-address": "0.0.2"
+  },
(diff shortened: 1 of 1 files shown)


9larsons (approved):
`question (non-blocking):` `/build` is an outlier here and `/dist` is the nx default, should we just be using that?

EvanHahn:
I tried to do that (see #27413) but it didn't work. Choosing not to investigate further, but happy to if you think that's important.

PR #27019, opened 2026-03-30
Added `verificationFlow`-gated verification test coverage

aileen:
ref [GVA-732](https://linear.app/ghost/issue/GVA-732/roll-out-to-ga-and-remove-old-logic-in-ghost)

This prep PR isolates the flag-aware verification test updates so the GA rollout PR can focus on removing the labs and legacy runtime paths. It adds a shared webhook test helper, makes the admin members webhook e2e test deterministic while keeping it gated behind `verificationFlow`, adds isolated batch-sending coverage for the flag-enabled path, and switches the importer test reset to `settingsCache`.

### DIFF
--- a/ghost/core/test/e2e-api/admin/__snapshots__/config.test.js.snap
+++ b/ghost/core/test/e2e-api/admin/__snapshots__/config.test.js.snap
@@ -29,6 +29,7 @@ Object {
       "tagsX": true,
       "themeTranslation": true,
       "transistor": true,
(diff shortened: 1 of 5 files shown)


aileen:
Closing this upstream copy because GitHub cannot retarget it onto the fork-only labs-base branch. The correct stacked prep PR is aileen/Ghost#6, which sits on top of the labs-base branch that mirrors local main.

PR #27025, opened 2026-03-31
Update Spanish translations in portal.json

bobvaneck:
Added two missing strings for the new retention offers. They were provided by a Pro customer.


ghost/i18n/locales/es/portal.json
```diff
@@ -72,7 +72,7 @@
     "Emails": "Correos electrónicos",
     "Emails disabled": "Correos electrónicos desactivados",
     "Ends {offerEndDate}": "Termina el {offerEndDate}",
-    "Enjoy {amountOff} off forever.": "",
+    "Enjoy {amountOff} off forever.": "Disfruta de un {amountOff} descuento para siempre",
```
cathysarisky: 🤖 AI: I think there might be a small grammar issue here: "Disfruta de un {amountOff} descuento para siempre" sounds a bit off in Spanish. Would "Disfruta de un descuento de {amountOff} para siempre" or "Disfruta de {amountOff} de descuento para siempre" be more natural while keeping the same meaning?

cathysarisky (commented):
Hey @bobvaneck - do you want to ask the customer about the comments below, or you want me to just merge it as-is?  Since we currently have nothing and they'd be seeing English, possibly poor Spanish wording is probably an improvement? :)

ghost/i18n/locales/es/portal.json
bobvaneck: Hmm this came directly from a Spanish speaking customer, but I have to agree that the last option is actually the best one, it just lacks the "un" part, so it should be "Disfruta de un {amountOff} de descuento para siempre".

cathysarisky, line 75:
```suggestion
    "Enjoy {amountOff} off forever.": "Disfruta de un {amountOff} de descuento para siempre.",
```

PR #27029, opened 2026-03-31
Fixed label filter replacing first selection when adding second

rob-ghost:
When using the members list label filter with `allowMultiple` enabled, selecting a second label value would replace the first filter instead of adding a new one. For example, filtering by "VIP" and then adding "Premium" would drop the "VIP" filter entirely, showing only "Premium" results instead of narrowing to members with both labels.

The root cause was in the `addFilter` and `addFilterWithOption` callbacks in the Filters component. Both functions unconditionally looked for an existing filter on the same field and updated it in place, which is correct for single-filter mode but wrong when multiple filters on the same field are allowed. The `addFilter` function also pre-populated `tempSelectedValues` from the existing filter every time the options popover opened, so the new filter always started with the old filter's values rather than an empty selection.

The fix introduces an `allowMultiple` guard in both callbacks. In single-filter mode the existing update-in-place behaviour is preserved. In multi-filter mode, `addFilter` now always starts with an empty selection and `addFilterWithOption` always creates a fresh filter entry, so each label choice becomes its own independent filter that combines with the others.

An E2E test covers the specific scenario: add a "VIP" label filter, then add a "Premium" label filter, and verify only the member carrying both labels remains visible.

### DIFF
--- a/apps/shade/src/components/ui/filters.tsx
+++ b/apps/shade/src/components/ui/filters.tsx
@@ -2214,51 +2214,56 @@ export function Filters<T = unknown>({
         [filters, onChange]
     );
 
+    const closeFilterPopover = useCallback(() => {
+        setAddFilterOpen(false);
+        setSelectedFieldKeyForOptions(null);
+        setTempSelectedValues([]);
+    }, []);
+
     const addFilter = useCallback(
         (fieldKey: string) => {
             const field = fieldsMap[fieldKey];
-            if (field && field.key) {
-                // For select and multiselect types, show options directly
-                if (field.type === 'select' || field.type === 'multiselect') {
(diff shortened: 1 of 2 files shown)


apps/shade/src/components/ui/filters.tsx
```diff
@@ -2238,85 +2246,63 @@ export function Filters<T = unknown>({
                         : field.type === 'boolean'
                             ? 'is'
                             : 'is');
```
kevinansfield: Nit: Could this be simplified further?


```suggestion
            const defaultOperator = field.defaultOperator ||
                (field.type === 'daterange' || field.type === 'numberrange' ? 'between' : 'is');
```

kevinansfield (approved):

PR #27043, opened 2026-04-01
Migrate comment moderation filters to new filter abstraction

jonatansberg:
## What changed

This reworks comment moderation filters in `apps/posts` to use the same canonical field/codec/NQL pipeline as members instead of the old ad-hoc query-param implementation.

It also keeps `id` and `thread` out of filter state:
- `filter` is now the only concern of the comment filter-state hook
- legacy `?id=is:<commentId>` deep links are handled separately at the page level
- single-comment mode still clears back to the full comments view

Follow-up work on this branch also tightened the shared value-source behavior so selected author/post filters remain representable when the backing record cannot be hydrated, without allowing synthetic `ID: ...` placeholders to win over real hydrated labels in the combined post/page source.

## Why

The previous comments filter implementation had a hand-rolled `buildNqlFilter()` path, custom URL shape, no NQL parsing, and browser-local date handling. That had drifted away from the mainline filter abstractions and made comments harder to reason about and maintain.

This brings comments back onto the current shared model while keeping the deep-link behavior that moderation flows still rely on.

## User impact

- comment moderation filters now round-trip through canonical `?filter=<NQL>` URLs
- filter parsing/serialization is aligned with members
- exact-date comment filtering uses site timezone semantics instead of browser-local assumptions
- legacy single-comment links via `?id=is:<commentId>` still work
- missing selected author/post resources stay visible in filter UI instead of collapsing into a broken state

## Validation

- `cd apps/posts && yarn vitest run test/unit/hooks/create-remote-value-source.test.tsx test/unit/hooks/create-combined-value-source.test.tsx test/unit/views/comments test/unit/utils/filter-normalization.test.ts`
- `cd apps/posts && yarn eslint --ext .ts,.tsx src/hooks/filter-sources/create-combined-value-source.ts src/hooks/filter-sources/use-post-resource-value-source.ts test/unit/hooks/create-combined-value-source.test.tsx`
- Earlier targeted verification on the main migration also included:
  - `cd apps/posts && yarn tsc --noEmit`
  - focused lint/test runs for the canonical comments filter modules

## Notes

- existing moderation link producers remain on `?id=is:<commentId>` in this pass
- `.claude/worktrees/` is an unrelated untracked local directory and is not part of this PR


codeCraft-Ritik (commented):
Excellent refactor! Bringing comment moderation filters in line with the canonical pipeline is a big improvement

kevinansfield (approved):

kevinansfield:
All working fine locally. Only thing I found was the comment text filtering not having any debounce so we fire an API request on every keypress but I think that was an existing issue?

PR #27045, opened 2026-04-01
Moved updated email verification webhook flow to GA

aileen:
closes [GVA-732](https://linear.app/ghost/issue/GVA-732/roll-out-to-ga-and-remove-old-logic-in-ghost)

The webhook-based verification flow is now the default behavior, so the `verificationFlow` labs flag and the old escalation email fallback were removed from Ghost. This updates the members, importer, and unit test coverage to expect webhook delivery by default, aligns the webhook configuration warnings with the GA rollout, and leaves the verification trigger maintaining a single supported path going forward.

### DIFF
--- a/ghost/core/core/server/services/verification-trigger.js
+++ b/ghost/core/core/server/services/verification-trigger.js
@@ -3,17 +3,7 @@ const DomainEvents = require('@tryghost/domain-events');
 const {MemberCreatedEvent} = require('../../shared/events');
 
 const messages = {
-    emailVerificationNeeded: `We're hard at work processing your import. To make sure you get great deliverability, we'll need to enable some extra features for your account. A member of our team will be in touch with you by email to review your account make sure everything is configured correctly so you're ready to go.`,
-    emailVerificationEmailSubject: `Email needs verification`,
-    emailVerificationEmailMessageImport: `Email verification needed for site: {siteUrl}, has imported: {amountTriggered} members in the last 30 days.`,
-    emailVerificationEmailMessageAdmin: `Email verification needed for site: {siteUrl} has added: {amountTriggered} members through the Admin client in the last 30 days.`,
-    emailVerificationEmailMessageAPI: `Email verification needed for site: {siteUrl} has added: {amountTriggered} members through the API in the last 30 days.`
-};
-
-const verificationMessageBySource = {
-    api: messages.emailVerificationEmailMessageAPI,
-    admin: messages.emailVerificationEmailMessageAdmin,
-    import: messages.emailVerificationEmailMessageImport
+    emailVerificationNeeded: `We're hard at work processing your import. To make sure you get great deliverability, we'll need to enable some extra features for your account. A member of our team will be in touch by email to review your account and make sure everything is configured correctly so you're ready to go.`
 };
 
 class VerificationTrigger {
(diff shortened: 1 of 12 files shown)


ghost/core/test/legacy/api/admin/members-importer.test.js
```diff
@@ -246,78 +244,71 @@ describe('Members Importer API', function () {
...
 
-        mockManager.assert.sentEmail({
-            subject: 'Email needs verification'
-        });
-
-        // Don't reset email verification required flag, as the following test
-        // relies on that value being true already.
-    });
-
-    it('Can still import members once email verification is required but does not send email', async function () {
```
aileen, line 284: combined those two tests, to test the full flow instead of relying on the result of the previous test

ghost/core/test/legacy/api/admin/members-importer.test.js
```diff
@@ -22,8 +21,7 @@ describe('Members Importer API', function () {
     });
 
     beforeEach(function () {
-        emailMockReceiver = mockManager.mockMail();
-        mockManager.mockLabsDisabled('verificationFlow');
+        mockManager.mockMail();
```
aileen, line 24: Still needed to assert the `Your member import is complete` email

ghost/core/core/server/services/verification-trigger.js
sam-lord, line 87: thank you Sonar for telling us this 😂

ghost/core/core/server/services/verification-trigger.js
```diff
@@ -260,36 +225,30 @@ class VerificationTrigger {
...
-                });
+        try {
+            webhookWasSent = await this._sendVerificationWebhook({
+                amountTriggered: amount,
+                threshold: threshold ?? amount,
+                method: method ?? source ?? 'import'
+            });
+        } catch (error) {
+            // `sendVerificationWebhook` already logs delivery failures.
+            return {needsVerification: false};
```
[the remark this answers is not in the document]
aileen: I'll add a test for this and ensure the correct behavior on webhook sending failure

sam-lord (approved):
Looks great :)

ghost/core/test/legacy/api/admin/members-importer.test.js
```diff
@@ -246,78 +244,71 @@ describe('Members Importer API', function () {
...
+        assert.equal(secondJsonResponse.meta.stats.invalid.length, 0);
 
         assert.equal(settingsCache.get('email_verification_required'), true, 'Email verification should still be required');
 
         // Don't send another email
-        emailMockReceiver.assertSentEmailCount(0);
+        assert.equal(receivedWebhookRequests.length, 1, 'Expected no further webhook requests after second import');
 
         // reset email verification required flag for next test
         await restoreEmailVerificationUtils();
```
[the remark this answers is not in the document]
aileen: Fixed in b26eab7: wrapped both importer email-verification tests in try/finally so restoreEmailVerificationUtils() always runs and test-side verification state cannot leak into later cases.

ghost/core/core/server/services/verification-trigger.js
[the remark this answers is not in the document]
aileen: Dismissing: this fail-open behavior is intentional for the current GA flow. If webhook delivery throws, we leave  unset so the customer stays unblocked and Ghost retries the webhook on the next qualifying add/import trigger instead of latching a blocked state during a transient webhook outage.

ghost/core/core/server/services/verification-trigger.js
aileen: Lol, right?

PR #27058, opened 2026-04-01
Updated Tinybird CI to trigger & follow infra workflow

JoeeGrigg:
Ghost CI now triggers & waits for the dispatched traffic analytics workflow and mirrors its final result.

This moves the Tinybird deploy check out of this repo, over to an internally managed repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -738,11 +738,6 @@ jobs:
         image: tinybirdco/tinybird-local:latest
         ports:
           - 7181:7181
-    env:
-      TINYBIRD_HOST: ${{ secrets.TINYBIRD_HOST }}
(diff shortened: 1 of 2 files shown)


cmraible (commented):
PR looks good! Have you seen these checks fail at all? I just want to make sure if they do fail, that it's reported properly here and prevents merging. 

Should be able to test this by adding some gibberish to the other repo and triggering this workflow again.

JoeeGrigg:
> PR looks good! Have you seen these checks fail at all? I just want to make sure if they do fail, that it's reported properly here and prevents merging.
> 
> Should be able to test this by adding some gibberish to the other repo and triggering this workflow again.

Yep I tested it last week, before making the tests pass I first tested it failing and seems to work well. I'll get this merged 🎉

PR #27085, opened 2026-04-02
Removed unused "destroy automated email" endpoint

EvanHahn:
closes https://linear.app/ghost/issue/NY-1201

We aren't using this endpoint today, and its existence makes a future change more difficult. Let's remove it.


troyciesco (approved):
lgtm - just noting we still have the "delete automated emails" permission in `ghost/core/core/server/data/schema/fixtures/fixtures.json`, but i think we still want that based on the other work you're doing.

EvanHahn:
Didn't know about that...thank you!

PR #27086, opened 2026-04-02
Split automated emails into pair for future extensibility

EvanHahn:
closes https://linear.app/ghost/issue/NY-1188

_I recommend reviewing this commit-by-commit._

We currently have one table for welcome emails: `automated_emails`. Instead, we want a table for automations and a table for all the emails in that automation.

This change isn't useful on its own, but will be followed up with the ability to add more than one email to an automation.


ghost/core/core/server/data/migrations/versions/6.26/2026-04-01-20-46-31-split-automated-emails-into-welcome-email-tables.js
```diff
@@ -0,0 +1,257 @@
...
+    updated_at: {type: 'dateTime', nullable: true},
+    '@@INDEXES@@': [
+        ['slug'],
+        ['status']
+    ]
+};
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        // 1. Create welcome_email_automations table (without FK to emails, to break circular reference)
```
cmraible, line 55: I would probably split this migration into a couple of separate migrations. The main smell for me is that this mixes DDL and DML in a single migration. I wouldn't say this is a hard rule (a few existing migrations do violate this), but unless we have a good reason to mix the two, I think it's worth splitting them up. Beyond just being a bit easier to reason about, MySQL has some quirks where e.g. DML operations like creating tables are committed immediately (regardless of any transactions). The risk there is that parts of the migration could succeed, while others fail, leaving the DB in a partially migrated state. 

As written today (though I think some of this will change with the data model changes you mentioned), I'd split this into probably 3 migrations:
1. Create both tables
2. Backfill from automated_emails, along with any code changes needed for welcome emails to continue to function with these new tables
3. Add/update foreign keys and drop automated_emails

cmraible (changes requested):
Throwing a bit of a wrench in your plans, sorry! The changes themselves all make sense (though I haven't looked too closely beyond the first commit), but I think we should split up this one migration into a couple more focused migrations — see comment below.

EvanHahn:
Closing in favor of #27183, #27184, #27185, and #27186.

PR #27175, opened 2026-04-06
🌐 Merged fragment translation keys in Portal notifications

rob-ghost:
## Summary
- Merged split-sentence translation keys in Portal's `notification.js` into complete sentences using `@doist/react-interpolate`
- Removed fragment keys (`"Click here to retry"`, `"You've successfully subscribed to"`, `"Signup error: Invalid link"`, `"Could not sign in. Login link expired."`) that prevented translators from reordering words for different languages
- Added combined interpolated keys with inline `<a>` and `<strong>` tags so translators have full sentence context
- Updated all 62 locale `portal.json` files and `context.json` with descriptions for the new keys

### Keys changed
| Old (removed) | New (added) |
|---|---|
| `"Could not sign in. Login link expired."` + `"Click here to retry"` | `"Could not sign in. Login link expired. <a>Click here to retry</a>"` |
| `"You've successfully subscribed to"` + site title | `"You've successfully subscribed to <strong>{siteTitle}</strong>"` |
| `"Signup error: Invalid link"` + `"Click here to retry"` | `"Signup error: Invalid link. <a>Click here to retry</a>"` |

### Why
Splitting sentences across multiple `t()` calls prevents translators from reordering words to match the grammar of their target language. Using `@doist/react-interpolate` (already a Portal dependency) allows embedding React elements within a single translatable string.

## Test plan
- [ ] Start Portal and trigger sign-in error notification (use expired magic link) — verify retry link renders and works
- [ ] Trigger successful signup — verify site title appears in bold in the notification
- [ ] Trigger signup error (use expired signup link) — verify retry link renders and works
- [ ] Verify i18n lint passes: `yarn workspace @tryghost/i18n lint:translations`


cathysarisky (changes requested):
Hey @rob-ghost - excellent thinking! There's one issue, though:  These strings have translations already in many languages, and your proposal will wipe them out, causing non-English sites to see English until someone files a new PR, and it gets approved and merged.  

I think the right thing to do here is to write a script that constructs translations for the new strings using our our "previous best" values.  That avoids an initial regression.  If you want me to do it, I can, but I think that work would be best placed right in this PR.

rob-ghost:
Thanks @cathysarisky I gave it a go to merge together the keys whose locales already had translations for the keys we're replacing.

apps/portal/src/components/notification.js
```diff
@@ -40,19 +41,34 @@ const NotificationText = ({type, status, context}) => {
...
         );
     } else if (type === 'signup' && status === 'success') {
         return (
             <p>
-                {t('You\'ve successfully subscribed to')} <br /><strong>{context.site.title}</strong>
+                <Interpolate
+                    mapping={{
+                        strong: <strong />
+                    }}
+                    string={t('You\'ve successfully subscribed to <br /><strong>{siteTitle}</strong>', {siteTitle: context.site.title})}
```
cathysarisky, line 59: ok, this breaks my heuristic that a br means a new string... 😆 

Is having the br necessary here?  Will it survive a variety of languages (which could be adjusted by translators) and site title lengths (which cannot?   Would it be better to let the browser make the decision about when to break the line?  And does the layout break if a translator removes it?  I'm imagining that the zh and zh-Hant translators would probably remove it, for example.

apps/portal/src/components/notification.js
cathysarisky, line 32: typo - you meant signup, right? :)  🎵 🎵

apps/portal/src/components/notification.js
```diff
@@ -40,19 +41,34 @@ const NotificationText = ({type, status, context}) => {
     } else if (type === 'signin' && status === 'error') {
         return (
             <p>
-                {t('Could not sign in. Login link expired.')} <br /><a href={signinPortalLink} target="_parent">{t('Click here to retry')}</a>
+                <Interpolate
+                    mapping={{
+                        a: <a href={signinPortalLink} target="_parent" />
+                    }}
+                    string={t('Could not sign in. Login link expired. <br /><a>Click here to retry</a>')}
+                />
```
cathysarisky, line 49: To play devil's advocate, the fact that there's a br might really suggest that these are two separate ideas that can be safely translated as separate strings.  And then we'd keep the enforced common phrasing for "Click here to retry", which really does seem to be a separate idea.

cathysarisky (commented):
THANK YOU for preserving the existing translations! 🙏 
A couple comments below.

cathysarisky (approved):
Marking as approved so that I'm not blocking you, but see comments about error in variable name and some questions about whether this swings a little too far in the other direction. :)

apps/portal/src/components/notification.js
rob-ghost: Looking at this with fresh eyes and I think you're right, this would revert this PR and make it a no-op so I'll close this PR

rob-ghost:
Closing this because the two translation keys were actually correct to be split up which makes this a no-op PR.

apps/portal/src/components/notification.js
cathysarisky: @rob-ghost - I think it's perhaps worth keeping "You've successfully subscribed to {siteTitle}".  That's an improvement :)

cathysarisky:
> Closing this because the two translation keys were actually correct to be split up which makes this a no-op PR.

There was also this one, which might be worth keeping? https://github.com/TryGhost/Ghost/pull/27175/changes#diff-b980c9767edc8137443b5b075a307bb3bca189e537e14f1da20e9ae0e9684904R55-R59

rob-ghost:
@cathysarisky I'm putting up a new one with just that change now

PR #27170, opened 2026-04-06
🌐 Backfilled empty i18n context.json strings

troyciesco:
no ref

- backfills empty i18n context hints in `context.json`
- future PR will prevent empty hints from being added (https://github.com/TryGhost/Ghost/pull/27171)

PR #27176, opened 2026-04-06
🌐 Filled empty context descriptions and added CI guard

rob-ghost:
## Summary
- Filled in meaningful context descriptions for all 33 translation keys in `context.json` that previously had empty strings
- Added a CI guard in `generate-context.js` that rejects empty descriptions, preventing new untranslatable keys from being merged
- Every key now tells translators where it appears and what it does (e.g., Portal retention offer, signin email, comments UI)

### CI guard
When `CI=true`, `generate-context.js` now checks for empty descriptions after validating that `context.json` is up to date. If any keys have empty descriptions, the script lists them and exits with a non-zero status. This runs as part of `yarn translate` which is already called during `yarn test` in the i18n package.

## Test plan
- [ ] Verify `yarn workspace @tryghost/i18n test` passes
- [ ] Verify CI guard catches empty descriptions: temporarily blank a description in `context.json`, then run `CI=true node generate-context.js` from `ghost/i18n/` — should fail with a list of the empty key(s)


rob-ghost:
Closing in favour of https://github.com/TryGhost/Ghost/pull/27170 and https://github.com/TryGhost/Ghost/pull/27171

PR #27179, opened 2026-04-06
🌐 Added ESLint rule to detect adjacent t() calls in JSX

rob-ghost:
## Summary

- Added a custom ESLint rule `no-adjacent-t-calls` that detects adjacent `t()` calls in JSX — a pattern that splits sentences across multiple translation keys, preventing translators from reordering words for different languages
- Rule placed in shared location (`ghost/i18n/eslint-rules/`) so it can be reused across apps
- Enabled as a **warning** in `comments-ui` (not an error, since some adjacent `t()` calls are independent UI elements rather than split sentences)
- Updated lint-staged config so the `--rulesdir` flag is passed correctly during pre-commit

### How it works

The rule flags sibling JSX children that both contain `t()` calls, e.g.:

```jsx
// ⚠️ Flagged: split sentence
<p>{t('Could not sign in.')} <a>{t('Click here')}</a></p>

// ✅ Good: single translation key with interpolation
<p><Interpolate string={t('Could not sign in. <a>Click here</a>')} mapping={{a: <a />}} /></p>
```

### Why warn instead of error

Some adjacent `t()` calls are genuinely independent UI elements (e.g., a label next to a button), not split sentences. The rule is warning-only to surface potential issues without blocking CI. Future work could refine the heuristics or add `eslint-disable` comments for known false positives.

## Test plan

- [x] `yarn workspace @tryghost/comments-ui lint` passes (0 errors, warnings only)
- [x] Rule correctly flags known split-sentence patterns
- [x] Rule does not produce errors that would block CI

### DIFF
--- a/.lintstagedrc.cjs
+++ b/.lintstagedrc.cjs
@@ -5,6 +5,11 @@ const FLAT_CONFIG_WORKSPACES = [
     'apps/admin'
 ];
 
+// Workspaces that use legacy eslintrc but need extra flags (e.g. --rulesdir)
+const LEGACY_SCOPED_WORKSPACES = [
+    {path: 'apps/comments-ui', extraFlags: ['--rulesdir', path.join(__dirname, 'ghost/i18n/eslint-rules')]}
+];
+
 function normalize(file) {
     return file.split(path.sep).join('/');
 }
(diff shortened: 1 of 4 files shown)


rob-ghost:
I actually don't love the UX for this so I'm going to close it

PR #27186, opened 2026-04-06
Dropped `automated_emails` table

EvanHahn:
closes https://linear.app/ghost/issue/NY-1188
ref https://github.com/TryGhost/Ghost/pull/27185

This removes the old automated emails table now that we aren't using it anymore.


ghost/core/core/server/data/migrations/versions/6.27/2026-04-06-16-29-31-drop-automated-emails-table.js
```diff
@@ -0,0 +1,46 @@
...
+    async function up(knex) {
+        const exists = await knex.schema.hasTable('automated_emails');
+        if (!exists) {
+            logging.warn('Skipping dropping table automated_emails - does not exist');
+            return;
+        }
+
+        logging.info('Dropping table: automated_emails');
+        await commands.deleteTable('automated_emails', knex);
+    },
```
cmraible: suggestion: use the dropTables migration util

cmraible (commented):
One minor suggestion on the migration, otherwise LGTM!

EvanHahn:
Moving this to draft until #27185 is merged. I'll regenerate the migration and address the PR review comment.

ghost/core/core/server/data/migrations/versions/6.27/2026-04-06-16-29-31-drop-automated-emails-table.js
EvanHahn: Done. This does make the migration irreversible, but I think that's fine—it's not like we can't add another migration later to add it back if we need.

EvanHahn: Actually, I think we want this migration to be reversible. Keeping this as is.

PR #27191, opened 2026-04-06
Updated welcome email renderer to use custom design settings when enabled

cmraible:
closes https://linear.app/ghost/issue/NY-1197/use-design-settings-from-the-database-when-rendering-welcome-emails

## Summary

- Updated the welcome email renderer to use custom design settings (background color, button style, header image, footer content, badge, etc.) from the database instead of hardcoded defaults
- All new behavior is gated behind the `welcomeEmailsDesignCustomization` labs flag — existing welcome emails are unchanged when the flag is off
- The service wrapper reinitializes when the labs flag changes, so toggling the flag takes effect without a restart

## Changes

**Renderer** (`member-welcome-email-renderer.js`):
- Reads `designSettings` and passes them through to `getEmailDesign()` (when flag is on) or uses hardcoded defaults (when flag is off)
- Reuses `registerHelpers` and the newsletter styles partial for consistent rendering with newsletter emails
- Passes header image, footer content, badge, and header title settings to the wrapper template

**Service** (`service.js`):
- Loads `emailDesignSetting` relation when fetching automated emails (flag-gated)
- Caches design settings at load time alongside other email data
- Wrapper class tracks labs flag state and reinitializes when it changes

**Template** (`wrapper.hbs`):
- Added `post-content-row` / `post-content` classes for newsletter-consistent content styling
- Added conditional footer content, badge, and powered-by-Ghost sections

## Test plan

- [x] Existing snapshot tests pass unchanged with the flag off (no regression)
- [x] New snapshot tests cover flag-on rendering
- [x] Unit tests verify design settings are passed through correctly vs. ignored when flag is off
- [x] Integration tests verify cached design settings are used (not re-fetched at send time)
- [x] Integration test verifies service reinitializes when labs flag changes


cmraible:
@Coderabbit review

ghost/core/core/server/services/member-welcome-emails/service.js
```diff
@@ -93,21 +94,30 @@ class MemberWelcomeEmailService {
         return this.#defaultNewsletterSenderOptions;
     }
 
+    #useDesignCustomization() {
```
troyciesco, line 97: nit: we don't need to change it here and in the other file if we can't think of anything demonstrably better, but everytime i see `use` i think it's a React hook lol

ghost/core/core/server/services/member-welcome-emails/member-welcome-email-renderer.js
```diff
@@ -108,25 +141,33 @@ class MemberWelcomeEmailRenderer {
...
+     * @param {Object} [options.designSettings] - Email design settings loaded from the database
      * @param {Object} options.member - Member data (name, email)
      * @param {Object} options.siteSettings - Site settings (title, url, accentColor)
      * @returns {Promise<{html: string, text: string, subject: string}>}
      */
-    async render({lexical, subject, member, siteSettings}) {
+    async render({lexical, subject, designSettings = {}, member, siteSettings}) {
+        const useDesignCustomization = labs.isSet('welcomeEmailsDesignCustomization');
+
+        designSettings = useDesignCustomization ? {
```
troyciesco: I'm not 100% sure where/how this is happening yet but i don't want to lose track of it:

when i have the labs flag _off_, i don't seem to get spacing above the start of the content

Image


i thought maybe it had something to do with logic for toggling the header, but with labs flag on and the image and publication title toggled off, i still get decent spacing:

Image

cmraible: The missing spacing at the top is present on main currently - ignoring since it's not _technically_ broken, and this will override that behavior once shipped.

ghost/core/core/server/services/member-welcome-emails/service.js
troyciesco: nvm this is getting removed later

troyciesco (approved):

PR #27202, opened 2026-04-07
Spike: CI Improvements

acburdine:
# Summary

This PR contains a few different improvements/optimizations to the CI pipeline:

- [ ] Update/Fix Nx caching with shared-fs-cache plugin 

This change will allow caching of built Nx artifacts within the same job run, reducing the amount of duplicate `build` steps invoked across all of the disparate jobs

2. [x] Standardize Nx Task Graph/utilize in setup + downstream jobs

This change updates the nx "base commit" detection to use the standard nx-set-shas workflow (which will look at the last _successful_ commit instead of just the commits contained within a given git push). Additionally, it exports affected projects from the setup job so that downstream jobs can take advantage of it rather than relying on the `changed_x` file path detection.

3. [x] Standardize app acceptance tests + convert to matrix build

This change standardizes the script naming of acceptance tests in the various `apps/*` packages that have playwright tests, and combines the different playwright jobs into a single acceptance test job that runs them all with a matrix.

Making the PR to validate the CI changes themselves, but will leave in draft until the PR is ready.


acburdine:
most of this work has been incorporated into things I think, so will close

PR #27174, opened 2026-04-06
🌐 Consolidated commenting-disabled translation keys into single interpolated string

rob-ghost:
The commenting-disabled box in comments-ui previously rendered its message using three separate translation keys: "You can't post comments in this publication.", "Contact support", and "for more information." This fragmented approach made it difficult for translators to produce natural-sounding sentences, since word order and grammar vary across languages. Splitting a sentence into independent fragments forces translators to work around English syntax rather than expressing the idea naturally in their own language.

This change consolidates those three keys into a single interpolated string: `"You can't post comments in this publication. <a>Contact support</a> for more information."` The `<a>` tag is handled at render time by `@doist/react-interpolate`, which swaps it for the actual mailto link element. Translators now see the full sentence and can reorder or restructure it freely while keeping the link in the right place.

All 63 locale files have been updated to remove the two obsolete keys and add the new combined key (with an empty value, ready for translation). The context file has also been updated with descriptions for the new key and a couple of previously undocumented keys.

PR #27203, opened 2026-04-07
🌐 Combined portal subscribe notification translation key

rob-ghost:
The Portal subscribe success notification previously split its message across two translation keys and a hardcoded `<strong>` tag — `t("You've successfully subscribed to")` followed by a raw `<strong>{context.site.title}</strong>`. This made it impossible for translators to reorder the site title within the sentence, which is necessary for many languages where the grammar requires the site name to appear in a different position.

This change combines the phrase and the site title into a single translation key: `"You've successfully subscribed to <strong>{siteTitle}</strong>"`. The `@doist/react-interpolate` library (already used elsewhere in Portal) handles rendering the `<strong>` tag from the translated string. All 60+ locale files have been updated to use the new key format, preserving existing translations while embedding the `<strong>{siteTitle}</strong>` placeholder in each.

This follows the same pattern established in #27174 for the commenting-disabled translation keys.

### DIFF
--- a/apps/portal/package.json
+++ b/apps/portal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/portal",
-  "version": "2.67.10",
+  "version": "2.67.11",
   "license": "MIT",
   "repository": "https://github.com/TryGhost/Ghost",
   "author": "Ghost Foundation",
--- a/apps/portal/src/components/notification.js
+++ b/apps/portal/src/components/notification.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import Interpolate from '@doist/react-interpolate';
 import Frame from './frame';
 import AppContext from '../app-context';
 import NotificationStyle from './notification.styles';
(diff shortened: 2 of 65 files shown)


cathysarisky (approved):
LGTM!  Thanks for making things better for our non-English sites! :)

rob-ghost:
@cathysarisky Thanks for the approval, I'm going to merge.

CodeRabbit highlighted some issues in the translations, but none were introduced in this PR, they just become evident thanks to the string now being combined correctly, which is exactly what we want to see.

I won't fix the translations here because I'm not a native speaker and I can't verify any of the AI output, but hopefully spotting and fixing these grammatical issues will be easier now thanks to the interpolation.

cathysarisky:
> CodeRabbit highlighted some issues in the translations, but none were introduced in this PR, they just become evident thanks to the string now being combined correctly, which is exactly what we want to see.
> 
> I won't fix the translations here because I'm not a native speaker and I can't verify any of the AI output, but hopefully spotting and fixing these grammatical issues will be easier now thanks to the interpolation.

Agreed!  Code Rabbit is just giving you some good justification for why this is the right thing to do! :)   Translators can take it from here.

PR #27213, opened 2026-04-07
Bumped upload- and download-artifact GitHub Action versions

EvanHahn:
no ref

This updates upload- and download-artifact GitHub Actions to their latest versions. It also pins to a SHA for security and reproducibility.

Some of these involve upgrading major versions, which I don't think should cause any problems based on reading the release notes.


EvanHahn:
Looks like this has already been applied. Abandoning.

PR #27215, opened 2026-04-07
Bumped Docker login GitHub Action to latest version

EvanHahn:
no ref

This bumps `docker/login-action` to the latest version, v4.1.0. I don't think the major version bump should be a problem.

It also pins to a specific SHA, which is useful for security and reproducibility.


EvanHahn:
Looks like this has already been applied. Abandoning.

PR #27217, opened 2026-04-08
Improved storage path validation

kevinansfield:
ref https://linear.app/ghost/issue/ONC-1595/

- shared path normalization across local storage reads and URL conversion so image path handling stays consistent

### DIFF
--- a/ghost/core/core/server/adapters/storage/LocalStorageBase.js
+++ b/ghost/core/core/server/adapters/storage/LocalStorageBase.js
@@ -13,7 +13,8 @@ const messages = {
     notFound: 'File not found',
     notFoundWithRef: 'File not found: {file}',
     cannotRead: 'Could not read file: {file}',
-    invalidUrlParameter: `The URL "{url}" is not a valid URL for this site.`
+    invalidUrlParameter: `The URL "{url}" is not a valid URL for this site.`,
+    invalidPathParameter: 'The path "{path}" is not valid for this storage.'
 };
 
 class LocalStorageBase extends StorageBase {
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/adapters/storage/LocalStorageBase.js
```diff
@@ -228,10 +295,15 @@ class LocalStorageBase extends StorageBase {
...
         options = options || {};
 
-        // remove trailing slashes
-        options.path = (options.path || '').replace(/\/$|\\$/, '');
+        let normalizedPath;
+
+        try {
+            normalizedPath = this._normalizeStorageRelativePath(options.path);
+        } catch (err) {
+            return Promise.reject(err);
```
allouis: If we make this an `async` function we can remove the try/catch

ghost/core/core/server/adapters/storage/LocalStorageBase.js
```diff
@@ -131,23 +202,24 @@ class LocalStorageBase extends StorageBase {
...
-        // TODO: remove absolute path support once all callers pass relative paths
-        if (targetDir && !targetDir.startsWith(this.storagePath)) {
-            targetDir = path.join(this.storagePath, targetDir);
+        let filePath;
+
+        try {
+            filePath = this._resolveAndValidateStoragePath(targetDir, fileName);
+        } catch (err) {
+            if (err instanceof errors.IncorrectUsageError) {
+                return Promise.resolve(false);
```
allouis: Why do we want to return `false` here instead of erroring?

ghost/core/core/server/adapters/storage/LocalStorageBase.js
```diff
@@ -38,6 +39,80 @@ class LocalStorageBase extends StorageBase {
...
     }
 
+    /**
+     * Normalizes a relative storage path and rejects traversal outside the storage root.
+     *
+     * @param {string} filePath
+     * @returns {string}
+     */
+    _normalizeStorageRelativePath(filePath, options = {}) {
+        const {
```
allouis: I find this injection of which error message to throw kinda weird/confusing, I would prefer to see the callers handle this themselves.

We could have the contract so that it returns `string | null` and then callers can do `if (result === null) throw ...`

allouis (commented):
Overall improvements look good, but I think we can clean this up a little

ghost/core/core/server/adapters/storage/LocalStorageBase.js
kevinansfield: `exists()` is intended to return a boolean, if we detect path traversal outside of the storage base then by definition the file doesn't exist and `false` makes sense and means callers don't have to handle errors. The other errors types re-thrown on 221 are from genuinely catastrophic errors from the path validation so  re-throwing those makes sense.

allouis: Cool, makes sense!

ghost/core/core/server/adapters/storage/LocalStorageBase.js
kevinansfield: Returning `string | null` feels far to implicit to me, I like the explicit error on incorrect usage but I agree passing the error details into the function feels off 🤔

kevinansfield: Updated so the helper function always throws the default error, then for the single URL case the caller catches and re-throws its custom error.

allouis (approved):

PR #27232, opened 2026-04-08
🐛 Fixed members subscription filter returning wrong results for multi-sub members

rob-ghost:
When a member has multiple Stripe subscriptions (for example, they cancel and re-subscribe before the cancellation period ends), the admin UI displays only the "most relevant" subscription. However, the member list filters were evaluating against every subscription in the `members_stripe_customers_subscriptions` table via the `members_stripe_customers` join. This meant filtering by `subscriptions.status:active` could match a member whose displayed subscription was actually canceled, and vice versa.

This change introduces a `members_current_subscription` lookup table that stores a single row per member pointing at their resolved "best" subscription, and a `members_resolved_subscription` database VIEW that encapsulates the subscription priority logic. The VIEW is the single source of truth for resolution, used by the backfill migration, the runtime upsert in `linkSubscription`, and test utilities. The member model's `subscriptions` filter relation now joins through the lookup table so all `subscriptions.*` filters operate on the single resolved subscription per member.

## Resolution priority

0. **`incomplete` and `incomplete_expired` subscriptions are excluded entirely.** These represent subs whose first payment never completed (3DS abandoned, card declined, checkout abandoned mid-flow). Ghost does not consider the customer subscribed in those states — the admin API already strips them from `member.subscriptions` for the same reason. A member with only an `incomplete` sub has no resolved current sub.
1. **Active statuses beat inactive ones.** Active = `active`, `trialing`, `past_due`, `unpaid` — matches Ghost's existing convention used across `member-repository`, `router-controller`, etc.
2. **Most recent `created_at` wins.** Tiers in Ghost are discrete products rather than hierarchical, so price/MRR doesn't factor in — whichever subscription the member most recently signed up for or was moved to represents their current state, regardless of whether it's an "upgrade" or "downgrade". We use `created_at` rather than `start_date` because `start_date` can be in the future for scheduled subscriptions; `created_at` is monotonic, can never be in the future, and reflects when the action was actually taken.
3. **`id` ascending** as a stable final tiebreaker for the rare case where two subscriptions share the same `created_at`.

The admin API exposes the resolved subscription as `current_subscription` on member responses (always loaded — costs one cheap extra query through the tiny lookup table). The frontend members list reads this directly. A `mostRelevantSubscription` fallback is kept in `apps/posts` for forwards-compat during BE/FE deploy skew and can be removed in a follow-up once the field has rolled out everywhere.

## Why an association table

A separate association table was chosen over a FK column on `members` because of how the filter infrastructure works. The bookshelf filter plugin (`mongo-knex`) generates JOINs for `subscriptions.*` filters using a `manyToMany` config: `members` → join table → target table. Adding a FK column to `members` wouldn't fit this pattern without modifying the filter plugin itself.

## Migrations

Three versioned migrations handle setup for existing installations: one creates the `members_current_subscription` table, a second creates the `members_resolved_subscription` VIEW, and a third backfills the lookup table by selecting from the VIEW. This requires a two-step deploy -- the platform team pre-applies the migrations on production before the code change is merged, so the table is populated before filters start using it.

The `getTables()` helper in `commands.js` was also updated to use `SHOW FULL TABLES WHERE Table_type = 'BASE TABLE'` on MySQL, so that database VIEWs are excluded from table listings.

## Notes for reviewers / release

**Filter semantics change.** Saved NQL filters using `subscriptions.*` previously matched if *any* of a member's subscriptions met the criteria. After this change, they match against the resolved subscription only. For multi-sub members the result set may change. This was the intended fix for the original bug but is worth calling out in release notes.

**Audit of mutation paths.** Every backend path that mutates `members_stripe_customers_subscriptions` either routes through `linkSubscription` (which calls `_updateCurrentSubscription`) or destroys the member entirely. The latter is covered by `cascadeDelete: true` on the `member_id` FK in the lookup table, so the lookup row is removed alongside the member. No standalone reconciler is needed at this time.

**Comped members.** Members with `status='comped'` who don't have a Stripe subscription are absent from the resolved view and therefore from `subscriptions.*` filter results. This is the correct semantics — `subscriptions.*` is about Stripe subscriptions, and the `comped` status lives on the member. Covered by a test in `subscription-resolution.test.ts`.

## Local verification

A seed script populates a small set of members covering each resolution
scenario (active vs cancelled, upgrade, downgrade, future-start, comped, etc.):

```
pnpm dev                        # start the dev environment
pnpm reset:data:empty           # ensure tiers/products exist
pnpm seed:multi-sub-scenarios   # seed the verification members
```

Members appear under `multisub-*@verify.test`. The script prints each
scenario's expected resolution alongside the actual one so any mismatch is
visible immediately. Re-run safely — it deletes the existing scenario members
on each run.

fixes https://linear.app/ghost/issue/BER-3345


ghost/core/core/server/data/migrations/init/1-create-tables.js
```diff
@@ -14,6 +15,14 @@ module.exports.up = async (options) => {
...
         await commands.createTable(table, connection);
     }));
+
+    // Create database VIEWs (depends on tables existing first)
+    for (const [name, sql] of Object.entries(views)) {
+        logging.info('Creating view: ' + name);
+        await connection.schema.createViewOrReplace(name, function (view) {
+            view.as(connection.raw(sql));
+        });
+    }
```
kevinansfield: Nit: Would this work as a separate init file `3-create-views.js`?

rob-ghost: We'd have to move it up the chain because `2-create-fixtures.js` depends on the views being created.

So we could make it:

1. Create tables
2. Create views
3. Create fixtures

rob-ghost: Correction, fixtures doesn't actually depend on the views.

Tables -> Views -> Fixtures still makes sense though so I made that change.

ghost/core/core/server/data/schema/commands.js
```diff
@@ -524,8 +524,8 @@ async function getTables(transaction = db.knex) {
         const response = await transaction.raw('select * from sqlite_master where type = "table"');
         return _.reject(_.map(response, 'tbl_name'), name => name === 'sqlite_sequence');
     } else if (client === 'mysql2') {
-        const response = await transaction.raw('show tables');
-        return _.flatten(_.map(response[0], entry => _.values(entry)));
+        const response = await transaction.raw('show full tables where Table_type = \'BASE TABLE\'');
+        return _.map(response[0], entry => _.values(entry)[0]);
```
kevinansfield, line 528: Was this change fixing an issue or just a cleanup?

rob-ghost: When we add the `view` it shows up in the list of tables. Since its not a real table, it should be excluded.

This changes the logic from `show tables` to `show full tables` which doesn't include views.

The view isn't intended to be used in exports etc so we want it filtered out and we don't want to have to update this code + tests every time we add a new view we don't want included.

ghost/core/core/server/data/schema/views.js
```diff
@@ -0,0 +1,26 @@
...
+                ROW_NUMBER() OVER (
+                    PARTITION BY msc.member_id
+                    ORDER BY
+                        CASE WHEN mscs.status IN ('active', 'trialing', 'past_due', 'unpaid') THEN 0 ELSE 1 END,
+                        mscs.current_period_end DESC
+                ) as rn
+            FROM members_stripe_customers_subscriptions mscs
+            JOIN members_stripe_customers msc ON msc.customer_id = mscs.customer_id
+        ) ranked
+        WHERE rn = 1
```
rob-ghost, line 58: Since we're synthesising our "most relevant subscription" rule into the database here, we want to make sure that we're confident in the rules we're appling.

For example, how do we resolve concurrent monthly or yearly subscriptions? What about tiers?

What we ported for this was considered a quick resolution on the frontend side of things to decide what to show, but we might need deeper consideration for our priority order here.

rob-ghost: Based on use cases the view covered most of them well, the outstanding concern was tier resolution. Between a paid and comped subscription, what should win?

We chose whatever had the highest paid option as the ranking, so now Gold > Bronze when resolving the ranking of the subscription.

rob-ghost: Additional problem, now there's drift between this definition of "most relevant" / current subscription and the FE logic.

Which leads us down the path of having the backend be the source-of-truth for a members current subscription using this view, such that we don't get this drift again.

So we'll join and return the `current_subscription` via the API and fallback to the old logic until that gets returned for backwards compatability.

apps/posts/src/views/members/member-query-params.ts
```diff
@@ -89,6 +89,20 @@ export function buildMemberOperationParams({nql, search}: BuildMemberOperationPa
...
+ * Resolve the member's current subscription, preferring the field returned by
+ * the backend. Falls back to local resolution for older API responses that
+ * don't yet expose `current_subscription` (deploy skew between backend and
+ * admin assets). Once the field has been live everywhere for long enough,
+ * `mostRelevantSubscription` and this fallback can be removed.
+ */
+export function getCurrentSubscription(member: Member): MemberSubscription | null {
+    if (member.current_subscription !== undefined) {
+        return member.current_subscription;
+    }
```
rob-ghost, line 109: **Question:** Is an `undefined` check exhaustive enough? Can `current_subscription` also be `null` and is that meaningfully different or should we be checking for any value vs no value here?

ghost/core/core/server/models/member.js
```diff
@@ -255,6 +255,17 @@ const Member = ghostBookshelf.Model.extend({
...
     },
 
+    currentSubscription() {
+        return this.belongsToMany(
+            'StripeCustomerSubscription',
+            'members_current_subscription',
+            'member_id',
+            'subscription_id',
+            'id',
+            'id'
```
rob-ghost, line 279: **Question:** Its unclear what these two `id` parameters refer to, is there any way to make it more obvious what tables / columns these are referring to?

ghost/core/core/server/models/member.js
```diff
@@ -279,6 +290,15 @@ const Member = ghostBookshelf.Model.extend({
...
         }
 
+        // currentSubscription is a 1:1 relation through members_current_subscription
+        // — flatten the bookshelf collection to a single object (or null).
+        if (defaultSerializedObject.currentSubscription !== undefined) {
+            defaultSerializedObject.current_subscription = Array.isArray(defaultSerializedObject.currentSubscription)
+                ? (defaultSerializedObject.currentSubscription[0] || null)
+                : defaultSerializedObject.currentSubscription;
+            delete defaultSerializedObject.currentSubscription;
+        }
```
rob-ghost: **Question:** Why is it a collection by default? Can we not hint that this will be a single value in the query or are all joins always a collection and we need to pull the first element from?

ghost/core/core/server/services/members/members-api/repositories/member-repository.js
```diff
@@ -1540,6 +1538,39 @@ module.exports = class MemberRepository {
...
+                // No subscriptions remain — remove the lookup row
+                await trx('members_current_subscription')
+                    .where({member_id: memberId})
+                    .del();
+            }
+        } catch (e) {
+            logging.error(`Failed to update members_current_subscription for member ${memberId}`);
+            logging.error(e);
+            throw e;
+        }
```
rob-ghost, line 1647: **Question:** Should this error logging follow our proposal for [structured error logging](https://www.notion.so/ghost/Proposal-Structured-Logging-Standardization-33551439c03081338f6ae0154b40e74f?v=68c64d14c3024017ba0648822c8e6bb1&source=copy_link)?

ghost/core/test/e2e-api/admin/members.test.js
```diff
@@ -716,6 +718,67 @@ describe('Members API', function () {
...
+            .get(`/members/?filter=subscriptions.status:canceled`)
+            .expectStatus(200);
+
+        const canceledEmails = canceledBody.members.map(m => m.email);
+        assert.ok(!canceledEmails.includes(email), 'Member with active+canceled subs should NOT appear in subscriptions.status:canceled filter');
+
+        // Clean up: remove the test member so subsequent tests aren't affected
+        await agent
+            .delete(`/members/${memberId}/`)
+            .expectStatus(204);
```
rob-ghost, line 805: **Suggestion:** Tests shouldn't be concerned about cleaning up after themselves. Why does this break other tests, and what changes can we make to our test structure so our tests can't be polluted requiring cleanup?

ghost/core/test/unit/server/services/members/members-api/repositories/member-repository.test.js
```diff
@@ -247,6 +247,7 @@ describe('MemberRepository', function () {
         });
 
         beforeEach(async function () {
+            sinon.stub(MemberRepository.prototype, '_updateCurrentSubscription').resolves();
```
rob-ghost, line 512: **Question:** Why stub this out? We should probably fix whatever this breaks in the tests so we don't have to concern ourselves with stubs.

ghost/core/core/server/data/schema/views.js
```diff
@@ -0,0 +1,60 @@
...
+//      subscription the member most recently signed up for or was moved to
+//      is the one that represents their current state, regardless of whether
+//      that's an "upgrade" or "downgrade".
+//
+//      We use `created_at` rather than `start_date` because `start_date` can
+//      be in the future for scheduled subscriptions — a sub created earlier
+//      but scheduled to start later would otherwise out-rank a more recently
+//      created sub that's already in effect. `created_at` is monotonic, can
+//      never be in the future, and reflects "when this action happened" from
+//      Ghost's perspective.
```
kevinansfield, line 33: I expect very edge-case but this feels a little leaky. `created_at` is currently set to the local date at the time the bookshelf record is created, so subscriptions created via imports or out-of-order webhook requests could resolve as most-relevant despite being created earlier than the most-recently-created subscription.

kevinansfield (approved):

ghost/core/core/server/data/schema/views.js
rob-ghost: Yeah, Stripe subscriptions do have a `created` but we're not using that, instead using `created_at: Date.now()` when creating the Bookshelf model.

One thing to note: we don't import Subscriptions, and if we did and ended up with two concurrent active subscriptions which required the `created_at` tiebreak I think we're in bigger trouble!

But you're right to callout that its not bulletproof, I just don't think the alternatives are any better unfortunately.

PR #27236, opened 2026-04-08
🐛 Fixed member edited webhook missing tiers

9larsons:
ref https://linear.app/ghost/issue/NY-1216/

## Summary

The `member.edited` webhook payload was missing `tiers` because the webhook serializer used the raw Bookshelf model from the event, which only had `labels` loaded. The Admin API response was correct (it does a fresh `read()` with all relations), but the webhook was not.

The webhook serializer now loads `products`, `labels`, and `newsletters` on the member model before serializing, so `tiers` is included in the payload.

Also fixed the stripe-mocker to return the `type` field on price objects, matching real Stripe API behavior (needed for the comp flow to work in tests).

### Root cause

The `member.edited` event fires with the Bookshelf model from `memberRepository.update()`, which only has `labels` in its `withRelated`. The `serializeMember` function only sets `tiers` when `json.products` exists on the model — and it didn't, because `products` was never loaded.

### Payload: before → after

**Before** (comped member):
```json
{
  "member": {
    "current": {
      "id": "...",
      "uuid": "...",
      "email": "comped-test@example.com",
      "name": "Comped Test Member",
      "status": "comped",
      "comped": true,
      "subscribed": true,
      "subscriptions": [],
      "labels": [],
      "avatar_image": null,
      "email_count": 0,
      "email_opened_count": 0,
      "email_open_rate": null,
      "last_seen_at": null,
      "created_at": "...",
      "updated_at": "..."
    },
    "previous": {
      "status": "free",
      "updated_at": "..."
    }
  }
}
```

**After** (comped member):
```json
{
  "member": {
    "current": {
      "id": "...",
      "uuid": "...",
      "email": "comped-test@example.com",
      "name": "Comped Test Member",
      "status": "comped",
      "comped": true,
      "subscribed": true,
      "subscriptions": [],
      "tiers": [
        {
          "id": "...",
          "name": "Default Product",
          "slug": "default-product",
          "active": true,
          "type": "paid",
          "currency": "usd",
          "monthly_price": 500,
          "yearly_price": 5000
        }
      ],
      "labels": [],
      "newsletters": [...],
      "avatar_image": null,
      "email_count": 0,
      "email_opened_count": 0,
      "email_open_rate": null,
      "last_seen_at": null,
      "created_at": "...",
      "updated_at": "..."
    },
    "previous": {
      "status": "free",
      "updated_at": "..."
    }
  }
}
```

Changes:
- **Added**: `tiers` — the Ghost products/tiers the member has access to (source of truth for member access)
- **Added**: `newsletters` — the newsletters the member is subscribed to
- **Unchanged**: `subscriptions` remains `[]` (Stripe relations not loaded; documenting/fixing this separately)

## Test plan

- [x] New e2e test: creates a free member, comps them via Admin API, verifies the webhook includes `tiers` with correct shape (id, name, slug)
- [x] Existing member webhook tests (added, deleted, edited) pass with updated snapshots


ghost/core/test/e2e-webhooks/members.test.js
```diff
@@ -159,4 +160,43 @@ describe('member.* events', function () {
                 }
             });
     });
+
+    it('member.edited event includes tiers when comped', async function () {
+        mockManager.mockStripe();
+
+        const webhookURL = 'https://test-webhook-receiver.com/member-comped/';
```
EvanHahn: nit: I think this is fine but I like to use domains that could never be registered, just in case.

```suggestion
        const webhookURL = 'https://test-webhook-receiver.example/member-comped/';
```

ghost/core/test/e2e-webhooks/members.test.js
```diff
@@ -159,4 +160,43 @@ describe('member.* events', function () {
...
+                }]
+            })
+            .expectStatus(200);
+
+        await webhookMockReceiver.receivedRequest();
+
+        const webhookPayload = webhookMockReceiver.body.body;
+        const current = webhookPayload.member.current;
+
+        assert.ok(current.tiers.length > 0, 'Webhook should include tiers');
```
EvanHahn: nit: could be worth testing what `current.tiers` actually is.

ghost/core/test/utils/stripe-mocker.js
```diff
@@ -297,6 +297,17 @@ class StripeMocker {
             }
         }
 
+        if (resource === 'prices') {
+            if (!id) {
```
EvanHahn: nit: this conditional can be combined.

```js
if ((resource === 'prices') && !id) {
    // ...
```

ghost/core/core/server/services/webhooks/serialize.js
```diff
@@ -26,12 +31,22 @@ module.exports = (event, model) => {
...
             return apiFramework
                 .serializers
                 .handle
                 .output(model, {docName: docName, method: 'read'}, api.serializers.output, frame)
                 .then(() => {
-                    return frame.response[docName][0];
+                    const result = frame.response[docName][0];
+
+                    if (docName === 'members') {
+                        delete result.subscriptions;
```
EvanHahn: Two thoughts about this:

1. Are we sure this breaking change is safe?
1. In general, I'm against mutating values like this. I'd instead prefer to use `_.omit(result, ['subscriptions'])` or something like it. Probably fine in this case but wanted to flag.

ghost/core/core/server/services/webhooks/serialize.js
```diff
@@ -26,12 +31,22 @@ module.exports = (event, model) => {
                 };
             }
 
+            if (docName === 'members') {
```
EvanHahn: nit: I think this is fine but could be more clear. I'd consider putting this in a `switch` with the `posts` and `pages` checks above. Something like this:

```js
switch (docName) {
case 'posts':
case 'pages':
    // ...
    break;
case 'members':
    // ...
    await model.load(MEMBER_WITH_RELATED);
    break;
default:
    break;
}
```

Feel free to ignore.

EvanHahn (approved):
Several non-blocking comments. The `subscriptions` removal is my biggest question.

ghost/core/test/e2e-webhooks/members.test.js
9larsons: Done — switched to .example per RFC 2606.

ghost/core/test/e2e-webhooks/members.test.js
9larsons: Added assertions for id, name, and slug on the tier object, and that exactly one tier is present.

ghost/core/test/utils/stripe-mocker.js
9larsons: Done — combined into a single conditional.

ghost/core/core/server/services/webhooks/serialize.js
9larsons: Agreed on both points. Restored subscriptions — removing it is a separate discussion and shouldn't be bundled with this fix.

ghost/core/core/server/services/webhooks/serialize.js
9larsons: Adopted the switch — applied it to both the current and previous serialization blocks for consistency.

PR #27238, opened 2026-04-08
Fixed welcome email body font after enabling design customization

cmraible:
closes https://linear.app/ghost/issue/NY-1209/font-changes-for-existing-welcome-emails-when-customization-is-enabled

Enabling the `welcomeEmailsDesignCustomization` labs flag switched welcome emails onto the new template path, but the renderer always used the serif content class instead of honoring the stored body font setting. This keeps existing welcome emails on the default sans-serif styling until someone explicitly chooses a different body font in the customize flow.


cmraible:
@coderabbit review

troyciesco (approved):

PR #27253, opened 2026-04-08
Further highlight utility functions in database migration skill

EvanHahn:
ref 25e1433a622ca46484bfc26d0e694c90f7321e53

I found that my agent wouldn't use these utility functions by default. I tweaked the instructions to better highlight that.


cmraible (approved):
Love this!

PR #27293, opened 2026-04-09
Improve performance of core acceptance/legacy CI jobs

acburdine:
## Summary

- remove nx cache step + unnecessary env var from setup job that hasn't done anything since the Nx v22 update
- remove explicit TS package build step in favor of nx automatic build tracking

## Why this change

As part of #27202, I was looking into ways to leverage the Nx build cache across jobs in a given CI run, since I noticed that certain packages (the frontend code in apps/* primarily) were being built multiple times throughout the various CI jobs. What I ultimately realized after a bit more investigation is that the Core Acceptance + Legacy tests jobs are building more of the repo than they need to, since the `--exclude=ghost-admin` doesn't actually prevent any work from being done (ghost-admin is a dependency of `@tryghost/admin` so it gets built anyways).

## What the fix does

By updating the nx config in ghost/core to dependent on the built target for the test:ci:* scripts (and switching the jobs themselves to use `yarn nx run ghost:test:ci:*`), Nx will automatically build the dependencies it needs to. Overall, this should shave an average of 3-4 minutes off of all of the Acceptance/Legacy test builds.

---

> [!NOTE]
> **Low Risk**
> Low risk CI-only change, but could cause unexpected rebuilds or missing prereqs if Nx target dependencies are misconfigured, affecting test job reliability/timings.
> 
> **Overview**
> **CI pipeline is simplified to reduce unnecessary work.** The setup job drops the unused `NX_REJECT_UNKNOWN_LOCAL_CACHE` env var and removes the GitHub Actions `Nx cache` step.
> 
> **Acceptance and legacy test jobs now rely on Nx dependency tracking.** The explicit “build TS packages” step is removed, and Ghost core `test:ci:*` scripts are executed via `yarn nx run ghost:test:ci:*`, with `ghost/core/package.json` updated so `build:tsc`, `lint`, `test:unit`, and the `test:ci:*` targets depend on `^build` instead of a hardcoded `@tryghost/parse-email-address` build.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 130791d8797b9d04e26a7705275a4c9033e13835. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -438,9 +438,6 @@ jobs:
         env:
           DEPENDENCY_CACHE_KEY: ${{ needs.job_setup.outputs.dependency_cache_key }}
 
-      - name: Build TS packages
-        run: yarn nx run-many -t build --exclude=ghost-admin
-
       - name: Set timezone (non-UTC)
         uses: szenius/set-timezone@1f9716b0f7120e344f0c62bb7b1ee98819aefd42 # v2.0
         with:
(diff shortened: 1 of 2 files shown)


ErisDS:
@acburdine Is there any reason why we should not just merge this? Seems great to me!

acburdine:
@ErisDS Nope, feel free to merge!

PR #27294, opened 2026-04-09
Added gift member status filter in admin

mike182uk:
ref https://linear.app/ghost/issue/BER-3481

Added gift member status filter in admin to allow filtering members who have a gift subscription

### DIFF
--- a/apps/posts/src/views/members/components/members-filters.tsx
+++ b/apps/posts/src/views/members/components/members-filters.tsx
@@ -11,6 +11,7 @@ import {
 } from '../use-member-filter-fields';
 import {getSettingValue, useBrowseSettings} from '@tryghost/admin-x-framework/api/settings';
 import {getSiteTimezone} from '@src/utils/get-site-timezone';
+import {useBrowseConfig} from '@tryghost/admin-x-framework/api/config';
 import {useBrowseNewsletters} from '@tryghost/admin-x-framework/api/newsletters';
 import {useBrowseOffers} from '@tryghost/admin-x-framework/api/offers';
 import {useBrowseTiers} from '@tryghost/admin-x-framework/api/tiers';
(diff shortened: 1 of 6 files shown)


apps/posts/src/views/members/components/members-filters.tsx
```diff
@@ -67,6 +69,7 @@ const MembersFilters: React.FC<MembersFiltersProps> = ({
     const emailTrackOpens = getSettingValue<boolean>(settings, 'email_track_opens') === true;
     const emailTrackClicks = getSettingValue<boolean>(settings, 'email_track_clicks') === true;
     const siteTimezone = getSiteTimezone(settings);
+    const giftSubscriptionsEnabled = configData?.config?.labs?.giftSubscriptions === true;
```
rob-ghost, line 72: FYI there's a hook for checking labs flags: `useLabsFlag` which abstracts away the config concerns.

rob-ghost: Actually, that doesn't seem to be used. There's a `useLabs` hook which uses the app context and I _think_ its a record of `string: bool` for labs flags but I'm not 100%

sagzy (approved):

apps/posts/src/views/members/components/members-filters.tsx
mike182uk: Hm, the only `useLabs` I'm seeing is in the `comments-ui` app

It looks like each app is sourcing this data in a different way:

- `comments-ui/useLabs`
- `stats/useLabsFlag`
- `admin-x-settings/useFeatureFlag`
- `posts/useFeatureFlag`

and each one seems to have different needs / contexts

we can't use `posts/useFeatureFlag` here even though we are in the posts app because its tied to the post analytics provider for some reason, and looks to be more used for routing (returns a `redirect` component)

agent followed similar approach to [this](https://github.com/TryGhost/Ghost/blob/c6f98c7cc15b5104109fce570f10022eda2f79c2/apps/posts/src/views/members/members.tsx#L36) thats already in the `posts` app

rob-ghost: You're totally right I keep getting mixed up with which hooks are in which app. I ran into the exact same thing before 😅 and I'm pretty sure I used the same method as you!

Nevermind!

PR #27297, opened 2026-04-09
Improve Nx base commit detection, remove unnecessary CI steps

acburdine:
## Summary

- add nrwl/nx-set-shas github action to set HEAD/BASE commit SHAs on main branch CI runs. 
- simplify unit test job by adding build as a dependent target for test:unit scripts.

## Why this change

Another split-out from #27202, this simplifies the CI jobs running that use `nx affected` to conditionally run targets on changed packages.

First, it adds the nrwl/nx-set-shas action for determining the BASE_COMMIT on main branch CI runs. The advantage of the set-shas action (on the main branch at least) is that it doesn't look at the commits merged as part of a particular CI run, it iterates through the CI action runs to find the last commit the CI workflow was successful on. This ensures that if the build fails for any reason, we won't inadvertently skip any lint/unit test runs (whereas the current implementation would end up skipping those checks). The set-shas action would work on pull-requests too, except we're only doing a shallow clone in the setup job and the set-shas action requires a deeper clone to be able to inspect the commit tree on PRs, so I left the pr base commit detection as-is.

Additionally, this PR simplifies the unit test job by adding build as a dependent target for test:unit scripts. This way, build jobs and unit test jobs can all run in parallel in the same Nx command rather than being split up over multiple ones.

## Potential Drawbacks

- Adding build as a dependency for `test:unit` means that for local devs, running test:unit with nx will end up needing `build` to be run as well. Operating under the assumption that build needs to be run before unit tests anyways, this shouldn't be much of an issue. If unit tests _don't_ require the build step to be run, then this PR could be simplified even further.
- If the main branch CI workflow fails for a reason unrelated to unit/lint jobs, the next run of CI will potentially run lint/test on packages it doesn't need to. That said, future PRs may expand the use of `nx affected` to cover additional types of tests, so having the base commit be the one with the last successful CI run seems prudent to cover future use-cases.

---

> [!NOTE]
> **Medium Risk**
> Changes CI SHA detection and how `nx affected` is parameterized, which can alter which projects get linted/tested (risk of unexpected skips or extra work). Also makes `test:unit` always run after `build`, affecting both CI and local workflows.
> 
> **Overview**
> Improves CI reliability for `nx affected` by switching push runs to `nrwl/nx-set-shas` and standardizing on `NX_BASE`/`NX_HEAD` env vars (including exposing `nx_base` from `job_setup`).
> 
> Simplifies lint/unit-test jobs by removing explicit `--base` arguments and (for unit tests) dropping the pre-step that computed affected projects + built them.
> 
> Updates `nx.json` so `test:unit` **depends on** `build`, ensuring required builds happen automatically when running affected unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b8f430a4008d541f6bf3dd89d5446a8468fb7088. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,14 +45,16 @@ jobs:
       IS_SIX: ${{ github.ref == 'refs/heads/6.x' }}
       IS_SIX_PR: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.ref == '6.x' }}
     permissions:
+      actions: read
       contents: read
-      pull-requests: read
     steps:
       - name: Checkout current commit
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           ref: ${{ env.HEAD_COMMIT }}
-          fetch-depth: 2
+          fetch-depth: 0
+          # fetch a treeless clone to improve checkout speed, the job will fetch contents later if needed
+          filter: 'tree:0'
 
       - name: Output GitHub context
         if: env.RUNNER_DEBUG == '1'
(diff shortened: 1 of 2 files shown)


acburdine:
note: now that #27295 is merged this change is even more important; main branch CI runs have the potential to be canceled which could potentially caused lint/unit test runs to be skipped since each main branch CI run only looks at the commits that were a part of its own push

acburdine:
sonarqube failure is a red herring I believe, it's flagging code that was already there.

(might make sense to refactor that part of the job at some point but it's not a _new_ issue)

9larsons (approved):
Looks great! If you'd like to move the setup job to a deeper clone, `depth: 1000` would only take ~7MB more which wouldn't appreciably slow us down. I was just looking at this a couple days ago and was shocked at the full depth being close to 400MB.

acburdine:
@9larsons tried 500 which should definitely have been enough, but I think the way the nx-set-shas action works for Pull Requests, it requires a full git tree in order to find the merge-base.

The way it's working now (push events fetch from last successful run, prs from API) _should_ be ok for now - the paths-filter actions actually end up fetching all the necessary commits between HEAD and BASE to do the diff so it ends up resolving whatever commits the job needs

acburdine:
Ok, finally got everything working as intended I think:

switched to using `fetch-depth: 0` with `filter: tree:0` - this gets us a treeless clone that has enough history for the nx shas bit to work, while still only taking ~8-10s to clone

.github/workflows/ci.yml
```diff
@@ -194,8 +190,14 @@ jobs:
       - name: Install dependencies
         run: bash .github/scripts/install-deps.sh
 
+      - name: Determine Affected Projects
```
9larsons, line 193: This is no longer used, right?

9larsons (approved):
Nice changes! Aside from cleaning up `affected_projects` it's good to merge.

.github/workflows/ci.yml
acburdine: Actually this is prep for future PRs 😅 added it here to verify Nx affected worked with the treeless clone

I removed the affected detection in the unit test job, the idea in the setup job is to use affected to determine whether certain jobs need to run at all

PR #27313, opened 2026-04-09
✨ Added design customization options for welcome emails

cmraible:
The functionality is already live on main behind the labs flag; this change just promotes the flag to GA so welcome emails customization will be enabled by default.


cmraible:
Reverting to draft, as we've decided to wait on this until Monday

troyciesco (approved):

PR #27314, opened 2026-04-09
Moved WORKDIR to /var/lib/ghost and added community Docker target

rob-ghost:
refs [BER-3398](https://linear.app/ghost/issue/BER-3398), [BER-3399](https://linear.app/ghost/issue/BER-3399)

The community Docker image (`docker-library/ghost`) is maintained by external contributors, and updates consistently lag behind Ghost releases because they depend on community PRs through `docker-library/official-images`. We want to publish our own `tryghost/ghost` image to Docker Hub that's a zero-change migration for the tens of thousands of self-hosters currently using the community image.

To make that work, our production Dockerfile needs to match the conventions the community image established — primarily `WORKDIR /var/lib/ghost` (so existing volume mounts and compose files just work) and content seeding on first boot (so a fresh named volume gets default themes and scaffolding). This PR makes those alignment changes and adds a `community` build target that layers the entrypoint, config, and VOLUME declaration on top of the existing `full` target.

The WORKDIR move from `/home/ghost` to `/var/lib/ghost` also cleaned up some dead code from the legacy Pro source-build era — unused `default/` directory creation and theme copies that nothing references. The `log/` directory creation was moved to `Dockerfile.pro` where it belongs since Ghost itself writes logs to `content/logs/`, not a top-level `log/` dir.

The community target is not built or published in CI yet — that comes with [BER-3400](https://linear.app/ghost/issue/BER-3400) when we're ready to push to Docker Hub. For now, it can be built locally with `docker build --target community`.

Merge [Ghost-Moya#210](https://github.com/TryGhost/Ghost-Moya/pull/210) after this lands to align the Pro container and PR preview tooling with the new WORKDIR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -883,6 +883,11 @@ jobs:
       - name: Prepare Docker build context
         run: mv ghost/core/package/ /tmp/ghost-production/
 
+      - name: Add community Docker files to build context
+        run: |
+          cp docker/community/docker-entrypoint.sh /tmp/ghost-production/
+          cp docker/community/config.production.json /tmp/ghost-production/
+
       - name: Determine push strategy
         id: strategy
         run: |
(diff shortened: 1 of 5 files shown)


rob-ghost:
Closing as this project has been paused.

PR #27315, opened 2026-04-09
Fixed welcome email customization closing without a save warning

cmraible:
closes https://linear.app/ghost/issue/NY-1218/

## What this changes

- fixed the welcome email customization modal so `Escape` no longer bypasses the dirty-state flow
- kept Escape handling local to the welcome email UI instead of changing shared Shade overlay behavior globally
- fixed nested color picker popovers so the first `Escape` closes the picker without also opening the unsaved-changes dialog
- fixed nested font select menus so the first `Escape` closes the menu without also opening the unsaved-changes dialog
- kept the confirmation dialog scoped so its own `Escape` handling does not cascade back into the parent modal

## Why

The original bug was not limited to the top-level customize modal. The same Escape propagation issue also showed up in nested portalled controls inside the modal, which could still skip straight to the save warning flow. This PR now covers the full interaction stack so `Escape` always closes the topmost UI first.

## Testing

- added E2E coverage for the customize modal dirty-state Escape flow
- added E2E coverage for the color picker Escape flow
- added E2E coverage for the font select Escape flow
- ran `PLAYWRIGHT_HTML_OPEN=never yarn workspace @tryghost/e2e test e2e/tests/admin/settings/member-welcome-emails.test.ts`
- ran `PLAYWRIGHT_HTML_OPEN=never yarn workspace @tryghost/e2e test:single "Escape closes welcome email"`

### DIFF
--- a/apps/admin-x-settings/src/components/settings/email-design/design-fields/body-font-field.tsx
+++ b/apps/admin-x-settings/src/components/settings/email-design/design-fields/body-font-field.tsx
@@ -11,7 +11,11 @@ export const BodyFontField = () => {
                 <SelectTrigger className="w-[180px]">
                     <SelectValue />
                 </SelectTrigger>
-                <SelectContent>
+                <SelectContent
+                    onEscapeKeyDown={(event) => {
+                        event.stopPropagation();
+                    }}
+                >
                     {FONT_OPTIONS.map(opt => (
                         <SelectItem key={opt.value} value={opt.value}>{opt.label}</SelectItem>
                     ))}
--- a/apps/admin-x-settings/src/components/settings/email-design/color-picker-field.tsx
+++ b/apps/admin-x-settings/src/components/settings/email-design/color-picker-field.tsx
@@ -114,7 +114,13 @@ const ColorPickerField: React.FC<ColorPickerFieldProps> = ({title, value, onChan
                     </div>
                 </div>
             </PopoverTrigger>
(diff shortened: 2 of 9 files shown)


apps/admin-x-settings/src/components/settings/email-design/design-fields/body-font-field.tsx
```diff
@@ -11,7 +11,11 @@ export const BodyFontField = () => {
                 <SelectTrigger className="w-[180px]">
                     <SelectValue />
                 </SelectTrigger>
-                <SelectContent>
+                <SelectContent
+                    onEscapeKeyDown={(event) => {
+                        event.stopPropagation();
+                    }}
```
cmraible, line 17: This feels kind of fragile, since we're fixing this in the consuming component. I wonder if this should be the default behavior in Shade so we don't have to add this everywhere these components are used, but I wasn't sure if there might be unintended consequences of doing that 🤔

troyciesco: i was gonna agree but i looked a bit and i think maybe this is the right approach?

because even though it's not a native `select` element the [radix primitive this is built on ](https://www.radix-ui.com/primitives/docs/components/select) still has the keyboard functionality, so it's not like the underlying component isn't handling it at _all_.

it looks like radix also did it this way on purpose: https://github.com/radix-ui/primitives/pull/2761

troyciesco (approved):

PR #27318, opened 2026-04-09
Updated CI to mirror playwright image to GHCR

cmraible:
## Summary

This updates E2E CI to mirror the pinned Playwright container image into GHCR and pull that mirrored image during test runs.

The change was motivated by intermittent CI failures when pulling directly from Microsoft Container Registry:

```text
Error response from daemon: pull access denied for mcr.microsoft.com/playwright, repository does not exist or may require 'docker login': denied: ... <title>Microsoft</title> ... The request is blocked.
```

To reduce that flakiness, CI now:
- computes the pinned Playwright version in `job_setup`
- conditionally runs a dedicated mirror job only when the Playwright version changes
- mirrors `mcr.microsoft.com/playwright:v<version>-noble` to `ghcr.io/tryghost/playwright:v<version>-noble`
- cleans up older mirrored Playwright versions in GHCR with package-specific retention rules


cmraible:
These errors seem to have been transient, so I'm going to leave this here for now, rather than spend the time and effort to validate that it's doing everything properly.

PR #27341, opened 2026-04-10
Improve performance of core acceptance/legacy CI jobs

acburdine:
no issue
This removes the explicit TS package build step in favor of nx automatic build triggers. Because of the way this command was invoked, all of the frontend packages were being built on every CI run of acceptance/legacy tests, which adds about 3-4 minutes per job. The only package ghost/core depends on is parse-email-address, and so by leveraging Nx's automatic dependent target running, we can ensure that only the necessary packages are built. Leveraging Nx now also means that if more lecal packages with build targets are added as ghost/core deps in the future, the jobs will continue to work as designed.

This is a repeat of #27293 which was partially reverted during the pnpm cutover)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -452,9 +452,6 @@ jobs:
         env:
           DEPENDENCY_CACHE_KEY: ${{ needs.job_setup.outputs.dependency_cache_key }}
 
-      - name: Build TS packages
-        run: pnpm nx run-many -t build --exclude=ghost-admin
-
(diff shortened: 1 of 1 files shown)


9larsons:
LGTM! Thanks for redoing this. I'm going to patch the Tinybird Tests requirement to not run on forks and then we can rebase this and get it merged.

9larsons (approved):
Notes in comment above.

9larsons:
https://github.com/TryGhost/Ghost/pull/27344

PR #27345, opened 2026-04-11
Replaced monobundle with pnpm deploy

9larsons:
ref https://linear.app/ghost/issue/PLA-27

## Summary
- Replace the custom `monobundle.js` + `npm pack` workflow with pnpm's native `deploy` command
- Add `ghost/core/scripts/pack.js` to post-process the deployed package for Ghost-CLI compatibility
- Replace the old `.npmignore` inversion rules with a `files` allowlist in `ghost/core/package.json`
- Remove `monobundle.js`, `prepack`/`postpack`, `pack:standalone`, and `pack:tarball`

## What pack.js still handles
1. Strip pnpm peer suffixes from dependency versions in the deployed `package.json` ([pnpm#6269](https://github.com/pnpm/pnpm/issues/6269))
2. Pack private workspace packages (`@tryghost/i18n`, `@tryghost/parse-email-address`) into `components/*.tgz` for Ghost-CLI installs
3. Copy the root `packageManager` into the deployed `package.json` so Corepack and release validation use the pinned pnpm version
4. Merge root `pnpm.overrides` into the deployed `package.json` so standalone installs resolve the same transitive versions as the workspace
5. Copy `.npmrc` and `pnpm-workspace.yaml`, then build the final `ghost-<version>.tgz`

## CI / validation
- [x] `pnpm archive`
- [x] Extracted tarball + `CI=true pnpm install --ignore-scripts --prod --prefer-offline`
- [x] `ghost install local --archive ...` followed by HTTP 200 from the running site
- [x] `ghost update --archive ...` followed by HTTP 200 from the running site
- [x] CI now keeps the `packageManager` tarball check and uses Ghost-CLI install/update smoke tests rather than broader tarball-internal assertions


ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,84 @@
...
+
+fs.writeFileSync(pkgPath, JSON.stringify(pkg, null, 2) + '\n');
+
+// 3. Create tarball
+const version = pkg.version;
+const tgzPath = path.join(CORE_DIR, `ghost-${version}.tgz`);
+
+console.log(`\nCreating tarball: ghost-${version}.tgz`);
+execSync(
+    `tar czf ${tgzPath} --exclude='./node_modules' -s '/^\\./package/' .`,
```
acburdine: @9larsons if it'd be easier from a BSD/GNU standpoint to not need the path transformation logic, I can probably update the CLI to work with or without the subfolder 🤔

9larsons: Sounds good - I'm just exploring atm. I didn't realize we had some 'cross requires' in the `admin` app w/ `posts` and `stats` that's causing trouble so I'm going to try to unpick that first.

.github/workflows/ci.yml
```diff
@@ -783,7 +787,7 @@ jobs:
 
       - run: node .github/scripts/bump-version.js canary
 
-      - run: pnpm archive
+      - run: pnpm --filter ghost archive
```
acburdine: this might need to be `pnpm nx run ghost:archive` to trigger all the dependent builds

or just keep it as `pnpm archive` since that's what the root level `archive` script ends up calling 😅

acburdine: `pnpm --filter ghost archive` works on the "build artifacts" job since it runs `build:production` as a prior step (tbh, the build artifacts job could probably be done the same way as this job where it run `pnpm archive` instead of separate build:production and archive calls)

ghost/core/package.json
```diff
@@ -23,10 +23,27 @@
     "access": "public",
     "registry": "https://registry.npmjs.org/"
   },
+  "files": [
```
EvanHahn, line 26: praise: I much prefer `files` to `npmignore`—nice.

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+
+const CORE_DIR = path.resolve(__dirname, '..');
+const ROOT_DIR = path.resolve(CORE_DIR, '../..');
+const DEPLOY_DIR = path.join(CORE_DIR, 'package');
+
+// 1. Run pnpm deploy
+// inject-workspace-packages is enabled only for deploy (not workspace-wide)
+// because it conflicts with packages that have build outputs in their files field.
+console.log('Running pnpm deploy...');
+execSync(`rm -rf ${DEPLOY_DIR}`, {stdio: 'inherit'});
```
EvanHahn, line 30: nits:

- If `DEPLOY_DIR` is weird in any way, this could cause problems. If it has a space, it won't be removed. It could also theoretically cause other issues.
- I believe this doesn't work on Windows.

[`fs.rmSync`](https://nodejs.org/api/fs.html#fsrmsyncpath-options) seems like a better option.

```suggestion
fs.rmSync(DEPLOY_DIR, {recursive: true, force: true});
```

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+execSync(`pnpm --filter ghost deploy ${DEPLOY_DIR} --prod --config.inject-workspace-packages=true`, {
```
EvanHahn, line 31: nit: like my comment above, a weird `DEPLOY_DIR` could cause problems here; spaces in the directory, for example. It's less ergonomic but `spawnSync` might be a safer alternative.

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+    cwd: ROOT_DIR,
+    stdio: 'inherit'
+});
+
+// 2. Fix package.json
+console.log('\nPost-processing package.json...');
+const pkgPath = path.join(DEPLOY_DIR, 'package.json');
+const pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
```
EvanHahn, line 39: nit: consider using [`fs-extra`'s `readJsonSync`][0], which we have installed. This comment applies here and in a few other places in this file.

[0]: https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/readJson-sync.md

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    const depDir = path.join(DEPLOY_DIR, 'node_modules', key);
+    if (!fs.existsSync(depDir)) {
+        console.warn(`  Warning: ${key} not found in node_modules, skipping`);
+        continue;
+    }
+
+    const depPkg = JSON.parse(fs.readFileSync(path.join(depDir, 'package.json'), 'utf8'));
+    const slug = key.replaceAll('@', '').replaceAll('/', '-');
```
EvanHahn, line 73: nit: This normalizes `@foo/bar` and `foobar` the same. This is unlikely to be an issue, but there's are more robust normalization strategies that sidestep this problem with ~3 additional lines.

I like [the one DefinitelyTyped uses][0], which converts `@foo/bar` to `foo__bar`, but I mostly care that we come up with something where collisions won't happen. You may wish to see [npm's rules for package names][1].

Could also be done in a follow-up PR. 

[0]: https://github.com/microsoft/DefinitelyTyped-tools/blob/8d0404adb08aaaba2ad6a9b0521ee18f59ce1e4e/packages/dts-gen/src/names.ts
[1]: https://github.com/npm/validate-npm-package-name?tab=readme-ov-file#naming-rules

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+    const tgzName = `${slug}-${depPkg.version}.tgz`;
+
+    console.log(`  Packing ${key} → components/${tgzName}`);
+    execSync(`npm pack --pack-destination ${componentsDir}`, {cwd: depDir, stdio: 'pipe'});
```
EvanHahn, line 77: nit: like I've mentioned elsewhere, if `componentsDir` has a space in it (or some other flag, like `--dry-run`), this could break. I recommend using `spawnSync` or some alternative that avoids this problem.

ghost/core/package.json
```diff
@@ -23,10 +23,27 @@
...
+    "index.js",
+    "ghost.js",
+    "loggingrc.js",
+    "MigratorConfig.js",
+    "nodemon.json",
+    "tsconfig.json"
+  ],
   "scripts": {
-    "archive": "pnpm pack:standalone && pnpm pack:tarball",
-    "pack:standalone": "rm -rf package; find . -maxdepth 1 -name 'ghost-*.tgz' -delete; npm pack && tar xzf ghost-*.tgz && cp ../../pnpm-lock.yaml package/ && cp ../../.npmrc package/.npmrc && echo '\nfrozen-lockfile=false\nshamefully-hoist=true' >> package/.npmrc && rm ghost-*.tgz && rm -f pnpm-lock.yaml pnpm-workspace.yaml .npmrc",
```
EvanHahn, line 28: praise: glad this was removed.

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+// moment to 2.24.0 so moment-timezone doesn't pull in a separate copy). pnpm deploy
+// creates a standalone context where these don't apply, so we merge them in.
+if (rootPkg.pnpm?.overrides || rootPkg.overrides) {
+    const rootOverrides = rootPkg.pnpm?.overrides || rootPkg.overrides;
+    pkg.pnpm = pkg.pnpm || {};
+    pkg.pnpm.overrides = {...rootOverrides, ...pkg.pnpm.overrides};
+    console.log(`  Merged ${Object.keys(rootOverrides).length} root overrides into package.json`);
+}
+
+fs.writeFileSync(pkgPath, JSON.stringify(pkg, null, 2) + '\n');
```
EvanHahn, line 100: - nit: is that trailing newline necessary?

```suggestion
fs.writeFileSync(pkgPath, JSON.stringify(pkg, null, 2));
```

- nit: if you're already going to use `fs-extra` elsewhere (see a prior comment), you could use [`writeJsonSync`](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/writeJson-sync.md) here too.

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+fs.writeFileSync(npmrcDst, npmrc);
+
+const workspaceSrc = path.join(ROOT_DIR, 'pnpm-workspace.yaml');
+const workspaceDst = path.join(DEPLOY_DIR, 'pnpm-workspace.yaml');
+fs.copyFileSync(workspaceSrc, workspaceDst);
+
+console.log('Copied .npmrc (with frozen-lockfile=false) and pnpm-workspace.yaml');
+
+// 3. Create tarball
+// ghost-cli expects a package/ prefix inside the tarball.
```
EvanHahn, line 119: thought: if relevant, this is just how npm packages work. You can see this structure by running, say, `npm pack react`, and looking at the result.

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+const version = pkg.version;
+const tgzPath = path.join(CORE_DIR, `ghost-${version}.tgz`);
+
+// Remove node_modules — both the tarball and Docker build install their own
+execSync(`rm -rf ${path.join(DEPLOY_DIR, 'node_modules')}`);
+
+console.log(`\nCreating tarball: ghost-${version}.tgz`);
+execSync(
+    `tar czf ${tgzPath} package`,
```
EvanHahn, line 128: - nit: this will be my final time complaining that, if `tgzPath` is weird, this command could get weird.
- thought: could this use `npm pack`?

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+
+console.log('Copied .npmrc (with frozen-lockfile=false) and pnpm-workspace.yaml');
+
+// 3. Create tarball
+// ghost-cli expects a package/ prefix inside the tarball.
+const version = pkg.version;
+const tgzPath = path.join(CORE_DIR, `ghost-${version}.tgz`);
+
+// Remove node_modules — both the tarball and Docker build install their own
+execSync(`rm -rf ${path.join(DEPLOY_DIR, 'node_modules')}`);
```
EvanHahn, line 124: nit: I'm a broken record, but if that path is weird, this command could get weird.

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+
+console.log(`\nCreating tarball: ghost-${version}.tgz`);
+execSync(
+    `tar czf ${tgzPath} package`,
+    {cwd: CORE_DIR, stdio: 'inherit'}
+);
+
+const size = (fs.statSync(tgzPath).size / 1024 / 1024).toFixed(1);
+console.log(`\nDone: ${tgzPath} (${size} MB)`);
```
EvanHahn, line 133: nit: to be extremely pedantic, this is MiB, not MB.

```suggestion
console.log(`\nDone: ${tgzPath} (${size} MiB)`);
```

Please feel free to ignore and/or to hate me for leaving this comment

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+
+    console.log(`  Packing ${key} → components/${tgzName}`);
+    execSync(`npm pack --pack-destination ${componentsDir}`, {cwd: depDir, stdio: 'pipe'});
+    pkg.dependencies[key] = `file:components/${tgzName}`;
+}
+
+// Carry over root-level fields that pnpm deploy doesn't preserve.
+const rootPkg = JSON.parse(fs.readFileSync(path.join(ROOT_DIR, 'package.json'), 'utf8'));
+
+// packageManager — required by corepack (Dockerfile) and verified by CI release checks.
```
EvanHahn, line 84: nit: If it's required, why are we allowed to skip it? Should this be an error instead?

.github/workflows/ci.yml
```diff
@@ -798,13 +801,19 @@ jobs:
       - name: Clean Install
         run: |
           DIR=$(mktemp -d)
-          ghost install local -d $DIR --archive $(pwd)/ghost/core/ghost.tgz
+          ghost install local -d "$DIR" --archive "$(pwd)/ghost/core/ghost.tgz"
```
EvanHahn, line 804: - praise: this is a good improvement. (Same comment applies in other places you've added quoting.)
- praise: this is a better way to test that the install works.

.github/workflows/ci.yml
```diff
@@ -1722,12 +1727,7 @@ jobs:
         run: npm install -g npm@11
 
       - name: Verify tarball contents
-        run: |
-          echo "Tarball contents:"
-          tar -tzf ghost-*.tgz | head -20
-          tar -tzf ghost-*.tgz | grep -q 'package/.npmrc' || { echo "::error::.npmrc not found in tarball"; exit 1; }
-          tar -tzf ghost-*.tgz | grep -q 'package/pnpm-lock.yaml' || { echo "::error::pnpm-lock.yaml not found in tarball"; exit 1; }
-          tar -xOf ghost-*.tgz package/package.json | jq -e '.packageManager' >/dev/null || { echo "::error::packageManager not found in packaged package.json"; exit 1; }
+        run: tar -xOf ghost-*.tgz package/package.json | jq -e '.packageManager' >/dev/null
```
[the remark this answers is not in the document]
EvanHahn: I agree with the spirit of this comment. Seems like we shouldn't remove the previous verifications. Maybe that's covered by the checks you added earlier?

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env node
+
+/**
+ * pack.js — Build a distributable Ghost tarball using pnpm deploy.
+ *
```
EvanHahn, line 5: thought: Could be useful to state where this tarball will be used. It's implied this is just for Ghost CLI, but I don't actually know! Feel free to ignore if this should be obvious.

ghost/core/scripts/pack.js
```diff
@@ -0,0 +1,133 @@
...
+fs.mkdirSync(componentsDir, {recursive: true});
+
+for (const [key, val] of Object.entries(pkg.dependencies || {})) {
+    if (typeof val !== 'string' || !val.includes('file:')) {
+        continue;
+    }
+
+    const depDir = path.join(DEPLOY_DIR, 'node_modules', key);
+    if (!fs.existsSync(depDir)) {
+        console.warn(`  Warning: ${key} not found in node_modules, skipping`);
```
EvanHahn, line 68: nit: seems like this should be an error, not a warning. If that dependency doesn't exist on disk, that seems like it could be bad.

EvanHahn (changes requested):
Lots of little nitpicks, but looks great overall. Definitely an improvement.

ghost/core/scripts/pack.js
9larsons: I've moved to `execFileSync` which I think accomplishes the goal.

ghost/core/scripts/pack.js
9larsons: Haha, I appreciate these!

ghost/core/scripts/pack.js
9larsons: I believe the existing pattern follows npm's conventions. I think we'd have to add a step after `npm pack` to handle the different naming, though I agree hardening this is not a bad idea.

ghost/core/scripts/pack.js
9larsons: `npm pack` is really tailored around building a package to publish on npm, which isn't our intent here. `npm pack` has hard-coded exclusions for `pnpm-lock.yaml` and we'd need to repack to include it. 

I think as long as we're using Ghost-CLI, this'll be the pattern we need (`pnpm deploy + tar czf`).

.github/workflows/ci.yml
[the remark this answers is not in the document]
9larsons: Agreed - added `job_ghost-cli` to `publish_ghost` and a few file checks in the bottom of pack.js.

apps/activitypub/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/activitypub",
-  "version": "3.1.13",
+  "version": "3.1.11",
```
EvanHahn: Should we revert the changes in this file?

EvanHahn (approved):

apps/activitypub/package.json
9larsons: Yep! Not sure how that snuck in. Reverted.

PR #27348, opened 2026-04-11
Updated i18n/admin jobs to leverage nx affected

acburdine:
## Summary

- remove unused dorny/paths-filter actions/output
- remove additional unused path-filter outputs
- update i18n/admin tests to use nx affected to determine if they should run

## Why

For i18n, we had to check if several different paths had changed to determine if the job should run. If more packages start to trigger i18n tests, the condition has to be updated and get more complicated. By using nx `tags` to dictate to the task graph which packages should trigger i18n tests, the configuration gets a lot simpler.

For admin, it was a simple matter of updating the condition to look in the affected packages list.

### Note about tags

The i18n dependency is a bit of a tricky one as far as Nx is concerned because there's a bit of "cyclical dependency" going on. Each of the packages with translations depends on the code in i18n, but we also want i18n tests to run whenever one of the packages that depend on it changes.

My original approach to solving this was to add each of the i18n-using packages as `implicitDependencies` in the i18n package (which would tell the nx task graph that if any of the packages changes, i18n tests should run). The downside of that approach is it creates cyclical dependencies; while nx allows this, it means that now a change in any of the i18n apps ends up telling nx that _all_ of the i18n apps have changed. (i.e. a change in portal triggers a "change" in i18n, which triggers a change in all packages that depend on i18n). Nx's package isolation boundaries also make it so we can't say "i18n depends on the _files_ in these other packages".

Tagging seemed like ultimately the simplest approach that leverages the task graph well and requires minimal configuration on the part of each package - we'll just need to remember in the future if more i18n packages are added that they will also need the i18n tag.


acburdine:
Also worth mentioning that the code as written doesn't detect changes to the github actions files yet. I have that locally in `nx.json` and will commit it, just wanted to push one commit without it first to verify the accepted logic functions as designed (nx.json triggers "accepted" in all packages)

acburdine:
@9larsons (or @rob-ghost since you're also doing CI work) this is ready for review whenever you have a minute 😄

.github/workflows/ci.yml
```diff
@@ -268,11 +255,8 @@ jobs:
     name: i18n
     if: |
         needs.job_setup.outputs.is_tag == 'true'
-        || needs.job_setup.outputs.changed_comments_ui == 'true'
-        || needs.job_setup.outputs.changed_signup_form == 'true'
-        || needs.job_setup.outputs.changed_sodo_search == 'true'
-        || needs.job_setup.outputs.changed_portal == 'true'
-        || needs.job_setup.outputs.changed_core == 'true'
+        || needs.job_setup.outputs.changed_i18n_apps == 'true'
+        || contains(fromJSON(needs.job_setup.outputs.affected_projects), '@tryghost/i18n')
```
9larsons: Should we just tag this with i18n as well? It would clean up the logic slightly.

acburdine: that's fair, can do that

apps/portal/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/portal",
-  "version": "2.68.9",
+  "version": "2.68.10",
```
9larsons, line 3: I bumped this to pass CI's app version checks.

nx.json
```diff
@@ -1,7 +1,12 @@
 {
     "$schema": "./node_modules/nx/schemas/nx-schema.json",
     "namedInputs": {
-        "default": ["{projectRoot}/**/*", "{workspaceRoot}/ghost/tsconfig.json"]
+        "default": [
```
9larsons, line 4: `suggestion:` I think we want `package-lock.json` and `pnpm-lock.yaml` in here as well.

nx.json
```diff
@@ -1,7 +1,12 @@
 {
     "$schema": "./node_modules/nx/schemas/nx-schema.json",
     "namedInputs": {
```
9larsons, line 3: `suggestion:` What about splitting this into `default` and `global` like nx scaffolds by [default](https://nx.dev/docs/reference/inputs#named-input-conventions)?

```
  "namedInputs": {                                                                                                                                                                                                                                         
    "sharedGlobals": [                                                            
      "{workspaceRoot}/pnpm-lock.yaml",                                                                                                                                                                                                                    
      "{workspaceRoot}/pnpm-workspace.yaml",                                                                                                                                                                                                               
      "{workspaceRoot}/.npmrc",                                                                                                                                                                                                                            
      "{workspaceRoot}/ghost/tsconfig.json"                                                                                                                                                                                                                
    ],                                                                                                                                                                                                                                                     
    "default": ["{projectRoot}/**/*", "sharedGlobals"]                                               
  } 
```

.github/workflows/ci.yml
```diff
@@ -170,24 +152,34 @@ jobs:
...
+
+          I18N_PROJECTS=$(pnpm -s nx show projects ${AFFECTED_ARG} --projects 'tag:i18n' --sep=, | tr -d '\n')
+          echo "affected_i18n_projects=${I18N_PROJECTS}" >> "$GITHUB_OUTPUT"
 
     outputs:
       affected_projects: ${{ steps.affected.outputs.affected_projects }}
-      changed_admin: ${{ steps.changed.outputs.admin }}
+      affected_projects_str: ${{ steps.affected.outputs.affected_projects_str }}
+      changed_i18n_apps: ${{ steps.affected.outputs.affected_i18n_projects != '' }}
+      changed_ci: ${{ steps.changed.outputs.ci }}
```
9larsons: `question:` I think this needs to be cleaned up. Is anything using `changed_ci`?

9larsons (changes requested):
A question and a few suggestions. Looks like a solid change!

nx.json
acburdine: I believe those are [factored in by default](https://nx.dev/docs/reference/inputs) but I'll double check

.github/workflows/ci.yml
acburdine: no, I added it in case it needed to be used elsewhere later, but I can remove it until it's actually needed

acburdine: removed 👍🏻

acburdine:
@9larsons feedback is addressed 👍🏻

9larsons (approved):
Made a minor change and bumped Portal so we can get this merged. Thanks!

PR #27311, opened 2026-04-09
Skipped CI build and E2E jobs for non-code changes

ErisDS:
## Summary

- Narrowed the shared paths-filter anchor from .github/** to only CI-relevant paths (.github/workflows/**, .github/actions/**, .github/scripts/**), so changes to renovate config, CODEOWNERS, codecov.yml, issue templates, and other non-CI metadata no longer trigger all test suites
- Expanded the any-code filter to exclude docs, IDE config, AI agent config, and other non-code files
- Wired the changed_any_code gate into job_build_artifacts, job_build_e2e_public_apps, job_build_e2e_image, and job_e2e_tests - these previously ran unconditionally on every non-tag PR

### What this skips for non-code PRs

The entire Docker build + E2E chain (~90 runner-minutes across 8 shards + 3 build jobs), plus lint and unit tests. For a config-only PR, CI now runs just Setup and the gate job.

### What is safe

- job_required_tests uses if: always() and treats skipped dependencies as passing
- trigger_cd and publish_ghost already check needs.job_build_artifacts.result == success, so they skip when build is skipped
- Test files live inside ghost/, apps/, and e2e/, so they still trigger any-code and their respective package filters

## Test plan

- [ ] Verify this PRs own CI run passes (it touches .github/workflows/ci.yml so all suites should still trigger via shared)
- [ ] Verify a non-code PR (e.g. renovate.json5 change, docs update) skips build + E2E jobs
- [ ] Verify a code PR still runs the full pipeline including E2E

PR #27349, opened 2026-04-11
Included CI-relevant paths in nx.json named inputs

acburdine:
## Summary

Adds filepaths for shared ci/other flows to default nx named inputs. This allows us to remove the dependency on the 'shared' filepath detection in the ci paths-filter and exclusively leverage nx affected to determine if builds should run

## Why

Relates to #27348 - once we start to use `nx affected` to conditionally skip/trigger jobs, we will need some way to dictate to `nx` that changes to the `.github/` folder should trigger reruns of packages in the workspace. Instead of looking at changes in _all_ `.github` files however, I borrowed the list of relevant files from #27311 so things are a bit cleaner.

## Potential Drawbacks

Because nx.json affects local development as well as CI runs, anytime the Github CI-related files are changed, it will also invalidate the local nx build cache.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -1,7 +1,15 @@
 {
     "$schema": "./node_modules/nx/schemas/nx-schema.json",
     "namedInputs": {
-        "default": ["{projectRoot}/**/*", "{workspaceRoot}/ghost/tsconfig.json"]
+        "default": [
+            "{projectRoot}/**/*",
+            "{workspaceRoot}/package.json",
+            "{workspaceRoot}/ghost/tsconfig.json",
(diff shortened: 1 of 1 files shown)


acburdine:
I also looked into whether it's possible to have a separate set of "named inputs" that are only taken into account in CI, but unfortunately it looks like the `nx show projects` command doesn't take into account the `--configuration` flag needed to make that work 🤔

acburdine:
hmm actually I think I can take the approach I did initially in #27202 and it'll work better than adding all of the github ci files to named inputs

PR #27357, opened 2026-04-12
🐛 Fixed RTL languages rendering left-to-right in newsletter emails

betschki:
Newsletters for sites with an RTL publication language (Persian, Arabic, Hebrew, Urdu) rendered left-to-right because the email template never emitted lang/dir on the root `<html>` element.

Portal already calls i18next's built-in dir() for the same purpose; the email renderer just never wired it through. The renderer now passes the resolved locale and direction into the template context, the template sets html lang/dir, the inline body style picks up direction, and the feedback button row mirrors with the document.

Defaults to ltr when no dir helper is provided so existing renders are byte-identical for LTR locales.

Before:
Screenshot 2026-04-12 at 18-04-31 Ghost Admin - Test

After:
Screenshot 2026-04-12 at 18-03-15 Ghost Admin - Test

- Why are you making it?
- What does it do?


marceloomens:
I'd love to see this PR merged soonest, as I currently have an issue with Ghost newsletter RTL support in production. Please let me know if there's anything I can do in support of moving this PR along.

ghost/core/core/server/services/email-service/email-templates/partials/styles.hbs
```diff
@@ -494,7 +494,7 @@ h3.latest-posts-header {
 .manage-subscription {
     font-size: 15px;
     font-weight: 400;
-    text-align: right;
+    text-align: end;
```
EvanHahn: `text-align: end` and `text-align: start` [are unsupported in some email clients](https://www.caniemail.com/features/css-text-align/). We may need to work around this by interpolating a value.

ghost/core/core/server/services/email-service/email-renderer.js
```diff
@@ -196,6 +197,7 @@ class EmailRenderer {
      * @param {object} dependencies.labs
      * @param {{Post: object}} dependencies.models
      * @param {Function} dependencies.t
+     * @param {Function} [dependencies.dir] Returns 'rtl' or 'ltr' for a given locale (i18next's `i18n.dir`)
```
EvanHahn: - nit: could you use a proper type like `() => 'rtl' | 'ltr'` instead of just `Function`?
- nit: could you make this function required? I see no reason to omit it and handle that case. That'll also let you remove the "Falls back to ltr when dir helper is not provided" test below.

ghost/core/core/server/services/email-rendering/partials/base-styles.hbs
```diff
@@ -22,6 +22,7 @@ body {
     -ms-text-size-adjust: 100%;
     -webkit-text-size-adjust: 100%;
     color: #15212A;
+    direction: {{site.direction}};
```
EvanHahn, line 25: I don't know much about this property, but [MDN discourages it](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/direction) in favor of the `dir` HTML attribute (which you've set). Is this necessary?

ghost/core/test/unit/server/services/email-service/email-renderer.test.js
```diff
@@ -2400,6 +2431,47 @@ describe('Email renderer', function () {
...
+        });
+
+        it('Sets site.direction to rtl for Arabic, Hebrew, and Urdu', async function () {
+            for (const locale of ['ar', 'he', 'ur']) {
+                settings.locale = locale;
+                const data = await templateDataWithSettings({});
+                assert.equal(data.site.locale, locale, `expected locale ${locale}`);
+                assert.equal(data.site.direction, 'rtl', `expected rtl for ${locale}`);
+            }
+        });
```
EvanHahn: nit: can you combine these two tests into one? I think you can remove the first test and add `fa` into the locale array.

ghost/core/test/unit/server/services/email-service/email-renderer.test.js
```diff
@@ -1374,6 +1375,35 @@ describe('Email renderer', function () {
...
+
+        it('Renders RTL <html> attributes for Arabic, Hebrew, and Urdu', async function () {
+            for (const locale of ['ar', 'he', 'ur']) {
+                customSettings.locale = locale;
+                const post = createModel(basePost);
+                const newsletter = createModel(baseNewsletter);
+                const response = await emailRenderer.renderBody(post, newsletter, null, {});
+                assert.match(response.html, new RegExp(`<html lang="${locale}" dir="rtl">`), `expected rtl <html> for ${locale}`);
+            }
+        });
```
EvanHahn: nit: can you combine these two tests into one? I think you can remove the first test and add `fa` into the locale array.

EvanHahn (changes requested):
Mostly looks good but I have a few comments. Thanks for doing this!

ghost/core/core/server/services/email-rendering/partials/base-styles.hbs
marceloomens: In my experience duplicating this (so both the `dir="rtl"` HTML attribute and `direction: rtl;` style) gives the most consistent result across clients.

I'm personally also in favor of "just the HTML attribute". Or better still, modern clients implying "dir=rtl" based on the `lang` attribute.

I would prefer to be pragmatic, aim for cross-client consistency, and duplicate this.

betschki: I like @marceloomens's approach. Any objection to that @EvanHahn?

EvanHahn:
Thanks for the edits. I'll take another look next week.

marceloomens:
I'd love to give this PR a bump (@EvanHahn). Self-hosting Ghost, I was able to implement a hackish solution to this aim. Now I've migrated to Magic Pages and @betschki (rightly so) doesn't want to do anything custom for a single site, my newsletters just aren't the same to a reader of RTL languages :-)

EvanHahn (approved):
Looks like there are some merge conflicts but this should be good otherwise!

betschki:
@EvanHahn thank you − should be mergeable now

ghost/core/content/themes/casper
EvanHahn, line 1: One last thing: could you revert these changes to submodules?

```sh
git checkout main ghost/core/content/themes/casper ghost/core/content/themes/source
```

EvanHahn (changes requested):
One last thing before we can merge!

EvanHahn:
I don't think I have edit access to your branch. If you grant that, I can get this over the finish line. Otherwise, if you could address [my (hopefully quick) comment](https://github.com/TryGhost/Ghost/pull/27357#pullrequestreview-4181351548), we can get this merged.

betschki:
Good catch, thank you @EvanHahn − reverted them now

EvanHahn (approved):
Thanks! Queuing this for merge.

EvanHahn:
Could you merge `main` into this branch?

EvanHahn (approved):

EvanHahn:
Thanks so much!

PR #27362, opened 2026-04-12
Fixed boot with fresh development container

EvanHahn:
ref f186b6adeb58412915012ac49c078a5fc96797c2

This fixes two bugs when booting a fresh development container:

- Fixes `Cannot find module '/home/ghost/.github/scripts/enforce-package-manager.js'` by copying all of `.github/scripts`, not just the `install-deps.sh` script
- `ghost/i18n` required `@tryghost/debug` but did not declare it, which meant a fresh container couldn't start


9larsons (approved):
Great catch! Thanks.

PR #27370, opened 2026-04-13
Added link preview support for gift subscription links

minimaluminium:
https://linear.app/ghost/issue/BER-3524/ghost-core-show-gift-specific-og-meta-data-for-link-previews

When a gift subscription link is shared, it now shows a rich link preview with Open Graph meta tags and a dynamically generated image.                         
                                                                                                                                                                                                                     
  - Added a new `/gift/:token` route that serves OG meta tags (title, description, image) and redirects visitors to the Portal redemption flow                                                                       
  - Added `/gift/:token/og-image` route that generates a branded PNG on-the-fly using the publication's accent color, tier name, and duration                                                                        
  - Image matches the design of the Portal redemption modal (same icon, accent color at 6% opacity background)                                                                                                       
  - Generated images are cached in-memory and via HTTP cache headers


ghost/core/core/server/web/gift/controller.js
```diff
@@ -0,0 +1,125 @@
...
+    return str
+        .replace(/&/g, '&amp;')
+        .replace(/"/g, '&quot;')
+        .replace(/</g, '&lt;')
+        .replace(/>/g, '&gt;');
+}
+
+async function fetchGiftData(token) {
+    const giftServiceWrapper = require('../../services/gifts');
+    await giftServiceWrapper.init();
```
sagzy: Gift service is initialised in https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/boot.js - we shouldn't need to initialise it here

ghost/core/core/server/web/gift-preview/controller.js
```diff
@@ -0,0 +1,143 @@
+const errors = require('@tryghost/errors');
+const logging = require('@tryghost/logging');
+const {generateGiftPreviewImage} = require('./image');
+
+function getCadenceLabel(cadence, duration) {
+    const unit = cadence === 'month' ? 'month' : 'year';
```
sagzy: nit: unnecessary variable, could do

```
return duration === 1 ? `1 ${cadence}` : `${duration} ${cadence}s`;
```

ghost/core/core/server/web/gift-preview/controller.js
```diff
@@ -0,0 +1,143 @@
...
+            throw new errors.NotFoundError({message: `Tier not found: ${gift.tierId}`});
+        }
+    } catch (err) {
+        logging.warn(`Gift preview: failed to load required gift data, redirecting to homepage`, err);
+
+        return res.redirect(302, siteUrl + '/');
+    }
+
+    const cadenceLabel = getCadenceLabel(gift.cadence, gift.duration);
+    const ogTitle = `A gift membership to ${siteTitle}`;
```
sagzy, line 50: Is there an easy way to localise these strings using the i18n service? That'd be pretty nice for non-English sites

mike182uk: I think we should be able to do this. Should we do it as part of the i18n task we have for the feature?

sagzy: Could do yes

ghost/core/core/server/web/gift-preview/controller.js
mike182uk: Fixed in https://github.com/TryGhost/Ghost/pull/27370/commits/232c7dc40b987da6989cb38b45f6b57150322731

PR #27376, opened 2026-04-13
Skipped nx shas step in CI for tag workflow runs

cmraible:
refs https://github.com/TryGhost/Ghost/actions/runs/24363396931/job/71148722202

## Summary

Skip Nx SHA and affected-project setup during tag-triggered CI runs.

## Problem

The `v6.29.0` release failed in `job_setup` on the tag-triggered `CI` workflow. After `#27297`, `job_setup` started using `nrwl/nx-set-shas` to derive `NX_BASE`, and then passed that value into `paths-filter` and Nx affected-project detection.

On a tag push, `github.ref_name` is the tag itself, so `nx-set-shas` tried to resolve workflow history for `v6.29.0` as though it were a branch. That failed, so it fell back to the empty-tree hash (`4b825dc...`). `dorny/paths-filter` then errored because that object is a tree, not a commit.

## Why this fix

Tag-triggered release runs do not need Nx base resolution, changed-file detection, or affected-project calculation:
- the downstream jobs that use those outputs are already skipped on tags
- the release path only needs `job_setup` to succeed so artifact build/publish jobs can proceed

Instead of trying to infer a “correct” comparison base for tags, this change skips the tag-incompatible diff logic entirely on tag runs while preserving the existing behavior for PRs and branch pushes.

**Changes**

- Skip `Set SHAs for Nx Commands` on tag runs
- Skip `Determine added packages` on tag runs
- Skip `Determine changed packages` on tag runs
- Skip `Determine Affected Projects` on tag runs

**Result**

Branch and PR CI continue to use Nx/affected diffing as before, while tag-triggered releases no longer fail in setup due to an invalid `NX_BASE`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,7 @@ jobs:
           echo "GITHUB_CONTEXT: ${{ toJson(github.event) }}"
 
       - name: Set SHAs for Nx Commands
+        if: env.IS_TAG != 'true'
         uses: nrwl/nx-set-shas@afb73a62d26e41464e9254689e1fd6122ee683c1 # v5.0.1
         with:
           main-branch-name: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.ref || github.ref_name }}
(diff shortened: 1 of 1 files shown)


cmraible:
@Coderabbit review

EvanHahn (approved):
We talked about this on a call and, while I don't have all the context, I think this is good enough to unblock release. (And may be good enough for all time!)

troyciesco (approved):

PR #27377, opened 2026-04-13
Revert "Bumped version to 6.30.0-rc.0"

cmraible:
ref https://github.com/TryGhost/Ghost/pull/27376

This reverts commit 60ba2126a3232af35eb49cea0affcabf4e03f433.

The `v6.29.0` release failed after creating the tag. This reverts the version in package.json back to `v6.29.0` so that we can release `v6.29.1`.


EvanHahn (approved):
We talked about this on a call.

PR #27378, opened 2026-04-13
Removed package directory cleanup from `pack:tarball` script

cmraible:
ref https://github.com/TryGhost/Ghost/actions/runs/24369399597/job/71170929638

The CI workflow is failing at the "Prepare docker build context" step on runs on tags because `ghost/core/package/` doesn't exist:

```mv: cannot stat 'ghost/core/package/': No such file or directory```

At this stage in the workflow, `ghost/core/package/` doesn't exist, because it was deleted by the `pack:tarball` script a few steps prior in the same workflow by the `&& rm -rf package` addition to the `pack:tarball` script. Prior to the pnpm migration, the `package` directory wasn't cleaned up at all. This commit returns to that state, so that `ghost/core/package/` should still exist once it gets to the "Prepare docker build context" step.


9larsons (approved):
Seems reasonable! I'd like to clean up after packing but it's not a big deal right now and shouldn't block this.

PR #27381, opened 2026-04-14
Added gift subscription label to Portal account page

sagzy:
ref https://linear.app/ghost/issue/BER-3486

## Summary

This PR lets Portal distinguish gift members from complimentary members on the Account page.

Previously, `/members/api/member` did not include the member `status`, so Portal could only infer zero-value subscriptions as "complimentary". That caused both comped and gift memberships to display the same label.

With this change:
- the members API now returns `status`
- Portal uses `member.status` to differentiate gift vs complimentary copy
- gift memberships now display `Gift subscription` on the Account page, with an expiry date

### Backend
- Added `status` to the `/members/api/member` response

### Portal
- Updated Account page plan label logic to use `member.status` for copy
- Gift members now show:
  - `Gift subscription - Expires {date}`
  - `Gift subscription`
- Complimentary members continue to show:
  - `Complimentary - Expires {date}`
  - `Complimentary`
- Renamed the shared helper from `getCompExpiry` to `getSubscriptionExpiry`


apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
```diff
@@ -33,12 +33,20 @@ const PaidAccountActions = () => {
...
-        const compExpiry = getCompExpiry({member});
-        if (isComplimentary) {
-            if (compExpiry) {
-                label = `${t('Complimentary')} - ${t('Expires {expiryDate}', {expiryDate: compExpiry})}`;
+        const subscriptionExpiry = getSubscriptionExpiry({member});
+        const isGiftMember = member?.status === 'gift';
+
+        if (isGiftMember) {
+            if (subscriptionExpiry) {
+                label = `${t('Gift subscription')} - ${t('Expires {expiryDate}', {expiryDate: subscriptionExpiry})}`;
```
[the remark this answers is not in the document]
sagzy: Intentional: helps to reuse the existing translations + add fewer new keys to translate

apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
```diff
@@ -33,12 +33,20 @@ const PaidAccountActions = () => {
...
+            } else {
+                label = t('Gift subscription');
+            }
+        } else if (isComplimentary) {
+            if (subscriptionExpiry) {
+                label = `${t('Complimentary')} - ${t('Expires {expiryDate}', {expiryDate: subscriptionExpiry})}`;
```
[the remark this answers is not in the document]
sagzy: Intentional: helps to reuse the existing translations + add fewer new keys to translate

apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
```diff
@@ -33,12 +33,20 @@ const PaidAccountActions = () => {
...
-                label = `${t('Complimentary')} - ${t('Expires {expiryDate}', {expiryDate: compExpiry})}`;
+        const subscriptionExpiry = getSubscriptionExpiry({member});
+        const isGiftMember = member?.status === 'gift';
+
+        if (isGiftMember) {
+            if (subscriptionExpiry) {
+                label = `${t('Gift subscription')} - ${t('Expires {expiryDate}', {expiryDate: subscriptionExpiry})}`;
+            } else {
+                label = t('Gift subscription');
+            }
```
mike182uk: Gift will always have an expiry right?

apps/portal/src/utils/fixtures-generator.js
```diff
@@ -184,6 +185,12 @@ export function getMemberData({
         email_suppression,
         newsletters
     };
+
+    if (status !== undefined) {
+        member.status = status;
+    }
```
mike182uk, line 195: Hm, wonder if we should do something more robust here seeming though we pass `paid` as a a param already?

ghost/core/core/server/services/members/utils.js
```diff
@@ -25,6 +25,7 @@ module.exports.formattedMemberResponse = function formattedMemberResponse(member
         unsubscribe_url: member.unsubscribe_url,
         subscribed: !!member.subscribed,
         subscriptions: member.subscriptions || [],
+        status: member.status,
```
mike182uk, line 28: What if we didn't return `status` and instead something like `gift: true` or something, like we do for `paid`? Would this be overkill? I was thinking returning status might be leaking implementation details a little, but may be overthinking it 🤔

apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
sagzy: Good shout, updated in https://github.com/TryGhost/Ghost/pull/27381/changes/8d30614cc497b390e7c402cd4f2bfd7086e81239

apps/portal/src/utils/fixtures-generator.js
sagzy: Yep, updated in https://github.com/TryGhost/Ghost/pull/27381/changes/8d30614cc497b390e7c402cd4f2bfd7086e81239

ghost/core/core/server/services/members/utils.js
sagzy: (Discussing on Slack)

mike182uk (approved):

PR #27382, opened 2026-04-14
Added job for cleaning up consumed gifts

mike182uk:
ref https://linear.app/ghost/issue/BER-3473

Added job for cleaning up consumed gifts so that members with gift subscriptions are correctly downgraded to `free` when their gift subscription has run out

### DIFF
--- a/ghost/core/core/server/services/gifts/gift-bookshelf-repository.ts
+++ b/ghost/core/core/server/services/gifts/gift-bookshelf-repository.ts
@@ -76,12 +76,10 @@ export class GiftBookshelfRepository implements GiftRepository {
     }
 
     async findPendingConsumption(): Promise<Gift[]> {
-        const today = new Date();
-
-        today.setUTCHours(0, 0, 0, 0);
+        const now = new Date();
 
         const collection = await this.model.findAll({
-            filter: `status:redeemed+consumes_at:<'${today.toISOString()}'`
+            filter: `status:redeemed+consumes_at:<'${now.toISOString()}'`
         });
 
         return collection.models.map(model => this.toGift(model));
--- a/ghost/core/core/server/services/gifts/gift-bookshelf-repository.ts
+++ b/ghost/core/core/server/services/gifts/gift-bookshelf-repository.ts
@@ -8,10 +8,15 @@ type BookshelfDocument<T> = {
     toJSON(): T;
 };
 
+type BookshelfCollection<T> = {
+    models: BookshelfDocument<T>[];
+};
+
 type BookshelfModel<T> = {
     add(data: Partial<T>, unfilteredOptions?: unknown): Promise<T>;
     transaction<R>(callback: (transacting: unknown) => Promise<R>): Promise<R>;
     findOne(data: Record<string, unknown>, unfilteredOptions?: unknown): Promise<BookshelfDocument<T> | null>;
+    findAll(unfilteredOptions?: unknown): Promise<BookshelfCollection<T>>;
 };
 
 type GiftRow = {
@@ -59,7 +64,7 @@ export class GiftBookshelfRepository implements GiftRepository {
             token
         }, {require: false, ...options});
 
-        return this.toGift(model);
+        return model ? this.toGift(model) : null;
     }
 
     async getByPaymentIntentId(paymentIntentId: string): Promise<Gift | null> {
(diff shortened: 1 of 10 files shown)


ghost/core/core/server/services/gifts/gift-bookshelf-repository.ts
```diff
@@ -59,15 +64,27 @@ export class GiftBookshelfRepository implements GiftRepository {
...
+        return model ? this.toGift(model) : null;
+    }
+
+    async findPendingConsumption(): Promise<Gift[]> {
+        const today = new Date();
+
+        today.setUTCHours(0, 0, 0, 0);
+
+        const collection = await this.model.findAll({
+            filter: `status:redeemed+consumes_at:<'${today.toISOString()}'`
```
sagzy: question: why not filter by consumes_at < now? (instead of today start of the day)

ghost/core/core/server/services/gifts/gift-service.ts
```diff
@@ -269,13 +269,55 @@ export class GiftService {
...
-                    status: 'free'
-                }, {id: gift.redeemerMemberId, transacting});
+                const member = await this.deps.memberRepository.get({id: gift.redeemerMemberId}, {transacting});
+
+                if (member?.get('status') === 'gift') {
+                    await this.deps.memberRepository.update({
+                        products: [],
+                        status: 'free'
+                    }, {id: gift.redeemerMemberId, transacting});
+                }
```
sagzy, line 279: Good catch here, the extra check also covers things such as member upgraded to paid in the meantime

ghost/core/core/server/services/members/jobs/clean-consumed-gifts.js
```diff
@@ -0,0 +1,57 @@
...
+        await giftService.init();
+
+        const {consumedCount, updatedMemberCount} = await giftService.service.processConsumed();
+
+        const cleanupEndDate = new Date();
+
+        debug(`Consumed ${consumedCount} gifts, updated ${updatedMemberCount} members in ${cleanupEndDate.valueOf() - cleanupStartDate.valueOf()}ms`);
+
+        if (parentPort) {
+            parentPort.postMessage(`Consumed ${consumedCount} gifts, updated ${updatedMemberCount} members in ${cleanupEndDate.valueOf() - cleanupStartDate.valueOf()}ms`);
```
sagzy: nit: could extract the repeated log into a variable

ghost/core/core/server/services/members/jobs/index.js
```diff
@@ -3,52 +3,40 @@ const jobsService = require('../../jobs');
...
+    jobsService.addJob({
+        at: `${s} ${m} ${h} * * *`,
+        job: path.resolve(__dirname, jobFile),
+        name
+    });
+
+    hasScheduled[key] = true;
+
+    return true;
+}
```
sagzy, line 28: nice one 🧹

sagzy (approved):
Looking good @mike182uk! Left one question and one nit

Also think we could use an index for the query:
```
SELECT
	*
FROM
	gifts
WHERE
	status = "redeemed"
	AND consumes_at < NOW() /* or today start of day */
```

Claude suggested a composite index:
```
CREATE INDEX gifts_status_consumes_at_idx
ON gifts (status, consumes_at);
```

ghost/core/core/server/services/gifts/gift-bookshelf-repository.ts
mike182uk: This was to ensure the sub lasted the whole day instead of cutting off part way through (which might not be an issue?)

e.g

a gift was redeemed on April 15 2025 for 1 year - `consumes_at` is set to April 15 2026

With the current approach:

- job runs April 15 2026 at 03:42 AM UTC
- today = April 15 2026 00:00:00 UTC
- filter = `consumes_at` < `2026-04-15T00:00:00.000Z`
- gift not processed - its `consumes_at` equals the cutoff (not less than)
- job runs again April 16 2026 at 02:17 AM UTC
- today = April 16 2026 00:00:00 UTC
- gift processed - `consumes_at` is less than the cutoff

With `new Date()` (now):

- job runs April 15 2026 at 03:42 AM UTC
- filter = `consumes_at` < `2026-04-15T03:42:00.000Z`
- gift processed - `consumes_at` (midnight) is less than 03:42

sagzy: When a gift is redeemed, we set the `consumes_at` date time to `now + duration of the gift`. For example, if I redeem a 1-year gift on `April 15 2025 15:46:23`, then `consumes_at` is set to `April 15 2026 15:46:23`.

Therefore, on `April 15 2026 15:46:23`, the receiver has actually used up 1 full year. So imo we can mark the gift as consumed as soon as the job runs, without having to wait for `April 16 2026`

mike182uk: ah my bad, i didn't check `redeem` properly. I've updated this in https://github.com/TryGhost/Ghost/pull/27382/commits/576756a4f3d6a5d9e38a78913f8acbe0e31a0d6f

PR #27385, opened 2026-04-14
Downgraded member when gift subscription is refunded

mike182uk:
ref https://linear.app/ghost/issue/BER-3474

When a gift is refunded, the member is downgrade to `free` and thier product access is removed

### DIFF
--- a/ghost/core/test/unit/server/services/gifts/gift-service.test.ts
+++ b/ghost/core/test/unit/server/services/gifts/gift-service.test.ts
@@ -532,11 +532,12 @@ describe('GiftService', function () {
             assert.equal(result, true);
             sinon.assert.calledOnce(giftRepository.save);
 
-            const saved = giftRepository.save.getCall(0).args[0];
+            const [saved, options] = giftRepository.save.getCall(0).args;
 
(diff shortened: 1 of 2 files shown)


ghost/core/test/unit/server/services/gifts/gift-service.test.ts
```diff
@@ -549,6 +550,57 @@ describe('GiftService', function () {
...
+            });
+
+            giftRepository.getByPaymentIntentId.resolves(gift);
+            memberRepository.update.rejects(new Error('Cannot remove product with active subscription'));
+
+            const service = createService();
+            await assert.rejects(
+                () => service.refund('pi_456'),
+                {message: 'Cannot remove product with active subscription'}
+            );
```
sagzy, line 601: nit: in that test, should we also assert that the gift remained unchanged (e.g., status: redeemed)

mike182uk: Added in https://github.com/TryGhost/Ghost/pull/27385/commits/0cf27e09cc46766ff2da974bef084c26f439719d

sagzy (approved):

PR #27389, opened 2026-04-14
Refactored gift repository to have dedicated update method

mike182uk:
no ref

In the gift repository we decided to have a dedicated `update` method instead of a generic `save` method that also fell back to creating


sagzy (approved):
💯

PR #27393, opened 2026-04-14
Added gift subscription label in Admin member page

sagzy:
closes https://linear.app/ghost/issue/BER-3482

## Summary

Update the Admin member details page to treat gift subscriptions as a distinct subscription type instead of displaying them as complimentary subscriptions.

For gift subscriptions, we now display:

`Gift subscription – Expires {date}`

with no action menu to remove the subscription:

CleanShot 2026-04-14 at 20 52 05@2x

## What changed

- Added gift-specific subscription normalization in the Admin subscription data helper
- Added a `giftExpiry` helper sourced from `sub.tier.expiry_at`
- Updated the subscription label logic so gift members render `Gift subscription`
- Updated validity details so gift subscriptions render `Expires {date}`
- Hid the action menu for gift subscriptions on the member details page
- Kept complimentary subscriptions unchanged:
  - still render as `Complimentary`
  - still use the complimentary expiry path
  - still show the `Remove complimentary subscription` action
- Tightened `isComplimentary` so it only applies to `comped` members with manual subscriptions

### DIFF
--- a/ghost/admin/app/components/gh-member-settings-form.hbs
+++ b/ghost/admin/app/components/gh-member-settings-form.hbs
@@ -183,6 +183,8 @@
                                                     </li>
                                                 </GhDropdown>
                                             </span>
+                                        {{else if sub.isGift}}
+                                        {{! No action menu for gift subscriptions }}
                                         {{else}}
                                             <span class="action-menu">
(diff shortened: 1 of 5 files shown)


sagzy:
As the Admin page shows a history of subscriptions, this is not going to be sufficient. Closing for another solution

PR #27399, opened 2026-04-14
🎨 Added Preview tab to welcome email editor

cmraible:
ref https://linear.app/ghost/issue/NY-1228/add-the-frontend-editpreview-toggle-to-the-welcome-email-edit-modal

## What this changes

- adds an `Edit` / `Preview` toggle to the welcome email modal in Admin
- lets admins preview unsaved subject and body changes before saving
- loads the preview from the automated email preview endpoint and renders it in an iframe sized to the email content
- preserves unsaved draft state when switching between edit and preview
- shows inline validation and preview loading/error states in the modal
- ignores stale out-of-order preview responses so older requests cannot overwrite a newer draft

## Why

The welcome email editor previously only exposed the raw editing experience, which made it harder to verify the final email output before saving. This change adds an in-context preview flow so admins can check both copy and rendered layout while keeping their draft intact.

## Implementation notes

- adds a dedicated `usePreviewWelcomeEmail` mutation and preview response types in `admin-x-framework`
- updates the modal header layout to host the segmented mode switch
- normalizes preview HTML links to open safely in a new tab
- measures iframe height so the modal owns scrolling instead of the preview document
- extends acceptance and browser E2E coverage for preview rendering and draft preservation

### DIFF
--- a/apps/admin-x-settings/src/components/settings/membership/member-emails/member-email-editor.tsx
+++ b/apps/admin-x-settings/src/components/settings/membership/member-emails/member-email-editor.tsx
@@ -27,7 +27,7 @@ const baseEditorStyles = cn(
     // Dark mode
     'dark:text-white dark:selection:bg-[rgba(88,101,116,0.99)]',
     // Placeholder styling
-    '[&_.koenig-lexical-editor-input-placeholder]:font-sans! [&_.koenig-lexical-editor-input-placeholder]:text-xl [&_.koenig-lexical-editor-input-placeholder]:tracking-tight',
+    '[&_.koenig-lexical-editor-input-placeholder]:font-sans! [&_.koenig-lexical-editor-input-placeholder]:text-[1.6rem] [&_.koenig-lexical-editor-input-placeholder]:tracking-tight',
     // Headings dark mode
     '[&_:is(h2,h3)]:dark:text-white',
     // Inputs
@@ -37,7 +37,7 @@ const baseEditorStyles = cn(
     // Settings panel
     '[&_[data-kg-card-selected]]:isolate',
     // Content typography
-    '[&_:is(p,blockquote,aside,ul,ol)]:font-sans! [&_:is(p,blockquote,aside,ul,ol)]:text-xl [&_:is(p,blockquote,aside,ul,ol)]:tracking-tight',
+    '[&_:is(p,blockquote,aside,ul,ol)]:tracking-tight',
     // Reset content typography inside card captions to match Koenig's caption styles
     '[&_figcaption_:is(p,blockquote,aside,ul,ol)]:text-[1.4rem] [&_figcaption_:is(p,blockquote,aside,ul,ol)]:tracking-[.025em]',
     '[&_figcaption_p]:mb-0',
@@ -49,7 +49,7 @@ const baseEditorStyles = cn(
     // Keep settings panel copy compact
     '[&_[data-kg-settings-panel]_p]:!mb-0',
     // Nested-editor (callout, etc.) fixes: align placeholder with text
-    '[&_.not-kg-prose>div]:font-sans! [&_.not-kg-prose>div]:tracking-tight! [&_.not-kg-prose>div]:text-xl! [&_.not-kg-prose>div]:leading-[1.6]!',
+    '[&_.not-kg-prose>div]:font-sans! [&_.not-kg-prose>div]:tracking-tight! [&_.not-kg-prose>div]:text-[1.6rem]! [&_.not-kg-prose>div]:leading-[1.6]!',
     '[&_.kg-inherit-styles_p]:mb-0!',
     '[&_.kg-inherit-styles]:pt-[3px]!',
     // CTA card: keep sponsor label at its intended 12.5px size
--- a/apps/admin-x-settings/src/components/settings/membership/member-emails/use-welcome-email-preview.ts
+++ b/apps/admin-x-settings/src/components/settings/membership/member-emails/use-welcome-email-preview.ts
@@ -1,5 +1,5 @@
 import {JSONError} from '@tryghost/admin-x-framework/errors';
-import {useCallback, useRef, useState} from 'react';
+import {useRef, useState} from 'react';
 
 import {type WelcomeEmailDraft, getWelcomeEmailValidationErrors} from './welcome-email-validation';
 
@@ -63,12 +63,12 @@ export const useWelcomeEmailPreview = ({automatedEmailId, previewWelcomeEmail, s
     const [previewState, setPreviewState] = useState<WelcomeEmailPreviewState>({status: 'idle'});
     const previewRequestIdRef = useRef(0);
 
-    const exitPreview = useCallback(() => {
+    const exitPreview = () => {
         previewRequestIdRef.current += 1;
         setPreviewState({status: 'idle'});
-    }, []);
+    };
 
-    const runPreview = useCallback(async (draft: WelcomeEmailDraft) => {
+    const enterPreview = async (draft: WelcomeEmailDraft) => {
         const requestId = previewRequestIdRef.current + 1;
         previewRequestIdRef.current = requestId;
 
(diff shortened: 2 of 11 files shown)


apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-preview-frame.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+    const previewHtml = previewState.status === 'success' ? previewState.html : '';
+
+    useEffect(() => {
+        cleanupPreviewMeasurement();
+        setIsPreviewReady(false);
+        setPreviewHeight(null);
+    }, [cleanupPreviewMeasurement, previewHtml, previewState.status]);
+
+    useEffect(() => cleanupPreviewMeasurement, [cleanupPreviewMeasurement]);
```
troyciesco, line 41: i haven't verified what everything is doing yet but i think this can just be one effect, since the second one is basically just a cleanup function. like this:

```
useEffect(() => {
    setIsPreviewReady(false);
    setPreviewHeight(null);

    return cleanupPreviewMeasurement;
}, [cleanupPreviewMeasurement, previewHtml, previewState.status]);

```

apps/admin-x-settings/src/hooks/use-welcome-email-preview.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (!state.subject?.trim()) {
+        newErrors.subject = 'A subject is required';
+    }
+
+    if (isEmptyLexical(state.lexical)) {
+        newErrors.lexical = 'Email content is required';
+    }
+
+    return newErrors;
+};
```
cmraible, line 84: Feels a bit weird to export this from the hook file - maybe this should be in a util file instead, and imported here and in the modal

apps/admin-x-settings/src/hooks/use-welcome-email-preview.ts
```diff
@@ -0,0 +1,183 @@
+import {JSONError} from '@tryghost/admin-x-framework/errors';
```
cmraible, line 1: maybe we should move this file into a `/hooks` directory closer to the modal where it's used, rather than a top level catch-all directory for hooks

apps/admin-x-settings/src/hooks/use-welcome-email-preview.ts
```diff
@@ -0,0 +1,183 @@
...
+}) => {
+    const [previewState, setPreviewState] = useState<WelcomeEmailPreviewState>({status: 'idle'});
+    const previewRequestIdRef = useRef(0);
+
+    const exitPreview = useCallback(() => {
+        previewRequestIdRef.current += 1;
+        setPreviewState({status: 'idle'});
+    }, []);
+
+    const runPreview = useCallback(async (draft: WelcomeEmailDraft) => {
```
cmraible: Generally: lots of useCallbacks in here, are these necessary?

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-preview-frame.tsx
cmraible: Agreed, I collapsed these two into a single useEffect

apps/admin-x-settings/src/hooks/use-welcome-email-preview.ts
cmraible: Pulled this out into a separate util file

apps/admin-x-settings/src/components/settings/membership/member-emails/use-welcome-email-preview.ts
```diff
@@ -0,0 +1,183 @@
+import {JSONError} from '@tryghost/admin-x-framework/errors';
```
cmraible: moved this directly into the member-emails/ directory so it's more colocated with the (only) component that uses it.

apps/admin-x-settings/src/hooks/use-welcome-email-preview.ts
cmraible: I ended up removing _all_ of these useCallbacks, in favor of plain functions.

apps/admin-x-settings/src/components/settings/membership/member-emails/member-email-editor.tsx
```diff
@@ -27,7 +27,7 @@ const baseEditorStyles = cn(
     // Dark mode
     'dark:text-white dark:selection:bg-[rgba(88,101,116,0.99)]',
     // Placeholder styling
-    '[&_.koenig-lexical-editor-input-placeholder]:font-sans! [&_.koenig-lexical-editor-input-placeholder]:text-xl [&_.koenig-lexical-editor-input-placeholder]:tracking-tight',
+    '[&_.koenig-lexical-editor-input-placeholder]:font-sans! [&_.koenig-lexical-editor-input-placeholder]:text-[1.6rem] [&_.koenig-lexical-editor-input-placeholder]:tracking-tight',
```
cmraible, line 30: Should this be xl instead of 1.6rem?

cmraible: we don't have a text utility that specifically maps to 1.6, so this is fine here.

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-preview-frame.tsx
```diff
@@ -0,0 +1,151 @@
...
+            window.cancelAnimationFrame(previewMeasureFrameRef.current);
+        }
+
+        previewMeasureFrameRef.current = window.requestAnimationFrame(() => {
+            previewMeasureFrameRef.current = null;
+            syncPreviewHeight();
+        });
+    }, [syncPreviewHeight]);
+
+    const handlePreviewLoad = useCallback(() => {
```
cmraible: do we need these useCallbacks in the component?

troyciesco (approved):

e2e/tests/admin/settings/member-welcome-emails.test.ts
```diff
@@ -142,6 +176,84 @@ test.describe('Ghost Admin - Member Welcome Emails', () => {
...
+                expect(welcomeEmail.Subject).toBe(personalizedSubject);
+                expect(welcomeEmail.Subject).not.toContain('{first_name}');
+            });
+        } finally {
+            const freeWelcomeEmail = await getFreeWelcomeEmail(page);
+            await restoreWelcomeEmail(page, {
+                ...freeWelcomeEmail,
+                subject: DEFAULT_FREE_WELCOME_EMAIL_SUBJECT
+            });
+        }
```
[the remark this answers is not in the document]
cmraible: This is fine, it's the only field that's edited in this test.

e2e/tests/admin/settings/member-welcome-emails.test.ts
```diff
@@ -142,6 +176,84 @@ test.describe('Ghost Admin - Member Welcome Emails', () => {
...
+                response.url().includes('/preview/') &&
+                response.request().method() === 'POST'
+        );
+
+        await welcomeEmailsSection.modalPreviewTab.click();
+        await previewResponse;
+
+        await welcomeEmailsSection.modalSubjectInput.clear();
+        await welcomeEmailsSection.modalSubjectInput.fill(customSubject);
+        await expect(welcomeEmailsSection.modalPreviewSubjectInput).toHaveValue(customSubject);
```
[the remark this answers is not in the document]
cmraible: There is no request triggered by the subject edit. I don't think this is a valid issue.

apps/admin-x-settings/src/components/settings/membership/member-emails/welcome-email-preview-frame.tsx
cmraible: These weren't there for performance reasons, just for the dependency graph, which made it harder to grok IMO. Refactored here to simplify and remove the useCallbacks.

PR #27402, opened 2026-04-15
Ensure sqlite3 is installed for ghost:core unit tests

acburdine:
this PR adds `--force` to pnpm install for the unit tests job, since sqlite3 appears to be required for ghost:core unit tests (see test failures on #27348, #27401)

this wasn't caught during the changes in #27386 likely because the ghost/core package wasn't changed, so it was never fully verified - some of the future work around `nx affected` detection should be able to catch this going forward.

~second commit is temporary, to verify force install fixes unit tests~ second commit removed, see https://github.com/TryGhost/Ghost/actions/runs/24446750005/job/71425107814 to verify unit test job works 👍🏻

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -349,7 +349,11 @@ jobs:
           cache: pnpm
 
       - name: Install dependencies
-        run: pnpm install --frozen-lockfile
+        # sqlite3 is an optionalDependency. Without --force, pnpm may skip
+        # installing/linking it when restoring from a cached store. --force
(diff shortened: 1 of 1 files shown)


9larsons (approved):
Thanks! I should probably adjust the core unit test job to run on all CI changes.

acburdine:
> I should probably adjust the core unit test job to run on all CI changes.

I can take care of that in some of the `nx affected` work, got some ideas on how to make it all work pretty smoothly

PR #27407, opened 2026-04-15
Fixed reference to missing install deps script

mike182uk:
ref https://github.com/TryGhost/Ghost/pull/27386

As part of the above PR `.github/scripts/install-deps.sh` was removed, but it is still referenced in the Dockerfile which causes `pnpm dev` to fail:

```
=> ERROR [ghost-dev stage-0 11/14] RUN --mount=type=cache,target=/root/.local/share/pnpm/store,id=pnpm-store     bash .github/scripts/install-deps.sh                   0.6s
------
 > [ghost-dev stage-0 11/14] RUN --mount=type=cache,target=/root/.local/share/pnpm/store,id=pnpm-store     bash .github/scripts/install-deps.sh:
0.345 bash: .github/scripts/install-deps.sh: No such file or directory
------
Dockerfile:37

--------------------

  36 |     # Install deps with a persistent pnpm store cache to speed up rebuilds

  37 | >>> RUN --mount=type=cache,target=/root/.local/share/pnpm/store,id=pnpm-store \

  38 | >>>     bash .github/scripts/install-deps.sh

  39 |

--------------------

target ghost-dev: failed to solve: process "/bin/sh -c bash .github/scripts/install-deps.sh" did not complete successfully: exit code: 127
```

### DIFF
--- a/docker/ghost-dev/Dockerfile
+++ b/docker/ghost-dev/Dockerfile
@@ -35,7 +35,7 @@ RUN corepack enable
 
 # Install deps with a persistent pnpm store cache to speed up rebuilds
 RUN --mount=type=cache,target=/root/.local/share/pnpm/store,id=pnpm-store \
-    bash .github/scripts/install-deps.sh
+    pnpm install --frozen-lockfile --prefer-offline
 
 # Copy entrypoint script that optionally loads Tinybird config
 COPY docker/ghost-dev/entrypoint.sh entrypoint.sh


9larsons (approved):

troyciesco:
@mike182uk @9larsons fyi that script is also referenced here (L17 and L23): https://github.com/TryGhost/Ghost/blob/main/docker/development.entrypoint.sh#L17

maybe needs the same fix?

9larsons:
Yep, on it. I was going to follow up with that to let this merge (I'm assuming this one works?).

mike182uk:
> I'm assuming this one works?

Yep confirmed working on my local install

PR #27408, opened 2026-04-15
🐛 Fixed members inheriting stale suppression state on resignup

jonatansberg:
## Summary

When a suppressed member was deleted and someone later signed up with the same email, the new member was created with `email_disabled=false` even though the suppression record still existed. The send filter then included them in every newsletter.

Mailgun rejected each send with a 607, triggering the suppression handler. But the handler silently swallowed the `ER_DUP_ENTRY` error from trying to re-insert the existing suppression record, never reaching the dispatch that would have set `email_disabled=true`. The cycle repeated indefinitely.

On Tangle, 224 members were in this state.

## Changes

- **`member-bread-service.add()`** now checks the suppression list before creating a member and sets `email_disabled` accordingly, mirroring what `edit()` already does
- **`mailgun-email-suppression-list` `handleEvent`** restructured so `EmailSuppressedEvent` dispatches even when `Suppression.add()` throws `ER_DUP_ENTRY` — this self-heals any members who drifted into the broken state, correcting them on their next bounce/complaint. Non-`ER_DUP_ENTRY` errors still skip the dispatch.

## Why the self-heal matters

No backfill migration needed — the 224 affected members on Tangle (and any equivalents on other Pro sites) will self-correct on the next newsletter send: Mailgun generates the 607, the handler now dispatches `EmailSuppressedEvent` despite the duplicate suppression record, and `email_disabled` gets flipped to `true`.

## Test plan

- [x] Added unit tests for `add()` suppression check (both suppressed and clean cases)
- [x] Added unit tests for `edit()` suppression check (including the case where email isn't being changed)
- [x] Added integration regression test for ONC-1640: pre-creates a suppression record, fires a 605 event for a member with `email_disabled=false`, verifies the member is disabled via the model (not the API, which would mask the bug)
- [ ] CI passes
- [ ] Manual verification on staging

ref ONC-1640

### DIFF
--- a/ghost/core/core/server/services/members/members-api/services/member-bread-service.js
+++ b/ghost/core/core/server/services/members/members-api/services/member-bread-service.js
@@ -348,6 +348,11 @@ module.exports = class MemberBREADService {
         let model;
 
         try {
+            if (data.email) {
+                const isSuppressed = (await this.emailSuppressionList.getSuppressionData(data.email))?.suppressed;
+                data.email_disabled = !!isSuppressed;
+            }
+
             const attribution = await this.memberAttributionService.getAttributionFromContext(options?.context);
             if (attribution) {
                 data.attribution = attribution;
--- a/ghost/core/core/server/services/members/members-api/services/member-bread-service.js
+++ b/ghost/core/core/server/services/members/members-api/services/member-bread-service.js
@@ -348,7 +348,7 @@ module.exports = class MemberBREADService {
         let model;
 
         try {
-            if (data.email) {
+            if (data.email && data.email_disabled === undefined) {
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/services/members/members-api/services/member-bread-service.js
```diff
@@ -348,6 +348,11 @@ module.exports = class MemberBREADService {
         let model;
 
         try {
+            if (data.email) {
+                const isSuppressed = (await this.emailSuppressionList.getSuppressionData(data.email))?.suppressed;
+                data.email_disabled = !!isSuppressed;
+            }
```
jonatansberg: We might need to check if email_disabled is already set on data so that it is possible to create new members with email_disabled = false even if there are suppression rules

PR #27411, opened 2026-04-15
Added gift subscriptions indexes

mike182uk:
no ref

Added 2 new indexes on the gift subscriptions job to make querying for pending consumption / expired gifts from the clean up job more efficient

### DIFF
--- a/ghost/core/core/server/data/migrations/versions/6.31/2026-04-15-15-18-28-add-gifts-status-indexes.js
+++ b/ghost/core/core/server/data/migrations/versions/6.31/2026-04-15-15-18-28-add-gifts-status-indexes.js
@@ -1,13 +1,13 @@
 const {createNonTransactionalMigration} = require('../../utils');
-const commands = require('../../../schema/commands');
+const {addIndex, dropIndex} = require('../../../schema/commands');
 
 module.exports = createNonTransactionalMigration(
     async function up(knex) {
-        await commands.addIndex('gifts', ['status', 'consumes_at'], knex);
-        await commands.addIndex('gifts', ['status', 'expires_at'], knex);
+        await addIndex('gifts', ['status', 'consumes_at'], knex);
+        await addIndex('gifts', ['status', 'expires_at'], knex);
     },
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/data/migrations/versions/6.31/2026-04-15-15-18-28-add-gifts-status-indexes.js
```diff
@@ -0,0 +1,13 @@
+const {createNonTransactionalMigration} = require('../../utils');
+const commands = require('../../../schema/commands');
+
+module.exports = createNonTransactionalMigration(
+    async function up(knex) {
+        await commands.addIndex('gifts', ['status', 'consumes_at'], knex);
+        await commands.addIndex('gifts', ['status', 'expires_at'], knex);
+    },
+    async function down(knex) {
+        await commands.dropIndex('gifts', ['status', 'consumes_at'], knex);
```
sagzy: Historically, we've added a try/catch around drop index for composite indices

Example: https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/data/migrations/versions/5.75/2023-11-27-15-55-add-members-newsletters-index.js

sagzy (changes requested):
Looks like you forgot to push the schema changes? :)

ghost/core/core/server/data/migrations/versions/6.31/2026-04-15-15-18-28-add-gifts-status-indexes.js
mike182uk: Hm, i think this is only when the index contains a FK to another table? In our case the fields aren't FK's so the error that the try/catch is handling shouldn't happen?

sagzy: Yep I think you're right, I didn't look closely enough at what we were doing in that catch block

sagzy (approved):

PR #27412, opened 2026-04-15
Added endpoints for managing drip campaigns

EvanHahn:
closes https://linear.app/ghost/issue/NY-1229

This adds `GET` and `PUT` `/automated_emails/:id/sequence` to read and write welcome email automations.

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/validators/input/automated_email_sequences.js
+++ b/ghost/core/core/server/api/endpoints/utils/validators/input/automated_email_sequences.js
@@ -1,0 +1,101 @@
+// Filename must match the docName specified in ../../../automated-email-sequences.js
+/* eslint-disable ghost/filenames/match-regex */
+
+const {ValidationError} = require('@tryghost/errors');
+
+const MAX_EMAILS = 10;
+const MAX_DELAY_DAYS = 7;
+
+const messages = {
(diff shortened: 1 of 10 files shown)


ghost/core/core/server/api/endpoints/utils/validators/input/automated_email_sequences.js
```diff
@@ -0,0 +1,101 @@
+// Filename must match the docName specified in ../../../automated-email-sequences.js
+/* eslint-disable ghost/filenames/match-regex */
+
+const {ValidationError} = require('@tryghost/errors');
+
+const MAX_EMAILS = 10;
+const MAX_DELAY_DAYS = 7;
```
troyciesco, line 7: these could totally be placeholder, just want to comment on it before i forget that we should have a product convo about what these values should be soon-ish.

given that these are strictly for welcome sequences they probably _shouldn't_ be too long...but i could see up to even 30 days being reasonable maybe? (e.g. "How did your free month go?", and maybe your sequence is only immediate welcome -> wait 30d -> check-in)

EvanHahn:
This is definitely not going to be merged as is, after we've changed a lot of the designs. Closing. It's still here as a reference if, for some reason, we really need it.

PR #27413, opened 2026-04-15
Update parse-email-address output directory to dist

EvanHahn:
ref https://github.com/TryGhost/Ghost/pull/27018#pullrequestreview-4096070157

Let's use Nx's standard output dir for this package. This fixes an issue with Nx's caching.


EvanHahn:
Giving up on this. See #27018.

PR #27428, opened 2026-04-16
Added `reminder_sent_at` column to `gifts` table

mike182uk:
ref https://linear.app/ghost/issue/BER-3525

Added `reminder_sent_at` column to `gifts` table to track when a reminder has been sent

### DIFF
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -1252,7 +1252,7 @@ module.exports = {
         consumed_at: {type: 'dateTime', nullable: true},
         expired_at: {type: 'dateTime', nullable: true},
         refunded_at: {type: 'dateTime', nullable: true},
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/data/schema/schema.js
```diff
@@ -1252,6 +1252,7 @@ module.exports = {
         consumed_at: {type: 'dateTime', nullable: true},
         expired_at: {type: 'dateTime', nullable: true},
         refunded_at: {type: 'dateTime', nullable: true},
+        reminder_sent_at: {type: 'dateTime', nullable: true},
```
sagzy: naming: `consumption_reminder_sent_at` so that it's more explicit?

mike182uk: Discussed on slack, gone with `consumes_soon_reminder_sent_at`

sagzy (approved):

PR #27430, opened 2026-04-16
Switched to more correct JSDoc types for primitives

EvanHahn:
no ref

This is a types-only change that should have no user impact.

This is, effectively, the whole diff:

```diff
-@param {String} foo
+@param {string} foo
```

`String` is less correct than `string`, and [will not be supported by TypeScript in an upcoming version][0].

All I did was run `sed` on a bunch of files.

[0]: https://devblogs.microsoft.com/typescript/progress-on-typescript-7-december-2025/#javascript-checking-and-jsdoc-compatibility


EvanHahn:
This patch does not contain a migration.

EvanHahn:
Choosing to ignore comments that are existing issues, including spots I missed. (I still think this change is an improvement.)

PR #27436, opened 2026-04-16
 Added a job to remind members before their gift subscription ends

mike182uk:
ref https://linear.app/ghost/issue/BER-3525

 - sends a one-time reminder email to the gift redeemer 7 days before their paid access ends (`consumes_at`)
 - introduces `processReminders` on `GiftService` + daily `send-gift-reminders` worker, gated on the `giftSubscriptions` labs flag
 - skips gifts that consume within the next 3 days (`GIFT_REMINDER_FLOOR_DAYS`) - a reminder with too little time to act on is worse than no reminder
 - records `consumes_soon_reminder_sent_at` inside the locked transaction before sending the email: any duplicate reminder path (e.g. the scheduler integration shipping in a follow-up PR) serialises on the row lock and no-ops, guaranteeing at-most-once delivery at the cost of a possible missed reminder on SMTP failure
 - tier lookup happens before the transaction so a missing tier is a recoverable failure - the next run will pick the gift up again after the admin restores the tier
 - per-gift errors are caught and counted as `failedCount` so a single transient failure doesn't abort the whole batch
 - email-disabled and missing-email redeemers are skipped but still marked as reminded to prevent retry loops


ghost/core/core/server/services/gifts/email-templates/gift-reminder.ts
```diff
@@ -0,0 +1,33 @@
...
+export function renderText(data: GiftReminderData): string {
+    const greeting = data.memberName ? `Hi ${data.memberName},` : 'Hi,';
+
+    return `${greeting}
+
+Your gift subscription to ${data.siteTitle} ends on ${data.gift.consumesAt}.
+
+Gift subscription: ${data.gift.tierName} • ${data.gift.cadenceLabel}
+
+To keep your access, upgrade to a paid subscription before your gift ends:
```
sagzy: copy (suggestion): "To keep your access, _continue_ with a paid subscription before your gift ends"

Reasoning: when you redeem a gift, you become a paid subscriber via a gift. You can then _continue_ with a paid subscription (on the same tier), or _upgrade_ to a higher tier.

ghost/core/core/server/services/gifts/gift-email-service.ts
```diff
@@ -108,6 +117,44 @@ export class GiftEmailService {
         });
     }
 
+    async sendReminder({memberEmail, memberName, tierName, cadence, duration, consumesAt}: ReminderData): Promise<void> {
+        const siteDomain = this.siteDomain;
+        const siteUrl = this.urlUtils.getSiteUrl();
+        const siteTitle = this.settingsCache.get('title') ?? siteDomain;
+
+        const unit = cadence === 'month' ? 'month' : 'year';
+        const cadenceLabel = duration === 1 ? `1 ${unit}` : `${duration} ${unit}s`;
```
sagzy: suggestion: remove `unit`, use `cadence` directly

No idea why agents keep adding this everywhere

ghost/core/core/server/services/gifts/gift-service.ts
```diff
@@ -3,6 +3,26 @@ import logging from '@tryghost/logging';
...
+const MS_PER_DAY = 24 * 60 * 60 * 1000;
+
+/**
+ * Narrow a Bookshelf attribute value (`unknown` from `member.get()`) to `string | null`.
+ * Returns `null` for any value that isn't a string, so callers can safely pass the
+ * result to APIs expecting `string | null`.
+ */
+function asStringOrNull(value: unknown): string | null {
+    return typeof value === 'string' ? value : null;
+}
```
sagzy: suggestion: remove this util, and use either `member.get('email')!` when the domain guarantees a valid value, or `member.get('name') ?? null` where value is a valid string, empty string or null

ghost/core/core/server/services/gifts/gift-service.ts
```diff
@@ -3,6 +3,26 @@ import logging from '@tryghost/logging';
...
+
+/**
+ * Narrow a Bookshelf attribute value to a boolean. Bookshelf may return `0`/`1`
+ * for tinyint columns on some drivers, so coerce defensively.
+ */
+function asBoolean(value: unknown): boolean {
+    return value === true || value === 1;
+}
```
sagzy: suggestion: remove this utility, use `!!` if needed

ghost/core/core/server/services/gifts/gift-service.ts
```diff
@@ -268,10 +314,16 @@ export class GiftService {
                 throw new errors.NotFoundError({message: `Tier not found: ${redeemed.tierId}`});
             }
 
+            const memberEmail = asStringOrNull(member.get('email'));
+
+            if (!memberEmail) {
+                throw new errors.InternalServerError({message: `Redeemer member ${member.id} has no email address`});
+            }
```
sagzy: This shouldn't be possible really, as it's ensured by the domain. Should we use non-null assertion `member.get('email')!`?

ghost/core/core/server/services/gifts/gift-service.ts
```diff
@@ -89,6 +124,17 @@ interface GiftServiceDeps {
...
+type ReminderResult =
+    | {outcome: 'skip'}
+    | {
+        outcome: 'send';
+        memberEmail: string;
+        memberName: string | null;
+        cadence: 'month' | 'year';
+        duration: number;
+        consumesAt: Date;
+    };
```
sagzy: Should we simplify that to:

```
type ReminderResult = 
    | {
        memberEmail: string;
        memberName: string | null;
        cadence: 'month' | 'year';
        duration: number;
        consumesAt: Date;
    }
   | null;
```

Returns a result → success
Return null → failure

sagzy: (the outcome feels unnecessary, as it's essentially only two values: skip or send)

sagzy: This should allow to simplify the code in `sendReminderForGift`

ghost/core/core/server/services/gifts/gift.ts
```diff
@@ -199,4 +203,11 @@ export class Gift {
             expiredAt: new Date()
         });
     }
+
+    remind(): Gift {
```
sagzy: naming: bit on the fence on the simpler `remind()` / `sendReminder()` etc. versus the more explicit but more verbose `remindConsumesSoon()` / `sendConsumesSoonReminder()` etc.

I think the simple version works ok today as we have a single reminder. Part of me still likes the more explicit version though!

Curious what you think. Is the explicit version way too verbose to you?

mike182uk: Hm, i was trying to keep inline with the other methods which are all single word verbs (`consume`, `redeem`, `refund`, `expire`) 

`sendReminder` feels off because the method doesn't actually send anything

The agent originally had `markReminderSent` which is more truthful to what it does, but breaks the convention

sagzy: Sorry previous comment was a bit unclear. The naming comment was more generally about: shall we name all code related to "consumes soon reminder" either:
- explicitely (but more verbosely), with e.g. `sendConsumesSoonReminder`
- or more simply (but less explicitely) with e.g. `sendReminder`

On second thought, I'm thinking let's keep it simple as there is only one reminder on the domain, so `sendReminder` is clear enough

What do you think?

mike182uk: Hm, but nothing is being `sent` here?

ghost/core/core/server/services/gifts/email-templates/gift-reminder.ts
mike182uk: Nice one, updated in https://github.com/TryGhost/Ghost/pull/27436/commits/020b1fb7228ea9e0cd3b5fbafa1d068357715309

ghost/core/core/server/services/gifts/gift-email-service.ts
mike182uk: I've updated this here and in the other methods also → https://github.com/TryGhost/Ghost/pull/27436/commits/61f836e3cad07a3f30e2d721c1d1432f110b8b8c

ghost/core/core/server/services/gifts/gift-service.ts
mike182uk: Have removed this in https://github.com/TryGhost/Ghost/pull/27436/commits/ba94dc8fe035285a5eb25b9f2e1e07da0c223cb7

ghost/core/core/server/services/gifts/gift-service.ts
mike182uk: Yeh theres no specific need for the outcome string, either returning the object or `null` should suffice → https://github.com/TryGhost/Ghost/pull/27436/commits/f13a767d523fdc4aec501485bd2849d596c794ef

ghost/core/core/server/services/gifts/gift-service.ts
mike182uk: Fixed in https://github.com/TryGhost/Ghost/pull/27436/commits/4ec904e605e996ff84e275d4d2030d6e2aeef2d9

ghost/core/core/server/services/gifts/gift-service.ts
mike182uk: Updated in https://github.com/TryGhost/Ghost/pull/27436/commits/35008bbe0f46313b0a29dd12245238c133edb50e

ghost/core/core/server/services/gifts/gift.ts
sagzy: Yep sorry, I commented here, but it was more of a general comment on naming across the changeset.

So I'm saying, either include `consumesSoon` in all the names, or keep it implicit for simplicity:

So this particular method would be:
- `remindConsumesSoon()` (explicit, more verbose)
- `remind()` (implicit, simpler)

sagzy: and on second thought, I prefer the simpler version. So TL;DR: happy with how things are in the PR 👍

PR #27437, opened 2026-04-16
Started scheduling automation polls with scheduler

EvanHahn:
closes https://linear.app/ghost/issue/NY-1191
ref https://github.com/TryGhost/Ghost/pull/27421
ref https://github.com/TryGhost/Ghost/pull/27519

Before, we used `setTimeout` to schedule automation polls (with a TODO).

Now, we properly use the scheduler.


EvanHahn:
@coderabbitai review

ghost/core/test/integration/services/member-welcome-emails.test.js
```diff
@@ -141,30 +141,32 @@ describe('Member Welcome Emails Integration', function () {
...
-            assert.ok(run.welcome_email_automation_id);
-            assert.ok(run.next_welcome_email_automated_email_id);
-            assert.equal(run.step_started_at, null);
-            assert.equal(run.step_attempts, 0);
-            assert.equal(run.exit_reason, null);
-
-            const timestamp = parseDatabaseDate(run.ready_at);
-            assert(timestamp >= before);
-            assert(timestamp <= after);
+            await db.knex.transaction(async (trx) => {
```
EvanHahn, line 144: I recommend reviewing this with whitespace changes disabled. I had to put this in a transaction to avoid the race condition where a member is created and we pick up the run before we can assert on it.

ghost/core/core/boot.js
```diff
@@ -385,7 +388,13 @@ async function initServices() {
...
         statsService.init(),
         explorePingService.init(),
-        giftService.init()
+        giftService.init(),
+        new WelcomeEmailAutomationsService().init({
+            domainEvents,
+            apiUrl,
+            schedulerAdapter,
+            schedulerIntegration
+        })
```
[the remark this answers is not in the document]
EvanHahn: Choosing not to address this.

ghost/core/core/server/api/endpoints/automations.js
```diff
@@ -0,0 +1,19 @@
+const domainEvents = require('@tryghost/domain-events');
+const StartAutomationsPollEvent = require('../../services/welcome-email-automations/events/start-automations-poll-event');
+
+/** @type {import('@tryghost/api-framework').Controller} */
+const controller = {
+    docName: 'automations',
+
+    poll: {
```
troyciesco, line 8: i _think_ this might also need a permissions field? https://github.com/TryGhost/framework/blob/main/packages/api-framework/lib/pipeline.js#L110

it looks like every other endpoint also has one. It seems like endpoints similar to this one use a value of `false` but i'm not 100% sure

troyciesco (changes requested):
requesting changes instead of fixing myself because i'm not completely certain I'm correct, but I wanna call it out sooner than later.

ghost/core/core/server/api/endpoints/automations.js
EvanHahn: Great catch. Added an integration test to avoid this kind of problem in the future, and also fixed the problem. See 691c698e8109eeb069c7e563e6c4900df19004ff.

EvanHahn:
Updated the permissions and added some additional tests.

ghost/core/core/boot.js
troyciesco, line 421: idk if this meets the criteria of "will interfere with tests" but heads up that this will run during tests now

troyciesco (approved):

PR #27438, opened 2026-04-16
Persisted scheduled email analytics fetch and improved debug UI

9larsons:
## Summary

- Persists the scheduled/backfill email analytics fetch to the database so it survives Ghost restarts. Uses the existing `metadata` column on the `jobs` table to store the begin/end time window as JSON; `finished_at` continues to track progress within that window (same as other fetch types), but since the scheduled job has a fixed time window — not computed relative to now — both the start and end times need explicit persistence.
- On startup, `restoreScheduled()` is called once (on the first `startFetch()` cycle) to restore any in-progress schedule from the database.
- Adds custom date range support to the debug UI (`/posts/analytics/:id/debug`). A "Custom Date Range" option now appears alongside the existing "Refetch Analytics" button, with datetime inputs for arbitrary begin/end dates. The API endpoint accepts optional `begin`/`end` parameters, falling back to the existing computed defaults.

## Changes

**Backend:**
- `queries.js` — Added `getJobMetadata()`/`setJobMetadata()`, updated `getJobData()` to include `metadata`
- `email-analytics-service.js` — `schedule()` persists metadata, `cancelScheduled()`/`fetchScheduled()` clear it on completion, new `restoreScheduled()` method
- `email-analytics-service-wrapper.js` — Calls `restoreScheduled()` once on first fetch cycle
- `emails.js` — `scheduleAnalytics` endpoint accepts optional `begin`/`end` params

**Frontend:**
- `debug.js` — Custom date range state and actions
- `debug.hbs` — "Custom Date Range" toggle with datetime-local inputs

## Test plan

- [ ] Start Ghost, navigate to a post's debug page, schedule a fetch with default dates — verify it appears in the Overview tab
- [ ] Restart Ghost — verify the scheduled fetch is restored and continues processing
- [ ] Cancel a scheduled fetch — verify it stops and metadata is cleared
- [ ] Use custom date range to schedule — verify custom dates are sent and used
- [ ] Schedule with only one custom date (begin or end) — verify the other falls back to default

### DIFF
--- a/ghost/admin/app/components/posts/debug.hbs
+++ b/ghost/admin/app/components/posts/debug.hbs
@@ -375,10 +375,10 @@
                         <td>Last event time:</td>
                         <td>{{ this.analyticsStatus.missing.lastEventTimestamp }}</td>
                     </tr>
+                    <tr>
+                        <td colspan="2"><hr></td>
+                    </tr>
                     {{#if (and this.analyticsStatus this.analyticsStatus.scheduled this.analyticsStatus.scheduled.schedule) }}
-                        <tr>
-                            <td colspan="2"><hr></td>
-                        </tr>
                         <tr>
                             <td>Analytics Scheduled running:</td>
                             <td class="gh-email-debug-settings-icon">
@@ -409,11 +409,29 @@
                             </tr>
                         {{/unless}}
                     {{else}}
-                        <tr>
-                            <td colspan="2">
-                                <button type="button" class="gh-email-debug-schedule-analytics" {{on "click" this.scheduleAnalytics }}>{{svg-jar "reload"}}Refetch Analytics</button>
-                            </td>
-                        </tr>
+                        {{#if this.showCustomSchedule}}
+                            <tr>
+                                <td><label for="custom-begin-date">Begin:</label></td>
+                                <td><input id="custom-begin-date" type="datetime-local" value={{this.customBeginDate}} {{on "input" this.updateCustomBeginDate}} /></td>
+                            </tr>
+                            <tr>
+                                <td><label for="custom-end-date">End:</label></td>
+                                <td><input id="custom-end-date" type="datetime-local" value={{this.customEndDate}} {{on "input" this.updateCustomEndDate}} /></td>
+                            </tr>
+                            <tr>
+                                <td colspan="2">
+                                    <button type="button" class="gh-email-debug-schedule-analytics" {{on "click" this.scheduleAnalytics }}>{{svg-jar "reload"}}Schedule Custom Refetch</button>
+                                    <button type="button" class="gh-email-debug-schedule-analytics" {{on "click" this.toggleCustomSchedule }}>Cancel</button>
+                                </td>
+                            </tr>
+                        {{else}}
+                            <tr>
+                                <td colspan="2">
+                                    <button type="button" class="gh-email-debug-schedule-analytics" {{on "click" this.scheduleAnalytics }}>{{svg-jar "reload"}}Refetch Analytics</button>
+                                    <button type="button" class="gh-email-debug-schedule-analytics" {{on "click" this.toggleCustomSchedule }}>Custom Date Range</button>
+                                </td>
+                            </tr>
+                        {{/if}}
                     {{/if}}
                 </tbody>
             </table>
--- a/ghost/admin/app/components/posts/debug.hbs
+++ b/ghost/admin/app/components/posts/debug.hbs
@@ -412,11 +412,11 @@
                         {{#if this.showCustomSchedule}}
                             <tr>
                                 <td><label for="custom-begin-date">Begin:</label></td>
-                                <td><input id="custom-begin-date" type="datetime-local" value={{this.customBeginDate}} {{on "input" this.updateCustomBeginDate}} /></td>
+                                <td><input id="custom-begin-date" type="datetime-local" value={{this.customBeginDate}} {{on "input" this.updateCustomBeginDate}} style="border: 1px solid #ddd; padding: 4px 8px; border-radius: 4px;" /></td>
                             </tr>
                             <tr>
                                 <td><label for="custom-end-date">End:</label></td>
-                                <td><input id="custom-end-date" type="datetime-local" value={{this.customEndDate}} {{on "input" this.updateCustomEndDate}} /></td>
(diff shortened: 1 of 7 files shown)


ghost/admin/app/components/posts/debug.hbs
EvanHahn, line 1: I didn't review this closely. Looks good at a glance.

ghost/core/core/server/api/endpoints/emails.js
```diff
@@ -158,12 +158,23 @@ const controller = {
...
         async query(frame) {
-            const model = await models.Email.findOne(frame.data, frame.options);
-            return emailAnalytics.service.schedule({
-                begin: model.get('created_at'),
-                end: new Date(Math.min(Date.now() - 60 * 60 * 1000, model.get('created_at').getTime() + 24 * 60 * 60 * 1000 * 7))
-            });
+            const {begin: beginParam, end: endParam, ...findOptions} = frame.options;
+            const model = await models.Email.findOne(frame.data, findOptions);
+
+            // Use custom dates if provided, otherwise compute defaults from the email
```
EvanHahn: nit: I think it's clear without this comment.

```suggestion
```

ghost/core/core/server/services/email-analytics/lib/queries.js
```diff
@@ -127,6 +127,48 @@ module.exports = {
...
+            if (result === 0 && metadata) {
+                await db.knex('jobs').insert({
+                    id: new ObjectID().toHexString(),
+                    name: jobName,
+                    metadata: value,
+                    updated_at: new Date().toISOString(),
+                    created_at: new Date().toISOString(),
+                    status: 'queued'
+                });
+            }
```
EvanHahn: nit: I think this pair of things should be done in a single transaction.

ghost/core/core/server/services/email-analytics/lib/queries.js
```diff
@@ -127,6 +127,48 @@ module.exports = {
...
+    async setJobMetadata(jobName, metadata) {
+        try {
+            const value = metadata ? JSON.stringify(metadata) : null;
+            const result = await db.knex('jobs').update({metadata: value, updated_at: new Date()}).where('name', jobName);
+            if (result === 0 && metadata) {
+                await db.knex('jobs').insert({
+                    id: new ObjectID().toHexString(),
+                    name: jobName,
+                    metadata: value,
+                    updated_at: new Date().toISOString(),
```
EvanHahn: question: I've seen inconsistent usage across Ghost here. Should `updated_at` be null here, because it hasn't technically been updated?

ghost/core/core/server/services/email-analytics/email-analytics-service.js
```diff
@@ -253,6 +257,46 @@ module.exports = class EmailAnalyticsService {
...
+            this.queries.setJobMetadata('email-analytics-scheduled', null);
+        }
+    }
+
+    /**
+     * Restores a previously persisted scheduled fetch from the database.
+     * Called once on startup to resume an in-progress backfill after a restart.
+     */
+    async restoreScheduled() {
+        if (this.#fetchScheduledData && this.#fetchScheduledData.schedule) {
```
EvanHahn: There's a race condition here.

restoreScheduled() is called
this.queries.getJobData('email-analytics-scheduled') starts (but doesn't complete)
restoreScheduled() is called again

In addition, if there's no job data, we return and never set #fetchScheduledData.

Not sure if this is a problem in a wider context, but there's an opportunity for a bug here.

Personally, I'd remove this whole check and just assert that callers avoid calling this twice.

ghost/core/core/server/services/email-analytics/email-analytics-service.js
```diff
@@ -253,6 +257,46 @@ module.exports = class EmailAnalyticsService {
                     jobName: 'email-analytics-scheduled'
                 };
             }
+            this.queries.setJobMetadata('email-analytics-scheduled', null);
```
EvanHahn: question: Is there any way we can update this to avoid having to remember to call `this.queries.setJobMetadata('email-analytics-scheduled', null)` in all the relevant spots? Maybe not.

ghost/admin/app/components/posts/debug.js
```diff
@@ -293,6 +296,27 @@ export default class Debug extends Component {
...
 
+    @action
+    toggleCustomSchedule() {
+        this.showCustomSchedule = !this.showCustomSchedule;
+        if (this.showCustomSchedule && !this.customBeginDate) {
+            this.customBeginDate = moment(this.email?.createdAtUTC).format('YYYY-MM-DDTHH:mm');
+            const createdAt = moment(this.email?.createdAtUTC);
+            const maxEnd = moment.min(moment().subtract(1, 'hour'), createdAt.clone().add(7, 'days'));
+            this.customEndDate = maxEnd.format('YYYY-MM-DDTHH:mm');
+        }
```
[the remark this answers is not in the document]
EvanHahn: Seems right?

ghost/admin/app/components/posts/debug.js
```diff
@@ -311,8 +335,19 @@ export default class Debug extends Component {
...
         let statsUrl = this.ghostPaths.url.api(`/emails/${this.post.email.id}/analytics`);
+        if (this.customBeginDate || this.customEndDate) {
+            const params = new URLSearchParams();
+            if (this.customBeginDate) {
+                params.set('begin', new Date(this.customBeginDate).toISOString());
+            }
+            if (this.customEndDate) {
+                params.set('end', new Date(this.customEndDate).toISOString());
+            }
+            statsUrl = `${statsUrl}?${params.toString()}`;
```
EvanHahn: nit: This is probably fine, but if `statsUrl` already has query params, this will fail.

EvanHahn (approved):
Several non-blocking comments.

ghost/core/core/server/services/email-analytics/email-analytics-service.js
```diff
@@ -234,6 +234,10 @@ module.exports = class EmailAnalyticsService {
                 end
             }
         };
+        this.queries.setJobMetadata('email-analytics-scheduled', {
+            begin: begin.toISOString(),
+            end: end.toISOString()
+        });
```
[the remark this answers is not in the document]
9larsons: Let's not worry about this here.

ghost/admin/app/components/posts/debug.js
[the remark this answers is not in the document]
9larsons: Handled.

PR #27439, opened 2026-04-16
🐛 Fixed admin deep-link redirect trailing slash and aligned dev/prod

rob-ghost:
Opening a Ghost admin deep link such as `/ghost/members/import` was meant to redirect to the hash-routed SPA URL `/ghost/#/members/import`. In production that flow first passed through `prettyUrls`, which appended a trailing slash before the deep-link redirect middleware ran. The browser therefore ended up on `/ghost/#/members/import/`, and React Router's hash routes do not match that form — so the user landed on an empty shell instead of the intended screen. The bug stayed invisible during local development because dev used a bespoke Vite plugin to handle these redirects, and that plugin did not share the Express middleware ordering, so CI was the first place it surfaced.

The fix has two parts. First, in `ghost/core/core/server/web/admin/app.js` the `redirectAdminUrls` middleware is now mounted before `prettyUrls`, so the captured path no longer picks up a trailing slash. `redirect-admin-urls.js` also defensively trims a trailing slash (handling the case where a query string is present, or a future ordering change leaks a slash through) before it builds the hash URL. Two regression tests cover the trailing-slash cases, and all six tests in that file pass.

The second part removes the dev/prod divergence that let the bug hide. The Vite dev server now serves the admin from its own `/__admin-dev__/*` prefix; the bespoke `vite-deep-links.ts` plugin is deleted; and the Caddy dev gateway rewrites the `/ghost` and `/ghost/` HTML entry to Vite while letting `/ghost/*` deep-link paths fall through to Ghost's Express middleware. That means dev and prod now share exactly one redirect implementation and one set of URL semantics, so a regression like this can't sneak past local testing again. A new Playwright test in `e2e/tests/admin/deep-links.test.ts` asserts on the `Location` header using a raw request context, which keeps it robust across both dev and build modes.

Production build output and CDN paths are unchanged — only the dev gateway's internal routing and the ordering of two Express middlewares move.


jonatansberg (commented):
Something about the `/__admin-dev__` prefix gives me an slight ick. Outside of that (pointless) aesthetic concern having a separate prefix seems like a sensible approach to simplify this mess! I just have to get over my own hangups 😂   

There a few relevant rabbit comments that you might want to look at, then I think we should be good to go!

PR #27453, opened 2026-04-18
Added Preview toggle to welcome email editor (version 2)

cmraible:
refs https://github.com/TryGhost/Ghost/pull/27399

This is a small UX change stacked on top of https://github.com/TryGhost/Ghost/pull/27399.

This moves the Subject input back into the Email Content (editor) tab, so that all editing UI surface remains in the "Email Content" tab, and all the previewing UI surface remains in the "Preview" tab. I think this is more intuitive than leaving the body of the email in the editor in the Email Content tab, and the Subject input in the Preview tab.

image
image

### DIFF
--- a/apps/admin-x-settings/package.json
+++ b/apps/admin-x-settings/package.json
@@ -54,6 +54,7 @@
     "@uiw/react-codemirror": "4.25.2",
     "clsx": "2.1.1",
     "lucide-react": "0.577.0",
+    "lexical": "0.13.1",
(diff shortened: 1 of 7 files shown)


cmraible:
Opted to go with https://github.com/TryGhost/Ghost/pull/27399

> TLDR: If we go this route then as an end user I interpret this as the post title and I'd expect it to be hideable and be able to set the subject of the email separately (defaulted to the title), just like in newsletters

PR #27468, opened 2026-04-20
Fixed Stripe Service coming up if billing portal manager errors

sam-lord:
ref ONC-1656

Prevents the whole Stripe Service from erroring if the billing portal manager encounters an error while trying to update the billing portal configuration. This can happen if the billing portal configuration that Ghost has created has been made the default one by the Stripe Account owner.

### DIFF
--- a/ghost/core/core/server/services/stripe/billing-portal-manager.js
+++ b/ghost/core/core/server/services/stripe/billing-portal-manager.js
@@ -1,3 +1,5 @@
+const logging = require('@tryghost/logging');
+
 const CONFIGURATION_ID_SETTING = 'stripe_billing_portal_configuration_id';
 
 const DEFAULT_FEATURES = {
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/services/stripe/billing-portal-manager.js
```diff
@@ -91,7 +93,13 @@ class BillingPortalManager {
                 const configuration = await this.api.createBillingPortalConfiguration(this.getConfigurationOptions());
                 return configuration.id;
             }
-            throw err;
+
+            logging.error('Failed to update the billing portal configuration', {
```
aileen, line 97: nitpick: if we wanted to monitor those logs, it might be useful to have some further information in them, so we could action on it.

aileen (approved):

ghost/core/core/server/services/stripe/billing-portal-manager.js
sam-lord: I think the error itself should help! That will have a message w plenty of context

PR #27476, opened 2026-04-21
Updated gift subscription token generation

mike182uk:
no ref

Updated gift subscription token generation to use an alphabet without special chars. Also moved generation to gift service instead of having inline in the payment service

### DIFF
--- a/ghost/core/core/server/services/gifts/gift-service.ts
+++ b/ghost/core/core/server/services/gifts/gift-service.ts
@@ -6,9 +6,6 @@ import type {GiftRepository} from './gift-repository';
 import tpl from '@tryghost/tpl';
 import {GIFT_REMINDER_FLOOR_DAYS, GIFT_REMINDER_LEAD_DAYS} from './constants';
 
-const GIFT_TOKEN_ALPHABET = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
-const GIFT_TOKEN_LENGTH = 12;
-
 const MS_PER_DAY = 24 * 60 * 60 * 1000;
 
 const errorMessages = {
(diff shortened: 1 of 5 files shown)


ghost/core/core/server/services/gifts/gift-service.ts
```diff
@@ -1,10 +1,14 @@
+import crypto from 'node:crypto';
 import errors from '@tryghost/errors';
 import logging from '@tryghost/logging';
 import {Gift} from './gift';
 import type {GiftRepository} from './gift-repository';
 import tpl from '@tryghost/tpl';
 import {GIFT_REMINDER_FLOOR_DAYS, GIFT_REMINDER_LEAD_DAYS} from './constants';
 
+const GIFT_TOKEN_ALPHABET = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+const GIFT_TOKEN_LENGTH = 12;
```
sagzy: nit: could inline those in `generateToken()` as not used anywhere else

ghost/core/core/server/services/gifts/gift-service.ts
```diff
@@ -124,6 +128,16 @@ export class GiftService {
         this.deps = deps;
     }
 
+    generateToken(): string {
```
sagzy, line 128: Might be worth adding an inline comment on the entropy. From Claude calculations:

```
Combinations: 62¹² ≈ 3.23 × 10²¹ (~3.23 sextillion)
Entropy: 12 × log₂(62) ≈ ~71.45 bits
```

ghost/core/core/server/services/gifts/gift-service.ts
sagzy: (you can then also more easily drop the _token_ from the name to calm down bots :p)

mike182uk: Good shout, moved inline: https://github.com/TryGhost/Ghost/pull/27476/commits/944afd59dcf7e3d2b62dcbc7cb31a1ae7d304005

ghost/core/core/server/services/gifts/gift-service.ts
mike182uk: Cool, added comment: https://github.com/TryGhost/Ghost/pull/27476/commits/3b1038fb08f52f3bae4c79ccc99980c734271dc3

sagzy (approved):

PR #27477, opened 2026-04-21
feat: update excerpt length to 2000 chars

azan-n:

- Why are you making it?
The database already supports upto 2000 chars on the excerpt.

- What does it do?

Allows 2000 characters to be set on the excerpt.

When migrating from other platforms, I have found many WP users have long excerpts. 

### DIFF
--- a/ghost/admin/app/validators/post.js
+++ b/ghost/admin/app/validators/post.js
@@ -61,8 +61,8 @@ export default BaseValidator.create({
     },
 
     customExcerpt(model) {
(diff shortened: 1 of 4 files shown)


azan-n:
closes #27478

azan-n:
I'm happy to work on this further once I get a go ahead from the admins. Thank you!

9larsons:
We don't intend to wholesale expand the excerpt length. The purpose is to keep it reasonably short because of the posts API payload size for large pulls. If anything, we could allow this to be configurable, but I'd still question the need.

PR #27483, opened 2026-04-21
Updated share button in newsletters to show on paid/member posts, and not show on email-only

troyciesco:
no ref

Some small tweaks to the visibility of the share button when it's toggled on for newsletters:
- if it's a paid/members-only post, we want to show the share button. idea is that shares of that post can drive help drive conversion for more members
- if it's an email-only newsletter, do _not_ show the share button. idea is an email-only newsletter isn't meant to be shared as a link

paid preview:
image
paid email:
image
member email:
image

email only:
image


EvanHahn (approved):
LGTM assuming this is the product direction we want.

PR #27486, opened 2026-04-21
🐛 Fixed appearance of some welcome email image captions

EvanHahn:
closes https://linear.app/ghost/issue/NY-1233
ref https://github.com/TryGhost/Ghost/pull/27485

| Before | After |
|---|---|
| Before screenshot | After screenshot |

Welcome emails use `<figure>` and `<figcaption>` for images and their captions, which some email clients don't support. That causes the styling to be messed up in those clients.

[Newsletter rendering had already fixed this.][0] This patch fixes it the same way.

In the long term, we should unify these renderers. In the short term, let's fix this bug.

[0]: https://github.com/TryGhost/Ghost/blob/d26bfdb0b20f029a82709782804164d621848d3f/ghost/core/core/server/services/email-service/email-renderer.js#L514-L544


ghost/core/core/server/services/email-rendering/finalize.js
```diff
@@ -1,10 +1,39 @@
...
+    // and <figcaption>][0]. To work around this, change the tags to <div>s.
+    // Juice should have properly styled them.
+    //
+    // [0]: https://www.caniemail.com/features/html-semantics/
+    $ = cheerio.load(juicedHtml, null, false);
+    $('figure, figcaption').each((_, el) => {
+        el.tagName = 'div';
+    });
+
+    return $.html();
```
EvanHahn, line 27: Using Cheerio has the side-effect of changing the way HTML entities are encoded. For example, `&copy;` becomes `&#xA9;`. These are equivalent. That's part of why I [recently changed the tests to be resilient to this](https://github.com/TryGhost/Ghost/pull/27485).

ghost/core/core/server/services/email-rendering/finalize.js
```diff
@@ -1,10 +1,39 @@
...
 const htmlToPlaintext = require('@tryghost/html-to-plaintext');
 
+/**
+ * @param {string} html
+ * @returns {string}
+ */
+const finalizeHtml = (html) => {
+    // Add a class to each figcaption so we can style them in the email.
+    let $ = cheerio.load(html, null, false);
+    $('figcaption').addClass('kg-card-figcaption');
```
EvanHahn, line 12: Lifted from the newsletter renderer: https://github.com/TryGhost/Ghost/blob/74bdef061f120f3f35362dcc30f0e24ef7f8389a/ghost/core/core/server/services/email-service/email-renderer.js#L515-L517

ghost/core/core/server/services/email-rendering/finalize.js
```diff
@@ -1,10 +1,39 @@
...
+
+    // Many email clients, like Outlook and Yahoo, [lack support for <figure>
+    // and <figcaption>][0]. To work around this, change the tags to <div>s.
+    // Juice should have properly styled them.
+    //
+    // [0]: https://www.caniemail.com/features/html-semantics/
+    $ = cheerio.load(juicedHtml, null, false);
+    $('figure, figcaption').each((_, el) => {
+        el.tagName = 'div';
+    });
```
EvanHahn, line 25: Lifted from the newsletter renderer: https://github.com/TryGhost/Ghost/blob/74bdef061f120f3f35362dcc30f0e24ef7f8389a/ghost/core/core/server/services/email-service/email-renderer.js#L544-L545

EvanHahn:
@coderabbit review

ghost/core/core/server/services/email-rendering/finalize.js
```diff
@@ -1,10 +1,39 @@
+const cheerio = require('cheerio');
```
troyciesco, line 1: i think cheerio might be breaking apostrophes, maybe more (that we account for in the [newsletter email renderer](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/services/email-service/email-renderer.js#L569) so maybe another point in favor of making these 1:1)

this is the local code in outlook 365, windows 11 - the web ones seem fine
Image

---

on a production site:
Image

troyciesco (changes requested):

ghost/core/core/server/services/email-rendering/finalize.js
EvanHahn: Great catch. Fixed in 9380f55c209ab9bd20e492b668cda1fa2ee11acf. I manually tested this on 32 different Parcel screenshots and it all looks good.

troyciesco (approved):
up to you if you care about Sonar's recommendation for `replaceAll` (I don't really, since `replace` matches the other email renderer)

PR #27491, opened 2026-04-21
WIP: automated email endpoints handling multiple automated emails

EvanHahn:
*I don't think we actually want this, but publishing just in case we do.*


EvanHahn:
Don't think we ever plan to merge this as is.

PR #27507, opened 2026-04-22
Optimize redis cache miss coalescing

vershwal:
Concurrent requests for the same cache key that all encounter a miss each independently call `fetchData()`, causing a "thundering herd" — N identical queries to the underlying data source when only 1 is needed. This adds request coalescing so that `fetchData()` runs exactly once per cache miss, and all concurrent callers share the same result.

### What changed

- Added a `currentlyExecutingReads` Map to `AdapterCacheRedis` that tracks in-flight fetch promises per key
- On a cache miss, the first caller stores its `fetchData()` promise in the Map; subsequent callers for the same key return the existing promise instead of spawning a new fetch
- The promise is cleaned up via `.finally()` so the next request after completion goes through normally
- Errors from `fetchData()` are caught and logged within the promise chain (matching the existing error-swallowing behavior), so all concurrent callers receive `undefined` and the key is freed for retry on the next request

### Why coalescing is scoped to cache misses only

An earlier approach ([#19627](https://github.com/TryGhost/Ghost/pull/19627)) coalesced the *entire* `get()` flow — including the Redis read, TTL check, and refresh-ahead logic. This means even cache hits were serialized: if 10 requests arrived concurrently for a cached key, 9 of them would wait for the first one's full Redis round-trip + TTL check to complete before returning.

Cache hits are the fast path and don't benefit from serialization — Redis handles concurrent reads efficiently. The expensive part is `fetchData()` (the database or API call behind the cache), and that only runs on a miss. By scoping coalescing to the miss branch:

- **Cache hits stay fast** — each request reads from Redis independently, no artificial queuing
- **Cache misses are protected** — `fetchData()` runs once, all waiters share the result
- **No extra private method needed** — the change stays within the existing `get()` method

### Why the Map is keyed by `internalKey` (not the external `key`)

The cache uses prefix-hash rotation for invalidation — `reset()` changes the prefix hash so all existing keys become invisible. If the coalescing Map were keyed by the external key (e.g. `'foo'`), a `reset()` during an in-flight `fetchData()` would cause two bugs:

1. **Incorrect coalescing** — a post-reset caller looks up `'foo'`, gets a new `internalKey` (`prefix:newHash:foo`), misses, but finds the pre-reset promise under `'foo'` and joins it instead of fetching fresh data
2. **Stale write to new generation** — the original `this.set(key, data)` call re-derives the internal key at write time using the *new* prefix hash, so stale data from the pre-reset fetch lands in the new cache generation

Keying by `internalKey` (which includes the prefix hash) fixes both: after a reset, a new caller gets a different `internalKey` so it won't join the stale promise, and the pre-reset writeback targets the old generation key directly via `this.cache.set(internalKey, data)` — a harmless orphan that nobody will read.

When `internalKey` is null (timeout or Redis failure during lookup), coalescing is skipped entirely — the degraded path doesn't need this protection.

### Unit tests (`adapter-cache-redis.test.js`)
- [x] Concurrent cache misses call `fetchData` only once, all callers receive the value
- [x] `fetchData` rejection propagates to all concurrent callers (all get `undefined`, error logged once)
- [x] After a coalesced fetch rejection, the Map is cleaned up and the next call retries successfully
- [x] Does not coalesce fetches across a `prefix_hash` cycle (reset)
- [x] Concurrent misses for different keys fetch independently (coalescing is per-key, not global)

### Integration test (`adapter-cache-redis.test.js`)
- [x] Concurrent cache misses against real Redis call `fetchData` only once, subsequent cached read also works

### DIFF
--- a/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
+++ b/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
@@ -65,6 +65,7 @@ class AdapterCacheRedis extends BaseCacheAdapter {
         this.refreshAheadFactor = config.refreshAheadFactor || 0;
         this.getTimeoutMilliseconds = config.getTimeoutMilliseconds || null;
         this.currentlyExecutingBackgroundRefreshes = new Set();
+        this.currentlyExecutingReads = new Map();
         this._keyPrefix = config.keyPrefix || '';
         this._prefixHashInitInFlight = null;
         this.redisClient.on('error', this.handleRedisError);
@@ -240,9 +241,19 @@ class AdapterCacheRedis extends BaseCacheAdapter {
                 }
                 return result;
             } else {
-                const data = await fetchData();
-                await this.set(key, data); // We don't use `internalKey` here because `set` handles it
-                return data;
+                if (this.currentlyExecutingReads.has(key)) {
+                    return this.currentlyExecutingReads.get(key);
+                }
+                const fetchPromise = fetchData().then(async (data) => {
+                    await this.set(key, data);
+                    return data;
+                }).catch((err) => {
+                    logging.error(err);
+                }).finally(() => {
+                    this.currentlyExecutingReads.delete(key);
+                });
+                this.currentlyExecutingReads.set(key, fetchPromise);
+                return fetchPromise;
             }
         } catch (err) {
             logging.error(err);
--- a/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
+++ b/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
@@ -241,32 +241,30 @@ class AdapterCacheRedis extends BaseCacheAdapter {
                 }
                 return result;
             } else {
-                if (internalKey && this.currentlyExecutingReads.has(internalKey)) {
+                if (!internalKey) {
+                    return fetchData();
+                }
+                if (this.currentlyExecutingReads.has(internalKey)) {
                     return this.currentlyExecutingReads.get(internalKey);
                 }
-                const fetchPromise = fetchData().then(async (data) => {
-                    if (internalKey) {
-                        try {
-                            debug('set', internalKey);
-                            await this.cache.set(internalKey, data);
-                        } catch (err) {
-                            logging.error(err);
-                        }
-                    } else {
-                        await this.set(key, data);
-                    }
-                    return data;
-                }).catch((err) => {
+                const fetchPromise = fetchData();
+                const resultPromise = fetchPromise.catch((err) => {
                     logging.error(err);
-                }).finally(() => {
-                    if (internalKey) {
-                        this.currentlyExecutingReads.delete(internalKey);
-                    }
                 });
-                if (internalKey) {
-                    this.currentlyExecutingReads.set(internalKey, fetchPromise);
-                }
-                return fetchPromise;
+                fetchPromise.then(async (data) => {
+                    try {
+                        debug('set', internalKey);
+                        await this.cache.set(internalKey, data);
+                    } catch (err) {
+                        logging.error(err);
+                    }
+                }).catch(() => {
+                    // fetchData rejection — already logged by resultPromise
+                }).finally(() => {
+                    this.currentlyExecutingReads.delete(internalKey);
+                });
+                this.currentlyExecutingReads.set(internalKey, resultPromise);
+                return resultPromise;
             }
         } catch (err) {
             logging.error(err);
--- a/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
+++ b/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
@@ -241,18 +241,31 @@ class AdapterCacheRedis extends BaseCacheAdapter {
                 }
                 return result;
             } else {
-                if (this.currentlyExecutingReads.has(key)) {
-                    return this.currentlyExecutingReads.get(key);
+                if (internalKey && this.currentlyExecutingReads.has(internalKey)) {
+                    return this.currentlyExecutingReads.get(internalKey);
                 }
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
```diff
@@ -230,9 +231,32 @@ class AdapterCacheRedis extends BaseCacheAdapter {
...
+                }
+                const fetchPromise = fetchData().then(async (data) => {
+                    if (internalKey) {
+                        try {
+                            debug('set', internalKey);
+                            await this.cache.set(internalKey, data);
+                        } catch (err) {
+                            logging.error(err);
+                        }
+                    } else {
```
allouis: If `internalKey` is falsy, then it means that our connection to redis either errored or it timed out, I wonder if we can take that as a signal to side step talking with Redis?

e.g. 

```ts
if (!internalKey) {
  return fetchData();
}

if (this.currentlyExecutingReads.has(internalKey)) {
  return this.currentlyExecutingReads.get(internalKey);
}

// 
```

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
```diff
@@ -230,9 +231,32 @@ class AdapterCacheRedis extends BaseCacheAdapter {
...
-                await this.set(key, data); // We don't use `internalKey` here because `set` handles it
-                return data;
+                if (internalKey && this.currentlyExecutingReads.has(internalKey)) {
+                    return this.currentlyExecutingReads.get(internalKey);
+                }
+                const fetchPromise = fetchData().then(async (data) => {
+                    if (internalKey) {
+                        try {
+                            debug('set', internalKey);
+                            await this.cache.set(internalKey, data);
```
allouis: This `await` adds some latency to the response, and I think the only reason we need to wait for it is because of the `fetchPromise` coordination

But if we had something like:

```ts
const fetchPromise = fetchData()

fetchPromise.then(async (data) => {
  await this.cache.set(internalKey, data);
}).finally(() => {
  this.currentlyExecutingReads.delete(internalKey);
});

this.currentlyExecutingReads.set(internalKey, fetchPromise);

return fetchPromise;
```

Then the data can be returned to the caller as soon as it's available, and the promise is still only cleared after we've cached it

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
vershwal: Good one, this make it much cleaner. Addressed in https://github.com/TryGhost/Ghost/pull/27507/changes/c4dcff7bf3164861165c4626eac351befa25cc99

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
vershwal: Fair point. Done!!

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
```diff
@@ -230,9 +231,30 @@ class AdapterCacheRedis extends BaseCacheAdapter {
...
-                await this.set(key, data); // We don't use `internalKey` here because `set` handles it
-                return data;
+                if (!internalKey) {
+                    return fetchData();
+                }
+                if (this.currentlyExecutingReads.has(internalKey)) {
+                    return this.currentlyExecutingReads.get(internalKey);
+                }
+                const fetchPromise = fetchData();
+                const resultPromise = fetchPromise.catch((err) => {
```
allouis, line 251: This swallows the error, which is different to the existing funtionality

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
```diff
@@ -230,9 +231,30 @@ class AdapterCacheRedis extends BaseCacheAdapter {
...
+                    logging.error(err);
+                });
+                fetchPromise.then(async (data) => {
+                    try {
+                        debug('set', internalKey);
+                        await this.cache.set(internalKey, data);
+                    } catch (err) {
+                        logging.error(err);
+                    }
+                }).catch(() => {
```
allouis, line 261: If we put the `logging.error` in this catch, we can get rid of `resultPromise` as well as the `try/catch` around `this.cache.set`!

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
vershwal: I think this preserves the existing behaviour. Before this PR, `fetchData` rejections were caught by the outer `try/catch`, logged, and the caller got `undefined`. The `.catch` here does the same: logs and returns `undefined`. It just moved into the promise chain because `return fetchPromise` without `await` skips the outer `try/catch`.

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
vershwal: Dropping `resultPromise` and returning `fetchPromise` directly means `fetchData` rejections reach the caller instead of being swallowed. Before this PR they were swallowed. `resultPromise` keeps that behaviour, and also ensures coalesced callers sharing the same promise from the Map don't get a raw rejection.

allouis (approved):

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
allouis: I missed the outer try catch, you're right :)

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
allouis: Same as above!

PR #27509, opened 2026-04-22
Updated gift subscription reminder email copy and layout

minimaluminium:
no issues

Follow-up to #27436 with a few design tweaks to the new gift reminder email. Also aligns the CTA with the in-progress "Continue subscription" button on the Portal account page.

- Rewrote the body copy and renamed the CTA *"Manage subscription"* → *"Continue subscription"* to match the in-progress Portal account button
- Removed the conditional *"Hi {name},"* greeting so the h1 leads directly
- Replaced the cadence row with a *"Price after gift ends"* row, and wired tier currency + monthly/yearly price through the reminder flow


ghost/core/core/server/services/gifts/gift-service.ts
```diff
@@ -491,6 +492,16 @@ export class GiftService {
...
 
+        // Throw before the transaction so the gift isn't marked as reminded;
+        // the next run recovers after an admin restores pricing.
+        const tierPrice = gift.cadence === 'month' ? tier.monthlyPrice : tier.yearlyPrice;
+
+        if (tierPrice === null || tier.currency === null) {
+            throw new errors.NotFoundError({
+                message: `Tier missing ${gift.cadence}ly pricing for gift: ${gift.tierId}`
+            });
+        }
```
sagzy, line 503: nit: currency/price would be missing only happen if the tier was "Free". We should probably do that check on the gift purchase rather than here, but it also doesn't harm to have an extra check here 👍

ghost/core/test/unit/server/services/gifts/gift-email-service.test.js
```diff
@@ -152,40 +152,48 @@ describe('GiftEmailService', function () {
...
-                sinon.assert.match(msg[field], sinon.match('1 year'));
                 sinon.assert.match(msg[field], sinon.match('23 Apr 2026'));
+                sinon.assert.match(msg[field], sinon.match('$100.00/year'));
                 sinon.assert.match(msg[field], sinon.match('https://example.com/#/portal/account'));
             }
         });
 
-        it('uses a generic greeting when the member has no name', async function () {
-            await service.sendReminder({...reminderData, memberName: null});
+        it('does not include a greeting', async function () {
```
sagzy: suggestion: delete this test entirely instead of testing for something that isn't present

ghost/core/test/unit/server/services/gifts/gift-email-service.test.js
```diff
@@ -152,40 +152,48 @@ describe('GiftEmailService', function () {
...
             const msg = mailer.send.getCall(0).args[0];
 
-            sinon.assert.match(msg.text, sinon.match(/^Hi,/));
+            sinon.assert.match(msg.text, sinon.match(/^Your gift subscription is ending soon/));
+            sinon.assert.match(msg.html, sinon.match(value => !/Hi,|Hi [^,]+,/.test(value)));
+            sinon.assert.match(msg.text, sinon.match(value => !/Hi,|Hi [^,]+,/.test(value)));
         });
 
-        it('includes the member name when provided', async function () {
+        it('uses the new CTA copy "Continue subscription"', async function () {
```
sagzy: nit (naming): 'renders a "Continue subscription" CTA'?

minimaluminium: Fixed in https://github.com/TryGhost/Ghost/pull/27509/commits/abb2596709d0cf579b3e664c8dfd4ba94d1710b4

ghost/core/test/unit/server/services/gifts/gift-email-service.test.js
minimaluminium: Removed in https://github.com/TryGhost/Ghost/pull/27509/commits/abb2596709d0cf579b3e664c8dfd4ba94d1710b4

ghost/core/core/server/services/gifts/gift-service.ts
minimaluminium: 👍 fair point, leaving in as a safety net for now

PR #27511, opened 2026-04-22
Standardize app acceptance tests with matrix builds

acburdine:
no issue
- standardize playwright test command naming in apps
- switch app acceptance tests to run in ci with a matrix build

## Summary

Following on from the changes in #27348, this PR updates the app acceptance tests to use a standardized matrix setup using common script names, instead of needing to use essentially the same job configuration repeated multiple times.


.github/workflows/ci.yml
```diff
@@ -174,15 +157,20 @@ jobs:
           I18N_PROJECTS=$(pnpm -s nx show projects ${AFFECTED_ARG} --projects 'tag:i18n' --sep=, | tr -d '\n')
           echo "affected_i18n_projects=${I18N_PROJECTS}" >> "$GITHUB_OUTPUT"
 
+          # NOTE: remove stats exclusion if/when stats has playwright tests added
+          PLAYWRIGHT_PROJECTS=$(pnpm -s nx show projects ${AFFECTED_ARG} \
+            --withTarget test:acceptance \
+            --exclude @tryghost/stats \
```
acburdine: this bit is here because apps/stats has the scripts/configuration for playwright tests, but no playwright tests exist, causing the job to error unless stats is excluded.

An easier approach would be to just remove the unnecessary playwright tests/configuration from apps/stats if it's not going to have acceptance tests added, but I wasn't sure the long-term plan there so I went with this route.

.github/workflows/ci.yml
```diff
@@ -576,91 +564,15 @@ jobs:
...
-    runs-on: ubuntu-latest
-    needs: [job_setup]
-    if: needs.job_setup.outputs.is_tag == 'true' || needs.job_setup.outputs.changed_comments_ui == 'true'
-    name: Comments-UI tests
+    if: needs.job_setup.outputs.affected_playwright_projects != '[]'
+    name: App Playwright Acceptance Tests
+    strategy:
+      fail-fast: true
+      matrix:
+        app: ${{ fromJSON(needs.job_setup.outputs.affected_playwright_projects) }}
```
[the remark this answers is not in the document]
9larsons: I think I agree with this, though I'm open to the alternative.

.github/workflows/ci.yml
9larsons: I'd say let's go ahead and rip out the playwright tests/config from stats while we're here - if you don't mind - as the goal there is to combine stats and posts with the new React-based admin app.

9larsons (changes requested):
No concerns, this is good cleanup. You had posed a question and I think we should do that, now that you have the added context.

.github/workflows/ci.yml
acburdine: fixed!

acburdine:
@9larsons I fixed all the merge conflicts, any other blockers to getting this merged?

9larsons (approved):
Sorry this sat for so long!

PR #27513, opened 2026-04-22
Fixed Shade Storybook preview hanging on pnpm install

peterzimon:
- admin-x-design-system and signup-form pin @storybook/react-vite@8.6.14, so pnpm hoists that older version into .pnpm/node_modules
- Storybook's CLI NODE_PATH picked up the v8 hoisted copy before shade's own v10.3.3, causing the generated vite-app.js to use the v8 preview template and import nonexistent paths like @storybook/react/dist/entry-preview.mjs
- the preview iframe then failed to load and only the sidebar rendered
- resolve @storybook/react-vite and addon packages via require.resolve from the config dir so the framework and addons come from shade's own v10 tree regardless of what pnpm hoists

### DIFF
--- a/apps/shade/.storybook/main.tsx
+++ b/apps/shade/.storybook/main.tsx
@@ -1,22 +1,34 @@
+import {createRequire} from 'node:module';
 import {dirname} from 'node:path';
 import {fileURLToPath} from 'node:url';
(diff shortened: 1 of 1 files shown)


peterzimon:
Obsolete

PR #27516, opened 2026-04-22
🐛 Prevented Escape keypresses in welcome emails Customize modal from exiting settings

cmraible:
ref https://linear.app/ghost/issue/NY-1234/flaky-member-welcome-emails-e2e-test

## Why this change?

Pressing `Escape` inside the welcome emails "Customize" flow could exit settings instead of just dismissing the active UI layer. That made the interaction flaky in E2E and, more importantly, let a nested control bypass the expected unsaved-changes flow.

## Reproduction
1. Go to Settings > Welcome emails
2. Open the "Customize" modal
3. Go to the Design tab
4. In rapid succession: click on one of the color swatches and immediately hit Escape

## Expected behavior
Escape should close the color picker, but leave the Customize modal open

## Actual behavior
If you time it just right, the Customize modal will close, then Settings will close, and you'll land on the Analytics screen (or where ever you were before opening settings to begin with).

## Root cause

`admin-x-settings` has a page-level `Escape` handler that navigates away from settings when no modal is open.

The welcome email customize UI is built with Shade/Radix dialogs and popovers, but the page-level guard was still only checking for the legacy `#modal-backdrop` element. That created two failure modes:

1. The customize modal and dirty-confirm dialog were not always recognized as "an open modal", so the page-level handler could still react to `Escape`.
6. The color picker popover mounts through a portal. Right after opening it, there is a brief window where the popover has been requested but its content has not mounted yet, so its `onEscapeKeyDown` handler cannot intercept the keypress. In that gap, `Escape` can bubble to the page-level handler and exit settings.

## Why this fix is correct

This fix closes the gap at the right layers instead of weakening the global shortcut.

- `MainContent` now treats open Radix `dialog` and `alertdialog` elements as active modals, not just the legacy backdrop. That keeps the page-level `Escape` shortcut disabled while the customize modal or unsaved-changes confirmation is open.
- `ColorPickerField` attaches an early capture-phase `Escape` listener as soon as the popover is opened. That means the first `Escape` is claimed by the color picker even if the popover content has not mounted yet. Once the popover is mounted, the normal popover escape handling still applies, and the temporary listener is removed when the popover closes or the component unmounts.

Together, those changes preserve the intended dismissal order:

1. First `Escape`: close the nested color picker.
2. Next `Escape`: show or dismiss the customize modal's unsaved-changes confirmation (if dirty).
7. Only when no modal is open should the settings page itself handle `Escape`.

## Test coverage

The PR adds:

- unit coverage for both the mounted-popover case and the immediate-after-open race
- e2e coverage proving `Escape` closes the color picker first and still respects the unsaved-changes confirmation flow

### DIFF
--- a/apps/admin-x-settings/src/main-content.tsx
+++ b/apps/admin-x-settings/src/main-content.tsx
@@ -30,13 +30,21 @@ const MainContent: React.FC = () => {
     const navigateAway = (escLocation: string) => {
         window.location.hash = escLocation;
     };
+    const hasOpenModal = () => {
+        if (document.getElementById('modal-backdrop')) {
+            return true;
+        }
+
+        return Boolean(document.querySelector(
+            '[role="dialog"][data-state="open"], [role="alertdialog"][data-state="open"]'
+        ));
+    };
 
     useEffect(() => {
         const handleKeyDown = (event: KeyboardEvent) => {
             if (event.key === 'Escape') {
                 // Don't navigate away if a modal is open - let the modal handle ESC
-                const modalBackdrop = document.getElementById('modal-backdrop');
-                if (modalBackdrop) {
+                if (hasOpenModal()) {
                     return;
                 }
 
--- a/apps/admin-x-settings/src/main-content.tsx
+++ b/apps/admin-x-settings/src/main-content.tsx
@@ -10,6 +10,7 @@ import {useGlobalData} from './components/providers/global-data-provider';
 import {useRouting} from '@tryghost/admin-x-framework/routing';
 
 const EMPTY_KEYWORDS: string[] = [];
+const OPEN_SHADE_MODAL_SELECTOR = ':is([role="dialog"], [role="alertdialog"])[data-state="open"]';
 
 const Page: React.FC<{children: ReactNode}> = ({children}) => {
     return <>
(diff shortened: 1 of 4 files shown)


e2e/tests/admin/settings/member-welcome-emails.test.ts
```diff
@@ -381,7 +381,7 @@ test.describe('Ghost Admin - Welcome Email Customize Button', () => {
         await expect(page).toHaveURL(/\/ghost\/#\/settings\/memberemails$/);
     });
 
-    test.skip('Escape closes welcome email color picker without bypassing unsaved changes confirmation', async ({page}) => {
```
cmraible, line 384: only changes here are un-skipping this test, and updating the names to follow the convention, as pointed out by coderabbit.

apps/admin-x-settings/src/main-content.tsx
```diff
@@ -30,13 +31,21 @@ const MainContent: React.FC = () => {
...
+        return Boolean(document.querySelector(OPEN_SHADE_MODAL_SELECTOR));
+    };
 
     useEffect(() => {
         const handleKeyDown = (event: KeyboardEvent) => {
             if (event.key === 'Escape') {
                 // Don't navigate away if a modal is open - let the modal handle ESC
-                const modalBackdrop = document.getElementById('modal-backdrop');
-                if (modalBackdrop) {
+                if (hasOpenModal()) {
```
troyciesco, line 48: interesting TIL - i was surprised the linter wasn't yelling about `hasOpenModal` not being in the dep array on L63, but it's because there's no state or reactivity or anything in the function, so it doesn't need to be. 

I'd always thought every function/variable got declared there but I guess not! Probably because like 98% of the time they _do_ have React stuff going on.

troyciesco (approved):

PR #27519, opened 2026-04-22
Created "poll automations" permission for scheduler

EvanHahn:
ref https://linear.app/ghost/issue/NY-1191
ref https://github.com/TryGhost/Ghost/pull/27437

This creates a new permission, "Poll automations", which the Scheduler Integration role gets.

This will be used in [an upcoming change][0]. It's a separate patch because it contains a database migration.

[0]: https://github.com/TryGhost/Ghost/pull/27437


gregoryk07 (approved):

ghost/core/core/server/data/migrations/versions/6.33/2026-04-22-20-53-24-add-automations-poll-permission-to-scheduler-integration.js
```diff
@@ -0,0 +1,13 @@
+const {combineTransactionalMigrations, addPermission, addPermissionToRole} = require('../../utils');
+
+module.exports = combineTransactionalMigrations(
```
troyciesco: There's actually an `addPermissionWithRoles` helper that can be used here which already combines all these functions:
```
const {addPermissionWithRoles} = require('../../utils');

module.exports = addPermissionWithRoles({
    name: 'Poll automations',
    action: 'poll',
    object: 'automation'
}, [
    'Scheduler Integration'
]);
```

- [example](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/data/migrations/versions/5.34/2023-01-30-07-27-add-mentions-permission.js)
- [more recent example](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/data/migrations/versions/6.25/2026-03-30-20-22-25-add-email-design-setting-permissions.js) (but adds multiple so still has `combineTransactionalMigrations`)

what's in this PR works too but i think it'd be the only instance of it in the codebase, so probably best to change it to make it easier for agents/humans later. 

feel free to dismiss "requested changes" if there was a reason for doing it this way tho!

troyciesco (changes requested):

ghost/core/core/server/data/migrations/versions/6.33/2026-04-22-20-53-24-add-automations-poll-permission-to-scheduler-integration.js
EvanHahn: TIL! Done in 7d2932ef3e56fb85fbd207beddf2410d76093490.

troyciesco (approved):

PR #27526, opened 2026-04-23
Updated stats app to include gift members in segmentation

mike182uk:
ref https://linear.app/ghost/issue/BER-3491

- the stats app segmented members as `paid`/`free`/`comped` and ignored the new `gift` status, so `gift` members weren't reflected in analytics KPIs, tooltips, or breakdowns
- `gift` is now counted in the `Total Members` and `Paid Members` KPIs, following the same treatment as `comped` (both grant access to paid content)
- the `Paid Members` tooltip gets a new `Gift` breakdown row, and the `Paid Subscription` Breakdown pie gets a new `Gift` slice - both gated behind the `giftSubscriptions` labs flag so sites not using the feature see no UI change
- also added per-day `gift_delta` to the backend `members-stats` service so the time-series is complete end-to-end - the frontend relies on this for accurate tooltip values and cadence-pie deltas over the selected range

### DIFF
--- a/apps/stats/src/views/Stats/Growth/components/growth-kpis.tsx
+++ b/apps/stats/src/views/Stats/Growth/components/growth-kpis.tsx
@@ -15,6 +15,7 @@ type ChartDataItem = {
     free: number;
     paid: number;
     comped: number;
+    gift: number;
     mrr: number;
     paid_subscribed?: number;
     paid_canceled?: number;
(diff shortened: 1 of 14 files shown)


apps/stats/src/views/Stats/Growth/components/growth-kpis.tsx
```diff
@@ -112,8 +123,9 @@ const GrowthKPIs: React.FC<{
 }> = ({chartData: allChartData, totals, initialTab, currencySymbol, isLoading, onTabChange}) => {
     const validatedInitialTab = isValidTab(initialTab) ? initialTab : 'total-members';
     const [currentTab, setCurrentTab] = useState<KpiTab>(validatedInitialTab);
-    const {range} = useGlobalData();
+    const {range, data: config} = useGlobalData();
     const {appSettings} = useAppContext();
+    const giftSubscriptionsEnabled = config?.labs?.giftSubscriptions === true;
```
sagzy, line 128: nit: there is a `useLabsFlag` hook that could be used here too. But also we're deleting that flag-related code in a week, so doesn't matter :D

sagzy (approved):

PR #27530, opened 2026-04-23
Added cache hit/miss/error/reset metrics

allouis:
ref https://linear.app/ghost/issue/HKG-1712

## Why

There's no way to answer basic questions about the redis cache in
production today: hit rate, reset frequency, error counts, whether
background refresh is working. The adapter only emits `debug()` logs,
which only help when investigating a single request.

## What

Emits metrics via `@tryghost/metrics` at every instrumentation point
in `AdapterCacheRedis`:

- `cache-hit`, `cache-miss` — on every `get`
- `cache-timeout` — when `getTimeoutMilliseconds` fires
- `cache-error` — failures in `get`, `set`, `reset` (with `operation` field)
- `cache-background-refresh-triggered` — when a refresh starts
- `cache-background-refresh-succeeded` — with duration in ms
- `cache-background-refresh-failed` — when the fetch or the cache write throws
- `cache-reset` — with duration

### DIFF
--- a/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
+++ b/ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
@@ -2,6 +2,7 @@ const crypto = require('crypto');
 const BaseCacheAdapter = require('@tryghost/adapter-base-cache');
 const errors = require('@tryghost/errors');
 const logging = require('@tryghost/logging');
+const metrics = require('@tryghost/metrics');
 const debug = require('@tryghost/debug')('redis-cache');
 const cacheManager = require('cache-manager');
 const redisStoreFactory = require('./redis-store-factory');
@@ -22,6 +23,7 @@ class AdapterCacheRedis extends BaseCacheAdapter {
      * @param {number} [config.getTimeoutMilliseconds] - default timeout for cache get operations in *milliseconds*
      * @param {number} [config.refreshAheadFactor] - 0-1 number to use to determine how old (as a percentage of ttl) an entry should be before refreshing it
      * @param {string} [config.keyPrefix] - prefix to use when building a unique cache key, e.g.: 'some_id:image-sizes:'
+     * @param {string} [config.featureName] - name of the cache feature (e.g. 'postsPublic') used for dashboard filtering
      * @param {boolean} [config.reuseConnection] - specifies if the redis store/connection should be reused within the process
      */
     constructor(config) {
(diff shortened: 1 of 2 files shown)


allouis:
@coderabbitai review

allouis:
@coderabbitai review

allouis:
@coderabbitai review

allouis:
@coderabbitai full review

allouis:
@coderabbitai review

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
```diff
@@ -22,6 +23,7 @@ class AdapterCacheRedis extends BaseCacheAdapter {
      * @param {number} [config.getTimeoutMilliseconds] - default timeout for cache get operations in *milliseconds*
      * @param {number} [config.refreshAheadFactor] - 0-1 number to use to determine how old (as a percentage of ttl) an entry should be before refreshing it
      * @param {string} [config.keyPrefix] - prefix to use when building a unique cache key, e.g.: 'some_id:image-sizes:'
+     * @param {string} [config.featureName] - name of the cache feature (e.g. 'postsPublic') used for dashboard filtering. Falls back to keyPrefix with any trailing ':' removed.
```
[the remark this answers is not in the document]
allouis: fixed!

ghost/core/core/server/adapters/lib/redis/AdapterCacheRedis.js
```diff
@@ -67,10 +69,24 @@ class AdapterCacheRedis extends BaseCacheAdapter {
         this.currentlyExecutingBackgroundRefreshes = new Set();
         this.currentlyExecutingReads = new Map();
         this._keyPrefix = config.keyPrefix || '';
+        this._featureName = config.featureName;
```
vershwal, line 72: We don't have `featureName` set on any of the cache configs today (`postsPublic`, `tagsPublic`, `linkRedirectsPublic`, `imageSizes`). Are we planning to add it as part of this change?
The PR description also mentions:

> falling back to keyPrefix with any trailing : stripped

…but I see the fallback was intentionally dropped in the CodeRabbit thread, and the existing test `omits feature when featureName is not set` actually asserts that no fallback runs.
Without the `featureName`, every emission in production will go out without the feature label, which means we can't distinguish `postsPublic` vs `tagsPublic` vs the others on the dashboard(not great for our visibility goal)

Even if we set `featureName` explicitly in config, I'd lean toward having some kind of safety net so a missing config never silently produces unlabelled metrics. Maybe throwing on construction if it's not provided. Curious what you're thinking?

allouis: Ah yeah, the description wasn't updated after the CodeRabbit thread - I need to stop including implementation specifics in descriptions!

---

> Without the featureName, every emission in production will go out without the feature label, which means we can't distinguish postsPublic vs tagsPublic vs the others on the dashboard(not great for our visibility goal)

IMO These metrics are valuable without breaking them down into specific features - we will still have visbility on misses/hits/errors etc...

RE: throwing if it's not provided, I'm not a fan of this - this would be a breaking change for anyone using the Redis cache, and would also make rolling this out mean that we need coordination with Pro!

My plan was to provide the `featureName` via config from Zuul, so we can ship this first, and then follow up with the config

vershwal: If we provide the `featureName` via config from Zuul, then I think that's good enough

vershwal (approved):

PR #27532, opened 2026-04-23
Handled gift subscriptions in member CSV import / export

mike182uk:
ref https://linear.app/ghost/issue/BER-3483

- Gift members previously exported as `free` members (`status` = `gift` was not mapped to any CSV column), so a round-trip silently converted them
- Added a `gift_id` column to the members CSV export, populated for members with an active (redeemed, non-consumed, non-refunded) gift attached
- On import, `gift_id` reassigns the gift to the row's member via `GiftService.reassignRedeemer()`: locks the gift, sets the member's `status` to `gift`, and applies the gift's tier with `expiry_at` = `consumes_at`
- Did not expose `gift_id` as a user-mappable field in the import UI; the importer force-recognises the column so exported CSVs round-trip regardless of the user's mapping
- `gift_id` is mutually exclusive with `import_tier` and `complimentary_plan` - combining them produces a row-level error
- `reassignRedeemer()` locks the destination member with `forUpdate` and rejects rows where `status` === `'paid'`, so a CSV row can't silently change a paying member onto the gift tier
- Added a `'missing-consumes-at'` guard in `Gift.checkReassignable()` to refuse reassignment when a redeemed gift has no `consumes_at` (would propagate `expiry_at: null`)
- Intentionally kept the CSV surface minimal (`gift_id only`, no buyer/cadence/duration metadata) — lifecycle data remains queryable via the gifts table

### DIFF
--- a/e2e/tests/admin/members/export.test.ts
+++ b/e2e/tests/admin/members/export.test.ts
@@ -16,7 +16,8 @@ const DOWNLOADED_CONTENT_FIELDS = [
     'created_at,',
     'deleted_at,',
     'labels,',
-    'tiers'
+    'tiers,',
+    'gift_subscription'
 ];
 
 const MEMBERS_FIXTURE = [
--- a/e2e/tests/admin/members/export.test.ts
+++ b/e2e/tests/admin/members/export.test.ts
@@ -17,7 +17,7 @@ const DOWNLOADED_CONTENT_FIELDS = [
     'deleted_at,',
     'labels,',
     'tiers,',
-    'gift_subscription'
+    'gift_id'
 ];
 
 const MEMBERS_FIXTURE = [
--- a/ghost/core/core/server/api/endpoints/utils/serializers/output/members.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/output/members.js
@@ -37,7 +37,7 @@ const CSV_HEADERS = [
     'deleted_at',
     'labels',
     'tiers',
-    'gift_subscription'
+    'gift_id'
 ];
 
 /**
@@ -64,9 +64,9 @@ function formatMemberForCSV(member) {
     // Only comped = true should result in 'true', otherwise empty string
     const complimentaryPlan = member.comped === true ? 'true' : '';
 
-    // Convert boolean 'false' to empty string for tests to pass
-    // Only gift_subscription = true should result in 'true', otherwise empty string
-    const giftSubscription = member.gift_subscription === true ? 'true' : '';
+    // Gift members carry the gift id so an exported CSV can be re-imported and reassigned
+    // back to a (possibly new) member record via the gifts table
+    const giftId = member.gift_id || '';
 
     // Convert subscribed boolean to string representation
     const subscribedToEmails = member.subscribed === true ? 'true' : 'false';
@@ -83,7 +83,7 @@ function formatMemberForCSV(member) {
         deleted_at: member.deleted_at || null,
         labels: labels,
         tiers: tiers,
-        gift_subscription: giftSubscription
+        gift_id: giftId
     };
 }
 
--- a/ghost/core/core/server/api/endpoints/utils/serializers/output/members.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/output/members.js
@@ -36,7 +36,8 @@ const CSV_HEADERS = [
     'created_at',
     'deleted_at',
     'labels',
-    'tiers'
+    'tiers',
+    'gift_subscription'
 ];
 
 /**
@@ -62,7 +63,11 @@ function formatMemberForCSV(member) {
     // Convert boolean 'false' to empty string for tests to pass
     // Only comped = true should result in 'true', otherwise empty string
     const complimentaryPlan = member.comped === true ? 'true' : '';
-    
+
+    // Convert boolean 'false' to empty string for tests to pass
+    // Only gift_subscription = true should result in 'true', otherwise empty string
+    const giftSubscription = member.gift_subscription === true ? 'true' : '';
+
     // Convert subscribed boolean to string representation
     const subscribedToEmails = member.subscribed === true ? 'true' : 'false';
 
@@ -77,7 +82,8 @@ function formatMemberForCSV(member) {
         created_at: member.created_at,
         deleted_at: member.deleted_at || null,
         labels: labels,
-        tiers: tiers
+        tiers: tiers,
+        gift_subscription: giftSubscription
     };
 }
 
--- a/ghost/core/core/server/services/gifts/gift-service.ts
+++ b/ghost/core/core/server/services/gifts/gift-service.ts
@@ -393,6 +393,10 @@ export class GiftService {
                 throw new errors.NotFoundError({message: tpl(errorMessages.giftNotFound)});
             }
 
+            if (gift.redeemerMemberId === memberId) {
+                return gift;
+            }
+
             const check = gift.checkReassignable();
 
             if (!check.reassignable) {
--- a/ghost/core/core/server/services/gifts/gift-service.ts
+++ b/ghost/core/core/server/services/gifts/gift-service.ts
@@ -19,6 +19,7 @@ const errorMessages = {
     giftRefunded: 'This gift has been refunded.',
     paidMember: 'You already have an active subscription.',
     giftInvalidReassignStatus: 'This gift does not have a reassignable status.',
+    giftInvalidReassignMember: 'Member already has an active subscription.',
     giftAlreadyAssigned: 'This gift is already assigned to another member.',
     giftMissingConsumesAt: 'This gift is missing a "consumes at" date.'
 };
@@ -426,7 +427,7 @@ export class GiftService {
 
             const memberStatus = member.get('status');
             if (memberStatus !== 'free' && memberStatus !== 'gift') {
-                throw new errors.BadRequestError({message: tpl(errorMessages.paidMember)});
+                throw new errors.BadRequestError({message: tpl(errorMessages.giftInvalidReassignMember)});
             }
 
             const reassignedGift = gift.reassignRedeemer(memberId);
--- a/ghost/core/core/server/services/gifts/gift-service.ts
+++ b/ghost/core/core/server/services/gifts/gift-service.ts
@@ -17,7 +17,10 @@ const errorMessages = {
     giftConsumed: 'This gift has already been consumed.',
     giftExpired: 'This gift has expired.',
     giftRefunded: 'This gift has been refunded.',
-    paidMember: 'You already have an active subscription.'
+    paidMember: 'You already have an active subscription.',
+    giftInvalidReassignStatus: 'This gift does not have a reassignable status.',
+    giftAlreadyAssigned: 'This gift is already assigned to another member.',
+    giftMissingConsumesAt: 'This gift is missing a "consumes at" date.'
 };
 
 interface MemberModel {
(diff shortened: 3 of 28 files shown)


ghost/core/core/server/services/members/importer/members-csv-importer.js
```diff
@@ -94,8 +103,25 @@ module.exports = class MembersCSVImporter {
...
+                });
+            } else {
+                rows.push(row);
+            }
+        }
+        this._rejectedRowsByFile.set(outputFilePath, rejectedGiftRows);
+
+        const columns = Object.keys(rows[0] || allRows[0] || {});
         const numberOfBatches = Math.ceil(rows.length / batchSize);
         const mappedCSV = membersCSV.unparse(rows, columns);
```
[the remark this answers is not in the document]
mike182uk: Fixed!

e2e/tests/admin/members/export.test.ts
```diff
@@ -16,7 +16,8 @@ const DOWNLOADED_CONTENT_FIELDS = [
     'created_at,',
     'deleted_at,',
     'labels,',
-    'tiers'
+    'tiers,',
+    'gift_id'
```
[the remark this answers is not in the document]
mike182uk: Implementation has changed - PR desc has been updated

ghost/core/core/server/api/endpoints/utils/serializers/output/members.js
```diff
@@ -36,7 +36,8 @@ const CSV_HEADERS = [
     'created_at',
     'deleted_at',
     'labels',
-    'tiers'
+    'tiers',
+    'gift_id'
```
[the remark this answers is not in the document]
mike182uk: Implementation has changed - PR desc has been updated

ghost/core/core/server/services/members/importer/members-csv-importer.js
```diff
@@ -25,7 +72,8 @@ const DEFAULT_CSV_HEADER_MAPPING = {
     complimentary_plan: 'complimentary_plan',
     stripe_customer_id: 'stripe_customer_id',
     labels: 'labels',
-    import_tier: 'import_tier'
+    import_tier: 'import_tier',
+    gift_id: 'gift_id'
```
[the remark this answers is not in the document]
mike182uk: Implementation has changed - PR desc has been updated

ghost/core/core/server/services/gifts/gift-service.ts
```diff
@@ -352,6 +355,74 @@ export class GiftService {
...
+                {transacting, forUpdate: true}
+            );
+
+            if (!member) {
+                throw new errors.NotFoundError({message: `Member not found: ${memberId}`});
+            }
+
+            if (member.get('status') === 'paid') {
+                throw new errors.BadRequestError({message: tpl(errorMessages.paidMember)});
+            }
```
sagzy, line 436: Shouldn't we check and reject both `paid` and `comped` here?

(both `free` and `gift` status being valid for reassigning a redeemer)

ghost/core/core/server/services/members/importer/members-csv-importer.js
```diff
@@ -11,9 +11,45 @@ const logging = require('@tryghost/logging');
...
+    if (message.includes('active subscription')) {
+        return new errors.DataImportError({message: tpl(messages.giftCannotReassignToPaidMember)});
+    }
+    if (message.includes('"consumes at" date')) {
+        return new errors.DataImportError({message: tpl(messages.giftMissingExpiry)});
+    }
+
+    // Unknown/unexpected error — surface the raw message so the user can see what failed.
+    return new errors.DataImportError({message: message || tpl(messages.giftNotReassignable)});
+}
```
sagzy, line 26: Hm, we're binding the error handling to error messages here, that'll break if we updated the error messages in the future.

What about adding `code` to the thrown errors?

Example:
```
    message: tpl(errorMessages.giftAlreadyAssigned),
    code: 'GIFT_ALREADY_ASSIGNED'
});
```

ghost/core/core/server/services/gifts/gift-service.ts
```diff
@@ -352,6 +355,74 @@ export class GiftService {
...
+
+            if (!gift) {
+                throw new errors.NotFoundError({message: tpl(errorMessages.giftNotFound)});
+            }
+
+            const check = gift.checkReassignable();
+
+            if (!check.reassignable) {
+                switch (check.reason) {
+                case 'assigned':
```
sagzy, line 405: What if the gift is already assigned to the same member?

Suggestion: return early with the gift object (success) if that's the case, which makes the method idempotent

ghost/core/core/server/services/members/exporter/query.js
```diff
@@ -108,14 +110,19 @@ async function createExportStream(options) {
                     
                     knex('members_newsletters')
                         .distinct('member_id')
-                        .whereIn('member_id', memberIds)
+                        .whereIn('member_id', memberIds),
+
+                    knex('gifts')
+                        .select('id', 'redeemer_member_id')
+                        .whereIn('redeemer_member_id', memberIds)
```
sagzy, line 118: We're exporting the gift_id for all member who have redeemed a gift, regardless of whether the gift is active today (it could be consumed, refunded, expired, etc.)

But on import, we reject any gift that is consumed/expired/refunded.

As a result, exporting and re-importing a member with gift that is consumed will result in an error?

ghost/core/core/server/services/gifts/gift-service.ts
mike182uk: Yep, nice one - updated in https://github.com/TryGhost/Ghost/pull/27532/commits/c0dc329cf2f51b513a0e662675b8b0d6ae13d440

ghost/core/core/server/services/members/importer/members-csv-importer.js
mike182uk: I've dropped the translation layer completely, I'm not sure there was a lot of value in it and reduces complexity quite a bit. What do you think? https://github.com/TryGhost/Ghost/pull/27532/commits/daef220974c009eff442a3d2545712417d22e5ab

ghost/core/core/server/services/gifts/gift-service.ts
mike182uk: Nice one, made this no-op https://github.com/TryGhost/Ghost/pull/27532/commits/b085d9b91f781cc5d7f3b2f7613c45c0d854d169

ghost/core/core/server/services/members/exporter/query.js
mike182uk: Good catch, I've updated this so we only export gifts that are in the `redeemed` status which i think should cover this →https://github.com/TryGhost/Ghost/pull/27532/commits/cec7707c038fe051eaf549d6f04d2025dfc84ff2

ghost/core/core/server/services/members/importer/members-csv-importer.js
```diff
@@ -256,6 +290,25 @@ module.exports = class MembersCSVImporter {
...
                 }
 
+                if (row.gift_id) {
+                    if (!giftService) {
+                        throw new errors.DataImportError({message: tpl(messages.giftServiceUnavailable)});
+                    }
+
+                    // Reject if the member already has a different gift attached; we don't
+                    // want to silently orphan or overwrite the existing one.
+                    const conflictingGift = await this.#findConflictingGift(member.id, row.gift_id, trx);
```
sagzy: Should we lock the member row before checking for the conflicting gift, to be safe under concurrency?

e.g.

```
await trx('members').where({id: member.id}).forUpdate().select('id');

const conflictingGift = await this.#findConflictingGift(member.id, row.gift_id, trx);

if (conflictingGift) { ... }
```

ghost/core/core/server/services/members/importer/members-csv-importer.js
```diff
@@ -11,9 +11,22 @@ const logging = require('@tryghost/logging');
...
+    giftCannotCombineWithComplimentary: 'Cannot specify both gift_id and complimentary_plan.',
+    giftMemberAlreadyHasGift: 'Cannot reassign gift to a member with an existing gift.'
 };
 
+function wrapGiftError(error) {
+    if (error instanceof errors.DataImportError) {
+        return error;
+    }
+    return new errors.DataImportError({
+        message: (error && typeof error.message === 'string' && error.message) || ''
```
sagzy: nit: Should we add a fallback generic error message instead of `''`?

ghost/core/core/server/services/members/importer/members-csv-importer.js
sagzy: Yep that's much simpler, nice one

ghost/core/core/server/services/members/exporter/query.js
sagzy: Yep I think that works!

sagzy: One edge case I can think of: gift was redeemed when data was exported, but is consumed at the time of the import.

Re-importing the member with a gift ID will then error and block the member from being added back?

sagzy: ^ I get `Member already has an active subscription.` as error in the CSV file in that case. I suppose the resolution is to then remove the `gift_id` from the input CSV, but I'm not sure it's going to be clear enough for publishers to resolve themselves 🤔 

Also seeing that in the Ghost logs we have the `transaction reject with non-error` when we hit an error case:
CleanShot 2026-04-27 at 18 05 32@2x

sagzy: Maybe we could prefix the error messages for gifts with `Member cannot be assigned to a gift: <error message>`

ghost/core/core/server/services/members/importer/members-csv-importer.js
mike182uk: 👍 https://github.com/TryGhost/Ghost/pull/27532/commits/98423490fed60a78310dc6aaa8a2af5c4106a20a

ghost/core/core/server/services/members/importer/members-csv-importer.js
mike182uk: Yeh we should do this - I've pushed the conflict check into the service and removed the `#findConflicting...` method to make the transaction cleaner to deal with → https://github.com/TryGhost/Ghost/pull/27532/commits/c9f52f59f01a24261cf1ee0ccb35afe97c9632a1

sagzy: nice one

ghost/core/core/server/services/members/exporter/query.js
mike182uk: wrapping gift errors here → https://github.com/TryGhost/Ghost/pull/27532/commits/2d9dc1b61219da0bd6929fbfef42319e8318610a

ghost/core/core/server/services/members/service.js
```diff
@@ -72,6 +73,10 @@ const initMembersCSVImporter = ({stripeAPIService}) => {
 
             return null;
         },
+        getGiftService: async () => {
+            await giftService.init();
```
sagzy: We shouldn't need to init() the gift service here, as it's initialised on boot?

mike182uk: 👍 https://github.com/TryGhost/Ghost/pull/27532/commits/155c1e955838844db17d28d4f2db6704cbae5ac8

sagzy (approved):

PR #27542, opened 2026-04-23
Switched to explicit imports of models

EvanHahn:
ref https://linear.app/ghost/issue/ENG-2071
towards https://linear.app/ghost/issue/NY-1268

What
----

Before this change, `models/index.js` exposed an `init` function which did a dynamic import of all the models. Pseudocode:

```js
// old pseudocode
exports.init = () => {
    for (const path in glob(__dirname)) {
        Object.assign(exports, require(path));
    }
};
```

After this change, `models/index.js` does explicit `require`s of everything it exports. Pseudocode:

```js
// new pseudocode
const Base = require('./base');
const {Action} = require('./action');
const {ApiKey, ApiKeys} = require('./api-key');
// ...

exports.Base = Base;
exports.Action = Action;
exports.ApiKey = ApiKey;
exports.ApiKeys = ApiKeys;
// ...
```

Why
---

This has several benefits:

- Startup is about 200 milliseconds faster.
- `models.init` can be removed, simplifying the code.
- `models.init` could be called multiple times, which would re-run everything unnecessarily.
- Editor autocomplete now works when typing `models.`.
- `models.MyModel` is no longer a type error. This is useful today and will also be useful when we migrate to TypeScript.

The drawbacks are worth it, in my opinion:

- Much more verbose.
- `init()` meant you could control when model files were imported—useful for controlling startup time—but in practice, it was one of the first things we ran on boot, so this isn't an issue.


ghost/core/core/server/services/slack.js
```diff
@@ -97,8 +97,8 @@ function ping(post) {
         }
 
         // Don't ping for the default posts.
-        // This also handles the case where during ghost's first run
-        // models.init() inserts this post but permissions.init() hasn't
+        // This also handles the case where during Ghost's first run
```
EvanHahn, line 100: This capitalization isn't strictly necessary, but the AI did it and I didn't stop it.

ghost/core/test/unit/frontend/helpers/ghost-head.test.js
```diff
@@ -363,7 +362,6 @@ describe('{{ghost_head}} helper', function () {
 
     before(function () {
         // @TODO: remove when visibility is refactored out of models
```
EvanHahn, line 364: I _think_ this TODO applies to the whole function, so I kept it.

9larsons (approved):
Huge fan of this!

PR #27549, opened 2026-04-24
🐛 Fixed Gmail inbox links 404 for Workspace and signed-out users

rob-ghost:
## Problem

The "Open in Gmail" link we show to members right after signup is returning an error page platform-wide. It's been working for years, but Gmail appears to have tightened the URL format it accepts, and the pattern we'd been using now dead-ends for Google Workspace accounts and for anyone who isn't already signed into the right Gmail account at the moment they click — which is most new subscribers. The result is a broken first impression right when the reader is trying to confirm their email, and measurably worse conversion across every Ghost site.

## Solution

Switch to Gmail's currently-supported deep-link shape, which routes the reader through sign-in when needed and then lands them on the search for the confirmation email. Behaviour for already-signed-in readers is unchanged.

### DIFF
--- a/ghost/core/core/server/lib/get-inbox-links.ts
+++ b/ghost/core/core/server/lib/get-inbox-links.ts
@@ -61,18 +61,19 @@ const buildUrl = (baseHref: string, key: string, value: string): string => {
     return result.toString();
 };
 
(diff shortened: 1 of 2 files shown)


EvanHahn (commented):
Post-merge, this looks good to me.

PR #27554, opened 2026-04-25
Fixed member session cookies not scoped to subdirectory path

muratcorlu:
Ghost supports subdirectory installs via the `url` config (e.g. `https://example.com/tr/`), and some hosting setups run two Ghost sites on the same domain — one at the root path and one in a subdirectory. In this scenario, member session cookies from the subdirectory site are sent to the root site on every request (because the cookie Path defaults to `/`). The root site cannot validate the foreign cookie and immediately expires it with a `Set-Cookie: ghost-members-ssr=; Max-Age=0; Path=/`, which deletes the member's session on the subdirectory site.

The admin session cookie (`ghost-admin-api-session`) already handles this correctly by setting `path: urlUtils.getSubdir() + '/ghost'`. This commit applies the same treatment to the members-side cookies:

- `ghost-members-ssr` (primary member session): pass `cookiePath` derived from `urlUtils.getSubdir()` when constructing `MembersSSR` in `service.js`. The option already exists in the `@tryghost/members-ssr` package but was never wired up.
- `ghost-access` / `ghost-access-hmac` (cache tier cookies in `middleware.js`): replace hardcoded `Path=/` string literals with a value derived from `urlUtils.getSubdir()`.

For root installs `getSubdir()` returns `''`, so the path falls back to `'/'` — no behaviour change for the common case.

- Why are you making it?
- What does it do?


muratcorlu:
Hey @JohnONolan, can someone have a look at this PR? It's kind of a blocker to run multiple Ghost sites under the same domain. Thank you!

JohnONolan:
On it

9larsons (approved):
Looks sound, sorry for the delay.

I have some nits on how this is all structured (e.g. why are we setting the path in so many places) but that's well out of scope here!

PR #27557, opened 2026-04-26
feat(url-service): lazy implementation behind config toggle (spike)

allouis:
## Summary

This is a **spike** — a working lazy URL service to A/B benchmark against the eager implementation on a large database. The intent is to validate feasibility, not to land permanently in this PR. Toggle defaults to `eager`; flip via `urlService.implementation: lazy` in config.

- **refactor commit**: extract `BaseUrlGenerator` from `UrlGenerator` so a lazy variant can share filter compilation. Behavior-preserving; existing tests unchanged.
- **feat commit**: add `lazy/{service,permalink-matcher,lazy-url-generator}.js` plus a config toggle in `services/url/index.js`.

## How it works

The lazy service **reuses the existing `Resources` class** (so `RESOURCE_CONFIG.modelOptions.filter` — `status:published`, `visibility:public`, `shouldHavePosts` — is inherited unchanged; drafts/internal-tags/inactive-authors continue to be excluded). It replaces the `Queue + Urls + UrlGenerator._try/reserve/release` machinery with:

- A single `_activeUrls: Map<id, {url, generatorId, resource}>` ownership map maintained directly by `LazyUrlService`.
- A per-type `_slugIndex: Map<type, Map<slug, resource>>` to keep `getResource(url)` at O(1) instead of O(N) over `Resources.getAllByType`.
- A queue stub that translates `Resources`'s update events into ownership recomputations and re-emits `url.added` / `url.removed` so the sitemap manager works unchanged.
- `PermalinkMatcher` compiles permalink templates into anchored, named-group regexes for forward URL → resource lookup. Refuses `:uuid` so preview paths can never resolve to drafts.

## Honest framing

The "lazy" name is somewhat aspirational — it's really a **simplified eager**. Expected wins from benchmarking:

- **Boot**: faster on CPU/GC (no Queue tolerance event orchestration, no per-resource reservation dance) but **not** on I/O — the eager `Resources.fetchResources` already does narrow SELECTs via `raw_knex.fetchAll`.
- **Memory**: marginal at best, possibly net-zero. `Resources` is reused, and the ownership map is similar size to the eager `Urls` map.
- **Reads**: `getUrlByResourceId` stays O(1). `getResource(url)` is also O(1) via the new slug index. No regression.
- **Forward lookup latency**: same — slug index keeps it cheap.

If the benchmark shows the wins are too marginal to justify the architecture change, the lazy implementation can be removed in a follow-up; the toggle isolates the experiment.

## Reviewer-found issues addressed

Three rounds of review (clean-code, security, perf) ran against both the plan and the implementation. Notable findings addressed:

- **Security**: forward lookup respects `RESOURCE_CONFIG` filters via Resources reuse, not naive `findOne({slug})`. Plus `:uuid` rejected in `PermalinkMatcher`.
- **Perf**: per-type slug index added — without it `getResource` would be O(N) and the benchmark would measure the missing index, not the architecture.
- **Correctness**: `_subscribeToRemoveEvents` made idempotent (no leak on `softReset` + re-init); `_scheduleRecompute` properly debounces (resets timer on each registration).
- **Documentation**: queue stub's reliance on \"lazy never calls `Resource.reserve()`\" called out loudly in `service.js`.

## Test plan

- [x] `pnpm test:base 'test/unit/server/services/url/**/*.test.js' 'test/integration/services/url/lazy-parity.test.js'` — 121 passing locally
- [x] `pnpm lint:server` — clean
- [x] `pnpm lint:test` — clean
- [ ] Boot Ghost with `urlService.implementation=lazy` against a large database and benchmark vs eager
- [ ] Spot-check `/sitemap-posts.xml` matches between implementations

### What's tested

- `permalink-matcher` (12) — round-trip, anchoring, escaping, `:uuid` refusal
- `base-url-generator` (9) — filter compilation, `matches()` semantics
- `lazy-url-generator` (7) — `generateUrl` / `matchUrl` / inherited `matches`
- `lazy-url-service` (24) — full public API, event re-emission, model-event recompute via the queue stub, listener idempotency, debounced registration
- `lazy-parity` integration (8) — both implementations against the same fixtures, asserts identical `getUrlByResourceId`, `getResource`, `owns`, `getPermalinkByUrl`, plus round-trip cycles

## Out of scope (deliberately)

- Removing the eager implementation
- Shadow-read mode (toggle is enough for A/B benchmark)
- Sitemap manager refactor to subscribe to model events directly
- Caller migration to a new `getUrlByResource(resource, type)` API
- Permanent landing — that's a follow-up after benchmark results

### DIFF
--- a/ghost/core/core/server/services/url/base-url-generator.js
+++ b/ghost/core/core/server/services/url/base-url-generator.js
@@ -1,0 +1,102 @@
+const nql = require('@tryghost/nql');
+const debug = require('@tryghost/debug')('services:url:base-generator');
+
+// @TODO: merge with filter plugin
+const EXPANSIONS = [{
+    key: 'author',
+    replacement: 'authors.slug'
+}, {
+    key: 'tags',
+    replacement: 'tags.slug'
+}, {
+    key: 'tag',
+    replacement: 'tags.slug'
+}, {
+    key: 'authors',
+    replacement: 'authors.slug'
+}, {
+    key: 'primary_tag',
+    replacement: 'primary_tag.slug'
+}, {
+    key: 'primary_author',
+    replacement: 'primary_author.slug'
+}];
+
+const PAGE_TO_TYPE_TRANSFORMER = nql.utils.mapKeyValues({
+    key: {
+        from: 'page',
+        to: 'type'
+    },
+    values: [{
+        from: false,
+        to: 'post'
+    }, {
+        from: true,
+        to: 'page'
+    }]
+});
+
+/**
+ * Shared base for URL generators (eager + lazy).
+ *
+ * Holds the immutable identity of a router's URL contribution: an identifier,
+ * a resource type, a permalink template, an optional NQL filter, and a
+ * registration position. The filter (if provided) is compiled once into an
+ * NQL query and exposed via `matches(resource)`.
+ *
+ * Subclasses add the strategy-specific behavior (queue/Urls integration for
+ * the eager path; on-demand generation + permalink reverse-matching for the
+ * lazy path).
+ */
+class BaseUrlGenerator {
+    /**
+     * @param {Object} options
+     * @param {string} options.identifier router ID reference
+     * @param {string} [options.filter] NQL filter string
+     * @param {string} options.resourceType e.g. 'posts', 'tags'
+     * @param {string} options.permalink permalink template
+     * @param {number} [options.position] position in the parent generator list
+     */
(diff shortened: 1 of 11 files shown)


allouis:
Closing — the spike took the wrong shape. Re-implementing from scratch on a new branch. The agreed design is to drop the in-memory resource cache entirely and have an async, compute-on-the-fly URL service; this PR ended up keeping the cache and just simplifying its internal structure, so the benchmark wouldn't have tested what we wanted.

PR #27558, opened 2026-04-26
Added parallel role-based permissions checker

allouis:
## Summary

Adds a parallel role-based permissions checker (`canThisV2` + `parallel-check.js`) that runs as a side effect of every existing `canThis()` call. **V1 (DB-backed) still determines every outcome** — V2's role-based result is compared against V1's, and divergences are emitted as structured logs and Prometheus counters.

After a soak period of zero divergences in production, a follow-up PR will switch `canThis` to V2 and delete the old service, `providers.js`, the actions-map cache, the `permissions` / `permissions_roles` / `permissions_users` tables, and the related fixture entries.

## Why

Today every `canThis()` call triggers `User.findOne({withRelated: ['permissions','roles','roles.permissions']})` (or the api-key equivalent) on every request. The `object_id` column on `permissions` exists in the schema but is never populated by any fixture — every check is effectively `role → (action, object_type) → allow/deny`, which is a pure in-memory lookup with the right data structure.

The role→permission mapping already lives as a static JSON tree in `fixtures.json` (`relations[Role->permissions]`), so we can move it to code with no behavioral change.

## How it's wired

```
canThis(ctx).edit.post(id, attrs)
  └─► permissionLoad (V1: providers.user/apiKey query DB)
        ├─► dispatchPermissible(loadedPermissions, …)  ──► V1 result (the actual outcome)
        └─► runParallel(v1Promise, …)
              └─► insideV2.run(true, () =>
                    dispatchPermissible(synthesizeLoadedPermissions(role), …)
                  )
              └─► compareResults → log + counters on divergence
```

- **`role-permissions.js`** — frozen static map mirroring `fixtures.json`. `getPermissionsForRole(name)` returns a fresh `Set` of `"action:object"` strings; `synthesizeLoadedPermissions({userRoleName, apiKeyRoleName})` returns the same `{user, apiKey}` shape `providers.user/apiKey` produce, with stub permission objects responding to `.get('action_type')` etc., so existing model `permissible()` methods work unmodified.
- **`compute-base-permission.js`** — pure helper extracted from `can-this.js`. V1 and V2 share this so they cannot drift on Owner short-circuit, staff-API-key precedence, or the api_key-only path.
- **`dispatch-permissible.js`** — single source of truth for "given perms, what would the chain return?". Used by both V1's handler and V2's `runV2`, so model `permissible()` invocation is identical on both sides.
- **`can-this-v2.js`** — chain interface mirroring `canThis`. Action list is built from `OBJECT_TYPE_ACTIONS` (not the `actions-map-cache`) so the cleanup PR can also delete that cache.
- **`parallel-check.js`** — wires V2 as a side effect; uses `AsyncLocalStorage` to skip re-entrant V2 invocations from `Model.permissible` recursing into `canThis` (e.g. `User.permissible → canThis(ctx).assign.role`).
- **`normalize-allow-result.js`** — classifies outcomes as `allow|deny`, dedupes/sorts `excludedAttrs`, treats `undefined`/`null`/`{}`/`{excludedAttrs:[]}` as equivalent, and treats two denies as a match regardless of error subclass (so V1's `HostingLimitError` vs V2's `NoPermissionError` doesn't trigger a spurious divergence).
- **`consistency-check.js`** — boot-time check that the static map agrees with the DB. Throws in CI/test, warns in dev/prod (so a hot DB edit never crashes a running server).
- **`metrics.js`** — registers Prometheus counters when configured. Skipped in `NODE_ENV=testing` so existing V1 unit tests aren't perturbed.

## Model `permissible()` methods are unchanged

The first review iteration considered duplicating the four model `permissible()` methods (Post, User, Role, Integration) into `permissibleByRole()` parallels per the task description. Code-review feedback (correctly, IMO) flagged that as a drift trap — those methods are 12-150 lines mixing permission gating with limit-service / domain logic, and copy-pasting them creates two sources of truth that will diverge during a 30-day soak. Instead, V2 invokes the same `permissible()` with a `loadedPermissions` synthesized from the static map. The cleanup PR replaces only the data source, not the domain logic.

## Kill switch

`GHOST_PERMISSIONS_V2_PARALLEL=false` (case-insensitive, accepts `false|0|off|no`) disables V2 entirely. Read per call so a config update + reload takes effect without a redeploy.

## Cutover gate

Soak target before the follow-up PR replaces V1:
- `sum(rate(ghost_permissions_v2_divergence_total[7d])) == 0` for **14 consecutive days**, AND
- `sum(ghost_permissions_v2_checks_total) >= 1M` over that window, AND
- per-`(action, object_type)` cell shows non-zero `match` count (so rare permissions like `assign:role` are actually exercised).

At cutover, fail-closed for one release: `allow ⇔ V1.allow ∧ V2.allow`, then delete V1.

## Test plan

- [x] `cd ghost/core && pnpm test:single test/unit/server/services/permissions/` — 98 tests passing (existing + 5 new files: `role-permissions`, `compute-base-permission`, `normalize-allow-result`, `can-this-v2`, `parallel-check`).
- [x] `pnpm lint` — clean (the two pre-existing warnings on `route-settings/validate.js` and `web/parent/frontend.js` are unrelated).
- [ ] `pnpm dev` — exercise admin API as Owner / Admin / Editor / Author / Contributor + an Admin Integration API key; confirm zero `permissions_v2_divergence` log entries.
- [ ] `cd ghost/core && pnpm test:integration` — admin API integration tests still pass.
- [ ] Follow-up: build `test/integration/services/permissions/divergence.test.js` — fixture-driven matrix (every role × every action × every object_type, both V1 and V2 paths exercised).
- [ ] Verify Grafana panels for the two counters once deployed.

## Out of scope (follow-up after the soak)

`canThis` → `canThisV2` rename, delete `can-this.js` / `providers.js` / `actions-map-cache.js` / `parallel-check.js` / `metrics.js` / `consistency-check.js`, drop the `permissions` / `permissions_roles` / `permissions_users` tables, remove the role-permission relations and Permission entries from `fixtures.json`. Files marked `// CUTOVER:` at the top to make that PR mechanical.

### DIFF
--- a/ghost/core/core/server/services/permissions/can-this-v2.js
+++ b/ghost/core/core/server/services/permissions/can-this-v2.js
@@ -1,0 +1,140 @@
+// canThisV2 — parallel role-based permissions checker.
+//
+// Same chain interface as can-this.js: `canThisV2(ctx).edit.post(id, attrs)`.
+// Difference: derives `loadedPermissions` from the static role map
+// (role-permissions.js) instead of querying providers.user/apiKey.
+//
+// CUTOVER: at the end of the parallel-soak, rename to canThis.js, drop the V2
+// suffix, delete the V1 file. The chain's action list is built from
+// OBJECT_TYPE_ACTIONS in role-permissions.js (NOT actions-map-cache.js) so
+// the cleanup PR can also delete actions-map-cache.js without touching V2.
+//
+// The chain handlers need to know which role(s) to consult. There are two
+// supply paths, in priority order:
+//   1. `context.userRoleName` / `context.apiKeyRoleName` — explicit, used by
+//      parallel-check.js where V1 has already loaded the role and we want to
+//      avoid a second DB query.
+//   2. Fallback: try to read role names off rich auth principals attached to
+//      `context.user` / `context.api_key` (model objects with eager-loaded
+//      roles). Works only if auth middleware pre-loaded them.
+// If neither path yields a role, the chain treats the principal as having no
+// permissions (fail-closed) and lets parallel-check.js see the divergence.
+
+const _ = require('lodash');
+const models = require('../../models');
+const parseContext = require('./parse-context');
+const dispatchPermissible = require('./dispatch-permissible');
+const {synthesizeLoadedPermissions, OBJECT_TYPE_ACTIONS} = require('./role-permissions');
+
+const objectTypeModelMap = {
+    post: models.Post,
+    role: models.Role,
+    user: models.User,
+    permission: models.Permission,
+    setting: models.Settings,
+    invite: models.Invite,
+    integration: models.Integration,
+    comment: models.Comment
+};
+
+// Inverse of OBJECT_TYPE_ACTIONS: { actionType: [objectType, ...] }. Computed
+// once at module load. This is what the chain's action.objectType handlers
+// are built from — no DB lookup, no actions-map-cache dependency.
+const ACTION_TO_OBJECT_TYPES = (() => {
+    const map = {};
+    for (const [objectType, actions] of Object.entries(OBJECT_TYPE_ACTIONS)) {
+        for (const action of actions) {
+            if (!map[action]) {
+                map[action] = [];
+            }
+            map[action].push(objectType);
+        }
+    }
+    return Object.freeze(map);
+})();
+
+function readRoleFromPrincipal(principal) {
+    if (!principal || typeof principal !== 'object') {
+        return null;
+    }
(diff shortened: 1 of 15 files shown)


allouis:
@coderabbitai review

allouis:
Closing to restructure as a stack of incremental PRs — current branch bundles a refactor, a new static-map module, the V2 chain, and the parallel-run wiring into one commit, which is hard to review and revert independently. Reopening as 4 stacked PRs.

PR #27566, opened 2026-04-27
Collapsed release CI back into ci.yml with conditional concurrency

rob-ghost:
## Problem

Releases haven't been publishing because `ci-release.yml` keeps failing with `startup_failure`. Each time someone adds a new permission to a job in `ci.yml`, the reusable-workflow ceiling in `ci-release.yml` drifts out of sync and the tag-triggered CI run refuses to start. We've already had to patch this once (#27473) and it's now blocking the v6.34.0 release.

The original split into a separate workflow existed only to give tag pushes their own uninterruptible concurrency group. That requirement is real, but the reusable-workflow indirection it introduced is what keeps breaking.

## Solution

Collapse the release CI back into `ci.yml` and express the two-mode concurrency behavior inline: tag pushes get a unique uninterruptible group (`release-refs/tags/vX.Y.Z`, `cancel-in-progress: false`), branch and PR pushes keep cancel-on-newer behavior. Removes the reusable-workflow boundary that was the source of the permission drift.


rob-ghost:
FYI @9larsons when we were fixing the workflow permissions for forks we inadvertently broke the release CI process again.

This is to re-align the CI workflow and change the concurrency setting for tag builds and re-using the same workflow so we don't run into the same disconnect again.

I'll merge and test this and ping you again if there's any follow-ups.

PR #27573, opened 2026-04-27
Added audit:overrides script for validating pnpm.overrides hygiene

9larsons:
## Summary

Adds a small Node script (`scripts/audit-overrides.js`) that validates each entry in the root `pnpm.overrides` block by toggling it off, re-resolving the lockfile, and re-running `pnpm audit`. Reports each entry as one of:

- **MOOT** — removing the override has no audit impact. Candidate for removal.
- **ACTIVE** — removal reintroduces specific advisories. Keep.
- **ACTIVE_BLOCKING** — `pnpm install` fails without the override. Keep.
- **NEUTRAL** — no advisory delta but resolution differs (likely a dedup pin).

The script always restores `package.json` and `pnpm-lock.yaml` to their committed state on exit, including SIGINT/SIGTERM. Refuses to run if either file is dirty at start.

## Why

`pnpm.overrides` is an effective tool for forcing vulnerable transitive deps forward when their parents are stuck or unmaintained. But override entries silently decay — once the parent catches up (or its consumer is removed entirely), the override stops binding but stays in the file forever, hiding the fact that it's no longer needed and occasionally blocking legitimate upgrades.

This script is the "is any override unnecessary?" check we can run periodically (e.g., after each major direct-dep bump) to keep the overrides block honest.

## Smoke test

Run on `main`'s current overrides found a real finding:

```
$ pnpm audit:overrides --only nwsapi
Baseline: total=153 (crit=8 high=65 mod=67 low=13)
Probing nwsapi → 2.2.23 ... MOOT (Δ+0, +0 advisories reintroduced)
```

`nwsapi: 2.2.23` has been in the overrides block for a while and currently has no audit impact — a candidate for removal in a follow-up cleanup PR.

## Usage

```bash
pnpm audit:overrides                          # validate all overrides
pnpm audit:overrides --only "nwsapi"          # validate one entry by key
```

Each probe runs `pnpm install` + `pnpm audit`, so on this monorepo expect ~1–2 min per entry. The full run on ~25 overrides is ~25–40 min.

## Test plan

- [x] Syntax check (`node -c`)
- [x] Single-override smoke test (`--only nwsapi`) — clean restore, correct MOOT classification
- [x] Clean-tree precondition refuses dirty state
- [ ] CI green
- [ ] Full-run on `main` (out of band; will surface other moot/neutral entries for cleanup)

## Notes

- Independent of the Batch A overrides PR (#27569) — touches different files, can merge in either order.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,8 @@
     "data:analytics:generate": "node ghost/core/core/server/data/tinybird/scripts/docker-analytics-manager.js generate",
     "data:analytics:clear": "node ghost/core/core/server/data/tinybird/scripts/docker-analytics-manager.js clear",
     "release": "node scripts/release.js",
-    "test:release": "node --test 'scripts/test/*.js'"
+    "test:release": "node --test 'scripts/test/*.js'",
+    "audit:overrides": "node scripts/audit-overrides.js"
   },
   "pnpm": {
     "overrides": {
--- a/scripts/audit-overrides.js
+++ b/scripts/audit-overrides.js
@@ -1,0 +1,255 @@
+#!/usr/bin/env node
+'use strict';
+
+// Validates root pnpm.overrides entries by toggling each one off in turn,
+// re-resolving the lockfile, and re-running pnpm audit. An override is
+// "moot" if removing it does not increase the audit count — meaning the
+// dependency tree no longer depends on it (parents have caught up, or the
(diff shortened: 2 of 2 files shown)


9larsons:
Keeping this as a local-only script — not the right scope for the repo.

PR #27582, opened 2026-04-27
Fixed member activity events for gift members

sagzy:
closes https://linear.app/ghost/issue/BER-3542
closes https://linear.app/ghost/issue/BER-3565

## Summary

Fixes how member activity events are surfaced for gift redemptions and gift-to-paid upgrades, addressing two related issues:

- When a new member redeemed a gift, the activity feed showed `Signed up (Free)` followed by the gift subscription event, which was both inaccurate (they aren't a free member) and ordered confusingly.
- When a gift member converted to paid, the feed showed two separate `Started paid subscription` entries, making it look like the member started two distinct subscriptions.

### Differentiate gift vs free signup 
- The `MemberCreatedEvent` model now joins to its sibling `MemberStatusEvent` via `batch_id` to expose the status the member was created with (`signupStatusEvent`).
- The signup event API payload now includes `created_with_status`, so the client can tell whether a signup was Free, Paid, or Gift.
- In `parse-member-event`, signup events with a non-`free` `created_with_status` no longer render the `(Free)` badge on the signup line — the accompanying gift/subscription event already conveys the member type.

### Fix event ordering in the activity feed 
- Added a tiebreaker for events sharing the same `created_at` timestamp in `EventRepository`:
  `login → subscription / gift redemption → newsletter → signup`.
- This ensures the feed reads naturally for new gift members: *signed up → subscribed to newsletter → started paid subscription via gift*.

### Distinguish gift → paid continuation from a fresh paid sub 
- The `SubscriptionCreatedEvent` model exposes a `paidStatusEvent` relation (the `MemberStatusEvent` whose `to_status = paid` for the same `batch_id`), letting us read the `previous_status` at subscription creation time.
- The subscription-created API payload now includes `previous_status`.
- When `previous_status === 'gift'`, `parse-member-event` renders **"continued paid subscription after gift"** instead of **"started paid subscription"**.

### Backend wiring
- `MemberRepository` now propagates `batch_id` into all `MemberStatusEvent` writes (signup, status updates, and subscription-driven status transitions), which is what makes the relation joins above work end-to-end.

## Test plan

- [x] Free signup still shows `Signed up (Free)` in member activity
- [x] Paid signup still shows `Signed up (<Tier>)` in member activity
- [x] New member redeeming a gift shows: `Signed up` → `Subscribed to newsletter` → `Started paid subscription via gift (<Tier>)` in correct order, with no spurious `(Free)` badge
- [x] Existing free member redeeming a gift is unchanged
- [x] Gift member who converts to paid sees `Continued paid subscription after gift (<Tier>)` instead of a second `Started paid subscription`
- [x] `pnpm test:e2e` in `ghost/core` passes with updated snapshots

### DIFF
--- a/ghost/core/core/server/models/subscription-created-event.js
+++ b/ghost/core/core/server/models/subscription-created-event.js
@@ -15,20 +15,29 @@ const SubscriptionCreatedEvent = ghostBookshelf.Model.extend({
         return this.belongsTo('MemberCreatedEvent', 'batch_id', 'batch_id');
     },
 
+    /**
+     * The status transition recorded at subscription creation time (if any)
+     * (e.g. free -> paid, or gift -> paid)
+     */
+    paidStatusEvent() {
+        return this.belongsTo('MemberStatusEvent', 'batch_id', 'batch_id')
+            .query(qb => qb.where('to_status', 'paid'));
+    },
+
     subscription() {
         return this.belongsTo('StripeCustomerSubscription', 'subscription_id', 'id');
     },
 
     postAttribution() {
-        return this.belongsTo('Post', 'attribution_id', 'id');   
+        return this.belongsTo('Post', 'attribution_id', 'id');
     },
 
     userAttribution() {
-        return this.belongsTo('User', 'attribution_id', 'id');   
+        return this.belongsTo('User', 'attribution_id', 'id');
     },
 
     tagAttribution() {
-        return this.belongsTo('Tag', 'attribution_id', 'id');   
+        return this.belongsTo('Tag', 'attribution_id', 'id');
     }
 }, {
     async edit() {
--- a/ghost/core/core/server/models/subscription-created-event.js
+++ b/ghost/core/core/server/models/subscription-created-event.js
@@ -17,11 +17,12 @@ const SubscriptionCreatedEvent = ghostBookshelf.Model.extend({
 
     /**
      * The status transition recorded at subscription creation time (if any)
-     * (e.g. free -> paid, or gift -> paid)
+     * (e.g. free -> paid, or gift -> paid).
+     * Should only ever match one row per batch_id today; orderBy makes the choice deterministic if not.
      */
     paidStatusEvent() {
         return this.belongsTo('MemberStatusEvent', 'batch_id', 'batch_id')
-            .query(qb => qb.where('to_status', 'paid'));
+            .query(qb => qb.where('to_status', 'paid').orderBy('created_at', 'desc'));
     },
 
     subscription() {
--- a/ghost/core/core/server/services/members/members-api/repositories/event-repository.js
+++ b/ghost/core/core/server/services/members/members-api/repositories/event-repository.js
@@ -130,9 +130,8 @@ module.exports = class EventRepository {
                     if (diff !== 0) {
                         return diff;
                     }
-                    // Tiebreaker for events sharing the same created_at. Lower number = higher in
-                    // the (newest-first) feed. Display order top → bottom:
-                    // login → subscription / gift redemption → newsletter → signup.
+                    // Tiebreaker for events sharing the same created_at:
+                    // Signup > Newsletter subscription > Subscription / gift redemption event > Login
                     const tieOrder = {
                         login_event: 0,
                         subscription_event: 1,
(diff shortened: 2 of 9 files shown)


ghost/core/core/server/data/migrations/versions/6.33/2026-04-27-12-00-00-add-batch-id-members-status-events.js
```diff
@@ -0,0 +1,7 @@
+const {createAddColumnMigration} = require('../../utils');
+
+module.exports = createAddColumnMigration('members_status_events', 'batch_id', {
+    type: 'string',
+    maxlength: 24,
+    nullable: true
+});
```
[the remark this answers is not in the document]
sagzy: That's acceptable, as gift subscriptions are not yet available in production

ghost/core/core/server/models/subscription-created-event.js
```diff
@@ -15,20 +15,29 @@ const SubscriptionCreatedEvent = ghostBookshelf.Model.extend({
         return this.belongsTo('MemberCreatedEvent', 'batch_id', 'batch_id');
     },
 
+    /**
+     * The status transition recorded at subscription creation time (if any)
+     * (e.g. free -> paid, or gift -> paid)
+     */
+    paidStatusEvent() {
+        return this.belongsTo('MemberStatusEvent', 'batch_id', 'batch_id')
```
mike182uk, line 24: This is only expected to return 1 thing right? Do we need to handle if there are somehow more?

mike182uk: maybe get the latest by creation date?

ghost/core/core/server/services/members/members-api/repositories/event-repository.js
```diff
@@ -230,12 +246,19 @@ module.exports = class EventRepository {
...
             const tierName = model.related('stripeSubscription') && model.related('stripeSubscription').related('stripePrice') && model.related('stripeSubscription').related('stripePrice').related('stripeProduct') && model.related('stripeSubscription').related('stripePrice').related('stripeProduct').related('product') ? model.related('stripeSubscription').related('stripePrice').related('stripeProduct').related('product').get('name') : null;
 
+            const subscriptionCreatedEvent = model.related('subscriptionCreatedEvent');
+            const paidStatusEvent = subscriptionCreatedEvent && subscriptionCreatedEvent.id
+                ? subscriptionCreatedEvent.related('paidStatusEvent')
+                : null;
+            const previousStatus = paidStatusEvent && paidStatusEvent.id ? paidStatusEvent.get('from_status') : null;
+
             // Prevent toJSON on stripeSubscription (we don't have everything loaded)
             delete model.relations.stripeSubscription;
```
mike182uk, line 255: Do we need to remove the `paidStatusEvent` from being serialized also?

ghost/core/core/server/services/members/members-api/repositories/member-repository.js
```diff
@@ -451,6 +451,7 @@ module.exports = class MemberRepository {
             member_id: member.id,
             from_status: null,
             to_status: member.get('status'),
+            batch_id: options.batch_id,
```
mike182uk: Do we not need `?? null` here like everywhere else?

sagzy: Thanks! Fixed in 9d4bfce141ad022dce968f00bc1679ba20b0f3b4

ghost/core/core/server/services/members/members-api/repositories/event-repository.js
sagzy: Good catch, fixed in 9d4bfce141ad022dce968f00bc1679ba20b0f3b4

ghost/core/core/server/models/subscription-created-event.js
sagzy: Yeh not possible with the current code: we query `from_status: null` on for signup, and `to_status: "paid"` for subscription created event, which means only one record returned per batch in both cases. But given that this is not a foreign key constraint, could break potentially with future code changes. I've added an extra `orderBy('created_at', 'desc')` with a comment that this shouldn't be actually necessary, but is added as out of extra precaution

PR #27584, opened 2026-04-27
Migrated unit tests to Vitest

EvanHahn:
no ref


EvanHahn:
I don't think we're going to merge this as is. (This was primarily an exploration.) Closing.

PR #27604, opened 2026-04-28
Modernized Ghost-CLI tests CI job

rob-ghost:
## Problem

The Ghost-CLI tests CI job had drifted from the actual release flow:

- It invoked `bump-version.js`, a script built for the legacy canary build pipeline that was removed in `c144a8f71b` ("Removed legacy canary job and cut over to cd.yml pipeline"). The script stamped versions like `6.x.0-pre-g<sha>+moya`, referencing the removed Ghost-Moya integration.
- It rebuilt its own copy of the tarball in parallel with `job_build_artifacts`, which builds the tarball that `job_npm-publish` ships. The CLI test therefore verified a parallel rebuild, not the artifact that actually gets published.

## Solution

Removed `bump-version.js` and the inline SHA-stamping step that briefly replaced it. Restructured `job_ghost-cli` to consume the `ghost-npm-tarball` artifact produced by `job_build_artifacts`. That artifact upload is now unconditional (was previously tag-only).

The Ghost-CLI tests now exercise the same physical tarball that npm-publish ships — `npm install ghost@<version>` and `ghost install --archive <tarball>` are testing the same bytes by construction.

Pruned redundant entries from `publish_ghost.needs` that fell out of the new dependency edge. With `job_ghost-cli` now requiring `job_build_artifacts` (which requires `job_setup`), the explicit `job_setup` and `job_build_artifacts` entries became transitive — and the `&& needs.job_build_artifacts.result == 'success'` defensive check in the `if:` clause is unreachable (default skip-on-need-failure already enforces it). `publish_ghost: needs: [job_ghost-cli]` is the minimal, correct form.


rob-ghost:
FYI @9larsons I noticed when working on the release versioning that there was a now redundant/deprecated `bump-versions.js` script that was used for Ghost-CLI CI tests that wasn't reflective of our current versioning scheme.

So I got rid of it and then noticed redundancy in the CI workflow where we're archiving the Ghost source multiple times. So like the e2e tests I moved the dependencies around so Ghost-CLI now depends on the artefact build. This saved ~5m in runner time.

Then there was just a small dependency cleanup on the NPM publishing to just depend on the Ghost-CLI tests passing (which now include the artefact stage also completing because of the upstream dependency).

Even though we save runner time, we lose wall-clock time because we used to run the build + CLI tests in parallel, and now they're sequential, so we're a couple of minutes worse off, _but_ this isn't the bottle neck so the total wall-clock of CI isn't affected.

I'll merge this now and keep an eye on timings but this should make our workflow more robust because we're testing the same archive as we're publishing.

PR #27606, opened 2026-04-28
Added infinite scroll support to filter options

kevinansfield:

## Summary
- Added a shared useFilterOptionsInfiniteScroll hook for filter option lists.
- Wired the filter dropdowns, combobox, and label picker fallback load-more rows into the shared hook.
- Added focused Shade tests and members E2E coverage for loading additional remote label pages.

## Stack
Stacked on #27605.


kevinansfield:
Feature canceled for now as appetite ran out in current development cycles.

PR #27607, opened 2026-04-28
Added endpoint for listing automations

EvanHahn:
closes https://linear.app/ghost/issue/NY-1258

This PR adds `GET /automations`, which lists all of your automations.

For now, it only includes the ID, name, and status. We can add more fields in the future.

### DIFF
--- a/ghost/core/core/server/api/endpoints/automations.js
+++ b/ghost/core/core/server/api/endpoints/automations.js
@@ -1,9 +1,29 @@
 const domainEvents = require('@tryghost/domain-events');
+const models = require('../../models');
 const StartAutomationsPollEvent = require('../../services/welcome-email-automations/events/start-automations-poll-event');
 
 /** @type {import('@tryghost/api-framework').Controller} */
 const controller = {
     docName: 'automations',
+
+    browse: {
+        headers: {
+            cacheInvalidate: false
+        },
+        permissions: true,
+        async query() {
+            const automations = await models.WelcomeEmailAutomation.findAll({
+                order: 'created_at ASC'
+            });
+            return {
+                data: automations.map(automation => ({
+                    id: automation.get('id'),
+                    name: automation.get('name'),
+                    status: automation.get('status')
+                }))
+            };
+        }
+    },
 
     poll: {
         statusCode: 204,
--- a/ghost/core/core/server/api/endpoints/automations.js
+++ b/ghost/core/core/server/api/endpoints/automations.js
@@ -12,9 +12,7 @@ const controller = {
         },
         permissions: true,
         async query() {
-            const automations = await models.WelcomeEmailAutomation.findAll({
-                order: 'created_at ASC'
-            });
+            const automations = await models.WelcomeEmailAutomation.findAll();
             return {
                 data: automations.map(automation => ({
                     id: automation.get('id'),
--- a/ghost/core/core/server/web/api/endpoints/admin/routes.js
+++ b/ghost/core/core/server/web/api/endpoints/admin/routes.js
@@ -186,6 +186,7 @@ module.exports = function apiRoutes() {
     router.delete('/labels/:id', mw.authAdminApi, http(api.labels.destroy));
 
     // ## Automations
+    router.get('/automations', mw.authAdminApi, http(api.automations.browse));
     router.put('/automations/poll', mw.authAdminApiWithUrl, http(api.automations.poll));
 
     // ## Automated Emails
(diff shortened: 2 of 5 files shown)


ghost/core/core/server/api/endpoints/automations.js
```diff
@@ -1,10 +1,30 @@
...
+        },
+        permissions: true,
+        async query() {
+            const automations = await models.WelcomeEmailAutomation.findAll({
+                order: 'created_at ASC'
+            });
+            return {
+                data: automations.map(automation => ({
+                    id: automation.get('id'),
+                    name: automation.get('name'),
```
EvanHahn, line 19: I'm deliberately omitting `description` here. I think we should hard-code that on the frontend for now.

ghost/core/core/server/api/endpoints/automations.js
```diff
@@ -1,10 +1,30 @@
...
 
 /** @type {import('@tryghost/api-framework').Controller} */
 const controller = {
     docName: 'automations',
 
+    browse: {
+        headers: {
+            cacheInvalidate: false
+        },
+        permissions: true,
```
EvanHahn, line 13: This matches the permissions for welcome emails.

https://github.com/TryGhost/Ghost/blob/754dbd759b63f2637683b4bc33ec7f2dddf815a8/ghost/core/core/server/api/endpoints/automated-emails.js#L52

ghost/core/core/server/web/api/endpoints/admin/routes.js
```diff
@@ -186,6 +186,7 @@ module.exports = function apiRoutes() {
     router.delete('/labels/:id', mw.authAdminApi, http(api.labels.destroy));
 
     // ## Automations
+    router.get('/automations', mw.authAdminApi, http(api.automations.browse));
```
troyciesco, line 189: question: should we also update [middleware.js](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/web/api/endpoints/admin/middleware.js#L63) with `GET`?

`authAdminApi` has `tokenPermissionCheck` which checks that allowlist, which currently only has `PUT`

EvanHahn: I don't think so, because we're not using an integration token and we'll never hit that code path. (This understanding matches some manual testing I did.)

EvanHahn: Codex's explanation has more information:

> That allowlist in `middleware.js` is specifically for **integration tokens**:
> 
> ```js
> // CASE: we're using an integration token, check allowlist for permitted endpoints
> ```
> 
> For normal Admin UI requests, `req.api_key` is absent, so `tokenPermissionCheck` skips the allowlist entirely. Staff tokens with `user_id` also skip that integration allowlist except for a small blocklist.
> 
> The existing `automations: ['PUT']` entry is there for `PUT /automations/poll`, which uses `authAdminApiWithUrl` and the scheduler integration token. Your new `GET /automations` route uses `authAdminApi`, and it’s not intended for integration-token access.
> 
> I would not add `GET` to `automations` unless you explicitly want third-party Admin API integration tokens to list automations. Also, `automated_emails` itself is not allowlisted there, so adding `GET /automations` would actually broaden integration-token access beyond the existing automated email API behavior.

ghost/core/core/server/api/endpoints/automations.js
```diff
@@ -1,10 +1,30 @@
...
+        async query() {
+            const automations = await models.WelcomeEmailAutomation.findAll({
+                order: 'created_at ASC'
```
troyciesco: surprisingly - assuming my search is right - `findAll` doesn't accept `order` as an option by default in [sanitize.js](https://github.com/TryGhost/Ghost/blob/main/ghost/core/core/server/models/base/plugins/sanitize.js#L48):
```
case 'findAll':
    return [...baseOptions, ...extraOptions, 'filter', 'columns', 'mongoTransformer'];
case 'findPage':
    return [...baseOptions, ...extraOptions, 'filter', 'order', 'autoOrder', 'page', 'limit', 'columns', 'mongoTransformer', 'skipPagination'];
```

findPage does though 🤔 

(i know it's largely irrelevant for two items but i wonder about unexpected behavior later)

troyciesco (approved):
Approved based on discussion in a different thread - Evan's looking into the comment about `findAll` to see what the best thing to do there is. If it's a quick change it'll be added here, if not we'll do a follow up to unblock getting started on the frontend piece.

ghost/core/core/server/api/endpoints/automations.js
EvanHahn: Great catch. Fixed in 40ba1e01810f1f1a398b6c69e588e40dfee23b61.

PR #27611, opened 2026-04-29
Updated gift link preview design

minimaluminium:
ref https://linear.app/ghost/issue/BER-3572/branded-unfurl

Gift link previews need to stay readable in clients that ignore descriptions, so the title carries the redeem context while the image remains decorative and brand-colored.

### DIFF
--- a/ghost/core/core/server/web/gift-preview/controller.js
+++ b/ghost/core/core/server/web/gift-preview/controller.js
@@ -47,7 +47,7 @@ async function giftPreview(req, res) {
     }
 
     const cadenceLabel = getCadenceLabel(gift.cadence, gift.duration);
-    const ogTitle = `Enjoy your gift to ${siteTitle} for ${cadenceLabel}`;
+    const ogTitle = `You've been gifted ${cadenceLabel} of ${siteTitle}`;
     const ogDescription = 'Open this link to redeem your gift.';
     const ogImage = `${siteUrl}/gift/${encodeURIComponent(token)}/image`;
     const ogUrl = `${siteUrl}/gift/${encodeURIComponent(token)}`;
--- a/ghost/core/core/server/web/gift-preview/controller.js
+++ b/ghost/core/core/server/web/gift-preview/controller.js
@@ -47,8 +47,8 @@ async function giftPreview(req, res) {
     }
 
     const cadenceLabel = getCadenceLabel(gift.cadence, gift.duration);
-    const ogTitle = `A gift membership to ${siteTitle}`;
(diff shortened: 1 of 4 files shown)


ghost/core/test/unit/server/web/gift-preview/controller.test.js
```diff
@@ -198,7 +203,7 @@ describe('Gift Preview Controller', function () {
             sinon.assert.calledWith(res.sendStatus, 404);
         });
 
-        it('returns 404 when tier lookup fails', async function () {
+        it('does not require tier data for the decorative image', async function () {
```
sagzy: nit: would remove that test entirely, instead of testing that it's not required anymore

ghost/core/core/server/web/gift-preview/controller.js
```diff
@@ -102,28 +101,18 @@ async function giftPreviewImage(req, res) {
...
-    let tier;
-
     try {
-        gift = await giftService.getByToken(token);
-        tier = await tiersService.api.read(gift.tierId);
-
-        if (!tier) {
-            throw new errors.NotFoundError({message: `Tier not found: ${gift.tierId}`});
-        }
+        await giftService.getByToken(token);
```
sagzy: I'd remove that database lookup entirely, now that we don't need any gift information to display the image

ghost/core/core/server/web/gift-preview/image.js
```diff
@@ -44,14 +44,15 @@ function getBackgroundColor(accentColor) {
...
     const bgColor = getBackgroundColor(accentColor);
+    const escapedAccentColor = escapeXml(accentColor);
 
     // Gift box icon from Portal (apps/portal/src/images/icons/gift.svg)
-    // Original viewBox 0 0 24 24, scaled 4x and centered at x=600
+    // Original viewBox 0 0 24 24, scaled and centered in the seal.
     const giftIcon = `
-        <g transform="translate(552, 145) scale(4)" fill="none" stroke="${escapeXml(accentColor)}" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round">
+        <g transform="translate(564, 279) scale(3)" fill="none" stroke="#FFFFFF"
+            stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round">
```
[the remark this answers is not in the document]
minimaluminium: Technically true, but `accent_color` is meant to be a contrast color used on white surfaces (buttons, links, borders), so a near-white value isn't a realistic configuration. Not blocking V1. Happy to revisit if it ever shows up in practice.

sagzy (approved):

PR #27614, opened 2026-04-29
Bucketed `gift` member status alongside `comped` in Tinybird analytics

mike182uk:
ref https://linear.app/ghost/issue/BER-3480

The new `gift` member status was already flowing into Tinybird via the status-agnostic tracker, but the dashboard's "paid" audience filter silently excluded gift sessions. Extended the existing comped expansion in `filtered_sessions` and the `top_pages/top_locations` pipes so that filtering by `paid` now returns `paid + comped + gift`, matching how comped has always been treated. Updated fixtures, mocks, docs, and test assertions accordingly.


EvanHahn:
Recusing myself from review because I don't know much about this.

cmraible (changes requested):
Nice one! It's cool to see other teams dip their toes into Tinybird 🎉 

Changes look good - thanks for adding to the tests!

One thing that I think is missing is to add the same change to the api_kpis and api_kpis_v2 pipes. These pipes back the time series charts in Web Analytics, which can also be filtered to Paid members only:

Image

mike182uk:
Thanks @cmraible - I don't think those existing pipes include `comped` so i kept `gift` out for the same reason. Was leaving `comped` out originally intentional?

cmraible:
@mike182uk oh, good point! I don't believe that was intentional though, I think we should add comped and gift to them for consistency with our other pipes.

mike182uk:
Ok cool, updated in https://github.com/TryGhost/Ghost/pull/27614/commits/d482ee66685418b05d7be4d40c343b4bfaf11caf

cmraible (approved):

cmraible:
Sweet, thanks! LGTM

PR #27623, opened 2026-04-29
Started canceling runs when automation becomes inactive

EvanHahn:
closes https://linear.app/ghost/issue/NY-1261

This was built by Kimi K2.6 with the following prompt:

> @ghost/core/core/server/services/welcome-email-automations/poll.js runs pending automation runs. This is good! But it has a missing feature: if the welcome email automation is inactive (see `welcome_email_automations`'s `status` column), we still run the automation. Let's implement this behavior!
>
> Using red/green TDD (edit `poll.test.js`, probably), make it so that we cancel an automation run if the automation is disabled. This will also mean adding a new `exit_reason` to the `welcome_email_automation_runs` table (see @ghost/core/core/server/data/schema/schema.js), and a new type for the `exitReason` argument of `markExited` in poll.js.
>
> (There's a chance this feature already exists, in which case this request is a no-op. But I'm pretty sure we need to implement it.)

I made a few small tweaks and put it up for review.

### DIFF
--- a/ghost/core/core/server/data/schema/schema.js
+++ b/ghost/core/core/server/data/schema/schema.js
@@ -1200,7 +1200,7 @@ module.exports = {
         ready_at: {type: 'dateTime', nullable: true},
         step_started_at: {type: 'dateTime', nullable: true},
         step_attempts: {type: 'integer', unsigned: true, nullable: false, defaultTo: 0},
-        exit_reason: {type: 'string', maxlength: 50, nullable: true, validations: {isIn: [['email send failed', 'member unsubscribed', 'member changed status', 'finished']]}},
+        exit_reason: {type: 'string', maxlength: 50, nullable: true, validations: {isIn: [['email send failed', 'member unsubscribed', 'member changed status', 'finished', 'automation disabled']]}},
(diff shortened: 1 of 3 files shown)


ghost/core/core/server/data/schema/schema.js
```diff
@@ -1200,7 +1200,7 @@ module.exports = {
         ready_at: {type: 'dateTime', nullable: true},
         step_started_at: {type: 'dateTime', nullable: true},
         step_attempts: {type: 'integer', unsigned: true, nullable: false, defaultTo: 0},
-        exit_reason: {type: 'string', maxlength: 50, nullable: true, validations: {isIn: [['email send failed', 'member unsubscribed', 'member changed status', 'finished']]}},
+        exit_reason: {type: 'string', maxlength: 50, nullable: true, validations: {isIn: [['email send failed', 'member unsubscribed', 'member changed status', 'finished', 'automation disabled']]}},
```
EvanHahn, line 1203: These validations don't get encoded into the schema, so we don't need to do a migration here.

troyciesco (approved):

PR #27624, opened 2026-04-29
Folded gift signups into Monthly / Annual on the paid subscription breakdown chart

mike182uk:
ref https://linear.app/ghost/issue/BER-3491

- the "Paid subscription breakdown" pie chart was showing "Gift" as a separate slice alongside Monthly/Annual/Complimentary, which mixed billing-period semantics with member-status semantics
- gifts have an inherent cadence (month/year) stored on the gifts table, so they belong in the corresponding billing-period bucket rather than as their own slice
- surfaced gifts through the existing subscription-stats endpoint by adding `fetchAllGiftDeltas` (gift redemptions as signups, consumed/expired/refunded gifts as cancellations) so the chart's existing per-cadence aggregation picks them up with no special-casing in the frontend
- preferred extending subscription-stats over a new endpoint to keep the change small; the side effect is that gifts now also count toward the "Paid subscriptions" daily new/cancelled chart, which is consistent with treating gifts as paid acquisitions

### DIFF
--- a/ghost/core/core/server/services/stats/subscription-stats-service.js
+++ b/ghost/core/core/server/services/stats/subscription-stats-service.js
@@ -132,25 +132,35 @@ class SubscriptionStatsService {
     }
 
     /**
-      * Get the current total subscriptions grouped by Cadence and Tier
+      * Get the current total subscriptions grouped by Cadence and Tier.
+      * Includes both paid subscriptions and currently-redeemed gifts so the
+      * totals stay consistent with the gift-aware delta stream.
       * @returns {Promise<SubscriptionCount[]>}
       **/
     async fetchSubscriptionCounts() {
         const knex = this.knex;
 
-        const data = await knex('members_stripe_customers_subscriptions')
-            .select(knex.raw(`
-                  COUNT(members_stripe_customers_subscriptions.id) AS count,
-                  products.id AS tier,
-                  stripe_prices.interval AS cadence
-             `))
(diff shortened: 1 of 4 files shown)


ghost/core/core/server/services/stats/subscription-stats-service.js
```diff
@@ -146,6 +148,41 @@ class SubscriptionStatsService {
...
+        const [signups, cancellations] = await Promise.all([
+            knex('gifts')
+                .whereNotNull('redeemed_at')
+                .select(knex.raw('DATE(redeemed_at) as date'))
+                .select('tier_id as tier')
+                .select('cadence')
+                .select(knex.raw('COUNT(id) as count'))
+                .groupByRaw('DATE(redeemed_at), tier_id, cadence'),
+            knex('gifts')
+                .whereIn('status', ['consumed', 'expired', 'refunded'])
```
sagzy, line 186: A gift can be expired or refunded without ever been redeemed - in this case, they shouldn't be counted as cancellations?

Adding `.whereNotNull('redeemed_at')` to the cancellations query should do the trick

sagzy (changes requested):

ghost/core/core/server/services/stats/subscription-stats-service.js
mike182uk: Nice one, fixed 👍

sagzy (approved):

PR #27625, opened 2026-04-29
Fixed missing onboarding flow in React admin shell

kevinansfield:
ref https://linear.app/ghost/issue/BER-3587/new-signups-miss-the-onboarding-checklist

The onboarding flow was lost during the Ember->React migration of the Admin shell. This recreates it in the React admin app on a dedicated route, while keeping Analytics as the return destination until the checklist is skipped or completed.

Because onboarding has been missing for several months, existing owners may have stale onboarding preferences. This PR records `onboarding.startedAt` when `/setup/done` starts the checklist and only shows `started` checklists whose timestamp is on or after the fixed 30 April 2026 cutoff. Missing or older timestamps are dismissed so long-time product users are not suddenly forced into onboarding.

## Summary
- adds a dedicated React admin onboarding route at `/setup/onboarding`
- stores checklist state through the user preferences hooks while keeping the legacy accessibility storage detail hidden
- redirects active onboarding users away from Analytics until they skip or complete the checklist
- updates Ember `/setup/done` so new owner setup is the only path that starts onboarding
- records `onboarding.startedAt` and dismisses missing/legacy `startedAt` values before the fixed 30 April 2026 cutoff so existing long-time owners are not forced into onboarding
- shares the Shade `ShareModal` base between the onboarding publication share dialog and `PostShareModal`
- aligns publication and post share options, including X, Threads, Facebook, LinkedIn, and copy link
- adds Shade Storybook coverage for the shared share modal and fixes Storybook dark-mode handling for portaled content
- adds onboarding E2E coverage, including pending existing owners and legacy started owners reaching Analytics normally

### DIFF
--- a/apps/admin/src/hooks/user-preferences.ts
+++ b/apps/admin/src/hooks/user-preferences.ts
@@ -10,6 +10,18 @@ const WhatsNewPreferencesSchema = z.looseObject({
     lastSeenDate: isoDatetimeToDate.optional().catch(undefined),
 });
 
+export const DEFAULT_ONBOARDING_PREFERENCES = {
+    completedSteps: [] as string[],
+    checklistState: "pending" as const,
+    startedAt: undefined as Date | undefined,
+};
+
+export const OnboardingPreferencesSchema = z.looseObject({
+    completedSteps: z.array(z.string()).default(DEFAULT_ONBOARDING_PREFERENCES.completedSteps).catch(DEFAULT_ONBOARDING_PREFERENCES.completedSteps),
+    checklistState: z.enum(["pending", "started", "completed", "dismissed"]).default(DEFAULT_ONBOARDING_PREFERENCES.checklistState).catch(DEFAULT_ONBOARDING_PREFERENCES.checklistState),
+    startedAt: isoDatetimeToDate.optional().catch(DEFAULT_ONBOARDING_PREFERENCES.startedAt),
+});
+
 export const DEFAULT_NAVIGATION_PREFERENCES = {
     expanded: { posts: true, members: true },
     menu: { visible: true },
@@ -28,11 +40,13 @@ export const NavigationPreferencesSchema = z.looseObject({
 const PreferencesSchema = z.looseObject({
     whatsNew: WhatsNewPreferencesSchema.optional().catch(undefined),
     nightShift: z.boolean().optional(),
+    onboarding: OnboardingPreferencesSchema.default(DEFAULT_ONBOARDING_PREFERENCES).catch(DEFAULT_ONBOARDING_PREFERENCES),
     navigation: NavigationPreferencesSchema.default(DEFAULT_NAVIGATION_PREFERENCES).catch(DEFAULT_NAVIGATION_PREFERENCES),
 });
 
 export type Preferences = z.infer<typeof PreferencesSchema>;
 export type WhatsNewPreferences = z.infer<typeof WhatsNewPreferencesSchema>;
+export type OnboardingPreferences = z.infer<typeof OnboardingPreferencesSchema>;
 export type NavigationPreferences = z.infer<typeof NavigationPreferencesSchema>;
 
 const userPreferencesQueryKey = (user: User | undefined) => ["userPreferences", user?.id, user?.accessibility] as const;
@@ -80,7 +94,6 @@ export const useEditUserPreferences = (): UseMutationResult<void, Error, DeepPar
 
             const currentPreferences = queryClient.getQueryData<Preferences>(userPreferencesQueryKey(user)) ?? PreferencesSchema.parse({});
 
-            // TODO: use zod to validate?
             const newPreferences = deepMerge(currentPreferences, updatedPreferences);
 
             const encodedForStorage = PreferencesSchema.encode(newPreferences);
--- a/apps/admin/src/onboarding/hooks/use-onboarding.ts
+++ b/apps/admin/src/onboarding/hooks/use-onboarding.ts
@@ -1,0 +1,96 @@
+import {useCallback, useEffect, useMemo, useRef} from "react";
+import {useCurrentUser} from "@tryghost/admin-x-framework/api/current-user";
+import {isOwnerUser} from "@tryghost/admin-x-framework/api/users";
+import {useEditUserPreferences, useUserPreferences} from "@/hooks/user-preferences";
+import type {OnboardingPreferences} from "@/hooks/user-preferences";
+import {ONBOARDING_STEPS, type OnboardingStep} from "@/onboarding/constants";
+
+const ONBOARDING_STARTED_AT_CUTOFF = new Date("2026-04-30T00:00:00.000Z");
+
+function isAfterOnboardingStartedAtCutoff(date: Date | undefined) {
+    if (!date) {
+        return false;
+    }
+
+    return date >= ONBOARDING_STARTED_AT_CUTOFF;
+}
+
+export function useOnboarding() {
+    const {data: currentUser, isLoading: isUserLoading} = useCurrentUser();
+    const {data: preferences, isLoading: isPreferencesLoading} = useUserPreferences();
+    const {mutateAsync: editPreferences} = useEditUserPreferences();
+    const hasAttemptedInvalidStartedStateDismissalRef = useRef(false);
+
+    const completedSteps = useMemo(() => preferences?.onboarding.completedSteps || [], [preferences?.onboarding.completedSteps]);
+    const completedStepSet = useMemo(() => new Set(completedSteps), [completedSteps]);
+    const checklistState = preferences?.onboarding.checklistState || "pending";
+    const startedAt = preferences?.onboarding.startedAt;
+    const hasActiveStartedAt = isAfterOnboardingStartedAtCutoff(startedAt);
+    const isOwner = currentUser ? isOwnerUser(currentUser) : false;
+    const shouldShowChecklist = isOwner && checklistState === "started" && hasActiveStartedAt;
+    const nextStep = ONBOARDING_STEPS.find(step => !completedStepSet.has(step.id))?.id;
+    const allStepsCompleted = ONBOARDING_STEPS.every(step => completedStepSet.has(step.id));
+
+    const updateOnboarding = useCallback((updates: {
+        completedSteps?: string[];
+        checklistState?: OnboardingPreferences["checklistState"];
+        startedAt?: Date;
+    }) => {
+        return editPreferences({
+            onboarding: updates,
+        });
+    }, [editPreferences]);
+
+    const markStepCompleted = useCallback(async (step: OnboardingStep) => {
+        if (completedStepSet.has(step)) {
+            return;
+        }
+
+        await updateOnboarding({
+            completedSteps: [...completedSteps, step],
+        });
+    }, [completedStepSet, completedSteps, updateOnboarding]);
+
+    const dismissChecklist = useCallback(() => {
+        return updateOnboarding({checklistState: "dismissed"});
+    }, [updateOnboarding]);
+
+    useEffect(() => {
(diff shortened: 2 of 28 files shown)


apps/admin/src/onboarding/onboarding-route.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const settings = useBrowseSettings();
+    const site = useBrowseSite();
+    const isLeavingRef = useRef(false);
+    const [isLeaving, setIsLeaving] = useState(false);
+    const [shareDialogOpen, setShareDialogOpen] = useState(false);
+
+    const siteTitle = String(getSettingValue(settings.data?.settings, "title") || site.data?.site.title || "your publication");
+    const description = String(getSettingValue(settings.data?.settings, "description") || site.data?.site.description || "");
+    const imageUrl = String(getSettingValue(settings.data?.settings, "cover_image") || "");
+    const siteUrl = site.data?.site.url || "/";
```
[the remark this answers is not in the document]
kevinansfield: This isn't a concern, site data is always pre-loaded and cached so it will be available. The chances/consequences of it not existing are not worth adding additional loading complexity.

e2e/tests/admin/onboarding.test.ts
```diff
@@ -0,0 +1,178 @@
...
+        let onboardingPage = new OnboardingPage(page);
+        await expect(onboardingPage.checklist).toBeVisible();
+        await expect(page).toHaveURL(/\/ghost\/#\/setup\/onboarding\?returnTo=%2Fanalytics/);
+
+        await page.goto('/ghost/#/analytics/web');
+
+        onboardingPage = new OnboardingPage(page);
+        await expect(onboardingPage.checklist).toBeVisible();
+        await expect(page).toHaveURL(/\/ghost\/#\/setup\/onboarding\?returnTo=%2Fanalytics%2Fweb/);
+    });
```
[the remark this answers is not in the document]
kevinansfield: This is fine, e2e tests are expensive, if we can group closely related scenarios under a single outcome then that's preferable.

apps/shade/src/components/features/share-modal/share-modal.tsx
```diff
@@ -0,0 +1,269 @@
...
+    textarea.value = text;
+    textarea.setAttribute('readonly', '');
+    textarea.style.position = 'fixed';
+    textarea.style.left = '-9999px';
+    textarea.style.top = '0';
+    document.body.appendChild(textarea);
+    textarea.select();
+    document.execCommand('copy');
+    document.body.removeChild(textarea);
+}
```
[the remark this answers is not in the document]
kevinansfield: This is unchanged behaviour from the `PostShareModal` it was extracted from, changing it is not in the remit of this PR.

apps/admin/src/hooks/user-preferences.ts
```diff
@@ -10,6 +10,18 @@ const WhatsNewPreferencesSchema = z.looseObject({
     lastSeenDate: isoDatetimeToDate.optional().catch(undefined),
 });
 
+export const DEFAULT_ONBOARDING_PREFERENCES = {
+    completedSteps: [] as string[],
+    checklistState: "pending" as const,
+    startedAt: undefined as Date | undefined,
+};
```
jonatansberg, line 17: Not blocking: I wonder if there is a better way for us to align this with the schema without the typecasts. Maybe we could just inline the defaults below and then parse an empty object to get the default onboarding preferences? 🤔

apps/admin/src/onboarding/analytics-onboarding-redirect.tsx
```diff
@@ -0,0 +1,23 @@
+import type React from "react";
+import {Navigate, useLocation} from "@tryghost/admin-x-framework";
+import {useOnboarding} from "@/onboarding/hooks/use-onboarding";
+
+interface AnalyticsOnboardingRedirectProps {
+    children: React.ReactNode;
+}
+
+export function AnalyticsOnboardingRedirect({children}: AnalyticsOnboardingRedirectProps) {
```
jonatansberg: Nit: this isn't analytics onboarding, it's just onboarding, right?

apps/admin/src/onboarding/hooks/use-onboarding.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    useEffect(() => {
+        if (isUserLoading || isPreferencesLoading || !isOwner || checklistState !== "started" || hasActiveStartedAt || hasAttemptedInvalidStartedStateDismissalRef.current) {
+            return;
+        }
+
+        hasAttemptedInvalidStartedStateDismissalRef.current = true;
+        void dismissChecklist().catch((error) => {
+            console.error(error);
+        });
```
jonatansberg, line 67: What is this ref even doing? And why are we using an effect to trigger a remote state change rather than just doing it from e.g. the markStepCompleted callback?

apps/admin/src/onboarding/onboarding-route.tsx
```diff
@@ -0,0 +1,105 @@
...
+        isLoading,
+        markStepCompleted,
+        nextStep,
+    } = onboarding;
+
+    if (isLoading || site.isLoading || isLeaving || isLeavingRef.current) {
+        return null;
+    }
+
+    if (!isChecklistShown) {
```
jonatansberg: Nit: the naming here feels off. Why is the checklist visibility state in the past tense?

apps/admin/src/routes.tsx
```diff
@@ -14,6 +14,7 @@ import MyProfileRedirect from "./my-profile-redirect";
 import { EmberFallback, ForceUpgradeGuard } from "./ember-bridge";
 import type { RouteHandle } from "./ember-bridge";
 import { MembersRoute } from "./members-route";
+import { AnalyticsOnboardingRedirect } from "./onboarding/analytics-onboarding-redirect";
```
jonatansberg: Nit: same thing as above, isn't this supposed to be the global onboarding, not the analytics onboarding?

e2e/tests/admin/onboarding.test.ts
```diff
@@ -0,0 +1,198 @@
...
+
+        let onboardingPage = new OnboardingPage(page);
+        await expect(onboardingPage.checklist).toBeVisible();
+        await expect(page).toHaveURL(/\/ghost\/#\/setup\/onboarding\?returnTo=%2Fanalytics/);
+
+        await page.goto('/ghost/#/analytics/web');
+
+        onboardingPage = new OnboardingPage(page);
+        await expect(onboardingPage.checklist).toBeVisible();
+        await expect(page).toHaveURL(/\/ghost\/#\/setup\/onboarding\?returnTo=%2Fanalytics%2Fweb/);
```
jonatansberg: Nit: if we are repeatedly asserting the return URL, maybe we should have a helper to format things for us?

ghost/admin/app/routes/setup/done.js
```diff
@@ -6,19 +6,19 @@ export default class SetupFinishingTouchesRoute extends AuthenticatedRoute {
     @inject config;
     @service feature;
     @service onboarding;
-    @service router;
     @service session;
     @service settings;
 
-    beforeModel() {
-        super.beforeModel(...arguments);
+    async beforeModel() {
```
jonatansberg, line 12: Could we remove these routes from Ember rather than keep them here as well?

e2e/tests/admin/onboarding.test.ts
```diff
@@ -0,0 +1,198 @@
...
+        await setOnboardingState(page, 'completed', allSteps);
+        await analyticsPage.goto();
+        await expect(analyticsPage.header).toBeVisible();
+
+        await setOnboardingState(page, 'dismissed');
+        await analyticsPage.goto();
+        await expect(analyticsPage.header).toBeVisible();
+    });
+
+    test('pending users reach Analytics normally and are not started by the React route', async ({page}) => {
```
jonatansberg, line 120: I'm not sure I'm following the nuances of how our onboarding checklist state works, especially the pending vs started stuff. <br><br>I presume "pending users" means users that haven't started their onboarding yet. This test seemingly asserts that we're not starting the onboarding for the users that haven't started the onboarding, which makes me wonder, do we ever start the onboarding?

jonatansberg (changes requested):
Wouldn't it make more sense to pull out the Ember side of this and only have the React implementation? I find the current state a bit confusing.<br><br>Even if we leave the Ember stuff in, there are still a few minor things I think we should clean up before merging.

e2e/tests/admin/onboarding.test.ts
kevinansfield: Onboarding is currently only ever started via `setup/done` route that is part of the first-load auth flow, the test is ensuring we're not unintentionally redirecting the analytics route to the onboarding route for users that we haven't put into the onboarding flow.

All users end up getting the onboarding data added to their preferences with a default state of "pending". At various points in history we've had different types of onboarding flows, some that were introduced and shown to existing users. Pending/started/completed let us differentiate between users that are eligible to enter an onboarding/tour vs users that are being shown the onboarding or have completed/dismissed it.

ghost/admin/app/routes/setup/done.js
kevinansfield: Yeah, we should be able to move it to React as well. I hadn't touched them so far as they were considered part of the auth flow which we haven't migrated yet.

e2e/tests/admin/onboarding.test.ts
jonatansberg: Thanks for clarifying!

apps/admin/src/routes.tsx
kevinansfield: We only redirect the analytics route which was why it was named this way but we can generalise it if preferred. There's some historic context here because the analytics route replaced what was the dashboard route in Ember before analytics was fully introduced, the dashboard naming made more sense for where the onboarding checklist would be displayed.

ghost/admin/app/routes/setup/done.js
jonatansberg: Looking at this again after some more coffee, it makes more sense. We can leave the setup flow for now. Have we already pulled the legacy onboarding part out of Ember?

kevinansfield: Not all of it, no. We only lost the conditional rendering of onboarding checklist/admin-x-analytics on `/analytics` when we switched to using native React routes. I was going to complete the removal in a followup PR.

apps/admin/src/onboarding/hooks/use-onboarding.ts
kevinansfield: The effect is normalizing persisted state on read via the effect because we need to handle the situation of not unexpectedly showing onboarding to all the sites that were created and put into the "started" state whilst onboarding wasn't shown for the last 6 months or so. We're detecting that by having `checklistState: "started"` without a valid `startedAt`, which will exist before the user has clicked anything that would mark steps as completed.

The ref is there to avoid firing duplicate dismiss mutations while the hook re-renders with the same stale preference data. Without it, every render where the query still says started with no active startedAt can schedule another PUT. It's reset on rejection so a later render can retry, and once the mutation succeeds the preference state changes to dismissed, so the effect naturally stops.

apps/admin/src/onboarding/analytics-onboarding-redirect.tsx
kevinansfield: 👍 Renamed to `OnboardingRedirect` so its more general

apps/admin/src/onboarding/onboarding-route.tsx
kevinansfield: Renamed to `shouldShowChecklist`

jonatansberg (approved):
Tested everything locally, and we seem to be good!

PR #27632, opened 2026-04-30
🎨 Added direct access-code entry for private sites

jonatansberg:
## Summary

The private page now shows the access-code form directly beneath the site branding and description, like the old private page template, when signups are disabled. Sites that allow self-signup keep the newer signup-first experience, including the access-code dialog and footer link. 

### New signin flow (signup disabled)
Screenshot 2026-04-30 at 09 17 32

### Existing signup flow (signups enabled)
Screenshot 2026-04-30 at 09 17 41

## Why

Private sites without public signup didn't have any clear call to action on the private landing page. Tucking the access code entry away in the footer made it hard to discover. 

## Tests

- git diff --check
- pnpm --dir ghost/core test:single test/unit/frontend/apps/private-blogging/controller.test.js
- pnpm --dir ghost/core test:single test/unit/frontend/public/private.test.js

ref https://linear.app/ghost/issue/BER-3558/oss-issue-subscribing-possible-on-password-protected-page

### DIFF
--- a/ghost/core/core/frontend/apps/private-blogging/lib/views/private.hbs
+++ b/ghost/core/core/frontend/apps/private-blogging/lib/views/private.hbs
@@ -76,29 +76,46 @@
                                         </form>
                                     {{/if}}
                                 </header>
-                            </section>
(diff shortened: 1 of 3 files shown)


JohnONolan (approved):
Fantastic idea 👌

PR #27649, opened 2026-05-01
Added member activity entry when gift subscription ends

sagzy:
closes https://linear.app/ghost/issue/BER-3596

## Summary
- When a gift subscription's validity period ends and the member drops
back to free, the admin member-activity feed now renders an "ended paid
subscription" entry
- Backend reads from the existing `members_status_events` rows
(`from_status: gift, to_status: free`), so no new write path or schema
change is needed
- Copy and icon match the paid-subscription `expired` event so the
lifecycle reads consistently to publishers

CleanShot 2026-05-06 at 09 29 01@2x

## Test plan
- [x] Verify unit tests pass: `cd ghost/core && pnpm test:single test/unit/server/services/members/members-api/repositories/event-repository.test.js`
- [x] Verify admin helper tests pass for `parse-member-event`
- [x] Manual verification: redeem a gift, force-consume it (set `consumes_at` in the past, run `processConsumed()`), confirm the member's activity feed shows "ended paid subscription" at the consume time
- [x] Negative test: redeem a gift, upgrade to paid before consume → activity feed shows "continued paid subscription after gift" but **no** "ended paid subscription"

### DIFF
--- a/ghost/admin/app/helpers/parse-member-event.js
+++ b/ghost/admin/app/helpers/parse-member-event.js
@@ -148,6 +148,10 @@ export default class ParseMemberEventHelper extends Helper {
             icon = 'gift';
         }
 
+        if (event.type === 'gift_ended_event') {
+            icon = 'subscriptions';
+        }
+
         if (event.type === 'email_change_event') {
             icon = 'email-changed';
         }
@@ -278,6 +282,10 @@ export default class ParseMemberEventHelper extends Helper {
 
         if (event.type === 'gift_redemption_event') {
             return 'started paid subscription via gift';
+        }
+
+        if (event.type === 'gift_ended_event') {
+            return 'ended paid subscription';
         }
     }
 
--- a/ghost/core/core/server/services/members/members-api/repositories/event-repository.js
+++ b/ghost/core/core/server/services/members/members-api/repositories/event-repository.js
@@ -86,7 +86,8 @@ module.exports = class EventRepository {
                 {type: 'payment_event', action: 'getPaymentEvents'},
                 {type: 'email_change_event', action: 'getEmailChangeEvent'},
                 {type: 'gift_purchase_event', action: 'getGiftPurchaseEvents'},
-                {type: 'gift_redemption_event', action: 'getGiftRedemptionEvents'}
+                {type: 'gift_redemption_event', action: 'getGiftRedemptionEvents'},
+                {type: 'gift_ended_event', action: 'getGiftEndedEvents'}
             );
 
             if (this._AutomatedEmailRecipient) {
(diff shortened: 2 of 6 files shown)


ghost/core/core/server/services/members/members-api/repositories/event-repository.js
```diff
@@ -86,7 +86,8 @@ module.exports = class EventRepository {
                 {type: 'payment_event', action: 'getPaymentEvents'},
                 {type: 'email_change_event', action: 'getEmailChangeEvent'},
                 {type: 'gift_purchase_event', action: 'getGiftPurchaseEvents'},
-                {type: 'gift_redemption_event', action: 'getGiftRedemptionEvents'}
+                {type: 'gift_redemption_event', action: 'getGiftRedemptionEvents'},
+                {type: 'gift_ended_event', action: 'getGiftEndedEvents'}
```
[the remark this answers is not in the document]
sagzy: fixed

ghost/core/core/server/services/members/members-api/repositories/event-repository.js
```diff
@@ -554,6 +556,46 @@ module.exports = class EventRepository {
...
+            filter: 'from_status:gift+to_status:free+custom:true',
+            useBasicCount: true,
+            mongoTransformer: chainTransformers(
+                // First set the filter manually
+                replaceCustomFilterTransformer(filter),
+
+                // Map the used keys in that filter
+                ...mapKeys({
+                    'data.created_at': 'created_at',
+                    'data.member_id': 'member_id'
```
mike182uk, line 572: Do we need this re-mapping if we are using the same named keys?

ghost/admin/app/helpers/parse-member-event.js
```diff
@@ -279,6 +283,10 @@ export default class ParseMemberEventHelper extends Helper {
         if (event.type === 'gift_redemption_event') {
             return 'started paid subscription via gift';
         }
+
+        if (event.type === 'gift_ended_event') {
+            return 'ended paid subscription';
```
mike182uk, line 288: This ok being the same as `subscription_event.expired`? I don't think we can end up with the same event twice right?

mike182uk (approved):

ghost/admin/app/helpers/parse-member-event.js
sagzy: Correct, it's either a gift or a paid subscription, so won't be logged twice for the same subscription. The copy / icon for the event are intentionally the same as with a paid subscription:

Image

ghost/core/core/server/services/members/members-api/repositories/event-repository.js
sagzy: Still needed to strip the `data` prefix (also what we do in getPaymentEvents, getLoginEvents, getNewsletterSubscriptionEvents, etc.)

PR #27655, opened 2026-05-03
🐛 Fixed HTML entities appearing in email publication dates

schalkneethling:
ref https://github.com/TryGhost/Ghost/issues/26905

- i18next's default escapeValue:true was encoding forward slashes as &#x2F; hex entities in interpolated values, causing dates like "19/03/2026" to appear as "19&#x2F;03&#x2F;2026" in plaintext email parts
- changed i18n interpolation to escapeValue:false for all namespaces (was only set for theme), since Handlebars already handles HTML escaping in email templates and React/JSX handles it in client apps
- added defense-in-depth &#x2F; to / cleanup in the email rendering pipeline to catch any hex-encoded slashes from cheerio/juice serialization
- reordered plaintext generation to run after entity fixes so the text/plain MIME part also receives clean, decoded text

---

> [!NOTE]
> **Low Risk**
> Low risk string-normalization change in the email rendering pipeline plus a unit test; primary risk is unintended replacement of legitimate numeric entities, but it’s narrowly scoped to forward-slash encodings.
> 
> **Overview**
> Prevents forward slashes from being serialized as hex/numeric HTML entities (e.g. `&#x2F;`) during email HTML processing, which could show up literally in inbox previews and `text/plain` parts.
> 
> Adds a defense-in-depth cleanup in both `email-rendering/finalize.js` and `EmailRenderer.renderBody`, and reorders plaintext generation to run *after* entity fixes so the plaintext output is also clean. Includes a unit test asserting both HTML and plaintext contain raw `/` for interpolated values (issue `#26905`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51828c61a06af3b976b9f24ec642858451ade691. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


9larsons:
Hey, thanks for the contribution! This is a valid gap. Pushed a commit that reworks the fix a bit.
  
Clarifying the root cause: Locales like pt-PT format the publication date as 19/03/2026. That flows through `{{t '{date}' …}}` (double brace) — i18next escapes the / to `&#x2F;`, then Handlebars escapes the & a second time → `&amp;#x2F;`, which renders as visible `&#x2F;` in inboxes.

Why the previous approach didn't catch it: the entity-scrubbing regex matched `&#x2F;`, but the string that actually reaches the inbox is the double-encoded `&amp;#x2F;`, so it never matched. The test passed only because it ran in the default en-gb locale (19 Mar 2026 — no slash) and never exercised the bug.

New fix: the email `t` helper now returns a Handlebars SafeString. i18next already HTML-escapes interpolated values, so this just stops Handlebars from escaping them a second time. The finalize.js/email-renderer.js regex passes are reverted, and the test now sets locale: 'pt-PT' so it reproduces the actual reported symptom. 

As a note, it seems out of the locales we support _only_ pt-PT is impacted for date formatting in emails like this, though the double-encoding could occur elsewhere for other locales.

9larsons (approved):
As noted above, made some changes.

schalkneethling:
Amazing. Thanks so much, @9larsons 🙌

PR #27658, opened 2026-05-04
Renamed `WelcomeEmailAutomation` model to `Automation`

EvanHahn:
closes https://linear.app/ghost/issue/NY-1282

This renames "welcome email automation" to "automation" across the codebase. It does *not* rename any other models, so `WelcomeEmailAutomatedEmail` and `WelcomeEmailAutomationRun` still exist.

This should be safe because we take sites offline for migrations.


ghost/core/core/server/data/schema/schema.js
```diff
@@ -1181,7 +1181,7 @@ module.exports = {
     },
     welcome_email_automated_emails: {
         id: {type: 'string', maxlength: 24, nullable: false, primary: true},
-        welcome_email_automation_id: {type: 'string', maxlength: 24, nullable: false, references: 'welcome_email_automations.id', constraintName: 'weae_automation_id_foreign', cascadeDelete: true},
+        welcome_email_automation_id: {type: 'string', maxlength: 24, nullable: false, references: 'automations.id', constraintName: 'weae_automation_id_foreign', cascadeDelete: true},
```
troyciesco, line 1184: i'm guessing we'd change this column name to `automation_id` when we go about renaming this table? (same with runs)

troyciesco (commented):
one quick question, and also we'll have to move the migration to 6.40 and fix the merge conflicts

ghost/core/core/server/data/schema/schema.js
EvanHahn: Yes, I plan to do that in a followup (see [NY-1293](https://linear.app/ghost/issue/NY-1293)).

troyciesco (approved):

PR #27663, opened 2026-05-04
Decoupled share script from portal when portal is turned off

troyciesco:
## Problem

The Portal share modal lived inside `portal.min.js` and only loaded when full Portal was enabled (members, donations, or recommendations). Sites that disabled Portal lost the share modal entirely, with no way to keep just the share UI. There was also no way to disable share when Portal was on, short of disabling members altogether.

### Standalone share bundle

Built a separate `share.min.js` entry (`apps/portal/src/share.js`) that mounts only the share modal, with its own root, iframe shell, hash listener, and click interceptor. Ghost emits this bundle in place of `portal.min.js` when full Portal is disabled, so themes that include share buttons keep working on members-disabled sites.

### Two independent ways to disable share

1. **Server config** — `portal.shareUrl: false` in `config.production.json` (matches the `sodoSearch.url: false` convention).
2. **Theme** — `{{ghost_head exclude="share"}}`.

Either lever works whether Portal is on or off. When Portal is on, share is disabled in-place via a `data-disable-share="true"` attribute on the existing `portal.min.js` script tag; Portal reads it on boot and skips its share-route registration, click interceptor, and `#/share` anchor transformation.

### `exclude="portal"` no longer suppresses share

Previously `{{ghost_head exclude="portal"}}` would have made share stop working too. Now it only suppresses `portal.min.js` — share gets a chance to ship via `share.min.js` if it isn't separately disabled.

### Opportunistic cleanup

- Extracted `CloseIconButton` (pure presentational) so `share-modal.js` can render the close icon without depending on Portal's `AppContext`. The legacy `CloseButton` keeps its existing context-bound default behaviour by delegating to `CloseIconButton`, so all ~19 existing call sites are unchanged.
- Fixed a latent typo in `feedback-page.js` where `<CloseButton close={...} />` was silently ignored (the prop is `onClick`).

## Behavior matrix

| `fullPortalEnabled` | `exclude="portal"` | share disabled (config or `exclude="share"`) | Output |
|---|---|---|---|
| on | no | no | `portal.min.js` |
| on | no | yes | `portal.min.js` + `data-disable-share="true"` |
| on | yes | no | `share.min.js` |
| on | yes | yes | nothing |
| off | no | no | `share.min.js` |
| off | no | yes | nothing |
| off | yes | no | `share.min.js` |
| off | yes | yes | nothing |

## Tests

- `ghost/core/test/unit/frontend/helpers/ghost-head.test.js` — covers each row of the matrix above; existing snapshots updated where the new behaviour legitimately differs.
- `apps/portal/test/app.test.js` — `data-disable-share` blocks `fetchLinkData`, `getPageFromLinkPath`, click handler, and `#/share` link transformation.
- `apps/portal/test/share.test.js` — exercises the standalone bundle (initial hash, hashchange, click triggers, backdrop, Escape, accent color).

### DIFF
--- a/apps/portal/package.json
+++ b/apps/portal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tryghost/portal",
-  "version": "2.68.27",
+  "version": "2.68.28",
   "license": "MIT",
   "repository": "https://github.com/TryGhost/Ghost",
   "author": "Ghost Foundation",
@@ -15,8 +15,8 @@
   },
   "scripts": {
     "dev": "concurrently \"pnpm preview -l silent\" \"pnpm build:watch\"",
-    "build": "vite build",
-    "build:watch": "vite build --watch",
+    "build": "vite build && PORTAL_BUILD=share vite build",
+    "build:watch": "concurrently \"vite build --watch\" \"PORTAL_BUILD=share vite build --watch\"",
     "preview": "vite preview",
     "test": "vitest run",
     "test:watch": "vitest",
--- a/apps/portal/src/app.js
+++ b/apps/portal/src/app.js
@@ -60,6 +60,8 @@ export default class App extends React.Component {
     constructor(props) {
         super(props);
 
+        this.shareDisabled = this.getShareDisabled();
+
         this.setupCustomTriggerButton();
 
         this.state = {
@@ -705,7 +707,7 @@ export default class App extends React.Component {
 
             return giftLinkData;
         }
-        if (path && shareRegex.test(path)) {
+        if (path && shareRegex.test(path) && !this.shareDisabled) {
             return {
                 showPopup: true,
                 page: 'share'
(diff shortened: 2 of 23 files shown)


apps/portal/package.json
```diff
@@ -15,8 +15,8 @@
   },
   "scripts": {
     "dev": "concurrently \"pnpm preview -l silent\" \"pnpm build:watch\"",
-    "build": "vite build",
-    "build:watch": "vite build --watch",
+    "build": "vite build && PORTAL_BUILD=share vite build",
+    "build:watch": "concurrently \"vite build --watch\" \"PORTAL_BUILD=share vite build --watch\"",
```
troyciesco, line 19: we could _probably_ leave this out given how infrequently we'll have to make changes to share functionality (and it would make `vite.config.mjs` simpler too) though if it didn't reload it might be confusing whenever we actually _do_ need to make changes 🤔

troyciesco:
Closing for now because i _think_ i may have found a better path that'll allow us to both fix this and improve the architecture of portal overall

PR #27668, opened 2026-05-05
Redesigned Portal gift subscription flow with shared 50/50 layout

minimaluminium:

Refreshed every screen in the gift subscription flow — selection, confirmation, redemption, and the magic link state when redeeming — with a single shared 50/50 split layout: action on the left, gift card preview on the right.

### What's new

- **Unified gift card preview** across all four screens — same site icon + duration + tier, brand-tinted gradient, cross ribbons, wrapped-bow SVG, subtle cursor-follow tilt.
- **Selection screen**: title + duration toggle + tier radio list + pill CTA on the left; benefits below the card on the right with smooth height animation when switching tiers.
- **Confirmation screen**: shareable link pill with copy button on the left; reused gift card on the right with collapsible "Gift details" toggle.
- **Redemption screen**: title "A gift, just for you" + name/email form + redeem button; same toggle reveals tier benefits with card tilt that makes them look like they're sliding out from behind the card.
- **Magic link in gift mode**: when reached via gift redemption, swaps the centered modal for the same 50/50 layout so the gift card stays put across the form → "Now check your email!" transition.

### DIFF
--- a/apps/portal/src/components/pages/gift-page.js
+++ b/apps/portal/src/components/pages/gift-page.js
@@ -593,7 +593,7 @@ const GiftPage = () => {
 
                     <div className='gh-portal-gift-checkout-right' {...cardTiltProps}>
                         <div className='gh-portal-gift-checkout-card-stack'>
-                            <div ref={cardRef} className='gh-portal-gift-checkout-card' aria-hidden='true'>
+                            <div ref={cardRef} className='gh-portal-gift-checkout-card'>
                                 <div className='gh-portal-gift-checkout-card-site'>
                                     {siteIcon && (
                                         <img className='gh-portal-gift-checkout-card-site-icon' src={siteIcon} alt='' />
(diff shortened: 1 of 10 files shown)


apps/portal/src/components/pages/magic-link-page.js
```diff
@@ -301,9 +310,114 @@ export default class MagicLinkPage extends React.Component {
...
+    }
+
     render() {
-        const {otcRef} = this.context;
+        const {otcRef, lastPage, pageData} = this.context;
         const showOTCForm = !!otcRef;
+        const isGiftMode = lastPage === 'gift' && !!pageData?.gift;
+
+        if (isGiftMode) {
+            return this.renderGiftLayout(showOTCForm);
```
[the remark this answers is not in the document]
minimaluminium: We actually rely on `lastPage === 'gift'` here on purpose. The redeemGift action explicitly sets lastPage: 'gift' when transitioning to the magic link page, and popup-modal.js uses the same 'gift' sentinel. So lastPage here isn't auto-derived from the previous page name — it's a manually-set marker the action chose. Switching to 'giftRedemption' would actually break the gift-mode rendering. Leaving as-is.

apps/portal/src/components/pages/gift-page.js
```diff
@@ -1,357 +1,637 @@
...
+
+    if (!activePrice) {
+        return '';
+    }
+
+    const currencySymbol = getCurrencySymbol(activePrice.currency);
+    return `${currencySymbol}${formatNumber(getStripeAmount(activePrice.amount))}`;
+}
+
+function getDurationLabel(selectedInterval) {
```
sagzy, line 462: nit: there is an existing `getGiftDurationLabel` that we could have re-used here

apps/portal/src/components/pages/gift-success-page.js
```diff
@@ -1,145 +1,211 @@
...
 
+const ChevronIcon = () => (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth="2.5" strokeLinecap="round" strokeLinejoin="round" aria-hidden="true">
+        <polyline points="6 9 12 15 18 9"/>
+    </svg>
+);
+
+function getDurationLabel(cadence) {
+    return cadence === 'month' ? '1 month' : '1 year';
+}
```
sagzy, line 91: nit: there is an existing getGiftDurationLabel that we could have re-used here

sagzy (commented):
Refactoring opportunity: we could extract a `<GiftCard>` and / or  `<GiftDetailsToggle>` components, given that all 4 screens use the same right-side design. Not blocking, and can be done in a follow-up PR

PR #27670, opened 2026-05-05
🐛 Fixed bookmark favicon storage on object-storage adapters

rob-ghost:
## Problem
Bookmark cards stopped rendering favicons on Pro because the bookmark fetcher passed an absolute filesystem path to the storage adapter. Investigation revealed the bug was the user-visible symptom of a wider design issue: the storage adapter contract was permissive — accepting absolute paths, paths already prefixed with the storage root, decorative leading slashes, etc. — and its consumers (the bookmark fetcher, the importer image handler, the external-media-inliner, and the image-sizes middleware) each violated it in different ways. The two adapters (LocalStorageBase and S3Storage) tolerated the violations differently, so callers got away with it on local dev while producing malformed bucket keys on object-storage backends.

## Solution
Tighten the contract: both adapters now require canonical relative paths via a shared `assertCanonicalRelativePath` validator. Migrate all four callers to pass the canonical shape. Drops the silent tolerance helpers from `LocalStorageBase`. Mutating methods (`save` / `saveRaw` / `delete` / `read`) throw on bad input; `exists()` is lenient and returns `false` since it's a query, not a mutation.

The empty string is the only "non-relative-looking" input the validator accepts, and only for `targetDir` arguments — it represents the storage root itself, which the importer needs for files at the top of an import zip.

## Migration
Forward-only. No existing storage objects move or rename; URLs already stored in posts continue to resolve to whatever object they referenced before. New writes from previously-broken callers (importer + media-inliner) now land at the keys their referencing URLs expect.

## Comparison with #27671
This is the strict half of a strict-vs-permissive design debate. PR #27671 is the permissive alternative — same user-visible fix, but the adapter absorbs the multi-shape inputs instead of forcing each caller to be correct. Both PRs are open so the contract direction can be picked deliberately.

ref https://linear.app/ghost/issue/ONC-1673

### DIFF
--- a/ghost/core/core/frontend/web/middleware/handle-image-sizes.js
+++ b/ghost/core/core/frontend/web/middleware/handle-image-sizes.js
@@ -111,14 +111,19 @@ module.exports = function handleImageSizes(req, res, next) {
         return redirectToOriginal();
     }
 
-    storageInstance.exists(req.url).then((exists) => {
+    // Strip the leading slashes that Express + the format/size regex remove
+    // in the URL space so storage adapters see canonical relative paths.
+    const storageRelativePath = req.url.replace(/^\/+/, '');
+    const originalImagePath = imagePath.replace(/^\/+/, '');
+
+    storageInstance.exists(storageRelativePath).then((exists) => {
         if (exists) {
             return;
         }
 
-        return imageSize.getOriginalImagePath(imagePath)
-            .then((storagePath) => {
-                return storageInstance.read({path: storagePath});
+        return imageSize.getOriginalImagePath(originalImagePath)
+            .then((originalRelativePath) => {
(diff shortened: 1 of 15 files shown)


rob-ghost:
Closing in favour of #27671, which lands the same user-visible fix (favicons restored) with a much smaller diff by absorbing the multi-shape inputs at the adapter layer rather than tightening the contract and migrating callers.

The strict approach in this PR is still defensible — it makes the storage adapter contract explicit and would prevent future callers from reintroducing similar bugs — but the cost (validator module, file-vs-directory distinction, three caller migrations, ~25 test fixtures touched) outweighs the benefit at this point. Re-opening this if we decide later that the permissive contract is causing problems.

PR #27671, opened 2026-05-05
🐛 Tolerated absolute and pre-prefixed paths in S3Storage

rob-ghost:
## Problem
Bookmark cards stopped rendering favicons on Pro (ONC-1673). Investigation showed this is the user-visible slice of a wider pattern: three callers — the bookmark fetcher, the importer image handler, and the external-media-inliner — build storage `targetDir` arguments by joining `getContentPath(...)` or `getTargetDir(storage.storagePath)` with extra segments, producing either absolute filesystem paths or paths that already include the storagePath prefix. `LocalStorageBase` tolerates both shapes; `S3Storage` doesn't — `path.posix.join` concatenates them onto storagePath verbatim, producing malformed bucket keys with the filesystem prefix embedded or the storagePath duplicated. The `exists()` probe then targets a different key than the eventual write, defeating uniqueness checks; on stricter bucket policies the malformed HEAD throws non-NotFound errors, surfacing as the favicon fallback users have been reporting.

## Solution
Tolerate the three input shapes callers actually pass — absolute filesystem path, path already prefixed with the storage root, and decorative leading slash — by canonicalising each before composing the bucket key. Mirrors how `LocalStorageBase` absorbs the same variants. Existing path-traversal protection still fires for `..` segments.

## Migration
Forward-only. No existing S3 objects move or rename; URLs already stored in posts continue to resolve to whatever object they pointed at before. Existing callers passing properly-relative paths see no behaviour change. Importer-side, where post HTML previously referenced a key that the file was never actually written to, the fix heals new writes by landing them at the matching key.

ref https://linear.app/ghost/issue/ONC-1673

### DIFF
--- a/ghost/core/core/server/adapters/storage/S3Storage.ts
+++ b/ghost/core/core/server/adapters/storage/S3Storage.ts
@@ -395,7 +395,7 @@ export default class S3Storage extends StorageBase {
             });
         }
 
-        const pathWithStorage = path.posix.join(this.storagePath, relativePath);
+        const pathWithStorage = path.posix.join(this.storagePath, this.toCanonicalRelativePath(relativePath));
 
         if (!pathWithStorage.startsWith(this.storagePath + '/') && pathWithStorage !== this.storagePath) {
             throw new errors.IncorrectUsageError({
@@ -408,6 +408,42 @@ export default class S3Storage extends StorageBase {
         }
 
         return `${this.tenantPrefix}/${pathWithStorage}`;
+    }
+
+    private toCanonicalRelativePath(input: string): string {
+        return this.fromAbsoluteFilesystemPath(input)
+            ?? this.fromStoragePathPrefixed(input)
+            ?? this.fromLeadingSlashPath(input)
+            ?? input;
+    }
+
+    private fromAbsoluteFilesystemPath(input: string): string | null {
+        if (!path.posix.isAbsolute(input)) {
+            return null;
+        }
+        const marker = `/${this.storagePath}/`;
+        const idx = input.lastIndexOf(marker);
+        if (idx !== -1) {
+            return input.slice(idx + marker.length);
+        }
+        if (input.endsWith(`/${this.storagePath}`)) {
+            return '';
+        }
+        return null;
+    }
+
+    private fromStoragePathPrefixed(input: string): string | null {
+        if (input === this.storagePath || input.startsWith(`${this.storagePath}/`)) {
+            return path.posix.relative(this.storagePath, input);
+        }
+        return null;
+    }
+
+    private fromLeadingSlashPath(input: string): string | null {
+        if (!path.posix.isAbsolute(input)) {
+            return null;
+        }
+        return input.replace(/^\/+/, '');
     }
 
     private isNotFound(error: unknown): boolean {
--- a/ghost/core/test/unit/server/adapters/storage/s3-storage.test.ts
+++ b/ghost/core/test/unit/server/adapters/storage/s3-storage.test.ts
@@ -494,6 +494,102 @@ describe('S3Storage', function () {
         );
     });
 
+    describe('absolute and pre-prefixed paths are normalised before building keys', function () {
+        // Some legacy callers build their target by joining `getContentPath(...)`
+        // with an extra segment, producing an absolute filesystem path. Without
+        // normalisation the absolute prefix gets concatenated into the bucket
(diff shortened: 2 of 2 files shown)


rob-ghost:
@allouis tagging you since you've been active in this code recently, I would value your input on this!

With the rollout to S3 storage for images we've seen a latent bug in how absolute paths are handled, which caused bookmark favicons to break. That's a symptom of a wider issue: there's drift in how S3Storage and LocalStorage adapters deal with absolute and relative file paths.

I opted to fix this in S3Storage rather than the favicon flow specifically because I suspect there's other path-resolution wrongness elsewhere that isn't user-visible yet (image imports look like they're storing the absolute on-disk path inside the bucket key, after the prefix).

I opted to stay permissive and add support for absolute paths alongside the other shapes rather than harden what the adapter accepts, because that would require 3+ changes in other places to prevent regressions. You might prefer that solution to this one. Let me know what you think!

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -410,6 +410,42 @@ export default class S3Storage extends StorageBase {
...
+        }
+        return null;
+    }
+
+    private fromLeadingSlashPath(input: string): string | null {
+        if (!path.posix.isAbsolute(input)) {
+            return null;
+        }
+        return input.replace(/^\/+/, '');
+    }
```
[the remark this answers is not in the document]
rob-ghost: Out of scope, its existing behaviour.

ghost/core/core/server/adapters/storage/S3Storage.ts
```diff
@@ -410,6 +410,42 @@ export default class S3Storage extends StorageBase {
         return `${this.tenantPrefix}/${pathWithStorage}`;
     }
 
+    private toCanonicalRelativePath(input: string): string {
```
allouis, line 413: I think we could implement the path transformations inline here, because they all boil down to "get everything after the storage path" or "remove leading slash" - something like:

```typescript
toCanonicalRelativePath(input) {
    const marker = this.storagePath + '/';                                                                                                                                               
    const parts = input.split(marker);       
    if (parts.length > 1) {                                                                                                                                                              
        return parts[parts.length - 1];
    }                                                                                                                                                                                    
    return input.replace(/^\/+/, '');
}
```

Or I guess the `lastIndexOf` approach works here too - I think it's effectively the same

allouis (commented):
I would prefer to fix the call site, and potentially even do something like

```
if (relativePath.startsWith('/') || relativePath.includes(storagePath)) {
  throw IncorrectUsageError('Pass a relative path');
}
```

Because I think supporting all these cases adds some complication to the storage layer, and it continues this pattern of consumers knowing way too much information about storage paths etc….

That said - I do get your reasoning. How would you feel about doing both? We fix the callers but add this support, or is that overkill?

ghost/core/core/server/adapters/storage/S3Storage.ts
rob-ghost: I originally had this implementation, but I found myself having to document the different cases and outcomes in comments so I wanted to see if I could encode the accepted inputs and their outputs into the code itself, hence the different function names etc.

Its more verbose in the code, but also more explicit and literal, so we don't have to rely on comments to explain what we accept or transform or for convention to remember what we accept or pass.

The other method of achieving the same goal is to not _allow_ "invalid" input which would require this kind of manipulation, but that would require changing all the call sites, which I wanted to avoid (in this specific design).

Regarding the suggestion vs the current implementation, we could go with either, I think the outcome is the same. I'll hold off actioning anything right now, I'll reply to the outer suggestion about updating call-sites first.

rob-ghost:
Thanks for taking a look @allouis!

I spent a while on the stricter approach of restricting inputs and updating call-sites but it really spiralled out of control. We have a few distinct paths which all pass something different and need special consideration.

For example, in the "import" path we want images stored at their absolute locations at the root, not relative, so we need some special handling for that case or to update the importer.

Its possible, and I have a PR which demonstrates the scope of change [here](https://github.com/TryGhost/Ghost/pull/27670) but its quite extensive.

In this case I think this PR is a pragmatic choice to fix the issue, and aligning paths is something that requires more time / consideration outside of on-call work.

What do you think?

allouis:
Thank you for exploring the alternative!

I think you're right, I didn't envisage the scope being quite so big - I'm happy with the approach in this PR!

allouis (approved):

PR #27675, opened 2026-05-05
Translated gift subscription Open Graph preview

mike182uk:
ref https://linear.app/ghost/issue/BER-3529

- the social media preview rendered at `/gift/:token` (title, description, noscript fallback) was hardcoded English, leaving non-English sites with mismatched OG metadata when gift links are shared
- wrapped user-facing strings with `t()` and added 5 new keys to the ghost namespace
- duration label uses the established `1 X` / `{n} Xs` two-key split (matching `portal.json`) rather than appending `s` to a bare unit, since plural rules and word order vary by locale
- initialised i18n in the controller test before requiring the controller so the destructured `t` import resolves to the live `i18next` instance


ghost/i18n/locales/context.json
```diff
@@ -421,6 +424,10 @@
     "{amount} off for first {number} months.": "Portal offer page, showing a discount",
     "{amount} off for first {period}.": "Portal offer page, showing a discount",
     "{amount} off forever.": "Portal offer page, showing a discount",
+    "{count} month_one": "Singular form of the month duration label used in the gift subscription Open Graph preview, e.g. '1 month'. i18next picks this form when {count} is 1 (English).",
+    "{count} month_other": "Plural form of the month duration label used in the gift subscription Open Graph preview, e.g. '3 months'. Some languages have additional plural forms (few, many) that translators can add as needed.",
+    "{count} year_one": "Singular form of the year duration label used in the gift subscription Open Graph preview, e.g. '1 year'. i18next picks this form when {count} is 1 (English).",
+    "{count} year_other": "Plural form of the year duration label used in the gift subscription Open Graph preview, e.g. '2 years'. Some languages have additional plural forms (few, many) that translators can add as needed.",
```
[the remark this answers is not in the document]
mike182uk: Skipping — this finding is invalid by design.

`context.json` is auto-generated from `en/ghost.json` by `ghost/i18n/generate-context.js`:

```js
const BASE_PATH = './locales/en';
// ...
for (const file of files) {
    const data = require(filePath);
    for (const key of Object.keys(data)) {
        newContext[key] = context[key] || '';
    }
}
```

It only mirrors keys present in the English locale. Since English's CLDR plural categories are just `one` and `other`, those are the only `{count} month_*` / `{count} year_*` entries that belong in `context.json`.

If I added `_few` / `_many` entries manually:
1. The next `pnpm translate` run would strip them out (the script rebuilds `newContext` from scratch using only English keys).
2. CI would fail with `"context.json is out of date"`.

This also isn't Slovak-specific — Russian, Polish, Czech, Croatian, Serbian, Ukrainian, and Bosnian all use the same `one`/`few`/`many`/`other` set; Arabic adds `zero` and `two` on top. Documenting per-locale variants would balloon the file and immediately desync with every translate run.

Translators see the `_one` / `_other` descriptions and apply the same context to whatever additional plural forms their language's CLDR rules require. Crowdin's plural editor surfaces one description across all form slots natively.

Verified: ran `pnpm --filter @tryghost/i18n translate` — `context.json` is unchanged, no warnings. No code changes needed.

ghost/i18n/locales/ar/ghost.json
```diff
@@ -1,4 +1,16 @@
...
+    "{count} month_two": "",
+    "{count} month_few": "",
+    "{count} month_many": "",
+    "{count} month_other": "",
+    "{count} year_zero": "",
+    "{count} year_one": "",
+    "{count} year_two": "",
+    "{count} year_few": "",
+    "{count} year_many": "",
+    "{count} year_other": "",
```
sagzy, line 13: I'm not sure we need these? 🤔  Wouldn't it be sufficient to have:

```
"{count} month": "",
"{count} months": "",
"{count} year": "",
"{count} years": "",
```

ghost/core/core/server/web/gift-preview/controller.js
```diff
@@ -1,9 +1,14 @@
...
+const {t} = require('../../services/i18n');
 
 function getCadenceLabel(cadence, duration) {
-    return duration === 1 ? `1 ${cadence}` : `${duration} ${cadence}s`;
+    if (cadence === 'year') {
+        return t('{count} year', {count: duration});
+    }
+
+    return t('{count} month', {count: duration});
 }
```
sagzy, line 12: nit: works today, but would need adapting if we introduce 3 months, 6 months, etc.

ghost/i18n/locales/bg/ghost.json
```diff
@@ -1,4 +1,8 @@
 {
+    "{count} month_one": "",
+    "{count} month_other": "",
+    "{count} year_one": "",
+    "{count} year_other": "",
```
sagzy, line 5: Also curious that it outputs different translation strings here 🤔

ghost/core/core/server/web/gift-preview/controller.js
sagzy: Update: actually I'm not sure anymore by looking at the generated translations lol. Does it cover plural forms already?

ghost/i18n/locales/ar/ghost.json
mike182uk: Its what our i18n tool generates - [this](https://github.com/TryGhost/Ghost/pull/27675#discussion_r3190229872) might explain it better

ghost/core/core/server/web/gift-preview/controller.js
mike182uk: Yep, thats what the `year_many` is for

sagzy: Ah ok! TIL. All good then!

sagzy (approved):

ghost/i18n/locales/bg/ghost.json
sagzy: update: understood now, never mind me :)

PR #27692, opened 2026-05-06
Added `GET automations/:id` route with placeholder static payload

cmraible:
refs https://linear.app/ghost/issue/NY-1265/create-endpoint-for-reading-a-single-automation

This adds the scaffolding for the GET `automations/:id` endpoint in the Admin API, with a hardcoded payload. The intent here is simply to unblock development of the frontend, while we iron out the rest of the schema. This will eventually be updated to return automations from the database, once the schema is in place.

### DIFF
--- a/ghost/core/core/server/api/endpoints/automations.js
+++ b/ghost/core/core/server/api/endpoints/automations.js
@@ -43,9 +43,9 @@ const controller = {
                 updated_at: '2026-05-05T00:00:00.000Z',
                 actions: [{
                     id: '67f3f3f3f3f3f3f3f3f3f3f4',
-                    type: 'wait',
+                    type: 'delay',
                     data: {
-                        wait_hours: 24
+                        delay_hours: 24
                     }
                 }, {
                     id: '67f3f3f3f3f3f3f3f3f3f3f5',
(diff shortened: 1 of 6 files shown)


ghost/core/test/e2e-api/admin/automations.test.js
cmraible, line 127: This was causing secretlint to fail, hence the change in .secretlintrc.json, which carves an exception for string interpolation with ${}

ghost/core/core/server/api/endpoints/automations.js
```diff
@@ -24,6 +24,49 @@ const controller = {
...
+                        email_sender_email: null,
+                        email_sender_reply_to: null,
+                        email_design_setting_id: '680000000000000000000001'
+                    }
+                }],
+                edges: [{
+                    source_action_id: '67f3f3f3f3f3f3f3f3f3f3f4',
+                    target_action_id: '67f3f3f3f3f3f3f3f3f3f3f5'
+                }]
+            };
```
cmraible, line 66: @troyciesco this is sort of a guess of what you'd need on the frontend, feel free to adjust it as needed

troyciesco (approved):

PR #27697, opened 2026-05-06
Diskless redirects: introduce storage abstraction (Phase 1)

allouis:
## Summary

Phase 1 of the [Diskless Redirects Config](https://linear.app/ghost/project/diskless-redirects-config-c4c5916f50a8) project. Replaces the legacy `CustomRedirectsAPI` with a Repository-pattern split — `RedirectsService` orchestrating a `RedirectsStore` (FileStore for now, GCSStore in Phase 2). No infrastructure change, only a refactor and one user-visible flip.

The store boundary is the seam Phase 2 plugs into. `activate()` is independently callable so the future cross-instance notifier can rebuild in-memory state without re-persisting.

## User-visible change

The admin redirects download endpoint now always returns YAML (`redirects.yaml`) regardless of the original upload format. Previously JSON uploads downloaded as JSON, YAML uploads as YAML — making the download → edit → re-upload loop awkward when the format varied. YAML is the recommended format per the docs and the only one the existing tooling generates examples for.

Self-hosters who SSH-edit `redirects.yaml` directly on disk are unaffected. Self-hosters who use the admin upload endpoint will see their YAML comments / anchors / formatting stripped after the first upload (FileStore writes JSON regardless of input format). The replaced YAML is preserved as a timestamped backup in `content/data/`.

## Stack (best reviewed commit-by-commit)

1. \`Improved tests for CustomRedirectsAPI before refactor\` — characterization tests, locks current behaviour
2. \`Added store contract tests and RedirectConfig typedef\` (HKG-1692) — the seam
3. \`Extracted redirect-config parsing and added YAML serialisation\` (HKG-1693) — pure-function parser, hardened against list-shaped uploads, multi-line values, special chars
4. \`Implemented FileStore for redirects\` (HKG-1694) — repository pattern, atomic writes, timestamped backups
5. \`Added RedirectsService with standalone activate()\` (HKG-1695) — thin orchestrator, per-item skip-and-log on bad rules
6. \`Wired RedirectsService and FileStore into the redirects module\` (HKG-1696) — module boundary swap, exports unchanged
7. \`🎨 Redirects download now returns YAML\` (HKG-1697) — the user-visible flip
8. \`Removed legacy CustomRedirectsAPI\` (HKG-1698) — dead code sweep

## Test plan

- [x] \`pnpm test:single test/unit/server/services/custom-redirects/\` (55 tests)
- [x] \`pnpm test:single test/e2e-api/admin/redirects.test.js\` (download + upload)
- [x] \`pnpm test:single test/legacy/api/admin/redirects.test.js\`
- [x] \`pnpm lint\` clean (only pre-existing unrelated warnings)
- [x] CodeRabbit review on this PR
- [ ] Manual upload + download verification on a test instance with a YAML config


allouis:
@coderabbitai review

allouis:
Addressed CodeRabbit's 5 findings. Force-pushed.

@coderabbitai review

allouis:
@coderabbitai — for visibility, here's where each of the round-1 findings was addressed in the force-pushed stack (head 2c48126):

| Finding | Resolved in |
| --- | --- |
| `file-store.js` `replaceAll()` ordering — backup before durable write | `xkquqozr` (Implemented FileStore for redirects). Same-path case copies to backup before atomic rename; different-path case writes new canonical first, then moves old to backup. New test `preserves the previous redirects on disk when the atomic write fails` simulates a write failure and asserts the canonical path still serves the original redirects. |
| `redirect-config-parser.js` — accepts mappings with no 301/302 sections | `otqpxlsl` (Extracted redirect-config parsing). Added explicit `Object.hasOwn(configYaml, '301') \|\| Object.hasOwn(configYaml, '302')` guard. New test rejects `foo: bar` with `BadRequestError`. |
| `redirects-service.js` `replace()` — persisted-but-partially-active state | `xoluwmkk` (Added RedirectsService with standalone activate). Added `_verifyAllLoadable()` pre-flight that loads every entry into a throwaway `DynamicRedirectManager` and throws `ValidationError` on the first rejection, before `store.replaceAll()` runs. Factory injectable for tests. New test asserts no persist + no live-manager call when the dry run rejects. |
| `validation.js` — accepts non-string `from` / `to` | `upxwxozl` (Added store contract tests). Replaced truthy check with `typeof === 'string' && trim().length > 0`. New tests cover object `from`, object `to`, and whitespace-only `from`. |
| `file-store.test.js` — non-deterministic backup-collision test | `xkquqozr`. Test now injects a fixed `getBackupFilePath` via the constructor so both writes resolve to the same backup target deterministically, independent of wall-clock seconds. |

Please verify when convenient.

allouis:
Addressed the round-2 nit on _loadIntoManager null-return logging. Force-pushed to head 469b70c.

@coderabbitai review

allouis:
Round-3 fixes pushed (head 1513dbd):

- validation.js — guard entry shape (`null` / non-object) before property access; raise ValidationError instead of TypeError. Tests added for null + primitive entries.
- redirects-service.js — `_verifyAllLoadable` wraps the dry-run `addRedirect` in try/catch and rethrows as ValidationError, unifying the throw and null-return failure modes.
- redirects.js endpoint — extracted `parseUpload(content, ext)` helper that case-insensitively normalises the extension and accepts `.yml` as an alias, keeping the upload query method within the project's max-api-complexity rule.

@coderabbitai review

allouis:
Round-4 fixes pushed (head c173d01):

- file-store.js — when the post-write `_backup()` of a legacy yaml fails, the freshly-written `redirects.json` is rolled back so the canonical read returns the surviving yaml consistently rather than an unreachable new json. New test simulates the failure and asserts the rollback.
- store-contract.js — added a `getAll()` defensive-copy contract test so any store implementation must hand callers data they can mutate without affecting internal state. Mirrors the existing `replaceAll` input-mutation test.

@coderabbitai review

allouis:
Round-5 nits pushed (head 1962ffb):

- redirect-config-parser.js — `parseYaml` now rejects non-string redirect targets (`: 123`, `: ~`) at parse time so the error points at the YAML structure rather than a coerced entry. Tests added for both cases.
- file-store.test.js — collision test now uses distinct yaml payloads on the two passes and asserts the stable backup path contains the second-pass content. The previous assertion couldn't actually distinguish overwrite from no-op.
- e2e-api/admin/redirects.test.js — added `upload (yaml)` case so the extension-driven YAML branch is covered alongside the existing JSON upload.

@coderabbitai review

allouis:
Round-6 nit pushed (head 2ed05af):

- e2e `upload (yaml)` test now follows the upload with a download and asserts the just-uploaded entry round-trips through the live config. Status + body alone couldn't distinguish a successful upload from a silent no-op.

@coderabbitai review

allouis:
@coderabbitai review

allouis:
@coderabbitai review

allouis:
@coderabbitai review

allouis:
@coderabbitai review

ghost/core/core/server/services/custom-redirects/redirects-service.ts
```diff
@@ -0,0 +1,171 @@
...
+     * that survives `validate()` but fails the manager's stricter
+     * buildRegex rejects the whole upload. Without this,
+     * `store.replaceAll()` could succeed and the per-item skip path
+     * would leave disk and memory partially synchronised.
+     */
+    async replace(redirects: RedirectConfig[]): Promise<void> {
+        this.validate(redirects);
+        this._verifyAllLoadable(redirects);
+        await this.store.replaceAll(redirects);
+        this._loadIntoManager(redirects, {validatePerItem: false});
```
vershwal: We validate/dry-run before writing, but after the write this still goes through `_loadIntoManager()`, which catches per-item failures and only logs them. So if the live manager rejects something that the dry-run accepted, the upload could still return success while the store and in-memory redirects are different.

Would it be safer for this post-write load to throw if any redirect fails? Since validation has already passed, that feels like an unexpected state we should surface.

vershwal (commented):
Nice refactor overall. Just one comment on the replace() flow, mostly around making sure the store and in-memory redirects don’t drift if the post-write load fails.

vershwal (approved):

PR #27700, opened 2026-05-06
Adding dislikes to comments UI and admin

weylandswart:
Ref https://linear.app/ghost/issue/BER-3549/create-comment-dislikes-private-lab-flag https://linear.app/ghost/issue/BER-3550/adjust-ui-for-liking-and-disliking-comments-in-comments-ui https://linear.app/ghost/issue/BER-3551/adjust-ui-for-comments-in-admin https://linear.app/ghost/issue/BER-3553/and-comment-dislikes-modal-to-admin

Adds downvotes/dislikes to comments behind the commentDislikes feature flag. Also adds a measure for these to Admin in the Comments list.

Screenshot 2026-05-19 at 19 24 10

Screenshot 2026-05-19 at 19 24 21

Screenshot 2026-05-19 at 19 24 40


ghost/core/core/server/data/migrations/versions/6.40/2026-04-22-00-00-add-comment-dislikes-table.js
```diff
@@ -0,0 +1,9 @@
+const {addTable} = require('../../utils');
+
+module.exports = addTable('comment_dislikes', {
+    id: {type: 'string', maxlength: 24, nullable: false, primary: true},
+    comment_id: {type: 'string', maxlength: 24, nullable: false, unique: false, references: 'comments.id', cascadeDelete: true},
+    member_id: {type: 'string', maxlength: 24, nullable: false, unique: false, references: 'members.id', cascadeDelete: true},
+    created_at: {type: 'dateTime', nullable: false},
+    updated_at: {type: 'dateTime', nullable: false}
+});
```
jonatansberg, line 9: Have we considered modeling this as another field on the like rather than a separate table? Right now, the way we delete existing (dis)likes if you click both buttons is racy, meaning you can end up both liking and disliking a given comment. While there are other ways to solve for that, just keeping it as one row in the same table makes it way easier. 

It would also mean we can do a single aggregation to get the net score, which should be a slight performance benefit. 

Let's chat about it in the mob.

jonatansberg (commented):
Overall, I think this looks pretty good. One architectural question that I think we should dig into before proceeding. 

The rabbit comment about UI accessibility should also be addressed.

PR #27702, opened 2026-05-06
Improved Ember dev server startup

jonatansberg:
## Summary

- Skip Ember test and hinting trees during normal development builds.
- Exclude Mirage files from normal development builds unless explicitly opted in.
- Keep `/tests` support available with `EMBER_INCLUDE_TESTS=true`.

## Why

`ember serve` was building test and Mirage assets by default in development, pulling large dev-only dependencies such as Faker into the default startup path. This keeps the normal dev server focused on the app graph while preserving explicit test-server support.

## Validation

- Ran isolated `ember serve --port 4210 --live-reload false` profiling before and after the change.
- Before: ~31.4s Ember build, 777 Broccoli steps, ~2.85 GB max RSS, 16.6 MiB Faker/Mirage webpack chunk.
- After: server reached `Build successful`, ~19.8s profiled Ember build, 716 Broccoli steps, ~1.68 GB max RSS.
- Commit hooks ran eslint and secret scanning successfully.

### DIFF
--- a/ghost/admin/ember-cli-build.js
+++ b/ghost/admin/ember-cli-build.js
@@ -14,6 +14,7 @@ const environment = EmberApp.env();
 const isDevelopment = environment === 'development';
 const isProduction = environment === 'production';
 const isTesting = environment === 'test';
+const includeTestAssets = !isProduction && (isTesting || process.env.EMBER_INCLUDE_TESTS === 'true');
 
 const postcssImport = require('postcss-import');
 const postcssCustomProperties = require('postcss-custom-properties');
@@ -58,7 +59,7 @@ const codemirrorAssets = function () {
     };
 
     // put the files in vendor ready for importing into the test-support file
-    if (environment === 'development') {
+    if (environment === 'development' && includeTestAssets) {
         config.vendor = codemirrorFiles;
     }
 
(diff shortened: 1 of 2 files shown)


ghost/admin/ember-cli-build.js
```diff
@@ -261,11 +265,11 @@ module.exports = function (defaults) {
 
     // pull things we rely on via lazy-loading into the test-support.js file so
     // that tests don't break when running via http://localhost:4200/tests
-    if (app.env === 'development') {
+    if (app.env === 'development' && includeTestAssets) {
         app.import('vendor/codemirror/lib/codemirror.js', {type: 'test'});
     }
 
-    if (app.env === 'development' || app.env === 'test') {
+    if (includeTestAssets) {
```
kevinansfield: I wonder if it's preferable to use `app.env` as that represents the type of build that's being generated, whereas the module-level environment check only represents the environment the build script is run from 🤔 It might not matter with our current setup but it was definitely an issue in the past for these imports

jonatansberg: Codex updated the post-constructor import guards to evaluate against `app.env`, so those test-only imports follow the build type Ember is generating. The constructor options still use the pre-constructor `EmberApp.env()` value because `app` does not exist yet there.

kevinansfield (approved):

PR #27706, opened 2026-05-06
Removed Portal billing section for gift members

sagzy:
closes https://linear.app/ghost/issue/BER-3613

Hide the "Billing info & receipts" section in Portal's account home page for gift members.

Gift members do not have a card on file or an active Stripe subscription, so the section's "Update" button (which calls `manageBilling`) led to a dead end. Complimentary members were already excluded for the same reason — gift members were an oversight.

## Changes

`BillingSection` already short-circuits when `isComplimentary` is true. Added the same guard for `isGiftMember`, which checks `member.status === 'gift'`.

## Test plan

- [x] Added unit tests covering the section's visibility for paid, gift, and complimentary members
- [x] `pnpm test` passes (25/25)
- [x] `pnpm lint` and `tsc --noEmit` clean
- [x] Manually verify a gift member no longer sees "Billing info & receipts" in Portal
- [x] Manually verify a regular paid member still sees it and "Update" still works

### DIFF
--- a/apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
+++ b/apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
@@ -145,12 +145,12 @@ const PaidAccountActions = () => {
         return null;
     };
 
-    const BillingSection = ({defaultCardLast4, isComplimentary}) => {
+    const BillingSection = ({defaultCardLast4, isComplimentary, isGift}) => {
         const {action} = useContext(AppContext);
         const label = action === 'manageBilling:running' ? (
             <LoaderIcon className='gh-portal-billing-button-loader' />
         ) : t('Update');
-        if (isComplimentary) {
+        if (isComplimentary || isGift) {
             return null;
         }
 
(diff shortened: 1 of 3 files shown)


apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
```diff
@@ -203,7 +204,7 @@ const PaidAccountActions = () => {
                     </div>
                     <PlanUpdateButton isPaid={isPaid} />
                 </section>
-                <BillingSection isComplimentary={isComplimentary} defaultCardLast4={defaultCardLast4} />
+                <BillingSection isComplimentary={isComplimentary} isGift={isGift} defaultCardLast4={defaultCardLast4} />
```
mike182uk: Would we be better just conditionally rendering this in the parent instead of passing state?

```ts
{{isPaidMember && <BillingSection defaultCardLast4={defaultCardLast4} />}}
```

mike182uk (approved):

apps/portal/src/components/pages/AccountHomePage/components/paid-account-actions.js
sagzy: yes good one, updated!

PR #27708, opened 2026-05-06
Updated Shade Avatar implementation

weylandswart:
Ref https://linear.app/ghost/issue/DES-1318/avatars-should-work-with-fallback-initials

- Updated Shade's Avatar implementation to support initial fallback
- Removed bespoke Member and Comment avatar implementations


peterzimon (approved):
Very cool

PR #27713, opened 2026-05-06
Introduced UrlServiceFacade and migrated callers

allouis:
## Summary

Introduces `UrlServiceFacade` — a resource-based abstraction in front of the eager `UrlService` — and migrates every in-tree caller across. Behaviour-preserving; no flag, no functional change. Eager URL service is unchanged underneath.

**Stacked on #27712** (the test PRs land first to provide regression safety nets for commits 1, 6, and 7 of this stack).

ref https://linear.app/ghost/issue/HKG-1738/

## Why

This is the preparatory refactor for an opt-in lazy URL routing experiment (#27714, stacked on this PR). The existing eager `UrlService` precomputes a full resource → URL map at boot through an 8-stage pipeline; we want to be able to swap in an on-demand backend without touching individual callers, but the call sites today are inconsistent — some use `getUrlByResourceId(id)`, others use `getResource(path)` with a different envelope shape.

This PR introduces a thin resource-based facade in front of the eager service and migrates every in-tree caller across, so a single backend boundary exists when the lazy implementation lands.

It also resolves an inconsistency in the resolved-resource shape: the routing-level type (`'posts'`/`'pages'`/`'tags'`/`'authors'`) now wins over any DB type field on the underlying record.

## What it does

- Adds `core/server/services/url/url-service-facade.ts` (new TS file). Exposes `getUrlForResource(resource, opts)`, `ownsResource(routerId, resource)`, `resolveUrl(path)`, `getResourceById(id)`, `hasFinished()`, and lifecycle pass-throughs.
- Migrates every caller of `urlService.getUrlByResourceId(id)` and `urlService.getResource(path)` to the new facade.
- Switches `audience-feedback` and `comments` URL helpers from id-only to taking a post object, so the helpers carry the full record into the URL service. Adds a small `toPlain(modelOrObj)` helper for callers that may have either a Bookshelf model or a plain hash.
- Includes a regression fix for `?fields=url` Content API requests: the API output URL serializer now passes `id` explicitly into the resource (the `attrs` spread alone wouldn't have it after `fields=url` strips everything but `url`, and the eager id-fallback would return `/404/` for every record).

## How to review

**Please review commit-by-commit** (the GitHub Commits tab). The full diff is a sweep across ~10 files, but each commit is small and isolated:

1. **Changed audience-feedback and comment URL helpers to take a post object** — preparatory signature change + `toPlain()` helper.
2. **Added `UrlServiceFacade`** (TS) — facade class only, no callers yet.
3. **Migrated routing controllers and RouterManager to UrlServiceFacade** — first batch.
4. **Migrated API output URL serializer to UrlServiceFacade** — `output/utils/url.js`. Includes the `?fields=url` regression fix and a regression test for it.
5. **Migrated frontend helpers and meta to UrlServiceFacade** — `{{tags}}`, `{{authors}}`, `meta/url`, `meta/author-url`.
6. **Migrated slack/indexnow/comments/audience-feedback to UrlServiceFacade** — last forward-lookup callers.
7. **Migrated reverse URL lookups to facade.resolveUrl** — `getResource(path)` → `facade.resolveUrl(path)`. `url-translator.getTypeAndIdFromPath` and `content-stats-service.getResourceTitle` become async; the ripple stops at their existing async callers.

Three pinning tests in earlier PRs in this stack (#27710, #27711, #27712); they catch regressions in commits 1, 6, and 7 specifically.

## Test plan

- [x] Lint clean
- [x] `pnpm test:unit` passes
- [x] CI green (Acceptance, Legacy, E2E across the previous umbrella PR)
- [ ] Local smoke: `/`, RSS, sitemap, comment notification email

### DIFF
--- a/ghost/core/core/server/api/endpoints/utils/serializers/output/utils/url.js
+++ b/ghost/core/core/server/api/endpoints/utils/serializers/output/utils/url.js
@@ -3,7 +3,16 @@ const urlUtils = require('../../../../../../../shared/url-utils');
 const localUtils = require('../../../index');
 
 const forPost = (id, attrs, frame) => {
-    attrs.url = urlService.getUrlByResourceId(id, {absolute: true});
+    // `forPost` is shared between the posts and pages mappers (pages.js
+    // delegates to posts.js). The router-level resource type therefore has
+    // to be derived from `attrs.type` ('post' / 'page') rather than hardcoded
+    // — otherwise the facade would dispatch a page record against the posts
+    // router. `id` is passed separately because callers may filter `attrs`
+    // via `fields=url`, which strips every attribute except the requested
+    // ones; without the explicit `id`, the eager facade's id-based fallback
+    // hits /404/ for every record.
+    const type = attrs.type === 'page' ? 'pages' : 'posts';
+    attrs.url = urlService.facade.getUrlForResource({...attrs, id, type}, {absolute: true});
 
     /**
      * CASE: admin api should serve preview urls
@@ -44,7 +53,7 @@ const forPost = (id, attrs, frame) => {
 
 const forUser = (id, attrs, options) => {
     if (!options.columns || (options.columns && options.columns.includes('url'))) {
-        attrs.url = urlService.getUrlByResourceId(id, {absolute: true});
+        attrs.url = urlService.facade.getUrlForResource({...attrs, id, type: 'authors'}, {absolute: true});
     }
 
     return attrs;
@@ -52,7 +61,7 @@ const forUser = (id, attrs, options) => {
 
 const forTag = (id, attrs, options) => {
     if (!options.columns || (options.columns && options.columns.includes('url'))) {
-        attrs.url = urlService.getUrlByResourceId(id, {absolute: true});
+        attrs.url = urlService.facade.getUrlForResource({...attrs, id, type: 'tags'}, {absolute: true});
     }
 
     return attrs;
--- a/ghost/core/core/server/services/url/url-service-facade.ts
+++ b/ghost/core/core/server/services/url/url-service-facade.ts
@@ -1,0 +1,166 @@
+/**
+ * UrlServiceFacade
+ *
+ * Sits in front of the URL service so callers can use a stable, resource-based
+ * interface regardless of the underlying implementation. The facade can be
+ * built with two backends:
+ *
+ *  - urlService:     the legacy eager UrlService that precomputes a full
+ *                    resource → URL map at boot.
+ *  - lazyUrlService: an on-demand implementation (LazyUrlService) that
+ *                    computes URLs and ownership per call.
+ *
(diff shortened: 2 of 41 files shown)


ghost/core/core/server/api/endpoints/utils/serializers/output/utils/url.js
```diff
@@ -3,7 +3,11 @@ const urlUtils = require('../../../../../../../shared/url-utils');
 const localUtils = require('../../../index');
 
 const forPost = (id, attrs, frame) => {
-    attrs.url = urlService.getUrlByResourceId(id, {absolute: true});
+    // `id` is passed separately because callers may filter `attrs` via
+    // `fields=url`, which strips every attribute except the requested ones.
+    // Without this explicit `id`, the eager facade's id-based fallback hits
+    // /404/ for every post.
+    attrs.url = urlService.facade.getUrlForResource({...attrs, id, type: 'posts'}, {absolute: true});
```
vershwal: `forPost` is shared between posts and pages — `mappers/posts.js:45` calls it with the post or page model. The hard-coded `type: 'posts'` here overwrites whatever `attrs.type` already had (`'page'`), so pages now get tagged as `posts` resources.

In **eager** mode this is invisible (the facade only reads `resource.id`), so the PR's behaviour-preserving claim still holds. In **lazy** mode (#27714) the facade dispatches off `resource.type`, and pages will all start being routed as posts. It's your call if you want to fix this in this PR or the next one.

```js
const type = attrs.type === 'page' ? 'pages' : 'posts';
attrs.url = urlService.facade.getUrlForResource({...attrs, id, type}, {absolute: true});

File: ghost/core/core/server/services/url/url-service-facade.ts [hunk 2]
```diff
@@ -0,0 +1,162 @@
...
+ */
+
+/**
+ * Routing-level resource. `type` is one of the plural router keys
+ * ('posts', 'pages', 'tags', 'authors'). Concrete records carry additional
+ * fields (slug, published_at, primary_tag, ...) used by permalink templates.
+ */
+export interface Resource {
+    type: string;
+    id?: string;
```
vershwal: Here id is optional, but line 92 does `getUrlByResourceId(resource.id, options)` — passing `undefined` here would return `/404/` silently. We should make this required or add a check in `getUrlByResourceId`.

ghost/core/core/server/api/endpoints/utils/serializers/output/utils/url.js
allouis: Good catch! Fixed by deriving the router type from `attrs.type, and added a regression test in url.test.js!

ghost/core/core/server/services/url/url-service-facade.ts
allouis: Fixed!

vershwal (approved):

PR #27714, opened 2026-05-06
Added experimental lazy URL routing service behind config flag

allouis:
## Summary

Experimental opt-in lazy URL routing service behind `config.lazyRouting` (default `false`). When the flag is on, the existing eager `UrlService` is replaced by an on-demand `LazyUrlService` that resolves URLs from the DB the first time each one is asked for.

**Stacked on #27713 — that PR must merge first.**

ref https://linear.app/ghost/issue/HKG-1738/

## Why

The eager URL service preloads every post / page / tag / author URL at boot through an 8-stage pipeline. On large sites that's a non-trivial chunk of cold-start time and resident memory. The hypothesis: an on-demand implementation can be fast enough in production while shrinking boot time and RSS.

This PR ships the implementation gated by a flag so we can measure against real traffic before deciding whether to default it on.

## What flips on with `config.lazyRouting: true`

- `UrlServiceFacade` routes through `LazyUrlService` instead of the eager service.
- `SiteMapManager` populates from the DB on first sitemap request (per type, paginated). Mirrors the eager URL service's resource queries: posts/pages by `status:published+type:`, tags/authors by `visibility:public`, plus `withRelated: ['tags', 'authors']` for posts so `:primary_tag`/`:primary_author` permalinks resolve.
- `bridge.start()` skips `urlService.queue.start` (no precompute).
- `route-settings.setFromFilePath` skips the readiness poll, since the lazy service is always finished.

Default-off. Flag-off path is unchanged from `main` after #27713.

## How to review

**Please review commit-by-commit.** Each commit is independently reviewable:

1. **Added DB-driven sitemap population for the lazyRouting flag** — sitemap can't subscribe to the `url.added` / `url.removed` events the lazy backend doesn't fire, so it populates from the DB on first request. Includes a generation-token guard against `routers.reset`-mid-populate races.
2. **Added LazyUrlService class for on-demand URL resolution** (TS) — pure for forward lookups; one injected `findResource` hook for reverse lookups. Same NQL `EXPANSIONS` table and `page:true|false` transformer as the eager `UrlGenerator`.
3. **Added config.lazyRouting flag to swap in LazyUrlService at boot** — wires the lazy backend into the facade and skips the eager init / shutdown when the flag is on.

## Test plan

- [x] Unit tests pass for both flag states
- [x] CI green on the previous umbrella PR
- [ ] Smoke test with `lazyRouting: true` locally
- [ ] Run benchmarks against a 250K-post production clone (tracked in the Linear project document, not in this repo)

### DIFF
--- a/ghost/core/core/frontend/services/sitemap/site-map-manager.js
+++ b/ghost/core/core/frontend/services/sitemap/site-map-manager.js
@@ -1,5 +1,7 @@
 const DomainEvents = require('@tryghost/domain-events');
+const config = require('../../../shared/config');
 const {URLResourceUpdatedEvent} = require('../../../shared/events');
+const toPlain = require('../../../server/lib/common/to-plain');
 const IndexMapGenerator = require('./site-map-index-generator');
 const PagesMapGenerator = require('./page-map-generator');
 const PostsMapGenerator = require('./post-map-generator');
@@ -21,6 +23,20 @@ class SiteMapManager {
         this.tags = options.tags || this.createTagsGenerator(options);
         this.index = options.index || this.createIndexGenerator(options);
 
+        // The lazy URL service does not fire url.added / url.removed /
+        // URLResourceUpdatedEvent. When that mode is active the sitemap
+        // populates itself from the database on first request instead.
+        this._lazyRouting = options.lazyRouting === undefined
+            ? config.get('lazyRouting')
+            : options.lazyRouting;
+        this._populated = false;
+        this._populating = null;
+        // Each routers.reset bumps this generation. An in-flight populate
+        // captures the generation it started in; if that generation is no
+        // longer current when the populate settles, its result is discarded
+        // so we don't mark a stale (potentially mid-reset) lookup as ready.
+        this._populationGeneration = 0;
+
         events.on('router.created', (router) => {
             if (router.name === 'StaticRoutesRouter') {
                 this.pages.addUrl(router.getRoute({absolute: true}), {id: router.identifier, staticRoute: true});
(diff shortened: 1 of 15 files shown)


allouis:
Rebased onto current main now that the refactor PR (#27763) has merged. The PR is now just the 3 lazy-feature commits: HKG-1769 (DB-driven sitemap), HKG-1770 (LazyUrlService), HKG-1771 (config flag wire-up).

@coderabbitai review

ghost/core/core/server/services/url/index.js
```diff
@@ -22,7 +23,56 @@ if (process.env.NODE_ENV.startsWith('test')){
...
+// on. The eager UrlService stays in the require graph either way so test code
+// and partially-migrated callers continue to work. The model layer is only
+// pulled in when the flag is on so flag-off boot keeps its existing
+// require-graph shape.
+let lazyUrlService = null;
+if (config.get('lazyRouting')) {
+    const models = require('../../models');
+
+    const loadOne = async (Model, query) => {
+        const result = await Model.findOne(query, {require: false});
```
vershwal: I think we may need to preload tags and authors here for posts, similar to the sitemap path. Otherwise reverse lookups for permalink shapes like /:primary_tag/:slug/ may not have enough data to pass the NQL/filter check.

ghost/core/core/frontend/services/sitemap/site-map-manager.js
```diff
@@ -86,6 +110,106 @@ class SiteMapManager {
...
+            this._loadType(models.Author, 'authors', this.users, facade)
+        ]);
+    }
+
+    async _loadType(Model, type, generator, facade) {
+        const baseOptions = browseOptionsFor(type);
+        let page = 1;
+        // eslint-disable-next-line no-constant-condition
+        while (true) {
+            const result = await Model.findPage({...baseOptions, page});
```
vershwal, line 175: The first sitemap request in lazy mode now has to wait for this full DB populate. On very large sites this could mean a lot of paginated queries before the request can return.

No change required for now but we should test this for a large site - mostly want to make sure the first /sitemap.xml request doesn’t become a very slow request when the flag is enabled.

ghost/core/core/server/services/url/index.js
```diff
@@ -22,7 +23,56 @@ if (process.env.NODE_ENV.startsWith('test')){
...
+    // that here so reverse lookups can't return drafts / private resources
+    // even though we now hit the DB on demand. We also disambiguate posts
+    // vs pages — both share `models.Post` and a permalink template like
+    // `/:slug/` could otherwise return either record.
+    //
+    // For tags and authors we use the public-facing scoped models
+    // (TagPublic, Author) which carry the shouldHavePosts gate, so reverse
+    // lookups can't return tags/users with no published posts (and in
+    // particular can't expose staff User accounts via a guessed slug).
+    const findResource = (type, query) => {
```
vershwal: The `LazyUrlService` tests look good, but I think this production wiring could use one small test as well. The service tests stub `findResource`, so they don’t catch regressions in the post/page split, `status: 'published',` public visibility filters, or the unknown type fallback.

We could add a small test around this function.

ghost/core/core/frontend/services/sitemap/site-map-manager.js
```diff
@@ -86,6 +110,106 @@ class SiteMapManager {
...
         return this[type].getXml(page);
     }
+
+    /**
+     * Populate the sitemap from the database. Used when the lazy URL service
+     * is active and the URL-added/-removed events are not firing. Idempotent;
+     * a second call is a no-op while the first is in flight or has finished.
+     */
+    async ensurePopulatedFromDatabase() {
+        if (!this._lazyRouting || this._populated) {
```
vershwal, line 120: Small concern here: once `_populated` is true in lazy mode, I don’t see anything invalidating it when `posts/tags/authors` change. Does this mean a newly published post won’t show up in the sitemap until routers.reset or a restart?

Maybe we should either invalidate this on model events, or add a small TTL so the sitemap freshness window is bounded.

ghost/core/core/frontend/services/sitemap/site-map-manager.js
allouis: Yup! That's one of the metrics we're measuring - but I think this would be a tradeoff we're okay with

ghost/core/core/frontend/services/sitemap/site-map-manager.js
allouis: We still incrementally add/remove to the sitemap on the model events!

vershwal (approved):

PR #27752, opened 2026-05-07
Rewrote post analytics CSV export endpoint to stream results

cmraible:
fixes https://linear.app/tryghost/issue/ONC-1566/

I recommend reviewing this PR [one commit at a time](https://github.com/TryGhost/Ghost/pull/27752/changes/4eccefea02bf250ba63a63c4af694e835f8a9151).

Rewrites Ghost's post analytics CSV export to stream results to the client instead of buffering the entire CSV in memory before sending. This should fix the export endpoint timing out for sites with large post catalogs before returning even a single byte.

I've tested this manually with a large dataset, and the resulting CSV export is byte-for-byte identical with the current behavior on main. Even if this doesn't fix the timeouts, it is a step in the right direction to stream the CSV rather than buffering it all in memory.


ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
```diff
@@ -0,0 +1,48 @@
...
+type Row = Record<string, unknown>;
+
+// Streaming Transform that converts post objects into CSV bytes one row at a time.
+// Pipe a Readable of post objects in, pipe CSV bytes out.
+export function createCSVTransform(): Transform {
+    let isFirstChunk = true;
+    let fields: string[] | null = null;
+
+    return new Transform({
+        // We receive JS objects from the exporter, not bytes.
```
EvanHahn: nit: I think it's clear without this comment.

```suggestion
```

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
+import {Transform, TransformCallback} from 'stream';
+const papaparse = require('papaparse');
+
+type Row = Record<string, unknown>;
+
+// Streaming Transform that converts post objects into CSV bytes one row at a time.
+// Pipe a Readable of post objects in, pipe CSV bytes out.
+export function createCSVTransform(): Transform {
+    let isFirstChunk = true;
```
EvanHahn: nit: this variable can be removed. Makes the code slightly shorter and also fixes a type error later on.

Specifically, I'd (1) delete this line (2) delete the line that sets it to `false` (3) change `if (isFirstChunk)` to `if (fields === null)`.

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
+import {Transform, TransformCallback} from 'stream';
+const papaparse = require('papaparse');
+
+type Row = Record<string, unknown>;
+
+// Streaming Transform that converts post objects into CSV bytes one row at a time.
+// Pipe a Readable of post objects in, pipe CSV bytes out.
```
EvanHahn: Three nits about this comment:

- The part about "bytes" is incorrect. The output is strings.
- It'd be nice to make this a doc comment so editor hover works.
- I think "Streaming Transform" is a little unclear.

Here's how I'd rewrite it (take it or leave it):

```suggestion
/**
 * Converts post objects into a CSV, one row at a time.
 *
 * Pipe a `Readable` of post objects in, pipe CSV string out.
 */
```

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
```diff
@@ -0,0 +1,48 @@
...
+        transform(post: Row, _encoding: BufferEncoding, callback: TransformCallback) {
+            try {
+                if (isFirstChunk) {
+                    // Lock the column list from the first row's keys, then emit
+                    // the header + the row in a single chunk.
+                    fields = Object.keys(post);
+                    const csv = papaparse.unparse({fields, data: [post]}, {
+                        header: true,
+                        escapeFormulae: true,
+                        newline: '\r\n'
```
EvanHahn: nit: this is the default value and is arguably unnecessary. If you want to remove it:

```suggestion
```

Same comment applies below.

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
```diff
@@ -0,0 +1,48 @@
...
+// Streaming Transform that converts post objects into CSV bytes one row at a time.
+// Pipe a Readable of post objects in, pipe CSV bytes out.
+export function createCSVTransform(): Transform {
+    let isFirstChunk = true;
+    let fields: string[] | null = null;
+
+    return new Transform({
+        // We receive JS objects from the exporter, not bytes.
+        objectMode: true,
+        transform(post: Row, _encoding: BufferEncoding, callback: TransformCallback) {
```
EvanHahn: nit: because we're in object mode, the type of `BufferEncoding` is wrong. This is a problem with `@types/node`, not your code, but I'd change it to this:

```suggestion
        transform(post: Row, _encoding: unknown, callback: TransformCallback) {
```

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
+import {Transform, TransformCallback} from 'stream';
+const papaparse = require('papaparse');
```
EvanHahn: nit: if you rebase/merge `main`, you should be able to do this:

```suggestion
import {Transform, TransformCallback} from 'node:stream';
import * as papaparse from 'papaparse';
```

This has a few advantages:

- no eslint disable
- proper type checking for `papaparse`
- `node:stream` instead of `stream` for portability

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
```diff
@@ -0,0 +1,48 @@
...
+                    header: false,
+                    escapeFormulae: true,
+                    newline: '\r\n'
+                });
+
+                // papaparse never prepends or appends a newline around its output,
+                // so we add exactly one CRLF to separate this row from the previous.
+                callback(null, '\r\n' + csv);
+            } catch (err) {
+                callback(err as Error);
```
EvanHahn: nit: casts like this violate TypeScript's soundness. It's probably fine, but it's fairly easy to keep TS soundness with something like this:

```suggestion
                callback(err instanceof Error ? err : new Error(String(err)));
```

ghost/core/test/unit/api/canary/utils/serializers/output/posts-export-csv.test.ts
```diff
@@ -0,0 +1,94 @@
+import assert from 'node:assert/strict';
+import {Readable, PassThrough, Writable} from 'stream';
+import {createCSVTransform} from '../../../../../../../core/server/api/endpoints/utils/serializers/output/posts-csv-transform';
+const papaparse = require('papaparse');
+const postsSerializer = require('../../../../../../../core/server/api/endpoints/utils/serializers/output/posts');
+
+describe('Unit: posts CSV streaming transform', function () {
+    it('Transforms a stream of objects into CSV matching papaparse output', function (done) {
```
EvanHahn: nit: `done`-style callbacks don't work in Vitest. Fine while we're on Mocha, but this will be slightly harder to change in the future.

You can fix this in several ways, but I'd use [the `text` function of `node:stream/consumers`][0]. Something like this:

```typescript
it('Transforms a stream of objects into CSV matching papaparse output', async function () {
    const data = [
        {id: '1', title: 'First Post', status: 'published'},
        {id: '2', title: 'Second, "Quoted" Post', status: 'sent'}
    ];
    const source = Readable.from(data, {objectMode: true});

    const csvTransform = createCSVTransform();

    source.pipe(csvTransform);

    const csvOutput = await text(csvTransform);

    const expected = papaparse.unparse(data, {
        escapeFormulae: true,
        newline: '\r\n'
    });
    assert.equal(csvOutput, expected);
});
```

Same comment applies to all the tests in this file. 

[0]: https://nodejs.org/docs/latest-v22.x/api/webstreams.html#streamconsumerstextstream

ghost/core/test/unit/api/canary/utils/serializers/output/posts-export-csv.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+        csvTransform.on('data', (chunk: Buffer) => {
+            output += chunk.toString();
+        });
+        csvTransform.on('end', () => {
+            assert.equal(output, '');
+            done();
+        });
+
+        source.pipe(csvTransform);
```
EvanHahn: nit: Could also be simplified with `await text(csvTransform)` like I suggested above. Take it or leave it.

ghost/core/test/unit/api/canary/utils/serializers/output/posts-export-csv.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    });
+});
+
+describe('Unit: posts CSV export serializer', function () {
```
EvanHahn: questions:

- Should we test the happy path too? 
- Should this be in a separate file from the `describe` above?

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
EvanHahn, line 1: Left a bunch of comments on this file. If you want to see how I envision the finished version with all my comments applied, [here you go](https://gist.github.com/EvanHahn/e0134aee26c241f238d78ad294f7ce2b).

ghost/core/core/server/api/endpoints/utils/serializers/output/posts.js
```diff
@@ -54,7 +55,24 @@ module.exports = {
     },
 
     exportCSV(models, apiConfig, frame) {
-        frame.response = papaparse.unparse(models.data);
+        frame.response = function streamResponse(req, res, next) {
+            const csvTransform = createCSVTransform();
+
+            const datetime = (new Date()).toJSON().substring(0, 10);
```
EvanHahn: nit: `datetime` is the wrong name for a string that only includes the date.

ghost/core/core/server/api/endpoints/utils/serializers/output/posts.js
```diff
@@ -54,7 +55,24 @@ module.exports = {
...
+            res.setHeader('Content-Type', 'text/csv; charset=utf-8');
+            res.setHeader('Content-Disposition', `Attachment; filename="post-analytics.${datetime}.csv"`);
+            const cacheControl = res.getHeader('Cache-Control');
+            res.setHeader('Cache-Control', cacheControl ? `${cacheControl}, no-transform` : 'no-transform');
```
EvanHahn: - question: why is this necessary, exactly?
- nit: What if `Cache-Control` already contains `no-transform`?

ghost/core/core/server/services/posts/posts-exporter.js
```diff
@@ -37,31 +49,58 @@ class PostsExporter {
...
+    async #getExportContext() {
+        const [
+            newsletters,
+            labels,
+            tiers
+        ] = await Promise.all([
+            this.#models.Newsletter.findAll(),
+            this.#models.Label.findAll(),
+            this.#models.Product.findAll()
+        ]);
```
EvanHahn, line 160: As an unrelated follow-up, we should probably do this in a single transaction.

ghost/core/core/server/services/posts/posts-exporter.js
```diff
@@ -1,5 +1,38 @@
...
+        return null;
+    }
+
+    if (limit === undefined || limit === null) {
+        return DEFAULT_EXPORT_LIMIT;
+    }
+
+    const parsed = Number.parseInt(limit, 10);
+
+    if (Number.isNaN(parsed)) {
```
EvanHahn: nit: I'd change this to `!Number.isSafeInteger` to avoid imprecise numbers. For example, `parseInt('9999999999999999')` loses precision.

```suggestion
    if (!Number.isSafeInteger(parsed)) {
```

Alternatively, you could specify a max value, like `if (parsed > 10_000)` or something.

ghost/core/core/server/services/posts/posts-exporter.js
```diff
@@ -26,42 +59,128 @@ class PostsExporter {
...
+            });
+
+            if (posts.data.length === 0) {
+                break;
+            }
 
-        const newsletters = (await this.#models.Newsletter.findAll()).models;
-        const labels = (await this.#models.Label.findAll()).models;
-        const tiers = (await this.#models.Product.findAll()).models;
+            const mapped = this.#mapPosts(posts.data, exportContext);
```
EvanHahn: nit: I think this might do unnecessary work if `posts.data.length > remaining`. Might be worth cutting it off sooner.

ghost/core/core/server/services/posts/posts-exporter.js
```diff
@@ -1,5 +1,38 @@
...
+    'count.paid_conversions',
+    'count.clicks',
+    'count.positive_feedback',
+    'count.negative_feedback',
+    'email'
+];
+
+function parseExportLimit(limit) {
+    if (limit === 'all') {
+        return null;
```
EvanHahn: nit: instead of `null`, I think `Infinity` better represents what you mean. That'd let you avoid several `null` checks below.

ghost/core/core/server/services/posts/posts-exporter.js
```diff
@@ -26,42 +59,128 @@ class PostsExporter {
...
-        const newsletters = (await this.#models.Newsletter.findAll()).models;
-        const labels = (await this.#models.Label.findAll()).models;
-        const tiers = (await this.#models.Product.findAll()).models;
+            const mapped = this.#mapPosts(posts.data, exportContext);
+            const remaining = requestedLimit === null ? mapped.length : requestedLimit - emitted;
+
+            for (const row of mapped.slice(0, remaining)) {
+                emitted += 1;
+                yield row;
+            }
```
EvanHahn: nit: this works but you could shorten it a bit with `yield*`:

```suggestion
            const batch = mapped.slice(0, remaining);
            emitted += batch.length;
            yield* batch;
```

In addition to being shorter, I also think it's clearer, and it's also [a little faster in my testing][0].

[0]: https://gist.github.com/EvanHahn/c6dac85f0565ac1d6a138c65c876129b

ghost/core/core/server/services/posts/posts-exporter.js
```diff
@@ -26,42 +59,128 @@ class PostsExporter {
...
+     * @param {number|null} options.requestedLimit - null means unbounded
+     * @param {number} options.pageLimit
+     * @param {ExportContext} options.exportContext
+     * @returns {AsyncGenerator<object>}
+     */
+    async *#streamPosts({filter, order, requestedLimit, pageLimit, exportContext}) {
+        let page = 1;
+        let emitted = 0;
+
+        while (requestedLimit === null || emitted < requestedLimit) {
```
EvanHahn: nit: I think this is fine but could be shortened. If you take my suggestion above to make `requestedLimit` _always_ a number and `Infinity` if there's no limit:

```suggestion
        for (let page = 1; emitted < requestLimit; page++) {
```

Then you can remove the initialization of `page` above and the `page += 1` below.

If you think this `for` usage is too weird, ignore this comment.

ghost/core/test/unit/server/services/posts/posts-exporter.test.js
```diff
@@ -16,6 +18,14 @@ class SettingsCache {
     }
 }
 
+async function collectStream(stream) {
```
EvanHahn: nit: you can replace this whole function with [`Array.fromAsync`][0].

[0]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync

ghost/core/test/unit/server/services/posts/posts-exporter.test.js
```diff
@@ -274,6 +284,163 @@ describe('PostsExporter', function () {
...
+                    const pageData = posts.slice(start, start + options.limit);
+
+                    return {
+                        data: pageData.map(data => createModel({...data, ...options}))
+                    };
+                }
+            };
+
+            const stream = await exporter.export({limit: 75});
+            assert.ok(stream instanceof Readable || typeof stream.pipe === 'function');
```
EvanHahn: nit: should this use `stream.isReadable` instead?

ghost/core/core/server/models/base/plugins/crud.js
```diff
@@ -46,6 +46,37 @@ const applyManualPaginationWindow = (itemCollection, options) => {
...
+        .split(',')
+        .map(orderPart => orderPart.trim().toLowerCase())
+        .some((orderPart) => {
+            const [orderedField] = orderPart.split(/\s+/);
+            return fields.includes(orderedField);
+        });
+};
+
+const applyStableOrder = (model, order) => {
+    if (!order) {
```
EvanHahn: nit: isn't this case impossible?

ghost/core/core/server/models/base/plugins/crud.js
```diff
@@ -114,7 +145,9 @@ module.exports = function (Bookshelf) {
         findPage: async function findPage(unfilteredOptions) {
             const options = this.filterOptions(unfilteredOptions, 'findPage');
             const skipPagination = options.skipPagination === true;
+            const stableOrder = options.stableOrder === true;
```
EvanHahn: I disagree with adding this option for a few reasons:

- "Stable order" is model-dependent. What about models that don't have an `id` attribute? We have [at least one](https://github.com/TryGhost/Ghost/blob/3d5627e4326446edf4b6fded7d5e30582f3624d3/ghost/core/core/server/data/schema/schema.js#L324-L330), and [plan to add a second](https://github.com/TryGhost/Ghost/blob/3d5627e4326446edf4b6fded7d5e30582f3624d3/ghost/core/core/server/services/automations/temporary-fake-database.js#L78-L82).
- The code that does this requires parsing SQL, which is brittle. Tables and columns can have spaces, for example, which would trip this up. So would any capitalization in the column name. _Probably_ not a risk.
- There's already a reasonable alternative for this: add `id` to your order clause.

Broadly, I think this is a leaky and unnecessary abstraction.

Could the post exporter add `id desc` itself?

EvanHahn (changes requested):
This was great to review because

- It was split up into easy-to-review chunks.
- Streaming always makes me happy.
- I learned a few things (got more practice with transform streams, learned about the performance of `yield*` versus `yield`, the `*#` syntax for a private generator).

Lots of little nitpicks, but looks good overall.

cmraible:
@EvanHahn thanks for the thorough review, lots of good stuff here! Planning to address these items tomorrow 👍

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/d2f691e7774ec1bfa7f9a3df30c9699038b2e8c0

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/25234036c4e94e9306ad5385a1925caa3626e19e

EvanHahn:
@cmraible Lemme know when you want me to take another look. (I _think_ you're still working on this.)

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/808fbd0389cce4bc6c10e36a3e80555d22c22cd3

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/25234036c4e94e9306ad5385a1925caa3626e19e

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/808fbd0389cce4bc6c10e36a3e80555d22c22cd3

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Nice one, addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/e609d3efa47ea854469a4227e0cb007aa5710e33

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/946de76c88935b9dac3ef9441e8ec7d8b48fbb73

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Sweeeet, addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/e2ba871ad164c6397546603b61816e368a4ff5cc

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/68ae23c4921f3f009de9819633d488f992bdc0e6

ghost/core/test/unit/api/canary/utils/serializers/output/posts-export-csv.test.ts
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/df071c44dd05ca4d6236c696f23144fa248a0f19

ghost/core/test/unit/api/canary/utils/serializers/output/posts-export-csv.test.ts
cmraible: Also addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/df071c44dd05ca4d6236c696f23144fa248a0f19

ghost/core/test/unit/api/canary/utils/serializers/output/posts-export-csv.test.ts
cmraible: Yeah good call on both counts!

- Split into two files: https://github.com/TryGhost/Ghost/pull/27752/changes/9120153d7aae46bb037b5a9a312bbc9d426f79ee
- Added happy path: https://github.com/TryGhost/Ghost/pull/27752/changes/1eb31c98fd8acd07bfe98eee5ad745bd3443b63a

ghost/core/core/server/api/endpoints/utils/serializers/output/posts.js
cmraible: indeed it is! addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/44e5e7494af0db4cb460d13e6146a5f41f40386b

ghost/core/core/server/api/endpoints/utils/serializers/output/posts.js
cmraible: > question: why is this necessary, exactly?

I'm interpreting this as "why do we need to set headers here?" rather than "why do we need these specific headers?" but let me know if that interpretation is incorrect.

Normally Ghost's API framework handles setting these headers, based on the other values in the endpoint definition itself. Since we set this specific endpoint to `stream: true`, the framework doesn't set the headers or send the response itself, so we have to do it manually.

> nit: What if Cache-Control already contains no-transform?

Good shout, addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/0380fc21b56f409b545235948c9fe0f1dccf11e4

ghost/core/core/server/services/posts/posts-exporter.js
cmraible: Good thought - added a follow up tracked in https://linear.app/ghost/issue/NY-1288/fetch-newsletters-labels-and-products-in-a-single-transaction-without

ghost/core/core/server/services/posts/posts-exporter.js
cmraible: Ah good call! Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/fbcd95d4353f447771b559c75bfffe4a359528f2

ghost/core/core/server/services/posts/posts-exporter.js
cmraible: Indeed it does! Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/eccaaacbfb237aa045e534f7cb87456db0255068

ghost/core/core/server/services/posts/posts-exporter.js
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/df10b7abc070038544f3803041e515112cc205f3

ghost/core/core/server/services/posts/posts-exporter.js
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/c13bf5fbfd4d1a1f3411dd4adb37c57ba776f936

ghost/core/test/unit/server/services/posts/posts-exporter.test.js
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/44175c02724a21fd3bee366e1d7135a3ae27f7bd

ghost/core/test/unit/server/services/posts/posts-exporter.test.js
cmraible: Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/a384c10b495f22b645e53a385b2f61da2af4f39d

ghost/core/core/server/models/base/plugins/crud.js
cmraible: This is irrelevant now since we're removing this option entirely!

ghost/core/core/server/models/base/plugins/crud.js
cmraible: These are all reasonable points!

I do think there is a deeper problem here that should be solved at the Bookshelf or API Framework layer. Generally I don't think API consumers should need to worry about stable ordering for pagination when passing a custom `order` option. That said, I agree this could use more thought and this change is out of place in this PR. 

Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/ffef8fccfa762cf0f0dc8a81cd73e3a42554058d, so the posts exporter adds `id desc` itself - - functionally equivalent for this use-case, but with a smaller blast radius and less fragile.

ghost/core/core/server/services/posts/posts-exporter.js
cmraible: Nice, I'll take a for loop over a while loop any day. Addressed in https://github.com/TryGhost/Ghost/pull/27752/changes/c13bf5fbfd4d1a1f3411dd4adb37c57ba776f936

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Thanks! I think I've addressed all these inline 😄

cmraible:
@EvanHahn thanks! I'm pretty sure I've addressed everything; ready for another look whenever you have time!

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
```diff
@@ -0,0 +1,47 @@
...
+
+                    callback(null, csv);
+                    return;
+                }
+
+                // Subsequent rows reuse the locked-in field list and skip the header.
+                const csv = papaparse.unparse({fields, data: [post]}, {
+                    header: false,
+                    escapeFormulae: true,
+                    newline: '\r\n'
```
EvanHahn: nit: you're setting the newline here, but not above. I think I'd just remove this.

```suggestion
```

ghost/core/core/server/models/base/plugins/sanitize.js
```diff
@@ -47,7 +47,7 @@ module.exports = function (Bookshelf) {
             case 'findAll':
                 return [...baseOptions, ...extraOptions, 'filter', 'columns', 'mongoTransformer'];
             case 'findPage':
-                return [...baseOptions, ...extraOptions, 'filter', 'order', 'autoOrder', 'page', 'limit', 'columns', 'mongoTransformer', 'skipPagination'];
+                return [...baseOptions, ...extraOptions, 'filter', 'order', 'autoOrder', 'page', 'limit', 'columns', 'mongoTransformer', 'skipPagination', 'stableOrder'];
```
EvanHahn: nit: I think this can be reverted. I think it was left over from a previous change.

```suggestion
                return [...baseOptions, ...extraOptions, 'filter', 'order', 'autoOrder', 'page', 'limit', 'columns', 'mongoTransformer', 'skipPagination'];
```

ghost/core/core/server/services/posts/posts-exporter.js
```diff
@@ -1,5 +1,56 @@
...
+    const parsed = Number.parseInt(limit, 10);
+
+    if (!Number.isSafeInteger(parsed)) {
+        return DEFAULT_EXPORT_LIMIT;
+    }
+
+    return Math.max(parsed, 1);
+}
+
+function hasIdOrder(order) {
```
EvanHahn, line 37: I don't have a suggestion, but it feels brittle to parse a SQL sub-expression like this. If you can think of a cleaner way, great...but I cannot.

EvanHahn (approved):
LGTM with a few small nits/questions. Thanks for responding to all my comments!

ghost/core/core/server/models/base/plugins/sanitize.js
cmraible: Ah, good find - thank you! Addressed in https://github.com/TryGhost/Ghost/pull/27752/commits/35aa62ea158acb29e073e68902597a1baf2bba05

ghost/core/core/server/api/endpoints/utils/serializers/output/posts-csv-transform.ts
cmraible: Ah, yes. Removed in https://github.com/TryGhost/Ghost/pull/27752/commits/3aa4dc114cc13837a826a75f2179dab5e5bd04bf

ghost/core/core/server/services/posts/posts-exporter.js
cmraible: Yeah definitely agree, though I haven't been able to come up with a better option either 😕

PR #27755, opened 2026-05-07
Changed default dev command to skip public apps

cmraible:
## Summary

- changed the default pnpm dev stack to run Ghost and admin app dev servers only
- added pnpm dev:apps for the previous full stack with public app dev servers
- updated contributor and public app docs to describe the new commands


cmraible:
This didn't make as significant a difference as I was hoping.

PR #27777, opened 2026-05-07
Removed duplicate automations endpoint tests

EvanHahn:
towards https://linear.app/ghost/issue/NY-1265

_This is a test-only change._

We decided that these tests are mostly redundant with the E2E tests. Let's remove them.


EvanHahn:
We all agreed this would be a good idea (outside GitHub), so I'm going to merge this as soon as CI passes.

PR #27779, opened 2026-05-07
Made fake automations database more realistic

EvanHahn:
ref https://linear.app/ghost/issue/NY-1260
towards https://linear.app/ghost/issue/NY-1265

Several improvements:

- Uses more realistic IDs across the board
- Revisions now have `email_design_setting_id` set
- Uses TEXT timestamps, matching what we do in the real database
- Attach more realistic Lexical data
- Uses prepared statements that are easier to read/edit
- API now exists to get a singleton, or create a new database from scratch

### DIFF
--- a/ghost/core/core/server/services/automations/temporary-fake-database.js
+++ b/ghost/core/core/server/services/automations/temporary-fake-database.js
@@ -10,6 +10,9 @@
  * migration once we're sure this schema is correct.
  */
 
+const errors = require('@tryghost/errors');
+const ObjectId = require('bson-objectid').default;
+
 /**
  * @returns {import('node:sqlite').DatabaseSync}
  */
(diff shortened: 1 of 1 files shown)


ghost/core/core/server/services/automations/temporary-fake-database.js
```diff
@@ -18,28 +21,49 @@ function createTemporaryFakeAutomationsDatabase() {
...
+        }
+    });
+    const fakeEmailDesignSettingId = id();
+
     database.exec(`
 CREATE TABLE automations (
   id TEXT PRIMARY KEY,
-  created_at INTEGER NOT NULL,
-  updated_at INTEGER NOT NULL,
+  created_at TEXT NOT NULL,
```
cmraible, line 48: question: why `TEXT` instead of `DATETIME`?

cmraible (approved):

PR #7579, opened 2016-10-17
Add brute force protection

cobbspur:
This PR is to improve Ghost's spam prevention mechanism and use a persistent data store

This uses a more flexible system using brute-express. Initially I have implemented a global reset block on the password reset endpoint that will lock the endpoint out for a given IP if it exceeds  5 requests per hour regardless of the email address. The endpoint is locked out for 1 hour.

There is then a secondary protection mechanism that links an IP and an email address and locks the endpoint out for that matching pair after 5 password reset requests. This is not limited to requests within a given timeframe but in all time. The endpoint is locked out initially for 10 minutes (this increases in a fibonacci sequence up to a period of a week) e.g. 10, 10, 20, 30, 50, 80...

We have the option of resetting the count if such conditions are met. E.g we could clear the count if the password is successfully updated for an email + IP
- uses express-brute and brute-knex to prevent spamming of password reset endpoint

PR #27784, opened 2026-05-08
Removed dead password reset brute-force stub

lsinger:
## Summary

- Removed the unreachable `protectBruteForce` stub in `services/auth/passwordreset.js`. It read a `tokenSecurity[key].count` value that was never assigned anywhere in the file, so the `count >= 10` guard was always false and the in-file rate limit never triggered.
- Removed the matching `tokenSecurity` constant, the `tokenLocked` message, the `// @TODO` comment pointing at #7579 (2016), and the now-unused `protectBruteForce` export.
- Dropped the awaited call from `api/endpoints/authentication.js` since the function only forwarded `{options, tokenParts}` unchanged on the success path.

## Why

The actual brute-force protection for the password reset flow already lives at the route layer in `web/api/endpoints/admin/routes.js`:

- `PUT /authentication/password_reset` is wrapped by `shared.middleware.brute.globalBlock` (express-brute, DB-persisted in the `brute` table, 50 attempts/IP/hour by default — configurable via `spam.global_block`).
- `POST /authentication/password_reset` adds `globalReset` (per-IP) and `userReset` (per-username).

The TODO predated that middleware being wired up to this route, which is why the in-file counter was never finished — the protection landed at the route layer instead. The stub had no tests covering it (`grep` across `ghost/core/test` returns nothing), wasn't called from any other consumer, and recently caused a false-positive in a security report. Deleting it removes a misleading code path without changing behavior.

### DIFF
--- a/ghost/core/core/server/services/auth/passwordreset.js
+++ b/ghost/core/core/server/services/auth/passwordreset.js
@@ -10,7 +10,6 @@ const mail = require('../mail');
 
 const messages = {
     userNotFound: 'User not found.',
-    tokenLocked: 'Token locked',
     resetPassword: 'Reset Password',
     expired: {
         message: 'Cannot reset password.',
@@ -28,8 +27,6 @@ const messages = {
         help: 'Check if password reset link has been fully copied or request new password reset via the login form.'
     }
 };
-
-const tokenSecurity = {};
 
 function generateToken(email, settingsAPI, transaction) {
     const options = {context: {internal: true}, transacting: transaction};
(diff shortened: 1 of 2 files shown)


ghost/core/core/server/services/auth/passwordreset.js
```diff
@@ -79,18 +76,6 @@ function extractTokenParts(options) {
...
-function protectBruteForce({options, tokenParts}) {
-    if (tokenSecurity[`${tokenParts.email}+${tokenParts.expires}`] &&
-        tokenSecurity[`${tokenParts.email}+${tokenParts.expires}`].count >= 10) {
-        return Promise.reject(new errors.NoPermissionError({
-            message: tpl(messages.tokenLocked)
-        }));
-    }
-
-    return Promise.resolve({options, tokenParts});
-}
```
cmraible, line 92: note / question (non-blocking): I noticed that this unreachable code intends to add brute protection at the _token_ level (e.g. allow up to 10 attempts per token email/expiration pair), which is stricter than the current brute protection at the route level, which is just a generic IP level brute protection. Do you think we should add this stricter protection for this endpoint (at the route layer)?

cmraible (approved):
Thanks for raising this! Agree that this code is unreachable so it makes sense to remove it.

I left one comment/question on intent, but non-blocking — looks good to merge to my eyes 👍

ghost/core/core/server/services/auth/passwordreset.js
lsinger: @cmraible in my perception the current protections have been sufficient -- if you're seeing data to the contrary we should definitely discuss. 💯

PR #27790, opened 2026-05-08
🐛 Fixed product card rating clicks being silently swallowed after sibling card focus change

rob-ghost:
## Problem

Editing a post with a product card immediately following a file card: clicking a rating star on the product card after uploading a file to the file card silently does nothing. The card still appears interactive (settings panel visible, green selection border) but the click never registers.

This was reported in 2023 and has been open in BER triage. Ticket reproducible end-to-end with the `pr-1882` Koenig alpha installed by this PR.

## Solution

Bumps `@tryghost/koenig-lexical` from `1.8.1` to `1.8.2-pr1882.0`. The actual fix lives in TryGhost/Koenig#1882; this PR just pulls it in.

The Koenig change makes `KoenigCardWrapper`'s mousedown handler early-return for clickthrough-marked elements (rather than dispatching `SELECT_CARD_COMMAND`, which was synchronously deselecting sibling cards and causing a layout shift between mousedown and mouseup), and promotes the rating widget to that contract — clicking a star now both updates the rating AND puts the card back into editing state in a single gesture.

This PR will be repointed to the real `1.8.2` (or whichever version) once Koenig#1882 merges and publishes.

refs https://linear.app/ghost/issue/BER-3595
refs https://github.com/TryGhost/Koenig/pull/1882

### DIFF
--- a/apps/admin-x-framework/package.json
+++ b/apps/admin-x-framework/package.json
@@ -76,7 +76,7 @@
         "@playwright/test": "1.59.1",
         "@testing-library/jest-dom": "5.17.0",
         "@testing-library/react": "14.3.1",
-        "@tryghost/koenig-lexical": "1.8.1",
+        "@tryghost/koenig-lexical": "1.8.2-pr1882.0",
         "@types/react": "18.3.28",
         "@types/react-dom": "18.3.7",
         "@vitejs/plugin-react": "4.7.0",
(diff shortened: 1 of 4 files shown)


rob-ghost:
Closing this as we feel this is dealing with the symptoms rather than the cause and is only increasing complexity with click pass-through.

PR #27798, opened 2026-05-09
Extended {{contrast_text_color}} helper to support CSS variable

fueko:
### What
`{{contrast_text_color}}` helper now returns CSS variables instead of hardcoded  hex values, with a fallback to pure white/black for backward compatibility.

### Why
Theme developers often use slightly off-white or off-black text colors for  better visual consistency. Previously there was no way to override the  hardcoded `#FFFFFF`/`#000000` returned by this helper.

By returning `var(--ghost-contrast-color-white)` and  `var(--ghost-contrast-color-black)`, theme developers can now define these  variables in their CSS to match their typography:

```css
:root {
  --ghost-contrast-color-white: #F5F5F5;
  --ghost-contrast-color-black: #1A1A1A;
}
```

### Backward compatible
If the CSS variables are not defined, the fallback values `#FFFFFF` and  `#000000` are used — so existing themes are unaffected.

---

> [!NOTE]
> **Low Risk**
> Low risk: small, localized change to a theme helper’s return values; main risk is behavior/expectation changes where callers or tests assume raw hex strings.
> 
> **Overview**
> Updates the `{{contrast_text_color}}` helper to return CSS `var()` references for pure white/black results (`--ghost-contrast-color-white`/`--ghost-contrast-color-black`) with hex fallbacks, instead of always returning hardcoded `#FFFFFF`/`#000000`.
> 
> Invalid/exception paths now also return the white CSS variable fallback, while non-white/black contrast results continue to return the computed hex value.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee45b89c21a39d24ea4f84404383da53894c76ef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ghost/core/core/frontend/helpers/contrast_text_color.js
+++ b/ghost/core/core/frontend/helpers/contrast_text_color.js
@@ -2,10 +2,12 @@ const {textColorForBackgroundColor} = require('@tryghost/color-utils');
 
 module.exports = function contrast_text_color(color) { // eslint-disable-line camelcase
     const backgroundColor = (typeof color === 'string' && color.trim()) ? color.trim() : '#15171A';
-
     try {
-        return textColorForBackgroundColor(backgroundColor).hex();
+        const contrastColor = textColorForBackgroundColor(backgroundColor).hex();
+        if (contrastColor === '#FFFFFF') return 'var(--ghost-contrast-color-white, #FFFFFF)';
+        if (contrastColor === '#000000') return 'var(--ghost-contrast-color-black, #000000)';
+        return contrastColor;
     } catch (err) {
-        return '#FFFFFF';
+        return 'var(--ghost-contrast-color-white, #FFFFFF)';
     }
 };


9larsons:
Thanks for the PR. There's a real gap here, but I don't think we can take this change.

The helper's current contract is "returns a hex color string" and themes rely on that outside CSS property values. The most concrete break: `{{color_to_rgba (contrast_text_color @site.accent_color) 0.5}}` — `color_to_rgba` parses its input as a real color and throws on `var(...)`, silently falling back to its dark-gray default. The same applies to `<meta name="theme-color" content="…">`, which doesn't resolve CSS variables.

Today, themes can already substitute their own colors by overriding `--gh-private-accent-contrast-color` directly in their stylesheet. That covers fixed-accent themes and themes that want the same off-white/off-black for both branches. The one case it doesn't cover is a theme with a user-configurable accent that wants different off-white vs off-black per branch, since the theme can't tell in static CSS which branch the helper picked.

If that case turns out to be important, a new sibling helper that exposes the decision (e.g. returning "light" or "dark") would be a cleaner shape than narrowing this helper's contract — but I'd want to see a specific theme that needs it before adding API surface. Happy to discuss in an issue if so.

fueko:
Thanks @9larsons for the detailed explanation. I understand, and you’re right — I didn’t consider all of the scenarios where the helper’s return value is used outside of CSS property values. The examples with `color_to_rgba` and `theme-color` make the compatibility concerns much clearer.

It would still be nice to have a helper that solves this particular use case at some point, but it’s not a big deal since it can also be handled on the JavaScript side when needed.

Thanks for taking the time to explain the reasoning and the broader implications. I appreciate it.

PR #27801, opened 2026-05-11
Improved gift subscription link previews

minimaluminium:

Updated the gift preview image to match the current gift card design and use bundled typography for consistent server rendering.


ghost/core/core/server/web/gift-preview/controller.js
```diff
@@ -11,6 +11,28 @@ function getCadenceLabel(cadence, duration) {
...
+    return t(`You've been gifted a {duration}-month {tierName} membership to {siteTitle}`, {
+        duration,
+        tierName,
+        siteTitle,
+        interpolation: {escapeValue: false}
+    });
+}
+
+function getTierName(tier) {
+    return tier.name || tier.toJSON?.().name || '';
```
sagzy: nit: unnecessary helper - if Tiers API returns a non-null Tier object, we can expect `tier.name` to always be a string

ghost/core/core/server/web/gift-preview/image.js
```diff
@@ -21,69 +27,77 @@ function escapeXml(str) {
...
+    return `${text.slice(0, maxLength - 3).trim()}...`;
 }
 
-function buildSvg({accentColor}) {
-    const bgColor = getBackgroundColor(accentColor);
+function buildSvg({accentColor, siteTitle = 'Ghost', tierName = '', cadenceLabel = '1 year'}) {
     const escapedAccentColor = escapeXml(accentColor);
+    const orbImageHref = escapeXml(getGiftCardOrbImageHref());
+    const safeCadenceLabel = escapeXml(truncateText(cadenceLabel, 18));
+    const safeTierLabel = escapeXml(truncateText(tierName ? `${tierName} membership` : 'Gift membership', 38));
```
sagzy: Tier label is missing the i18n wrapper

ghost/core/core/server/web/gift-preview/fonts.conf
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
+<fontconfig>
+    <dir prefix="relative">.</dir>
+    <dir>/usr/share/fonts</dir>
+    <dir>/usr/local/share/fonts</dir>
+    <dir>/System/Library/Fonts</dir>
+    <dir>/Library/Fonts</dir>
+    <cachedir>/tmp/ghost-gift-preview-font-cache</cachedir>
```
sagzy: `/tmp` might not always be writable (e.g. for self-hosters). Think the simplest would be to omit `<cachedir>` and let fontconfig pick its default

ghost/core/core/server/overrides.js
```diff
@@ -12,3 +13,11 @@ const moment = require('moment-timezone');
...
 luxon.Settings.defaultZone = 'UTC';
 moment.tz.setDefault('UTC');
+
+/**
+ * Point fontconfig at the bundled Inter font used for gift subscription
+ * link previews. This must run before sharp/libvips initializes fontconfig
+ * (on first text render) — once fontconfig caches its config, changing the
+ * env var is a no-op.
+ */
+process.env.FONTCONFIG_FILE ??= path.join(__dirname, 'web/gift-preview/fonts.conf');
```
sagzy: This feels a bit itchy, as it changes the fontconfig behaviour globally for sharp/libvips 🤔 

Asked GPT 5.5 for an alternative, here's what it suggested (looks reasonable, but need to spike it to see if it actually works):



#### Preferred solution: Render text with Sharp’s per-call fontfile support instead of SVG <text>:

1. Remove this from overrides.js (line 23):


```js
process.env.FONTCONFIG_FILE ??= path.join(__dirname, 'web/gift-preview/fonts.conf');
```




2. Delete ghost/core/core/server/web/gift-preview/fonts.conf.



3. Add sharp as an explicit ghost/core dependency, or expose a helper from @tryghost/image-transform. I’d prefer explicit sharp only if this stays local to gift previews; otherwise extend @tryghost/image-transform.



4. Change image.js (line 53) so the SVG only contains the background/artwork. Then composite text overlays:



```js

const sharp = require('sharp');

const INTER_FONT_PATH = path.join(__dirname, 'Inter.ttf');

function textOverlay({text, left, top, width, height, fontSize, weight = 700}) {
    return {
        input: {
            text: {
                text: escapePango(text),
                font: `Inter ${weight}`,
                fontfile: INTER_FONT_PATH,
                width,
                height,
                rgba: true
            }
        },
        left,
        top
    };
}

const image = await sharp(Buffer.from(backgroundSvg), {animated: false})
    .resize(1200, 630, {withoutEnlargement: false})
    .composite([
        textOverlay({text: safeCadenceLabel, left: 64, top: 75, width: 1000, height: 100, fontSize: 84}),
        textOverlay({text: safeTierLabel, left: 66, top: 175, width: 1000, height: 70, fontSize: 44, weight: 500}),
        textOverlay({text: safeSiteTitle, left: 66, top: 500, width: 1000, height: 70, fontSize: 52})
    ])
    .png()
    .timeout({seconds: 10})
    .toBuffer();
```

ghost/core/core/server/web/gift-preview/image.js
```diff
@@ -21,75 +27,131 @@ function escapeXml(str) {
...
 
     if (cache.has(cacheKey)) {
         return cache.get(cacheKey);
     }
 
-    const imageTransform = require('@tryghost/image-transform');
+    const sharp = require('sharp');
+
+    sharp.cache(false);
+    sharp.concurrency(1);
```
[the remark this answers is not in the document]
sagzy: thanks! fixed

PR #27808, opened 2026-05-11
Removed unnecessary /setup/done route

kevinansfield:
## Summary
- keeps the minimal Ember onboarding service for setup handoffs
- starts the onboarding checklist directly from post-setup and ?firstStart=true
- redirects both entry points straight to /setup/onboarding?returnTo=/analytics
- removes the obsolete Ember setup.done route and the unused React /setup/done route


jonatansberg (commented):
Something about the shape of this feels a bit off to me. The `/setup/done` route feels a bit out of place to me in a React world. Since the only callers are Ember setup completion, could we keep the command at those Ember intent points, call the existing onboarding start service there, and navigate directly to `/setup/onboarding?returnTo=/analytics`? That avoids a React route whose only purpose is “run mutation in an effect, then redirect”.

```
// Ember setup/home command point
if (this.session.user.isOwnerOnly) {
    await this.onboarding.startChecklist();
}

window.location.hash = '/setup/onboarding?returnTo=/analytics';
```

That does mean keeping the tiny Ember onboarding service for now, but that seems like the better temporary migration boundary than adding a redirect-only React route with effect/ref dedupe logic.

kevinansfield:
It used to be that Ghost(Pro) landed on `/setup/done` once it handled the site creation, bypassing the other Ember setup screens. Maybe that's no longer the case though, I'll double-check. If it's no longer the case then we could drop the route entirely

kevinansfield:
Ok, the post-setup landing path is now `/ghost/?firstStart=true` so we can get rid of the `/setup/done` route entirely. I'll update the PR

PR #27809, opened 2026-05-11
Added noise effect to gift card preview and gift link preview image

minimaluminium:

Adds a subtle noise texture to the gift card preview and the gift link OG preview image, and reworks the card's notch into a brand-tinted element for a more polished look.

## Changes

**Portal (in-app gift card)**
- New noise texture overlay (`gift-card-noise.webp`) applied to `.gh-portal-gift-checkout-card` via a `::after` pseudo-element at 10% opacity
- Orb gradient moved from `::after` → `::before` to make room for the noise layer
- `isolation: isolate` added to the card so the noise stays contained in its own stacking context
- Notch converted from a `::before` pseudo-element to a real `<div class="gh-portal-gift-checkout-card-notch">` (with `aria-hidden`) — needed because both pseudo slots are now used for orb + noise
- Notch background switched from `rgba(0, 0, 0, 0.35)` to `color-mix(in srgb, var(--brandcolor) 65%, #000 35%)` so it picks up the site's accent color
- Card site footer's white background moved to a `::before` so the noise overlay can sit above it
- Portal version bumped to `2.68.37`

**Server (OG preview image)**
- New `gift-card-noise.png` tiled across the rendered 1200×630 image via Sharp's `composite({ tile: true })`
- New `buildNotchOverlay` SVG composite, tinted with `mixWithBlack(accentColor, 0.65)` to mirror the CSS `color-mix` on the front end
- New helpers `parseHexColor` / `mixWithBlack` to compute the notch color server-side
- Noise tile buffer is memoized via `getGiftCardNoiseTile()` so Sharp only resizes it once per process


ghost/core/core/server/web/gift-preview/image.js
```diff
@@ -141,11 +199,19 @@ async function generateGiftPreviewImage({accentColor = '#15171A', siteTitle, tie
     const sharp = require('sharp');
 
     const svg = buildSvg({accentColor});
+    const noiseTile = await getGiftCardNoiseTile(sharp);
```
sagzy: nit: works, but feels weird to pass sharp as parameter

We should be okay to lazy-require sharp in `getGiftCardNoiseTile`, as Node caches require() results

so e.g.

```
async function getGiftCardNoiseTile() {
    if (giftCardNoiseTile !== undefined) {
        return giftCardNoiseTile;
    }

    const sharp = require('sharp');
    giftCardNoiseTile = await sharp(GIFT_CARD_NOISE_PATH)
        .resize(192, 192, {kernel: 'nearest'})
        .png()
        .toBuffer();

    return giftCardNoiseTile;
}
```

sagzy: Updated in 22dcee577f5183caefbeee37cad2491753cbdf79

